PR #101, opened 2019-01-09
Configure readthedocs to use python3.6 and install dependencies

vxgmichel:
The parsec module now needs to be importable for sphinx-build to run.

This will be useful to document the API using `sphinx.ext.autodoc`.

Tested on readthedocs.


vxgmichel:
[Fix readthedoc documentation génération](https://trello.com/c/8DtPgDmd/219-fix-readthedoc-documentation-g%C3%A9n%C3%A9ration)

PR #200, opened 2019-02-19
Set debug log level when running tests in the CI

vxgmichel:


touilleMan:
As usual, appveyor doesn't agree with this PR :'-(

Anyway, debug log level is currently far too verbose, I suggest we drop trames logging in the transport layer
https://github.com/Scille/parsec-cloud/blob/f114cf326a23281e346f41b279ed1acaee0aec35/parsec/api/transport.py#L63-L73

It seems most of the logs are due to those loggers (especially when hypothesis start sending crazy buffers), and they aren't really useful given they display msgpack-encoded data...

PR #205, opened 2019-02-20
Remove copy from filesystem api

vxgmichel:
As decided [here](https://github.com/Scille/parsec-cloud/issues/120#issuecomment-465519751), `copy` should not be part of the filesystem API.

### DIFF
--- a/tests/core/fs/test_local_folder_fs.py
+++ b/tests/core/fs/test_local_folder_fs.py
@@ -102,8 +102,7 @@ def test_file_create(local_folder_fs, alice):
     }
 
 
-@pytest.mark.parametrize("type", ["copy", "move"])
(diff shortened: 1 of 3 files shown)


tests/core/fs/test_local_folder_fs.py
```diff
@@ -102,8 +102,7 @@ def test_file_create(local_folder_fs, alice):
     }
 
 
-@pytest.mark.parametrize("type", ["copy", "move"])
-def test_copy_folders_create_new_accesses(local_folder_fs, type):
+def test_copy_folders_create_new_accesses(local_folder_fs):
```
touilleMan: Shouldn't we rename this `test_move_folder_create_new_accesses` ?

vxgmichel: Done!

PR #222, opened 2019-02-25
Mountpoint manager works on a per-workspace basis

### LINKED ISSUE #212: Treat each workspace as a separate mount point, controlled independently
After a discussion with @touilleMan, the following proposal is submitted to the team (feel free to edit this comment to add more information):

## Treat each workspace as a separate mount point, controlled independently

A workspace would no longer be treated as a directory at the root of the mount point, but as a mount point itself. Conceptually, this seems to be a better fit than the current design, which feels a bit clumsy.

### Pros

- Greatly simplify filesystem logic as in many places, workspaces are treated very differently than regular folders:
  * Manifests are not recursively copied when a workspace is renamed.
  * It is not possible to create a file or a folder at the root of the mountpoint.
  * And more:
    ```console
    $ git grep workspace_manifest parsec/core/fs
    parsec/core/fs/local_folder_fs.py:                if is_workspace_manifest(child_manifest):
    parsec/core/fs/local_folder_fs.py:        if is_workspace_manifest(manifest):
    parsec/core/fs/local_folder_fs.py:        elif is_workspace_manifest(manifest):
    parsec/core/fs/local_folder_fs.py:        if not is_workspace_manifest(src_manifest):
    parsec/core/fs/local_folder_fs.py:            if is_workspace_manifest(src_ro_manifest):
    parsec/core/fs/local_folder_fs.py:            if is_workspace_manifest(src_manifest):
    parsec/core/fs/local_folder_fs.py:            if is_workspace_manifest(src_manifest):
    ```
- Solve existing issues with the current implementation of the file system module:
  * Issue #206 only happens when workspaces are involved.
  * Moving a folder from a workspace to another should copy the corresponding blocks (see [this comment](https://github.com/Scille/parsec-cloud/issues/120#issuecomment-465519751)). This issue is currently not trivial to solve, because the move operation might require the downloading of many blocks not present on the system. With independent mount points, this move operation simply disappears: the OS takes care of it by recursively copying the files (i.e. reading and writing their content) from one mount point to the other before removing the source.
  

### Cons

- Require a lot of refactoring on the core side
  * Each workspace gets its own mountpoint manager and filesystem manager.
  * The API is also going to change since workspace-specific functions have to be moved.
- Require new features in the GUI:
  * Each workspace/mountpoint has to be controlled independently.
  * The current status of each mountpoint has to be displayed.
- Require a lot of changes in the test modules, obviously.

### Questions

- Is having many mount points mounted and displayed in the file manager a problem:
  * from the user experience point of view?
  * from the technical point of view (i.e. technical limitation)?

touilleMan:
closes #212


tests/conftest.py
```diff
@@ -43,7 +43,7 @@ def pytest_addoption(parser):
         help="Use PostgreSQL backend instead of default memory mock",
     )
     parser.addoption("--runslow", action="store_true", help="Don't skip slow tests")
-    parser.addoption("--runfuse", action="store_true", help="Don't skip FUSE/WinFSP tests")
+    parser.addoption("--runmountpoint", action="store_true", help="Don't skip FUSE/WinFSP tests")
```
vxgmichel: Maybe switch to `--skipmountpoint` so the mountpoint tests run by default, now that mountpoints are supported on both windows and linux?

parsec/core/mountpoint/manager.py
```diff
@@ -43,28 +49,71 @@ def get_mountpoint_runner():
...
+                pass
+
+    async def unmount_all(self):
+        for workspace in list(self._mountpoint_tasks.keys()):
+            await self.unmount_workspace(workspace)
 
 
 @asynccontextmanager
-async def mountpoint_manager(fs, event_bus, mountpoint, nursery, *, debug: bool = False, **config):
+async def mountpoint_manager_factory(
```
vxgmichel: Why the renaming? Isn't `mountpoint_manager` explicit enough? Or `mountpoint_manager_context` since it is an async context?

parsec/core/mountpoint/manager.py
```diff
@@ -43,28 +49,71 @@ def get_mountpoint_runner():
...
-        return
+    if not enabled:
+        runner = disabled_runner
+    else:
+        runner = get_mountpoint_runner()
+        if not runner:
+            raise RuntimeError("Mountpoint support not available.")
+    curried_runner = partial(runner, config=config, fs=fs, event_bus=event_bus)
+
+    async with trio.open_nursery() as nursery:
```
vxgmichel: So the mountpoint manager doesn't use an external nursery anymore (`monitor_nursery` from logged core)? What was the problem?

parsec/core/mountpoint/manager.py
```diff
@@ -43,28 +49,71 @@ def get_mountpoint_runner():
...
+    async def mount_workspace(self, workspace: str):
+        if workspace in self._mountpoint_tasks:
+            raise MountpointAlreadyMounted(f"Workspace `{workspace}` already mounted.")
+
+        try:
+            await self._fs.stat(f"/{workspace}")
+        except FileNotFoundError as exc:
+            raise MountpointConfigurationError(f"Workspace `{workspace}` doesn't exist") from exc
+
+        mountpoint = self._base_mountpoint / f"{self._fs.device.user_id}-{workspace}"
```
vxgmichel: Why not use `self._base_mountpoint / f"{self._fs.device.user_id}/{workspace}"` instead?

vxgmichel:
Ooops the GUI has to be updated too:
```
  File "/home/vinmic/miniconda/envs/parsec/lib/python3.7/site-packages/parsec/core/gui/main_window.py", line 191, in login_with_password
    self.start_core()
  File "/home/vinmic/miniconda/envs/parsec/lib/python3.7/site-packages/parsec/core/gui/main_window.py", line 143, in start_core
    self.central_widget.set_core_attributes(core=self.core, portal=self.portal)
  File "/home/vinmic/miniconda/envs/parsec/lib/python3.7/site-packages/parsec/core/gui/central_widget.py", line 72, in set_core_attributes
    super().set_core_attributes(core, portal)
  File "/home/vinmic/miniconda/envs/parsec/lib/python3.7/site-packages/parsec/core/gui/core_widget.py", line 13, in set_core_attributes
    self.core = core
  File "/home/vinmic/miniconda/envs/parsec/lib/python3.7/site-packages/parsec/core/gui/central_widget.py", line 108, in core
    self.label_mountpoint.setText(str(self._core.mountpoint))
AttributeError: 'LoggedCore' object has no attribute 'mountpoint'
```

parsec/core/logged_core.py
```diff
@@ -34,8 +34,7 @@ class LoggedCore:
     local_db = attr.ib()
     event_bus = attr.ib()
     encryption_manager = attr.ib()
-    mountpoint = attr.ib()
```
vxgmichel: [This broke the GUI](https://github.com/Scille/parsec-cloud/pull/222#issuecomment-467816616).

vxgmichel:
It also breaks the CLI:
```python-traceback
  File "/home/vinmic/miniconda/envs/parsec/lib/python3.7/site-packages/parsec/core/cli/run.py", line 32, in _run_mountpoint
    mountpoint_display = click.style(str(core.mountpoint.absolute()), fg="yellow")
AttributeError: 'LoggedCore' object has no attribute 'mountpoint
```

touilleMan:
@vxgmichel bien vu, j'ai fixé dans 3b941ed3edb56bc58b510e42969b54d832582b09

tests/conftest.py
touilleMan: mountpoint tests are slow (and beside not thaaat stable right now), so I'd better keep it this way ;-)

parsec/core/mountpoint/manager.py
touilleMan: `factory` is to indicate this function is not the mountpoint manager (i.e. call this function won't start fuse operation for instance) but will provide an the mountpoint manager (which expose an API to start/stop mountpoints)

parsec/core/mountpoint/manager.py
touilleMan: Separation of concerns !
Better avoid mixin nurseries (and use different nursery for different things)

parsec/core/mountpoint/manager.py
touilleMan: Good point !
My thinking was to have a clear naming in for the file explorer (nautilus detect the mountpoint and expose it with the name of the folder)

parsec/core/mountpoint/manager.py
vxgmichel: Agreed! This was bothering me while I was working on #145

parsec/core/mountpoint/manager.py
vxgmichel: But there is typically only one parsec client running at a time on a given machine right? So the `user_id` doesn't seem to provide much information :thinking:

parsec/core/mountpoint/manager.py
vxgmichel: I see your point, although I don't think `factory` carries much information as a name. In my opinion, the following line feels a bit weird:
```python
async with mountpoint_manager_factory(...) as mountpoint_manager:
```
because we actually mean:
```python
async with mountpoint_management_context(...) as mountpoint_manager:
```
Not a big deal though, it's mostly nitpicking :smile:

vxgmichel:
Extra question: how do we deal with workspace being created, shared or deleted while the mountpoint manager is running?

touilleMan:
Good question, I see two solutions:

1) have a monitor task listening on `fs.workspace.loaded` / `fs.workspace.unloaded` and then calling the mountpoint manager.
2) let it to the user by providing mount/unmount button for each workspace in the Qt GUI.

I prefer solution 2 (seems like less work, and the mountpoint control from the GUI is needed anyway).
There is still the case of the end of sharing where we must force the unmount of the workspace but it shouldn't be too much of a trouble.

touilleMan:
https://github.com/Scille/parsec-cloud/pull/222#pullrequestreview-209922497

Go for it

> ^Cfuse: fuse_remove_signal_handlers: unknown session

I'll have a look on this...

PR #256, opened 2019-03-01
Add right access customization to core.share

touilleMan:


parsec/core/fs/remote_loader.py
```diff
@@ -34,7 +36,7 @@ def __init__(self, backend_cmds, encryption_manager, local_db):
 
         self.local_db.set(access, block)
 
-    async def load_manifest(self, access: ManifestAccess) -> None:
+    async def load_manifest(self, access: Union[ManifestAccess, WorkspaceManifestAccess]) -> None:
```
vxgmichel: So `WorkspaceManifestAccess` is not a `ManifestAcccess`?

touilleMan: This PR makes a distinction between `WorkspaceManifestAccess` (which contains admin/read/write rights along with key and id) and `ManifestAcccess`

But PR #259 drop `WorkspaceManifestAccess` because it was a poor idea to do this

So nothing to see here, move along :smile:

vxgmichel:
This PR is outdated by #259.

PR #266, opened 2019-03-06
Improve winfspy mountpoint

touilleMan:


touilleMan:
merged, ci failure tracked by #269

PR #288, opened 2019-03-11
Correct misspelling revocated -> revoked

touilleMan:
cf. https://en.wiktionary.org/wiki/revoke#English vs https://en.wiktionary.org/wiki/revocate#English


vxgmichel:
I just fixed the formatting issue.

PR #289, opened 2019-03-13
Vlob api add autho timestamp

touilleMan:


Max-7:
Already contained in https://github.com/Scille/parsec-cloud/pull/290

Issue #269, opened 2019-03-07
Flaky test with winfsp on appveyor


touilleMan:
https://ci.appveyor.com/project/touilleMan/parsec-cloud/builds/22872696

```
E           hypothesis.errors.Flaky: Unreliable assumption: An example which satisfied assumptions on the first run now fails it.
c:\python36-x64\lib\site-packages\hypothesis\core.py:748: Flaky
---------------------------- Captured stdout call -----------------------------
Falsifying example: run_state_machine(factory=FileOperationsStateMachine, data=data(...))
state = FileOperationsStateMachine()
state.init()
state.teardown()
Traceback (most recent call last):
  File "c:\python36-x64\lib\site-packages\hypothesis\core.py", line 682, in run
    falsifying_example.__expected_traceback,
  File "c:\python36-x64\lib\site-packages\hypothesis\core.py", line 558, in execute
    result = self.test_runner(data, run)
  File "c:\python36-x64\lib\site-packages\hypothesis\executors.py", line 56, in default_new_style_executor
    return function(data)
  File "c:\python36-x64\lib\site-packages\hypothesis\core.py", line 556, in run
    return test(*args, **kwargs)
  File "c:\python36-x64\lib\site-packages\hypothesis\stateful.py", line 108, in run_state_machine
    while should_continue.more():
  File "c:\python36-x64\lib\site-packages\hypothesis\internal\conjecture\utils.py", line 400, in more
    should_continue = biased_coin(self.data, p_continue)
  File "c:\python36-x64\lib\site-packages\hypothesis\internal\conjecture\utils.py", line 227, in biased_coin
    i = data.draw_bits(bits)
  File "c:\python36-x64\lib\site-packages\hypothesis\internal\conjecture\data.py", line 338, in draw_bits
    return int_from_bytes(self.draw_bytes(n // 8))
  File "c:\python36-x64\lib\site-packages\hypothesis\internal\conjecture\data.py", line 396, in draw_bytes
    self.__check_capacity(n)
  File "c:\python36-x64\lib\site-packages\hypothesis\internal\conjecture\data.py", line 368, in __check_capacity
    raise StopTest(self.testcounter)
hypothesis.errors.StopTest: 138
You can reproduce this example by temporarily adding @reproduce_failure('4.5.5', b'AAE=') as a decorator on your test case
```

touilleMan:
Seems like something unreliably reliable...
[...]

PR #290, opened 2019-03-13
Remove generate token

touilleMan:
follows #289


touilleMan:
see #269 for appveyor error

PR #291, opened 2019-03-13
Added get_permissions API in fs

Max-7:


touilleMan:
see #269 for appveyor error

PR #321, opened 2019-03-18
Rework crypto

touilleMan:
- rename `certified_*` -> `*_certificate`
- Introduce `RemoteDevicesManager`, responsible for fetching&validating
  user/device from the backend
- Rework crypto module to also take care of user/device/revoked_device
  certificates
- Remove `EncryptionManager` and `trustchain` module and divide their features
  between the `crypto` module and `RemoteDevicesManager`.
- user api only provides certificate, this ensure the client must verify
  the certificate before making use of the data it contains
- message and vlob api requires timestamp of signed document, this allow the
  backend to ensure the timestamp roughly correspond to the current time so
  a client fetching the message can make sure it wasn't signed with a date
  set in the future (or in the past)
- Add tests for invite_claim and test_{user|device}_get_invitation_creator


parsec/core/local_storage.py
```diff
@@ -35,16 +27,6 @@ def __enter__(self):
...
-    # User interface
-
-    def get_user(self, access: Access) -> Dict:
-        raw_user_data = self.persistent_storage.get_user(access)
-        return remote_user_serializer.loads(raw_user_data)
-
-    def set_user(self, access: Access, user: Dict) -> None:
-        raw = remote_user_serializer.dumps(user)
-        self.persistent_storage.set_user(access, raw)
-
```
vxgmichel: PersistentStorage update

tests/fuzzing/test_fuzz_core.py
```diff
@@ -232,7 +231,7 @@ class SkipCommand(Exception):
 
 @pytest.mark.trio
 @pytest.mark.slow
-@pytest.mark.skipif(os.name == "nt", reason="TODO: Cause freeze in winfsp so far...")
+@pytest.mark.skip  # TODO: cause unreliable freezes in ci !
```
vxgmichel: To update after #328 is merged.

parsec/core/local_storage.py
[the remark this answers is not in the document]
touilleMan: done in e584f67d

Issue #330, opened 2019-03-20
Regression in FSOnlineConcurrentTreeAndSync


vxgmichel:
```python-traceback
    def set_clean_manifest(self, access: Access, raw: bytes):
        if self._check_presence(self.dirty_conn, access):
>           raise ValueError("Cannot set clean manifest: a dirty manifest is already present")
E           ValueError: Cannot set clean manifest: a dirty manifest is already present

parsec/core/persistent_storage.py:255: ValueError
```

```python
Falsifying example: run_state_machine(factory=FSOnlineConcurrentTreeAndSync, data=data(...))

state = FSOnlineConcurrentTreeAndSync()
async def steps():
    v1 = await state.init()
    v2 = await state.register_fs2(force_after_init=v1)
    v3 = await state.register_fs1(force_after_init=v1)
    v4 = await state.create_workspace(fs=v3, name='a')
    v5 = await state.create_file(fs=v3, name='a', parent=v4)
    await state.sync_all_the_files()
    await state.update_file(fs=v3, path=v5)
    await state.update_file(fs=v2, path=v5)
    await state.sync_all_the_files()
    await state.update_file(fs=v3, path=v5)
    await state.sync_all_the_files()
    await state.teardown()
state.trio_run(steps)

You can reproduce this example by temporarily adding @reproduce_failure('4.5.5', b'AXicY2RgZGQAQjYgwcDIAiJBXGYgycwIYzEAAALXACM=') as a decorator on your test case
```

PR #336, opened 2019-03-21
 Check manifest status (clean/dirty) before setting them in local storage

vxgmichel:
Fix issue #330.

### DIFF
--- a/parsec/core/fs/local_folder_fs.py
+++ b/parsec/core/fs/local_folder_fs.py
@@ -90,9 +90,13 @@ class LocalFolderFS:
         # Always keep the user manifest locally
         if access == self.root_access:
             return self.set_dirty_manifest(access, manifest)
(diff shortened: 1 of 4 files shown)


parsec/core/fs/local_folder_fs.py
```diff
@@ -90,9 +90,13 @@ def set_clean_manifest(self, access: Access, manifest: LocalManifest, force=Fals
         # Always keep the user manifest locally
         if access == self.root_access:
             return self.set_dirty_manifest(access, manifest)
+        if manifest.need_sync:
+            raise ValueError("The given manifest is a clean manifest.")
```
touilleMan: "is not a clean manifest" plutôt ?

vxgmichel: Effectivement!

PR #349, opened 2019-03-25
Divide FS between UserFS and WorkspaceFS

touilleMan:


vxgmichel:
Related: #350

parsec/core/fs/userfs.py
```diff
@@ -0,0 +1,140 @@
...
+        self._local_file_fs = LocalFileFS(device, local_storage, self._local_folder_fs, event_bus)
+        self._remote_loader = RemoteLoader(backend_cmds, remote_devices_manager, local_storage)
+        self._syncer = Syncer(
+            device,
+            backend_cmds,
+            remote_devices_manager,
+            self._local_folder_fs,
+            self._local_file_fs,
+            event_bus,
+        )
```
vxgmichel: I think we'll need to have all the following objects instanciated once for each workspace fs:
- local folder fs
- local file fs
- syncer

Not sure about the remote loader though.

parsec/core/fs/workspacefs.py
```diff
@@ -0,0 +1,138 @@
...
+        self.local_storage = local_storage
+        self.backend_cmds = backend_cmds
+        self.event_bus = event_bus
+        self._local_folder_fs = _local_folder_fs
+        self._local_file_fs = _local_file_fs
+        self._remote_loader = _remote_loader
+        self._syncer = _syncer
+
+    def _cook_path(self, relative_path=""):
+        return FsPath(f"/{self.workpace_name}/{relative_path}")
```
vxgmichel: We'll have to tackle this change of meaning for root at some point and get rid of the cooking

tests/core/fs/test_fs_online_concurrent_user.py
```diff
@@ -0,0 +1,143 @@
...
+    hypothesis_settings,
+    backend_addr,
+    backend_factory,
+    server_factory,
+    local_storage_factory,
+    fs_factory,
+    alice,
+    alice2,
+):
+    class FSOnlineConcurrentUser(TrioRuleBasedStateMachine):
```
vxgmichel: Good idea to test this logic separately :+1:

PR #372, opened 2019-04-04
Fix serde timestamp unpacking on non-utc machine

touilleMan:


touilleMan:
CI failure not related (see https://github.com/Scille/parsec-cloud/issues/377)

PR #390, opened 2019-04-12
Implement atomic transactions for entry operations

vxgmichel:
Continue the work started with #370.

### DIFF
--- a/parsec/core/fs/entry_transactions.py
+++ b/parsec/core/fs/entry_transactions.py
@@ -1,0 +1,379 @@
+# Parsec Cloud (https://parsec.cloud) Copyright (c) AGPLv3 2019 Scille SAS
+
+import os
+import errno
+from typing import Tuple
+from collections import namedtuple
+from async_generator import asynccontextmanager
+
+from parsec.event_bus import EventBus
+from parsec.core.types import (
+    FsPath,
+    Access,
+    BlockAccess,
+    WorkspaceEntry,
+    LocalDevice,
+    LocalManifest,
+    ManifestAccess,
+    LocalFileManifest,
+    LocalFolderManifest,
+    LocalWorkspaceManifest,
+    LocalUserManifest,
+    FileCursor,
+)
+from parsec.core.local_storage import LocalStorage, LocalStorageMissingEntry
+from parsec.core.fs.utils import (
+    is_file_manifest,
+    is_folder_manifest,
+    is_folderish_manifest,
+    is_workspace_manifest,
+    is_user_manifest,
+)
+from parsec.core.fs.remote_loader import RemoteLoader
+
+
+Entry = namedtuple("Entry", "access manifest")
+
+
+def from_errno(errno, message=None, filename=None, filename2=None):
+    if message is None:
+        message = os.strerror(errno)
+    return OSError(errno, message, filename, None, filename2)
+
+
+class EntryTransactions:
+    def __init__(
+        self,
+        workspace_entry: WorkspaceEntry,
+        local_storage: LocalStorage,
+        remote_loader: RemoteLoader,
+        event_bus: EventBus,
+    ):
+        self.workspace_entry = workspace_entry
+        self.local_storage = local_storage
+        self.remote_loader = remote_loader
+        self.event_bus = event_bus
+
+    # Rights logic
+
+    def _check_write_rights(self, path: FsPath):
+        if not self.workspace_entry.write_right:
+            raise from_errno(errno.EACCES, str(path))
+
+    # Look-up logic
+
+    async def _get_manifest(self, access: Access) -> LocalManifest:
+        try:
+            return self.local_storage.get_manifest(access)
+        except LocalStorageMissingEntry as exc:
+            return await self.remote_loader.load_manifest(exc.access)
+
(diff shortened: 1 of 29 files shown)


tests/core/fs/test_fs_sync.py
```diff
@@ -582,7 +581,8 @@
 
     # First create data locally
     with freeze_time("2000-01-02"):
-        obj_id = await alice_fs.file_create("/w/foo.txt")
+        fd = await alice_fs.file_create("/w/foo.txt")
```
touilleMan: Useful to have `file_create` returning the entry id

Given returning file descriptor is also useful, I would prefer to be able to do:
```python
file_id = fs.file_create("/foo.txt")
file_id, fd = fs.file_create("/bar.txtb, open=True)
```

tests/core/fs/test_fs_sync.py
```diff
@@ -17,7 +18,8 @@
         assert stat1 == stat2
 
         cooked_children = {}
-        if stat1["is_folder"]:
+        # XXX: type root should not exist
```
touilleMan: Same than below

parsec/core/fs/workspacefs.py
```diff
@@ -5,119 +5,147 @@
...
+from parsec.core.fs.entry_transactions import EntryTransactions
 from parsec.core.fs.local_folder_fs import FSManifestLocalMiss, FSMultiManifestLocalMiss
 
 
 class WorkspaceFS:
     def __init__(
         self,
-        workpace_name,
+        workspace_name,
+        workspace_entry,
```
touilleMan: `workspace_entry` also contains name

parsec/core/fs/fs.py
```diff
@@ -194,12 +214,12 @@ def _get_workspace(self, path):
         cooked_path = FsPath(path)
         if not cooked_path.is_workspace():
             # Will fail with correct exception
-            workpace_name = path
+            workspace_name = path
```
touilleMan: Damn typo, it has reproduce like cockroaches !!!!

parsec/core/fs/fs.py
```diff
@@ -66,60 +66,76 @@ def _get_workspace(self, path):
         return workspace, subpath
 
     async def stat(self, path: str) -> dict:
+        # XXX: This method should be splitted in several methods:
```
touilleMan: better use `TODO` everywhere to simplify grep

parsec/core/fs/entry_transactions.py
```diff
@@ -0,0 +1,382 @@
...
+            async with self.local_storage.lock_manifest(access) as manifest:
+                yield Entry(access, manifest)
+
+        # The entry has been deleted while we were waiting for the lock
+        except LocalStorageMissingEntry:
+            raise from_errno(errno.ENOENT, filename=str(path))
+
+    @asynccontextmanager
+    async def _lock_parent_entry(self, path):
+        # Source is root
```
touilleMan: Should add a comment about why the locking strategy is parent -> children only

tests/fuzzing/test_fuzz_core.py
```diff
@@ -140,7 +140,7 @@ class SkipCommand(Exception):
     elif x < 20:
         path = fs_state.get_new_path()
         try:
-            await core.fs.file_create(path)
+            await core.fs.touch(path)
```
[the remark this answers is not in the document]
vxgmichel: Well, those general tests should rely on the [high-level pathlib-like interface](https://github.com/Scille/parsec-cloud/wiki/Workspace-file-system-architecture#workspace-path-interface). The issue is that this interface doesn't exist yet so a temporary solution was to add `touch` and sed the tests. 

We could also use `file_create(path, open=False)` if we add the extra switch, but again high-level tests should not directly use the low level transactions, so we'll have to remember to change it later.

tests/core/fs/test_fs_sync.py
vxgmichel: Is there a use case I'm not aware of? Because it wasn't used anywhere in the code base. By the way `folder_create` no longer returns the entry id either.

As far as I understand the concept of entry id only leaks out of the FS because we want to send notifications to other devices without using plain path information right?

Also I can add the `open` flag to `file_create`, but I don't think it's going to be used in the code base, except maybe for some tests.

parsec/core/fs/workspacefs.py
vxgmichel: OK I'll change that

parsec/core/fs/fs.py
vxgmichel: Yea I noticed this fix in your PR too ^^

parsec/core/fs/fs.py
vxgmichel: Alright I'll fix it

parsec/core/fs/entry_transactions.py
vxgmichel: No problem

tests/core/fs/test_fs_sync.py
vxgmichel: Done! (see e55ed7f and 1c3d68c)

vxgmichel:
Next step is to bypass the FS object. More specifically, the mountpoints should have a reference to their workspace file system and use the transaction API directly.

PR #412, opened 2019-04-21
Fix fs.share

touilleMan:


parsec/core/fs/fs.py
```diff
@@ -270,15 +270,14 @@ def _iter_workspaces(self):
...
+        if not subpath:
             # Will fail with correct exception
             workspace_name = path
-        else:
-            workspace_name = cooked_path.workspace
+        workspace_entry = self._get_workspace_entry_from_name(workspace_name)
 
         await self._user_fs.workspace_share(
-            workspace_name,
+            workspace_entry.access.id,
```
vxgmichel: How could the tests miss that? :thinking:

touilleMan: The tests are already ported to `UserFS` api... :thinking:

PR #418, opened 2019-04-24
Updated FS handling in the GUI

Max-7:
Merge #416 first.


parsec/core/gui/workspaces_widget.py
```diff
@@ -89,24 +99,23 @@ def delete_workspace(self, workspace_button):
...
         )
         if not new_name:
             return
         try:
-            self.jobs_ctx.run(
-                self.core.fs.workspace_rename, current_file_path, os.path.join("/", new_name)
-            )
+            workspace_id = workspace_button.workspace_fs.workspace_entry.access.id
+            self.jobs_ctx.run(self.core.fs._user_fs.workspace_rename, workspace_id, new_name)
+            workspace_button.workspace_fs = self.core.fs._user_fs.get_workspace(workspace_id)
```
vxgmichel: Getting a new workspace object shouldn't be necessary in the long run, but maybe this solves the mutation issue you were talking about? In this case, a comment `TODO: remove later` would be nice.

parsec/core/gui/workspaces_widget.py
```diff
@@ -119,56 +128,13 @@ def create_workspace_clicked(self):
         if not workspace_name:
             return
         try:
-            self.jobs_ctx.run(self.core.fs.workspace_create, os.path.join("/", workspace_name))
+            self.jobs_ctx.run(self.core.fs._user_fs.workspace_create, workspace_name)
```
vxgmichel: This `core.fs._user_fs.do_stuff` seems painful. Maybe we should start with adding a `user_fs` shortcut to `core` to prepare the removal of the `fs` interface.

parsec/core/gui/workspaces_widget.py
Max-7: Yeah I realized that while updating workspace sharing and it will be removed.

PR #454, opened 2019-05-15
Updated black, added flake8 in pre-commit config

c4ffein:


vxgmichel:
Ooops, I approved too fast: black is also pinned in `setup.py`

c4ffein:
missing file

PR #455, opened 2019-05-15
Flake8 in pre commit

c4ffein:
missing setup.py : updated black, added flake8

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -305,11 +305,10 @@ test_requirements = [
     "tox",
     "wheel",
     "Sphinx",
-    "flake8",
+    "flake8==3.7.7",
     "hypothesis==4.5.5",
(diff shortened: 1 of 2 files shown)


setup.py
```diff
@@ -308,7 +308,8 @@ def _extract_libs_cffi_backend():
     "flake8",
     "hypothesis==4.5.5",
     "hypothesis-trio==0.3.0",
-    "black==18.9b0",  # Pin black to avoid flaky style check
+    "black==19.3b0",  # Pin black to avoid flaky style check
+    "flake8==3.7.7",
```
vxgmichel: `flake8` is already part of the list (L308)

c4ffein: Fixed

PR #464, opened 2019-05-24
Tests new userfs

frossigneux:


parsec/core/fs/workspacefs/workspacefs.py
```diff
@@ -216,6 +225,8 @@ def workspace_name(self) -> str:
         source = FsPath(source)
         destination = FsPath(destination)
         real_destination = destination
+        if _destinsrc(source, destination):
+            raise OSError(errno.EINVAL)
```
vxgmichel: This change would deserve its own commit right? Also a message argument like `"Cannot move a directory '%s' into itself '%s'"` would be nice.

parsec/core/sync_monitor.py
```diff
@@ -61,7 +61,7 @@ def running(self):
         with trio.CancelScope() as self._monitoring_cancel_scope:
             self.event_bus.send("sync_monitor.reconnection_sync.started")
             try:
-                await self.fs.full_sync()
+                await self.user_fs._syncer.full_sync()
```
vxgmichel: The `_syncer` object is legacy and shoudn't be used directly. What about `self.user_fs.sync()`?

parsec/core/sync_monitor.py
```diff
@@ -105,19 +105,19 @@ def _on_entry_updated(event, workspace_id=None, id=None):
 
         for id, (first_updated, last_updated) in updated_entries.items():
             if now - first_updated > MAX_WAIT:
-                if id == self.fs.user_fs.user_manifest_access.id:
-                    await self.fs.sync("/")
+                if id == self.user_fs.user_fs.user_manifest_access.id:
+                    await self.user_fs.sync("/")
```
vxgmichel: `user_fs.sync()` doesn't take any argument as far as I can tell.

parsec/core/sync_monitor.py
```diff
@@ -105,19 +105,19 @@ def _on_entry_updated(event, workspace_id=None, id=None):
...
                 else:
-                    await self.fs.sync_by_id(id)
+                    await self.user_fs._syncer.sync_by_id(id)
```
vxgmichel: The `_syncer` object is legacy and shoudn't be used directly. `sync_by_id` is available as a method of `WorkspaceFS` but you would also need the workspace id to get the corresponding instance.

parsec/core/sync_monitor.py
```diff
@@ -105,19 +105,19 @@ def _on_entry_updated(event, workspace_id=None, id=None):
...
                 break
 
         else:
             for id, (_, last_updated) in updated_entries.items():
                 if now - last_updated > MIN_WAIT:
-                    if id == self.fs.user_fs.user_manifest_access.id:
-                        await self.fs.sync("/")
+                    if id == self.user_fs.user_fs.user_manifest_access.id:
+                        await self.user_fs.sync("/")
```
vxgmichel: Same comment.

parsec/core/sync_monitor.py
```diff
@@ -105,19 +105,19 @@ def _on_entry_updated(event, workspace_id=None, id=None):
...
                     else:
-                        await self.fs.sync_by_id(id)
+                        await self.user_fs._syncer.sync_by_id(id)
```
vxgmichel: Same comment.

PR #477, opened 2019-06-05
Fix crash due to deprecated access in workspace_widget

touilleMan:


touilleMan:
Appveyor failure not related to this PR

PR #484, opened 2019-06-11
Fix user removed from all realms when unshared from a single realm wi…

touilleMan:
…th postgres


touilleMan:
Appveyor failure not related to this PR (planned to be fixed by #482 )

PR #497, opened 2019-06-17
Realm role with certificates

touilleMan:


parsec/core/fs/workspacefs/workspacefs.py
```diff
@@ -357,6 +357,9 @@ def encryption_revision(self) -> int:
         if remote_manifest is None:
             return
 
+        # Create in backend the realm corresponding to this workspace
+        await self.remote_loader.create_realm(entry_id)
+
```
vxgmichel: So this operation is indempotent? Is this blocks still useful then?
```python
        # Get workspace manifest
        try:
            workspace_manifest = self.local_storage.get_manifest(self.workspace_id)
        # Not available locally so nothing to synchronize
        except LocalStorageMissingError:
            return
        # Half hacky solution: we sync the workspace manifest but don't
        # update the user manifest. This should be fine because synchronization
        # is idempotent, but it's quite ugly...
        if workspace_manifest.is_placeholder:
            await self.minimal_sync(self.workspace_id)
```

Issue #503, opened 2019-06-19
The monitor sync should not trigger recursive synchronization


vxgmichel:
Only the manifest that has changed (remotely or locally) should get synchronized when the monitor sync decides  that the time is right (depending on the MIN_WAIT/MAX_WAIT policy).

Issue #502, opened 2019-06-19
Block size needs to be strictly lower than the transport buffer size


vxgmichel:
There is a 1MB maximum buffer size defined in `parsec.api.transport`, 

https://github.com/Scille/parsec-cloud/blob/b07fe93ff19863b9c49abba542f202710584ff37/parsec/api/transport.py#L38

and 1MB maximum block size defined in `parsec.core.fs.workspacefs.sync_transactions`:

https://github.com/Scille/parsec-cloud/blob/b07fe93ff19863b9c49abba542f202710584ff37/parsec/core/fs/workspacefs/sync_transactions.py#L27

The limits appear to be in conflict when uploading a 1MB block:
```python-tb
  File "/home/vinmic/parsec-cloud/parsec/core/sync_monitor.py", line 74, in _monitoring
    await self._monitoring_loop()
  File "/home/vinmic/parsec-cloud/parsec/core/sync_monitor.py", line 106, in _monitoring_loop
    await self._monitoring_tick(updated_entries)
  File "/home/vinmic/parsec-cloud/parsec/core/sync_monitor.py", line 147, in _monitoring_tick
    await workspace.sync_by_id(id)
  File "/home/vinmic/parsec-cloud/parsec/core/fs/workspacefs/workspacefs.py", line 459, in sync_by_id
    await self.sync_by_id(entry_id, remote_changed=remote_changed, recursive=True)
  File "/home/vinmic/parsec-cloud/parsec/core/fs/workspacefs/workspacefs.py", line 441, in sync_by_id
    manifest = await self._sync_by_id(entry_id, remote_changed=remote_changed)
  File "/home/vinmic/parsec-cloud/parsec/core/fs/workspacefs/workspacefs.py", line 417, in _sync_by_id
    await self._upload_blocks(new_remote_manifest)
  File "/home/vinmic/parsec-cloud/parsec/core/fs/workspacefs/workspacefs.py", line 346, in _upload_blocks
    await self.remote_loader.upload_block(access, data)
  File "/home/vinmic/parsec-cloud/parsec/core/fs/remote_loader.py", line 137, in upload_block
    raise FSError(f"Cannot upload block: {exc}") from exc
parsec.core.fs.exceptions.FSError: Cannot upload block: {'reason': 'Invalid message format', 'status': 'invalid_msg_format'}
```
A quick workaround is to lower the block size or increase the buffer size.

@touilleMan is there more to it?

vxgmichel:
[...]

Issue #501, opened 2019-06-19
LocalStorageMissingError when receiving notifications from a non-loaded workspace


vxgmichel:
This happens when receiving a `remote_changed` notification from a workspace that has never been loaded. This mean the workspace manifest does not exist in the local storage and an exception is raised. Simply ignoring this exception seems to be an acceptable fix.

```python-tb
Traceback (most recent call last):
 File "/snap/parsec/97/venv/lib/python3.6/site-packages/parsec/core/gui/trio_thread.py", line 65, in _run_fn
 self.ret = await self._fn(*self._args, **self._kwargs)
 File "/snap/parsec/97/venv/lib/python3.6/site-packages/parsec/core/gui/main_window.py", line 47, in _do_run_core
 qt_on_ready.emit(core, core_jobs_ctx)
 File "/snap/parsec/97/venv/lib/python3.6/site-packages/async_generator/_util.py", line 53, in __aexit__
 await self._agen.athrow(type, value, traceback)
 File "/snap/parsec/97/venv/lib/python3.6/site-packages/parsec/core/logged_core.py", line 100, in logged_core_factory
 monitor_nursery.cancel_scope.cancel()
 File "/snap/parsec/97/venv/lib/python3.6/site-packages/trio/_core/_run.py", line 506, in __aexit__
 raise combined_error_from_nursery
 File "/snap/parsec/97/venv/lib/python3.6/site-packages/parsec/core/sync_monitor.py", line 153, in monitor_sync
 await sync_monitor.run(task_status=task_status)
 File "/snap/parsec/97/venv/lib/python3.6/site-packages/parsec/core/sync_monitor.py", line 54, in run
 await self._monitoring()
 File "/snap/parsec/97/venv/lib/python3.6/site-packages/parsec/core/sync_monitor.py", line 74, in _monitoring
 await self._monitoring_loop()
 File "/snap/parsec/97/venv/lib/python3.6/site-packages/parsec/core/sync_monitor.py", line 106, in _monitoring_loop
 await self._monitoring_tick(updated_entries)
 File "/snap/parsec/97/venv/lib/python3.6/site-packages/parsec/core/sync_monitor.py", line 147, in _monitoring_tick
 await workspace.sync_by_id(id)
 File "/snap/parsec/97/venv/lib/python3.6/site-packages/parsec/core/fs/workspacefs/workspacefs.py", line 432, in sync_by_id
 workspace_manifest = self.local_storage.get_manifest(self.workspace_id)
[...]

PR #505, opened 2019-06-19
Fix several synchronization issues

vxgmichel:
Several hot fixes for issue #501, #502 and #503.


parsec/core/fs/workspacefs/sync_transactions.py
```diff
@@ -24,7 +24,7 @@
 
 __all__ = "SyncTransactions"
 
-DEFAULT_BLOCK_SIZE = 1 * 1024 * 1024  # 1 MB
+DEFAULT_BLOCK_SIZE = 512 * 1024  # 1 MB
```
touilleMan: du coup c'est plus 1mo

PR #541, opened 2019-06-27
Improve RAID5 blockstore to support read/create under single blockstore failure

touilleMan:


touilleMan:
@c4ffein I've improved the tests to cover the cases you're suggesting (especially considering checksum blockstore is handled diffrently than other ones)

touilleMan:
appveyor fail not related to this PR

PR #547, opened 2019-06-28
Vlob versioning core managed at deeper level

c4ffein:


parsec/core/fs/remote_loader.py
```diff
@@ -393,3 +393,38 @@ def __init__(
...
+        return await super().load_manifest(entry_id, version=version, timestamp=self.timestamp)
+
+    async def upload_manifest(self, *e, **ke):
+        raise FSError(f"Cannot upload manifest through a timestamped remote loader")
+
+    async def _vlob_create(self, *e, **ke):
+        raise FSError(f"Cannot create vlob through a timestamped remote loader")
+
+    async def _vlob_update(self, *e, **ke):
+        raise FSError(f"Cannot update vlob through a timestamped remote loader")
```
vxgmichel: FYI, `_vlob_create` and `_vlob_update` are only internal methods meant to be called through `upload_manifest`.

parsec/core/fs/workspacefs/workspacefs_timestamped.py
```diff
@@ -41,16 +45,25 @@ def __init__(self, workspacefs: WorkspaceFS, timestamp: pendulum.Pendulum):
...
-        self.remote_loader = workspacefs.remote_loader
-        self.file_transactions = workspacefs.file_transactions
-        self.entry_transactions = workspacefs.entry_transactions
+        self.remote_loader = workspacefs.remote_loader.to_timestamped(timestamp)
+        self.file_transactions = FileTransactions(
+            self.workspace_id, self.local_storage, self.remote_loader, self.event_bus
+        )
+        self.entry_transactions = EntryTransactions(
+            self.workspace_id,
+            self.get_workspace_entry,
```
vxgmichel: You could also mock `get_workspace_entry().role` to `WorkspaceRole.READER` to make sure no write operations is performed (instead of using `_throw_permission_error`)

parsec/core/local_storage.py
```diff
@@ -261,3 +263,106 @@ def create_cursor(self, entry_id: EntryID) -> FileDescriptor:
...
+
+    @asynccontextmanager
+    async def lock_manifest(self, entry_id: EntryID):
+        async with self.lock_entry_id(entry_id):
+            yield self.get_manifest(entry_id)
+
+    def _check_lock_status(self, entry_id: EntryID) -> None:
+        task = self.locking_tasks.get(entry_id)
+        # TODO: remove `task is None` to ensure that the lock is taken
+        assert task is None or task == hazmat.current_task()
```
vxgmichel: Why re-implement the locking methods?

parsec/core/local_storage.py
```diff
@@ -261,3 +263,106 @@ def create_cursor(self, entry_id: EntryID) -> FileDescriptor:
...
+        assert isinstance(entry_id, EntryID)
+        self._check_lock_status(entry_id)
+        self.base_manifest_cache.pop(entry_id, None)
+        self.local_manifest_cache.pop(entry_id, None)
+
+    # Block interface
+
+    def set_dirty_block(self, block_id: BlockID, block: bytes) -> None:
+        # should not return anything? no sense to write dirty blocks from ltimestamped? raise error?
+        return
```
vxgmichel: Yes, this method should never be called.

parsec/core/local_storage.py
```diff
@@ -261,3 +263,106 @@ def create_cursor(self, entry_id: EntryID) -> FileDescriptor:
...
+
+    def clear_block(self, block_id: BlockID) -> None:
+        # better not uncache a maybe used block by the actual workspace than to have to reload it?
+        return
```
vxgmichel: I would also raise an error here.

parsec/core/local_storage.py
```diff
@@ -261,3 +263,106 @@ def create_cursor(self, entry_id: EntryID) -> FileDescriptor:
...
+        except KeyError:
+            raise LocalStorageMissingError(entry_id)
+
+    def get_manifest(self, entry_id: EntryID) -> LocalManifest:
+        """Raises: LocalStorageMissingError"""
+        assert isinstance(entry_id, EntryID)
+        try:
+            return self.local_manifest_cache[entry_id]
+        except KeyError:
+            raise LocalStorageMissingError
```
vxgmichel: `get_manifest` may also return a local manifest from the remote manifest in `base_manifest_cache` if it exists

parsec/core/fs/remote_loader.py
c4ffein: Isn't it safer to mock them anyway, just in case? They could be used by non redefined unherited methods.

PR #550, opened 2019-06-30
Fix SecretKey.__repr__ to avoid leaking key

touilleMan:


touilleMan:
appveyor failure not related

Issue #553, opened 2019-07-01
TODOs in the file system module


vxgmichel:
This meta-issue lists all TODOs related to the fs modules

## Worskpace FS

- [x] Move reverse lookup logic (`get_entry_path`) to entry_transactions - [TODO](https://github.com/Scille/parsec-cloud/blob/ae350a871bf4f60c50d1b36bbf960cc66ea36111/parsec/core/fs/workspacefs/entry_transactions.py#L193) - see #554 
- [x] Investigate the looping policy of the `_sync_by_id` method - see #554
- [x] Investigate the `sync` method policy on synchronizing the path itself - [TODO](https://github.com/Scille/parsec-cloud/blob/ae350a871bf4f60c50d1b36bbf960cc66ea36111/parsec/core/fs/workspacefs/workspacefs.py#L481) - see PR #631 

## Synchronization and deletion

- [x] Prevent non synchronized manifest from being cleared when they are deleted - [TODO](https://github.com/Scille/parsec-cloud/blob/ae350a871bf4f60c50d1b36bbf960cc66ea36111/parsec/core/fs/workspacefs/entry_transactions.py#L325) - see #571
- [x] Implement a mechanism to synchronize removed entries (either by sync or deletion) - [comment](https://github.com/Scille/parsec-cloud/blob/ae350a871bf4f60c50d1b36bbf960cc66ea36111/parsec/core/fs/workspacefs/sync_transactions.py#L102) - PR #608 

## Base/current manifest handling

- [x] Use a single API call to set the base manifest and the current manifest - [TODO](https://github.com/Scille/parsec-cloud/blob/ae350a871bf4f60c50d1b36bbf960cc66ea36111/parsec/core/fs/workspacefs/sync_transactions.py#L261)
- [x] Ensure the invariant: base_manifest.version == local_manifest.base_version - [TODO](https://github.com/Scille/parsec-cloud/blob/ae350a871bf4f60c50d1b36bbf960cc66ea36111/parsec/core/fs/workspacefs/sync_transactions.py#L263)

Those two points are resolved with #587.

## Local storage

- [x] ~Investigate an asynchronous interface for the local storage (especially for sqlite)~ - deferred to #635
- [x] Move local storage module inside the fs module (similar to the remote loader) - PR #602
- [x] Investigate a per-workspace local storage solution - PR #602
[...]

PR #554, opened 2019-07-02
Cleaning up the fs module

vxgmichel:
Related: #553


parsec/core/fs/workspacefs/workspacefs.py
```diff
@@ -327,17 +327,21 @@ def to_timestamped(self, timestamp: Pendulum):
...
-        elif await self.is_file(source):
-            await self.copyfile(source, real_destination)
-            await self.unlink(source)
-            return
-        raise NotImplementedError
+
+        # Copy file
+        await self.copyfile(source, real_destination)
+        await self.unlink(source)
+        return
```
frossigneux: `return` final un peu superflu...

PR #555, opened 2019-07-02
Use realm_get_role_certificates in core

touilleMan:


parsec/backend/drivers/memory/realm.py
```diff
@@ -118,16 +118,14 @@ def _get_realm(self, organization_id, realm_id):
         author: DeviceID,
         realm_id: UUID,
         since: pendulum.Pendulum,
-    ) -> Tuple[RealmGrantedRole]:
-        """
-        Raises:
-            RealmNotFoundError
-            RealmAccessError
-        """
+    ) -> List[bytes]:
```
vxgmichel: No dedicated type for certificates?

touilleMan: Nobody ain't no time for this !

touilleMan: trust is using a type here is a bit cumbersome given the code dealing with certificates as bytes is faily small (deserialization then passing to db functions) and only in two places (`backend/realm.py` and `backend/<driver>/realm.py`)

Issue #557, opened 2019-07-02
Investigate trio loop sanity


vxgmichel:
We need a tool to measure the amount of time each task run before before giving the control back to the trio loop. We want to make sure to get some warnings when this time never exceeds a certain limit (say 100 ms like [asyncio debug mode](https://docs.python.org/3/library/asyncio-dev.html#debug-mode)) as this would be the sign of incorrect IO management (or heavy CPU task).

Trio reference: https://trio.readthedocs.io/en/latest/reference-hazmat.html?highlight=instrument#debugging-and-instrumentation

vxgmichel:
My findings:
```
2019-07-03 11:05:21 [warning  ] The previous step for task `parsec.core.sync_monitor.monitor_sync` took 1.195 s
Resuming afer:
  File "/home/vinmic/parsec-cloud/parsec/core/sync_monitor.py", line 156, in monitor_sync
    await sync_monitor.run(task_status=task_status)
```
And
```
2019-07-03 11:04:43 [warning  ] The last step for task `__main__.amain.<locals>.claim_task` took 0.092 s [parsec.core.monitoring]
```

vxgmichel:
It turns out the `merge_buffers` function is very slow, even for low number of buffers. Consider the following test:
```python
def test_merge_buffers_performance():
    blocks = []
    dirty_size = 4 * 1024
    clean_size = 512 * 1024
    total_size = 2 * 1024 * 1024
    for offset in range(0, total_size, dirty_size):
        block = Buffer(offset, offset + dirty_size, None)
        blocks.append(block)

    merged = merge_buffers_with_limits_and_alignment(blocks, 0, total_size, clean_size)

    assert merged.start == 0
    assert merged.end == merged.size == total_size
    assert len(merged.spaces) == 4
    assert sum(len(space.buffers) for space in merged.spaces) == 512
    buffers = [buff.buffer for space in merged.spaces for buff in space.buffers]
    assert buffers == blocks
```

Here the merging of 512 buffers takes half a second (note that size does not matter here as we are not dealing with the data itself). Running this test with `pytest-profiling` shows that:
- 99 % of the time is spent sorting buffers using `sorted`
- 96 % of the time is spent comparing the items to sort (`__lt__` method)
- 84 % of the time is spent converting the buffer objects to tuple in order to make the comparison (using `attrs_to_tuple`)

vxgmichel:
The `merge_buffer` performance issue is fixed in commit 3e2ec1a.

PR #559, opened 2019-07-03
Add a trio instrument for monitoring blocking tasks

vxgmichel:
See issue #557 
Also, fix performance issue in `merge_buffers`.


parsec/core/fs/buffer_ordering.py
```diff
@@ -218,7 +218,11 @@ def _merge_in_contiguous_space(overlaid_contiguous_spaces, buff):
 
         unsorted_spaces += trimmed
 
-    return ContiguousSpace(start, end, sorted(unsorted_spaces))
+    # Comparing attrs instances is costy and causes this sorting process to be
+    # really slow (0.5 seconds for 500 items). Using a lambda key speeds up the
+    # process by a factor of 20.
+    sorted_spaces = sorted(unsorted_spaces, key=lambda x: (x.start, x.end))
```
touilleMan: ouuuuuuiiiiiiiiii \o/

parsec/core/logged_core.py
```diff
@@ -41,6 +42,9 @@ class LoggedCore:
 async def logged_core_factory(
     config: CoreConfig, device: LocalDevice, event_bus: Optional[EventBus] = None
 ):
+    # Add task monitoring instrument
+    trio.hazmat.add_instrument(TaskMonitoringInstrument())
```
touilleMan: Should be moved elsewhere given the instrument is not really related to this function (and we are going to push multiple instruments if we call this function multiple time, like when the user does login/logout in the gui)

We should have instead:
- `@fixture(autouse=True)` defined in `tests/conftest.py` that returns the instrument (pytest-trio automatically collect instruments returned by fixtures and pass them to the `trio.run()`)
- for the cli, create a new `parsec.utils.trio_run` function that should be use instead of the original `trio.run` and `trio_asyncio.run` (so our `trio_run` should have a `with_asyncio=False` param to determine if we need to run a trio_asyncio loop on top of the trio loop)

parsec/core/monitoring.py
```diff
@@ -0,0 +1,45 @@
+# Parsec Cloud (https://parsec.cloud) Copyright (c) AGPLv3 2019 Scille SAS
```
touilleMan: Not specific to core, so should be in `parsec/monitoring.py`

Issue #513, opened 2019-06-20
Add a parsec:// protocol


Max-7:
Instead of having a huge URL (currently using ws:// protocol) to bootstrap an organization or add a user/device, we could implement or own parsec:// protocol and register this protocol to the operating system so that it opens Parsec automatically if such a link is clicked.
URL should give some information about the associated operation, so that the GUI can display the right window and autofill the URL form field.

Max-7:
https://docs.microsoft.com/fr-fr/windows/uwp/launch-resume/handle-uri-activation

vxgmichel:
On ubuntu (and maybe more distros) this can be done using the `x-scheme-handler`:
1. add support for an URI argument to the parsec command (e.g. `parsec parsec://foo`)
2. change the `Exec` entry of the `.desktop` file to `parsec %u` to pass the argument
2. add `MimeType=x-scheme-handler/parsec;` to the `.desktop` file
3. run `xdg-mime default parsec.desktop x-scheme-handler/parsec` to set up the scheme handler
4. test with `xdg-open parsec://foo`

[More info](https://askubuntu.com/a/1023143/400096).

touilleMan:
last step here https://github.com/Scille/parsec-cloud/pull/684

Max-7:
Done.

PR #567, opened 2019-07-05
Replace ws:// url scheme by parsec://

touilleMan:


touilleMan:
related to #513

touilleMan:
travis ci failure not related

parsec/types.py
```diff
@@ -56,8 +56,8 @@ def __init__(self, raw: str):
 
         self._split = urlsplit(raw)
 
-        if self._split.scheme not in ("ws", "wss"):
-            raise ValueError("Backend addr must start with ws:// or wss://")
+        if self._split.scheme not in ("parsec", "parsecs"):
+            raise ValueError("Backend addr must start with parsec:// or parsecs://")
```
vxgmichel: Do we really need both `parsec` and `parsecs`?

touilleMan: `parsecs` for parsec-over-ssl mode, otherwise we could set the ssl with a url parameter

touilleMan: what do you think of 778006c ?

vxgmichel: much better! what's the point of supporting both ws and wss though?

touilleMan: ssl certificates are complex to set up (must ask a certificate from an authority or create our own and transfert it on each client machine) and can be handled by multiple components (for instance on heroku the ssl layer is directly handled by the infrastructure and our application receive clear data)
Hence it's simpler to consider ssl as an optional layer that can be easily setup according to the needs.

vxgmichel: If ssl is the default, shouldn't the option be `no_ssl`?

touilleMan: done.

touilleMan:
Travis CI fail not related

PR #579, opened 2019-07-18
Remove api realm_get_role & improve idempotence on realm_update_roles

touilleMan:
Right now, `realm_update_roles` accept role certificates specifying the same role as the current one (e.g. Bob is MANAGER, Alice can do `realm_update_roles('bob', RealRole.MANAGER)`, this won't cause any error and the role certificates list will contain one more certificate)

This PR fix this behavior by not storing the new role certificate if it doesn't provide a new role. However the request is still considered a success.

I think it would be cleaner to return an error in such case, but this could break more things (having a new type of error that is triggered in subtle cases...) and doesn't seems really useful right now

@vxgmichel @max what do you think ?


vxgmichel:
> This PR fix this behavior by not storing the new role certificate if it doesn't provide a new role. However the request is still considered a success.

Yes, idempotency sounds like a good choice for this purpose.

> I think it would be cleaner to return an error in such case

You think so? Is there a use case for that?

touilleMan:
> You think so? Is there a use case for that?

It feels strange to have alice sending a role certificate for bob without trouble, then could find any trace of this when listing the roles certificates later on.
In theory the backend should return an error if the role was already granted, then the GUI should discard this error considering it usecase ("need to have user X with role Y", not "need to have user X with role Y given by device Z").

PR #589, opened 2019-07-23
Fix manifest check lock status

frossigneux:


parsec/core/gui/workspaces_widget.py
```diff
@@ -382,6 +412,7 @@ def reencrypt_workspace(self, workspace_id, user_revoked, role_revoked):
             ThreadSafeQtSignal(self, "workspace_reencryption_success"),
             ThreadSafeQtSignal(self, "workspace_reencryption_error"),
             _reencrypt,
+            core=self.core,
```
vxgmichel: Maybe move the GUI stuff to another PR and get @Max-7 to review it?

tests/core/conftest.py
```diff
@@ -23,28 +23,29 @@ def _local_storage_factory(device, user_manifest_in_v0=False, force=True):
         if not user_manifest_in_v0:
             user_manifest = initial_user_manifest_state.get_user_manifest_v1_for_device(device)
             user_manifest = user_manifest.evolve(author=device_id)
-            local_storage.set_manifest(device.user_manifest_id, user_manifest)
+            async with local_storage.lock_entry_id(device.user_manifest_id):
+                local_storage.set_manifest(device.user_manifest_id, user_manifest)
```
vxgmichel: Maybe use `check_lock_status=False` here since it affects the user manifest?

PR #596, opened 2019-07-26
Timestamp param only for core cli "run" cmd 

touilleMan:


touilleMan:
Appveyor failure not related

PR #598, opened 2019-07-26
POC benchmark

touilleMan:


parsec/core/cli/profile.py
```diff
@@ -0,0 +1,88 @@
...
+    "-t",
+    default="api_4k_chunks",
+    type=click.Choice(("mountpoint", "api", "api_4k_chunks")),
+)
+def benchmark(config, device, mountpoint, stats_out, type, **kwargs):
+    config = config.evolve(mountpoint_enabled=True)
+    if mountpoint:
+        config = config.evolve(mountpoint_base_dir=Path(mountpoint))
+    with cli_exception_handler(config.debug):
+        trio_run(_run_mountpoint, config, device, stats_out, type)
```
vxgmichel: Should this really be a CLI module? Maybe we should add specific benchmark tests to the test suite ? Maybe use [pytest-benchmark](https://github.com/ionelmc/pytest-benchmark/)?

parsec/core/fs/workspacefs/file_transactions.py
```diff
@@ -233,4 +235,5 @@ def _attempt_read(self, manifest: LocalFileManifest, start: int, end: int):
 
     async def fd_flush(self, fd: FileDescriptor) -> None:
         # No-op
-        pass
+        async with self._load_and_lock_file(fd) as (entry_id, manifest):
+            self.local_storage.set_manifest(entry_id, manifest)
```
vxgmichel: This should do the trick :+1:

parsec/core/fs/workspacefs/file_transactions.py
```diff
@@ -172,7 +173,8 @@ def _attempt_read(self, manifest: LocalFileManifest, start: int, end: int):
 
             # Atomic change
             self.local_storage.set_dirty_block(block_access.id, padded_content)
-            self.local_storage.set_manifest(entry_id, manifest)
+            # self.local_storage.set_manifest(entry_id, manifest)
```
vxgmichel: Remove this line?

parsec/core/local_storage.py
```diff
@@ -117,6 +117,12 @@ def set_manifest(self, entry_id: EntryID, manifest: LocalManifest) -> None:
         self.persistent_storage.set_manifest(entry_id, raw)
         self.local_manifest_cache[entry_id] = manifest
 
+    def set_manifest_cache_only(self, entry_id: EntryID, manifest: LocalManifest) -> None:
+        assert isinstance(entry_id, EntryID)
+        assert manifest.author == self.device_id
+        self._check_lock_status(entry_id)
+        self.local_manifest_cache[entry_id] = manifest
```
vxgmichel: This method could also store the `entry_id` in a set of cache-only ids. The `set_manifest` method would also discard the corresponding id from this set. Then a `set_manifest` call have to be issued for each id in the set when `__exit__`ing the local storage.

parsec/core/cli/profile.py
touilleMan: The only reason I did this way was it was the fastest to have something running
This should be moved to misc before merging

touilleMan: or in tests as you said

parsec/core/local_storage.py
touilleMan: I've modified the code to keep a set of ids and flushing them in LocalStorage's `__exit__`

However I'm still not sure how to handle `LocalStorage.clear_manifest` & `LocalStorage.clear_memory_cache`.
They are only used during test so I'm wondering if you shouldn't have a `TestLocalStorage` child class implementing those methods instead
I've considered their desired behavior is to get rid of the  data, so we lose it if it was only in cache.

parsec/core/fs/workspacefs/file_transactions.py
touilleMan: I've improved this to avoid doing the costly serialization if the manifest's cache wasn't ahead of the disk

parsec/core/local_storage.py
```diff
@@ -109,13 +113,26 @@ def get_manifest(self, entry_id: EntryID) -> LocalManifest:
...
+    def set_manifest(
+        self, entry_id: EntryID, manifest: LocalManifest, cache_only: bool = False
+    ) -> None:
         assert isinstance(entry_id, EntryID)
         assert manifest.author == self.device_id
         self._check_lock_status(entry_id)
+        if not cache_only:
+            raw = local_manifest_serializer.dumps(manifest)
+            self.persistent_storage.set_manifest(entry_id, raw)
+        self.local_manifest_cache[entry_id] = manifest
```
vxgmichel: Maybe discard the `entry_id` as well:

    self.cache_ahead_of_persistance_ids.discard(entry_id)

touilleMan: done

vxgmichel:
LGTM!

PR #601, opened 2019-07-29
Optimize persistant storage & add garbage collector test

touilleMan:
- configure sqlite with WAL mode, isolation=None and synchronous=OFF
- blocks are stored in sqlite
- Use integer for accessed_on column (faster than text in db + time.time() much faster than pendulum.now())

before:
before

after:
after

Bench is done on a copy of a 10mo file, getting a ~30% improvement

[profs.zip](https://github.com/Scille/parsec-cloud/files/3444084/profs.zip)

### DIFF
--- a/parsec/core/persistent_storage.py
+++ b/parsec/core/persistent_storage.py
@@ -1,11 +1,10 @@
 # Parsec Cloud (https://parsec.cloud) Copyright (c) AGPLv3 2019 Scille SAS
 
 import os
+from time import time
 from pathlib import Path
 from contextlib import contextmanager
 from sqlite3 import Connection, connect as sqlite_connect
-
-import pendulum
 
 from parsec.core.types import EntryID, BlockID
 from parsec.crypto import SecretKey, encrypt_raw_with_secret_key, decrypt_raw_with_secret_key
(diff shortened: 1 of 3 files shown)


parsec/core/persistent_storage.py
```diff
@@ -211,48 +218,49 @@ def _is_block(self, conn: Connection, path: Path, block_id: BlockID):
...
-        if not changes:
-            raise LocalStorageMissingError(block_id)
+            cursor.execute("""SELECT data FROM blocks WHERE block_id = ?""", (str(block_id),))
+            ciphered, = cursor.fetchone()
+            cursor.execute("END")
 
-        ciphered = self._read_file(block_id, path)
         return decrypt_raw_with_secret_key(self.local_symkey, ciphered)
 
     def _set_block(self, conn: Connection, path: Path, block_id: BlockID, raw: bytes):
```
vxgmichel: A few things need to be removed:
- the `path` argument for `_set_block` and `_get_block`
- the `_dirty_db_files` and `_clean_db_files` attributes

touilleMan: done \o/
(also removed `get_block_cache_size` which is meaningless and wasn't used anyway)

PR #646, opened 2019-08-29
Provide hash digest through crypto module

touilleMan:


parsec/core/fs/remote_loader.py
```diff
@@ -215,7 +215,7 @@ def __init__(
             raise FSError(f"Cannot decrypt block: {exc}") from exc
 
         # TODO: let encryption manager do the digest check ?
-        assert sha256(block).digest() == access.digest, access
+        assert hashdigest(block) == access.digest, access
```
vxgmichel: Shouldn't this be moved out of the remote loader?

PR #659, opened 2019-09-05
Revoke user

touilleMan:


parsec/backend/memory/user.py
```diff
@@ -187,11 +185,13 @@ def _get_user_devices(
...
-            results = [
-                user_id
-                for user_id in results
-                if not user_is_revoked(org._devices[user_id].values())
-            ]
+            now = pendulum.now()
+
+            def _user_is_revoked(user_id):
+                revoked_on = org._users[user_id].revoked_on
+                return revoked_on is not None and revoked_on <= now
```
vxgmichel: `revoked_on <= now` : so a device can be revoked in the future?

parsec/backend/user.py
```diff
@@ -645,7 +680,7 @@ def _on_organization_events(
                 user = await self.get_user(client_ctx.organization_id, client_ctx.device_id.user_id)
 
             except UserNotFoundError as exc:
-                raise RuntimeError("User `{client_ctx.device_id.user_id}` disappeared !") from exc
+                raise RuntimeError(f"User `{client_ctx.device_id.user_id}` disappeared !") from exc
```
vxgmichel: Nice catch!

parsec/backend/app.py
```diff
@@ -247,7 +247,7 @@ def __init__(self, config, event_bus=None):
                     if organization.root_verify_key != expected_rvk:
                         result_req = handshake.build_rvk_mismatch_result_req()
 
-                    elif device.revoked_on:
+                    elif user.revoked_on and user.revoked_on <= pendulum_now():
```
vxgmichel: Same here?

parsec/api/protocol/user.py
```diff
@@ -19,31 +19,28 @@
...
-    "device_revoke_serializer",
 )
 
 
 #### Access user API ####
 
 
-class UserGetReqSchema(BaseReqSchema):
-    user_id = UserIDField(required=True)
+class TrustchainSchema(UnknownCheckedSchema):
```
vxgmichel: So the trustchain doesn't use the `SCHEMA_CLS` API?

parsec/backend/postgresql/init_tables.sql
```diff
@@ -28,6 +28,12 @@ CREATE TABLE user_ (
     -- NULL if certifier is the Root Verify Key
     user_certifier INTEGER,
     created_on TIMESTAMPTZ NOT NULL,
+    -- NULL if not yet revoked
+    revoked_on TIMESTAMPTZ,
```
vxgmichel: So `TIMESTAMPTZ` works in with postgresql but not sqlite?

parsec/backend/user.py
```diff
@@ -108,6 +102,13 @@ def public_key(self) -> PublicKey:
     revoked_user_certifier: DeviceID = None
 
 
+@attr.s(slots=True, frozen=True, auto_attribs=True)
+class Trustchain:
+    users: Tuple[bytes, ...]
+    revoked_users: Tuple[bytes, ...]
+    devices: Tuple[bytes, ...]
```
vxgmichel: :+1: for explicitly defining this trustchain concept

tests/core/fs/workspacefs/test_workspace_fs.py
```diff
@@ -388,7 +388,7 @@
                 "children": {
                     "bar": {
                         "base_version": 1,
-                        "blocksize": 524288,
+                        "blocksize": 524_288,
```
vxgmichel: Really? :smile:

parsec/core/backend_connection/cmds.py
```diff
@@ -48,12 +48,7 @@
     device_cancel_invitation_serializer,
     device_create_serializer,
 )
-from parsec.core.types import (
-    UnverifiedRemoteUser,
-    UnverifiedRemoteDevice,
-    UnverifiedRealmRole,
-    EntryID,
-)
```
vxgmichel: What's an "unverified remote"?

parsec/backend/memory/realm.py
```diff
@@ -195,7 +195,13 @@ def _get_realm(self, organization_id, realm_id):
...
         if encryption_revision != realm.status.encryption_revision + 1:
             raise RealmEncryptionRevisionError("Invalid encryption revision")
-        if per_participant_message.keys() ^ realm.roles.keys():
+        now = pendulum.now()
+        not_revoked_roles = set()
+        for user_id in realm.roles.keys():
+            user = await self._user_component.get_user(organization_id, user_id)
+            if not user.revoked_on or user.revoked_on > now:
+                not_revoked_roles.add(user_id)
+        if per_participant_message.keys() ^ not_revoked_roles:
```
vxgmichel: Funny, I didn't know about that:
```python
>>> {"a": 1, "b": 2}.keys() ^ {"b", "c"}
{'a', 'c'}

>>> isinstance({}.keys(), collections.abc.Set)                             
True

File: parsec/core/fs/remote_loader.py [hunk 10]
```diff
@@ -101,7 +96,7 @@ def __init__(
 
                 # Make sure author had the right to do this
                 existing_user_role = current_roles.get(unsecure_certif.user_id)
-                if not current_roles and unsecure_certif.user_id == author.user_id:
+                if not current_roles and unsecure_certif.user_id == author.device_id.user_id:
```
vxgmichel: In some other parts of the code base, `author` refers to a `DeviceID`. Maybe we should have some naming convention to differentiate a `DeviceID` from a `VerifiedRemoteDevice`.

vxgmichel: I also noticed:
```
± git grep "remote_device_manager" | wc
     11      40    1100
± git grep "remote_devices_manager" | wc
     55     235    5625
```
This is going to bite us for sure :)

parsec/core/remote_devices_manager.py
```diff
@@ -19,12 +16,10 @@
...
-from parsec.core.types import (
-    UnverifiedRemoteDevice,
-    UnverifiedRemoteUser,
-    VerifiedRemoteDevice,
-    VerifiedRemoteUser,
-)
+from parsec.core.trustchain import TrustchainContext, TrustchainError
+
+
+DEFAULT_CACHE_VALIDITY = 60 * 60  # 1h
```
vxgmichel: Can we justify this value? Why not 5 minutes or 5 hours?

parsec/core/trustchain.py
```diff
@@ -0,0 +1,258 @@
...
+    pass
+
+
+class TrustchainContext:
+    def __init__(self, root_verify_key: VerifyKey, cache_validity: int):
+        self.root_verify_key = root_verify_key
+        self.cache_validity = cache_validity
+        self._users_cache = {}
+        self._devices_cache = {}
+        self._revoked_users_cache = {}
```
vxgmichel: So the `Trustchain` class isn't used here?

tests/backend/realm/test_reencryption_maintenance.py
```diff
@@ -37,23 +37,53 @@
 
 
 @pytest.mark.trio
-async def test_start_bad_per_participant_message(backend, alice_backend_sock, alice, bob, realm):
-    # Add bob for more fun !
+async def test_start_bad_per_participant_message(
+    backend, alice_backend_sock, alice, bob, adam, realm
+):
+    # Bob used to be part of the realm
```
vxgmichel: Seems like Bob's not fun anymore :/

parsec/backend/postgresql/init_tables.sql
touilleMan: yep, see https://www.postgresql.org/docs/10/datatype-datetime.html:
> Note
> The SQL standard requires that writing just timestamp be equivalent to timestamp without time zone, and PostgreSQL honors that behavior. **timestamptz** is accepted as an abbreviation for timestamp with time zone; this is a PostgreSQL extension.

parsec/backend/memory/user.py
touilleMan: in theory yes, but it's not used for the moment (so the check is just to keep things logical)

parsec/backend/app.py
touilleMan: same ;-)

tests/core/fs/workspacefs/test_workspace_fs.py
touilleMan: black did this, not me ! ^^

parsec/core/backend_connection/cmds.py
touilleMan: Old concept that stores both the certificate (the binary data) and a timestamp from when the data was received.
The idea was to validate the certificate, which end up in creating a `VerifiedRemoteDevice` structure.

Now all this stuff is replaced by the `DeviceCertificateContent` class

parsec/core/remote_devices_manager.py
touilleMan: No scientific reason here

parsec/core/trustchain.py
touilleMan: No, `Trustchain` class is only defined in the backend to return data for the trustchain field in some commands (this field only provide a subset of the entire trustchain that is useful to validate the certificate provided in the command)

tests/backend/realm/test_reencryption_maintenance.py
touilleMan: ?

parsec/api/protocol/user.py
touilleMan: No, the `Trustchain` class is just a helper internal to the backend (like `UserInvitation` for instance)

parsec/core/fs/remote_loader.py
touilleMan: > This is going to bite us for sure :)

Yep, pretty sure we need a cleanup there

PR #681, opened 2019-09-15
Fix dependencies

touilleMan:


parsec/core/gui/new_version.py
```diff
@@ -19,15 +18,24 @@
 RELEASE_URL = "https://github.com/Scille/parsec-build/releases/latest"
 
 
+def _extract_version_tuple(raw):
+    match = re.match(r"^.*v([0-9]+)\.([0-9]+)\.([0-9]+)", raw)
+    if match:
+        return match.groups()
```
vxgmichel: No mapping to `int`?

touilleMan: done.

PR #725, opened 2019-10-08
WiP: New versions method in workspacefs

c4ffein:
- Added versions method to workspacefs
- Added UT


parsec/core/fs/workspacefs/workspacefs.py
```diff
@@ -440,6 +441,193 @@ def get_encryption_revision(self) -> int:
...
+                    tree,
+                    entry_id,
+                    max(early, timestamp),
+                    late if next_version not in versions else min(late, versions[next_version][0]),
+                    version,
+                    next_version,
+                )
+
+        return_tree = {}
+        root_manifest = await self.transactions._get_manifest_from_path(FsPath("/"))
```
vxgmichel: `workspace_manifest = await self.local_storage.get_manifest(self.workspace_id)`

c4ffein: Couldn't we imagine some special case where the root manifest has been garbage collected? I wasn't sure... I will change this anyway, and put the whole algorithm in a new module as it's indeed a big enough piece of code

vxgmichel: >  Couldn't we imagine some special case where the root manifest has been garbage collected? I wasn't sure...

Oh right. Then:

    workspace_manifest = await self.transactions._get_manifest(self.workspace_id)

PR #734, opened 2019-10-14
Add role certificate validity check when loading manifest in core

### LINKED ISSUE #727: Core should verify the chain of signatures to validate the role of the user (writer, owner etc.)
Currently, the verification of the user roles is performed only by the back-end. The same verification should be performed by users after receiving content from metadata server.

This mechanism is similar with validating administrative signed operations, which happen both at the back-end and at the users side.

touilleMan:
fix #727

### DIFF
--- a/parsec/core/fs/remote_loader.py
+++ b/parsec/core/fs/remote_loader.py
@@ -53,6 +53,24 @@ class RemoteLoader:
         self.backend_cmds = backend_cmds
         self.remote_device_manager = remote_device_manager
         self.local_storage = local_storage
+        self._realm_role_certificates_cache = None
+        self._realm_role_certificates_cache_timestamp = None
+
+    async def _get_user_realm_role_at(self, user_id: UserID, timestamp: Pendulum):
+        if (
+            not self._realm_role_certificates_cache
+            or self._realm_role_certificates_cache_timestamp <= timestamp
+        ):
+            cache_timestamp = pendulum_now()
+            self._realm_role_certificates_cache, _ = await self._load_realm_role_certificates()
+            # Set the cache timestamp in two times to avoid invalid value in case of exception
+            self._realm_role_certificates_cache_timestamp = cache_timestamp
+
+        for certif in reversed(self._realm_role_certificates_cache):
+            if certif.user_id == user_id and certif.timestamp <= timestamp:
+                return certif.role
+        else:
+            return None
 
     async def _load_realm_role_certificates(self, realm_id: Optional[EntryID] = None):
         try:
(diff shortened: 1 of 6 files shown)


tests/core/fs/test_fs_sync.py
```diff
@@ -79,7 +79,8 @@
 @pytest.mark.trio
 @pytest.mark.parametrize("type", ["file", "folder"])
 async def test_new_empty_entry(type, running_backend, alice_user_fs, alice2_user_fs):
-    wid = await create_shared_workspace("w", alice_user_fs, alice2_user_fs)
+    with freeze_time("2000-01-01"):
+        wid = await create_shared_workspace("w", alice_user_fs, alice2_user_fs)
```
vxgmichel: Why is that necessary?

parsec/core/fs/remote_loader.py
```diff
@@ -325,13 +343,26 @@ def __init__(
...
+        )
+        if role_at_timestamp is None:
+            raise FSError(
+                f"Manifest was created at {expected_timestamp} by `{expected_author}` "
+                "which had no right to access the workspace at that time"
+            )
+        elif role_at_timestamp == RealmRole.READER:
+            raise FSError(
+                f"Manifest was created at {expected_timestamp} by `{expected_author}` "
+                "which had write right on the workspace at that time"
```
vxgmichel: `which had no write right`? `which had no right to write in the workspace`?

tests/core/fs/workspacefs/test_invalid_data.py
```diff
@@ -0,0 +1,112 @@
...
+
+            # This should trigger FSError
+            with pytest.raises(FSError) as exc:
+                await workspace.sync()
+            str(exc.value) == exc_msg
+
+            # Also test timestamped workspace
+            with pytest.raises(FSError) as exc:
+                await workspace.to_timestamped(options["backend_timestamp"])
+            str(exc.value) == exc_msg
```
vxgmichel: Maybe parametrize the `testbed` fixture and let the test catch the `FSError` exception? e.g:
```python
@pytest.mark.trio
async def test_empty_blob(testbed):
    with pytest.raises(FSError) as context:
        await testbed.run(blob=b"")
    assert context.value.message == "Cannot decrypt vlob: The nonce must be exactly 24 bytes long"
```

tests/core/fs/test_fs_sync.py
touilleMan: Because otherwise shared workspace is created with current time, but after that we do a `workspace.sync()` with time frozen at 2000-01-03.
So the timestamp consistency test failed given there is no role certificate available at the time we upload the data

parsec/core/fs/remote_loader.py
touilleMan: wrong wording or poor style ?

tests/core/fs/workspacefs/test_invalid_data.py
touilleMan: That was exactly what I was doing before adding the timestamped part ^^

Before that we had only one call raising an exception so it was logical to let the caller test handle it
But with two errors being raised, I should have two methods (well 3 in fact: `init`, `run_workspace`, `run_timestamped_workspace`) that have to be called by the test with exception handling. This is much more cumbersome (especially considering the methods may not be easily idempotent) so I just KISS

PR #736, opened 2019-10-15
[WIP] Add bakend check on user revocation during realm role update

touilleMan:
DO NOT MERGE YET !

@vxgmichel suggested the real trouble here is the fact to process a role change we have to check for 1) the user role 2) whether the user has been revoked

It would be more logical to have 2) implying 1), hence a role change only needs 1).
This is not possible given user roles are signed by workspace owner and user revocations are signed by organization admin (so overlapping is not always possible)

A solution would be to issue a message informing about the revocation to each owner of workspace the user is part. Then the user would be gradually removed from each workspace each time a message is processed.

Another solution would be to have owner take care of removing revoked user from workspace before doing a re-encryption.
This approach has the benefit of simplicity (only have to add a new step on core side before actually doing the re-encryption) and security (checks are done on client side, we don't send the new key to users which have been revoked).

### DIFF
--- a/newsfragments/736.bugfix.rst
+++ b/newsfragments/736.bugfix.rst
@@ -1,0 +1,1 @@
+Add check in backend to prevent non-removal role update on revoked user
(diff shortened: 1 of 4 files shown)


vxgmichel:
Fixed in #2525

PR #744, opened 2019-10-18
[GUI] Workspace time changing from list versions

c4ffein:
Bugs still exist because of actual implementation of backend attributing timestamps to files


parsec/core/fs/remote_loader.py
```diff
@@ -597,11 +597,15 @@ def __init__(self, remote_loader: RemoteLoader, timestamp: Pendulum):
...
-        return await super().load_manifest(entry_id, version=version, timestamp=self.timestamp)
+        return await super().load_manifest(
+            entry_id,
+            version=version,
+            timestamp=None
+            if version is not None
+            else self.timestamp
+            if timestamp is None
+            else timestamp,
+        )
```
vxgmichel: A comment explaining why a timestamped remote loader might need to load manifest from yet another timestamp?

parsec/core/gui/file_history_dialog.py
```diff
@@ -17,29 +17,48 @@ class FileHistoryDialog(QDialog, Ui_FileHistoryDialog):
...
+            workspace_fs=self.workspace_fs,
             path=path,
         )
 
     def add_history(self):
         versions_dict = self.versions_job.ret
         self.versions_job = None
         if not versions_dict:
-            return  # TODO : something something before
+            return  # TODO : error before?
```
vxgmichel: What happened in this  case?

parsec/core/mountpoint/manager.py
```diff
@@ -112,7 +107,10 @@ def get_path_in_mountpoint(
         if timestamp is None:
             workspace = self._get_workspace(workspace_id)
         else:
-            workspace = await self._get_workspace_timestamped(workspace_id, timestamp)
+            try:
+                workspace = self._get_workspace_timestamped(workspace_id, timestamp)
+            except MountpointNotMounted:
+                return await self.remount_workspace_new_timestamp(workspace_id, None, timestamp)
```
vxgmichel: This is a bit confusing, as it doesn't `remount` but simply `mount` a new workspace, which what this method does.

parsec/core/mountpoint/manager.py
```diff
@@ -134,6 +132,47 @@ def get_path_in_mountpoint(
         await self._mountpoint_tasks[(workspace_id, timestamp)].cancel_and_join()
         del self._mountpoint_tasks[(workspace_id, timestamp)]
 
+    async def remount_workspace_new_timestamp(
+        self, workspace_id: EntryID, original_timestamp: Pendulum, target_timestamp: Pendulum
+    ) -> PurePath:
+        # TODO : use different workspaces for temp mount
+        if original_timestamp == target_timestamp:
+            return
```
vxgmichel: Incorrect return value.

parsec/core/mountpoint/manager.py
```diff
@@ -134,6 +132,47 @@ def get_path_in_mountpoint(
...
+        self._timestamped_workspacefs[(workspace_id, target_timestamp)] = new_workspace
+
+        curried_runner = partial(
+            self._runner,
+            new_workspace,
+            self.base_mountpoint_path,
+            config=self.config,
+            event_bus=self.event_bus,
+        )
+        runner_task = await start_task(self._nursery, curried_runner)
```
vxgmichel: Is really necessary to duplicate the runner logic? I felt it would be much simpler to implement a `_load_workspace_timestramped` async method and let `mount_workspace` do its job.

parsec/core/mountpoint/manager.py
c4ffein: This method can either remount implying it would unmount the workspace at the original timestamp, or only mount a new timestamped workspace at target_timestamp without unmounting anything if the original_timestamp is at None. Better add a comment then, right?

parsec/core/mountpoint/manager.py
c4ffein: Actually I would prefer to refactor the whole class in a new PR, wouldn't it be better?

parsec/core/fs/remote_loader.py
c4ffein: Done!

parsec/core/mountpoint/manager.py
```diff
@@ -112,19 +119,14 @@ def get_path_in_mountpoint(
         if timestamp is None:
             workspace = self._get_workspace(workspace_id)
         else:
-            workspace = await self._get_workspace_timestamped(workspace_id, timestamp)
+            try:
+                workspace = self._get_workspace_timestamped(workspace_id, timestamp)
```
vxgmichel: Hmm, `workspace` is useless here right? This block is going to raise in any case right?
```
        if (workspace_id, timestamp) in self._mountpoint_tasks:
            raise MountpointAlreadyMounted(f"Workspace `{workspace_id}` already mounted.")
```
So why not structure it like so:
```python
    async def mount_workspace(self, workspace_id: EntryID, timestamp: Pendulum = None) -> PurePath:
        if (workspace_id, timestamp) in self._mountpoint_tasks:
            raise MountpointAlreadyMounted(f"Workspace `{workspace_id}` already mounted.")
        if timestamp is not None:
            return await self.remount_workspace_new_timestamp(workspace_id, None, timestamp)
        workspace = self._get_workspace(workspace_id)
        runner_task = await self._mount_workspace_helper(workspace)
        return runner_task.value
```

parsec/core/mountpoint/manager.py
```diff
@@ -134,6 +136,55 @@ def get_path_in_mountpoint(
...
+        try:
+            new_workspace = await workspace.to_timestamped(target_timestamp)
+        except FSWorkspaceTimestampedTooEarly as exc:
+            raise MountpointConfigurationWorkspaceFSTimestampedError(
+                f"Workspace `{workspace_id}` didn't exist at `{target_timestamp}`",
+                workspace_id,
+                target_timestamp,
+                workspace.get_workspace_name(),
+            ) from exc
+        self._timestamped_workspacefs[(workspace_id, target_timestamp)] = new_workspace
```
vxgmichel: This could also be a helper: 
```python
    async def _load_workspace_timestamped(self, workspace_id: EntryID, timestamp: Pendulum, original_timestamp: Optional[Pendulum] = None):
        workspace = self._get_workspace(workspace_id)
        try:
            new_workspace = await workspace.to_timestamped(target_timestamp)
        except FSWorkspaceTimestampedTooEarly as exc:
            raise MountpointConfigurationWorkspaceFSTimestampedError(
                f"Workspace `{workspace_id}` didn't exist at `{target_timestamp}`",
                workspace_id,
                target_timestamp,
                workspace.get_workspace_name(),
            ) from exc
        self._timestamped_workspacefs[(workspace_id, target_timestamp)] = new_workspace
        return new_workspace

PR #748, opened 2019-10-21
Update manifest timestamp before the upload

vxgmichel:
Remote manifests need to be timestamped right before the upload.

This fixes a causality break here:
https://github.com/Scille/parsec-cloud/blob/9bfe85aa2d0e9f30dfe727a76ae13ccc3851973d/parsec/core/fs/workspacefs/workspacefs.py#L528-L529

Here the new remote manifest is already timestamped but the manifests for placeholder children have yet to be generated.


tests/core/fs/workspacefs_timestamped/test_sync.py
```diff
@@ -0,0 +1,25 @@
+# Parsec Cloud (https://parsec.cloud) Copyright (c) AGPLv3 2019 Scille SAS
+
+import pytest
+
+
+@pytest.mark.trio
+async def test_sync_timestamp_consistency(alice_workspace):
+    # Add more files and dirs
+    await alice_workspace.mkdir("/test_sync_dir")
```
touilleMan: Ça aurait pas été bien de mettre des `freeze_time` ici ?

vxgmichel: Justement non, le freeze time cache le probleme vu que tous les uploads ont le meme timestamp.

touilleMan: alors un commentaire pour expliquer ça ? ^^

vxgmichel: :+1:

PR #801, opened 2019-11-14
[PoC] Use service nurseries instead of regular nurseries

vxgmichel:
Service nurseries provide more guarantees as their child tasks are protected from cancellation until the body of nursery has exited. The test suite doesn't seem to have any problem with this change. Also bump trio to 0.13.0.


touilleMan:
not a POC anymore :trollface:

PR #806, opened 2019-11-19
Partial refactor of versioning_helpers

c4ffein:


parsec/core/fs/workspacefs/versioning_helpers.py
```diff
@@ -50,109 +51,187 @@ class ManifestDataAndPaths(NamedTuple):
...
-        FSRemoteManifestNotFound
+    Contains a manifest and his known current, previous and next paths.
+    Unlike ManifestDataAndPaths, attributes are mutable.
+    """
+
+    def __init__(self, manifest, source_path=None, destination_path=None, current_path=None):
+        self.manifest = manifest
+        self.source = source_path
+        self.destination = destination_path
+        self.current = current_path
```
vxgmichel: Maybe use `attrs` for this class, so you have a syntax similar to the other data classes.

Also, those would have been a good use case for [dataclasses](https://docs.python.org/3/library/dataclasses.html) if we were using python 3.7.

parsec/core/fs/workspacefs/versioning_helpers.py
```diff
@@ -50,109 +51,187 @@ class ManifestDataAndPaths(NamedTuple):
...
+                return self._manifest_cache[entry_id][version].manifest
             if timestamp:
                 return next(
                     (
-                        t[2]
-                        for t in manifest_cache[entry_id].values()
-                        if t[0] and t[1] and t[0] <= timestamp <= t[1]
+                        t.manifest
+                        for t in self._manifest_cache[entry_id].values()
+                        if t.early and t.late and t.early <= timestamp <= t.late
```
vxgmichel: It looks much better :+1:

parsec/core/fs/workspacefs/versioning_helpers.py
```diff
@@ -50,109 +51,187 @@ class ManifestDataAndPaths(NamedTuple):
...
-        return versions_list_cache[entry_id]
 
-    async def _get_past_path(entry_id: EntryID, version=None, timestamp=None) -> FsPath:
+class VersionsListCache:
+    """
+    Caches results of the remote_loader.list_versions calls (working on EntryIDs).
+    """
+
+    def __init__(self, remote_loader):
+        self._versions_list_cache = {}
```
vxgmichel: How long does this cache live in memory?

parsec/core/fs/workspacefs/versioning_helpers.py
```diff
@@ -50,109 +51,187 @@ class ManifestDataAndPaths(NamedTuple):
...
-                        t[2]
-                        for t in manifest_cache[entry_id].values()
-                        if t[0] and t[1] and t[0] <= timestamp <= t[1]
+                        t.manifest
+                        for t in self._manifest_cache[entry_id].values()
+                        if t.early and t.late and t.early <= timestamp <= t.late
                     )
                 )
+            return
         except (KeyError, StopIteration):
```
vxgmichel: The catching scope is a bit wide. Maybe manage the cache miss and the `StopIteration` separately?

parsec/core/fs/workspacefs/versioning_helpers.py
```diff
@@ -50,109 +51,187 @@ class ManifestDataAndPaths(NamedTuple):
...
             if timestamp:
                 return next(
                     (
-                        t[2]
-                        for t in manifest_cache[entry_id].values()
-                        if t[0] and t[1] and t[0] <= timestamp <= t[1]
+                        t.manifest
+                        for t in self._manifest_cache[entry_id].values()
+                        if t.early and t.late and t.early <= timestamp <= t.late
                     )
```
vxgmichel: Those parenthesis are not necessary.

parsec/core/fs/workspacefs/versioning_helpers.py
```diff
@@ -50,109 +51,187 @@ class ManifestDataAndPaths(NamedTuple):
...
+        try:
+            name = next(
+                name
+                for name, child_id in parent_manifest.children.items()
+                if child_id == current_id
+            )
+        except StopIteration:
+            raise EntryNotFound(entry_id)
+        else:
+            parts.append(name)
```
vxgmichel: This part could be written with a `for-else`, although it's a matter of taste:
```python
# Find the child name
for name, child_id in parent_manifest.children.items():
    if child_id == current_id:
        parts.append(name)
        break
else:
    raise EntryNotFound(entry_id)
```

parsec/core/fs/workspacefs/versioning_helpers.py
```diff
@@ -167,39 +246,39 @@ class ManifestDataAndPaths(NamedTuple):
...
-                except (FSRemoteManifestNotFoundBadVersion, EntryNotFound):
-                    pass
+            async def _populate_source_path(data, entry_id, timestamp):
+                data.source = await _try_get_path_at_timestamp(entry_id, timestamp)
+
+            async def _populate_destination_path(data, entry_id, timestamp):
+                data.destination = await _try_get_path_at_timestamp(entry_id, timestamp)
+
+            async def _populate_current_path(data, entry_id, timestamp):
+                data.current = await _try_get_path_at_timestamp(entry_id, timestamp)
```
vxgmichel: So those attributes are only mutated once?

c4ffein: Yes, I still use this class as a way to pass something that can be modified later by the populate_x_path methods, is it ok or is there a much cleaner way?

parsec/core/fs/workspacefs/versioning_helpers.py
c4ffein: Changed it to yours as I find it cleaner too

parsec/core/fs/workspacefs/versioning_helpers.py
c4ffein: Thanks, missed that obvious thing...

parsec/core/fs/workspacefs/versioning_helpers.py
c4ffein: Done, and I splitted this code to another method to make the main one more readable!

parsec/core/fs/workspacefs/versioning_helpers.py
c4ffein: It is only needed during the list_version function execution, so I added a comment to explain

parsec/core/fs/workspacefs/versioning_helpers.py
c4ffein: Done!

PR #835, opened 2019-11-27
Fix WinFSP's read_directory operation when a marker is provided

touilleMan:


touilleMan:
appveyor ci failure not related to this PR

PR #843, opened 2019-11-29
Create backend_app_factory to replace BackendApp init/teardown ugliness

touilleMan:


tests/backend/user/test_user_revoke.py
```diff
@@ -49,6 +50,9 @@ def bob_revocation_from_alice(alice, bob):
             await spy.wait_with_timeout(
                 "user.revoked", {"organization_id": bob.organization_id, "user_id": bob.user_id}
             )
+            # `user.revoked` event schedules connection cancellation, so wait
+            # for things to settle down to make sure the cancellation is done
+            await trio.testing.wait_all_tasks_blocked()
```
vxgmichel: Is there a reason why we don't have something in the code base to provide this causality?

touilleMan: I can't see how to enforce this... the `user.rekoved` event occurs at a totally arbitrary time and trio code handling backend connection is agnostic to events.

PR #853, opened 2019-12-03
Better exception logging in sentry

vxgmichel:
Sentry seems to have problems recognizing exceptions generated with:
```python
logger.error(msg, exc_info=exc)
```

... and also remove pkcs11


parsec/backend/cli/run.py
```diff
@@ -294,18 +294,18 @@ def run_cmd(
                     try:
                         await backend.handle_client(stream)
 
-                    except Exception as exc:
+                    except Exception:
                         # If we are here, something unexpected happened...
-                        logger.error("Unexpected crash", exc_info=exc)
```
c4ffein: Ok pour passer en exception, mais on veut pas garder exc?..

parsec/core/gui/win_registry.py
```diff
@@ -101,8 +101,8 @@ def _remove_keys(handle):
 def remove_nav_link():
     try:
         _remove_nav_link()
-    except OSError as exc:
-        logger.error("Error while removing navbar registry values: {}".format(str(exc)))
+    except OSError:
+        logger.exception("Error while removing navbar registry values")
```
c4ffein: Pareil..?

Issue #811, opened 2019-11-20
Inconsistent test: test_folder_operations (hypothesis, flaky)


vxgmichel:
Seen on appveyor (`mountpoint\test_folder_operations.py::test_folder_operations`):
```
[00:11:31] ..\tests\core\mountpoint\test_folder_operations.py:229: 
[00:11:31] _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
[00:11:31] c:\python36-x64\lib\site-packages\hypothesis_trio\stateful.py:258: in run_state_machine_as_test
[00:11:31]     return original(state_machine_factory, settings=settings)
[00:11:31] c:\python36-x64\lib\site-packages\hypothesis\stateful.py:131: in run_state_machine_as_test
[00:11:31]     run_state_machine(state_machine_factory)
[00:11:31] c:\python36-x64\lib\site-packages\hypothesis\stateful.py:89: in run_state_machine
[00:11:31]     @given(st.data())
[00:11:31] _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
[00:11:31] 
[00:11:31] self = <hypothesis.core.StateForActualGivenExecution object at 0x0000007F59B9D588>
[00:11:31] message = 'Unreliable assumption: An example which satisfied assumptions on the first run now fails it.'
[00:11:31] 
[00:11:31]     def __flaky(self, message):
[00:11:31]         if len(self.falsifying_examples) <= 1:
[00:11:31] >           raise Flaky(message)
[00:11:31] E           hypothesis.errors.Flaky: Unreliable assumption: An example which satisfied assumptions on the first run now fails it.
[00:11:31] 
[00:11:31] c:\python36-x64\lib\site-packages\hypothesis\core.py:748: Flaky
[00:11:31] ---------------------------- Captured stdout call -----------------------------
[00:11:31] Falsifying example: run_state_machine(factory=FolderOperationsStateMachine, data=data(...))
[00:11:31] state = FolderOperationsStateMachine()
[00:11:31] v1 = state.init()
[00:11:31] v2 = state.touch(name='a', parent=v1)
[00:11:31] v3 = state.mkdir(name='a', parent=v1)
[00:11:31] state.unlink(path=v2)
[...]

PR #854, opened 2019-12-03
Fix user manifest v0 creation

touilleMan:


touilleMan:
CI error not related (see #811)

PR #894, opened 2019-12-20
Improved Readme with modified download links

c4ffein:


c4ffein:
> Could you squash the commits before merging?

Yup don't worry, that's what I planned from the beginning for this kind of PR, and also why I left those kind of commit messages :)

PR #966, opened 2020-02-14
Added a --no-clean option to run_testenv.sh

Max-7:
It allows me to source the same env in another shell so I can start Parsec with arguments.


Max-7:
Useless, the script already prints the variables to export.

Issue #984, opened 2020-02-20
Set default config variable when running testenv


Max-7:
gui_first_launch=false
gui_check_version_at_startup=false
gui_tray_enabled=false

PR #988, opened 2020-02-20
Set default gui variables for easier debugging

Max-7:
#984


vxgmichel:
Tested on windows, works fine :+1:

PR #1017, opened 2020-03-03
Make winfsp-tests-x86 optional for tests on windows

touilleMan:


tests/core/mountpoint/test_winfsp_tests.py
```diff
@@ -3,14 +3,30 @@
...
+        def file_system_tempdir():
+            pass
 
-# Constants
+    else:
+        raise
+
+else:
+    supported_test_cases_marker = pytest.mark.parametrize("test_case", winfsp_tests.TEST_CASES)
+    file_system_tempdir = winfsp_tests.file_system_tempdir
```
vxgmichel: Why not simply use `pytest.skip("...", allow_module_level=True) ` like here?

https://github.com/Scille/parsec-cloud/blob/763e050cdbbc5cf9ec49c5e9b307b92939b0221e/tests/core/mountpoint/test_winfsp_tests.py#L6-L7

touilleMan: Nice ! I didn't know about this ;-)

PR #1023, opened 2020-03-05
Fix sync monitor (with rare elegance !)

touilleMan:


touilleMan:
Related to #1013

parsec/core/sync_monitor.py
```diff
@@ -159,63 +158,66 @@ def set_remote_change(self, entry_id: EntryID) -> bool:
 
         now = timestamp()
 
-        if self.due_time > now:
+        # math.inf means due_time hasn't been set yet (or there is nothing
+        # to sync, in which case it's safe to continue the function)
+        if self.due_time != math.inf and self.due_time > now:
```
vxgmichel: Isn't it a bit counter intuitive? I don't understand why the `math.inf` special case is necessary, I would expect continuity over `due_time` here

parsec/core/sync_monitor.py
```diff
@@ -150,7 +149,7 @@ def set_remote_change(self, entry_id: EntryID) -> bool:
         self.due_time = timestamp()
         return True
 
-    async def tick(self) -> float:
+    async def tick(self, do_sync=True) -> float:
```
vxgmichel: Why is this `do_sync` argument necessary? What's the point of ticking if not for syncing? At least, let's have a docstring and an explaination

parsec/core/sync_monitor.py
```diff
@@ -358,10 +365,15 @@ def _on_sharing_updated(sender, new_entry, previous_entry):
...
-            with trio.move_on_at(min(wait_times)):
+            wait_time = min(wait_times)
+            if wait_time == math.inf:
+                task_status.idle()
+            with trio.move_on_at(wait_time) as cancel_scope:
                 await early_wakeup.wait()
                 early_wakeup = trio.Event()
             wait_times.clear()
+            if cancel_scope.cancelled_caught:
+                task_status.awake()
```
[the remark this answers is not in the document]
touilleMan: `task_status.awake()` is already called in case we have a early wakeup, only a sleep that came to it end need to call awake() here

parsec/core/sync_monitor.py
touilleMan: During the init step of the monitor, we call tick for the userfs (and also for each workspacefs present). We don't want this tick to actually do sync work (because it means the init will take more time), we only want to know the due_time.

parsec/core/sync_monitor.py
touilleMan: Yup I spend a good amount of time trying to make this clearer but didn't come to anything...

The idea is due_time can be:
- in the past, in which case we should do a synchronization
- in the future, in which case we have nothing to do yet
- in the infinite future, in which case we have nothing to do OR the due_time hasn't been computed

If due_time hasn't been computed, we should do the synchronization (given `self._refresh_checkpoint()` has just been called, `self.remote_changes` and `self.local_changes` has just been modified)
If we have nothing to do, we can also do a synchronization (which will do nothing and set the due_time to math.inf, so nothing is modified)

I considered using None as value for uninitialized due_time, but it feel really buggy (must check everywhere before doing comparisons)

parsec/core/backend_connection/authenticated.py
```diff
@@ -248,26 +257,47 @@ def _cancel_manager_connect(self):
...
+                    monitors_states[idx] = "IDLE"
+                    if all(state == "IDLE" for state in monitors_states):
+                        self._monitors_idle_event.set()
 
-                    async def _bootstrap_monitor(monitor_cb):
-                        await monitors_nursery.start(monitor_cb)
+                def _awake():
+                    monitors_states[idx] = "AWAKE"
+                    if self._monitors_idle_event.is_set():
+                        self._monitors_idle_event = trio.Event()
```
vxgmichel: Shoudn't `_awake` be idempotent just like `_idle`?

parsec/core/sync_monitor.py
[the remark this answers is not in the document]
vxgmichel: Hmm, but `awake` is also called here:
https://github.com/Scille/parsec-cloud/pull/1023/commits/a4932211c58c837aba0b9b60172b2a6f1a79eb79#diff-5f7596c95b3b87cd0cecd47d5dc28b07R309
Wouldn't it be safer to make `awake` idempotent and call it whenever we wake up?

parsec/core/sync_monitor.py
vxgmichel: Then why use tick to compute the due time? Why not simply make sure that `sync_context.due_time` is always the right value?

parsec/core/sync_monitor.py
vxgmichel: I think I'm missing the point of this extra logic... Why can't we stick to the existing logic of:
- Tick: is it time for me to wake up?
- If no: return
- If yes: synchronize what needs to be synchronize

PR #1030, opened 2020-03-11
Use Azure pipelines for CI

### LINKED ISSUE #1031: Migrate to azure pipelines for CI

touilleMan:
fix #1031


.azure-pipelines.yml
```diff
@@ -0,0 +1,298 @@
...
+trigger:
+  branches:
+    include:
+    - '*'
+  tags:
+    include:
+    - '*'
+
+variables:
+  python.version: '3.7'
```
vxgmichel: Bumping to 3.7? Does snap support that?

parsec/core/gui/ui/__init__.py
```diff
@@ -0,0 +1 @@
+# Parsec Cloud (https://parsec.cloud) Copyright (c) AGPLv3 2019 Scille SAS
```
vxgmichel: Was that really necessary ? :thinking:

touilleMan: pre-commit did this, not me ;-)

.azure-pipelines.yml
touilleMan: snap build with 3.6, so we should stick to 3.6 !

PR #1035, opened 2020-03-13
Make sure that winfsp is not required at application startup

### LINKED ISSUE #976: Add regression tests to make sure that winfsp is not required at application startup
It's important that parsec can start even if winfsp is not properly installed. This way, a user might be prompted for updating parsec, which might fix the winfsp installation. In general, a lot of useful checks can be performed at startup and we want to make sure those are going to run.

We need some regression tests to make sure that this constraint is respected, as it is easy to break the rule by adding some misplaced imports to winfspy.

vxgmichel:
Fix #976


.azure-pipelines.yml
```diff
@@ -189,6 +188,14 @@ jobs:
         --junitxml=test-results-gui.xml
     workingDirectory: $(Build.SourcesDirectory)/empty
     displayName: 'Tests GUI'
+  - bash: |
+      set -eux
+      py.test $(pytest.base_args) \
+        ../tests/ --runslow -m slow \
+        -n auto --max-worker-restart=0 -x \
+        --junitxml=test-results-memory.xml
```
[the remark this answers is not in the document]
vxgmichel: :+1:

.azure-pipelines.yml
```diff
@@ -170,11 +170,10 @@ jobs:
   - bash: |
       set -eux
       py.test $(pytest.base_args) \
-        ../tests/ --runmountpoint --runslow \
-        -n auto --max-worker-restart=0 -x \
+        ../tests/ -x \
```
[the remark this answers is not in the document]
vxgmichel: I thought parallelism was only useful for slow tests. Should I re-enable it?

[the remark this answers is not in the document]
touilleMan: yup, pytest-xdist uses subprocesses so it's definitely useful ! ^^

touilleMan:
missing newsfragment !

parsec/core/gui/app.py
```diff
@@ -80,7 +84,31 @@ def _before_quit():
                 continue
 
 
-def run_gui(config: CoreConfig, start_arg: str = None):
+@contextmanager
+def fail_on_first_exception(diagnose, kill_window):
```
touilleMan: Wouldn't it be more explanatory to call this something like `diagnose_exceptions_catcher`, then have a if in `run_gui`:
```python
if diagnose:
    with diagnose_exceptions_catcher(kill_window):
        return app.exec_()
else:
    return app.exec_()
```

tests/test_cli.py
```diff
@@ -67,15 +70,17 @@ def test_share_workspace(tmpdir, alice, bob):
...
     cooked_cmd = ("python -m parsec.cli " + cmd).split()
-    ret = subprocess.run(
-        cooked_cmd, stdout=subprocess.PIPE, stderr=subprocess.PIPE, cwd=CWD, env=env
-    )
-    print(ret.stdout.decode())
-    print(ret.stderr.decode())
+    kwargs = {}
+    if capture:
+        kwargs["stdout"] = kwargs["stderr"] = subprocess.PIPE
+    ret = subprocess.run(cooked_cmd, cwd=CWD, env=env, timeout=timeout, **kwargs)
```
touilleMan: Timeout here is a great idea !

parsec/core/mountpoint/manager.py
```diff
@@ -19,45 +19,37 @@
...
 
 def get_mountpoint_runner():
+    # Windows
     if os.name == "nt":
-        try:
-            import winfspy  # noqa: test import is working
 
-            logging.getLogger("winfspy").setLevel(logging.WARNING)
+        try:
+            import_function("winfspy")
```
touilleMan: why not using good'old `import winfspy` ?

parsec/core/gui/app.py
```diff
@@ -153,9 +181,11 @@ def kill_window(*args):
         # QTimer wakes up the event loop periodically which allows us to close
         # the window even when it is in background.
         timer = QTimer()
-        timer.start(400)
-        timer.timeout.connect(lambda: None)
+        timer.start(1000 if diagnose else 400)
```
touilleMan: Not sure about this... In diagnostic mode we run parsec for 1s, then stop concluding everything was fine.
In theory 1s is far enough for the application to start, but considering what antivirus like Avast are doing I would feel better if we start waiting after the application has reached a certain idle state

parsec/core/gui/app.py
vxgmichel: :+1:

parsec/core/mountpoint/manager.py
vxgmichel: So I can easily mock it in the tests. Any problem with that?

parsec/core/gui/app.py
vxgmichel: Good point! The tricky thing with this command is that we want to test the most out of it while interacting as little as possible with the regular workflow (in order to avoid false positive). It's far from perfect at the moment and it can be improved over time.

Still I'll have a quick look at it and see if I can find a better stop condition for the diagnose run.

parsec/core/mountpoint/manager.py
touilleMan: no it's fine, but we should add a comment about this then ;-(

parsec/core/gui/app.py
touilleMan: 👍

vxgmichel: From the Qt docs on [app.exec](https://doc.qt.io/qt-5/qcoreapplication.html#exec):
> To make your application perform idle processing (by executing a special function whenever there are no pending events), use a QTimer with 0 timeout. More advanced idle processing schemes can be achieved using processEvents().

A timer seems to be ok for this purpose. I'll leave it as it is for the moment.

Issue #952, opened 2020-02-03
Migrate to new UI design


Max-7:

Max-7:
Waiting for the mock ups.

Max-7:
When finished :
 - Test #974

Max-7:
Mockups : https://xd.adobe.com/view/7f4a75d1-9d42-4690-56fa-68114221c105-097b/

UI should use Material icons (https://material.io/resources/icons/?style=baseline). Since icons can be downloaded as SVG, it can be interesting to be able to set up the colors directly in the code instead of using an external software. This provides more flexibility and has the added benefit of not boring the UI programmer (myself) to death by having to correct every icon manually.

Max-7:
- [ ] Transition icons to material
- [ ] Icon color decided in the code
- [ ] Redo info dialogs (warnings, question, text input, ...). See #985 
- [ ] Background greyd out with modals
- [ ] Redo main menu
- [ ] Redo settings
- [ ] Redo login page
- [ ] Redo users and devices pages
- [ ] Redo workspaces page
- [ ] Redo share workspace dialog
- [ ] Redo user/device invite
- [ ] Redo file navigation
- [ ] Redo changelog/about/...

Max-7:
Merged.

PR #1044, opened 2020-03-19
Moved GUI to the new design: new style, new translations, new icons, …

Max-7:
#952


touilleMan:
image
When the app is not fullscreen, it seems there is some overlapping issues

touilleMan:
image
top-left icon is not correctly displayed

touilleMan:
image

Minimize tray is broken:
- select "Minimize in tray" option
- restart Parsec
- no tray icon are displayed :'-(
- closing the Parsec windows doesn't shut down the application (so we end up with a running application in the background that we can no longer bring to the foreground ^^)

PR #1046, opened 2020-03-20
Update nsis installer to retry install after warn parsec is already started

touilleMan:


packaging/win32/installer.nsi
```diff
@@ -94,14 +94,19 @@ Var StartMenuFolder
...
 
-# Check for running Parsec instance.
-Function .onInit
+Function checkParsecRunning
     System::Call 'kernel32::OpenMutex(i 0x100000, b 0, t "parsec-cloud") i .R0'
     IntCmp $R0 0 notRunning
         System::Call 'kernel32::CloseHandle(i $R0)'
         MessageBox MB_OK|MB_ICONEXCLAMATION "Parsec is running. Please close it first" /SD IDOK
-        Abort
+        Call checkParsecRunning
```
vxgmichel: Such marouflage, but much elegance!

PR #1054, opened 2020-03-25
More complex updater using json from GitHub API

c4ffein:


parsec/core/gui/new_version.py
```diff
@@ -28,18 +26,49 @@ def _extract_version_tuple(raw):
...
+    latest_from_head = _extract_version_tuple(resolved_url)
     current_version = _extract_version_tuple(__version__)
-    if lastest_version and current_version and current_version < lastest_version:
-        return lastest_version
-    else:
-        return None
+    if latest_from_head and current_version and current_version < latest_from_head:
+        json_releases = json.loads((await _async_get(api_url)).read())
+
+        current_arch = QSysInfo().currentCpuArchitecture()
```
vxgmichel: We've been using `is_64bits = sys.maxsize > 2 ** 32` in winfspy, but this may be cleaner if Qt is available...

parsec/core/gui/new_version.py
```diff
@@ -28,18 +26,49 @@ def _extract_version_tuple(raw):
...
-    # urlopen automatically follows redirections
-    resolved_url = await trio.to_thread.run_sync(
-        lambda: urlopen(Request(url, method="HEAD")).geturl()
-    )
+def _tuple_to_version_number(version):
+    return ".".join([str(i) for i in version])
+
+
+async def _async_get(url, method="GET"):
+    return await trio.to_thread.run_sync(lambda: urlopen(Request(url, method=method)))
```
vxgmichel: We should consider an actual async library for http requests at some point.

parsec/core/gui/new_version.py
```diff
@@ -28,18 +26,49 @@ def _extract_version_tuple(raw):
...
+
+        current_arch = QSysInfo().currentCpuArchitecture()
+        if current_arch == "x86_64":
+            win_version = "win64"
+        elif current_arch == "i386":
+            win_version = "win32"
+        else:
+            return (latest_from_head, url)
+
+        latest = ((0, 0, 0), "")
```
vxgmichel: I'm being picky here, but why not use the following?
```python
latest_version = (0, 0, 0)
latest_url = ""
```

c4ffein: Done!

parsec/core/gui/new_version.py
c4ffein: Yup, we should...

PR #1055, opened 2020-03-25
Monitors raises BackendNotAvailable instead of FS* errors

touilleMan:


parsec/core/sync_monitor.py
```diff
@@ -356,7 +360,7 @@ def _on_sharing_updated(sender, new_entry, previous_entry):
     async def _ctx_action(ctx, meth):
         try:
             return await getattr(ctx, meth)()
-        except FSBackendOfflineError:
+        except BackendNotAvailable:
             raise
```
vxgmichel: We could have:
```python
except BackendNotAvailable:
    raise
except FSBackendOfflineError as exc:
    logger.exception("Unexpected FSBackendOfflineError exception")
    raise BackendNotAvailable from exc

PR #1056, opened 2020-03-26
Increase verbosity and add duration reports for CI tests

vxgmichel:
Also fixes some inconsistent tests. In particular, this first test can take a bit of time to run:

https://dev.azure.com/Scille/parsec/_build/results?buildId=146&view=logs&j=bfccef3f-45d4-504d-9a7f-a8a74922e6bf&t=85724bb3-46c9-5fad-b9a2-5b861fe1fc50&l=48
`6.62s call     tests/test_cli.py::test_gui_with_diagnose_option[Standard environement]`

A 10 seconds timeout was a bit low.

### DIFF
--- a/.azure-pipelines.yml
+++ b/.azure-pipelines.yml
@@ -14,6 +14,7 @@ variables:
   winfsp.version: '1.7.20038-pre'
   pytest.base_args: |
     --log-level=DEBUG \
+    --durations=10 -v \
     --cov=parsec --cov-config=../setup.cfg --cov-append --cov-report= \
 
 jobs:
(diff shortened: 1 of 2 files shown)


.azure-pipelines.yml
```diff
@@ -88,6 +88,7 @@ jobs:
       py.test $(pytest.base_args) \
         ../tests/ --runmountpoint --runslow \
         -n auto --max-worker-restart=0 -x \
+        --durations=10 -v \
```
touilleMan: Why not put this in `pytest.base_args` ?

PR #1060, opened 2020-03-26
Check newsfragment on PR

### LINKED ISSUE #1033: CI should check for newsfragment
We forget to write a newsfragment too frequently when writing PRs. The CI should enforce the existence of an actual newsfragment. As we want to sometimes write a PR without a newsfragment, we should also add the category "ignore" : that would just be deleted by the releaser script, but let pass the CI.

c4ffein:
Fix #1033

### DIFF
--- a/.azure-pipelines.yml
+++ b/.azure-pipelines.yml
@@ -48,6 +48,8 @@ jobs:
       for FILENAME in newsfragments/*
       do
         # If file never existed in master, consider as a new newsfragment
+        # Cannot git diff against master branch here given newsfragments removed in master will be considered as
+        # new items in our branch
         if ! git log master --exit-code -- newsfragments/"$FILENAME" > /dev/null
         then exit 0; fi
       done
--- a/.azure-pipelines.yml
+++ b/.azure-pipelines.yml
@@ -43,6 +43,18 @@ jobs:
       pre-commit run --all-files --show-diff-on-failure
     displayName: 'Pre-commit hooks check'
   - bash: |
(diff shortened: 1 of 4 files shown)


.azure-pipelines.yml
```diff
@@ -42,6 +42,18 @@ jobs:
   - bash: |
       pre-commit run --all-files --show-diff-on-failure
     displayName: 'Pre-commit hooks check'
+  - bash: |
+      if ([[ $1 = "refs/heads/master" ]] || [[ $1 = "refs/tags/"* ]])
+      then exit 0; fi
+      for FILENAME in newsfragments/*
```
touilleMan: To avoid the for loop:
`[[ $(git --no-pager diff master --name-only -- newsfragments/ | grep -v README.rst | wc -l) -eq 0 ]]`

.azure-pipelines.yml
```diff
@@ -42,6 +42,18 @@ jobs:
   - bash: |
       pre-commit run --all-files --show-diff-on-failure
     displayName: 'Pre-commit hooks check'
+  - bash: |
+      if ([[ $1 = "refs/heads/master" ]] || [[ $1 = "refs/tags/"* ]])
```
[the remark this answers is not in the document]
c4ffein: Mixed that we code used to test locally, now fixed

.azure-pipelines.yml
c4ffein: Added clearer comment

PR #1088, opened 2020-04-01
Fix broken change in boto3

bitlogik:
With the option  --blockstore 's3:s3.amazonaws.com:eu-central-1:bucket:AccessKey:SecretKey', botocore called by Parsec said :  
`ValueError: Invalid endpoint: s3.amazonaws.com`

It was not possible at all to connect to any s3 service providers, and Parsec kept saying "Invalid endpoint". The botocore now requires a leading "https://", because the *is_valid_endpoint_url* method in botocore/utils.py "must have at least a scheme and a hostname" for the given url. In the case "https://" was added in the endpoint_url blockstorage parameter,  Parsec complained : 
`Error: Invalid value for "--blockstore" / "-b": Invalid S3 config, must be 's3:<endpoint_url>:<region>:<bucket>:<key>:<secret>'`

This fixes this issue and now allows to connect to a S3 service again.


bitlogik:
We're investigating to also fix the #1094 issue, and will update this PR with a fix.

bitlogik:
There are major versions conflicts with the docutils package dependency. We are working on it to fix this side effect.

bitlogik:
The pipelines results seems inconsistent. The right configuration seems to be docutils at 0.15. Still, with the exact same code, one time Linux is installing docutils 0.16 and Windows docutils 0.15. So Linux fails the pip check with `botocore 1.15.34 has requirement docutils<0.16,>=0.10, but you have docutils 0.16`. An other time, this is inverted Windows fails because docutils is 0.16, and Linux gets docutils 15.2 and can install Parsec.
Is there cache management for the Azure pipelines which can explain this? Or this is due to our squashing commits and forced changes?

bitlogik:
What can explain this ?

**Linux pipeline**
Boostrap
```
Collecting docutils>=0.13.1
  Downloading docutils-0.16-py2.py3-none-any.whl (548 kB)
```
Install
`ERROR: botocore 1.15.34 has requirement docutils<0.16,>=0.10, but you'll have docutils 0.16 which is incompatible.
`

**Windows pipeline**
Bootstrap
```
Collecting docutils>=0.13.1
  Downloading docutils-0.16-py2.py3-none-any.whl (548 kB)
```
Install :
```
  Attempting uninstall: docutils
    Found existing installation: docutils 0.16
    Uninstalling docutils-0.16:
      Successfully uninstalled docutils-0.16
...
+ pip check 'parsec[all]'
No broken requirements found.
```

bitlogik:
So `pip install -U wheel pip setuptools twine` installs docutils 0.16. Then Windows uninstalls it, but not Linux.
Did the tests passed for the previous stable release?
I'm totally lost and clueless so far, I'll try on a fresh Ubuntu 18 machine.

bitlogik:
Checked on a fresh 18.04 :
Simulating a run of of Linux pipeline 555080395 of the commit 385710ca235cb9
```
apt-get install -y python3-pip

git clone https://github.com/bitlogik/parsec-cloud.git
cd parsec-cloud
git checkout bitlogik-patch-1 [commit 385710ca235cb9 = 

pip3 install -U wheel pip setuptools twine
python3  setup.py bdist_wheel
pip3  install $(ls dist/parsec_cloud-*.whl)[all]
...
Attempting uninstall: docutils
    Found existing installation: docutils 0.16
    Uninstalling docutils-0.16:
      Successfully uninstalled docutils-0.16
...
pip check parsec[all]
No broken requirements found.
```

Something looks like broken in the Linux Azure pipeline, it doesn't uninstall packages right after building. Why?

bitlogik:
To recap this PR :

- Fixes S3 endpoint_url specified in CLI

- Updates S3 client boto3, fixing issue #1094 

- Workaround to fix the Linux test pipeline not properly removing incompatible dependencies. Triggered by the boto3 package update. Issue #1097

touilleMan:
Hi @bitlogik !

Thanks for your hard work here ;-)

I've modified your PR to change the fix for the docutils issue (see here f22e58b6ed96977101f77cff80b120ac6606dd5e), if you're ok with the changes, just update your current PR with it.

Beside before we can merge this PR, can you [sign the contributor agreement](https://github.com/crossbario/crossbar/blob/master/CONTRIBUTING.md#send-us-a-contributor-assignment-agreement) ?

bitlogik:
We are committed to work to improve the Parsec project, I signed and sent the contributor agreement.

The PR is changed with you changes. Except, docutils is not set up to v0.16, as this would require testing, there should be reasons why botocore is specifying less than 0.16, might be some breaking changes, so I considered a conservative approach, one step a a time, only upgrade docutils from v0.14 to 0.15.

touilleMan:
I've updated your branch a bit: the idea is to keep the endpoint_url optional (this is convenient, besides we should avoid breaking backward compatibility)
I've also added an escape mechanism to allow configuration like `s3:https\\:foo.com:abc:123`, I guess it's logical to do this way (it allow the use of non-ssl endpoints, and the swift backend would also benefit from it)

> so I considered a conservative approach, one step a a time, only upgrade docutils from v0.14 to 0.15.

I agree

touilleMan:
aaaand merged ! 👍

PR #1106, opened 2020-04-04
[GUI] Prevent dialog opening if parent is hidden

### LINKED ISSUE #1105: Prevent dialogs from opening if their parent is hidden

Max-7:
Closes #1105

### DIFF
--- a/newsfragments/1105.bugfix.rst
+++ b/newsfragments/1105.bugfix.rst
@@ -1,0 +1,1 @@
+Prevent dialog opening if parent is hidden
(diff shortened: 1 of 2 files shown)


newsfragments/1105.bugfix.rst
```diff
@@ -0,0 +1 @@
+Prevent dialog opening if parent is hidden
```
touilleMan, line 1: Not really clear it's about the GUI, what about:

> Correct GUI to prevent dialog opening if the parent window is hidden

Max-7:
Deprecated.

PR #1112, opened 2020-04-07
Update nsis installer for windows builds

### LINKED ISSUE #1057: Winfsp installation should be silent
As part of the parsec installation process, winfsp installer is run only if it's not already installed. This is fine for the moment but this installation does not run silently and is confusing for the user (see #1050). This can easily be fixed by using:

    msiexec /i winfsp-installer.msi /qn INSTALLLEVEL=1000

Related to #977

vxgmichel:
Closes #1057.

### DIFF
--- a/packaging/win32/freeze_parsec.py
+++ b/packaging/win32/freeze_parsec.py
@@ -57,12 +57,12 @@ def main(parsec_source):
     if not TOOLS_VENV_DIR.is_dir():
         print(f"### Create tool virtualenv ###")
         run(f"python -m venv {TOOLS_VENV_DIR}")
-        run(f"{ TOOLS_VENV_DIR / 'Scripts/pip' } install wheel")
+        run(f"{ TOOLS_VENV_DIR / 'Scripts/python' } -m pip install wheel")
 
     if not WHEELS_DIR.is_dir():
         print(f"### Generate wheels from Parsec&dependencies ###")
         run(
-            f"{ TOOLS_VENV_DIR / 'Scripts/pip' } wheel {parsec_source}[core] --wheel-dir {WHEELS_DIR}"
+            f"{ TOOLS_VENV_DIR / 'Scripts/python' } -m pip wheel {parsec_source}[core] --wheel-dir {WHEELS_DIR}"
         )
 
     # Now we actually generate the build target
@@ -80,7 +80,7 @@ def main(parsec_source):
     print(f"### Installing wheels in temporary virtualenv ###")
     run(f"python -m venv {build_venv_dir}")
     wheels = " ".join(map(str, WHEELS_DIR.glob("*.whl")))
-    run(f"{ build_venv_dir / 'Scripts/pip' } install {wheels}")
+    run(f"{ build_venv_dir / 'Scripts/python' } -m pip install {wheels}")
 
     # Move build virtualenv's site-packages to the build and patch imports
     print(f"### Move site-packages to embedded distribution ###")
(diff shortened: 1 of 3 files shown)


packaging/win32/freeze_parsec.py
```diff
@@ -57,12 +57,12 @@ def main(parsec_source):
     if not TOOLS_VENV_DIR.is_dir():
         print(f"### Create tool virtualenv ###")
         run(f"python -m venv {TOOLS_VENV_DIR}")
-        run(f"{ TOOLS_VENV_DIR / 'Scripts/pip' } install wheel")
+        run(f"{ TOOLS_VENV_DIR / 'Scripts/python' } -m pip install wheel")
```
touilleMan, line 60: Any reason to do so ?

packaging/win32/installer.nsi
```diff
@@ -122,13 +122,11 @@ Function .onInit
...
-    ;Run the uninstaller
+    ;Run the uninstaller sequentially and silently
+    ;https://nsis.sourceforge.io/Docs/Chapter3.html#installerusageuninstaller
     uninst:
       ClearErrors
-      IfSilent +3
-      Exec $R0
-      Goto +2
-      Exec "$R0 /S"
+      ExecWait '"$R0" /S _?=$INSTDIR'
```
touilleMan, line 129: A sarcastic comment is lacking here :trollface: 

More seriously, what is displayed to the user while the uninstall is done in the background ? doesn't it look like the installer is frozen ? (especially given uninstalling parsec is quiet slow...)

packaging/win32/installer.nsi
```diff
@@ -234,7 +232,7 @@ SectionEnd
 !macro InstallWinFSP
     SetOutPath "$TEMP"
     File ${WINFSP_INSTALLER}
-    ExecWait "msiexec /i ${WINFSP_INSTALLER}"
+    ExecWait "msiexec /i ${WINFSP_INSTALLER} /qn INSTALLLEVEL=1000"
```
touilleMan, line 235: image

(a comment would be welcome to understand this option...)

packaging/win32/freeze_parsec.py
vxgmichel: My environment had no pip shortcut for some reason, and I guess it's more robust this way.

packaging/win32/installer.nsi
vxgmichel: Actually the uninstallation is only slow because of the GUI. On my (quite slow) VM, it only takes a couple of seconds. During this period, nothing is show two the user: the dialog is gone, and later the installer pops up.

packaging/win32/installer.nsi
vxgmichel: Not really sure how it works, but that's how you tell msiexec to not stop when a user input is required but instead go through the whole process. I think I stole that from winfuse CI.

packaging/win32/freeze_parsec.py
touilleMan: 👍

packaging/win32/installer.nsi
touilleMan: so it's so slow because it displays useless information that scrolls too fast for the user to read them anyway ? O_o

packaging/win32/installer.nsi
touilleMan: guess it's still better than use uuid4 as name for your options
> to enable the verbose mode, just pass the `--6b76bea6d3334e1da24e12dcc0336c1b` option (this of course disable the `--bd2e6c050730442393343403d02a2705` quiet mode)

packaging/win32/installer.nsi
vxgmichel: Exactly :D

PR #1115, opened 2020-04-07
API v2: simplified invitation system

### LINKED ISSUE #1119: Implement API v2.0 (new authentication + invite api)
Macro issue, depends on #1118,  #1037 and #1078

touilleMan:
fix #1119


tests/core/test_invite.py
```diff
@@ -0,0 +1,292 @@
...
+from parsec.api.protocol import DeviceID, DeviceName, HumanHandle, HandshakeInvitedOperation
+from parsec.backend.invite import DeviceInvitation, UserInvitation
+from parsec.core.backend_connection import (
+    backend_invited_cmds_factory,
+    backend_authenticated_cmds_factory,
+)
+from parsec.core.types import BackendInvitationAddr, LocalDevice
+from parsec.core.invite import (
+    InvitePeerResetError,
+    claimer_retreive_info,
```
Max-7, line 15: "retrieve" in English :p

parsec/backend/client_context.py
```diff
@@ -0,0 +1,162 @@
...
+        super().__init__(transport, handshake)
+        self.organization_id = organization_id
+        self.is_admin = is_admin
+        self.device_id = device_id
+        self.human_handle = human_handle
+        self.public_key = public_key
+        self.verify_key = verify_key
+
+        self.event_bus_ctx = None  # Overwritten in BackendApp.handle_client
+        self.channels = trio.open_memory_channel(100)
```
vxgmichel, line 71: Why 100? Why not `math.inf`?

touilleMan: This channel is consumed by the client, hence we don't want to flood the server in case the client subscribe to events but then doesn't listen to them

tests/core/test_invite.py
touilleMan: 7e537da6bdf1a4518b820394b1bfd7a4d3f4a177 -__-'''

parsec/api/data/invite.py
```diff
@@ -0,0 +1,119 @@
...
+    # Big endian number extracted from bits [20, 40[
+    greeter_sas = (hmac_as_int >> 20) % 2 ** 20
+
+    return claimer_sas, greeter_sas
+
+
+def generate_sas_code_candidates(valid_sas: int, size: int = 3) -> List[int]:
+    candidates = {valid_sas}
+    while len(candidates) < size:
+        candidates.add(randint(0, 2 ** 20 - 1))
```
vxgmichel, line 39: Would it make sense to generate candidates that always differ from `valid_sas`?

For instance we could make sure that the first 5 bits are always different from that `valid_sas` so that the candidates all start with a different character.

parsec/core/cli/invitation.py
```diff
@@ -0,0 +1,395 @@
...
+    return new_device
+
+
+async def _do_claim_device(initial_ctx):
+    async with spinner("Initializing connection with greeter for claiming device"):
+        in_progress_ctx = await initial_ctx.do_wait_peer()
+
+    choices = in_progress_ctx.generate_greeter_sas_choices(size=3)
+    for i, choice in enumerate(choices):
+        display_choice = click.style(str(choice), fg="yellow")
```
vxgmichel, line 269: > `str(choice)`

So we're not using the 32-symbol alphabet here?
```python
In [1]: dict(enumerate("ABCDEFGHJKLMNPQRSTUVWXYZ23456789"))
Out[1]: 
{0: 'A',
 1: 'B',
 2: 'C',
 3: 'D',
 4: 'E',
 5: 'F',
 6: 'G',
 7: 'H',
 8: 'J',
 9: 'K',
 10: 'L',
 11: 'M',
 12: 'N',
 13: 'P',
 14: 'Q',
 15: 'R',
 16: 'S',
 17: 'T',
 18: 'U',
 19: 'V',
 20: 'W',
 21: 'X',
 22: 'Y',
 23: 'Z',
 24: '2',
 25: '3',
 26: '4',
 27: '5',
 28: '6',
 29: '7',
 30: '8',
 31: '9'}
```

parsec/core/cli/invitation.py
```diff
@@ -0,0 +1,395 @@
...
+        click.echo(f" {i} - {display_choice}")
+    code = click.prompt(
```
vxgmichel, line 271: Is the sync prompt gonna be a problem at some point? Maybe wrap it with `trio.to_thread.run_sync()`?

Too bad `prompt-toolkit` doesn't support trio :)

parsec/core/cli/invitation.py
touilleMan: I found the 7 digits number code both simple to read and the easiest to maintain

I guess the advantage of your technique is to reduce the size of the code (only need 4 characters), @Max-7 preference on this ?

(obviously if we change this, we should modify `generate_greeter_sas_choices` so that it returns a 4 characters string, hence the complexity is hidden for the CLI/GUI)

parsec/api/data/invite.py
touilleMan: > Would it make sense to generate candidates that always differ from valid_sas?

`candidates` is a set, so we are guaranteed it's the case

> For instance we could make sure that the first 5 bits are always different from that valid_sas so that the candidates all start with a different character.

I don't think it's a good idea given the user is supposed to check the entire code so similar starting character shouldn't be an issue (and advanced user noticing a pattern and figuring out he only have to check for the first character is a security concern !)

parsec/core/cli/invitation.py
touilleMan: The cli commands that make use of user input are run with task monitor disabled (`trio_run(_greet_invitation, config, device, token, monitor_tasks=False)`) in order to avoid meaningless warnings.

I don't remember clearly the issue, but using `trio.to_thread.run_sync()` to run the prompt wasn't possible. I'll retry this to investigate further ;-)

parsec/core/cli/invitation.py
vxgmichel: I'm all for 4 symbols from a 32-symbols alphabet. Compare the following:
```
C9VC : 0085986
QZKW : 0665326
XCUU : 0608341
X4ZB : 0057173
Y866 : 0947158
NVGG : 0203372
T5J9 : 1024881
MHSW : 0671979
TGKJ : 0271569
RNNQ : 0471439
```

parsec/api/data/invite.py
vxgmichel: > `candidates` is a set, so we are guaranteed it's the case

Oh, right!

> the user is supposed to check the entire code so similar starting character shouldn't be an issue
> (and advanced user noticing a pattern and figuring out he only have to check for the first character is a security concern !)

vxgmichel: > `candidates` is a set, so we are guaranteed it's the case

Oh, right!

> the user is supposed to check the entire code so similar starting character shouldn't be an issue

Fair point.

> (and advanced user noticing a pattern and figuring out he only have to check for the first character is a security concern !)

I really don't find it realistic nor a security concern.

parsec/core/cli/invitation.py
vxgmichel: > I don't remember clearly the issue, but using trio.to_thread.run_sync() to run the prompt wasn't possible. I'll retry this to investigate further ;-)

Oh really? I'm quite curious about it!

touilleMan: No troubles in fact, I've change the code ;-)

parsec/core/cli/invitation.py
touilleMan: 👍

PR #1129, opened 2020-04-17
Support replace_if_exists flag in the rename winfsp operation

### LINKED ISSUE #1128: Some "save" operations fail on windows
Seen on:
- Photos (utility to edit photos on windows)
- Photoshop

This is related to the mechanism used by the third party application to safely save files. This mechanism might use the `replace_if_exists` flag in the `rename` winfsp operation. It turns out this flag is not properly supported at the moment

vxgmichel:
Fix #1128


touilleMan:
BTW, do you think there is a way we could have test on this behavior ?

vxgmichel:
> BTW, do you think there is a way we could have test on this behavior ?

Yes, I was planning to add this before merging :wink:

PR #1158, opened 2020-05-12
Invitation v2 for the GUI

### LINKED ISSUE #1084: Invite a User / Administrator
Feed back of a novice user : It is important to separate the invitation of a simple user from the invitation of an administrator (responsible for the management of all the people in an organization).
image
We propose to suppress the check box in the menu "Inviter un utilisateur", decouple the invitations functions and to create instead two menu : Invite a user / Invite an administrator (responsible for an organization).
The Windows "Invite an administrator" must explain that an Administrator is a user that has special right to manage the organization, which is a group of person, and as a consequence to invite people in the organization. The message must  explain that a good practice is to have two Administrator is case of one being compromize.

### LINKED ISSUE #1120: New authentication in GUI
Depends on #1119

### LINKED ISSUE #1154: Update users list to look more like the devices list

### LINKED ISSUE #1169: GUI: Responsive boostrap organization widget
bootstrap_not_responsive

Max-7:
Closes #1120 
Closes #1154 
Closes #1084 
Closes #1169 
Based on #1115


parsec/core/gui/invite_user_widget.py
```diff
@@ -1,206 +1,276 @@
...
 logger = get_logger()
 
 
-async def _do_registration(core, device, new_user_id, token, is_admin):
-    try:
-        new_user_id = UserID(new_user_id)
-    except ValueError as exc:
-        raise JobResultError("registration-invite-bad-value") from exc
+async def _do_invite_user(device, config, email):
+    async with backend_authenticated_cmds_factory(
```
touilleMan, line 28: Don't want to use `core.backend_cmds` instead ? this would have two small benefits: a) slightly better performances and b) if this fail due to poor connection, the whole application is considered offline

parsec/core/gui/invite_user_widget.py
```diff
@@ -1,206 +1,276 @@
...
+        addr=device.organization_addr,
+        device_id=device.device_id,
+        signing_key=device.signing_key,
+        keepalive=config.backend_connection_keepalive,
+    ) as cmds:
+        rep = await cmds.invite_list()
+        if rep["status"] != "ok":
+            raise JobResultError(rep["status"])
+        return [inv for inv in rep["invitations"] if inv["type"] == InvitationType.USER]
```
touilleMan, line 56: No love for device invitations ?
(I think it would be better to let the filter be done on the gui side, or to return a tuple `(<list of user invitations>, <optional device invitation>)`)

touilleMan: BTW this let me to a question: wouldn't it be better to change the `invitation_list` api from this
```python
{
    "invitations": [
        <user invitation 1>,
        <device invitation>,
        <user invitation 2>,
    ]
}
```

to this:
```python
{
    "user_invitations": [<user invitation 1>, <user invitation 2>],
    "device_invitation": <device invitation>
 }
```
This would enforce the fact there could be at most only one device invitation at a time, and would avoid weird filtering on client side to sort client/device invitations

What do you think ? (and @vxgmichel )

parsec/core/gui/invite_user_widget.py
```diff
@@ -1,206 +1,276 @@
...
 
-        token = core_generate_invitation_token()
-        try:
-            addr = BackendOrganizationClaimUserAddr.build(
-                self.core.device.organization_addr, user_id=new_user_id
-            )
-        except ValueError as exc:
-            show_error(self, _("TEXT_INVITE_USER_WRONG_PARAMETERS"), exception=exc)
-            return
+    def _on_list_invitations_success(self, job):
```
touilleMan, line 233: I would be good to listen on `backend.invite.status_changed` event to know when to update the list
This typically useful to remove inivatiton deleted on other devices, or to switch the invitation status from idle to ready (and back again) when peer start claiming the invitation (it could be good to also send a gui notification in such case)

PS: `backend.invite.status_changed` doesn't exist yet, update should be made here first: 
https://github.com/Scille/parsec-cloud/blob/91116b4b277588b05634fc9fc54ae21a6b79aae3/parsec/core/backend_connection/authenticated.py#L75

parsec/core/gui/claim_user_widget.py
```diff
@@ -1,207 +1,660 @@
...
+        GetGreeterSas = 3
+        SignifyTrust = 4
+        GetClaimerSas = 5
+        WaitPeerTrust = 6
+        ClaimUser = 7
 
-    except InviteClaimBackendOfflineError as exc:
-        raise JobResultError("backend-offline", info=str(exc)) from exc
+    def __init__(self):
+        self.main_oob_send, self.main_oob_recv = trio.open_memory_channel(0)
```
touilleMan, line 46: Not sure about the naming: "oob" stands for "Out of bound" (which refer to the fact something is passed outside of the application)...

I think it would be better to enforce a naming like "worker" for the coroutine actually doing the work and "driver" for the coroutines requesting operations (would be cool to add a bit of documentation about this in the Claimer docstring to easily get the basic concepts of the module)

So we could name them `self.worker_send_channel`, `self.worker_receive_channel`, `self.driver_send_channel`, `self.driver_receive_channel`.

touilleMan: Another consideration is the driver channel that is shared between multiple producers (i.e. multiple Widgets that will trigger multiple jobs). This can cause issues if a job get cancelled before it can retrieve the answer to his request, hence the next job will get it instead of the answer it was expecting...

The cleanest solution for this would be to only keep the driver channel, and send in this channel a reply channel in addition with the command to execute.

However in case something goes wrong (and a job got cancelled), the channel should not be reused given the widget reset recreate Claimer. I wonder if we could take advantage of that to the [close system in trio channel](https://trio.readthedocs.io/en/stable/reference-core.html#trio.open_memory_channel) to ensure we cannot end up in a subtle weird state.

parsec/core/gui/claim_user_widget.py
```diff
@@ -1,207 +1,660 @@
...
+        try:
+            async with backend_invited_cmds_factory(addr=addr) as cmds:
+                r = await self.main_oob_recv.receive()
 
-    try:
-        save_device_with_password(config.config_dir, device, password)
+                assert r == self.Step.RetrieveInfo
+                try:
+                    initial_ctx = await claimer_retrieve_info(cmds)
+                    await self.job_oob_send.send((True, None, initial_ctx.claimer_email))
```
touilleMan, line 57: Would be better to create a class to describe what is exchanged in the channel given the tuple starts to be big...

parsec/core/gui/claim_user_widget.py
```diff
@@ -1,207 +1,660 @@
...
+                except Exception as exc:
+                    await self.job_oob_send.send((False, exc, None))
 
-    except LocalDeviceAlreadyExistsError as exc:
-        raise JobResultError("user-exists") from exc
+                r = await self.main_oob_recv.receive()
 
-    return device, password
+                assert r == self.Step.WaitPeer
```
touilleMan, line 63: The asserts are a really good idea to ensure the state machine is respected ;-)

parsec/core/gui/invite_user_widget.py
vxgmichel: LGTM!

touilleMan:
image
Copy email to the clipboard button stay greyed once clicked, would be better to be able to click multiple times on it

touilleMan:
image
Not really interesting to display the invitation url here (especially given it cannot be fully displayed), I think we should only show the email address (and it would be better to display it entirely... but I guess it would require to change the layout of the users display entirely so that's a story for another PR ^^)

touilleMan:
image
button should be named "new invitation" or something similar

Issue #1082, opened 2020-04-01
Warning window relatively to Acrobate Reader DC bug


tleblond:
Feed back of a novice user april the 1st 2020 : "_People generaly hates that a software changes the default settings. The warning window relatively to Acrobate Reader DC's bug  is unnecessarily anxiety-provoking, user place before a choice of which he does not control the causes and the consequences._"
![image](https://user-images.githubusercontent.com/9845177/78163776-1f9c6800-7449-11ea-9849-ef222039f90a.png)

The previous message is not understandable for an npovice user and should be modified as follow :
In english :
> By default, the Acrobate Reader DC software presents an uncorrected bug (reference of the bug) which prohibits the opening of a PDF file directly in a Parsec enclave. We suggest that you fix this bug.
> If you do not correct this bug, you will still be able to use Parsec, but you will not be able to execute a PDF file on Acrobate Reader DC directly from a Parsec enclave. The solution will then be:
> - either download the files locally before launching them
> - or to install another PDF reader which does not present this bug, like Sumatra

In french :
> Par défaut, le logiciel Acrobate Reader DC présente un bug non corrigé (référence du bug) qui interdit l'ouverture d'un fichier PDF directement dans une enclave Parsec. Nous vous proposons de corriger ce bug.
> Si vous ne corrigez pas ce bug, vous pourrez toujours utiliser Parsec, mais vous ne pourrez pas exécuter un fichier PDF sur Acrobate Reader DC directement depuis une enclave Parsec. La solution sera alors  :
> - soit de télécharger les fichiers en local avant de les lancer 
> - soit d'installer un autre lecteur de PDF qui ne présente pas ce bug, comme Sumatra

The choice button must be :
- Oui, rendre compatible / Yes, make compatible
- Non / No

Max-7:
The message is meant to be anxiety-inducing. Adobe containers are a security feature, not a bug. The bug is only that they don't work well with Parsec mountpoints and the only solution we found is to disable them, but it is not something that should be done lightly. We could improve the message, explaining what containers are and why they don't work well with Parsec (thought it could become too technical), but absolutely not by presenting them as a bug we could fix for the user.

vxgmichel:
Outdated after #1159

touilleMan:
Will be fixed by #1159

Issue #1081, opened 2020-04-01
Acrobat Reader DC does not open a PDF file if another PDF from another workspace is already open


vxgmichel:
Somehow, Acrobat Reader DC does not open a PDF file from a workspace if another PDF from another workspace is already open.

In order to reproduce:
- Open a PDF file using Acrobat Reader (from a workspace or elsewhere on the system)
- Go to a different workspace
- Try to open a PDF file from this workspace using either the file explorer or the application
- Acrobat is expected to pop up and the PDF is expected to open in a new tab
- Instead nothing happens
- Note that it works fine if the file is open directly from acrobat reader

I have absolutely zero idea why this happens...

vxgmichel:
This is due to an Acrobat Reader bug and can be worked around by disabling the protected mode. However, this is not an acceptable solution for us.

The mountpoints-as-drive approach should be to be reconsidered.

touilleMan:
Will be fixed by #1159

Issue #1058, opened 2020-03-26
Workspace name translations


vxgmichel:
The way workspace names are handled is currently inconsistent.

On one hand, the suffix `(shared by X)` is added directly to the workspace entry:

https://github.com/Scille/parsec-cloud/blob/3e1f1a0ebb1f21c42aa8e879737c037a62c7597d/parsec/core/fs/userfs/userfs.py#L795

It appears in both the GUI and the mountpoint directory name but it is not translated to French.

On the other hand, the suffix `(shared with others)` is defined in the Qt translations:

https://github.com/Scille/parsec-cloud/blob/dc5a0cd66339a3dc7e5f22140c302cb14533058b/parsec/core/gui/tr/parsec_en.po#L182

It only appears in the GUI but not in the mountpoint directory name and it is translated to French.

My opinion is that we should separate two distinct issue:

- the identification of the workspace in the GUI: the GUI has the tools to display extra information (the user who created the workspace, the user who shared it, the date of creation and so on).

- the identification of the workspace as a mountpoint directory: a string has to be crafted in a way to avoid confusion if two workspaces have the same name. If we choose to include information like `shared by`, it has to be translated too.

I suggest we stop adding extra information to the workspace name in the workspace entry as it is confusing and error prone. For instance, here's what happens if Bob shares a workspace with Alice and Alice shares it with Clara:

     /home/clara/Parsec/bob_workspace (shared by bob) (shared by alice)

One issue with changing this is that we have to deal with backward compatibility, and in particular workspaces that already contains `(shared with X)` information in their names.

Max-7:
I would also add that when remounting a timestamped workspace (a workspace at a given time) in the GUI, the GUI creates its own name for it (something like workspace_name-temporary-date). On the file system, the workspace is simply named with its usual name (adding a counter eventually like "workspace (2)".

touilleMan:
Partially implemented in #1159

Issue #928, opened 2020-01-22
Investigate path length limit on Windows


Max-7:
And the errors it can create when syncing file from Linux.

Max-7:
I created a ~300 characters path (3 folders of ~100 characters each) on Linux. In it, I put a 200 characters long file (a .txt file) and a few characters long file (the "control group").

No problem on Linux, I've tried a few apps and all of them could open the file.

On Windows, the Parsec GUI acts as usual, and even the Windows Explorer seems to have no trouble displaying large folder names. Errors arise when trying to perform more challenging tasks.

Opening the file leads to different results depending on the app.
For exemple, with Notepad++:
- when trying to open the file by double-clicking on it (.txt are open by default with it), a dialog from EXPLORE.exe appears and warns that the directory name is invalid, nothing else is done).
- when trying to open using the context menu (right click -> Edit with Notepad++), nothing happens.
- when trying to open using the Windows context menu (right click -> Open With...), nothing happens.
- when trying to open it from Notepad++ (menu File/Open), nothing happens

Other apps (Microsoft Word, LibreOffice, WordPad, Acrobat Reader, Chrome, ...) have the same kind of issues (sometimes signaling that the file is not found/can not be opened)

Ironically, the good old Notepad on Windows has absolutely no trouble opening and editing the file at all.

Trying with the file with a short name leads to the same results.

Windows Explorer, while it's able to display the file, can not do anything else. It's impossible to copy/paste a file inside the folder, rename a file or copy a file already in the folder. These operations fail without anything happening (not even an error dialog). It is possible though to copy a file from the folder to another destination.

The GUI works perfectly fine and does all of these operations as expected.

It seems perfectly possible to create an app on Windows handling long file names (don't know the real limit but it's probably given by the file system, see https://en.wikipedia.org/wiki/Filename#Comparison_of_filename_limitations); and indeed, Parsec, WinFSP, and even Notepad don't seem to have problems, but most apps won't be able to open the file, and even the Windows Explorer will have problem getting anything done (other than displaying the files).

These problems could be perceived by a user as being Parsec's fault (especially since Windows Explorer and most other apps do not even display an error message).

Max-7:
Limit path length in the mountpoint?

PR #1159, opened 2020-05-12
Workspaces as windows drives

vxgmichel:
Replacing:  #1142
Fixing: #928, #1058, #1081, #1082, #925, #913, #914, #912, #1126

Todo:
- [x] Use reader role to mark the drive as read only (with remounting if necessary)
- [x] Use parsec icon as drive icon
- [x] Implement a mecanism to enable/disable workspaces in the GUI
- [x] Add enabled/disabled workspace status to the configuration
- [x] Update tests


parsec/core/mountpoint/manager.py
```diff
@@ -222,23 +234,74 @@ async def get_timestamped_mounted(self):
...
 
+            # A nursery dedicated to new workspace events
+            async with trio.open_service_nursery() as mount_nursery:
+
+                # Setup new workspace events
+                events = []
+                if mount_on_workspace_created:
+                    events = [("fs.workspace.created", on_new_workspace)]
+                if mount_on_workspace_shared:
+                    events += [("sharing.updated", on_new_workspace)]
```
touilleMan, line 289: What about umount on sharing lost when `mount_on_workspace_shared=False` ?

parsec/core/mountpoint/winfsp_runner.py
```diff
@@ -20,80 +22,56 @@
...
-
-async def _bootstrap_mountpoint(base_mountpoint_path: Path, workspace_name) -> Path:
-    # Mountpoint can be a drive letter, in such case nothing to do
-    if str(base_mountpoint_path) == base_mountpoint_path.drive:
-        return
-
-    # On Windows, only mountpoint's parent must exists
-    trio_base_mountpoint_path = trio.Path(base_mountpoint_path)
-    await trio_base_mountpoint_path.mkdir(exist_ok=True, parents=True)
+BASE = "PQRSTUVWXYZHIJKLMNO"
```
touilleMan, line 25: Should be good to rename this to a more comprehensible name like `DRIVE_LETTER_CANDIDATES` ?

parsec/core/mountpoint/winfsp_runner.py
```diff
@@ -20,80 +22,56 @@
...
+assert len(BASE) == 19
```
touilleMan, line 26: Why 19 is so important ? (I guess it's for the algo in `candidates`, but would be good to document this)

parsec/core/mountpoint/winfsp_runner.py
```diff
@@ -20,80 +22,56 @@
...
-    # Find a suitable path where to mount the workspace. The check we are doing
-    # here are not atomic (and the mount operation is not itself atomic anyway),
-    # hence there is still edgecases where the mount can crash due to concurrent
-    # changes on the mountpoint path
-    for tentative in count(1):
-        if tentative == 1:
-            dirname = workspace_name
-        else:
-            dirname = f"{workspace_name} ({tentative})"
+def candidates(index, length, nearest=5):
```
touilleMan, line 29: Better naming ? (for instance `_workspace_id_to_drive_letter`)
Adding typing would be good too, and we should replace the str return type by a `List[str]` (better to inform we return multiple drive letters)

And document this non-trivial code ?

parsec/core/mountpoint/winfsp_runner.py
```diff
@@ -20,80 +22,56 @@
...
+    for _ in BASE:
+        result += BASE[index % len(BASE)]
+        index += length
+    return result
 
-        mountpoint_path = base_mountpoint_path / dirname
-        trio_mountpoint_path = trio_base_mountpoint_path / dirname
 
+async def _get_available_drive(index, length) -> Path:
+    drives = [Path(f"{letter}:\\") for letter in candidates(index, length)]
```
touilleMan, line 41: What if there is no valid letter ? (I guess we should return instead an `Optional[Path]`, and have caller code raise an exception if None is returned)

parsec/core/mountpoint/winfsp_runner.py
```diff
@@ -110,23 +88,26 @@ async def winfsp_mountpoint_runner(
...
+        unicodedata.normalize("NFKD", f"{workspace_name.capitalize()}")
         .encode("ascii", "ignore")[:32]
         .decode("ascii")
     )
     volume_serial_number = _generate_volume_serial_number(device, workspace_fs.workspace_id)
     operations = WinFSPOperations(event_bus, volume_label, fs_access)
     # See https://docs.microsoft.com/en-us/windows/desktop/api/fileapi/nf-fileapi-getvolumeinformationa  # noqa
     fs = FileSystem(
-        str(mountpoint_path.absolute()),
+        mountpoint_path.drive,
```
touilleMan, line 110: What happens if the drive has been taken in the meantime ?

parsec/core/mountpoint/winfsp_runner.py
```diff
@@ -154,25 +135,40 @@ async def winfsp_mountpoint_runner(
...
-        event_bus.send("mountpoint.started", mountpoint=mountpoint_path)
-        task_status.started(mountpoint_path)
+                    # Restart workspace if necessary
+                    if workspace_fs.is_read_only() != fs.volume_params["read_only_volume"]:
+                        restart = partial(fs.restart, read_only_volume=workspace_fs.is_read_only())
+                        await trio.to_thread.run_sync(restart)
 
-        await trio.sleep_forever()
+                    # Wait and reset waiter
+                    await waiter.wait()
```
touilleMan, line 170: Shouldn't `waiter.wait()` be first in the loop ?

parsec/core/mountpoint/winfsp_runner.py
```diff
@@ -154,25 +135,40 @@ async def winfsp_mountpoint_runner(
...
+
+            # Notify the manager that the mountpoint is ready
+            event_bus.send("mountpoint.started", **event_kwargs)
+            task_status.started(mountpoint_path)
 
-        # Run fs start in a thread, as a cancellable operation
-        # This is because fs.start() might get stuck for while in case of an IRP timeout
-        await trio.to_thread.run_sync(fs.start, cancellable=True)
+            # Start recording `sharing.updated` events
+            with event_bus.waiter_on("sharing.updated") as waiter:
```
touilleMan, line 159: This doesn't seems to be restricted to winfspy, wouldn't it be better to put it in the generic manager along with the event handler dealing with workspace creation/sharing automount ?

tests/core/mountpoint/test_base.py
```diff
@@ -275,8 +275,10 @@ async def test_get_path_in_mountpoint(base_mountpoint, alice_user_fs, event_bus)
         bar_path = mountpoint_manager.get_path_in_mountpoint(wid, FsPath("/bar.txt"))
 
         assert isinstance(bar_path, PurePath)
-        expected = base_mountpoint / f"mounted_wksp" / "bar.txt"
-        assert str(bar_path) == str(expected.absolute())
+        # Windows uses drives, not base_mountpoint
+        if os.name != "nt":
+            expected = base_mountpoint / f"mounted_wksp" / "bar.txt"
```
touilleMan, line 280: no need for the fstring, and a single string would save one Path construction ^^

parsec/core/gui/workspaces_widget.py
```diff
@@ -432,6 +433,13 @@ def unmount_workspace(self, workspace_id, timestamp=None):
             timestamp=timestamp,
         )
 
+    def update_workspace_config(self, workspace_id, state):
+        if state:
+            self.disabled_workspaces -= {workspace_id}
```
touilleMan, line 438: .discard()/.add() ?

vxgmichel: We can't, `disabled_workspaces` is a `frozenset`.

tests/core/mountpoint/test_base.py
vxgmichel: :+1:

parsec/core/mountpoint/winfsp_runner.py
vxgmichel: It's kind of restricted to winfsp as it is relasted to the fact that we can't change the `readonly` status of the drive at runtime.

Good point though, this reminds me that I still have to investigate how read only mountpoints work on linux.

parsec/core/mountpoint/winfsp_runner.py
vxgmichel: No:
- First we start the mountpoint
- Then we connect to the events

That means we could miss an event in between those two points. I think the following logic prevents race condition in general:
```python
with listen() as event:
    while True:
        if fix_needed():
            fix()
        event.wait_and_reset()
```

parsec/core/mountpoint/winfsp_runner.py
vxgmichel: At the moment, the mount task will simply fail with a `MountpointDriverCrash` exception.

Should we investigate a better solution?

parsec/core/mountpoint/winfsp_runner.py
vxgmichel: Good point, we probably want a specific exception for that.

parsec/core/mountpoint/winfsp_runner.py
vxgmichel: :+1:

parsec/core/mountpoint/winfsp_runner.py
vxgmichel: Cause it's prime, but a comment is definitely need :+1:

parsec/core/mountpoint/winfsp_runner.py
vxgmichel: :+1:

parsec/core/mountpoint/manager.py
vxgmichel: I was thinking about it too, I'll investigate.

parsec/core/mountpoint/manager.py
```diff
@@ -198,13 +210,13 @@ async def remount_workspace_new_timestamp(
                 raise MountpointNotMounted(f"Workspace `{workspace_id}` not mounted.")
 
             await self._mountpoint_tasks[(workspace_id, original_timestamp)].cancel_and_join()
-            del self._mountpoint_tasks[(workspace_id, original_timestamp)]
         return runner_task.value
 
-    async def mount_all(self, timestamp: Pendulum = None):
+    async def mount_all(self, timestamp: Pendulum = None, exclude: list = ()):
```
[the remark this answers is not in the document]
vxgmichel: :+1:

parsec/core/mountpoint/manager.py
```diff
@@ -134,15 +140,22 @@ async def _load_workspace_timestamped(
...
+                return await self._runner(
+                    self.user_fs,
+                    workspace_fs,
+                    self.base_mountpoint_path,
+                    config=self.config,
+                    event_bus=self.event_bus,
+                    task_status=task_status,
+                )
+            finally:
+                self._mountpoint_tasks.pop(key, None)
```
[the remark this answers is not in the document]
vxgmichel: :+1:

parsec/core/mountpoint/manager.py
```diff
@@ -134,15 +140,22 @@ async def _load_workspace_timestamped(
...
+
+        key = (workspace_fs.workspace_id, timestamp)
```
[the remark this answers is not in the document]
vxgmichel: :+1:

parsec/core/mountpoint/manager.py
```diff
@@ -51,6 +54,9 @@ def get_mountpoint_runner():
         logging.getLogger("winfspy").setLevel(logging.WARNING)
         from parsec.core.mountpoint.winfsp_runner import winfsp_mountpoint_runner
 
+        # Now is a good time to perform some cleanup in the registry
```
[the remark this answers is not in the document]
vxgmichel: :+1: 

I put it here is because we're in windows land at this point, but that's not really a good reason.

parsec/core/mountpoint/fuse_runner.py
```diff
@@ -106,7 +107,14 @@ async def fuse_mountpoint_runner(
     )
 
     try:
-        event_bus.send("mountpoint.starting", mountpoint=mountpoint_path)
+
+        # Prepare event information
+        event_kwargs = {
```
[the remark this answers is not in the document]
vxgmichel: :+1:

parsec/core/gui/workspaces_widget.py
```diff
@@ -322,13 +334,30 @@ def add_workspace(self, workspace_fs, ws_entry, users_roles, files, timestamped)
...
         workspace_name = self.jobs_ctx.run_sync(workspace_fs.get_workspace_name)
+
+        # Temporary code to fix the workspace names edited by
+        # the previous naming policy (the userfs used to add
+        # `(shared by <device>)` at the end of the workspace name)
+        token = " (shared by "
+        if token in workspace_name:
+            workspace_name, *_ = workspace_name.split(token)
+            self.jobs_ctx.submit_job(
+                ThreadSafeQtSignal(self, "ignore_success", QtToTrioJob),
```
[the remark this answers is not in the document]
vxgmichel: I don't think so, renaming should never conflict with mounting logic (since workspaces are identified with their id).

parsec/core/gui/workspace_button.py
```diff
@@ -232,5 +297,36 @@ def reload_workspace_name(self, workspace_name):
...
+
+    def reload_workspace_mounted_status(self):
+        is_mounted = self.workspaces_widget.is_workspace_mounted(
+            self.workspace_fs.workspace_id, self.timestamp
+        )
+        self.switch_button.setChecked(is_mounted)
+        self.on_switch_toggled(is_mounted)
+
+    def on_mountpoint_started(self, workspace_id, timestamp):
+        if workspace_id != self.workspace_id or timestamp != self.timestamp:
```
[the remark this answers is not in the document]
vxgmichel: I don't think so, that mountpoint manager uses `(workspace_id, timestamp)` as a key for triggering `WorkspaceNotMounted` and `WorkspaceAlreadyMounted`.

parsec/core/gui/workspace_button.py
```diff
@@ -117,11 +132,34 @@ def __init__(
...
             self.label_shared.hide()
         self.reload_workspace_name(self.workspace_name)
+        self.reload_workspace_mounted_status()
+
+    @property
+    def is_shared(self):
+        return len(self.users_roles) > 1
+
+    @property
+    def owner(self):
```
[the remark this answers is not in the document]
vxgmichel: Right. 

I think @Max-7 is going to re-work that part anyway :grin:

parsec/core/gui/workspace_button.py
```diff
@@ -36,27 +39,39 @@ def __init__(
...
-        self.is_shared = is_shared
         self.reencrypting = None
         self.setCursor(QCursor(Qt.PointingHandCursor))
         self.widget_empty.layout().addWidget(EmptyWorkspaceWidget())
         self.setContextMenuPolicy(Qt.CustomContextMenu)
         self.customContextMenuRequested.connect(self.show_context_menu)
         files = files or []
 
+        # Add switch button
+        # TODO: integrated properly in the .ui files
```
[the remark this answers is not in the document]
vxgmichel: Yes, @Max-7 FTW

parsec/core/gui/tr/parsec_en.po
```diff
@@ -260,8 +260,17 @@ msgstr "Only a workspace owner can reencrypt."
...
+msgstr "private"
+
+msgid "TEXT_WORKSPACE_IS_SHARED_WITH_user"
+msgstr "shared with {user}"
+
+msgid "TEXT_WORKSPACE_IS_SHARED_WITH_n_USERS"
+msgstr "shared with {n} users"
+
+msgid "TEXT_WORKSPACE_IS_OWNED_BY_user"
+msgstr "owned by {user}"
```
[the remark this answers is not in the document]
vxgmichel: https://github.com/Scille/parsec-cloud/pull/1159#discussion_r431760724

parsec/core/gui/settings_widget.py
```diff
@@ -19,16 +18,13 @@ def __init__(self, core_config, jobs_ctx, event_bus, *args, **kwargs):
...
-            self.widget_misc.hide()
-        else:
-            if not win_registry.is_acrobat_reader_dc_present():
-                self.widget_misc.hide()
-            else:
-                self.check_acrobat_container.setChecked(
-                    not win_registry.get_acrobat_app_container_enabled()
-                )
+
+        # TODO: remove the old acrobat compatibility section
```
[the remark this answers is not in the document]
vxgmichel: Yes, @Max-7 probably wants to update the UI files here.

parsec/core/gui/main_window.py
```diff
@@ -387,25 +388,22 @@ def showMaximized(self, skip_dialogs=False):
...
+            # Update from parsec `<1.13` to `>=1.13`
+            if LooseVersion(self.config.gui_last_version) < "1.13":
+
+                # Revert the acrobat reader workaround
+                if (
+                    platform.system() == "Windows"
+                    and win_registry.is_acrobat_reader_dc_present()
+                    and not win_registry.get_acrobat_app_container_enabled()
+                ):
+                    win_registry.del_acrobat_app_container_enabled()
```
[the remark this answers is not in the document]
vxgmichel: We don't have that many users yet so removing the workaround when updating from under `1.13` to above `1.13` seems like an acceptable thing to do.

> What is the behavior if we uninstall Parsec instead of updating it ?

The workaround will stay in place until the new parsec installation.

parsec/core/fs/userfs/userfs.py
```diff
@@ -792,7 +792,7 @@ async def _process_message_sharing_granted(
         # Finally insert the new workspace entry into our user manifest
         workspace_entry = WorkspaceEntry(
             # Name are not required to be unique across workspaces, so no check to do here
-            name=f"{msg.name} (shared by {msg.author.user_id})",
+            name=f"{msg.name}",
```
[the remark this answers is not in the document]
vxgmichel: :+1:

parsec/core/mountpoint/winfsp_runner.py
touilleMan: no I think it's fine (as long as the GUI can handle this properly of course ! ^^)

parsec/core/mountpoint/winfsp_runner.py
touilleMan: 👍

parsec/core/gui/workspaces_widget.py
touilleMan: but `-=` is supposed to be an in-place operator... 

```
>>> x = {1, 2}
>>> y = x
>>> x -= {2}
>>> x
{1}
>>> y
{1}
>>> x = frozenset({1})
{}>>> x
frozenset({1})
>>> y = x
>>> x += {2}
  File "<stdin>", line 1, in <module>
TypeError: unsupported operand type(s) for +=: 'frozenset' and 'set'
>>> x -= {2}
>>> x
frozenset({1})
>>> y
frozenset({1})
>>> x -= {1}
>>> x
frozenset()
>>> y
frozenset({1})
```

O_o

This feels really inconsistent to me, I would prefer to use `disabled_workspaces.difference({workspace_id})` and pretend I've never witnessed this whole craziness ^^

vxgmichel: You can't add sets with `+`, you're supposed to `or` them
```python
>>> {1} + {2}
  File "<stdin>", line 1, in <module>
TypeError: unsupported operand type(s) for +=: 'set' and 'set'
>>> {1} | {2}
{1, 2}
```
> but -= is supposed to be an in-place operator...

But "in-place" doesn't mean mutating the existing data structure:
```python
>>> a = (1,)                                                                              
>>> a += (2,)                                                                             
>>> a                                                                                     
(1, 2)

>>> a = 1                                                                                 
>>> a += 2                                                                                
>>> a                                                                                     
3
```

parsec/core/mountpoint/winfsp_runner.py
vxgmichel: Apparently there is no equivalent to `read_only_volume` on fuse.

tests/core/mountpoint/test_base.py
```diff
@@ -55,26 +62,24 @@ async def test_mount_unknown_workspace(base_mountpoint, alice_user_fs, event_bus
...
     ) as mountpoint_manager:
-
         await mountpoint_manager.mount_workspace(wid)
+        bar_txt = get_path_in_mountpoint(mountpoint_manager, wid, "/bar.txt")
         assert await bar_txt.exists()
 
 
 @pytest.mark.trio
 @pytest.mark.mountpoint
+@pytest.mark.skipif(os.name == "nt", reason="Windows uses drive")
```
touilleMan, line 82: Is there a test "drive already used" for windows ?

parsec/core/gui/workspaces_widget.py
```diff
@@ -176,18 +177,24 @@ def __init__(self, core, jobs_ctx, event_bus, **kwargs):
...
+        self.unmount_success.connect(self.on_unmount_success)
+        self.unmount_error.connect(self.on_unmount_error)
+
         self.workspace_reencryption_success.connect(self._on_workspace_reencryption_success)
         self.workspace_reencryption_error.connect(self._on_workspace_reencryption_error)
 
+        self.reset_required = False
         self.reset_timer = QTimer()
-        self.reset_timer.setInterval(1000)
+        self.reset_timer.setInterval(self.RESET_TIMER_THRESHOLD)
```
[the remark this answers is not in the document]
vxgmichel: This timer is not used for polling, it is used for throttling the requests to reset the GUI.

tests/core/mountpoint/test_base.py
vxgmichel: I'll add one!

PR #1165, opened 2020-05-28
Use 4 symbols from a 32-symbol alphabet as SAS code

vxgmichel:


tests/api/test_types.py
```diff
@@ -104,3 +105,23 @@ def test_valid_human_handle(email, label):
...
     with pytest.raises(ValueError):
         HumanHandle(email, label)
+
+
+def test_sas_code():
+    assert SASCode.from_int(0x0) == SASCode("AAAA")
+    assert SASCode.from_int(0x1) == SASCode("BAAA")
+    # [...]
+    assert SASCode.from_int(0x84001) == SASCode("BASS")
+    # [...]
```
touilleMan, line 115: no test on `1337` ? :trollface:

parsec/core/cli/invitation.py
```diff
@@ -6,7 +6,7 @@
 from functools import partial
 
 from parsec.utils import trio_run
-from parsec.cli_utils import cli_exception_handler, spinner, operation
+from parsec.cli_utils import cli_exception_handler, spinner, operation, aprompt, aconfirm
```
touilleMan, line 9: Why is this part of the PR ? (it seems a commit of mine was included...)

tests/api/test_types.py
vxgmichel: `1337` is not valid, but `L337` is :wink:

PR #1174, opened 2020-06-05
Anonymous device

touilleMan:


parsec/core/local_device.py
```diff
@@ -0,0 +1,242 @@
...
+        return self.device_label or str(self.device_id.device_name)
+
+    @property
+    def slug(self) -> str:
+        # Drop the `.keys` suffix
+        return self.key_file_path.name.rsplit(".", 1)[0]
+
+    @property
+    def slughash(self) -> str:
+        return sha256(self.slug.encode()).hexdigest()
```
vxgmichel, line 118: Necessary code duplication?
https://github.com/Scille/parsec-cloud/pull/1174/commits/6e39a87268553f364a2564fe8545e882654254b1#diff-310fbdc384848dedf0feca35dfb5f58cR80-R88

parsec/core/backend_connection/cmds.py
```diff
@@ -679,6 +681,8 @@ async def organization_bootstrap(
         root_verify_key=root_verify_key,
         user_certificate=user_certificate,
         device_certificate=device_certificate,
+        redacted_user_certificate=redacted_user_certificate,
+        redacted_device_certificate=redacted_device_certificate,
```
vxgmichel, line 685: Funny we missed that!

parsec/core/invite/greeter.py
```diff
@@ -358,10 +358,8 @@ class DeviceGreetInProgress4Ctx:
     _shared_secret_key: SecretKey
     _cmds: BackendInvitedCmds
 
-    async def do_create_new_device(
-        self, author: LocalDevice, device_name: DeviceName, device_label: Optional[str]
-    ) -> None:
-        device_id = DeviceID(f"{author.user_id}@{device_name}")
+    async def do_create_new_device(self, author: LocalDevice, device_label: Optional[str]) -> None:
+        device_id = DeviceID(f"{author.user_id}@{DeviceName.new()}")
```
vxgmichel, line 362: A `DeviceID.new_from_user_id()` method might make sense?

parsec/core/invite/organization.py
```diff
@@ -0,0 +1,65 @@
...
+from parsec.crypto import SigningKey
+from parsec.api.data import UserCertificateContent, DeviceCertificateContent, UserProfile
+from parsec.api.protocol import HumanHandle, DeviceID
+from parsec.core.types import LocalDevice, BackendOrganizationAddr
+from parsec.core.local_device import generate_new_device
+from parsec.core.backend_connection import APIV1_BackendAnonymousCmds
+
+
+async def bootstrap_organization(
+    cmds: APIV1_BackendAnonymousCmds,
```
vxgmichel, line 15: So we're still using APIV1 in some non-legacy cases? Are those API V1 vs V2 use cases documented somewhere?

parsec/core/cli/bootstrap_organization.py
```diff
@@ -1,75 +1,46 @@
...
-
-    config_dir = Path(config_dir) if config_dir else get_default_config_dir(os.environ)
-    debug = "DEBUG" in os.environ
-    configure_logging(log_level="DEBUG" if debug else "WARNING")
-
-    with cli_exception_handler(debug):
-        trio_run(_bootstrap_organization, debug, device, addr, config_dir, force, password)
+    with cli_exception_handler(config.debug):
+        # Disable task monitoring given user prompt will block the coroutine
+        trio_run(_bootstrap_organization, config, addr, password, force)
```
vxgmichel, line 46: I'm a bit confused by those two `bootstrap_organizations` commands. It seems like there is one legacy (APIV1) command and one new command, but they both use the `apiv1_backend_anonymous_cmds_factory`, is that normal?

Also, how does one decide when to use which? Does this have to do with the backend version? In this case, why do we want to support old backend versions?

parsec/core/cli/utils.py
```diff
@@ -46,50 +47,61 @@ def wrapper(config_dir, *args, **kwargs):
...
-        raw_org, raw_device_id = parts
-        return (OrganizationID(raw_org), DeviceID(raw_device_id), val)
-    else:
-        raise ValueError("Must follow format `[<organization>:]<user_id>@<device_name>`")
+def format_available_devices(devices: List[AvailableDevice]) -> str:
+    # Try to shorten the slughash to make it easier to work with
+    slughashes = [d.slughash for d in devices]
+    for slughash_len in range(3, 64):
+        if len({h[:slughash_len] for h in slughashes}) == len(slughashes):
+            break
```
vxgmichel, line 55: Good idea!

parsec/core/cli/utils.py
```diff
@@ -46,50 +47,61 @@ def wrapper(config_dir, *args, **kwargs):
...
+
+        all_available_devices = list_available_devices(config.config_dir)
+        devices = []
+        for device in all_available_devices:
+            if device.slughash.startswith(device_slughash):
+                devices.append(device)
 
-        organization_id, device_id, slugname = kwargs["device"]
-        devices = [
-            (o, d, t, kf)
```
vxgmichel, line 71: Bye `o`, `d`, `t`, and `kf`, we won't miss you~ (especially you `kf`)

tests/test_cli.py
```diff
@@ -150,6 +150,27 @@ def _running(cmd, wait_for=None, env={}):
...
+    for _ in range(10):
+        sleep(0.1)
+        stdout = p.live_stdout.read()
+        if wait_txt in stdout:
+            return stdout
+    else:
+        raise AssertionError("Too slow")
+
+
+def _wait_for_regex(p, regex):
```
vxgmichel, line 163: Good idea!

parsec/core/gui/parsec_application.py
```diff
@@ -6,7 +6,7 @@
 
 
 class ParsecApp(QApplication):
-    connected_devices = []
+    connected_devices = set()
```
vxgmichel, line 9: such speed

parsec/test_utils/organization.py
```diff
@@ -2,70 +2,56 @@
...
-    password,
-    administration_token,
-    force,
-):
+    config_dir: Path,
+    backend_address: BackendAddr,
+    password: str,
+    administration_token: str,
+    force: bool,
+) -> Tuple[LocalDevice, LocalDevice, LocalDevice]:
```
vxgmichel, line 42: We lost quite a bit of parametrization here, but I suspect it wasn't so useful in the first place.

touilleMan: yep, the existing params weren't really used.
Besides there provide a kind of "fake genericity" given you can choose the name of the devices, by not their number or their relationship...
So I felt like YAGNI here ;-)

parsec/core/cli/utils.py
touilleMan: image

tests/backend/test_apiv1_organization.py
```diff
@@ -26,20 +26,32 @@ async def organization_create(sock, organization_id, expiration_date=None):
...
+        "bootstrap_token": bootstrap_token,
+        "user_certificate": user_certificate,
+        "device_certificate": device_certificate,
+        "root_verify_key": root_verify_key,
+    }
+
+    if redacted_user_certificate is not _missing:
+        data["redacted_user_certificate"] = redacted_user_certificate
+    if redacted_device_certificate is not _missing:
+        data["redacted_device_certificate"] = redacted_device_certificate
```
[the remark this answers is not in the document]
touilleMan: `organization_bootstrap` command has been updated with new `reducted_user/device_certificate` fields (let say this change is made in API 1.3)

Of course old core using API 1.2 don't know about those fields so they are missing in their payload.

On the contrary, the new version of the core using API 1.3 is required to provide the redacted fields given it creates user/device manifest containing human_handle/device_label fields (which are the fields to be redacted).

So the redacted fields should be mandatory (but are not for backward compatibility reason), hence `None` is not allowed given knowing this field exists means you are not a legacy core !

parsec/core/local_device.py
```diff
@@ -0,0 +1,242 @@
...
+        return str(self.human_handle or self.device_id.user_id)
+
+    @property
+    def device_display(self) -> str:
+        return self.device_label or str(self.device_id.device_name)
+
+    @property
+    def slug(self) -> str:
+        # Drop the `.keys` suffix
+        return self.key_file_path.name.rsplit(".", 1)[0]
```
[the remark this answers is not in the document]
touilleMan: Cool, I didn't know about this ;-)

parsec/core/local_device.py
touilleMan: Yep, the `LocalDevice` and `AvailableDevice` both implement this.
But given it's really trivial stuff I find it better than introducing a shared 1-liner function

parsec/core/invite/greeter.py
touilleMan: You touch something interesting: the DeviceID(f"{xxx.user_id}@whatever")` pattern is used multiple times throughout the codebase, I wonder if it we could provide a method to `UserID` instead to further simplify the use:
```python
author.user_id.to_device_id(DeviceName.new())
```
(not sure about the naming though...)

parsec/core/backend_connection/cmds.py
touilleMan: It's because we didn't used those fields until now ! (see next comment)

parsec/core/invite/organization.py
touilleMan: APIV1 is still useful for administration-related stuff (e.g. organization create&status) and for organization bootstrap

The rational behind this is adiministration related stuff would be much better exposed with a REST API, so we didn't provide it as part of the APIv2. Hence we could provide a REST API in the future, then totally deprecate the APIv1.

The only part that could be discussed is the organization bootstrap process. This process should be done by a Parsec client in a pretty similar fashion that the user/device claim (i.e. an invited party connect to the backend to end up with a valid device).
So it could make sense to consider the organization bootstrap command should be provided by APIv2.
However from an implementation standpoint it's slightly different (the device/user claim doesn't use the same kind of token than the organization boostrap, the claiming process works completely differently with multiple steps where the organization bootstrap has only one) so sticking with APIv1 seems like a good idea given it simplify the codebase and allows us to consider the switch to REST API in the future.

parsec/core/cli/bootstrap_organization.py
touilleMan: see https://github.com/Scille/parsec-cloud/pull/1174#discussion_r440265323

tl;dr: apiv1 is only used for administration authentications (typically organization create/stats) and for organization bootstrap

parsec/core/local_device.py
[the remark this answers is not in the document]
touilleMan: 09c8546a

parsec/core/invite/greeter.py
vxgmichel: > `author.user_id.to_device_id(DeviceName.new())`

Looks good! Could even be `author.user_id.new_device_id()`

touilleMan: `author.user_id.to_device_id(DeviceName.new())` is more useful than `author.user_id.new_device_id()` given it allows to do `author.user_id.to_device_id("laptop")` which is used pretty often in the codebase

PR #1175, opened 2020-06-05
Custom rsync

vamonte:


parsec/core/cli/run.py
```diff
@@ -55,3 +57,121 @@ def run_mountpoint(config, device, mountpoint, timestamp, **kwargs):
...
+        await _import_file(workspace_fs, path, absolute_path)
+        await workspace_fs.sync()
+    return folder_manifest
+
+
+async def _clear_directory(parent_absolute_path, path, workspace_fs, folder_manifest):
+    local_children_keys = [c.parts[-1] for c in path.iterdir()]
+    for name, entry_id in folder_manifest.children.items():
+        if name not in local_children_keys:
+            absolute_path = f"{parent_absolute_path}{name}"
```
touilleMan, line 119: Would be better to use type `Path` for `parent_absolute_path` (well I think we should use this everywhere it's possible !), hence we would avoid weird behavior if the `parent_absolute_path` lack a slash

```python
absolute_path = (parent_absolute_path / name).absolute()
```

parsec/core/cli/run.py
```diff
@@ -55,3 +57,121 @@ def run_mountpoint(config, device, mountpoint, timestamp, **kwargs):
...
+            if await workspace_fs.is_dir(absolute_path):
+                await workspace_fs.rmtree(absolute_path)
+            else:
+                await workspace_fs.unlink(absolute_path)
+            await workspace_fs.sync()
+
+
+async def _sync_directory_content(parent, source, workspace_fs, manifest):
+    for path in source.iterdir():
+        name = path.parts[-1]
```
touilleMan, line 130: `parth.name`?

parsec/core/cli/run.py
```diff
@@ -55,3 +57,121 @@ def run_mountpoint(config, device, mountpoint, timestamp, **kwargs):
...
+        absolute_path = f"{parent}{name}"
+        is_dir = path.is_dir()
+        is_dir_updating = False
+        folder_manifest = None
+        if await workspace_fs.exists(absolute_path):
+            entry_id = manifest.children.get(name)
```
touilleMan, line 136: `name in manifest.children` is not equivalent to `await workspace_fs.exists(absolute_path)` ?

parsec/core/cli/run.py
```diff
@@ -55,3 +57,121 @@ def run_mountpoint(config, device, mountpoint, timestamp, **kwargs):
...
+
+
+async def _update_file(workspace_fs, entry_id, path, absolute_path):
+    remote_file_manifest = await workspace_fs.remote_loader.load_manifest(entry_id)
+    local_access_digests = _local_access_digests(workspace_fs, path)
+    remote_access_digests = [access.digest for access in remote_file_manifest.blocks]
+    if local_access_digests != remote_access_digests:
+        await _import_file(workspace_fs, path, absolute_path)
+        print("update %s" % absolute_path)
+        await workspace_fs.sync()
```
touilleMan, line 97: @vxgmichel is there a way to use the transaction to ask for sync of only a certain file/folder ? this would come handy here !

vxgmichel: Yes, `workspace_fs.sync_by_id(entry_id, remote_changed=False, recursive=False)`:
https://github.com/Scille/parsec-cloud/blob/f8f7f9e2dfcd26633772e53c181c05c60b683b06/parsec/core/fs/workspacefs/workspacefs.py#L576-L578

parsec/core/cli/run.py
```diff
@@ -55,3 +57,121 @@ def run_mountpoint(config, device, mountpoint, timestamp, **kwargs):
...
+        await workspace_fs.touch(dest)
+    except FileExistsError:
+        await workspace_fs.truncate(dest, 0)
+    with open(src, "rb") as fd:
+        read_size = 0
+        while True:
+            chunk = fd.read(DEFAULT_BLOCK_SIZE)
+            if not chunk:
+                break
+            await workspace_fs.write_bytes(dest, chunk, read_size)
```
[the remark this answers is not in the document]
vxgmichel: > In fact I'm wondering if using workspace_fs.transactions.file_open is not even better for readability given fd_resize/fd_write are just as readable thantruncate /write_bytes, but we have a more fine control of what is done and the lock on the file from the open might offer interesting guarantees (@vxgmichel any opinion on this ?)

We had this plan of implementing a workspace async file object with an interface like those presented in the `io` module:
https://docs.python.org/3.8/library/io.html

It would be returned by `workspace_fs.open`:
```python
async with workspace_fs.file_open(fspath, "wb") as f:
    await f.write(some_binary_data)
```

parsec/core/cli/run.py
```diff
@@ -55,3 +63,173 @@ def run_mountpoint(config, device, mountpoint, timestamp, **kwargs):
         config = config.evolve(mountpoint_base_dir=Path(mountpoint))
     with cli_exception_handler(config.debug):
         trio_run(_run_mountpoint, config, device, timestamp)
+
+
+async def _import_file(workspace_fs: WorkspaceFS, src: FsPath, dest: FsPath):
```
vxgmichel, line 68: All rsync related functions should probably go to their own module (e.g. `parsec/core/cli/rsync.py`)

parsec/core/cli/run.py
```diff
@@ -55,3 +63,173 @@ def run_mountpoint(config, device, mountpoint, timestamp, **kwargs):
...
+    try:
+        await workspace_fs.touch(dest)
+    except FileExistsError:
+        pass
```
vxgmichel, line 72: Syncing a file of size 0 would fail here, right? I think a call to `await workspace_fs.truncate(dest, 0)` is necessary.

parsec/core/cli/run.py
```diff
@@ -55,3 +63,173 @@ def run_mountpoint(config, device, mountpoint, timestamp, **kwargs):
...
+async def _update_file(
+    workspace_fs: WorkspaceFS, entry_id: EntryID, path: AnyPath, absolute_path: AnyPath
+):
+    remote_file_manifest = await workspace_fs.remote_loader.load_manifest(entry_id)
+    local_access_digests = _local_access_digests(workspace_fs, path)
+    remote_access_digests = [access.digest for access in remote_file_manifest.blocks]
+    if local_access_digests != remote_access_digests:
+        await _import_file(workspace_fs, path, FsPath(str(absolute_path)))
+        print("update %s" % absolute_path)
+        await workspace_fs.sync_by_id(entry_id, remote_changed=False, recursive=False)
```
vxgmichel, line 105: Here, writing could be optimized to only write the blocks with different digests. This would help with the sync as only the necessary blocks would be uploaded.

vxgmichel:
Thanks to @Sithi5 we now have a better workspace file interface (https://github.com/Scille/parsec-cloud/pull/1190)

That should help a bit with this PR :)

parsec/core/cli/rsync.py
```diff
@@ -21,22 +21,20 @@ async def _import_file(workspace_fs: WorkspaceFS, src: FsPath, dest: FsPath):
...
         await workspace_fs.touch(dest)
     except FileExistsError:
-        await workspace_fs.truncate(dest, 0)
-    finally:
-        for offset, chunk in _chunks_from_path(workspace_fs, src):
-            await workspace_fs.write_bytes(dest, chunk, offset)
+        pass
+    f = await workspace_fs.open_file(dest, "w")
+    await f.write(b"".join(_chunks_from_path(workspace_fs, src)))
+    await f.close()
```
vxgmichel, line 27: Two comments:
- There's an async context manager for workspace files to make sure the file is closed
- Joining the chunks for a single write might fill up the RAM, it's better to call `f.write` repeatedely
```python
async with workspace_fs.open_file(dest, "w") as dest_f:
    for chunk in _chunks_from_path(workspace_fs, src):
        await dest_f.write(chunk)
```

PR #1180, opened 2020-06-09
Realm stats

Sithi5:
Stuck in backend part (file memory/realm.py), looking for a way to get all components of realm to get their size.


Sithi5:
Backend is now working, calculating realm size with block (need to add the metadata also)

Sithi5:
metadata added

parsec/api/protocol/realm.py
```diff
@@ -62,6 +63,15 @@ class RealmStatusRepSchema(BaseRepSchema):
...
 realm_status_serializer = CmdSerializer(RealmStatusReqSchema, RealmStatusRepSchema)
 
+class RealmStatsReqSchema(BaseReqSchema):
+    realm_id = fields.UUID(required=True)
+
+
+class RealmStatsRepSchema(BaseRepSchema):
+    data_size = fields.Integer(required=True)
+
+realm_stats_serializer = CmdSerializer(RealmStatsReqSchema, RealmStatsRepSchema)
```
touilleMan, line 73: nitpicking: 2 `\n` between top-level definitions ;-p

black autoformater doesn't enforce this here given we are between a class definition and a variable declaration (see https://black.readthedocs.io/en/stable/the_black_code_style.html#empty-lines), but it's the style we use in most of the code

parsec/backend/memory/block.py
```diff
@@ -15,7 +15,6 @@
     BlockInMaintenanceError,
 )
 
-
```
touilleMan, line 18: Leeeave `\n` alone !
image

parsec/backend/realm.py
```diff
@@ -78,6 +79,8 @@ class RealmStatus:
     def in_maintenance(self) -> bool:
         return bool(self.maintenance_type)
 
+class RealmStats:
```
touilleMan, line 82: ` @attr.s(slots=True, frozen=True, auto_attribs=True) ` is needed

without it, `size` is a class attribute (by opposition of a class instance attribute). This means you cannot do `RealmStats(size=42)`

tests/backend/realm/test_realm_stats.py
```diff
@@ -0,0 +1,39 @@
...
+
+import pytest
+from uuid import uuid4
+
+from tests.backend.common import realm_create, realm_stats
+from tests.backend.common import vlob_create, block_create
+
+
+@pytest.mark.trio
+async def test_realm_stats_ok(backend, alice, alice_backend_sock, administration_backend_sock, realm):
```
touilleMan, line 11: `administration_backend_sock` fixture is not needed for this test
`backend` fixture is indirectly needed by `alice_backend_sock` so no need to declare it here as well

same thing for `alice`

tests/backend/realm/test_realm_stats.py
```diff
@@ -0,0 +1,39 @@
...
+    }
+
+    # Create new data
+    await block_create(alice_backend_sock, realm_id=realm, block_id=uuid4(), block=b"1234")
+    rep = await realm_stats(alice_backend_sock, realm_id=realm)
+    assert rep == {
+        "status": "ok",
+        "data_size": 8,
+    }
+    await block_create(alice_backend_sock, realm_id=realm, block_id=uuid4(), block=b"1234")
```
touilleMan, line 32: No need for the remaining of the test: this is cheap to do when using the memory backend, but thinks are much slower in PostgreSQL, so better not waste resources ;-)

tests/backend/realm/test_realm_stats.py
```diff
@@ -0,0 +1,39 @@
+# Parsec Cloud (https://parsec.cloud) Copyright (c) AGPLv3 2019 Scille SAS
+
+import pytest
+from uuid import uuid4
+
+from tests.backend.common import realm_create, realm_stats
+from tests.backend.common import vlob_create, block_create
+
+
```
touilleMan, line 9: Other interesting tests:
- test stats not allowed for user with no access on the realm
- test stats not allowed for user with similar device_id but on a different organization (see https://github.com/Scille/parsec-cloud/blob/b735aede870925e724360594155ce426e41c2796/tests/backend/realm/test_vlob_access.py#L328-L335)
- test stat on unknown realm_id

parsec/backend/memory/realm.py
```diff
@@ -2,6 +2,7 @@
 
 import attr
 import pendulum
+from sys import getsizeof
```
touilleMan, line 5: Doesn't seems to be used in the code...

parsec/backend/memory/realm.py
```diff
@@ -102,6 +107,20 @@ async def get_status(
             raise RealmAccessError()
         return realm.status
 
+    async def get_stats(
+        self, organization_id: OrganizationID, author: DeviceID, realm_id: UUID
+    ) -> RealmStats:
+        from pprint import pprint
```
touilleMan, line 113: useless import ;-)

parsec/backend/memory/realm.py
```diff
@@ -102,6 +107,20 @@ async def get_status(
...
+        RealmStats.data_size = 0
+        for (_, _), blockmeta_value in self._block_component._blockmetas.items():
+            if blockmeta_value.realm_id == realm_id:
+                RealmStats.data_size += blockmeta_value.size
+        for (_, _), vlobmeta_value in self._vlob_component._vlobs.items():
```
touilleMan, line 118: I think we shouldn't mix metadata & data size: it's much better to provide two separate `metadata_size` & `data_size` (so client can decide if he wants to display those data separately or together)

besides, data size is much bigger than metadata size (~500 time bigger in our production server) so adding metadata size to data size won't change much.

parsec/api/protocol/realm.py
```diff
@@ -62,6 +63,15 @@ class RealmStatusRepSchema(BaseRepSchema):
 
 realm_status_serializer = CmdSerializer(RealmStatusReqSchema, RealmStatusRepSchema)
 
+class RealmStatsReqSchema(BaseReqSchema):
+    realm_id = fields.UUID(required=True)
+
+
+class RealmStatsRepSchema(BaseRepSchema):
+    data_size = fields.Integer(required=True)
```
touilleMan, line 71: would be better to name this `blocks_size` (so we can easily tell how this is calculated, and we can add a `vlobs_size` latter on)

parsec/backend/memory/realm.py
```diff
@@ -102,6 +107,20 @@ async def get_status(
...
+            if vlobmeta_value.realm_id == realm_id:
+                RealmStats.data_size += sum(len(blob) for (blob, _, _) in vlobmeta_value.data)
```
touilleMan, line 120: nice use of the list comprehension :smiley:

.pytest_cache/v/cache/lastfailed
```diff
@@ -0,0 +1,3 @@
+{
+  "tests/backend/realm/test_realm_stats.py::test_realm_stats_ok": true
+}
```
touilleMan, line 3: tu as dû faire un `git add .pytest_cache` par erreur

parsec/backend/postgresql/realm_queries/get.py
```diff
@@ -89,6 +122,38 @@ async def query_get_status(
...
+    blocks = await conn.fetch(
+        *_q_get_blocks_size_from_realm(organization_id=organization_id, realm_id=realm_id)
+    )
+    vlobs = await conn.fetch(
+        *_q_get_vlob_size_from_realm(organization_id=organization_id, realm_id=realm_id)
+    )
+    RealmStats.blocks_size = 0
+    RealmStats.vlobs_size = 0
+
+    for block in blocks:
```
c4ffein, line 148: I helped @Sithi5 write this code, but I didn't realized it was for the sql part... It seems highly inefficient to me, and we should use sql SUM instead, WDYT @touilleMan ?

Sithi5: I updated my sql part to use SUM instead. It seems indeed more appropriate

c4ffein: Seems perfect! You can add more complex UT or make the existing ones more complete if you want, for example creating multiple `vlobs` or `blocks`, but I'll accept the PR as is :)

PR #1181, opened 2020-06-09
PyPI Azure CI

c4ffein:


.azure-pipelines.yml
```diff
@@ -269,6 +269,34 @@ jobs:
...
+  - bash: |
+      set -eux
+      sudo apt install python python-pip
+      pip install twine
+    displayName: 'Bootstrap'
+  - bash: |
+      set -eux
+      python setup.py generate_pyqt sdist
+  - bash: |
+      echo $SNAPCRAFT_CREDENTIALS | twine upload dist/*
```
touilleMan, line 290: Wouldn't it be better to use the create a `~/.pypirc` instead ? (see https://packaging.python.org/specifications/pypirc/#using-a-pypi-token)

.azure-pipelines.yml
```diff
@@ -269,6 +269,34 @@ jobs:
...
+  - checkout: self
+    submodules: true
+  - bash: |
+      set -eux
+      sudo apt install python python-pip
+      pip install twine
+    displayName: 'Bootstrap'
+  - bash: |
+      set -eux
+      python setup.py generate_pyqt sdist
```
touilleMan, line 288: No `bdist_wheel` ?

.azure-pipelines.yml
```diff
@@ -269,6 +269,34 @@ jobs:
...
+  - bash: |
+      echo $SNAPCRAFT_CREDENTIALS | twine upload dist/*
+    displayName: 'Build PyPI'
+    workingDirectory: $(Agent.TempDirectory)
+  - publish: $(Agent.TempDirectory)/dist
```
touilleMan, line 293: What's the use for this ? (if the twine upload failed we stop before being here, if not we have the dist on pypi so it's easy to get them from there)

.azure-pipelines.yml
touilleMan, line 288: `generate_pyqt` shouldn't be needed here (calling `python setup.py bdist_wheel` should depend on it automatically)

However I didn't see such dependency in the code, so if it is not in place I think it's better to have this on multiple lines:
```python
python setup.py sdist
python setup.py generate_pyqt  # Needed before bdist_wheel
python setup.py bdist_wheel
```
Here I considered sdist doesn't need this cooking (I'm not sure it is the case...)

We may want to have a deeper look at the setup.py configuration in a further PR ;-)

.azure-pipelines.yml
```diff
@@ -269,6 +269,34 @@ jobs:
 #################################################################################
 
 
+- job: 'linux_pypi_build'
+  displayName: 'Linux PyPI build'
+  condition: startsWith(variables['build.sourceBranch'], 'refs/tags/')
+  timeoutInMinutes: 60
+  pool:
+    vmImage: 'ubuntu-18.04'
```
touilleMan, line 277: use `ubuntu-latest` (`ubuntu-18.04` is for building snap... because snap :trollface: )

.azure-pipelines.yml
```diff
@@ -269,6 +269,34 @@ jobs:
...
+  steps:
+  - checkout: self
+    submodules: true
+  - bash: |
+      set -eux
+      sudo apt install python python-pip
```
touilleMan, line 283: `sudo apt install twine` to avoid the two steps ?

.azure-pipelines.yml
c4ffein: So we would include the pypirc in our private vars and echo it to the file?

touilleMan: I think it would be less cumbersome to do:
```shell
echo "
[pypi]
username = __token__
password = $SNAPCRAFT_CREDENTIALS
" > ~/.pypirc
```

.azure-pipelines.yml
c4ffein: Corrected

.azure-pipelines.yml
c4ffein: I used the Snap code as a base so I kept it, as we already have the snaps in the artifacts, why not have the wheels too, just in case. But that's true, it is already on pypi, that's kind of useless...

.azure-pipelines.yml
c4ffein: Yup, I wanted to keep that one short, and didn't took time to correct the dependencies in setup.py , but it's true we should absolutely clean that one

I still prefer to generate the widgets before the sdist, and doing everything in one line seems to be ok

.azure-pipelines.yml
c4ffein: Isn't it bad if we generate wheels for a too high Python version? I'm not sure how this generation works under the hood, but aren't we supposed to launch bdist_wheel on a distribution with python 3.6?

touilleMan: Parsec is currently pure-python (you can see this by the `none-any` part in the name of the wheel generated) so this is not a trouble.
Building binary wheel is much more complicated (see manylinux), we will have plenty of fun when we'll need this ;-)

.azure-pipelines.yml
c4ffein: Is there really an advantage to use twine installed through apt instead of pip?

c4ffein: http://manpages.ubuntu.com/manpages/cosmic/man1/twine.1.html
Even the ubuntu documentation recommend to use pip :/

.azure-pipelines.yml
c4ffein: Done

.azure-pipelines.yml
c4ffein: Actually generate_pyqt is useless, widgets are generated nevertheless

.azure-pipelines.yml
c4ffein: Fixed!

.azure-pipelines.yml
```diff
@@ -269,6 +269,53 @@ jobs:
...
+      set -eux
+      pip install --user virtualenv
+      python -m venv venv
+      source venv/bin/activate
+      pip install dist/*.whl[all]
+      xvfb run parsec core gui
+      deactivate
+  - bash: |
+      source venv/bin/activate
+      (sleep 3; killall xvfb-run) & xvfb-run parsec core gui; test $? -eq 143 || exit -1
```
touilleMan, line 299: what's the sleep&killall for ? (we should add a comment about this)

also it would be better to add the `set -eux` at the begining of this script to know what was executed in case of error

.azure-pipelines.yml
```diff
@@ -269,6 +269,53 @@ jobs:
...
+      set -eux
+      python setup.py sdist bdist_wheel
+  - bash: |
+      set -eux
+      pip install --user virtualenv
+      python -m venv venv
+      source venv/bin/activate
+      pip install dist/*.whl[all]
+      xvfb run parsec core gui
+      deactivate
```
touilleMan, line 296: Is this really needed ? (I genuinely don't know ^^)
I would guess using a different shell reset the env variables (hence reset the virtual env config) and the working directory

.azure-pipelines.yml
touilleMan: amen ! ^^

.azure-pipelines.yml
```diff
@@ -269,6 +269,53 @@ jobs:
...
+  - bash: |
+      set -eux
+      python setup.py sdist bdist_wheel
+  - bash: |
+      set -eux
+      pip install --user virtualenv
+      python -m venv venv
+      source venv/bin/activate
+      pip install dist/*.whl[all]
+      xvfb run parsec core gui
```
touilleMan, line 295: Why do we run this here instead than in the next bash entry ? (seems more logical to keep all the test commands in the same part)

.azure-pipelines.yml
c4ffein: Otherwise we would have the GUI staying open forever... We just want to check we can launch it, kill it, and get the correct error code. A better test would simulate clicks to check the window do appear on screen, then quit the app. We should probably also use 

We can't have set -eux as the killing of the running xvfb would appear as an error. But we can have an exit -1 in case we can't obtain the expected error code.

.azure-pipelines.yml
c4ffein: Yup I'll remove those lines!

.azure-pipelines.yml
c4ffein: I used those test commands here because I wanted to test pip install in a -eux shell. I just forgot to remove the xvfb command...

.azure-pipelines.yml
touilleMan: you should be able to do `set -eu` to keep comand info without stopping on crash
Also you should add a comment about why we need this

Also you should divide this line into multiples (I only realized now that it was `(sleep 3; killall xvfb-run) & xvfb-run [...]` and not `(sleep 3; killall xvfb-run) && xvfb-run [...]` xD)

c4ffein: :crying_cat_face:

PR #1188, opened 2020-06-12
Fix trio deadlock when importing a file from a parsec workspace

### LINKED ISSUE #1185: Trio deadlock when importing a file from a parsec workspace
The culprit:
https://github.com/Scille/parsec-cloud/blob/adc66c9b548fc4ed53903f9be0d914bb204da5b7/parsec/core/gui/files_widget.py#L99-L109

The fix: use [trio.open_file](https://trio.readthedocs.io/en/stable/reference-io.html#trio.open_file)

Sithi5:
Fix #1185

### DIFF
--- a/newsfragments/1188.bugfix.rst
+++ b/newsfragments/1188.bugfix.rst
@@ -1,0 +1,1 @@
+Fixed deadlock when importing a file from a parsec workspace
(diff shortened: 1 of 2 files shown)


newsfragments/1188.bugfix.rst
```diff
@@ -0,0 +1 @@
+Fixed deadlock when importing a file from a parsec workspace
```
vxgmichel, line 1: `Fixed` -> `Fix`

PR #1190, opened 2020-06-15
Implement workspace file interface

Sithi5:


parsec/core/fs/workspacefs/workspacefile.py
```diff
@@ -0,0 +1,179 @@
...
+        self._name = None
+        mode.lower()
+        # preventing to open in write and read in same time or write and append or open with no mode
+        if (("w" in mode or "a" in mode) and "r" in mode) or (
+            ("w" in mode and "a" in mode) or re.search("[arw]", mode) is None
+        ):
+            raise ValueError("must have exactly one of create/read/write/append mode")
+        # preventing to open with non-existant mode
+        elif re.search("[^arwb]", mode) is not None:
+            raise ValueError("invalid mode: ", mode)
```
vxgmichel, line 29: It seems like this logic could be simplified, especially the regex part. Why not something like this:
```python
# Read + append is invalid
if "r" in mode and "a" in mode:
    raise ValueError("...")
# Read + write mode is invalid
if "r" in mode and "w" in mode:
    raise ValueError("...")
# Write + append mode is invalid
if "w" in mode and "a" in mode:
    raise ValueError("...")
# No mode selected
if not any(c in mode for c in "rwa"):
    raise ValueError("...")
```
If you find it too verbose, this should also work:
```python
# Exactly one of `r`, `w`, and `a` should be provided
if sum(c in mode for c in "rwa") == 1:
    raise ValueError(...)
```

parsec/core/fs/workspacefs/workspacefile.py
```diff
@@ -0,0 +1,179 @@
...
+            raise FSUnsupportedOperation
+        currentOffset = self._offset
+        # correct size :
+        if size < -1:
+            raise ValueError("read length must be non-negative or -1")
+        # if first read, need to get the fileLenght of the file to move the offset if EOF
+        if self._fileLenght == -1:
+            await self.set_filelenght()
+
+        # reading until EOF : set offset to end (equal to size)
```
vxgmichel, line 74: Please capitalize your comments: https://www.python.org/dev/peps/pep-0008/#comments

parsec/core/fs/workspacefs/workspacefile.py
Sithi5: The regex seems usefull to me because I try to catch all the other characters to raise my error. The sum method for checking that the file have been opened with exactly one correct mode is really usefull though.

parsec/core/fs/workspacefs/workspacefile.py
```diff
@@ -0,0 +1,179 @@
...
+
+    async def read(self, size: int = -1):
+        """
+        Raises:
+            FSUnsupportedOperation
+        """
+        # check if readable
+        if self.readable() is False:
+            raise FSUnsupportedOperation
+        currentOffset = self._offset
```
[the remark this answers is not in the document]
Sithi5: `CurrentOffset` is a tmp var to store the offset because I am modifying it before returning the transactions_read. I will rename it as `current_offset`.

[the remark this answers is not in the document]
Sithi5: `       current_offset = self._offset
        self._offset += await self._transactions.fd_write(self._fd, data, self._offset)
        nbWrittedBytes = self._offset - current_offset
        return nbWrittedBytes`
In write_bytes function I use it to calculate the number of bytes written by the fd_write transaction.
I will rename `nbWrittedBytes` as `nb_writted_bytes`.

parsec/core/fs/workspacefs/workspacefile.py
```diff
@@ -0,0 +1,159 @@
...
+        elif size < 0:
+            raise FSOffsetError
+        elif size > 0:
+            await self._transactions.fd_resize(self._fd, size, truncate_only=False)
+            return size
+
+    def writable(self):
+        return "w" in self._mode or "a" in self._mode
+
+    async def write_bytes(self, data: bytes):
```
vxgmichel, line 142: Isn't this method called `read`? Also you did not document the return type

parsec/core/fs/workspacefs/workspacefile.py
```diff
@@ -0,0 +1,159 @@
...
+        if self._first_write is True and "w" in self._mode:
+            await self._transactions.fd_resize(self._fd, 0, truncate_only=True)
+            self._first_write = False
+        # Preparing offset at EOF if open with append mode
+        elif "a" in self._mode:
+            self._offset = await self.get_filelenght()
+        current_offset = self._offset
+        self._offset += await self._transactions.fd_write(self._fd, data, self._offset)
+        nb_writted_bytes = self._offset - current_offset
+        return nb_writted_bytes
```
vxgmichel, line 159: Again, why not:
```python
result = await self._transactions.fd_write(self._fd, data, self._offset)
self._offset += result
return result
```

parsec/core/gui/files_widget.py
```diff
@@ -98,12 +98,13 @@ async def _do_import(workspace_fs, files, total_size, progress_signal):
             progress_signal.emit(src.name, current_size)
 
             async with await trio.open_file(src, "rb") as f:
+                dstf = await workspace_fs.open_file(dst, "w")
```
vxgmichel, line 101: Maybe use a more meaningful name?

parsec/core/fs/workspacefs/workspacefile.py
```diff
@@ -0,0 +1,159 @@
...
+
+import os
+import re
+from parsec.core.fs.exceptions import FSUnsupportedOperation, FSOffsetError
+
+
+class WorkspaceFile:
+    """methods of this object should act like trio async file :
+    "https://github.com/python-trio/trio/blob/master/trio/_file_io.py"
+    using transactions"""
```
vxgmichel, line 11: Also it would be nice to document how `mode` works here, especially since `b` is ignored as all the methods deal with bytes.

tests/core/fs/workspacefs/test_workspacefile.py
```diff
@@ -0,0 +1,324 @@
...
+    d = tmp_path / "triofoo"
+    d.mkdir()
+    p = d / "triobar"
+    p.write_text("Hello World !")
+    return p
+
+
+@pytest.fixture
+@pytest.mark.trio
+async def random_text():
```
vxgmichel, line 22: This doesn't have to be an async fixture

tests/core/fs/workspacefs/test_workspacefile.py
```diff
@@ -0,0 +1,324 @@
...
+    if (f is None and alice_workspace is None) or (triof is None and trio_file is None):
+        raise ValueError
+    if f is None:
+        f = await alice_workspace.open_file("/foo/bar", "rb")
+    if triof is None:
+        triof = await trio.open_file(trio_file, "rb")
+    output = await f.read(size)
+    output = output.decode("utf-8")
+    trio_output = await triof.read(size)
+    trio_output = trio_output.decode("utf-8")
```
vxgmichel, line 44: I don't think you need to decode the output of both files, comparing the raw output is fine

tests/core/fs/workspacefs/test_workspacefile.py
```diff
@@ -0,0 +1,324 @@
...
+    await triof.seek(500, io.SEEK_END)
+    await f.seek(500, io.SEEK_END)
+    assert f.tell() == await triof.tell() == 510
+
+
+@pytest.mark.trio
+async def test_truncate(alice_workspace, trio_file, random_text):
+    f = await alice_workspace.open_file("/foo/bar", "wb")
+    triof = await trio.open_file(trio_file, "wb")
+    await write_in_both_files(triof, f, "thisis10ch".encode("utf-8"))
```
vxgmichel, line 277: 10 CH? :grin: 
15CH

parsec/core/fs/workspacefs/workspacefile.py
```diff
@@ -0,0 +1,159 @@
...
+        if size < -1:
+            raise ValueError("read length must be non-negative or -1")
+        # Reading until EOF : set offset to end (equal to size)
+        if size == -1:
+            self._offset = await self.get_filelenght()
+            return await self._transactions.fd_read(self._fd, size, current_offset)
+            # Reading size : set offset to current_offset + size
+        else:
+            self._offset += size
+            return await self._transactions.fd_read(self._fd, size, current_offset)
```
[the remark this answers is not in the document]
Sithi5: I actually forgot this case ^^.

parsec/core/fs/workspacefs/workspacefile.py
```diff
@@ -0,0 +1,229 @@
...
+    mode("a") -> append file.
+    mode("b") -> bytes, have to be used with another mode, ignored, class is already working with
+    bytes.
+    mode("w") -> write file.
+    mode("r") -> read file.
+    """
+
+    def __init__(self, fd: int, transactions, path: FsPath, mode: str = "r"):
+        self._fd = fd
+        self._name = None
```
vxgmichel, line 48: Unused?

parsec/core/fs/workspacefs/workspacefile.py
```diff
@@ -0,0 +1,229 @@
...
+        return self
+
+    async def __aexit__(self, *args):
+        self._state = FileState.CLOSED
+        await self.close()
+
+    async def ainit(self):
+        # Calling aenter even without context manager
+        if self._state == FileState.INIT:
+            await self.__aenter__()
```
vxgmichel, line 77: Maybe the other way around?
```python

    async def ainit(self):
        if self._state == FileState.INIT:
            # Checking if the 'w' open mode have been selected, if yes, truncate the file
            if "w" in self._mode:
                await self.truncate(0)
            self._state = FileState.OPEN


    async def __aenter__(self):
        await self.ainit()
        return self
```

parsec/core/fs/workspacefs/workspacefile.py
```diff
@@ -0,0 +1,229 @@
...
+            await self._transactions.fd_close(self._fd)
+
+    @property
+    def closed(self):
+        return self._state == FileState.CLOSED
+
+    async def file_stat(self):
+        """Getting file stat"""
+        stats = await self._transactions.fd_info(self._fd, self._path)
+        return stats
```
vxgmichel, line 95: Maybe make it an internal method (`_stat` or `_file_stat`)? Especially since it's not protected by `check_state`

parsec/core/fs/workspacefs/workspacefile.py
```diff
@@ -0,0 +1,229 @@
...
+
+    async def get_filelength(self):
+        """Getting file length from file stat"""
+        stats = await self.file_stat()
+        return stats["size"]
```
vxgmichel, line 100: Same here (`_get_size` for instance)

parsec/core/fs/workspacefs/workspacefile.py
```diff
@@ -0,0 +1,229 @@
...
+    OPEN = 1
+    CLOSED = 2
+
+
+def check_state(func):
+    @functools.wraps(func)
+    def wrapper(self, *args, **kwargs):
+        if self.closed:
+            raise ValueError("I/O operation on closed file.")
+
```
vxgmichel, line 22: Maybe check for the `INIT` state too:
```python
if self._state == FileState.INIT:
    raise ValueError("I/O operation on a non initialized file")
```

parsec/core/fs/workspacefs/workspacefile.py
Sithi5: It was suppose to be implemented later but finaly the `name` method return the path. I remove this attribute

parsec/core/fs/workspacefs/workspacefile.py
```diff
@@ -0,0 +1,234 @@
...
+    INIT = 0
+    OPEN = 1
+    CLOSED = 2
+
+
+def check_state(func):
+    @functools.wraps(func)
+    def wrapper(self, *args, **kwargs):
+        if self.closed or self._state == FileState.INIT:
+            raise ValueError("I/O operation on closed file.")
```
vxgmichel, line 21: Wrong error message in the case `self._state == FileState.INIT`

parsec/core/gui/files_widget.py
```diff
@@ -98,12 +98,13 @@ async def _do_import(workspace_fs, files, total_size, progress_signal):
...
 
             async with await trio.open_file(src, "rb") as f:
+                dest_file = await workspace_fs.open_file(dst, "w")
                 read_size = 0
                 while True:
                     chunk = await f.read(DEFAULT_BLOCK_SIZE)
                     if not chunk:
                         break
-                    await workspace_fs.write_bytes(dst, chunk, read_size)
+                    await dest_file.write(chunk)
```
vxgmichel, line 107: `dest_file` is never closed here, maybe use the context manager?

parsec/core/fs/workspacefs/workspacefile.py
```diff
@@ -0,0 +1,234 @@
...
+        return self._state == FileState.CLOSED
+
+    @check_state
+    async def file_stat(self):
+        """Getting file stat"""
+        stats = await self._transactions.fd_info(self._fd, self._path)
+        return stats
+
+    @check_state
+    async def get_filelength(self):
```
vxgmichel, line 102: I think `size` is a better name than `length`, maybe `get_size()` or `file_size()`?

PR #1193, opened 2020-06-16
[GUI] Sexier log in screen

### LINKED ISSUE #1130: Update login screen (alternative to combo box)
Currently, the user chooses the device he wants to log into using a combo box.

While it's ugly, it's also useless most of the time since users will rarely have more than one device per machine. In addition, the combo box does not let you see precisely which organization/device you are login into.

To replace it, devices should be displayed as a list (in the same way users are listed). When a device is selected, the prompt for the password appears. This allows us to directly prompt for the password if only one device is found on the machine.

Max-7:
Closes #1130


bitlogik:
Does it fix https://github.com/Scille/parsec-cloud/issues/1169 ?

tests/core/gui/test_workspaces.py
```diff
@@ -30,7 +29,8 @@ async def test_add_workspace(
     monkeypatch.setattr(
         "parsec.core.gui.workspaces_widget.get_text_input", lambda *args, **kwargs: (workspace_name)
     )
-    await aqtbot.mouse_click(w_w.button_add_workspace, QtCore.Qt.LeftButton)
+    async with aqtbot.wait_signal(w_w.list_success, timeout=2000):
+        await aqtbot.mouse_click(w_w.button_add_workspace, QtCore.Qt.LeftButton)
```
vxgmichel, line 33: Why not use `aqtbot.wait_until`? @touilleMan and I noticed it to be more robust than `wait_signal`?

tests/core/gui/test_workspaces.py
```diff
@@ -65,7 +64,8 @@ async def test_rename_workspace(
 
     # Create a workspace and make sure the workspace is displayed
     core = logged_gui.test_get_core()
-    await core.user_fs.workspace_create("Workspace1")
+    async with aqtbot.wait_signal(w_w.list_success, timeout=2000):
+        await core.user_fs.workspace_create("Workspace1")
```
vxgmichel, line 68: Same as https://github.com/Scille/parsec-cloud/pull/1193/files#r468053106

tests/core/gui/test_workspaces.py
Max-7: I forgot that wait_until also as a timeout parameter. It's been fixed.

tests/core/gui/test_workspaces.py
Max-7: Fixed.

Issue #1191, opened 2020-06-15
Revoked account logging exception


AntoineDupre:
I got the following exception from the gui logging from a revoked account:

```
020-06-15 14:34:24 [exception] Unhandled exception in fuse mountpoint [parsec.core.mountpoint.fuse_operations] 
Traceback (most recent call last):
  File "/home/antdup/parsec-cloud/parsec/core/backend_connection/transport.py", line 163, in _do_handshake
    handshake.process_result_req(result_req)
  File "/home/antdup/parsec-cloud/parsec/api/protocol/handshake.py", line 380, in process_result_req
    raise HandshakeRevokedDevice(data["help"])
parsec.api.protocol.handshake.HandshakeRevokedDevice: Device has been revoked

The above exception was the direct cause of the following exception:

Traceback (most recent call last):
  File "/home/antdup/parsec-cloud/parsec/core/fs/remote_loader.py", line 69, in _backend_cmds
    return await getattr(self.backend_cmds, cmd)(*args, **kwargs)
  File "/home/antdup/parsec-cloud/parsec/core/backend_connection/expose_cmds.py", line 32, in wrapper
    async with self.acquire_transport(allow_not_available=True) as transport:
  File "/home/antdup/parsec-cloud/venv/lib/python3.6/site-packages/async_generator/_util.py", line 34, in __aenter__
    return await self._agen.asend(None)
  File "/home/antdup/parsec-cloud/parsec/core/backend_connection/apiv1_authenticated.py", line 246, in _acquire_transport
    raise self._status_exc
  File "/home/antdup/parsec-cloud/parsec/core/backend_connection/apiv1_authenticated.py", line 152, in _run_manager
    await self._manager_connect()
  File "/home/antdup/parsec-cloud/parsec/core/backend_connection/apiv1_authenticated.py", line 189, in _manager_connect
    async with self._acquire_transport(ignore_status=True, force_fresh=True) as transport:
  File "/home/antdup/parsec-cloud/venv/lib/python3.6/site-packages/async_generator/_util.py", line 34, in __aenter__
    return await self._agen.asend(None)
  File "/home/antdup/parsec-cloud/parsec/core/backend_connection/apiv1_authenticated.py", line 249, in _acquire_transport
    async with self._transport_pool.acquire(force_fresh=force_fresh) as transport:
  File "/home/antdup/parsec-cloud/venv/lib/python3.6/site-packages/async_generator/_util.py", line 34, in __aenter__
[...]

PR #1205, opened 2020-06-17
Gui revoked user logging

AntoineDupre:
Widgets that require to fetch information from `get_user_and_devices` can get a `RemoteDevicesManagerError` if the user has been revoked. This error is not handle in `devices_widget`.

This PR try to change how this error is handled. Now, no popup is displayed by the `users_widgets` or the `devices_widget`  if the exception is raised and the devices (or users ) list is filled with an error message. 

There is now a popup saying that the user is revoked when the event bus get the revoked user notification. (This popup is displayed also after logging in).

Also `devices_widget` and `users_widget` are not filled after logging anymore (when the widget is instantiated)  but on display time (only the first time)

linked issue #1191

### DIFF
--- a/parsec/core/gui/central_widget.py
+++ b/parsec/core/gui/central_widget.py
@@ -10,7 +10,9 @@ from parsec.core.gui.users_widget import UsersWidget
 from parsec.core.gui.devices_widget import DevicesWidget
 from parsec.core.gui.menu_widget import MenuWidget
 from parsec.core.gui.lang import translate as _
+from parsec.core.gui.custom_dialogs import show_error
 from parsec.core.gui.ui.central_widget import Ui_CentralWidget
+
 
 from parsec.api.protocol import (
     HandshakeAPIVersionError,
@@ -171,6 +173,8 @@ class CentralWidget(QWidget, Ui_CentralWidget):
                 )
             elif isinstance(cause, HandshakeRevokedDevice):
                 tooltip = _("TEXT_BACKEND_STATE_REVOKED_DEVICE")
+                notif = ("REVOKED", tooltip)
+                self.new_notification.emit(*notif)
             elif isinstance(cause, HandshakeOrganizationExpired):
                 tooltip = _("TEXT_BACKEND_STATE_ORGANIZATION_EXPIRED")
             else:
@@ -190,7 +194,8 @@ class CentralWidget(QWidget, Ui_CentralWidget):
             self.new_notification.emit(*notif)
 
     def on_new_notification(self, notif_type, msg):
-        pass
+        if notif_type == "REVOKED":
+            show_error(self, msg)
 
     def show_mount_widget(self):
         self.clear_widgets()
--- a/parsec/core/gui/devices_widget.py
+++ b/parsec/core/gui/devices_widget.py
@@ -1,7 +1,7 @@
 # Parsec Cloud (https://parsec.cloud) Copyright (c) AGPLv3 2019 Scille SAS
 
 from PyQt5.QtCore import pyqtSignal, Qt, QTimer
-from PyQt5.QtWidgets import QWidget, QMenu, QGraphicsDropShadowEffect
+from PyQt5.QtWidgets import QWidget, QMenu, QGraphicsDropShadowEffect, QLabel
 from PyQt5.QtGui import QColor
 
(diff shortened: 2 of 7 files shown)


parsec/core/gui/devices_widget.py
```diff
@@ -88,7 +93,13 @@ def __init__(self, core, jobs_ctx, event_bus, *args, **kwargs):
         self.list_error.connect(self.on_list_error)
         self.line_edit_search.textChanged.connect(self.filter_timer.start)
         self.filter_timer.timeout.connect(self.on_filter_timer_timeout)
-        self.reset()
+        self.initialized = False
+
+    def show(self):
+        if not self.initialized:
```
Max-7, line 99: I'm guessing that you're checking for initialization to avoid resetting the widget, but does it make that much of a difference? I usually avoid adding logic for performance reason because it has bitten us in the ass in the past.

I would also add that the guy who wrote the original code is a moron and it seems like the device list is never reset. Ideally we would like to reset() on an event (I have no idea if a `device.new` event exists) but resetting when shown would still be better than nothing.

parsec/core/gui/devices_widget.py
```diff
@@ -134,7 +148,13 @@ def on_list_success(self, job):
             )
 
     def on_list_error(self, job):
-        pass
+        status = job.status
+        if status == "error":
+            self._flush_devices_list()
+            self.initialized = False
+            label = QLabel(_("TEXT_USER_LIST_RETRIEVABLE_FAILURE"))
```
Max-7, line 155: Should probably be something with `DEVICE_LIST` instead of `USER_LIST`.

tests/core/gui/test_revoked_user.py
```diff
@@ -0,0 +1,64 @@
...
+async def test_revoked_notification(
+    aqtbot, running_backend, autoclose_dialog, logged_gui, alice_user_fs, bob
+):
+    now = pendulum.now()
+    revoked_device_certificate = RevokedUserCertificateContent(
+        author=alice_user_fs.device.device_id, timestamp=now, user_id=bob.user_id
+    ).dump_and_sign(alice_user_fs.device.signing_key)
+    central_widget = logged_gui.test_get_central_widget()
+    assert central_widget is not None
+    async with aqtbot.wait_signals([central_widget.new_notification]):
```
Max-7, line 47: Should be `wait_signal` since there's only one.

parsec/core/gui/devices_widget.py
AntoineDupre: The original idea there is to avoid getting errors when you just log  because the widget can not be filled (Like the exception reported in the issue #1191 ).In fact now the central widget handles this event errors so it probably does not matter to call `reset` at the end of `__init__` .

The event approach you suggest songs to be the right thing to do.  With my initialized flag, If there is no error at the first display, the list is not going to be refilled on the second one. An easy fix is to remove this initialized flag and always requests devices on display.

About events, is it the `central_widget` that is in charge to listen to `device.new` event (or `user.new` in the case of a user event and the `user_widget`) and then call `reset` on the `devices_widget`, or the `devices_widget` has to listen itself the events_bus ? (For example, i think to call reset from the central widget if the revoked user event happened to clear the local cache)

parsec/core/gui/central_widget.py
```diff
@@ -191,7 +195,8 @@ def _on_connection_state_changed(self, status, status_exc):
             self.new_notification.emit(*notif)
 
     def on_new_notification(self, notif_type, msg):
-        pass
+        if notif_type == "REVOKED":
+            show_error(None, msg)
```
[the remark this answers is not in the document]
AntoineDupre: no reason, i changed it

parsec/core/gui/devices_widget.py
AntoineDupre: True, i added this new text entry:

```
TEXT_DEVICE_LIST_RETRIEVABLE_FAILURE
Failed to retrieve the device list from the server.
Impossible de récupérer la liste des appareils depuis le serveur.
```

tests/core/gui/test_revoked_user.py
AntoineDupre: fixed

AntoineDupre:
@Max-7 Is it ok to merged

PR #1208, opened 2020-06-17
Silence test_file_operations flaky error

touilleMan:
related to #1207... though doesn't really fix it :(

PR #1210, opened 2020-06-18
Disable reject_irp_prior_to_transact0 option for WinFSP

### LINKED ISSUE #1207: Inconsistent test: test_file_operations with WinFSP
https://dev.azure.com/Scille/parsec/_build/results?buildId=1098&view=logs&j=bfccef3f-45d4-504d-9a7f-a8a74922e6bf&t=0730e104-ab2a-596a-34ce-ce16eed3b088&l=47091

```python
    class FileOperationsStateMachine(RuleBasedStateMachine):
        @initialize()
        def init(self):
            [...]
            self.mountpoint_service = mountpoint_service_factory(_bootstrap)
            [...]
            self.oracle_fd = os.open(self.oracle_file_path, os.O_RDWR | os.O_CREAT)
            self.fd = os.open(self.file_path, os.O_RDWR | os.O_CREAT)  # Error here !
```
The error seems to come from the fact the mountpoint return from init while not fully initialized (hence failing to create a file right away).
Adding a `time.sleep(0.01)` before the os.open make the error disappear.

touilleMan:
fix #1207 the proper way (not like #1208 :trollface: )

### DIFF
--- a/newsfragments/1210.bugfix.rst
+++ b/newsfragments/1210.bugfix.rst
@@ -1,0 +1,1 @@
+Fix error on Windows when using the mountpoint right after (<0.01s) it has been mounted.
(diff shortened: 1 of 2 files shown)


touilleMan:
@vxgmichel this is also done for fuse so I'm not sure... but let's try this with e4ed5c46 ! ^^

PR #1215, opened 2020-06-19
Restrict read access for parsec directories to the current user

### LINKED ISSUE #940: Default rights on mountpoint
Currently the mountpoint has 755 rights on it, which allows a user on the same machine to see files contained in the mountpoint if Parsec is started.

vxgmichel:
Fix #940.

### DIFF
--- a/parsec/core/config.py
+++ b/parsec/core/config.py
@@ -137,10 +137,13 @@ def config_factory(
     )
 
     # Make sure the directories exist on the system
-    core_config.config_dir.mkdir(parents=True, exist_ok=True)
-    core_config.data_base_dir.mkdir(parents=True, exist_ok=True)
-    core_config.cache_base_dir.mkdir(parents=True, exist_ok=True)
-    core_config.mountpoint_base_dir.mkdir(parents=True, exist_ok=True)
+    core_config.config_dir.mkdir(mode=0o700, parents=True, exist_ok=True)
+    core_config.data_base_dir.mkdir(mode=0o700, parents=True, exist_ok=True)
+    core_config.cache_base_dir.mkdir(mode=0o700, parents=True, exist_ok=True)
+
+    # Mountpoint base directory is not used on windows
+    if os.name != "nt":
+        core_config.mountpoint_base_dir.mkdir(mode=0o700, parents=True, exist_ok=True)
 
     return core_config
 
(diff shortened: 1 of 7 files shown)


parsec/core/config.py
```diff
@@ -137,10 +137,13 @@ def config_factory(
     )
 
     # Make sure the directories exist on the system
-    core_config.config_dir.mkdir(parents=True, exist_ok=True)
-    core_config.data_base_dir.mkdir(parents=True, exist_ok=True)
-    core_config.cache_base_dir.mkdir(parents=True, exist_ok=True)
-    core_config.mountpoint_base_dir.mkdir(parents=True, exist_ok=True)
+    core_config.config_dir.mkdir(mode=0o700, parents=True, exist_ok=True)
```
touilleMan, line 140: What is the behavior on windows ? (python documentation doesn't talk about: https://docs.python.org/3.7/library/pathlib.html#pathlib.Path.mkdir)

From my experience it doesn't seem to do anything:
```
$ python
Python 3.7.7 (tags/v3.7.7:d7c567b08f, Mar 10 2020, 10:41:24) [MSC v.1900 64 bit (AMD64)] on win32                                                                                   
Type "help", "copyright", "credits" or "license" for more information.                                                                                                              
>>> import pathlib
>>> pathlib.Path('foo').mkdir(mode=0o700)
>>> pathlib.Path('foo').stat
<bound method Path.stat of WindowsPath('foo')>
>>> pathlib.Path('foo2').stat()
os.stat_result(st_mode=16895, st_ino=6473924464938213, st_dev=4168444147, st_nlink=1, st_uid=0, st_gid=0, st_size=0, st_atime=1592813875, st_mtime=1592813875, st_ctime=1592813875)
>>> quit()
$ ls -lh
[...]
drwxr-xr-x 1 gbleu 197121    0 Jun 22 10:17  foo/
[...]
```

vxgmichel: It doesn't:

> Although Windows supports chmod(), you can only set the file’s read-only flag with it (via the stat.S_IWRITE and stat.S_IREAD constants or a corresponding integer value). All other bits are ignored.

https://docs.python.org/3.8/library/os.html#os.chmod

vxgmichel:
> I'm still a bit worried by the mode=0o700 in every mkdir, it feels fragile and easy to forget during new feature development...

Yes but I guess it's not that big of a deal either, especially if the parent directories (namely config, data and cache) are already `0o700`. Plus those directories should not contain sensitive information, I changed it because I thought it's good practice.

> A solution could be do to monkeypatching or create a custom Path class inheriting from pathlib.Path to overwrite the mkdir method, but it seems like nailing with a howitzer

Yes I don't think that's necessary. Maybe a `safe_mkdir` helper in `parsec/utils`?

PR #1223, opened 2020-06-23
Update WinFSP

bitlogik:
The WinFSP package version is outdated and is not a "stable" version. This change now uses the latest (as of today) stable release. This fixes many things, so potentially preventing a lot of future issue using WinFSP.


bitlogik:
> 3 days after the winfsp release, that was fast :)

Well actually, that was a coincidence, we just looked for new versions of WinFSP to eventually update it, and they had just released a new stable version. 👍

PR #1226, opened 2020-06-24
Events as enums

AntoineDupre:
### About
This PR change events on the event_bus from string to enum. Except in the enum definition, an event shall not be refereed as a string.

There is 3 different event definitions:
 - `CoreEvents` [core.core_events](https://github.com/Scille/parsec-cloud/blob/events_as_enums/parsec/core/core_events.py#L6) : events used only in the core
 - `BackendEvents` [backend.backend_events](https://github.com/Scille/parsec-cloud/blob/events_as_enums/parsec/backend/backend_events.py#L8) : events used only in the backend. They are converted to string in the [postgresql.handler](https://github.com/Scille/parsec-cloud/blob/events_as_enums/parsec/backend/postgresql/handler.py#L211) (and the other way arround [there](https://github.com/Scille/parsec-cloud/blob/events_as_enums/parsec/backend/postgresql/handler.py#L190) ).
 - `Events` (looking for a better name) [api.protocol.events](https://github.com/Scille/parsec-cloud/blob/events_as_enums/parsec/api/protocol/events.py#L11)  : Events send between the backend and the clients. They are (like BackendEvent) converted to string in the `postgresql.handler`. Those events are also serialized/deserialized when they are sent to the websocket with [EnumCheckedConstant](https://github.com/Scille/parsec-cloud/blob/events_as_enums/parsec/serde/fields.py#L185)

### Main changes

A lot of files in this PR are just modified due to the renaming. Here i am pointing few of them that have important changes:

#### SpiedEvents
[SpiedEvents](https://github.com/Scille/parsec-cloud/blob/events_as_enums/tests/event_bus_spy.py#L147) can now be built from an Enum. They can still be built with a string, maybe this can be removed.

#### QtSignals
Some qt signals have been changed to use `CoreEvents`   instead of strings:
```python
class FilesWidget(QWidget, Ui_FilesWidget):
    fs_updated_qt = pyqtSignal(CoreEvents, UUID)
    fs_synced_qt = pyqtSignal(CoreEvents, UUID)
```

#### Schema Serialization/deserialization
[EnumCheckedConstant](https://github.com/Scille/parsec-cloud/blob/events_as_enums/parsec/serde/fields.py#L185) can now be `_serialized`
[Here](https://github.com/Scille/parsec-cloud/blob/events_as_enums/parsec/api/protocol/events.py#L67)  are the new schema for the EventsTypes
[type_field](https://github.com/Scille/parsec-cloud/blob/events_as_enums/parsec/serde/schema.py#L147) is now not overwritten if the field already exist in the data (otherwise it can be reset with the not serialized object)

#### Postgresql handler
[Here](https://github.com/Scille/parsec-cloud/blob/events_as_enums/parsec/backend/postgresql/handler.py#L211) is the serialization
[Here](https://github.com/Scille/parsec-cloud/blob/events_as_enums/parsec/backend/postgresql/handler.py#L188)  is the deserialization.


newsfragments/1226.misc.rst
```diff
@@ -0,0 +1,2 @@
+Change events on the event_bus from string to python enum
```
touilleMan, line 1: No need to talk about this change in the release changelog given it doesn't affect the user (so an empty `1226.empty.rst` would be enough)

parsec/backend/app.py
```diff
@@ -246,7 +248,7 @@ async def _handle_client_loop(self, transport, client_ctx):
                     continue
 
             if get_log_level() <= LOG_LEVEL_DEBUG:
-                client_ctx.logger.debug("Response", rep=_filter_binary_fields(req))
+                client_ctx.logger.debug("Response", rep=_filter_binary_fields(rep))
```
touilleMan, line 251: Good catch !

parsec/backend/backend_events.py
```diff
@@ -0,0 +1,18 @@
+# Parsec Cloud (https://parsec.cloud) Copyright (c) AGPLv3 2019 Scille SAS
+
+from enum import Enum
+
+__all__ = "BackendEvents"
```
touilleMan, line 5: `__all__` should be a list or a tuple (beside I'm not sure it is needed here, from my point of view `__all__` is only useful in `__init__.py` to re-expose imports)

parsec/api/protocol/events.py
```diff
@@ -1,60 +1,61 @@
...
-EVENTS = (
-    "pinged",
-    "realm.roles_updated",
-    "realm.vlobs_updated",
-    "realm.maintenance_started",
-    "realm.maintenance_finished",
-    "message.received",
-    "invite.status_changed",
-)
+class EventsType(Enum):
```
touilleMan, line 11: Shouldn't we call this just `Events` ? (especially given we have `BackendEvents`)

AntoineDupre: `Events` seems better

newsfragments/1226.misc.rst
AntoineDupre: fixed

parsec/backend/backend_events.py
AntoineDupre: that's a mistake, thanks

parsec/api/protocol/events.py
AntoineDupre: fixed

parsec/backend/invite.py
```diff
@@ -150,7 +152,7 @@ def _on_status_changed(event, organization_id, greeter, token, status):
             else:  # Invitation deleted or back to idle
                 self._claimers_ready[organization_id].discard(token)
 
-        self._event_bus.connect("invite.status_changed", _on_status_changed)
+        self._event_bus.connect(Events.invite_status_changed, _on_status_changed)
```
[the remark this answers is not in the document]
AntoineDupre: Done, now `BackendEvent` mirror `Event`, there is not anymore `Event` sent to the backend event bus. The conversion is done on [backend.events](https://github.com/Scille/parsec-cloud/blob/events_as_enums/parsec/backend/events.py#L93)

parsec/api/protocol/events.py
```diff
@@ -1,60 +1,61 @@
...
-EVENTS = (
-    "pinged",
-    "realm.roles_updated",
-    "realm.vlobs_updated",
-    "realm.maintenance_started",
-    "realm.maintenance_finished",
-    "message.received",
-    "invite.status_changed",
-)
+class Events(Enum):
```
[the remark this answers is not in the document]
AntoineDupre: renamed

parsec/backend/memory/invite.py
```diff
@@ -226,7 +235,7 @@ async def _new(
             org.invitations[invitation.token] = invitation
 
         await self._send_event(
-            "invite.status_changed",
+            Events.invite_status_changed,
```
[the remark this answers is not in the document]
AntoineDupre: fixed

parsec/serde/schema.py
```diff
@@ -99,17 +100,27 @@ def get_obj_type(self, obj):
 
     def _get_schema(self, type):
         if self._instantiated_schemas is None:
+            # Assert all schemas key are Enum:
+            if not all([isinstance(k, Enum) for k in self.type_schemas]):
```
touilleMan, line 104: I always do `in self.types_schemas.keys()` (feels more explicit) but maybe it's kind of pointless...

touilleMan: However I would do an even stronger check here:
```python
enum_types = {type(k) for k in self.type_schemas.keys()}
if len(enum_types) != 1 or not issubclass(enum_types[0], Enum):
    raise ValueError …
```
This way we also enforce the keys are all from the same enum ;-)

parsec/serde/schema.py
```diff
@@ -99,17 +100,27 @@ def get_obj_type(self, obj):
...
+                raise ValueError("type_schemas key can only be an Enum")
             self._instantiated_schemas = {
                 k: v if isinstance(v, Schema) else v() for k, v in self.type_schemas.items()
             }
+
             if self.fallback_type_schema and not isinstance(self.fallback_type_schema, Schema):
                 self._instantiated_fallback_schema = self.fallback_type_schema()
 
+        if not isinstance(type, Enum):
```
touilleMan, line 113: If type can be Enum or str, we should add typing to the method signature

parsec/serde/schema.py
```diff
@@ -99,17 +100,27 @@ def get_obj_type(self, obj):
...
+            for k in self._instantiated_schemas:
```
touilleMan, line 114: would be simpler to do this when building `_instantiated_schemas`:
```python
self._instantiated_schemas: Dict[Union[Enum, str], Schema]  # Important to type this attribute given it key pattern is exotic
for k, v in self.type_schemas.item():
    schema_instance = v if isinstance(v, Schema) else v()
    self._instantiated_schemas[k] = schema_instance
    self._instantiated_schemas[k.value] = schema_instance
```

parsec/serde/schema.py
```diff
@@ -144,7 +155,7 @@ def _dump(self, obj, update_fields=True, **kwargs):
             return MarshalResult(None, {"_schema": "Unsupported object type: %s" % obj_type})
 
         result = schema.dump(obj, many=False, update_fields=update_fields, **kwargs)
-        if result.data:
+        if result.data and self.type_field not in result.data:
```
touilleMan, line 158: Didn't we agree to remove this part altogether ? (considering it's the schema responsibility to do the serialization, so doing `result.data[self.type_field] = obj_type` after it is a bad idea)

We should instead have a `assert self.type_field in result.data` check (another equivalent safety check would be to ensure all the schema defined in `self._instantiated_schemas` define a `CheckedEnumConstant` field named after `self.type_field`, this is a bit harder to implement but s0 much Rox0r for sP33D !)

tests/backend/user/test_apiv1_user_invite.py
```diff
@@ -67,7 +68,9 @@ async def test_user_invite_timeout(mock_clock, backend, apiv1_alice_backend_sock
     with backend.event_bus.listen() as spy:
         async with user_invite(apiv1_alice_backend_sock, user_id=mallory.user_id) as prep:
 
-            await spy.wait_with_timeout("event.connected", {"event_name": "user.claimed"})
+            await spy.wait_with_timeout(
+                "event.connected", {"event_name": BackendEvent.user_claimed}
```
touilleMan, line 72: We should also have a `EventBusMetaEvent` Enum (or a better/shorter name 😄 ) with EVENT_CONNECTED/EVENT_DISCONNECTED.
This would allow to add proper typing to the `EventBus` class and remove this dirty check:
https://github.com/Scille/parsec-cloud/blob/335bdb699684f593454320ba53d63f7501ea95c0/parsec/event_bus.py#L47-L49

touilleMan: also we could rename `event_name` param into `event_type` (calling it `name` make is feel like it's a string...)

parsec/api/data/message.py
```diff
@@ -6,6 +6,14 @@
 from parsec.serde import fields, post_load, OneOfSchema
 from parsec.api.data.entry import EntryID, EntryIDField
 from parsec.api.data.base import BaseAPISignedData, BaseSignedDataSchema
+from enum import Enum
+
+
+class MessageContentType(Enum):
```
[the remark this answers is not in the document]
AntoineDupre: Yes, that's the idea

parsec/backend/postgresql/handler.py
```diff
@@ -177,10 +182,13 @@ def _on_notification(self, connection, pid, channel, payload):
         data.pop("__id__")  # Simply discard the notification id
         signal = data.pop("__signal__")
         logger.debug("notif received", pid=pid, channel=channel, payload=payload)
+        # Convert strings to enums
+        if signal in (evt.value for evt in BackendEvent):
```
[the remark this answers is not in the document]
AntoineDupre: Yes it's better, here we are suppose to get only `BackendEvent` as signal

tests/core/test_messages_monitor.py
```diff
@@ -68,12 +70,12 @@ async def test_process_while_offline(
...
         with trio.fail_after(60):  # autojump, so not *really* 60s
             await spy.wait(
-                "backend.connection.changed",
+                CoreEvent.backend_connection_changed,
                 {"status": BackendConnStatus.READY, "status_exc": None},
             )
             await alice_core.wait_idle_monitors()
         assert alice_core.backend_status == BackendConnStatus.READY
-        spy.assert_event_occured("pinged", {"ping": "hello from Bob !"})
+        spy.assert_event_occured(Event.pinged, {"ping": "hello from Bob !"})
```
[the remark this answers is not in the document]
AntoineDupre: fixed

parsec/core/backend_connection/apiv1_authenticated.py
```diff
@@ -9,14 +9,15 @@
...
 from parsec.api.data import EntryID
-from parsec.api.protocol import DeviceID
+from parsec.api.protocol import DeviceID, Event
 from parsec.core.types import BackendOrganizationAddr
 from parsec.core.backend_connection import cmds
 from parsec.core.backend_connection.transport import apiv1_connect, TransportPool
 from parsec.core.backend_connection.exceptions import BackendNotAvailable, BackendConnectionRefused
 from parsec.core.backend_connection.expose_cmds import expose_cmds_with_retrier
 from parsec.core.backend_connection.authenticated import BackendConnStatus
 from parsec.api.protocol import APIV1_AUTHENTICATED_CMDS
```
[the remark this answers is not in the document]
AntoineDupre: fixed

parsec/core/gui/files_widget.py
```diff
@@ -242,9 +243,9 @@ def __init__(self, core, jobs_ctx, event_bus, *args, **kwargs):
         self.loading_dialog = None
         self.import_progress.connect(self._on_import_progress)
 
-        self.event_bus.connect("fs.entry.updated", self._on_fs_entry_updated_trio)
-        self.event_bus.connect("fs.entry.synced", self._on_fs_entry_synced_trio)
-        self.event_bus.connect("sharing.updated", self._on_sharing_updated_trio)
+        self.event_bus.connect(CoreEvent.fs_entry_updated, self._on_fs_entry_updated_trio)
+        self.event_bus.connect(CoreEvent.fs_entry_synced, self._on_fs_entry_synced_trio)
+        self.event_bus.connect(CoreEvent.sharing_updated, self._on_sharing_updated_trio)
         self.event_bus.connect("fs.entry.downsynced", self._on_entry_downsynced_trio)
```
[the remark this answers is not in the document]
AntoineDupre: yes it is ;), fixed

parsec/serde/schema.py
AntoineDupre: that's better, fixed

parsec/serde/fields.py
```diff
@@ -189,6 +190,9 @@ def __init__(self, constant: Enum, **kwargs):
         super().__init__(**kwargs)
         self.constant = constant
 
+    def _serialize(self, value, attr, obj):
```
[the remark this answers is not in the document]
AntoineDupre: i added it:

```python
    def _serialize(self, value, attr, obj):
        return self.constant
```

parsec/core/gui/main_window.py
```diff
@@ -450,7 +451,7 @@ def on_tab_notification(self, widget, event):
         idx = self.tab_center.indexOf(widget)
         if idx == -1 or idx == self.tab_center.currentIndex():
             return
-        if event in ["sharing.updated"]:
+        if event in [CoreEvent.sharing_updated]:
```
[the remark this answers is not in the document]
AntoineDupre: fixed

tests/event_bus_spy.py
```diff
@@ -143,6 +144,8 @@ def _cook_event_params(self, event):
         elif isinstance(event, tuple):
             event = event + (ANY,) * (3 - len(event))
             return SpiedEvent(*event)
+        elif isinstance(event, Enum):
```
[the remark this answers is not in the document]
AntoineDupre: fixed

parsec/core/gui/central_widget.py
```diff
@@ -26,11 +27,11 @@
...
-        "mountpoint.stopped",
-        "mountpoint.remote_error",
-        "mountpoint.unhandled_error",
-        "sharing.updated",
+        CoreEvent.backend_connection_changed,
+        CoreEvent.mountpoint_stopped,
+        CoreEvent.mountpoint_remote_error,
+        CoreEvent.mountpoint_unhandled_error,
+        CoreEvent.sharing_updated,
         "fs.entry.file_update_conflicted",
```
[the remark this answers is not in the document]
AntoineDupre: i don't like it ! 
fixed

parsec/backend/postgresql/handler.py
```diff
@@ -196,7 +204,8 @@ async def send_signal(conn, signal, **kwargs):
     # Add UUID to ensure the payload is unique given it seems Postgresql can
     # drop duplicated NOTIFY (same channel/payload)
     # see: https://github.com/Scille/parsec-cloud/issues/199
-    raw_data = b64encode(packb({"__id__": uuid4().hex, "__signal__": signal, **kwargs})).decode(
+    raw_signal = signal.value if isinstance(signal, Enum) else signal
```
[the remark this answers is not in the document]
AntoineDupre: fixed

tests/core/test_cross_cores_sync.py
```diff
@@ -29,13 +30,13 @@ def _on_entry_synced(sender, id, path):
         if synced == to_sync:
             event.set()
 
-    core.signal_ns.connect("fs.entry.synced", _on_entry_synced)
+    core.signal_ns.connect(CoreEvent.fs_entry_synced, _on_entry_synced)
```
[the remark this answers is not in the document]
AntoineDupre: It does not exist in the `core`, it is used only in `test_cross_cores_sync.py`. There all the tests are skipped ("Recursive sync strategy need to be reworked"). I guess we need to clean this at some point.

parsec/core/fs/userfs/userfs.py
```diff
@@ -756,7 +761,7 @@ async def _process_message(
             await self._process_message_sharing_revoked(msg)
 
         elif isinstance(msg, PingMessageContent):
-            self.event_bus.send("pinged", ping=msg.ping)
+            self.event_bus.send(Event.pinged, ping=msg.ping)
```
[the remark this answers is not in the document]
AntoineDupre: i renamed it to `message.pinged`

tests/backend/user/test_apiv1_user_invite.py
AntoineDupre: I added a new Enum in `parsec.event_bus`:
```python
class MetaEvent(Enum):
    EVENT_CONNECTED = "event.connected"
    EVENT_DISCONNECTED = "event.disconnected"
```

`MetaEvent` is a name like `CoreEvent` 


I also renamed `event_name` param into `event_type`

parsec/serde/schema.py
AntoineDupre: Now schema are inspected when `_instanciated_schemas` is created:
```python
        for k, v in self.type_schemas.items():
            schema_instance = v if isinstance(v, Schema) else v()
            if self.type_field not in schema_instance.fields:
                raise ValueError(f"{schema_instance} needs to define '{self.type_field}' field")
```

parsec/api/protocol/events.py
```diff
@@ -1,60 +1,61 @@
...
-EVENTS = (
-    "pinged",
-    "realm.roles_updated",
-    "realm.vlobs_updated",
-    "realm.maintenance_started",
-    "realm.maintenance_finished",
-    "message.received",
-    "invite.status_changed",
-)
+class Event(Enum):
```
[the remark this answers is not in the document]
AntoineDupre: `Event` renamed to `APIEvent`

I also put in uppercase all the `Enum` attributes

parsec/backend/backend_events.py
```diff
@@ -0,0 +1,26 @@
...
+
+from enum import Enum
+
+__all__ = ("BackendEvent",)
+
+
+class BackendEvent(Enum):
+    """ Backend internal events"""
+
+    device_claimed = "device.claimed"
```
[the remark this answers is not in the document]
AntoineDupre: It is now in uppercase

parsec/core/core_events.py
```diff
@@ -0,0 +1,30 @@
+# Parsec Cloud (https://parsec.cloud) Copyright (c) AGPLv3 2019 Scille SAS
+
+from enum import Enum
+
+
+class CoreEvent(Enum):
+    # Backend
+    backend_connection_changed = "backend.connection.changed"
```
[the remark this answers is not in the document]
AntoineDupre: It is now in uppercase

parsec/serde/schema.py
AntoineDupre: fixed

parsec/serde/schema.py
AntoineDupre: done

parsec/serde/schema.py
```diff
@@ -90,26 +92,35 @@ def get_obj_type(self, obj):
...
-    type_field_remove = True
-    type_schemas = {}
-    fallback_type_schema = None
-    _instantiated_schemas = None
-    _instantiated_fallback_schema = None
+    type_field: str = "type"
+    type_schemas: Dict[Enum, Schema] = {}
+    fallback_type_schema: Schema = None
+    _instantiated_schemas: Dict[Union[Enum, str], Schema] = None
+    _instantiated_fallback_schema: Dict[Union[Enum, str], Schema] = None
```
touilleMan, line 99: bad typing: fallback is a single schema (so should be `_instantiated_fallback_schema: Schema`)

parsec/serde/schema.py
```diff
@@ -201,3 +208,19 @@ def validate(self, data, many=None, partial=None):
...
+    _instantiated_schemas = None
+    _instantiated_fallback_schema = None
+
+    def _get_schema(self, type):
+        if self._instantiated_schemas is None:
+            self._instantiated_schemas = {
+                k: v if isinstance(v, Schema) else v() for k, v in self.type_schemas.items()
+            }
+
+            if self.fallback_type_schema and not isinstance(self.fallback_type_schema, Schema):
```
touilleMan, line 223: If we have `self.fallback_type_schema = Myschema()`, then we don't go in this if, hence `self._instantiated_fallback_schema` is`None`...

parsec/serde/schema.py
```diff
@@ -90,26 +92,35 @@ def get_obj_type(self, obj):
...
+                raise ValueError(f"{schema_instance} needs to define '{self.type_field}' field")
+            self._instantiated_schemas[k] = schema_instance
+            self._instantiated_schemas[k.value] = schema_instance
 
-    def _get_schema(self, type):
-        if self._instantiated_schemas is None:
-            self._instantiated_schemas = {
-                k: v if isinstance(v, Schema) else v() for k, v in self.type_schemas.items()
-            }
             if self.fallback_type_schema and not isinstance(self.fallback_type_schema, Schema):
```
touilleMan, line 113: Shouldn't we put this out of the for loop ? (`self._instantiated_fallback_schema` should be instantiated once)
and if we have `self.fallback_type_schema = Myschema()`, then we don't go in this if, hence `self._instantiated_fallback_schema` is `None`...

parsec/core/fs/userfs/userfs.py
```diff
@@ -756,7 +761,7 @@ async def _process_message(
             await self._process_message_sharing_revoked(msg)
 
         elif isinstance(msg, PingMessageContent):
-            self.event_bus.send("pinged", ping=msg.ping)
+            self.event_bus.send(CoreEvent.PINGED, ping=msg.ping)
```
touilleMan, line 764: https://github.com/Scille/parsec-cloud/pull/1226#discussion_r446528076 ?

AntoineDupre: renamed (i thought we were speaking about the event value string `"message.pinged"` and not the event name. )

parsec/event_bus.py
```diff
@@ -45,16 +50,16 @@ def connection_context(self):
 
     def send(self, event: str, **kwargs):
```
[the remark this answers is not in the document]
AntoineDupre: fixed

parsec/serde/schema.py
AntoineDupre: fixed

parsec/serde/schema.py
AntoineDupre: I removed `and not isinstance(self.fallback_type_schema, Schema)`, do you think we shall have some extra assertion on the type ?

parsec/serde/schema.py
AntoineDupre: fixed

parsec/event_bus.py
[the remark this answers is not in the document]
touilleMan: In fact there is also typing error for the other methods of this class...

parsec/serde/schema.py
touilleMan: I think it's fine to consider `self.fallback_type_schema` must be a `Schema`, and in such case we don't use assert to enforce this, considering it's a static typing job to make this check

besides, if we pass an instance of schema instead of a schema class, we will have an exception anyway when trying to `instantiate` the instance (well except if the instance is callable, but then the user has really really fucked up ^^)

PR #1228, opened 2020-06-24
Mark flaky windows mountpoint tests as flaky using pytest-rerunfailures

vxgmichel:
Use the flaky mark from pytest-rerunfailures to deal with flaky tests in a more flexible way.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -17,8 +17,6 @@ repos:
     rev: v2.2.3  # Use the ref you want to point at
     hooks:
     - id: mixed-line-ending
-      args:
-        - "--fix=lf"
     - id: trailing-whitespace
 -   repo: local
     hooks:
--- a/setup.cfg
+++ b/setup.cfg
@@ -52,3 +52,14 @@ filterwarnings =
      default::DeprecationWarning:tests\.(.*)|parsec\.(.*)
      # Ignore our todo warnings
      ignore:TODO:UserWarning:parsec|tests
+markers =
+    linux: marks tests as linux only
+    win32: marks tests as Windows only
+    gui: marks tests as GUI (select with '--rungui')
+    slow: marks tests as slow (select with '--runslow')
+    mountpoint: marks tests as mountpoint (select with '--runmountpoint')
+    postgresql
+    raid0_blockstore
+    raid1_blockstore
+    raid5_blockstore
+    backend_not_populated
--- a/setup.cfg
+++ b/setup.cfg
@@ -55,10 +55,10 @@ filterwarnings =
 markers =
     linux: marks tests as linux only
     win32: marks tests as Windows only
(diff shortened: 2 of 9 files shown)


setup.cfg
```diff
@@ -52,3 +52,14 @@ filterwarnings =
      default::DeprecationWarning:tests\.(.*)|parsec\.(.*)
      # Ignore our todo warnings
      ignore:TODO:UserWarning:parsec|tests
+markers =
+    linux: marks tests as linux only
+    win32: marks tests as Windows only
+    gui: marks tests as GUI (select with '--rungui')
+    slow: marks tests as slow (select with '--runslow')
+    mountpoint: marks tests as mountpoint (select with '--runmountpoint')
+    postgresql
```
touilleMan, line 61: "Select with `--postgresql`" ?

.pre-commit-config.yaml
```diff
@@ -17,8 +17,6 @@ repos:
     rev: v2.2.3  # Use the ref you want to point at
     hooks:
     - id: mixed-line-ending
-      args:
```
touilleMan, line 20: Not needed anymore ?

setup.cfg
vxgmichel: So the `@pytest.mark.postgresql` tests are skipped when `--postgresql` is not provided?

.pre-commit-config.yaml
vxgmichel: This kept messing with my commits on windows, and I don't understand the point of it. Simply using `mixed-line-ending` should be enough right?

vxgmichel: By the way, the default is `auto`: https://github.com/pre-commit/pre-commit-hooks#mixed-line-ending

setup.cfg
touilleMan: yup (on top of having backend_factory using the postgresql implementation instead of the memory mock one)

.pre-commit-config.yaml
touilleMan: It's probably an issue with your git config on windows, you should customize `core.autocrlf=false` (using `true` make git convert back and forth the encoding before doing the commit, hence having the pre-commit complaining given it does it check before the conversion happens)

I think we should keep this check because it's far too easy to add `CRLF` otherwise (typically when creating a new file on windows, which is not fixed by `mixed-line-ending=auto`)

touilleMan: One possible solution would be to do `mixed-line-ending=auto` on local and `mixed-line-ending=lf` in the CI (it should be fairly easy to do that by adding a sed in the azure script)

vxgmichel: Done! 4149287

vxgmichel:
Merged: https://github.com/Scille/parsec-cloud/commit/48b17b8bf86621c428acee949ee9d5d328f4c31d

PR #1231, opened 2020-06-29
Add missing backend rep check in core bootstrap_organization

touilleMan:


parsec/core/invite/organization.py
```diff
@@ -66,5 +67,9 @@ async def bootstrap_organization(
         redacted_user_certificate=redacted_user_certificate,
         redacted_device_certificate=redacted_device_certificate,
     )
+    if rep["status"] in ("not_found", "already_bootstrapped"):
+        raise InviteNotAvailableError()
```
Max-7, line 71: Can we have another exception ? In the GUI it makes sense to be able to differentiate these cases and display an error message accordingly.

Issue #918, opened 2020-01-15
Deployment scripts for on-premises backend install


stefan-contiu:
We need an automated deployment mechanism for organizations that want to deploy the back-end (metadata service & db) internally.

Options :
* docker compose: https://docs.docker.com/compose/
* ansible: https://www.ansible.com/

c4ffein:
Will be useful for certification

c4ffein:
Implemented by #1233

PR #1233, opened 2020-06-30
Made a docker-compose

c4ffein:


c4ffein:
Implements #918

docker-compose.yml
```diff
@@ -0,0 +1,43 @@
...
+
+    networks:
+      - parsec_net
+  postgres:
+    image: "postgres"
+    volumes:
+      - postgres-data:/var/lib/postgresql/data/
+    environment:
+      - POSTGRES_USER=parsec
+      - POSTGRES_PASSWORD=DBPASS  # Change for prod
```
touilleMan, line 33: As I understand, Docker-Compose doesn't allow you to define global variables that could be overwritten when running the docker-compose file ?

docker-compose.yml
```diff
@@ -0,0 +1,43 @@
...
+services:
+  parsec:
+    build:
+      context: .
+      dockerfile: packaging/docker/Dockerfile
+    environment:
+      - PARSEC_HOST=0.0.0.0
+      - PARSEC_DB=postgresql://parsec@postgres/parsec?password=DBPASS  # Change db password for prod
+      - PARSEC_BLOCKSTORE=POSTGRESQL  # This is a demo...
+      - PARSEC_LOG_LEVEL=DEBUG  # Change for prod
```
touilleMan, line 11: I think `DEBUG` is too verbose for most usecase, better stick with `INFO`

docker-compose.yml
```diff
@@ -0,0 +1,43 @@
...
+      - PARSEC_ADMINISTRATION_TOKEN=CH4NG3M3  # Change for prod
```
touilleMan, line 12: we should use `s3cr3t` here given it's the default token when running with `--dev`

packaging/docker/wait-for-postgres.sh
```diff
@@ -0,0 +1,22 @@
+#!/bin/sh
+# wait-for-postgres.sh
+
+# Make Parsec wait for Postgres as it is not resilient enough
```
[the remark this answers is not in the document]
vamonte: https://docs.docker.com/compose/startup-order/

docker-compose.yml
```diff
@@ -0,0 +1,43 @@
+version: '2'
```
vamonte, line 1: The last version of the compose file format is 3.8. 
Why do you use the version 2 ?
https://docs.docker.com/compose/compose-file/

docker-compose.yml
```diff
@@ -0,0 +1,43 @@
...
+      - POSTGRES_DB=parsec
+    networks:
+      - parsec_net
+
+
+volumes:
+  postgres-data:
+
+networks:
+  parsec_net:
```
vamonte, line 43: Why do you set one custom network ? 
One default network is always created.
https://docs.docker.com/compose/networking/

packaging/docker/wait-for-postgres.sh
```diff
@@ -0,0 +1,22 @@
...
+# As recommended as a quickfix even by Docker documentation
+# Will be rendered useless when docker-compose is updated
+
+set -e
+
+host="$1"
+shift
+cmd="$@"
+
+apt-get update -y; apt-get install postgresql-client -y  # I'm ashamed.
```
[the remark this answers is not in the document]
c4ffein: Well, this file is only temporary, until we make Parsec able to wait for Postgres by default...

docker-compose.yml
```diff
@@ -0,0 +1,43 @@
...
+      - PARSEC_LOG_LEVEL=DEBUG  # Change for prod
+      - PARSEC_ADMINISTRATION_TOKEN=CH4NG3M3  # Change for prod
+    ports:
+      - 6777:6777
+
+    #### TODO : that part could be useless if Parsec is made to retry connections ####
+    depends_on:
+      - "postgres"
+    entrypoint: []
+    command: ["/bin/sh", "/wait-for-postgres.sh", "postgresql://parsec@postgres/parsec?password=DBPASS", "parsec", "backend", "run"]
```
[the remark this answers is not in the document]
c4ffein: Nope, the point of docker-compose is to launch a predefined postgres instance, for the easiest deploy. That's what the certifiers want to easily deploy our app if I understood @stefan-contiu correctly

packaging/docker/wait-for-postgres.sh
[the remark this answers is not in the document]
c4ffein: yup, will be included

docker-compose.yml
c4ffein: Right, they could still modify it if needed

docker-compose.yml
c4ffein: Old habit, changed that

docker-compose.yml
c4ffein: For exemple, I have tested my file as version 3.3 on Solus, it's still working as expected. But docker-compose 3.8 is still not supported on Solus (while the docker core is obtained through the docker repo, docker-compose is obtained through the OS repos, which tend to be not recent enough on some distributions).

Why not keep it a low number, and benefit from docker-compose backward compatibility to make it work on as much platforms as possible? Otherwise, I would still tend to go for a non-latest version, like the 3.3 that seems to work perfectly on all my setups.

@vamonte

PR #1235, opened 2020-06-30
 Immediate deletion in winfsp cleanup operation 

### LINKED ISSUE #1232: Investigate flakiness of marked mountpoint tests
Following up on #811, #1036, #1131, #1225, #1230, #1228, https://github.com/Scille/winfspy/issues/24.

Now that #1228 is merged, four mountpoint tests are marked as flaky:
- `test_mountpoint_revoke_access`
- `test_folder_operations`
- `test_winfstests`
- `test_winfsp_tests` (although there's been no occurrences of flakiness for this one)

The issue for at least some of them seems to be related to windows deletion not being performed correctly (see the analysis here https://github.com/Scille/parsec-cloud/issues/1230#issuecomment-651091027)

vxgmichel:
~Might fix #1232~
Does fix #1232 \o/

### DIFF
--- a/parsec/core/mountpoint/winfsp_operations.py
+++ b/parsec/core/mountpoint/winfsp_operations.py
@@ -233,12 +233,6 @@ class WinFSPOperations(BaseFileSystemOperations):
         if isinstance(file_context, OpenedFile):
             self.fs_access.fd_close(file_context.fd)
 
-        if file_context.deleted:
-            if isinstance(file_context, OpenedFile):
-                self.fs_access.file_delete(file_context.path)
-            else:
-                self.fs_access.folder_delete(file_context.path)
-
     @handle_error
     def get_file_info(self, file_context):
         stat = self.fs_access.entry_info(file_context.path)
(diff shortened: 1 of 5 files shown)


parsec/core/mountpoint/winfsp_operations.py
```diff
@@ -233,12 +233,6 @@ def close(self, file_context):
         if isinstance(file_context, OpenedFile):
             self.fs_access.fd_close(file_context.fd)
```
touilleMan, line 234: Is it ok to do `fd_close` on a deleted file ? (should it trigger an exception ?)

Anyway I think a comment about this would be good here ^^

vxgmichel: It raises a `FSInvalidFileDescriptor` (`errno.EBADF`, `ntstatus.STATUS_INVALID_HANDLE`)

I don't know if it's a problem or not, the test suite seems to agreee with it though :thinking:

vxgmichel: Actually deleting a file does not delete the underlying manifest, it simply removed this entry from the parent manifest.

vxgmichel: b6876b9

bitlogik:
Thanks God! Congrats. 🎉 After 6 issues opened, 7 months passed, hundreds of testing pipelines failures, good to have this root cause found and fixed 👍 👍 😃

PR #1244, opened 2020-07-02
[GUI] Fixed missing reject method on file history

### LINKED ISSUE #1239: AttributeError: 'FileHistoryWidget' object has no attribute 'reject'
Sentry Issue: [PARSEC-QQZ](https://sentry.io/organizations/scille/issues/1756961116/?referrer=github_integration)

```
AttributeError: 'FileHistoryWidget' object has no attribute 'reject'
  File "parsec\core\gui\file_history_widget.py", line 136, in on_get_version_error
    self.reject()
```

Max-7:
Closes #1239

### DIFF
--- a/parsec/core/gui/file_history_widget.py
+++ b/parsec/core/gui/file_history_widget.py
@@ -51,6 +51,7 @@ class FileHistoryWidget(QWidget, Ui_FileHistoryWidget):
         super().__init__()
         self.setupUi(self)
         self.jobs_ctx = jobs_ctx
+        self.dialog = None
         update_version_list.connect(self.reset_dialog)
         self.get_versions_success.connect(self.on_get_version_success)
         self.get_versions_error.connect(self.on_get_version_error)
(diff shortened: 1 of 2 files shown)


parsec/core/gui/file_history_widget.py
```diff
@@ -161,4 +162,5 @@ def exec_modal(
         d = GreyedDialog(
             w, title=_("TEXT_FILE_HISTORY_TITLE_name").format(name=path.name), parent=parent
         )
+        w.dialog = d
```
touilleMan, line 165: Wouldn't it better to create GreyedDialog in `FileHistoryWidget.__init__` so that `FileHistoryWidget.dialog` can never be `None` ?

Max-7: It might. I originally designed these widgets to be independant (showing them without a GreyedDialog). Later on when GreyedDialog was added, it seemed easier to add a classmethod to do the job rather than update the widget itself.
It would also be weird to have a widget construct itself and its parent in the __init__. I'm not very satisfied with this dialog variable but it is, imo, the best choice.

PR #1252, opened 2020-07-08
Add support for confined (temporary) files

### LINKED ISSUE #990: Provide file sync filter config to ignore tilde lock files
Multiple softwares (Excel, Open office, Emacs to name the more common ones :trollface: ) rely on temporary files to mark a file as opened (typically opening `my_file.doc` in Microsoft Word creates a `~$my_file.doc` lock file) an prevent concurrent edit on it.

Currently there is no special treatment for those files whatsoever.
This has multiple drawbacks:
- useless synchronization: Those file has no real value in the long run, so we are just slowing down the Parsec client and eating up more metadata (and data, but those files are typically really small so this is not a big deal) in the metadata server.
- other devices may retrieve the lock file which would emit strange error (the software would see the lock file and claim the file is already opened, where it's not the case on the device)
- lock file may contain configuration specific to the machine it has been created on, which would further lead to strange errors
- multiple devices opening the same file at the same time may end up with a conflict on the lock file

A simple solution to this would disable synchronization on certain files based on there name pattern.
Obviously the pattern should be configurable (given it could depend very much) and have a sensitive value by default (typically ignoring files with pattern `~$.*`)

vxgmichel:
Fix #990

Add support for confined (i.e temporary) files and directories.
In this context, confined means files that are not meant to be synchronized with other clients

### DIFF
--- a/newsfragments/990.feature.rst
+++ b/newsfragments/990.feature.rst
@@ -1,0 +1,3 @@
+Add support for confined (i.e temporary) files and directories.
+In this context, confined means files that are not meant to be
+synchronized with other clients
--- a/parsec/core/config.py
+++ b/parsec/core/config.py
@@ -80,6 +80,7 @@ class CoreConfig:
     gui_confirmation_before_close: bool = True
     gui_workspace_color: bool = False
     gui_allow_multiple_instances: bool = False
+    gui_show_confined: bool = False
 
     ipc_socket_file: Path = None
     ipc_win32_mutex_name: str = "parsec-cloud"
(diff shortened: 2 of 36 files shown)


bitlogik:
Great progress ! 👍 

I propose that the confined file detection is based on files name :

- starting with "~$" or
- starting with ".~" or
- ending with ".tmp"

In the [initial proposal](https://github.com/Scille/parsec-cloud/issues/990#issuecomment-625159723), I even suggested it was "*.tmp" files which start with "\~". We can let this point as just ".tmp" ending.

One of the main use case for this evolution is to filter temporary Microsoft Word (or any Office) documents. When they are edited, they are copied with a prepended "$~" file name (no "tmp" suffix).

vxgmichel:
@bitlogik 
> I propose that the confined file detection is based on files name :
> [...]

Thanks for the suggestion! I'm definitely planning to implement this. Here's my todo list for the moment:
- [x] Add confined entry set to local manifest
- [ ] Add filtered entry set to local manifest (filtering entries from other clients)
- [ ] Define a confinement pattern list
- [ ] Add dedicated icons in the GUI for confined files
- [ ] Make the pattern list configurable (optional)

Issue #1220, opened 2020-06-23
File version display confusing


bitlogik:
When importing a new file in a workspace, the file creation is considered a version, then the file competion (uploaded?) is considered a second version.

![image](https://user-images.githubusercontent.com/32455462/85394073-4cdf4a80-b54e-11ea-9eb3-1d986b7cffaa.png)

It is shown to the user as there were 2 versions of the file, but there is only one. The version displayed can be changed to consider the file creation or file adding ("0 bytes") as the first file input, not as a first version. And "Version 2" (file filled) will become the first version.

```
File creation (or "New file")   Size : 0 b
Version 1                       Size : 2 Mb
...
```

So by only changing the display of the version in the GUI (Version1 -> Creation, Version2 -> Version 1), it will be much more clearer for the user and closer from the actual experience.

bitlogik:
Also, some files display for the history only the "Version 2", starting their history at this point.
The versions numbers should really be shifted. "Version 2" needs to be "Version 1". And if there is a "Version 1", this should be displayed as "Version 0", or more friendly as "File creation" or "New file".

PR #1255, opened 2020-07-09
File versions display update

bitlogik:
Quickly fix https://github.com/Scille/parsec-cloud/issues/1220

### DIFF
--- a/newsfragments/1220.misc.rst
+++ b/newsfragments/1220.misc.rst
@@ -1,0 +1,1 @@
+Update the file version display in the GUI.
--- a/parsec/core/gui/file_history_widget.py
+++ b/parsec/core/gui/file_history_widget.py
@@ -33,7 +33,7 @@ class FileHistoryButton(QWidget, Ui_FileHistoryButton):
         super().__init__()
         self.setupUi(self)
         if version <= 1:
-            version_str = "file creation"
+            version_str = _("TEXT_FILE_HISTORY_CREATION")
         else:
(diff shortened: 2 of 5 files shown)


bitlogik:
We forget that English is not the only language. So this requires some GUI translation. Like _("TEXT_FILE_CREATION"). Or we can use TEXT_FILE_CREATED_HEADER so the string is already present and translated in gui/tr.
It would display : "Version Created" if version == 1 (when dir or 0 size)
Then Version 1, Version 2, ...

bitlogik:
I guess that we have the work here to do again. As this looks fully in conflict with #1283

bitlogik:
OK so the #1283 fix doesn't conflict with this pull request. Still, as it added a test about the displayed version, this pull request will be amended so that it makes this test pass. I think especially at this
https://github.com/Scille/parsec-cloud/pull/1283/files#diff-cb0f3b4073387b5cd7dd2f50c117b185R101
Which should be replaced by "data creation" or _("TEXT_FILE_HISTORY_CREATION")

Is there "_" tr function in the tests? Are the tests in English GUI ?

bitlogik:
_Is there "\_" tr function in the tests?_
Not sure, but all GUI tests are checked against plain English without tr _
_Are the tests in English GUI ?_ Yes

We updated this pull request according to the answers above.
It was marked as "passed" before, because it was based on an incompatible head. This rebase including the updated test should handle the stuff.

PR #1258, opened 2020-07-09
Made Parsec able to retry first db connection, cleaner docker-compose

c4ffein:


parsec/backend/cli/run.py
```diff
@@ -227,6 +227,26 @@ def handle_parse_result(self, ctx, opts, args):
     envvar="PARSEC_DB_MAX_CONNECTIONS",
     help="Maximum number of connections to the database if using PostgreSQL",
 )
+@click.option(
+    "--db-first-tries-number",
+    default=1,
+    show_default=True,
+    envvar="PARSEC_DB_FIRST_TRIES_NUMBER",
+    help=(
+        "Number of first tries allowed to connect to the database if using PostgreSQL, "
```
touilleMan, line 236: Given PostgreSQL is the only "real" choice, I guess we can skip the "if using PostgreSQL" part and just say `Number of tries allowed during initial database connection (0 is unlimited)` and `Number of second waited between tries during initial database connection`

parsec/backend/postgresql/handler.py
```diff
@@ -166,15 +176,32 @@ async def init(self, nursery):
...
+                        )
+                        task_status.started()
+                        await trio.sleep_forever()
+            except OSError as err:
+                if self.first_tries_number == 1:
+                    raise err
+                if self.first_tries_number > 1:
+                    self.first_tries_number -= 1
+                logger.warning(
+                    "first db connection failed",
```
touilleMan, line 198: `initial database connection failed`

parsec/backend/postgresql/handler.py
```diff
@@ -166,15 +176,32 @@ async def init(self, nursery):
...
+                    tries_remaining=self.first_tries_number
+                    if self.first_tries_number
+                    else "unlimited",
```
touilleMan, line 201: why not using unicode `∞` :trollface:

parsec/backend/postgresql/handler.py
```diff
@@ -166,15 +176,32 @@ async def init(self, nursery):
...
+                    # would only complicate stuff to include it into the connection pool
+                    async with triopg.connect(self.url) as self.notification_conn:
+                        await self.notification_conn.add_listener(
+                            "app_notification", self._on_notification
+                        )
+                        task_status.started()
+                        await trio.sleep_forever()
+            except OSError as err:
+                if self.first_tries_number == 1:
+                    raise err
```
touilleMan, line 194: can use `raise` without err (this way we can also have `except OSError:`)

parsec/backend/cli/run.py
c4ffein: Done

parsec/backend/postgresql/handler.py
c4ffein: Changed

parsec/backend/postgresql/handler.py
c4ffein: :thinking:

parsec/backend/postgresql/handler.py
c4ffein: changed

Issue #1267, opened 2020-07-12
Flaky error: gui/test_revoked_user.py::test_revoked_notification


touilleMan:
https://dev.azure.com/Scille/parsec/_build/results?buildId=1924&view=logs&j=bfccef3f-45d4-504d-9a7f-a8a74922e6bf&t=cac1cc85-f7fc-5d1e-c90b-1188f789d25b&l=152

```
tests/core/gui/test_revoked_user.py::test_revoked_notification FAILED    [ 81%]

================================== FAILURES ===================================
__________________________ test_revoked_notification __________________________

value = <trio.Nursery object at 0x00000144BFF15F98>

    async def yield_(value=None):
>       return await _yield_(value)

c:\hostedtoolcache\windows\python\3.6.8\x64\lib\site-packages\async_generator\_impl.py:106: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
c:\hostedtoolcache\windows\python\3.6.8\x64\lib\site-packages\async_generator\_impl.py:99: in _yield_
    return (yield _wrap(value))
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

aqtbot = <tests.core.gui.conftest.AsyncQtBot object at 0x00000144BFF151D0>
running_backend = AppServer(entry_point=<bound method BackendApp.handle_client of <parsec.backend.app.BackendApp object at 0x00000144BFE...onnection_factory=<function server_factory.<locals>._server_factory.<locals>.connection_factory at 0x00000144C12377B8>)
autoclose_dialog = <tests.core.gui.conftest.autoclose_dialog.<locals>.DialogSpy object at 0x00000144C1246EB8>
logged_gui = <parsec.core.gui.main_window.MainWindow object at 0x00000144BF464948>
alice_user_fs = <parsec.core.fs.userfs.userfs.UserFS object at 0x00000144BFF12A20>
bob = LocalDevice(organization_addr=BackendOrganizationAddr(url=parsec://bob.dev1.example.com:9999/CoolOrg?no_ssl=true&rvk=2..._key=<parsec.crypto.SecretKey object at 0x144c138d468>, local_symkey=<parsec.crypto.SecretKey object at 0x144c138d858>)

[...]

PR #1272, opened 2020-07-15
Gui flaky test revoked user

AntoineDupre:
fix issue #1267

I ran it 50 times in the pipeline [there](https://dev.azure.com/Scille/parsec/_build/results?buildId=2001&view=logs&j=bfccef3f-45d4-504d-9a7f-a8a74922e6bf&t=1b8d0bba-f11a-562f-077c-1aaa5fdf35ee)

### DIFF
--- a/tests/core/gui/test_revoked_user.py
+++ b/tests/core/gui/test_revoked_user.py
@@ -1,36 +1,36 @@
 # Parsec Cloud (https://parsec.cloud) Copyright (c) AGPLv3 2019 Scille SAS
 
-import pendulum
 import pytest
 from PyQt5.QtWidgets import QLabel
-
-from parsec.api.data import RevokedUserCertificateContent
 
 
 @pytest.mark.gui
 @pytest.mark.trio
 async def test_revoked_notification(
-    aqtbot, running_backend, autoclose_dialog, logged_gui, alice_user_fs, bob
+    aqtbot, running_backend, backend, autoclose_dialog, logged_gui, alice, bob
 ):
-    now = pendulum.now()
-    revoked_device_certificate = RevokedUserCertificateContent(
-        author=alice_user_fs.device.device_id, timestamp=now, user_id=bob.user_id
-    ).dump_and_sign(alice_user_fs.device.signing_key)
-    central_widget = logged_gui.test_get_central_widget()
 
+    central_widget = logged_gui.test_get_central_widget()
     assert central_widget is not None
-    async with aqtbot.wait_signal(central_widget.new_notification, timeout=3000):
-        await alice_user_fs.backend_cmds.user_revoke(revoked_device_certificate)
+
+    await backend.user.revoke_user(
+        organization_id=alice.organization_id,
+        user_id=bob.user_id,
+        revoked_user_certificate=b"dummy",
+        revoked_user_certifier=alice.device_id,
+    )
 
     # Assert dialog
-    assert len(autoclose_dialog.dialogs) == 1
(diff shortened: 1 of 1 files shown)


parsec/core/sync_monitor.py
```diff
@@ -231,7 +231,7 @@ async def tick(self) -> float:
                 del self._local_changes[entry_id]
                 try:
                     await self._sync(entry_id)
-                except FSBackendOfflineError as exc:
+                except (FSBackendOfflineError, BackendConnectionError) as exc:
```
touilleMan, line 234: BackendConnectionError != BackendNotAvailable
for instance we could have a `BackendConnectionRefused` because our device has been revoked (or because our organization has expired)

So I don't think we should do this conversion, especially given `BackendNotAvailable` is a subclass of `BackendConnectionError`, hence calling code expecting  `BackendNotAvailable` should also expects `BackendConnectionError` (but, of course, does have to expect the Spanish inquisition :trollface: )

tests/core/gui/test_revoked_user.py
```diff
@@ -36,21 +36,28 @@ async def test_revoked_notification(
     ).dump_and_sign(alice_user_fs.device.signing_key)
     central_widget = logged_gui.test_get_central_widget()
     assert central_widget is not None
-    async with aqtbot.wait_signal(central_widget.new_notification):
+    async with aqtbot.wait_signal(central_widget.new_notification, timeout=3000):
```
touilleMan, line 39: default timeout is 1s, why is this not enough ?

tests/core/gui/test_revoked_user.py
```diff
@@ -36,21 +36,28 @@ async def test_revoked_notification(
...
 
     # Assert dialog
     assert len(autoclose_dialog.dialogs) == 1
     dialog = autoclose_dialog.dialogs[0]
     assert dialog[0] == "Error"
     assert dialog[1] == "This device is revoked"
 
-    # Assert device and user widgets
-    devices_w = logged_gui.test_get_devices_widget()
     users_w = logged_gui.test_get_users_widget()
```
touilleMan, line 48: I think we should use the brand new shinny `logged_gui.test_switch_to_devices_widget` (which really click on the page instead of just getting the widget)

AntoineDupre: it's simplify things a lot ;)

tests/core/gui/test_revoked_user.py
AntoineDupre: 1second is generating random timeout.

touilleMan: can't we avoid this by using `aqtbot.wait_until(callbable)` instead of `wait_signal` ? (`wait_until` is much more reliable)

tests/core/gui/test_revoked_user.py
```diff
@@ -2,31 +2,15 @@
...
-    async with aqtbot.wait_signals([lw.login_with_password_clicked, tabw.logged_in]):
-        await aqtbot.mouse_click(lw.button_login, QtCore.Qt.LeftButton)
-
-    return gui
+from tests.common import customize_fixtures
 
 
 @pytest.mark.gui
 @pytest.mark.trio
+@customize_fixtures(logged_gui_as_admin=False)
```
touilleMan, line 13: no need to set this to false given it's the value by default ;-)

AntoineDupre: Can we keep it to explicitly point that we are logged as bob ?

Issue #1271, opened 2020-07-15
Can not delete invitation in the gui


AntoineDupre:
Delete a pending invitation raise the following error: 

```
Traceback (most recent call last):
  File "/home/antdup/parsec-cloud/parsec/core/gui/trio_thread.py", line 94, in _run_fn
    result = await self._fn(*self._args, **self._kwargs)
  File "/home/antdup/parsec-cloud/parsec/core/gui/users_widget.py", line 173, in _do_cancel_invitation
    await core.delete_invitation(token=token)
TypeError: delete_invitation() missing 1 required positional argument: 'reason'
Traceback (most recent call last):
  File "/home/antdup/parsec-cloud/parsec/core/gui/trio_thread.py", line 94, in _run_fn
    result = await self._fn(*self._args, **self._kwargs)
  File "/home/antdup/parsec-cloud/parsec/core/gui/users_widget.py", line 173, in _do_cancel_invitation
    await core.delete_invitation(token=token)
TypeError: delete_invitation() missing 1 required positional argument: 'reason'
```

AntoineDupre:
merged

PR #1273, opened 2020-07-15
Fix cancel user invitation

AntoineDupre:
#1271

### DIFF
--- a/parsec/core/logged_core.py
+++ b/parsec/core/logged_core.py
@@ -198,14 +198,14 @@ class LoggedCore:
             token=rep["token"],
         )
 
-    async def delete_invitation(self, token: UUID, reason: InvitationDeletedReason) -> None:
+    async def delete_invitation(
+        self, token: UUID, reason: InvitationDeletedReason = InvitationDeletedReason.CANCELLED
+    ) -> None:
         """
         Raises:
             BackendConnectionError
         """
-        rep = await self._backend_conn.cmds.invite_delete(
-            token=token, reason=InvitationDeletedReason.CANCELLED
-        )
+        rep = await self._backend_conn.cmds.invite_delete(token=token, reason=reason)
         if rep["status"] != "ok":
             raise BackendConnectionError(f"Backend error: {rep}")
 
--- a/tests/core/gui/users_widget/test_invite.py
+++ b/tests/core/gui/users_widget/test_invite.py
@@ -136,3 +136,50 @@ async def test_revoke_user_not_allowed(
 
     assert autoclose_dialog.dialogs == [("Error", "Could not revoke this user.")]
     assert alice_w.user_info.is_revoked is False
+
+
+@pytest.mark.gui
+@pytest.mark.trio
+@customize_fixtures(backend_has_email=True, logged_gui_as_admin=True)
+async def test_cancel_user_invitation(
+    aqtbot,
+    logged_gui,
+    running_backend,
+    backend,
+    autoclose_dialog,
+    monkeypatch,
+    alice,
+    email_letterbox,
+):
+
+    email = "i@like.coffee"
+
+    # Patch dialogs
+    monkeypatch.setattr("parsec.core.gui.users_widget.get_text_input", lambda *x, **y: email)
+    monkeypatch.setattr(
+        "parsec.core.gui.users_widget.ask_question",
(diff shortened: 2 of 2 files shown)


tests/core/gui/test_invite_user.py
```diff
@@ -0,0 +1,66 @@
+# Parsec Cloud (https://parsec.cloud) Copyright (c) AGPLv3 2019 Scille SAS
```
touilleMan, line 1: should move this test into `core/gui/users_widget`

tests/core/gui/test_invite_user.py
```diff
@@ -0,0 +1,66 @@
...
+from parsec.core.gui.lang import translate as _
+
+from tests.common import customize_fixtures
+
+
+@pytest.mark.gui
+@pytest.mark.trio
+@customize_fixtures(logged_gui_as_admin=True)
+async def test_cancel_user_invitation(
+    qapp, aqtbot, logged_gui, running_backend, backend, autoclose_dialog, monkeypatch, alice
```
touilleMan, line 16: `qapp` fixture needed ?

tests/core/gui/test_invite_user.py
```diff
@@ -0,0 +1,66 @@
...
+    # Patch dialogs
+    monkeypatch.setattr("parsec.core.gui.users_widget.get_text_input", lambda *x, **y: email)
+    monkeypatch.setattr(
+        "parsec.core.gui.users_widget.ask_question",
+        lambda *x, **y: _("TEXT_USER_INVITE_CANCEL_INVITE_ACCEPT"),
+    )
+    u_w = await logged_gui.test_switch_to_users_widget()
+
+    # Force send_email to false
+    _invite_new = u_w.core._backend_conn.cmds.invite_new
```
touilleMan, line 30: no need for this you can use the decorator `@customize_fixtures(backend_has_email=True)`, then the fixture `email_letterbox` (see `tests/core/gui/users_widgets/test_invite.py`)

tests/core/gui/test_invite_user.py
```diff
@@ -0,0 +1,66 @@
...
+    async with aqtbot.wait_signals(
+        [
+            user_invitation_w.button_cancel.clicked,
+            user_invitation_w.cancel_clicked,
+            u_w.cancel_invitation_success,
+        ]
+    ):
+        await aqtbot.mouse_click(user_invitation_w.button_cancel, QtCore.Qt.LeftButton)
+
+    # Force gui update
```
touilleMan, line 61: Do we need this ? I would say the list should be updated once we've clicked on the cancel button (hence no need to click again on the list users button)

parsec/core/gui/users_widget.py
```diff
@@ -170,7 +170,7 @@ async def _do_list_users_and_invitations(core):
 
 async def _do_cancel_invitation(core, token):
     try:
-        await core.delete_invitation(token=token)
+        await core.delete_invitation(token=token, reason=InvitationDeletedReason.CANCELLED)
```
touilleMan, line 173: Funny thing is `core.delete_invitation` doesn't use the `reason` parameters ! https://github.com/Scille/parsec-cloud/blob/39866ad773c2333dc10b2cdb8a1ff8b7b553b099/parsec/core/logged_core.py#L206-L208

I think we should modify `core.delete_invitation` to use the `reason` param and to provide it with a default value of `InvitationDeletedReason.CANCELLED` (hence nothing to change in `_do_cancel_invitation`)

tests/core/gui/test_invite_user.py
AntoineDupre: it's much better :)

tests/core/gui/test_invite_user.py
AntoineDupre: yes it is better to wait on signal `u_w.list_success` to be sure that reset has been done

tests/core/gui/test_invite_user.py
AntoineDupre: removed

tests/core/gui/test_invite_user.py
AntoineDupre: i moved the test into `core/gui/users_widget/test_invite.py`

parsec/core/gui/users_widget.py
AntoineDupre: fixed

Issue #1274, opened 2020-07-16
GUI: change device password broken


AntoineDupre:
```
Traceback (most recent call last):
  File "/home/antdup/parsec-cloud/parsec/core/gui/devices_widget.py", line 54, in change_password
    self.change_password_clicked.emit(self.device_name)
AttributeError: 'DeviceButton' object has no attribute 'device_name'
```

AntoineDupre:
merged

PR #1275, opened 2020-07-16
Fix device password change

AntoineDupre:
issue  #1274

### DIFF
--- a/parsec/core/gui/devices_widget.py
+++ b/parsec/core/gui/devices_widget.py
@@ -17,7 +17,7 @@ from parsec.core.gui.ui.device_button import Ui_DeviceButton
 
 
 class DeviceButton(QWidget, Ui_DeviceButton):
-    change_password_clicked = pyqtSignal(str)
+    change_password_clicked = pyqtSignal()
 
     def __init__(self, device_info, is_current_device):
         super().__init__()
@@ -51,7 +51,7 @@ class DeviceButton(QWidget, Ui_DeviceButton):
         menu.exec_(global_pos)
 
     def change_password(self):
-        self.change_password_clicked.emit(self.device_name)
+        self.change_password_clicked.emit()
 
 
 async def _do_invite_device(core):
@@ -117,7 +117,7 @@ class DevicesWidget(QWidget, Ui_DevicesWidget):
                 else:
                     w.show()
 
-    def change_password(self, device_name):
+    def change_password(self):
         PasswordChangeWidget.exec_modal(core=self.core, parent=self)
 
     def invite_device(self):
--- a/tests/core/gui/test_password_change.py
+++ b/tests/core/gui/test_password_change.py
@@ -149,3 +149,27 @@ async def test_change_password_success(
 
     async with aqtbot.wait_signals([lw.login_with_password_clicked, tabw.logged_in]):
         await aqtbot.mouse_click(lw.button_login, QtCore.Qt.LeftButton)
+
+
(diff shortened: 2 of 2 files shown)


tests/core/gui/test_password_change.py
```diff
@@ -149,3 +149,24 @@ def _create_change_password_dialog():
 
     async with aqtbot.wait_signals([lw.login_with_password_clicked, tabw.logged_in]):
         await aqtbot.mouse_click(lw.button_login, QtCore.Qt.LeftButton)
+
+
+@pytest.mark.gui
+@pytest.mark.trio
+async def test_change_password_canceled(
```
c4ffein, line 156: Why name it cancel? It just checks the dialog is open?

parsec/core/gui/devices_widget.py
```diff
@@ -51,7 +51,7 @@ def show_context_menu(self, pos):
         menu.exec_(global_pos)
 
     def change_password(self):
-        self.change_password_clicked.emit(self.device_name)
+        self.change_password_clicked.emit(self.device_info.device_label)
```
touilleMan, line 54: shouldn't it be `self.device_info.device_display` given `device_label` can be `None` ?

besides `self.change_password_clicked` is connected to `DevicesWidget.change_password` which make no use of it parameter `device_name`... so do we really need it ?

Max-7, line 54: The argument is not actually used (since it's only possible to change the password of the current device) and can be removed.

tests/core/gui/test_password_change.py
AntoineDupre: Because it uses `autoclose_dialog` fixture that close the dialog window, so it test that the widget is display and also that nothing happen if it's simply  closed

parsec/core/gui/devices_widget.py
AntoineDupre: good point !

tests/core/gui/test_password_change.py
c4ffein: Right, missed that one

tests/core/gui/test_password_change.py
```diff
@@ -149,3 +149,24 @@ def _create_change_password_dialog():
...
+    aqtbot, running_backend, logged_gui, qt_thread_gateway, autoclose_dialog
+):
+    d_w = logged_gui.test_get_devices_widget()
+
+    assert d_w is not None
+    async with aqtbot.wait_signal(d_w.list_success):
```
touilleMan, line 162: instead of waiting on signals which is cumbersome, not reliable and error prone (rumors say it also gives cancer... :trollface: ) it's better to use `aqtbot.wait_until(callable)`:

```python
def _devices_displayed():
        assert d_w.isVisible()
        assert d_w.layout_devices.count() == 2

await aqtbot.wait_until(_devices_displayed)
```

(btw pay attention to use `widget.isVisible()` and not `not widget.isHidden()`, the two are not related)

parsec/core/gui/devices_widget.py
AntoineDupre: fixed

AntoineDupre: max spots it 1 minute before you ;)

tests/core/gui/test_password_change.py
AntoineDupre: fixed, thanks for the info

PR #1277, opened 2020-07-16
Fix docker-compose migration

c4ffein:


parsec/backend/postgresql/handler.py
```diff
@@ -71,13 +97,26 @@ class MigrationResult:
 
 
 async def apply_migrations(
-    url: str, migrations: List[MigrationItem], dry_run: bool
+    url: str, first_tries_number, first_tries_sleep, migrations: List[MigrationItem], dry_run: bool
```
touilleMan, line 100: missing typing ;-)

PR #1279, opened 2020-07-16
Add invitation_already_deleted result to api handshake and allow unkn…

touilleMan:
…own api handshake result


parsec/api/protocol/handshake.py
```diff
@@ -295,6 +299,17 @@ def build_bad_identity_result_req(self, help="Invalid handshake information") ->
             {"handshake": "result", "result": "bad_identity", "help": help}
         )
 
+    def build_already_deleted_invitation_result_req(
+        self, help="Invitation already used or cancelled"
```
vxgmichel, line 303: It might also be useful to make the distinction between:
- invitation already completed (e.g. a user tries to use the same invitation link on several device)
- invitation cancelled (e.g. the greeter cancelled the invitation)

touilleMan:
@Max-7 waiting for your call to merge this ;-)

touilleMan:
too old (and should be re-implemented in Rust)

PR #1280, opened 2020-07-16
Greatly improve tests/core/gui/user_widget/test_greet.py

touilleMan:


parsec/core/gui/claim_user_widget.py
```diff
@@ -532,31 +532,38 @@ def __init__(self, jobs_ctx, config, addr):
         self.retrieve_info_success.connect(self._on_retrieve_info_success)
         self.retrieve_info_error.connect(self._on_retrieve_info_error)
         self.claimer = Claimer()
+        self._count__ = 0
         self._run_claimer()
 
     def _run_claimer(self):
+        print("_run_claimer", self._count__)
```
Max-7, line 539: Remove the print and the count variable.

touilleMan: :+1:

tests/core/gui/test_revoked_user.py
```diff
@@ -13,14 +15,25 @@ async def test_revoked_notification(
...
     await backend.user.revoke_user(
         organization_id=alice.organization_id,
         user_id=bob.user_id,
         revoked_user_certificate=b"dummy",
         revoked_user_certifier=alice.device_id,
     )
 
-    # Assert dialog
+    # Make sure we try to reach the backend
+    await aqtbot.run(central_widget.show_users_widget)
```
touilleMan, line 34: Why is this necessary ? The event listener connection should close as soon as the revoked event is processed in the backend

tests/core/gui/test_revoked_user.py
```diff
@@ -13,14 +15,25 @@ async def test_revoked_notification(
     central_widget = logged_gui.test_get_central_widget()
     assert central_widget is not None
 
+    # Make sure we're connected
+    def _connected():
+        assert central_widget.connection_status == BackendConnStatus.READY
```
touilleMan, line 20: Wouldn't it be better to wait for the `central_widget.connection_state_changed` signal instead ?

Or even better waiting on:
```python
    def _connection_online():
        assert central_widget.menu.label_connection_state.text() == "Connected"
    await aqtbot.wait_until(_connection_online)
```

touilleMan:
image

PR #1281, opened 2020-07-17
Spontaneous organization bootstrap

touilleMan:


touilleMan:
no longer a draft ^^ @NicoTuxx can you have a look on this too ?

parsec/backend/webhooks.py
```diff
@@ -0,0 +1,54 @@
...
+        url, method="POST", headers={"content-type": "application/json; charset=utf-8"}, data=data
+    )
+    try:
+        with urlopen(req, timeout=30) as rep:
+            if rep.getcode() != 200:
+                logger.warning(
+                    "webhook bad return status", url=url, data=data, return_status=rep.getcode()
+                )
+
+    except URLError as exc:
```
c4ffein, line 25: Shouldn't we have a mechanism for keeping track of failed webhook, and retry later? Otherwise, could we have another mechanism that would enable the BMS to get all or specific non-received bootstrap infos?

touilleMan: Keeping track of send weebhook seems too much effort (especially given it's not something that should occur that often, and because it's not a perfect solution given the webhook is not part of the atomic transaction creating the organization).

I think it's much better to have the BMS do periodical checks to make sure it is synced with the backend. And given it's not a vital system we can wait for the new REST administration API to be ready before adding this to the BMS system ;-)

c4ffein: :+1:

PR #1282, opened 2020-07-20
Fix issue where backend-http-interface only work with test

Sithi5:


parsec/backend/app.py
```diff
@@ -247,7 +247,12 @@ async def handle_client(self, stream):
                 pass
 
     async def handle_client_http(self, stream, event, conn):
-        status_code, headers, data = get_method_and_execute(event.target, self.backend_addr)
+        if hasattr(self, "running_backend_addr"):
+            status_code, headers, data = get_method_and_execute(
+                url=event.target, backend_addr=self.running_backend_addr
+            )
+        else:
+            status_code, headers, data = get_method_and_execute(url=event.target, backend_addr=None)
```
c4ffein, line 255: Nice fix in theory, but if that was the correct attribute, shouldn't we provide a "guessed" `backend_addr` instead if it hasn't been defined by the user? (If I remember correctly from what @touilleMan told me months ago, `backend_addr` is used in case the local config of the server is different from what is accessible from the outside)

Also, where is supposed to be defined `self.backend_addr` in the nominal case? Arent't we just supposed to use `self.config.backend_addr` instead, for which existence is guaranteed?

parsec/backend/http/router.py
```diff
@@ -35,9 +35,9 @@ def get_method(url: bytes):
     return method
 
 
-def get_method_and_execute(url: bytes, backend_addr):
+def get_method_and_execute(url: bytes, backend_addr=None):
     method = get_method(url)
-    return method(url, backend_addr=backend_addr)
+    return method(url=url, backend_addr=backend_addr)
```
c4ffein, line 40: Same : really bad to have it optional here.

parsec/backend/http/templates/404.html
```diff
@@ -1,7 +1,10 @@
-{% extends "base.html" %} {% block title %}{{ super() }}Introuvable :'({% endblock %} {% block body %} {{ super() }}
+{% extends "base.html" %}
+{% block title %}{{ super() }}Introuvable :'({% endblock %}
+{% block body %}
+{{ super() }}
```
c4ffein, line 4: cleaner!

parsec/backend/http/templates/base.html
```diff
@@ -9,6 +9,8 @@
 </head>
 
 <body class="home">
+{% block body %}
+{% endblock %}
```
c4ffein, line 13: I'm disappointed at myself that I missed that one on the last review

tests/conftest.py
```diff
@@ -619,7 +619,7 @@ def running_backend_ready(request):
 @pytest.fixture
 async def running_backend(server_factory, backend_addr, backend, running_backend_ready):
 
-    backend.backend_addr = backend_addr
+    backend.running_backend_addr = backend_addr
```
c4ffein, line 622: Why?

parsec/backend/app.py
c4ffein:
```status_code, headers, data = get_method_and_execute(
    event.target, self.config.backend_addr if self.config.backend_addr else "https://localhost"+self.config.port
)```

File: parsec/backend/app.py [hunk 6]
```diff
@@ -24,7 +24,6 @@
 from parsec.backend.memory import components_factory as mocked_components_factory
 from parsec.backend.postgresql import components_factory as postgresql_components_factory
 from parsec.backend.http.router import get_method_and_execute
-
```
c4ffein, line 27: Better keep parsec and h11 imports separate

tests/scripts/run_testenv.py
```diff
@@ -128,7 +128,7 @@ async def restart_local_backend(administration_token, backend_port):
     pattern = f"parsec.* backend.* run.* -P {backend_port}"
     command = (
         f"{sys.executable} -Wignore -m parsec.cli backend run -b MOCKED --db MOCKED "
-        f"-P {backend_port} --administration-token {administration_token}"
+        f"-P {backend_port} --administration-token {administration_token} --backend-addr parsec://localhost"
```
touilleMan, line 131: `parsec://localhost:{backend_port}?no_ssl=true`

c4ffein, line 131: So we don't have address set by default at parsec://localhost if not provided?

parsec/backend/http/templates/404.html
touilleMan, line 4: no need for the super part given base block is empty

parsec/backend/http/router.py
```diff
@@ -94,10 +84,12 @@ def is_route(url: bytes):
 
 def _redirect_to_parsec(url: bytes, backend_addr, *arg, **kwarg):
     """Redirect the http invite request to a parsec url request"""
-    location = _create_parsec_url(backend_addr=backend_addr) + "/"
+    if not backend_addr:
+        return 501, {}, b""
```
touilleMan, line 88: should add a message in the body `Url redirection is not available`

Issue #1270, opened 2020-07-15
File history does not display authors properly 


AntoineDupre:
File history does not display authors properly 
![history](https://user-images.githubusercontent.com/1070501/87525568-f74c1880-c689-11ea-85cc-96d9dd0ebfd8.png)

Max-7:
VersionLister.list() returns a UserID instead of a UserInfo.

We could use LoggedCore.get_user_info() to retrieve the human handle from the UserID but it seems to me that it would be way better to have VersionLister.list() return the UserInfo directly as it would seriously limit the number of queries made to the backend.

bitlogik:
We made some tests on Parsec 12 days ago, and we noticed that. As there are many work in progress on this topic we were not sure if this was a bug or just a remaining task to clear. Author and device are not "decoded" in the file history listing.

When you wrote "we could use LoggedCore.get_user_info() from the backend", I thought that the main goal of this obfuscated id, was to hide the real info from the server, so that the server never knows the info names. Did I miss anything?
Indeed, on top of  avoiding a query to the server to get a content which is known locally, for the sake of logic, this user info should be get locally, not from the server.

Max-7:
Fixed by https://github.com/Scille/parsec-cloud/pull/1283, closing the issue.

PR #1283, opened 2020-07-20
Fix gui file history name

AntoineDupre:
#1270

### DIFF
--- a/newsfragments/1270.bugfix.rst
+++ b/newsfragments/1270.bugfix.rst
@@ -1,0 +1,1 @@
+Display author name in file history instead of DeviceID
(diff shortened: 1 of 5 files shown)


parsec/core/logged_core.py
```diff
@@ -116,6 +116,10 @@ async def get_user_info(self, user_id: UserID) -> UserInfo:
             created_on=user_certif.timestamp,
         )
 
+    async def get_user_info_from_device_id(self, device_id: DeviceID) -> UserInfo:
```
Max-7, line 119: The DeviceID class has a user_id property. You don't need the `get_user_info_from_device_id` method and the `split()`, you can just call `get_user_info(creator.user_id)`

AntoineDupre: thanks for the tip

PR #1286, opened 2020-07-21
Add pre-release support to misc/releaser.py

touilleMan:


misc/releaser.py
```diff
@@ -30,27 +32,126 @@ class ReleaseError(Exception):
...
 
 
-Version = namedtuple("Version", "full,short,major,minor,patch,extra")
-Version.__eq__ = lambda self, other: self.full == other.full
-Version.__lt__ = lambda self, other: (self.major, self.minor, self.patch) < (
-    other.major,
-    other.minor,
-    other.patch,
-)
+class Version:
```
vxgmichel, line 35: Why not simply use `from distutils.version import LooseVersion`?

misc/releaser.py
```diff
@@ -30,27 +32,126 @@ class ReleaseError(Exception):
...
+assert Version("1.2.3+dev") < Version("1.2.4-rc1")
+assert Version("1.2.4-rc1") < Version("1.2.4-rc1+dev")
+assert Version("1.2.4-rc1+dev") < Version("1.2.4")
+assert Version("1.2.3") < Version("1.2.3-10-g3b5f5762")
+assert Version("1.2.3-10-g3b5f5762") < Version("1.2.4-b42")
+assert Version("1.2.4-b42") < Version("1.2.4-b42-10-g3b5f5762")
+assert Version("1.2.4-b42-10-g3b5f5762") < Version("1.2.4")
+assert Version("1.2.3-rc10+dev") < Version("1.2.3")
+assert Version("1.2.3-b10+dev") < Version("1.2.3-rc1+dev")
+assert Version("1.2.3-b10+dev") < Version("1.2.3+dev")
```
vxgmichel, line 154: This should go to a `test_version` function so we can run `pytest misc/releaser.py`

touilleMan: The thing is releaser is use in the Q&A job of the CI, so in parallel with the Unittest job...

I think it's better to keep the asserts inline in the file given there is not much of them (so doesn't clutter too much the file and doesn't slow down the speed) and the script is not supposed to be modified often. This way we have a standalone script that ensure it integrity as part of it run ;-)

misc/releaser.py
touilleMan: because... I didn't remember where this class was ! (I was looking for `type(sys.version_info)`)

But now that Version is here and fine tuned, I don't think we should switch to `distutils.version.LooseVersion`

PR #1289, opened 2020-07-22
Email invite link

Sithi5:


parsec/backend/invite.py
```diff
@@ -201,8 +201,11 @@ def generate_invite_email(
             use_ssl=backend_addr.use_ssl,
         )
     )
+    https_invite_link = parsec_invite_link.replace("parsec://", "")
+    split_invite_link = https_invite_link.split("/")
+    https_invite_link = "https://" + split_invite_link[0] + "/api/redirect/" + split_invite_link[1]
     body = [
-        line.format(sender_name=sender_name, invite_link=invite_link, parsec_url=parsec_url)
+        line.format(sender_name=sender_name, invite_link=parsec_invite_link, parsec_url=parsec_url)
```
Ironicbay, line 208: Here the `invite_link` need to be `https_invite_link` , not `parsec_invite_link` (this corresponds to the clickable button)

parsec/backend/invite.py
```diff
@@ -201,8 +201,11 @@ def generate_invite_email(
             use_ssl=backend_addr.use_ssl,
         )
     )
+    https_invite_link = parsec_invite_link.replace("parsec://", "")
+    split_invite_link = https_invite_link.split("/")
+    https_invite_link = "https://" + split_invite_link[0] + "/api/redirect/" + split_invite_link[1]
```
Ironicbay, line 206: `https` or `http` ? Only `http` works for me

parsec/backend/invite.py
```diff
@@ -191,7 +193,7 @@ def generate_invite_email(
 ) -> Message:
     mail_text = MAIL_TEXT[email_config.language]
     parsec_url = "https://parsec.cloud/get-parsec"
-    invite_link = str(
+    parsec_invite_link = str(
```
touilleMan, line 196: Better to use the `BackendInvitationAddr.to_url` method than `str` conversion (it's kind of more explicit)

parsec/backend/invite.py
```diff
@@ -191,7 +193,7 @@ def generate_invite_email(
...
         BackendInvitationAddr(
```
touilleMan, line 197: It's better to create this with the `BackendInvitationAddr.build` class method:
```python
BackendInvitationAddr.build(
  backend_addr=backend_addr,
  organization_id=organization_id,
  invitation_type=InvitationType.USER,
  token=invitation.token
)
```

parsec/backend/invite.py
```diff
@@ -201,8 +203,16 @@ def generate_invite_email(
             use_ssl=backend_addr.use_ssl,
         )
     )
+    http_invite_link = parsec_invite_link.replace("parsec://", "")
```
touilleMan, line 206: This cooking is too fragile (for instance what if `http_invite_link` contains multiple `/` ?)

Ideally we should create a new class `BackendHTTPRedirectionAddr`, so we could do something like:
```python
invite_addr = BackendInvitationAddr.build(...)
http_redirection_addr = BackendHTTPRedirectionAddr.build(
    invite_addr=invite_addr
)
```

or we could even add a helper method in `BackendInvitationAddr` to simplify this:
```python
invite_addr = BackendInvitationAddr.build(...)
http_redirection_addr = invite_addr.to_http_redirection_addr()
```

The trouble is `BackendAddr` (the base class all addr classes inherit from) consider the scheme is always `parsec://`, so modifying this would be hazardous (and implementing a `BackendHTTPRedirectionAddr` without making it inherit from `BackendAddr` means copy/pasting a long non-trivial code).
In the end I don't think it is worth it given this redirection url is not used a lot.

All in all, I think we should only add the `BackendInvitationAddr.to_http_redirection_url()` method which would return the redirection as a `str` (note the naming: we call `addr` the object and `url` the string representation, hence the `BacendInvitationAddr.to_url()` method for instance).

And of course the `to_http_redirection_url()` should use `urlsplit`/`urlunsplit` to cook the url (hence the complex corner cases should be taken care of by the stdlib instead of ourselves ^^)

parsec/backend/invite.py
```diff
@@ -201,8 +203,16 @@ def generate_invite_email(
...
+    split_invite_link = http_invite_link.split("/")
+    http_invite_link = "http://" + split_invite_link[0] + "/api/redirect/" + split_invite_link[1]
     body = [
-        line.format(sender_name=sender_name, invite_link=invite_link, parsec_url=parsec_url)
+        line.format(
+            sender_name=sender_name,
+            http_invite_link=http_invite_link,
+            parsec_invite_link=parsec_invite_link,
```
touilleMan, line 213: I thin it's better to do the `invite_addr.to_url()` here instead sooner (hence we have access in this function of the addr object which provide useful methods instead of just the string url)

parsec/backend/invite.py
```diff
@@ -201,8 +203,16 @@ def generate_invite_email(
             use_ssl=backend_addr.use_ssl,
         )
     )
+    http_invite_link = parsec_invite_link.replace("parsec://", "")
+    split_invite_link = http_invite_link.split("/")
+    http_invite_link = "http://" + split_invite_link[0] + "/api/redirect/" + split_invite_link[1]
```
touilleMan, line 208: How do we now it's `https` ?

(We should use the `backend.config.backend_addr` variable to know which scheme to use)

parsec/backend/invite.py
Sithi5: With something like that : 
`class BackendHTTPRedirectionAddr(BackendActionAddr):
    def __init__(
        self,
        invitation_type: InvitationType,
        token: UUID,
        **kwargs,
    ):
        if not hostname.startswith("parsec://"):
            hostname = "parsec://" + hostname
        super().__init__(**kwargs)
        self._organization_id = organization_id
        self._invitation_type = invitation_type
        self._token = token`
it should be alright to inherit from BackendAddr no ?
Because it will be more convenient for me to create this class for my PR https://github.com/Scille/parsec-cloud/pull/1294

Sithi5: I'm actually thinking about creating the `BackendHTTPRedirectionAddr`, with inheritance from `BackendInvitationAddr` or from `BackendActionAddr` (in the second case, I will duplicate a big part of BackendinvitationAddr, but both part will be independant in the future...). I don't think that BackendAddr considering `parsec://` scheme is a problem : cf `BackendInvitationAddr.to_http_redirection_url()`

Sithi5: Nevermind, I have no use of the class in the other PR finally ^^

parsec/backend/invite.py
```diff
@@ -191,18 +193,19 @@ def generate_invite_email(
...
-            invitation_type=InvitationType.USER,
-            token=invitation.token,
-            hostname=backend_addr.hostname,
-            port=backend_addr.port,
-            use_ssl=backend_addr.use_ssl,
-        )
+    backend_invitation_addr = BackendInvitationAddr.build(
+        backend_addr=backend_addr,
+        organization_id=organization_id,
+        invitation_type=InvitationType.USER,
```
touilleMan, line 199: `generate_invite_email` is used for both user and device invitation (the signature of the invitation parameter is wrong, it should be `invitation: Invitation`)

Hence we should deduce the type of the invitation from the `invitation` parameter:
```python
invitation_type=InvitationType.USER if isinstance(invitation, UserInvitation) else InvitationType.DEVICE
```

However I think it would be better to add a `UserInvitation.type` and `DeviceInvitation.type` constant attribute (the if/else sentence is cumbersome and error prone if another kind of invitation get introduced in the future):
```python
class UserInvitation:
    TYPE = InvitationType.USER  # Note this is a class attribute (not to be confounded with the instance attributes declared in a similar fashion with class decorated with `@attr.s`)
```

parsec/backend/invite.py
```diff
@@ -191,18 +193,19 @@ def generate_invite_email(
...
+        token=invitation.token,
     )
     body = [
-        line.format(sender_name=sender_name, invite_link=invite_link, parsec_url=parsec_url)
+        line.format(
+            sender_name=sender_name,
+            http_invite_link=backend_invitation_addr.to_http_redirection_url(),
```
touilleMan, line 205: image

parsec/backend/invite.py
Sithi5: Done 👍

Issue #1141, opened 2020-04-29
RC channel for Windows updater


c4ffein:
As we want our employees to test the pre-releases before publishing them as stable, we should make it available to them through an automatic update channel. Actually, the windows updater checks for pre-release status on GitHub, and only includes stable releases. We should include a settings to allow for automatic install of pre-release. Making it available through the GUI isn't mandatory though : we could only make it settable through config files as it is something interesting for our employees, or other developers.

PR #1324, opened 2020-07-30
Added optional RC channel updater

c4ffein:


c4ffein:
Implements #1141

parsec/core/gui/new_version.py
```diff
@@ -47,28 +55,32 @@ def _fetch_json_releases():
...
-            return (latest_from_head, url)
+            return latest_from_head, url
 
-        latest_version = (0, 0, 0)
+        latest_version = Version("0.0.0")
         latest_url = ""
 
-        for release in json_releases:
-            try:
+        try:
```
touilleMan, line 63: Why the switch between try and for ? given the except does a return I don't see the difference...

tests/core/gui/test_new_version.py
```diff
@@ -399,4 +408,64 @@ def test_windows_update_newest_announced_is_draft():
...
+    assert smallcheck() == ((1, 9, 0), gui_check_version_url)
+
+
+@patch("parsec.core.gui.new_version.__version__", new="1.7.0")
+@patch(
+    "parsec.core.gui.new_version.urlopen",
+    new=urlopener(
+        head_version="1.9.0",
+        api_json={
+            "v1.9.0-rc3": {"draft": False, "prerelease": False, "32": True, "64": True},
```
touilleMan, line 420: is it expected to have `prerelease: False` for a rc release ? (and also `32: True, 64: True` ?)

parsec/core/cli/run.py
```diff
@@ -22,12 +22,15 @@
     # Let the GUI handle the parsing of the url to display dialog on error
     @click.argument("url", required=False)
     @click.option("--diagnose", "-d", is_flag=True)
+    @click.option("--rc-update", "-r", is_flag=True)
```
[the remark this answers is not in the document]
c4ffein: Removed

parsec/core/config.py
```diff
@@ -130,6 +132,7 @@ def config_factory(
         gui_first_launch=gui_first_launch,
         gui_last_version=gui_last_version,
         gui_check_version_at_startup=gui_check_version_at_startup,
+        gui_rc_update_enabled=gui_rc_update_enabled,
```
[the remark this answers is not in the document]
c4ffein: Right, changed that

parsec/core/gui/new_version.py
```diff
@@ -17,25 +17,33 @@
 from parsec.core.gui.ui.new_version_info import Ui_NewVersionInfo
 from parsec.core.gui.ui.new_version_available import Ui_NewVersionAvailable
 
+from misc.releaser import Version, RELEASE_REGEX
```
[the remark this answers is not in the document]
c4ffein: Went for the cool kid, works well!

parsec/core/gui/new_version.py
c4ffein: As it was finally decided to make a return at the first json inconsistency, as it means that data can't be trusted and we should redirect the user to the latest version in any case, we might as well put the iteration on `json_releases` in the try block

c4ffein: Also, it covers the highly unlikely case where `json_releases` is not iterable

tests/core/gui/test_new_version.py
c4ffein: No, it is expected to have `prerelease: True` for a prerelease, from RC to alpha. But as doing a manual release as obligated by hardware signing is a boring, simple process, it is also expected that we might end up forgetting to set that tiny checkbox at the end of it one day or another. So this PR is made to work whether we forget that checkbox or not, and we also test that :)

parsec/core/gui/new_version.py
touilleMan: 👍 

(note we should also add a `except  JSONDecodeError` in `_fetch_json_releases` to protect against json inconsistency)

tests/core/gui/test_new_version.py
touilleMan: Ha ! I didn't read the test's name 😄

parsec/core/gui/new_version.py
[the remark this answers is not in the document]
touilleMan: I think we should also add packaging in the `requirements.txt`

parsec/core/gui/new_version.py
c4ffein: Also done

parsec/core/gui/new_version.py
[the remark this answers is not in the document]
c4ffein: Also done

PR #1341, opened 2020-08-04
Invite an user that is already member of the organization is now forbiden

vamonte:


tests/backend/invite/test_base.py
```diff
@@ -337,6 +337,14 @@ async def test_delete(
             pass
 
 
+@pytest.mark.trio
+async def test_user_invitation_already_member(alice, bob, backend, alice_backend_sock):
+    rep = await invite_new(
+        alice_backend_sock, type=InvitationType.USER, claimer_email="bob@example.com"
```
AntoineDupre, line 343: maybe get `claimer_email` from bob human_handle:
```python
    rep = await invite_new(
         alice_backend_sock, type=InvitationType.USER, claimer_email=bob.human_handle.email
    )
```

parsec/backend/postgresql/user_queries/find.py
```diff
@@ -88,6 +108,32 @@ async def query_find(
     return results, len(all_results)
 
 
+@query()
+async def query_retrieve_human(
+    conn, organization_id: OrganizationID, arguments: List[Tuple[str, str, str]]
+) -> HumanFindResultItem:
```
AntoineDupre, line 114: It seems that in can return None, so the return type shall be `Optional[HumanFindResultItem]`

parsec/backend/postgresql/user_queries/find.py
```diff
@@ -88,6 +108,32 @@ async def query_find(
...
+        q = _q_retrieve_human_factory(factory_arguments)
+        args = q(**kwargs)
+        result = await conn.fetchrow(*args)
+        if result:
+            return HumanFindResultItem(
+                user_id=UserID(result["user_id"]),
+                human_handle=HumanHandle(email=result["email"], label=result["label"]),
+                revoked=result["revoked"],
+            )
+    return None
```
AntoineDupre, line 134: It this statement is there to be explicit ?

tests/core/gui/users_widget/test_invite.py
```diff
@@ -39,6 +39,18 @@ def _new_invitation_displayed():
         await aqtbot.wait_until(_new_invitation_displayed)
         assert not autoclose_dialog.dialogs
 
+        monkeypatch.setattr(
+            "parsec.core.gui.users_widget.get_text_input", lambda *args, **kwargs: "bob@example.com"
```
AntoineDupre, line 43: maybe use bob human_handle to get bob's mail

parsec/backend/invite.py
```diff
@@ -249,6 +249,8 @@ def _to_http_redirection_url(client_ctx, invitation):
         if msg["type"] == InvitationType.USER:
             if client_ctx.profile != UserProfile.ADMIN:
                 return invite_new_serializer.rep_dump({"status": "not_allowed"})
+            if await self.user_already_member(client_ctx.organization_id, msg["claimer_email"]):
```
touilleMan, line 252: I think we shouldn't have this separated from `new_for_user` given we cannot guarantee atomicity this way

(hence `new_for_user` should raise a new kind of exception to signify the user already exists)

On top that this would remove the need for the `user_already_member` & `retrieve_human` methods (and there memory/postgresql implementations and the abstract declaration) so it should globally simplify this PR ;-)

parsec/backend/postgresql/invite.py
```diff
@@ -462,9 +462,15 @@ async def _do_new_user_invitation(
...
 class PGInviteComponent(BaseInviteComponent):
-    def __init__(self, dbh: PGHandler, *args, **kwargs):
+    def __init__(self, dbh: PGHandler, user_component: PGUserComponent, *args, **kwargs):
         super().__init__(*args, **kwargs)
         self.dbh = dbh
+        self.user_component = user_component
+
+    async def user_already_member(self, organization_id: OrganizationID, email: str) -> bool:
+        async with self.dbh.pool.acquire() as conn, conn.transaction():
+            r = await self.user_component.retrieve_human(organization_id, email)
```
touilleMan, line 472: note here we acquire a connection and create a transaction... then call `self.user_component.retrieve_human` that itself acquire a connection and create a transaction !

We should instead use the `query_retrieve_human` function that takes a conn object and is expected to be executed in an existing transaction context if needed.

parsec/backend/postgresql/user_queries/find.py
```diff
@@ -88,6 +108,32 @@ async def query_find(
     return results, len(all_results)
 
 
+@query()
+async def query_retrieve_human(
+    conn, organization_id: OrganizationID, arguments: List[Tuple[str, str, str]]
```
touilleMan, line 113: This `arguments` param seems far to broad for it own good !

It would be better to have multiple query with each the exact parameters needed (this improve the readability of the caller code, simplify the implementation of this function and remove the need for the `_q_retrieve_human_factory` or at least for the need to call it dynamically with a lru cache !)

parsec/backend/postgresql/user_queries/find.py
```diff
@@ -88,6 +108,32 @@ async def query_find(
...
+    if len(arguments) == 0:
+        return
+
+    kwargs = {"organization_id": organization_id, "now": pendulum_now()}
+    factory_arguments = tuple()
+    for (field, operator, value) in arguments:
+        key = field.replace(".", "_")
+        factory_arguments += (field, operator, f"${key}")
+        kwargs[key] = value
+    else:
```
touilleMan, line 124: why the `else` here (given there is no `break` in the for, we always go into the `else`)

parsec/backend/postgresql/invite.py
```diff
@@ -474,10 +475,13 @@ async def new_for_user(
         created_on: Optional[Pendulum] = None,
     ) -> UserInvitation:
         """
-        Raise: Nothing
+        Raise: InvitationAlreadyMemberError
         """
         created_on = created_on or pendulum_now()
         async with self.dbh.pool.acquire() as conn, conn.transaction():
+            human = await query_retrieve_active_human_by_email(conn, organization_id, claimer_email)
+            if human and not human.revoked:
```
touilleMan, line 483: given we retrieve an active human, check on `human.revoked` seems redundant

parsec/backend/postgresql/user_queries/find.py
```diff
@@ -1,14 +1,28 @@
...
 from parsec.backend.postgresql.utils import query
 from parsec.backend.postgresql.queries import Q, q_organization_internal_id
 
+_q_retrieve_active_human_by_email = Q(
+    f"""
+SELECT
+    user_.user_id,
+    human.email,
+    human.label,
+    user_.revoked_on IS NOT NULL AND user_.revoked_on <= $now as revoked
```
touilleMan, line 17: Given we return the "active" human if any, we should move this line on the filder part of the query and modify it to be:
```sql
user_.revoked_on IS NULL OR user_.revoked_on > $now
```

parsec/backend/postgresql/user_queries/find.py
```diff
@@ -88,6 +102,23 @@ async def query_find(
     return results, len(all_results)
 
 
+@query()
+async def query_retrieve_active_human_by_email(
+    conn, organization_id: OrganizationID, email: str
+) -> Optional[HumanFindResultItem]:
```
touilleMan, line 108: There could be multiple user with the same human handle (but at most one which is not revoked).

Given the query currently doesn't filter on the revoked field (but only return a boolean is_revoked), I guess we can have trouble in case we have multiple user with the same human handle.

This should be solved by the changes I proposed on the query (i.e. filter on revoked and so return only zero on one row) 😃

parsec/backend/postgresql/user_queries/find.py
```diff
@@ -88,6 +102,23 @@ async def query_find(
...
+    result = await conn.fetchrow(
+        *_q_retrieve_active_human_by_email(
+            organization_id=organization_id, now=pendulum_now(), email=email
+        )
+    )
+    if result:
+        return HumanFindResultItem(
```
touilleMan, line 115: Though I think it's a good idea to introduce new classes to represent complex values, I think here we could avoid doing this now that we have removed the is_revoked field (and hence we can simplify by returning only a `Optional[UserID]` which would be enough for our current needs)

tests/backend/invite/test_base.py
```diff
@@ -337,6 +337,14 @@ async def test_delete(
             pass
 
 
+@pytest.mark.trio
+async def test_user_invitation_already_member(alice, bob, backend, alice_backend_sock):
```
touilleMan, line 341: would be good to add another similar test (or to parametrize this one ^^) that create another revoked user with similar human handle to make sure the invitation is allowed in such case

parsec/backend/memory/invite.py
```diff
@@ -173,6 +174,14 @@ async def new_for_user(
         claimer_email: str,
         created_on: Optional[Pendulum] = None,
     ) -> UserInvitation:
+        """
+        Raise: InvitationAlreadyMemberError
+        """
+        org = self._user_component._organizations[organization_id]
+        for _, user in org.users.items():
+            is_revoked = user.revoked_on and user.revoked_on <= pendulum_now()
```
touilleMan, line 182: don't we already have a `user.is_revoked` property to do this check ? (I can't retrieve it in the codebase but I thought I saw something like this... @AntoineDupre any idea ?)

AntoineDupre: @touilleMan I don't think so. I found that we test if a user is revoked in different places in the same way (like [there](https://github.com/Scille/parsec-cloud/blob/master/parsec/backend/handshake.py#L112) )

It can be nice to added it as there is an `is_expired` for the organization:
```python
@attr.s(slots=True, frozen=True, auto_attribs=True)
class User:
 ...
    @property
    def is_revoked(self):
        return user.revoked_on and user.revoked_on <= pendulum_now()
```

touilleMan: @AntoineDupre yep it would be good to add this then

btw I think a method is better than a property here given it result depends of side effects (so doing `is_revoked = user.is_revoked [...] if is_revoked [...] if is_revoked [...]` is not equivalent to `if user.is_revoked [...] if user.is_revoked [...]`)

parsec/backend/postgresql/invite.py
```diff
@@ -474,10 +475,16 @@ async def new_for_user(
...
         """
-        Raise: Nothing
+        Raise: InvitationAlreadyMemberError
         """
         created_on = created_on or pendulum_now()
         async with self.dbh.pool.acquire() as conn, conn.transaction():
+            user_id = await query_retrieve_active_human_by_email(
+                conn, organization_id, claimer_email
+            )
+            print(user_id)
```
touilleMan, line 485: debug print to remove ;-)

tests/backend/invite/test_base.py
```diff
@@ -337,6 +338,25 @@ async def test_delete(
             pass
 
 
+@pytest.mark.trio
+@pytest.mark.parametrize("is_revoked", [True, False])
+async def test_user_invitation_already_member(alice, bob, backend, alice_backend_sock, is_revoked):
+    with patch.object(User, "is_revoked") as is_revoked_mock:
```
touilleMan, line 344: I don't understand why the mocks are needed (and mocking the postgresql implementation seems scary given we remove a whole query...)

I think it should be enough to just use the bob fixture (so we have an existing user) then revoke (if the parametrize says so) at the beginning of the test.

parsec/core/invite/exceptions.py
```diff
@@ -20,6 +20,11 @@ def __init__(self, msg="Invitation already used"):
         super().__init__(msg)
 
 
+class InviteAlreadyMemberError(InviteError):
+    def __init__(self, msg="Invitation is already a member"):
```
touilleMan, line 24: `Invited user is already part of the organization` ?

touilleMan (approved):

Issue #1330, opened 2020-07-31
Make offline mode more obvious to the users


vxgmichel:
@tleblond reported that users don't necessarily realize that the reason why a document fails to open is because the system is offline at the moment. This could be solved by making offline mode more obvious to the users.

stefan-contiu:
Stress the case when the user is not in the GUI.

AntoineDupre:
Fixed in PR #1342 

Now a systray notification is sent when one logged in organization backend is disconnected. This makes the disconnection obvious even if the parsec client is not directly in used. (It does not wok for user that logged in through the CLI)

PR #1342, opened 2020-08-05
Display systray notification on backend disconnected event

AntoineDupre:
Display systray notification to make offline mode more obvious to the users

#1330

### DIFF
--- a/parsec/core/gui/central_widget.py
+++ b/parsec/core/gui/central_widget.py
@@ -206,7 +206,9 @@ class CentralWidget(QWidget, Ui_CentralWidget):
         if disconnected:
             self.systray_notification.emit(
                 "Parsec",
-                f"{self.core.device.organization_id}:" + _("TEXT_BACKEND_STATE_DISCONNECTED"),
+                _("TEXT_SYSTRAY_BACKEND_DISCONNECT_organization").format(
+                    organization=self.core.device.organization_id
+                ),
                 5000,
             )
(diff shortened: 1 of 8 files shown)


parsec/core/gui/central_widget.py
```diff
@@ -194,10 +198,17 @@ def _on_connection_state_changed(self, status, status_exc):
...
             notif = ("ERROR", tooltip)
+            disconnected = True
 
         self.menu.set_connection_state(text, tooltip, icon)
         if notif:
             self.new_notification.emit(*notif)
+        if disconnected:
+            self.systray_notification.emit(
+                "Parsec",
+                f"{self.core.device.organization_id}:" + _("TEXT_BACKEND_STATE_DISCONNECTED"),
```
Max-7, line 209: The formatting here is done mostly in the code. I would rather we let the translator do all the formatting (as much as possible) with something like:
```python
_("TEXT_SYSTRAY_BACKEND_DISCONNECT_organization").format(organization=self.core.device.organization_id)
```
which allows us to use `{organization}` in the translation entry in the PO file.

AntoineDupre: It's better, i defined it like that in PO:
```python
 msgid "TEXT_SYSTRAY_BACKEND_DISCONNECT_organization"                                                                                                                      
 msgstr "{organization}: Disconnected" 
```

PR #1248, opened 2020-07-07
Mypy on core fs

AntoineDupre:
Add mypy check to parsec.core.fs

PR #1358, opened 2020-08-10
Make author field mandatory for all data classes

vxgmichel:
... except for `UserCertificateContent` and `DeviceCertificateContent`

To merge after #1248

### DIFF
--- a/tests/core/fs/storage/test_workspace_storage.py
+++ b/tests/core/fs/storage/test_workspace_storage.py
@@ -5,6 +5,7 @@ from pathlib import Path
 import pytest
 from pendulum import now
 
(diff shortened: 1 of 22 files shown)


tests/core/fs/storage/test_workspace_storage.py
```diff
@@ -19,13 +20,14 @@
 )
 
 
-def create_manifest(device, type=LocalWorkspaceManifest):
+def create_manifest(device, type=LocalWorkspaceManifest, legacy=False):
+    author = None if legacy else device.device_id
```
touilleMan, line 24: a comment about what this legacy option is about would be interesting here (and maybe even rename `legacy` into `legacy_base_manifest_author_is_none`)

PR #1359, opened 2020-08-11
Add support for confined and filtered (temporary) files and directories

### LINKED ISSUE #990: Provide file sync filter config to ignore tilde lock files

vxgmichel:
Fix #990

Follow-up on #1252, closed by mistake

Add support for confined and filtered (i.e temporary) files and directories.
In this context:
- confined entries: local files and directories that are not meant to be synchronized with other clients
- filtered entries: remote files and directories synced by other clients that should not appear locally

It includes:
- [x] Add confined entry set to local manifest
- [x] Add filtered entry set to local manifest (filtering entries from other clients)
- [x] Define a confinement pattern list
- [x] Add dedicated icons in the GUI for confined files
- [x] Update sync monitor to manage the syncing of no-longer-confined entries
- [x] Implement the procedure for changing filter pattern
- [x] Make the pattern list configurable (optional)

### DIFF
--- a/parsec/core/config.py
+++ b/parsec/core/config.py
@@ -54,8 +54,8 @@ class CoreConfig:
     data_base_dir: Path
     cache_base_dir: Path
     mountpoint_base_dir: Path
+    prevent_sync_pattern_path: Optional[Path] = None  # Use `default_pattern.ignore` by default
     preferred_org_creation_backend_addr: BackendAddr
-
     debug: bool = False
 
     backend_max_cooldown: int = 30
@@ -82,6 +82,7 @@ class CoreConfig:
     gui_confirmation_before_close: bool = True
     gui_workspace_color: bool = False
     gui_allow_multiple_instances: bool = False
+    gui_show_confined: bool = False
 
     ipc_socket_file: Path = None
     ipc_win32_mutex_name: str = "parsec-cloud"
@@ -95,6 +96,7 @@ def config_factory(
     data_base_dir: Path = None,
     cache_base_dir: Path = None,
     mountpoint_base_dir: Path = None,
+    prevent_sync_pattern_path: Optional[Path] = None,
     mountpoint_enabled: bool = False,
     disabled_workspaces: FrozenSet[EntryID] = frozenset(),
     backend_max_cooldown: int = 30,
@@ -112,6 +114,7 @@ def config_factory(
     gui_workspace_color: bool = False,
     gui_allow_multiple_instances: bool = False,
     preferred_org_creation_backend_addr: Optional[BackendAddr] = None,
+    gui_show_confined: bool = False,
     environ: dict = {},
     **_,
 ) -> CoreConfig:
@@ -132,6 +135,7 @@ def config_factory(
         data_base_dir=data_base_dir,
         cache_base_dir=cache_base_dir or get_default_cache_base_dir(environ),
         mountpoint_base_dir=get_default_mountpoint_base_dir(environ),
+        prevent_sync_pattern_path=prevent_sync_pattern_path,
         mountpoint_enabled=mountpoint_enabled,
         disabled_workspaces=disabled_workspaces,
         backend_max_cooldown=backend_max_cooldown,
@@ -150,6 +154,7 @@ def config_factory(
         gui_workspace_color=gui_workspace_color,
         gui_allow_multiple_instances=gui_allow_multiple_instances,
         preferred_org_creation_backend_addr=preferred_org_creation_backend_addr,
+        gui_show_confined=gui_show_confined,
         ipc_socket_file=data_base_dir / "parsec-cloud.lock",
         ipc_win32_mutex_name="parsec-cloud",
     )
@@ -193,6 +198,11 @@ def load_config(config_dir: Path, **extra_config) -> CoreConfig:
         pass
 
     try:
+        data_conf["prevent_sync_pattern_path"] = Path(data_conf["prevent_sync_pattern_path"])
+    except (KeyError, ValueError):
+        pass
+
+    try:
         data_conf["disabled_workspaces"] = frozenset(map(EntryID, data_conf["disabled_workspaces"]))
     except (KeyError, ValueError):
         pass
@@ -231,6 +241,7 @@ def save_config(config: CoreConfig):
             {
                 "data_base_dir": str(config.data_base_dir),
                 "cache_base_dir": str(config.cache_base_dir),
+                "prevent_sync_pattern": str(config.prevent_sync_pattern_path),
                 "telemetry_enabled": config.telemetry_enabled,
                 "disabled_workspaces": list(map(str, config.disabled_workspaces)),
                 "backend_max_cooldown": config.backend_max_cooldown,
@@ -245,6 +256,7 @@ def save_config(config: CoreConfig):
                 "gui_workspace_color": config.gui_workspace_color,
                 "gui_allow_multiple_instances": config.gui_allow_multiple_instances,
                 "preferred_org_creation_backend_addr": config.preferred_org_creation_backend_addr.to_url(),
+                "gui_show_confined": config.gui_show_confined,
             },
             indent=True,
         )
--- a/parsec/core/fs/storage/manifest_storage.py
+++ b/parsec/core/fs/storage/manifest_storage.py
@@ -1,8 +1,10 @@
 # Parsec Cloud (https://parsec.cloud) Copyright (c) AGPLv3 2019 Scille SAS
 
+import re
+
 import trio
 from structlog import get_logger
-from typing import Dict, Tuple, Set, Optional, Union
+from typing import Dict, Tuple, Set, Optional, Union, Pattern
 from async_generator import asynccontextmanager
 
 from parsec.core.fs.exceptions import FSLocalMissError
@@ -10,6 +12,8 @@ from parsec.core.types import EntryID, ChunkID, LocalDevice, BaseLocalManifest, 
 from parsec.core.fs.storage.local_database import LocalDatabase
 
 logger = get_logger()
+
+EMPTY_PATTERN = r"^\b$"  # Do not match anything (https://stackoverflow.com/a/2302992/2846140)
 
 
 class ManifestStorage:
@@ -87,6 +91,58 @@ class ManifestStorage:
                   checkpoint INTEGER NOT NULL
                 );
                 """
+            )
+            # Singleton storing the prevent_sync_pattern
+            cursor.execute(
+                """
+                CREATE TABLE IF NOT EXISTS prevent_sync_pattern
+                (
+                  _id INTEGER PRIMARY KEY NOT NULL,
+                  pattern TEXT NOT NULL,
+                  fully_applied INTEGER NOT NULL  -- Boolean
+                );
+                """
+            )
+            # Set the default "prevent sync" pattern if it doesn't exist
+            cursor.execute(
+                """
+                INSERT OR IGNORE INTO prevent_sync_pattern(_id, pattern, fully_applied)
+                VALUES (0, ?, 0)""",
+                (EMPTY_PATTERN,),
+            )
+
+    # "Prevent sync" pattern operations
+
+    async def get_prevent_sync_pattern(self) -> Tuple[Pattern, bool]:
+        async with self._open_cursor() as cursor:
+            cursor.execute("SELECT pattern, fully_applied FROM prevent_sync_pattern WHERE _id = 0")
+            reply = cursor.fetchone()
+            pattern, fully_applied = reply
+            return (re.compile(pattern), bool(fully_applied))
+
+    async def set_prevent_sync_pattern(self, pattern: Pattern):
+        """Set the "prevent sync" pattern for the corresponding workspace
+
+        This operation is idempotent,
+        i.e it does not reset the `fully_applied` flag if the pattern hasn't changed.
+        """
+        async with self._open_cursor() as cursor:
+            cursor.execute(
+                """UPDATE prevent_sync_pattern SET pattern = ?, fully_applied = 0 WHERE _id = 0 AND pattern != ?""",
+                (pattern.pattern, pattern.pattern),
+            )
+
+    async def mark_prevent_sync_pattern_fully_applied(self, pattern):
+        """Mark the provided pattern as fully applied.
+
+        This is meant to be called after one made sure that all the manifests in the
+        workspace are compliant with the new pattern. The applied pattern is provided
+        as an argument in order to avoid concurrency issues.
+        """
+        async with self._open_cursor() as cursor:
+            cursor.execute(
+                """UPDATE prevent_sync_pattern SET fully_applied = 1 WHERE _id = 0 AND pattern = ?""",
+                (pattern.pattern,),
             )
 
     # Checkpoint operations
--- a/parsec/core/fs/storage/workspace_storage.py
+++ b/parsec/core/fs/storage/workspace_storage.py
@@ -2,7 +2,7 @@
 
 from pathlib import Path
 from collections import defaultdict
-from typing import Dict, Tuple, Set, Optional, Union, AsyncIterator, NoReturn
+from typing import Dict, Tuple, Set, Optional, Union, AsyncIterator, NoReturn, Pattern
 
 import trio
 from trio import hazmat
@@ -63,6 +63,11 @@ class BaseWorkspaceStorage:
         # Manifest and block storage
         self.block_storage = block_storage
         self.chunk_storage = chunk_storage
+
+        # Pattern attributes
+        # Set by `_load_prevent_sync_pattern` in WorkspaceStorage.run()
+        self._prevent_sync_pattern: Pattern
+        self._prevent_sync_pattern_fully_applied: bool
 
     def _get_next_fd(self) -> FileDescriptor:
         self.fd_counter += 1
@@ -152,6 +157,14 @@ class BaseWorkspaceStorage:
             if not miss_ok:
                 raise
 
+    # "Prevent sync" pattern interface
+
+    def get_prevent_sync_pattern(self) -> Pattern:
+        return self._prevent_sync_pattern
+
+    def get_prevent_sync_pattern_fully_applied(self) -> bool:
+        return self._prevent_sync_pattern_fully_applied
+
 
 class WorkspaceStorage(BaseWorkspaceStorage):
     """Manage the access to the local storage.
@@ -213,7 +226,7 @@ class WorkspaceStorage(BaseWorkspaceStorage):
                         async with ChunkStorage.run(device, data_localdb) as chunk_storage:
 
                             # Instanciate workspace storage
-                            yield cls(
+                            instance = cls(
                                 device,
                                 path,
                                 workspace_id,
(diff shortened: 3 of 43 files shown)


parsec/core/types/manifest.py
```diff
@@ -450,22 +457,144 @@ def match_remote(self, remote_manifest: RemoteFileManifest) -> bool:
...
         @post_load
         def make_obj(self, data):
             data.pop("type")
+            data.setdefault("confined_entries", frozenset())
+            data.setdefault("filtered_entries", frozenset())
             return LocalFolderManifest(**data)
 
     base: RemoteFolderManifest
     children: FrozenDict[EntryName, EntryID]
+    confined_entries: FrozenSet[EntryID]
```
touilleMan, line 596: Should put a comment here about the difference between confined and filtered entries

parsec/core/types/manifest.py
```diff
@@ -500,25 +632,37 @@ def to_stats(self):
...
-
     # Remote methods
 
     @classmethod
-    def from_remote(cls, remote: RemoteFolderManifest) -> "LocalFolderManifest":
-        return cls(base=remote, need_sync=False, updated=remote.updated, children=remote.children)
+    def from_remote(
+        cls,
+        remote: RemoteFolderManifest,
+        pattern_filter: Pattern,
```
touilleMan, line 641: pattern mandatory here ?

parsec/core/types/manifest.py
```diff
@@ -575,31 +728,45 @@ def to_stats(self):
...
-
     # Remote methods
 
     @classmethod
-    def from_remote(cls, remote: RemoteWorkspaceManifest) -> "LocalWorkspaceManifest":
-        return cls(base=remote, need_sync=False, updated=remote.updated, children=remote.children)
+    def from_remote(
+        cls,
+        remote: RemoteFolderManifest,
+        pattern_filter: Pattern,
```
touilleMan, line 737: pattern mandatory here ?

parsec/core/types/manifest.py
```diff
@@ -266,14 +266,19 @@ def evolve_and_mark_updated(self: LocalManifestTypeVar, **data) -> LocalManifest
 
     @classmethod
     def from_remote(
-        cls: Type[LocalManifestTypeVar], remote: BaseRemoteManifest
+        cls: Type[LocalManifestTypeVar],
+        remote: BaseRemoteManifest,
+        pattern_filter: Optional[Pattern] = None,
+        local_manifest: Optional["BaseLocalManifest"] = None,
```
touilleMan, line 272: It's kind of weird to create a `BaseLocalManifest` while providing another `BaseLocalManifest`... (this is especially a trouble given we use only part of the provided `local_manifest`, without explicitly saying so in the method signature)

It would be better to use `evolve` (or to create a specialized `evolve_...` method), but I guess the trouble is we are in a kind of middle ground here given we keep keep only part of the local_manifest (so the `evolve_...` method would reset the `need_sync` and `children` attributes, so not really an evolve...)

In the end I think the trouble is we are trying to put part of the local/remote merge logic in the manifest class... wouldn't it be better to only use `from_remote `/ `evolve` and keep the more complex logic in the `sync_transaction.merge_manifests` function ?

parsec/core/fs/storage/manifest_storage.py
```diff
@@ -88,6 +92,47 @@ async def _create_db(self):
                 );
                 """
             )
+            # Singleton storing the pattern_filter
+            cursor.execute(
```
touilleMan, line 96: We don't have a migration system for the sqlite database. I guess we are lucky this time given we can work around with this `CREATE TABLE IF NOT EXISTS` ;-)

parsec/core/fs/storage/manifest_storage.py
```diff
@@ -88,6 +92,47 @@ async def _create_db(self):
...
+        async with self._open_cursor() as cursor:
+            cursor.execute("SELECT pattern, fully_applied FROM pattern_filter WHERE _id = 0")
+            reply = cursor.fetchone()
+            pattern, fully_applied = reply
+            return (re.compile(pattern), bool(fully_applied))
+
+    async def set_pattern_filter(self, pattern: Pattern):
+        async with self._open_cursor() as cursor:
+            cursor.execute(
+                """UPDATE pattern_filter SET pattern = ?, fully_applied = 0 WHERE _id = 0 AND pattern != ?""",
```
touilleMan, line 126: what the need for the `AND pattern != ?` ? Is it because we can avoid a more costly write this way (and `pattern == ?` is often true)

parsec/core/fs/storage/manifest_storage.py
```diff
@@ -11,6 +13,8 @@
 
 logger = get_logger()
 
+EMPTY_PATTERN = r"^/b$"  # Do not match anything
```
touilleMan, line 16: PR author hates him: a young programmer found a pattern that match the "do not match anything" regex !
```python
>>> re.match(r'^/b$', '/b')
<_sre.SRE_Match object; span=(0, 2), match='/b'>
```

parsec/core/fs/storage/manifest_storage.py
```diff
@@ -88,6 +92,47 @@ async def _create_db(self):
...
+                  fully_applied INTEGER NOT NULL  -- Boolean
+                );
+                """
+            )
+            # Set the default pattern filter if it doesn't exist
+            cursor.execute(
+                """
+                INSERT OR IGNORE INTO pattern_filter(_id, pattern, fully_applied)
+                VALUES (0, ?, 0)""",
+                (EMPTY_PATTERN,),
```
touilleMan, line 111: I guess it's better to store the empty pattern in database than return it from the code if the SELECT in `get_pattern_filter` returns no row given we have the guarantee the returned pattern is the one that has been used (and not something else given we are running a new version of Parsec where `EMPTY_PATTERN` has changed)

parsec/core/fs/storage/workspace_storage.py
```diff
@@ -152,6 +156,15 @@ async def clear_chunk(self, chunk_id: ChunkID, miss_ok: bool = False) -> None:
             if not miss_ok:
                 raise
 
+    # Pattern filter interface
+
+    def get_pattern_filter(self) -> Pattern:
+        assert self._pattern_filter is not None
```
touilleMan, line 162: `self._pattern_filter` is defined as Optional and is initialized to `None`...

I guess the thing is we have `_load_pattern_filter` that is called after `__init__` which set `_pattern_filter`.
In such case I think we should only declare the type in `__init__`:
```python
def __init__(self, [...]):
    [...]
    self._pattern_filter: Pattern  # Defined by BaseWorkspaceStorage._pattern_filter as part of init

async def run(self, [...]):
    self._pattern_filter = turbo_smurfing_the_stuff()
```

@AntoineDupre you did this kind of trick something in a previous PR but I couldn't remember where...

parsec/core/fs/storage/manifest_storage.py
```diff
@@ -88,6 +92,47 @@ async def _create_db(self):
...
+            return (re.compile(pattern), bool(fully_applied))
+
+    async def set_pattern_filter(self, pattern: Pattern):
+        async with self._open_cursor() as cursor:
+            cursor.execute(
+                """UPDATE pattern_filter SET pattern = ?, fully_applied = 0 WHERE _id = 0 AND pattern != ?""",
+                (pattern.pattern, pattern.pattern),
+            )
+
+    async def set_pattern_filter_fully_applied(self, pattern):
```
touilleMan, line 130: wouldn't it be simpler to just have `async def set_pattern_filter(self, pattern: Pattern, fully_applied: bool=False):` ?

EDIT: after wondering about concurrency issue, I think I understand better the logic here: we have a two steps mechanism
- declare the pattern as changed by calling `set_pattern_filter()`
- then crawl the data to apply it, this is outside of the scope of this api
- finally mark the pattern as fully applied by calling `set_pattern_filter_fully_applied()`, the pattern must be provided here to make sure it hasn't been changed in the meantime (by calling `set_pattern_filter()` again)

If those assumptions are true, I think we should rename a bit this:
`change_pattern_filter()`
`mark_pattern_filter_fully_applied()`
and document how the two must be used together (and how concurrency supported handled)

parsec/core/fs/workspacefs/entry_transactions.py
```diff
@@ -161,8 +170,10 @@ async def entry_info(self, path: FsPath) -> dict:
         self.check_read_rights(path)
 
         # Fetch data
-        manifest = await self._get_manifest_from_path(path)
-        return manifest.to_stats()
+        manifest, confined = await self._get_manifest_from_path(path)
+        stats = manifest.to_stats()
```
touilleMan, line 174: so `to_stats()` is no longer all the stats...

Could it be a trouble for `FileTransactions.fd_info()` (the other place where this `to_stats()` is used) ?

(I would prefer not to have this `to_stats()` method given it is some kind of arbitrary conversion that depend of who is needed them, but I understant it's not that easy here given we have two places that need the stats... and now maybe the-same-stats-but-not-100%-the-same...)

parsec/core/fs/workspacefs/sync_transactions.py
```diff
@@ -240,6 +240,23 @@ async def synchronization_step(
         # Fetch and lock
         async with self.local_storage.lock_manifest(entry_id) as local_manifest:
 
+            # Go through the parent chain
+            current_manifest = local_manifest
+            while not is_workspace_manifest(current_manifest):
+                parent_manifest = await self.local_storage.get_manifest(current_manifest.parent)
```
touilleMan, line 257: So we access each parent manifest here, could it be possible that the manifest is no longer present in the local_storage ? (currently we keep all the manifest in the local storage, but is this a strong guarantee or just an implementation detail that could silently break the code here if changed ?...)

And could it be possible the parent manifest has never been in the local storage ? for instance when using `WorkspaceFS.sync_by_id()` ?

parsec/core/fs/workspacefs/sync_transactions.py
```diff
@@ -240,6 +240,23 @@ async def synchronization_step(
...
+                parent_manifest = cast(LocalFolderishManifests, parent_manifest)
+
+                # The entry is not confined
+                if current_manifest.id not in parent_manifest.confined_entries:
+                    current_manifest = parent_manifest
+                    continue
+
+                # Send synced event
+                self._send_event(
+                    CoreEvent.FS_ENTRY_CONFINED, entry_id=entry_id, cause_id=parent_manifest.id
```
touilleMan, line 267: so this event is triggered every time an entry should have been synchronized but is filtered ? I guess it's useful because we can listen to `FS_ENTRY_SYNCED `and `FS_ENTRY_CONFINED `and get the same behavior as when we used to listen to `FS_ENTRY_SYNCED` alone

parsec/core/sync_monitor.py
```diff
@@ -139,9 +140,19 @@ async def _load_changes(self) -> bool:
 
     def set_local_change(self, entry_id: EntryID) -> bool:
         # Ignore local changes in read only mode
+        wake_up = False
         if self.read_only:
-            return
+            return wake_up
 
+        # Pop confined entries related to current entry_id
+        confined_entries = self._confined_entries.pop(entry_id, ())
```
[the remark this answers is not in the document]
touilleMan: btw we may want to a have a special vocab:
- `confined entry` to refer to a manifest (folder or file) that is confined
- `confinement point` to refer to an entry that name provoke the confinement of itself and it children if any

Using this convention, we can see that synchronizing a confined entry depend on the presence of confinement points in it path.

parsec/core/fs/storage/manifest_storage.py
vxgmichel: Good catch! It's a typo, it should be [`r'^\b$'`](https://stackoverflow.com/a/2302992/2846140)

parsec/core/fs/workspacefs/entry_transactions.py
vxgmichel: b89946e15d95d68291748803f1033db8abc61d05

parsec/core/sync_monitor.py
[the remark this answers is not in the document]
vxgmichel: b89946e15d95d68291748803f1033db8abc61d05

parsec/core/resources/default_pattern.ignore
```diff
@@ -0,0 +1,83 @@
...
+.goutputstream-*
+
+### macOS ###
+
+# General
+.DS_Store
+.AppleDouble
+.LSOverride
+
+# Icon must end with two \r
```
touilleMan, line 38: What does this mean ?

(`Icon` seems like a legit folder name for a regular user...)

parsec/core/resources/default_pattern.ignore
```diff
@@ -0,0 +1,83 @@
...
+._*
+
+# Files that might appear in the root of a volume
+.DocumentRevisions-V100
+.fseventsd
+.Spotlight-V100
+.TemporaryItems
+.Trashes
+.VolumeIcon.icns
+.com.apple.timemachine.donotpresent
```
touilleMan, line 51: thanks they didn't called this shit ` .com.apple.timemachine.dotnotpresent ` :trollface:

parsec/core/logged_core.py
```diff
@@ -40,6 +45,55 @@
...
+
+
+def get_pattern_filter(pattern_filter_path: Optional[Path] = None) -> Pattern:
+    pattern = None
+    # Get the pattern from the path defined in the core config
+    if pattern_filter_path is not None:
+        pattern = _get_pattern_filter(pattern_filter_path)
+    # Default to the pattern from the ignore file in the core resources
+    if pattern is None:
+        with importlib_resources.path(core_resources, "default_pattern.ignore") as path:
```
touilleMan, line 90: I don't think this is zipsafe, we should use `importlib.resources.read_text` instead

parsec/core/logged_core.py
```diff
@@ -40,6 +45,55 @@
...
+        return None
+    try:
+        regex = "|".join(
+            fnmatch.translate(line.strip())
+            for line in data.splitlines()
+            if line.strip() and not line.strip().startswith("#")
+        )
+    except ValueError as exc:
+        logger.warning(
+            f"Could not parse the file containing the filename patterns " f"to ignore: {exc}"
```
touilleMan, line 70: `" f"` due to black

and shouldn't we use `exc_info=exc` to pass the exception ?

parsec/core/logged_core.py
```diff
@@ -40,6 +45,55 @@
...
+        )
+        return None
+    try:
+        return re.compile(regex)
+    except re.error as exc:
+        logger.warning(
+            f"Could not compile the file containing the filename patterns "
+            f"to ignore into a regex pattern: {exc}"
```
touilleMan, line 78: same as above

parsec/core/logged_core.py
```diff
@@ -40,6 +45,55 @@
...
+    try:
+        data = pattern_filter_path.read_text()
+    except OSError as exc:
+        logger.warning(
+            f"Path to the file containing the filename patterns "
+            f"to ignore is not properly defined: {exc}"
+        )
+        return None
+    try:
+        regex = "|".join(
```
touilleMan, line 63: time to use a for loop I guess:
```python
regex = []
for line in data.splitlines():
    line = line.strip()
    if line and not line.startswith("#"):
        regex.append(fnmatch.translate(line))
regex = "|".join(regex)
```

parsec/core/config.py
```diff
@@ -54,8 +54,8 @@ class CoreConfig:
     data_base_dir: Path
     cache_base_dir: Path
     mountpoint_base_dir: Path
+    pattern_filter_path: Optional[Path] = None  # None means: use the default filter path
```
touilleMan, line 57: Wouldn't it be simpler for the core to directly store here the pattern object (and let the configuration loader do the dirty work of retrieving/building the pattern) ?

parsec/core/logged_core.py
vxgmichel: It seems like it is:
> Exiting the context manager cleans up any temporary file created when the resource needs to be extracted from e.g. a zip file.

https://docs.python.org/3/library/importlib.html#importlib.resources.path

parsec/core/logged_core.py
vxgmichel: Good catch!

parsec/core/logged_core.py
vxgmichel: :+1:

parsec/core/logged_core.py
vxgmichel: :+1:

parsec/core/resources/default_pattern.ignore
vxgmichel: No idea, I just assumed it was in the list for a reason. I'll remove it.

parsec/core/config.py
vxgmichel: Does it really make any difference? I wanted to avoid putting too much logic in the configuration module

parsec/core/resources/default_pattern.ignore
vxgmichel: https://stackoverflow.com/a/30755378/2846140

parsec/core/resources/default_pattern.ignore
vxgmichel: :smile:

parsec/core/logged_core.py
vxgmichel: e55ed1ec6d0c085d18bfdc8c964be92f342860b4

parsec/core/logged_core.py
vxgmichel: e55ed1ec6d0c085d18bfdc8c964be92f342860b4

parsec/core/resources/default_pattern.ignore
vxgmichel: e55ed1ec6d0c085d18bfdc8c964be92f342860b4

parsec/core/config.py
touilleMan: I guess we can rework this later if it's too much of a hassle... especially given this PR is far to long for it own good ;-)

In theory, this could be useful to provide a custom filter in test without having to create a file

parsec/core/logged_core.py
touilleMan: That's pretty interesting !

(I'm still not a huge fan of thinking the underlying layers have to create a temp file on the FS so we can access it, but I don't have any serious reason to support this claim :smile: so let's leave this as it is !)

touilleMan (approved):
:+1:

Issue #1309, opened 2020-07-27
Rollback to PyQt 5.14.2 might break Windows production in some environments [investigating]


bitlogik:
Following, the [PyQT5 issue fixing](https://github.com/Scille/parsec-cloud/pull/1308), and the [issues related to the binaries package builds](https://github.com/Scille/parsec-cloud/commit/2fbe93b4c2fa99f8647640b5622587ef00fbae01), we'd like to report a potential issue when you'll build the package for Windows. Full software testing is part of the delivery, but sadly, there's no test dedicated to binary package production, so you mostly discover the issue once the new version is tagged for released.

The Windows installer production script seems to be broken by this PyQT5 rollback.
One of our pipeline was OK for Windows, we just merged 23c939e9d53c4d1e0a1fa6ed70219a68a1ffae3b, and it now fails, as follow :

```
Building wheels for collected packages: parsec-cloud, msgpack, pendulum, PyPika
  Building wheel for parsec-cloud (setup.py): started
  Building wheel for parsec-cloud (setup.py): finished with status 'error'
  Running setup.py clean for parsec-cloud
  ERROR: Command errored out with exit status 1:
  ImportError: DLL load failed while importing sip: The parameter is incorrect.
```

Full log (from the failure point) :
```

Collecting pytz>=0a
  Using cached pytz-2020.1-py2.py3-none-any.whl (510 kB)
  Saved c:\test_user\builds\parsec\parsec-cloud\packaging\win32\build\wheels\pytz-2020.1-py2.py3-none-any.whl
Collecting pycparser
  Downloading pycparser-2.20-py2.py3-none-any.whl (112 kB)
  Saved c:\test_user\builds\parsec\parsec-cloud\packaging\win32\build\wheels\pycparser-2.20-py2.py3-none-any.whl
Skipping attrs, due to already being wheel.
Skipping click, due to already being wheel.
Skipping wsproto, due to already being wheel.
Skipping toastedmarshmallow, due to already being wheel.
Skipping PyNaCl, due to already being wheel.
Skipping trio, due to already being wheel.
Skipping python-interface, due to already being wheel.
Skipping async-generator, due to already being wheel.
Skipping sentry-sdk, due to already being wheel.
Skipping structlog, due to already being wheel.
Skipping importlib-resources, due to already being wheel.
Skipping colorama, due to already being wheel.
Skipping async-exit-stack, due to already being wheel.
Skipping outcome, due to already being wheel.
[...]

PR #1360, opened 2020-08-11
Pin pyqt5-sip dependency to avoid issues in the future

vxgmichel:
Address https://github.com/Scille/parsec-cloud/issues/1309#issuecomment-666334980:
> And related to this matter, I suggest to define the pyQt5-sip version to 12.8.0, in the relevant place (setup.py?). This would avoid breaking change issue in the future.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -313,13 +313,13 @@ test_requirements = [
 ]
 
 
-PYQT_DEP = "PyQt5==5.14.2"
+PYQT_DEPS = ["PyQt5==5.14.2", "pyqt5-sip==12.8.0"]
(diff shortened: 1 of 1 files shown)


setup.py
```diff
@@ -321,6 +321,7 @@ def run(self):
     "core": [
         PYQT_DEP,
```
touilleMan, line 322: Would it be better to put it in the PYQT_DEP ? (maybe renaming this to `PYQT_DEPS`)

vxgmichel: 25e3033

touilleMan: :+1:

Issue #1356, opened 2020-08-07
Crash on key device read/write with long path in Windows


bitlogik:
Testing the v1.14.0 tag released in Windows, the GUI test is failing during the tests/core/gui/test_bootstrap_organization test.

It is not clear why this is failing. There's a failed check it is logged on, connection closing which seems "normal", and there's a file not found. One theory is that the cause is the too long path to access the device keys file. The key file path used by save_device_with_password, is 291 bytes, far beyond the 255-260 bytes windows limit. So it can't log because it can't find the devices key.

If this is the case, one solution is to use relative paths. [Details here](https://stackoverflow.com/questions/55146969/filenotfounderror-long-file-path-python-filepath-longer-than-255-characters#comment97046893_55146969).
Also one can use partial ID, no need of the full ID, it can be shortened, 4-8 bytes is OK (4 bytes : up to 65000 users). The current lastdir/file.key is very huge (2x83 bytes) : `1d5289ee43#NewOrg#a345ac5f34f04e18a23f695b58366529@612ad7d536dd43208a9adda2048c8ca4\1d5289ee43#NewOrg#a345ac5f34f04e18a23f695b58366529@612ad7d536dd43208a9adda2048c8ca4.keys
`

Here's the full log :
```
$py.test --log-level=DEBUG --durations=10 -v tests --runmountpoint --runslow --rungui -m gui -x
...
tests/core/gui/test_bootstrap_organization.py::test_bootstrap_organization FAILED [  2%]
================================== FAILURES ===================================
_________________________ test_bootstrap_organization _________________________
value = <trio.Nursery object at 0x00000279869DDCC8>
    async def yield_(value=None):
>       return await _yield_(value)
c:\python37\lib\site-packages\async_generator\_impl.py:106: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
c:\python37\lib\site-packages\async_generator\_impl.py:99: in _yield_
    return (yield _wrap(value))
tests\core\gui\test_bootstrap_organization.py:76: in test_bootstrap_organization
    await aqtbot.wait_until(_bootstrap_done)
tests\core\gui\conftest.py:174: in wrapper
    return await self.qt_thread_gateway.send_action(_action)
tests\core\gui\conftest.py:63: in send_action
    return reply.result()
[...]

PR #1364, opened 2020-08-12
Start to solve #1356

bitlogik:
Here's our modest proposal to fix the long path issue in Windows.

Can you look at if this is necessary to do the same in the workspace methods ?

### DIFF
--- a/newsfragments/1356.bugfix.rst
+++ b/newsfragments/1356.bugfix.rst
@@ -1,0 +1,1 @@
+Fix issue with long path names for keys in Windows
--- a/parsec/core/local_device.py
+++ b/parsec/core/local_device.py
@@ -4,6 +4,8 @@ import attr
 from typing import List, Optional
 from pathlib import Path
 from hashlib import sha256
+from os import fspath
+from os import name as osname
 
 from parsec.serde import BaseSchema, fields, MsgpackSerializer
 from parsec.crypto import (
@@ -83,11 +85,28 @@ def generate_new_device(
     )
 
 
+def fix_dir_win(path: Path) -> Path:
+    if osname == "nt":
+        return Path("\\\\?\\" + fspath(path.resolve()))
+    else:
+        return path
+
+
+def decorator_fix_windir(func):
+    def wrapper(*args, **kwargs):
+        path = func(*args, **kwargs)
+        return fix_dir_win(path)
+
+    return wrapper
+
+
+@decorator_fix_windir
 def get_key_file(config_dir: Path, device: LocalDevice) -> Path:
     slug = device.slug
     return get_devices_dir(config_dir) / slug / f"{slug}.keys"
 
 
+@decorator_fix_windir
 def get_devices_dir(config_dir: Path) -> Path:
     return config_dir / "devices"
 
@@ -132,7 +151,7 @@ def list_available_devices(config_dir: Path) -> List[AvailableDevice]:
     devices = []
     for device_folder_path in candidate_pathes:
         slug = device_folder_path.name
-        key_file_path = device_folder_path / f"{slug}.keys"
+        key_file_path = fix_dir_win(device_folder_path / f"{slug}.keys")
 
         try:
             organization_id, device_id = LocalDevice.load_slug(slug)
--- a/tests/core/test_local_device.py
+++ b/tests/core/test_local_device.py
@@ -3,6 +3,8 @@
 import pytest
 from uuid import UUID
 from pathlib import Path
+from os import fspath
+from os import name as osname
 
 from parsec.crypto import SigningKey
 from parsec.serde import packb, unpackb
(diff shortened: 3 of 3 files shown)


vxgmichel:
Hi @bitlogik, thanks for making this PR!

I took a different approach to #1356, as seen in this other PR: #1366 

My guess is that the `\\?\` fix you propose is only useful for the test called `test_list_devices_support_legacy_file_with_meaningful_name` since we're testing the problematic behavior that should no longer exists (the fix can be seen [here](https://github.com/Scille/parsec-cloud/pull/1366/files#diff-93841ed9968868e2042dfe9d9dbe93faR292-R295)). 

I haven't  been able to reproduce the issue #1356 on my windows machine though, could you test that #1366 actually solves the problem?

Thanks a lot for your help :)

bitlogik:
> the \\?\ fix you propose is only useful for the test called test_list_devices_support_legacy_file_with_meaningful_name

Well, this PR is suppose to fix get_key_file and get_devices_dir which are used in many places in the app: in gui/login_widget calling  list_available_devices, in gui/main_window calling try_login,... Are you sure these updated functions in core/local_device.py are only used in tests?

The issue is that pathlib is not always correctly joining long paths, because windows long path needs to adhere on strict rules (one is it shall begin with "\\?\") and sometimes path concatenation drops this required header.

Your PR is good, as it changes the bad practice of having unnecessary long path, but it is not a fix for this current issue. It can reduce the risks linked to #1356 , but not fixing the root cause. Imagine the following, more as a thought experiment than a practical behavior : if a user has a very long user name in Windows, and have a very long organization name. We think that this PR makes this case working right, as #1366 is still facing #1356.

We tested, and #1366 are passing the tests, as it works around the issue.
Anyway, both changes are required 1) reducing the folder and files names of local keys files 2) fixing the long path concatenation issue in Windows

vxgmichel:
>> My guess is that the \\?\ fix you propose is only useful for the test called test_list_devices_support_legacy_file_with_meaningful_name since we're testing the problematic behavior that should no longer exists (the fix can be seen here).

> Well, this PR is suppose to fix get_key_file and get_devices_dir which are used in many places in the app: in gui/login_widget calling list_available_devices, in gui/main_window calling try_login,... Are you sure these updated functions in core/local_device.py are only used in tests?

Sorry for not being clear, what I meant is that in PR #1366 the only case of very long path in the CI should appear in the test I mentioned. That's why I used the `\\?\` trick specifically for this particular test.

 > Your PR is good, as it changes the bad practice of having unnecessary long path, but it is not a fix for this current issue. It can reduce the risks linked to #1356 , but not fixing the root cause. 

I understand your point about the #1366 not fixing the root cause. Here's my opinion about it :)

 The root cause is either windows not supporting correctly very long paths, or python not providing a transparent path conversion layer for very long paths. In this regard, any fix in our code base is a workaround for issues that is not the responsibility of parsec. My opinion about workarounds is that we introduce them in the code base only if they solve problems that actually happens. In other words, we try to avoid introducing workarounds to solve problems preemptively.

In this particular case, here's a device keys filename with PR #1366:

    C:\Users\User\AppData\Roaming\Parsec\config\devices\0ef87a2eab794122b11b2dbf7d5d51d7.keys

It's about 90 chars long, so a user would have to have a username of about 170 chars to hit the 256 chars limit. This seem rather unrealistic and I don't think it's enough to justify the introduction of a workaround. 

I understand that this piece is quite opinionated so I can bring up the issue at the next meeting to see if our colleagues share the same opinion !

> We tested, and #1366 are passing the tests, as it works around the issue.

Thanks for testing this, it's much appreciated :smiley:

bitlogik:
We're kind of agree on many things here. I admit that during the investigation, I planned to open an issue in Python. Because there's an issue in the way pathlib is handling long path in Windows. Handling properly long path on Windows is very difficult and pathlib concatenation is not compliant with it. Whether this is strictly a bug or just an annoying behavior is an other topic. I finally thought that fixing this in Python will be a very complex process, uncertain, very long, eventually leading to a fix in Python v2.9. So I agree the root cause is an issue in pathlib, not in parsec. Then it is a matter of priority and philosophy, ans this PR is indeed a fix of an underlying problem out of parsec, but which can affect parsec.

Also I wrongly assume the key file path included the user name or device name, something like config/devices/DEVICE_NAME/0ef87a2eab794122b11b2dbf7d5d51d7.keys. If there's no such things, the issue happening probability is even lower.

This PR is only a very small overhead to fix the long path on Windows, and it will be implemented on our public forked version of Parsec. There will be some differences, and this little change will be one.

vxgmichel:
> I admit that during the investigation, I planned to open an issue in Python. Because there's an issue in the way pathlib is handling long path in Windows.

> This PR is only a very small overhead to fix the long path on Windows, and it will be implemented on our public forked version of Parsec. There will be some differences, and this little change will be one.

No problem :) Then I would suggest monkeypatching `pathlib` in order to integrate your fix in a transparent way. This way, not only device files but all parsec-related local files are protected against this long path handling issue.

Since #1366 is now merged, I'll close #1365 and #1364. Thanks again for your help !

PR #1366, opened 2020-08-13
Devices keys filenames are no longer meaningful

vxgmichel:
Key files used to be stored in a directory named after the device slug in a file also named after the same device slug. The total length is more than 160 characters depending on the organization name.

Key files are now stored directly in the devices directory using the device slughash and the `.keys` extension. The total length is 37
characters.

This fixes issue #1356.

### DIFF
--- a/parsec/core/local_device.py
+++ b/parsec/core/local_device.py
@@ -204,8 +204,11 @@ def _load_device_file(key_file_path: Path) -> Optional[AvailableDevice]:
 def _iter_available_devices(config_dir: Path) -> Iterator[AvailableDevice]:
     # Set of seen slugs
     seen = set()
-    # Loop over `.keys` files in devices directory and subdirectories
-    for key_file_path in get_devices_dir(config_dir).rglob("*.keys"):
+    # Consider `.keys` files in devices directory and subdirectories
+    key_file_paths = list(get_devices_dir(config_dir).rglob("*.keys"))
+    # Sort paths so the discovery order is deterministic
+    # In the case of duplicate files, that means only the first discovered device is considered
+    for key_file_path in sorted(key_file_paths):
         # Load the device file
         device = _load_device_file(key_file_path)
         # Try with the legacy deserializer if necessary
--- a/parsec/core/local_device.py
+++ b/parsec/core/local_device.py
@@ -49,7 +49,9 @@ class LocalDevicePackingError(LocalDeviceError):
     pass
 
 
-class DeviceFileSchema(BaseSchema):
+class LegacyDeviceFileSchema(BaseSchema):
+    """Schema for legacy device files where the filename contains complementary information."""
+
     type = fields.CheckedConstant("password", required=True)
     salt = fields.Bytes(required=True)
     ciphertext = fields.Bytes(required=True)
@@ -61,12 +63,26 @@ class DeviceFileSchema(BaseSchema):
     human_handle = HumanHandleField(allow_none=True, missing=None)
     device_label = fields.String(allow_none=True, missing=None)
 
+
+class DeviceFileSchema(LegacyDeviceFileSchema):
+    """Schema for device files that does not rely on the filename for complementary information."""
+
+    # Override those fields to make them required (although `None` is still valid)
+    human_handle = HumanHandleField(required=True, allow_none=True)
+    device_label = fields.String(required=True, allow_none=True)
+
     # Store device ID, organization ID and slug in the device file
     # For legacy versions, this information is available in the file name
-    device_id = DeviceIDField(allow_none=True, missing=None)
-    organization_id = OrganizationIDField(allow_none=True, missing=None)
-    slug = fields.String(allow_none=True, missing=None)
+    device_id = DeviceIDField(required=True)
+    organization_id = OrganizationIDField(required=True)
+    slug = fields.String(required=True)
 
+
+legacy_key_file_serializer = MsgpackSerializer(
+    LegacyDeviceFileSchema,
+    validation_exc=LocalDeviceValidationError,
+    packing_exc=LocalDevicePackingError,
+)
 
 key_file_serializer = MsgpackSerializer(
     DeviceFileSchema, validation_exc=LocalDeviceValidationError, packing_exc=LocalDevicePackingError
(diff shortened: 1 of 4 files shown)


parsec/core/local_device.py
```diff
@@ -124,35 +142,40 @@ def slughash(self) -> str:
...
-
-        try:
-            organization_id, device_id = LocalDevice.load_slug(slug)
-            # Not a valid slug, ignore this folder
-        except ValueError:
-            continue
+    for key_file_path in key_file_paths:
 
         try:
             data = key_file_serializer.loads(key_file_path.read_bytes())
```
touilleMan, line 154: Wouldn't it be better to have two separate `DeviceFileSchema` classes: `DeviceFileSchema` and `LegacyDeviceFileSchema` (or `DeviceFileSchemaV1` and `DeviceFileSchemaV2`)

this would allow us to have a strong border between old and new formats (which should help to deprecate the old format)
```python
class DeviceFileV2Schema(DeviceFileV1Schema):
    # Store device and organization ID in the device file
    # For legacy versions, this information is available in the file name
    device_id = DeviceIDField(required=True)
    organization_id = OrganizationIDField(required=True)
    root_verify_key_hash = fields.String(required=True)


key_file_v1_serializer = MsgpackSerializer(
    DeviceFileV1Schema, validation_exc=LocalDeviceValidationError, packing_exc=LocalDevicePackingError
)

key_file_v2_serializer = MsgpackSerializer(
    DeviceFileV2Schema, validation_exc=LocalDeviceValidationError, packing_exc=LocalDevicePackingError
)

def load_device_file_v1(key_file_path):
    slug = key_file_path.steam
    try:
        organization_id, device_id = LocalDevice.load_slug(slug)
        # Not a valid slug, ignore this folder
    except ValueError:
        return None

    try:
        data = key_file_v1_serializer.loads(key_file_path.read_bytes())
    except (FileNotFoundError, LocalDeviceError):
        # Not a valid device file, ignore this folder
        return None

    return AvailableDevice(
        key_file_path=key_file_path,
        organization_id=organization_id,
        device_id=device_id,
        human_handle=data["human_handle"],
        device_label=data["device_label"],
    )



def load_device_file_v2():
    try:
        data = key_file_v2_serializer.loads(key_file_path.read_bytes())
    except (FileNotFoundError, LocalDeviceError):
        # Not a valid device file, ignore this folder
        return None

    return AvailableDevice(
        key_file_path=key_file_path,
        organization_id=data["organization_id"],
        device_id=data["device_id"],
        human_handle=data["human_handle"],
        device_label=data["device_label"],
    )

def list_available_devices(config_dir: Path) -> List[AvailableDevice]: 
    [...]
    for key_file_path in key_file_paths:
        for loader in (load_device_file_v1, load_device_file_v2):
            available_device = loader(key_file_path)
            if available_device:
                devices.append(available_device)
```

parsec/core/types/local_device.py
```diff
@@ -91,17 +90,21 @@ def slughash(self) -> str:
...
+        rvk_hash, raw_org_id, raw_device_id = slug.split("#")
+        return rvk_hash, OrganizationID(raw_org_id), DeviceID(raw_device_id)
 
     @property
     def root_verify_key(self):
         return self.organization_addr.root_verify_key
 
+    @property
+    def root_verify_key_hash(self):
+        return sha256(self.root_verify_key.encode()).hexdigest()[:10]
```
touilleMan, line 106: why not the whole hash ? (and why not return bytes ? ^^)

parsec/core/local_device.py
touilleMan: By the way, wouldn't it be better the run a conversion routine before actually retrieving the available devices ?
This would have the benefit of keeping the legacy format handling in a single function (this would solve the issue I was talking about in the `get_key_file` function) and should be really simple to implement given the key file is immutable and all data are available.

On top of that if we don't delete the old file after conversion, we are still compatible with previous version of Parsec  \o/

parsec/core/local_device.py
```diff
@@ -84,8 +97,12 @@ def generate_new_device(
...
 
 def get_key_file(config_dir: Path, device: LocalDevice) -> Path:
-    slug = device.slug
-    return get_devices_dir(config_dir) / slug / f"{slug}.keys"
+    """Return the keyfile path for a given device.
+
+    Note that the filename does not carry any intrinsic meaning.
+    Here, we simply use the slughash to avoid name collision.
+    """
+    return get_devices_dir(config_dir) / f"{device.slughash}.keys"
```
touilleMan, line 105: Does this work with legacy key file when calling this function to retrieve the file in question ?
This is what is done in the GUI when changing the password...

vxgmichel: Oopsie, addressed here: 5017400cbeccb8427f9bdab2044dbeabeeade29c

parsec/core/local_device.py
```diff
@@ -54,6 +61,12 @@ class DeviceFileSchema(BaseSchema):
     human_handle = HumanHandleField(allow_none=True, missing=None)
     device_label = fields.String(allow_none=True, missing=None)
 
+    # Store device and organization ID in the device file
+    # For legacy versions, this information is available in the file name
+    device_id = DeviceIDField(allow_none=True, missing=None)
+    organization_id = OrganizationIDField(allow_none=True, missing=None)
+    root_verify_key_hash = fields.String(allow_none=True, missing=None)
```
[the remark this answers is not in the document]
vxgmichel: In this context `root_verify_key_hash` is the first 10 chars of the hex representation of the hash, so a string makes sense. We could store the actual hash of the rvk, I don't know if it's useful or not. 

Another way to approach the problem would be to store the `slug` instead since it already contains all this info. Any opinion about that?

parsec/core/types/local_device.py
vxgmichel: > Wouldn't it be better to have two separate DeviceFileSchema classes: DeviceFileSchema and LegacyDeviceFileSchema (or DeviceFileSchemaV1 and DeviceFileSchemaV2)

I don't really have a strong opinion about it, but there's already a compatibility layer in `DeviceFileSchema`:
```
    # Since human_handle/device_label has been introduced, device_id is
    # redacted (i.e. user_id and device_name are 2 random uuids), hence
    # those fields have been added to the device file so the login page in
    # the GUI can use them to provide useful information.
    human_handle = HumanHandleField(allow_none=True, missing=None)
    device_label = fields.String(allow_none=True, missing=None)
```

Does that mean we'd need `V1`, `V2` and `V3`?  Maybe it's a bit overkill considering the amount of compatibility logic?

parsec/core/local_device.py
vxgmichel: > Wouldn't it be better to have two separate DeviceFileSchema classes: DeviceFileSchema and LegacyDeviceFileSchema (or DeviceFileSchemaV1 and DeviceFileSchemaV2)

I don't really have a strong opinion about it, but there's already a compatibility layer in `DeviceFileSchema`:
```
    # Since human_handle/device_label has been introduced, device_id is
    # redacted (i.e. user_id and device_name are 2 random uuids), hence
    # those fields have been added to the device file so the login page in
    # the GUI can use them to provide useful information.
    human_handle = HumanHandleField(allow_none=True, missing=None)
    device_label = fields.String(allow_none=True, missing=None)
```

Does that mean we'd need `V1`, `V2` and `V3`?  Maybe it's a bit overkill considering the amount of compatibility logic?

vxgmichel: > By the way, wouldn't it be better the run a conversion routine before actually retrieving the available devices ?
This would have the benefit of keeping the legacy format handling in a single function

Do you mean re-writing the device files from old to new format? Good question :thinking: 

On one hand we would get rid of the old format more easily, but on the other hand writing in the a read function can be a bit dangerous. Plus, we wouldn't be able to fix the `human_handle` and `device_label` fields so we'd need a compatibility layer anyway.

> ... (this would solve the issue I was talking about in the get_key_file function) and should be really simple to implement given the key file is immutable and all data are available.

l used another approach to solve this issue, see 5017400cbeccb8427f9bdab2044dbeabeeade29c

vxgmichel: Done! 
c809c86

parsec/core/local_device.py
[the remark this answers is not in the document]
vxgmichel: c809c86

parsec/core/local_device.py
```diff
@@ -112,51 +153,77 @@ def short_user_display(self) -> str:
...
+        device_label=data["device_label"],
+        slug=data["slug"],
+    )
+
+
+def _iter_available_devices(config_dir: Path) -> Iterator[AvailableDevice]:
+    # Set of seen slugs
+    seen = set()
+    # Loop over `.keys` files in devices directory and subdirectories
+    for key_file_path in get_devices_dir(config_dir).rglob("*.keys"):
```
touilleMan, line 208: This could be troublesome in case we have multiple files with the same slug (if a user has copied a key file for instance)

We should enforce that the iteration is done in a sorted way so calling multiple time `get_key_file` always provide the same path (and we should document this as well of course ^^)

PR #1368, opened 2020-08-13
Fix broken Ubuntu Azure pipelines

bitlogik:
Ubuntu system package needs to be updated, else there can be some broken downloaded package links. e.g. postgresql `Failed to fetch http://...postgresql amd64.deb  404  Not Found [IP: xxxx 80]`

### DIFF
--- a/.azure-pipelines.yml
+++ b/.azure-pipelines.yml
@@ -38,6 +38,7 @@ jobs:
       versionSpec: '$(python.version)'
   - bash: |
       set -eux
+      sudo apt-get update
       python --version
       pip install -r pre-requirements.txt
     displayName: 'Bootstrap'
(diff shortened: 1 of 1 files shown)


vxgmichel:
Hi @bitlogik! Thanks for fixing our CI :sweat_smile: 

One question though, it seems to me that only one `apt-get update` is necessary, i.e the one before `apt-get install`:
```
      sudo apt-get install \
        postgresql-$(postgresql.version) \
        desktop-file-utils  # Provides `update-desktop-database` used by `tests/scripts/run_testenv.sh`
```
Am I missing something?

bitlogik:
Yes, you're right. This is just a matter of security and future proof, this way the apt system will be OK in any Linux VM pipeline. For example if one day `install -y xvfb xorg` is un-commented. Note that `apt update` is just a matter of seconds, it only updates the APT database, especially the download links to packages. This does not update all the installed packages of the system, this operation is called *upgrade* in Debian/Ubuntu and is much more longer and stressful for the system.

vxgmichel:
Alright, thanks for the insight!

PR #1372, opened 2020-08-17
Invitation: inconsistent backend replies

AntoineDupre:
Always close connection, never returns, when an invitation is cancelled after the invitation handshake.

Add a retrier on the client side on BackendNotAvailable for invitation commands.

### DIFF
--- a/parsec/backend/app.py
+++ b/parsec/backend/app.py
@@ -27,7 +27,7 @@ from parsec.backend.handshake import do_handshake
 from parsec.backend.memory import components_factory as mocked_components_factory
 from parsec.backend.postgresql import components_factory as postgresql_components_factory
 from parsec.backend.http import HTTPRequest
-from parsec.backend.invite import InvitationAlreadyDeletedError
+from parsec.backend.invite import CloseInviteConnection
 
 
 logger = get_logger()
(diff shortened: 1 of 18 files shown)


vxgmichel (changes requested):
`InviteAlreadyUsedError` should be removed completely, LGTM otherwise :+1:

parsec/backend/app.py
```diff
@@ -303,6 +304,15 @@ async def _handle_client_loop(self, transport, client_ctx):
                         "reason": "Invalid message.",
                     }
 
+                except InvitationAlreadyDeletedError:
```
touilleMan, line 307: I'm not sure about this approach given we are creating a special case for this kind of exception...

At first I was thinking about plugging into the event callback responsible to close the connection:
```python
class BaseClientContext:
    await def signify_and_wait_for_certain_death(self):
        self.connection_end_requested.set()
        await trio.sleep_forever()

[...]

async def watch_for_connection_end_requested():
    await client_ctx.connection_end_requested.wait()
    cancel_scope.cancel()
```

But it's cumbersome to connect the client_ctx with the watchdog function and we need a nursery for the watchdog which wasn't needed previously.

So in the end I think the "close my connection please" exception is a good idea, but it should be a generic exception (and the invite cmd function should do the translation from `InvitationAlreadyDeletedError` into this generic one)

touilleMan (changes requested):

parsec/backend/app.py
AntoineDupre: Update: `InvitationAlreadyDeletedError` are now converted to a `CloseInviteConnection`. This exception is handled directly in `handle_client_websocket`

```python
try:
    with trio.CancelScope() as cancel_scope:
        with self.event_bus.connection_context() as event_bus_ctx:
            ...
            await self._handle_client_loop(transport, client_ctx)
except CloseInviteConnection:
    pass
finally:
    with trio.CancelScope(shield=True):
        await self.invite.claimer_left(...)
await transport.aclose()
```

vxgmichel (approved):
LGTM!

PR #1373, opened 2020-08-17
Display invitation user before user list

Sithi5:


parsec/core/gui/users_widget.py
```diff
@@ -375,10 +375,7 @@ def _on_list_success(self, job):
         self._flush_users_list()
 
         current_user = self.core.device.user_id
-        for user_info in users:
-            self.add_user(user_info=user_info, is_current_user=current_user == user_info.user_id)
-
-        for invitation in invitations:
+        for invitation in reversed(invitations):
```
Max-7, line 378: Why the `reversed`? We only care that invitations are displayed before the users, we don't care about the order in which the invitations are displayed.

Sithi5: I use reverse to put latest user invitation in first

PR #1374, opened 2020-08-17
Invite user now show a confirm popup

Sithi5:


parsec/core/gui/tr/parsec_fr.po
```diff
@@ -673,14 +673,15 @@ msgstr "utilisateur@hote.com"
...
-msgstr "Annuler l'invitation"
-
 msgid "TEXT_USER_INVITE_CANCEL_INVITE_QUESTION_CONTENT"
 msgstr ""
 "Êtes-vous sûr de vouloir annuler cette invitation ? Le lien d'invitation "
 "qu'a reçu l'utilisateur par email deviendra invalide."
 
+msgid "TEXT_USER_INVITE_SUCCESS_email"
+msgstr ""
+"L'invitation à rejoindre votre organisation a été envoyé avec succès à l'adresse : <b>\"{email}\"</b>."
```
Max-7, line 683: "EnvoyéE".
In both French and English I would simply remove the colon and the quotes around the email.

parsec/core/gui/users_widget.py
```diff
@@ -257,6 +257,7 @@ def invite_user(self):
         if not user_email:
             return
 
+        self._user_email = user_email
```
Max-7, line 260: It is better to get the email from the return value of `_do_invite_user` rather than store it as an attribute. Storing this kind of attributes can lead to a lot of weird errors later on.

parsec/core/gui/tr/parsec_fr.po
Sithi5: My bad ^^

parsec/core/gui/users_widget.py
Sithi5: Done

PR #1367, opened 2020-08-13
Workspace file fs improvement

Sithi5:

PR #1383, opened 2020-08-19
Refactor read_bytes and write_bytes to use workspace file interface

vxgmichel:


parsec/core/fs/workspacefs/workspacefs.py
```diff
@@ -423,14 +408,14 @@ async def copyfile(
...
         """
-        await self.touch(target_path, exist_ok=exist_ok)
-        offset = 0
-        while True:
-            buff = await self.read_bytes(source_path, length, offset * length)
-            if not buff:
-                break
-            await self.write_bytes(target_path, buff, offset * length)
-            offset += 1
+        write_mode = "wb" if exist_ok else "xb"
```
Sithi5, line 411: Can use "wb" mode in this case because it will also create the file if it doesn't exist

tests/core/fs/workspacefs/test_workspace_fs.py
```diff
@@ -230,44 +230,27 @@ async def test_read_bytes(alice_workspace):
     assert await alice_workspace.read_bytes("/foo/bar") == b""
 
     await alice_workspace.write_bytes("/foo/bar", b"abcde")
-    assert await alice_workspace.read_bytes("/foo/bar", size=3) == b"abc"
```
Sithi5, line 233: Maybe can be replaced by `assert await alice_workspace.read_bytes("/foo/bar") == b"abcde"`?

Sithi5 (approved):
LGTM

parsec/core/fs/workspacefs/workspacefile.py
```diff
@@ -87,7 +87,8 @@ async def ainit(self):
                 _, self._fd = await self._transactions.file_create(self._path, open=True)
             else:
                 try:
-                    _, self._fd = await self._transactions.file_open(self._path)
+                    mode = "rw" if self.writable() else "r"
```
touilleMan, line 90: `b` mode is not taken into account here ?

parsec/core/fs/workspacefs/workspacefile.py
```diff
@@ -49,7 +49,7 @@ class WorkspaceFile:
     mode("r") -> read file.
     """
 
-    def __init__(self, transactions, path, mode: str = "r"):
+    def __init__(self, transactions, path, mode: str = "rb"):
```
touilleMan, line 52: The docstring about the `b` mode seems kind of weird: `bytes, have to be used with another mode, ignored, class is already working with bytes.`
Why are we saying it is ignored when checks are done in the write method that depend on this param ?

parsec/core/fs/workspacefs/workspacefs.py
```diff
@@ -324,41 +324,26 @@ async def truncate(self, path: AnyPath, length: int) -> None:
         path = FsPath(path)
         await self.transactions.file_resize(path, length)
 
-    async def read_bytes(self, path: AnyPath, size: int = -1, offset: int = 0) -> bytes:
+    async def open_file(self, path: AnyPath, mode="rb"):
```
touilleMan, line 327: no return typing ?

parsec/core/fs/workspacefs/workspacefs.py
```diff
@@ -423,14 +408,14 @@ async def copyfile(
...
-            buff = await self.read_bytes(source_path, length, offset * length)
-            if not buff:
-                break
-            await self.write_bytes(target_path, buff, offset * length)
-            offset += 1
+        write_mode = "wb" if exist_ok else "xb"
+        async with await self.open_file(source_path, mode="rb") as source:
+            async with await self.open_file(target_path, mode=write_mode) as target:
+                while True:
+                    data = await source.read(length)
```
touilleMan, line 415: `length` feels like a bad name for this parameter (it seems it there to copy only part of the file, where it is there to controle the size of the buffer for the copy)... so we should rename this `buffer_size` or even remove this parameter entirely given we don't use it in our codebase ;-)

tests/core/fs/test_fs_idempotent_sync.py
```diff
@@ -115,7 +115,8 @@ async def try_to_create_already_exists(self, type, path):
         @rule(path=BadPath)
         async def try_to_update_file(self, path):
             with pytest.raises(OSError):
-                await self.workspace.write_bytes(path, data=b"a", offset=0)
+                async with await self.workspace.open_file(path, "rb+"):
```
touilleMan, line 118: `+` is not documented in the `WorkspaceFile` docstring

tests/core/fs/test_fs_offline_restart_and_rwfile.py
```diff
@@ -43,9 +45,9 @@ async def atomic_read(self, size, offset):
             content=st.binary(max_size=PLAYGROUND_SIZE),
         )
         async def atomic_write(self, offset, content):
-            await self.workspace.write_bytes(
-                "/foo.txt", data=content, offset=offset, truncate=False
-            )
+            async with await self.workspace.open_file("/foo.txt", "rb+") as f:
```
touilleMan, line 48: shouldn't it be `wb` mode there ?

tests/core/fs/workspacefs/test_workspace_fs.py
touilleMan, line 233: Those tests are no longer relevant, but shouldn't we do them for `WorkspaceFile` instead ?

tests/core/test_sync_monitor_stateful.py
```diff
@@ -187,7 +187,8 @@ async def _bob_update_file(self, wid, file_path, create_file=False):
                 await wfs.touch(file_path)
             else:
                 data = self.get_next_data()
-                await wfs.write_bytes(file_path, data)
+                if await wfs.exists(file_path):
```
touilleMan, line 190: In which case would we be here with `file_path` not existing ?

I have the feeling we this `if exists` is not a good idea given we rely on the tested component to determine if we should do the test...

(besides this implies `wfs.write_bytes` behavior has changed... is this expected ?)

parsec/core/fs/workspacefs/workspacefile.py
vxgmichel: No, because that's not same definition of mode.
- `transactions.file_open` is low-level and uses the following `mode` definition: `{os.O_RDONLY: "r", os.O_WRONLY: "w", os.O_RDWR: "rw"}`
- `workspace.open_file` is high level and uses the same definition as the `open` python built-in

I can add a comment explaining that.

parsec/core/fs/workspacefs/workspacefile.py
vxgmichel: @Sithi5 Could you fix the docstring?

parsec/core/fs/workspacefs/workspacefs.py
vxgmichel: @Sithi5 Could you fix the typing for `open_file` and `WorkspaceFile`?

parsec/core/fs/workspacefs/workspacefs.py
vxgmichel: Configuring the `buffer_size` copy seems ok to me, I'll rename it :)

tests/core/fs/test_fs_idempotent_sync.py
vxgmichel: @Sithi5 Could you fix the docstring?

tests/core/fs/test_fs_offline_restart_and_rwfile.py
vxgmichel: No because we do not want to create the file if it doesn't exist. The previous version of `write_bytes` used to match this behavior but it was a mistake as `write_bytes` is expected to create a file if it doesn't exist (as in the `pathlib.Path.write_bytes` definition)

tests/core/fs/workspacefs/test_workspace_fs.py
vxgmichel: Those tests already exist, see PR #1367

tests/core/test_sync_monitor_stateful.py
vxgmichel: > this implies wfs.write_bytes behavior has changed... is this expected ?

Yes, as explained in this comment: https://github.com/Scille/parsec-cloud/pull/1383#discussion_r474081558

I simply adapted the code to match the previous behavior, should I investigate this in more depth?

parsec/core/fs/workspacefs/workspacefile.py
```diff
@@ -87,7 +87,8 @@ async def ainit(self):
                 _, self._fd = await self._transactions.file_create(self._path, open=True)
             else:
                 try:
-                    _, self._fd = await self._transactions.file_open(self._path)
+                    mode = "rw" if self.writable() else "r"
+                    _, self._fd = await self._transactions.file_open(self._path, mode)
                 except FileNotFoundError:
                     if sum(c in self._mode for c in "aw") == 1:
```
touilleMan, line 93: I'd rather do
```python
if 'a' in self._mode or 'w' in self._mode
```
(especially given we already checked that we cannot have both 'a' and 'w' at the same time)

beside for the sum, it would be better not to take advantage of the fact `int(True) == 1` given it's more of a legacy detail that a real feature:
```python
        if sum(1 for c in mode if c in "rwax") != 1:
```

tests/core/test_sync_monitor_stateful.py
touilleMan: The `create_file` param seems to indicate the file should already be present, so I think we should remove the `wfs.exists check` here

parsec/core/fs/workspacefs/workspacefile.py
Sithi5: Yes it seems better, I modify that

tests/core/fs/test_fs_idempotent_sync.py
Sithi5: Yes, done

parsec/core/fs/workspacefs/workspacefs.py
Sithi5: I did it for `open_file` and for `readable/writable/seekable` method of WorkspaceFile.

parsec/core/fs/workspacefs/workspacefile.py
Sithi5: done

parsec/core/fs/workspacefs/workspacefile.py
vxgmichel: > it would be better not to take advantage of the fact int(True) == 1 given it's more of a legacy detail that a real feature

Well, `True == 1` and `False == 1` are more than a legacy detail to me, I don't think it's something that could be discarded from the language. But I'm fine with both approaches, just being pedantic here :)

PR #1386, opened 2020-08-20
Handle pagination in GUI

Sithi5:


bitlogik:
We are in the process of porting this branch in our fork. Here are some issues we noticed about this branch in progress.

parsec/backend/postgresql/user_queries/find.py has a query with an extra parenthesis, breaks the postgresql queries on line 69.

No need limit and offset for human count _q_count_total_human in parsec/backend/postgresql/user_queries/find.py
_q_count_total_human is even called without these parameters.  
parsec/backend/postgresql/user_queries/find.py :  
change L44  remove extra params
delete L57 from request

"now" parameter is not always in count query _q_count_total_human, leading to error
```
if omit_revoked:
    total = await conn.fetch(*q(organization_id=organization_id, now=pendulum_now()))
else:
    total = await conn.fetch(*q(organization_id=organization_id))

```

q/_q_count_total_human fetch.row is returning a <Record count> object. To get the integer, one should use fetchrow and get the first element (index 0)

_q_factory is not called with limit and and offset, `LIMIT {limit} OFFSET {offset`} should be added after ORDER BY L26

We worked to [add the sort to be  fully done by postgresql](https://github.com/bitlogik/guardata/commit/05514d0a539e7cc9386ebe346cb7a1214c8b0fc6) (sort is not as good as before). Now, it just lacks the memory mocked path to perform the same (the order was changed) in order to pass the test_apiv1_human_find test.

The changes are available in the paging branch in guardata, the last 4 commit (additional coming), ones following:
https://github.com/bitlogik/guardata/commit/54df63672809fa49617830b1903fa2ee6617af85
It now lacks to perform the same ordering in memory part for tests.

https://github.com/bitlogik/guardata/commits/listing

About retro-compatibility, how an "old" client would react when the server is sending back more total number thant the response elements list?

Note there can be some breaking change about retro-compatibility, as the server response changed sign

bitlogik:
Oh yes I didn't even list all the issues, for example the offset computation was wrong.

More changes :
https://github.com/bitlogik/guardata/commits/listing
With these, up to fd50eedc, it pass all tests in tests/backend/user/

bitlogik:
... all tests with --postgresql, else it fails because the memory can now sort only for "humans" as the "label" attr used to sort the result. I recommend to sort in memory only on user_id as it is done in postgresql and in other memory parts. This way, the sort can be performed whatever the results.

Sithi5:
Thank you for your comment, 
Another current issue is that in `postgresql` as well as in `memory` we currently find by `human_label` , `email` and `userId`. In some case we need to find by `userId` but for the GUI users filter it is actually a problem.
About the retro-compatibility it should be Ok because we was actually getting all users so `total` should be the same.

bitlogik:
The last issue "memory test sort" was finally an easy fix. https://github.com/bitlogik/guardata/commit/c7322a2a915f6189df81c96dfa581d1a8033082a

bitlogik:
One more error, at least.
As the sorting changed, the greeting test fails at the end when it reads the users list to confirm the user was actually added.
```
   def _greet_done():
        assert not gu_w.isVisible()
        assert autoclose_dialog.dialogs == [
            ("", "The user was successfully greeter in your organization.")
        ]
        # User list should be updated
        assert u_w.layout_users.count() == 4
        user_widget = u_w.layout_users.itemAt(3).widget()
        assert isinstance(user_widget, UserButton)
>       assert user_widget.user_info.human_handle == self.requested_human_handle
E       AssertionError: assert <HumanHandle ...example.com> > == <HumanHandle ...pfry@pe.com> >
E         At index 0 diff: 'bob@example.com' != 'pfry@pe.com'

tests/client/gui/users_widget/test_greet.py:292: AssertionError
```

bitlogik:
Actually, the issue in this test is an issue in the test script. Sometimes the new person invited is listed at the end, as it should, sometimes at first (from invitations at first, then as a user). I'm trying to solve this testing issue, which is related to a race condition lack of refresh, something related to https://github.com/Scille/parsec-cloud/issues/1428.
For now I can't get any reliable position index, as the new person can be at the end or at first. I think this can be solvable by doing like `assert itemAt(0) or itemAt(3)` and also in looping into all users to check if the new one is among the listed. Still, I'm investigating as there is a issue in the test reproducibility, as the order of the user listed at this place in the script is fluctuating (unde mergitur).

bitlogik:
OK, I get it. The 3 existing users have their name as id (adam, alice, bob), and the new invited user gets for her id, 32 hex chars. So when looking the test list in users widget, 68% of the time (0..9a) it is at position 0, and 32% position 3 (bcdef) [actually it can be anywhere for example afxxxxx will list it in pos 2].
Since we now sort only by userid, this test must be changed accordingly. I'll look if we can change the way the userid is computed in this tests scripts. Either given random id for 3 exsitng, or from the name for the invited one.

bitlogik:
Finally fixed it by looping through all users. Since this test is done in the GUI (not API), control the user id of the invited user is almost impossible (IMO it would require importing backend code at least). At the end, it loops to all 4 users to see if the expected new one is among there.
https://github.com/bitlogik/guardata/commit/c4af7c186a25ec1982615fec5a07b2423ba7c371

I think now, we're done with all tests. Still, I'll manually check the behavior about timeout from server response (when the GU can't reach the server, or the query packets are discarded and don't reach the server). Because the spinner was added in this branch, and timeout were removed.

Sithi5:
Thank you for your help, every tests seems to work correctly. I think the use of `find()` instead of `startwith()` is maybe less userfriendly, I will try to find a way in the sql query to behave like `startwith`. Another problem is that for now we have this type of behaviour 
image
because we also filter with the `userId`.

bitlogik:
Personally I'm choosing "find" which is more user friendly, as it can find any string, not just the start. That would be one of the differences with our implementation flavour.
To have "startswith" with a postgresql query, I would recomment ILIKE x% ,maybe with a f-string like I did for some search query (so instead of $x you get {x} and the value is populated when building the query , not during the query execution).

I think I'll remove the search in the user id (will search in other fields), I need to see what are the side effect. Mostly "non-human" users won't get find. But I understand nonhuman is only for outsiders, so they don't need to seek for any user. Also without userid and working only on humans, this can bring back the sort by human names which is more natural.
The device search has a really strange behavior, maybe this is because of the same reason, it search on user (or device id?)which is a UUID hex. I'll investigate on this, may after userid search removal.

As we fear, the spinner can loop forever when there's no response from server (and no error during the query). We'll add back the timeout in our client, to avoid this annoying behavior. I need to fully confirm this before trying to solve it. But for now we have a urgent matter as the CICD on Gitlab is totally broken for windows.

Sithi5:
see https://github.com/Scille/parsec-cloud/pull/1467 and https://github.com/Scille/parsec-cloud/pull/1452

PR #1392, opened 2020-08-24
aqtbot was not defined in reencryption test

bitlogik:


bitlogik:
[Nothing to see here]()

PR #1394, opened 2020-08-25
[GUI] Fixed error message when workspace is unshared and added tests

### LINKED ISSUE #1385: No workspace name info when removed from a workspace
The modal show up directly after connecting to the organization and there is no info of which workspace we have been revoked from. Maybe the title of the modal should also be different ? ex : "Revoked from `workspacename`"
image

Max-7:
Closes #1385

### DIFF
--- a/parsec/core/gui/tr/parsec_en.po
+++ b/parsec/core/gui/tr/parsec_en.po
@@ -7,8 +7,8 @@ msgid ""
 msgstr ""
 "Project-Id-Version: PROJECT VERSION\n"
 "Report-Msgid-Bugs-To: EMAIL@ADDRESS\n"
-"POT-Creation-Date: 2020-08-14 06:36+0200\n"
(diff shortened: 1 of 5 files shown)


parsec/core/gui/tr/parsec_en.po
```diff
@@ -1523,68 +1524,6 @@ msgstr "The file is synced."
 msgid "TEXT_FILE_ITEM_IS_NOT_SYNCED_TOOLTIP"
 msgstr "The file is not synced yet."
 
-msgid "TEXT_BOOTSTRAP_ORG_INSTRUCTIONS_url-organization"
```
AntoineDupre, line 1526: Is this clean up of unused messages ?

Max-7: Yep it is.

AntoineDupre (approved):
LGTM

PR #1397, opened 2020-08-25
[GUI] Display workspace as private if all users it's shared with are revoked

### LINKED ISSUE #1387: Fixed workspace marked as shared when all users it's been shared with are revoked

Max-7:
Closes #1387

### DIFF
--- a/parsec/core/gui/workspace_button.py
+++ b/parsec/core/gui/workspace_button.py
@@ -122,7 +122,7 @@ class WorkspaceButton(QWidget, Ui_WorkspaceButton):
 
     @property
     def is_shared(self):
-        return len(self.users_roles) > 1
+        return sum(1 for _, u in self.users_roles.values() if u.revoked_on is None) > 1
 
     @property
     def owner(self):
--- a/parsec/core/gui/workspace_button.py
+++ b/parsec/core/gui/workspace_button.py
@@ -122,6 +122,7 @@ class WorkspaceButton(QWidget, Ui_WorkspaceButton):
 
     @property
     def is_shared(self):
+        # Filter out users that have been revoked
         return sum(1 for _, u in self.users_roles.values() if u.revoked_on is None) > 1
 
     @property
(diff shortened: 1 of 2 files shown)


parsec/core/gui/workspace_button.py
```diff
@@ -122,7 +122,7 @@ def __init__(
 
     @property
     def is_shared(self):
-        return len(self.users_roles) > 1
+        return sum(1 for _, u in self.users_roles.values() if u.revoked_on is None) > 1
```
touilleMan, line 125: Would be good to add a comment here to underline we need this only to filter out users that have been revoked since the last workspace re-encryption

tests/core/gui/test_workspace_sharing.py
```diff
@@ -288,3 +290,48 @@ def _no_workspace_listed():
...
+    wid = await core.user_fs.workspace_create("Workspace")
+
+    def _workspace_not_shared_listed():
+        assert w_w.layout_workspaces.count() == 1
+        wk_button = w_w.layout_workspaces.itemAt(0).widget()
+        assert isinstance(wk_button, WorkspaceButton)
+        assert wk_button.label_title.text() == "Workspace (private)"
+        assert wk_button.label_title.toolTip() == "Workspace (private)"
+        assert not wk_button.is_shared
+        wk_button.name == "Workspace"
```
touilleMan, line 313: assert missing ?

tests/core/gui/test_workspace_sharing.py
```diff
@@ -288,3 +290,48 @@ def _no_workspace_listed():
...
+
+    await aqtbot.wait_until(_workspace_not_shared_listed, timeout=2000)
```
touilleMan, line 315: timeout need to be overwritten ? (that's kind of fishy to me... ^^)

touilleMan (approved):

tests/core/gui/test_workspace_sharing.py
Max-7: Not really. We have a timer to update the workspace list set to 1s to avoid making to many refreshes. I guess it would be possible to overwrite the timer but I find it easier to just set the timeout to 2s.

PR #1399, opened 2020-08-26
Gui file link on opened organization

### LINKED ISSUE #1398: GUI: A new tab is opened on parsec file link click
A new login tab is open when clicking a parsec file link even if the current tab is from the same organization.

AntoineDupre:
fix #1398

### DIFF
--- a/parsec/core/gui/main_window.py
+++ b/parsec/core/gui/main_window.py
@@ -492,7 +492,6 @@ class MainWindow(QMainWindow, Ui_MainWindow):
             tab.show_login_widget()
             self.on_tab_state_changed(tab, "login")
             self.switch_to_tab(self.tab_center.count() - 1)
-            idx = self.tab_center.count() - 1
 
     def go_to_file_link(self, action_addr):
         devices = list_available_devices(self.config.config_dir)
@@ -554,6 +553,18 @@ class MainWindow(QMainWindow, Ui_MainWindow):
         )
         self.switch_to_login_tab()
 
+    def show_create_org_widget(self, action_addr):
+        self.switch_to_login_tab()
+        self._on_create_org_clicked(action_addr)
+
+    def show_claim_user_widget(self, action_addr):
+        self.switch_to_login_tab()
+        self._on_claim_user_clicked(action_addr)
+
+    def show_claim_device_widget(self, action_addr):
+        self.switch_to_login_tab()
+        self._on_claim_device_clicked(action_addr)
+
     def add_instance(self, start_arg: Optional[str] = None):
         action_addr = None
         if start_arg:
@@ -562,32 +573,25 @@ class MainWindow(QMainWindow, Ui_MainWindow):
             except ValueError as exc:
                 show_error(self, _("TEXT_INVALID_URL"), exception=exc)
 
-        # Do not open a new logging widget if the organisation is already opened
-        if (
-            action_addr
-            and isinstance(action_addr, BackendOrganizationFileLinkAddr)
-            and self.tab_center.count()
-        ):
-            self.go_to_file_link(action_addr)
-            return
-
-        self.switch_to_login_tab()
-
         self.show_top()
-        if action_addr and isinstance(action_addr, BackendOrganizationFileLinkAddr):
-            # Organization is not connected, login is required
-            return
-        elif action_addr:
-            if isinstance(action_addr, BackendOrganizationBootstrapAddr):
-                self._on_create_org_clicked(action_addr)
-            elif isinstance(action_addr, BackendInvitationAddr):
-                if action_addr.invitation_type == InvitationType.USER:
-                    self._on_claim_user_clicked(action_addr)
-                elif action_addr.invitation_type == InvitationType.DEVICE:
-                    self._on_claim_device_clicked(action_addr)
-                else:
-                    show_error(self, _("TEXT_INVALID_URL"))
-                    return
+        if not action_addr:
+            self.switch_to_login_tab()
+        elif isinstance(action_addr, BackendOrganizationFileLinkAddr):
+            self.go_to_file_link(action_addr)
+        elif isinstance(action_addr, BackendOrganizationBootstrapAddr):
+            self.show_create_org_widget(action_addr)
+        elif (
+            isinstance(action_addr, BackendInvitationAddr)
+            and action_addr.invitation_type == InvitationType.USER
+        ):
+            self.show_claim_user_widget(action_addr)
+        elif (
+            isinstance(action_addr, BackendInvitationAddr)
+            and action_addr.invitation_type == InvitationType.DEVICE
+        ):
+            self.show_claim_device_widget(action_addr)
+        else:
+            show_error(self, _("TEXT_INVALID_URL"))
 
     def close_current_tab(self, force=False):
         if self.tab_center.count() == 1:
--- a/parsec/core/gui/main_window.py
+++ b/parsec/core/gui/main_window.py
@@ -483,6 +483,17 @@ class MainWindow(QMainWindow, Ui_MainWindow):
         if not ParsecApp.has_active_modal():
             self.tab_center.setCurrentIndex(idx)
 
+    def switch_to_login_tab(self):
+        idx = self._get_login_tab_index()
+        if idx != -1:
+            self.switch_to_tab(idx)
+        else:
+            tab = self.add_new_tab()
+            tab.show_login_widget()
+            self.on_tab_state_changed(tab, "login")
+            self.switch_to_tab(self.tab_center.count() - 1)
+            idx = self.tab_center.count() - 1
+
     def go_to_file_link(self, action_addr):
         devices = list_available_devices(self.config.config_dir)
         found_org = False
@@ -541,9 +552,7 @@ class MainWindow(QMainWindow, Ui_MainWindow):
                 organization=action_addr.organization_id
             ),
         )
-        idx = self._get_login_tab_index()
-        if idx != -1:
-            self.switch_to_tab(idx)
+        self.switch_to_login_tab()
 
     def add_instance(self, start_arg: Optional[str] = None):
         action_addr = None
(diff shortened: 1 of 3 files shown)


touilleMan:
fix #1398

parsec/core/gui/main_window.py
```diff
@@ -483,6 +483,17 @@ def switch_to_tab(self, idx):
...
+    def switch_to_login_tab(self):
+        idx = self._get_login_tab_index()
+        if idx != -1:
+            self.switch_to_tab(idx)
+        else:
+            tab = self.add_new_tab()
+            tab.show_login_widget()
+            self.on_tab_state_changed(tab, "login")
+            self.switch_to_tab(self.tab_center.count() - 1)
+            idx = self.tab_center.count() - 1
```
touilleMan, line 495: What's the need for getting `idx` here ?

parsec/core/gui/main_window.py
```diff
@@ -553,19 +562,21 @@ def add_instance(self, start_arg: Optional[str] = None):
...
+            action_addr
+            and isinstance(action_addr, BackendOrganizationFileLinkAddr)
+            and self.tab_center.count()
+        ):
+            self.go_to_file_link(action_addr)
+            return
+
+        self.switch_to_login_tab()
 
         self.show_top()
```
touilleMan, line 576: We don't do `self.show_top()` in case of file link addr and logged in organization ? wouldn't this cause some issues if te parsec gui is in background or minimized when clicking on the link ?

parsec/core/gui/main_window.py
```diff
@@ -553,19 +562,21 @@ def add_instance(self, start_arg: Optional[str] = None):
...
-            tab = self.add_new_tab()
-            tab.show_login_widget()
-            self.on_tab_state_changed(tab, "login")
-            self.switch_to_tab(self.tab_center.count() - 1)
-            idx = self.tab_center.count() - 1
+        # Do not open a new logging widget if the organisation is already opened
+        if (
+            action_addr
+            and isinstance(action_addr, BackendOrganizationFileLinkAddr)
+            and self.tab_center.count()
```
touilleMan, line 569: why the check on `self.tab_center.count()` given we iterate on this count in `got_to_file_link` ?

(and is there actual case where we end up with no widget in `tab_center` ? I though we would at least always have a logging widget present there...)

parsec/core/gui/main_window.py
```diff
@@ -553,19 +562,21 @@ def add_instance(self, start_arg: Optional[str] = None):
...
+        ):
+            self.go_to_file_link(action_addr)
+            return
+
+        self.switch_to_login_tab()
 
         self.show_top()
         if action_addr and isinstance(action_addr, BackendOrganizationFileLinkAddr):
```
touilleMan, line 577: I seems weird to have this check which is really similar to the one above, do you think we could factorize this ?
typically to end up with something like:
```python
if not action_addr:
    self.show_login_widget()  # Display login, creating the widget if neeeded
elif isinstance(action_addr, BackendOrganizationFileLinkAddr):
    self.go_to_file_link(action_addr)  # Find the wiget logged in with the organization or fallback to `self.show_login_widget(available_device_for_the_needed_organization, start_arg)` with an additional dialog
elif isintance(action_addr, BackendOrganizationBootstrapAddr):
    self.show_create_org_widget(action_addr)
elif isintsance(..):  # Similar stuff for device/user invitation urls
...
else:
    show_error(self, _("TEXT_INVALID_URL"))
```

parsec/core/gui/main_window.py
```diff
@@ -483,6 +483,17 @@ def switch_to_tab(self, idx):
         if not ParsecApp.has_active_modal():
             self.tab_center.setCurrentIndex(idx)
 
+    def switch_to_login_tab(self):
```
touilleMan, line 486: would be good to have an optional parameter to be able to provide the available device that sould be pre-selected and an `start_arg` parameter
This way clicking on a link while not connected would redirect to the connection screen then do the expected behavior once the user has typed his password ;-)

(this maybe a too big change for this PR btw ^^)

parsec/core/gui/main_window.py
```diff
@@ -553,19 +562,21 @@ def add_instance(self, start_arg: Optional[str] = None):
...
-        idx = self._get_login_tab_index()
-        if idx != -1:
-            self.switch_to_tab(idx)
-        else:
-            tab = self.add_new_tab()
-            tab.show_login_widget()
-            self.on_tab_state_changed(tab, "login")
-            self.switch_to_tab(self.tab_center.count() - 1)
-            idx = self.tab_center.count() - 1
+        # Do not open a new logging widget if the organisation is already opened
```
touilleMan, line 565: Not sure about the comment here: if we don't go into the if we open the logging widget, and if we go into the if we call a function that maybe open the logging widget !

So I guess this comment would be better within the `go_to_file_link` function ?

tests/core/gui/test_main_window.py
```diff
@@ -0,0 +1,194 @@
...
+@pytest.fixture
+async def logged_gui_with_files(
+    aqtbot, running_backend, backend, autoclose_dialog, logged_gui, bob, monkeypatch
+):
+    w_w = await logged_gui.test_switch_to_workspaces_widget()
+
+    assert logged_gui.tab_center.count() == 1
+
+    monkeypatch.setattr(
+        "parsec.core.gui.workspaces_widget.get_text_input", lambda *args, **kwargs: ("w1")
```
touilleMan, line 19: Why the parenthesis around `("w1")` ? it looks like we are returning a tuple so I think we should remove them

tests/core/gui/test_main_window.py
```diff
@@ -0,0 +1,194 @@
...
+    )
+    monkeypatch.setattr(
+        "parsec.core.gui.files_widget.get_text_input", lambda *args, **kwargs: ("dir1")
+    )
+
+    await aqtbot.mouse_click(w_w.button_add_workspace, QtCore.Qt.LeftButton)
+
+    def workspace_button_ready():
```
touilleMan, line 27: nit-picking: start the private function name with an underscore ?

tests/core/gui/test_main_window.py
```diff
@@ -0,0 +1,194 @@
...
+        assert w_w.layout_workspaces.count() == 1
+        wk_button = w_w.layout_workspaces.itemAt(0).widget()
+        assert not isinstance(wk_button, QtWidgets.QLabel)
+
+    await aqtbot.wait_until(workspace_button_ready)
+
+    f_w = logged_gui.test_get_files_widget()
+    wk_button = w_w.layout_workspaces.itemAt(0).widget()
+    async with aqtbot.wait_exposed(f_w), aqtbot.wait_signal(f_w.folder_changed):
```
touilleMan, line 36: we should not use `wait_signal` in the tests because it is cumbersome and buggy (most of the time we are waiting for the callback binded to the event to be executed, but we wait for the event in spite of a better thing to wait on)

```python
def _wait_files_widget_ready():
    assert f_w.isVisible()
    assert f_w.table_files.rowCount() == 4  # Wait for the list of files to display the correct number of items
await aqtbot.wait_until(_wait_files_widget_ready)
```

tests/core/gui/test_main_window.py
```diff
@@ -0,0 +1,194 @@
...
+        await aqtbot.mouse_click(wk_button, QtCore.Qt.LeftButton)
+
+    def _entry_available():
+        assert f_w.workspace_fs.get_workspace_name() == "w1"
+        assert f_w.table_files.rowCount() == 1
```
touilleMan, line 41: Damn ! I've spent 5mn crafting the example in the comment above, just to find it here ! :sob:

tests/core/gui/test_main_window.py
```diff
@@ -0,0 +1,194 @@
...
+    await aqtbot.wait_until(_entry_available)
+
+    def folder_ready():
+        assert f_w.isVisible()
+        assert f_w.table_files.rowCount() == 2
+        folder = f_w.table_files.item(1, 1)
+        assert folder
+        assert folder.text() == "dir1"
+
+    async with aqtbot.wait_signal(f_w.folder_stat_success, timeout=3000):
```
touilleMan, line 52: Why the longer-than-defalt timeout here ? (it's always very fishy to have something like this...)

tests/core/gui/test_main_window.py
```diff
@@ -0,0 +1,194 @@
...
+@pytest.mark.trio
+async def test_file_link(
+    aqtbot, running_backend, backend, autoclose_dialog, logged_gui_with_files, bob, monkeypatch
+):
+    logged_gui, w_w, f_w = logged_gui_with_files
+    url = BackendOrganizationFileLinkAddr.build(
+        f_w.core.device.organization_addr, f_w.workspace_fs.workspace_id, f_w.current_directory
+    )
+
+    class AvailableDevice:
```
touilleMan, line 77: Not a fan of this kind of mock given it makes assumptions on the underlying implementation.

I think it would be better to create a fixture that returns a real `AvailableDevice` (which can be provided with a dummy file path for the moment to keep it implementation simple)

```python
@pytest.fixture
def alice_available_device(alice, tmpdir):
    key_file_path = Path(tmpdir) / "alice_device.key"
    # Given the fixture never provides the password, it's pretty pointless to
    # really save the device, hence I think we can just provide a dummy Path
    # and be gone with it ;-)
    _save_device_with_password(
    	key_file=key_file_path,
    	device=alice,
    	password="leJambonC'estTropBon"
    )
    return AvailableDevice(
	    key_file_path=key_file_path,
	    organization_id=alice.organization_id,
	    device_id=alice.device_id,
	    human_handle=alice.human_handle,
	    device_label=alice.device_label,
	    slug=alice.slug,
    )
```

touilleMan (changes requested):

parsec/core/gui/main_window.py
AntoineDupre: it's pointless, i removed it

parsec/core/gui/main_window.py
AntoineDupre: It's a mistake, `show_top() ` is now always called (for all kind of parsec link)

parsec/core/gui/main_window.py
AntoineDupre: that's true, this check is pointless,  it has been refactored

parsec/core/gui/main_window.py
AntoineDupre: That's better, i refactored it to follow this logic (i also updated unit test to cover all kind of link)

parsec/core/gui/main_window.py
AntoineDupre: removed, as you mentioned, this is already handled in `go_to_file_link`

tests/core/gui/test_main_window.py
AntoineDupre: :fearful: oops

tests/core/gui/test_main_window.py
AntoineDupre: fixed, thanks

tests/core/gui/test_main_window.py
AntoineDupre: As you notice it's done bellow, so in fact this check is pointless.

tests/core/gui/test_main_window.py
AntoineDupre: sorry :(

tests/core/gui/test_main_window.py
AntoineDupre: In fact this entire test is pointless because there is bellow some `wait_until` assertion

tests/core/gui/test_main_window.py
AntoineDupre: Thanks, it definitely better

parsec/core/gui/main_window.py
AntoineDupre: I opened an new issue for this #1405

touilleMan (approved):

PR #1401, opened 2020-08-26
Fuse and winfsp mountpoints now report 0 MB used over 1 TB

vxgmichel:
The winfsp and fuse mountpoints now always report 0 MB used over a 1 TB capacity. 

Those values are arbitrary but useful to the operating system, especially OSX.

### DIFF
--- a/parsec/core/mountpoint/fuse_operations.py
+++ b/parsec/core/mountpoint/fuse_operations.py
@@ -80,6 +80,18 @@ class FuseOperations(LoggingMixIn, Operations):
     def init(self, path: FsPath):
         pass
 
+    def statfs(self, path: FsPath):
+        # We have currently no way of easily getting the size of workspace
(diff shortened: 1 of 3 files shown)


parsec/core/mountpoint/fuse_operations.py
```diff
@@ -80,6 +80,18 @@ def schedule_exit(self):
     def init(self, path: FsPath):
         pass
 
+    def statfs(self, path: FsPath):
+        # We have currently no way of easily getting the size of workspace
+        # Also, the total size of a workspace is not limited
+        # For the moment let's settle on 0 MB used for 1 TB available
+        return {
+            "f_bsize": 512 * 1024,  # 512 K, i.e the default block size
```
touilleMan, line 88: 512 KB ?

touilleMan (approved):
image

PR #1403, opened 2020-08-27
[GUI] Allows creation of an org on a custom backend

### LINKED ISSUE #1390: Allow spontaneous bootstrap on a custom metadata server

Max-7:
Closes #1390

### DIFF
--- a/parsec/core/gui/tr/parsec_en.po
+++ b/parsec/core/gui/tr/parsec_en.po
@@ -7,8 +7,8 @@ msgid ""
 msgstr ""
 "Project-Id-Version: PROJECT VERSION\n"
 "Report-Msgid-Bugs-To: EMAIL@ADDRESS\n"
-"POT-Creation-Date: 2020-08-25 09:56+0200\n"
-"PO-Revision-Date: 2020-08-25 09:55+0200\n"
+"POT-Creation-Date: 2020-08-27 11:08+0200\n"
+"PO-Revision-Date: 2020-08-27 11:09+0200\n"
 "Last-Translator: \n"
 "Language: en\n"
 "Language-Team: en <LL@li.org>\n"
@@ -314,21 +314,6 @@ msgstr "Only a workspace owner can reencrypt a workspace."
 msgid "TEXT_WORKSPACE_CURRENTLY_REENCRYPTING_TOOLTIP"
 msgstr "Reencrypting the workspace"
 
-msgid "TEXT_WORKPACE_REENCRYPT_OFFLINE_ERROR"
-msgstr "Reencryption failure, can not reach the server."
-
-msgid "TEXT_WORKPACE_REENCRYPT_ACCESS_ERROR"
-msgstr "Reencryption failure, you don't have permission to reencrypt the workspace."
-
-msgid "TEXT_WORKPACE_REENCRYPT_NOT_FOUND_ERROR"
-msgstr "Reencryption failure, workspace not found."
-
-msgid "TEXT_WORKPACE_REENCRYPT_FS_ERROR"
-msgstr "Reencryption failure, workspace already in maintenance."
-
-msgid "TEXT_WORKPACE_REENCRYPT_UNKNOWN_ERROR"
-msgstr "Reencryption failure, an unexpected error occured."
-
 msgid "TEXT_WORKSPACE_IS_PRIVATE"
 msgstr "private"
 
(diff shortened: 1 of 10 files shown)


parsec/core/gui/tr/parsec_en.po
```diff
@@ -514,6 +499,23 @@ msgstr "Reencrypt the workspace"
...
+"workspace."
+
+msgid "TEXT_WORKPACE_REENCRYPT_NOT_FOUND_ERROR"
+msgstr "Reencryption failure, workspace not found."
+
+msgid "TEXT_WORKPACE_REENCRYPT_FS_ERROR"
+msgstr "Reencryption failure, workspace already in maintenance."
+
+msgid "TEXT_WORKSPACE_REENCRYPT_UNKOWN_ERROR"
+msgstr ""
```
touilleMan, line 517: not translated, is it expected ?

parsec/core/gui/tr/parsec_en.po
```diff
@@ -2332,3 +2340,7 @@ msgstr "Create a new workspace"
 
 msgid "ACTION_WORKSPACE_GOTO_FILE_LINK"
 msgstr "Go to a file using a link"
+
+#~ msgid "TEXT_WORKPACE_REENCRYPT_UNKNOWN_ERROR"
```
touilleMan, line 2344: Dead code to remove ?

parsec/core/gui/validators.py
```diff
@@ -38,9 +38,7 @@ def validate(self, string, pos):
             BackendAddr.from_url(string)
             return QValidator.Acceptable, string, pos
         except ValueError:
-            if pos > 3:
-                return QValidator.Invalid, string, pos
-            return QValidator.Intermediate, string, pos
+            return QValidator.Invalid, string, pos
```
touilleMan, line 41: The remove of the `pos > 3 check` comes from changes in https://github.com/Scille/parsec-cloud/pull/1402 ?

tests/core/gui/test_create_organization.py
```diff
@@ -278,3 +284,88 @@ def _modal_shown():
...
+    gui, aqtbot, running_backend, catch_create_org_widget, autoclose_dialog
+):
+    # The org creation window is usually opened using a sub-menu.
+    # Sub-menus can be a bit challenging to open in tests so we cheat
+    # using the keyboard shortcut Ctrl+N that has the same effect.
+    await aqtbot.key_clicks(gui, "n", QtCore.Qt.ControlModifier)
+
+    co_w = await catch_create_org_widget()
+    co_w.config = co_w.config.evolve(
+        preferred_org_creation_backend_addr=BackendAddr.from_url("parsec://localhost:1337")
```
touilleMan, line 302: Providing a dummy port on localhost should be done with the `unused_tcp_port` to ensure the port is really not in use.

Beside I think we should think of a better way to patch the config here: the config is supposed to be immutable so it would be better to first craft a config, then use to create the gui (this is usually done with the `customize_fixtures` decorator).

And last but not least, `preferred_org_creation_backend_addr` field from the `core_config` is already configured to be the `backend_addr` fixture. This is good because we don't risk any request to external addresses and we can use the offline system to achieve our goal "with rare elegance"®:
```python
@pytest.fixture
def preferred_org_creation_backend_addr(backend_addr):
    return addr_with_device_subdomain(backend_addr, 'preferred_org_creation')  # parsec://preferred_org_creation.example.com:9999?use_ssl=false

@pytest.fixture
def core_config(tmpdir, preferred_org_creation_backend_addr):
    # We must provide a custom backend address with a subname here given offline(backend_addr) would put offline all subdomain
    tmpdir = Path(tmpdir)
    return CoreConfig(
        [...]
        preferred_org_creation_backend_addr=preferred_org_creation_backend_addr,
    )

def tost_turbo_stuffing_the_stuff(preferred_org_creation_backend_addr, backend_addr):
    custom_addr = addr_with_device_subdomain(backend_addr, 'custom')  # parsec://custom.example.com:9999?use_ssl=false
    with offline(preferred_org_creation_backend_addr):
        # From now on, only preferred_org_creation.example.com:9999 is offline, so we can use custom.example.com:9999
        assert_org_creation_with_preferred_addr_doesnt_work()
        assert_org_creation_with_custom_addr_work(custom_addr)
```

touilleMan (changes requested):

parsec/core/gui/validators.py
Max-7: I think the `pos > 3` dated from way back when we didn't even have a `parsec://` scheme. No idea what was the point of it so I just removed it.

PR #1404, opened 2020-08-27
Fix email user invite generation

bitlogik:
There was a mismatch between the placeholder variable names in the templates and in the provided arguments in the code.


touilleMan:
thanks 👍

PR #1427, opened 2020-09-04
Flaky gui test event bus none

AntoineDupre:
Try to fix ci errors:
[Reset error](https://dev.azure.com/Scille/parsec/_build/results?buildId=3487&view=logs&j=5ba45810-c0e9-5996-abe7-b2cb0c9b6baa&t=98a05ba0-83e8-5985-e9d7-1cce7a13f66e)
[No core](https://dev.azure.com/Scille/parsec/_build/results?buildId=3382&view=logs&j=5ba45810-c0e9-5996-abe7-b2cb0c9b6baa&t=98a05ba0-83e8-5985-e9d7-1cce7a13f66e)
[Invitation widget timeout](https://dev.azure.com/Scille/parsec/_build/results?buildId=3603&view=logs&j=5ba45810-c0e9-5996-abe7-b2cb0c9b6baa&t=98a05ba0-83e8-5985-e9d7-1cce7a13f66e)
[increase mountpoint timeout](https://dev.azure.com/Scille/parsec/_build/results?buildId=3611&view=logs&j=5ba45810-c0e9-5996-abe7-b2cb0c9b6baa&t=98a05ba0-83e8-5985-e9d7-1cce7a13f66e)

Gui tests ran 10 times without failure there:
https://dev.azure.com/Scille/parsec/_build/results?buildId=3618&view=logs&j=5ba45810-c0e9-5996-abe7-b2cb0c9b6baa

still missing a fix:
https://dev.azure.com/Scille/parsec/_build/results?buildId=3622&view=logs&j=5ba45810-c0e9-5996-abe7-b2cb0c9b6baa&t=98a05ba0-83e8-5985-e9d7-1cce7a13f66e

### DIFF
--- a/parsec/core/gui/instance_widget.py
+++ b/parsec/core/gui/instance_widget.py
@@ -212,20 +212,19 @@ class InstanceWidget(QWidget):
 
     def show_central_widget(self):
         self.clear_widgets()
-        try:
-            central_widget = CentralWidget(
-                self.core,
-                self.core_jobs_ctx,
-                self.core.event_bus,
-                systray_notification=self.systray_notification,
-                parent=self,
-            )
-        except AttributeError:
-            # The core can be set to None at any time if do_run_core get an error, is cancelled or
-            # terminate.
-            if self.core is None:
-                return
-            raise
+        # The core can be set to None at any time if do_run_core get an error, is cancelled or
+        # terminate.
+        core = self.core
+        core_jobs_ctx = self.core_jobs_ctx
+        if core is None or core_jobs_ctx is None:
+            return
+        central_widget = CentralWidget(
+            core,
+            core_jobs_ctx,
+            core.event_bus,
+            systray_notification=self.systray_notification,
+            parent=self,
+        )
         self.layout().addWidget(central_widget)
         central_widget.logout_requested.connect(self.logout)
         central_widget.show()
--- a/parsec/core/gui/instance_widget.py
+++ b/parsec/core/gui/instance_widget.py
@@ -212,13 +212,20 @@ class InstanceWidget(QWidget):
 
     def show_central_widget(self):
         self.clear_widgets()
-        central_widget = CentralWidget(
-            self.core,
-            self.core_jobs_ctx,
-            self.core.event_bus,
-            systray_notification=self.systray_notification,
-            parent=self,
-        )
+        try:
+            central_widget = CentralWidget(
+                self.core,
+                self.core_jobs_ctx,
+                self.core.event_bus,
+                systray_notification=self.systray_notification,
+                parent=self,
+            )
+        except AttributeError:
+            # The core can be set to None at any time if do_run_core get an error, is cancelled or
+            # terminate.
+            if self.core is None:
+                return
+            raise
         self.layout().addWidget(central_widget)
         central_widget.logout_requested.connect(self.logout)
         central_widget.show()
--- a/tests/core/gui/test_create_organization.py
+++ b/tests/core/gui/test_create_organization.py
@@ -81,7 +81,7 @@ async def test_create_organization(
     # The org creation window is usually opened using a sub-menu.
     # Sub-menus can be a bit challenging to open in tests so we cheat
     # using the keyboard shortcut Ctrl+N that has the same effect.
-    await aqtbot.key_click(gui, "n", QtCore.Qt.ControlModifier)
+    await aqtbot.key_click(gui, "n", QtCore.Qt.ControlModifier, 200)
 
     co_w = await catch_create_org_widget()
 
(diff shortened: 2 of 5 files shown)


Max-7 (approved):

parsec/core/gui/instance_widget.py
```diff
@@ -212,13 +212,20 @@ def login_with_password(self, key_file, password):
...
+        try:
+            central_widget = CentralWidget(
+                self.core,
+                self.core_jobs_ctx,
+                self.core.event_bus,
+                systray_notification=self.systray_notification,
+                parent=self,
+            )
+        except AttributeError:
+            # The core can be set to None at any time if do_run_core get an error, is cancelled or
```
touilleMan, line 224: I think we should take some time to think about the states of this `InstanceWidget`

The fact that `self.core` can be set to `None` at any time makes things much more complicated.
For instance this code is not thread safe given the check on self.core not None is done in a different step than the access: https://github.com/Scille/parsec-cloud/blob/710ca19c946bface10616637604c862b218c3210/parsec/core/gui/instance_widget.py#L82-L85

On top of that we could have subtle error in:
https://github.com/Scille/parsec-cloud/blob/710ca19c946bface10616637604c862b218c3210/parsec/core/gui/instance_widget.py#L217-L218

If `self.core` gets executed after `self.core.event_bus` (for instance if we refactor the code following my previous comment ^^), we can end up with `self.core != None` when we access `self.core.event_bus`, then `self.core == None` when we retrieve the core...

So I wonder if we shouldn't avoid entirely to reset `self.core`. Considering the tab containing the `InstanceWidget` is going to be closed, we should be able to state that this class cannot be re-used (so it lifespan would be `creation -> login -> error`, `creation -> login -> ready -> error` or `creation -> login -> ready -> logout`).

parsec/core/gui/instance_widget.py
```diff
@@ -212,13 +212,20 @@ def login_with_password(self, key_file, password):
...
-        )
+        try:
+            central_widget = CentralWidget(
+                self.core,
+                self.core_jobs_ctx,
+                self.core.event_bus,
+                systray_notification=self.systray_notification,
+                parent=self,
+            )
+        except AttributeError:
```
touilleMan, line 223: It would be better to only try/except the access to the event_bus:
```python
try:
    core_event_bus = self.core.event_bus
except AttributeError:
    [...]
```

tests/core/gui/test_workspaces.py
```diff
@@ -235,7 +235,7 @@ def _mounted():
         assert wk_button.button_open.isEnabled()
         assert wk_button.switch_button.isChecked()
 
-    await aqtbot.wait_until(_mounted)
+    await aqtbot.wait_until(_mounted, timeout=2000)
```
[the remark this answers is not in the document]
AntoineDupre: Because it sometimes timeout, by increasing the timeout, i did not manage to reproduce the error.
https://dev.azure.com/Scille/parsec/_build/results?buildId=3611&view=logs&j=5ba45810-c0e9-5996-abe7-b2cb0c9b6baa&t=98a05ba0-83e8-5985-e9d7-1cce7a13f66e

bitlogik:
The root cause of the event_bus None during the test is when a timeout fires and the client is disconnect from the backend server during the initialization "fixture" of the test. Increasing the timeout is mandatory on some points, [especially in test_workspace_sharing](https://github.com/bitlogik/guardata/commit/53fc04e8bbb1d05d5c9353f708a8f244d0731639#diff-fec9aea8d02fafff13f9048e0dc4786f) where the listing of the workspaces (_workspace_shared_listed and similar) can take more than 1s. 

The errors in test_workpaces are different and must be solved to help in this flaky tests issue. https://github.com/Scille/parsec-cloud/issues/1428

tests/core/gui/test_create_organization.py
```diff
@@ -81,7 +81,7 @@ async def test_create_organization(
     # The org creation window is usually opened using a sub-menu.
     # Sub-menus can be a bit challenging to open in tests so we cheat
     # using the keyboard shortcut Ctrl+N that has the same effect.
-    await aqtbot.key_click(gui, "n", QtCore.Qt.ControlModifier)
+    await aqtbot.key_click(gui, "n", QtCore.Qt.ControlModifier, 200)
```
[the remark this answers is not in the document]
AntoineDupre: I notice that short cut signals are not fully bind when the test started.
I did not find something to assert that shortcuts are properly connected in QT. Sending the signal with a short delay makes the test quiet reliable.

parsec/core/gui/instance_widget.py
AntoineDupre:
```
Considering the tab containing the InstanceWidget is going to be closed,
```
This is not the case when user logs out. We could change the logout mecanism.

parsec/core/gui/instance_widget.py
AntoineDupre: I guess, we shall also do some sanity check with `core_jobs_ctx` to be sure that the widget is not built with None object.

PR #1435, opened 2020-09-08
[GUI] Display a spinner while loading the workspaces

### LINKED ISSUE #1432: Add a spinner when loading the list of workspaces

Max-7:
Closes #1432


AntoineDupre (approved):
LGTM

bitlogik:
The method _has_workspaces_displayed_ always returns False. This is because the WorkspaceButton instance is `self.layout_workspaces.itemAt(0).widget()` (else the item is just a QTWidget)

Then, the test for the spinner in _list_workspaces_ should be inverted to : `if not self.has_workspaces_displayed():`

PR #1440, opened 2020-09-09
Remove no longer used apiv1 authenticated mode from the core and remo…

touilleMan:
…ve deprecated `parsec core apiv1` commands from the cli


vxgmichel (approved):
LGTM!

PR #1441, opened 2020-09-09
macOS version

Ironicbay:


parsec/core/gui/parsec_application.py
```diff
@@ -13,6 +16,8 @@ def __init__(self):
         self.setOrganizationName("Scille")
         self.setOrganizationDomain("parsec.cloud")
         self.setApplicationName("Parsec")
+        path = os.path.join(os.path.dirname(sys.modules[__name__].__file__), 'parsec.png')
+        self.setWindowIcon(QIcon(path))
```
Max-7, line 20: The window icon set in the .ui file is not enough? Or is it a problem of the resource file not being compiled on Osx?

Ironicbay: Despite the current settings, the taskbar/dock icon was still the default Python icon. This was the only proper fix I could find to display a custom icon.

Max-7: Could you try `self.setWindowIcon(QIcon(":/logos/images/icons/parsec.png"))` instead? It would at least use the icon in the qrc file instead of using a path. It this doesn't work, then it's a problem of the resource file.

Ironicbay: Works fine ! I'll have to check if this addition is still necessary when we'll have Parsec in `.app` format,  this behaviour might change then.

.azure-pipelines.yml
```diff
@@ -467,3 +467,91 @@ jobs:
...
+  - checkout: self
+    submodules: true
+  - task: UsePythonVersion@0
+    inputs:
+      versionSpec: '$(python.version)'
+  - download: current
+    artifact: wheel
+  - bash: |
+      set -eux
+      cd packaging/macOS
```
touilleMan, line 494: you can use `workingDirectory: $(Build.SourcesDirectory)/packaging/macOS` instead

packaging/macOS/macos_pyinstaller.sh
```diff
@@ -0,0 +1,25 @@
+# Note : pyinstaller doesn't work with python 3.8 yet.
+# see https://github.com/pyinstaller/pyinstaller/issues/4311
+
+cd ../..
```
touilleMan, line 4: cd to relative directory is not a good idea given you don't know from where this script is going to be run:
```sh
$ pwd
/home/menahem
$ ./parsec-cloud/packaging/macOS/macos_pyinstaller.sh
!!! crash because cd ../.. leads to / given we are running the script from /home/menahem/
```

The solution is usually to do 
```
BASEDIR=$(dirname $(readlink -f "$0") )
cd $BASEDIR/../..
```

Besides, instead of using cd, you should check pushd/popd ;-)

packaging/macOS/my_hooks/hook-pendulum.py
```diff
@@ -0,0 +1,22 @@
+# Parsec Cloud (https://parsec.cloud) Copyright (c) AGPLv3 2019 Scille SAS
+
+# ------------------------------------------------------------------
+# Copyright (c) 2020 PyInstaller Development Team.
```
touilleMan, line 4: I think we can remove this license given the code  is really trivial and we modified it

touilleMan (commented):
`parsec/core/gui/parsec.png` should be moved to `parsec/core/resources`

tests/core/mountpoint/test_file_operations.py
```diff
@@ -13,6 +14,7 @@
 
 @pytest.mark.slow
 @pytest.mark.mountpoint
+@pytest.mark.skipif(sys.platform == "darwin", reason="Infinitely looping on macOS")
```
touilleMan, line 17: Should add a TODO here (and in every skipif) to easily retrieve this

tests/backend/test_postgres_db.py
```diff
@@ -24,7 +24,9 @@ async def test_postgresql_connection_not_ok(
     with pytest.raises(OSError) as exc:
         async with backend_factory(config={"db_url": postgresql_url}):
             pass
-    assert "[Errno 111] Connect call failed" in str(exc.value)
+    assert "[Errno 111] Connect call failed" in str(
```
touilleMan, line 27: Would be better to separate the error code between platforms:
```python
if platform.system() == "Darwin":
    errno = 61
else:
    errno = 111
assert f"[Errno {errno}] Connect call failed" in str(exc.value)
```
```

File: parsec/core/mountpoint/manager.py [hunk 7]
```diff
@@ -38,7 +39,7 @@
         import_function("winfspy")
     else:
         import_function("fuse")
-except (ImportError, RuntimeError):
+except (ImportError, RuntimeError, OSError):
```
touilleMan, line 42: why OSError ?

parsec/core/mountpoint/manager.py
```diff
@@ -264,6 +265,16 @@ async def safe_unmount_all(self):
             await self.safe_unmount(workspace_id, timestamp=timestamp)
 
 
+async def cleanup_macos_mountpoint_folder(base_mountpoint_path):
```
touilleMan, line 268: A comment would be welcome to explain why this is needed

parsec/core/mountpoint/manager.py
```diff
@@ -284,6 +295,8 @@ async def mountpoint_manager_factory(
     # Now is a good time to perform some cleanup in the registry
     if os.name == "nt":
         cleanup_parsec_drive_icons()
+    elif os.name == "posix" and _platform == "darwin" and os.path.isdir(base_mountpoint_path):
```
touilleMan, line 298: can't just do `elif sys.platform == "darwin":` ?

and what the need for `os.path.isdir(base_mountpoint_path)` ? is there usecase where it's not the case ?

parsec/core/mountpoint/fuse_runner.py
```diff
@@ -127,6 +130,21 @@ async def fuse_mountpoint_runner(
...
+                        "local": True,
+                        "volname": workspace_fs.get_workspace_name(),
+                        "volicon": Path(resources.__file__).absolute().parent / "parsec.icns",
+                    }
+                    # osxfuse-specific options :
+                    # - local : allows mountpoint to show up correctly in finder (+ desktop)
+                    # - volname : specify volume name (default is OSXFUSE [...])
+                    # - volicon : specify volume icon (default is macOS drive icon)
+
+                elif sys.platform.startswith("linux"):
```
touilleMan, line 145: no need to do startswith since python 3.3 (see https://docs.python.org/3/library/sys.html?highlight=sys#sys.platform), we can do `sys.platform == "linux"` ;-)

parsec/core/mountpoint/fuse_runner.py
```diff
@@ -210,8 +228,11 @@ async def _stop_fuse_thread(
     if fuse_thread_stopped.is_set() or not fuse_thread_started.is_set():
         return
     logger.info("Stopping fuse thread...", mountpoint=mountpoint_path)
-    # Schedule an exit in the fuse operations
-    fuse_operations.schedule_exit()
+    if sys.platform == "darwin":
+        await trio.run_process(["diskutil", "unmount", str(mountpoint_path)])
```
touilleMan, line 232: should explain why this is needed

parsec/core/mountpoint/fuse_runner.py
touilleMan: ...and we can just use `else` in fact here

packaging/macOS/parsec.spec
```diff
@@ -0,0 +1,61 @@
+# -*- mode: python ; coding: utf-8 -*-
```
touilleMan, line 1: 1 - License header is missing
2 - `# -*-` encoding line is not needed since python3 so it can be removed
3 - we should use black on this file just to clean it a bit (you can rename it as `parsec.py`, run black on it, then rename it back to `parsec.spec`)

touilleMan (changes requested):

parsec/core/mountpoint/manager.py
Ironicbay: macOS encounters an OSError here if OSXFUSE isn't installed, and crashes instead of showing the corresponding message. (Same thing in line 68)

parsec/core/mountpoint/manager.py
Ironicbay: I guessed there would possibly be situations where the mountpoint folder doesn't exist, such as the first use of Parsec on a computer.

packaging/macOS/macos_pyinstaller.sh
```diff
@@ -0,0 +1,27 @@
+# Note : pyinstaller doesn't work with python 3.8 yet.
```
touilleMan, line 1: missing `#! /bin/sh`

parsec/core/mountpoint/manager.py
```diff
@@ -284,6 +297,8 @@ async def mountpoint_manager_factory(
     # Now is a good time to perform some cleanup in the registry
     if os.name == "nt":
         cleanup_parsec_drive_icons()
+    elif _platform == "darwin" and os.path.isdir(base_mountpoint_path):
```
touilleMan, line 300: would be better to add a `try/except FileNotFoundError` surrounding the for os.listdir in `cleanup_macos_mountpoint_folder` instead

touilleMan (approved):

PR #1446, opened 2020-09-10
GUI: Force login on file link click with not connected organization

### LINKED ISSUE #1405: GUI: file link with not connected organization
If a parsec file link is clicked and the organization is not open in the parsec gui, the login screen is displayed. After login in, parsec is not switching to the file location

It could be nice to directly from the link display the login widget for this organization (where the user enter the password) and then to follow up after logging to the file link in the good workspace

AntoineDupre:
closes #1405

### DIFF
--- a/newsfragments/1405.feature.rst
+++ b/newsfragments/1405.feature.rst
@@ -1,0 +1,1 @@
+Display login and follow link on not logged organization file link click.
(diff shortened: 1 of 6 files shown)


Max-7 (approved):
LGTM

PR #1450, opened 2020-09-11
Improve test utils.organization

Sithi5:


parsec/test_utils/organization.py
```diff
@@ -116,16 +111,23 @@ async def initialize_test_organization(
...
+                        cmds=alice_cmds,
+                        author=alice_device,
+                        alice_core=alice_core,
+                        bob_core=bob_core,
+                        alice_ws_id=alice_ws_id,
+                        bob_ws_id=bob_ws_id,
+                        additional_users_number=additional_users_number,
+                    )
+                # Add additional random device for alice
+                if additional_devices_number > 0:
```
touilleMan, line 123: we can remove the if check and always call `_add_random_device` given it does nothing but a loop on `range(additional_devices_number )`

parsec/test_utils/organization.py
```diff
@@ -116,16 +111,23 @@ async def initialize_test_organization(
                 # Add additional random users
                 if additional_users_number > 0:
```
touilleMan, line 112: we can remove the if check and always call `_add_random_users` given it does nothing but a loop on `range(additional_users_number)`

parsec/test_utils/organization.py
```diff
@@ -136,51 +138,51 @@ async def initialize_test_organization(
     return (alice_device, other_alice_device, bob_device)
 
 
+async def _add_random_device(cmds, config_dir, password, force, device, additional_devices_number):
+    for _ in range(additional_devices_number):
+        device_label = "device_" + str(uuid4())[:9]
+        new_device = await _register_new_device(cmds=cmds, author=device, device_label=device_label)
+        save_device_with_password(config_dir, new_device, password, force=force)
```
touilleMan, line 145: do we really need to save the devices here ? I would say no given we don't save the users added from `_add_random_users`

touilleMan (approved):

PR #1458, opened 2020-09-15
[GUI] Ask password immediatly if only once device

### LINKED ISSUE #1456: Use defaut account if only one device on the machine
In case the user has only one device, we can skip the account selection process and directly ask for the password.

Max-7:
Closes #1456


AntoineDupre (approved):
LGTM

PR #1460, opened 2020-09-16
Disable validate button by default when chosing a password

### LINKED ISSUE #1459: OK button should be disabled when choosing a password when claiming a user
And possibly a device.

The "OK" button to finalize the user creation process is enabled by default even if no password has been entered. It should be disabled.

Max-7:
Closes #1459


AntoineDupre (approved):
tested; LGTM

PR #1462, opened 2020-09-16
handle the pagination in backend

Sithi5:


Sithi5:
https://github.com/Scille/parsec-cloud/pull/1467 new PR with memory pagination also

PR #1467, opened 2020-09-17
Handle pagination memory

Sithi5:


parsec/backend/memory/user.py
```diff
@@ -285,26 +275,33 @@ async def find_humans(
             )
             for user in data
         ]
-
+        results = sorted(
```
touilleMan, line 278: A comment should be added here to explain the reason (but I think this code is not needed, see my other comment below)

parsec/backend/memory/user.py
```diff
@@ -285,26 +275,33 @@ async def find_humans(
...
-            [res for res in results if res.human_handle],
-            key=lambda r: (r.human_handle.label.lower(), r.user_id.lower()),  # type: ignore
-        )
-        non_humans = sorted(
-            [res for res in results if not res.human_handle], key=lambda r: r.user_id.lower()
-        )
-
-        if omit_non_human:
+        if omit_non_human or query:
+            total = len(results)
```
touilleMan, line 286: `total` and the paginated `results` can be put before the if

parsec/backend/memory/user.py
```diff
@@ -253,26 +253,16 @@ async def find_humans(
     ) -> Tuple[List[HumanFindResultItem], int]:
         org = self._organizations[organization_id]
```
touilleMan, line 254: `query` typing is wrong, should be `Optional[str]`

parsec/backend/memory/user.py
```diff
@@ -253,26 +253,16 @@ async def find_humans(
...
-                    user_terms = (user.user_id.lower(),)
-
-                for qt in query_terms:
-                    if not any(ut.startswith(qt) for ut in user_terms):
-                        break
-                else:
-                    # All query term have match the current user
+            for user in temp_data:
+                if (
+                    str(user.human_handle.email).find(query) != -1
```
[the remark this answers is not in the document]
touilleMan: what about user whith `human_handle` set to `None` ? Do we have some test to ensure the behavior is ok ?

parsec/backend/memory/user.py
```diff
@@ -285,26 +275,33 @@ async def find_humans(
...
+            total = len(results)
+            results = results[(page - 1) * per_page : page * per_page]
+            humans = sorted(
+                [res for res in results if res.human_handle], key=lambda r: r.human_handle.label
+            )
+            non_humans = [res for res in results if not res.human_handle]
             results = [*humans, *non_humans]
 
-        return (results[(page - 1) * per_page : page * per_page], len(results))
+        if not results:
```
touilleMan, line 301: This shouldn't be needed given we can do `total = len(results)` even if results is an empty list

parsec/backend/memory/user.py
```diff
@@ -285,26 +275,33 @@ async def find_humans(
...
-        humans = sorted(
-            [res for res in results if res.human_handle],
-            key=lambda r: (r.human_handle.label.lower(), r.user_id.lower()),  # type: ignore
-        )
-        non_humans = sorted(
-            [res for res in results if not res.human_handle], key=lambda r: r.user_id.lower()
-        )
-
-        if omit_non_human:
+        if omit_non_human or query:
```
touilleMan, line 285: It's not clear why `query` is tested along with `omit_non_human`
I think we should instead to something like this at the beginning of the function:
```
# query is run against human handle field, hence non-human are automatically ignored
omit_non_human = omit_non_human or query is not None
```

touilleMan: As a matter of fact, it would be even better to handle the filter of non-human directly in the main `if query:` (https://github.com/Scille/parsec-cloud/pull/1467/files#diff-549a86cf6e3e6ca50a72cd4c24ea02baR261)
```python
# Retrieve all results
if query:
    results = []
    for user in org.users.values():
        # query is run against human handle field, hence non-human are automatically ignored
        if user.human_handle and str(user.human_handle.email).find(query) != -1 and ... :
            results.append(user)
else:
    results = org.users.values()
    if omit_non_human:
        results = [r for r in results if r.human_handle]
# Handle pagination and ordering
results = order(results, key=...)
paginated_results = results[(page - 1) * per_page : page * per_page]
return paginated_results, len(results)
```

parsec/backend/postgresql/user_queries/find.py
```diff
@@ -29,25 +29,29 @@ def _q_factory(query, omit_revoked):
...
-    user_.revoked_on IS NOT NULL AND user_.revoked_on <= $now
-FROM user_ LEFT JOIN human ON user_.human=human._id
-WHERE
+    user_.revoked_on IS NOT NULL AND user_.revoked_on <= $now,
+    count(*) OVER() AS full_count
+    FROM user_ LEFT JOIN human ON user_.human=human._id
+    WHERE
     user_.organization = { q_organization_internal_id("$organization_id") }
     { " ".join(conditions) }
+    ORDER BY user_id
```
touilleMan, line 52: I'm pretty sure this is broken:

users_id | human_handle
---------|--------------
1            | bob3
2            | bob2
3            | bob1

with `query=bob`, we retrieve the 3 results, but if we ask for a 2 items in the pagination, we will end up with `1, bob3` & `2, bob2`, so we will display to the user:
First page:
- bob3
- bob2

Second page:
- bob1

touilleMan: I think we shouldn't sort when asking without query parameter (no sort means we should get the results in the insertion order which should be in line with what is done in the memory implementation)

When asking with query parameter, we should only do the sort on sort on `human.label`

parsec/backend/postgresql/user_queries/find.py
```diff
@@ -97,45 +105,44 @@ async def query_find_humans(
...
-
-        else:
-            q = _q_human_factory(query=False, omit_revoked=False, omit_non_human=omit_non_human)
-            args = q(organization_id=organization_id, now=pendulum_now())
+        args = q(organization_id=organization_id, now=pendulum_now())
 
     raw_results = await conn.fetch(*args)
-
+    if raw_results:
+        total = raw_results[0][4]
```
touilleMan, line 127: You can access data through the name which is nicer: `raw_results[0]["full_count"]`

parsec/backend/postgresql/user_queries/find.py
```diff
@@ -97,45 +105,44 @@ async def query_find_humans(
...
         HumanFindResultItem(
             user_id=UserID(user_id),
             human_handle=HumanHandle(email=email, label=label),
             revoked=revoked,
         )
-        for user_id, email, label, revoked in raw_results
+        for user_id, email, label, revoked, _ in raw_results
         if email is not None
     ]
     non_humans = [
```
touilleMan, line 139: No need to divide between non-human and human for building results now \o/

parsec/backend/postgresql/user_queries/find.py
```diff
@@ -97,45 +105,44 @@ async def query_find_humans(
...
         HumanFindResultItem(user_id=UserID(user_id), human_handle=None, revoked=revoked)
-        for user_id, email, label, revoked in raw_results
+        for user_id, email, label, revoked, _ in raw_results
         if email is None
     ]
     results = [
-        *sorted(humans, key=lambda x: (x.human_handle.label.lower(), x.user_id.lower())),
+        *sorted(humans, key=lambda r: r.human_handle.label),
```
touilleMan, line 145: and no more need for sorting outside of postgresql also !

tests/backend/user/test_apiv1_human_find.py
```diff
@@ -62,7 +62,7 @@ async def test_isolation_from_other_organization(
     backend, alice, sock_from_other_organization_factory
 ):
     async with sock_from_other_organization_factory(backend) as sock:
-        rep = await human_find(sock, query=alice.user_id)
+        rep = await human_find(sock, query=alice.human_handle.label)
```
touilleMan, line 65: You should also add a call to human_find without `query` param to make sure we only have users from our ow organization this way

touilleMan: And this test also exists in `test_human_find.py` (i.e. the apiv2 version), it should be corrected as well

tests/backend/user/test_apiv1_human_find.py
```diff
@@ -207,55 +203,6 @@ async def test_search_multiple_user_same_human_handle(
     }
 
 
-@pytest.mark.trio
-async def test_search_with_non_humans(access_testbed, organization_factory, local_device_factory):
```
touilleMan, line 211: Here goes the ninja empire :'( https://youtu.be/mhB0jR6xVek?t=342

touilleMan (changes requested):

parsec/backend/memory/user.py
```diff
@@ -253,26 +253,16 @@ async def find_humans(
     ) -> Tuple[List[HumanFindResultItem], int]:
         org = self._organizations[organization_id]
 
+        total = 0
         if query:
+            temp_data = org.users.values()
```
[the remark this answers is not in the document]
Sithi5: done

parsec/backend/memory/user.py
```diff
@@ -253,26 +253,16 @@ async def find_humans(
...
-                    if not any(ut.startswith(qt) for ut in user_terms):
-                        break
-                else:
-                    # All query term have match the current user
+            for user in temp_data:
+                if (
+                    str(user.human_handle.email).find(query) != -1
+                    or str(user.human_handle.label).find(query) != -1
+                ):
                     data.append(user)
```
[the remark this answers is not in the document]
Sithi5: done

parsec/backend/memory/user.py
Sithi5: done

parsec/backend/memory/user.py
Sithi5: done

parsec/backend/memory/user.py
Sithi5: Done

parsec/backend/postgresql/user_queries/find.py
```diff
@@ -56,6 +60,10 @@ def _q_human_factory(query, omit_revoked, omit_non_human):
 async def query_find(
     conn, organization_id: OrganizationID, query: str, page: int, per_page: int, omit_revoked: bool
 ) -> Tuple[List[UserID], int]:
+    # if page >= 1:
```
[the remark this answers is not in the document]
Sithi5: Removed

parsec/backend/postgresql/user_queries/find.py
Sithi5: Done

parsec/backend/memory/user.py
touilleMan: but why removing the default value ?

parsec/backend/memory/user.py
```diff
@@ -283,28 +275,23 @@ async def find_humans(
...
         if omit_revoked:
             results = [res for res in results if not res.revoked]
-
-        # PostgreSQL does case insensitive sort
-        humans = sorted(
-            [res for res in results if res.human_handle],
-            key=lambda r: (r.human_handle.label.lower(), r.user_id.lower()),  # type: ignore
+        # Handle pagination
+        paginated_results = results[(page - 1) * per_page : page * per_page]
+        # Sort human by label
```
touilleMan, line 285: Sorting is only when query is in use, we should move this in the if part

touilleMan: As a matter of fact, it's especially important to do the sort before the pagination ! Is there some test to ensure there is no trouble with this ?

tests/backend/user/test_apiv1_human_find.py
```diff
@@ -8,6 +8,71 @@
 from tests.common import freeze_time, customize_fixtures
 
 
+class NonDeterministicOrderedResults(list):
```
touilleMan, line 11: A comment would be good to explain the need for this special kind of list

tests/backend/user/test_apiv1_human_find.py
```diff
@@ -183,17 +247,14 @@ async def test_search_multiple_user_same_human_handle(
...
-        "per_page": 100,
-        "page": 1,
-        "total": 3,
-    }
+    expected = [
+        {"user_id": nick1.user_id, "human_handle": nick1.human_handle, "revoked": True},
+        {"user_id": nick2.user_id, "human_handle": nick2.human_handle, "revoked": True},
+        {"user_id": nick3.user_id, "human_handle": nick3.human_handle, "revoked": False},
+    ]
+    assert rep["results"] == NonDeterministicOrderedResults(
```
touilleMan, line 255: you should test the entire rep:
```python
assert rep == {
    "status": "ok",
    "results": NonDeterministicOrderedResults([
        {"user_id": nick2.user_id, "human_handle": nick2.human_handle, "revoked": True},
        {"user_id": nick1.user_id, "human_handle": nick1.human_handle, "revoked": True},
        {"user_id": nick3.user_id, "human_handle": nick3.human_handle, "revoked": False},
    ], key=lambda x: x['human_handle']),
    "per_page": 100,
    "page": 1,
    "total": 3,
}
```

tests/backend/user/test_apiv1_human_find.py
```diff
@@ -293,9 +305,9 @@ async def test_pagination(access_testbed, organization_factory, local_device_fac
     assert rep == {
         "status": "ok",
         "results": [
-            {"user_id": blacky.user_id, "human_handle": blacky.human_handle, "revoked": False},
             {"user_id": godfrey1.user_id, "human_handle": godfrey1.human_handle, "revoked": False},
-            {"user_id": mike.user_id, "human_handle": mike.human_handle, "revoked": False},
+            {"user_id": richard.user_id, "human_handle": richard.human_handle, "revoked": False},
```
touilleMan, line 309: no need for `NonDeterministicOrderedResults` here ?

tests/backend/user/test_apiv2_human_find.py
```diff
@@ -8,6 +8,71 @@
 from tests.backend.common import human_find
 
 
+class NonDeterministicOrderedResults(list):
```
touilleMan, line 11: Would be better to keep this here and have `test_apiv1_human_find.py` import it from there instead of duplicate the code

tests/backend/user/test_apiv2_human_find.py
```diff
@@ -311,3 +323,14 @@ async def test_bad_args(access_testbed, organization_factory, local_device_facto
     for bad in [{"page": 0}, {"per_page": 0}, {"per_page": 101}]:
         rep = await human_find(sock, **bad)
         assert rep["status"] == "bad_message"
+
+
+@pytest.mark.trio
+async def test_human_handle_to_none(access_testbed, organization_factory, local_device_factory):
```
touilleMan, line 329: Better call it "non-human user" instead of "human handle to none" (that what is used in the rest of the codebase)

tests/backend/user/test_apiv2_human_find.py
```diff
@@ -311,3 +323,14 @@ async def test_bad_args(access_testbed, organization_factory, local_device_facto
...
+    binder, org, godfrey1, sock = access_testbed
+
+    richard = local_device_factory(base_human_handle=None, org=org)
```
touilleMan, line 332: you can use the awesome `@customize_fixtures(alice_has_human_handle=False)` decorator on the test to have fixture `alice` and `alice2` without human_handle (works also for bob and adam) !

tests/backend/user/test_apiv2_human_find.py
touilleMan: so the test should be called `test_find_with_query_ignore_non_human`

tests/backend/user/test_apiv2_human_find.py
```diff
@@ -311,3 +323,14 @@ async def test_bad_args(access_testbed, organization_factory, local_device_facto
...
+    await binder.bind_device(richard, certifier=godfrey1)
+
+    rep = await human_find(sock, query="richard")
```
touilleMan, line 335: I guess the best thing to do is to use the user_id as query to ensure the old behavior is no longer in place

touilleMan (commented):
no need to rename the test with `apiv2` given with provide the `apiv1` suffix only to mark them as deprecated (and soon-to-be-removed)

parsec/backend/postgresql/user_queries/find.py
```diff
@@ -84,6 +87,7 @@ async def query_find(
     all_results = await conn.fetch(*args)
     # TODO: should user LIMIT and OFFSET in the SQL query instead
```
touilleMan, line 88: TODO to remove I guess ^^

tests/backend/user/test_apiv1_human_find.py
Sithi5: done

tests/backend/user/test_apiv2_human_find.py
Sithi5: done

tests/backend/user/test_apiv2_human_find.py
Sithi5: done

tests/backend/user/test_apiv2_human_find.py
Sithi5: done

tests/backend/user/test_apiv2_human_find.py
Sithi5: done

tests/backend/user/test_apiv1_human_find.py
Sithi5: Not sure but it should be done with lines below, I assert that users from alice_backend_sock are different from the other organization to test the isolation.

tests/backend/user/test_apiv1_human_find.py
Sithi5: done

parsec/backend/memory/user.py
Sithi5: fixed

tests/backend/user/test_apiv1_human_find.py
Sithi5: I don't think so because the order should be the same as the insert order.

parsec/backend/postgresql/user_queries/find.py
Sithi5: Done

parsec/backend/memory/user.py
Sithi5: Done ! with a test as well
`    # Test sort is before pagination when pagination and test non-sensitive sort
    rep = await human_find(sock, page=1, per_page=1, query="BlaCkY")
    assert rep == {
        "status": "ok",
        "results": [
            {"user_id": blacky.user_id, "human_handle": blacky.human_handle, "revoked": False}
        ],
        "per_page": 1,
        "page": 1,
        "total": 4,
    }`

parsec/backend/postgresql/user_queries/find.py
```diff
@@ -2,140 +2,150 @@
...
-SELECT user_id
+SELECT
+    user_id,
+    count(*) OVER() AS full_count
 FROM user_
 WHERE
     organization = { q_organization_internal_id("$organization_id") }
     { " ".join(conditions) }
 ORDER BY user_id
+LIMIT { limit }
```
touilleMan, line 29: `limit` and `offset` are data provided by the user, even if they are sanitized during deserialization, we shouldn't trust them !

So here we expose ourselves to SQL injection, we must instead pass those variable through parameters.

On top of that, `_q_factory` has a `lru_cache`, so we shouldn't provide parameters that could vary a lot (offset and limit could have any value)

PR #1488, opened 2020-09-29
Display shared workspaces between two users

AntoineDupre:
See issue #1454

### DIFF
--- a/parsec/core/gui/workspaces_widget.py
+++ b/parsec/core/gui/workspaces_widget.py
@@ -80,7 +80,7 @@ async def _do_workspace_rename(core, workspace_id, new_name, button):
         raise JobResultError("rename-error") from exc
 
 
-async def _do_workspace_list(core, filter_user_id=None):
+async def _do_workspace_list(core):
     workspaces = []
 
     async def _add_workspacefs(workspace_fs, timestamped):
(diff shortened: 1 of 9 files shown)


tests/core/gui/test_workspaces.py
```diff
@@ -249,3 +251,74 @@ def _wk_opened():
...
+    assert u_w.layout_users.count() == 3
+    for i in range(u_w.layout_users.count()):
+        button = u_w.layout_users.itemAt(i).widget()
+        if not button.is_current_user and button.user_info.user_id == alice.user_id:
+            button.filter_user_workspaces_clicked.emit(button.user_info)
+            break
+    else:
+        raise ValueError("Can not find Alice user")
+
+    def _wokrspace_filtered():
```
Max-7, line 306: `workspace`

parsec/core/gui/workspaces_widget.py
```diff
@@ -422,6 +439,14 @@ def add_workspace(self, workspace_fs, ws_entry, users_roles, files, timestamped)
                 new_name=workspace_name,
                 button=None,
             )
+
+        if self.filter_user_info is not None:
+
+            for user_id in users_roles.keys():
```
Max-7, line 445:
```python
if self.filter_user_info.user_id not in users_roles:
    return
```

Max-7 (changes requested):

parsec/core/gui/workspaces_widget.py
```diff
@@ -241,6 +243,21 @@ def __init__(self, core, jobs_ctx, event_bus, **kwargs):
...
+    def remove_user_filter(self):
+        self.filter_user_info = None
+        self.shared_button.hide()
+        self.reset()
+
+    def set_user_info(self, user_info):
+        self.filter_user_info = user_info
+        self.shared_button.show()
+        self.shared_button.setText(
+            _("TEXT_WORKSPACE_FILTERED_user").format(user=user_info.human_handle.label)
```
Max-7, line 258: Should use `user_info.short_user_display` here. Some users may not have a `human_handle`.

Max-7 (approved):

PR #1490, opened 2020-09-29
[GUI] Prevent unecessary scrolling for users and devices

### LINKED ISSUE #1449: Users display keeps the scrolling bar when it's unecessary
users_scrolling_bug

Parsec opens in fullscreen, but when resizing the window, the vertical scrollbar should disappear if there are not a lot of users to display. Devices display does not have this problem.

Max-7:
Closes #1449


Sithi5 (approved):
LGTM 👍

PR #1500, opened 2020-10-05
Fix a possible deadlock when cancelling the mounting of a workspace on linux

vxgmichel:


parsec/core/mountpoint/fuse_runner.py
```diff
@@ -167,44 +172,56 @@ def _run_fuse_thread():
...
     while True:
+        # Tick for 10 ms
         await trio.sleep(0.01)
+        # Stat the root of the workspace
         try:
             stat = await trio_mountpoint_path.stat()
+        # The directory does not exist for some reason
         except FileNotFoundError:
-            continue
+            raise
```
AntoineDupre, line 194: What is the point here to catch and raise the exception, explicitness ?

AntoineDupre (approved):
LGTM

parsec/core/mountpoint/fuse_runner.py
vxgmichel: Otherwise it's caught by `except OSError: break` which doesn't really make sense if the directory does not exist

PR #1507, opened 2020-10-12
[GUI] Improved import files error messages

### LINKED ISSUE #1491: Confusing error message when failed to import a file
When importing a file from windows with a permission denied, it could be better to output the windows error in the GUI and the file/folder name in question instead of the generic error message that could be confusing.
image
image

Max-7:
Closes #1491 

- displays a specific error message for PermissionDenied errors
- displays different messages depending on the number of files imported

### DIFF
--- a/parsec/core/gui/files_widget.py
+++ b/parsec/core/gui/files_widget.py
@@ -89,6 +89,7 @@ async def _do_folder_create(workspace_fs, path):
 
 async def _do_import(workspace_fs, files, total_size, progress_signal):
     current_size = 0
+    errors = []
     for src, dst in files:
         try:
             if dst.parent != FsPath("/"):
@@ -107,9 +108,15 @@ async def _do_import(workspace_fs, files, total_size, progress_signal):
                         progress_signal.emit(src.name, current_size + read_size)
             current_size += read_size + 1
             progress_signal.emit(src.name, current_size)
-
         except trio.Cancelled as exc:
-            raise JobResultError("cancelled", last_file=dst) from exc
+            errors.append(exc)
+            raise JobResultError(
+                "cancelled", last_file=dst, file_count=len(files), exceptions=errors
+            ) from exc
+        except PermissionError as exc:
+            errors.append(exc)
+    if errors:
+        raise JobResultError("error", exceptions=errors, file_count=len(files))
 
 
 async def _do_remount_timestamped(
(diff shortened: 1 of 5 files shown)


parsec/core/gui/tr/parsec_fr.po
```diff
@@ -500,10 +500,17 @@ msgstr "Un répertoire du même nom existe déjà."
 msgid "TEXT_FILE_FOLDER_CREATE_ERROR_UNKNOWN"
 msgstr "Impossible de créer le nouveau répertoire."
 
-msgid "TEXT_FILE_IMPORT_ERROR"
+msgid "TEXT_FILE_IMPORT_ONE_PERMISSION_ERROR"
+msgstr ""
```
Sithi5, line 504: Forgot  french msg translation

parsec/core/gui/files_widget.py
```diff
@@ -764,7 +771,35 @@ def _on_import_success(self):
         self.import_job = None
 
     def _on_import_error(self):
+        def _display_import_error(file_count, exceptions=None):
+            print("WTF", exceptions, file_count)
```
Sithi5, line 775: print still usefull ?
If it is for the CLI, maybe changing the text for something like "Import files failed"

parsec/core/gui/files_widget.py
```diff
@@ -775,7 +810,15 @@ def _on_import_error(self):
                 silent=True,
             )
         else:
-            show_error(self, _("TEXT_FILE_IMPORT_ERROR"), exception=self.import_job.exc)
+            print(
```
Sithi5, line 813: Same here, print still usefull?

parsec/core/gui/files_widget.py
```diff
@@ -107,9 +108,15 @@ async def _do_import(workspace_fs, files, total_size, progress_signal):
...
             progress_signal.emit(src.name, current_size)
-
         except trio.Cancelled as exc:
-            raise JobResultError("cancelled", last_file=dst) from exc
+            errors.append(exc)
+            raise JobResultError(
+                "cancelled", last_file=dst, file_count=len(files), exceptions=errors
+            ) from exc
+        except PermissionError as exc:
+            errors.append(exc)
```
Sithi5, line 117: It should be possible to remove the if statement below and raise the exception in the `except PermissionError as exc` statement no ?

tests/core/gui/test_files.py
```diff
@@ -813,3 +814,72 @@ def _import_failed():
...
+
+    # Changing file permissions
+    os.chmod(temp_dir / "file01.txt", 000)
+    monkeypatch.setattr(
+        "PyQt5.QtWidgets.QFileDialog.getOpenFileNames",
+        classmethod(lambda *args, **kwargs: ([temp_dir / "file01.txt"], True)),
+    )
+
+    async with aqtbot.wait_signals(
+        [w_f.button_import_files.clicked, w_f.import_error], timeout=3000
```
Sithi5, line 839: CI has failed, maybe the timeout is too short ?

parsec/core/gui/files_widget.py
Sithi5: actually got both print when import fail

parsec/core/gui/tr/parsec_en.po
```diff
@@ -2376,3 +2383,12 @@ msgstr "Create a new workspace"
 msgid "ACTION_WORKSPACE_GOTO_FILE_LINK"
 msgstr "Go to a file using a link"
 
+#~ msgid "TEXT_FILE_IMPORT_PERMISSION_ERROR"
```
Sithi5, line 2386: Can be removed

parsec/core/gui/tr/parsec_fr.po
```diff
@@ -2438,3 +2445,12 @@ msgstr "Créer un nouvel espace de travail"
 msgid "ACTION_WORKSPACE_GOTO_FILE_LINK"
 msgstr "Trouver un fichier via son lien"
 
+#~ msgid "TEXT_FILE_IMPORT_PERMISSION_ERROR"
```
Sithi5, line 2448: Can be removed

Sithi5 (changes requested):

parsec/core/gui/files_widget.py
Max-7: This one has bothered me for some time. Since we can be importing more than one file, we have two options:
- we stop at the first error and not even try to import some files
- we try to import as many files as possible

I chose the later even though I'm not convinced that it's the best. I'll open an issue on the subject so we can discuss it.

Sithi5 (approved):
LGTM 👍

PR #1528, opened 2020-10-15
[GUI] Fix file open failure

### LINKED ISSUE #1525: No error message in GUI When failed to open a file
In this case the file is not opened because no sofware are registered to open this type of file.
image

Max-7:
Closes #1525


Sithi5 (approved):
LGTM ! 👍

PR #1535, opened 2020-10-23
Fix database corruption and add tests for disk full errors on sqlite operations

vxgmichel:
Fix issue #1537 and add more tests to detect database corruption after a failed write operation in a local sqlite database.

### DIFF
--- a/.azure-pipelines.yml
+++ b/.azure-pipelines.yml
@@ -153,6 +153,14 @@ jobs:
         -n auto --max-worker-restart=0 -x \
         --junitxml=test-results-postgresql.xml
     displayName: 'Tests postgresql'
+  # The full disk test are run separately since they rely on running a sqlite3 database
+  # on a fuse mountpoint which can deadlock if run with the rest of the test suite
+  - bash: |
+      set -eux
+      py.test $(pytest.base_args) \
+        tests --rundiskfull --runslow -m diskfull -x \
+        --junitxml=test-results-fulldisk.xml
+    displayName: 'Tests full disk'
   # TODO: run gui tests with xvfb
   # - bash: |
   #     set -eux
--- a/parsec/core/fs/storage/local_database.py
+++ b/parsec/core/fs/storage/local_database.py
@@ -118,6 +118,12 @@ class LocalDatabase:
         # Lock the access to the connection object
         async with self._lock:
 
+            # The connection might not be set
+            try:
+                self._conn
+            except AttributeError:
+                return
+
             # Commit and close
             await self._run_in_thread(self._conn.commit)
             self._conn.close()
--- a/parsec/core/fs/storage/local_database.py
+++ b/parsec/core/fs/storage/local_database.py
@@ -93,9 +93,7 @@ class LocalDatabase:
     # Operational error protection
 
     @contextmanager
-    def _manage_operational_error(
-        self, conn: Optional[Connection] = None, allow_commit: bool = False
-    ) -> Iterator[None]:
+    def _manage_operational_error(self, allow_commit: bool = False) -> Iterator[None]:
         """Close the local database when an operational error is detected
 
         Operational errors have to be treated with care since they usually indicate
@@ -110,11 +108,7 @@ class LocalDatabase:
         This way, we force the core to create new workspace storage objects and therefore
         discarding any uncommited data.
         """
-        # Get connection
-        if conn is None:
-            conn = self._conn
-        in_transaction_before = conn.in_transaction
-
+        in_transaction_before = self._conn.in_transaction
         # Safe context for operational errors
         try:
             try:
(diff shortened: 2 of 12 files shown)


parsec/core/fs/storage/local_database.py
```diff
@@ -118,9 +178,24 @@ async def _close(self) -> None:
...
+            del self._conn
+
+    async def _commit(self) -> None:
+        # Close the local database if an operational error is detected
+        with self._manage_operational_error(allow_commit=True):
+            await self._run_in_thread(self._conn.commit)
+
+    def _check_state(self) -> None:
+        try:
+            self._conn
```
touilleMan, line 196:
```python
if not hasattr(self, "_conn"):
```
?

parsec/core/fs/storage/local_database.py
```diff
@@ -129,52 +204,63 @@ async def open_cursor(self, commit: bool = True) -> AsyncIterator[Cursor]:
...
         # Lock the access to the connection object
         async with self._lock:
 
-            # Flush to disk
-            await self._run_in_thread(self._conn.commit)
+            # Check connection state
+            self._check_state()
+
+            # Vacuum disabled
+            if self.vacuum_threshold is None:
```
touilleMan, line 256: Why the vacuum_threshold check here ?

parsec/core/fs/workspacefs/workspacefile.py
```diff
@@ -106,8 +110,27 @@ def _check_open_state(self) -> None:
...
     async def close(self) -> None:
         """Close the file"""
-        if self._state != FileState.CLOSED:
-            await self._transactions.fd_close(self.fileno())
+        # Idempotency
+        if self._state == FileState.CLOSED:
+            return
+        # Make sure the state is set to CLOSED
+        try:
+            # Make sure the fd is closed is called even if the flushing fails
```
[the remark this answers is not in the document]
touilleMan: btw it could be good to explain here the implication of a failed flush (typically pointing to the `_manage_operational_error`'s comment)

touilleMan (approved):

parsec/core/fs/workspacefs/file_transactions.py
```diff
@@ -249,7 +249,7 @@ async def fd_resize(self, fd: FileDescriptor, length: int, truncate_only: bool =
                 return
 
             # Perform the resize operation
-            await self._manifest_resize(manifest, length)
+            await self._manifest_resize(manifest, length, cache_only=True)
```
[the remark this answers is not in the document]
vxgmichel: Actually no, this is just to be consistent with the `fd_write` method which is also cache-only. This way, using `write_bytes` typically flushes the cache only once, when closing the file (instead of flushing once when truncating to zero then once more when closing).

parsec/core/fs/storage/local_database.py
```diff
@@ -82,13 +85,69 @@ async def run(
...
+            # Raise an FSLocalStorageOperationalError
+            finally:
+                raise FSLocalStorageOperationalError from exception
 
     # Life cycle
 
     async def _create_connection(self) -> Connection:
         # Create directories
-        self.path.parent.mkdir(mode=0o700, parents=True, exist_ok=True)
+        await self.path.parent.mkdir(mode=0o700, parents=True, exist_ok=True)
```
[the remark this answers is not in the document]
vxgmichel: I don't think so, it doesn't have much to do with sqlite3 operational errors.

parsec/core/fs/storage/local_database.py
```diff
@@ -82,13 +85,69 @@ async def run(
...
+        in_transaction_before = conn.in_transaction
+
+        # Safe context for operational errors
+        try:
+            try:
+                yield
+
+            # Extra checks for end of transaction
+            finally:
+                end_of_transaction_detected = in_transaction_before and not conn.in_transaction
```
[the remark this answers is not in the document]
vxgmichel: Yes, typically this case:
```python
with local_database.open_cursor() as cursor:
    try:
        cursor.execute("something")
    finally:
        raise Something # For some reason
```
In this case, the `sqlite3.OperationalError` might get shadowed by the exception raised in the finally clause.

Still we can detect a potential rollback if the conn used to be in a transaction when the context was entered but is not when it's left.

parsec/core/fs/storage/local_database.py
```diff
@@ -82,13 +85,69 @@ async def run(
...
+                if not allow_commit and end_of_transaction_detected:
+                    raise OperationalError("A forbidden commit/rollback has been detected")
+
+        # An operational error has been detected
+        except OperationalError as exception:
+
+            # Mark the local database as closed
+            try:
+                del self._conn
```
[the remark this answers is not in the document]
vxgmichel: Right, I'll change that.

parsec/core/fs/storage/local_database.py
vxgmichel: Sure, why not.

parsec/core/fs/storage/local_database.py
vxgmichel: I thought it would make more sense to always raise an `FSLocalStorageClosedError` if the local database is closed, even if vacuum is disabled. But I don't have a strong opinion about it.

.azure-pipelines.yml
```diff
@@ -153,6 +153,14 @@ jobs:
         -n auto --max-worker-restart=0 -x \
         --junitxml=test-results-postgresql.xml
     displayName: 'Tests postgresql'
+  # The full disk test are run separately since they rely on running a sqlite3 database
+  # on a fuse mountpoint which can deadlock if run with the rest of the test suite
+  - bash: |
+      set -eux
+      py.test $(pytest.base_args) \
+        tests --rundiskfull --runslow -m diskfull -x \
+        --junitxml=test-results-memory.xml
```
touilleMan, line 162: same junitxml file name than another test ? won't this erase the previous results ?

parsec/core/fs/storage/local_database.py
touilleMan: 👍

touilleMan (approved):

.azure-pipelines.yml
vxgmichel: Ooopsie, fixed!  82485f0

PR #1563, opened 2020-11-05
Adding a test for file history

Sithi5:


tests/core/fs/workspacefs/test_file_history.py
```diff
@@ -0,0 +1,65 @@
...
+
+
+@pytest.mark.trio
+async def test_file_history(alice_workspace):
+    sync_by_id = alice_workspace.sync_by_id
+    entry = alice_workspace.get_workspace_entry()
+    wid = entry.id
+
+    # Empty workspace
+    await sync_by_id(wid)
```
touilleMan, line 16: I don't thing this is needed: alice_workspace fixture should ensure the user & workspaces manifests has been synced

However it seems it's not the case:
https://github.com/Scille/parsec-cloud/blob/c73f6a84755f894baa2ffa668dfe70dd84d18cfe/tests/core/fs/workspacefs/conftest.py#L9-L14

So I think the fixture (and also bob_workspace fixture) should be modified to also do a `alice_user_fs.sync()` before returning

tests/core/fs/workspacefs/test_file_history.py
```diff
@@ -0,0 +1,65 @@
...
+
+    # Creating files and writing first byte
+    await alice_workspace.touch("/f")
```
touilleMan, line 19: "writing first byte" doesn't seem true given touch only create the file

tests/core/fs/workspacefs/test_file_history.py
```diff
@@ -0,0 +1,65 @@
...
+
+    # Test version before first sync
+    f_versions = await VersionLister(alice_workspace).list(FsPath("/f"))
```
touilleMan, line 22: more readable to unpack the result here:
```python
versions_list, download_limit_reached = await VersionLister(alice_workspace).list(FsPath("/f"))
```

tests/core/fs/workspacefs/test_file_history.py
```diff
@@ -0,0 +1,65 @@
...
+
+    # Assert if task list have been completed
+    assert f_versions[1] is True
+
+    # Should have no version yet because we didn't sync
+    assert not f_versions[0]
+
+    # Syncronysing the files
```
touilleMan, line 30: No need to comment every line ;-)

I think you can just go:
```python
# Test versioning before first sync (hence should be empty)
versions_list, download_limit_reached = ...
assert download_limit_reached
assert not versions_list
```

tests/core/fs/workspacefs/test_file_history.py
```diff
@@ -0,0 +1,65 @@
...
+    # Empty workspace
+    await sync_by_id(wid)
+
+    # Creating files and writing first byte
+    await alice_workspace.touch("/f")
+
+    # Test version before first sync
+    f_versions = await VersionLister(alice_workspace).list(FsPath("/f"))
+
+    # Assert if task list have been completed
```
touilleMan, line 24: I don't thing this is what the second returned value is for:
> Returns:
>             A tuple containing a list of TimestampBoundedData and a bool indicating wether the
>            download limit has been reached

tests/core/fs/workspacefs/test_file_history.py
```diff
@@ -0,0 +1,65 @@
...
+
+    # Checking again, should have the first version now
+    f_versions = await VersionLister(alice_workspace).list(FsPath("/f"))
+
+    # Assert if task list have been completed
+    assert f_versions[1] is True
+    # Assert there is a version in the list
+    assert f_versions[0]
+    # Checking the first version of the list
+    assert getattr(f_versions[0][0], "version") == 1
```
touilleMan, line 41: Hard to read statement, it would be much simpler to write:
```python
assert versions_list[0].version == 1
```
or even:
```python
assert versions_list == [
    TimestampBoundedData(
        id=ANY,
        version=1,
        early=ANY,
        late=ANY,
        creator=alice.device_id,
        updated=ANY,
        is_folder=False,
        size=0,
        source=ANY,  # not sure what to put here...
        destination=ANY,  # not sure what to put here...
    )
]
```
This has the added benefit of checking the number of items in `versions_list`
The check against a `TimestampBoundedData` is a bit cumbersome, so it really depend on many field inside it we want to test and how complex to create this object is...

tests/core/fs/workspacefs/test_file_history.py
```diff
@@ -0,0 +1,65 @@
...
+    version_nb = 2
+    previous_late = None
+    for version in f_versions[0]:
+        assert version.version == version_nb
+        version_nb += 1
+        # Checking DateTime is coherent
+        if previous_late:
+            assert previous_late == version.early
+        previous_late = version.late
+    # File should have 21 versions (20 modification + 1 creation). version_nb - 1 because it is
```
touilleMan, line 63: Good comment ^^

touilleMan (changes requested):

touilleMan (approved):

PR #1564, opened 2020-11-06
[GUI] Made cut and copy asynchronous

### LINKED ISSUE #1560: Move copy/paste to the job system
It's currently using the job.run() function which freezes the GUI. It should instead start a job, and potentially display a loading dialog while the copy is being made.

Max-7:
Closes #1560


newsfragments/1560.bugfix.rst
```diff
@@ -0,0 +1 @@
+Made copy and cut of files asynchronous
```
touilleMan, line 1: `... in GUI` ?

parsec/core/gui/files_widget.py
```diff
@@ -70,6 +70,101 @@ async def _do_delete(workspace_fs, files, silent=False):
                 raise JobResultError("error", multi=len(files) > 1) from exc
 
 
+async def _do_copy_files(workspace_fs, current_directory, files, source_workspace):
+    last_exc = None
+    error_count = 0
+    for f in files:
```
touilleMan, line 76: `for src, src_type in files` ?

parsec/core/gui/files_widget.py
```diff
@@ -70,6 +70,101 @@ async def _do_delete(workspace_fs, files, silent=False):
...
+        src = f[0]
+        src_type = f[1]
+        file_name = src.name
+        base_name = pathlib.Path(src.name)
```
touilleMan, line 80: `base_name` is confusing given it seems to refer to https://docs.python.org/3.8/library/os.path.html#os.path.basename, why not `stem_name` ?

Besides I'm not sure why the conversion to `pathlib.Path` is needed (it gives the impression we are working on the entire path where only the file name is important here)
Last but not least, it seams `pathlib.Path.stem` is fairly simple:
https://github.com/python/cpython/blob/e74fb2d7666eea43ad738528a565bb56bc88c28d/Lib/pathlib.py#L855-L863

So I think we should just make this ourselves:
```python
file_name_with_extensions, *_ = src.name.split(".", 1)
```

parsec/core/gui/files_widget.py
```diff
@@ -70,6 +70,101 @@ async def _do_delete(workspace_fs, files, silent=False):
...
+        base_name = str(base_name)
+        while True:
+            try:
+                dst = current_directory / file_name
+                await workspace_fs.move(src, dst, source_workspace)
+                break
+            except FileExistsError:
+                # File already exists, we append a counter at the end of its name
+                file_name = "{} ({}){}".format(
+                    base_name, count, "".join(pathlib.Path(src.name).suffixes)
```
touilleMan, line 145: base_name is only used here, hence would be better to define it there too

touilleMan (changes requested):

parsec/core/gui/files_widget.py
Max-7: Not sure what you mean. It seems to me that it's better to get the file name without extensions before the loop instead of doing the same split over and over again inside the loop.

touilleMan: Given the while loop should do very few iteration (most of the time no iterations at all !) I don' think there is any performance issue.
On the other hand, this splitting logic put before the loop decrease the readability for the main case.

However now that the splitting logic is a single trivial one-liner, I think it's less useful than when the comment was added. So do as you feel ;-)

touilleMan (approved):
👍

PR #1571, opened 2020-11-10
Add event reason for entry updated

Sithi5:


parsec/core/core_events.py
```diff
@@ -34,3 +34,16 @@ class CoreEvent(Enum):
     # Others
     SHARING_UPDATED = "sharing.updated"
     USERFS_UPDATED = "userfs.updated"
+
+
+class FSEntryUpdatedReason(IntEnum):
```
touilleMan, line 39: Why `IntEnum` and not `Enum` ? Given we don't do serialization on this type, `Enum` seems good enough to me

Besides instead of using `auto` (which give an int as a value), we should use a string a value (to be consistent with wath is done for all the other enums in the codebase):
```
class FSEntryUpdatedReason(Enum):
    ENTRY_RENAME = "ENTRY_RENAME"
...
```

parsec/core/gui/files_widget.py
```diff
@@ -934,7 +934,7 @@ def _display_import_error(file_count, exceptions=None):
     def _on_fs_entry_synced_trio(self, event, id, workspace_id=None):
         self.fs_synced_qt.emit(event, id)
 
-    def _on_fs_entry_updated_trio(self, event, workspace_id=None, id=None):
+    def _on_fs_entry_updated_trio(self, event, workspace_id=None, id=None, **kwarg):
```
touilleMan, line 937: `reason` is a mandatory param for the event, so better to explicitly provide it here:
```python
def _on_fs_entry_updated_trio(self, event, reason, workspace_id=None, id=None):
...
```

parsec/core/fs/workspacefs/entry_transactions.py
```diff
@@ -261,7 +261,9 @@ async def entry_rename(
             await self.local_storage.set_manifest(parent.id, new_parent)
 
         # Send event
-        self._send_event(CoreEvent.FS_ENTRY_UPDATED, id=parent.id)
+        self._send_event(
+            CoreEvent.FS_ENTRY_UPDATED, id=parent.id, reason=FSEntryUpdatedReason.ENTRY_RENAME
```
touilleMan, line 265: I think we need to provide more informations about what is renamed here (typically `entry_id`, `entry_source_name` and `entry_destination_name`)

parsec/core/fs/workspacefs/entry_transactions.py
```diff
@@ -295,7 +297,9 @@ async def folder_delete(self, path: FsPath) -> EntryID:
             await self.local_storage.set_manifest(parent.id, new_parent)
 
         # Send event
-        self._send_event(CoreEvent.FS_ENTRY_UPDATED, id=parent.id)
+        self._send_event(
+            CoreEvent.FS_ENTRY_UPDATED, id=parent.id, reason=FSEntryUpdatedReason.FOLDER_DELETE
```
touilleMan, line 301: same as above, missing `entry_id` and `entry_name`

parsec/core/fs/workspacefs/entry_transactions.py
```diff
@@ -355,8 +361,10 @@ async def folder_create(self, path: FsPath) -> EntryID:
             await self.local_storage.set_manifest(parent.id, new_parent)
 
         # Send events
-        self._send_event(CoreEvent.FS_ENTRY_UPDATED, id=parent.id)
-        self._send_event(CoreEvent.FS_ENTRY_UPDATED, id=child.id)
+        self._send_event(CoreEvent.FS_ENTRY_UPDATED, id=parent.id, reason=None)
+        self._send_event(
+            CoreEvent.FS_ENTRY_UPDATED, id=child.id, reason=FSEntryUpdatedReason.FOLDER_CREATE
```
touilleMan, line 366: Here we can use another reason like `FSEntryUpdatedReason.FOLDER_CREATE_ENTRY_CREATION` (and it would be good to have a comment explaining the difference between `FOLDER_CREATE` and `FOLDER_CREATE_ENTRY_CREATION` in `FSEntryUpdatedReason` definition)

parsec/core/fs/workspacefs/entry_transactions.py
```diff
@@ -355,8 +361,10 @@ async def folder_create(self, path: FsPath) -> EntryID:
             await self.local_storage.set_manifest(parent.id, new_parent)
 
         # Send events
-        self._send_event(CoreEvent.FS_ENTRY_UPDATED, id=parent.id)
-        self._send_event(CoreEvent.FS_ENTRY_UPDATED, id=child.id)
+        self._send_event(CoreEvent.FS_ENTRY_UPDATED, id=parent.id, reason=None)
```
touilleMan, line 364: no reason is a bad reason !
We'd better have `reason=FSEntryUpdatedReason.FOLDER_CREATE` here (to be coherent with the fact `reason=FSEntryUpdatedReason.FOLDER_DELETE` is triggered for the parent)

Of cours this event should provide `entry_id` and `entry_name`

parsec/core/fs/workspacefs/entry_transactions.py
```diff
@@ -389,8 +397,10 @@ async def file_create(
             fd = self.local_storage.create_file_descriptor(child) if open else None
 
         # Send events
-        self._send_event(CoreEvent.FS_ENTRY_UPDATED, id=parent.id)
-        self._send_event(CoreEvent.FS_ENTRY_UPDATED, id=child.id)
+        self._send_event(CoreEvent.FS_ENTRY_UPDATED, id=parent.id, reason=None)
```
touilleMan, line 400: same as above

parsec/core/fs/workspacefs/sync_transactions.py
```diff
@@ -415,8 +415,16 @@ async def file_conflict(
                 await self.local_storage.set_manifest(parent_id, new_parent_manifest)
                 await self.local_storage.set_manifest(entry_id, other_manifest)
 
-                self._send_event(CoreEvent.FS_ENTRY_UPDATED, id=new_manifest.id)
-                self._send_event(CoreEvent.FS_ENTRY_UPDATED, id=parent_id)
+                self._send_event(
+                    CoreEvent.FS_ENTRY_UPDATED,
```
touilleMan, line 419: same as above: we shouldn't use the same reason for the conflicting entry and it parent (otherwise it would become harder to determine what to do with this event from a listener point of view)

So we should have `FSEntryUpdatedReason.SYNC_FILE_CONFLICT_ENTRY_CREATION` and `FSEntryUpdatedReason.SYNC_FILE_CONFLICT` (which would expose `entry_id` and `entry_name`)

touilleMan (changes requested):
I surprised there is almost no changes in the tests... I would expect we check for those events and there corresponding parameters in some dedicated tests...

parsec/core/core_events.py
```diff
@@ -34,3 +34,26 @@ class CoreEvent(Enum):
...
     SHARING_UPDATED = "sharing.updated"
     USERFS_UPDATED = "userfs.updated"
+
+
+class FSEntryUpdatedReason(Enum):
+    ENTRY_RENAME = "ENTRY_RENAME"
+    FOLDER_DELETE = "FOLDER_DELETE"
+    # Creation of the parent manifest (v1) when creating folder
+    FOLDER_CREATE = "FOLDER_CREATE"
+    # Creation of the child (v2)
```
touilleMan, line 44: not sure what the v1/v2 means...

parsec/core/fs/userfs/userfs.py
```diff
@@ -388,7 +394,14 @@ async def workspace_rename(self, workspace_id: EntryID, new_name: AnyEntryName)
                 updated_workspace_entry
             )
             await self.set_user_manifest(updated_user_manifest)
-            self.event_bus.send(CoreEvent.FS_ENTRY_UPDATED, id=self.user_manifest_id)
+            self.event_bus.send(
+                CoreEvent.FS_ENTRY_UPDATED,
+                id=self.user_manifest_id,
+                reason=FSEntryUpdatedReason.WORKSPACE_RENAME,
+                entry_id=updated_workspace_entry.id,
+                entry_source_name=EntryName(workspace_entry.name),
```
touilleMan, line 402: `workspace_entry.name` is already an `EntryName`

I guess you got bitten by this incorrect typing:
https://github.com/Scille/parsec-cloud/blob/641558f9938fd27fc5008c3502c4e78915b01a1e/parsec/api/data/manifest.py#L78

Here we should have `name: EntryName` (according to what is defined in the `WorkspaceEntry.SCHEMA_CLS`).

So can you do fix `WorkspaceEntry.name` typing ;-)

touilleMan: BTW I'm not sure about the `entry_source_name` here: unlike for FolderManifest we are not actually moving a workspace around but are just changing it name.
This is a subtle difference but the main idea is you can have multiple workspace with the same name (which is of course impossible within a folder !)

So I think we should use the `entry_old_name`/`entry_new_name` to mimic the name of the parameter of this function and to mark the difference with folder.

parsec/core/fs/workspacefs/entry_transactions.py
```diff
@@ -261,7 +261,14 @@ async def entry_rename(
             await self.local_storage.set_manifest(parent.id, new_parent)
 
         # Send event
-        self._send_event(CoreEvent.FS_ENTRY_UPDATED, id=parent.id)
+        self._send_event(
+            CoreEvent.FS_ENTRY_UPDATED,
+            id=parent.id,
+            reason=FSEntryUpdatedReason.ENTRY_RENAME,
+            entry_id=parent.children[source.name],
```
touilleMan, line 268: you can use `entry_id=source_entry_id` here

parsec/core/fs/userfs/userfs.py
```diff
@@ -366,7 +366,13 @@ async def workspace_create(self, name: AnyEntryName) -> EntryID:
             user_manifest = user_manifest.evolve_workspaces_and_mark_updated(workspace_entry)
             await self._create_workspace(workspace_entry.id, workspace_manifest)
             await self.set_user_manifest(user_manifest)
-            self.event_bus.send(CoreEvent.FS_ENTRY_UPDATED, id=self.user_manifest_id)
+            self.event_bus.send(
+                CoreEvent.FS_ENTRY_UPDATED,
+                id=self.user_manifest_id,
+                reason=FSEntryUpdatedReason.WORKSPACE_CREATE,
+                entry_id=workspace_entry.id,
```
touilleMan, line 373: should call this `workspace_id`/`workspace_name` (given `entry` is a more generic term that we'd better use in the case we deal with file/folder entry)

parsec/core/fs/userfs/userfs.py
touilleMan: aaaand same as above: we should call them `workspace_id`/`workspace_xxx_name`

tests/core/test_events_reason.py
```diff
@@ -0,0 +1,156 @@
+# Parsec Cloud (https://parsec.cloud) Copyright (c) AGPLv3 2019 Scille SAS
+import pytest
+
+from unittest.mock import ANY
+
+from parsec.core.core_events import CoreEvent, FSEntryUpdatedReason
+
+
+@pytest.mark.trio
+async def test_fs_update_event_reason(alice_user_fs, running_backend):
```
touilleMan, line 10: super cool test 👍 🥇 😄

parsec/core/fs/workspacefs/sync_transactions.py
```diff
@@ -415,8 +415,18 @@ async def file_conflict(
...
+                    id=parent_id,
```
touilleMan, line 420: Is there a reason why you changed the order of the events ?

parsec/core/fs/workspacefs/sync_transactions.py
```diff
@@ -415,8 +415,18 @@ async def file_conflict(
...
+                    reason=FSEntryUpdatedReason.SYNC_FILE_CONFLICT,
+                )
```
touilleMan, line 422: I would add `entry_id`, `entry_name`, `entry_conflict_id` and `entry_conflict_name` to this event

parsec/core/fs/workspacefs/entry_transactions.py
```diff
@@ -355,8 +374,16 @@ async def folder_create(self, path: FsPath) -> EntryID:
             await self.local_storage.set_manifest(parent.id, new_parent)
 
         # Send events
-        self._send_event(CoreEvent.FS_ENTRY_UPDATED, id=parent.id)
-        self._send_event(CoreEvent.FS_ENTRY_UPDATED, id=child.id)
+        self._send_event(
+            CoreEvent.FS_ENTRY_UPDATED, id=parent.id, reason=FSEntryUpdatedReason.FOLDER_CREATE
+        )
```
touilleMan, line 379: I would add the `entry_id`/`entry_name` fields in this event instead of `FOLDER_CREATE_ENTRY_CREATION`.
The rational is this event is the "main" one (the one we are actually listening for if we are interested in changes in a given directory), the `FOLDER_CREATE_ENTRY_CREATION` is only interesting if we are listening for any changes (like what does the sync monitor)

parsec/core/fs/workspacefs/entry_transactions.py
```diff
@@ -355,8 +374,16 @@ async def folder_create(self, path: FsPath) -> EntryID:
...
+        self._send_event(
+            CoreEvent.FS_ENTRY_UPDATED,
+            id=child.id,
+            reason=FSEntryUpdatedReason.FOLDER_CREATE_ENTRY_CREATION,
+            entry_id=child.id,
```
touilleMan, line 384: In fact I wouldn't even put the `entry_id`/`entry_name` fields in this event:
- `entry_id` is just the same as `id` so no need to keep it
- `entry_name` is meaningful only if we consider the parent directory (given a manifest doesn't store the file name)

parsec/core/fs/workspacefs/entry_transactions.py
```diff
@@ -389,8 +416,16 @@ async def file_create(
             fd = self.local_storage.create_file_descriptor(child) if open else None
 
         # Send events
-        self._send_event(CoreEvent.FS_ENTRY_UPDATED, id=parent.id)
-        self._send_event(CoreEvent.FS_ENTRY_UPDATED, id=child.id)
+        self._send_event(
+            CoreEvent.FS_ENTRY_UPDATED, id=parent.id, reason=FSEntryUpdatedReason.FILE_CREATE
+        )
```
touilleMan, line 421: same as above

parsec/core/fs/workspacefs/entry_transactions.py
```diff
@@ -425,6 +460,12 @@ async def file_resize(self, path: FsPath, length: int) -> EntryID:
 
             # Perform resize
             await self._manifest_resize(manifest, length)
-
+            self._send_event(
+                CoreEvent.FS_ENTRY_UPDATED,
+                id=manifest.id,
+                reason=FSEntryUpdatedReason.FILE_RESIZE,
+                entry_id=manifest.id,
+                entry_name=path.name,
```
touilleMan, line 468: `FSEntryUpdatedReason.FILE_RESIZE` is already triggered in `file_transaction.py` but without the `entry_name` field.
We'd better just remove `entry_id`/`entry_name` here to avoid such inconsistency ;-)

and as a matter of fact it seems you fixed a bug here given the event was missing here \o/

tests/core/test_events_reason.py
touilleMan: Can you add a test for the file access with a file descriptor too ? (`fd_resize` and `fd_write`)

parsec/core/fs/workspacefs/sync_transactions.py
```diff
@@ -415,8 +415,18 @@ async def file_conflict(
...
+                self._send_event(
+                    CoreEvent.FS_ENTRY_UPDATED,
+                    id=new_manifest.id,
+                    reason=FSEntryUpdatedReason.SYNC_FILE_CONFLICT_ENTRY_CREATION,
+                    entry_id=entry_id,
```
touilleMan, line 427: Same as above, I would remove the `entry_id` and `entry_name` entirely

touilleMan (changes requested):

parsec/core/fs/workspacefs/sync_transactions.py
Sithi5: It was just to have the same order than in workspace/files creations, the ENTRY event in second.

parsec/core/fs/workspacefs/sync_transactions.py
Sithi5: I am not sure about the `entry_conflict_id`, is it not the same as `parent_id`?

tests/core/test_events_reason.py
Sithi5: I also add some test for workspace creation etc

parsec/core/fs/userfs/userfs.py
```diff
@@ -366,7 +366,13 @@ async def workspace_create(self, name: AnyEntryName) -> EntryID:
             user_manifest = user_manifest.evolve_workspaces_and_mark_updated(workspace_entry)
             await self._create_workspace(workspace_entry.id, workspace_manifest)
             await self.set_user_manifest(user_manifest)
-            self.event_bus.send(CoreEvent.FS_ENTRY_UPDATED, id=self.user_manifest_id)
+            self.event_bus.send(
+                CoreEvent.FS_ENTRY_UPDATED,
+                id=self.user_manifest_id,
+                reason=FSEntryUpdatedReason.WORKSPACE_CREATE,
+                workspace_entry_id=workspace_entry.id,
```
Sithi5, line 373: `workspace_entry_id` instead of `workspace_id` because of conflict in `workspace_widget`

parsec/core/core_events.py
```diff
@@ -34,3 +34,26 @@ class CoreEvent(Enum):
     # Others
     SHARING_UPDATED = "sharing.updated"
     USERFS_UPDATED = "userfs.updated"
+
+
+class FSEntryUpdatedReason(Enum):
+    ENTRY_RENAME = "ENTRY_RENAME"
+    FOLDER_DELETE = "FOLDER_DELETE"
+    # Creation of the parent manifest when creating folder, this is the main event
```
Sithi5, line 42: Not so sure how to correctly explain the use of the two events there

parsec/core/fs/userfs/userfs.py
```diff
@@ -193,7 +193,7 @@ def __init__(
 
         now = pendulum_now()
         wentry = WorkspaceEntry(
-            name="<user manifest>",
+            name=EntryName("<user manifest>"),
```
Sithi5, line 196: New typing in `WorkspaceEntry`

parsec/core/fs/userfs/userfs.py
```diff
@@ -820,7 +837,7 @@ async def _process_message_sharing_granted(
         # Finally insert the new workspace entry into our user manifest
         workspace_entry = WorkspaceEntry(
             # Name are not required to be unique across workspaces, so no check to do here
-            name=msg.name,
+            name=EntryName(msg.name),
```
Sithi5, line 840: Same here, new typing in `WorkspaceEntry`

parsec/api/data/manifest.py
```diff
@@ -75,7 +75,7 @@ class SCHEMA_CLS(BaseSchema):
         def make_obj(self, data: Dict[str, Any]) -> "WorkspaceEntry":
             return WorkspaceEntry(**data)
 
-    name: str
+    name: EntryName
```
Sithi5, line 78: Not so sure about the fix of this typing because in the code we use this class with `name` as str in input.

newsfragments/1571.api.rst
```diff
@@ -0,0 +1 @@
+Fix message name typing, setting it to EntryName to comfort with WorkspaceEntry.
```
Max-7, line 1: It's not what the PR does. The PR adds the reason for an entry update.

parsec/core/fs/workspacefs/entry_transactions.py
```diff
@@ -389,8 +418,18 @@ async def file_create(
             fd = self.local_storage.create_file_descriptor(child) if open else None
 
         # Send events
-        self._send_event(CoreEvent.FS_ENTRY_UPDATED, id=parent.id)
-        self._send_event(CoreEvent.FS_ENTRY_UPDATED, id=child.id)
+        self._send_event(
```
Max-7, line 421: Could we have additional information? To add a file the GUI also needs its creation date, last update date, size, and if it's confined or not.

newsfragments/1571.api.rst
touilleMan: The PR does this, but not just that ;-)
@Sithi5 you should create 2 newsfragments (`1571.api.rst` and `1571.misc.rst`)

parsec/api/data/message.py
```diff
@@ -46,7 +47,7 @@ def get_obj_type(self, obj: Dict[str, object]) -> object:
 class SharingGrantedMessageContent(BaseMessageContent):
     class SCHEMA_CLS(BaseSignedDataSchema):
         type = fields.EnumCheckedConstant(MessageContentType.SHARING_GRANTED, required=True)
-        name = fields.String(required=True)
+        name = EntryNameField(validate=validate.Length(min=1, max=256), required=True)
```
touilleMan, line 50: Length is already checked by `EntryName`

btw it check the size of the utf8-encoded string given it correspond to the actual size of stored data, here you check the size of the string which can be different if the string contains non-ascii characters:
```python
>>> len("foo".encode("utf8"))
3
>>> len("ééé".encode("utf8"))
6
>>> "ééé".encode("utf8")
b'\xc3\xa9\xc3\xa9\xc3\xa9'
>>> len("🐍🐍🐍".encode("utf8"))
12
>>> "🐍🐍🐍".encode("utf8")
b'\xef\xbf\xbd\xef\xbf\xbd\xef\xbf\xbd'
```

touilleMan: I guess the trouble comes from `WorkspaceEntry`, `FolderManifest` and `WorkspaceManifest`, can you fix them ?

newsfragments/1571.misc.rst
```diff
@@ -0,0 +1 @@
+Add event reasons when an entry is updated for GUI.
```
touilleMan, line 1: The newsfragment will end up in the changelog which is supposed to be read by users. So would be better to avoid internal details.

So I would rewrite ` newsfragments/1571.api.rst`:
>  Fix message name typing, setting it to EntryName to comfort with WorkspaceEntry. 
as:
> Fix incorrect definitions of entry name type for workspace and folder manifests in api

I consider advanced user are aware of workspace/folder manifests (see https://docs.parsec.cloud/en/latest/architecture.html#parsec-data-model), and the fact the API has been modified is important to know given it can create incompatibilities between versions

Considering `newsfragments/1571.misc.rst`, I realize it only concern internal changes for the moment so I think we should just remove this file. It will be the job of the PR using those changes to improve the GUI to provide a newsfragement.

parsec/core/fs/workspacefs/entry_transactions.py
touilleMan: entry_size is really needed ? I would say it's always 0 here given the file has just been created !

BTW I'm not really sure it's a good idea to go this way and add so much information in the event, typically here it means we should add `entry_created`&`entry_updated` to all the `FS_ENTRY_UPDATED` events and `entry_size` to the `FILE_WRITE` and `FILE_RESIZE` flavor (and it's pretty easy to forget to update all the events when a new field is needed...)

I think it's much cleaner to consider the event is just a trigger that notifies the GUI it should in turn go fetch the information it needs (so doing a `workspace_fs.path_info(current_folder / event.entry_name)`)

parsec/core/fs/workspacefs/sync_transactions.py
touilleMan: 👍

touilleMan (changes requested):

touilleMan:
Closed in favor of #1582 that provide a cleaner approach

PR #1574, opened 2020-11-13
Gui key widget

### LINKED ISSUE #1530: Backup local keys on an external device (ie USB) ?
Provide a function "creation of a backup key" for saving the keys locate from the device :
- on Linux dans : /home/thierry/.config/parsec/devices
- on windows : C:\Users\Thierry\AppData\Roaming\parsec\config\devices
to a external device to store on a secure place.

AntoineDupre:
closes #1530


parsec/core/gui/keys_widget.py
```diff
@@ -0,0 +1,98 @@
...
+            show_error(self, translate("EXPORT_KEY_ERROR"), err)
+
+    def _on_import_key(self):
+        file_dialog = QFileDialog()
+        file_dialog.setFileMode(QFileDialog.ExistingFile)
+        key_file, _ = file_dialog.getOpenFileName(
+            parent=self,
+            caption="Import a key",
+            filter="Keys files (*.keys)",
+            initialFilter="Keys files (*.keys)",
```
Max-7, line 79: Those should be translated.

I can't find a reference to `initialFilter` in the documentation (https://doc.qt.io/qt-5/qfiledialog.html#getOpenFileName). I can see what it's supposed to be doing but where does it come from?

Weirdly enough, the flag `QDialogDialog.HideNameFilterDetails` seems to be set by default, which hide the extensions and just shows `Key files` instead of `Key files (*.keys)`. That may be disturbing for a user who renamed their key file.

I think the better option would be to add two filters: one for `*.keys`, one for `*` and set the `initialFilter` (wherever it comes from) to `*.keys`. This way the user still has the option to select any file if their own key file has a different extension.

parsec/core/gui/keys_widget.py
```diff
@@ -0,0 +1,98 @@
...
+            parent=self,
+            title=translate("ASK_IMPORT_KEY"),
+            message=(
+                f"{new_device.organization_id}<br>{new_device.human_handle.label}<br>{new_device.device_label}"
+            ),
+            button_texts=(translate("ACTION_IMPORT_YES"), translate("ACTION_IMPORT_NO")),
+        )
+        if rep == translate("ACTION_IMPORT_YES"):
+            _, key_name = os.path.split(new_device.key_file_path)
+            shutil.copyfile(
```
Max-7, line 94: We should add a warning if the same key files already exists (I don't mean just the name, but also the org/user/device) to avoid writing over an existing key file or duplicating one.

parsec/core/gui/main_window.py
```diff
@@ -240,6 +244,11 @@ def _show_settings(self):
         d = GreyedDialog(w, title=_("TEXT_SETTINGS_TITLE"), parent=self, width=1000)
         d.exec_()
 
+    def _on_manage_keys(self):
+        w = KeysWidget(config=self.config, parent=self)
+        d = GreyedDialog(w, title=_("TEXT_KEYS_DIALOG"), parent=self, width=1000)
```
Max-7, line 249: We should reload the login_widget device list (if any is displayed) to take into account potential changes to the device list.

parsec/core/gui/keys_widget.py
```diff
@@ -0,0 +1,98 @@
...
+        file_dialog.setFileMode(QFileDialog.ExistingFile)
+        key_file, _ = file_dialog.getOpenFileName(
+            parent=self,
+            caption="Import a key",
+            filter="Keys files (*.keys)",
+            initialFilter="Keys files (*.keys)",
+        )
+        if not key_file:
+            return
+        new_device = load_device_file(Path(key_file))
```
Max-7, line 83: Why are you loading the file here? To check if it's a valid Parsec key file? We should probably add some exception handling.

Max-7 (changes requested):

parsec/core/gui/forms/key_widget.ui
```diff
@@ -0,0 +1,143 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
Max-7, line 1: New .ui files should be added to `parsec-gui.pro`

parsec/core/gui/keys_widget.py
AntoineDupre: Good idea, 

It's strange that `initalFilter` is not in the documentation because it is clearly defined in the pytqt sip file like that: 
```python
getOpenFileName(parent: QWidget = None, caption: str = '', directory: str = '', filter: str = '', initialFilter: str = '', options: Union[QFileDialog.Options, QFileDialog.Option] = 0)
```

parsec/core/gui/keys_widget.py
AntoineDupre: Yes, it's to be sure that it is a valid parsec file. `load_device_file` does not raise exceptions, it returns `none` if loading the file fails.

parsec/core/gui/keys_widget.py
```diff
@@ -0,0 +1,98 @@
...
+                w.setParent(None)
+        devices = list_available_devices(self.config.config_dir)
+        for device in devices:
+            w = KeyWidget(device, parent=self)
+            w.export_clicked.connect(self._on_export_key)
+            self.scroll_content.layout().insertWidget(self.scroll_content.layout().count() - 1, w)
+
+    def _on_export_key(self, device):
+        file_dialog = QFileDialog()
+        output_directory = file_dialog.getExistingDirectory()
```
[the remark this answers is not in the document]
AntoineDupre: :+1:

parsec/core/gui/keys_widget.py
```diff
@@ -0,0 +1,98 @@
...
+            return
+        _, key_name = os.path.split(device.key_file_path)
+        try:
+            shutil.copyfile(device.key_file_path, os.path.join(output_directory, key_name))
+        except IOError as err:
+            show_error(self, translate("EXPORT_KEY_ERROR"), err)
+
+    def _on_import_key(self):
+        file_dialog = QFileDialog()
+        file_dialog.setFileMode(QFileDialog.ExistingFile)
```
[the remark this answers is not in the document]
AntoineDupre: :+1:

parsec/core/gui/keys_widget.py
```diff
@@ -0,0 +1,115 @@
...
+                    parent=self,
+                    title=translate("ASK_OVERWRITE_KEY"),
+                    message=translate("TEXT_OVERWRITE_KEY"),
+                    button_texts=(
+                        translate("ACTION_OVERWRITE_KEY_YES"),
+                        translate("ACTION_IMPORT_NO"),
+                    ),
+                )
+                if not rep == translate("ACTION_OVERWRITE_KEY_YES"):
+                    print("nop")
```
Max-7, line 108: Remove the print()

parsec/core/gui/main_window.py
AntoineDupre: fixed

parsec/core/gui/keys_widget.py
AntoineDupre: fixed

parsec/core/gui/keys_widget.py
```diff
@@ -0,0 +1,98 @@
...
+            self.scroll_content.layout().insertWidget(self.scroll_content.layout().count() - 1, w)
+
+    def _on_export_key(self, device):
+        file_dialog = QFileDialog()
+        output_directory = file_dialog.getExistingDirectory()
+        if not output_directory:
+            return
+        _, key_name = os.path.split(device.key_file_path)
+        try:
+            shutil.copyfile(device.key_file_path, os.path.join(output_directory, key_name))
```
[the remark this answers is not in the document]
AntoineDupre: Good idea, i did not find a way to set a default name and let the user specify the name. I am still using `QFileDialog.getExistingDirectory()`

parsec/core/gui/forms/key_widget.ui
AntoineDupre: fixed

parsec/core/gui/keys_widget.py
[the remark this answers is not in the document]
Max-7: You can let the user specify the name using `QFileDialog.getSaveFileName()` (https://doc.qt.io/qt-5/qfiledialog.html#getSaveFileName). The `dir` argument is not well named but it can take a whole path by default. For example:
`save_path, _ = QFileDialog.getSaveFileName(self, "Title", "/home/max/parsec-myorg-max-debian.keys")`
will open a dialog where the user will be able to chose where to save the keys (defaulting to his home directory), with a default name of "parsec-myorg-max-debian.keys" that the user will be able to change.

parsec/core/gui/keys_widget.py
```diff
@@ -0,0 +1,117 @@
...
+    def __init__(self, config, parent):
+        super().__init__(parent=parent)
+        self.config = config
+        self.setupUi(self)
+        self.reload_devices()
+        self.button_import_key.clicked.connect(self._on_import_key)
+
+    def reload_devices(self):
+        layout = self.scroll_content.layout()
+        for _ in range(layout.count()):
```
Max-7, line 51: Should iter to `layout.count() - 1` to not remove the vertical spacer

Max-7 (approved):

PR #1579, opened 2020-11-18
adding missing event

Sithi5:


touilleMan (approved):
Can you change the newsfragment to "1579.bugfix.rst" with "Fix synchronization potentially not triggered after a file resize"

PR #1580, opened 2020-11-18
adding test for entry-updated

Sithi5:


touilleMan (approved):
should rename the file to `test_event_entry_updated` given it's the only type of event we test ;-)

PR #1581, opened 2020-11-18
Adding backend option to enforce https

Sithi5:


tests/common.py
```diff
@@ -247,6 +247,7 @@ def cook_entry(entry):
     "logged_gui_as_admin",
     "logged_gui_create_two_workspaces",
     "fake_preferred_org_creation_backend_addr",
+    "use_enforce_https",
```
touilleMan, line 250: `backend_enforce_https` to mimic `backend_over_ssl`&`backend_has_webhook` ?

tests/backend/http/test_http.py
```diff
@@ -104,6 +104,19 @@ async def assert_stream_closed_on_peer_side(stream):
...
 
 
+@customize_fixtures(use_enforce_https=True)
+@pytest.mark.trio
+async def test_enforce_https(backend, backend_http_send):
+    assert backend.config.enforce_https is True
+    req = b"GET / HTTP/1.0\r\n\r\n"
+    status, headers, test = await backend_http_send(req=req)
+    assert status == (200, "OK")
+    print()
```
touilleMan, line 114: debug print to remove ?

tests/backend/http/test_http.py
```diff
@@ -104,6 +104,19 @@ async def assert_stream_closed_on_peer_side(stream):
         await stream.send_all(b"GET / HTTP/1.0\r\n\r\n")
 
 
+@customize_fixtures(use_enforce_https=True)
+@pytest.mark.trio
+async def test_enforce_https(backend, backend_http_send):
+    assert backend.config.enforce_https is True
```
touilleMan, line 110: no need to check this given it is the responsibility of `customize_fixture` to enable this stuff (and if it is not enabled, the test should fail by itself)

tests/backend/http/test_http.py
```diff
@@ -104,6 +104,19 @@ async def assert_stream_closed_on_peer_side(stream):
...
+    req = b"GET / HTTP/1.0\r\n\r\n"
+    status, headers, test = await backend_http_send(req=req)
```
touilleMan, line 112: I don't see what is tested here... how are we switching from http to https ?

touilleMan (changes requested):
I see the configuration... but where is the code to actually enforce https in the backend ? 😄

touilleMan:
> My PR isn't done yet ^^

image

bitlogik:
To help in the direction of strengthening your access server security, there also a couple of things you can do, on top of switching to HTTPS :
- Providing HSTS header flag, so that the client remember to use HTTPS (switch itself before sending a request)
- Use a 4096 bits RSA key for the server
- Similarly, use EC384 for EC group, not weak as sect163r1
- Use DNS Certification Authority Authorization (and DNSSEC)
- Use only the latest TLS "good" version (no more 1.0 nor 1.1)
- Use only a subset of encryption protocol, not recognized as "weak"
- and encryption protocol  which are "forward secrecy"
- Use OCSP stapling
- Fully disable session resumption only by cache

See the difference in ratings of the same service but with different proxy settings, these ratings are providing a view on the points to improve.

Analyzed by Qualys SSL Labs :
cloud.guardata.app : A+ score : https://www.ssllabs.com/ssltest/analyze.html?d=cloud.guardata.app
saas.parsec.cloud : B score : https://www.ssllabs.com/ssltest/analyze.html?d=saas.parsec.cloud&s=52.214.73.151

Analyzed by imirhil :
cloud.guardata.app : A+ score : https://tls.imirhil.fr/https/cloud.guardata.app
saas.parsec.cloud : B score : https://tls.imirhil.fr/https/saas.parsec.cloud

Maybe an opened issue dedicated to this matter would be relevant ? Let us know.

touilleMan:
@bitlogik thanks for the feedback

As you say, SSL support is a slippery slope and you end up very fast with a lot of options that should be taken into account (so in the end more code for less tested but security-sensitive features).
I think the line is clear on this topic: production deployment should serve Parsec behind a reverse proxy so all those advanced concerned shouldn't be a concern to us. The main reason Parsec server has SSL support is to allow simple deployment for test purpose and for the unittests.

Regarding this PR, I think `X-Forward-Proto` support on application side is pretty common and simple enough.

touilleMan (approved):

PR #1582, opened 2020-11-19
[GUI] Updated how the GUI handles files

### LINKED ISSUE #1567: GUI keeps reloading (causing blinking and unselecting) when importing/renaming/deleting files

Max-7:
Closes #1567 

Instead of using the FS_ENTRY_UPDATED as initially planned, we still use stats() but avoid clearing the whole table everytime, trying instead to find the differences. We also avoid displaying the spinner everytime we refresh the files to avoid the blinking effect.


touilleMan:
too old, Moore's law has made perf better since 2020 anyway :trollface:

PR #1586, opened 2020-11-23
Show version in macos app

Ironicbay:
If not mentioned, the displayed version of the macOS app will be 0.0.0 instead of the current one.

### DIFF
--- a/packaging/macOS/parsec.spec
+++ b/packaging/macOS/parsec.spec
@@ -1,6 +1,7 @@
 # Parsec Cloud (https://parsec.cloud) Copyright (c) AGPLv3 2019 Scille SAS
 
 import os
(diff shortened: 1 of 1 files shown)


packaging/macOS/parsec.spec
```diff
@@ -1,6 +1,7 @@
 # Parsec Cloud (https://parsec.cloud) Copyright (c) AGPLv3 2019 Scille SAS
 
 import os
+exec(open("../../parsec/_version.py", encoding="utf8").read())
```
touilleMan, line 4: should be written like https://github.com/Scille/parsec-cloud/blob/25025075426176385d8b1adb0b1f563071ee6fa8/setup.py#L10-L12 to explain why

touilleMan (approved):

Issue #1587, opened 2020-11-23
Run core without core extra requirements installed


touilleMan:
Currently there is two flavors of extra requirements:
- backend that should be installed to run the Parsec server
- core that should be installed to run the Parsec client

However the core extra requirements is only needed when running Parsec client on the desktop: it contains Qt, i18n and fuse/winfsp stuff, all of which is not strictly needed for running the actual "core" of Parsec.

This was fine so far but now that we want to provide Parsec on mobile we want to be able to install the bare minimum to run Parsec core.

touilleMan:
We could rename the extra requirements to make them more understandable, core and backend are some silly historical names at this point and I think we could replace them by `desktop` and `server`

This way pip install would become much more readable I think:
```shell
$ pip install parsec  # install the actual "core", i.e. the bare minimum needed to have a parsec client
$ pip install parsec[desktop]
$ pip install parsec[server]
```

FirelightFlagboy:
The `core` exist no more, I'm closing this issue

PR #1588, opened 2020-11-23
Remove gui psutil dead code

touilleMan:
related to #1587


touilleMan:
besides we could also entirely remove the need for psutil by using `_win32` module:
```python
import _win32
PROCESS_QUERY_LIMITED_INFORMATION = 0x1000 # https://docs.microsoft.com/en-us/windows/win32/procthread/process-security-and-access-rights

def pid_is_running():
    try:
        _winapi.OpenProcess(PROCESS_QUERY_LIMITED_INFORMATION , False, 5192)
    except OSError:  # TODO: better error handling, this could mean the process exists but we don't have access to it !
        return False
    return True
```

This is basically what is done in psutil (see https://github.com/giampaolo/psutil/blob/994c429c010049a6a0556ce3b0d1af1f86f27867/psutil/arch/windows/process_utils.c#L146-L171)
However as it name may suggest `_win32` is a hidden module specific to CPython on Windows (it is not part of the Python standard and so may not be available to other implementations...) so it would make the code less portable, so I'm not sure it is worth it...

Max-7 (approved):

PR #1589, opened 2020-11-25
[GUI] Add a warning when a user choses their password

### LINKED ISSUE #525: Add messages to warn the user that a password lost means files lost
Since there's absolutely no way to retrieve a lost password, the user has to be well informed that losing a password means losing the data if it has not been shared with another user or if the user only has one device or if it has not been synced yet.

Max-7:
Closes #525 
Replaces old #614


Sithi5:
Is there a way to edit password afterward ?

Sithi5 (approved):
LGTM 👍

PR #1595, opened 2020-12-03
Improve stats computation in backend

touilleMan:


touilleMan:
wow that's pretty old ^^

PR #1600, opened 2020-12-09
[GUI] Use clipboard content as default URL when joining an org

### LINKED ISSUE #1598: Use clipboard when joining an org to prefill the URL field

Max-7:
Closes #1598


parsec/core/gui/main_window.py
```diff
@@ -271,11 +271,18 @@ def _on_finished(ret):
         )
 
     def _on_join_org_clicked(self):
+        default_url = ""
+        try:
+            # Get the clipboard text, try and convert it to parsec addr and use it as default value
+            default_url = BackendActionAddr.from_url(desktop.get_clipboard()).to_url()
+        except:
```
AntoineDupre, line 278: Can you specify which kind of exception you want to catch here ? 

Maybe you don't need also to catch `BaseException` here and get only:
 `except Exception:`

tests/core/gui/test_main_window.py
```diff
@@ -708,3 +714,98 @@ async def test_link_file_unknown_org(
...
+    await aqtbot.key_click(gui, "o", QtCore.Qt.ControlModifier, 200)
+    text_input_w = await catch_text_input_widget()
+    assert text_input_w
+
+    assert text_input_w.line_edit_text.text() == invitation_organization_link
+
+
+@pytest.mark.gui
+@pytest.mark.trio
+async def test_join_organization_device_link_in_clipboard(
```
AntoineDupre, line 745: It seems that only the invitation link is different from the previous test `test_join_organization_org_link_in_clipboard`. Maybe those two tests can be merged in one parameterized test.

tests/core/gui/test_main_window.py
```diff
@@ -708,3 +714,98 @@ async def test_link_file_unknown_org(
...
+    await aqtbot.key_click(gui, "o", QtCore.Qt.ControlModifier, 200)
+    text_input_w = await catch_text_input_widget()
+    assert text_input_w
+
+    assert text_input_w.line_edit_text.text() == invitation_device_link
+
+
+@pytest.mark.gui
+@pytest.mark.trio
+async def test_join_organization_user_link_in_clipboard(
```
AntoineDupre, line 769: same here

tests/core/gui/test_main_window.py
```diff
@@ -708,3 +714,98 @@ async def test_link_file_unknown_org(
...
+    await aqtbot.key_click(gui, "o", QtCore.Qt.ControlModifier, 200)
+    text_input_w = await catch_text_input_widget()
+    assert text_input_w
+
+    assert text_input_w.line_edit_text.text() == invitation_user_link
+
+
+@pytest.mark.gui
+@pytest.mark.trio
+async def test_join_organization_other_text_in_clipboard(
```
AntoineDupre, line 793: same here

AntoineDupre (changes requested):

AntoineDupre (approved):

PR #1607, opened 2020-12-15
Fix unhandled trio.Cancelled in fuse mountpoint when trio loop gets c…

touilleMan:
…losed during processing

### DIFF
--- a/newsfragments/1607.bugfix.rst
+++ b/newsfragments/1607.bugfix.rst
@@ -1,0 +1,1 @@
+Fix unhandled error on linux/macOS when logout occures during mountpoint processing.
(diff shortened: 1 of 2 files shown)


bitlogik:
We know this is already merged, but we would suggest to put the new caught exception trio.Cancelled before the generic Exception catch. Not 100% sure but Python seems to inspect the except clauses sequentially, and this would void this new exception catch.

touilleMan:
trio.Cancelled doesn't inherit from Exception but from BaseException:
```python
>>> issubclass(trio.Cancelled, Exception)
False
>>> issubclass(trio.Cancelled, BaseException)
True
```

The idea is exception needed for "system usage" (as opposed by user exceptions) such as `StopIteration` or `SystemExit` all inherit from `BaseException` so that they can go through `except Exception` blocks user is allowed to put.

bitlogik:
OK great. Now it's sure this is OK. We mixed BaseException and Exception.

PR #1612, opened 2020-12-18
Fix macOS packaging in CI

touilleMan:


.azure-pipelines.yml
```diff
@@ -401,7 +403,7 @@ jobs:
 
 - job: 's2_macos_app_build'
   displayName: 'MacOS app build'
-  condition: startsWith(variables['build.sourceBranch'], 'refs/tags/')
+  # condition: startsWith(variables['build.sourceBranch'], 'refs/tags/')
```
touilleMan, line 406: Don't forget to uncomment this before merging !

packaging/macOS/README.md
```diff
@@ -10,15 +10,18 @@ To release an app on MacOS, even without dealing with the App Store, we must go
 2. __Sign the app__: Using an active Apple Developer account, we sign our code and its dependencies with a custom certificate created using the account.
 3. __Notarize the app__: Notarizing the app is equivalent to asking Apple to approve it for distribution. Since MacOS 10.14.5, all applications downloaded outside of the App Store must be notarized to run. See https://developer.apple.com/documentation/xcode/notarizing_macos_software_before_distribution
 4. __Staple the app__: An app notarization will be checked with an online request when a user tries to execute it. Stapling "attaches" the notarization to the app bundle itself, so that this check can be performed offline.
+5. __Bundle into .dmg__: A nice format for distribution in MacOS is a `.dmg` bundle, which compresses the app and allows the user to simply drag and drop the app inside their `/Applications` folder.
```
touilleMan, line 13: image

packaging/macOS/launch_script.py
```diff
@@ -3,9 +3,12 @@
 import parsec.cli
 import sys
 import os
+import locale
 
 os.environ["SENTRY_URL"] = "https://863e60bbef39406896d2b7a5dbd491bb@sentry.io/1212848"
 os.environ["PREFERRED_ORG_CREATION_BACKEND_ADDR"] = "parsec://saas.parsec.cloud/"
 
+locale.setlocale(locale.LC_ALL, "")
```
touilleMan, line 11: A comment to explain why this is needed would be welcome here

packaging/macOS/README.md
```diff
@@ -93,12 +96,35 @@ $ rm dist/parsec.zip
...
 $ spctl --assess --type execute -vvv "dist/parsec.app"
 ```
+
+
+### 6 - Bundle into .dmg
+
+The app is now ready to be bundled for distribution. To this end, we create a drag & drop `.dmg` folder.
+A `.dmg` background is needed at this point, many neutral backgrounds can be found with a quick search, until we eventually design our own.
+
+#### Create the .dmg
```
touilleMan, line 114: Can't we do this with a command line instead ? (a quick search seems to indicate `hdiutil` is the right tool for this job  https://apple.stackexchange.com/a/50040)

packaging/macOS/README.md
```diff
@@ -93,12 +96,35 @@ $ rm dist/parsec.zip
...
 
 ```shell
 $ spctl --assess --type execute -vvv "dist/parsec.app"
 ```
+
+
+### 6 - Bundle into .dmg
+
+The app is now ready to be bundled for distribution. To this end, we create a drag & drop `.dmg` folder.
+A `.dmg` background is needed at this point, many neutral backgrounds can be found with a quick search, until we eventually design our own.
```
touilleMan, line 112: Can't we just store a background image in our sources instead ? Doesn't seem like a good idea to have to search for a background image each time we want to do the packaging ;-)

.azure-pipelines.yml
```diff
@@ -410,7 +412,7 @@ jobs:
     submodules: true
   - task: UsePythonVersion@0
     inputs:
-      versionSpec: '$(python.version)'
+      versionSpec: '$(python.version_for_macos_packaging)'
   - bash: |
```
touilleMan, line 416: instead of packaging the `Parsec.app` folder in a tar.bz2, shouldn't we just do the dmg packaging here ?

PR #1613, opened 2020-12-21
[GUI] Used user inputs to make password validation stronger

### LINKED ISSUE #1601: Improve the password validator by adding some user_inputs
The zxcvbn library allow to set some user_inputs in its check method to improve its results.
In our case, we may use the user's data (name, mail, org_name) as user_inputs to avoid one password like "Name_Mail_Orgname".

```python
from zxcvbn import zxcvbn

results = zxcvbn('JohnSmith123', user_inputs=['John', 'Smith'])
```

https://github.com/dwolfhub/zxcvbn-python

Max-7:
Closes #1601


parsec/core/gui/claim_device_widget.py
```diff
@@ -380,6 +381,9 @@ def __init__(self, jobs_ctx, claimer, device_email):
         self.label_wait.hide()
         self.button_ok.clicked.connect(self._on_claim_clicked)
 
+    def _on_device_text_changed(self, device_text):
+        self.widget_password.user_inputs = [device_text]
```
touilleMan, line 385: I don't understand why we want to use the device name as password...

 edit: I understand now (we use the non-password inputs to check the security of the password), but it's not really clear at first look
We should add a comment here about this

parsec/core/gui/password_validation.py
```diff
@@ -34,23 +36,36 @@ def get_password_strength_text(password_score):
...
 
 
 class PasswordStrengthWidget(QWidget, Ui_PasswordStrengthWidget):
     def __init__(self, parent=None):
         super().__init__(parent)
         self.setupUi(self)
+        self._user_inputs = []
+
+    @property
+    def user_inputs(self):
```
touilleMan, line 53: `user_inputs` feels too vague for me (it took me a while to understand we were expecting to receive everything **but** the password)
We should rename this to something like `password_strength_context_strings`
On top of that it's unclear `user_inputs` is a list of strings, typing would be welcome here ;-) (another possibility would be to use `*user_inputs` as parameter so that it's obvious we receive a list of string)

And last, I'm not a big fan of using the @property setter (using a function makes it clear we are not hacking anything, and we don't have to provide a getter on the sanitized data that make expose it as a public attribute where it is supposed to be only internally used)

parsec/core/gui/password_validation.py
```diff
@@ -34,23 +36,36 @@ def get_password_strength_text(password_score):
...
+        return self._user_inputs
+
+    @user_inputs.setter
+    def user_inputs(self, user_inputs):
+        self._user_inputs = []
+        # User inputs are passed raw, we have to clean them up a bit
+        # We split them up on the most common characters
+        for i in user_inputs:
+            self._user_inputs.extend([x for x in re.split(r"\W+", i) if x and len(x) > 3])
```
touilleMan, line 62: why not the more pythonic
```python
self._user_inputs += [x for x in re.split(r"\W+", i) if x and len(x) > 3]
```
?

parsec/core/gui/password_validation.py
```diff
@@ -34,23 +36,36 @@ def get_password_strength_text(password_score):
...
+    @property
+    def user_inputs(self):
+        return self._user_inputs
+
+    @user_inputs.setter
+    def user_inputs(self, user_inputs):
+        self._user_inputs = []
+        # User inputs are passed raw, we have to clean them up a bit
+        # We split them up on the most common characters
+        for i in user_inputs:
```
touilleMan, line 61: `i` feels like a number, maybe another variable ?

parsec/core/gui/password_validation.py
touilleMan: Regarding my last rant, `PasswordChoiceWidget` also use `user_inputs` getter, however it could use a `PasswordStrengthWidget.get_password_strength(password)` method instead

tests/core/gui/test_password_validation.py
```diff
@@ -25,3 +26,56 @@ def test_password_text(core_config):
...
+    assert not p.label_mismatch.isHidden()
+    assert p.label_mismatch.text() == "Does not match the password."
+
+
+@pytest.mark.gui
+def test_password_choice_widget_with_user_inputs(qtbot, core_config):
+    switch_language(core_config, "en")
+
+    p = PasswordChoiceWidget(parent=None)
+    p.user_inputs = ["william.j.blazkowicz@wolfenstein.de"]
```
touilleMan, line 74: should be `@wolfenstein.at`
https://en.wikipedia.org/wiki/Hohenwerfen_Castle
:trollface:

touilleMan (changes requested):

tests/core/gui/test_password_validation.py
Max-7: Nop! 
https://en.wikipedia.org/wiki/Wewelsburg

touilleMan: xD

parsec/core/gui/password_validation.py
Max-7: Renamed it to `excluded_strings`. It was named `user_inputs` because zxcvbn (the module used for password validation) calls it `user_inputs`.

touilleMan: 👍

touilleMan (approved):

PR #1617, opened 2020-12-30
Fix macos installation

C1em:


touilleMan:
used to improve #1612

PR #1621, opened 2021-01-07
[GUI] Improved workspace loading (less reencryption queries)

### LINKED ISSUE #1619: The more workspaces, the slower they are to appear in GUI on startup
with Parsec v2.2.2:

```
> /home/emmanuel/projects/parsec-cloud2/parsec/core/remote_devices_manager.py(148)get_user_and_devices()
-> rep = await self._backend_cmds.user_get(user_id)
(Pdb) up
> /home/emmanuel/projects/parsec-cloud2/parsec/core/remote_devices_manager.py(96)get_user()
-> verified_user, verified_revoked_user, _ = await self.get_user_and_devices(
(Pdb) 
> /home/emmanuel/projects/parsec-cloud2/parsec/core/fs/remote_loader.py(235)get_user()
-> return await self.remote_devices_manager.get_user(user_id, no_cache=no_cache)
(Pdb) 
> /home/emmanuel/projects/parsec-cloud2/parsec/core/fs/workspacefs/workspacefs.py(209)get_reencryption_need()
-> _, revoked_user = await self.remote_loader.get_user(user_id, no_cache=True)
(Pdb) 
> /home/emmanuel/projects/parsec-cloud2/parsec/core/gui/workspaces_widget.py(135)_add_workspacefs()
-> reenc_needs = await workspace_fs.get_reencryption_need()
```

GUI's workspace widgets check for reencryption needs each time when created.
However reencryption needs check doesn't use cache (otherwise a recently revoked user would not be detected as a reencryption need).
This lead to 2 + N requests (with N the number of user that currently have a role in the workspace) to the backend for each workspaces to:
- get realm status (to make sure the realm is not already beeing reencrypted)
- get realm role certificates (single request to the backend gets us all the certificates)
- get user revoked certificate for each member

The bad part is the GUI doesn't display the widget until the check is done, and no cache means this is growing linearly with the number of workspaces :'-(

To solve this, the simplest solution would be to make this reencryption need check asynchronous in the GUI (so first display all the workspaces, run the check in the background, then finally add the "reencryption needed" warning where needed)

Another potential quick win would be to do this check only when the current user is owner of the workspace (given only there her can proceed to the reencryption). Currently the check is done for everybody (so everybody is see that the workspace need reencryption... not sure if it's a good or a bad thing) but clicking on the reencrypt button pop a dialog explaining only owner can do that.

However this is not 100% perfect given it doesn't solve the « `(2 + N) * W` requests to the backend each time the workspaces widget page is loaded » which seems a bit overkill for a check that is very rarely needed.
On top of that, that greedy check doesn't protect us against a malicious backend that wants to lie by omision (hence hiding us the fact that a revocation certificate has been issued toward a given user).
So I think in the long run it migh be worth it to rework the cache system (for instance caching all the user/device/role certifications involved in a workspace once, then listening for backend events to update them)

Max-7:
Closes #1619

### DIFF
--- a/newsfragments/1619.bugfix.rst
+++ b/newsfragments/1619.bugfix.rst
@@ -1,0 +1,1 @@
+Improved workspace loading performance (less query for reencryption)
--- a/parsec/core/gui/workspaces_widget.py
+++ b/parsec/core/gui/workspaces_widget.py
@@ -129,14 +129,7 @@ async def _do_workspace_list(core):
             # reencryption operation
             pass
 
-        # Get reencryption needs
-        reenc_needs = None
-        try:
-            reenc_needs = await workspace_fs.get_reencryption_need()
-        except FSBackendOfflineError:
-            pass
-
-        workspaces.append((workspace_fs, ws_entry, users_roles, files, timestamped, reenc_needs))
+        workspaces.append((workspace_fs, ws_entry, users_roles, files, timestamped))
 
     user_manifest = core.user_fs.get_user_manifest()
     available_workspaces = [w for w in user_manifest.workspaces if w.role]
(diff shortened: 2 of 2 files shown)


touilleMan (approved):
:+1:

PR #1624, opened 2021-01-22
[GUI] Add extra cancel user invitation button

AntoineDupre:
Add extra cancel user invitation button

invitation

see issue #1622


touilleMan:
Shouldn't we also remove the "no entry" icon ? Keeping the icon seems to signify it does something different than the "cancel invitation" button so it's still a bit unclear

Max-7:
This was done by another PR, invitation buttons look like this now:
invitation

PR #1634, opened 2021-02-03
Postgres improved connection

### LINKED ISSUE #1618: Forced LISTEN query termination on postgres end is not detected by the backend
The backend uses a long lasting LISTEN query to pass all the events:
https://github.com/Scille/parsec-cloud/blob/b80cdb3922da9bf1614900885d8244b294cafd1c/parsec/backend/postgresql/handler.py#L229-L231

If for whatever reason the PostgreSQL server decides to terminate this LISTEN query, the backend doesn't seem to be notified about it (no exception is raised), so it becomes "blind": no event can be provided to it and it cannot tell if it's because there is no currently no events or because it cannot see them !

We should find a way to make sure this doesn't occur:
- ideally to get an exception is if the query is terminated (so we can log the trouble and restart the query)
- restarting the query periodically. However it doesn't look like a very good solution given it brings concurrency issues when the query is restarted and we should look for short query time to minimize "blindness time"... but the shorter the query time, the bigger the concurrency issues !

touilleMan:
fix #1618

### DIFF
--- a/tests/backend/test_postgres_db.py
+++ b/tests/backend/test_postgres_db.py
@@ -13,13 +13,13 @@ def records_filter_debug(records):
     return [record for record in records if record.levelname != "DEBUG"]
 
 
-async def wait_for_listener(conn, to_terminate=False, timeout=3.0):
+async def wait_for_listeners(conn, to_terminate=False, timeout=3.0):
     with trio.fail_after(timeout):
         while True:
             rows = await conn.fetch(
                 "SELECT pid FROM pg_stat_activity WHERE query ILIKE 'listen %' AND state ILIKE 'idle'"
             )
-            if not to_terminate and rows or to_terminate and not rows:
+            if (not to_terminate and rows) or (to_terminate and not rows):
                 return [r["pid"] for r in rows]
 
 
@@ -39,11 +39,12 @@ async def test_postgresql_notification_listener_terminated(postgresql_url, backe
         with pytest.raises(ConnectionError):
 
             async with backend_factory(config={"db_url": postgresql_url}):
-                pid, = await wait_for_listener(conn)
+                pid, = await wait_for_listeners(conn)
                 value, = await conn.fetchrow("SELECT pg_terminate_backend($1)", pid)
                 assert value
-                await trio.sleep(1)
-                assert False, "Should be cancelled by now"
+                # Wait to get cancelled by the backend app
+                with trio.fail_after(3):
+                    await trio.sleep_forever()
 
 
 @pytest.mark.trio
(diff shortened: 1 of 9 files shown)


tests/backend/test_postgres_db.py
```diff
@@ -3,19 +3,49 @@
...
+
+    async with triopg.connect(postgresql_url) as conn:
+
+        with pytest.raises(ConnectionError):
+
+            async with backend_factory(config={"db_url": postgresql_url}):
+                pid, = await wait_for_listener(conn)
+                value, = await conn.fetchrow("SELECT pg_terminate_backend($1)", pid)
+                assert value
+                await trio.sleep(1)
```
touilleMan, line 45: I find clearer to use `fail_after` with a comment:

```python
with trio.fail_after(1):
    # Wait for backend to take into account the connection termination
    trio.sleep_forever()
```

tests/backend/test_postgres_db.py
```diff
@@ -30,39 +60,92 @@ async def test_postgresql_connection_not_ok(
...
+            )
+            # Connect to PostgreSQL database
+            async with triopg.connect(postgresql_url) as conn:
+                # Wait for the backend to be connected
+                pid, = await wait_for_listener(conn)
+                # Terminate the backend listener connection
+                value, = await conn.fetchrow(f"SELECT pg_terminate_backend($1)", pid)
+                assert value
+                # The _run_backend should raise a Connection error and cancel the nursery
+                await trio.sleep(3)
```
touilleMan, line 105: same as above

tests/backend/test_postgres_db.py
```diff
@@ -30,39 +60,92 @@ async def test_postgresql_connection_not_ok(
...
+                assert False
+
+
+@pytest.mark.trio
+@pytest.mark.postgresql
+async def test_retry_policy_allow_retry(postgresql_url, unused_tcp_port, asyncio_loop):
+    host = "localhost"
+    port = unused_tcp_port
+    app_config = BackendConfig(
```
touilleMan, line 114: Shouldn't we create a `backend_config_factory` (or a `backend_config_default_params`) fixture for this ? (would be shared between this file's tests and `backend_factory`)

tests/backend/test_postgres_db.py
```diff
@@ -30,39 +60,92 @@ async def test_postgresql_connection_not_ok(
...
+
+            # Test for 10 cycles
+            for _ in range(10):
+                # Wait for the backend to be connected
+                pid, = await wait_for_listener(conn)
+                # Terminate the backend listener connection
+                value, = await conn.fetchrow("SELECT pg_terminate_backend($1)", pid)
+                assert value
+                # Wait for the listener to terminate
+                await wait_for_listener(conn, to_terminate=True)
```
touilleMan, line 148: might be interesting to store the pid of the terminated connections to ensure a fresh connection appears for each loop

vxgmichel:
@touilleMan all comments addressed here: 0f3e29e

PR #1636, opened 2021-02-08
WIP: Improve backend actives users stats

### LINKED ISSUE #1635: Improve the response of organization stats API.
In parsec-bms we have one daily task that uses the organizations stats API to retrieve their uses. (users, storages etc.) The retrieved usages are used to build the monthly billing.
Currently the organizations stats API response use this format:
```
{'data_size': xxxxx, 'metadata_size': xxxxx, 'status': 'ok', 'users': xxxxx}
```

The users attribute is the sum of all the organization's users. (actives and revoqued). The problem is that we don't want to count the revoqued users in our billing system.

Is it possible to improve the response API to split the users attribute in two new attributes ? (actives_users, revoqued_users)

AntoineDupre:
closes #1635


touilleMan:
already implemented, closing

PR #1637, opened 2021-02-08
Cleanup deprecations in backend addrs

### LINKED ISSUE #1604: Backend addresses no longer need rvk field
Since api v2, root verify key is provided to invitee by inviter as part of the invite/claim process, hence there is no need for the `rvk` in the address.

touilleMan:
closes #1604


parsec/core/gui/central_widget.py
```diff
@@ -199,10 +230,10 @@ def handle_event(self, event, **kwargs):
...
             )
         elif event == CoreEvent.SHARING_UPDATED:
-            new_entry = kwargs["new_entry"]
-            previous_entry = kwargs["previous_entry"]
-            new_role = getattr(new_entry, "role", None)
-            previous_role = getattr(previous_entry, "role", None)
+            new_entry: WorkspaceEntry = kwargs["new_entry"]  # type: ignore
+            previous_entry: Optional[WorkspaceEntry] = kwargs["previous_entry"]  # type: ignore
+            new_role = new_entry.role
+            previous_role = previous_entry.role if previous_entry is not None else None
```
Max-7, line 236: Are we sure that if `new_entry` and `previous_entry` are not None, they will always have a `role` attribute? The use of `getattr` here makes me think that it's not always the case.

parsec/core/gui/central_widget.py
```diff
@@ -305,44 +341,59 @@ def _on_organization_stats_success(self, job):
...
-    def go_to_file_link(self, workspace_id, path, mount=False):
+    def go_to_file_link(self, addr: BackendOrganizationFileLinkAddr, mount: bool = True) -> None:
+        """
+        Raises:
+            GoToFileLinkBadOrgazationIDError
+            GoToFileLinkBadWorkspaceIDError
+            GoToFileLinkPathDecryptionError
+        """
+        if addr.organization_id != self.core.device.organization_id:
+            raise GoToFileLinkBadOrgazationIDError
```
Max-7, line 362: `Orgazation`

Max-7 (changes requested):

parsec/core/gui/central_widget.py
touilleMan: 👍

parsec/core/gui/central_widget.py
touilleMan: `new_entry` and `previous_entry` are of type `WorkspaceEntry` which has a `role` attribute, so we are sure of this ;-)

I think the use of `getattr(previous_entry, "role", None)` was to simply handle the case when `previous_entry` is `None` (and then the same pattern was applied for `new_entry` for consistency)

parsec/core/gui/main_window.py
```diff
@@ -512,106 +532,106 @@ def add_new_tab(self):
...
+                        ),
+                    )
+                    break
+            else:
+                show_error(
+                    self,
+                    _("TEXT_FILE_LINK_NOT_IN_ORG_organization").format(
+                        organization=file_link_addr.organization_id
+                    ),
+                )
```
vxgmichel, line 574: Might be a bit easier to read as:
```python
# We're done here
if not file_link_addr:
    return

# Find the device corresponding to the organization in the link
for available_device in list_available_devices(self.config.config_dir):
    if available_device.organization_id == file_link_addr.organization_id:
        break

# Organization not found
else:
    show_error(...)

# Select the corresponding device
login_w = self.tab_center.widget(idx).get_login_widget()
login_w._on_account_clicked(available_device)

# Set the path in advance
instance_widget = self.tab_center.widget(idx)
instance_widget.set_workspace_path(file_link_addr)

# Prompt the user for the need to log in first
show_info(...)

File: parsec/core/gui/main_window.py [hunk 4]
```diff
@@ -512,106 +532,106 @@ def add_new_tab(self):
...
-            _("TEXT_FILE_LINK_PLEASE_LOG_IN_organization").format(
-                organization=action_addr.organization_id
-            ),
-        )
-        self.switch_to_login_tab(action_addr)
+            except GoToFileLinkPathDecryptionError:
+                show_error(self, _("TEXT_INVALID_URL"))
+                return
+            else:
+                self.switch_to_tab(idx)
```
vxgmichel, line 611: Isn't it confusing to switch tab only if processing the file link succeeds? That means the error is going to be prompted in an unrelated tab.

parsec/core/types/backend_address.py
```diff
@@ -78,7 +84,24 @@ def from_url(cls, url: str):
         else:
             params = {}
 
-        path = unquote_plus(split.path)
+        if allow_http_redirection and split.scheme in ("http", "https"):
+            # `no_ssl` is defined by http/https scheme and shouldn't be
+            # overwritten by the query part of the url
+            params["no_ssl"] = ["true" if split.scheme == "http" else "false"]
```
vxgmichel, line 90: I'm a bit confused about this.. 

A `parsec` URL only needs to know about SSL if it plans to reach the backend, which is not the case for links to a file within a workspace. But we do need to use the proper protocol (`http` or `https`) if we want the backend to make the `http(s)` to `parsec` redirection. And we're assuming that the redirection and the backend API will always share the ssl/no_ssl config (since they run on same port using the same protocol), so we use the `http` or `https` information to set the `no_ssl` parameter in the backend address, possibly overwriting the existing one.

Am I getting this right?

vxgmichel, line 90: Also, why are parameters a list of string instead of a simple string?

parsec/core/types/backend_address.py
```diff
@@ -78,7 +84,24 @@ def from_url(cls, url: str):
...
+            # Remove the `/redirect/` path prefix
+            split_path = path.split("/", 2)
+            if split_path[:2] != ["", "redirect"]:
+                raise ValueError("HTTP to Parsec redirection URL must have a `/redirect/...` path")
+            try:
+                path = f"/{split_path[2]}"
+            except IndexError:
+                path = ""
```
vxgmichel, line 98: Isn't this a `ValueError` similar to the ones raised below?

parsec/core/types/backend_address.py
```diff
@@ -137,7 +160,12 @@ def use_ssl(self):
         return self._use_ssl
 
 
-class OrganizationParamsFixture(BackendAddr):
+class BackendOrganizationAddr(BackendAddr):
+    """
+    Represent the URL to access an organization within a backend
+    (e.g. ``parsec://parsec.example.com/MyOrg?rvk=7NFDS4VQLP3XPCMTSEN34ZOXKGGIMTY2W2JI2SPIHB2P3M6K4YWAssss``)
```
vxgmichel, line 166: Is this really a valid URL? I doesn't contain any action to perform.

parsec/crypto.py
```diff
@@ -157,27 +157,47 @@ def encrypt_for_self(self, data: bytes) -> bytes:
 # Helpers
 
 
+# Binary encoder/decoder for url use.
+# Notes:
+# - We replace padding char `=` by a simple `s` (which is not part of
+#   the base32 table so no risk of collision) to avoid copy/paste errors
+#   and silly escaping issues when carrying the key around.
+# - We could be using base64url (see RFC 4648) which would be more efficient,
+#   but backward compatibility prevent us from doing it :'(
```
vxgmichel, line 166: Funny how `base64.urlsafe_b64encode` still includes some `=`: 
```python
In [3]: base64.urlsafe_b64encode(b"a")
Out[3]: b'YQ=='
```

vxgmichel (changes requested):
A single important change is requested (about qt and core threads), but LGTM otherwise :+1:

parsec/crypto.py
touilleMan: yes, ulrsafe is not about the `=` character but about removing `+` and `/` (see https://docs.python.org/3/library/base64.html#base64.urlsafe_b64decode)

copy/paste error is much less of an issue now that the invite link is passed trough mail

parsec/core/gui/main_window.py
touilleMan: 👍

parsec/core/gui/main_window.py
touilleMan: yep, good catch !

parsec/core/types/backend_address.py
touilleMan: only `BackendActionAddr` and children contain action parameter

parsec/core/types/backend_address.py
touilleMan: yup, the idea is `ValueError` is the standard exception if the provided URL doesn't respect the expected format

parsec/core/types/backend_address.py
touilleMan: it's because an url param can be repeated multiple times (e.g. `http://example.com/foo?spam=1&spam=2`, which `parse_qs` would turn into `{"spam": [1, 2]}`)

touilleMan: Yes, as you say this imply the http:// and parsec:// urls should use the same netloc (but this seems pretty resonable)

> possibly overwriting the existing one.

No, the `no_ssl` param is just not known. So we use the http vs https information to determine `no_ssl`. On an totally unrelated level, a `no_ssl` param could be provided as part of the url (because why not ? typically an attacker might want to mess with this...), but it shouldn't be taken into account.

vxgmichel (approved):
LGTM :+1:

Max-7 (approved):

PR #1638, opened 2021-02-10
Disable core file logging in packaging

### LINKED ISSUE #1616: Limit log size in client
Currently logs in `~/.config/parsec/parsec-core.log` grow without limit. This is not much of an issue given the amount of logs is rather low, but it's always better to avoid infinite grow ;-)

touilleMan:
closes #1616

### DIFF
--- a/packaging/macOS/launch_script.py
+++ b/packaging/macOS/launch_script.py
@@ -1,9 +1,11 @@
 # Parsec Cloud (https://parsec.cloud) Copyright (c) AGPLv3 2019 Scille SAS
 
-import parsec.cli
 import sys
 import os
 import locale
+
+from parsec.cli import cli
+
 
 os.environ["SENTRY_URL"] = "https://863e60bbef39406896d2b7a5dbd491bb@sentry.io/1212848"
 os.environ["PREFERRED_ORG_CREATION_BACKEND_ADDR"] = "parsec://saas.parsec.cloud/"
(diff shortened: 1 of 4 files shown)


packaging/macOS/launch_script.py
```diff
@@ -1,16 +1,17 @@
...
 os.environ["SENTRY_URL"] = "https://863e60bbef39406896d2b7a5dbd491bb@sentry.io/1212848"
 os.environ["PREFERRED_ORG_CREATION_BACKEND_ADDR"] = "parsec://saas.parsec.cloud/"
 
 # The locale values appear to be None by default when launching the app from the
 # Finder. If not set, Click will report an encoding error and the app won't launch.
 locale.setlocale(locale.LC_ALL, "")
 
-sys.argv = [sys.argv[0], "core", "gui", *sys.argv[1:]]
-parsec.cli.cli.main()
+cli(args=[sys.argv[0], "core", "gui", *sys.argv[1:]])
```
Ironicbay, line 17: sys.argv[0] isn't needed : `cli(args=["core", "gui", *sys.argv[1:]])` works as intended

Ironicbay (changes requested):

packaging/macOS/launch_script.py
touilleMan: good catch !

Ironicbay (approved):

PR #1639, opened 2021-02-10
Remove log filter param

touilleMan:


parsec/logging.py
```diff
@@ -31,16 +30,6 @@ def configure_logging(log_level=None, log_format=None, log_file=None, log_filter
...
-    if log_filter:
-        log_filter = re.compile(log_filter)
-
-        def dropper(logger, method_name, event_dict):
-            if not log_filter.match(str(event_dict)):
-                raise structlog.DropEvent
-            return event_dict
-
-        shared_processors.append(dropper)
-
```
AntoineDupre, line 43: This feature is removed because it's a not used feature ?

AntoineDupre (approved):

parsec/logging.py
touilleMan: yup, got bitten by YAGNI here ;-)

PR #1644, opened 2021-02-25
Add force option to diskutil unmount calls

Ironicbay:
This can prevent errors during login after a crash, or in various situations where a mountpoint can't unmount without it (when a terminal's cwd is inside the mountpoint for instance)

### DIFF
--- a/parsec/core/mountpoint/fuse_runner.py
+++ b/parsec/core/mountpoint/fuse_runner.py
@@ -232,7 +232,7 @@ async def _stop_fuse_thread(
         # The schedule_exit() solution doesn't work on macOS, instead freezes the application for
         # 120 seconds before a timeout occurs. The solution used is to call this function (macOS
         # equivalent to fusermount) in a subprocess to unmount.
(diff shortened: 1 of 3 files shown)


parsec/core/mountpoint/fuse_runner.py
```diff
@@ -232,7 +232,7 @@ async def _stop_fuse_thread(
         # The schedule_exit() solution doesn't work on macOS, instead freezes the application for
         # 120 seconds before a timeout occurs. The solution used is to call this function (macOS
         # equivalent to fusermount) in a subprocess to unmount.
-        await trio.run_process(["diskutil", "unmount", str(mountpoint_path)])
+        await trio.run_process(["diskutil", "unmount", "force", str(mountpoint_path)])
```
touilleMan, line 235: funny the option is `force` and not `--force` ?

touilleMan (approved):

parsec/core/mountpoint/fuse_runner.py
Ironicbay: Indeed: https://ss64.com/osx/diskutil.html

PR #1645, opened 2021-02-25
[GUI] Detect caps lock when entering a password

### LINKED ISSUE #1643: Add a caps lock status indicator when entering a password

Max-7:
Closes #1643 

A few things to note:
- no tests. It tried using qtbot.keyClick() with Qt.Key_CapsLock but it didn't work. I could load a native library and try to set caps lock with ctypes but it seems overkill to just check if a small message is displayed or not. And anyway, I'm not sure that this would work with the CI or if it requires a physical keyboard connected or something like that.
- we check if caps lock is on when creating the object and then again when a letter is entered, but this means that if the user turns caps lock off, the message will still be displayed until he enters another letter. It could be improved but would require overriding keyPressedEvent, which is more painful that just using a slot.

### DIFF
--- a/parsec/core/gui/desktop.py
+++ b/parsec/core/gui/desktop.py
@@ -1,9 +1,8 @@
 # Parsec Cloud (https://parsec.cloud) Copyright (c) AGPLv3 2019 Scille SAS
 
 import platform
-import os
 
-from ctypes import cdll, c_uint, byref, c_void_p, c_char_p, c_int, c_short
+from ctypes import cdll, c_uint, byref, c_void_p, c_int, c_short
 
 from PyQt5.QtCore import QUrl, QFileInfo, QSysInfo, QLocale
 from PyQt5.QtGui import QDesktopServices, QGuiApplication, QClipboard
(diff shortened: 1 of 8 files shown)


parsec/core/gui/desktop.py
```diff
@@ -40,3 +44,49 @@ def get_locale_language():
...
+                if not native_library:
+                    return False
+                native_library.XOpenDisplay.restype = c_void_p
+
+            state = c_uint()
+            XkbUseCoreKdb = c_uint(0x0100)
+            CapsLockMask = 0x1
+            display = native_library.XOpenDisplay(None)
+            if display:
+                native_library.XkbGetIndicatorState(display, XkbUseCoreKdb, byref(state))
```
AntoineDupre, line 87: This line segfault on my machine (running from a conda environment):
```[1]    29154 segmentation fault (core dumped) ```

The library looks to be properly loaded:
```
<CDLL 'libX11.so', handle 55d874e894c0 at 0x7fd068505410>
```

I'll investigate it. Did you try it on another linux env ?

File: parsec/core/gui/desktop.py [hunk 2]
```diff
@@ -40,3 +44,49 @@ def get_locale_language():
...
+
+    global native_library
+
+    if platform.system() == "Windows":
+        try:
+            if not native_library:
+                native_library = cdll.LoadLibrary("User32")
+                native_library.GetKeyState.restype = c_short
+
+            VK_CAPITAL = c_int(0x14)
```
touilleMan, line 66: Hold my beer !

parsec/core/gui/tr/parsec_fr.po
```diff
@@ -2558,3 +2560,6 @@ msgstr "Créer un nouvel espace de travail"
 msgid "ACTION_WORKSPACE_GOTO_FILE_LINK"
 msgstr "Trouver un fichier via son lien"
 
+#~ msgid "TEXT_WORKSPACE_SHesARING_OFFLINE"
```
touilleMan, line 2563: dead stuff to remove ?

parsec/core/gui/desktop.py
Max-7: Only on my VM running debian.

touilleMan:
Closing, will be implemented in the new web-based GUI instead

PR #1647, opened 2021-02-27
Add EntryID.new() method so EntryID(None) correctly raise a TypeError…

touilleMan:
… instead of believing we want a new EntryID


vxgmichel (approved):
:+1:

PR #1655, opened 2021-03-04
Add the number of workspaces in the organizations stats

vamonte:


newsfragments/1655.feature.rst
```diff
@@ -0,0 +1 @@
+Add the number of workspaces in the organization stats
```
touilleMan, line 1: there is the `api` category instead of `feature`

On top of that given we changed the api, we should bump the api minor version from `1.2` to `1.3` (see `parsec/api/version.py`) and tell about this 1.3 version in the newsfragment

parsec/backend/memory/organization.py
```diff
@@ -100,7 +104,11 @@ async def stats(self, id: OrganizationID) -> OrganizationStats:
 
         users = len(self._user_component._organizations[id].users)
 
-        return OrganizationStats(users=users, data_size=data_size, metadata_size=metadata_size)
+        workspaces = len(list(filter(lambda ks: ks[0] == id, self._realm_component._realms.keys())))
```
touilleMan, line 107: more pythonic to use list comprehension:
```python
workspaces = len([realm_id for organization_id, realm_id  in self._realm_component._realms.keys()
 if organization_id == id])
```

touilleMan (changes requested):

touilleMan (approved):

PR #1656, opened 2021-03-08
Http invitation links

vamonte:
After some discussion about this issue, we decided that we don't have to handle the http(s) links in the parsec client.
Indeed, the http(s) url has been build only to redirect the user to the client and avoid the copy/paste action.
In case of the button doesn't work, we have to provide to the users the true parsec url in the mail.
In this PR, I improve the invalid url message to explain the expected format and I add the parsec url in the mail in case of the user need to copy past it in the client.

### DIFF
--- a/parsec/backend/invite.py
+++ b/parsec/backend/invite.py
@@ -151,20 +151,19 @@ def generate_invite_email(
     reply_to: Optional[str],
     greeter_name: Optional[str],  # Noe for device invitation
     organization_id: OrganizationID,
-    invitation_url: str,
-    parsec_url: str,
+    invitation_addr: BackendInvitationAddr,
 ) -> Message:
     html = get_template("invitation_mail.html").render(
         greeter=greeter_name,
         organization_id=organization_id,
-        invitation_url=invitation_url,
-        parsec_url=parsec_url,
+        invitation_url=invitation_addr.to_http_redirection_url(),
+        parsec_url=invitation_addr.to_url(),
     )
     text = get_template("invitation_mail.txt").render(
         greeter=greeter_name,
         organization_id=organization_id,
-        invitation_url=invitation_url,
-        parsec_url=parsec_url,
+        invitation_url=invitation_addr.to_http_redirection_url(),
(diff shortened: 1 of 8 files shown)


parsec/core/gui/tr/parsec_en.po
```diff
@@ -1596,7 +1596,7 @@ msgid "TEXT_JOIN_ORG_INVALID_URL"
 msgstr "The link is invalid."
 
 msgid "TEXT_INVALID_URL"
-msgstr "The link is invalid."
+msgstr "The link is invalid.  (expected format : parsec://hote:port/organisation?action=action&..)"
```
Max-7, line 1599: ":" required a space before and after in French, but only after in English. You did the exact opposite. Could you also remove the dot?

`The link is invalid (expected format: parsec://hote:port/organisation?action=action&..)`

Max-7 (approved):

parsec/core/gui/tr/parsec_en.po
```diff
@@ -1596,7 +1596,7 @@ msgid "TEXT_JOIN_ORG_INVALID_URL"
 msgstr "The link is invalid."
 
 msgid "TEXT_INVALID_URL"
-msgstr "The link is invalid."
+msgstr "The link is invalid (expected format: parsec://hote:port/organisation?action=action&..)"
```
touilleMan, line 1599: `host:port`

parsec/backend/invite.py
```diff
@@ -305,6 +320,7 @@ def _to_http_redirection_url(client_ctx, invitation):
                     reply_to=reply_to,
                     organization_id=client_ctx.organization_id,
                     invitation_url=_to_http_redirection_url(client_ctx, invitation),
+                    parsec_url=_to_url(client_ctx, invitation),
```
touilleMan, line 323: why not provide `invitation_addr` instead of `invitation_url`&`parsec_url` ? (so we could have a single `_to_invitation_addr` function)

touilleMan:
@vamonte replaced by #1637 which allow http:// link to be converted into parsec:// ?

PR #1660, opened 2021-03-16
Fix typos

touilleMan:


parsec/backend/invite.py
```diff
@@ -245,7 +245,7 @@ def __init__(self, event_bus: EventBus, config: BackendConfig):
         # information in database so that we default to no claimer present
         # (which is the most likely when a backend is restarted) .
         #
-        # However there is multiple way this list can go out of sync:
+        # However there is multiple ways this list can go out of sync:
```
vxgmichel, line 248: `There are multiple ways`?

vxgmichel (commented):
LGTM

vxgmichel (approved):

parsec/backend/invite.py
touilleMan: ><'''

PR #1666, opened 2021-03-22
Improve the long request management (peer request)

vamonte:


tests/scripts/debug_websocket.py
```diff
@@ -0,0 +1,120 @@
...
+        device_id=alice_device.device_id,
+        signing_key=alice_device.signing_key,
+    ) as alice_cmds:
+        invitation_addr = await invite_new_user(alice_cmds, alice_device)
+        transport = await connect_as_invited(invitation_addr)
+        async with trio.open_nursery() as nursery:
+            nursery.start_soon(_invite_1_claimer_wait_peer, transport)
+            # nursery.start_soon(_api_ping, transport)
+            nursery.start_soon(_websocket_ping, transport)
+            nursery.start_soon(invite_1_greeter_wait_peer, alice_cmds, invitation_addr._token)
```
touilleMan, line 116: wyi you can use `invitation_addr.token`

parsec/backend/utils.py
```diff
@@ -13,21 +13,36 @@
...
 ):
     def wrapper(fn):
-        assert not hasattr(fn, "_api_info")
-        fn._api_info = {"cmd": cmd, "handshake_types": handshake_types}
-        return fn
+        if long_request:
+
+            @wraps(fn)
+            async def wrapped(self, client_ctx, *args, **kwargs):
+                with trio.move_on_after(PEER_EVENT_MAX_WAIT):
```
touilleMan, line 34: I don't think we should have this timeout here.

This is part of the apiv1invite/claim commands, but hasn't been added to the apiv2 commands (from what I saw in the code there is absolutely no timeout handilng in the backend, the core or the gui regarding apiv2 invite/claim process).

I think if a timeout has to be implemented, we'd better have it on the client side, typically in the GUI (given it has to deal anyway with user getting bored and manually cancelling the invite/claim by himself)

On top of that `events_listen` is another long time command, but should never timeout given we know receiving events is an infrequent operation, otherwise we would spend most of the socket traffic retrying command after a timeout... (note backend keep a buffer of to-send-to-peer events, so that timeout request won't make the client miss an event arriving to the backend at the wrong).

touilleMan: @vxgmichel any idea about this ?

tests/backend/test_ws_events.py
```diff
@@ -0,0 +1,113 @@
...
+                assert isinstance(event, Ping)
+                assert backend_transport1 is backend_transport2
+                client_transport2, event = await next_ws_proto_related_event()
+                assert isinstance(event, Pong)
+                assert client_transport1 is client_transport2
+
+            # send another request in the steam to be sure that will be managed by the backend
+            raw_req = ping_serializer.req_dumps({"cmd": "ping", "ping": "foo"})
+            await transport.send(raw_req)
+            await trio.testing.wait_all_tasks_blocked()
```
touilleMan, line 104: Instead of this, we could move the nursery so that we wait for it coroutine completion (i.e. `_cmd` given it is the only one scheduled) before moving on:
```python
assert testing_stuff
with trio.open_nursery() as nursery:
    nursery.start_soon(_cmd)
    await do_stuff
    assert testing_stuff
    await transport.send(raw_req)  # Now _cmd is going to receive a response
assert events_listen_rep == {...}
```

touilleMan (changes requested):

touilleMan (approved):

PR #1667, opened 2021-03-22
Allow read while reencryption

### LINKED ISSUE #1650: Allow read access during Workspace reencryption
Currently starting a workspace reencryption make the backend refuse any read/write operations to the corresponding realm (so cannot upload new blocks and vlob, cannot read existing blocks and vlob and cannot access or modify the sharing informations)

Blocking the modification operations is needed to prevent users from uploading new data encrypted with the old key. Depending of the encryption implementation it could end up with the new data not reencrypted (so this new data would appear as corrupted given once reencryption is done, Parsec clients only keep the new key) or could prevent the reencryption from being ever done (if new data with the old key keep being added).

On the other hand, read-only access to data during reencryption seems reasonable; especially given the backend keep side by side the original data and it reencrypted version (it's up to the PostgreSQL database administrator to choose when to actually remove the old data to prevent unexpected data loss).

vxgmichel:
Fix #1650.

~Still a WIP, missing~:
- [x] tests
- [x] postresql backend implementation

### DIFF
--- a/mypy.ini
+++ b/mypy.ini
@@ -59,3 +59,12 @@ disallow_untyped_defs=False
 disallow_any_decorated=False
 disallow_any_explicit=False
 
+
+[mypy-parsec.backend.postgresql.*]
+ignore_errors = False
+# disallow_untyped_defs=True
+# disallow_any_unimported=True
+# disallow_any_decorated=True
+# disallow_any_explicit=True
+# disallow_any_generics=True
+# disallow_subclassing_any=True
--- a/mypy.ini
+++ b/mypy.ini
@@ -59,12 +59,11 @@ disallow_untyped_defs=False
 disallow_any_decorated=False
 disallow_any_explicit=False
 
-
 [mypy-parsec.backend.*]
-ignore_errors = False
-# disallow_untyped_defs=True
-# disallow_any_unimported=True
-# disallow_any_decorated=True
-# disallow_any_explicit=True
-# disallow_any_generics=True
-# disallow_subclassing_any=True
+ignore_errors=False
+disallow_untyped_defs=False
+disallow_any_unimported=False
+disallow_any_decorated=False
+disallow_any_explicit=False
+disallow_any_generics=False
+disallow_subclassing_any=False
--- a/mypy.ini
+++ b/mypy.ini
@@ -60,7 +60,7 @@ disallow_any_decorated=False
 disallow_any_explicit=False
 
 
-[mypy-parsec.backend.postgresql.*]
+[mypy-parsec.backend.*]
 ignore_errors = False
 # disallow_untyped_defs=True
 # disallow_any_unimported=True
--- a/parsec/backend/client_context.py
+++ b/parsec/backend/client_context.py
@@ -1,11 +1,13 @@
 # Parsec Cloud (https://parsec.cloud) Copyright (c) AGPLv3 2019 Scille SAS
 
 from uuid import UUID
+from enum import Enum
 from typing import Optional, Union, Tuple, Set, Dict
 
 import trio
 
 from parsec.crypto import VerifyKey, PublicKey
+from parsec.event_bus import EventBusConnectionContext
 from parsec.api.version import ApiVersion
 from parsec.api.transport import Transport
 from parsec.api.data import UserProfile
@@ -43,6 +45,7 @@ class AuthenticatedClientContext(BaseClientContext):
         "organization_id",
         "device_id",
         "human_handle",
+        "device_label",
         "profile",
         "public_key",
         "verify_key",
@@ -61,6 +64,7 @@ class AuthenticatedClientContext(BaseClientContext):
         organization_id: OrganizationID,
         device_id: DeviceID,
         human_handle: Optional[HumanHandle],
+        device_label: Optional[str],
         profile: UserProfile,
         public_key: PublicKey,
         verify_key: VerifyKey,
@@ -70,11 +74,12 @@ class AuthenticatedClientContext(BaseClientContext):
         self.profile = profile
         self.device_id = device_id
         self.human_handle = human_handle
+        self.device_label = device_label
         self.public_key = public_key
         self.verify_key = verify_key
 
-        self.event_bus_ctx = None  # Overwritten in BackendApp.handle_client
-        self.channels = trio.open_memory_channel[Tuple[str, Dict[str, object]]](100)
+        self.event_bus_ctx: EventBusConnectionContext
+        self.channels = trio.open_memory_channel[Tuple[Enum, Dict[str, object]]](100)
         self.realms: Set[UUID] = set()
         self.events_subscribed = False
 
@@ -103,8 +108,7 @@ class AuthenticatedClientContext(BaseClientContext):
 
     @property
     def device_display(self) -> str:
-        # The device label is not known here
-        return str(self.device_id.device_name)
+        return str(self.device_label or self.device_id.device_name)
 
     @property
     def send_events_channel(self):
--- a/parsec/backend/client_context.py
+++ b/parsec/backend/client_context.py
@@ -1,7 +1,9 @@
 # Parsec Cloud (https://parsec.cloud) Copyright (c) AGPLv3 2019 Scille SAS
 
+from uuid import UUID
+from typing import Optional, Union, Tuple, Set, Dict
+
 import trio
-from typing import Optional
 
 from parsec.crypto import VerifyKey, PublicKey
 from parsec.api.version import ApiVersion
@@ -14,6 +16,8 @@ from parsec.api.protocol import (
     DeviceName,
     DeviceID,
     HumanHandle,
+    HandshakeType,
+    APIV1_HandshakeType,
 )
 from parsec.backend.invite import Invitation
 
@@ -30,7 +34,7 @@ class BaseClientContext:
         return self.handshake.backend_api_version
 
     @property
-    def handshake_type(self) -> str:
+    def handshake_type(self) -> Union[HandshakeType, APIV1_HandshakeType]:
         return self.handshake.answer_type
 
 
@@ -70,8 +74,8 @@ class AuthenticatedClientContext(BaseClientContext):
         self.verify_key = verify_key
 
         self.event_bus_ctx = None  # Overwritten in BackendApp.handle_client
-        self.channels = trio.open_memory_channel(100)
-        self.realms = set()
+        self.channels = trio.open_memory_channel[Tuple[str, Dict[str, object]]](100)
+        self.realms: Set[UUID] = set()
         self.events_subscribed = False
 
         self.conn_id = self.transport.conn_id
@@ -99,7 +103,8 @@ class AuthenticatedClientContext(BaseClientContext):
 
     @property
     def device_display(self) -> str:
-        return self.device_label or str(self.device_id.device_name)
+        # The device label is not known here
+        return str(self.device_id.device_name)
 
     @property
     def send_events_channel(self):
--- a/parsec/backend/handshake.py
+++ b/parsec/backend/handshake.py
@@ -127,6 +127,7 @@ async def _do_process_authenticated_answer(
         organization_id=organization_id,
         device_id=device_id,
         human_handle=user.human_handle,
+        device_label=device.device_label,
         profile=user.profile,
         public_key=user.public_key,
         verify_key=device.verify_key,
@@ -182,11 +183,11 @@ async def _process_invited_answer(
         result_req = handshake.build_bad_identity_result_req()
         return None, result_req, _make_error_infos("Bad invitation")
 
-    expected_invitation_type: Type
-    if handshake.answer_data["invitation_type"] == InvitationType.USER:
-        expected_invitation_type = UserInvitation
-    else:  # Device
-        expected_invitation_type = DeviceInvitation
+    expected_invitation_type: Type = (
+        UserInvitation
+        if handshake.answer_data["invitation_type"] == InvitationType.USER
+        else DeviceInvitation
+    )
     if not isinstance(invitation, expected_invitation_type):
         result_req = handshake.build_bad_identity_result_req()
         return None, result_req, _make_error_infos("Bad invitation")
--- a/parsec/backend/handshake.py
+++ b/parsec/backend/handshake.py
@@ -1,10 +1,12 @@
 # Parsec Cloud (https://parsec.cloud) Copyright (c) AGPLv3 2019 Scille SAS
 
-from typing import Tuple, Dict, Optional
 from pendulum import now as pendulum_now
+from typing import Tuple, Dict, Optional, cast, Type
 
 from parsec.api.transport import Transport
 from parsec.api.protocol import (
+    DeviceID,
+    OrganizationID,
     ProtocolError,
     InvitationType,
     HandshakeType,
@@ -87,8 +89,8 @@ async def _do_process_authenticated_answer(
     backend, transport: Transport, handshake: ServerHandshake, handshake_type
 ) -> Tuple[Optional[BaseClientContext], bytes, Optional[Dict]]:
 
-    organization_id = handshake.answer_data["organization_id"]
-    device_id = handshake.answer_data["device_id"]
+    organization_id = cast(OrganizationID, handshake.answer_data["organization_id"])
+    device_id = cast(DeviceID, handshake.answer_data["device_id"])
     expected_rvk = handshake.answer_data["rvk"]
 
     def _make_error_infos(reason):
@@ -136,8 +138,8 @@ async def _do_process_authenticated_answer(
 async def _process_invited_answer(
     backend, transport: Transport, handshake: ServerHandshake
 ) -> Tuple[Optional[BaseClientContext], bytes, Optional[Dict]]:
-    organization_id = handshake.answer_data["organization_id"]
-    invitation_type = handshake.answer_data["invitation_type"]
+    organization_id = cast(OrganizationID, handshake.answer_data["organization_id"])
+    invitation_type = cast(InvitationType, handshake.answer_data["invitation_type"])
     token = handshake.answer_data["token"]
 
     def _make_error_infos(reason):
@@ -180,6 +182,7 @@ async def _process_invited_answer(
         result_req = handshake.build_bad_identity_result_req()
         return None, result_req, _make_error_infos("Bad invitation")
 
+    expected_invitation_type: Type
     if handshake.answer_data["invitation_type"] == InvitationType.USER:
         expected_invitation_type = UserInvitation
     else:  # Device
@@ -206,7 +209,7 @@ async def _apiv1_process_authenticated_answer(
 async def _apiv1_process_anonymous_answer(
     backend, transport: Transport, handshake: ServerHandshake
 ) -> Tuple[Optional[BaseClientContext], bytes, Optional[Dict]]:
-    organization_id = handshake.answer_data["organization_id"]
+    organization_id = cast(OrganizationID, handshake.answer_data["organization_id"])
     expected_rvk = handshake.answer_data["rvk"]
 
     def _make_error_infos(reason):
--- a/parsec/backend/http.py
+++ b/parsec/backend/http.py
@@ -2,7 +2,7 @@
 
 import re
 import attr
-from typing import List, Dict, Optional, Callable, Match, Awaitable
+from typing import List, Dict, Optional
 import mimetypes
 from urllib.parse import parse_qs, urlsplit, urlunsplit, urlencode
 import importlib_resources
@@ -68,27 +68,16 @@ class HTTPResponse:
 class HTTPComponent:
     def __init__(self, config: BackendConfig):
         self._config = config
-        # Map the path pattern to the corresponding handler
-        self.routes: Dict[str, Callable[[HTTPRequest, Match], Awaitable[HTTPResponse]]] = {
-            self._http_root_pattern: self._http_root,
-            self._http_redirect_pattern: self._http_redirect,
-            self._http_static_pattern: self._http_static,
-        }
 
     async def _http_404(self, req: HTTPRequest) -> HTTPResponse:
         data = get_template("404.html").render()
         return HTTPResponse.build_html(404, data=data)
 
-    _http_root_pattern = r"^/?$"
-
-    async def _http_root(self, req: HTTPRequest, match: Match) -> HTTPResponse:
+    async def _http_root(self, req: HTTPRequest) -> HTTPResponse:
         data = get_template("index.html").render()
         return HTTPResponse.build_html(200, data=data)
 
-    _http_redirect_pattern = r"^/redirect/(?P<path>.*)$"
-
-    async def _http_redirect(self, req: HTTPRequest, match: Match) -> HTTPResponse:
-        path = match["path"]
+    async def _http_redirect(self, req: HTTPRequest, *, path: str) -> HTTPResponse:
         if not self._config.backend_addr:
             return HTTPResponse.build(501, data=b"Url redirection is not available")
         backend_addr_split = urlsplit(self._config.backend_addr.to_url())
@@ -105,10 +94,7 @@ class HTTPComponent:
 
         return HTTPResponse.build(302, headers={b"location": location_url.encode("ascii")})
 
-    _http_static_pattern = r"^/static/(?P<path>.*)$"
-
-    async def _http_static(self, req: HTTPRequest, match: Match) -> HTTPResponse:
-        path = match["path"]
+    async def _http_static(self, req: HTTPRequest, *, path: str) -> HTTPResponse:
         if path == "__init__.py":
             return HTTPResponse.build(404)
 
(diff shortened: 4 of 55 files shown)


parsec/core/fs/remote_loader.py
```diff
@@ -417,20 +425,51 @@ async def load_manifest(
...
         if timestamp is not None and version is not None:
             raise FSError(
                 f"Supplied both version {version} and timestamp `{timestamp}` for manifest "
                 f"`{entry_id}`"
             )
         # Download the vlob
-        workspace_entry = self.get_workspace_entry()
+        current_workspace_entry = self.get_workspace_entry()
+        if workspace_entry is None:
+            workspace_entry = self.get_workspace_entry()
```
touilleMan, line 451: Why two call to `get_workspace_entry` ?
Given there is no `await` between the two, both call should return the same entry, but I think it would be better to do
```python
workspace_entry = workspace_entry or current_workspace_entry
```

parsec/core/fs/userfs/userfs.py
```diff
@@ -201,10 +202,15 @@ def __init__(
             role_cached_on=now,
             role=WorkspaceRole.OWNER,
         )
+
+        async def get_previous_entry() -> NoReturn:
```
touilleMan, line 206: `_get_previous_entry` ?

parsec/core/fs/remote_loader.py
```diff
@@ -417,20 +425,51 @@ async def load_manifest(
...
                 entry_id,
                 version=version,
                 timestamp=timestamp if version is None else None,
             )
+        # Special case for loading manifest while in maintenance
+        if (
+            rep["status"] == "in_maintenance"
+            and workspace_entry.encryption_revision == current_workspace_entry.encryption_revision
+        ):
+            previous_workspace_entry = await self.get_previous_workspace_entry()
```
touilleMan, line 464: Shouldn't we provide explicitly the revision number of the workspace entry we want ?
Currently we return the n-1 workspace entry version, which might be `current_workspace_entry` if in the meantime (given we are calling an async function) a reencryption has started !

I'm not sure if this behavior is buggy (it would say we call `_load_manifest` with `current_workspace_entry == previous_entry` so this call try to do the same thing it caller has tried, then call `_load_manifest` another time, this time with a correct `previous_entry` value), but at least it should be documented given how slippy the slope is ^^

parsec/core/fs/remote_loader.py
```diff
@@ -87,12 +87,14 @@ def __init__(
         device: LocalDevice,
         workspace_id: EntryID,
         get_workspace_entry: Callable[[], WorkspaceEntry],
+        get_previous_workspace_entry: Callable[[], Awaitable[WorkspaceEntry]],
```
touilleMan, line 90: We should document the exception this function could raise
Ideally we should have `get_workspace_entry` and `get_previous_workspace_entry` never raise any exception (only returning `None` if the workspace entry cannot be found), so a single comment here would do the trick

However I guess it's hard to have `get_previous_workspace_entry` not raising any exception (because if it call ends up with a `FSBackendOfflineError` and we return a `None`, then we are most likely going to give the wrong information to the user...)

So maybe we should allow `FSError`, but this should be documented, and given documentation of a callback is messy, we should also document the call site that only an `FSError` might be raised.

parsec/core/fs/userfs/userfs.py
```diff
@@ -283,21 +289,26 @@ async def workspace_storage_task(
...
-        # WorskpaeFS to load it each time it is needed
+        # WorkspaceFS to load it each time it is needed
         def get_workspace_entry() -> WorkspaceEntry:
             user_manifest = self.get_user_manifest()
             workspace_entry = user_manifest.get_workspace_entry(workspace_id)
             if not workspace_entry:
                 raise FSWorkspaceNotFoundError(f"Unknown workspace `{workspace_id}`")
             return workspace_entry
 
+        async def get_previous_workspace_entry() -> WorkspaceEntry:
```
touilleMan, line 300: Would be good to add a small docstring explaining we are retrieving previous encryption revision version of the workspace entry

parsec/core/fs/remote_loader.py
```diff
@@ -417,20 +425,51 @@ async def load_manifest(
...
+    async def _load_manifest(
+        self,
+        entry_id: EntryID,
+        version: Optional[int] = None,
+        timestamp: Optional[DateTime] = None,
+        expected_backend_timestamp: Optional[DateTime] = None,
+        workspace_entry: Optional[WorkspaceEntry] = None,
+    ) -> BaseRemoteManifest:
         if timestamp is not None and version is not None:
             raise FSError(
```
touilleMan, line 444: I would have raised a `ValueError` or a `RuntimeError` (or use an assert !) instead given this should never occur and it result from a bug in the code (hence we want it not be catched !)

parsec/core/fs/remote_loader.py
touilleMan: The alternative would be to pass a `encryption_revision` parameter to `get_previous_workspace_entry`.
At first this seems like a cleaner api, and you have the guaranty you won't be spiraling in nested `_load_manifest` callbacks.
On the other hand the "double reencryption" scenario I was talking to is less transparent (given the second call to `_load_manifest` end up with the backend saying we are requesting a n-2 encryption revision which never allowed)

The scenario:
```
encryption_revision == 1
<---- start encryption, encription_revision == 2
call _load_manifest(previous_encription_revision=None)
    --> get_manifest(encription_revision=2)
   <-- ko, revision not available (should use encription_revision==1)
    call _load_manifest(previous_encription_revision=1)
        <---- start encryption, encription_revision == 3
        --> get_manifest(encription_revision=1)
        <-- ko, revision not available (should use encription_revision==2)
        ???
```
At `???` we could choose to do another `_load_manifest(previous_revision=2)` (the behavior of the current code) or to raise an `FSBadEncryptionRevision` exception (the behavior we would most likely have if `get_previous_workspace_entry` would take a revision number)

Another scenario would be:
```
encryption_revision == 1
<---- start reencryption, encription_revision == 2
call _load_manifest(previous_encription_revision=None)
    --> get_manifest(encription_revision=2)
   <-- ko, revision not available (should use encription_revision==1)
    call _load_manifest(previous_encription_revision=1)
        <---- end reencryption
        --> get_manifest(encription_revision=1)
        <-- ko, revision not available (should use encription_revision==2)
        ???
```
Here the current behavior is to raise a `FSBadEncryptionRevision` (given we no longer are in maintenance), so closer to the behavior with `get_previous_workspace_entry` taking a revision number.

I didn't check how this `FSBadEncryptionRevision` is handled, so no definitive opinion on this though ;-)

parsec/backend/memory/vlob.py
```diff
@@ -159,26 +152,48 @@ def _check_realm_access(
...
+        if expected_maintenance and not realm.status.in_maintenance:
+            raise VlobNotInMaintenanceError(f"Realm `{realm_id}` not under maintenance")
+
+        # Encryption revision checking
+        if encryption_revision is not None:
+
+            # Special case of reading while in maintenance
+            if read_only and realm.status.in_maintenance:
+
+                # The vlob is not available yet for the current revision
```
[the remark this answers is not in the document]
vxgmichel: Except it's not only `read`, it's also `poll_changes` (and potentially `list_versions`).

See 5769bda for an improved version.

parsec/core/fs/remote_loader.py
vxgmichel: 5769bda

parsec/core/fs/userfs/userfs.py
vxgmichel: 5769bda

parsec/core/fs/remote_loader.py
vxgmichel: > Currently we return the n-1 workspace entry version, which might be `current_workspace_entry` if in the meantime (given we are calling an async function) a reencryption has started !

...which might be the PREVIOUS `current_workspace_entry`, but the next call to `get_workspace_entry()` should return a new `current_workspace_entry` which has an encryption revision strictly greater than `previous_workspace_entry`. 

I'll add the following assert to make sure we don't fall in an infinite loop because of some other bug:
```python
assert previous_workspace_entry.encryption_revision < self.get_workspace_entry().encryption_revision
```

vxgmichel: 5769bda

parsec/core/fs/remote_loader.py
vxgmichel: 5769bda

parsec/core/fs/userfs/userfs.py
vxgmichel: 5769bda

parsec/core/fs/remote_loader.py
vxgmichel: 5769bda

parsec/core/fs/remote_loader.py
```diff
@@ -417,20 +426,50 @@ async def load_manifest(
...
-                f"Supplied both version {version} and timestamp `{timestamp}` for manifest "
-                f"`{entry_id}`"
-            )
+        assert (
+            timestamp is None or version is None
+        ), "Either timestamp or version argument should be provided"
+        # Get the current and requested workspace entry
+        # They're usually the same, except when loading from a workspace while it's in maintenance
+        current_workspace_entry = self.get_workspace_entry()
+        workspace_entry = current_workspace_entry if workspace_entry is None else workspace_entry
```
touilleMan, line 435: workspace_entry is an object or `None`, why don't use the simpler `or` construct instead ?
```python
workspace_entry = workspace_entry or current_workspace_entry
```

touilleMan (changes requested):

parsec/backend/memory/block.py
```diff
@@ -35,28 +35,26 @@ def register_components(
...
+        allowed_roles = (RealmRole.OWNER, RealmRole.MANAGER, RealmRole.CONTRIBUTOR)
+        if read_only:
+            allowed_roles += (RealmRole.READER,)
+
         if realm.roles.get(user_id) not in allowed_roles:
             raise BlockAccessError()
 
-        if realm.status.in_maintenance:
+        # Reading blocks is allowed while reencrypting the workspace
+        if realm.status.in_reencryption and not read_only:
```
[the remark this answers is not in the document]
vxgmichel: Right! it's been correctly implemented in the postgresql version though

parsec/core/fs/remote_loader.py
vxgmichel: I personally find it more explicit :)

parsec/backend/realm.py
```diff
@@ -79,6 +79,14 @@ class RealmStatus:
     def in_maintenance(self) -> bool:
         return bool(self.maintenance_type)
 
+    @property
```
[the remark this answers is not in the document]
vxgmichel: Or maybe we could make `get_realm_status` return a `RealmStatus` in postgresql? :innocent:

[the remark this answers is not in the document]
touilleMan: image

tests/core/mountpoint/test_file_failures.py
```diff
@@ -93,11 +95,19 @@ async def test_remote_error_event(
 
         trio_w = trio.Path(mountpoint_manager.get_path_in_mountpoint(wid, FsPath("/")))
 
-        # Switch the mountpoint in maintenance...
-        await bob_user_fs.workspace_start_reencryption(wid)
-
         def _testbed():
-            # ...accessing workspace data in the backend should endup in remote error
+
+            # Simulate network failure
```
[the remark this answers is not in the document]
vxgmichel: c81a5ff18844a6a31649bd01de4c1271cd932df6

parsec/backend/memory/block.py
```diff
@@ -35,28 +36,29 @@ def register_components(
...
+        self._check_realm_access(organization_id, realm_id, user_id, OperationKind.DATA_WRITE)
 
-    def _check_realm_access(self, organization_id, realm_id, user_id, allowed_roles):
+    def _check_realm_access(self, organization_id, realm_id, user_id, operation_kind):
         try:
             realm = self._realm_component._get_realm(organization_id, realm_id)
         except RealmNotFoundError:
             raise BlockNotFoundError(f"Realm `{realm_id}` doesn't exist")
 
+        allowed_roles = (RealmRole.OWNER, RealmRole.MANAGER, RealmRole.CONTRIBUTOR)
```
touilleMan, line 50: I would replace this by multiple if
```python
if operation_kind == OperationKind.DATA_READ:
    allowed_roles = (...)
elif operation_kind == OperationKind.DATA_WRITE:
    allowed_roles = (...)
elif operation_kind == OperationKind.MAINTENANCE:
    allowed_roles = (...)
else:
    allowed_roles = ()
```
This avoid the combination trick which grow increasingly unclear with the number of possible operation_kind

parsec/backend/memory/factory.py
```diff
@@ -22,7 +25,9 @@
 
 @asynccontextmanager
 async def components_factory(config: BackendConfig, event_bus: EventBus):
-    (send_events_channel, receive_events_channel) = trio.open_memory_channel(math.inf)
+    (send_events_channel, receive_events_channel) = trio.open_memory_channel[
```
touilleMan, line 28: the parenthesis around `(send_events_channel, receive_events_channel) =` was probably added by formatting, I think we should remove it to save a bit more space

parsec/backend/memory/factory.py
```diff
@@ -59,10 +64,12 @@ async def _dispatch_event():
...
     }
-    for component in (organization, user, invite, message, realm, vlob, ping, block):
-        component.register_components(**components)
+    for component in components.values():
+        method = getattr(component, "register_components", None)
+        if method is not None:
+            method(**components)
 
-    async with trio.open_service_nursery() as nursery:
+    async with open_service_nursery() as nursery:
```
touilleMan, line 72: why remove the `trio.` ? I'm not against it, but it seems to be the way we access `open_service_nursery` in the rest of the codebase

parsec/backend/memory/factory.py
```diff
@@ -22,7 +25,9 @@
 
 @asynccontextmanager
 async def components_factory(config: BackendConfig, event_bus: EventBus):
-    (send_events_channel, receive_events_channel) = trio.open_memory_channel(math.inf)
+    (send_events_channel, receive_events_channel) = trio.open_memory_channel[
+        Tuple[str, Dict[str, object]]
```
touilleMan, line 29: I have the feeling something's broken here: we use consider event have `str` here (and in `_send_event/_dispatch_event`) this is probably because it must be serialized to be send over postgresql listen/notify system. However the event bus is supposed to receive an Enum.

My guess is we should be bitten by this once this PR and #1680 are merged

parsec/backend/webhooks.py
```diff
@@ -3,7 +3,8 @@
 import trio
 from typing import Optional
 from structlog import get_logger
-from urllib.request import urlopen, Request, URLError
+from urllib.error import URLError
```
touilleMan, line 6: Is there a reason for this ?

parsec/backend/user.py
```diff
@@ -151,10 +151,6 @@ def __repr__(self):
     creator: DeviceID
     created_on: pendulum.DateTime = attr.ib(factory=pendulum.now)
 
-    @property
-    def organization_id(self) -> OrganizationID:
-        return self.user_id.organization_id
```
touilleMan, line 156: This was totally broken, right ?

parsec/backend/memory/realm.py
```diff
@@ -114,22 +115,23 @@ async def get_stats(
         realm = self._get_realm(organization_id, realm_id)
         if author.user_id not in realm.roles:
             raise RealmAccessError()
-        RealmStats.blocks_size = 0
```
touilleMan, line 117: Ooooups ^^

parsec/backend/memory/block.py
touilleMan: on top of that, this is what is done in `memory/vlob.py` ^^

parsec/backend/memory/vlob.py
```diff
@@ -137,60 +138,87 @@ def _get_vlob(self, organization_id, vlob_id):
...
+        # All roles can do read-only operation
+        elif operation_kind == OperationKind.DATA_READ:
+            allowed_roles = (
+                RealmRole.OWNER,
+                RealmRole.MANAGER,
+                RealmRole.CONTRIBUTOR,
+                RealmRole.READER,
+            )
+        # All roles except reader can do write operation
+        else:
```
touilleMan, line 170: would be better to have an explicit `elif` on `OperationKind.DATA_WRITE`, then a final `else` with no allowed roles

This way the code is safe if we add new operation kind

parsec/backend/postgresql/realm_queries/get.py
```diff
@@ -135,20 +135,21 @@ async def query_get_stats(
...
-    RealmStats.vlobs_size = 0
-    if "sum" in blocks_size[0] and blocks_size[0]["sum"] is not None:
-        RealmStats.blocks_size = blocks_size[0]["sum"]
-    if "sum" in vlobs_size[0] and vlobs_size[0]["sum"] is not None:
-        RealmStats.vlobs_size = vlobs_size[0]["sum"]
 
-    return RealmStats
+    blocks_size = 0
+    vlobs_size = 0
+    if "sum" in blocks_size_rep[0] and blocks_size_rep[0]["sum"] is not None:
```
touilleMan, line 147: given we only use the first row, why not do `conn.fetchrow` instead of `blocks_size_rep[0]` ?

touilleMan: on top of that, if a row is present, it must contain the `sum` field so not need to check for it presence

parsec/backend/postgresql/user_queries/get.py
```diff
@@ -304,9 +303,9 @@ async def query_get_user_with_devices_and_trustchain(
             d.redacted_device_certificate if redacted else d.device_certificate
             for d in user_devices
         ],
-        trustchain_device_certificates=trustchain.devices,
-        trustchain_user_certificates=trustchain.users,
-        trustchain_revoked_user_certificates=trustchain.revoked_users,
+        trustchain_device_certificates=list(trustchain.devices),
```
touilleMan, line 306: I think it should be `GetUserAndDevicesResult`'s definition that should be corrected instead to use tuples instead of lists, we love immutability around here ;-)

parsec/core/fs/userfs/userfs.py
```diff
@@ -201,10 +202,15 @@ def __init__(
             role_cached_on=now,
             role=WorkspaceRole.OWNER,
         )
+
+        async def _get_previous_entry() -> NoReturn:
```
touilleMan, line 206: I wouldn't use a `NoReturn` here given UserRemoteLoader's `get_previous_entry` is supposed to return the entry info, it's just * this * implementation that is not functional

If we want to sugnify that `get_previous_entry` is never going to be called for this instance (because userfs is never going to be reencrypted), then we should raise a `RuntimeError` or `AssertionError` instead of the `NotImplementedError`

touilleMan (changes requested):

parsec/backend/client_context.py
```diff
@@ -1,7 +1,9 @@
 # Parsec Cloud (https://parsec.cloud) Copyright (c) AGPLv3 2019 Scille SAS
 
+from uuid import UUID
+from typing import Optional, Union, Tuple, Set, Dict
+
```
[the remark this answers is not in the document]
vxgmichel: Let's wait for [isort](https://pypi.org/project/isort/) integration.

parsec/backend/client_context.py
```diff
@@ -70,8 +74,8 @@ def __init__(
         self.verify_key = verify_key
 
         self.event_bus_ctx = None  # Overwritten in BackendApp.handle_client
-        self.channels = trio.open_memory_channel(100)
-        self.realms = set()
+        self.channels = trio.open_memory_channel[Tuple[str, Dict[str, object]]](100)
```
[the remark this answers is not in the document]
vxgmichel: Yep: https://github.com/python-trio/trio-typing/issues/6#issuecomment-485056719

parsec/backend/http.py
```diff
@@ -111,23 +111,25 @@ async def _http_static(self, req: HTTPRequest, path: str) -> HTTPResponse:
...
-        for path_pattern, route in self.ROUTE_MAPPING:
-            match = re.match(path_pattern, req.path)
-            if match:
-                route_args = match.groupdict()
-                break
-        else:
-            route = HTTPComponent._http_404
-            route_args = {}
+        # Root GET request
+        if re.match(r"^/?$", req.path):
```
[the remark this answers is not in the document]
vxgmichel: Yea, except different routes have a different capture scheme with different arguments, so there has to be some standardization before we we can use a mapping to dispatch the handlers.

parsec/backend/memory/factory.py
vxgmichel: > to save a bit more space

"Y'a pas de petits profits"

parsec/backend/memory/factory.py
vxgmichel: Because it's added to the trio namespace with monkey patching, and trio doesn't like it.

> but it seems to be the way we access open_service_nursery in the rest of the codebase

Not in the `parsec.core.fs` module where mypy is enabled :)

parsec/backend/webhooks.py
vxgmichel: That's [where it lives](https://docs.python.org/3.8/library/urllib.error.html#urllib.error.URLError).

Also, mypy.

parsec/backend/user.py
vxgmichel: Yup

parsec/backend/memory/realm.py
vxgmichel: That was a fun one ^^

parsec/backend/postgresql/realm_queries/get.py
vxgmichel: I did not want to change the logic while messing with the typing.

Should I look into it?

mypy.ini
```diff
@@ -50,3 +50,12 @@ disallow_untyped_defs=False
 disallow_any_decorated=False
 disallow_any_explicit=False
 
+
+[mypy-parsec.backend.*]
+ignore_errors = False
+# disallow_untyped_defs=True
```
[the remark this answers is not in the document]
vxgmichel: 9ddb38c5be108a3fc89d6ff4ab2404d92401c0ed

parsec/backend/client_context.py
```diff
@@ -70,8 +74,8 @@ def __init__(
         self.verify_key = verify_key
 
         self.event_bus_ctx = None  # Overwritten in BackendApp.handle_client
```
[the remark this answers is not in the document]
vxgmichel: 9ddb38c5be108a3fc89d6ff4ab2404d92401c0ed

parsec/backend/client_context.py
```diff
@@ -99,7 +103,8 @@ def user_display(self) -> str:
 
     @property
     def device_display(self) -> str:
-        return self.device_label or str(self.device_id.device_name)
+        # The device label is not known here
+        return str(self.device_id.device_name)
```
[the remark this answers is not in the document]
vxgmichel: 9ddb38c5be108a3fc89d6ff4ab2404d92401c0ed

parsec/backend/handshake.py
```diff
@@ -87,8 +89,8 @@ async def _do_process_authenticated_answer(
     backend, transport: Transport, handshake: ServerHandshake, handshake_type
 ) -> Tuple[Optional[BaseClientContext], bytes, Optional[Dict]]:
 
-    organization_id = handshake.answer_data["organization_id"]
-    device_id = handshake.answer_data["device_id"]
+    organization_id = cast(OrganizationID, handshake.answer_data["organization_id"])
```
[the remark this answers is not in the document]
vxgmichel: 9ddb38c5be108a3fc89d6ff4ab2404d92401c0ed

parsec/backend/handshake.py
```diff
@@ -180,6 +182,7 @@ def _make_error_infos(reason):
         result_req = handshake.build_bad_identity_result_req()
         return None, result_req, _make_error_infos("Bad invitation")
 
+    expected_invitation_type: Type
```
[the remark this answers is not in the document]
vxgmichel: 9ddb38c5be108a3fc89d6ff4ab2404d92401c0ed

parsec/backend/memory/block.py
vxgmichel: 9ddb38c5be108a3fc89d6ff4ab2404d92401c0ed

parsec/backend/memory/factory.py
vxgmichel: 9ddb38c5be108a3fc89d6ff4ab2404d92401c0ed

parsec/backend/memory/factory.py
vxgmichel: 9ddb38c5be108a3fc89d6ff4ab2404d92401c0ed

parsec/backend/memory/vlob.py
vxgmichel: 9ddb38c5be108a3fc89d6ff4ab2404d92401c0ed

parsec/backend/postgresql/user_queries/get.py
vxgmichel: 9ddb38c5be108a3fc89d6ff4ab2404d92401c0ed

parsec/core/fs/userfs/userfs.py
vxgmichel: 9ddb38c5be108a3fc89d6ff4ab2404d92401c0ed

parsec/backend/http.py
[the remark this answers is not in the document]
vxgmichel: 11d796eccd0b853c6b597c62a153d149590a8ee0

parsec/backend/postgresql/realm_queries/get.py
vxgmichel: 11d796eccd0b853c6b597c62a153d149590a8ee0

touilleMan (approved):
image

PR #1672, opened 2021-03-30
Bump pyqt to 5.15.2

Ironicbay:


packaging/macOS/launch_script.py
```diff
@@ -2,20 +2,14 @@
 
 import sys
 import os
-import locale
 
 from parsec.cli import cli
+from click import core
 
+# This prevents click from raising encoding errors at startup
+core._verify_python3_env = lambda: None
```
touilleMan, line 10: It would be better to wrap the original `_verify_python3_env` so that we could log an error when it raises an exception instead of silence it. This would allow us to have more insight of when those errors occurs (OS version, locales used etc.)

However the crash occurs before we initialize sentry logging, so it would be a mess to catch the exception, save it on the side, then retrieve and log it after sentry initialization...
On top of that, we don't really care about unicode in click here (we are expecting to receive no arguments from argv or a single url which is should not contain unicode, similarly we output ascii logs to stdout but nobody read them anyway when starting the application this way)

So I think it's fine to just mock the way you did, but you may want to add my comment to the code in order to explicit why we consider this ok ;-)

setup.py
```diff
@@ -318,7 +318,7 @@ def run(self):
 ]
 
 
-PYQT_DEPS = ["PyQt5==5.14.2", "pyqt5-sip==12.8.0"]
+PYQT_DEPS = ["PyQt5==5.15.2", "pyqt5-sip==12.8.0"]
```
touilleMan, line 321: no need to update pyqt5-sip too ?

touilleMan (approved):

PR #1673, opened 2021-03-30
Add BackendAddr.__hash__

touilleMan:


vxgmichel (approved):
When is a backend address hashed exactly?

LGTM otherwise!

touilleMan:
> When is a backend address hashed exactly?

This is currently needed in a 3rd party project using parsec-cloud as foundation ;-)
Anyway, it's a good thing to have __eq__ and __hash__ for address given they are basically wrappers over strings

PR #1674, opened 2021-03-30
Add params to setup human/device label/email from the cli in `parsec …

touilleMan:
…core bootstrap_organization`

### DIFF
--- a/parsec/core/cli/bootstrap_organization.py
+++ b/parsec/core/cli/bootstrap_organization.py
@@ -13,11 +13,14 @@ from parsec.core.invite import bootstrap_organization as do_bootstrap_organizati
 from parsec.core.cli.utils import core_config_options
 
 
-async def _bootstrap_organization(config, addr, password, force):
-    label = await aprompt("User fullname")
-    email = await aprompt("User email")
-    human_handle = HumanHandle(email=email, label=label)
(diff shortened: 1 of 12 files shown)


parsec/core/cli/bootstrap_organization.py
```diff
@@ -26,21 +29,30 @@ async def _bootstrap_organization(config, addr, password, force):
             )
 
         device_display = click.style(new_device.slughash, fg="yellow")
+        # We don't have to worry about overwritting an existing keyfile
+        # given their names are base on the device's slughash which is intended
+        # to be globally unique.
         with operation(f"Saving device {device_display}"):
             save_device_with_password(
-                config_dir=config.config_dir, device=new_device, password=password, force=force
+                config_dir=config.config_dir, device=new_device, password=password
```
vxgmichel, line 37: > We don't have to worry about overwritting an existing keyfile
> given their names are base on the device's slughash which is intended
> to be globally unique.

Maybe the `force` argument should also be removed from `save_device_with_password` then?

vxgmichel (approved):
LGTM!

parsec/core/cli/bootstrap_organization.py
touilleMan: 👍 8a666202d3c55cf13ca241a450cc22f182ecbad4

PR #1676, opened 2021-04-01
Bump winfspy version to 0.8.2

vxgmichel:
And winfsp version to 1.8.20304.

### DIFF
--- a/packaging/win32/installer.nsi
+++ b/packaging/win32/installer.nsi
@@ -33,7 +33,7 @@
 !define LICENSE_FILEPATH "${PARSEC_FREEZE_BUILD_DIR}\LICENSE.txt"
 !define INSTALLER_FILENAME "parsec-${PROGRAM_VERSION}-${PROGRAM_PLATFORM}-setup.exe"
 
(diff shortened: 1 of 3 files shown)


packaging/win32/installer.nsi
```diff
@@ -33,7 +33,7 @@
 !define LICENSE_FILEPATH "${PARSEC_FREEZE_BUILD_DIR}\LICENSE.txt"
 !define INSTALLER_FILENAME "parsec-${PROGRAM_VERSION}-${PROGRAM_PLATFORM}-setup.exe"
 
-!define WINFSP_INSTALLER "winfsp-1.7.20172.msi"
+!define WINFSP_INSTALLER "winfsp-1.8.20304.msi"
```
touilleMan, line 36: so WinFSP 1.8 will only be provided to new users (given we test winfsp presence between 1.4 and 2.0), I guess it's ok

touilleMan (approved):

packaging/win32/installer.nsi
vxgmichel: Yes, until we implement a better update workflow for both winfsp and parsec :)

PR #1680, opened 2021-04-07
Improve event_bus typing

touilleMan:


C1em (approved):

parsec/event_bus.py
```diff
@@ -15,40 +33,54 @@ class MetaEvent(Enum):
...
         if self._filter and not self._filter(event, **kwargs):
             return
         self._event_result = (event, kwargs)
         self._event_occured.set()
 
-    async def wait(self):
+    async def wait(self) -> Tuple[Enum, Dict[str, Any]]:
         await self._event_occured.wait()
-        return self._event_result
+        return self._event_result  # type: ignore
```
vxgmichel, line 66: Use `assert self._event_result is not None` instead to tell mypy that the `None` type should not be considered from this point on.

parsec/event_bus.py
```diff
@@ -15,40 +33,54 @@ class MetaEvent(Enum):
     EVENT_DISCONNECTED = "event.disconnected"
 
 
+CustomEvent = TypeVar("CustomEvent", bound=Enum)
+AllEvent = Union[MetaEvent, CustomEvent]
+
+
+class EventCallback(Protocol):
+    def __call__(self, event: Enum, **kwargs: Any) -> None:
```
vxgmichel, line 41: One alternative to `Any` is to use `object`. It guarantees that those objects are not used for anything, and simply passed around. This way, it's up to the event handlers to cast the values they get from event bus according to a given "protocol" (e.g. keyword argument "x" of event "y" is supposed to be of type "z"). 

This would allow mypy to catch more errors since casting has to be added explicitly by the devs, instead of the implicitness of `Any`.

parsec/event_bus.py
```diff
@@ -15,40 +33,54 @@ class MetaEvent(Enum):
...
+        ...
+
+
 class EventWaiter:
-    def __init__(self, filter):
+    def __init__(self, filter: Optional[EventFilterCallback]):
         self._filter = filter
         self._event_occured = trio.Event()
-        self._event_result = None
+        self._event_result: Optional[Tuple[Enum, Dict[str, Any]]] = None
```
vxgmichel, line 54: Maybe this attribute should be splited as:
```python
self._event_result_type: Optional[Enum] = None
self._event_result_values: Dict[str, Any] = {}
```

Also `Enum` is a bit generic.. Is there a base class for event types?

vxgmichel (approved):
A few comments, typing looks good otherwise :+1:

parsec/event_bus.py
touilleMan: > Also Enum is a bit generic.. Is there a base class for event types?

Good catch, it should be an `AllEvent` !

> Maybe this attribute should be splited as:

Not sure: the non-splitted form make it clear the event is present or not, splitting it seems only better because it makes the typing information smaller, which doesn't seems like a very compelling argument to me ;-)

vxgmichel: > smaller

I'd say "more readable" but alright :)

parsec/event_bus.py
touilleMan: 👍

parsec/event_bus.py
touilleMan: I didn't know that, seems pretty good to me ;-(

touilleMan: cf7d9408ba8492afd17f56a7de61931e5780c0c1

parsec/event_bus.py
touilleMan: cf7d9408ba8492afd17f56a7de61931e5780c0c1

PR #1685, opened 2021-04-09
Normalize filenames

Ironicbay:


vxgmichel:
The failing test has nothing to do with your PR, it's might be related to PR #1691.

I'm investigating this.

parsec/api/data/entry.py
```diff
@@ -37,8 +37,12 @@ def new(cls: Type[EntryIDTypeVar]) -> EntryIDTypeVar:
 class EntryName(str):
     __slots__ = ()
 
-    def __init__(self, raw: str):
+    def __new__(cls, raw: str) -> "EntryName":
+        import unicodedata
```
touilleMan, line 41: import should be on top of the file (import is costly, even when the module has already been imported)

```
$ python -m timeit -s "pass" "import unicodedata"
10000000 loops, best of 3: 0.0796 usec per loop
$ python -m timeit -s "import unicodedata" "pass"
100000000 loops, best of 3: 0.00513 usec per loop
```

tests/api/test_types.py
```diff
@@ -125,3 +125,16 @@ def test_sas_code():
     for invalid in ["", "AAA", "AAAAA", "aaaa", "AAAI", "AAAO", "AAA0", "AAA1"]:
         with pytest.raises(ValueError):
             SASCode(invalid)
+
+
+def test_entryname_normalization():
+    nfc_str = "àáâäæãåāçćčèéêëēėęîïíīįìłñńôöòóœøōõßśšûüùúūÿžźż"
+    nfd_str = (
```
touilleMan, line 132: Funny to see black doesn't format the two lines the same way even though there look the same length 😄 

I guess it can be explained by:
```python
>>> len("àáâäæãåāçćčèéêëēėęîïíīįìłñńôöòóœøōõßśšûüùúūÿžźż")
47
>>> len("àáâäæãåāçćčèéêëēėęîïíīįìłñńôöòóœøōõßśšûüùúūÿžźż")
89
```

tests/api/test_types.py
```diff
@@ -125,3 +125,16 @@ def test_sas_code():
...
+        "àáâäæãåāçćčèéêëēėęîïíīįìłñńôöòóœøōõßśšûüùúūÿžźż"
+    )
+
+    for i in range(len(nfc_str)):
+        assert nfc_str[i] != nfd_str[i]
```
touilleMan, line 137: Well I guess my previous joking comment was of use in the end ^^
so we cannot access item by index this way for the nfd:
```python
>>> nfd_str = "àáâäæãåāçćčèéêëēėęîïíīįìłñńôöòóœøōõßśšûüùúūÿžźż"
>>> nfc_str = "àáâäæãåāçćčèéêëēėęîïíīįìłñńôöòóœøōõßśšûüùúūÿžźż"
>>> len(nfd_str)
89
>>> len(nfc_str)
47
>>> nfc_str[0]
'à'
>>> nfd_str[0]
'a'
>>> nfd_str[1]
'̀'
>>> nfd_str[88]
'̇'
>>> nfd_str[87]
'z'
>>> nfc_str[46]
'ż'
```

And we cannot use a index trick (i.g. `nfc_str[i] != nfd_str[2 *i] + nfd_str[2 * i + 1]`) given a) `2 * len(nfc_str) != len(nfd_str)` and b) there is no guarantee a nfd character is composed of only two codepoints

I think we should just do `assert nfd_str != nfc_str` (just to be sure an auto-formatter doesn't come and do nfc normalization on this source file 😄 )

touilleMan (changes requested):

PR #1694, opened 2021-04-19
Increase timeout for running_backend fixture

vxgmichel:
Tested with `tests/backend/test_organization_stats.py::test_organization_stats`.

Measurements of lowest non-failing timeout on my machine:
- memory backend: 65-70 ms
- postgresql 10 backend: 220-230 ms
- postgresql 12 backend: 680-700 ms

Reproduce with:
```bash
while PGINSTALLATION=/usr/lib/postgresql/10/bin pytest --postgresql tests/backend/test_organization_stats.py; do :; done;
```


touilleMan (approved):
LDNGTM (looking definitly not good to me, but we need a hotfix for the other PRs anyway 😞 )

PR #1695, opened 2021-04-20
Run the upload blocks operation in one coroutine pool

vamonte:


tests/core/fs/workspacefs/test_workspace_fs.py
```diff
@@ -438,3 +441,77 @@ async def mockup(*args):
     with running_backend.offline():
         with pytest.raises(FSBackendOfflineError):
             await alice_workspace.get_reencryption_need()
+
+
+@pytest.mark.trio
+async def test_upload_blocks(alice_user_fs, alice2_user_fs, running_backend, monkeypatch):
+    wid = await create_shared_workspace("w", alice_user_fs, alice2_user_fs)
```
touilleMan, line 448: no need for this heavy helper function here given we don't have to share the workspace with another user (given alice and alice2 are two devices of the same user)
So we can do instead:
```python
wid = await alice_user_fs.workspace_create("w")
await alice_user_fs.sync()
await alice2_user_fs.sync()
```

tests/core/fs/workspacefs/test_workspace_fs.py
```diff
@@ -438,3 +441,77 @@ async def mockup(*args):
...
+
+    await alice_workspace.write_bytes(fspath, b"aaaaaa" * DEFAULT_BLOCK_SIZE)
+
+    info = await alice_workspace.path_info(fspath)
+
+    assert info["base_version"] == 1
+    assert info["is_placeholder"] is False
+    assert info["need_sync"] is True
+
+    assert await alice2_workspace.read_bytes(fspath) == bytearray()
```
touilleMan, line 466: I don't think this line is needed: the fact that `alice2_workspace` is isolated from `alice_workspace` and requires an explicit sync call is used in multiple other tests so there is no reason we should doubt on it here

tests/core/fs/workspacefs/test_workspace_fs.py
```diff
@@ -438,3 +441,77 @@ async def mockup(*args):
     with running_backend.offline():
         with pytest.raises(FSBackendOfflineError):
             await alice_workspace.get_reencryption_need()
+
+
+@pytest.mark.trio
+async def test_upload_blocks(alice_user_fs, alice2_user_fs, running_backend, monkeypatch):
```
touilleMan, line 447: `test_backend_block_upload_error_during_sync` ?

tests/core/fs/workspacefs/test_workspace_fs.py
```diff
@@ -438,3 +441,77 @@ async def mockup(*args):
...
+    wid = await create_shared_workspace("w", alice_user_fs, alice2_user_fs)
+    alice_workspace = alice_user_fs.get_workspace(wid)
+
+    alice2_workspace = alice2_user_fs.get_workspace(wid)
+
+    fspath = "/taz"
+
+    await alice_workspace.touch(fspath)
+    await alice_workspace.sync()
```
touilleMan, line 456: Should add this comment ;-)
```python
# Placeholder sync always starts by a "minimal sync" (i.e. an empty manifest is synchronized so that
# parent folder don't have to wait for the full sync before syncing itself).
# Here we force the minimal sync so to avoid this corner-case in the rest of the test.
```

tests/core/fs/workspacefs/test_workspace_fs.py
```diff
@@ -438,3 +441,77 @@ async def mockup(*args):
...
+
+    await alice_workspace.write_bytes(fspath, b"aaaaaa" * DEFAULT_BLOCK_SIZE)
```
touilleMan, line 458: `b"aaaaaa" * DEFAULT_BLOCK_SIZE` look strange: it is equivalent to `b"a" * 6 * DEFAULT_BLOCK_SIZE`, but `b"aaaaaa"` looks like it is made of an arbitrary number of characters.

What do you think of doing something like this instead ?
```python
# Fill the file with multiple blocks worth of data so that parallel upload will occurs
TAZ_V2_BLOCKS = 6
await alice_workspace.write_bytes(fspath, b"a" * DEFAULT_BLOCK_SIZE * TAZ_V2_BLOCKS)
```

tests/core/fs/workspacefs/test_workspace_fs.py
```diff
@@ -438,3 +441,77 @@ async def mockup(*args):
...
+    info = await alice_workspace.path_info(fspath)
+
+    assert info["base_version"] == 1
+    assert info["is_placeholder"] is False
+    assert info["need_sync"] is True
+
+    assert await alice2_workspace.read_bytes(fspath) == bytearray()
+
+    # mock the synchronization_step method to retrieve the new manifest
+    vanilla_transaction_synchronization_step = alice_workspace.transactions.synchronization_step
```
touilleMan, line 469: Can't we just use a block counter in the mock_bcakend_block_create to know when to raise an error without having to also mock the synchronization step ?

```python
blocks_create_before_crash = TAR_V2_BLOCKS // 2  # Crash in the middle of the upload

async def mock_backend_block_create(*args, **kwargs):
    nonlocal blocks_create_before_crash
    if blocks_create_before_crash == 0:
        raise BlockNotFoundError()
    blocks_create_before_crash -= 1
    return await vanilla_backend_block_create(*args, **kwargs)

monkeypatch.setattr(running_backend.backend.block, "create", mock_backend_block_create)
```

tests/core/fs/workspacefs/test_workspace_fs.py
```diff
@@ -438,3 +441,77 @@ async def mockup(*args):
...
+
+    info = await alice_workspace.path_info(fspath)
+
+    assert info["base_version"] == 1
+    assert info["is_placeholder"] is False
+    assert info["need_sync"] is True
+
+    monkeypatch.setattr(running_backend.backend.block, "create", vanilla_backend_block_create)
+
+    await alice2_workspace.sync()
```
touilleMan, line 506:
```python
# Given sync has failed, alice2 should still see file in v1
```
Besides I think you should keep this test before the unmock (checking alice2 view is part of the checks done on the view of alice)
And you should use path_info instead of read_bytes (given it gives you the size attribute that sould be enough:

```python
with pytest.raises(FSError):
    await alice_workspace.sync()

# File sync has failed, so it info should be the same
info = await alice_workspace.path_info(fspath)
assert info["base_version"] == 1
assert info["need_sync"] is True
# Other device doesn't see the modifications that have failed to be synced
await alice2_workspace.sync()
info = await alice2_workspace.path_info(fspath)
assert info["base_version"] == 1
assert info["need_sync"] is False
assert info["size"] == 0
```
(note I remove the check on `is_placeholder` to be consistent with the "avoid this corner-case in the rest of the test" comment I made above 😄 )

touilleMan (approved):

tests/core/fs/workspacefs/test_workspace_fs.py
vamonte: I used a similar logic in the first version of my test, but I removed it because with a counter, if we had several files in the workspace, we are not sure on wich file the error will be raised.

vamonte: Not a probleme, in this new version, we use a new workspace.

PR #1696, opened 2021-04-20
Replace Pendulum by a custom DateTime class (mainly to no longer have…

touilleMan:
… the huge pytzdata dependency)

@vxgmichel you need to review this :trollface:

### DIFF
--- a/parsec/api/data/base.py
+++ b/parsec/api/data/base.py
@@ -2,7 +2,7 @@
 
 import attr
 from typing import Optional, Any, TypeVar, Type
-from pendulum import DateTime
+from parsec.datetime import DateTime
 
 from parsec.serde import (
     BaseSchema,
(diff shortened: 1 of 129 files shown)


touilleMan:
Size of win32 package without/with this patch:
```
$ find build-master/parsec-v2.2.4+dev-win64/ | wc -l
931
$ find build/parsec-v2.2.4+dev-win64/ | wc -l
154
$ du -sh build-master/parsec-v2.2.4+dev-win64/
96M     build-master/parsec-v2.2.4+dev-win64/
$ du -sh build/parsec-v2.2.4+dev-win64/
94M     build/parsec-v2.2.4+dev-win64/
```

vxgmichel:
> Size of win32 package without/with this patch: [...]

Did you measure the speed up at install time ? So we see what we're buying at this cost :)

touilleMan:
will be done in Rust instead

PR #1697, opened 2021-04-23
MacOS app fixes

### LINKED ISSUE #1519: Window put in background with red cross doesn't pop up when clicking dock icon

Ironicbay:
This aims at enhancing the app's behaviour to be more like usual Mac apps:

- Clicking the red X on Mac apps closes the window and leaves the process running in background. In this situation, clicking the Dock icon should pop the window back up, which wasn't the case. This is handled here with the `ApplicationActivate` event.
- Removal of the systray for MacOS as there is already is a default menu on the system with the same functions.
- Removal of the system notification when minimizing the app for MacOS, as this is default on most apps.

The app behaviour when `Minimize in tray` option is unchecked remains unchanged.

Closes #1519

### DIFF
--- a/parsec/core/gui/app.py
+++ b/parsec/core/gui/app.py
@@ -171,7 +171,7 @@ def run_gui(config: CoreConfig, start_arg: str = None, diagnose: bool = False):
             # Another instance of Parsec already started, nothing more to do
             return
 
-        if systray_available():
+        if systray_available() and sys.platform != "darwin":
             systray = Systray(parent=win)
             win.systray_notification.connect(systray.on_systray_notification)
             systray.on_close.connect(win.close_app)
--- a/parsec/core/gui/app.py
+++ b/parsec/core/gui/app.py
@@ -171,6 +171,7 @@ def run_gui(config: CoreConfig, start_arg: str = None, diagnose: bool = False):
             # Another instance of Parsec already started, nothing more to do
             return
 
+        # Systray is not displayed on MacOS, having natively a menu with similar functions.
         if systray_available() and sys.platform != "darwin":
             systray = Systray(parent=win)
             win.systray_notification.connect(systray.on_systray_notification)
--- a/parsec/core/gui/main_window.py
+++ b/parsec/core/gui/main_window.py
@@ -680,7 +680,7 @@ class MainWindow(QMainWindow, Ui_MainWindow):
         if self.minimize_on_close and not self.need_close:
             self.hide()
             event.ignore()
-            if not self.minimize_on_close_notif_already_send:
+            if not self.minimize_on_close_notif_already_send and platform.system() != "Darwin":
                 self.minimize_on_close_notif_already_send = True
                 self.systray_notification.emit(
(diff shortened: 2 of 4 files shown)


parsec/core/gui/app.py
```diff
@@ -171,7 +171,7 @@ def emit(self, *args):
             # Another instance of Parsec already started, nothing more to do
             return
 
-        if systray_available():
+        if systray_available() and sys.platform != "darwin":
```
touilleMan, line 174: Should add a comment to explain why not macOS

parsec/core/gui/main_window.py
```diff
@@ -680,7 +680,7 @@ def closeEvent(self, event):
         if self.minimize_on_close and not self.need_close:
             self.hide()
             event.ignore()
-            if not self.minimize_on_close_notif_already_send:
+            if not self.minimize_on_close_notif_already_send and platform.system() != "Darwin":
```
touilleMan, line 683: why the check is done on `sys.platform` in app.py and on `platform.system()` here ?

It seems so far the codebase use `sys.platform` and `platform.system()` interchangeably (and sometime even `os.name`, which is definitely not a good idea given it cannot tell appart Linux from macOS).
From what I understand the difference between the two is `sys.platform` is set at compilation time where `platform.system()` relies on `uname` to return the value (see https://github.com/python/cpython/blob/2f01c562be913004e1d46fc02705c51b81638d67/Lib/platform.py#L1061).

I would say it's better to use `sys.platform` so that we are guaranteed on the returned values (https://docs.python.org/fr/3/library/sys.html#sys.platform)

And I'm going to open an issue about the need to standardize the rest of the codebase ;-)

touilleMan: And more comment is needed to explain why macOS do it differently

parsec/core/gui/parsec_application.py
```diff
@@ -24,6 +26,10 @@ def __init__(self):
 
     def event(self, e):
         """Handle macOS FileOpen events."""
+        if sys.platform == "darwin" and e.type() == QEvent.ApplicationActivate:
```
touilleMan, line 29: According to https://doc.qt.io/qt-5/qevent.html, `QEvent.ApplicationActivate` is deprecated (and should be replaced by `ApplicationStateChange`), any idea why is it so ?

touilleMan (changes requested):

parsec/core/gui/parsec_application.py
Ironicbay: There are three events related to this issue: `ApplicationActivate`, `ApplicationDeactivate` and `ApplicationStateChange`.

The events `ApplicationDeactivate` and `ApplicationStateChange` are tied to whenever the window state changes from foreground to background. Inversely, the events `ApplicationActivate` and `ApplicationStateChange` can be caught when switching from background to foreground, or clicking the dock icon.

Even though `ApplicationActivate` is said to be deprecated, it's seemingly the only event we can use for this particular case, `ApplicationStateChange` not being specific enough

touilleMan: Great explanation ! You should add it to the code ;-)

touilleMan (approved):

PR #1709, opened 2021-04-27
Make explicit psutil&winreg imports in parsec.core.win_registry

touilleMan:
image

### DIFF
--- a/parsec/core/win_registry.py
+++ b/parsec/core/win_registry.py
@@ -3,7 +3,6 @@
 import os
 import string
 import sys
-from importlib import import_module
 import importlib_resources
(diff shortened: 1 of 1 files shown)


vxgmichel (approved):
LGTM :+1: 

We should mentioned that this PR is motivated by the fact that pyinstaller automatically chooses what module to include based on explicit imports (see #1690)

touilleMan:
> We should mentioned that this PR is motivated by the fact that pyinstaller automatically chooses what module to include based on explicit imports (see #1690)

Good idea, see 69c8d09a2126d6560fdeca002359bacac4e746e7 ;-)

PR #1710, opened 2021-04-27
Improve ipc_server bootstrap behavior when error occurs

touilleMan:


parsec/core/gui/app.py
```diff
@@ -49,41 +49,48 @@ def _before_quit():
 
 
 async def _start_ipc_server(config, main_window, start_arg, result_queue):
-    new_instance_needed_qt = ThreadSafeQtSignal(main_window, "new_instance_needed", object)
```
vxgmichel, line 52: I think this function should be called `_run_ipc_server` since it's a long running task.

parsec/core/gui/app.py
```diff
@@ -49,41 +49,48 @@ def _before_quit():
...
+            return {"status": "ok"}
+
+        while True:
             try:
+                async with run_ipc_server(
+                    _cmd_handler,
+                    config.ipc_socket_file,
+                    win32_mutex_name=config.ipc_win32_mutex_name,
+                ):
+                    result_queue.put_nowait("started")
```
vxgmichel, line 70: Maybe it's time to clean the queue communication? An enum would be nice here.

parsec/core/gui/app.py
```diff
@@ -49,41 +49,48 @@ def _before_quit():
...
+                    # IPC server has closed, retry to create our own
+                    continue
+
+                # We have successfuly noticed the other running application
+                result_queue.put_nowait("already_running")
+                break
+
+    except Exception:
+        result_queue.put_nowait("error")
+        raise
```
vxgmichel, line 93: What is the implication of the `raise` here? As far as I can tell the corresponding `run_ipc_server` job is not used so this exception is likely lost. The exception could also go in the queue.

More generally, the fact that we're mixing job communication and custom queue communication makes it a bit hard to read.

vxgmichel (approved):
LGTM, just a few comments about the implementation of the function.

parsec/core/gui/app.py
touilleMan: >  job is not used so this exception is likely lost

trio_thread module handles uncatched exceptions of all submitted jobs:
https://github.com/Scille/parsec-cloud/blob/2d877db087a1203c3024b963af0f356a42f39ef3/parsec/core/gui/trio_thread.py#L132-L136

But I agree this is a bit hidden, I'll add a comment about this

> More generally, the fact that we're mixing job communication and custom queue communication makes it a bit hard to read.

Yes, we should rewrite this code so that `run_gui` is divided in two callbacks (one before and one after ipc server has started). However it's not trivial to do given the `app.exec_` at the very end of the function (i.e. the Qt loop is not started until then, so no signal can be handled...)
Another solution would be to start the ipc server in a thread not related to the trio_thread stuff, but it's also messy because we want to run ipc_server stuff in the trio loop (which is handled by trio_trhead)

parsec/core/gui/app.py
touilleMan: :+1:

parsec/core/gui/app.py
touilleMan: :+1:

Issue #1701, opened 2021-04-23
Serializing the manifest of a large file might take up to 500ms


vxgmichel:
The following call has been measured taking 500 ms of CPU time for the manifest corresponding to freshly imported 600 MB file:
https://github.com/Scille/parsec-cloud/blob/125ed7252c8f15d0be1805201eb81a0be2c93eda/parsec/core/fs/storage/manifest_storage.py#L276

Todo: 
- [ ] investigate if the reshaped version can be serialized faster
- [ ] check the size of the produced binary, as it might explain a slow call to the database here:
https://github.com/Scille/parsec-cloud/blob/125ed7252c8f15d0be1805201eb81a0be2c93eda/parsec/core/fs/storage/manifest_storage.py#L279-L296
(it's been measured taking 350 ms)

vxgmichel:
A few measurements:
- dumping a remote file manifest with 2500 blocks: ~ 150 ms
- dumping a local manifest with 0 blocks and a base manifest with 2500 blocks: ~ 250 ms
- dumping a local manifest with 2500 blocks and a base manifest with 2500 blocks: ~ 500 ms

Issue #1700, opened 2021-04-23
Updating the chunks table might take up to 600 ms


vxgmichel:
The following two calls have caused warnings because they were blocking the loops for 600 ms:

https://github.com/Scille/parsec-cloud/blob/125ed7252c8f15d0be1805201eb81a0be2c93eda/parsec/core/fs/storage/chunk_storage.py#L123-L128

https://github.com/Scille/parsec-cloud/blob/125ed7252c8f15d0be1805201eb81a0be2c93eda/parsec/core/fs/storage/chunk_storage.py#L101-L106

This only happens when the database is growing in size, e.g when writing a file bigger than 500 MB

Issue #1699, opened 2021-04-23
The ChunkStorage.clear_old_blocks method might block the event loop for 250 ms


vxgmichel:
The following call can take up to 250 ms

https://github.com/Scille/parsec-cloud/blob/125ed7252c8f15d0be1805201eb81a0be2c93eda/parsec/core/fs/storage/chunk_storage.py#L174-L181

Also, there's a concurrency issue here:

https://github.com/Scille/parsec-cloud/blob/125ed7252c8f15d0be1805201eb81a0be2c93eda/parsec/core/fs/storage/chunk_storage.py#L189-L196

The amount of blocks to clean up is computed without holding the lock to the database, so concurrent call to `set_chunk` end up cleaning up more blocks that expected.

PR #1713, opened 2021-04-28
Fix blocking IO calls to sqlite when uploading large files

### LINKED ISSUE #1698: Closing the connection to the local database might take up to 600 ms
Hence, the following calls should use `LocalDatabase._run_in_thread`:

https://github.com/Scille/parsec-cloud/blob/125ed7252c8f15d0be1805201eb81a0be2c93eda/parsec/core/fs/storage/local_database.py#L190

https://github.com/Scille/parsec-cloud/blob/125ed7252c8f15d0be1805201eb81a0be2c93eda/parsec/core/fs/storage/local_database.py#L279

vxgmichel:
Also remove thread pool executor from the local database module since trio now caches its threads.

Fix #1698, #1699, #1700 and partially #1701.

### DIFF
--- a/parsec/core/fs/storage/chunk_storage.py
+++ b/parsec/core/fs/storage/chunk_storage.py
@@ -100,7 +100,7 @@ class ChunkStorage:
         async with self._open_cursor() as cursor:
             # Use a thread as executing a statement that modifies the content of the database might,
             # in some case, block for several hundreds of milliseconds
-            await trio.to_thread.run_sync(
+            await self.localdb.run_in_thread(
                 cursor.execute,
                 """
                 UPDATE chunks SET accessed_on = ? WHERE chunk_id = ?;
@@ -124,7 +124,7 @@ class ChunkStorage:
         async with self._open_cursor() as cursor:
             # Use a thread as executing a statement that modifies the content of the database might,
             # in some case, block for several hundreds of milliseconds
-            await trio.to_thread.run_sync(
+            await self.localdb.run_in_thread(
                 cursor.execute,
                 """INSERT OR REPLACE INTO
                 chunks (chunk_id, size, offline, accessed_on, data)
@@ -136,7 +136,7 @@ class ChunkStorage:
         async with self._open_cursor() as cursor:
             # Use a thread as executing a statement that modifies the content of the database might,
             # in some case, block for several hundreds of milliseconds
-            await trio.to_thread.run_sync(
+            await self.localdb.run_in_thread(
                 cursor.execute, "DELETE FROM chunks WHERE chunk_id = ?", (chunk_id.bytes,)
             )
             cursor.execute("SELECT changes()")
@@ -190,7 +190,7 @@ class BlockStorage(ChunkStorage):
             # Insert the chunk
             # Use a thread as executing a statement that modifies the content of the database might,
             # in some case, block for several hundreds of milliseconds
-            await trio.to_thread.run_sync(
+            await self.localdb.run_in_thread(
                 cursor.execute,
                 """INSERT OR REPLACE INTO
                 chunks (chunk_id, size, offline, accessed_on, data)
@@ -211,7 +211,7 @@ class BlockStorage(ChunkStorage):
             limit = extra_blocks + self.block_limit // 10
             # Use a thread as executing a statement that modifies the content of the database might,
             # in some case, block for several hundreds of milliseconds
-            await trio.to_thread.run_sync(
+            await self.localdb.run_in_thread(
                 cursor.execute,
                 """
                 DELETE FROM chunks WHERE chunk_id IN (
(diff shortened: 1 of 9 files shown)


vxgmichel:
Note: this PR decreases the performance when writing large files using the mountpoint:
- about 4.8 MB/s before the PR
- about 3.3 MB/s after the PR

parsec/core/types/manifest.py
```diff
@@ -670,6 +676,9 @@ def to_stats(self):
     def parent(self):
         return self.base.parent
 
+    def size_estimate(self):
+        return 1 + len(self.base.children) + len(self.children)
```
AntoineDupre, line 680: I don't really get the logic here. Why ` len(self.base.children) + len(self.children)` is not enough

Also, it seems that this logic (introduced in [this commit](https://github.com/Scille/parsec-cloud/commit/7d54659e759276b37a85dac80ddb8c1f5d1d45f6) ) is not used anymore (removed [there](https://github.com/Scille/parsec-cloud/commit/f8294eccb121a12bb523302ccda4dc8fff9205e9#diff-0a806323e83144ea32278ce2ffd433c2af0473ac5095bd58d2788b336b8931ec) )

AntoineDupre (approved):

parsec/core/types/manifest.py
vxgmichel: Good catch! I left it on purpose so we can discuss its usefulness, as this estimate can be useful for other serde operation in the code (for remote manifests for instance).

The size estimate is an estimate of the number of entries that the manifest is referencing (as it should grow roughly linearly with the serialization time). I'm considering the manifest itself as an entry so empty file/folder/workspace manifests have the same size as a user manifest, i.e. 1. 

It's quite arbitrary though, and it can be changed/removed.

parsec/core/fs/storage/local_database.py
```diff
@@ -127,7 +86,7 @@ def _manage_operational_error(self, allow_commit: bool = False) -> Iterator[None
 
             # Close the sqlite3 connection
             try:
-                self._conn.close()
+                await self.run_in_thread(self._conn.close)
```
touilleMan, line 89: Just to be sure: given the call is now async, it is impacted by `trio.Cancelled` exception
It should have an impact for the main case (i.e. when an `OperationalError` is raised), but in more complex case (`OperationalError` then `trio.Cancelled`) the behavior is different

parsec/core/fs/storage/manifest_storage.py
```diff
@@ -269,37 +269,46 @@ async def _ensure_manifest_persistent(self, entry_id: EntryID) -> None:
...
             manifest = self._cache[entry_id]
-
-            # Dump and decrypt the manifest
-            ciphered = manifest.dump_and_encrypt(self.device.local_symkey)
-
-            # Insert into the local database
-            cursor.execute(
-                """INSERT OR REPLACE INTO vlobs (vlob_id, blob, need_sync, base_version, remote_version)
-                VALUES (
+            pending_chunks_ids = [
```
touilleMan, line 274: The behavior has changed here:
- in the old code `self._cache_ahead_of_localdb[entry_id]` was computed at the end of the operations
- in the new code, `self._cache_ahead_of_localdb[entry_id]` is computed at the beginning of the operations

On top of that, the old code blocked the trio loop (so no concurrent operations could change `self._cache_ahead_of_localdb[entry_id]` in our back) and now we are doing our stuff in another thread (so `self._cache_ahead_of_localdb[entry_id]` can be modified at any time)

I guess we need more comment explaining the intent here ;-)

vxgmichel: That's exactly right :grin: 

The overall idea is explained here: https://github.com/Scille/parsec-cloud/issues/1715

Also note the comment below, line 310:
> # Tag entry as up-to-date only if no new manifest has been written in the meantime

parsec/core/fs/storage/chunk_storage.py
```diff
@@ -169,28 +178,45 @@ async def clear_all_blocks(self) -> None:
         async with self._open_cursor() as cursor:
             cursor.execute("DELETE FROM chunks")
 
-    async def clear_old_blocks(self, limit: int) -> None:
+    # Upgraded set method
+
+    async def set_chunk(self, chunk_id: ChunkID, raw: bytes) -> None:
+        assert isinstance(raw, (bytes, bytearray))
```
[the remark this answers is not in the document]
vxgmichel: c4bb4eb50061e9a08cad5f2606bd6dfcdee96e17

parsec/core/types/manifest.py
vxgmichel: 0ac8ce05b053026fdfec8e7242a425a82b0f47bc

parsec/core/fs/storage/local_database.py
vxgmichel: 4ba56f6623e2f17bea1d06876885307bfb6c61b4

PR #1716, opened 2021-04-30
Fix a fuse issue where the shutdown procedure blocks forever

vxgmichel:
The problem has been observed on linux, usually when trying to exit the application after the writing of a large file, during the synchronization. It happens sporadically.


Ironicbay (approved):
LGTM !

PR #1717, opened 2021-05-04
Fix a fuse issue where the shutdown procedure blocks forever, second edition

vxgmichel:
PR #1716 introduced some freezing in the CI.

This PR implements another approach to fix the issue.

### DIFF
--- a/parsec/core/mountpoint/fuse_runner.py
+++ b/parsec/core/mountpoint/fuse_runner.py
@@ -1,6 +1,5 @@
 # Parsec Cloud (https://parsec.cloud) Copyright (c) AGPLv3 2016-2021 Scille SAS
 
-from parsec.core.core_events import CoreEvent
 import os
 import sys
 import trio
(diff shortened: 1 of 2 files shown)


parsec/core/mountpoint/fuse_runner.py
```diff
@@ -232,13 +232,26 @@ async def _stop_fuse_thread(
 
     # Fuse exit in macOS
     if sys.platform == "darwin":
+
         # The schedule_exit() solution doesn't work on macOS, instead freezes the application for
         # 120 seconds before a timeout occurs. The solution used is to call this function (macOS
         # equivalent to fusermount) in a subprocess to unmount.
-        await trio.run_process(["diskutil", "unmount", "force", str(mountpoint_path)])
+        process_args = ["diskutil", "unmount", "force", str(mountpoint_path)]
+        process = await trio.open_process(process_args)
```
Ironicbay, line 240: What's the point of using `open_process` instead of `run_process` here ?

vxgmichel: it's basically the same logic as the one with the daemon thread below, we fire a background task that is supposed to cause the FS thread to stop, and we wait for the corresponding event while restarting the background task if necessary.

vxgmichel: See this code within the loop:
```python
            if process.poll() is not None:
                process = await trio.open_process(process_args)
```

Ironicbay (approved):

PR #1719, opened 2021-05-05
Fix conditional snap&pypi upload in CI when depending jobs has been cancelled

touilleMan:
see: https://docs.microsoft.com/fr-fr/azure/devops/pipelines/process/conditions?view=azure-devops&tabs=yaml#run-for-the-main-branch-if-succeeding


vxgmichel (approved):
LGTM :+1:

PR #1728, opened 2021-05-11
Fix newsfragment check in CI for first commit (adding the `+dev` suff…

touilleMan:
…ix to version) after tag

### DIFF
--- a/.azure-pipelines.yml
+++ b/.azure-pipelines.yml
@@ -84,8 +84,7 @@ jobs:
       pre-commit run --all-files --show-diff-on-failure
     displayName: 'Pre-commit hooks check'
   - bash: |
-      if ([[ $(Build.SourceBranch) = "refs/heads/master" ]] || [[ $(Build.SourceBranch) = "refs/tags/"* ]])
-      then exit 0; fi
+      # PRs must contain a newsfragment
       for FILENAME in newsfragments/*
       do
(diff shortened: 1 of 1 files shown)


Ironicbay (approved):

.azure-pipelines.yml
```diff
@@ -94,6 +94,10 @@ jobs:
         if git log origin/master --exit-code -- "$FILENAME" > /dev/null
         then exit 0; fi
       done
+      # Special case if we are the first commit after a tag (commit that
+      # set `+dev` version prefix, hence no newsfragment provided)
+      if ([[ $(git log --oneline $(git describe --tags --abbrev=0).. | wc -l) -eq 1 ]]);
+      then exit 0; fi
```
vxgmichel, line 100: Looks a bit fragile to me, maybe simply run the newsfragment check if it comes from a PR (since the `+dev` commit can be pushed directly to master).

vxgmichel (changes requested):

.azure-pipelines.yml
touilleMan: 👍 see 5127369bc24f0e085cb0c56682cf0186993cbda6

vxgmichel (approved):

touilleMan:
> It might be a bit more clear to use conditions:

I wish I knew that earlier :'(

PR #1735, opened 2021-05-17
Add commit signature check in CI

touilleMan:


.azure-pipelines.yml
```diff
@@ -75,6 +75,28 @@ jobs:
...
+      import json
+      import re
+      match = re.match(r'refs/pull/([0-9]+)/merge', '$(Build.SourceBranch)')
+      if match:
+        pr_id = match.group(1)
+        r = urlopen(f'https://api.github.com/repos/Scille/parsec-cloud/pulls/{pr_id}/commits')
+        bad_commits = [c['sha'] for c in json.load(r) if c['commit']['verification']['reason'] == 'invalid']
+        if bad_commits:
+          raise SystemExit(f'''Invalid signatures in commits: {', '.join(bad_commits)}''')
+      "
```
vxgmichel, line 98: Maybe move to a dedicated python script and use:
```bash
./misc/check_commit_signatures_for_pr $(System.PullRequest.PullRequestNumber)
```

vxgmichel (approved):
A single comment about moving the logic to a dedicated python script.

.azure-pipelines.yml
touilleMan: For the moment I only consider this code as a little sanity check so it's fine to inline it in the azure pipeline (we shouldn't need this as a separate script because gpg shouldn't have created this bad commit in the first place).

vxgmichel: I don't really agree, this CI file is already 600 lines long and hard to navigate. Also inline code is hard to read for a couple of reasons:
- no syntax highlighting and other IDE support
- it's a bash string, with all the quote and substitution mess that comes with it

I'm not going to fight over it though, I already validated this PR so it's fine if you want to merge it as it is :)

touilleMan: The script is 10 lines long, almost half of the lines are for a comment explaining why this mess is needed.
I'm not convinced it would be more readable in a separate file (given you'd have to switch between the two, with the explanation comment missing in one of the places)

Of course it would be must better if we could split this file between jobs ;-)

PR #1739, opened 2021-05-17
Quick'n dirty fix to prevent QTimer from waiting too much during test…

touilleMan:
…s (which lead to slow tests and possible unexpected timeouts)


vxgmichel:
PR #1742 has been merged instead.

PR #1741, opened 2021-05-18
Add pop-up to download latest version on MacOS

Ironicbay:


parsec/core/gui/new_version.py
```diff
@@ -145,7 +149,7 @@ def __init__(self, jobs_ctx, event_bus, config, **kwargs):
         super().__init__(**kwargs)
         self.setupUi(self)
 
-        if sys.platform != "win32":
+        if sys.platform != "win32" and sys.platform != "darwin":
```
touilleMan, line 152:
```python
        if sys.platform not in ("win32", "darwin"):
```

parsec/core/gui/new_version.py
```diff
@@ -74,8 +74,12 @@ def _fetch_json_releases():
                     continue
                 if release["prerelease"] and not check_pre:
                     continue
+                if sys.platform == "darwin":
```
touilleMan, line 77: This would be better moved into line 61 (if current_arch ... set win_version):

```python
if sys.platform == "win32" and current_arch == "x86_64":
    installer_suffix = "-win64-setup.exe"
elif sys.platform == "win32" and current_arch == "i386":
    installer_suffix = "-win32-setup.exe"
elif sys.platform == "darwin":
    installer_suffix = "-macos-amd64.dmg"
else:
    # Generic fallback
    return latest_from_head, url
[...]
```

touilleMan (changes requested):

parsec/core/gui/new_version.py
```diff
@@ -20,78 +22,95 @@
...
+            logger.error("Cannot deserialize releases info from API", exc_info=exc, api_url=api_url)
+            return None
+        except HTTPError:
+            return None
+
+    return await trio.to_thread.run_sync(_do_http_request)
+
+
+async def do_check_new_version(
+    api_url: str, allow_prerelease: bool = False
```
Ironicbay, line 70: Deleting the `url` argument causes errors elsewhere making the pop-up not show at all

```
File "parsec/core/gui/trio_thread.py", line 95, in _run_fn
    result = await self._fn(*self._args, **self._kwargs)
TypeError: do_check_new_version() got an unexpected keyword argument 'url'
```

touilleMan (approved):
lgtm ;-)

PR #1742, opened 2021-05-18
Replace QTimer in GUI by QtToTrioJobScheduler.submit_delayed_job

### LINKED ISSUE #1740: Avoid QTimer in the GUI
see #1739

We already avoid `QTimer` as much as possible given it lead to complex spaghetti code and polling. Two more reasons to avoid it:
- The GUI tests are greatly slowed down when waiting for something that is triggered by a `QTimer`
- The GUI tests can even end up in hard to reproduce timeout (typically with a timeout of 3s and a `QTimer` triggered every second), see for instance https://dev.azure.com/Scille/parsec/_build/results?buildId=6032&view=logs&j=5ba45810-c0e9-5996-abe7-b2cb0c9b6baa&t=b18150ab-3a6d-5272-0b08-a0dc06a2acd4

The place we use `QTimer`:
```
$ rg QTimer parsec/core/gui
parsec/core/gui/workspaces_widget.py
6:from PyQt5.QtCore import pyqtSignal, QTimer, Qt
240:        self.reset_timer = QTimer()

parsec/core/gui/files_widget.py
11:from PyQt5.QtCore import Qt, pyqtSignal, QTimer
305:        self.update_timer = QTimer()

parsec/core/gui/app.py
14:from PyQt5.QtCore import QTimer, Qt
215:        # QTimer wakes up the event loop periodically which allows us to close
217:        timer = QTimer()
222:            diagnose_timer = QTimer()

parsec/core/gui/central_widget.py
4:from PyQt5.QtCore import pyqtSignal, QTimer
81:        self.organization_stats_timer = QTimer()

parsec/core/gui/workspace_sharing_widget.py
5:from PyQt5.QtCore import QCoreApplication, pyqtSignal, QEvent, QTimer
122:        self.status_timer = QTimer()
```

> parsec/core/gui/app.py

From what I recall, this one is hard to remove given it aims at forcing
Qt loop to do something so we can close the main window event if it is in the background...
We may want to put more investigation into the reason of this (I'd like to see
more documentation about this in the code given it's quiet exotic ^^), but
no need to remove `QTimer` from here I think.

> parsec/core/gui/workspaces_widget.py

This is the bad boy that cause all the GUI test to slow down and timeout, should
definitively remove it and replace it by core event handling.

> parsec/core/gui/files_widget.py

Pretty sure this one cause the same kind of issues that `workspaces_widget.py`

> parsec/core/gui/central_widget.py

This one is here only to send a `organization_stats_timer` command every 5s to
the backend...
Seems pretty wasteful, I think we could trigger the stat command when the window
is created.
I would even remove this entirely and add a this information in the logout menu
(so that the command is send only when the menu is opened)

> parsec/core/gui/workspace_sharing_widget.py

This could be replaced by events from the sync monitor

touilleMan:
The non-hacky brother of #1739

fix #1740

### DIFF
--- a/parsec/core/gui/app.py
+++ b/parsec/core/gui/app.py
@@ -213,8 +213,8 @@ def run_gui(config: CoreConfig, start_arg: Optional[str] = None, diagnose: bool 
         # QTimer wakes up the event loop periodically which allows us to close
         # the window even when it is in background.
         timer = QTimer()
-        timer.start(1000 if diagnose else 400)
-        timer.timeout.connect(kill_window if diagnose else lambda: None)
+        timer.start(400)
+        timer.timeout.connect(lambda: None)
 
         if diagnose:
             diagnose_timer = QTimer()
--- a/parsec/core/gui/files_widget.py
+++ b/parsec/core/gui/files_widget.py
@@ -237,7 +237,7 @@ class Clipboard:
 
 
 class FilesWidget(QWidget, Ui_FilesWidget):
-    RELOAD_FILES_LIST_DELAY = 1  # 1s
+    RELOAD_FILES_LIST_THROTTLE_DELAY = 1  # 1s
 
     fs_updated_qt = pyqtSignal(CoreEvent, UUID)
     fs_synced_qt = pyqtSignal(CoreEvent, UUID)
@@ -623,15 +623,15 @@ class FilesWidget(QWidget, Ui_FilesWidget):
         elif file_type == FileType.Folder:
             self.load(self.current_directory / file_name)
 
-    def schedule_reload(self):
-        self.load(self.current_directory, delay=self.RELOAD_FILES_LIST_DELAY)
+    def reload(self):
+        self.load(self.current_directory, throttle_delay=self.RELOAD_FILES_LIST_THROTTLE_DELAY)
 
-    def load(self, directory, default_selection=None, delay: float = 0):
+    def load(self, directory, default_selection=None, throttle_delay: float = 0):
         self.table_files.clear()
         self.spinner.show()
         self.jobs_ctx.submit_throttled_job(
             "files_widget.load",
-            delay,
+            throttle_delay,
             ThreadSafeQtSignal(self, "folder_stat_success", QtToTrioJob),
             ThreadSafeQtSignal(self, "folder_stat_error", QtToTrioJob),
             _do_folder_stat,
@@ -962,7 +962,7 @@ class FilesWidget(QWidget, Ui_FilesWidget):
         if ws_id != workspace_id:
             return
         if id == self.current_directory_uuid:
-            self.schedule_reload()
+            self.reload()
 
     def _on_fs_synced_qt(self, event, uuid):
         if not self.workspace_fs:
@@ -986,7 +986,7 @@ class FilesWidget(QWidget, Ui_FilesWidget):
             return
 
         if self.current_directory_uuid == uuid or self.table_files.has_file(uuid):
-            self.schedule_reload()
+            self.reload()
 
     def _on_sharing_updated_trio(self, event, new_entry, previous_entry):
         self.sharing_updated_qt.emit(new_entry, previous_entry)
--- a/parsec/core/gui/files_widget.py
+++ b/parsec/core/gui/files_widget.py
@@ -636,9 +636,7 @@ class FilesWidget(QWidget, Ui_FilesWidget):
         elif file_type == FileType.Folder:
             self.load(self.current_directory / file_name)
 
-    def reload(self, default_selection=None, delay=None):
-        if delay is None:
-            delay = self.RELOAD_FILES_LIST_THROTTLE_DELAY
+    def reload(self, default_selection=None, delay=RELOAD_FILES_LIST_THROTTLE_DELAY):
         self.jobs_ctx.submit_throttled_job(
             "files_widget.reload",
             delay,
--- a/parsec/core/gui/files_widget.py
+++ b/parsec/core/gui/files_widget.py
@@ -156,10 +156,21 @@ async def _do_move_files(workspace_fs, target_dir, source_files, source_workspac
 async def _do_folder_stat(workspace_fs, path, default_selection):
     stats = {}
     dir_stat = await workspace_fs.path_info(path)
+    # Retrieve children info, this is not an atomic operation so our view
+    # on the folder might be slightly non-causal (typically if a file is
+    # created during while we do the loop it won't appear in the final result,
+    # but later update on existing files will appear).
+    # We consider this fine enough given a change in the folder should lead
+    # to an event from the corefs which in turn will re-trigger this stat code
     for child in dir_stat["children"]:
         try:
             child_stat = await workspace_fs.path_info(path / child)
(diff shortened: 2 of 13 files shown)


parsec/core/gui/app.py
```diff
@@ -215,8 +215,8 @@ def kill_window(*args):
         # QTimer wakes up the event loop periodically which allows us to close
         # the window even when it is in background.
         timer = QTimer()
-        timer.start(1000 if diagnose else 400)
-        timer.timeout.connect(kill_window if diagnose else lambda: None)
+        timer.start(400)
+        timer.timeout.connect(lambda: None)
```
vxgmichel, line 219: Any reason for removing `diagnose`?

parsec/core/gui/files_widget.py
```diff
@@ -628,13 +626,15 @@ def item_activated(self, file_type, file_name):
         elif file_type == FileType.Folder:
             self.load(self.current_directory / file_name)
 
-    def reload(self):
-        self.load(self.current_directory)
+    def schedule_reload(self):
+        self.load(self.current_directory, delay=self.RELOAD_FILES_LIST_DELAY)
```
vxgmichel, line 630: Renaming to `schedule_reload` is no longer justified.

parsec/core/gui/trio_thread.py
```diff
@@ -200,6 +205,62 @@ def _run_job(self, job, *args, sync=False):
...
+            nonlocal qt_on_success, qt_on_error, args, kwargs
+            submitted = trio.current_time()
+
+            # Only modify `_scheduled_jobs_args` from the trio
+            # thread to avoid concurrent acces with the Qt thread
+            self._scheduled_jobs_args[throttling_id] = (qt_on_success, qt_on_error, args, kwargs)
+            task_status.started()
+
+            last_executed = self._scheduled_jobs_last_executed.get(throttling_id, 0)
+            await trio.sleep_until(last_executed + delay)
```
vxgmichel, line 233: Maybe add a comment to show that sleeping is only needed if a similar job ran in the last `delay` seconds.

parsec/core/gui/trio_thread.py
```diff
@@ -200,6 +205,62 @@ def _run_job(self, job, *args, sync=False):
...
+
+            # Make sure fn has not been executed while we were waiting
+            job_args = self._scheduled_jobs_args.pop(throttling_id, None)
+            last_executed = self._scheduled_jobs_last_executed.get(throttling_id, 0)
+            if not job_args or last_executed > submitted:
+                return
```
vxgmichel, line 239: There's a bug here: `job_args` has been popped but is discarded if `last_executed > submitted`, which means the other scheduled job will have nothing to pop from `_scheduled_jobs_args`

vxgmichel: Also I'm not a big fan of `submitted` as it adds an extra timestamp (and extra race conditions).

Instead maybe use `last_executed + delay > trio.current_time()` which doesn't require the extra variable.

vxgmichel (changes requested):
A few comments, LGTM otherwise :+1:

parsec/core/gui/app.py
touilleMan: diagnose is already handled in the next lines, I didn't see the need to add it here (we could also remove the `if diagnose` stuff and keep this timer multi-purposed)

parsec/core/gui/trio_thread.py
touilleMan: Good catch !

> Instead maybe use last_executed + delay > trio.current_time() which doesn't require the extra variable.

:+1:

parsec/core/gui/trio_thread.py
touilleMan: yup, I love comments !

parsec/core/gui/files_widget.py
touilleMan: :+1:

vxgmichel (approved):
:+1:

vxgmichel:
Aouch: https://dev.azure.com/Scille/parsec/_build/results?buildId=6069&view=logs&j=5ba45810-c0e9-5996-abe7-b2cb0c9b6baa&t=b18150ab-3a6d-5272-0b08-a0dc06a2acd4&l=54

vxgmichel:
@touilleMan GUI tests should now be consistent \o/ (at least `pytest --rungui -m gui --count 10` runs on my machine).

Could you review the changes I made?

tests/core/gui/test_main_window.py
```diff
@@ -110,6 +110,8 @@ def _workspace_button_ready():
         assert w_w.layout_workspaces.count() == 1
         wk_button = w_w.layout_workspaces.itemAt(0).widget()
         assert not isinstance(wk_button, QtWidgets.QLabel)
+        # Important: this is necessary for the incoming click to reliably work
```
touilleMan, line 113: Not sure why this is important: this check box enable workspace mounting, however I don't think we test the mountpoint in GUI test
(on top of that I don't remember the rule for a new workspace to be mounted or not...)

parsec/core/gui/files_widget.py
```diff
@@ -635,20 +636,31 @@ def item_activated(self, file_type, file_name):
         elif file_type == FileType.Folder:
             self.load(self.current_directory / file_name)
 
-    def reload(self):
-        self.load(self.current_directory, throttle_delay=self.RELOAD_FILES_LIST_THROTTLE_DELAY)
+    def reload(self, default_selection=None, delay=None):
```
touilleMan, line 639:
```python
def reload(self, default_selection=None, delay=RELOAD_FILES_LIST_THROTTLE_DELAY):
```
?

touilleMan (commented):
lgtm (but can't auto-approve my own PR 😢 )

tests/core/gui/test_files.py
```diff
@@ -107,14 +107,7 @@ def _do_selection():
...
         async def reset_selection(self):
-            for r in f_w.table_files.selectedRanges():
-                top = r.topRow()
-                bottom = r.bottomRow()
-                await aqtbot.run(
-                    f_w.table_files.setRangeSelected,
-                    QtWidgets.QTableWidgetSelectionRange(top, 0, bottom, 0),
-                    False,
-                )
+            await aqtbot.run(f_w.table_files.reset)
```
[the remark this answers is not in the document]
vxgmichel: Both!

tests/core/gui/test_main_window.py
vxgmichel: It is important as left click is disabled if the check box is checked (i.e the workspace is not mounted). This was the reason why some tests were inconsistent, when the click happened to be sent before the the check box was enabled.

tests/core/gui/test_files.py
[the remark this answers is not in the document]
vxgmichel: I didn't look much into it but I noticed some selection could remain in some cases after a call to `reset_selection`. Simply using the dedicated qt call fixed the issue.

parsec/core/gui/files_widget.py
vxgmichel: d1812a2

PR #1748, opened 2021-05-27
Improve MacOS mountpoint cleanup function

Ironicbay:


parsec/core/mountpoint/manager.py
```diff
@@ -268,15 +268,18 @@ async def safe_unmount_all(self):
 
 
 async def cleanup_macos_mountpoint_folder(base_mountpoint_path):
-    # In case of a crash on macOS, workspaces don't unmount correctly and leave empty directories
-    # in the default mount folder. This function is used to clean these anytime a login occurs.
+    # In case of a crash on macOS, workspaces don't unmount correctly and leave
+    # empty mountpoints or directories in the default mount folder. This
+    # function is used to unmount or delete these anytime a login occurs.
     for dirs in os.listdir(base_mountpoint_path):
```
touilleMan, line 274: `dirs` ending with a `s` suggest it is an array, what about `mountpoint_name` instead (and `mountpoint_path` instead of `dir_path`) ?

parsec/core/mountpoint/manager.py
```diff
@@ -268,15 +268,18 @@ async def safe_unmount_all(self):
...
         dir_path = str(base_mountpoint_path) + "/" + str(dirs)
         stats = os.statvfs(dir_path)
         if stats.f_blocks == 0 and stats.f_ffree == 0 and stats.f_bavail == 0:
-            await trio.run_process(["diskutil", "unmount", "force", dir_path])
-            if dirs in os.listdir(base_mountpoint_path):  # checking if there is something to delete
-                await trio.run_process(["rm", "-d", dir_path])  # otherwise an empty dir remains
+            try:
+                await trio.run_process(["diskutil", "unmount", "force", dir_path])
+                await trio.run_process(["rm", "-d", dir_path])
+            except:
```
touilleMan, line 281: Raw `except` is not a good idea:
- we catch `BaseException` which is used internally by the python interpreter and low level stuff (e.g. `SystemExit`, `StopIteration`, `trio.Cancelled`etc.)
- we have no idea which exception is expected, so code is hard to refactor and to understand

touilleMan: On top of that, it would be good to add a warning log of the exception (warning log is not returned to Sentry, but is collected as part of the breadcrumbs that are provide as context if an error log occurs not too long after)

touilleMan (changes requested):

parsec/core/mountpoint/manager.py
```diff
@@ -268,15 +268,18 @@ async def safe_unmount_all(self):
...
+        mountpoint_path = str(base_mountpoint_path) + "/" + str(mountpoint_name)
+        stats = os.statvfs(mountpoint_path)
         if stats.f_blocks == 0 and stats.f_ffree == 0 and stats.f_bavail == 0:
-            await trio.run_process(["diskutil", "unmount", "force", dir_path])
-            if dirs in os.listdir(base_mountpoint_path):  # checking if there is something to delete
-                await trio.run_process(["rm", "-d", dir_path])  # otherwise an empty dir remains
+            try:
+                await trio.run_process(["diskutil", "unmount", "force", mountpoint_path])
+                await trio.run_process(["rm", "-d", mountpoint_path])
+            except BaseException as e:
```
touilleMan, line 281: `except:` is equivalent to `except BaseException:`, so we're still catching far too broad here 😄

parsec/core/mountpoint/manager.py
```diff
@@ -268,15 +268,18 @@ async def safe_unmount_all(self):
...
+                logger.warning(f"MacOS cleanup exception: {e}")
```
touilleMan, line 282: you'd better use `exc_info` field to provide the exception in the log (this way the log message will automatically display the traceback, and sentry can do black magic on it !)

touilleMan (changes requested):

parsec/core/mountpoint/manager.py
```diff
@@ -268,15 +268,19 @@ async def safe_unmount_all(self):
...
         if stats.f_blocks == 0 and stats.f_ffree == 0 and stats.f_bavail == 0:
-            await trio.run_process(["diskutil", "unmount", "force", dir_path])
-            if dirs in os.listdir(base_mountpoint_path):  # checking if there is something to delete
-                await trio.run_process(["rm", "-d", dir_path])  # otherwise an empty dir remains
+            try:
+                await trio.run_process(["diskutil", "unmount", "force", mountpoint_path])
+                if Path(mountpoint_path).is_dir():
+                    await trio.run_process(["rm", "-d", mountpoint_path])
+            except (CalledProcessError, FileNotFoundError) as exc:
+                logger.warning(f"MacOS cleanup exception", exc_info=exc)
```
touilleMan, line 283: would be good to add `mountpoint_path` as part of the log (structlog allow to pass arbitrary kwargs for this)

parsec/core/mountpoint/manager.py
```diff
@@ -268,15 +268,19 @@ async def safe_unmount_all(self):
...
-    # In case of a crash on macOS, workspaces don't unmount correctly and leave empty directories
-    # in the default mount folder. This function is used to clean these anytime a login occurs.
-    for dirs in os.listdir(base_mountpoint_path):
-        dir_path = str(base_mountpoint_path) + "/" + str(dirs)
-        stats = os.statvfs(dir_path)
+    # In case of a crash on macOS, workspaces don't unmount correctly and leave
+    # empty mountpoints or directories in the default mount folder. This
+    # function is used to unmount or delete these anytime a login occurs.
+    for mountpoint_name in os.listdir(base_mountpoint_path):
+        mountpoint_path = str(base_mountpoint_path) + "/" + str(mountpoint_name)
```
touilleMan, line 275: `base_mountpoint_path` is a `pathlib.Path` (given it's the type used in `CoreConfig`)
so you can do
```python
mountpoint_path = base_mountpoint_path / mountpoint_name
```

parsec/core/mountpoint/manager.py
```diff
@@ -268,15 +268,19 @@ async def safe_unmount_all(self):
...
+        stats = os.statvfs(mountpoint_path)
```
touilleMan, line 276: Cannot use `pathlib.Path.stat` instead ? (and especially the trio version `trio.Path.stat`, which guarantees us the trio loop won't be stuck if the stat is done on a folder that is still a valid mountpoint)

parsec/core/mountpoint/manager.py
```diff
@@ -268,15 +268,19 @@ async def safe_unmount_all(self):
...
         if stats.f_blocks == 0 and stats.f_ffree == 0 and stats.f_bavail == 0:
-            await trio.run_process(["diskutil", "unmount", "force", dir_path])
-            if dirs in os.listdir(base_mountpoint_path):  # checking if there is something to delete
-                await trio.run_process(["rm", "-d", dir_path])  # otherwise an empty dir remains
+            try:
+                await trio.run_process(["diskutil", "unmount", "force", mountpoint_path])
+                if Path(mountpoint_path).is_dir():
+                    await trio.run_process(["rm", "-d", mountpoint_path])
```
touilleMan, line 281: why don't we use directly the python api for removing directory ?
Trio provide an async wrapper over it, so we could do:
```python
try:
    await trio.run_process(["diskutil", ...])
except CalledProcessError as exc:
    logger.warning("Error during mountpoint cleanup: disktutil unmount failed", mountpoint_path=mountpoint_path, exc_info=exc)

try:
    await trio.Path(mountpoint_path).rmdir()
except FileNotFoundError:
    pass

```

(btw `base_mountpoint_path` is pathlib.Path)

parsec/core/mountpoint/manager.py
touilleMan: And what about `"Error during mountpoint cleanup: disktutil unmount failed"` for the log wording:
- no need to talk about MacOS (given we always know the platform the application runs on)
- we mention the "diskutil" command given it's a bit exotic stuff

parsec/core/mountpoint/manager.py
```diff
@@ -300,15 +304,7 @@ async def mountpoint_manager_factory(
...
-        try:
-            await cleanup_macos_mountpoint_folder(base_mountpoint_path)
-        except FileNotFoundError:
-            pass
-        except CalledProcessError:
-            # The unmount can fail if the app wasn't properly closed and will
-            # block login until reboot. Cleanup is skipped in such cases and
-            # should work properly after the next user's reboot.
-            pass
+        await cleanup_macos_mountpoint_folder(base_mountpoint_path)
```
touilleMan, line 307: Comment about `CalledProcessError` exception has been removed, is it because it was incorrect or because the behavior is no longer true ?

touilleMan (changes requested):

parsec/core/mountpoint/manager.py
Ironicbay: My mistake, I thought it was a duplicate of the comment inside the cleanup function. I'll add the missing information in the latter now that the exceptions have moved

parsec/core/mountpoint/manager.py
```diff
@@ -268,15 +268,36 @@ async def safe_unmount_all(self):
...
-            await trio.run_process(["diskutil", "unmount", "force", dir_path])
-            if dirs in os.listdir(base_mountpoint_path):  # checking if there is something to delete
-                await trio.run_process(["rm", "-d", dir_path])  # otherwise an empty dir remains
+    # In case of a crash on macOS, workspaces don't unmount correctly and leave empty
+    # mountpoints or directories in the default mount folder. This function is used to
+    # unmount and/or delete these anytime a login occurs. In some rare and so far unknown
+    # conditions, the unmount call can fail, raising a CalledProcessError, hence the
+    # exception below. In such cases, the issue stops occurring after a system reboot,
+    # making the exception catching a temporary solution.
+    if not base_mountpoint_path.is_dir():
```
touilleMan, line 277: would be more elegant to use `trio.Path` everywhere
```python
base_mountpoint_path = trio.Path(base_moutpoint_path)
try:
    mountpoint_names = await base_mountpoint_path.iterdir()
except FileNotFoundError:
    # Unlike with `pathlib.Path.iterdir` which returns a lazy itertor, `trio.Path.iterdir` does FS access and may raise FileNotFoundError
    return
for mountpoint_name in mountpoint_names:
    [...]
```

touilleMan (approved):
lgtm ;-)

PR #1749, opened 2021-05-28
Improve macOS CI

Ironicbay:


.azure-pipelines.yml
```diff
@@ -253,22 +253,26 @@ jobs:
...
-  #     set -eux
-  #     py.test $(pytest.base_args) \
-  #       tests/backend --postgresql --runslow \
-  #       -n auto --max-worker-restart=0 -x \
-  #       --junitxml=test-results-postgresql.xml
-  #   displayName: 'Tests postgresql'
+  - bash: |
+      set -eux
+      py.test $(pytest.base_args) \
+        tests --runmountpoint --runslow \
```
touilleMan, line 259: `--runmountpoint` enable mountpoint related tests, but this is already tested at line 265... is it expected ?

.azure-pipelines.yml
```diff
@@ -253,22 +253,26 @@ jobs:
...
   - bash: |
       set -eux
       py.test $(pytest.base_args) \
         tests --runmountpoint --runslow -m mountpoint -x \
         --junitxml=test-results-mountpoint.xml
     displayName: 'Tests mountpoint'
+  - bash: |
+      set -eux
+      py.test $(pytest.base_args) \
+        tests/backend tests/test_cli.py --postgresql --runslow \
```
touilleMan, line 272: why testing also `tests/test_cli.py` with postgresql enabled ? (we normally only tests `tests/backend` with postgresql)

tests/test_scripts.py
```diff
@@ -69,7 +69,10 @@ def run_testenv():
 
 
 @pytest.mark.slow
-@pytest.mark.skipif(sys.platform == "win32", reason="causes a freeze in appveyor for some reasons")
+@pytest.mark.skipif(
+    "linux" not in sys.platform,
+    reason="causes a freeze in appveyor for some reasons, and crashes on Catalina",
```
touilleMan, line 74: ah !

.azure-pipelines.yml
Ironicbay: I couldn't figure if there was a difference with the `-m mountpoint` option in line 266. If it's redundant I will remove the mountpoint tests then

.azure-pipelines.yml
Ironicbay: I copied the lines from the linux tests (180th line). Should we fix it on both ?

tests/test_scripts.py
Ironicbay: It works fine on Big Sur though, there has always been issues with the test_env on Catalina. I guess the `skipif` will remain until we can run tests on a 11.x image

.azure-pipelines.yml
Ironicbay: Nevermind, I get it now

.azure-pipelines.yml
touilleMan: @vxgmichel @Max-7 Any idea why we enable postgresql in `test_cli.py` ?

touilleMan (approved):
all good ^^

PR #1759, opened 2021-06-07
Reword telemetry enable dialogue

### LINKED ISSUE #1745: Reword the Error reporting enabling feature
image

`Error reporting` is a very aggressive term (the user has the feeling something when wrong and this is an error message !)

The title (and the enable button) should be `Enable Telemetry`.

On top of that the `No`  button should be in grey instead of blue to indicate it is the not recommended option ;-)

touilleMan:
fix #1745


parsec/core/gui/custom_dialogs.py
```diff
@@ -191,20 +191,42 @@ def get_text_input(
...
             b.clicked_self.connect(self._on_button_clicked)
             b.setCursor(Qt.PointingHandCursor)
-            if radio_mode:
-                self.layout_radios.addWidget(b)
-            else:
-                self.layout_buttons.insertWidget(1, b)
+            b.setStyleSheet(
+                "QPushButton {background-color: darkgrey;} QPushButton:hover {background-color: grey;}"
+            )
+            b.apply_style()
```
Max-7, line 212: apply_style() should not be needed.

parsec/core/gui/tr/parsec_fr.po
```diff
@@ -1641,22 +1641,21 @@ msgstr "L'URL n'est pas valide."
...
 msgstr ""
-"PARSEC peut automatiquement envoyer des données quand une erreur est "
-"rencontrée. Cette information nous aide à améliorer votre expérience "
-"utilisateur.<br /><br />\n"
-"\n"
-"Autorisez-vous PARSEC à envoyer automatiquement des rapports d'erreur ? "
-"Vous pourrez désactiver cette fonctionnalité dans la configuration."
+"La télémétrie est utilisée dans Parsec pour automatiquement envoyer "
+"des rapports d'erreur anonymes.<br />\n"
+"Ces information nous aide à améliorer votre expérience utilisateur.<br />\n"
```
Max-7, line 1651: `Ces informationS nous aideNT`

parsec/core/gui/tr/parsec_fr.po
```diff
@@ -2464,11 +2463,11 @@ msgstr "Vérifier la disponibilité d'une nouvelle version au démarrage"
...
 
-msgid "TEXT_SETTINGS_ERROR_REPORTING_TITLE"
-msgstr "Rapport d'erreur"
+msgid "TEXT_SETTINGS_TELEMETRY_TITLE"
+msgstr "Télémétrie"
 
-msgid "TEXT_SETTINGS_CHECK_SEND_CRASH_REPORT"
-msgstr "Envoyer les rapports d'erreur"
+msgid "TEXT_SETTINGS_ENABLE_TELEMETRY"
+msgstr "Activer la télémétrie pour l'envoie automatique de rapports d'erreur"
```
Max-7, line 2470: `envoi`

parsec/core/gui/custom_dialogs.py
```diff
@@ -191,20 +191,42 @@ def get_text_input(
...
+            # Add "yes" button
+            b = Button(yes_text)
             b.clicked_self.connect(self._on_button_clicked)
             b.setCursor(Qt.PointingHandCursor)
-            if radio_mode:
-                self.layout_radios.addWidget(b)
-            else:
-                self.layout_buttons.insertWidget(1, b)
+            b.setStyleSheet(
+                "QPushButton {background-color: darkgrey;} QPushButton:hover {background-color: grey;}"
```
Max-7, line 210: A bit worried that it makes it look like a disabled button. It should be ok in this context since I don't think there's a case where a button in the question dialog can be disabled.

Max-7 (changes requested):

parsec/core/gui/custom_dialogs.py
touilleMan: I think its clear enough given the button change when it is hovered (which wouldn't be the case for a disabled button)

parsec/core/gui/tr/parsec_fr.po
touilleMan: 👍

parsec/core/gui/tr/parsec_fr.po
touilleMan: 👍

parsec/core/gui/custom_dialogs.py
touilleMan: 👍

Max-7 (approved):

PR #1762, opened 2021-06-11
Add sec-websocket-protocol header support in websocket handshake

touilleMan:


touilleMan:
won't be needed given we are switching toward pure http

PR #1766, opened 2021-06-16
Organization users limit

vamonte:


parsec/backend/memory/user.py
```diff
@@ -37,12 +38,16 @@ def __init__(self, send_event, *args, **kwargs):
...
-        pass
+        self._organization_component = other_components["organization"]
 
     async def create_user(
         self, organization_id: OrganizationID, user: User, first_device: Device
     ) -> None:
         org = self._organizations[organization_id]
+        users_limit = self._organization_component._organizations[organization_id].users_limit
+
+        if users_limit and users_limit <= len(org.users):
```
touilleMan, line 49: is the limit on all users or only on active users ?

parsec/backend/organization.py
```diff
@@ -94,17 +99,23 @@ async def api_organization_create(self, client_ctx, msg):
...
                 expiration_date=expiration_date,
+                users_limit=default_users_limit,
             )
 
         except OrganizationAlreadyExistsError:
             return {"status": "already_exists"}
 
-        rep = {"bootstrap_token": bootstrap_token, "status": "ok"}
+        rep = {
+            "users_limit": default_users_limit,
```
touilleMan, line 115: I don't think returning `users_limit` is needed (the caller will call `organization_status` when it needs this information, and adding this information open the door to add all the other organization configuration stuff...)

parsec/backend/organization.py
```diff
@@ -328,6 +340,8 @@ async def api_organization_bootstrap(self, client_ctx, msg):
         except (OrganizationNotFoundError, OrganizationInvalidBootstrapTokenError):
             return {"status": "not_found"}
 
+        organization = await self.get(client_ctx.organization_id)
```
touilleMan, line 343: Same as for `organization_create`: the client is used to call `organization_stats` whenever `users_limit` info is needed

On top of that it is a bit dangerous to do this call here because we can endup in a situation where `self.bootstrap` succeeded, but `self.get` failed (e.g. PostgreSQL DB entered in maintenance in the meantime), then returning the wrong status to the client

Not we shouldn't have this trouble with `await self.webhooks.on_organization_bootstrap` given in case of error, this function is designed to log the error and return without exceptions.

...

Now that I've written this explanation, I'm kind of sad it is not in the codebase as a comment... maybe you could add it for me ? 😄

parsec/backend/postgresql/organization.py
```diff
@@ -99,14 +99,32 @@
...
+_q_check_users_limit = Q(
+    f"""
+    SELECT
+        organization._id
+    FROM organization
+    LEFT JOIN
+        user_ on user_.organization = organization._id
+    WHERE
+        organization = { q_organization_internal_id("$organization_id") }
+    GROUP BY organization._id
```
touilleMan, line 111: Why the `GROUP BY` ? the `WHERE` condition should have left us with only one value for `organization._id`

parsec/backend/postgresql/user.py
```diff
@@ -46,6 +48,9 @@ async def create_user(
         self, organization_id: OrganizationID, user: User, first_device: Device
     ) -> None:
         async with self.dbh.pool.acquire() as conn:
```
touilleMan, line 50: A transaction block is missing here

In fact the transaction block is already surrounding the `query_create_user` (that's what does the `@query(in_transaction=True)` decorator)

So I think the user limit check should be pushed into `query_create_user` 

The rational here is:
- `PGUserComponent` only implements the expected interface and calling `query_xxx`. The connection object is also created here given we want the `query_xxx` functions to be stateless.
- `query_xxx` is the actual implementation of the feature. It sets up the transaction context (given a feature operation should atomic)
- `query_xxx` makes use of multiple query functions (such as `send_signal` and all the `_q_yyy` instances of `Q` class), those query function should be combined together so they require a connection object and never deal with the transaction state

parsec/backend/user.py
```diff
@@ -530,6 +534,8 @@ async def _api_user_create(self, client_ctx, msg):
 
         except UserAlreadyExistsError as exc:
             return {"status": "already_exists", "reason": str(exc)}
+        except UserLimitReached:
+            return {"status": "not_allowed", "reason": "User's limit reached"}
```
touilleMan, line 538: `User's limit reached` ? shouldn't it be `Users limit reached` ?

poke @Max-7 who knows how to spell "retreive" :trollface:

parsec/core/gui/tr/parsec_en.po
```diff
@@ -2490,3 +2490,6 @@ msgstr "Create a new workspace"
 
 msgid "ACTION_WORKSPACE_GOTO_FILE_LINK"
 msgstr "Go to a file using a link"
+
+msgid "USERS_LIMIT_REACHED"
+msgstr "User's limit reached"
```
touilleMan, line 2495: same as above

parsec/core/gui/users_widget.py
```diff
@@ -313,10 +313,13 @@ def add_user(self, user_info, is_current_user):
...
         button.show()
 
-    def add_user_invitation(self, email, invite_addr):
+    def add_user_invitation(self, email, invite_addr, greet_disabled):
         button = UserInvitationButton(email, invite_addr)
         self.layout_users.addWidget(button)
         button.greet_clicked.connect(self.greet_user)
+        button.setDisabled(greet_disabled)
+        if greet_disabled:
+            button.setToolTip(_("USERS_LIMIT_REACHED"))
```
touilleMan, line 322: Given the message is about the user limit, shouldn't the parameter passed to this function be also named after that ? (i.e. there is no other reasons for greet to be disabled, and if in the future there is, we will end up with an incorrect tool tip)

parsec/backend/postgresql/organization.py
```diff
@@ -99,14 +99,32 @@
...
+    HAVING users_limit is NULL OR count(user_) < users_limit;
+"""
```
touilleMan, line 113: The point of this query is to return a whether we've reached the user limit or not, I think might be able to simplify this into:
```SQL
SELECT
    organization.users_limit is NULL
    OR (
        SELECT count(*) FROM user_ WHERE user_.organization = organization._id
    ) < organization.users_limit
FROM organization
WHERE organization.organization_id = $organization_id
```

(I didn't try this query, so I might be completely broken 😄 )

parsec/backend/postgresql/organization.py
```diff
@@ -99,14 +99,32 @@
...
 
+_q_check_users_limit = Q(
+    f"""
+    SELECT
+        organization._id
+    FROM organization
+    LEFT JOIN
+        user_ on user_.organization = organization._id
+    WHERE
+        organization = { q_organization_internal_id("$organization_id") }
```
touilleMan, line 110: `organization.organization_id = $organization_id` should do the trick here (that's basically what `q_organization_internal_id` does, but with an aditional `SELECT` we don't need here given we already have our hands on the organization table)

touilleMan (changes requested):

parsec/api/protocol/organization.py
```diff
@@ -52,6 +53,7 @@ class OrganizationBootstrapWebhookSchema(BaseSchema):
     device_label = fields.String(allow_none=True, required=True)
     human_email = fields.String(allow_none=True, required=True)
     human_label = fields.String(allow_none=True, required=True)
+    users_limit = fields.Integer(allow_none=True, required=True)
```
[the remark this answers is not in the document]
vamonte: The expiration date is always set by the SaaS instead of the user's limit that is set by the backend. The SaaS need to know the limit to send its mails.

parsec/api/protocol/organization.py
```diff
@@ -52,6 +53,7 @@ class OrganizationBootstrapWebhookSchema(BaseSchema):
     device_label = fields.String(allow_none=True, required=True)
     human_email = fields.String(allow_none=True, required=True)
     human_label = fields.String(allow_none=True, required=True)
+    active_users_limit = fields.Integer(allow_none=True, required=True)
```
touilleMan, line 56: following https://github.com/Scille/parsec-cloud/pull/1766#discussion_r657746730

> The expiration date is always set by the SaaS instead of the user's limit that is set by the backend. The SaaS need to know the limit to send its mails.

The default user limit is configured in the backend once, so the SAAS should know about it.

I'm reluctant to add this field given it would mean we should also add `user_profile_outsider_allowed` (given it is also a per-organization configuration set by the backend) and later on all the other configuration needed...

I think it would be simpler to consider the default values known by the SAAS (of course this mean if the default values evolve, the SAAS should be re-deployed in sync with the backend) or just to have the SAAS send an `organization_status` command to the backend every time it receive a webhook (the only complexity here is to make sure we handle errors such as backend not available by logging the content of the webhook so we can manually retry later)

parsec/api/protocol/organization.py
```diff
@@ -12,6 +12,7 @@ class APIV1_OrganizationCreateReqSchema(BaseReqSchema):
 class APIV1_OrganizationCreateRepSchema(BaseRepSchema):
     bootstrap_token = fields.String(required=True)
     expiration_date = fields.DateTime(allow_none=True, required=False)
+    active_users_limit = fields.Integer(allow_none=True, required=True)
```
touilleMan, line 15: I don't think we should add this here (given how easy it is to call `organization_status` to have this information and all the other configuration)

(of course the same rational can be used against the `expiration_date` field... but this one is already there :trollface:  and we never use it !)

parsec/backend/organization.py
```diff
@@ -202,6 +213,7 @@ async def apiv1_administration_organization_stats(self, client_ctx, msg):
     @catch_protocol_errors
     async def api_organization_update(self, client_ctx, msg):
         msg = apiv1_organization_update_serializer.req_load(msg)
+        print(msg)
```
touilleMan, line 216: debug stuff to remove

parsec/backend/organization.py
touilleMan: ping !

parsec/backend/cli/run.py
```diff
@@ -332,6 +332,12 @@ def handle_parse_result(self, ctx, opts, args):
 organization_id, device_id, device_label (can be null), human_email (can be null), human_label (can be null)
 """,
 )
+@click.option(
+    "--organization-default-users-limit",
```
touilleMan, line 336: Sorry to be nerdy about naming, but I can't help myself ! 😄 

First, the name should contain the fact that we only count non-revoked users (so I guess `--organization-default-active-users-limit`)

Second, I don't think "default" is the good word here:
- it makes confusing to talk about the default value of the this parameter (i.e. there is no limit on active users if this variable is not configured)
- a "default" implied it could be overwritten. However this is something that can't be done during organization creation (the `organization_create` command doesn't have a field to change that, and the spontaneous bootstrap doesn't allow that for obvious reason ^^). So this option can be modified later on, but it initial value is always the one configured here.

Hence I would call it `--organization-initial-active-users-limit` ;-)

And this way we can have:
```python
    help="Non-revoked users limit used to configure newly created organizations (default: no limit)",
```

tests/backend/test_apiv1_organization_status.py
```diff
@@ -23,7 +23,16 @@ async def organization_status(sock, organization_id):
...
+async def organization_config(sock):
+    raw_rep = await sock.send(
+        organization_config_serializer.req_dumps({"cmd": "organization_config"})
+    )
+    raw_rep = await sock.recv()
+    return organization_config_serializer.rep_loads(raw_rep)
+
+
 async def organization_update(sock, organization_id, **fields: Dict):
+    print(fields)
```
touilleMan, line 35: debug print

touilleMan (changes requested):

parsec/backend/postgresql/organization.py
vamonte: @touilleMan I understand your comment about the aim of this query but I don't know if it's simpler to use one subquery instead of one HAVING condition.

parsec/backend/postgresql/organization.py
vamonte: The GROUP BY is mandatory if we use the HAVING condition

parsec/api/protocol/organization.py
vamonte: @touilleMan Ok, I will set the default value in the SAAS.

touilleMan (approved):
lgtm
image

touilleMan:
https://www.youtube.com/watch?v=P3ALwKeSEYs

PR #1768, opened 2021-06-17
Add the outsider profile's count in the organization_stats response

vamonte:


parsec/backend/memory/organization.py
```diff
@@ -103,7 +104,13 @@ async def stats(self, id: OrganizationID) -> OrganizationStats:
             if vlob_organization_id == id:
                 data_size += blockmeta.size
 
-        users = len(self._user_component._organizations[id].users)
+        users = 0
```
touilleMan, line 107: out of curiosity I was wondering how to keep this without a for loop (you know... because.... reasons... 😄 )

```python
from functools import reduce
stats = reduce(lambda c, user: {'users': c['users'] + 1, 'outsiders': c['outsiders'] + int(user.profile == UserProfile.OUTSIDER)}, data, {'users': 0, 'outsiders': 0})
```
:/

So in the end I think the for loop is much more readable here !

parsec/backend/postgresql/organization.py
```diff
@@ -61,15 +62,21 @@
...
 
-
 _q_get_stats = Q(
     f"""
 SELECT
     (
         SELECT COUNT(*)
         FROM user_
-        WHERE user_.organization = { q_organization_internal_id("$organization_id") }
+        WHERE user_.organization = { q_organization_internal_id("$organization_id") } AND
```
touilleMan, line 71: We can use array and tuple to fetch profile and revocation date in a single query:
```sql
SELECT ARRAY(SELECT (revoked_on, profile) FROM user_ WHERE organization = 42);
```

This would allow us to crunch better stats:
```python
{
  ...
  "users": ...,
  "active_users": ...,
  "users_per_profile_detail": {
     "ADMINISTRATOR": {"active": ..., "revoked": ...},
     "STANDARD": {"active": ..., "revoked": ...},
     "OUTSIDER": {"active": ..., "revoked": ...},
  }
}
```

parsec/backend/postgresql/organization.py
```diff
@@ -61,15 +62,21 @@
 """
 )
 
-
 _q_get_stats = Q(
     f"""
 SELECT
     (
         SELECT COUNT(*)
         FROM user_
```
touilleMan, line 70: So user stats also count revoked users, isn't it a trouble for the SAAS ?

touilleMan (changes requested):

parsec/backend/postgresql/organization.py
vamonte: Nope

parsec/api/protocol/organization.py
```diff
@@ -65,6 +65,8 @@ class OrganizationStatsRepSchema(BaseRepSchema):
     data_size = fields.Integer(required=True)
     metadata_size = fields.Integer(required=True)
     users = fields.Integer(required=True)
+    active_users = fields.Integer(required=True)
+    users_per_profile_detail = fields.Dict(required=True)
```
touilleMan, line 69: `Dict` field is too broad, we should better specify the structure:
```python
class UsersPerProfileDetailItem(BaseSchema):
    profile = UserProfileField(required=True)
    active = fields.Integer(required=True)
    revoked = fields.Integer(required=True)

[...]
users_per_profile_detail = fields.List(UsersPerProfileDetailItem, required=True)
```

The rational for using a list of structs here is it's complex to define a dict with only user profile values as keys 

Having the profile value within the struct also allows for specialization (for instance we can have a `UserPerAdminProfileDetail` which define additional fields compared to the default `UserPerProfileDetail`).


Another solution would be to define a `Nested` class with each user profile as field, this would have the added benefit over the List approach of guaranteeing they won't be any user profile duplication:

```python
class UsersPerProfileDetailItem(BaseSchema):
    profile = UserProfileField(required=True)
    active = fields.Integer(required=True)
    revoked = fields.Integer(required=True)

class UsersPerProfileDetail(BaseSchema):
    ADMIN = fields.Nested(UsersPerProfileDetailItem, required=True)
    STANDARD = fields.Nested(UsersPerProfileDetailItem, required=True)
    OUTSIDER = fields.Nested(UsersPerProfileDetailItem, required=True)

[...]
users_per_profile_detail = fields.Nested(UsersPerProfileDetail, required=True)

```

However with this approach the returned `users_per_profile_detail` dict keys are `str` instead of `UserProfile` which is unexpected.

So in the end I prefer the list-based approach

@vxgmichel what do you think ?

parsec/api/protocol/organization.py
```diff
@@ -81,6 +83,8 @@ class APIV1_OrganizationStatsRepSchema(BaseRepSchema):
     metadata_size = fields.Integer(required=True)
     users = fields.Integer(required=True)
     workspaces = fields.Integer(required=True)
+    active_users = fields.Integer(required=True)
+    users_per_profile_detail = fields.Dict(required=True)
```
touilleMan, line 87: same as above

parsec/backend/memory/organization.py
```diff
@@ -103,7 +103,16 @@ async def stats(self, id: OrganizationID) -> OrganizationStats:
...
+        active_users = 0
+        users_per_profile_detail = self._init_users_per_profile_detail()
```
touilleMan, line 108: `_init_users_per_profile_detail` shouldn't be declared in the parent class: we should try to limit the parent class to api handling and interface definitions.
Here given the helper is rather short and simple (i.e. `detail = {p.value: {"active": 0, "revoked": 0}for p in UserProfile}`), it's just more readable to keep it inlined ;-)

parsec/backend/organization.py
```diff
@@ -79,7 +79,9 @@ class OrganizationStats:
     data_size: int
     metadata_size: int
     users: int
+    active_users: int
     workspaces: int
+    users_per_profile_detail: Dict[str, Counter]
```
touilleMan, line 84: key should be of type `UserProfile`

touilleMan: See my comment about `organization_stats` command schema: we'd better use more explicit types here

```python
@attr.s(slots=True, frozen=True, auto_attribs=True)
class UsersPerProfileDetailItem:
    active: int
    revoked: int

@attr.s(slots=True, frozen=True, auto_attribs=True)
class OrganizationStats:
    [...]
    users_per_profile_detail: List[UsersPerProfileDetailItem]

```

parsec/core/types/organizations.py
```diff
@@ -1,13 +1,16 @@
...
+from typing import Dict, Counter
 
 
 @attr.s(frozen=True, slots=True, auto_attribs=True)
 class OrganizationStats:
     users: int
+    active_users: int
     data_size: int
     metadata_size: int
+    users_per_profile_detail: Dict[str, Counter]
```
touilleMan, line 13: Same remark as `OrganizationStats` def in the backend ;-)

tests/backend/test_apiv1_organization_stats.py
```diff
@@ -19,6 +21,7 @@ async def organization_stats(sock, organization_id):
 
 
 @pytest.mark.trio
+@customize_fixtures(adam_profile=UserProfile.OUTSIDER)
```
touilleMan, line 24: I think we should split this `test_organization_stats` into:
- `test_organization_stats_data`: this test without the `customize_fixture` (given we don't care at all about users profiles here)
- `test_organization_stats_users`: a brand new test that creates 2 users of each type, check the stats, revoke user 1 user of each type, check the stats

tests/backend/test_organization_stats.py
```diff
@@ -6,15 +6,28 @@
 from pendulum import now as pendulum_now
 
 from parsec.core.backend_connection import BackendConnectionError
+from parsec.api.data import UserProfile
 from parsec.core.logged_core import OrganizationStats
+from tests.common import customize_fixtures
 
 
 @pytest.mark.trio
+@customize_fixtures(adam_profile=UserProfile.OUTSIDER)
```
touilleMan, line 15: same as for apv1 test

parsec/backend/postgresql/organization.py
```diff
@@ -61,15 +61,28 @@
...
-        WHERE user_.organization = { q_organization_internal_id("$organization_id") }
+        WHERE user_.organization = { q_organization_internal_id("$organization_id") } AND
+             user_.profile != '{ UserProfile.OUTSIDER.value }'
     ) users,
+    (
+        SELECT COUNT(*)
+        FROM user_
+        WHERE user_.organization = { q_organization_internal_id("$organization_id") } AND
+             user_.profile = '{ UserProfile.OUTSIDER.value }'
+    ) outsiders,
```
touilleMan, line 85: We don't need to crunch this `outsiders` and the `users` results column given the `SELECT (revoked_on, profile) ...` should give us all the data to retrieve everything ;-)

edit: `result['users']` and `result['outsiders']` are not used in the `stats` function, so those two results columns can safely be removed

parsec/backend/postgresql/organization.py
```diff
@@ -61,15 +61,28 @@
...
 _q_get_stats = Q(
     f"""
 SELECT
+    (
+        SELECT ARRAY(
+            SELECT (revoked_on, profile)
+            FROM user_
+            WHERE organization = { q_organization_internal_id("$organization_id") }
+        )
+    )_users,
```
touilleMan, line 73: What about `users` as result column name ? (given the other `users` result column is not needed anymore, see next comment)

touilleMan (changes requested):

parsec/backend/postgresql/organization.py
```diff
@@ -197,8 +197,30 @@ async def stats(self, id: OrganizationID) -> OrganizationStats:
...
             await self._get(conn, id)  # Check organization exists
             result = await conn.fetchrow(*_q_get_stats(organization_id=id))
+            users = 0
+            active_users = 0
+            users_per_profile_detail = {p: {"active": 0, "revoked": 0} for p in UserProfile}
+            for u in result["users"]:
+                revoked_on, profile = u
+                users += 1
+                if revoked_on:
+                    users_per_profile_detail[UserProfile[profile]]["revoked"] += 1
```
touilleMan, line 207: I didn't know you could do `UserProfile[profile]` and use a `str` as key, this is really cool 😃 

That mean all the `STR_TO_xxx` can be removed \o/
https://github.com/Scille/parsec-cloud/blob/99d94710cf868dafabdc106a830991e05d62e345/parsec/backend/postgresql/utils.py#L17-L24

parsec/core/types/organizations.py
```diff
@@ -1,13 +1,17 @@
...
+from parsec.api.protocol import UsersPerProfileDetailItem
 
 
 @attr.s(frozen=True, slots=True, auto_attribs=True)
 class OrganizationStats:
     users: int
+    active_users: int
     data_size: int
     metadata_size: int
+    users_per_profile_detail: List[UsersPerProfileDetailItem]
```
touilleMan, line 14: image

tests/backend/test_apiv1_organization_stats.py
```diff
@@ -70,3 +96,152 @@ async def test_organization_stats(
...
+    local_device_factory,
+):
+
+    async with backend_factory(populated=False) as backend:
+        binder = backend_data_binder_factory(backend)
+
+        org = organization_factory("IFD")
+        device = local_device_factory(
+            org=org,
+            base_device_id="godfrey@d1",
```
touilleMan, line 116: no need for `base_device_id` and `base_human_handle` here (this is only needed for tests doing user search)

tests/backend/test_apiv1_organization_stats.py
```diff
@@ -70,3 +96,152 @@ async def test_organization_stats(
...
+            {"profile": UserProfile.STANDARD, "active": 0, "revoked": 0},
+            {"profile": UserProfile.OUTSIDER, "active": 0, "revoked": 0},
+        ],
+        "data_size": 0,
+        "metadata_size": ANY,
+        "workspaces": 1,
+    }
+
+    nick1 = local_device_factory(
+        base_device_id="el_murcielago_enmascarado_ii@d1",
```
touilleMan, line 146: same as above: no need for custom device_id/human_handle

tests/backend/test_apiv1_organization_stats.py
```diff
@@ -70,3 +96,152 @@ async def test_organization_stats(
...
+        base_human_handle="Rodolfo Guzman Huerta",
+        profile=UserProfile.ADMIN,
+        org=org,
+    )
+    await binder.bind_device(nick1, certifier=godfrey1)
```
touilleMan, line 151: It might be more readable to use a loop here:
```python
expected_stats = {
        "status": "ok",
        "users": 1,
        "active_users": 1,
        "users_per_profile_detail": [
            {"profile": UserProfile.ADMIN, "active": 1, "revoked": 0},
            {"profile": UserProfile.STANDARD, "active": 0, "revoked": 0},
            {"profile": UserProfile.OUTSIDER, "active": 0, "revoked": 0},
        ],
        "data_size": 0,
        "metadata_size": ANY,
        "workspaces": 2,
    }

for profile in UserProfile:
    device = local_device_factory(profile=profile, org=org)
    await binder.bind_device(device, certifier=godfrey1, initial_user_manifest_in_v0=True)
    expected_stats["users"] += 1
    expected_stats["active_users"] += 1
    expected_stats["users_per_profile_detail"][profile]["active"] += 1
    stats = await organization_stats(sock, org.organization_id)
    assert stats == expected_stats

    await binder.bind_revocation(device.user_id, certifier=godfrey1)
    expected_stats["active_users"] -= 1
    expected_stats["users_per_profile_detail"][profile]["active"] -= 1
    expected_stats["users_per_profile_detail"][profile]["revoked"] += 1
    stats = await organization_stats(sock, org.organization_id)
    assert stats == expected_stats

# Also make sure stats are isolated between organizations
device = local_device_factory(org=otherorg, profile=UserProfile.ADMIN)
await binder.bind_organization(otherorg, device, initial_user_manifest_in_v0=True)
stats = await organization_stats(sock, org.organization_id)
assert stats == expected_stats
```

tests/backend/test_apiv1_organization_stats.py
```diff
@@ -70,3 +96,152 @@ async def test_organization_stats(
...
+
+        org = organization_factory("IFD")
+        device = local_device_factory(
+            org=org,
+            base_device_id="godfrey@d1",
+            base_human_handle="Godfrey Ho <godfrey.ho@ifd.com>",
+            profile=UserProfile.ADMIN,
+        )
+        with freeze_time("2000-01-01"):
+            await binder.bind_organization(org, device)
```
touilleMan, line 121: Can use `initial_user_manifest_in_v0=True` to avoid creating a workspace for the user manifest
And no need to freeze time here given we don't check the creation date ;-)

touilleMan (approved):

PR #1785, opened 2021-07-13
Bsl licensed backend

touilleMan:
Example of projects with similar license:

[MariaDB's MaxScale](https://github.com/mariadb-corporation/MaxScale/blob/88dea54851c25d8600e6c49beee4284b243e0cc9/LICENSE25.TXT)

[ZeroTierOne](https://github.com/zerotier/ZeroTierOne/blob/6faca86bb424d0b9643b6efa50571f73310d8276/LICENSE.txt)

[CouchBase](https://github.com/couchbase/kv_engine/blob/5d1a62e1493537a922b08bd29351826986fadd04/licenses/BSL-Couchbase.txt)

[CockroachDB](https://github.com/cockroachdb/cockroach/blob/b3b850e89a012e79203c79c93db637cb067a30cc/licenses/BSL.txt)

[Sentry](https://github.com/getsentry/sentry/blob/75ef01f6497d352dad9a7e30c9b9abc0b120422d/LICENSE)


misc/license_headers.py
```diff
@@ -9,40 +9,9 @@
...
-        if path.is_dir():
-            for f in chain(Path(path).glob("**/*.py"), Path(path).glob("**/*.sql")):
-                if need_skip(f):
-                    continue
-                yield f
-        elif path.is_file():
-            yield path
-        elif not path.exists():
-            raise SystemExit(f"Error: Path `{path}` doesn't exist !")
+THIS_YEAR = "2021"  # Change me when new year comes ;-)
```
vxgmichel, line 13: Why not use `date.today()` like below?

touilleMan: Because this would cause a CI breakage every January 1rst which requires modifying each file to solve :'(
So I think it's better to manually control when the change should occur

PR #1788, opened 2021-07-20
Change QR Code color and removed the logo

### LINKED ISSUE #1787: Remove logo and change color of QR code

Max-7:
Closes #1787


parsec/core/gui/qrcode_widget.py
```diff
@@ -21,21 +21,6 @@
 
 
 def generate_qr_code(text):
-    global PARSEC_LOGO
```
touilleMan, line 24: `PARSEC_LOGO` global var definition should be removed now that it is unused

parsec/core/gui/qrcode_widget.py
```diff
@@ -21,21 +21,6 @@
...
-
-    if not PARSEC_LOGO:
-        img = QImage(":/logos/images/logos/parsec2.png")
```
touilleMan, line 27: `:/logos/images/logos/parsec2.png` doesn't seem to be used anywhere else in the code base, so shouldn't this file be removed as well ?

touilleMan (changes requested):

parsec/core/gui/qrcode_widget.py
Max-7: Yep, it has been removed.

touilleMan (approved):

PR #1790, opened 2021-07-21
Replace STR_TO_XXX by Enum.__getitem__ in backend postgresql data fetching

touilleMan:


vamonte (changes requested):
Seems good to me but the CI is broken.

vamonte (approved):

PR #1793, opened 2021-07-22
Fix the GUI freeze in the invitation process on MacOS

Ironicbay:


parsec/core/gui/app.py
```diff
@@ -204,6 +204,11 @@ def emit(self, *args):
         win.show_top()
         win.new_instance_needed.emit(start_arg)
 
+        if sys.platform == "darwin":
```
touilleMan, line 207: A small comment would be welcome here
```
# macFUSE is not bundled with Parsec and must be manually installed by the user
# so we detect early such need to provide a help dialogue ;-)
# TODO: provide a similar mechanism on Windows&Linux to handle mountpoint runner not available
```

parsec/core/gui/instance_widget.py
```diff
@@ -44,6 +43,19 @@ async def _do_run_core(config, device, qt_on_ready):
                 qt_on_ready.emit(core, core_jobs_ctx)
 
 
+def macfuse_installation_widget(window):
```
touilleMan, line 46: A more explicit name would be welcome, like `ensure_macfuse_available_or_show_dialogue`

touilleMan (approved):

PR #1800, opened 2021-08-03
Improve api data compat check and introduce api protocol compat check

touilleMan:


vamonte (changes requested):
Seems good to me. These regular expressions could be tricky but I haven't another ideas to avoid them.
The CI is broken.

touilleMan:
> These regular expressions could be tricky

Yup I'm not really happy about them either.

The solution would be to change `parsec.api.protocol.cmds` dicts to enums, then to use those enums as `EnumCheckedConst` fields in the command serializers schema. This should make the data validation more robust (it would also open the way to also have enums for response status \o/), however it brings a much bigger change which I'd rather avoid for the moment.

tests/schemas/builder.py
```diff
@@ -48,8 +58,16 @@ def field_to_spec(field):
     # Handle nested/meta/special types
     if isinstance(field, CheckedConstant):
         spec["value"] = field.constant
-    if isinstance(field, EnumCheckedConstant):
+    elif isinstance(field, EnumCheckedConstant):
         spec["value"] = field.constant.value
+        spec["enum_type"] = type(field.constant).__name__
```
vxgmichel, line 63: Should this field be called `enum_constant_type` so it's not confused with the `enum_type` of enum fields?

tests/schemas/builder.py
```diff
@@ -117,14 +135,113 @@ def generate_core_data_specs():
...
+    cmd_serializers = {}
+    for item_name in dir(protocol_mod):
+        item = getattr(protocol_mod, item_name)
+        if not isinstance(item, CmdSerializer):
+            continue
+        # This is where things start to be hacky: given we don't use enum
+        # for command name, and given we sometime use the same schema for
+        # APIv1 and v2 definitions, we rely on a weak naming convention
+        # to do the identification...
+        match = re.match(r"^(?P<name>(apiv1_)?[a-z][a-z0-9_]+)_serializer$", item_name)
```
vxgmichel, line 188: So those regex define what is a valid name for different parts of the API? Maybe it would makes sense to define them together somewhere else with some comments and some examples of valid names to make it clearer.

vxgmichel (approved):
A few comments, LGTM otherwise

parsec/serde/fields.py
```diff
@@ -61,29 +61,33 @@
...
             raise ValidationError(f"Invalid role `{value}`")
 
-    return type(
-        f"{enum.__name__}Field", (Field,), {"_serialize": _serialize, "_deserialize": _deserialize}
-    )
+
+# Using inheritance to define enum fields allows for easy instrospection detection
+# which is useful when checking api changes in tests (see tests/schemas/builder.py)
+def enum_field_factory(enum):
+    return type(f"{enum.__name__}Field", (BaseEnumField,), {"ENUM": enum})
```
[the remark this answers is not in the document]
touilleMan: Yup ;-)

tests/schemas/builder.py
touilleMan: I think it's fine given 1) it's the same kind of data that goes in both case and 2) the `spec["type"]` field already discriminate (the value being either `EnumCheckedConstant` or `MyEnumField`)

vamonte (approved):

tests/schemas/builder.py
touilleMan: > So those regex define what is a valid name for different parts of the API?

yup, this is totally sub-optimal, but otherwise we would have to change plenty of code in the api.
The good think is we check naming for all `CmdSerializer`, so a wrong naming will be caught here (hence it should be obvious enough how to fix the naming) and given we also check unused `CmdSerializer` we should be also safe against name shadowing.

Note we are not defining a scheme for the api command, but for the variables implementing them. So I don't think it's not that bad to have this kind of hack ;-)

PR #1802, opened 2021-08-10
Fix backend error when user vlob_create with a non-existing realm id

touilleMan:


vxgmichel (approved):
So the fix is about returning a `not_allowed` status instead of `not_found` when a realm does not exist? What does it fix exactly? 

LGTM otherwise :+1:

PR #1805, opened 2021-08-12
Fix concurrent merge involving prevent sync entry

touilleMan:


tests/core/fs/workspacefs/test_sync_transactions.py
```diff
@@ -106,6 +106,60 @@ def test_merge_folder_manifests():
...
+
+    # Now merging should detect the duplication
+    merged_manifest = merge_manifests(
+        local_author=my_device,
+        prevent_sync_pattern=prevent_sync_pattern,
+        local_manifest=local_manifest,
+        remote_manifest=remote_manifest_v2,
+        force_apply_pattern=False,
+    )
+    assert merged_manifest.children
```
touilleMan, line 160: test is not finished !

Now that the merge is ok, we should check the resulting manifest

parsec/core/types/manifest.py
```diff
@@ -547,11 +554,17 @@ def _restore_local_confinement_points(
...
             return self
+        # Create a set for fast lookup in order to make sure no entry gets duplicated.
+        # This might happen when a synchronized entry is renamed to a confined name locally.
+        self_entry_ids = set(self.children.values())
         previously_local_confinement_points = {
             name: entry_id
             for name, entry_id in other.children.items()
-            if entry_id in other.local_confinement_points
-            or entry_id in self.remote_confinement_points
+            if entry_id not in self_entry_ids
```
touilleMan, line 563: So if an entry has been renamed in local to a confined name, we revert the rename here ? (even if the remote change that triggered the sync is not related to the entry, typically a new entry added to the folder manifest)

vxgmichel: The rename is not reverted here, this method is mostly used to create `local_from_remote`, i.e a local manifest that corresponds to the remote once the local context is applied (but without merging with the content of the local manifest). Then this `local_from_remote` is merged with the `local` manifest and potential conflicts get resolved.

In the case we're interested in, this should happen:
- `local_from_remote` contains some changes 
- `local_manifest` contains a rename to a confinement point
- if the changes in `local_from_remote` do not include a rename of the sensitive entry, the merge gets resolved without conflict, and the new manifest includes both the remote changes and the rename to confinement.
- if `local_from_remote` includes a rename of the sensitive entry, the merge finds a conflict that gets resolved like other similar conflicts (in this case renaming the entry with a special suffix indicating what happened)

tests/core/fs/workspacefs/test_sync_transactions.py
vxgmichel: 071d0f49e5e87811f91bb8b57d7dbb9a0f355d27

parsec/core/types/manifest.py
touilleMan: 👍

vxgmichel (approved):

PR #1810, opened 2021-08-17
Remove APIv1 & replace organization's expirate_date & introduce the REST API \o/

touilleMan:
image


parsec/api/protocol/base.py
```diff
@@ -32,11 +30,11 @@ class MessageSerializationError(SerdePackingError, ProtocolError):
     pass
 
 
-def packb(data: Dict[str, object]) -> bytes:
+def packb(data: Mapping) -> bytes:  # type: ignore[type-arg]
     return _packb(data, MessageSerializationError)
 
 
-def unpackb(data: bytes) -> Dict[str, object]:
+def unpackb(data: bytes) -> dict:  # type: ignore[type-arg]
```
vxgmichel, line 37: Why are those changes necessary?

parsec/backend/postgresql/factory.py
```diff
@@ -51,23 +51,29 @@ async def _send_event(
...
+        "invite": invite,
+        "message": message,
+        "realm": realm,
+        "vlob": vlob,
+        "ping": ping,
+        "block": block,
+        "blockstore": blockstore,
+    }
+    for component in components.values():
+        method = getattr(component, "register_components", None)
```
vxgmichel, line 69: So `register_components` so different components can know about one another? And so far it's only used for `http` to know about `organization`?

vxgmichel (approved):
LGTM, only a couple of questions :+1:

parsec/backend/postgresql/factory.py
touilleMan: yes, because so far the postgresql components rely on SQL requests to access all the data (i.e. unlike the memory implementation, no component owned the underlying data).

However the administration api implemented in the `http` component is very closely related to the `organization` component.
This is because `http` component is related to a way to access data (i.e. through HTTP, hence the REST API being put there), where all the others components provide way to manipulate data and way to access data through the websocket API.

In theory we should have instead:
- a class exposing way to manipulate data
- a class exposing websocket API (maybe even a class per authentication type !)
- a class exposing HTTP API
- `http` module would only contain a class exposing the static HTTP routes
- `organization` module would contain all the three class types

However this seems too cumbersome to implement for the moment given how small the REST API currently is, so using the component registration system seems a much simpler solution ;-)

parsec/api/protocol/base.py
touilleMan: Because `unpackb` can return dict with non-str as keys, so it should be `Dict[object, object]`, however it seems `str` is not considered `object` by mypy when used as argument :/

```
parsec/backend/app.py:395: error: Argument 1 to "packb" has incompatible type "Dict[str, str]"; expected "Mapping[object, object]"  [arg-type]
```

So we endup with a `def packb(data: Mapping) -> bytes:` signature style, which is still not accepted by mypy
```
parsec/api/protocol/base.py:33: error: Missing type parameters for generic type "Mapping"  [type-arg]
```

Last step is to suppress this error with the `# type: ignore[type-arg]`

The funny thing is `packb`/`unpackb` from `api.protocol.base` module only call `packb`/`unpackb` from `serde.packing` module which have unparametrized `Mapping` and `dict` as param/return values (but mypy is not as strict in this module)

So in the end I got bored of pretending typing has any value in those functions (given we manipulate arbitrary dicts and pass them to a function that specify it wants arbitrary dicts ^^), so I just `type: ignore` the hell out of this :smile:

PR #1811, opened 2021-08-18
Implement a robust way to deal with users with slightly shifted clocks

### LINKED ISSUE #1738: Causality issue with parsec timestamps
A specific issue: #1681
--------------

Two kind of timestamps are involved in this problem:

#### 1. The timestamp in uploaded manifests follows the following scenario:
 - A device wants to upload a new or updated manifests
 - The device restamps the manifest to its current clock
 - The device encrypts the manifest as a vlob
 - The device sends the vlob to the backend along with the exact manifest timestamp in clear
 - The backend receives the request
 - The backend checks that the timestamp is in the ballpark of its current clock (+/- 30 minutes)
 - The backend checks that the corresponding user has write access to the corresponding realm
 - The backend accepts the request and update the database with the vlob and the timestamp
 - The backend notifies other clients that a new manifest is available

#### 2. The timestamp in the role certificate
 - A device wants to change the role of another user for a given realm
 - The device creates the corresponding realm role certificate
 - The certificate is signed and timestamped with the current device clock
 - The device sends the certificate to the backend
 - The backend receives the request
 - The backend checks that the timestamp is in the ballpark of its current clock (+/- 30 minutes)
 - The backend validates the role certificate and add it to the trustchain
 - The backend notifies other clients that the trust chain needs to be updated

The ballpark check being done by checking the absolute difference between the given timestamp and the current clock, there is no guarantee that two events that happen in a similar "ballpark" time window will have their timestamps matching the order in which they occurred. In short, when two events happen in the same 30 minutes window, there is no way to reliably tell which one has been processed first.

When a client downloads the manifest, it's going to compare its timestamp to the timestamps in the different role certificates. Since order is not guaranteed, the client might conclude that the client had no right to write the corresponding realm at that time, event though the backend was completely legitimate in accepting it.

vxgmichel:
Fix #1738

### DIFF
--- a/newsfragments/1811.bugfix.rst
+++ b/newsfragments/1811.bugfix.rst
@@ -1,0 +1,1 @@
+Fix sporadic errors that might occur when clients with slightly different clocks perform actions sur as reading/writing to workspaces and changing roles concurrently.
--- a/parsec/api/protocol/base.py
+++ b/parsec/api/protocol/base.py
@@ -70,6 +70,13 @@ class ErrorRepSchema(BaseRepSchema):
     errors = fields.Dict(allow_none=True)
 
 
+class RequireGreaterTimestampRepSchema(BaseRepSchema):
+    status: fields.CheckedConstant = fields.CheckedConstant(
+        "require_greater_timestamp", required=True
+    )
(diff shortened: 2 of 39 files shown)


touilleMan (approved):
image

PR #1817, opened 2021-08-27
Remove useless expiration_date field in organization_config (if org i…

touilleMan:
…s expired you cannot execute this command anyway !)


vxgmichel (approved):
Looks very straightforward, :+1:

PR #1822, opened 2021-08-30
Fix macos/Windows file open

Max-7:


vxgmichel:
I think it would be more consistent with the existing code base to implement this fix using the trio job scheduler and `trio.to_thread.run_sync()`:
```python
async def _open_files(paths):
    status = True
    for path in paths:
        status &= await trio.to_thread.run_sync(
            lambda: QDesktopServices.openUrl(QUrl.fromLocalFile(QFileInfo(path).absoluteFilePath())) 
        )
    return status

[...]

self.jobs_ctx.submit_job(
    self._open_files_success,
    self._open_files_error,
    _open_files,
    paths,
)
```

vxgmichel (approved):
LGTM :+1:

PR #1827, opened 2021-09-02
Filter invitation and revoked users

### LINKED ISSUE #1744: Add a way to hide workspaces from the GUI
This would be useful when the user has a lot of workspaces, especially when some of them are not used anymore (kind of a palliative for the missing remove workspace feature ^^)

AntoineDupre:
fix #1744

### DIFF
--- a/parsec/core/gui/tr/parsec_fr.po
+++ b/parsec/core/gui/tr/parsec_fr.po
@@ -2532,6 +2532,12 @@ msgstr "Filtrer les utilisateurs"
 msgid "ACTION_FILTER_LIST_USERS"
 msgstr "Rechercher utilisateurs"
 
+msgid "ACTION_FILTER_REVOKED_USERS"
+msgstr "Masquer les utilisteurs révoqués"
+
+msgid "ACTION_FILTER_PENDING_INVITATION"
+msgstr "Masquer les invitations"
+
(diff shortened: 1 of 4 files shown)


parsec/core/gui/tr/parsec_fr.po
```diff
@@ -2532,6 +2532,12 @@ msgstr "Filtrer les utilisateurs"
 msgid "ACTION_FILTER_LIST_USERS"
 msgstr "Rechercher utilisateurs"
 
+msgid "ACTION_FILTER_REVOKED_USERS"
+msgstr "Masquer les utilisteurs révoqués"
```
Max-7, line 2536: Missing a letter.

Max-7 (changes requested):
filter-users-layout-padding

Could you add some spacing to each layout so that elements are not so close to each other?

There's also a lot of blinking taking place, which is a bit weird because buttons and checkboxes are also bliking, not just the list of users.

AntoineDupre:
> filter-users-layout-padding
> 
> Could you add some spacing to each layout so that elements are not so close to each other?
> 
> There's also a lot of blinking taking place, which is a bit weird because buttons and checkboxes are also bliking, not just the list of users.

fixed

Max-7:
users_filtering

PR #1832, opened 2021-09-06
Changed use of tempfile in run_testenv.sh

alexandrediasldev:


touilleMan (approved):

tests/scripts/run_testenv.fish
```diff
@@ -1,9 +1,4 @@
 #!/usr/bin/env fish
-: '
-Create a temporary environment and initialize a test setup for parsec.
-
-Run `tests/scripts/run_testenv.fish --help` for more information.
-'
```
vxgmichel, line 6: Why did we loose this glorious docstring?

touilleMan: Do you have any idea how to do docstring in fish ?

Any time I want to `set/unset` a variable or declare a function I'm banging my head against the crazy "fish is not a sh shell, hence we choose to do this in our own (not better, not worst, just different) way" -__-'''

Live's too short, I'm not going to look fish documentation on how to do docstring (and I'm not turning this into a comment either ^^)

touilleMan: but on the other hand fish default configuration is super convenient (autocomplete, advanced globbing etc.)
I guess I'm just waiting for a ripbash :trollface:

PR #1829, opened 2021-09-04
Fix sync monitor freeze on placeholders regression

touilleMan:

PR #1834, opened 2021-09-06
Fix behavior of memory implementation of vlob's maintenance_save_reen…

touilleMan:
…cryption_batch whith invalid id in batch

based on #1829

### DIFF
--- a/parsec/backend/vlob.py
+++ b/parsec/backend/vlob.py
@@ -274,7 +274,7 @@ class BaseVlobComponent:
                 {"status": "not_allowed"}
             )
 
-        # Not need to catch VlobNotFoundError given unknown vlob/version in batch are ignored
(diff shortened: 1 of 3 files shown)


parsec/backend/vlob.py
```diff
@@ -274,7 +274,8 @@ async def api_vlob_maintenance_save_reencryption_batch(self, client_ctx, msg):
                 {"status": "not_allowed"}
             )
 
-        except (VlobNotFoundError, VlobRealmNotFoundError) as exc:
+        # Not need to catch VlobNotFoundError given unknown vlob/version in batch are ignored
```
vxgmichel, line 277: Typo here, `No need [...]`

vxgmichel (approved):
LGTM :+1:

PR #1835, opened 2021-09-07
Improve the tests of the merge_folder_children method to check some s…

vamonte:
…pecial cases


parsec/core/fs/workspacefs/sync_transactions.py
```diff
@@ -43,12 +43,16 @@ def get_filename(manifest: LocalFolderishManifests, entry_id: EntryID) -> Option
...
 
 def get_conflict_filename(
-    filename: EntryName, filenames: List[EntryName], author: DeviceID
+    filename: EntryName, filenames: List[EntryName], content_conflict: bool = False
 ) -> EntryName:
     counter = count(2)
-    new_filename = full_name(filename, f"conflicting with {author}")
+    suffix = "filename conflict"
+    if content_conflict:
+        suffix = "file's content conflict"
```
vxgmichel, line 51: I'd personally use `name conflict` and `content conflict`, until we implement better internationalization for those suffixes.

vxgmichel (changes requested):
A blocking issue with `:` not being supported on windows and a few suggestions but LGTM otherwise :+1:

vxgmichel (approved):
One minor comment, :+1:  otherwise!

parsec/core/fs/workspacefs/sync_transactions.py
```diff
@@ -43,12 +44,14 @@ def get_filename(manifest: LocalFolderishManifests, entry_id: EntryID) -> Option
...
-    filename: EntryName, filenames: List[EntryName], author: DeviceID
+    filename: EntryName, filenames: Iterable[EntryName], suffix: str = NAME_CONFLICT_SUFFIX
 ) -> EntryName:
     counter = count(2)
-    new_filename = full_name(filename, f"conflicting with {author}")
-    while new_filename in filenames:
-        new_filename = full_name(filename, f"conflicting with {author} - {next(counter)}")
+    suffix = f"Parsec - {suffix}"
+    new_filename = full_name(filename, suffix)
+    filename_set = set(filenames)
```
touilleMan, line 52: Bot sure if conversion to a set is worth it here:
- Name clash is unlikely so we are going to do a single lookup most of the time (and each additional lookup is a order of magnitude more unlikely than the previous one)
- conversion is a O(n) operation that allow to do lookup in O(log(n))
- without conversion, lookup is O(n)
- O(n) is roughly equivalent to O(log(n)) if n is small enough (say n<100) which is the most common case for folder manifest

touilleMan (approved):
LGTM ;-)

PR #1838, opened 2021-09-08
Add a native MacOS menu

Ironicbay:


Ironicbay:
Current look:
Capture d’écran 2021-09-10 à 15 31 14
Capture d’écran 2021-09-10 à 15 31 21
Capture d’écran 2021-09-10 à 15 31 26
Capture d’écran 2021-09-10 à 15 31 31
Capture d’écran 2021-09-10 à 15 31 38

It's the current app menu duplicated here, which seems to be standard for many apps.
The `Preferences` and `Quit Parsec` are standard with standard keybinds, which is something that was asked for at some point.

touilleMan (approved):
👌 al dente PR

PR #1839, opened 2021-09-08
Increase autojump clock threshold in test_sync_monitor.py (due to thr…

touilleMan:
…ead use in local data storage)


vxgmichel (approved):
LGTM :+1:

PR #1852, opened 2021-09-14
Add a way to hide unmounted workspaces

### LINKED ISSUE #1744: Add a way to hide workspaces from the GUI

### LINKED ISSUE #3056: Add a way to hide workspaces from the GUI
Forced to create a new issue because #1744 was incorrectly used for another PR.

Max-7:
Closes #1744
Closes #3056 (second one was created because 1744 was used incorrectly for another newsfragment).

Add a checkbox to hide unmounted workspaces. The checkbox is checked by default. Ideally it should remember its previous state but it requires a config by device which is quite bothersome. This solution should be good enough for now.

### DIFF
--- a/newsfragments/3056.feature.rst
+++ b/newsfragments/3056.feature.rst
@@ -1,0 +1,1 @@
+Add a way to filter out unmounted workspaces
--- a/parsec/core/gui/forms/workspaces_widget.ui
+++ b/parsec/core/gui/forms/workspaces_widget.ui
@@ -6,8 +6,8 @@
    <rect>
     <x>0</x>
     <y>0</y>
-    <width>636</width>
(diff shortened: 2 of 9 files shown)


touilleMan:
Can we move the checkbox on the search line ? It could be more elegant
image

Max-7:
> Can we move the checkbox on the search line ? It could be more elegant image

No, for two reasons:
1. The users pages uses the same layout
2. It adds too many things on the same line, which restricts the minimum width of the window.

Ironicbay (approved):
👍

PR #1857, opened 2021-09-16
Improve conflict messages

vamonte:


parsec/core/fs/workspacefs/sync_transactions.py
```diff
@@ -44,10 +45,14 @@ def get_filename(manifest: LocalFolderishManifests, entry_id: EntryID) -> Option
 
 
 def get_conflict_filename(
-    filename: EntryName, filenames: Iterable[EntryName], suffix: str = NAME_CONFLICT_SUFFIX
+    filename: EntryName,
+    filenames: Iterable[EntryName],
+    core_config: CoreConfig,
+    suffix_key: str = FILENAME_CONFLICT_KEY,
```
vxgmichel, line 51: Maybe make this argument mandatory so `merge_folder_children` has to pass `FILENAME_CONFLICT_KEY` explicitly? (cause I don't think it makes sense for one of the two keys to be the default)

vxgmichel (approved):
LGTM :+1: 

Maybe add a test that changes the core config from `en` to `fr`?

PR #1867, opened 2021-09-23
Improve the workspace sharing UX.(Disable some realm roles for the ou…

vamonte:
…tsider profile)


Max-7 (approved):

touilleMan (approved):
Only the newsfragment is missing :thumbup:

PR #1870, opened 2021-09-30
Correct errors in .po files

Ironicbay:


Delantonio (changes requested):
Information is uncountable in english which means that it can only be used in singular

Delantonio (approved):

PR #1888, opened 2021-11-05
Improve file download

alexandrediasldev:


parsec/core/fs/remote_loader.py
```diff
@@ -351,15 +356,41 @@ def __init__(
...
+               FSBackendOfflineError
+               FSWorkspaceInMaintenance
+           """
+        blocks_iter = iter(blocks)
+        channels: Tuple[
+            "MemorySendChannel[BlockAccess]", "MemoryReceiveChannel[BlockAccess]"
+        ] = open_memory_channel(0)
+        send_channel, receive_channel = channels
+
+        async def _loader(send_channel: "MemorySendChannel[BlockAccess]") -> None:
```
vxgmichel, line 380: Why is the `send_channel` type in quotes?

parsec/core/fs/remote_loader.py
```diff
@@ -351,15 +356,41 @@ def __init__(
...
+                while True:
+                    access = next(blocks_iter, None)
+                    if not access:
+                        break
+                    await self.load_block(access)
+                    await send_channel.send(access)
+
+        nursery.start_soon(_loader, send_channel)
+        for _ in range(3):
+            nursery.start_soon(_loader, send_channel.clone())
```
vxgmichel, line 391: The trio docs suggests:
```python
async with send_channel:
    for _ in range(4):
        nursery.start_soon(_loader, send_channel.clone())
```

parsec/core/fs/storage/chunk_storage.py
```diff
@@ -96,6 +96,44 @@ async def is_chunk(self, chunk_id: ChunkID) -> bool:
...
+
+        local_chunk_ids = []
+        bytes_id_list = [(id.bytes,) for id in chunk_id]
+
+        async with self._open_cursor() as cursor:
+            # Can't use execute many with SELECT so we have to make a temporary table filled with the needed chunk_id
+            # and intersect it with the normal table
+            # create random name for the temporary table to avoid asynchronous errors
+            table_name = "temp" + secrets.token_hex(12)
+            table_name = "".join(chr for chr in table_name if chr.isalnum())
```
vxgmichel, line 109: Isn't `table_name` already alphanum only? If it acts as a safety net then an `assert` would be a better fit.

parsec/core/fs/storage/chunk_storage.py
```diff
@@ -96,6 +96,44 @@ async def is_chunk(self, chunk_id: ChunkID) -> bool:
...
+            {table_name} (chunk_id)
+            VALUES (?)""",
+                iter(bytes_id_list),
+            )
+
+            cursor.execute(
+                f"""SELECT chunk_id FROM chunks INTERSECT SELECT chunk_id FROM {table_name}"""
+            )
+
+            manifest_rows = cursor.fetchall()
```
vxgmichel, line 129: `manifest`? I'm not sure this name correctly represents this variable.

parsec/core/fs/storage/chunk_storage.py
```diff
@@ -96,6 +96,44 @@ async def is_chunk(self, chunk_id: ChunkID) -> bool:
...
+            cursor.execute(f"""DROP TABLE IF EXISTS {table_name}""")
+        for row in manifest_rows:
```
vxgmichel, line 131: An extra line feed would be nice to show we're out of the `cursor` context.

parsec/core/fs/storage/chunk_storage.py
```diff
@@ -96,6 +96,44 @@ async def is_chunk(self, chunk_id: ChunkID) -> bool:
...
+            index = bytes_id_list.index(row)
+            local_chunk_ids.append(chunk_id[index])
```
vxgmichel, line 133: This logic of looking back at the bytes seems a bit backward, I'd rather do:
```python
return [ChunkID(bytes=id_bytes) for (id_bytes,) in intersect_rows]
```
The `bytes` argument is not supported at the moment but this `UUID4` class can easily be modified as:
```python
class UUID4(UUID):
    __slots__ = ()

    def __init__(self, *args, **kwargs):
        if not args and not kwargs:
            super().__init__(bytes=uuid4().bytes)
        else:
            super().__init__(*args, **kwargs)
```

parsec/core/fs/workspacefs/entry_transactions.py
```diff
@@ -180,6 +181,35 @@ async def _lock_parent_manifest_from_path(
             await self._load_manifest(entry_id)
 
     # Transactions
+    async def entry_missing_data(
+        self, path: FsPath, limit: int
+    ) -> Tuple[int, int, List[BlockAccess]]:
+        missing_blocks = []
+        manifest, confinement_point = await self._get_manifest_from_path(path)
+        manifest: LocalFileManifest
+        accessible_chunks = []
```
vxgmichel, line 190: `accessible_chunks` (in the sense of `chunks with an access`) should be called `blocks`

parsec/core/fs/workspacefs/entry_transactions.py
```diff
@@ -180,6 +181,35 @@ async def _lock_parent_manifest_from_path(
...
+                    total_size += chunk.raw_size
+
+        accessible_chunk_ids = [chunk.id for chunk in accessible_chunks]
+        missing_size = 0
+
+        local_chunk_ids = await self.local_storage.block_storage.get_local_chunk_ids(
+            accessible_chunk_ids
+        )
+        for chunk in accessible_chunks:
+            if chunk.id not in local_chunk_ids:
```
vxgmichel, line 207: Use a `set` for `O(1)` lookups.

vxgmichel (changes requested):
LGTM :+1:

Just a few nitpicks :)

parsec/core/fs/remote_loader.py
alexandrediasldev: Mypy assumes it's a generic type but the interpreter raise an error, to fix that we put the type in quotes.
See https://stackoverflow.com/questions/55076778/why-isnt-this-function-type-annotated-correctly-error-missing-type-parameters for more info

PR #1891, opened 2021-11-09
[GUI] Better error handling with a device using smartcards on a versi…

Max-7:
…on of Parsec that does not support smartcards


parsec/core/gui/tr/parsec_en.po
```diff
@@ -327,6 +327,11 @@ msgstr "Cannot create recovery device"
 msgid "TEXT_DEVICE_UNLOCK_FOR_RECOVERY_LABEL"
 msgstr "Please enter the device password."
 
+msgid "TEXT_UNLOCK_ERROR_SMARTCARD_NOT_AVAILABLE"
+msgstr ""
+"This device was created with a smartcard but this version of Parsec does "
```
touilleMan, line 332: "This device is protected by smartcard authentication, which is not supported by this version of Parsec."

PR #1893, opened 2021-11-10
Fix click errors on MacOS 11 & 12 and update MacOS PyInstaller

Ironicbay:
When building on MacOS 11 and 12, the `click` error persists even with the current workaround.

The default encoding when launching apps from the Finder/taskbar is `ascii` regardless of which is used in Python, hence the error [here](https://github.com/pallets/click/blob/main/src/click/_unicodefun.py). Setting the correct flags in the `.spec` file solves the issue and allows the previous fix to be removed.

Probably linked to #1703.

### DIFF
--- a/packaging/macOS/parsec.spec
+++ b/packaging/macOS/parsec.spec
@@ -62,6 +62,8 @@ app = BUNDLE(coll,
             'CFBundleURLSchemes': ['parsec']
          }
       ],
+      # Encoding is specified here to force UTF-8 when launching app instead of
+      # default ascii, which caused issues with click
       'LSEnvironment': {
          'LANG': 'en_US.UTF-8',
          'LC_CTYPE': 'en_US.UTF-8'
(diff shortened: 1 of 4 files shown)


.azure-pipelines.yml
```diff
@@ -434,7 +434,7 @@ jobs:
 
 - job: 's2_macos_app_build'
   displayName: 'MacOS app build'
-  condition: startsWith(variables['build.sourceBranch'], 'refs/tags/')
+  # condition: startsWith(variables['build.sourceBranch'], 'refs/tags/')
```
touilleMan, line 437: Don't forget to re-enable this before merge !

packaging/macOS/parsec.spec
```diff
@@ -61,6 +61,10 @@ app = BUNDLE(coll,
             'CFBundleURLName': 'com.scille.parsec',
             'CFBundleURLSchemes': ['parsec']
          }
-      ]
+      ],
+      'LSEnvironment': {
```
touilleMan, line 65: Should add a comment to explain why this is needed (given it looks strange not to use the user's locale but to force English everywhere)

touilleMan (approved):

PR #1899, opened 2021-11-18
Remove Babel from requirements (only needed as setup requirements for…

touilleMan:
… wheel build)


touilleMan:
Babel is kind of big dependency (~8mo) so better only install it when needed ;-)

PR #1903, opened 2021-11-22
Add first rust test to the python test suite

vxgmichel:


oxidation/libparsec_python/src/crypto.rs
```diff
@@ -16,8 +20,16 @@ impl HashDigest {
     }
 
     #[staticmethod]
-    fn from_data(data: &[u8]) -> PyResult<HashDigest> {
-        Ok(Self(parsec_api_crypto::HashDigest::from_data(data)))
+    fn from_data(py: Python, data: PyObject) -> PyResult<HashDigest> {
+        let copy;
+        let bytes = match data.extract::<&PyByteArray>(py) {
+            Ok(x) => {
+                copy = x.to_vec();
```
touilleMan, line 27: Using [unsafe as_bytes](https://docs.rs/pyo3/0.15.1/pyo3/types/struct.PyByteArray.html#method.as_bytes) seems ok here given the GIL is held for the duration of the function and `&[u8]` doesn't leave the function body

oxidation/libparsec_python/src/crypto.rs
```diff
@@ -16,8 +20,12 @@ impl HashDigest {
     }
 
     #[staticmethod]
-    fn from_data(data: &[u8]) -> PyResult<HashDigest> {
-        Ok(Self(parsec_api_crypto::HashDigest::from_data(data)))
+    fn from_data(py: Python, data: PyObject) -> PyResult<HashDigest> {
+        let bytes = match data.extract::<&PyByteArray>(py) {
+            Ok(x) => unsafe { x.as_bytes() },
```
touilleMan, line 25: should add a comment to explain why this is ok here ;-)

touilleMan (approved):

PR #1912, opened 2021-11-26
[GUI] Better display disabled items in combo boxes

### LINKED ISSUE #1864: Better handling of disallowed role selection with outsider
image

The current message is rather violent, it would be better to simply disallow the Owner/Manager options with a tooltip text explaning outsiders cannot have this kind of role.

Max-7:
Closes #1864 

not_showing_disabled_items
show_disabled_items


vxgmichel (approved):
LGTM :+1: 

Although the striking is a bit much isn't it? I think the grey disabled color plus the tooltip should be enough.

PR #1914, opened 2021-11-26
Added detection for deprecated versions of macfuse

### LINKED ISSUE #1907: No detection if deprecated version of osxfuse is installed
Python doesn't raise an `OSError` as expected currently if an old (and not functioning) version of osxfuse is installed.

Error example with version 3.11.2: 
```
kext load failed: -603947007
mount_osxfuse: the file system is not available (255)
```

In this scenario, no warning happen during installation (since fuse can be imported) and mounting doesn't work.

Ironicbay:
Closes #1907

### DIFF
--- a/parsec/core/gui/instance_widget.py
+++ b/parsec/core/gui/instance_widget.py
@@ -7,6 +7,7 @@ import trio
 from structlog import get_logger
 from PyQt5.QtCore import pyqtSignal, pyqtSlot
 from PyQt5.QtWidgets import QWidget, QVBoxLayout, QApplication
+from packaging.version import Version
 
 from parsec.event_bus import EventBus
 from parsec.api.protocol import HandshakeRevokedDevice
@@ -34,6 +35,8 @@ from parsec.core.gui.central_widget import CentralWidget
 
 logger = get_logger()
 
+MIN_MACFUSE_VERSION = Version("4.2.3")
+
 
 async def _do_run_core(config, device, qt_on_ready):
     # Quick fix to avoid MultiError<Cancelled, ...> exception bubbling up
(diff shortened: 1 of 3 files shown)


parsec/core/gui/instance_widget.py
```diff
@@ -47,6 +47,24 @@ async def _do_run_core(config, device, qt_on_ready):
                 await trio.sleep_forever()
 
 
+def check_macfuse_version() -> bool:
+    import os
+
+    macfuse_plist_path = "/Library/Filesystems/macfuse.fs/Contents/Info.plist"
+    if not os.path.exists(macfuse_plist_path):
+        # osxfuse installed, not macfuse
```
touilleMan, line 55: osxfuse and macfuse are not the same thing ???? O_o

(maybe add a small comment about this, I guess one is a legacy version of the other 🤔 )

parsec/core/gui/instance_widget.py
```diff
@@ -47,6 +47,24 @@ async def _do_run_core(config, device, qt_on_ready):
...
+        return False
+
+    import plistlib
+    import requests
+
+    local_version = plistlib.readPlist(macfuse_plist_path)["CFBundleVersion"]
+    remote_version = requests.get("https://api.github.com/repos/osxfuse/osxfuse/releases/latest")
```
touilleMan, line 62: this is going to make the GUI freeze during the whole query so we should avoid this (typically using a job)

parsec/core/gui/instance_widget.py
```diff
@@ -47,6 +47,24 @@ async def _do_run_core(config, device, qt_on_ready):
...
+    if local_version < remote_version.json()["name"].split()[1]:
```
touilleMan, line 63: So we ask the user to update his version of macfuse every time a new release occurs upstream ?

Shouldn't we just set a minimal macfuse version (that we update each time we do a new Parsec release) and compare to this one ? this would have to benefits:
- no need for a http request ;-)
- less annoyance to the user considering how new releases [are rather abundant](https://github.com/osxfuse/osxfuse/releases), besides it should be more acceptable for the user to update his macfuse version when he choose to update Parsec rather than at any random time

touilleMan (changes requested):

parsec/core/gui/instance_widget.py
```diff
@@ -34,6 +34,8 @@
 
 logger = get_logger()
 
+min_macfuse_version = "4.2.3"
```
touilleMan, line 37: constant should be uppercase

parsec/core/gui/instance_widget.py
```diff
@@ -47,6 +49,26 @@ async def _do_run_core(config, device, qt_on_ready):
...
+        # it probably means that its previous version, osxfuse, is installed
+        # instead (in a "osxfuse.fs" directory). Any current MacOS version is
+        # incompatible with osxfuse.
+        return False
+
+    import plistlib
+
+    local_version = plistlib.readPlist(macfuse_plist_path)["CFBundleVersion"]
+
+    if local_version < min_macfuse_version:
```
touilleMan, line 67: comparing strings is dangerous !

`1.10 > 1.2` but `"1.10" < "1.2"`

We should use `from packaging.version.Version` (see what is done in `parsec/core/gui/new_version.py`)

touilleMan (approved):

PR #1915, opened 2021-11-26
Add pyup config

alexandrediasldev:


setup.py
```diff
@@ -271,85 +270,28 @@ def run(self):
...
-    "docutils>=0.12,<0.16",
-    # Documentation generation requirements
-    "sphinx==2.4.3",
-    "sphinx-intl==2.0.0",
-    "sphinx-rtd-theme==0.4.3",
-    "psutil==5.8.0",
-]
+def get_requirement_from_file(file):
+    with open(file) as requirements_txt:
+        return str(requirements_txt.read()).split("\n")
```
touilleMan, line 275: no need for the `str` conversion

but encoding should be specified in the open instead (otherwise system default encoding might be used instead of utf8).

(and you can fix the lacking encoding option when reading README and HISTORY btw.)

touilleMan: And parsing is wrong here: we can't have comments in the requirement.txt file

in theory we should use `pkg_resources.parse_requirements` to parse a requirements.txt file, however pkg_resources is a big module (slow to load) and depends of setuptools so it really seems like heavy artillery for the job

good thing is [pkg_resources.parse_requirements is simple enough](https://github.com/pypa/pip/blob/a800ae2a7d680c0ed5ba21476f722e6416a5ce0d/src/pip/_vendor/pkg_resources/__init__.py#L3075-L3094)

So I think we should just copy the parser code in `get_requirement_from_file`:
```python
def _maybe_add_as_continuation(current_line, previous_lines):
    try:
        last_line = previous_lines[-1]
    except IndexError:
        return False
    if last_line.endswith("\\"):
        previous_lines[-1]` = last_line[:-1] .strip() + current_line
        return True
    else:
        return False

lines = []
for line in requirements.iterlines():
    line = line.strip()
    if not line or line.startswith("#"):
        continue
    # Drop comments -- a hash without a space may be in a URL.
    line = line[:line.find(' #')]
    # If there is a line continuation, drop it, and append the next line.
    if not _maybe_add_as_continuation((line, lines):
        lines.append(line)
_maybe_add_as_continuation("", lines)  # In case file finish with \\ and no further line
```
Note I've changed continuation handling from the original code given:
1) it seems to me it wasn't handling multiple line continuation properly
2) the [removal of the continuation character)(https://github.com/pypa/pip/blob/a800ae2a7d680c0ed5ba21476f722e6416a5ce0d/src/pip/_vendor/pkg_resources/__init__.py#L3089) also remove the previous character, which imply it must be a space but this is [not what the spec says](https://pip.pypa.io/en/stable/reference/requirements-file-format/#line-continuations)

touilleMan (changes requested):

setup.py
vxgmichel: > however pkg_resources is a big module (slow to load) and depends of setuptools so it really seems like heavy artillery for the job

I beg to differ :)
```python
In [1]: import setuptools

In [2]: %time import pkg_resources
CPU times: user 4 µs, sys: 0 ns, total: 4 µs
Wall time: 4.77 µs
```

vxgmichel (approved):
LGTM :+1:

PR #1916, opened 2021-11-26
[GUI] Added a navigation bar when navigating a workspace

Max-7:
folder_navigation_bar


vxgmichel (commented):
I have a suggestion: what about discarding `root` and make the workspace name clickable:
`Documents ● Alice workspace > Directory 1 > Directory 2`

Also the navigation bar blinks when entering a new directory.

Also:
```
2022-01-27T11:23:40.926188Z [error    ] Exception in Qt slot           
  File "/home/vinmic/repos/parsec-cloud/parsec/core/gui/central_widget.py", line 384, in _on_organization_stats_success
    self.menu.show_organization_stats(
  File "/home/vinmic/repos/parsec-cloud/parsec/core/gui/menu_widget.py", line 58, in show_organization_stats
    self.label_organization_name.setText(organization_id.str)
AttributeError: 'builtins.OrganizationID' object has no attribute 'str'
```

Also:
```
QLayout: Attempting to add QLayout "" to NavigationBarWidget "navigation_bar_widget", which already has a layout
```

Also I don't think those path items have to look like hyperlinks, having the mouse cursor changing is probably enough.

parsec/core/gui/navigation_bar_widget.py
```diff
@@ -0,0 +1,128 @@
...
+            label = None
+            if idx != len(parts) - 1:
+                obj_name = part.replace(" ", "_")
+                label = ClickableLabel(part)
+                label.clicked.connect(self._on_label_clicked(idx))
+                label.setObjectName(f"label_{obj_name}")
+                label.setStyleSheet(
+                    f"#label_{obj_name} {{ color: #999999; }} #label_{obj_name}:hover {{ color: #0092FF; }}"
+                )
+            else:
```
vxgmichel, line 105: Is this `if-else` really necessary? I mean clicking the current directory seems fine to me

vxgmichel (approved):
LGTM :+1:

parsec/core/gui/navigation_bar_widget.py
Max-7: It's just that clicking it does absolutely nothing so I would rather avoid showing it as clickable.

PR #1924, opened 2021-12-01
Added bindings for SigningKey and VerifyKey

Max-7:


tests/rust/test_signing_and_verify_key.py
```diff
@@ -0,0 +1,45 @@
...
+
+    MESSAGE = b"My message"
+
+    rs_signed = rs_sk.sign(MESSAGE)
+    py_signed = py_sk.sign(MESSAGE)
+
+    assert rs_signed == py_signed
+
+    assert rs_vk.verify(rs_signed) == py_vk.verify(py_signed)
+    assert rs_vk.verify(py_signed) == py_vk.verify(rs_signed)
```
touilleMan, line 39: should also check behavior on invalid data to verify (typically testing too small data and data with incorrect signature)

oxidation/libparsec_python/src/crypto.rs
```diff
@@ -52,3 +52,90 @@ impl HashDigest {
...
+        match parsec_api_crypto::VerifyKey::try_from(data) {
+            Ok(h) => Ok(Self(h)),
+            Err(err) => Err(PyValueError::new_err(err)),
+        }
+    }
+
+    fn verify<'p>(&self, py: Python<'p>, signed: &[u8]) -> PyResult<&'p PyBytes> {
+        match self.0.verify(signed) {
+            Ok(v) => Ok(PyBytes::new(py, &v)),
+            Err(err) => Err(PyValueError::new_err(err)),
```
touilleMan, line 114: In Python implementation, a `nacl.exceptions.CryptoError` is raised

it is a bit messy to import `nacl` exception from here (given we don't explicitly depend on pynacl, but on on the other hand we cannot do `from parsec.crypto import CryptoError` given `parsec.crypto` depends on us...

Another solution would be to create a separate `parsec.crypto_exceptions` module (or replace `parsec/crypto.py` by a folder to have `parsec/crypto/impl.py` and `parsec/crypto/exceptions.py`)
@vxgmichel any opinion on this ?

touilleMan (changes requested):

oxidation/crates/parsec_api_crypto/src/rustcrypto/sign.rs
```diff
@@ -99,7 +99,7 @@ impl VerifyKey {
     pub const ALGORITHM: &'static str = "ed25519";
     pub const SIZE: usize = PUBLIC_KEY_LENGTH;
 
-    pub fn unsecure_unwrap<'a>(&self, signed: &'a [u8]) -> Option<&'a [u8]> {
+    pub fn unsecure_unwrap(signed: &[u8]) -> Option<&[u8]> {
```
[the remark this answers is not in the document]
touilleMan: and `parsec_api_crypto/tests/sign.rs` should be modified to call `VerifyKey::unsecure_unwrap`

oxidation/libparsec_python/src/crypto.rs
jojoquivogn: A way to do it is to use `use pyo3::import_exception;` and then `import_exception!(nacl.exceptions, CryptoError);`. It work perfectly for me @touilleMan

Max-7:
Done by #1950

PR #1926, opened 2021-12-02
Added entry get block by type

alexandrediasldev:


parsec/core/fs/workspacefs/entry_transactions.py
```diff
@@ -198,17 +199,30 @@ async def entry_missing_data(
...
         local_chunk_ids_set = set(local_chunk_ids)
-        for chunk in blocks:
-            if chunk.id not in local_chunk_ids_set:
-                assert chunk.access is not None
-                missing_blocks.append(chunk.access)
-                missing_size += chunk.raw_size
 
-        return missing_size, total_size, missing_blocks
+        for chunk in blocks:
+            assert chunk.is_block
```
vxgmichel, line 217: Could you turn `is_block` property into a method? I think it's confusing and bad move on my part :sweat_smile:

parsec/core/fs/workspacefs/entry_transactions.py
```diff
@@ -198,17 +199,30 @@ async def entry_missing_data(
...
+            if chunk.id in local_and_remote_chunk_ids_set:
+                local_and_remote_blocks.append(chunk.access)
+            elif chunk.id in local_chunk_ids_set:
+                local_blocks.append(chunk.access)
+            else:
+                remote_blocks.append(chunk.access)
+
+        return local_and_remote_blocks, local_blocks, remote_blocks
```
vxgmichel, line 225: So `total_size` is no longer returned? I was thinking it might be interesting to return:
- `file_size`: `manifest.size`
- `proper_blocks_size`: `sum(chunk.raw_size for chunks in manifest.blocks for chunk in chunks if chunk.is_block())`
- `pending_chunks_size`: `sum(chunk.raw_size for chunks in manifest.blocks for chunk in chunks if not chunk.is_block())`

vxgmichel: This makes 6 values to return so maybe a `namedtuple` would be nice to avoid confusion.

parsec/core/fs/workspacefs/entry_transactions.py
```diff
@@ -198,17 +199,30 @@ async def entry_missing_data(
...
+        remote_blocks = []
+        local_and_remote_blocks = []
+
+        # To avoid concurrency problems local storage is called first
+        local_and_remote_chunk_ids = await self.local_storage.block_storage.get_local_chunk_ids(
+            block_ids
+        )
+        local_and_remote_chunk_ids_set = set(local_and_remote_chunk_ids)
+
+        local_chunk_ids = await self.local_storage.chunk_storage.get_local_chunk_ids(block_ids)
```
vxgmichel, line 213: I would call those variables:
- `block_ids`
- `local_and_remote_block_ids`
- `local_only_block_ids`

And then compute `remote_only_block_ids` as:
```python
remote_only_block_ids = block_ids - local_and_remote_block_ids - local_only_block_ids
```
Also if you use a mapping id-to-block called `block_dict` you can then do:
```python
local_only_blocks = [block_dict[block_id].access for block_id in local_only_block_ids]
remote_only_blocks = [block_dict[block_id].access for block_id in remote_only_block_ids]
local_and_remote_blocks = [block_dict[block_id].access for block_id in local_and_remote_block_ids]
```

To sum it up:
```python
block_dict = {chunk.id: chunk for chunks in manifest.blocks for chunk in chunks if chunk.is_block()}
block_ids = set(block_dict)

local_and_remote_chunk_ids = set(await self.local_storage.block_storage.get_local_chunk_ids(block_ids))
local_only_chunk_ids = set(await self.local_storage.chunk_storage.get_local_chunk_ids(block_ids))
remote_only_block_ids = block_ids - local_and_remote_block_ids - local_only_block_ids

local_only_blocks = [block_dict[block_id].access for block_id in local_only_block_ids]
remote_only_blocks = [block_dict[block_id].access for block_id in remote_only_block_ids]
local_and_remote_blocks = [block_dict[block_id].access for block_id in local_and_remote_block_ids]
```
What do you think?

vxgmichel (commented):
A few comments but LGTM otherwise :+1:

parsec/core/fs/workspacefs/entry_transactions.py
alexandrediasldev: block_dict needs to not be a list comprehension to be able to break when we reach the limit argument passed to the function apart from that LGTM

vxgmichel (approved):
LGTM :+1:

PR #1940, opened 2021-12-07
Added share_workspace --recipiant option and human_find command

Delantonio:
share_workspace --recipiant option is used to share a workspace using the human handle of a user (name or email address) instead of the user_id
The human_find command print user_id and human_handle corresponding to the query

### DIFF
--- a/parsec/core/cli/human_find.py
+++ b/parsec/core/cli/human_find.py
@@ -12,25 +12,27 @@ async def _human_find(
     config,
     device,
     query,
+    omit_revoked,
+    omit_non_human: bool = False,
     page: int = 1,
     per_page: int = 100,
-    omit_revoked: bool = False,
-    omit_non_human: bool = False,
 ):
     async with logged_core_factory(config, device) as core:
         user_info_tab, nb = await core.find_humans(
             query, page, per_page, omit_revoked, omit_non_human
         )
     for user in user_info_tab:
-        click.echo(f"{user.human_handle} - UserID: {user.user_id}")
+        is_revoked = " (revoked)" if user.revoked_on is not None else ""
+        click.echo(f"{user.human_handle} - UserID: {user.user_id}{is_revoked}")
     if not nb:
         click.echo("No human found!")
 
 
-@click.command(short_help="search user id from email address")
+@click.command(short_help="Retrieve user ID from human email/label")
 @click.argument("query", type=str)
+@click.option("--include-revoked", is_flag=True)
 @core_config_and_device_options
 @cli_command_base_options
-def human_find(config, device, query, **kwargs) -> dict:
+def human_find(config, device, query, include_revoked, **kwargs) -> dict:
     with cli_exception_handler(config.debug):
-        trio_run(_human_find, config, device, query)
+        trio_run(_human_find, config, device, query, not include_revoked)
--- a/parsec/core/cli/human_find.py
+++ b/parsec/core/cli/human_find.py
@@ -8,7 +8,15 @@ from parsec.core.cli.utils import cli_command_base_options, core_config_and_devi
 from parsec.cli_utils import cli_exception_handler
 
 
-async def _human_find(config, device, query, page, per_page, omit_revoked, omit_non_human):
+async def _human_find(
+    config,
+    device,
+    query,
+    page: int = 1,
+    per_page: int = 100,
+    omit_revoked: bool = False,
+    omit_non_human: bool = False,
+):
     async with logged_core_factory(config, device) as core:
         user_info_tab, nb = await core.find_humans(
             query, page, per_page, omit_revoked, omit_non_human
@@ -20,18 +28,9 @@ async def _human_find(config, device, query, page, per_page, omit_revoked, omit_
 
 
 @click.command(short_help="search user id from email address")
-@click.option("--query", required=True, type=str, help="Search through names and email addresses")
+@click.argument("query", type=str)
 @core_config_and_device_options
 @cli_command_base_options
-def human_find(
-    config,
-    device,
-    query: str = None,
-    page: int = 1,
-    per_page: int = 100,
-    omit_revoked: bool = False,
-    omit_non_human: bool = False,
-    **kwargs,
-) -> dict:
+def human_find(config, device, query, **kwargs) -> dict:
     with cli_exception_handler(config.debug):
-        trio_run(_human_find, config, device, query, page, per_page, omit_revoked, omit_non_human)
+        trio_run(_human_find, config, device, query)
--- a/parsec/core/cli/human_find.py
+++ b/parsec/core/cli/human_find.py
@@ -1,0 +1,37 @@
+# Parsec Cloud (https://parsec.cloud) Copyright (c) AGPLv3 2016-2021 Scille SAS
+
+import click
+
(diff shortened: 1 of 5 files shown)


parsec/core/cli/human_find.py
```diff
@@ -0,0 +1,37 @@
...
+            query, page, per_page, omit_revoked, omit_non_human
+        )
+    for user in user_info_tab:
+        click.echo(f"{user.human_handle} - UserID: {user.user_id}")
+    if not nb:
+        click.echo("No human found!")
+
+
+@click.command(short_help="search user id from email address")
+@click.option("--query", required=True, type=str, help="Search through names and email addresses")
```
AntoineDupre, line 23: Maybe query can be a`click.argument instead of being a `click.option` to be used directly (as it is the main entry for this command)

parsec/core/cli/human_find.py
```diff
@@ -0,0 +1,37 @@
...
+@core_config_and_device_options
+@cli_command_base_options
+def human_find(
+    config,
+    device,
+    query: str = None,
+    page: int = 1,
```
AntoineDupre, line 30: The following options can not be configured from the cli. Instead of setting default value there, in the cli command definition, it's better to remove those argument and set default value to the helper `_find_humans`

Ex
```python
def human_find(config, device, query):

async def _human_find(config, device, query, page=1, per_page=100, omit_revoked=False, omit_non_human=Flase):
```

AntoineDupre (changes requested):

AntoineDupre (approved):
:+1:

parsec/core/cli/share_workspace.py
```diff
@@ -13,24 +13,36 @@
...
+                raise RuntimeError("Specify the user more precisely or use the --user_id option")
+            user_id = user_info_tab[0].user_id
         await core.user_fs.workspace_share(workspace.id, user_id, user_role)
 
 
 @click.command(short_help="share workspace")
 @click.option("--workspace-name")
-@click.option("--user-id", type=UserID, required=True)
+@click.option("--user-id", type=UserID)
+@click.option("--recipiant", help="Name or email to whom the workspace is shared")
```
touilleMan, line 36: "[...]is shared with" ?

parsec/core/cli/share_workspace.py
```diff
@@ -13,24 +13,36 @@
 WORKSPACE_ROLE_CHOICES = {"NONE": None, **{role.value: role for role in WorkspaceRole}}
 
 
-async def _share_workspace(config, device, name, user_id, user_role):
+async def _share_workspace(config, device, name, user_id, recipiant, user_role):
+    if recipiant and user_id or not recipiant and not user_id:
+        raise click.ClickException("Either recipiant or user_id should be used, but not both")
```
touilleMan, line 18: Should indicate the option name used by the user for better understanding: "[...] --recipiant or --user-id [...]"

parsec/core/cli/share_workspace.py
```diff
@@ -13,24 +13,36 @@
...
     async with logged_core_factory(config, device) as core:
         workspace = core.find_workspace_from_name(name)
+        if recipiant:
+            user_info_tab, nb = await core.find_humans(recipiant, 1, 100, False, False)
+            if nb == 0:
+                raise RuntimeError("This recipiant doesn't exist, use the email address")
+            if nb != 1:
+                for user in user_info_tab:
+                    click.echo(f"{user.human_handle} - UserID: {user.user_id}")
+                raise RuntimeError("Specify the user more precisely or use the --user_id option")
```
touilleMan, line 28: `--user-id`

parsec/core/cli/share_workspace.py
```diff
@@ -13,24 +13,36 @@
...
-async def _share_workspace(config, device, name, user_id, user_role):
+async def _share_workspace(config, device, name, user_id, recipiant, user_role):
+    if recipiant and user_id or not recipiant and not user_id:
+        raise click.ClickException("Either recipiant or user_id should be used, but not both")
     async with logged_core_factory(config, device) as core:
         workspace = core.find_workspace_from_name(name)
+        if recipiant:
+            user_info_tab, nb = await core.find_humans(recipiant, 1, 100, False, False)
+            if nb == 0:
+                raise RuntimeError("This recipiant doesn't exist, use the email address")
```
touilleMan, line 24: I think "Unknown recipiant" is good enough (the "use the email address" part is misleading if the command has already been used with an email address that happened to not be part of the existing human handles)

parsec/core/cli/human_find.py
```diff
@@ -0,0 +1,36 @@
...
+        user_info_tab, nb = await core.find_humans(
+            query, page, per_page, omit_revoked, omit_non_human
+        )
+    for user in user_info_tab:
+        click.echo(f"{user.human_handle} - UserID: {user.user_id}")
+    if not nb:
+        click.echo("No human found!")
+
+
+@click.command(short_help="search user id from email address")
```
touilleMan, line 30: "Retreive user ID from human email/label"

parsec/core/cli/human_find.py
```diff
@@ -0,0 +1,36 @@
...
+    per_page: int = 100,
+    omit_revoked: bool = False,
+    omit_non_human: bool = False,
+):
+    async with logged_core_factory(config, device) as core:
+        user_info_tab, nb = await core.find_humans(
+            query, page, per_page, omit_revoked, omit_non_human
+        )
+    for user in user_info_tab:
+        click.echo(f"{user.human_handle} - UserID: {user.user_id}")
```
touilleMan, line 25: would be interesting to also add the revoked information in the output:
```
$ parsec core human_find zack
Zack Snyder <zack@example.com> - UserID: acd17910ca5b4d3dbce4cc29a79fd2f8
Zack Snyder <zack@example.com> - UserID: c7c7f53f7e8744d48ed3d106877bb20f (revoked)
Zack Effron <zack.effron@example.com> - UserID: 078d3e952698428290f036a0b37d3860
```

parsec/core/cli/human_find.py
```diff
@@ -0,0 +1,36 @@
...
+    if not nb:
+        click.echo("No human found!")
+
+
+@click.command(short_help="search user id from email address")
+@click.argument("query", type=str)
+@core_config_and_device_options
```
touilleMan, line 32: Would should add a `--include-revoked` flag, so that by default we only display non-revoked ids (the typical usecase for this comand being to retreive the current user_id for the given human handle)

```
$ parsec core human_find zack
Zack Snyder <zack@example.com> - UserID: acd17910ca5b4d3dbce4cc29a79fd2f8
$ parsec core human_find zack --include-rekoved
Zack Snyder <zack@example.com> - UserID: acd17910ca5b4d3dbce4cc29a79fd2f8
Zack Snyder <zack@example.com> - UserID: c7c7f53f7e8744d48ed3d106877bb20f (revoked)
```

touilleMan: And I don't think the other params (such as the pagination and omit_non_human) are useful to expose: they don't fit the usecase (pagination) or are too complex to explain (omit_non_human)

touilleMan (changes requested):

touilleMan (approved):
lgtm

PR #1943, opened 2021-12-07
[oxidation] Binding secret Key

jojoquivogn:


parsec/core/local_device.py
```diff
@@ -311,10 +311,8 @@ def _decrypt_ciphertext(data: dict) -> bytes:
 
         try:
             key, _ = derivate_secret_key_from_password(password, data["salt"])
-            print(key)
-            print("data= ", data)
             return key.decrypt(data["ciphertext"])
-        except CryptoError as exc:
+        except ValueError as exc:
```
jojoquivogn, line 315: should i left as a ValueError or send a CryptoError with rust

oxidation/libparsec_python/src/crypto.rs
```diff
@@ -2,8 +2,11 @@
 
 use pyo3::basic::CompareOp;
 use pyo3::exceptions::PyValueError;
+use pyo3::import_exception;
 use pyo3::prelude::*;
-use pyo3::types::{PyByteArray, PyBytes};
+use pyo3::types::{PyByteArray, PyBytes, PyType};
+
+import_exception!(nacl.exceptions, CryptoError);
```
vxgmichel, line 9: Nice find!

oxidation/libparsec_python/src/crypto.rs
```diff
@@ -52,3 +55,60 @@ impl HashDigest {
...
+    }
+
+    #[classmethod]
+    fn generate(_cls: &PyType) -> PyResult<SecretKey> {
+        Ok(SecretKey(parsec_api_crypto::SecretKey::generate()))
+    }
+
+    #[getter]
+    fn secret<'p>(&self, py: Python<'p>) -> PyResult<&'p PyBytes> {
+        Ok(PyBytes::new(py, self.0.as_ref()))
```
vxgmichel, line 80: What about returning a `PyResult<&u8>` and letting pyo3 perform the `PyBytes` conversion?

oxidation/libparsec_python/src/crypto.rs
```diff
@@ -52,3 +55,60 @@ impl HashDigest {
...
+    }
+
+    pub fn encrypt<'p>(&self, py: Python<'p>, data: &[u8]) -> Result<&'p PyBytes, PyErr> {
```
vxgmichel, line 83: Why a `Result` and not a `PyResult` like other methods?

oxidation/libparsec_python/src/crypto.rs
```diff
@@ -52,3 +55,60 @@ impl HashDigest {
...
+        }
+    }
+
+    pub fn hmac<'p>(
+        &self,
+        py: Python<'p>,
+        data: &[u8],
+        digest_size: usize,
+    ) -> PyResult<&'p PyByteArray> {
+        Ok(PyByteArray::new(py, &self.0.hmac(data, digest_size)))
```
vxgmichel, line 100: Why returning a `PyByteArray`? Do we need the result to be mutable?

parsec/core/fs/workspacefs/file_transactions.py
```diff
@@ -111,7 +111,7 @@ async def _read_chunk(self, chunk: Chunk) -> bytes:
         return data[chunk.start - chunk.raw_offset : chunk.stop - chunk.raw_offset]
 
     async def _write_chunk(self, chunk: Chunk, content: bytes, offset: int = 0) -> int:
-        data = padded_data(content, offset, offset + chunk.stop - chunk.start)
+        data: bytes = padded_data(content, offset, offset + chunk.stop - chunk.start)
```
vxgmichel, line 114: I don't think that's necessary, probably an artefact

parsec/crypto.py
```diff
@@ -45,40 +45,62 @@
...
+        return self.secret == other.secret
 
     @classmethod
     def generate(cls) -> "SecretKey":
         return cls(random(SecretBox.KEY_SIZE))
 
     def __repr__(self):
         # Avoid leaking the key in logs
-        return f"<{type(self).__module__}.{type(self).__qualname__} object at {hex(id(self))}>"
+        return "SecretKey"
```
vxgmichel, line 70: Why changing the existing implementation?

parsec/crypto.py
```diff
@@ -137,6 +159,11 @@ def generate(cls, *args, **kwargs) -> "SigningKey":
     def __eq__(self, other):
         return isinstance(other, _SigningKey) and self._signing_key == other._signing_key
 
+    # Quick fix, waiting for binding signingKey
```
vxgmichel, line 162: How's that a quick fix? Looks fine to me.

tests/rust/test_secret_key.py
```diff
@@ -0,0 +1,39 @@
...
+
+    # We cannot assert "ciphered_rst == ciphered_py" because the encryption is not exactly the same
+
+    # Verify with both decryption
+    assert rst.decrypt(ciphered_rst) == py.decrypt(ciphered_py)
+    assert rst.decrypt(ciphered_py) == py.decrypt(ciphered_rst)
+
+    # TO-DO: test for hmac
+    # first we need to implement hmac in rust
+    # assert rst.hmac(data, digest_size) == py.hmac(data)
```
vxgmichel, line 35: Seems like `hmac` is implemented now :)

vxgmichel (commented):
A few comments and questions, but LGTM otherwise :+1:

oxidation/libparsec_python/src/crypto.rs
jojoquivogn: i just forgot to put PyResult instead of Result

parsec/crypto.py
jojoquivogn: we cannot use the existing implementation until secretKey and SigningKey are finished because of `tests/test_types.py::test_keys_dont_leak_on_repr`

parsec/core/fs/workspacefs/file_transactions.py
jojoquivogn: forget to remove that one

tests/rust/test_secret_key.py
jojoquivogn: `    // TODO
    pub fn hmac(&self, _data: &[u8], _digest_size: usize) -> Vec<u8> {
        // // blake2b(data, digest_size=digest_size, key=self, encoder=RawEncoder)
        // let key = blake2b::Key::from_slice(&self.0);
        // blake2b::derive_from_key(
        //     subkey: &mut [u8],
        //     subkey_id: u64,
        //     ctx: [u8; 8],
        //     key,
        // );
        vec![]
    }`
Not yet unfortunatly

parsec/crypto.py
jojoquivogn: cf: see the comment above

oxidation/libparsec_python/src/crypto.rs
jojoquivogn: no right, will change this

tests/rust/test_secret_key.py
jojoquivogn: done

oxidation/crates/parsec_api_crypto/src/rustcrypto/secret.rs
```diff
@@ -35,23 +39,19 @@ impl SecretKey {
...
-        // // blake2b(data, digest_size=digest_size, key=self, encoder=RawEncoder)
-        // let key = blake2b::Key::from_slice(&self.0);
-        // blake2b::derive_from_key(
-        //     subkey: &mut [u8],
-        //     subkey_id: u64,
-        //     ctx: [u8; 8],
-        //     key,
-        // );
-        vec![]
+    pub fn hmac(&self, data: &[u8]) -> Vec<u8> {
```
touilleMan, line 50: why remove the `digest_size` param ? it is used in the Python implementation

in theory we could truncate the returned vec to the size we want, but
1) it feels very hacky to do that in the caller and it may reduce hash randomness in a subtle way (for instance if the  underlying algorithm is later changed)
2) blake2 handling of `digest_size` is not as simple as trunking the output, so better not play smarter than him...

oxidation/crates/parsec_api_crypto/src/rustcrypto/secret.rs
```diff
@@ -35,23 +39,19 @@ impl SecretKey {
...
+        let mut hasher = Blake2bMac256::new_from_slice(&self.0).unwrap();
```
touilleMan, line 51: We should avoid the unwrap here, or comment why it is ok if it never fails

oxidation/crates/parsec_api_crypto/src/sodiumoxide/secret.rs
```diff
@@ -31,15 +31,17 @@ impl SecretKey {
     }
 
     pub fn decrypt(&self, ciphered: &[u8]) -> Result<Vec<u8>, &'static str> {
-        let nonce_slice = ciphered.get(..NONCEBYTES).ok_or("Invalid data size")?;
+        let nonce_slice = ciphered
+            .get(..NONCEBYTES)
+            .ok_or("The nonce must be exactly 24 bytes long")?;
```
touilleMan, line 36: Not sure we want to explain too much on why the decryption failed: data are either right or totally broken so the caller should not be tempted to return a detailed status that could help an attacker...

What do you think ?

oxidation/libparsec_python/src/crypto.rs
touilleMan: @Max-7 you might want to see this ;-)

parsec/core/fs/workspacefs/file_transactions.py
```diff
@@ -50,7 +50,7 @@ def padded_data(data: bytes, start: int, stop: int) -> bytes:
     if start <= stop <= 0:
         return b"\x00" * (stop - start)
     if 0 <= start <= stop:
-        return data[start:stop]
+        return bytes(data)[start:stop]
```
touilleMan, line 53: why this is needed ? can't we do like what is done in HashDigest binding instead ?

https://github.com/Scille/parsec-cloud/blob/24a2851b33063d55f2ce956c0decf52f06f199ac/oxidation/libparsec_python/src/crypto.rs#L24-L31

parsec/crypto.py
touilleMan: so is there a TODO to add ? from my point of view it is fine to keep the repr of all the crypto classes to something dead simple like you did.

I would just display it as something like `SecretKey(<redacted>)` (given the parenthesis with attributes inside is the most common way to represent an object)

touilleMan (changes requested):

oxidation/crates/parsec_api_crypto/src/sodiumoxide/secret.rs
jojoquivogn: we can't change this error message as long as cryptoBox send back this error

tests/rust/test_secret_key.py
```diff
@@ -0,0 +1,37 @@
...
+    ciphered_py = py.encrypt(MESSAGE)
+
+    # We cannot assert "ciphered_rs == ciphered_py"
+    # Encryption is not idempotent (a nonce is involved), hence comparison is not possible
+
+    # Verify with both decryption
+    assert rs.decrypt(ciphered_rs) == py.decrypt(ciphered_py)
+    assert rs.decrypt(ciphered_py) == py.decrypt(ciphered_rs)
+
+    # check if we get the sane hmac
```
touilleMan, line 32: "the same"

parsec/crypto.py
touilleMan: I would also remove the "quick fix" comment given the change is fine (and saying quick fix imply something is fishy ^^)

oxidation/crates/parsec_api_crypto/src/sodiumoxide/secret.rs
```diff
@@ -31,15 +31,17 @@ impl SecretKey {
...
         let nonce = Nonce::from_slice(nonce_slice).ok_or("Invalid data size")?;
         let plaintext =
             open(&ciphered[NONCEBYTES..], &nonce, &self.0).or(Err("Decryption error"))?;
         Ok(plaintext)
     }
 
     // TODO
-    pub fn hmac(&self, _data: &[u8], _digest_size: usize) -> Vec<u8> {
+    pub fn hmac(&self, _data: &[u8], digest_size: usize) -> Vec<u8> {
```
touilleMan, line 44: no leading `_` should lead to warning given the variable is not used ?

touilleMan: Also can you add a `unimplemented!()` in the body

touilleMan (approved):

PR #1944, opened 2021-12-07
Updating importlib resources 5.4.0

alexandrediasldev:


parsec/backend/http.py
```diff
@@ -177,7 +177,13 @@ async def _http_static(self, req: HTTPRequest, **kwargs: str) -> HTTPResponse:
         try:
             # Note we don't support nested resources, this is fine for the moment
             # and it prevent us from malicious path containing `..`
-            data = importlib_resources.read_binary(http_static_module, path)
+            path_base = importlib_resources.files(http_static_module).resolve(strict=True)
+            full_path = path_base.joinpath(path).resolve(strict=True)
+            if str(full_path).startswith(str(path_base)):
+                data = full_path.read_bytes()
+            else:
+                raise ValueError()
```
vxgmichel, line 185: I see a few problems with this implementation:
- `.files()` returns a `Traversable` object that is not guaranteed to have a `resolve` method
- since it is a sensitive code from a security standpoint, the logic should probably be made more clear

For instance here's how `importlib_resources._legacy` implements `read_binary`:
```python
(importlib_resources.files(package) / normalize_path(resource)).read_bytes()
```
with `normalize_path` being defined as such:
https://github.com/python/importlib_resources/blob/bc6f0d19007b33568a44d67eb56358f63b6b7a97/importlib_resources/_legacy.py#L30-L40

vxgmichel (changes requested):
A single change request, LGTM otherwise :+1:

vxgmichel (approved):
LGTM :+1:

PR #1952, opened 2021-12-12
Fix regression in test_expired_organization introduced in #1894

vxgmichel:
Thanks @alexandrediasldev for the report


alexandrediasldev (approved):
LGTM :+1:

Issue #1937, opened 2021-12-04
Add size limit for device_label


touilleMan:
Unlike human_handle's label which is limited to 255 bytes, device_label [is arbitrary sized](https://github.com/Scille/parsec-cloud/blob/f82e3d566aaeb0948bbe155aa59518802149de1f/parsec/api/data/certif.py#L131)

We should add a similar check, and implement it handling in the GUI.
Regarding backward compatibility, I guess we can say a >255 bytes device_label is just begging for trouble anyway ^^

PR #1972, opened 2021-12-20
Fix device_label size

jojoquivogn:
https://github.com/Scille/parsec-cloud/issues/1937

### DIFF
--- a/parsec/api/protocol/types.py
+++ b/parsec/api/protocol/types.py
@@ -98,10 +98,21 @@ class DeviceID(str):
         return cls(f"{uuid4().hex}@{uuid4().hex}")
 
 
+class DeviceLabel(str):
+    def __new__(cls, device_label: str) -> "DeviceLabel":
+        device_label = normalize("NFC", device_label)
+
+        if not 0 < _bytes_size(device_label) < 255:
+            raise ValueError("Invalid device_label")
+
+        return super(DeviceLabel, cls).__new__(cls)
+
+
 OrganizationIDField = fields.str_based_field_factory(OrganizationID)
 UserIDField = fields.str_based_field_factory(UserID)
 DeviceNameField = fields.str_based_field_factory(DeviceName)
 DeviceIDField = fields.str_based_field_factory(DeviceID)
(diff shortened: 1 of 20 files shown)


parsec/api/protocol/types.py
```diff
@@ -98,6 +98,22 @@ def new(cls: Type[DeviceIDTypeVar]) -> DeviceIDTypeVar:
...
+    def __new__(cls, device_label: str) -> "DeviceLabel":
+        device_label = normalize("NFC", device_label)
+
+        if not 0 < _bytes_size(device_label) < 255:
+            raise ValueError("Invalid device_label")
+
+        return super(DeviceLabel, cls).__new__(cls, device_label)
+
+
+class DeviceLabelField(fields.String):  # type: ignore
```
jojoquivogn, line 111: `# type: ignore` for avoid mypy strange error (only quick fix): 
```
parsec/api/protocol/types.py:111: error: Class cannot subclass 'String' (has type 'Any')  [misc]
parsec/api/protocol/types.py:111: error: Base type String becomes "Any" due to an unfollowed import  [no-any-unimported]
```

parsec/api/protocol/types.py
```diff
@@ -98,6 +98,22 @@ def new(cls: Type[DeviceIDTypeVar]) -> DeviceIDTypeVar:
         return cls(f"{uuid4().hex}@{uuid4().hex}")
 
 
+class DeviceLabel(str):
+    def __new__(cls, device_label: str) -> "DeviceLabel":
+        device_label = normalize("NFC", device_label)
+
+        if not 0 < _bytes_size(device_label) < 255:
+            raise ValueError("Invalid device_label")
```
touilleMan, line 106: message should be `"Invalid device label"`

parsec/api/protocol/types.py
touilleMan: why not using `str_based_field_factory` ?

```python
DeviceLabelField = fields.str_based_field_factory(DeviceLabel)
```

touilleMan (approved):

touilleMan (approved):
image

PR #1980, opened 2021-12-31
Rename extra core and backend to desktop and server

### LINKED ISSUE #1587: Run core without core extra requirements installed

alexandrediasldev:


touilleMan:
Outdated since we switched to poetry ;-)

PR #1989, opened 2022-01-06
Expose the workspace cache size in the configuration file

### LINKED ISSUE #1983: Make the default block cache size configurable
At the moment, this value is hardcoded at 500MB:
```python
DEFAULT_BLOCK_CACHE_SIZE = 512 * 1024 * 1024
```
It could easily be exposed through the config file.

vxgmichel:
Fix #1983

### DIFF
--- a/parsec/core/config.py
+++ b/parsec/core/config.py
@@ -12,6 +12,7 @@ import base64
 from parsec.api.data import EntryID
 from parsec.core.types import BackendAddr
 
+DEFAULT_WORKSPACE_STORAGE_CACHE_SIZE = 512 * 1024 * 1024
 
 logger = get_logger()
 
@@ -60,6 +61,7 @@ class CoreConfig:
 
     sentry_url: Optional[str] = None
     telemetry_enabled: bool = True
+    workspace_storage_cache_size: int = DEFAULT_WORKSPACE_STORAGE_CACHE_SIZE
 
     gui_last_device: Optional[str] = None
     gui_tray_enabled: bool = True
@@ -95,6 +97,7 @@ def config_factory(
     backend_max_connections: int = 4,
     sentry_url: str = None,
     telemetry_enabled: bool = True,
+    workspace_storage_cache_size: int = DEFAULT_WORKSPACE_STORAGE_CACHE_SIZE,
     debug: bool = False,
     gui_last_device: str = None,
     gui_tray_enabled: bool = True,
@@ -122,11 +125,14 @@ def config_factory(
             "parsec://localhost:6777?no_ssl=true"
         )
 
+    if mountpoint_base_dir is None:
+        mountpoint_base_dir = get_default_mountpoint_base_dir(environ)
+
     data_base_dir = data_base_dir or get_default_data_base_dir(environ)
     core_config = CoreConfig(
         config_dir=config_dir or get_default_config_dir(environ),
         data_base_dir=data_base_dir,
-        mountpoint_base_dir=get_default_mountpoint_base_dir(environ),
+        mountpoint_base_dir=mountpoint_base_dir,
         prevent_sync_pattern_path=prevent_sync_pattern_path,
         mountpoint_enabled=mountpoint_enabled,
         disabled_workspaces=disabled_workspaces,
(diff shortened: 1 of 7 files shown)


parsec/core/config.py
```diff
@@ -122,18 +125,22 @@ def config_factory(
             "parsec://localhost:6777?no_ssl=true"
         )
 
+    if mountpoint_base_dir is None:
```
touilleMan, line 128: This is an unrelated fix ?

parsec/core/fs/storage/chunk_storage.py
```diff
@@ -230,6 +238,14 @@ async def set_chunk(self, chunk_id: ChunkID, raw: bytes) -> None:
...
+            # Perform cleanup if necessary
+            await self.cleanup(cursor)
+
+    async def cleanup(self, cursor: Optional[Cursor] = None) -> None:
+
+        # Update database
+        async with self._reenter_cursor(cursor) as cursor:
+
             # Count the chunks
             cursor.execute("SELECT COUNT(*) FROM chunks")
```
touilleMan, line 250: Not directly related to this PR, but it seems very costly to me to have this request being made every time we insert a new chunk (the query itself should be very fast, be we pay the query parsing, python-to-c boundary crossing with GIL release, SQlite lock acquire etc.). Wouldn't it be better to add the `SELECT COUNT(*) FROM chunks` as an aditional `VALUES` to the `INSERT OR REPLACE` query instead ?

parsec/core/fs/storage/workspace_storage.py
```diff
@@ -251,14 +250,21 @@ async def run(
         data_base_dir: Path,
         device: LocalDevice,
         workspace_id: EntryID,
-        cache_size: int = DEFAULT_BLOCK_CACHE_SIZE,
+        cache_size: int = DEFAULT_WORKSPACE_STORAGE_CACHE_SIZE,
         vacuum_threshold: int = DEFAULT_CHUNK_VACUUM_THRESHOLD,
```
touilleMan, line 254: so this parameter is only used for `data_localdb` and not for `cache_locadb`, do we need to change it name to `data_vacuum_threshold` ?

touilleMan (approved):

parsec/core/config.py
vxgmichel: Yes, I noticed that the `mountpoint_base_dir` argument was unused.

parsec/core/fs/storage/workspace_storage.py
vxgmichel: Right :+1:

parsec/core/fs/storage/chunk_storage.py
vxgmichel: I don't think it's worth it, the previous query takes about 50 times longer so there's not much too gain:
```python
In [1]: import sqlite3, uuid, time

In [2]: path = "/tmp/tmpqh61dj4l/share/parsec/9b3bc58589#Org#4418c3e2a3144db2baa1c9ae6041ed46@31de682070ea489f9bf201cf625af8ea/e190f567-0fe3-4f
   ...: a2-b104-736752f4825f/workspace_cache-v1.sqlite"

In [3]: c = sqlite3.connect(path)

In [6]: data = b"a" * 512 * 1024

In [7]: %%time
   ...: c.execute(
   ...:   """INSERT OR REPLACE INTO
   ...:   chunks (chunk_id, size, offline, accessed_on, data)
   ...:   VALUES (?, ?, ?, ?, ?)""",
   ...:   (uuid.uuid4().bytes, len(data), False, time.time(), data),
   ...: )
   ...: 
   ...: 
CPU times: user 0 ns, sys: 3.77 ms, total: 3.77 ms
Wall time: 7.58 ms
Out[7]: <sqlite3.Cursor at 0x7eff41acac70>

In [8]: %time print(c.execute("SELECT COUNT(*) FROM chunks").fetchone())
(956,)
CPU times: user 117 µs, sys: 5 µs, total: 122 µs
Wall time: 167 µs

parsec/core/fs/storage/workspace_storage.py
vxgmichel: 1ebdd468e

PR #1996, opened 2022-01-12
Fix compatiblity issue with older backend introduced in PR #1894

vxgmichel:


alexandrediasldev (approved):
LGTM

parsec/api/protocol/base.py
```diff
@@ -80,13 +80,23 @@ class RequireGreaterTimestampRepSchema(BaseRepSchema):
...
     # even though `timestamp_out_of_ballpark` would be more explicit
     status: fields.CheckedConstant = fields.CheckedConstant("bad_timestamp", required=True)
-    ballpark_client_early_offset = fields.Float(required=True, allow_none=False)
-    ballpark_client_late_offset = fields.Float(required=True, allow_none=False)
-    client_timestamp = fields.DateTime(required=True, allow_none=False)
-    backend_timestamp = fields.DateTime(required=True, allow_none=False)
+    ballpark_client_early_offset = fields.Float(required=False, missing=None)
+    ballpark_client_late_offset = fields.Float(required=False, missing=None)
+    client_timestamp = fields.DateTime(required=False, missing=None)
+    backend_timestamp = fields.DateTime(required=False, missing=None)
```
touilleMan, line 99: Can't we use a `@pre_load` schema decorator instead ? This way all the legacy handling would be contained there instead of being scattered across the codebase.

On top of that it will be much easier to remove this corner case in the future once the versions<2.4 become old enoughh

touilleMan: After more discussion with @vxgmichel, it appear those values cannot be easily generated if they are missing: this would imply choosing an arbitrary ballpark, then a `client_timestamp` (while the client may have kept the original `client_timestamp`...) and do a cheating aritmetic so that the `backend_timestamp` makes things out of ballpark :/

parsec/api/protocol/base.py
```diff
@@ -80,13 +80,23 @@ class RequireGreaterTimestampRepSchema(BaseRepSchema):
 
 
 class TimestampOutOfBallparkRepSchema(BaseRepSchema):
+    """This schema has been added to API version 2.4.
```
touilleMan, line 83: We should add a `TODO: This backward compatibility should be removed once Parsec<2.4 support is dropped`

touilleMan (approved):

parsec/api/protocol/base.py
vxgmichel: :+1:

vxgmichel: 3eab1f05ffae0aa8d3997b1b95dea3b113f00a3d

PR #2028, opened 2022-02-11
Prevent Qt from parsing special characters in navigation bar

### LINKED ISSUE #2025: Parse errors on some folder names in the navigation bar
When opening folders or files inside some folders, the following error can occur:
`Could not parse stylesheet of object ClickableLabel({address}, name = "{folder name}")`

Ironicbay:
Fixes #2025 

With the current code, if a folder has special characters in its name, Qt is unable to parse them during the `setStyleSheet` call resulting in errors and the name not being displayed in the navigation bar.

Proposed solution is to replace said name by the numerical index in the loop, not changing the navigation bar behaviour and correctly setting the colors.

### DIFF
--- a/parsec/core/gui/navigation_bar_widget.py
+++ b/parsec/core/gui/navigation_bar_widget.py
@@ -95,7 +95,7 @@ class NavigationBarWidget(QWidget):
                 )
             label = None
             if idx != len(parts) - 1:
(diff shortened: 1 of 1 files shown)


touilleMan (approved):
👌

Max-7 (approved):

PR #2042, opened 2022-02-21
[oxidation] Fix missing timestamp param in WorkspaceEntry::new

touilleMan:


oxidation/crates/parsec_api_types/src/manifest.rs
```diff
@@ -205,20 +205,19 @@ pub struct WorkspaceEntry {
 }
 
 impl WorkspaceEntry {
-    pub fn new(name: &EntryName) -> Self {
-        let now = Utc::now();
+    pub fn new(name: &EntryName, timestamp: &DateTime<Utc>) -> Self {
```
TimeEngineer, line 208: It is better to give the ownership directly, because it is required in the function
In that case, developper can give name without reallocation and DateTime<Utc> implement Copy

touilleMan: Good point !

I missed the fact DateTime was a copy while having a look at the source code given it is declared as the regular `#[derive(Copy)]`: https://docs.rs/chrono/latest/src/chrono/datetime.rs.html#648-649 😄

touilleMan: done.

TimeEngineer: I suggested to change `name: &EntryName` to `name: EntryName` too

touilleMan: done.

TimeEngineer: I guess the rule of thumb is something like "pass Copy by value and Clone by reference unless you are making a constructor" - Yes, specifically the function signature must match the usage of the arguments
A copiable value is often light enough to be as fast as a referencing/dereferencing.

You can find an idiomatic example for a builder where he/she uses string as input: https://rust-unofficial.github.io/patterns/patterns/creational/builder.html

TimeEngineer (approved):

PR #2047, opened 2022-02-22
update handshake challenge

jojoquivogn:


parsec/api/protocol/handshake.py
```diff
@@ -461,8 +476,16 @@ def __init__(
     def process_challenge_req(self, req: bytes) -> bytes:
         self.load_challenge_req(req)
         challenge = self.challenge_data["challenge"]
+
         assert isinstance(challenge, bytes)
-        answer = self.user_signkey.sign(challenge)
+
+        # TO-DO remove the else for the next release
+        if self.backend_api_version.version >= 2 and self.backend_api_version.revision >= 5:
```
vxgmichel, line 483: The following code looks a bit cleaner:
 `if self.backend_api_version >= (2, 5)`

parsec/api/protocol/handshake.py
```diff
@@ -346,7 +353,15 @@ def build_result_req(self, verify_key: Optional[VerifyKey] = None) -> bytes:
             try:
                 answer = self.answer_data["answer"]
                 assert isinstance(answer, bytes)
-                returned_challenge = verify_key.verify(answer)
+
+                # TO-DO remove the else for the next release
```
vxgmichel, line 357: I don't think it's necessary to remove the `else` statement in the next release, since supporting older clients doesn't seem to be a problem in that case (contrary to the newer client/older backend scenario which has a security issue)

tests/backend/test_compatibility.py
```diff
@@ -66,3 +77,58 @@ class OlderHandshakeChallengeSchema(BaseSchema):
     # Backend API < 2.4 with newer clients
     data = older_handshake_challenge_serializer.dumps(old_data)
     assert handshake_challenge_serializer.loads(data) == compat_data
+
+
+# This test would be useless when all clients and server will be up to date
```
vxgmichel, line 82: As mentioned above, I would keep this test for older client/newer backend compatibility.

tests/backend/test_compatibility.py
```diff
@@ -66,3 +77,58 @@ class OlderHandshakeChallengeSchema(BaseSchema):
...
+def test_handshake_challenge_client_server_compatibility(mallory, alice, monkeypatch):
+    ash = ServerHandshake()
```
vxgmichel, line 84: Explicit naming please :)

tests/backend/test_compatibility.py
```diff
@@ -66,3 +77,58 @@ class OlderHandshakeChallengeSchema(BaseSchema):
...
+        "ballpark_client_late_offset": BALLPARK_CLIENT_LATE_OFFSET,
+    }
+
+    monkeypatch.setattr(bch, "SUPPORTED_API_VERSIONS", [client_version])
+
+    bsh.build_challenge_req()
+    answer_req = bch.process_challenge_req(packb(req))
+
+    answer = handshake_answer_serializer.loads(answer_req)
+    assert mallory.verify_key.verify(answer["answer"]) == challenge
```
vxgmichel, line 134: Nice test :+1:

vxgmichel (approved):
Minor comments and a question for @touilleMan , but LGTM otherwise :+1:

parsec/api/protocol/handshake.py
```diff
@@ -346,7 +353,15 @@ def build_result_req(self, verify_key: Optional[VerifyKey] = None) -> bytes:
...
+                if self.client_api_version.version >= 2 and self.client_api_version.revision >= 5:
```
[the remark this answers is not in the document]
touilleMan: the original code is broken when we reach version `3.0`

parsec/api/protocol/handshake.py
```diff
@@ -461,8 +476,16 @@ def __init__(
     def process_challenge_req(self, req: bytes) -> bytes:
         self.load_challenge_req(req)
         challenge = self.challenge_data["challenge"]
+
         assert isinstance(challenge, bytes)
-        answer = self.user_signkey.sign(challenge)
+
+        # TO-DO remove the else for the next release
+        if self.backend_api_version.version >= 2 and self.backend_api_version.revision >= 5:
+            # TO-DO Need to use "BaseSignedData" ?
```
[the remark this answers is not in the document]
touilleMan: This seems fine enough to me ;-)

The only drawback I see is the fact that sign + dump must be done by the caller instead of behind provided as a clear api. But given this is only used in a single place (and this is not going to change), it is not a big issue.

If you want to avoid this, you could create your own class `AnswerSerializer` inheriting `MsgpackSerializer` which would define `dump_and_sign/verify_and_load` methods.

PR #2050, opened 2022-02-23
oxidation api protocol schema

TimeEngineer:


oxidation/crates/parsec_api_protocol/src/lib.rs
```diff
@@ -0,0 +1,44 @@
...
+pub use message::*;
+pub use organization::*;
+pub use ping::*;
+pub use realm::*;
+pub use user::*;
+pub use vlob::*;
+
+use serde::{Deserialize, Serialize};
+
+#[macro_export]
```
touilleMan, line 25: no need for the macro_export given the macro won't be used outside of the crate

you can add a `pub(crate) use impl_api_protocol_dump_load;` after the macro definition instead (cf. `new_uuid_type` macro)

touilleMan (approved):

PR #2051, opened 2022-02-24
Oxidation of remote manifests

Max-7:
Depends on #2048


touilleMan (approved):
lgtm

parsec/api/data/entry.py
```diff
@@ -34,7 +34,7 @@ class EntryID(UUID4):
 
 class EntryName(StrBased):
     # Ignore the REGEX
-    REGEX = None
+    REGEX = re.compile(".*")
```
touilleMan, line 37: Should modify typing for `StrBased` to `Optional[Pattern[str]]` instead

Max-7: Fixed

parsec/core/mountpoint/winfsp_operations.py
```diff
@@ -363,6 +365,11 @@ def read_directory(self, file_context, marker):
         # NOTE: The "." and ".." directories should ONLY be included
         # if the queried directory is not root
 
+        print(
+            f"READ DIR marker: {marker}, file ctx: {file_context.path!r} (is_root: {file_context.path.is_root()})"
+        )
+        print(f"READ DIR stats: {stat!r}")
+
```
vxgmichel, line 372: Debug prints

PR #2052, opened 2022-02-25
Improve schemas documentation & fix compat tests

### LINKED ISSUE #2300: Improve schemas documentation & fix compat tests

touilleMan:
fix #2300


vxgmichel (approved):
LGTM :+1:

tests/schemas/builder.py
```diff
@@ -144,15 +144,27 @@ def generate_api_data_specs():
...
 
-    return {data_cls.__name__: data_class_to_spec(data_cls) for data_cls in data_classes}
+    return {
+        **{data_cls.__name__: data_class_to_spec(data_cls) for data_cls in data_classes},
+        # Hack to include device file serialization
+        **{
+            serializer.schema.__class__.__name__: {
+                "serializing": "msgpack"
+                if serializer.__class__.__name__.startswith("Msgpack")
+                else "error!!!",
```
vxgmichel, line 162: Should this be an `assert`?

parsec/api/protocol/base.py
```diff
@@ -128,10 +128,19 @@ class TimestampOutOfBallparkRepSchema(BaseRepSchema):
...
+    backend_timestamp = fields.DateTime(required=False, allow_none=False)
+
+    @post_load
+    def make_obj(self, data: Dict[str, Any]) -> Dict[str, Any]:  # type: ignore[misc]
+        # Cannot use `missing=None` with `allow_none=False`
+        data.setdefault("ballpark_client_early_offset", None)
+        data.setdefault("ballpark_client_late_offset", None)
+        data.setdefault("client_timestamp", None)
+        data.setdefault("backend_timestamp", None)
+        return data
```
vxgmichel, line 143: If the logic doesn't change, shouldn't we keep the original version as it seems more explicit?

vxgmichel (commented):

parsec/api/protocol/base.py
touilleMan: My commit comment was wrong: the original code uses `allow_none=True`, cf. what is done in marshmallow:

```python
        # If missing=None, None should be considered valid by default
        if allow_none is None:
            if missing is None:
                self.allow_none = True
            else:
                self.allow_none = False
        else:
            self.allow_none = allow_none
```

This is typically for this kind of "too smart for it own good" code that we need to be explicit about required/allow_none !

I'm changing the commit comment then

tests/schemas/builder.py
```diff
@@ -83,7 +89,15 @@ def field_to_spec(field):
 
 
 def schema_to_spec(schema):
-    return {"fields": {k: field_to_spec(v) for k, v in schema.fields.items()}}
+    if isinstance(schema, OneOfSchema):
+        if schema.fallback_type_schema:
+            breakpoint()
```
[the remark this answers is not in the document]
touilleMan: done

tests/schemas/builder.py
touilleMan: good idea, done !

PR #2058, opened 2022-03-01
oxidation add test api protocol

TimeEngineer:


touilleMan (changes requested):

oxidation/crates/parsec_api_protocol/src/block.rs
```diff
@@ -27,8 +27,14 @@ impl_api_protocol_dump_load!(BlockCreateReqSchema);
...
-pub struct BlockCreateRepSchema {
-    pub status: Status,
+#[serde(tag = "status", rename_all = "snake_case")]
+pub enum BlockCreateRepSchema {
+    Ok,
+    AlreadyExists,
+    NotFound,
+    Timeout,
+    NotAllowed,
+    InMaintenance,
```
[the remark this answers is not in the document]
TimeEngineer: Done :+1:

oxidation/crates/parsec_api_protocol/src/events.rs
```diff
@@ -34,7 +34,6 @@ pub enum APIEvent {
 
 #[derive(Debug, Clone, Serialize, Deserialize, PartialEq, Eq)]
 pub struct EventsPingedRepSchema {
```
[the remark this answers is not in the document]
TimeEngineer: Done :+1:

oxidation/crates/parsec_api_protocol/src/realm.rs
```diff
@@ -143,8 +160,16 @@ impl_api_protocol_dump_load!(RealmUpdateRolesReqSchema);
  */
 
 #[derive(Debug, Clone, Serialize, Deserialize, PartialEq, Eq)]
-pub struct RealmUpdateRolesRepSchema {
-    pub status: Status,
+#[serde(tag = "status", rename_all = "snake_case")]
+pub enum RealmUpdateRolesRepSchema {
+    Ok,
+    NotAllowed { reason: Option<String> },
```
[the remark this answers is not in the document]
TimeEngineer: Done :+1:

oxidation/crates/parsec_api_protocol/src/vlob.rs
```diff
@@ -64,18 +69,27 @@ impl_api_protocol_dump_load!(VlobReadReqSchema);
...
+pub enum VlobReadRepSchema {
+    Ok {
+        version: u64,
+        #[serde_as(as = "Bytes")]
+        blob: Vec<u8>,
+        author: DeviceID,
+        timestamp: DateTime,
+        // This field is used by the client to figure out if its role certificate cache is up-to-date enough
+        // to be able to perform the proper integrity checks on the manifest timestamp.
+        // The `missing=None` argument is used to provide compatibilty of new clients with old backends.
```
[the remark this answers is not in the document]
TimeEngineer: Done :+1:

oxidation/crates/parsec_api_protocol/src/block.rs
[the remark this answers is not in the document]
touilleMan: 1) The fix should be done on all the enum reps.

2) on second thought, I think `#[non_exhaustive]` is not the right way to go here:

`#[non_exhaustive]` typically use by library to create a future-proof API. in our case those enums are strictly intern to libparsec, so there is no need for future-proofing (if a new enum value is added compilitaion will be broken until the the match at caller site is updated, which is something trivial to do)

The need we have here is that in the future the server might send to our client a response with an unknown status, in this cas we don't want to have a serialization error but to fallback to a default error type.

So in the end we should have:
- an `UnknownError { error: String}` enum value for each rep enum
- a test for each rep enum to ensure unknown rep status are correctly translated to `UnknownError`

oxidation/crates/parsec_api_protocol/src/lib.rs
```diff
@@ -10,36 +11,24 @@ mod realm;
...
-}
-
-pub(crate) use impl_api_protocol_dump_load;
-
-#[derive(Debug, Clone, Serialize, Deserialize, PartialEq, Eq)]
-#[serde(rename_all = "lowercase")]
-pub enum Status {
-    Ok,
-}
+pub use events::APIEvent;
```
touilleMan, line 25: What is the need for the `use xxx::*;` then `pub use xxx::{yyy, zzz}` pattern ?

is it to allow to have simpler use in the sub modules (e.g. can do `use crate::yyy` instead of `use crate::xxx::yyy`) ?

What's the advantage over the regular `pub use xxx::°` ?

oxidation/crates/parsec_api_protocol/src/cmds.rs
```diff
@@ -0,0 +1,202 @@
+// Parsec Cloud (https://parsec.cloud) Copyright (c) BSLv1.1 (eventually AGPLv3) 2016-2021 Scille SAS
+
+macro_rules! cmds_bundle {
+    ($cmds: ident, [$(($path: path, $module: ident, $cmd: ident $(, $rename: literal)? $(,)?),)*]) => {
+
+        ::paste::paste! {
+            pub mod $cmds {
+                use serde::{Deserialize, Serialize};
```
touilleMan, line 8: Better use `::serde::...` (so with leading :: ) to ensure the path is absolute.

Here it is not useful given the macro is private, but I think it is better to always write it this way to avoid copy/paste or refactoring errors

touilleMan: and same thing for use of `rpm_serde::`

oxidation/crates/parsec_api_protocol/src/cmds.rs
```diff
@@ -0,0 +1,202 @@
+// Parsec Cloud (https://parsec.cloud) Copyright (c) BSLv1.1 (eventually AGPLv3) 2016-2021 Scille SAS
+
+macro_rules! cmds_bundle {
```
touilleMan, line 3: Can you add a comment (typically showing the load/dumps code for client/server side) to explain the rational behind this weird nested modules ? ;-)

oxidation/crates/parsec_api_protocol/tests/test_block.rs
```diff
@@ -19,39 +19,105 @@ fn serde_block_create_req() {
...
+    let req = authenticated_cmds::block_create::Req {
         block_id: "57c629b69d6c4abbaf651cafa46dbc93".parse().unwrap(),
         realm_id: "1d3353157d7d4e95ad2fdea7b3bd19c5".parse().unwrap(),
         block: b"foobar".to_vec(),
     };
 
-    let schema = BlockCreateReqSchema::load(&data).unwrap();
+    let expected = authenticated_cmds::AnyCmdReq::BlockCreate(req.clone());
+
+    let schema = authenticated_cmds::AnyCmdReq::loads(&data).unwrap();
```
touilleMan, line 30: the name `schema` is not valid, should be `data` (especially given we have `data2` later on ^^)

This `schema` name is a bit everywhere in the tests. I suggest the following convention:
- `raw` for serialized stuff (this is already what we use in the codebase and python tests), `raw_xxx` if there is multiple raw elements in the test
- `data` for deserialized
- `req` for `authenticated_cmds::block_create::Req`
- `any_req` for `authenticated_cmds::AnyCmdReq`
- `expected` alone when the expected stuff is obvious, other wise `expected_xxx`

oxidation/crates/parsec_api_protocol/src/cmds.rs
TimeEngineer: Weird ? :joy: Ok :+1:

oxidation/crates/parsec_api_protocol/src/lib.rs
TimeEngineer: _is it to allow to have simpler use in the sub modules (e.g. can do use crate::yyy instead of use crate::xxx::yyy) ?_ - yes
_What's the advantage over the regular pub use xxx::° ?_ - It's easier to import as you said

oxidation/crates/parsec_api_protocol/src/cmds.rs
TimeEngineer: Done :+1:

oxidation/crates/parsec_api_protocol/src/cmds.rs
TimeEngineer: Done :+1:

oxidation/crates/parsec_api_protocol/tests/test_block.rs
TimeEngineer: Done :+1:

oxidation/crates/parsec_api_protocol/src/block.rs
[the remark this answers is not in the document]
TimeEngineer: Done :+1:
But only one test is enough as long as the macro checks if UnknownError is there

oxidation/crates/parsec_api_protocol/src/lib.rs
touilleMan: Considering:

```rust
// spam/mod.rs
use foo::*;
pub use foo::Public1;

// spam/foo.rs
struct Private1 {};
pub struct Public1 {};
pub struct Public2 {};
```

Here we have `Public2` this is us marked `pub`, but the caller has to do `spam::foo::Public2` to access it.
This might be a good thing (for instance if the API is divided between high level and a low level parts, this way low level stuff is in a sub module to ease readability)

However in our current case there is no need for this, so I don't see the value of separating `use` vs `pub use` over just doing `pub use foo::*` (given `pub use` only expose pub stuff anyway)

oxidation/crates/parsec_api_protocol/src/lib.rs
```diff
@@ -10,36 +11,60 @@ mod realm;
...
+            }
 
-#[derive(Debug, Clone, Serialize, Deserialize, PartialEq, Eq)]
-#[serde(rename_all = "lowercase")]
-pub enum Status {
-    Ok,
+            pub fn loads(buf: &[u8]) -> Self {
+                match ::rmp_serde::from_read_ref(buf) {
+                    Ok(res) => res,
+                    Err(e) => Self::UnknownError {
```
touilleMan, line 60: So this is the part that ensure `UnknownError` must be part of the Rep ^^

oxidation/crates/parsec_api_protocol/src/cmds.rs
```diff
@@ -0,0 +1,205 @@
...
+        (
+            crate::invite,
+            invite_4_claimer_communicate,
+            Invite4ClaimerCommunicate,
+            "invite_4_claimer_communicate",
+        ),
+    ]
+);
+
+#[test]
```
touilleMan, line 196: What do you think of moving this test within the macro definition so that we run this check for each command !

I think it's interesting to have this test even if the `impl_dumps_loads_for_rep` already fail to compile if `UnknownError` is not present given it makes it very explicit that `UnknownError` must be defined for each rep.

touilleMan (approved):

oxidation/crates/parsec_api_protocol/src/cmds.rs
TimeEngineer: Done :+1:

PR #2064, opened 2022-03-04
Oxidation local manifest binding

TimeEngineer:


Max-7 (changes requested):
A lot (and I mean A LOT) of tests are missing. You are only testing the constructor and the evolve() of LocalFileManifest, LocalFolderManifest, LocalWorkspaceManifest and LocalUserManifest. Yet each of these classes as a lot more methods (is_reshaped, assert_integrity, from_remote, to_remote, shenanigans with the confinement, ...).
You are also missing tests for Chunk and ChunkID.

Every function that exists in Python has to be compared to its Rust implementation to make sure that their behavior is the same.

tests/core/fs/test_fs_online_user.py
```diff
@@ -70,7 +70,7 @@ async def rename_workspace(self, workspace, new_name):
                 await self.user_fs.workspace_rename(wid, EntryName(new_name))
             else:
                 with pytest.raises(FSWorkspaceNotFoundError):
-                    await self.user_fs.workspace_rename(workspace, EntryName(new_name))
+                    await self.user_fs.workspace_rename(EntryID.new(), EntryName(new_name))
```
TimeEngineer, line 73: Why ?

oxidation/crates/parsec_client_types/src/local_manifest.rs
```diff
@@ -55,6 +56,77 @@ pub struct Chunk {
...
+        });
+
+        Ok(result)
+    }
+
+    pub fn is_block(&self) -> bool {
+        // Requires an access
+        if let Some(access) = &self.access {
+            // Pseudo block
+            !(!self.is_pseudo_block()
```
touilleMan, line 102: double `!` brings poor readibility for me, I would prefer applying De Morgan's laws:
```rust
if let Some(access) = &self.access {
  // Pseudo block
  self.is_pseudo_block()
  // Offset inconsistent
  && self.raw_offset == access.offset
  // Size inconsistent
  && u64::from(self.raw_size) == access.size
} else {
  false
}
```

oxidation/crates/parsec_client_types/src/local_manifest.rs
```diff
@@ -55,6 +56,77 @@ pub struct Chunk {
...
+                || self.raw_offset != access.offset
+                // Size inconsistent
+                || u64::from(self.raw_size) != access.size)
+        } else {
+            false
+        }
+    }
+
+    pub fn is_pseudo_block(&self) -> bool {
+        if self.start != self.raw_offset {
```
touilleMan, line 113: missing comments `# Not left aligned/# Not right aligned` comments, @vxgmichel love those comments 😄

oxidation/crates/parsec_client_types/src/local_manifest.rs
```diff
@@ -111,6 +183,99 @@ impl From<LocalFileManifest> for LocalFileManifestData {
 
 impl_local_manifest_dump_load!(LocalFileManifest);
 
+impl LocalFileManifest {
+    pub fn get_chunks(&self, block: usize) -> Vec<Chunk> {
```
touilleMan, line 187: the return value should be `Vec<&Chunk>` considering how get_chunks is used in the python codebase

oxidation/crates/parsec_client_types/src/local_manifest.rs
```diff
@@ -111,6 +183,99 @@ impl From<LocalFileManifest> for LocalFileManifestData {
 
 impl_local_manifest_dump_load!(LocalFileManifest);
 
+impl LocalFileManifest {
```
touilleMan, line 186: image

oxidation/crates/parsec_client_types/src/local_manifest.rs
```diff
@@ -111,6 +183,99 @@ impl From<LocalFileManifest> for LocalFileManifestData {
...
+        })
+    }
+
+    pub fn to_remote(
+        &self,
+        author: DeviceID,
+        timestamp: DateTime,
+    ) -> Result<FileManifest, &'static str> {
+        self.assert_integrity();
+        assert!(self.is_reshaped());
```
touilleMan, line 243: I think the assert is here given `to_remote` is supposed to be used only in a specific situation

However we should be able to modify this into a better api once we start recoding the code calling this method, typically having to_remote returinng an error and having the caller doing an `.unwrap_or_else(|| unreachable!())`

Hence I think you should replace the assert by a `return Err("Need reshape")`

oxidation/crates/parsec_client_types/src/local_manifest.rs
```diff
@@ -111,6 +183,99 @@ impl From<LocalFileManifest> for LocalFileManifestData {
...
+            version: self.base.version + 1,
+            created: self.base.created,
+            updated: self.updated,
+            size: self.size,
+            blocksize: self.blocksize,
+            blocks,
+        })
+    }
+
+    pub fn match_remote(&self, remote_manifest: &FileManifest) -> Result<bool, &'static str> {
```
touilleMan, line 268: I think we can use `-> bool` as return value, this would make this method more pleasant to use and we don't care that the match failed because the file manifest is not reshaped.

oxidation/crates/parsec_client_types/src/local_manifest.rs
```diff
@@ -111,6 +183,99 @@ impl From<LocalFileManifest> for LocalFileManifestData {
...
+        timestamp: DateTime,
+    ) -> Result<FileManifest, &'static str> {
+        self.assert_integrity();
+        assert!(self.is_reshaped());
+
+        let blocks = self
+            .blocks
+            .iter()
+            .map(|chunks| chunks[0].get_block_access())
+            .collect::<Result<Vec<_>, _>>()?
```
touilleMan, line 249: it seems the collect cannot fail as long as `self.is_reshaped() == true`

That would mean that we could use a `map_err` to return `Err("Need reshape")`, and remove the `is_reshaped` call entirely \o/

oxidation/crates/parsec_client_types/src/local_manifest.rs
```diff
@@ -184,6 +349,64 @@ impl From<LocalFolderManifest> for LocalFolderManifestData {
 
 impl_local_manifest_dump_load!(LocalFolderManifest);
 
+impl LocalFolderManifest {
+    pub fn _filter_local_confinement_points(&self) -> Self {
```
touilleMan, line 353: protip: in Python methods starting with an underscore are private by convention, so no need to declare them `pub` in Rust ;-)

(an we can remove the underscore from the naming to look more Rust I guess ?)

oxidation/crates/parsec_client_types/src/local_manifest.rs
```diff
@@ -184,6 +349,64 @@ impl From<LocalFolderManifest> for LocalFolderManifestData {
 
 impl_local_manifest_dump_load!(LocalFolderManifest);
 
+impl LocalFolderManifest {
```
touilleMan, line 352: missing `new_placeholder/evolve_children_and_mark_updated`

oxidation/crates/parsec_client_types/src/local_manifest.rs
touilleMan: There is plenty of methods missing (e.g. `new_placeholder/apply_prevent_sync_pattern/evolve_children_and_mark_updated` etc.) for all the local manifests structs

they have ended up in the python binding code, but there should be implemented here !

touilleMan (changes requested):

tests/rust/test_local_manifest.py
```diff
@@ -0,0 +1,738 @@
...
+            isinstance(b2, BlockAccess)
+            and b1.id == b2.id
+            and b1.offset == b2.offset
+            and b1.size == b2.size
+            for (b1, b2) in zip(py.blocks, rs.blocks)
+        )
+
+    kwargs = {
+        "base": FileManifest(
+            **{
```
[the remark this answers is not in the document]
Max-7: Done.

tests/rust/test_manifest.py
```diff
@@ -301,18 +316,99 @@ def _assert_workspace_manifest_eq(py, rs):
     py_signed_and_encrypted = py_wm.dump_sign_and_encrypt(signing_key, secret_key)
     rs_signed_and_encrypted = rs_wm.dump_sign_and_encrypt(signing_key, secret_key)
 
-    py_decrypted_and_verified = _PyWorkspaceManifest.decrypt_verify_and_load(
+    WorkspaceManifest.decrypt_verify_and_load(
         py_signed_and_encrypted, secret_key, signing_key.verify_key, py_wm.author, py_wm.timestamp
     )
-    rs_decrypted_and_verified = WorkspaceManifest.decrypt_verify_and_load(
-        rs_signed_and_encrypted, secret_key, signing_key.verify_key, rs_wm.author, rs_wm.timestamp
+    _PyWorkspaceManifest.decrypt_verify_and_load(
```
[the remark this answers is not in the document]
Max-7: Changed the comparaison function to check for the types of both objects and realized that `_PyWorkspaceManifest.crypt_verify_and_load()` returns a Rust type, not a Python type. This is not really a problem as it will happen only when testing. I added the comparaison back (but with `==` instead of the function).

oxidation/crates/parsec_api_types/src/ext_types.rs
```diff
@@ -87,7 +87,7 @@ impl std::fmt::Display for DateTime {
         write!(
             f,
             "{}",
-            self.0.to_rfc3339_opts(chrono::SecondsFormat::Micros, true)
+            self.0.to_rfc3339_opts(chrono::SecondsFormat::Secs, false)
```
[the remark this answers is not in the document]
TimeEngineer: Because python doesn't display in error message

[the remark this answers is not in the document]
TimeEngineer: Because python doesn't display in error message

oxidation/crates/parsec_api_types/src/manifest.rs
```diff
@@ -128,8 +136,22 @@ impl std::fmt::Debug for EntryName {
     }
 }
 
+#[derive(Debug, Clone, Copy, Serialize, Deserialize)]
```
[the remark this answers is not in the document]
TimeEngineer: Done :+1:

oxidation/crates/parsec_client_types/src/local_manifest.rs
TimeEngineer: Done :+1:

oxidation/crates/parsec_client_types/src/local_manifest.rs
TimeEngineer: Done :+1:

oxidation/crates/parsec_client_types/src/local_manifest.rs
TimeEngineer: Done :+1:

touilleMan (approved):

touilleMan (changes requested):
Block.offset_size > 0 is now a thing: https://github.com/Scille/parsec-cloud/pull/2171

Max-7 (approved):

oxidation/crates/parsec_client_types/src/local_manifest.rs
```diff
@@ -82,14 +82,14 @@ impl Chunk {
...
         Ok(Self {
             id: ChunkID::from(*block_access.id),
             raw_offset: block_access.offset,
-            raw_size,
+            raw_size: block_access.size,
             start: block_access.offset,
-            stop,
+            stop: (block_access.offset + u64::from(block_access.size))
+                .try_into()
+                .unwrap(),
```
touilleMan, line 92: `from_block_access` return a `Result`, so better replace this unwrap by a proper error handling

touilleMan (approved):
lgtm except https://github.com/Scille/parsec-cloud/pull/2064/commits/9a56095318d9ff12b25ca942c02383537e7ef9a8#r848534744 to correct

oxidation/crates/parsec_client_types/src/local_manifest.rs
touilleMan: > It can't fail ! (>= 0) + (> 0) = (> 0)

if it can't fail, we should use the `unwrap_or(|| unreachable!())` syntax and, more importantly, explain why the unwrap is ok in this case

> And if you want to handle overflow, we should change all addition maybe ?

That's a valid point, in the current context the need of handling overflow is really obvious given `block_access.offset/size` can come third party (hence a malicious user can create a block_access.size/offset with `std::u64::MAX`, then all the Parsec client reading those data would crash !)

Can you do a check on the rest of the Rust codebase to see where overflow may occur ? (typically open another PR with `// TODO: what to do with overflow` comments where we need to think about)

touilleMan: image

PR #2070, opened 2022-03-10
fix .fuse_hidden file

jojoquivogn:


tests/core/mountpoint/test_folder_operations.py
```diff
@@ -213,7 +213,11 @@ def move_folder(self, src, dst_parent, dst_name):
         def iterdir(self, path):
             expected_exc = None
             try:
-                expected_children = {x.name for x in path.to_oracle().iterdir()}
+                expected_children = {
+                    x.name
+                    for x in path.to_oracle().iterdir()
+                    if not x.name.__contains__(".fuse_hidden")
```
touilleMan, line 219: `".fuse_hidden" in x.name` is the pythonic way to do this ^^

but In the current case I would use `x.name.startswith(".fuse_hidden")` instead

tests/core/mountpoint/test_folder_operations.py
```diff
@@ -213,7 +213,11 @@ def move_folder(self, src, dst_parent, dst_name):
         def iterdir(self, path):
             expected_exc = None
             try:
-                expected_children = {x.name for x in path.to_oracle().iterdir()}
+                expected_children = {
+                    x.name
+                    for x in path.to_oracle().iterdir()
```
touilleMan, line 218: are you sure the filter should be done on the oracle ?

I would say the `.fuse_hidden` files appear in the fuse mountpoint, hence in the thing we are testing (the oracle is a regular folder that is suppose to provide us the expected behavior the mountpoint should have)

PR #2075, opened 2022-03-14
[GUI] Fixed type error when listing files

Max-7:


vxgmichel (approved):
LGTM :+1: 

Maybe a test to catch those issues? :)

PR #2101, opened 2022-03-18
Fix diskfull tests

vxgmichel:


parsec/core/fs/workspacefs/workspacefile.py
```diff
@@ -123,14 +123,14 @@ async def close(self) -> None:
                     # Flush the file (typically causes the manifest to be reshaped)
                     await self._transactions.fd_flush(self.fileno())
                 except FSLocalStorageClosedError:
-                    return
+                    pass
```
touilleMan, line 126: The return was causing `self._state` to not ending up on the right state, and this was causing issue for the tests.

Is there more trouble you saw in this ?

vxgmichel: The issue was not about `self._state` not being in the right state (because the `self._state` affectation is in a `finally` clause), but it was about silently ignoring the current exception. Just like this:
```python
In [1]: def f():
   ...:     try:
   ...:         1/0
   ...:     finally:
   ...:         return
   ...: 

In [2]: f()
# f() do not raise
```

touilleMan: but havin `except: pass` also silence the exception, in your exemple:
```python
def f():
  try:
    1/0
  except Exception:
    pass

f()
# f() do not raise
```

so I don't get the difference between a `pass` and a `return` here given in both case the remaining code to execute in is finally blocks :/

(I would say `pass` is better here because it is more resistant to refactoring mistake though)

touilleMan (approved):

PR #2111, opened 2022-03-22
Add handshake protocol

TimeEngineer:


oxidation/crates/parsec_api_protocol/src/error.rs
```diff
@@ -0,0 +1,49 @@
...
+use thiserror::Error;
+
+use crate::ApiVersion;
+
+#[derive(Error, Debug, PartialEq)]
+pub enum HandshakeError {
+    #[error("{0}")]
+    FailedChallenge(String),
+    #[error("Handshake bad administration token {0:?}")]
+    BadAdministrationToken(Option<String>),
```
touilleMan, line 13: Regarding all those `Option<String>` fields, I think we should remove them altogether: they are based on the `help` field from handshake which is similar to the `reason` field in the command response. In both case those fields are too broad to be used in a programmatic way (no guarantee what the field value would be, and parsing super cumbersome if it contains variable such as user id) and too low-level to be displayed to the user.

oxidation/crates/parsec_api_protocol/src/handshake.rs
```diff
@@ -0,0 +1,659 @@
...
+    fn default() -> Self {
+        Self {
+            challenge_size: 48,
+            supported_api_version: vec![API_V2_VERSION, API_V1_VERSION],
+        }
+    }
+}
+
+impl ServerHandshakeStalled {
+    pub fn new(challenge_size: usize) -> Self {
```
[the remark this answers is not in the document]
touilleMan: I've created an issue about the change in the python codebase: #2118

oxidation/crates/parsec_api_protocol/src/handshake.rs
```diff
@@ -0,0 +1,659 @@
...
+        .dumps()?;
+
+        Ok(ServerHandshakeChallenge {
+            supported_api_version: self.supported_api_version,
+            challenge,
+            raw,
+        })
+    }
+    #[cfg(feature = "test")]
+    pub fn build_challenge_req_with_challenge(
```
touilleMan, line 172: would be better to have `build_challenge_req` and `build_challenge_req_with_challenge` rely on the same internal `_build_challenge_req_with_challenge` function instead of duplicating their internals (this is to ensure there won't be drifting between the two in later refactoring)

oxidation/crates/parsec_api_protocol/src/handshake.rs
```diff
@@ -0,0 +1,659 @@
...
+    pub fn build_challenge_req(self) -> Result<ServerHandshakeChallenge, HandshakeError> {
+        let mut challenge = vec![0; self.challenge_size];
+        thread_rng().fill(&mut challenge[..]);
+
+        let raw = Handshake::Challenge {
+            challenge: challenge.clone(),
+            supported_api_versions: self.supported_api_version.clone(),
+            ballpark_client_early_offset: Some(BALLPARK_CLIENT_EARLY_OFFSET),
+            ballpark_client_late_offset: Some(BALLPARK_CLIENT_LATE_OFFSET),
+            backend_timestamp: Some(DateTime::now()),
```
touilleMan, line 161: Using `Datetime::now()` is a side effect that make testing hard (hence `test_good_authenticated_handshake_client` marked `#[ignore]` I guess ^^)

We should instead have `fn build_challenge_req(self, timestamp: DateTime)`
- this way testing is trivial
- note we don't pass a `Option<Datetime>`, this is because timestamp field is not a mandatory field only because it has been introduced later (yeah, yet another backward compatibility trick !). So we want to always include this timestamp field, but we cannot assume it is always here.

oxidation/crates/parsec_api_protocol/src/error.rs
```diff
@@ -0,0 +1,49 @@
+// Parsec Cloud (https://parsec.cloud) Copyright (c) BSLv1.1 (eventually AGPLv3) 2016-2021 Scille SAS
+
+use parsec_api_types::DateTime;
+use thiserror::Error;
+
+use crate::ApiVersion;
+
+#[derive(Error, Debug, PartialEq)]
+pub enum HandshakeError {
+    #[error("{0}")]
```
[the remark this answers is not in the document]
TimeEngineer: I don't know what are the specification

oxidation/crates/parsec_api_protocol/src/handshake.rs
TimeEngineer: I agree :+1;

oxidation/crates/parsec_api_protocol/src/handshake.rs
TimeEngineer: Ok I will add this later

oxidation/crates/parsec_api_protocol/src/error.rs
TimeEngineer: Done :+1:

oxidation/crates/parsec_api_protocol/src/handshake.rs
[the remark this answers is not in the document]
TimeEngineer: Done :+1:

oxidation/crates/parsec_api_protocol/src/handshake.rs
TimeEngineer: Done :+1:

oxidation/crates/parsec_api_protocol/src/handshake.rs
```diff
@@ -0,0 +1,659 @@
...
+}
+
+#[derive(Debug)]
+pub struct InvitedClientHandshakeChallenge {
+    pub backend_api_version: ApiVersion,
+    pub client_api_version: ApiVersion,
+    pub raw: Vec<u8>,
+}
+
+pub trait BaseClientHandshake: Sized {
```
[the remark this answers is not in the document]
TimeEngineer: Done :+1:

oxidation/crates/parsec_api_protocol/src/handshake.rs
TimeEngineer: Done :+1:
I changed the field directly instead

oxidation/generate_test_data/protocol/handshake.py
```diff
@@ -0,0 +1,39 @@
+# Parsec Cloud (https://parsec.cloud) Copyright (c) BSLv1.1 (eventually AGPLv3) 2016-2021 Scille SAS
+# flake8: noqa
+
+from utils import *
+
+display(
```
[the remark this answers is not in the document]
touilleMan: Yes, but this file should also be a proof that the data was generated from the Python codebase ;-)

oxidation/crates/parsec_api_protocol/src/error.rs
```diff
@@ -0,0 +1,49 @@
...
+
+use parsec_api_types::DateTime;
+use thiserror::Error;
+
+use crate::ApiVersion;
+
+#[derive(Error, Debug, PartialEq)]
+pub enum HandshakeError {
+    #[error("Failed chalenge {0:?}")]
+    FailedChallenge(String),
```
touilleMan, line 11: I guess we can also remove the `String` field here too, and just have a `"Handshake invalid answer (bad signature or challenge value)"` constant error message.

oxidation/crates/parsec_api_protocol/src/error.rs
```diff
@@ -0,0 +1,49 @@
...
+    #[error("Handshake revoked device")]
+    RevokedDevice,
+    #[error("Handshake out of ballpark {0:x?}")]
+    OutOfBallpark(ChallengeDataReport),
+    #[error("No overlap between client API versions {client_versions:?} and backend API versions {backend_versions:?}")]
+    APIVersion {
+        client_versions: Vec<ApiVersion>,
+        backend_versions: Vec<ApiVersion>,
+    },
+    #[error("{0}")]
```
touilleMan, line 29: Can replace `String` by `&'static str`, and should have a `"Handshake invalid message: {0}"` error message

oxidation/crates/parsec_api_protocol/src/error.rs
```diff
@@ -0,0 +1,49 @@
...
+    InvalidMessage(String),
+    #[error("`verify_key` param must be provided for authenticated handshake")]
+    Authenticated,
```
touilleMan, line 32: `Authenticated` is too broad name, `MissingVerifyKey` seems better considering the error message.

And why not instead considering this is a generic error and use ```InvalidMessage("`verify_key` param must be provided for authenticated handshake")``` ?

oxidation/crates/parsec_api_protocol/src/handshake.rs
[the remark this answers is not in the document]
touilleMan: no need for a `challenge_size` attribute in `ServerHandshakeStalled` struct, you can just have a `const HANDSHAKE_CHALLENGE_SIZE: usize = 48;` at the top of `handshake.rs`

oxidation/crates/parsec_api_protocol/src/error.rs
TimeEngineer: I can't, because there is somewhere I use format! to build dynamic error

oxidation/crates/parsec_api_protocol/src/error.rs
TimeEngineer: Done :+1:

oxidation/crates/parsec_api_protocol/src/handshake.rs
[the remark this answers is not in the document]
TimeEngineer: Done :+1:

oxidation/crates/parsec_api_protocol/src/handshake.rs
touilleMan: > Done 

I don't get it, why is there still a `DateTime::now()` call here ? 🤔

oxidation/crates/parsec_api_protocol/src/error.rs
```diff
@@ -0,0 +1,47 @@
...
+    #[error("No overlap between client API versions {client_versions:?} and backend API versions {backend_versions:?}")]
+    APIVersion {
+        client_versions: Vec<ApiVersion>,
+        backend_versions: Vec<ApiVersion>,
+    },
+    #[error("Handshake invalid message: {0}")]
+    InvalidMessage(String),
+}
+
+impl From<&'static str> for HandshakeError {
```
touilleMan, line 33: I will prefere to use `HandshakeError::InvalidMessage(my_error.into())` directly in the caller instead of `my_error.into()`: the 1rst form is much more explicit while the 2nd encourage misleading behavior (typically when doing copy/paste when developing a new feature, one can be lead into thinking `my_error.into()` is the default way to create a `HandshakeError`, where it is only a special corner case)

oxidation/crates/parsec_api_types/src/ext_types.rs
```diff
@@ -140,6 +141,13 @@ impl<'de> serde::Deserialize<'de> for DateTime {
     }
 }
 
+impl Sub for DateTime {
+    type Output = Duration;
```
touilleMan, line 145: We should re-expose `chrono::Duration` then (the idea would be to never rely on chrono outside of here)

oxidation/crates/parsec_api_protocol/src/handshake.rs
```diff
@@ -0,0 +1,653 @@
...
+        answer: Vec<u8>,
+    },
+    #[serde(rename = "INVITED")]
+    Invited {
+        client_api_version: ApiVersion,
+        organization_id: OrganizationID,
+        invitation_type: InvitationType,
+        token: InvitationToken,
+    },
+    #[serde(rename = "signed_answer")]
```
touilleMan, line 81: `AnswerSchema` is different from `HandshakeAuthenticatedAnswerSchema`/`HandshakeInvitedAnswerSchema` !

`AnswerSchema` is use to build the `HandshakeAuthenticatedAnswerSchema.answer` field (by serializing then signing)

touilleMan (changes requested):

oxidation/crates/parsec_api_protocol/src/handshake.rs
TimeEngineer: Because I changed the field in the test, then the problem is solved

touilleMan: Yes, but having side effect here is not good, so we should change this api nevertheless ;-)

(an modifying `AuthenticatedClientHandshakeStalled` in the test is not ideal anyway)

oxidation/crates/parsec_api_types/src/ext_types.rs
TimeEngineer: Done :+1:

oxidation/crates/parsec_api_protocol/src/handshake.rs
TimeEngineer: Done :+1:

oxidation/crates/parsec_api_protocol/src/handshake.rs
```diff
@@ -0,0 +1,649 @@
...
+            challenge_size: 48,
+            supported_api_version: vec![API_V2_VERSION, API_V1_VERSION],
+        }
+    }
+}
+
+impl ServerHandshakeStalled {
+    fn _build_challenge_req_with_challenge(
+        self,
+        challenge: Vec<u8>,
```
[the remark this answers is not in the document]
TimeEngineer: Done :+1:

oxidation/crates/parsec_api_protocol/src/handshake.rs
TimeEngineer: I changed, but it doesn't change the tests, because it's on the python side that we don't choose the timestamp

oxidation/crates/parsec_api_protocol/src/error.rs
touilleMan: > it won't exist anymore when we remove them all

Then we should add a ```// TODO: temporary code to handle `&'static str` kind of errors```

Do you have an idea how complex it would be to clean the code from the need of this ?
I would naively guess the handshake is only used by the transport, which at this point is another unmerged PR, so better modify the transport PR than adding this legacy code handler.

> From trait allow the operator ? to auto convert from an error to an other error

That's precisely why this code is dangerous ;-)

touilleMan (changes requested):

oxidation/crates/parsec_api_protocol/src/error.rs
TimeEngineer: This is not dangerous, If you can't autoconvert that means you didn't handle correctly your error

TimeEngineer: Naively you will want to remove this implementation, and it will find where you are doing autoconversions,
whereas if you don't do this auto conversion, you will have to change all the map_error

touilleMan: The change was pretty trivial (see e975b6f03d607c1b46b5ec881141333e5e01a62b): replace `?` by `.map_err(|err| HandshakeError::InvalidMessage(err.into()))?` in a single file.

Adding temporary handler for this is not worth it ;-)

oxidation/crates/parsec_api_protocol/tests/test_handshake.rs
```diff
@@ -0,0 +1,874 @@
...
+
+    let mut ch = AuthenticatedClientHandshakeStalled::new(
+        alice.organization_id().clone(),
+        alice.device_id.clone(),
+        alice.signing_key.clone(),
+        alice.root_verify_key().clone(),
+    );
+
+    // Fix ballpark failure
+    ch.client_timestamp = DateTime::from_f64_with_us_precision(1647884434.47775);
```
touilleMan, line 105: This is why we should not have side effect on datetime (cf. https://github.com/Scille/parsec-cloud/pull/2111#discussion_r843721057)

`AuthenticatedClientHandshakeStalled::new` should have a `client_timestamp: Datetime` paremeter

oxidation/crates/parsec_api_protocol/src/handshake.rs
```diff
@@ -0,0 +1,668 @@
...
+        // They will be missing for older backend so they cannot be strictly required.
+        // TODO: This backward compatibility should be removed once Parsec < 2.4 support is dropped
+        #[serde(default, deserialize_with = "maybe_field::deserialize_some")]
+        ballpark_client_early_offset: Option<f64>,
+        #[serde(default, deserialize_with = "maybe_field::deserialize_some")]
+        ballpark_client_late_offset: Option<f64>,
+        #[serde(default, deserialize_with = "maybe_field::deserialize_some")]
+        backend_timestamp: Option<DateTime>,
+    },
+    Answer(Box<Answer>),
```
touilleMan, line 127: Why box is needed here ?

touilleMan (changes requested):

oxidation/crates/parsec_api_protocol/src/handshake.rs
TimeEngineer: This was due to the answer size, but it's resolved by boxing rvk field

TimeEngineer: Done

oxidation/crates/parsec_api_protocol/tests/test_handshake.rs
TimeEngineer: Done

oxidation/crates/parsec_api_protocol/src/handshake.rs
touilleMan: > This was due to the answer size

That's an interesting point, do you have a rule of thumb for determining what should be boxed ?

(FYI rvk is a ed25519 public key, so it size is at least 32 bytes, but RustCrypto cache aditional points on the Edward curve so total size is around ~200bytes)

oxidation/crates/parsec_api_protocol/src/error.rs
touilleMan: The format was needed because of the `help` field that is deprecated (deprecated precisely because non-structured error value is hard to handle and easy to break ^^)

Without this `help` handling, `&'static str` is possible, and it simplify a lot error conversion, so much win \o/
(see 59ea8fa904b99158d4f4653a754a059dee085f17)

oxidation/crates/parsec_api_protocol/src/handshake.rs
TimeEngineer: It's clippy the goat which gives this advice, the point is that an enum use the same size for all his variants, then it can be a waste of memory if a variant is larger than another

touilleMan: good point !

touilleMan (approved):
lgtm

PR #2113, opened 2022-03-22
Pki enrollment

AntoineDupre:


parsec/api/data/pki.py
```diff
@@ -0,0 +1,62 @@
...
+    requested_device_label: DeviceLabel
+
+
+@attr.s(slots=True, frozen=True, auto_attribs=True, kw_only=True, eq=False)
+class PkiEnrollmentRequest(BaseAPIData):
+    class SCHEMA_CLS(BaseSchema):
+        der_x509_certificate = fields.Bytes(require=True)
+        signature = fields.Bytes(required=True)
+        pki_request_info = fields.Bytes(required=True)  # Signature should be checked before loading
+        requested_human_handle = HumanHandleField(required=True)
```
touilleMan, line 37: This should go in `PkiEnrollmentRequestReqSchema`

parsec/backend/http.py
```diff
@@ -345,3 +369,41 @@ async def _api_check_auth_and_load_body(self, req: HTTPRequest, req_serializer):
...
+        if req.method == "GET":
+            return HTTPResponse.build_msgpack(200, data=b"")
+
+        body = await req.get_body()
+        msg = unpackb(body)
+        msg["cmd"] = "pki_enrollement_request"
+        rep = await self._pki_component.api_pki_enrollment_request(
+            msg, OrganizationID(organization_name)
+        )
+        return HTTPResponse.build_msgpack(200, data=packb(rep))
```
touilleMan, line 391: `packb` should go into `build_msgpack`

parsec/backend/memory/pki.py
```diff
@@ -0,0 +1,153 @@
...
+        pki_certificate.reply_object = pki_reply_bundle.reply_object
+        pki_certificate.reply_user_id = pki_reply_bundle.reply_user_id
+        pki_certificate.reply_timestamp = pendulum.now()
+        pki_certificate.reply_admin_user = pki_reply_bundle.reply_admin_user
+        return pki_certificate.reply_timestamp
+
+    async def pki_enrollment_get_reply(
+        self, certificate_id, request_id
+    ) -> Tuple[
+        Optional[PkiEnrollmentReply],
```
touilleMan, line 133: Too much optional here, the combinatorics is very high then !

I have the feeling this is in fact much simpler: either a reply is present and all field are here, either not yet and we only have the request timestamp !

but it is weird to return the request timestamp given the name of the method :(

So I would rename it `get_enrollment_state()` and it would return a `PkiEnrollementStateRequest` or `PkiEnrollementStateReply` (with both being a simple dataclass)

touilleMan (changes requested):

parsec/api/protocol/pki.py
```diff
@@ -0,0 +1,92 @@
...
+from parsec.api.data.pki import PkiEnrollmentReply
+from parsec.api.protocol.base import BaseRepSchema, BaseReqSchema, CmdSerializer
+from parsec.api.protocol.types import HumanHandleField
+from parsec.serde import fields
+
+
+# pki_enrolement_request
+
+
+class PkiEnrollmentRequestReqSchema(BaseReqSchema):
```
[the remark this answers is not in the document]
vxgmichel: Plus, the command is called `pki_enrollement_request` so the protocol should reflect that name.

parsec/api/data/pki.py
```diff
@@ -0,0 +1,62 @@
...
+    der_x509_certificate: bytes
+    signature: bytes
+    pki_request_info: bytes
+    requested_human_handle: HumanHandle
+
+
+@attr.s(slots=True, frozen=True, auto_attribs=True, kw_only=True, eq=False)
+class PkiEnrollmentReply(BaseAPIData):
+    class SCHEMA_CLS(BaseSchema):
+        der_x509_admin_certificate = fields.Bytes(required=True)
```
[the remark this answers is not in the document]
vxgmichel: `der_x509_certificate` sounds fine.

parsec/api/data/pki.py
```diff
@@ -0,0 +1,62 @@
...
+
+from parsec.api.data.base import BaseAPIData
+from parsec.api.protocol import HumanHandle, HumanHandleField
+from parsec.api.protocol.types import DeviceLabel, DeviceLabelField
+from parsec.crypto import PublicKey, VerifyKey
+from parsec.serde import BaseSchema, fields, post_load
+
+
+@attr.s(slots=True, frozen=True, auto_attribs=True, kw_only=True, eq=False)
+class PkiEnrollmentRequestInfo(BaseAPIData):
```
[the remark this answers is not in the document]
vxgmichel: Yes. It's an invitation in the sense that the requester has been provided with an URL that invites it to perform the enrollment procedure.

At the moment, I don't think the concepts between `requester` and `replier` are ever named in the code base, but it sounds fine to me (and it's separated from the greeter/claimer semantics)

parsec/api/data/pki.py
```diff
@@ -0,0 +1,62 @@
...
+    pki_request_info: bytes
+    requested_human_handle: HumanHandle
+
+
+@attr.s(slots=True, frozen=True, auto_attribs=True, kw_only=True, eq=False)
+class PkiEnrollmentReply(BaseAPIData):
+    class SCHEMA_CLS(BaseSchema):
+        der_x509_admin_certificate = fields.Bytes(required=True)
+        signature = fields.Bytes(required=True)
+        pki_reply_info = fields.Bytes(required=True)  # Signature should be checked before loading
```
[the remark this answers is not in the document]
vxgmichel: Yes, I'll take care of this part.

parsec/api/protocol/pki.py
```diff
@@ -0,0 +1,92 @@
...
+class PkiEnrollmentGetReplyReqSchema(BaseReqSchema):
+    certificate_id = fields.Bytes(required=True)
+    request_id = fields.UUID(required=True)
+
+
+class PkiEnrollmentGetReplyRepSchema(BaseRepSchema):
+    status = fields.String(required=True)
+    reply = fields.Nested(PkiEnrollmentReply.SCHEMA_CLS, required=True, allow_none=True)
+    timestamp = fields.DateTime(required=True)
+    admin_human_handle = HumanHandleField(required=True, allow_none=True)
```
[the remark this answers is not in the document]
vxgmichel: I think the original idea was that it could be useful when documenting something that goes wrong. But I guess the reply could be sent if an error occur in order to let the client investigate what happened (for instance, who else processed the requests while an admin try to accept/deny it).

parsec/api/protocol/pki.py
```diff
@@ -0,0 +1,92 @@
...
+from parsec.api.protocol.types import HumanHandleField
+from parsec.serde import fields
+
+
+# pki_enrolement_request
+
+
+class PkiEnrollmentRequestReqSchema(BaseReqSchema):
+    request = fields.Nested(PkiEnrollmentRequest.SCHEMA_CLS, required=True)
+    certificate_id = fields.Bytes(required=True)
```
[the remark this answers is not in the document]
vxgmichel: Then what about `get_reply`? Should the requester provide the `certificate_id` by performing the hash on its side even though this was never part of the `request`, or should the they provide the `der_certificate` and let the backend handle certificate identification on its own?

parsec/api/data/pki.py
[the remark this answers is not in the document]
vxgmichel: Also, I would add the we should never have `requester/replier/Request/Reply` on their own, they should always be prefixed with `Enrollment`. I would say the concepts are `enrollment`, `enrollment request(er)`, `enrollment reply(ier)`, and all this is part of the parsec module that deals with PKI.

Just nickpicking though

parsec/api/protocol/invite.py
```diff
@@ -48,6 +48,7 @@ class InvitationToken(UUID4):
 class InvitationType(Enum):
     USER = "USER"
     DEVICE = "DEVICE"
+    PKI = "PKI"
```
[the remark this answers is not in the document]
vxgmichel: Sounds good to me! As I said elsewhere it's still a bit like an invitation to me, in the sense that the user is invited to perform the pki enrollment procedure, but the `BackendPkiEnrollmentAddr` naming sounds quite explicit about that.

The original reason for this choice was to avoid code duplication (as those `BackendSomeActionAddr` are not trivial to implement), but keeping things separated is probably for the better.

parsec/core/cli/pki.py
```diff
@@ -105,6 +106,10 @@ async def _pki_enrollment_poll(
     save_device_with_selected_auth: Callable,
     extra_trust_roots: Sequence[Path],
 ):
+
+    # Use extra trust roots from both command line and config
+    extra_trust_roots = [*extra_trust_roots, *config.pki_extra_trust_roots]
```
touilleMan, line 112: Why not add this directly in the `@core_config_options` decorator ?

BTW I think it would be better to have the option name `--extra-trust-root` mention it is related to PKI, so for instance `--pki-extra-trust-root` ? (and adding a `help="Additional directory to look for PKI root certificate"`)

parsec/core/cli/pki.py
```diff
@@ -187,13 +210,16 @@ def _display_pending_enrollment(pending: PkiEnrollmentSubmitterSubmittedCtx):
...
             assert isinstance(ctx, PkiEnrollmentSubmitterAcceptedStatusCtx)
-            click.echo(
-                "Enrollment has been accepted on " + click.style(ctx.accepted_on, fg="yellow")
-            )
-            ctx = await ctx.finalize()
+            click.echo(_display_accepted_enrollment(ctx))
             if not dry_run:
-                await save_device_with_selected_auth(
-                    config_dir=config.config_dir, device=ctx.new_device
+                if not await aconfirm("Finalize device creation"):
```
touilleMan: What's the use of this confirmation ? I don't see a good usecase for it, especially given it seems redundant with the `--dry-run` option

parsec/core/types/pki.py
```diff
@@ -0,0 +1,152 @@
...
+
+    @classmethod
+    def list(cls, config_dir: Path) -> List["LocalPendingEnrollment"]:
+        result = []
+        for path in cls.iter_path(config_dir):
+            try:
+                result.append(cls.load(path.read_bytes()))
+            # Ignore invalid or missing files
+            except (DataError, OSError) as e:
+                print(repr(e))
```
touilleMan: print to remove ?

parsec/core/cli/pki.py
vxgmichel: Because the user has to check the information contained in the accepter certificate before finalizing the device creation (similar to what happens on the accepter side)

parsec/core/types/pki.py
vxgmichel: Ooopsie

parsec/core/pki/submitter.py
```diff
@@ -63,7 +70,29 @@ async def new(
...
     ) -> "PkiEnrollmentSubmitterSubmittedCtx":
-        # TODO: document exceptions !
+        """
+        Raises:
+            BackendNotAvailable
+            BackendProtocolError
+
+            BackendCmdsInvalidRequest
+            BackendCmdsInvalidResponse
+            BackendNotAvailable
```
touilleMan, line 80: Duplicated `BackendNotAvailable`

Funny enough, I don't see any reference to `BackendCmdsInvalidRequest`, `BackendCmdsInvalidResponse` and `BackendCmdsBadResponse` in the codebase (they are only referenced in `_send_cmd`'s documentation...). I think we should document `BackendConnectionError` instead, and add `BackendNotAvailable`/`BackendProtocolError`/`BackendOutOfBallparkError` sub errors (given they are useful to catch separatly)

touilleMan: And it would be good to also correct `_send_cmd` ;-)

parsec/core/pki/exceptions.py
```diff
@@ -23,7 +23,13 @@
 
 
 class PkiEnrollmentError(Exception):
-    pass
+    @property
+    def message(self):
+        """All PkiEnrollmentError should have a message as first argument."""
```
touilleMan, line 28: if it's a message, don't we want to do assert `isisnance(msg, str)` in the __init__ ?

parsec/backend/postgresql/__init__.py
```diff
@@ -30,6 +31,8 @@
     "PGRealmComponent",
     "PGVlobComponent",
     "PGBlockComponent",
+    "PGCertificateComponent",
```
touilleMan, line 34: doesn't exist anymore ^^

parsec/backend/postgresql/migrations/0010_add_pki_certificate_table.sql
```diff
@@ -0,0 +1,51 @@
+ -- Parsec Cloud (https://parsec.cloud) Copyright (c) BSLv1.1 (eventually AGPLv3) 2016-2021 Scille SAS
```
touilleMan, line 1: This stuff should also be added to `datamodel.sql` (in the current state, the codebase should not pass the `tests/backend/migrations/test_migrations.py::test_migrations` that checks that once the migrations have been applied, the resulting tables schemas in the database correspond to the one obtained by applying `datamodel.sql` on a blank database)

parsec/backend/postgresql/migrations/0010_add_pki_certificate_table.sql
```diff
@@ -0,0 +1,51 @@
...
+    accepter INTEGER REFERENCES device (_id),
+    accepted INTEGER REFERENCES device (_id),
+
+    enrollment_state enrollment_state NOT NULL DEFAULT 'SUBMITTED',
+    info_accepted pki_enrollment_info_accepted,
+    info_rejected pki_enrollment_info_rejected,
+    info_cancelled pki_enrollment_info_cancelled,
+
+    UNIQUE(organization, enrollment_id)
+);
```
touilleMan, line 51: missing trailling newline, should have been catched by pre-commit hook 🤔

parsec/backend/postgresql/migrations/0010_add_pki_certificate_table.sql
```diff
@@ -0,0 +1,51 @@
...
+
+    enrollment_id UUID NOT NULL,
+    submitter_der_x509_certificate BYTEA NOT NULL,
+    submitter_der_x509_certificate_sha1 BYTEA NOT NULL,
+
+    submit_payload_signature BYTEA NOT NULL,
+    submit_payload BYTEA NOT NULL,
+    submitted_on TIMESTAMPTZ NOT NULL,
+
+    accepter INTEGER REFERENCES device (_id),
```
touilleMan, line 42: why accepter and accepted are not part of `pki_enrollment_info_accepted` type ?

touilleMan: on top of that, I would rename `accepted` into `submitter_accepted_device` (or maybe `new_device` if it is defined inside `pki_enrollment_info_accepted`)

parsec/backend/postgresql/pki.py
```diff
@@ -0,0 +1,439 @@
...
+    PkiEnrollmentNoLongerAvailableError,
+    PkiEnrollmentNotFoundError,
+)
+from parsec.backend.postgresql import PGHandler
+from parsec.backend.postgresql.utils import Q, q_organization_internal_id, q_device_internal_id
+from parsec.backend.postgresql.user_queries.create import q_create_user
+
+_q_get_pki_enrollment_from_certificate_sha1 = Q(
+    f"""
+    SELECT * FROM pki_enrollment
```
touilleMan, line 37: Should name the values instead of using `SELECT *` to avoid mixing fields with compatible value types

parsec/backend/postgresql/pki.py
```diff
@@ -0,0 +1,439 @@
...
+    else:
+        assert False
+
+
+class PGPkiEnrollmentComponent(BasePkiEnrollmentComponent):
+    def __init__(self, dbh: PGHandler):
+        self.dbh = dbh
+
+    def register_components(self, **other_components):
+        self._user_component = other_components["user"]
```
touilleMan, line 204: I don't see where it is used...

parsec/backend/postgresql/pki.py
```diff
@@ -0,0 +1,439 @@
...
+            await send_signal(
+                conn, BackendEvent.PKI_ENROLLMENT_UPDATED, organization_id=organization_id
+            )
+
+    async def info(self, organization_id: OrganizationID, enrollment_id: UUID) -> PkiEnrollmentInfo:
+        """
+        Raises:
+            PkiEnrollmentNotFoundError
+        """
+        async with self.dbh.pool.acquire() as conn, conn.transaction():
```
touilleMan, line 317: No need for a transaction block here given we do a single query

parsec/backend/postgresql/pki.py
```diff
@@ -0,0 +1,439 @@
...
+                raise PkiEnrollmentNotFoundError()
+            else:
+                return _build_enrollment_info(row)
+
+    async def list(self, organization_id: OrganizationID) -> List[PkiEnrollmentListItem]:
+        """
+        Raises: Nothing !
+        """
+
+        async with self.dbh.pool.acquire() as conn, conn.transaction():
```
touilleMan, line 333: same, no need for transaction block

parsec/backend/postgresql/pki.py
```diff
@@ -0,0 +1,439 @@
...
+    ) -> None:
+        """
+        Raises:
+            PkiEnrollmentNotFoundError
+            PkiEnrollmentNoLongerAvailableError
+        """
+
+        async with self.dbh.pool.acquire() as conn, conn.transaction():
+            row = await conn.fetchrow(
+                *_q_get_pki_enrollment_for_update(
```
touilleMan, line 362: 👌

parsec/backend/postgresql/pki.py
```diff
@@ -0,0 +1,439 @@
...
+                    conn=conn, organization_id=organization_id, user=user, first_device=first_device
+                )
+
+            except UserAlreadyExistsError as exc:
+                raise PkiEnrollmentAlreadyExistError from exc
+
+            except UserActiveUsersLimitReached as exc:
+                raise PkiEnrollmentActiveUsersLimitReached from exc
+
+            assert user.user_certifier is not None
```
touilleMan, line 423: Add this comment `#  Only the user created during organization bootstrap has no certifier`

parsec/backend/postgresql/pki.py
```diff
@@ -0,0 +1,439 @@
...
+                        revoked_on=row["revoked_on"],
+                        revoked_user_certificate=row["revoked_user_certificate"],
+                        revoked_user_certifier=None,
+                    )
+                    # if row and row.get("revoked_on") and row.get("revoked_on") < submitted_on:
+                    if not user.is_revoked():
+                        raise PkiEnrollmentAlreadyEnrolledError(submitted_on)
+                else:
+                    assert False
+            await conn.execute(
```
touilleMan, line 296: should check the return value and exception to handle the (unlikely !) case of two concurrent submits with the same `enrollement_id`, hence one succeed and the other raising an `UniqueViolationError` (see what is done in `postgresql/user_queries/create.py` when creating user/device)

parsec/backend/postgresql/pki.py
```diff
@@ -0,0 +1,439 @@
...
+                elif enrollment_state == PkiEnrollmentStatus.ACCEPTED.value:
+                    # Previous attempt end successfully, we are not allowed to submit
+                    # unless the created user has been revoked
+                    assert row["accepted"] is not None and row["accepter"] is not None
+
+                    row = await conn.fetchrow(
+                        *_q_get_user_from_device_id(
+                            organization_id=organization_id.str,
+                            device_id=row["accepted"],
+                            # now=submitted_on,
```
touilleMan, line 276: dead code ?

parsec/backend/postgresql/pki.py
```diff
@@ -0,0 +1,439 @@
...
+from parsec.backend.postgresql.user_queries.create import q_create_user
+
+_q_get_pki_enrollment_from_certificate_sha1 = Q(
+    f"""
+    SELECT * FROM pki_enrollment
+    WHERE (
+        organization = { q_organization_internal_id("$organization_id") }
+        AND submitter_der_x509_certificate_sha1=$submitter_der_x509_certificate_sha1
+    )
+    ORDER BY enrollment_state ASC
```
touilleMan, line 42: Can we order by `_id` instead ? I feel ordering by state is not really clear, and I may break if we add more states...

This is especially true given the comment where this query is used say we are retrieving the last attempt ;-)

touilleMan: And we may want to rename this query `_q_get_last_pki_enrollment_from_certificate_sha1` and add a `LIMIT 1` to the SQL

parsec/backend/postgresql/pki.py
```diff
@@ -0,0 +1,439 @@
...
+                raise PkiEnrollmentIdAlreadyUsedError()
+
+            # Try to retrieve the last attempt with this x509 certificate
+            rep = await conn.fetch(
+                *_q_get_pki_enrollment_from_certificate_sha1(
+                    organization_id=organization_id.str,
+                    submitter_der_x509_certificate_sha1=submitter_der_x509_certificate_sha1,
+                )
+            )
+            for row in rep:
```
touilleMan, line 239: There is no break in this loop, hence we are checking against all the previous enrollments which seems wrong (we should only access the last enrollment, hence my proposal to rename the query `_q_get_last_pki_enrollment_from_certificate_sha1` and ensure it returns at most 1 result)

touilleMan: btw there should a test with enough ACCEPTED/CANCELLED/REJECTED enrollments (and ideally witch a check those old enrollments hasn't been altered) that should break in against this current implementation ;-)

parsec/backend/postgresql/pki.py
```diff
@@ -0,0 +1,439 @@
...
+        """
+        Raises:
+            PkiEnrollmentCertificateAlreadySubmittedError
+            PkiEnrollmentAlreadyEnrolledError
+        """
+        submitter_der_x509_certificate_sha1 = hashlib.sha1(submitter_der_x509_certificate).digest()
+        async with self.dbh.pool.acquire() as conn, conn.transaction():
+            # Assert enrollment_id not used
+            row = await conn.fetchrow(
+                *_q_get_pki_enrollment_for_update(
```
touilleMan, line 225: FOR UPDATED is not mandatory here (given we raise just after we get a result, and if we don't get a result there is nothing to lock so we are not protected against concurrent insertion, see my following comment about `UniqueViolationError`)

However there is nothing wrong with adding the FOR UPDATE, so I'm okay with using this here ;-)
(but maybe add a small comment)

parsec/backend/postgresql/pki.py
touilleMan: and the `_q_get_pki_enrollment_from_certificate_sha1` should do a `FOR UPDATE`

parsec/backend/postgresql/pki.py
```diff
@@ -0,0 +1,439 @@
...
+                ]:
+                    # Previous attempt was unsuccessful, so we are clear to submit a new attempt !
+                    pass
+                elif enrollment_state == PkiEnrollmentStatus.ACCEPTED.value:
+                    # Previous attempt end successfully, we are not allowed to submit
+                    # unless the created user has been revoked
+                    assert row["accepted"] is not None and row["accepter"] is not None
+
+                    row = await conn.fetchrow(
+                        *_q_get_user_from_device_id(
```
touilleMan, line 273: we should also have a FOR UPDATE here... I'm wondering if we shouldn't use a advisory lock instead to simplify this locking logic (so we would have one big lock for user/device creation per organization)

touilleMan (changes requested):

parsec/api/protocol/events.py
```diff
@@ -21,6 +22,7 @@ class APIEvent(Enum):
     REALM_MAINTENANCE_STARTED = "realm.maintenance_started"
     REALM_VLOBS_UPDATED = "realm.vlobs_updated"
     REALM_ROLES_UPDATED = "realm.roles_updated"
+    PKI_ENROLLMENT_UPDATED = "pki_enrollment.updated"
```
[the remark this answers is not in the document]
AntoineDupre: fixed

parsec/api/protocol/base.py
```diff
@@ -108,8 +108,14 @@ def __repr__(self) -> str:
             f"rep_schema={self._rep_serializer})"
         )
 
-    def __init__(self, req_schema_cls: Type[BaseSchema], rep_schema_cls: Type[BaseSchema]):
+    def __init__(
+        self,
+        req_schema_cls: Type[BaseSchema],
+        rep_schema_cls: Type[BaseSchema],
+        extra_rep_schema: Dict[str, Type[BaseSchema]] = {},
```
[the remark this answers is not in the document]
AntoineDupre: We have reached a consensus \o/

parsec/backend/memory/user.py
```diff
@@ -38,6 +38,7 @@ def __init__(self, send_event, *args, **kwargs):
 
     def register_components(self, **other_components) -> None:
         self._organization_component = other_components["organization"]
+        self._pki_component = other_components["pki"]
```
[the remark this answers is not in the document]
AntoineDupre: no

parsec/backend/postgresql/migrations/0010_add_pki_certificate_table.sql
AntoineDupre: It's better to move it into  `pki_enrollment_info_accepted`, i tried to mimic `memory` implementation where those field are outside of the Accepted structure

vxgmichel (approved):

parsec/backend/postgresql/migrations/0010_add_pki_certificate_table.sql
AntoineDupre: It appears that `REFERENCES` can not be used inside a `TYPE`

touilleMan (approved):

parsec/backend/postgresql/pki.py
AntoineDupre: There is now a new test that try the following sequence:

PR #2119, opened 2022-03-23
Add transport

TimeEngineer:


.env
```diff
@@ -0,0 +1 @@
+VERSION="v2.6.0+dev"
```
touilleMan, line 1: I guess this file wasn't supposed to be commited

oxidation/crates/parsec_api_protocol/src/transport.rs
```diff
@@ -0,0 +1,69 @@
...
+
+pub struct Transport {
+    pub write: SplitSink<WebSocketStream<TcpStream>, Message>,
+    pub read: SplitStream<WebSocketStream<TcpStream>>,
+}
+
+// Only because we need user-agent
+fn build_request_from_uri<T: AsRef<str>>(uri: T) -> Request<()> {
+    dotenv::dotenv().ok();
+    let version = std::env::var("VERSION").unwrap();
```
touilleMan, line 23: To get the version, in theory we should be using `CARGO_PKG_VERSION` env var that is set automatically by cargo during compilation.
However this won't work given we want the version of `libparsec` and not of `parsec_api_protocol` crate.
I think the easiest way to fix this is to have a dead simple `parsec_version` crate (to avoid cyclic dependencies) that will contain a const version field.

oxidation/crates/parsec_api_protocol/src/transport.rs
```diff
@@ -0,0 +1,69 @@
...
+    WebSocketStream,
+};
+use futures::{
+    stream::{SplitSink, SplitStream},
+    SinkExt, StreamExt,
+};
+use http::{Request, Uri};
+
+pub struct Transport {
+    pub write: SplitSink<WebSocketStream<TcpStream>, Message>,
```
touilleMan, line 16: I think we want write/read fields to be kept private

oxidation/crates/parsec_api_protocol/tests/test_transport.rs
```diff
@@ -0,0 +1,33 @@
+// Parsec Cloud (https://parsec.cloud) Copyright (c) BSLv1.1 (eventually AGPLv3) 2016-2021 Scille SAS
+
+use async_std::net::{TcpListener, TcpStream};
+use async_std::task;
+use parsec_api_protocol::Transport;
+use rstest::rstest;
+
+#[rstest]
+fn test_transport() {
+    smol::block_on(async {
```
touilleMan, line 10: Now use Parsec handshake instead of `b"foobar"` payloads ;-)

oxidation/crates/parsec_api_protocol/src/transport.rs
TimeEngineer: Yes of course, I forgot this part after I switched branch and clippy disagreed

TimeEngineer: Done :+1:

oxidation/crates/parsec_api_protocol/tests/test_transport.rs
TimeEngineer: I can't while handshake is not merged :grimacing:

oxidation/crates/platform_native/tests/test_transport.rs
```diff
@@ -0,0 +1,31 @@
...
+
+use async_std::net::TcpListener;
+use async_std::task;
+use libparsec_platform_native::Transport;
+use rstest::rstest;
+
+#[rstest]
+fn test_transport() {
+    smol::block_on(async {
+        let addr = "127.0.0.1:8080";
```
touilleMan, line 11: Should not pick a fixed port, this typically prevent concurrent tests

Instead should use `TcpListener::bind("127.0.0.1:0")` to have the OS assign a port for us

oxidation/crates/parsec_api_protocol/tests/test_transport.rs
TimeEngineer: Done :+1:

oxidation/crates/platform_native/tests/test_transport.rs
TimeEngineer: Done :+1:

oxidation/crates/parsec_api_protocol/src/transport.rs
touilleMan: So we can have the `parsec_version` crate going yolo and parse the `parsec/_version.py` file.
This is not as crazy as it seems given 1) the data we want to extract is very simple so a `^__version__ = "(.+)"$` regex is enough and 2) a failing the compilation if the regex search doesn't succeed should be good enough detection if the `_version.py` is modified in an unexpected way

note: we can improve the regex to `^__version__ = "(v[0-9]+\.[0-9]+\.[0-9]+.*)"$` if we want to be more pedantic, but I'm not sure it is worth it...

oxidation/crates/platform_native/src/transport.rs
```diff
@@ -0,0 +1,71 @@
...
+
+pub struct Transport {
+    write: SplitSink<WebSocketStream<TcpStream>, Message>,
+    read: SplitStream<WebSocketStream<TcpStream>>,
+}
+
+// Only because we need user-agent
+async fn build_client_from_uri(uri: Uri) -> (TcpStream, Request<()>) {
+    dotenv::dotenv().ok();
+    let version = std::env::var("VERSION").unwrap();
```
FirelightFlagboy, line 23:
```suggestion
    let version = std::env::var("VERSION").expect("missing VERSION env variable");
```

oxidation/crates/platform_native/src/transport.rs
```diff
@@ -0,0 +1,71 @@
...
+    let user_agent = format!("parsec/{version}");
+    let authority = uri.authority().unwrap().as_str();
```
FirelightFlagboy, line 25:
```suggestion
    let authority = uri.authority().expect("missing authority segment").as_str();
```

oxidation/crates/platform_native/src/transport.rs
```diff
@@ -0,0 +1,71 @@
...
+        Request::builder()
+            .header("Host", host)
+            .header("Connection", "Upgrade")
+            .header("Upgrade", "websocket")
+            .header("Sec-WebSocket-Version", "13")
+            .header("Sec-WebSocket-Key", generate_key())
+            .header("User-Agent", user_agent)
+            .uri(uri)
+            .body(())
+            .unwrap(),
```
FirelightFlagboy, line 42:
```suggestion
            .expect("failed to build request"),
```

oxidation/crates/platform_native/src/transport.rs
```diff
@@ -0,0 +1,71 @@
...
+    )
+}
+
+impl Transport {
+    pub async fn init_for_client<T: AsRef<str>>(uri: T) -> Self {
+        let uri = uri.as_ref().parse::<Uri>().unwrap();
```
FirelightFlagboy, line 48:
```suggestion
        let uri = uri.as_ref().parse::<Uri>().expect("cannot parse string as Uri");
```

oxidation/crates/platform_native/src/transport.rs
```diff
@@ -0,0 +1,71 @@
...
+        let (stream, request) = build_client_from_uri(uri).await;
+        let (ws_stream, _) = client_async(request, stream).await.unwrap();
```
FirelightFlagboy, line 50:
```suggestion
        let (ws_stream, _) = client_async(request, stream).await.expect("Failed to open websocket connection");
```

oxidation/crates/platform_native/src/transport.rs
```diff
@@ -0,0 +1,71 @@
...
+        let (write, read) = ws_stream.split();
+
+        Self { write, read }
+    }
+    pub async fn init_for_server(stream: TcpStream) -> Self {
+        let ws_stream = accept_async(stream).await.unwrap();
```
FirelightFlagboy, line 56:
```suggestion
        let ws_stream = accept_async(stream).await.expect("failed to open websocket connection");
```

oxidation/crates/platform_web/Cargo.toml
```diff
@@ -3,4 +3,8 @@ name = "libparsec_platform_web"
 version = "0.0.0"
 edition = "2021"
 
+[[test]]
+name = "integration"
+path = "tests/tests.rs"
+
```
FirelightFlagboy, line 9: not required, cargo will auto discover the test https://doc.rust-lang.org/cargo/reference/cargo-targets.html#target-auto-discovery.

but you may want to renamed the file to `tests/integration.rs`

FirelightFlagboy (changes requested):

oxidation/crates/platform_native/src/transport.rs
TimeEngineer: Done :+1:

oxidation/crates/platform_native/src/transport.rs
TimeEngineer: Done :+1:

PR #2125, opened 2022-03-25
Add trustchain

TimeEngineer:


oxidation/crates/parsec_api_types/src/error.rs
```diff
@@ -0,0 +1,32 @@
...
+    Crypto { exc: CryptoError },
+
+    #[error("Invalid serialization")]
+    Serialization,
+
+    #[error("Invalid author: expected `{expected}`, got `None`")]
+    Validation1 { expected: DeviceID },
+
+    #[error("Invalid author: expected `{expected}`, got `{got}`")]
+    Validation2 { expected: DeviceID, got: DeviceID },
```
FirelightFlagboy, line 23: `Validation2` is almost identical to `Validation1`.
Can those variant be merged together by using `Option` ?

TimeEngineer: Good catch !

TimeEngineer: The downside is adding Some around the id in the python error, which is a bit weird

TimeEngineer: The change imply:

`DataError(f"Invalid author: expected `{repr_expected_author}`, got `{data.author}`")` 

---->

```Python
if data.author is None:
    `DataError(f"Invalid author: expected `{repr_expected_author}`, got `{data.author}`")` 
else:
    `DataError(f"Invalid author: expected `{repr_expected_author}`, got `Some({data.author})`")` 
```

TimeEngineer: Another solution is to type `got` as `String`

TimeEngineer: My bad, we can access self data

FirelightFlagboy (approved):

Issue #2110, opened 2022-03-22
Custom backend url should be trimmed


Max-7:
When creating an org using a custom backend, the URL should be trimmed, as a space at the end will result in a "Cannot connect to the server" error and a space before will result in the URL shown as invalid.

Example:
"parsec://host:port" is alright
" parsec://host:port" (with a space before) shows the URL as invalid (cannot click the Next button)
"parsec://host:port " (with a space after) is allowed but results in an error when the Next button is clicked

PR #2141, opened 2022-04-05
trim custom backend addr

jojoquivogn:
https://github.com/Scille/parsec-cloud/issues/2110


tests/core/gui/test_create_organization.py
```diff
@@ -391,6 +391,20 @@ def _user_widget_ready():
 
     await aqtbot.wait_until(_user_widget_ready)
 
+    def _custom_bckend_addr_widget_is_valid():
```
Max-7, line 394: This test does absolutely nothing. The dialog is started using a bootstrap address, which means `line_edit_backend_addr` will be read only, therefore `key_clicks` will not be able to change it.

This test should be put instead in `test_create_organization_custom_backend()` (L503), but it is currently marked as `skip`.  It would probably be a good thing to remove the skip and actually have a test using a custom backend addr.

Max-7 (changes requested):

vxgmichel (approved):
LGTM :+1:

PR #2144, opened 2022-04-06
Improve fuse/winfsp operation error log on deadlock (adding exc_info)

touilleMan:


Max-7 (approved):

vxgmichel (commented):
I don't think that this exception will provide any useful information, as it can only be raised when the trio loop is unreachable:
https://github.com/Scille/parsec-cloud/blob/5dc92a199122709e727af6e3a9b4373411703cfa/parsec/core/mountpoint/thread_fs_access.py#L65

PR #1407, opened 2020-08-28
Made history async

c4ffein:

PR #2147, opened 2022-04-06
Fix aqtbot.key_clicks in gui tests

vxgmichel:
.. as it started to behave weirdly on some systems (i.e. mine).

It is now an async method and waits for the keys to actually be taken into account by the corresponding text widget.

### DIFF
--- a/tests/core/gui/test_file_history.py
+++ b/tests/core/gui/test_file_history.py
@@ -60,4 +60,4 @@ async def test_file_history(
         assert hb2_w.label_user.text() == "Boby McBobFace"
         assert hb2_w.label_version.text() == "2"
 
-    await aqtbot.wait_until(_history_displayed)
+    await aqtbot.wait_until(_history_displayed, timeout=3000)
--- a/tests/core/gui/test_validators.py
+++ b/tests/core/gui/test_validators.py
@@ -19,6 +19,7 @@ def test_device_label_validator(qtbot, core_config):
     le.show()
 
     qtbot.keyClicks(le, "abcd")
+    qtbot.wait_until(lambda: le.text() == "abcd")
     assert le.is_input_valid()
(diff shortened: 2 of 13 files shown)


tests/core/gui/test_file_history.py
```diff
@@ -60,4 +60,4 @@ def _history_displayed():
         assert hb2_w.label_user.text() == "Boby McBobFace"
         assert hb2_w.label_version.text() == "2"
 
-    await aqtbot.wait_until(_history_displayed)
+    await aqtbot.wait_until(_history_displayed, timeout=3000)
```
touilleMan, line 63: why a specific timeout here ? This is most likely a code smell...

tests/core/gui/test_validators.py
```diff
@@ -19,6 +19,7 @@ def test_device_label_validator(qtbot, core_config):
     le.show()
 
     qtbot.keyClicks(le, "abcd")
+    qtbot.wait_until(lambda: le.text() == "abcd")
```
touilleMan, line 22: I think we shoul put the asserts in the wait_until callback instead of just waiting for the text to be present (to avoid having to think about when the validation kicks in)

touilleMan (approved):

tests/core/gui/test_file_history.py
vxgmichel: It's just that the history is very slow at the moment.

touilleMan: can't wait for #1407 to be done :trollface:

tests/core/gui/test_validators.py
vxgmichel: Having the callback return a boolean value is actually the equivalent of asserting this value.

PR #2156, opened 2022-04-07
Only look at the latest entry for a certificate in a submit command

AntoineDupre:
Simplify submit command to look only at the latest entry for a certificate

### DIFF
--- a/parsec/backend/postgresql/pki.py
+++ b/parsec/backend/postgresql/pki.py
@@ -35,10 +35,16 @@ from parsec.backend.postgresql.user_queries.create import (
     q_take_user_device_write_lock,
 )
 
-_q_get_last_pki_enrollment_from_certificate_sha1 = Q(
+_q_get_last_pki_enrollment_from_certificate_sha1_for_update = Q(
     f"""
-    SELECT enrollment_id, enrollment_state, submitted_on, submitter_accepted_device, accepter
-    FROM pki_enrollment
+    SELECT
+        enrollment_id,
+        enrollment_state,
+        submitted_on,
+        submitter_accepted_device,
+        accepter
+    FROM
+        pki_enrollment
     WHERE (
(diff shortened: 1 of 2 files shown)


tests/backend/test_pki.py
```diff
@@ -502,3 +502,42 @@ async def test_pki_info_rejected(anonymous_backend_sock, mallory, alice_backend_
...
+    async def _revoke(user_id):
+        now = pendulum.now()
+        revocation = RevokedUserCertificateContent(
+            author=alice.device_id, timestamp=now, user_id=user_id
+        ).dump_and_sign(alice.signing_key)
+
+        rep = await user_revoke(alice_backend_sock, revoked_user_certificate=revocation)
+        assert rep == {"status": "ok"}
+
+    for _ in range(4):
```
touilleMan, line 538: Doing it two times should be enough ?

parsec/backend/postgresql/pki.py
```diff
@@ -35,14 +35,16 @@
     q_take_user_device_write_lock,
 )
 
-_q_get_pki_enrollment_from_certificate_sha1 = Q(
+_q_get_last_pki_enrollment_from_certificate_sha1 = Q(
```
touilleMan, line 38: Add a `for_update` suffix ?

touilleMan (approved):

parsec/backend/postgresql/pki.py
```diff
@@ -35,14 +35,16 @@
...
     f"""
-    SELECT * FROM pki_enrollment
+    SELECT enrollment_id, enrollment_state, submitted_on, submitter_accepted_device, accepter
```
touilleMan, line 40: Should also remove the select * from the other queries

touilleMan (commented):
I don't think the newsfragment has to be a bugfix, empty is ok given this is more of an improvement than a bug fix

PR #2162, opened 2022-04-11
[GUI] Added button to update subscription for SAAS users

Max-7:


parsec/core/gui/commercial.py
```diff
@@ -0,0 +1,10 @@
+# Parsec Cloud (https://parsec.cloud) Copyright (c) AGPLv3 2016-2021 Scille SAS
+
+from parsec.core.types import BackendAddr
+
+SAAS_HOSTNAME = "saas.parsec.cloud"
+SAAS_UPDATE_SUBSCRIPTION_URL = "https://parsec.cloud/comment-changer-dabonnement-parsec/"
```
vxgmichel, line 6: Is there an internationalized version of this link?

vxgmichel (approved):
One comment, LGTM otherwise :+1:

parsec/core/gui/commercial.py
Max-7: Nop.

Max-7: I did take your comment into account and I now retrieve the URL from the translations. Also update the `SAAS_HOSTNAME` to be a list, in case we want to add other hosts.

PR #2185, opened 2022-04-15
Add POC: generation of schema from json

TimeEngineer:


oxidation/crates/poc/src/lib.rs
```diff
@@ -0,0 +1,31 @@
...
+use std::fs::File;
+use std::io::Read;
+use syn::{parse_macro_input, LitStr};
+
+use crate::utils::quote_struct;
+
+#[proc_macro]
+pub fn parsec_schema(path: TokenStream) -> TokenStream {
+    let path = parse_macro_input!(path as LitStr);
+    let file_path = std::env::current_dir().unwrap().join(&path.value());
```
touilleMan, line 16: I there an official doc stating that `std::env::current_dir()`  always point to the crate source directory when running a proc_macro ? (and if yes, we should add a comment here ^^)
This is typically the kind code that breaks when compiling from another crate...

oxidation/crates/poc/src/parser.rs
```diff
@@ -0,0 +1,50 @@
+// Parsec Cloud (https://parsec.cloud) Copyright (c) BSLv1.1 (eventually AGPLv3) 2016-2021 Scille SAS
```
touilleMan, line 1: Is this really a custom json parser ????

We should use miniserde instead, this would come especially handy as our schema grows bigger to handle more features

oxidation/crates/poc/enum.json
```diff
@@ -0,0 +1,29 @@
+{
+    "name": "FooEnum",
+    "variants": [
+        {
+            "name": "Foo0",
+            "fields": [
+                {
+                    "name": "Bar0",
+                    "type": "String",
+                    "rename": "bar0"
```
touilleMan, line 10: rename is a concept used by serde when generating the rust source, so it's not something we want to avoid in the json schema

Typically we should only have the `name` field (that indicate the name that will be serialized) and if this given name  is an invalid Rust name (for instance `type`), then we automatically add a trailing underscore (so the field is access as `my_stuff.type_` in Rust)

oxidation/crates/poc/api_protocol.json
```diff
@@ -0,0 +1,45 @@
+{
+    "family": "authenticated",
+    "cmds": [
+        {
+            "name": "BlockRead",
+            "req": [
+                {
+                    "name": "block_id",
+                    "type": "BlockID"
```
touilleMan, line 9: we want to add an optional `introduced_in_revision` field that contain an integer, this will be used to know if the field is always present or if it may be missing (in which case the type in Rust will be `Maybe<...>`

oxidation/crates/poc/enum.json
```diff
@@ -0,0 +1,29 @@
...
+                },
+                {
+                    "name": "Bar1",
+                    "type": "Vec<u8>",
```
touilleMan, line 14: Type should be not related to Rust (we have another an implementation in Python after all ^^)

So the types are:
 - generics: `Map` `List` `Enum` `Nested`
 - scalars: `Boolean` `Integer` `Float` `String` `Bytes` 

We may want to add additional scalars in the future (e.g. non-zero integer, non-negative integers etc.) but it's easy to add them when we will need them

touilleMan: And as we say those types must be deserialized to propor structs:

```
enum AnyType {
  Scalar(ScalarType),
 ...
}

struct ScalarType {
  name: String
}

impl TryFrom<syn::Type> for ScalarType {
  ...
}

struct MapType {
  item_type: AnyType,
}

impl TryFrom<syn::Type> for MapType {
  ...
}

[...]
```

oxidation/crates/poc/src/parser.rs
```diff
@@ -0,0 +1,336 @@
...
+            #variant,
+        };
+    }
+    variants
+}
+
+#[derive(Deserialize)]
+pub(crate) struct Schema {
+    name: String,
+    rename_all: Option<String>,
```
touilleMan, line 186: as I said, we really want to be explicit in the schema so no `rename`/`rename_all`

oxidation/crates/poc/src/parser.rs
```diff
@@ -0,0 +1,336 @@
...
+    // Struct based
+    fields: Option<Vec<Field>>,
```
touilleMan, line 188: No need for the Option

touilleMan:
The json format we want to support:

```js
// We really REALLY want to be able to add comment to the json file !
// So a pre-cooking stage should occure to strip them before passing json to miniserde
// Handling escaping is complexe and error prone, so better just ignore this entirely and
// strip everything after a `//`.
// This should be okay given we don't expect to have `//` anywhere else in this file.
// And if by mistake this occurs, the resulting json will most likely be invalid so
// the compilation will just fail.


// crates/parsec_api_serial/data/user_manifest.json
{  // struct ApiData
    "name": "WorkspaceManifest",  // Name of the struct
    "type": "workspace_manifest",  // Value used in the `type` field
    "serializing": "zip+msgpack",  // Allowed values are msgpack/zip+msgpack (in Python we also have json, but just ignore this for the moment !)
    "fields": {
        "author": {  // struct ApiField.Scalar  (APiField is an enum discrimined on `type`)
                     // enum ApiField {Scalar{...}, Map{..., key_type: ..., nested_type: ...}, List{..., nested_type: ...}, Nested{..., nested_fields: [...]}}
            "type": "DeviceID"
        },
        "children": {  // struct ApiField.Map
            "type": "Map<EntryName, EntryID>"
        },
        "created": {
            "type": "DateTime",
            "introduced_in_revision": 2  // Optional field to indicate the field may be missing in the schema
        },
        "id": {
            "type": "EntryID"
        },
        "timestamp": {
            "type": "DateTime"
        },
        "updated": {
            "type": "DateTime"
        },
        "version": {
            "type": "Integer"
        }
    }
}


// crates/parsec_api_serial/protocol/authenticated_block_read.json
// We should have a `generate_api_protocol_familly!()` macro that takes all the
// cmds json schemas of a given familly at once, e.g.:
//     generate_api_protocol_familly!("Authenticated",
//       "authenticated_block_read.json",  // of course all the macro check each json has a compatible `familly` field
//       "authenticated_block_create.json",
//       ...)
// The alternative would be to have a single json containing all the req/rep for
// a given familly, but this would make the json file totally unreadable ^^
{  // struct ApiProtocolCmd
    "name": "BlockRead",
    "familly": "authenticated",
    "req": {  // struct ApiCmdReq
        "cmd": "block_read",
        "other_fields": [
            {
                "name": "block_id",
                "type": "Optional<BlockID>",
            }
        ]
    },
    "reps": [
        {  // struct ApiCmdRep
            "status": "ok",
            "other_fields": [
                {
                    "name": "block",
                    "type": "Optional<Bytes>",
                }
            ]
        },
        {  // struct ApiCmdRep
            "status": "not_allowed",
            "other_fields": []
        },
        ...
    ]
}


// Bonus: not to be implemented for the moment !

// crates/parsec_api_serial/types.json
// Type conversion on custom struct is a tricky part given it involves:
// - the type name used in the json `type` field
// - the actual type the data should be deserialized into (e.g. `DeviceID`)
// - the way to serialize/deserialize the actual type (e.g. `#[serde_as(as = "DisplayFromStr")]`)
// This schema defines those conversion for the custom types.
{  // struct ApiTypes
    "types": [
        {
            "name": "DeviceID",  // name used in the json schema
            "type": "::parsec_api_types::DeviceID",  // type must contain the full path given it is most likely defined in another crate
            "serde_as": "DisplayFromStr" // will be put verbatim in a `#[serde_as(as = "...")]` block
        },
        {
            "name": "Bytes",
            "type": "Vec<u8>",
            "serde_as": "Bytes",
        },
        {
            "name": "DateTime",
            "type": "::parsec_api_types::DateTime",
            // We want to separate DateTime's serialization code given it is very
            // specific to our current way of doing serialization
            "serde_as": "::parsec_api_serial::serialize_datetime_as_extension_1"
        },
        ...
    ]
}
```

oxidation/crates/poc/api_protocol.json
```diff
@@ -0,0 +1,45 @@
+{
+    "family": "authenticated",
+    "cmds": [
+        {
+            "name": "BlockRead",
+            "req": [
```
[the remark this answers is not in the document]
TimeEngineer: Done

oxidation/crates/poc/enum.json
TimeEngineer: Done

oxidation/crates/poc/api_protocol.json
TimeEngineer: Done

oxidation/crates/poc/src/parser.rs
TimeEngineer: Miniserde doesn't implement complex enum

oxidation/crates/poc/src/parser.rs
TimeEngineer: Done

oxidation/crates/poc/src/parser.rs
touilleMan: not sure what you mean...

My point is a json file for data can represent either a struct or an enum. So we should deserialize each in it own struct (so we should have a `struct StructSchema` and `struct EnumSchema`) instead of having a single `struct Schema` that represent both (representing both oblige to have optional fields which not clear which one in present in which case when using the struct).

oxidation/crates/parsec_api_types/src/ext_types.rs
```diff
@@ -410,3 +410,86 @@ pub mod maybe_field {
...
+
+        impl<'de, T, U> serde::de::Visitor<'de> for MaybeVisitor<T, U>
+        where
+            U: serde_with::DeserializeAs<'de, T>,
+            T: Default,
+        {
+            type Value = Maybe<T>;
+
+            fn expecting(&self, formatter: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+                formatter.write_str("maybe")
```
touilleMan, line 468: According to documentation, expecting should be:

> The message should complete the sentence "This Visitor expects to receive ...", for example the message could be "an integer between 0 and 64". The message should not be capitalized and should not end with a period.

Now I don't see when this expecting would be used in our usecase (given field not present is ok and field present mean we pass it to T), so maybe have an unreachable!() instead ?

PR #2189, opened 2022-04-19
[GUI] Added a recover device button when the list of devices is empty

### LINKED ISSUE #2184: Add a "recover a device" button to the GUI for empty parsec sessions
When a user wants to recover a device, they are likely to start with a fresh parsec install. In that case, the main prompts are:
- Create an organization
- Join an existing organization
A third option for recovering a device should be added to this list.

Max-7:
Closes #2184


vxgmichel (approved):
LGTM :+1:

PR #2194, opened 2022-04-20
Bump api version to 3.0 

vxgmichel:
so we can later break the compatibility with version 2.4 and lower

### DIFF
--- a/parsec/api/protocol/handshake.py
+++ b/parsec/api/protocol/handshake.py
@@ -263,6 +263,8 @@ class ServerHandshake:
         )
 
         # Use the correct serializer
+        # `settle_compatible_versions` is called before,
+        # so we already settled on a version from `self.SUPPORTED_API_VERSIONS`
         if client_api_version.version == 1:
             serializer = apiv1_handshake_answer_serializer
         else:
@@ -356,7 +358,7 @@ class ServerHandshake:
 
                 # Provides compatibility with API version 2.4 and below
                 # TODO: Remove once API v2.x is deprecated
-                if self.client_api_version.version == 2 and self.client_api_version.revision < 5:
+                if (2, 0) <= self.client_api_version < (2, 5):
                     returned_challenge = verify_key.verify(answer)
 
                 # Used in API v2.5+ and API v3.x
@@ -484,7 +486,7 @@ class AuthenticatedClientHandshake(BaseClientHandshake):
 
         # Provides compatibility with API version 2.4 and below
         # TODO: Remove once API v2.x is deprecated
-        if self.backend_api_version.version == 2 and self.backend_api_version.revision < 5:
+        if (2, 0) <= self.backend_api_version < (2, 5):
             answer = self.user_signkey.sign(challenge)
 
         # Used in API v2.5+ and API v3.x
--- a/parsec/api/protocol/handshake.py
+++ b/parsec/api/protocol/handshake.py
@@ -23,7 +23,7 @@ from parsec.api.protocol.invite import (
     InvitationType,
     InvitationTypeField,
 )
-from parsec.api.version import ApiVersion, API_V1_VERSION, API_V2_VERSION
+from parsec.api.version import ApiVersion, API_V1_VERSION, API_V2_VERSION, API_V3_VERSION
 
 
 class HandshakeError(ProtocolError):
@@ -211,7 +211,7 @@ handshake_result_serializer = serializer_factory(HandshakeResultSchema)
 
 class ServerHandshake:
     # Class attribute
-    SUPPORTED_API_VERSIONS = (API_V2_VERSION, API_V1_VERSION)
+    SUPPORTED_API_VERSIONS = (API_V1_VERSION, API_V2_VERSION, API_V3_VERSION)
     CHALLENGE_SIZE = 48
 
     def __init__(self) -> None:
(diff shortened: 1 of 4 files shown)


parsec/api/protocol/handshake.py
```diff
@@ -256,13 +256,17 @@ def process_answer_req(self, req: bytes) -> None:
...
+            self.SUPPORTED_API_VERSIONS, [client_api_version]
+        )
+
+        # Use the correct serializer
+        if client_api_version.version == 1:
             serializer = apiv1_handshake_answer_serializer
         else:
-            # Imcompatible version !
-            raise HandshakeAPIVersionError(self.SUPPORTED_API_VERSIONS, [client_api_version])
+            serializer = handshake_answer_serializer
```
touilleMan, line 269: what if `client_api_version.version > 3` ? I guess this is catched by `_settle_compatible_versions` (hence why it has been moved up), but it would be better to have a comment about that

parsec/api/protocol/handshake.py
```diff
@@ -355,12 +354,16 @@ def build_result_req(self, verify_key: Optional[VerifyKey] = None) -> bytes:
                 answer = self.answer_data["answer"]
                 assert isinstance(answer, bytes)
 
-                if self.client_api_version >= (2, 5):
+                # Provides compatibility with API version 2.4 and below
+                # TODO: Remove once API v2.x is deprecated
+                if self.client_api_version.version == 2 and self.client_api_version.revision < 5:
```
touilleMan, line 359: why not `if (2, 0) <= self.client_api_version <= (2, 5)` ?

touilleMan: btw this works because we don't support apiv1 authenticated handshake anymore, but in theory this function should 1) check api version (and typically set boolean flags used by the rest of the function) 2) do the `if self.answer_type == HandshakeType.AUTHENTICATED:`

parsec/api/protocol/handshake.py
```diff
@@ -479,14 +482,16 @@ def process_challenge_req(self, req: bytes) -> bytes:
...
 
-        # TO-DO remove the else for the next release
-        if self.backend_api_version >= (2, 5):
-            # TO-DO Need to use "BaseSignedData" ?
+        # Provides compatibility with API version 2.4 and below
+        # TODO: Remove once API v2.x is deprecated
+        if self.backend_api_version.version == 2 and self.backend_api_version.revision < 5:
+            answer = self.user_signkey.sign(challenge)
+
+        # Used in API v2.5+ and API v3.x
```
touilleMan, line 490: This is the part where we've already broken our "compatibility across revision" promise 😄 

So next time we should have create 3.0 instead of 2.5, and then only test on `self.backend_api_version.version` for the vast majority of compatibility handling code 

we would use `self.backend_api_version.revision` only to know if a new command (or, in an existing command,  a new field adding information without changing behavior if it is ignored/not provided) is available. Right ?

tests/backend/test_handshake.py
```diff
@@ -54,9 +55,11 @@ async def test_handshake_incompatible_version(backend, server_factory):
         assert unpackb(result_req) == {
             "handshake": "result",
             "result": "bad_protocol",
-            "help": "No overlap between client API versions {3.0} and backend API versions {"
-            + str(API_VERSION)
-            + ", 1.3}",
+            "help": "No overlap between client API versions {"
```
touilleMan, line 58: :/

```python
f"No overlap between client API versions {{{incompatible_version}}} and backend API version {{{', '.join(map(str, ServerHandshake.SUPPORTED_API_VERSIONS))}}}"
```

(and if you find `{{{x}}}` ugly... well that's why we are supposed to use back-tick everywhere 😄

touilleMan (approved):

parsec/api/protocol/handshake.py
vxgmichel: Exactly, I'm adding the comment

tests/backend/test_handshake.py
vxgmichel: I don't think it's ugly, it's fine with me!

> well that's why we are supposed to use back-tick everywhere

I think curly braces are used here because we're talking about sets of versions.

parsec/api/protocol/handshake.py
vxgmichel: > This is the part where we've already broken our "compatibility across revision" promise

On the contrary, our rule is fulfilled: we have backward compatibility going both ways across revisions.

I agree that it would have been simpler in terms of code to bump API v3 when we introduced the new signing mechanism but since we made sure that every thing remained compatible on both the server and client side then it's ok. 

> we would use self.backend_api_version.revision only to know if a new command (or, in an existing command, a new field adding information without changing behavior if it is ignored/not provided) is available. Right ?

I guess most changes would work as you describe, but we might need more advanced compatibility in some rare cases (like this one).

parsec/api/protocol/handshake.py
vxgmichel: > why not if (2, 0) <= self.client_api_version <= (2, 5) 

Sure, why not.

> btw this works because we don't support apiv1 authenticated handshake anymore, but in theory this function should 1) check api version (and typically set boolean flags used by the rest of the function) 2) do the if self.answer_type == HandshakeType.AUTHENTICATED:

I typically like to keep this compatibility as local as possible. Do you mean something like:
```python
use_old_signing_scheme = (2, 0) <= self.client_api_version <= (2, 5)
[...]
if self.answer_type == HandshakeType.AUTHENTICATED:
    [...]
    if use_old_signing_scheme:
        [...]
    else:
        [...]
```
or am I missing something? :thinking:

parsec/api/protocol/handshake.py
touilleMan: > On the contrary, our rule is fulfilled: we have backward compatibility going both ways across revisions.

I disagree:
 - we have Parsec implementations that are able to handle both <2.5 and >=2.5 so we have backward compatibility
 - BUT we don't have "any Parsec client implementation that support api v2.x is compatible with any Parsec server implementation thas support api v2.y"

So we have "compatibility across Parsec versions" but not "compatibility across revisions of of a given parsec api version"

PR #2195, opened 2022-04-20
Migrate to python 3.9

vxgmichel:
The tests in #2183 have been conclusive.


packaging/snap/snap/snapcraft.yaml
```diff
@@ -176,26 +176,26 @@ parts:
...
 
     build-packages:
-      - python3.8-venv
-      - python3.8
+      - python3.9-venv
+      - python3.9
     stage-packages:
       # Why so much packages ?
-      # Well because Python 3.8 is part of base image core20 so Snapcraft
+      # Well because Python 3.9 is part of base image core20 so Snapcraft
```
touilleMan, line 183: Is the comment still relevant with 3.9 instead of 3.8

(I guess we may want to keep this hack even if 3.9 is not part of base image core20 just in case we will endup in the same issue when switching to core22...)

parsec/core/gui/instance_widget.py
```diff
@@ -64,7 +64,8 @@ def check_macfuse_version() -> bool:
     import plistlib
 
     try:
-        local_version = Version(plistlib.readPlist(macfuse_plist_path)["CFBundleVersion"])
+        with open(macfuse_plist_path, "rb") as plist_file:
```
touilleMan, line 67: missing encoding parameter, it's better to be explicit on utf8 ^^

touilleMan (approved):

PR #2197, opened 2022-04-20
Migrate organization_bootstrap command to ANONYMOUS_CMDS

### LINKED ISSUE #2196: Migrate organization_bootstrap command to ANONYMOUS_CMDS
The `organization_bootstrap` command is the only command still in use on the client side from the APIv1 (as part of `APIV1_ANONYMOUS_CMDS`). Migrating this command to the new HTTP api for anonymous command would allow to finally get rid of all the legacy code related to APIv1 (once all the clients have the new `organization_bootstrap` implementation).

vxgmichel:
Fix #2196

### DIFF
--- a/parsec/api/protocol/organization.py
+++ b/parsec/api/protocol/organization.py
@@ -37,6 +37,24 @@ apiv1_organization_bootstrap_serializer = CmdSerializer(
 )
 
 
+class OrganizationBootstrapReqSchema(BaseReqSchema):
+    bootstrap_token = fields.String(required=True)
+    root_verify_key = fields.VerifyKey(required=True)
+    user_certificate = fields.Bytes(required=True)
+    device_certificate = fields.Bytes(required=True)
+    redacted_user_certificate = fields.Bytes(required=True)
+    redacted_device_certificate = fields.Bytes(required=True)
+
+
+class OrganizationBootstrapRepSchema(BaseRepSchema):
+    pass
+
+
+organization_bootstrap_serializer = CmdSerializer(
+    OrganizationBootstrapReqSchema, OrganizationBootstrapRepSchema
+)
+
+
 class OrganizationBootstrapWebhookSchema(BaseSchema):
     organization_id = OrganizationIDField(required=True)
     device_id = DeviceIDField(required=True)
--- a/parsec/backend/http.py
+++ b/parsec/backend/http.py
@@ -367,33 +367,51 @@ class HTTPComponent:
         return None, data
 
     async def _http_api_anonymous(self, req: HTTPRequest, **kwargs: str) -> HTTPResponse:
+        # Check whether the organization exists
         try:
             organization_id = OrganizationID(kwargs["organization_id"])
             await self._organization_component.get(organization_id)
-        except (ValueError, OrganizationNotFoundError):
+        except OrganizationNotFoundError:
+            organization_exists = False
+        except ValueError:
+            return HTTPResponse.build_msgpack(404, {})
+        else:
+            organization_exists = True
+
+        # Reply to GET
+        if req.method == "GET":
+            status = 200 if organization_exists else 404
+            return HTTPResponse.build_msgpack(status, {})
+
+        # Reply early to POST when the organization doesn't exists
+        if not organization_exists and not self._config.organization_spontaneous_bootstrap:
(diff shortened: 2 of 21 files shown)


parsec/api/protocol/organization.py
```diff
@@ -37,6 +37,32 @@ class APIV1_OrganizationBootstrapRepSchema(BaseRepSchema):
 )
 
 
+class OrganizationBootstrapReqSchema(BaseReqSchema):
+    bootstrap_token = fields.String(required=True)
+    root_verify_key = fields.VerifyKey(required=True)
+    user_certificate = fields.Bytes(required=True)
+    device_certificate = fields.Bytes(required=True)
+    # Same certificates than above, but expurged of human_handle/device_label
+    # Backward compatibility prevent those field to be required, however
```
touilleMan, line 46: No need for backward compatibility here given this command is brand new !

parsec/backend/organization.py
```diff
@@ -140,10 +142,20 @@ async def api_authenticated_organization_stats(self, client_ctx, msg):
...
+        # Use the correct serializer depending on the API
+        if isinstance(client_ctx, APIV1_AnonymousClientContext):
+            serializer = apiv1_organization_bootstrap_serializer
+        else:
+            assert isinstance(client_ctx, AnonymousClientContext)
+            serializer = organization_bootstrap_serializer
+
+        msg = serializer.req_load(msg)
+
         bootstrap_token = msg["bootstrap_token"]
```
touilleMan, line 159: should add a comment above where `if "redacted_user_certificate" in msg:` is done given this won't be needed once APIv1 is dropped

parsec/core/cli/bootstrap_organization.py
```diff
@@ -32,22 +31,19 @@ async def _bootstrap_organization(
         device_label_raw = await aprompt("Device label", default=platform.node())
         device_label = DeviceLabel(device_label_raw)
 
-    async with apiv1_backend_anonymous_cmds_factory(addr=addr) as cmds:
-        async with spinner("Bootstrapping organization in the backend"):
-            new_device = await do_bootstrap_organization(
-                cmds=cmds, human_handle=human_handle, device_label=device_label
-            )
+    async with spinner("Bootstrapping organization in the backend"):
+        new_device = await do_bootstrap_organization(
```
touilleMan, line 35: sooo much better not to have to create a cmds ! \o/

touilleMan (approved):

touilleMan (approved):

parsec/backend/http.py
```diff
@@ -367,33 +367,51 @@ async def _api_check_auth_and_load_body(self, req: HTTPRequest, req_serializer):
         return None, data
 
     async def _http_api_anonymous(self, req: HTTPRequest, **kwargs: str) -> HTTPResponse:
+        # Check whether the organization exists
```
touilleMan, line 370: Are we sure this is what we want to do ?

On the one hand it's tempting to have bootstrap as part of the anonymous family given they all refer to an organization_id, but on the other hand the fact the organization should not be bootstrapped (and may even not be created at all) make it very special (hence those non-trivial changes that may be error prone...)

I don't have a strong opinion, but I think it's interesting to think a bit about that before choosing our poison ;-)

touilleMan (approved):

parsec/backend/http.py
vxgmichel: Agreed, as a first implementation I chose to follow what was already done here:
https://github.com/Scille/parsec-cloud/blob/f4a624d2c0152842f4cd5d82b510728d85c3428a/parsec/backend/handshake.py#L194-L235

But we might want to change that, I don't have a strong opinion either.

touilleMan: > But we might want to change that, I don't have a strong opinion either.

So how do we decide if it's good or not ? :trollface:

vxgmichel: > the fact the organization should not be bootstrapped (and may even not be created at all) make it very special (hence those non-trivial changes that may be error prone...)

The fact that the organization should not be bootsrapped is already managed correctly in the command itself. The weird part is the spontaneous creation of the organization if the backend configuration allows it.

vxgmichel: Let's keep it this way for the moment and investigate later how we want to evolve this command.

PR #2212, opened 2022-04-29
 Add optional smartcard tests when parsec_ext is installed

vxgmichel:


tests/test_cli.py
```diff
@@ -737,7 +737,9 @@ async def test_pki_enrollment(tmp_path, mocked_parsec_ext_smartcard, backend, al
         async def _cli_invoke_in_thread(cmd: str):
             # We must run the command from another thread given it will create it own trio loop
             with trio.fail_after(1):
-                return await trio.to_thread.run_sync(runner.invoke, cli, cmd)
+                return await trio.to_thread.run_sync(
+                    lambda: runner.invoke(cli, cmd, env={"DEBUG": "1"})
```
touilleMan, line 741: good idea, maybe add a comment `# pass DEBUG env var for better output on crash`

tests/fixtures.py
```diff
@@ -778,12 +780,92 @@ async def _sock_from_other_organization_factory(
...
+        subprocess.run(command, shell=True, check=True, cwd=tmp_path)
+
+    # Concatenate the key and the certificate
+    combined_data = (tmp_path / "client_cert.pem").read_bytes() + (
+        tmp_path / "client_key.pem"
+    ).read_bytes()
+    (tmp_path / "client_combined.pem").write_bytes(combined_data)
+
+    # Return certificates and password
+    return tmp_path / "ca_cert.pem", tmp_path / "client_combined.pem", "test"
```
touilleMan, line 808: don't want to use `P@ssw0rd` for password ? (`test` seems too generic when debugging)

touilleMan (approved):
:+1:

tests/test_cli.py
vxgmichel: 4a1617256

tests/fixtures.py
vxgmichel: 4a1617256

PR #2218, opened 2022-05-02
Add storage

TimeEngineer:


parsec/core/gui/tr/parsec_en.po
```diff
@@ -2380,18 +2380,18 @@ msgstr "Reencrypt the workspace"
 msgid "ACTION_WORKSPACE_REENCRYPTION_CONFIRM"
 msgstr "Reencrypt"
 
-msgid "TEXT_WORKPACE_REENCRYPT_OFFLINE_ERROR"
```
touilleMan, line 2383: make a separate PR for this (this one is already big enough ^^)

TimeEngineer: Done :+1:

oxidation/crates/parsec_core_fs/src/storage/chunk_storage.rs
```diff
@@ -0,0 +1,404 @@
...
+            .first(conn)
+            .map_err(|_| FSError::QueryTable("chunks: get_nb_blocks"))
+    }
+
+    fn get_total_size(&self) -> FSResult<i64> {
+        let conn = &mut *self.conn().lock().expect("Mutex is poisoned");
+        chunks::table
+            .select(coalesce_total_size())
+            .first(conn)
+            .map_err(|_| FSError::QueryTable("chunks: get_total_size"))
```
FirelightFlagboy, line 69: could be useful to save the error returned by `first(conn)` for debug purpose

tests/core/test_invite.py
```diff
@@ -111,7 +113,7 @@ async def _run_greeter():
 
         await in_progress_ctx.do_create_new_device(author=alice, device_label=granted_device_label)
 
-    with trio.fail_after(1):
+    with trio.fail_after(3):
```
touilleMan, line 116: should not be needed, (well the correct way of doing this is to replace this by the new shinny `real_clock_timeout` which provide a consistent across tests (and configurable) timeout time

tests/core/test_invite.py
```diff
@@ -269,7 +272,7 @@ async def _run_greeter():
             profile=granted_profile,
         )
 
-    with trio.fail_after(1):
+    with trio.fail_after(3):
```
touilleMan, line 275: same

tests/conftest.py
```diff
@@ -87,6 +87,7 @@ def pytest_addoption(parser):
         "--realcrypto", action="store_true", help="Don't mock crypto operation to save time"
     )
     parser.addoption("--runrust", action="store_true", help="Don't skip rust tests")
+    parser.addoption("--runpy", action="store_true", help="Don't skip py tests")
```
touilleMan, line 90: what are `py` tests ? this should be more explicit
(and I don't see any test using this flag)

tests/common.py
```diff
@@ -30,7 +30,7 @@
 #   in case a mistake in the code lead to a deadlock :(
 # So the solution is to make this configurable: a good middleground by default
 # and a long long time on the CI.
-_SIDE_EFFECTS_TIMEOUT = 3
+_SIDE_EFFECTS_TIMEOUT = 4
```
touilleMan, line 33: 3s on a regular pc should be huge (we increase this to 10s on CI just to be sure), in theory we should be able to keep this to 1s or even lower.
So increasing this seems really smelly... what the reason ?

oxidation/crates/parsec_core_fs/src/error.rs
```diff
@@ -0,0 +1,66 @@
+// Parsec Cloud (https://parsec.cloud) Copyright (c) BSLv1.1 (eventually AGPLv3) 2016-2021 Scille SAS
+
+use thiserror::Error;
+use uuid::Uuid;
+
+use parsec_api_crypto::CryptoError;
+use parsec_api_types::FileDescriptor;
+
+#[derive(Error, Debug)]
+pub enum FSError {
```
FirelightFlagboy, line 10: Would be useful to add the errors returned by `diesel` to their proper variants.

I don't like to lose the info of the reason of the error return by the library `diesel`

oxidation/crates/parsec_core_fs/src/storage/chunk_storage.rs
```diff
@@ -0,0 +1,404 @@
...
+    }
+
+    fn is_chunk(&self, chunk_id: ChunkID) -> FSResult<bool> {
+        let conn = &mut *self.conn().lock().expect("Mutex is poisoned");
+        chunks::table
+            .select(count_star())
+            .filter(chunks::chunk_id.eq((*chunk_id).as_ref()))
+            .first::<i64>(conn)
+            .map_err(|_| FSError::QueryTable("chunks: is_chunk"))
```
FirelightFlagboy, line 78: could be useful to save the error returned by `first::<i64>(conn)` for debug purpose

oxidation/crates/parsec_core_fs/src/storage/chunk_storage.rs
```diff
@@ -0,0 +1,404 @@
...
+        let mut res = Vec::with_capacity(chunk_ids.len());
+
+        let conn = &mut *self.conn().lock().expect("Mutex is poisoned");
+        for bytes_id_list_chunk in bytes_id_list.chunks(SQLITE_MAX_VARIABLE_NUMBER) {
+            res.append(
+                &mut chunks::table
+                    .select(chunks::chunk_id)
+                    .filter(chunks::chunk_id.eq_any(bytes_id_list_chunk))
+                    .load::<Vec<u8>>(conn)
+                    .map_err(|_| FSError::QueryTable("chunks: get_local_chunk_ids"))?
```
FirelightFlagboy, line 97: could be useful to save the error returned by `load::<Vec<u8>>(conn)` for debug purpose

parsec/core/fs/storage/workspace_storage.py
```diff
@@ -421,13 +439,120 @@ async def mark_prevent_sync_pattern_fully_applied(self, pattern: Pattern[str]) -
...
 
+_PyWorkspaceStorage = WorkspaceStorage
+if not TYPE_CHECKING:
+    try:
+        from libparsec.types import WorkspaceStorage as _RsWorkspaceStorage
+    except:
+        pass
+    else:
+
+        class WorkspaceStorage:
```
touilleMan, line 463: better to define this before the `if not typechecking`, this save a useless level of indentation, eg:
```python
class SecretKey:
    [...]

_PySecretKey = SecretKey
if not TYPE_CHECKING:
    try:
        from libparsec.hazmat import SecretKey as _RsSecretKey
    except ImportError:
        pass
    else:
        SecretKey = _RsSecretKey
```

oxidation/crates/parsec_core_fs/src/storage/chunk_storage.rs
```diff
@@ -0,0 +1,404 @@
...
+
+    fn get_chunk(&self, chunk_id: ChunkID) -> FSResult<Vec<u8>> {
+        let accessed_on = DateTime::now().get_f64_with_us_precision() as f32;
+
+        let conn = &mut *self.conn().lock().expect("Mutex is poisoned");
+        let changes =
+            diesel::update(chunks::table.filter(chunks::chunk_id.eq((*chunk_id).as_ref())))
+                .set(chunks::accessed_on.eq(accessed_on))
+                .execute(conn)
+                .map_err(|_| FSError::UpdateTable("chunks: get_chunk"))?
```
FirelightFlagboy, line 121: could be useful to save the error returned by `execute(conn)` for debug purpose

oxidation/crates/parsec_core_fs/src/storage/chunk_storage.rs
```diff
@@ -0,0 +1,404 @@
...
+
+        if !changes {
+            return Err(FSError::LocalMiss(*chunk_id));
+        }
+
+        let ciphered = chunks::table
+            .select(chunks::data)
+            .filter(chunks::chunk_id.eq((*chunk_id).as_ref()))
+            .first::<Vec<u8>>(conn)
+            .map_err(|_| FSError::QueryTable("chunks: get_chunk"))?;
```
FirelightFlagboy, line 132: could be useful to save the error returned by `first::<vec<u8>>(conn)` for debug purpose

oxidation/crates/parsec_core_fs/src/storage/chunk_storage.rs
```diff
@@ -0,0 +1,404 @@
...
+        };
+
+        let conn = &mut *self.conn().lock().expect("Mutex is poisoned");
+        diesel::insert_into(chunks::table)
+            .values(&new_chunk)
+            .on_conflict(chunks::chunk_id)
+            .do_update()
+            .set(&new_chunk)
+            .execute(conn)
+            .map_err(|_| FSError::InsertTable("chunks: set_chunk"))?;
```
FirelightFlagboy, line 156: could be useful to save the error returned by `execute(conn)` for debug purpose

oxidation/crates/parsec_core_fs/src/storage/chunk_storage.rs
```diff
@@ -0,0 +1,404 @@
...
+        Ok(())
+    }
+
+    fn clear_chunk(&self, chunk_id: ChunkID) -> FSResult<()> {
+        let conn = &mut *self.conn().lock().expect("Mutex is poisoned");
+        let changes =
+            diesel::delete(chunks::table.filter(chunks::chunk_id.eq((*chunk_id).as_ref())))
+                .execute(conn)
+                .map_err(|_| FSError::DeleteTable("chunks: clear_chunk"))?
```
FirelightFlagboy, line 165: could be useful to save the error returned by `execute(conn)` for debug purpose

oxidation/crates/parsec_core_fs/src/storage/chunk_storage.rs
```diff
@@ -0,0 +1,404 @@
...
+        }
+
+        Ok(())
+    }
+
+    fn run_vacuum(&self) -> FSResult<()> {
+        let conn = &mut *self.conn().lock().expect("Mutex is poisoned");
+        sql_query("VACUUM;")
+            .execute(conn)
+            .map_err(|_| FSError::Vacuum)?;
```
FirelightFlagboy, line 179: could be useful to save the error returned by `execute(conn)` for debug purpose

oxidation/crates/parsec_core_fs/src/storage/chunk_storage.rs
```diff
@@ -0,0 +1,404 @@
...
+
+    fn block_limit(&self) -> u64 {
+        self.cache_size() / DEFAULT_BLOCK_SIZE
+    }
+
+    fn clear_all_blocks(&self) -> FSResult<()> {
+        let conn = &mut *self.conn().lock().expect("Mutex is poisoned");
+        diesel::delete(chunks::table)
+            .execute(conn)
+            .map_err(|_| FSError::DeleteTable("chunks: clear_all_blocks"))?;
```
FirelightFlagboy, line 197: could be useful to save the error returned by `execute(conn)` for debug purpose

oxidation/crates/parsec_core_fs/Cargo.toml
```diff
@@ -0,0 +1,27 @@
+[package]
```
touilleMan, line 1: crate should be put in `oxidation/libparsec/crate/core_fs` , and crate named `libparsec_core_fs` (see my comment https://github.com/Scille/parsec-cloud/pull/2334#discussion_r878904512)

oxidation/crates/parsec_core_fs/Cargo.toml
```diff
@@ -0,0 +1,27 @@
...
+name = "parsec_core_fs"
+version = "0.0.0"
+edition = "2021"
+license = " BUSL-1.1"
+autotests = false
+
+# [[test]]
```
touilleMan: dead code ?

oxidation/crates/parsec_core_fs/migrations/2022-04-26-103106_chunks/down.sql
```diff
@@ -0,0 +1,4 @@
+-- Parsec Cloud (https://parsec.cloud) Copyright (c) BSLv1.1 (eventually AGPLv3) 2016-2021 Scille SAS
+
+-- This file should undo anything in `up.sql`
+DROP TABLE IF EXISTS chunks;
```
touilleMan, line 4: You must create tests if you add a way to downgrade the datamodel.

I think it would be better to not provide any way to downgrade given it's not the goal of this PR (and YAGNI), and a poorly tested feature we might be temped to use in the future is far worst than no feature at all

oxidation/crates/parsec_core_fs/migrations/2022-04-26-103106_chunks/up.sql
```diff
@@ -0,0 +1,10 @@
+-- Parsec Cloud (https://parsec.cloud) Copyright (c) BSLv1.1 (eventually AGPLv3) 2016-2021 Scille SAS
```
touilleMan, line 1: date in the filename is wrong, you should have a look at which Parsec version introduced the change in the datamodel and use it release date instead

oxidation/crates/parsec_core_fs/src/lib.rs
```diff
@@ -0,0 +1,20 @@
...
+mod storage;
+mod workspacefs;
+
+#[macro_use]
+extern crate diesel;
+
+pub use error::*;
+pub use storage::*;
+
+pub enum Language {
```
touilleMan, line 17: `libparsec` should be separated to the GUI, so you should not handle language in the Rust code (besides error messages retuned by liparsec are technical one, so no need to translate them)

oxidation/crates/parsec_core_fs/src/storage/local_database.rs
```diff
@@ -0,0 +1,44 @@
+// Parsec Cloud (https://parsec.cloud) Copyright (c) BSLv1.1 (eventually AGPLv3) 2016-2021 Scille SAS
+
+use diesel::connection::SimpleConnection;
+use diesel::r2d2::{ConnectionManager, Pool, PooledConnection};
+use diesel::SqliteConnection;
+use std::path::PathBuf;
+
+use crate::error::{FSError, FSResult};
+
+pub const SQLITE_MAX_VARIABLE_NUMBER: usize = 999;
```
touilleMan, line 10: Should add a comment to explain what this is for, and the rational for this uncommon value

oxidation/crates/parsec_core_fs/src/storage/mod.rs
```diff
@@ -0,0 +1,12 @@
+// Parsec Cloud (https://parsec.cloud) Copyright (c) BSLv1.1 (eventually AGPLv3) 2016-2021 Scille SAS
+
+mod chunk_storage;
+mod local_database;
+mod manifest_storage;
+#[allow(dead_code)]
```
touilleMan, line 6: why ?

touilleMan (changes requested):

oxidation/crates/parsec_core_fs/src/workspacefs/file_transactions.rs
```diff
@@ -0,0 +1,33 @@
...
+    // The corresponding file is locked while performing the change (i.e. between
+    // the reading and writing of the corresponding manifest) in order to avoid
+    // race conditions and data corruption.
+
+    // The table below lists the effects of the 6 file transactions:
+    // - close    -> remove file descriptor from local storage
+    // - write    -> affects file content and possibly file size
+    // - truncate -> affects file size and possibly file content
+    // - read     -> no side effect
+    // - flush    -> no-op
```
FirelightFlagboy, line 28:
```suggestion
/// A stateless class to centralize all file transactions.
///
/// The actual state is stored in the local storage and file transactions
/// have access to the remote loader to download missing resources.
///
/// The exposed transactions all take a file descriptor as first argument.
/// The file descriptors correspond to an entry id which points to a file
/// on the file system (i.e. a file manifest).
///
/// The corresponding file is locked while performing the change (i.e. between
/// the reading and writing of the corresponding manifest) in order to avoid
/// race conditions and data corruption.
///
/// The table below lists the effects of the 6 file transactions:
/// - close    -> remove file descriptor from local storage
/// - write    -> affects file content and possibly file size
/// - truncate -> affects file size and possibly file content
/// - read     -> no side effect
/// - flush    -> no-op
pub struct FileTransactions {
```

use doc comments

oxidation/libparsec_python/src/addrs.rs
```diff
@@ -97,7 +97,7 @@ impl BackendAddr {
 
 #[pyclass]
 #[derive(PartialEq, Eq, Clone)]
-pub(crate) struct BackendOrganizationAddr(parsec_api_types::BackendOrganizationAddr);
+pub(crate) struct BackendOrganizationAddr(pub parsec_api_types::BackendOrganizationAddr);
```
FirelightFlagboy, line 100:
```suggestion
pub(crate) struct BackendOrganizationAddr(pub(crate) parsec_api_types::BackendOrganizationAddr);
```

shouldn't it be `pub(crate)` ? since the struct is only visible to the crate ?

parsec/core/fs/storage/workspace_storage.py
TimeEngineer: Changed

oxidation/crates/parsec_core_fs/Cargo.toml
TimeEngineer: Removed

oxidation/crates/parsec_core_fs/src/storage/mod.rs
TimeEngineer: Because I don't use user_storage for the moment

oxidation/crates/parsec_core_fs/src/storage/manifest_storage.rs
```diff
@@ -0,0 +1,628 @@
...
+
+        Ok(())
+    }
+
+    #[cfg(test)]
+    fn drop_db(&self) -> FSResult<()> {
+        let conn = &mut *self.conn.lock().expect("Mutex is poisoned");
+        sql_query("DROP TABLE IF EXISTS vlobs;")
+            .execute(conn)
+            .map_err(|_| FSError::DropTable("vlobs"))?;
```
FirelightFlagboy, line 153: save the error into the variant

oxidation/crates/parsec_core_fs/src/storage/manifest_storage.rs
```diff
@@ -0,0 +1,628 @@
...
+
+        sql_query("DROP TABLE IF EXISTS realm_checkpoints;")
+            .execute(conn)
+            .map_err(|_| FSError::DropTable("realm_checkpoints"))?;
```
FirelightFlagboy, line 157: save the error into the variant

oxidation/crates/parsec_core_fs/src/storage/manifest_storage.rs
```diff
@@ -0,0 +1,628 @@
...
+
+        sql_query("DROP TABLE IF EXISTS prevent_sync_pattern;")
+            .execute(conn)
+            .map_err(|_| FSError::DropTable("prevent_sync_pattern"))?;
```
FirelightFlagboy, line 161: save the error into the variant

oxidation/crates/parsec_core_fs/src/storage/manifest_storage.rs
```diff
@@ -0,0 +1,628 @@
...
+    pub fn get_prevent_sync_pattern(&self) -> FSResult<(Regex, bool)> {
+        let conn = &mut *self.conn.lock().expect("Mutex is poisoned");
+        let (re, fully_applied) = prevent_sync_pattern::table
+            .select((
+                prevent_sync_pattern::pattern,
+                prevent_sync_pattern::fully_applied,
+            ))
+            .filter(prevent_sync_pattern::_id.eq(0))
+            .first::<(String, _)>(conn)
+            .map_err(|_| FSError::QueryTable("prevent_sync_pattern: get_prevent_sync_pattern"))?;
```
FirelightFlagboy, line 190: save the error into the variant

oxidation/crates/parsec_core_fs/src/storage/manifest_storage.rs
```diff
@@ -0,0 +1,628 @@
...
+
+        let re = Regex::new(&re)
+            .map_err(|_| FSError::QueryTable("prevent_sync_pattern: corrupted pattern"))?;
```
FirelightFlagboy, line 193: save the error into the variant

oxidation/crates/parsec_core_fs/src/storage/manifest_storage.rs
```diff
@@ -0,0 +1,628 @@
...
+                    .eq(0)
+                    .and(prevent_sync_pattern::pattern.ne(pattern)),
+            ),
+        )
+        .set((
+            prevent_sync_pattern::pattern.eq(pattern),
+            prevent_sync_pattern::fully_applied.eq(false),
+        ))
+        .execute(conn)
+        .map_err(|_| FSError::UpdateTable("prevent_sync_pattern: set_prevent_sync_pattern"))?;
```
FirelightFlagboy, line 217: save the error into the variant

oxidation/crates/parsec_core_fs/src/lib.rs
TimeEngineer: This enum is used in FileTransactions

oxidation/crates/parsec_core_fs/src/storage/manifest_storage.rs
```diff
@@ -0,0 +1,628 @@
...
+        diesel::update(
+            prevent_sync_pattern::table.filter(
+                prevent_sync_pattern::_id
+                    .eq(0)
+                    .and(prevent_sync_pattern::pattern.eq(pattern)),
+            ),
+        )
+        .set(prevent_sync_pattern::fully_applied.eq(true))
+        .execute(conn)
+        .map_err(|_| {
```
FirelightFlagboy, line 239: save the error into the variant

oxidation/crates/parsec_core_fs/src/storage/manifest_storage.rs
```diff
@@ -0,0 +1,628 @@
...
+        // TODO: How to improve ?
+        // It is difficult to build a raw sql query with bind in a for loop
+        // Another solution is to query all data then insert
+        let conn = &mut *self.conn.lock().expect("Mutex is poisoned");
+        for (id, version) in changed_vlobs {
+            sql_query("UPDATE vlobs SET remote_version = ? WHERE vlob_id = ?;")
+                .bind::<diesel::sql_types::Integer, _>(version)
+                .bind::<diesel::sql_types::Binary, _>((**id).as_ref())
+                .execute(conn)
+                .map_err(|_| FSError::UpdateTable("vlobs: update_realm_checkpoint"))?;
```
FirelightFlagboy, line 277: save the error into the variant

oxidation/crates/parsec_core_fs/src/storage/manifest_storage.rs
```diff
@@ -0,0 +1,628 @@
...
+        }
+
+        diesel::insert_into(realm_checkpoint::table)
+            .values(&new_realm_checkpoint)
+            .on_conflict(realm_checkpoint::_id)
+            .do_update()
+            .set(&new_realm_checkpoint)
+            .execute(conn)
+            .map_err(|_| FSError::InsertTable("realm_checkpoint: update_realm_checkpoint"))?;
```
FirelightFlagboy, line 286: save the error into the variant

oxidation/crates/parsec_core_fs/src/storage/manifest_storage.rs
```diff
@@ -0,0 +1,628 @@
...
+                vlobs::base_version,
+                vlobs::remote_version,
+            ))
+            .filter(
+                vlobs::need_sync
+                    .eq(true)
+                    .or(vlobs::base_version.ne(vlobs::remote_version)),
+            )
+            .load::<(Vec<u8>, bool, i32, i32)>(conn)
+            .map_err(|_| FSError::QueryTable("vlobs: get_need_sync_entries"))?
```
FirelightFlagboy, line 321: save the error into the variant

oxidation/crates/parsec_core_fs/src/storage/manifest_storage.rs
```diff
@@ -0,0 +1,628 @@
...
+            return Ok(manifest.clone());
+        }
+
+        // Look into the database
+        let conn = &mut *self.conn.lock().expect("Mutex is poisoned");
+        let manifest = vlobs::table
+            .select(vlobs::blob)
+            .filter(vlobs::vlob_id.eq((*entry_id).as_ref()))
+            .first::<Vec<u8>>(conn)
+            .map_err(|_| FSError::QueryTable("vlobs: get_manifest"))?;
```
FirelightFlagboy, line 354: save the error into the variant

oxidation/crates/parsec_core_fs/src/storage/manifest_storage.rs
```diff
@@ -0,0 +1,628 @@
...
+
+        let manifest = LocalManifest::decrypt_and_load(&manifest, &self.local_symkey)
+            .map_err(|_| FSError::Crypto(CryptoError::Decryption))?;
```
FirelightFlagboy, line 357: save the error into the variant

oxidation/crates/parsec_core_fs/src/storage/manifest_storage.rs
```diff
@@ -0,0 +1,628 @@
...
+                        )
+                    )")
+                    .bind::<diesel::sql_types::Binary, _>(vlob_id)
+                    .bind::<diesel::sql_types::Binary, _>(ciphered)
+                    .bind::<diesel::sql_types::Bool, _>(manifest.need_sync())
+                    .bind::<diesel::sql_types::Integer, _>(manifest.base_version() as i32)
+                    .bind::<diesel::sql_types::Integer, _>(manifest.base_version() as i32)
+                    .bind::<diesel::sql_types::Binary, _>(vlob_id)
+                    .execute(conn)
+                    .map_err(|_| FSError::InsertTable("vlobs: _ensure_manifest_persistent"))?;
```
FirelightFlagboy, line 450: save the error into the variant

tests/common.py
TimeEngineer: I tried to run the tests in parallels :sweat_smile:

oxidation/crates/parsec_core_fs/src/storage/manifest_storage.rs
```diff
@@ -0,0 +1,628 @@
...
+
+                for pending_chunk_ids_chunk in pending_chunk_ids.chunks(SQLITE_MAX_VARIABLE_NUMBER)
+                {
+                    diesel::delete(
+                        chunks::table.filter(chunks::chunk_id.eq_any(pending_chunk_ids_chunk)),
+                    )
+                    .execute(conn)
+                    .map_err(|_| FSError::DeleteTable("chunks: clear_manifest"))?;
```
FirelightFlagboy, line 458: save the error into the variant

oxidation/crates/parsec_core_fs/src/storage/manifest_storage.rs
```diff
@@ -0,0 +1,628 @@
...
+            .lock()
+            .expect("Mutex is poisoned")
+            .remove(&entry_id)
+            .is_some();
+
+        // Remove from local database
+        let conn = &mut *self.conn.lock().expect("Mutex is poisoned");
+        let deleted = diesel::delete(vlobs::table.filter(vlobs::vlob_id.eq((*entry_id).as_ref())))
+            .execute(conn)
+            .map_err(|_| FSError::DeleteTable("vlobs: clear_manifest"))?
```
FirelightFlagboy, line 498: save the error into the variant

oxidation/crates/parsec_core_fs/src/storage/manifest_storage.rs
```diff
@@ -0,0 +1,628 @@
...
+                .iter()
+                .map(ChunkOrBlockID::as_ref)
+                .collect::<Vec<_>>();
+
+            for pending_chunk_ids_chunk in pending_chunk_ids.chunks(SQLITE_MAX_VARIABLE_NUMBER) {
+                diesel::delete(
+                    chunks::table.filter(chunks::chunk_id.eq_any(pending_chunk_ids_chunk)),
+                )
+                .execute(conn)
+                .map_err(|_| FSError::DeleteTable("chunks: clear_manifest"))?;
```
FirelightFlagboy, line 517: save the error into the variant

tests/core/gui/users_widget/test_greet.py
```diff
@@ -603,7 +603,7 @@ async def cancelled_step_5_provide_claim_info(self):
             # expected_message = translate("TEXT_GREET_USER_GET_REQUESTS_ERROR")
             await self._cancel_invitation()
 
-            await aqtbot.wait_until(partial(self._greet_restart, expected_message))
+            await aqtbot.wait_until(partial(self._greet_restart, expected_message), timeout=2000)
```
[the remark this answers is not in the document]
TimeEngineer: Because it fails on my machine, even with python version

FirelightFlagboy (changes requested):
Smaller PR Please, It's hard to read

tests/conftest.py
TimeEngineer: It is used to skip tests which can't be run with rust binding

oxidation/crates/parsec_core_fs/migrations/2022-04-26-103106_chunks/down.sql
TimeEngineer: I use the migrations to generate the schema.rs file, that's how diesel work

oxidation/crates/parsec_core_fs/src/storage/chunk_storage.rs
```diff
@@ -0,0 +1,404 @@
...
+use super::local_database::{SqliteConn, SQLITE_MAX_VARIABLE_NUMBER};
+use crate::error::{FSError, FSResult};
+use crate::extensions::coalesce_total_size;
+use crate::schema::chunks;
+
+#[derive(Insertable, AsChangeset)]
+#[diesel(table_name = chunks)]
+struct NewChunk<'a> {
+    pub chunk_id: &'a [u8],
+    pub size: i32,
```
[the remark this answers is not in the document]
TimeEngineer: It's file descriptor, most of C system type is int

oxidation/crates/parsec_core_fs/src/extensions.rs
```diff
@@ -0,0 +1,27 @@
...
+
+use diesel::backend::Backend;
+use diesel::expression::ValidGrouping;
+use diesel::query_builder::{AstPass, QueryFragment};
+use diesel::sql_types::BigInt;
+use diesel::{Expression, QueryResult};
+
+pub fn coalesce_total_size() -> CoalesceTotalSize {
+    CoalesceTotalSize
+}
```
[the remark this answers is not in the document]
TimeEngineer: This is diesel example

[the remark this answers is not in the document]
TimeEngineer: https://diesel.rs/guides/extending-diesel.html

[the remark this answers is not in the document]
FirelightFlagboy: I would prefer `CoalesceTotalSize::default()`

oxidation/crates/parsec_core_fs/src/error.rs
```diff
@@ -0,0 +1,66 @@
...
+    #[error("ConnectionError")]
+    Connection,
+
+    #[error("CreateTableError {0}")]
+    CreateTable(&'static str),
+
+    #[error("CreateDirError")]
+    CreateDir,
+
+    #[error("CryptoError: {0}")]
```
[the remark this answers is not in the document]
TimeEngineer: This error comes from below, it may have more information

oxidation/crates/parsec_core_fs/src/storage/chunk_storage.rs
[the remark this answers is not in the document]
FirelightFlagboy: What do you mean by 

> It's a file descriptor 

? the field's name is `size` not `file_descritor` 

> most of C system type is int

So, can't we cast the value to `i32` only when it's required ?

oxidation/crates/parsec_core_fs/src/storage/manifest_storage.rs
TimeEngineer: LocalManifest error is deprecated

oxidation/crates/parsec_core_fs/src/storage/manifest_storage.rs
TimeEngineer: Done :+1:

oxidation/crates/parsec_core_fs/src/storage/mod.rs
touilleMan: then you should add a comment ;-)

oxidation/libparsec_python/src/lib.rs
```diff
@@ -180,5 +182,9 @@ fn _libparsec(_py: Python, m: &PyModule) -> PyResult<()> {
     m.add_class::<certif::DeviceCertificate>()?;
     m.add_class::<trustchain::TrustchainContext>()?;
     m.add_function(wrap_pyfunction!(time::freeze_time, m)?)?;
+    // LocalDevice
+    m.add_class::<local_device::LocalDevice>()?;
+    // Storage
```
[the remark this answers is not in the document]
TimeEngineer: Done :+1:

oxidation/libparsec_python/src/addrs.rs
TimeEngineer: useless, because you can't access it anyway

oxidation/crates/parsec_core_fs/src/storage/chunk_storage.rs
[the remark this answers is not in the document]
TimeEngineer: Oh something went wrong in my mind, diesel doesn't accept u32 type

oxidation/crates/parsec_core_fs/src/error.rs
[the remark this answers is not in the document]
FirelightFlagboy: ok

oxidation/crates/parsec_core_fs/src/extensions.rs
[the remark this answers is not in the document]
FirelightFlagboy: thanks

oxidation/crates/parsec_core_fs/src/storage/chunk_storage.rs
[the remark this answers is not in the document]
FirelightFlagboy: diesel-rs/diesel#1907 :(
So we can't use usigned value

oxidation/crates/parsec_core_fs/src/storage/mod.rs
TimeEngineer: Ok !

oxidation/crates/parsec_core_fs/src/error.rs
```diff
@@ -0,0 +1,66 @@
...
+    #[error("QueryTableError: {0}")]
+    QueryTable(String),
+
+    #[error("PermissionError")]
+    Permission,
+
+    #[error("PoolError")]
+    Pool,
+
+    #[error("UpdateTableError {0}")]
```
FirelightFlagboy, line 50:
```suggestion
    #[error("UpdateTableError: {0}")]
```
missing `:` ?

oxidation/crates/parsec_core_fs/src/storage/manifest_storage.rs
FirelightFlagboy: Why it's deprecated ?

oxidation/crates/parsec_core_fs/src/workspacefs/file_transactions.rs
```diff
@@ -0,0 +1,33 @@
...
+    /// The corresponding file is locked while performing the change (i.e. between
+    /// the reading and writing of the corresponding manifest) in order to avoid
+    /// race conditions and data corruption.
+
+    /// The table below lists the effects of the 6 file transactions:
+    /// - close    -> remove file descriptor from local storage
+    /// - write    -> affects file content and possibly file size
+    /// - truncate -> affects file size and possibly file content
+    /// - read     -> no side effect
+    /// - flush    -> no-op
```
FirelightFlagboy:
```suggestion
/// A stateless class to centralize all file transactions.
///
/// The actual state is stored in the local storage and file transactions
/// have access to the remote loader to download missing resources.
///
/// The exposed transactions all take a file descriptor as first argument.
/// The file descriptors correspond to an entry id which points to a file
/// on the file system (i.e. a file manifest).
///
/// The corresponding file is locked while performing the change (i.e. between
/// the reading and writing of the corresponding manifest) in order to avoid
/// race conditions and data corruption.
///
/// The table below lists the effects of the 6 file transactions:
/// - close    -> remove file descriptor from local storage
/// - write    -> affects file content and possibly file size
/// - truncate -> affects file size and possibly file content
/// - read     -> no side effect
/// - flush    -> no-op
pub struct FileTransactions {
```
Move the doc comment above `pub struct FileTransactions` if the comment here to document concern the struct.
Unless the comment if for the field `workspace_id`

oxidation/crates/parsec_core_fs/src/remote_loader.rs
```diff
@@ -0,0 +1 @@
+// Parsec Cloud (https://parsec.cloud) Copyright (c) BSLv1.1 (eventually AGPLv3) 2016-2021 Scille SAS
```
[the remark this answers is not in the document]
TimeEngineer: Done :+1:

TimeEngineer:
> Why it's deprecated ?
Because it's a static str, not an error

oxidation/crates/parsec_core_fs/src/storage/chunk_storage.rs
```diff
@@ -0,0 +1,405 @@
...
+            res.append(
+                &mut chunks::table
+                    .select(chunks::chunk_id)
+                    .filter(chunks::chunk_id.eq_any(bytes_id_list_chunk))
+                    .load::<Vec<u8>>(conn)
+                    .map_err(|e| FSError::QueryTable(format!("chunks: get_local_chunk_ids {e}")))?
+                    .into_iter()
+                    .map(|chunk_id| {
+                        <[u8; 16]>::try_from(&chunk_id[..]).map_err(|e| {
+                            FSError::QueryTable(format!("chunks: corrupted chunk_id {e}"))
```
FirelightFlagboy: The error returned by the `try_from` will be `chunk_id`

FirelightFlagboy (changes requested):

oxidation/crates/parsec_core_fs/src/workspacefs/file_transactions.rs
TimeEngineer: Good catch

oxidation/crates/parsec_core_fs/src/storage/chunk_storage.rs
TimeEngineer: :man_facepalming: :+1:

TimeEngineer: Oh anyway it's nice to have

oxidation/libparsec/crates/core_fs/src/storage/manifest_storage.rs
```diff
@@ -0,0 +1,641 @@
...
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use parsec_api_crypto::HashDigest;
+    use parsec_api_types::{BlockAccess, Blocksize, DateTime, DeviceID, FileManifest};
+    use parsec_client_types::{Chunk, LocalFileManifest};
+
+    use super::super::local_database::SqlitePool;
```
FirelightFlagboy, line 548: super super

oxidation/libparsec/crates/core_fs/src/storage/mod.rs
```diff
@@ -0,0 +1,13 @@
+// Parsec Cloud (https://parsec.cloud) Copyright (c) BSLv1.1 (eventually AGPLv3) 2016-2021 Scille SAS
+
+mod chunk_storage;
+mod local_database;
+mod manifest_storage;
+// Not used for the moment
```
FirelightFlagboy, line 6:
```suggestion
/// Not used for the moment
```

oxidation/libparsec/crates/core_fs/src/workspacefs/file_transactions.rs
```diff
@@ -0,0 +1,33 @@
...
+    /// The corresponding file is locked while performing the change (i.e. between
+    /// the reading and writing of the corresponding manifest) in order to avoid
+    /// race conditions and data corruption.
+
+    /// The table below lists the effects of the 6 file transactions:
+    /// - close    -> remove file descriptor from local storage
+    /// - write    -> affects file content and possibly file size
+    /// - truncate -> affects file size and possibly file content
+    /// - read     -> no side effect
+    /// - flush    -> no-op
```
FirelightFlagboy: This comment is for `FileTransactions` or `workspace_id` ?

tests/core/gui/test_file_history.py
```diff
@@ -13,6 +14,7 @@ def catch_file_history_widget(widget_catcher_factory):
 
 @pytest.mark.gui
 @pytest.mark.trio
+@customize_fixtures(real_data_storage=True)
```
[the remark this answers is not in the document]
TimeEngineer: Could you be more explicit ?

tests/core/test_invite.py
TimeEngineer: Changed to real_clock_timeout

tests/conftest.py
TimeEngineer: Timestamped tests for example

oxidation/libparsec/crates/core_fs/src/storage/mod.rs
TimeEngineer: No

oxidation/libparsec/crates/core_fs/src/workspacefs/file_transactions.rs
TimeEngineer: File Transaction indeed

oxidation/libparsec/crates/core_fs/src/storage/chunk_storage.rs
```diff
@@ -0,0 +1,405 @@
...
+            res.append(
+                &mut chunks::table
+                    .select(chunks::chunk_id)
+                    .filter(chunks::chunk_id.eq_any(bytes_id_list_chunk))
+                    .load::<Vec<u8>>(conn)
+                    .map_err(|e| FSError::QueryTable(format!("chunks: get_local_chunk_ids {e}")))?
+                    .into_iter()
+                    .map(|chunk_id| {
+                        <[u8; 16]>::try_from(&chunk_id[..]).map_err(|e| {
+                            FSError::QueryTable(format!("chunks: corrupted chunk_id {e}"))
```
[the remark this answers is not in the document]
TimeEngineer: Nice to have

oxidation/libparsec/crates/core_fs/src/storage/chunk_storage.rs
```diff
@@ -0,0 +1,405 @@
...
+            res.append(
+                &mut chunks::table
+                    .select(chunks::chunk_id)
+                    .filter(chunks::chunk_id.eq_any(bytes_id_list_chunk))
+                    .load::<Vec<u8>>(conn)
+                    .map_err(|e| FSError::QueryTable(format!("chunks: get_local_chunk_ids {e}")))?
+                    .into_iter()
+                    .map(|chunk_id| {
+                        <[u8; 16]>::try_from(&chunk_id[..]).map_err(|_| {
+                            FSError::QueryTable(format!("chunks: corrupted chunk_id {chunk_id:?}"))
```
FirelightFlagboy, line 101: That wasn't what I mean, do we want to display the raw value of `chunk_id` ?

oxidation/libparsec/crates/core_fs/src/storage/chunk_storage.rs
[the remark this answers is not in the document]
FirelightFlagboy: Ok

oxidation/libparsec/crates/core_fs/src/storage/chunk_storage.rs
FirelightFlagboy: After reading the previous answer, we want to keep the `chunk_id` in the error message

oxidation/libparsec/crates/core_fs/src/storage/mod.rs
FirelightFlagboy: :(

oxidation/libparsec/crates/core_fs/src/workspacefs/file_transactions.rs
FirelightFlagboy: so move the doc comment above `pub struct FileTransactions`

FirelightFlagboy (approved):

oxidation/crates/parsec_core_fs/migrations/2022-04-26-103106_chunks/up.sql
TimeEngineer: Is it important ?

tests/core/gui/test_file_history.py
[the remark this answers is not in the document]
touilleMan: we want to keep using the in-memory mock for the pure python test, this is not possible if we put `@customize_fixtures(real_data_storage=True) ` in each test

So instead we should modify the `persistent_mockup` fixture so that it doesn't do anything if it detects the code is currently running with oxidation stuff

.pre-commit-config.yaml
```diff
@@ -7,6 +7,8 @@ repos:
     rev: v2.2.3  # Use the ref you want to point at
     hooks:
     - id: mixed-line-ending
+      args:
```
touilleMan, line 10: tl;dr: remove this we handle it in the CI

We patch this config file in the CI so that we won't have mixed ending merged. However we don't enforce that in this config file given it causes issue on Windows: On Windows git can be configured to checkout file with \r\n, but to convert them back into \n when doing the commit, however this doesn't work with pre-commit given it is run on the file on the disk.

oxidation/crates/parsec_api_types/src/id.rs
```diff
@@ -267,6 +267,9 @@ impl ToString for UserProfile {
     }
 }
 
+#[derive(Debug, Clone, Copy, PartialEq, Eq, Hash)]
+pub struct FileDescriptor(pub i32);
```
touilleMan, line 271: signed value ? what negative and 0 stand for ? (yes I'm asking for more comments ! ^^)

tests/conftest.py
touilleMan: We should instead introduce a  `is_oxidized` function that return true if we currently run with the oxidation stuff, and use this to skip tests on the fly:
- by modifying the fixtures when possible (cf. what I was saying about the storage mock fixture)
- otherwise by using `@pytest.mark.skipif(is_oxidized(), reason="Test not compatible with Oxidaziton extension")` decorator

tests/common.py
touilleMan: There is definitely something fishy here, you should investigate what is taking so long

oxidation/crates/parsec_core_fs/migrations/2022-04-26-103106_chunks/up.sql
touilleMan: not having a date is valid, having a date is valid. However having an incorrect date is dangerous since we may end up using it by mistake (typically if all the subsequent migrations have a correct date we will forget this one is a special snowflake)

So I would say it is better not to have date at all (since we use a number prefix in the backend postgresql migrations it would be better to be consistent), but is this possible with diesel ?

.pre-commit-config.yaml
TimeEngineer: How can I tell to my pre-commit ?

oxidation/crates/parsec_core_fs/src/lib.rs
touilleMan: Oh yes, this to know how to rename a file conflict: `foo.txt` -> `foo (conflict with Bar).txt` in english and `foo (conflit avec Bar).txt` in french

I don't like this stuff (because it brings translation issues in an unrelated codebase), but @vxgmichel said it's a killer feature so I guess it's here to stay 😄

touilleMan: You should add a `// Translations used to for File/Folder conflict renaming` comment

tests/core/test_sync_monitor_stateful.py
```diff
@@ -48,6 +48,8 @@ def recursive_compare_fs_dumps(alice_dump, bob_dump, ignore_need_sync=False):
 
 
 @pytest.mark.slow
+# This test runs into infinite loop when it uses real data storage
```
touilleMan, line 51: Are you sure this is an infinite loop ? hypothesis tests are slow, and can become very very slow if the code behind it is itself slow ^^

touilleMan (changes requested):

oxidation/crates/parsec_core_fs/migrations/2022-04-26-103106_chunks/up.sql
TimeEngineer: Of course it is possible ^^

.pre-commit-config.yaml
TimeEngineer: nvm

oxidation/crates/parsec_api_types/src/id.rs
FirelightFlagboy: `0` is a valid file descriptor :p.
but a negative one ... don't know

FirelightFlagboy (approved):

parsec/__init__.py
```diff
@@ -6,3 +6,10 @@
 import parsec.utils  # noqa
 
 __all__ = ("__version__",)
+
+try:
+    import libparsec  # noqa
+except:
```
touilleMan, line 12: `except ImportError`

oxidation/libparsec/crates/core_fs/src/schema.rs.patch
```diff
@@ -0,0 +1,27 @@
+diff --git a/oxidation/libparsec/crates/core_fs/src/schema.rs b/oxidation/libparsec/crates/core_fs/src/schema.rs
```
FirelightFlagboy, line 1: Why do we use a patch here ?

oxidation/libparsec/crates/core_fs/src/storage/chunk_storage.rs
```diff
@@ -0,0 +1,409 @@
...
+    fn get_local_chunk_ids(&self, chunk_ids: &[ChunkID]) -> FSResult<Vec<ChunkID>> {
+        let bytes_id_list = chunk_ids
+            .iter()
+            .map(|chunk_id| (**chunk_id).as_ref())
+            .collect::<Vec<_>>();
+
+        let mut res = Vec::with_capacity(chunk_ids.len());
+
+        let conn = &mut *self.conn().lock().expect("Mutex is poisoned");
+        for bytes_id_list_chunk in bytes_id_list.chunks(SQLITE_MAX_VARIABLE_NUMBER) {
```
touilleMan, line 91: Can you add a comment to explain what's going on here: the fact we are doing a query in a for loop seems like poor code where it is something we have really think about !

You can also mention:
```
The number of loop iteration is expected to be rather low: typically considering 4ko per chunk (i.e. the size of the buffer the Linux Kernel send us through Fuse), each query could perform ~4mo of data.

If performance ever becomes an issue, we could further optimize this by having the caller filter the chunks: a block containing multiple chunk means those correspond to local modification and hence can be filtered out. With this we would only provide the chunks corresponding to a synced block which are 512ko big, so each query performs up to 512Mo !
```

oxidation/libparsec/crates/core_fs/src/storage/chunk_storage.rs
```diff
@@ -0,0 +1,409 @@
...
+                accessed_on REAL, -- Timestamp
+                data BLOB NOT NULL
+            );",
+        )
+        .execute(conn)
+        .map_err(|e| FSError::CreateTable(format!("chunks {e}")))?;
+        Ok(())
+    }
+
+    #[cfg(test)]
```
touilleMan, line 45: 👍 for the cfg test

oxidation/libparsec/crates/core_fs/src/storage/chunk_storage.rs
```diff
@@ -0,0 +1,409 @@
...
+use crate::extensions::CoalesceTotalSize;
+use crate::schema::chunks;
+
+#[derive(Insertable, AsChangeset)]
+#[diesel(table_name = chunks)]
+struct NewChunk<'a> {
+    pub chunk_id: &'a [u8],
+    pub size: i32,
+    pub offline: bool,
+    pub accessed_on: Option<f32>,
```
touilleMan, line 21: could we use DateTime here instead ?

If not, could we use `DateTime` here if `DateTime` would implement `From<f64>` ? 😃 

Right now this is not super useful given there is few place where we do the conversion by hand, but it may become handy later on ^^

touilleMan, line 21: And we want to use `f64` here

oxidation/libparsec/crates/core_fs/src/storage/chunk_storage.rs
```diff
@@ -0,0 +1,409 @@
...
+use super::local_database::{SqliteConn, SQLITE_MAX_VARIABLE_NUMBER};
+use crate::error::{FSError, FSResult};
+use crate::extensions::CoalesceTotalSize;
+use crate::schema::chunks;
+
+#[derive(Insertable, AsChangeset)]
+#[diesel(table_name = chunks)]
+struct NewChunk<'a> {
+    pub chunk_id: &'a [u8],
+    pub size: i32,
```
touilleMan, line 19: [integer as stored as signed, up to 64bits](https://www.sqlite.org/datatype3.html), so this should be `i64` in theory.

However I think it would be better if we set it as unsigned, and even better as `NonZeroU64` (which is not strictly true, it should be more like `NonZeroU63` but I don't think Rust support this type and we should be pretty safe from hitting such upper limit)

touilleMan: Another approach is to stick with `NonZeroU32` which sqlite can store (unlike u64), and given chunk size is much murch smaller (typically ~512ko vs 2³²)

@vxgmichel @FirelightFlagboy what do you think ?

oxidation/libparsec/crates/core_fs/src/storage/manifest_storage.rs
```diff
@@ -0,0 +1,645 @@
...
+            Self::BlockID(id) => (**id).as_ref(),
+        }
+    }
+}
+
+#[derive(Insertable)]
+#[diesel(table_name = vlobs)]
+struct NewVlob<'a> {
+    vlob_id: &'a [u8],
+    base_version: i32,
```
touilleMan, line 38: why i32 and not i64 ? besides version is defined as u32 in manifest.rs

oxidation/libparsec/crates/core_fs/src/storage/manifest_storage.rs
```diff
@@ -0,0 +1,645 @@
...
+    remote_version: i32,
+    need_sync: bool,
+    blob: &'a [u8],
+}
+
+#[derive(Insertable, AsChangeset)]
+#[diesel(table_name = realm_checkpoint)]
+struct NewRealmCheckpoint {
+    _id: i32,
```
touilleMan, line 47: i64 ?

oxidation/libparsec/crates/core_fs/src/storage/manifest_storage.rs
```diff
@@ -0,0 +1,645 @@
...
+            cache: Mutex::new(HashMap::new()),
+            cache_ahead_of_localdb: Mutex::new(HashMap::new()),
+        };
+        instance.create_db()?;
+        Ok(instance)
+    }
+
+    /// Database initialization
+    pub fn create_db(&self) -> FSResult<()> {
+        let conn = &mut *self.conn.lock().expect("Mutex is poisoned");
```
touilleMan, line 101: how transactions are handled here ? (in the python code all those sql statements are executed in a single transaction)

touilleMan (changes requested):

oxidation/crates/parsec_api_types/src/id.rs
TimeEngineer: Do you mean I have to search the information ?

oxidation/libparsec/crates/core_fs/src/storage/manifest_storage.rs
```diff
@@ -0,0 +1,645 @@
...
+                        max(
+                            ?,
+                            IFNULL((SELECT remote_version FROM vlobs WHERE vlob_id=?), 0)
+                        )
+                    )")
+                    .bind::<diesel::sql_types::Binary, _>(vlob_id)
+                    .bind::<diesel::sql_types::Binary, _>(ciphered)
+                    .bind::<diesel::sql_types::Bool, _>(manifest.need_sync())
+                    .bind::<diesel::sql_types::Integer, _>(manifest.base_version() as i32)
+                    .bind::<diesel::sql_types::Integer, _>(manifest.base_version() as i32)
```
TimeEngineer, line 459: ohoh

oxidation/libparsec/crates/core_fs/src/storage/workspace_storage.rs
```diff
@@ -0,0 +1,333 @@
...
+    ) -> FSResult<()> {
+        self.manifest_storage
+            .set_manifest(entry_id, manifest, cache_only, removed_ids)
+    }
+
+    pub fn ensure_manifest_persistent(&self, entry_id: EntryID) -> FSResult<()> {
+        self.manifest_storage.ensure_manifest_persistent(entry_id)
+    }
+
+    #[allow(deprecated)]
```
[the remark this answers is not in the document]
TimeEngineer: Due to his parent

oxidation/libparsec_python/src/storage.rs
```diff
@@ -0,0 +1,351 @@
...
+            block
+        } else {
+            return Err(PyValueError::new_err(
+                "evolve_as_block: invalid input for block",
+            ));
+        };
+
+        py.allow_threads(|| {
+            self.0
+                .set_chunk(chunk_id.0, block)
```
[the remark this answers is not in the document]
TimeEngineer: Good catch

oxidation/libparsec/crates/core_fs/src/storage/chunk_storage.rs
```diff
@@ -0,0 +1,409 @@
...
+
+use parsec_api_crypto::SecretKey;
+use parsec_api_types::{ChunkID, DateTime, DEFAULT_BLOCK_SIZE};
+
+use super::local_database::{SqliteConn, SQLITE_MAX_VARIABLE_NUMBER};
+use crate::error::{FSError, FSResult};
+use crate::extensions::CoalesceTotalSize;
+use crate::schema::chunks;
+
+#[derive(Insertable, AsChangeset)]
```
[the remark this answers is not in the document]
TimeEngineer: If conflict, there is an update

oxidation/libparsec/crates/core_fs/src/storage/chunk_storage.rs
TimeEngineer: Can we improve later ?

oxidation/libparsec/crates/core_fs/src/storage/chunk_storage.rs
FirelightFlagboy: I'm in favor of using unsigned type when the value don't make any sense when negative

oxidation/libparsec/crates/core_fs/src/storage/manifest_storage.rs
TimeEngineer: I decided to run separately in order to handle which table's creation fail since it's called once.
There is batch_execute which perform one transaction

oxidation/libparsec/crates/core_fs/src/storage/chunk_storage.rs
FirelightFlagboy: Yes we can iterate, could you create the issue ?

TimeEngineer: It's not something you will use, it's an interface with Sqlite

TimeEngineer: I warn that even if we have an unsigned integer, the representation will be signed so a little dangerous

touilleMan: The integer size an signed/unsigned is a typical recipe for errors, we must do some thinking about it (and and comments to explain the rational)

But for the moment you can:
- add `TODO: think about type compat with sqlite`
- create an issue about that is the oxidation project

PR #2228, opened 2022-05-05
[GUI] Added spinner and edited error message when claiming a user/device

### LINKED ISSUE #2215: Add spinner for dialogs that requires a connection to the server
In some cases, the connection to the parsec server might take some time to time out, for instance.
https://docs.microsoft.com/en-us/troubleshoot/windows-client/networking/10060-connection-timed-out-with-proxy-server

In that case, the user might get a confusing dialog:
image012

In those cases, we should a spinner to let the user know that the content is currently being loaded.

### LINKED ISSUE #2216: Improve user message when the backend is not available while starting the invitation process
The following message:
```
msgid "TEXT_INVITATION_BACKEND_NOT_AVAILABLE"
msgstr "The connection to the server has been lost. Please restart the process."
```
might be updated to:
```
msgid "TEXT_INVITATION_BACKEND_NOT_AVAILABLE"
msgstr "The connection to the server has been lost. Please check your connection and restart the process."
```
to make it clear to the user that their connection is likely at fault.

Max-7:
Closes #2216 
Closes #2215


vxgmichel (approved):
Tested by adding a `trio.sleep` in the transport, LGTM :+1:

PR #2229, opened 2022-05-05
[GUI] Fixed file link in some instances

### LINKED ISSUE #2223: Blank page after opening a file link with no access when logged out
To replicate:

1.  Be logged out of the organization
2. Click link of file for which we don't have access
3. Log in with password

Which results in an entirely blank page. This doesn't occur if already logged in the org, instead raises an error and loads the widgets correctly.

This _might_ be mac-only.
Capture d’écran 2022-05-03 à 09 15 31

Max-7:
Closes #2223

### DIFF
--- a/newsfragments/2223.bugfix.rst
+++ b/newsfragments/2223.bugfix.rst
@@ -1,0 +1,1 @@
+Fixed clicking on a file link in some instances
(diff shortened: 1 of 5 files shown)


Ironicbay (approved):
LGTM :+1:

PR #2237, opened 2022-05-06
[GUI] Fixed an error message appearing when dragging a file

### LINKED ISSUE #2257: Dragging a file displays an error message

Max-7:
Closes #2257 

An error essage about EntryID not being serializable by pickle appeared when dragging a file. It didn't seem to have any incidence as the file could still be dragged and dropped, but it still better to fix it. We do so by ensuring that the EntryID stored in the widget UserData are converted to string. We could instead try to register EntryID to Qt's meta-object system but this would just be annoying and I don't know how to do it with PyQt.

### DIFF
--- a/newsfragments/2237.bugfix.rst
+++ b/newsfragments/2237.bugfix.rst
@@ -1,0 +1,1 @@
+Fixed an error message when dragging files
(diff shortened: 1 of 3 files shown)


vxgmichel (approved):
LGTM :+1: 

No tests for this?

PR #2239, opened 2022-05-09
[GUI] Handle cases where a user does not provide a certificate or a pin

Max-7:
Handled cases in the GUI where a user cancels the certificate selection or declines to provide the private key pin code. We usually do nothing in those cases instead of displaying an error.

Changes were only made in the pki enrollment and the login process, not for authentication change or device recovery.


vxgmichel (approved):
A couple of comments but LGTM :+1:

PR #2267, opened 2022-05-11
Add http proxy (which PAC !) support

### LINKED ISSUE #2217: Investigate proxy server support
At the moment, users behind proxies might experience this kind of error:
https://docs.microsoft.com/en-us/troubleshoot/windows-client/networking/10060-connection-timed-out-with-proxy-server

We should investigate and see what kind of proxy server support we can provide. A proper error detection and explicit message to explain to users that some forms of proxy are not supported (if any) would also be nice.

touilleMan:
fix #2217


parsec/core/backend_connection/transport.py
```diff
@@ -114,7 +127,280 @@ async def _connect(
...
 
 
-def _upgrade_stream_to_ssl(raw_stream, hostname):
+def cook_basic_auth_header(username: str, password: str) -> str:
+    step1 = f"{username}:{password}".encode("utf8")
+    step2 = b64encode(step1).decode("ascii")
+    return f"Basic {step2}"
+
+
+# Global vars to easily set up proxy from config file or cli params
```
vxgmichel, line 136: Is this a quick hack? It feels like such parameters are typically managed in the configuration and set through CLI params.

parsec/core/backend_connection/transport.py
```diff
@@ -114,7 +127,280 @@ async def _connect(
...
+            logger.warning(
+                "Found multiple proxies info in .PAC proxy config for target, only the first one is going to be used !",
+                target_url=url,
+                proxies=proxies,
+            )
+        # We don't handle multiple proxies so keep the first correct one and pray !
+        for proxy in proxies:
+            if proxy in ("DIRECT", ""):
+                # PAC explicitly told us not to use a proxy
+                return ""
```
vxgmichel, line 207: The special use of `""` as an explicit way to not use proxy feels weird, can we have a dedicated value for it? And maybe something that does not evaluate to `False` in order to not be confused with `None`?

parsec/core/backend_connection/transport.py
```diff
@@ -114,7 +127,280 @@ async def _connect(
...
+        logger.debug("Got .PAC config explictly specifying direct access", target_url=target_url)
+
+    # Fallback on direct proxy url config in environ variables & Windows registers table
+    if proxy_url is None:  # `proxy_url == ""` explicitly indicates no proxy should be use
+        proxy_url = _get_proxy_from_os_config(hostname, use_ssl)
+        if proxy_url:
+            logger.debug("Got proxy", proxy_url=proxy_url, target_url=target_url)
+
+    if proxy_url in (None, ""):
+        return None
```
vxgmichel, line 273: It feels like related things are too far apart. 

Shouldn't the following block from `_connect` replace the different `return None` statement?
```python
        logger.debug("Using direct (no proxy) connection", hostname=hostname, port=port)
        try:
            stream = await trio.open_tcp_stream(hostname, port)

        except OSError as exc:
            logger.debug(
                "Impossible to connect to backend", hostname=hostname, port=port, exc_info=exc
            )
            raise BackendNotAvailable(exc) from exc
```
This way the semantics of this function becomes `open_connection_using_proxy_config`.

vxgmichel (changes requested):
A couple of comments and change requests but LGTM otherwise :+1:

parsec/core/backend_connection/transport.py
touilleMan: removed, this was needed to provide pramas in the cli, but we can work with env var only instead

vxgmichel (changes requested):

parsec/core/backend_connection/transport.py
```diff
@@ -501,42 +224,21 @@ async def http_request(
 
     def _target():
         request = urllib.request.Request(url, data=data, headers=headers, method=method)
+        handlers = []
+        proxy_url = blocking_io_get_proxy(target_url=url, hostname=request.host)
+        if proxy_url:
+            logger.debug("Using proxy to connect", proxy_url=proxy_url, target_url=url)
+            handlers.append(urllib.request.ProxyHandler(proxies={request.type: proxy_url}))
```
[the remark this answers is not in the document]
vxgmichel: See https://docs.python.org/fr/3/library/urllib.request.html#urllib.request.ProxyHandler:
> To disable autodetected proxy pass an empty dictionary

parsec/core/backend_connection/transport.py
touilleMan: you you mean the returned `""` value vs `None`, I've made this more explicit by adding docstring in `_get_proxy_*` and having `get_proxy` getting rid of it (hence only `get_proxy` call `_get_proxy_*` and have to deal with the `""` special case)

parsec/core/backend_connection/transport.py
touilleMan: yup, all the proxy logic is now contained within `get_proxy` and `maybe_connect_through_proxy` (that internally uses `get_proxy` and wrap the given stream the same way adding ssl support works)

parsec/core/backend_connection/transport.py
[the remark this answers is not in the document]
touilleMan: Looks pretty similar to the else clause... is there something I'm missing ? :/

parsec/core/backend_connection/transport.py
vxgmichel: What I meant is that the semantics of:
```python
async def maybe_connect_through_proxy(
    hostname: str, port: int, use_ssl: bool
) -> Optional[trio.abc.Stream]:
```
feels a bit weird and that the following would make more sense:
```python
async def connect_maybe_through_proxy(
    hostname: str, port: int, use_ssl: bool
) -> trio.abc.Stream:
```

parsec/core/backend_connection/proxy.py
```diff
@@ -0,0 +1,294 @@
...
+def _get_proxy_from_environ_or_os_config(url: str, hostname: str) -> Optional[str]:
+    """
+    Returns:
+        proxy url to use
+        "" if no proxy should be used
+        `None` if no config was found
+    """
+    # Retrieve proxies from environ vars then (depending on platform) os configuration
+    if proxy_bypass(hostname):
+        return None
```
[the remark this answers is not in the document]
touilleMan: good point !

parsec/core/backend_connection/proxy.py
```diff
@@ -0,0 +1,294 @@
...
+        logger.debug("Got .PAC config explictly specifying direct access", target_url=target_url)
+
+    # Fallback on direct proxy url config in environ variables & Windows registers table
+    if proxy_url is None:  # `proxy_url == ""` explicitly indicates no proxy should be use
+        proxy_url = _get_proxy_from_environ_or_os_config(target_url, hostname)
+
+    if proxy_url in (None, ""):
+        return None
+
+    return proxy_url
```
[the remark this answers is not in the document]
touilleMan: yeah, this is much better !

parsec/core/backend_connection/transport.py
touilleMan: This is to mimic the wrapping system for SSL support.
On top of that I prefer having the proxy concepts hidden behind the `_connect` generic concept given it's a ugly detail we don't want to talk too much about 😄

parsec/core/backend_connection/proxy.py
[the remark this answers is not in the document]
touilleMan: done

parsec/core/backend_connection/proxy.py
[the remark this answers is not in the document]
touilleMan: done

vxgmichel (approved):
LGTM :+1:

PR #2270, opened 2022-05-12
Yolo/proposal bug report template

FirelightFlagboy:


.github/ISSUE_TEMPLATE/bug-report.yml
```diff
@@ -0,0 +1,51 @@
+name: '🐛 | Bug Report'
+description: >
+  File a bug report
+title: '[🐛 | Bug]: '
+labels: [bug, triage]
+assignees:
+  - touilleMan
```
touilleMan, line 7: but why meeeee ? :'(

.github/ISSUE_TEMPLATE/bug-report.yml
```diff
@@ -0,0 +1,51 @@
...
+      placeholder: 'Tell us what you see !'
+    validations:
+      required: true
+  - type: dropdown
+    id: version
+    attributes:
+      label: Version
+      description: 'What version of our software are you running ?'
+      options:
+        - 2.8.3
```
touilleMan, line 23: Why not a free field instead ? this is going to be a pain to keep this field up to date.

FirelightFlagboy: It would be easier to fill for the user but harder to maintain.

I can add a field that say `version not listed` or something

.github/ISSUE_TEMPLATE/bug-report.yml
FirelightFlagboy: because ?

FirelightFlagboy: So do you want to include another person to be able to do the triage of the bug ?

FirelightFlagboy: remove the field assignees

touilleMan (approved):

PR #2273, opened 2022-05-12
Feat/2271 rust oxidationlibparseccratesplatform async crate mvp

FirelightFlagboy:


newsfragments/2223.bugfix.rst
```diff
@@ -0,0 +1 @@
+Fixed clicking on a file link in some instances
```
touilleMan, line 1: This newsfragments don't seem related to this PR...

FirelightFlagboy: hum, I'll remove them

FirelightFlagboy:
wololo, I've made some cleanup @touilleMan

oxidation/crates/platform-async/Cargo.toml
```diff
@@ -0,0 +1,26 @@
+[package]
+name = "platform-async"
```
touilleMan, line 2: `libparsec_platform_async` (the idea is it's easy to see the correspondance with the path `oxidation/libparsec/crates/platform_async` while still behind explicit about the fact this is not a 3rd party crate)

And we use underscore instead of dash (because it's what is used in the rust code when doing `use libparsec_platform_async::*`, so better stay consistent)

oxidation/crates/platform-async/Cargo.toml
```diff
@@ -0,0 +1,26 @@
...
+
+[dependencies]
+flume = "0.10.12"
+futures = "0.3.21"
+futures-lite = { version = "1.12.0", default-features = false }
+
+[dev-dependencies]
+log = "0.4.17"
+
+[target.'cfg(target_arch = "wasm32")'.dependencies]
```
touilleMan, line 14: so far we use feature `web` vs `native` instead of relying on target_arch (given `wasm32` may refer to wasi, or to a node module)

I guess the advantage of the target_arch, though less consistent, is to avoid having to deal with feature and default_features in the crates but I have the feeling this is not a big deal given we have to set the crates once and for all with default_features=false for the dependencies and we're good ;-)

oxidation/crates/platform-async/src/task.rs
```diff
@@ -0,0 +1,17 @@
+// Parsec Cloud (https://parsec.cloud) Copyright (c) BSLv1.1 (eventually AGPLv3) 2016-2021 Scille SAS
+
+use std::future::Future;
+
+pub trait Taskable<T>: Future<Output = T> {
```
touilleMan: What's the need for this trait ?

I guess this is to abstract the fact we have 2 implementations of Task (desktop vs web). If this is the fact I think we are better without any trait at all.
This is the approach used for the crate `parsec_api_crypto` (that have rustcrypto and libsodium implementations).

This has be benefit of simplifying the code (no need to use the trait in the caller code, simpler types to work with, async traits add complexity)
The only drawback is it makes impossible to use both implementations at the same time, but this is not a concern in here ^^

oxidation/crates/platform-async/Cargo.toml
FirelightFlagboy: it don't have `libparsec` in the path :p 

for the crate name, I have a lot of counter example that use the kebab case for it

oxidation/crates/platform-async/src/platform.rs
```diff
@@ -0,0 +1,8 @@
+// Parsec Cloud (https://parsec.cloud) Copyright (c) BSLv1.1 (eventually AGPLv3) 2016-2021 Scille SAS
```
touilleMan: we use `<module_name>/mod.rs` style in the codebase

oxidation/crates/platform-async/src/platform/default.rs
```diff
@@ -0,0 +1,149 @@
+// Parsec Cloud (https://parsec.cloud) Copyright (c) BSLv1.1 (eventually AGPLv3) 2016-2021 Scille SAS
```
touilleMan: maybe call it "native.rs" or "tokio.rs" instead ?

oxidation/crates/platform-async/src/platform/default.rs
```diff
@@ -0,0 +1,149 @@
...
+    F: Future + Send + 'static,
+    F::Output: Send + 'static,
+{
+    let shared_state = SharedState::default();
+    let shared_state = Arc::new(Mutex::new(shared_state));
+
+    let task = Task::new(shared_state.clone());
+    let runnable = Runnable::new(future, shared_state);
+
+    let _handle = tokio::spawn(runnable);
```
touilleMan: would be useful in this kind of code to explain why dropping is okay

(btw, is there a rustonic way of showing drop without using the variable is not an error in the code ? using `drop(handle)` explicitly ?)

oxidation/crates/platform-async/Cargo.toml
FirelightFlagboy: I can use `target_family = "wasm"`. 

I prefer `target_*` over `feature = web / native` because it's more explicite that the architecture/capabilities will be different

oxidation/crates/platform-async/Cargo.toml
```diff
@@ -0,0 +1,26 @@
+[package]
```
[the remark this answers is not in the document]
FirelightFlagboy: I've moved the crate

oxidation/crates/platform-async/src/task.rs
```diff
@@ -0,0 +1,15 @@
...
+    /// Cancels the task.
+    fn cancel(&self) -> Option<T>;
+
+    /// Detaches the task to let it keep running in the backgroud.
+    fn detach(self);
+
+    /// Return `true` if the current task is finished.
+    fn is_finised(&self) -> bool;
```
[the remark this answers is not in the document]
FirelightFlagboy: done

touilleMan (changes requested):

oxidation/crates/platform-async/src/task.rs
FirelightFlagboy: It's to make a sure that any implementations will share a set of method

oxidation/crates/platform-async/src/platform/default.rs
FirelightFlagboy: clippy may don't like `drop(object)` https://rust-lang.github.io/rust-clippy/master/#drop_copy

oxidation/crates/platform-async/examples/eyeballs.rs
```diff
@@ -0,0 +1,137 @@
...
+
+    let task = spawn(attempt(
+        which,
+        targets,
+        failed_attempt,
+        join_set.clone(),
+        sender,
+    ));
+
+    join_set.lock().unwrap().push(task);
```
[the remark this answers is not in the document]
FirelightFlagboy: it's an example ;) but I take you comment into consideration when I'll be implementing the joinset

oxidation/crates/platform-async/src/platform.rs
FirelightFlagboy: I've moved the file.

I've looked if we can configure clippy for the whole project:

It's not possible at the moment but https://github.com/rust-lang/cargo/issues/5034

oxidation/libparsec/crates/platform-async/examples/eyeballs.rs
```diff
@@ -0,0 +1,137 @@
...
+
+#[derive(Clone, Copy, Debug)]
+enum Mode {
+    Fail(Duration),
+    Success(Duration),
+}
+
+#[tokio::main]
+async fn main() -> Result<(), ()> {
+    simple_logger::init_with_level(log::Level::Debug).expect("cannot initialise simple logger");
```
TimeEngineer, line 24: initialize

oxidation/libparsec/crates/platform-async/examples/eyeballs.rs
```diff
@@ -0,0 +1,137 @@
...
+    let join_set = Vec::with_capacity(targets.len());
+    let join_set = Arc::new(Mutex::new(join_set));
+    let (tx, mut rx) = mpsc::channel::<usize>(targets.len());
+
+    spawn_attempt(0, targets, failed_attempt, join_set.clone(), tx);
+    let result = dbg!(rx.recv().await);
+    for task in join_set.lock().unwrap().iter() {
+        task.cancel();
+    }
+    result.ok_or(())
```
TimeEngineer, line 57: What's the point to transform the option ?

oxidation/libparsec/crates/platform-async/examples/eyeballs.rs
```diff
@@ -0,0 +1,137 @@
...
+            failed_attempt[which].notify_waiters();
+        }
+        Ok(_) => {
+            log::info!("[#{which}] finished with success");
+            sender.send(which).await.unwrap();
+        }
+    }
+}
+
+async fn mock_load(mode: Mode) -> Result<(), ()> {
```
TimeEngineer, line 126: Better to use an Enum as Result to describe more clearly

oxidation/libparsec/crates/platform-async/src/platform/default.rs
```diff
@@ -0,0 +1,149 @@
...
+    F: Future + Send + 'static,
+    F::Output: Send + 'static,
+{
+    let shared_state = SharedState::default();
+    let shared_state = Arc::new(Mutex::new(shared_state));
+
+    let task = Task::new(shared_state.clone());
+    let runnable = Runnable::new(future, shared_state);
+
+    let _handle = tokio::spawn(runnable);
```
TimeEngineer, line 23: unused variable

TimeEngineer (changes requested):
Some minor change

oxidation/libparsec/crates/platform-async/examples/eyeballs.rs
FirelightFlagboy: corrected in 001824770

oxidation/libparsec/crates/platform-async/examples/eyeballs.rs
```diff
@@ -0,0 +1,137 @@
...
+        .iter()
+        .map(|_| Arc::new(Notify::new()))
+        .collect::<Vec<_>>();
+    let targets = Arc::new(targets);
+    let join_set = Vec::with_capacity(targets.len());
+    let join_set = Arc::new(Mutex::new(join_set));
+    let (tx, mut rx) = mpsc::channel::<usize>(targets.len());
+
+    spawn_attempt(0, targets, failed_attempt, join_set.clone(), tx);
+    let result = dbg!(rx.recv().await);
```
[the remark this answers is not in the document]
FirelightFlagboy: it's an example ;)

oxidation/libparsec/crates/platform-async/examples/eyeballs.rs
FirelightFlagboy: `Result` is an `Enum`, what would you prefer ?

oxidation/libparsec/crates/platform-async/examples/eyeballs.rs
FirelightFlagboy: none

oxidation/libparsec/crates/platform-async/src/platform/default.rs
```diff
@@ -20,7 +20,7 @@ where
     let task = Task::new(shared_state.clone());
     let runnable = Runnable::new(future, shared_state);
 
-    let _handle = tokio::spawn(runnable);
+    drop(tokio::spawn(runnable));
```
TimeEngineer, line 23: why drop ?
It reaches out of scope after the return

oxidation/libparsec/crates/platform-async/examples/eyeballs.rs
FirelightFlagboy: changed in bb524deee

TimeEngineer (changes requested):

oxidation/libparsec/crates/platform-async/examples/eyeballs.rs
TimeEngineer:
```Rust
MockResult {
    Success,
    Failed,
}
```
Result is an Enum, but generic Enum, why should I want to think about if there is something wrapped inside when I read a match arms ?
It's like why should I use Enum instead of bool ^^

TimeEngineer: You can also keep your Mode Enum
Do you know this pattern ?
```Rust
enum Enum {
    Foo(i32),
    Bar(i32),
}

fn fun() -> Enum {
    let e = Enum::Foo(5);
    let (Enum::Foo(delay) | Enum::Bar(delay)) = e;
    // sleep delay
    e
}
```

TimeEngineer: I'm wondering what's the best way for this type of structure
```rust
struct Foo {
    delay: i32
    mode: Mode
}
enum Mode {
    Success,
    Fail,
```
or the current implementation

FirelightFlagboy:
```rust
let e = Enum::Foo(5);
let (Enum::Foo(delay) | Enum::Bar(delay)) = e;
```
I know this pattern

FirelightFlagboy: > I'm wondering what's the best way for this type of structure
> 
> ```rust
> struct Foo {
>     delay: i32,
>     mode: Mode,
> }
> enum Mode {
>     Success,
>     Fail,
> }
> ```
> 
> or the current implementation

I like this structure, but keep in mind that this is an example

oxidation/libparsec/crates/platform-async/src/platform/native/task.rs
```diff
@@ -0,0 +1,149 @@
...
+    detached: bool,
+    value: Option<T>,
+    task_waker: Option<Waker>,
+    runnable_waker: Option<Waker>,
+}
+
+impl<T> Default for SharedState<T> {
+    fn default() -> Self {
+        Self {
+            canceled: false,
```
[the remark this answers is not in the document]
FirelightFlagboy: yes it can have multiple combination

oxidation/libparsec/crates/platform-async/src/platform/default.rs
FirelightFlagboy: I don't need the handle returned by tokio.

I've removed the drop in 847d250

oxidation/libparsec/crates/platform-async/src/native/timer.rs
```diff
@@ -0,0 +1,44 @@
...
+            sleep: Box::pin(sleep(duration)),
+        }
+    }
+}
+
+impl Future for Timer {
+    type Output = ();
+
+    fn poll(mut self: Pin<&mut Self>, cx: &mut Context<'_>) -> Poll<Self::Output> {
+        self.sleep.poll_unpin(cx)
```
touilleMan, line 42: Use `Pin::new(self).poll(cx)` to avoid needed future ? (cf. https://rust-lang-nursery.github.io/futures-api-docs/0.1.27/src/futures_util/future/mod.rs.html#531)

oxidation/libparsec/crates/platform-async/src/native/task.rs
```diff
@@ -0,0 +1,166 @@
...
+        state.canceled = true;
+        if let Some(ref waker) = state.runnable_waker {
+            waker.wake_by_ref();
+        }
+        state.value.take()
+    }
+
+    /// Detaches the task to let it keep running in the background
+    pub fn detach(self) {
+        self.shared_state.lock().unwrap().detached = true;
```
touilleMan, line 87: expect instead of unwrap ?

oxidation/libparsec/crates/platform-async/src/native/task.rs
```diff
@@ -0,0 +1,166 @@
...
+    shared_state: Arc<Mutex<SharedState<T>>>,
+}
+
+impl<T> Task<T> {
+    fn new(shared_state: Arc<Mutex<SharedState<T>>>) -> Self {
+        Self { shared_state }
+    }
+
+    /// Cancels the task
+    pub fn cancel(&self) -> Option<T> {
```
touilleMan, line 75: cancel is too close to what we use in trio, we should use another more explicit name (to show the fact the future is not longer going to be called anymore)

so maybe `abort` to use the tokio jargon

oxidation/libparsec/crates/platform-async/src/native/timer.rs
TimeEngineer: I prefer ctx instead of cx, because it is used in a lot of crates

FirelightFlagboy: `cx` is the default name of the parameters in the trait definition

FirelightFlagboy: https://doc.rust-lang.org/std/future/trait.Future.html

TimeEngineer: God :man_facepalming:

oxidation/libparsec/crates/platform-async/examples/eyeballs.rs
```diff
@@ -0,0 +1,134 @@
+// Parsec Cloud (https://parsec.cloud) Copyright (c) BSLv1.1 (eventually AGPLv3) 2016-2021 Scille SAS
+#![cfg(not(target_arch = "wasm32"))]
```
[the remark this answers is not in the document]
FirelightFlagboy: done in 77c6eba35

oxidation/libparsec/crates/platform-async/src/native/task.rs
FirelightFlagboy: done in 9de5859

FirelightFlagboy:
@touilleMan and @TimeEngineer

Could you review the changes that I've made

TimeEngineer (approved):

PR #2278, opened 2022-05-13
Pki submit email check

### LINKED ISSUE #2255: Check user email already used in pki submit request
Currently it's allowed to submit a PKI enrollment request with a user email already used. This triggers an error when the administrator accept the invitation (as email is unique for a human in an organization).

The e-mail address can be check on the submit request to directly fail when a user tries to join with an already used error.

AntoineDupre:
fixes #2255

### DIFF
--- a/parsec/api/protocol/pki.py
+++ b/parsec/api/protocol/pki.py
@@ -28,7 +28,7 @@ class PkiEnrollmentSubmitReqSchema(BaseReqSchema):
 
     submitter_der_x509_certificate = fields.Bytes(require=True)
     # Duplicated certificate email field. (The backend need to check if the email is used without loading the certificate)
-    submitter_der_x509_certificate_email = fields.String(required=True)
+    submitter_der_x509_certificate_email = fields.String(required=False)
     submit_payload_signature = fields.Bytes(required=True)
     submit_payload = fields.Bytes(required=True)  # Signature should be checked before loading
 
--- a/parsec/api/protocol/pki.py
+++ b/parsec/api/protocol/pki.py
@@ -27,6 +27,8 @@ class PkiEnrollmentSubmitReqSchema(BaseReqSchema):
     force = fields.Boolean(required=True)
 
     submitter_der_x509_certificate = fields.Bytes(require=True)
+    # Duplicated certificate email field. (The backend need to check if the email is used without loading the certificate)
+    submitter_der_x509_certificate_email = fields.String(required=True)
     submit_payload_signature = fields.Bytes(required=True)
     submit_payload = fields.Bytes(required=True)  # Signature should be checked before loading
 
--- a/parsec/api/protocol/pki.py
+++ b/parsec/api/protocol/pki.py
@@ -28,7 +28,9 @@ class PkiEnrollmentSubmitReqSchema(BaseReqSchema):
 
     submitter_der_x509_certificate = fields.Bytes(require=True)
     # Duplicated certificate email field. (The backend need to check if the email is used without loading the certificate)
-    submitter_der_x509_certificate_email = fields.String(required=False, missing=None)
+    submitter_der_x509_certificate_email = fields.String(
+        required=False, missing=None, allow_none=True
+    )
     submit_payload_signature = fields.Bytes(required=True)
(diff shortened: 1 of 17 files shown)


parsec/backend/postgresql/pki.py
```diff
@@ -200,6 +202,19 @@
...
 )
+_q_retrieve_active_human_by_email_for_update = Q(
+    f"""
+    SELECT
+        user_.user_id
+    FROM user_ LEFT JOIN human ON user_.human=human._id
+    WHERE
+        user_.organization = { q_organization_internal_id("$organization_id") }
+        AND human.email = $email
+        AND (user_.revoked_on IS NULL OR user_.revoked_on > $now)
```
touilleMan, line 213: The `user_.revoked_on > $now` is not needed: we consider revoked any user that have `revoked_on` not NULL

parsec/core/backend_connection/cmds.py
```diff
@@ -594,6 +594,7 @@ async def pki_enrollment_submit(
     enrollment_id: UUID,
     force: bool,
     submitter_der_x509_certificate: bytes,
+    submitter_der_x509_certificate_email: str,
```
touilleMan, line 597: `pki_enrollment_submit` is an anonymous command, so I think this function is never used (and same thing for `pki_enrollment_info`)

parsec/backend/pki.py
```diff
@@ -138,6 +142,9 @@ async def api_pki_enrollment_submit(
                 submit_payload_signature=msg["submit_payload_signature"],
                 submit_payload=msg["submit_payload"],
                 submitted_on=submitted_on,
+                submitter_der_x509_certificate_email=msg.get(
```
touilleMan, line 145: msg contains the default value even if the sent payload was missing it (it's the deserialization's job)

image

parsec/backend/memory/pki.py
```diff
@@ -61,6 +62,7 @@ async def submit(
         submit_payload_signature: bytes,
         submit_payload: bytes,
         submitted_on: DateTime,
+        submitter_der_x509_certificate_email: Optional[str] = None,
```
touilleMan, line 65: Should not have a default value given the parent declaration doesn't have it (I also think it's better not to have a default value here)

parsec/backend/memory/pki.py
```diff
@@ -107,6 +109,16 @@ async def submit(
                 # last one represent the current state of this x509 certificate.
                 break
 
+        # Optional check for client compatibility with version < 2.8.3
+        if submitter_der_x509_certificate_email:
+            # Assert email not used.
+            _, total = await self._user_component.find_humans(
```
touilleMan, line 115: would be better to use sync calls only here (for instance by introducing a `MemoryUserComponent._find_humans` sync method): this is allow us to have atomic operation given we don't have the for update locking and transaction that is available in postgresql

tests/backend/test_pki.py
```diff
@@ -29,6 +33,7 @@ async def _submit_request(
     signature=b"<signature>",
     request_id=None,
     force=False,
+    certif_email="another_bob_email",
```
touilleMan, line 36: don't want to use a real email format ? e.g. `another_bob@example.com`

touilleMan (changes requested):

vxgmichel (approved):
:+1: for adding compatibility tests

parsec/core/backend_connection/cmds.py
AntoineDupre: That's right, i removed.This field has been moved to the anonymous command

parsec/api/protocol/pki.py
```diff
@@ -27,6 +27,8 @@ class PkiEnrollmentSubmitReqSchema(BaseReqSchema):
     force = fields.Boolean(required=True)
 
     submitter_der_x509_certificate = fields.Bytes(require=True)
+    # Duplicated certificate email field. (The backend need to check if the email is used without loading the certificate)
+    submitter_der_x509_certificate_email = fields.String(required=False, missing=None)
```
touilleMan, line 31: can you add `allow_none=False` also ? this is not strictly needed, however I find is more explicit (and PR #2052 add `allow_none=...` everywhere `required=False`)

PR #2306, opened 2022-05-17
Remove unused coverage reports

vxgmichel:
Should we?


touilleMan:
Not sure what the point of this... do we gain speed ?

The good things about thoses test reports are we can have analytics of tests over time, see https://dev.azure.com/Scille/parsec/_build?definitionId=1&_a=summary&view=ms.vss-pipelineanalytics-web.new-build-definition-pipeline-analytics-view-cardmetrics

vxgmichel:
Nevermind, I thought this was triggering the coverage, except it doesn't.

PR #2327, opened 2022-05-19
Improve evolve_as_block speed

TimeEngineer:
In my machine:

# CPU
model name: AMD Ryzen 7 3800X 8-Core Processor
cache size: 512 KB
cpu MHz: 3515.208 (min) - 4397.87 (max)

# Result
Data Size: 524 KB
Before: ~ 10 ms
Now: ~ 0.27 ms

### DIFF
--- a/oxidation/crates/parsec_api_crypto/Cargo.toml
+++ b/oxidation/crates/parsec_api_crypto/Cargo.toml
@@ -18,7 +18,7 @@ libsodium-sys = { version = "0.2.7", optional = true }
 
 digest = { version = "0.10.1", optional = true }
 generic-array = { version = "0.14.4", optional = true, features = ["serde"] }
-sha2 = { version = "0.10.0", optional = true }
+sha2 = { version = "0.10.0", optional = true, features = ["asm"] }
 xsalsa20poly1305 = { version = "0.8.0", optional = true }
 ed25519-dalek = { version = "1.0.1", optional = true }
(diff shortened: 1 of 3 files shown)


FirelightFlagboy (approved):
could be interesting to crate a benchmark in rust https://doc.rust-lang.org/cargo/commands/cargo-bench.html

TimeEngineer:
> could be interesting to crate a benchmark in rust https://doc.rust-lang.org/cargo/commands/cargo-bench.html
Yes it could be great for many operations in Rust

Here the major improvement is binding, we should avoid Vec<u8> in input

touilleMan (approved):

oxidation/crates/parsec_api_crypto/Cargo.toml
```diff
@@ -18,7 +18,7 @@ libsodium-sys = { version = "0.2.7", optional = true }
 
 digest = { version = "0.10.1", optional = true }
 generic-array = { version = "0.14.4", optional = true, features = ["serde"] }
-sha2 = { version = "0.10.0", optional = true }
+sha2 = { version = "0.10.0", optional = true, features = ["asm"] }
```
touilleMan, line 21: asm feature compatible with wasm32 ?

TimeEngineer: Only x86, x86-64, and (partially) AArch64 architectures are currently supported.

PR #2341, opened 2022-05-22
Remove tcp mock from tests

### LINKED ISSUE #2337: Remove tcp mock from tests
shame, but needed to be compatible with the Rust bindings

touilleMan:
fix #2337


.azure-pipelines.yml
```diff
@@ -213,7 +213,7 @@ jobs:
     displayName: 'Install'
   - bash: |
       set -eux
-      py.test $(pytest.base_args) tests -n auto --runrust
+      python tests/run_parallel.py -n auto $(pytest.base_args) tests --runrust
```
vxgmichel, line 216: Why is `-n auto` still here if `run_parallel.py` already runs the tests in parallel?

vxgmichel: Nevermind, I get it now.

vxgmichel (approved):
We should be good to go

touilleMan:
image

PR #2354, opened 2022-05-23
[GUI] Added validators when entering file names

### LINKED ISSUE #2331: Missing feedback when trying to name/rename with invalid characters in GUI
### Version

2.9.0

Windows, Linux, MacOS

### What happened ?

When naming a new folder or renaming a file through the GUI with a forbidden character, a `ValueError` is raised from `EntryName` and isn't handled, which results in the user having no feedback on why the operation failed, and the error potentially being raised through telemetry.

### Relevant output

_No response_

Max-7:
Closes #2331

### DIFF
--- a/newsfragments/2331.misc.rst
+++ b/newsfragments/2331.misc.rst
@@ -1,0 +1,1 @@
+Added some feedback when entering a file name to avoid invalid names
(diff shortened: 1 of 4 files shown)


Ironicbay (approved):
LGTM 👍

PR #2363, opened 2022-05-25
Add parsec-sequestre-dump cli base

TimeEngineer:


oxidation/parsec_cli/parsec_sequestre_dump/Cargo.toml
```diff
@@ -0,0 +1,15 @@
...
+license = " BUSL-1.1"
+
+# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
+
+[dependencies]
+parsec_api_crypto = { path = "../../crates/parsec_api_crypto" }
+
+diesel = { version = "2.0.0-rc.0", features = ["sqlite"], git = "https://github.com/diesel-rs/diesel" }
+clap = { version = "3.1.18", features = ["derive"] }
+rmp-serde = "1.0.0"
```
touilleMan, line 14: I don't think `serde`&`rmp-serde`  are needed since we won't create our own schemas (so we need parsec_api_types` instead)

oxidation/parsec_cli/parsec_sequestre_dump/migrations/0001_vlob_atom/up.sql
```diff
@@ -0,0 +1,22 @@
...
+    -- author/timestamp are required to validate the consistency of blob
+    -- Care must be taken when exporting this field (and the device_ table) to
+    -- keep this relationship valid !
+    author INTEGER REFERENCES device (_id) NOT NULL,
+    -- this field is called created_on in Parsec datamodel, but it correspond to the timestamp field in the API
+    -- (the value is provided by the client when sending request and not created on backend side) so better
+    -- give it the better understandable name
+    timestamp REAL NOT NULL,
+
+    UNIQUE(vlob_id)
```
touilleMan, line 21: I made a mistake in the schema (I've since correct it in the issue ^^), this should be `UNIQUE(vlob_id, version)`

touilleMan, line 21: I made a mistake in the schema (I've since correct it in the issue ^^), this should be `UNIQUE(vlob_id, version)`

oxidation/parsec_cli/parsec_sequestre_dump/migrations/0001_vlob_atom/up.sql
```diff
@@ -0,0 +1,22 @@
...
+    blob BLOB NOT NULL,
+    size INTEGER NOT NULL,
+    -- author/timestamp are required to validate the consistency of blob
+    -- Care must be taken when exporting this field (and the device_ table) to
+    -- keep this relationship valid !
+    author INTEGER REFERENCES device (_id) NOT NULL,
+    -- this field is called created_on in Parsec datamodel, but it correspond to the timestamp field in the API
+    -- (the value is provided by the client when sending request and not created on backend side) so better
+    -- give it the better understandable name
+    timestamp REAL NOT NULL,
```
touilleMan, line 19: Every time I see the type of this timestamp field I fell stupid for having thought once it would be a good idea :'(

According to SQLite documentation (https://www.sqlite.org/datatype3.html), REAL is a 8-byte IEEE floating point number, so it is fine to store our millisecond precision datetime monster ^^

oxidation/parsec_cli/parsec_sequestre_dump/src/model.rs
```diff
@@ -0,0 +1,47 @@
...
+pub struct Data {
+    pub blocks: Vec<Block>,
+    pub vlob_atoms: Vec<VlobAtom>,
+    pub role_certificates: Vec<Vec<u8>>,
+    pub user_certificates: Vec<Vec<u8>>,
+    pub device_certificates: Vec<Vec<u8>>,
+}
+
+impl Data {
+    pub fn save(&self, key: &SecretKey, path: &str) {
```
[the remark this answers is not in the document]
touilleMan: Ha ! I just realized `save` here means save *to disk* and not save *to database*

Anyway, we should use the `parsec_api_types` schema to do the deserialization

(don't do the consistency checking on expected timestamp & expected author for now, this is not vital and we should first think a bit about the architecture & trustchani public api before doing this)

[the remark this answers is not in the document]
touilleMan: Aaaand one last thing: I think it would be better to keep a separation between the data fetched from sqlite (i.e. those structs) and the work we want to do on them (e.i. this method)

Hence you should have a separate file with functions (and not methods) for the save on disk logic

oxidation/parsec_cli/parsec_sequestre_dump/src/main.rs
```diff
@@ -0,0 +1,159 @@
...
+use parsec_api_crypto::SecretKey;
+
+use crate::{
+    model::{Block, Data, VlobAtom},
+    schema::{block, device, realm_role, user_ as user, vlob_atom},
+};
+
+#[derive(Parser, Debug)]
+#[clap(author, version, about, long_about = None)]
+struct Args {
```
FirelightFlagboy, line 20: could you add a *help* message to the options, It would be useful for the user when he make a mistake or use `--help` / `-h`

oxidation/parsec_cli/parsec_sequestre_dump/src/main.rs
```diff
@@ -0,0 +1,159 @@
...
+    #[clap(short, long)]
+    input: String,
```
FirelightFlagboy, line 22: 1. You could use a clap [`validator`](https://docs.rs/clap/latest/clap/struct.Arg.html#method.validator) to check if the file is present or not
2. Why not using a `PathBuf` / `Path` ?

oxidation/parsec_cli/parsec_sequestre_dump/src/main.rs
```diff
@@ -0,0 +1,159 @@
...
+
+    #[clap(short, long)]
+    key: String,
+
+    #[clap(short, long)]
+    output: String,
```
FirelightFlagboy, line 28: Why not using `PathBuf` / `Path` ?

oxidation/parsec_cli/parsec_sequestre_dump/src/model.rs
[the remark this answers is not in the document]
FirelightFlagboy: I would prefer a name like `dump`, That brings us back to the idea of dumping the database on the disk

oxidation/parsec_cli/parsec_sequestre_dump/src/model.rs
```diff
@@ -0,0 +1,47 @@
...
+        let data = rmp_serde::to_vec_named(self).unwrap();
```
FirelightFlagboy, line 36:
```suggestion
        let data = rmp_serde::to_vec_named(self).expect("failed to format data");
```

oxidation/parsec_cli/parsec_sequestre_dump/src/model.rs
```diff
@@ -0,0 +1,47 @@
...
+        let encrypted = key.encrypt(&data);
+        std::fs::write(path, &encrypted).unwrap();
```
FirelightFlagboy, line 38:
```suggestion
        std::fs::write(path, &encrypted).expect("failed to write data on the disk");
```

FirelightFlagboy (changes requested):
1. It's expected for the news fragment to be empty ? I mean, we're adding a new tool
2. So do we settle with the name `parsec_sequestre_dump`, I would have prefer `parsec_sequestre` it's shorter :)

oxidation/parsec_cli/parsec_sequestre_dump/Cargo.toml
```diff
@@ -0,0 +1,15 @@
...
+version = "0.0.0"
+edition = "2021"
+license = " BUSL-1.1"
+
+# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
+
+[dependencies]
+parsec_api_crypto = { path = "../../crates/parsec_api_crypto" }
+
+diesel = { version = "2.0.0-rc.0", features = ["sqlite"], git = "https://github.com/diesel-rs/diesel" }
```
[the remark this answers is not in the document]
TimeEngineer: Wonderful !

oxidation/parsec_cli/parsec_sequestre_dump/migrations/0005_info/up.sql
```diff
@@ -0,0 +1,33 @@
...
+    version INTEGER NOT NULL,  -- should be 1 for now
+    realm_id BLOB NOT NULL
+);
+
+INSERT INTO info VALUES (
+    87947, 1, 'realm_id'
+);
+
+
+INSERT INTO realm_role VALUES (
```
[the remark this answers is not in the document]
TimeEngineer: I wanted to test it

oxidation/parsec_cli/parsec_sequestre_dump/src/schema.rs
```diff
@@ -0,0 +1,58 @@
...
+        vlob_id -> Binary,
+        version -> Integer,
+        blob -> Binary,
+        size -> Integer,
+        author -> Integer,
+        timestamp -> Float,
+    }
+}
+
+joinable!(block -> device (author));
```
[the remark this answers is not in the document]
TimeEngineer: Yes, diesel generate this file from the current database

oxidation/parsec_cli/parsec_sequestre_dump/src/model.rs
```diff
@@ -0,0 +1,47 @@
...
+use parsec_api_crypto::SecretKey;
+use serde::{Deserialize, Serialize};
+
+#[derive(Debug, Queryable, Serialize, Deserialize, PartialEq)]
+pub struct Block {
+    pub block_id: Vec<u8>,
+    pub data: Vec<u8>,
+    pub author: i32,
+    pub size: i32,
+    pub created_on: f32,
```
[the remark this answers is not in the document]
TimeEngineer: Sqlite use String to store Timestamp ^^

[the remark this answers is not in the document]
touilleMan: yup we store the data in sqlite as float (as btw this should be a f64, see https://github.com/Scille/parsec-cloud/pull/2363#discussion_r882473832), but we want to manipulate in the code a datetime object.
So we need to put the conversion somewhere, and it would be best to have it here ;-)

touilleMan:
> 1. It's expected for the news fragment to be empty ? I mean, we're adding a new tool
> 
>     2. So do we settle with the name `parsec_sequestre_dump`, I would have prefer `parsec_sequestre` it's shorter :)

Yes I think it's better to keep the newsfragment empty for the moment given we are working on a small part of a new feature, so we will put it in the changelog when it is ready to use

oxidation/parsec_cli/parsec_sequestre_dump/src/model.rs
TimeEngineer: I will handle all unwrap later

PR #2366, opened 2022-05-27
Correct Schema

TimeEngineer:


oxidation/crates/parsec_client_types/schema/local_workspace_manifest.json
```diff
@@ -19,19 +19,40 @@
...
+            //   during the initial workspace creation
+            // This speculative information is useful during merge to understand if
+            // a data is not present in the placeholder compared with a remote because:
+            // a) the data is not locally known (speculative is True)
+            // b) the data is known, but has been locally removed (speculative is False)
+            // Prevented to be `required=True` by backward compatibility
             "name": "speculative",
             "type": "Boolean",
-            "introduced_in_revision": 6
+            "introduced_in_revision": 15
```
touilleMan, line 55: Watch out `revision 15` != `Parsec v1.15` !!!

There is no revision logic on the data, only on the protocol stuff (the rational is, unlike in protocol where client and server can evolve together out of an old format)

However we cannot just remove the `introduced_in_revision` field here (given it is used to mark that the field is maybe present), so you should just set `introduced_in_revision: 115` and add a comment `// hack to indicate this field was added in Parsec v1.15, given revision is meaningless for data`

touilleMan: and can you check for other place in the data where `introduced_in_revision` should be corrected

touilleMan: and don't worry for the hack thing, I'm currently writing some documentation on api versioning strategy, so we will fix this when this is ready ;-)

touilleMan (changes requested):

oxidation/crates/parsec_client_types/schema/local_workspace_manifest.json
TimeEngineer: Done

touilleMan (approved):
lgtm 👍

PR #2371, opened 2022-05-27
Poetry as package manager \o/

touilleMan:


touilleMan:
Artifacts to test: https://dev.azure.com/Scille/parsec/_build/results?buildId=12431&view=artifacts&pathAsName=false&type=publishedArtifacts

packaging/macOS/freeze_program.py
```diff
@@ -0,0 +1,186 @@
...
+        program_version = global_dict.get("__version__")
+    assert program_version.startswith("v")
+    program_version_without_v_prefix = program_version[1:]
+    print(f"### Detected Parsec version {program_version} ###")
+
+    # Bootstrap tools virtualenv
+    tools_python = TOOLS_VENV_DIR / "bin/python"
+    if not TOOLS_VENV_DIR.is_dir():
+        print("### Create tool virtualenv ###")
+        run(f"{ python } -m venv {TOOLS_VENV_DIR}")
```
FirelightFlagboy, line 71: TODO => Cache Venv

packaging/wheel/wheel_it.py
```diff
@@ -0,0 +1,88 @@
...
+import shutil
+import re
+
+
+# Fully-qualified path for the executable should be used with subprocess to
+# avoid unreliability (especially when running from within a virtualenv)
+python = shutil.which("python")
+poetry = shutil.which("poetry")
+if not python or not poetry:
+    raise SystemExit("Cannot find python or poetry commands :(")
```
FirelightFlagboy, line 15:
```suggestion
    raise SystemExit("Cannot find python and/or poetry commands :(")
```

packaging/win32/freeze_program.py
```diff
@@ -11,12 +13,19 @@
 from pathlib import Path
 
 
+# Fully-qualified path for the executable should be used with subprocess to
+# avoid unreliability (especially when running from within a virtualenv)
+python = shutil.which("python")
+if not python:
+    raise SystemExit("Cannot find python command :(")
```
FirelightFlagboy, line 20: We don't find python, in a python script ?

touilleMan: You would be surprised ^^

The python command may not be the one that has been used to start this script (it could be python3, or pypy)
There is a function in the stdlib to retrieve the command that started the script (hence retrieving *in theory* a python interpreter... unless we are in the very unlikely case the script has been packaged into a binary using e.g. pyinstaller)

So in the end I find it simpler and more readable to treat python as a regular external command

packaging/macOS/freeze_program.py
touilleMan: The ` if not TOOLS_VENV_DIR.is_dir()` is a way to reuse the virtualenv if it exists.
This is mostly done to ease the testing when modifying the script, otherwise this script is designed to be mostly a one shot run: typically when using it you want to start from scratch to be sure you're not packaging something outdated.

(On top of that pip is already caching the dependencies wheels)

FirelightFlagboy (approved):

packaging/macOS/freeze_program.py
FirelightFlagboy: The comment was more for the CI

touilleMan: I'd rather not use virtualenv cache (so only caching wheels from pip own cache) for packaging job to avoid misconfiguration or bug that could lead to a invalid generated release ;-)

PR #2372, opened 2022-05-30
[GUI] Prevent closing the dialog while creating an org

### LINKED ISSUE #2360: Improve organization creation popup UX

Max-7:
Closes #2360

### DIFF
--- a/newsfragments/2360.empty.rst
+++ b/newsfragments/2360.empty.rst
@@ -1,0 +1,1 @@
+Prevent closing the dialog while creating an organization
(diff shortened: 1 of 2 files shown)


newsfragments/2360.empty.rst
```diff
@@ -0,0 +1 @@
+Prevent closing the dialog while creating an organization
```
FirelightFlagboy, line 1: the fragment is not empty

Ironicbay (commented):
It's still possible to close the window at the password creation stage (with cmd-Q on mac, possibly with alt-f4 as well). If so, the org is created but unreachable and doesn't appear on the organisation list.

newsfragments/2360.empty.rst
Max-7: Yup, good catch. It's fixed.

PR #2404, opened 2022-06-01
Updated boto and botocore requirements

Max-7:
Dependabot struggles to update both at the same time (#2396 and #2399) because of what seems like a circular dependency. So we update them both manually at the same time.


Max-7:
See #2407

PR #2425, opened 2022-06-08
[GUI] Fixed performance issues when loading a folder

### LINKED ISSUE #2411: [🐛 | Bug]: Quadratic slowdown in GUI when opening big folders
### Version

2.8.0+

Windows

### What happened ?

On Windows, the folder view in the GUI is very slow to display big folder.
- it is not a network issue as the slow down is the same when going backward and re-entering the folder
- ~the issue is not present on Linux (tested with Parsec 2.9.2 on Ubuntu 20.04)~ edit: issue can be reproduced on Linux with a 4K screen (see comment below)
- ~the issue might be related to the `ItemDelegate.paint` (I've done some profiling and this part is called a lot, see screenshot)~ edit: removing `ItemDelegate` doesn't fix the slowdown, on the other hand using dummy data with `ItemDelegate` is instantaneous, so issue seems to come from scheduling the trio async function responsible for fetching the data

image

Max-7:
Closes #2411

### DIFF
--- a/newsfragments/2411.bugfix.rst
+++ b/newsfragments/2411.bugfix.rst
@@ -1,0 +1,1 @@
+Fixed performance issues when loading a folder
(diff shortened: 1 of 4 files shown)


vxgmichel (approved):
:tada:

PR #2427, opened 2022-06-08
[GUI] Clear file table when entering a new workspace

### LINKED ISSUE #2426: [🐛 | Bug]: The previously browsed file list is shown when a new workspace is being loaded
### Version

Latest

Windows, Linux, MacOS

### What happened ?

When switching between workspaces (or going back to the same workspace), the previously browsed file list appears while the new list is being loaded. The behavior makes sense when browsing to a subdirectory (i.e show the current directory while the subdirectory is being loaded) but the widget should be cleared when the files widget gets hidden.

Max-7:
Closes #2426

### DIFF
--- a/newsfragments/2426.bugfix.rst
+++ b/newsfragments/2426.bugfix.rst
@@ -1,0 +1,1 @@
+Do not display old files when entering a new workspace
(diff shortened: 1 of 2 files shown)


vxgmichel (approved):
LGTM :+1:

Issue #2419, opened 2022-06-03
Impl `parsec/core/backend_connection/authenticated.py` in Rust


FirelightFlagboy:
impl `parsec/core/backend_connection/authenticated.py` in Rust.

We'll be using `http` request instead of `WebSocket`.

For the http client, we'll be using `reqwest`

The request will required to be authenticated using a protocol that'll be define later.

PR #2431, opened 2022-06-09
Yolo/fix/wrap cmd rep load in result

FirelightFlagboy:
`Rep::load` now return a `Result` instead of including a variant `UnknownError`

this change is required for #2419 because, on the rust side it's expected that the load may failed, but looking for the specific variant is a anti pattern because we have the `Result` enum that do that for us.

### DIFF
--- a/oxidation/libparsec_python/src/protocol/block.rs
+++ b/oxidation/libparsec_python/src/protocol/block.rs
@@ -1,5 +1,6 @@
 // Parsec Cloud (https://parsec.cloud) Copyright (c) BSLv1.1 (eventually AGPLv3) 2016-2021 Scille SAS
 
+use pyo3::exceptions::PyValueError;
 use pyo3::import_exception;
 use pyo3::prelude::*;
 use pyo3::types::{PyBytes, PyType};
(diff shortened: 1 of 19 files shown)


oxidation/libparsec_python/src/protocol/block.rs
```diff
@@ -106,7 +107,9 @@ impl BlockCreateRep {
 
     #[classmethod]
     fn load(_cls: &PyType, buf: Vec<u8>) -> PyResult<Self> {
-        Ok(Self(block_create::Rep::load(&buf)))
+        block_create::Rep::load(&buf)
+            .map(Self)
+            .map_err(|e| PyErr::new::<PyValueError, _>(e.to_string()))
```
TimeEngineer, line 112: You can use directly PyValueError::new instead

oxidation/libparsec_python/src/protocol/events.rs
```diff
@@ -134,7 +134,7 @@ impl EventsListenRep {
     }
 
     #[classmethod]
-    #[pyo3(name = "OkRealmRolesUdpated")]
+    #[pyo3(name = "OkRealmRolesUpdated")]
```
TimeEngineer, line 137: Good catch

TimeEngineer (changes requested):

oxidation/libparsec_python/src/protocol/block.rs
FirelightFlagboy: 😯

oxidation/libparsec_python/src/protocol/events.rs
FirelightFlagboy: 😉

oxidation/libparsec_python/src/protocol/block.rs
FirelightFlagboy: done

PR #2433, opened 2022-06-09
format pre-commit + remove duplicate key

FirelightFlagboy:


.pre-commit-config.yaml
```diff
@@ -1,97 +1,95 @@
...
-        --tests,
-        --bins,
-        --lib,
-        --,
-        --deny=warnings,
-        --deny=clippy::self_named_module_files
-      ]
+      - id: fmt
+      - id: clippy
+        args:
```
touilleMan, line 62: not sure this `- --...` form is more readable 🤔

FirelightFlagboy: It was to normalize with the `args` of `black`.

PR #2446, opened 2022-06-09
Add WorkspaceStorage binding

TimeEngineer:


FirelightFlagboy (approved):

oxidation/libparsec_python/src/storage.rs
```diff
@@ -0,0 +1,326 @@
...
+
+    fn get_chunk<'py>(&self, py: Python<'py>, chunk_id: ChunkID) -> PyResult<&'py PyBytes> {
+        let chunk = py
+            .allow_threads(|| self.0.get_chunk(chunk_id.0))
+            .map_err(|_| FSLocalMissError::new_err(chunk_id))?;
+        Ok(PyBytes::new(py, &chunk))
+    }
+
+    // Pyo3 is inefficient with Vec<u8> but set_chunk must handle PyByteArray and PyBytes
+    // but can't safely call unsafe { block.as_bytes() } because it's called in multiple threads
```
touilleMan, line 213: I think it's more interesting here to keep using the unsafe trick and dropping the use of `py.allow_threads` instead

My guess is `set_chunk()` call should be very fast where the block vector is typically multiple Ko big

Can you run the tests with both approaches to have a rough idea of the perf difference ? (I guess this is going to be unfair for the unsafe approach given the chunks are really small in test, but it may at least show if the copy approach has a very performance drawback)

FirelightFlagboy (approved):

oxidation/libparsec_python/src/storage.rs
TimeEngineer: For a block of `4096B`:
```
allow_threads with Vec<u8>: 250µs
unsafe with as_bytes(): 150µs
```
When I move a file in my mountpoint, I see no big difference for a file of 186MB

touilleMan: That's intesting ;-)

I guess the copy is okay then !

touilleMan (approved):

PR #2448, opened 2022-06-10
Feat/serialization req new

### LINKED ISSUE #2449: Add `new` method on commands generated by `parsec_api_protocol`

FirelightFlagboy:
make `Req` struct impl method `new` to simplify their creation.

requirement for #2419 
closes #2449

### DIFF
--- a/oxidation/crates/parsec_serialization_format/src/parser/field.rs
+++ b/oxidation/crates/parsec_serialization_format/src/parser/field.rs
@@ -17,33 +17,18 @@ pub(crate) enum Vis {
 
 #[derive(Deserialize)]
 pub(crate) struct Field {
-    name: String,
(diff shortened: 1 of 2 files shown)


TimeEngineer:
I'm ok if you want a new method, but I don't understand why it is simpler ?

FirelightFlagboy:
> I'm ok if you want a new method, but I don't understand why it is simpler ?

It will be simpler for me when I generate the code that build the `Request` because without the `new` method I've to manage two way to create a new `Request`:

1. with `{ .. }` (the command `user_get` for example)
  ```rust
  Req { .. }
  ``` 
2. with `(..)` (then command `invite_new` for example)
  ```rust
  Req(..)
  ```

TimeEngineer (approved):

PR #2451, opened 2022-06-10
Make sha2 asm as feature

TimeEngineer:


oxidation/crates/parsec_api_crypto/Cargo.toml
```diff
@@ -62,6 +62,7 @@ use-rustcrypto = [
     "rand_07",
     "rand_08",
 ]
+asm = ["sha2?/asm"]
```
touilleMan, line 65: what is this syntax with a interrogation point ?

And I think you should add a comment to explain why this is needed

FirelightFlagboy: why not enabling the feature `sha2/asm` when the `target_arch` is not `wasm32` ?

touilleMan: @FirelightFlagboy the issue is not related to wasm32: the compilation seems to fail with msvc (see https://github.com/RustCrypto/hashes/issues/315)

but we could disable this feature for windows instead

FirelightFlagboy (approved):

oxidation/crates/parsec_api_crypto/Cargo.toml
TimeEngineer: In the doubt, we can leave this, we don't know about others platforms

PR #2486, opened 2022-06-20
Oxidation update libsodium

### LINKED ISSUE #2485: Oxidation update sodiumoxide serialization/deserialization
Sodiumoxide serialization/deserialization is not the same as RustCrypto.
In deserialization, it fills a buffer of length 32 instead of check if the data is correct

TimeEngineer:
closes #2485


FirelightFlagboy (commented):
What's the use of this PR ?

To be able to de/serialize Key pair ? If yes, for which use ?

FirelightFlagboy (approved):

PR #2499, opened 2022-06-22
Yolo/use yamllint

FirelightFlagboy:
Add `yamllint` configuration file, to enforce a style on `yaml` files


.yamllint.yml
```diff
@@ -0,0 +1,85 @@
+yaml-files:
```
touilleMan, line 1: What's the reason we don't use yamllint's default configuration ?

.landscape.yml
```diff
@@ -10,4 +10,3 @@ ignore-patterns:
   - (^|/)docs(/|$)
 python-targets:
   - 3
-
```
touilleMan, line 13: @vxgmichel do you know if this config file is still used ? I have the feeling we don't use this service since ages...

.pre-commit-config.yaml
```diff
@@ -1,97 +1,103 @@
 exclude: ^misc/bench.py$
 repos:
```
touilleMan, line 2: I guess we should add yamllint to pre-commit config

This should be fairly simple given yamllint repo contains a [pre-commit config](https://github.com/adrienverge/yamllint/blob/master/.pre-commit-hooks.yaml)
```yaml
- repo: https://github.com/adrienverge/yamllint
  rev: v1.26.3
  hooks:
  - id: yamllint
```

touilleMan (changes requested):

touilleMan:
I'd rather wait for #2371 to be merged before merging this one (given azure-pipeline.yml and snapcraft.yml gets modified in this PR)

.pre-commit-config.yaml
FirelightFlagboy: I've planned that in another PR

.yamllint.yml
FirelightFlagboy: 1. I find the default configuration a little to permissive for braces and bracket
2. I prefer the config to be explicite

.landscape.yml
vxgmichel: The commit is from 6 years ago, I wasn't event here :)

FirelightFlagboy: the file is removed

.yamllint.yml
touilleMan: I'd rather have one less configuration file to worry about

> I prefer the config to be explicit

I disagree: it's better to have one way to do the formatting and don't think about it (see for instance black for Python or rust-fmt compared to the mad house that is C formatting)

> I find the default configuration a little to permissive for braces and bracket

See my previous point ;-)

touilleMan (approved):

PR #2516, opened 2022-06-27
Yolo/rework addr

FirelightFlagboy:
- rename `to_http_domain_url` to `to_http_url_with_path` to better reflect what the method do
- rename `BaseBackendAddr::to_http_redirection_url` to `to_http_url` to better reflect what the method do
- correct some typo
- don't publicly expose `_to_url`
- more consistent spacing

### DIFF
--- a/oxidation/crates/parsec_api_types/src/addr.rs
+++ b/oxidation/crates/parsec_api_types/src/addr.rs
@@ -37,7 +37,7 @@ macro_rules! impl_common_stuff {
             }
 
             pub fn from_http_redirection(url: &str) -> Result<Self, AddrError> {
-                // For wathever reason, Url considers illegal changing scheme
+                // For whatever reason, Url considers illegal changing scheme
                 // from http/https to a custom one, so we cannot just use
                 // `Url::set_scheme` and instead have to do this hack :(
                 let url_with_forced_custom_scheme = format!("x{}", url);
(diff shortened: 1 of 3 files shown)


touilleMan:
> rename BaseBackendAddr::to_http_redirection_url to to_http_url to better reflect what the method do

I disagree, redirection url is a special case and we don't want to mess it with e.g. the http url used to send api command to the server.

FirelightFlagboy:
This method do nothing related to redirection

oxidation/crates/parsec_api_types/src/addr.rs
```diff
@@ -211,24 +212,22 @@ impl BaseBackendAddr {
         url
     }
 
-    pub fn to_http_redirection_url(&self) -> Url {
+    /// Create a url for http request.
+    pub fn to_http_url(&self) -> Url {
```
touilleMan, line 216: I think we said we could remove entirely this and only use `BaseBackendAddr::to_http_url_with_path(None)` instead

FirelightFlagboy: I was asking myself, if I wouldn't change its (for `to_http_url_with_path`) signature to

```rust
pub fn to_http_url_with_path(&self, path: &str) -> Url {}
```

touilleMan: > my comment above was for the case when we have both to_http_url and to_http_url_with_path

I think it's better to only have a single `to_http_url_with_path(path: Option<...>)`

FirelightFlagboy: Done, I've merged both

touilleMan (approved):

PR #2517, opened 2022-06-27
Fix workspace sneak peek dup

Max-7:


vxgmichel (approved):
LGTM :+1: 

Could you target branch 3.10 so we can gather all the fixes there?

PR #2519, opened 2022-06-27
[GUI] Don't overwrite files when importing a new file with the same name

### LINKED ISSUE #2504: [🐛 | Bug]: No confirmation when importing a file over an existing one
### Version

v2.10.0-rc1

Windows, Linux

### What happened ?

Importing a file over an existing one doesn't prompt the user for what to do, nor does it use a new name like `something (2).txt`.

The same happen with drag&drop from the file explorer.

However, drag&drop from the application **does** use a new name with an increment.

### Relevant output

_No response_

Max-7:
Quickfix to avoid overwriting an existing file when importing a new one with the same name: the new file is renamed to something like "file (2).ext".
Closes #2504

### DIFF
--- a/newsfragments/2504.bugfix.rst
+++ b/newsfragments/2504.bugfix.rst
@@ -1,0 +1,1 @@
+Don't overwrite files when importing a new file with the same name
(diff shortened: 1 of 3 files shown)


vxgmichel (approved):
LGTM :+1:

PR #2523, opened 2022-06-27
migrate QA job to github action

FirelightFlagboy:


touilleMan:
FYI #2371 change quiet a few things in azure-pipeline.yml

FirelightFlagboy:
> I added this commit [2541049](https://github.com/Scille/parsec-cloud/commit/2541049dabeb1102c6f40bad6477c85afc65d81d) as part of the 2.10 release, could you migrate it too?

Done

ci_utils/check_commit_signature.py
```diff
@@ -0,0 +1,24 @@
+# Parsec Cloud (https://parsec.cloud) Copyright (c) AGPLv3 2016-2021 Scille SAS
```
touilleMan, line 1: I'd rather have a `.github/scripts` folder instead of this `ci_utils`: one less entry in this too crowded root folder ;-)

(besidet that's [what is done in the doc example](https://docs.github.com/en/actions/learn-github-actions/essential-features-of-github-actions#adding-scripts-to-your-workflow) ^^)

touilleMan: Add moving ci scripts out of the config file is totally the right move for readability ! 😃

.github/workflows/quality-assurance.yml
```diff
@@ -0,0 +1,86 @@
...
+env:
+  python-version: "3.9"
+  rust-version: "1.61.0"
+  node-version: "16"
+
+jobs:
+  check-quality-assurance:
+    runs-on: ubuntu-22.04
+    steps:
+      - uses: actions/checkout@v3
```
touilleMan, line 15: We should pin github action with commit hash instead of tag.

Otherwise an attacker getting access to the action's repo can modify the tag and execute arbitrary code which is able to exfiltrate the secrets we use in our CI (in our case typically api token to upload to snap and pypi).

However it's your lucky day given we already have all the code to implement this ready: 😄 
- https://github.com/touilleMan/godot-python/blob/master/.github/workflows/build.yml
- https://github.com/touilleMan/godot-python/blob/6adbe46550c90c0f676b8835e854eddd1b654a97/.pre-commit-config.yaml#L19-L26
- https://github.com/touilleMan/godot-python/blob/master/misc/pin_github_actions.py

ci_utils/check_commit_signature.py
FirelightFlagboy: done

.github/workflows/quality-assurance.yml
```diff
@@ -0,0 +1,86 @@
...
+env:
+  python-version: "3.9"
+  rust-version: "1.61.0"
+  node-version: "16"
+
+jobs:
+  check-quality-assurance:
+    runs-on: ubuntu-22.04
+    steps:
+      - uses: actions/checkout@2541b1294d2704b0964813337f33b291d3f8596b
```
touilleMan, line 15: can you also add a comment to indicate which tag this commit is supposed to stand for (yeah the commit is safer but also less readable when we want to do an update ^^)
```
 - uses: actions/checkout@2541b1294d2704b0964813337f33b291d3f8596b  # tag foobar
```

.github/workflows/quality-assurance.yml
```diff
@@ -0,0 +1,86 @@
...
+          python .github/scripts/check_news_fragment.py ${{ github.ref_name }}
+
+      - name: Run cargo build to prepare rust's cache
+        run: cargo build --workspace
+
+      - name: Patch pre-commit for line-ending
+        run: |
+          sed -i '/id: mixed-line-ending/a\      args:\n\        - "--fix=lf"' .pre-commit-config.yaml
+
+      - uses: pre-commit/action@646c83fcd040023954eafda54b4db0192ce70507
```
touilleMan, line 84: This action depends itself of `actions/cache@v3` (see https://github.com/pre-commit/action/blob/main/action.yml).

Unfortunately we cannot pin the transitive dependency :(

Do you have an idea how action triggering sub action like this is common ? I would guess it's something rather rare (I've had a look at cache, setup-python and setup-node, which all don't do that).

Regarding what to do here, `actions/cache` is an official action from Github that is massively used. So the risk of it being used to leak data is very low (compared to using a action created by $RANDOM_DEV).

So I guess this is ok to use pre-commit/action here, but we should be careful of this and read the content of each action we want to add to our CI

.github/workflows/quality-assurance.yml
FirelightFlagboy: done for pinned actions version, for the pre-commit hook, I'll do this in another PR

.github/workflows/quality-assurance.yml
FirelightFlagboy: done

.github/workflows/quality-assurance.yml
```diff
@@ -0,0 +1,86 @@
...
+  rust-version: "1.61.0"
+  node-version: "16"
+
+jobs:
+  check-quality-assurance:
+    runs-on: ubuntu-22.04
+    steps:
+      - uses: actions/checkout@2541b1294d2704b0964813337f33b291d3f8596b # pin v3.0.2
+
+      - uses: actions/setup-python@d09bd5e6005b175076f227b13d9730d56e9dcfcb
```
touilleMan, line 17:
```suggestion
      - uses: actions/setup-python@d09bd5e6005b175076f227b13d9730d56e9dcfcb  # pin v4.0.0
```

.github/workflows/quality-assurance.yml
```diff
@@ -0,0 +1,86 @@
...
+        with:
+          python-version: ${{ env.python-version }}
+
+      - uses: actions/setup-node@eeb10cff27034e7acf239c5d29f62154018672fd
```
touilleMan, line 21:
```suggestion
      - uses: actions/setup-node@eeb10cff27034e7acf239c5d29f62154018672fd  # pin v3.3.0
```

touilleMan (changes requested):

.github/workflows/quality-assurance.yml
touilleMan: > We could also fork each actions used in our orga.

This seems like a good approach whenever we want to use exotic (i.e. not Github official) actions 👍

touilleMan (approved):
👍

PR #2525, opened 2022-06-27
Fix the propagation of the revoked status to other users

### LINKED ISSUE #2505: [🐛 | Bug]: Revoked status not propagated correctly to other users
### Version

v2.10.0-rc1

Windows

### What happened ?

If Alice gets revoked by Bob, Carl is supposed to get notified that Alice is revoked.

However, Carl does not see Alice as revoked in the user list, even though the "hide revoked user" button **does** behave correctly, hiding Alice profile. Carl can also share a workspace with Alice, and the snackbar indicates this has been done successfully. Then Alice **does** appear as revoked in the user list and **does not** appear in the list of the workspace roles. 

We should make sure that the workspace key has not been sent to Alice in the process.

### Relevant output

_No response_

vxgmichel:
Fix #2505

### DIFF
--- a/parsec/backend/user_type.py
+++ b/parsec/backend/user_type.py
@@ -54,7 +54,7 @@ class User:
         return attr.evolve(self, **kwargs)
 
     def is_revoked(self):
-        return self.revoked_on and self.revoked_on <= pendulum.now()
+        return self.revoked_on
 
     @property
     def public_key(self) -> PublicKey:
--- a/tests/backend/realm/test_roles.py
+++ b/tests/backend/realm/test_roles.py
@@ -521,10 +521,8 @@ async def test_update_roles_for_revoked_user(
     realm,
     realm_generate_certif_and_update_roles_or_fail,
     next_timestamp,
-    backend_data_binder_factory,
+    backend_data_binder,
 ):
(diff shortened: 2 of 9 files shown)


parsec/backend/user_type.py
```diff
@@ -54,7 +54,7 @@ def evolve(self, **kwargs):
         return attr.evolve(self, **kwargs)
 
     def is_revoked(self):
-        return self.revoked_on and self.revoked_on <= pendulum.now()
+        return self.revoked_on
```
touilleMan, line 57: 👍

tests/backend/realm/test_roles.py
```diff
@@ -510,3 +510,38 @@ async def test_update_roles_causality_checks(
...
+    backend,
+    alice,
+    bob,
+    alice_backend_sock,
+    realm,
+    realm_generate_certif_and_update_roles_or_fail,
+    next_timestamp,
+    backend_data_binder_factory,
+):
+    binder = backend_data_binder_factory(backend)
```
touilleMan, line 526: you can also directly use `backend_data_binder` fixture

touilleMan (approved):
👌

PR #2544, opened 2022-06-30
Feat/authenticated api server

### LINKED ISSUE #2624: Impl authenticated-api for server asgi

FirelightFlagboy:
The goal is to implement the authenticated part of the api using the http protocol.

# TODO

- [x] Add test on the server side
  - [x] Valid request
  - [x] `Api-Version`
    - [x] Missing header
    - [x] Incompatible version
    - [x] different but compatible version (revision changed)
  - [x] `Authorization`
    - [x] Missing header
    - [x] Invalid value
  - [x] `Author`
    - [x] Missing header
    - [x] `Author` not found in database
  - [x] `Timestamp`
    - [x] Missing header
  - [x] `Signature`
    - [x] Missing header
    - [x] Invalid Signature

# Questions

- [x] Were the `api_version` come from ?
  the `api_version` will be located in the header `API_VERSION`
- [x] Is the `Api-Version` Header received from the user is correctly used ?
  - [x] Need to settle with the api version used 
  - [x] Need to send the `Api-Version` that the serve will be using
- [x] What to do with the timestamp ?
  Wont be used right now
- [x] Should `parsec.api.protocol` have its own `version` module ?

depends on #2338 
closes #2624


parsec/backend/asgi/rpc.py
```diff
@@ -75,3 +86,114 @@ async def anonymous_api(raw_organization_id: str):
...
+    try:
+        organization = await backend.organization.get(org_id)
+        user, device = await backend.user.get_user_with_device(org_id, device_id)
+    except (OrganizationNotFoundError, UserNotFoundError, KeyError) as exc:
+        raise ValueError("User not found: {}", exc)
+
+    if organization.is_expired:
+        raise ValueError("organization is dead :(")
+
+    if user.revoked_on and user.revoked_on <= pendulum.now():
```
FirelightFlagboy, line 179:
```suggestion
    if user.revoked_on
```

parsec/backend/asgi/rpc.py
```diff
@@ -75,3 +86,114 @@ async def anonymous_api(raw_organization_id: str):
...
+
+    verify_key = device.verify_key()
+    if (
+        verify_key.verify(smessage=raw_device_id + raw_timestamp + body, signature=signature_bytes)
+        is None
+    ):
+        raise ValueError("Invalid signed message, can't auth :(")
+
+    client_ctx = AuthenticatedClientContext(
+        api_version=api_version,
```
FirelightFlagboy, line 190: use `_settle_compatible_versions`

https://github.com/Scille/parsec-cloud/blob/b5da56f550a3325173aee55fb7239041735ed1ed/parsec/api/protocol/handshake.py#L269-L271

parsec/backend/asgi/rpc.py
```diff
@@ -75,3 +86,114 @@ async def anonymous_api(raw_organization_id: str):
...
+    cmd = msg.get("cmd")
+    try:
+        if not isinstance(cmd, str):
+            raise KeyError()
+        cmd_func = backend.apis[ClientType.AUTHENTICATED][cmd]
+    except KeyError:
+        return rpc_msgpack_rep({"status": "unknown_command"})
+
+    rep = await cmd_func(client_ctx, msg)
+    return rpc_msgpack_rep(rep)
```
FirelightFlagboy, line 139: add a header `Api-Version` for the settled version used by the server

parsec/backend/asgi/rpc.py
```diff
@@ -75,3 +86,114 @@ async def anonymous_api(raw_organization_id: str):
...
+
+    # Unpack the verified body
+    if request.headers["Content-Type"] != "application/msgpack":
+        abort(400)
+    try:
+        msg = unpackb(body)
+    except SerdePackingError:
+        return rpc_msgpack_rep({"status": "invalid_msg_format"})
+
+    cmd = msg.get("cmd")
```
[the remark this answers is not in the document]
FirelightFlagboy: the sanitization is already done by [`unpackb`](https://github.com/Scille/parsec-cloud/blob/230c588ffd13ed8f0c5f3fa4a265bbac81ae17aa/parsec/serde/packing.py#L55-L77)

parsec/backend/asgi/rpc.py
```diff
@@ -75,3 +86,114 @@ async def anonymous_api(raw_organization_id: str):
...
+
+    if organization.is_expired:
+        raise ValueError("organization is dead :(")
+
+    if user.revoked_on and user.revoked_on <= pendulum.now():
+        raise ValueError("User is dead :(")
+
+    verify_key = device.verify_key()
+    if (
+        verify_key.verify(smessage=raw_device_id + raw_timestamp + body, signature=signature_bytes)
```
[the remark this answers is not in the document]
FirelightFlagboy: done in 60d0bb4cb

parsec/backend/asgi/rpc.py
FirelightFlagboy: done in 1a1d6f46c

parsec/backend/asgi/rpc.py
FirelightFlagboy: done in 61086fb60

oxidation/libparsec/crates/crypto/src/rustcrypto/sign.rs
```diff
@@ -120,14 +120,31 @@ impl VerifyKey {
...
+    pub fn verify_with_signature(
+        &self,
+        raw_signature: &[u8],
+        message: &[u8],
+    ) -> Result<Vec<u8>, CryptoError> {
+        if raw_signature.len() != Signature::BYTE_SIZE {
+            return Err(CryptoError::Signature);
+        }
+        let signature = Signature::try_from(raw_signature)
+            .expect("Precondition all ready check for the signature size");
```
vxgmichel, line 147: already checked

parsec/api/version.py
```diff
@@ -1,14 +1,62 @@
...
+        return cls(version, revision)
+
 
 API_V1_VERSION = ApiVersion(version=1, revision=3)
 API_V2_VERSION = ApiVersion(version=2, revision=7)
 API_V3_VERSION = ApiVersion(version=3, revision=1)
 API_VERSION = API_V3_VERSION
+
+
+class IncompatibleAPIVersionsError(Exception):
```
vxgmichel, line 27: Should inherit from `ProtocolError`

parsec/backend/handshake.py
```diff
@@ -59,7 +60,7 @@ async def do_handshake(
                 backend, handshake
             )
 
-    except ProtocolError as exc:
+    except (ProtocolError, IncompatibleAPIVersionsError) as exc:
```
vxgmichel, line 63: TODO: remove once `IncompatibleAPIVersionsError` inherit from `ProtocolError`

parsec/backend/asgi/rpc.py
```diff
@@ -75,3 +93,171 @@ async def anonymous_api(raw_organization_id: str):
...
+    try:
+        if request.headers["Content-Type"] != "application/msgpack":
+            raise InvalidHeaderValue("Content-Type", request.headers["Content-Type"])
+        user, device = await verify_auth_request(body, request.headers, organization, backend)
+    except (
+        IncompatibleAPIVersionsError,
+        MissingHeaderError,
+        OrganizationNotFoundError,
+        UserNotFoundError,
+        InvalidHeaderValue,
```
vxgmichel, line 139: Is it really necessary to bother with all those exceptions? Wouldn't it be simpler to call `abort` directly where those are raised?

vxgmichel (changes requested):

oxidation/libparsec/crates/crypto/src/rustcrypto/sign.rs
FirelightFlagboy: done

parsec/backend/handshake.py
FirelightFlagboy: added `TODO` comment

parsec/api/version.py
FirelightFlagboy: That would create a `cyclic` dependencies, maybe I should move the exception to `api/protocol/base.py` ?

parsec/backend/asgi/rpc.py
```diff
@@ -75,3 +93,171 @@ async def anonymous_api(raw_organization_id: str):
...
+    ) as e:
+        logger.warning(f"invalid request, cause: {e}")
+        abort(400)
+    except ValueError as e:
```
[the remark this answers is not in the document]
FirelightFlagboy: `ValueError` was used when I haven't declared specific expection for the failures points of the authentication.

in place of `ValueError` we now have `AuthenticationProtocolError`

parsec/backend/asgi/rpc.py
FirelightFlagboy: Now all of these exception are based on the same root exception `HttpProtocolError`, so this discussion is outdated

parsec/api/version.py
FirelightFlagboy: done

parsec/backend/asgi/__init__.py
```diff
@@ -20,7 +20,7 @@
 from parsec.backend.config import BackendConfig
 from parsec.backend.asgi.administration import administration_bp
 from parsec.backend.asgi.redirect import redirect_bp
-from parsec.backend.asgi.rpc import rpc_bp
+from parsec.backend.asgi.rpc import rpc_ano_bp, rpc_auth_bp
```
vxgmichel, line 23: Maybe we need better naming for those blueprints? I don't mind `rpc`, but `rpc_authenticated_blueprint` would look better IMO.

parsec/backend/asgi/error.py
```diff
@@ -0,0 +1,54 @@
...
+    def __str__(self) -> str:
+        return f"User {self.user_id} from {self.organization_id} is revoked"
+
+
+class CannotVerifySignature(AuthenticationProtocolError):
+    def __init__(self, internal_exception: Exception) -> None:
+        self.internal_exception = internal_exception
+
+    def __str__(self) -> str:
+        return f"Could not verify the signature, reason: {self.internal_exception}"
```
vxgmichel, line 54: Nice errors :+1: 
Maybe add an inheritance tree in the module docstring:
```
HttpProtocolError
 + MissingHeaderError
 + InvalidHeaderValue
 + AuthenticationProtocolError
   + DeviceNotFound
   + ...
```
Also some errors have the `Error` suffix and some don't. Maybe have `Error` everywhere?

parsec/backend/asgi/rpc.py
```diff
@@ -75,3 +104,184 @@ async def anonymous_api(raw_organization_id: str):
...
+        msg = unpackb(body)
+    except SerdePackingError:
+        return rpc_msgpack_rep({"status": "invalid_msg_format"})
+
+    cmd = msg.get("cmd")
+    try:
+        if not isinstance(cmd, str):
+            raise KeyError()
+        cmd_func = backend.apis[ClientType.AUTHENTICATED][cmd]
+    except KeyError:
```
vxgmichel, line 166: I'm not sure it makes sense to factorize `if not isinstance(cmd, str):` within this try-except. Maybe perform this check before and have a dedicated warning for this case?

tests/common/backend.py
```diff
@@ -555,6 +556,10 @@ def correct_addr(self, addr: BackendAddr) -> BackendAddr:
     async def connection_factory(self) -> trio.abc.Stream:
         return await trio.open_tcp_stream(self.addr.hostname, self.addr.port)
 
+    def test_client(self, use_cookies: bool = True) -> TestClientProtocol:
+        """Creates and returns a test client"""
+        return self.asgi_app.test_client(use_cookies)
```
vxgmichel, line 561: oooh nice :+1:

tests/backend/asgi/authenticated/test_timestamp.py
```diff
@@ -0,0 +1,38 @@
...
+    request, _reply_parser, _expected_reply = dummy_request_with_reply()
+
+    signature = signing_request(alice.signing_key, dev_id, timestamp, request)
+
+    headers = (
+        AuthenticatedHeadersBuilder(api_version=API_VERSION)
+        .with_author(dev_id)
+        .with_signature(signature)
+        .build()
+    )
```
vxgmichel, line 35: Looks like someone is having fun with the builder pattern :grin:

tests/backend/asgi/authenticated/utils.py
```diff
@@ -0,0 +1,116 @@
...
+    timestamp: Optional[DateTime] = None
+    signature: Optional[bytes] = None
+    author: Optional[DeviceID] = None
+
+    def __init__(
+        self,
+        api_version: Optional[ApiVersion] = API_VERSION,
+        content_type: Optional[str] = "application/msgpack",
+        authorization: Optional[str] = "PARSEC-SIGN-ED25519",
+    ) -> None:
```
vxgmichel, line 74: It's unclear why some attributes are set with the regular constructor and why others are set with the builder pattern.

Why not simply have a builder function?
```python
def build_authenticated_headers(
    api_version: Optional[ApiVersion] = API_VERSION,
    content_type: Optional[str] = "application/msgpack",
    authorization: Optional[str] = "PARSEC-SIGN-ED25519",
    timestamp: Optional[DateTime] = None,
    signature: Optional[bytes] = None,
    author: Optional[DeviceID] = None,
) -> Headers:
 [...]

**Review by vxgmichel — APPROVED** [review]
Very nice PR, and great test coverage ! 

LGTM :+1: 

A couple of minor nitpicks though.

File: parsec/backend/asgi/__init__.py (hunk 11 above)
Comment (FirelightFlagboy) [reply parsec/backend/asgi/__init__.py:23]: done

File: parsec/backend/asgi/error.py (hunk 12 above)
Comment (FirelightFlagboy) [reply parsec/backend/asgi/error.py:54]: I've added the suffix `Error` to those exceptions

File: parsec/backend/asgi/__init__.py (hunk 11 above)
Comment (touilleMan) [reply parsec/backend/asgi/__init__.py:23]: we don't need multiple blueprints here, `rpc_bp` is good enough

Blueprint is a tool to do deferred initialization on the app object, we don't use to do code separation (we split the code between files for that !)

**Review by touilleMan — CHANGES_REQUESTED** [review]

File: parsec/backend/asgi/error.py (hunk 12 above)
Comment (FirelightFlagboy) [reply parsec/backend/asgi/error.py:54]: Adding the _inheritance_ tree in the doc-string would be an extra cost to maintain, I prefer not to.
Modern IDE allow to show references of a item (on `vscode` => `peek references`)

File: tests/backend/asgi/authenticated/utils.py [hunk 17]
```diff
@@ -0,0 +1,116 @@
...
+
+
+class AuthenticatedHeadersBuilder:
+    timestamp: Optional[DateTime] = None
+    signature: Optional[bytes] = None
+    author: Optional[DeviceID] = None
+
+    def __init__(
+        self,
+        api_version: Optional[ApiVersion] = API_VERSION,
```
touilleMan: The `Optional` on those params means that the caller can pass `None` ?

(my guess is those params shouldn't be marked Optional as they all have a default value)

parsec/backend/asgi/rpc.py
FirelightFlagboy: `mypy` _require_ this check to validate that `cmd` is indeed a string

FirelightFlagboy: I've added more context to the raised exception, this would help with the logger just below in the `catch`

tests/backend/asgi/authenticated/utils.py
touilleMan: Or `None` stands for "don't add the corresponding header to the HTTP request" ?

parsec/backend/asgi/rpc.py
```diff
@@ -1,27 +1,38 @@
...
 
 
 def rpc_msgpack_rep(data: dict) -> Response:
     return Response(
         response=packb(data),
-        # Unlik REST, RPC doesn't use status to encode operational result
+        # Unlike REST, RPC doesn't use status to encode operational result
         status=200,
         content_type="application/msgpack",
     )
```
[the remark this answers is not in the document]
FirelightFlagboy: done

tests/backend/asgi/authenticated/utils.py
FirelightFlagboy: done

tests/backend/asgi/authenticated/test_api_version.py
```diff
@@ -0,0 +1,110 @@
...
+
+@pytest.mark.trio
+async def test_api_version_missing_header(running_backend: RunningBackend, alice: LocalDevice):
+    client = running_backend.test_client()
+
+    org_id = alice.organization_id
+    dev_id = alice.device_id
+    timestamp = DateTime.now()
+
+    request, _reply_parser, _expected_reply = dummy_request_with_reply()
```
touilleMan, line 26: use `_` for unused params: `request, _reply_parser, _ = dummy_request_with_reply()`

parsec/backend/asgi/error.py
vxgmichel: > Modern IDE allow to show references of a item

Except it's not a clear at first glance how the exceptions are structured. We have one those tree for pki exceptions already:
https://github.com/Scille/parsec-cloud/blob/890490c21d801fdc2090ec2a4ce4d3ecd159992e/parsec/core/pki/exceptions.py#L3-L41

But I won't fight for it either, it's up to you.

tests/backend/asgi/authenticated/test_api_version.py
```diff
@@ -0,0 +1,110 @@
...
+
+    signature = signing_request(alice.signing_key, dev_id, timestamp, request)
+
+    headers = build_authenticated_headers(
+        api_version=None, author=dev_id, timestamp=timestamp, signature=signature
+    )
+
+    rep = await send_authenticated_request(client, org_id, headers, request)
```
touilleMan, line 34: This code is too verbose and low level, instead you should do create a authenticated htttp client class with fixture to use it:

```python
class AuthenticatedHttpApiClient:
    def __init__(self, client: TestClientProtocol, device: LocalDevice):
        self.client = client
        self.device = device
        self.base_headers = {
            "Content-Type": "application/msgpack",
            "Api-Version": "PARSEC-SIGN-ED25519",
            [...]
            "Author": b64encode(device.device_id.encode("utf8")),
        }

    async def send_ping(self, ping, **kwargs):
        return await self.send({"cmd": "ping", "ping": ping}, ping_serializer, **kwargs)

    async def send(self, req, serializer, extra_headers: Dict[str, str]= {}, check_rep: bool = True):
        now = DateTime.now()
        body = serializer.req_dumps(req)
        headers = {
            **self.base_headers,
            "Timestamp": DateTime.now(),
            "Signature": signing_request(self.device.signing_key, self.device.device_id, now, body)
        }

        # Customize headers
        for k, v in extra_headers.items():
            if v is None:
                headers.pop(k, None)
            else:
                headers[k] = v

        rep = await self.client.post(f"/authenticated/{self.device.organization_id}", headers=headers, data=body)

        if check_rep:
            assert rep.status_code == 200
            rep_body = await rep.body()
            return serializer.rep_loads(rep_body)

        else:
            return rep

@pytest.fixture
def alice_http_client(alice, running_backend):
    client = running_backend.test_client()
    return AuthenticatedHttpApiClient(alice, client)


# and now you can do
# Test ok
rep = await alice_http_client.send_ping("foo")
assert rep == {"status": "ok", "pong": "foo"}

# Test error in transport
rep = await alice_http_client.send_ping("foo", check_rep=False, extra_headers={"timestamp": None})
assert rep.status_code == 400
```

touilleMan (changes requested):

parsec/backend/asgi/rpc.py
```diff
@@ -63,15 +93,194 @@ async def anonymous_api(raw_organization_id: str):
...
+    except AuthenticationProtocolError as e:
+        logger.warning(f"Failed to authenticate request because of {e}")
+        abort(401)
+
+    api_version = get_api_version(request.headers)
+
+    client_ctx = build_client_context(organization, user, device, api_version)
+
+    # Unpack the verified body
+    if request.headers["Content-Type"] != "application/msgpack":
```
touilleMan, line 152: `"application/msgpack"` should be a constant at the top of the file

parsec/backend/asgi/rpc.py
```diff
@@ -63,15 +93,194 @@ async def anonymous_api(raw_organization_id: str):
...
     rep = await cmd_func(client_ctx, msg)
     return rpc_msgpack_rep(rep)
+
+
+@rpc_blueprint.route("/authenticated/<raw_organization_id>", methods=["POST"])
+async def authenticated_api(raw_organization_id: str):
+    backend: BackendApp = g.backend
+    logger = current_app.logger.getChild(authenticated_api.__name__)
+
+    logger.debug(
```
touilleMan, line 116: structlog allow us to do structured logging, so we can do instead:
```python
logger.debug("Received authenticated request", org=raw_organization_id, device=request.headers.get("Author", "<missing>"))
```

touilleMan (changes requested):

parsec/backend/asgi/rpc.py
touilleMan: so according to [hypercorn documentation](https://pgjones.gitlab.io/hypercorn/how_to_guides/logging.html#how-to-log), it should be enough to use the `{Header}i` format in the access log format: https://github.com/Scille/parsec-cloud/blob/890490c21d801fdc2090ec2a4ce4d3ecd159992e/parsec/backend/asgi/__init__.py#L126

Given we already have the organization id in the endpoint, we only need to provide the device id by doing something like:
` "access_log_format": "%(h)s %(r)s %(s)s %(b)so %(D)sus (device_id:%({Author}i)s)",`

However this is not great given we don't return the actual DeviceID but a base64 of the DeviceID :(
On top of that all request unrelated to the authenticated route will have an ugly `(device_id:)` in the log...

The thing is we really want to have the device id in the log given it's really useful we looking at the log to see the different queries a given client has done...

touilleMan: This is getting too complex for this PR, can you:
- open an "provide device id in server access log for authenticated http request" issue
- remove all the logs in here (we don't need that level of debug)

tests/backend/asgi/authenticated/utils.py
FirelightFlagboy: > `None` stands for "don't add the corresponding header to the HTTP request" ?

Yes

parsec/backend/asgi/__init__.py
```diff
@@ -20,7 +20,7 @@
 from parsec.backend.config import BackendConfig
 from parsec.backend.asgi.administration import administration_bp
 from parsec.backend.asgi.redirect import redirect_bp
-from parsec.backend.asgi.rpc import rpc_bp
+from parsec.backend.asgi.rpc import rpc_blueprint
```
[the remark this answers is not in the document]
FirelightFlagboy: I can revert the change, but I need an arbitration on
> If we use the short suffix `bp` or long suffix `blueprint`

see related [discussion](https://github.com/Scille/parsec-cloud/pull/2544#discussion_r954963369)

parsec/backend/asgi/rpc.py
```diff
@@ -63,15 +93,194 @@ async def anonymous_api(raw_organization_id: str):
...
+        "Received call in the authenticated api for the organization {org_id} from the device={dev_id}".format(
+            org_id=raw_organization_id, dev_id=request.headers.get("Author", "Missing")
+        )
+    )
+    try:
+        organization_id = OrganizationID(raw_organization_id)
+    except ValueError:
+        logger.warning(f"Invalid organization id: {raw_organization_id}")
```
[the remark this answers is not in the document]
FirelightFlagboy: done

parsec/backend/asgi/rpc.py
FirelightFlagboy: I've removed all the logs, I let you open the issue because you have all the details

parsec/backend/asgi/rpc.py
FirelightFlagboy: done

tests/backend/asgi/authenticated/test_api_version.py
FirelightFlagboy: done

tests/backend/asgi/authenticated/utils.py
```diff
@@ -0,0 +1,93 @@
...
+        headers["Content-Type"] = content_type
+
+    if api_version is not None:
+        headers["Api-Version"] = api_version
+
+    if authorization is not None:
+        headers["Authorization"] = authorization
+
+    if author is not None:
+        headers["Author"] = b64encode(str(author).encode())
```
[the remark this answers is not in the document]
FirelightFlagboy: Just to remember, why we have 2 method (without `repr`) to represent `DeviceID` ? shouldn't 1 be enough ?

parsec/backend/asgi/__init__.py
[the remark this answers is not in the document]
touilleMan: >  I need an arbitration on

follow KIS and keep `bp` ;-)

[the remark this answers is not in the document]
FirelightFlagboy: done

tests/backend/asgi/authenticated/test_api_version.py
FirelightFlagboy: done

parsec/backend/asgi/rpc.py
```diff
@@ -63,15 +93,194 @@ async def anonymous_api(raw_organization_id: str):
...
+        return rpc_msgpack_rep({"status": "invalid_msg_format"})
+
+    cmd = msg.get("cmd")
+    try:
+        if not isinstance(cmd, str):
+            raise KeyError(f"`cmd` type is not a string but a {type(cmd).__name__}")
+        cmd_func = backend.apis[ClientType.AUTHENTICATED][cmd]
+    except KeyError as e:
+        logger.warning(f"Authenticated command {cmd} doesn't exist ({e})")
+        return rpc_msgpack_rep({"status": "unknown_command"})
```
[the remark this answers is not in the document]
FirelightFlagboy: done

tests/backend/asgi/authenticated/utils.py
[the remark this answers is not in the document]
touilleMan: `repr` is for debugging representation
`str` is for pretty user-facing representation

So while `str(x)` might be read as "cast x into a string" but in fact it is more like "give me a representation of x for user facing".

Now the next question is what is "user facing" and that's the trick: it depends.
For instance in the GUI we may want to represent `DeviceID("foo@bar")` as `foo@bar` or maybe as `foo working from bar`. To implement that the dev working on the GUI will of course modify the `__str__` method given it's the right way to go... and break the application ;-)

So the explicit approach is better here: with an `.str` attribute there is no way the GUI dev is going to think "ha ! I can change that arbitrary attribute without breaking the something in the code"

parsec/backend/asgi/rpc.py
touilleMan: I've created #2953

FirelightFlagboy: :+1:

tests/backend/asgi/authenticated/utils.py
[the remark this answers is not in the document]
FirelightFlagboy: That explain why we have `str` but I'm still missing pieces why we have `__str__` defined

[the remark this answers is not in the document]
touilleMan: > why we have __str__ defined

it is a shorthand for the GUI, but I agree it's misleading and we should remove it and instead rely on `<my_obj>.str` everywhere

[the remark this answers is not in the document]
FirelightFlagboy: Should I create an issue for that ?

parsec/backend/asgi/rpc.py
```diff
@@ -63,15 +95,173 @@ async def anonymous_api(raw_organization_id: str):
...
+
+    verify_key = device.verify_key
+
+    try:
+        # when the verify failed, `verify()` raise the exception `BadSignatureError`
+        verify_key.verify_with_signature(
+            signature=signature,
+            message=base64.b64encode(str(device_id).encode()) + raw_timestamp + body,
+        )
+    except Exception as e:
```
touilleMan, line 247: Don't catch global `Exception` but the actual type of exception you know are possible

parsec/backend/asgi/rpc.py
```diff
@@ -63,15 +95,173 @@ async def anonymous_api(raw_organization_id: str):
...
+async def verify_auth_request_ed25519(
+    body: bytes,
+    device_id: DeviceID,
+    raw_timestamp: bytes,
+    signature: bytes,
+    organization: Organization,
+    backend: BackendApp,
+) -> Tuple[User, Device]:
+    try:
+        user, device = await backend.user.get_user_with_device(
```
touilleMan, line 230: This can raises a `UserNotFoundError` which seems related to `DeviceNotFoundError` but it's not the case

So I'd rather have a catch on `UserNotFoundError` and raise a `DeviceNotFoundError` instead. This way the caller only have to deal with `DeviceNotFoundError`.

parsec/backend/asgi/rpc.py
```diff
@@ -63,15 +95,173 @@ async def anonymous_api(raw_organization_id: str):
...
+            organization.organization_id, device_id
+        )
+    except KeyError:
```
touilleMan, line 233: In which case can we have a `KeyError` ? (`get_user_with_device` should not raise such a common exception)

oxidation/libparsec/crates/crypto/src/rustcrypto/sign.rs
```diff
@@ -120,14 +120,31 @@ impl VerifyKey {
...
+    pub fn verify_with_signature(
+        &self,
+        raw_signature: &[u8],
+        message: &[u8],
+    ) -> Result<Vec<u8>, CryptoError> {
+        if raw_signature.len() != Signature::BYTE_SIZE {
+            return Err(CryptoError::Signature);
+        }
+        let signature = Signature::try_from(raw_signature)
+            .expect("Precondition all ready checked for the signature size");
```
touilleMan, line 147:
```suggestion
            .expect("Precondition already checked for the signature size");
```

oxidation/libparsec/crates/crypto/src/rustcrypto/sign.rs
```diff
@@ -120,14 +120,31 @@ impl VerifyKey {
...
+        self.verify_with_signature(
+            &signed[..Signature::BYTE_SIZE],
+            &signed[Signature::BYTE_SIZE..],
+        )
+    }
+
+    /// Verify a signature using the given [VerifyKey], `signature` and `message`
+    pub fn verify_with_signature(
+        &self,
+        raw_signature: &[u8],
```
touilleMan, line 140: shouldn't we have something like `raw_signature: &[u8:Signature::BYTE_SIZE]` ?

parsec/backend/asgi/rpc.py
FirelightFlagboy: done

parsec/backend/asgi/rpc.py
FirelightFlagboy: You listed the reason on https://github.com/Scille/parsec-cloud/pull/2544#discussion_r957336101
So I'll catch `UserNotFound` instead

parsec/backend/asgi/rpc.py
FirelightFlagboy: done

tests/common/backend.py
```diff
@@ -555,6 +565,109 @@ def correct_addr(self, addr: BackendAddr) -> BackendAddr:
...
+
+
+class AuthenticatedHttpApiClient:
```
touilleMan, line 573: Replace this by the class I suggested in https://github.com/Scille/parsec-cloud/pull/2544#discussion_r955070416

this has the added value of turning this:
```python
@pytest.mark.trio
async def test_signature_missing_header(alice_http_client: AuthenticatedHttpApiClient):
   timestamp = DateTime.now()

    request = alice_http_client.get_request()

    headers = alice_http_client.build_authenticated_headers(
        api_version=API_VERSION,
        author=alice_http_client.device_id,
        timestamp=timestamp,
    )

    rep = await alice_http_client.send_request(headers, request)
    assert rep.status_code == 400  # bad request
```

into this:

```python
@pytest.mark.trio
async def test_signature_missing_header(alice_http_client: AuthenticatedHttpApiClient):
    rep = await alice_http_client.send_ping(extra_headers={"Signature": None})
    assert rep.status_code == 400  # bad request
```

touilleMan: For more advanced customization of the headers, we should add special params to the send:

```python
class AuthenticatedHttpApiClient:
    [...]
    # now and before_send_hook for easy customization
    async def send(self, req, serializer, extra_headers: Dict[str, str]= {}, before_send_hook=None, now=None, check_rep: bool = True):
        now = now or DateTime.now()
        body = serializer.req_dumps(req)
        headers = {
            **self.base_headers,
            "Timestamp": now.to_rfc(),
            "Signature": signing_request(self.device.signing_key, self.device.device_id, now, body)
        }

        # Customize headers
        for k, v in extra_headers.items():
            if v is None:
                headers.pop(k, None)
            else:
                headers[k] = v

        args = {"url": f"/authenticated/{self.device.organization_id}", "headers": headers, "data": body}
        # Last chance to customize the request !
        if before_send_hook:
            # Passing as dict allow the hook to event modify the url param
            before_send_hook(args)
        rep = await self.client.post(**args)

        if check_rep:
            assert rep.status_code == 200
            rep_body = await rep.body()
            return serializer.rep_loads(rep_body)

        else:
            return rep
```
then

```python
@pytest.mark.trio
async def test_signature_invalid(alice_http_client: AuthenticatedHttpApiClient):
    timestamp = DateTime.now()

    request = alice_http_client.get_request()

    signature = alice_http_client.sign_request(timestamp.add(minutes=1), request)

    headers = alice_http_client.build_authenticated_headers(
        api_version=API_VERSION,
        author=alice_http_client.device_id,
        timestamp=timestamp,
        signature=signature,
    )

    rep = await alice_http_client.send_request(headers, request)
    assert rep.status_code == 401  # Unauthorized
```

becomes

```python
@pytest.mark.trio
async def test_signature_invalid(alice_http_client: AuthenticatedHttpApiClient):
    def _before_send_hook(args):
        args["headers"]["signature"] = "dummy"
    alice_http_client.send_ping(before_send_hook=_before_send_hook)
    assert rep.status_code == 401  # Unauthorized

# Or with a valid but in the future signature:
@pytest.mark.trio
async def test_signature_invalid(alice_http_client: AuthenticatedHttpApiClient):
    alice_http_client.send_ping(now=DateTime.now().add(minutes=1))
    assert rep.status_code == 401  # Unauthorized
```

touilleMan (changes requested):

oxidation/libparsec/crates/crypto/src/rustcrypto/sign.rs
FirelightFlagboy: Nop, that's a slice, what you're referring to is an array

oxidation/libparsec/crates/crypto/src/rustcrypto/sign.rs
FirelightFlagboy: done

parsec/backend/asgi/error.py
```diff
@@ -0,0 +1,54 @@
...
+        return f"Missing header {self.header}"
+
+
+class InvalidHeaderValueError(HttpProtocolError):
+    def __init__(self, header: str, invalid_value: str) -> None:
+        self.header = header
+        self.invalid_value = invalid_value
+
+    def __str__(self) -> str:
+        return f"Invalid header value for header {self.header}"
```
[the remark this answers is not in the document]
FirelightFlagboy: done

tests/backend/asgi/authenticated/utils.py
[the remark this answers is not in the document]
FirelightFlagboy: @touilleMan up

tests/common/backend.py
FirelightFlagboy: done

parsec/backend/asgi/error.py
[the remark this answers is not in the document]
touilleMan: it should be done in the other exceptions too

tests/backend/asgi/authenticated/utils.py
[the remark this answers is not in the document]
touilleMan: yup create an issue and we will discuss it next meeting

parsec/backend/asgi/rpc.py
```diff
@@ -42,56 +31,172 @@
...
+async def _do_handshake(
+    raw_organization_id: str,
+    backend: BackendApp,
+    allow_missing_organization: bool,
+    check_authentication: bool,
+) -> Tuple[ApiVersion, OrganizationID, Optional[Organization], Optional[User], Optional[Device]]:
+    # The anynimous HTTP API existed before the `Api-Version`/`Content-Type` fields
+    # where introduced, hence we have this workaround to provide backward compatibility
+    request.headers.setdefault("Api-Version", "3.0")
+    request.headers.setdefault("Content-Type", CONTENT_TYPE_MSGPACK)
```
FirelightFlagboy, line 83: Do we also want to add those headers in authenticated api ?
Maybe move them to `anonymous_api` ?

tests/backend/test_rpc_handshake.py
```diff
@@ -0,0 +1,243 @@
...
+        rep = await client.send_ping(
+            extra_headers={"Api-Version": None},
+            check_rep=False,
+        )
+        # Special case for anynous
+        assert rep.status_code == 200
+        rep.headers["Api-Version"] == "3.0"
+
+        # Client provides an incompatible api version
+        unknown_major = ApiVersion(version=5, revision=0)
```
FirelightFlagboy, line 69:
```suggestion
        unknown_major = ApiVersion(version=API_VERSION+5, revision=0)
```

tests/backend/test_rpc_handshake.py
```diff
@@ -0,0 +1,243 @@
...
+        too_old_major = ApiVersion(version=1, revision=0)
```
FirelightFlagboy, line 70:
```suggestion
```

tests/backend/test_rpc_handshake.py
```diff
@@ -0,0 +1,243 @@
...
+
+async def _test_authenticated_handshake_bad_signature_header(
+    client: AuthenticatedRpcApiClient, alice: LocalDevice, bob: LocalDevice
+):
+    # First test missing & plain bad headers
+    for expected_status_code, extra_headers in [
+        (401, {"Timestamp": None}),  # Missing Timestamp header
+        (401, {"Timestamp": "dummy"}),  # Bad Timestamp header
+        (401, {"Signature": None}),  # Missing Signature header
+        (401, {"Signature": "dummy"}),  # Bad Signature header
```
FirelightFlagboy, line 117:
```suggestion
        (401, {"Signature": b64encode(d"dummy")}),  # Bad Signature header
```

tests/backend/test_rpc_handshake.py
touilleMan: This test is independent of the actual versions supported by the server (cf. the `with patch(...)`)

tests/backend/test_rpc_handshake.py
touilleMan: This test is independent of the actual versions supported by the server (cf. the `with patch(...)`)

parsec/backend/asgi/rpc.py
touilleMan: At first I only patched the anonymous api, but it was cumbersome to test (special cases are never fun to work with !)

So I consider having the authenticated route slightly more forgiving than needed was a good tradeoff given the code&test simplification ;-)

tests/backend/test_rpc_handshake.py
touilleMan: the point is to have a plain wrong header, but also testing "half wrong" (with valid base64 but invalid content) is a good idea ^^

I've added this testcase

tests/common/rpc_api.py
```diff
@@ -0,0 +1,136 @@
...
+from tests.common import RunningBackend, OrganizationFullData
+
+
+class AuthenticatedRpcApiClient:
+    def __init__(self, client: TestClientProtocol, device: LocalDevice):
+        self.client = client
+        self.device = device
+        self.base_headers = Headers(
+            {
+                "Content-Type": "application/msgpack",
```
FirelightFlagboy, line 23:
```suggestion
                "Content-Type": CONTENT_TYPE_MSGPACK,
```

tests/common/rpc_api.py
```diff
@@ -0,0 +1,136 @@
...
+                "Api-Version": str(API_VERSION),
+                "Authorization": "PARSEC-SIGN-ED25519",
```
FirelightFlagboy, line 25:
```suggestion
                "Authorization": AUTHORIZATION_PARSEC_ED25519,
```

tests/common/rpc_api.py
```diff
@@ -0,0 +1,136 @@
...
+            return rep
+
+
+class AnonymousRpcApiClient:
+    def __init__(self, organization_id: OrganizationID, client: TestClientProtocol):
+        self.organization_id = organization_id
+        self.client = client
+        self.base_headers = Headers(
+            {
+                "Content-Type": "application/msgpack",
```
FirelightFlagboy, line 81:
```suggestion
                "Content-Type": CONTENT_TYPE_MSGPACK,
```

tests/common/rpc_api.py
touilleMan: can you revert those changes ? I think it's better to keep those value separated from the constant in the codebase

The idea is to be able to detect a backward incompatibility if we change by mistake the constants

tests/common/rpc_api.py
touilleMan: same

tests/common/rpc_api.py
touilleMan: same

tests/common/rpc_api.py
FirelightFlagboy: done

touilleMan (approved):

PR #2545, opened 2022-06-30
[GUI] Fixed snackbar visibility when window is not in focus and how snackbars are handled

### LINKED ISSUE #2530: [🐛 | Bug]: Snackbar is displayed when Parsec is hidden
### Version

2.10.0rc2

Windows

### What happened ?

Parsec was hidden when someone shared a workspace with me, the snackbar notification was displayed right in the middle of the screen. It happens both if Parsec is minimized or hidden by another window (it should only happen when the window is in focus).

### Relevant output

_No response_

Max-7:
Better fix for #2530 

Snackbars now have the MainWindow as a parent, meaning they will more closely follow its state instead of appearing above every other windows on Windows.

Also, the snackbars were stored in a global variable. This caused problems, especially while testing, because the C++ Qt object would be deleted when the window was closed at the end of a test while the Python object would stay alive. Snackbars are now stored in an object whose lifetime depends on the main window.

closes #2530

### DIFF
--- a/parsec/core/gui/main_window.py
+++ b/parsec/core/gui/main_window.py
@@ -11,7 +11,6 @@ from PyQt5.QtWidgets import QWidget, QMainWindow, QMenu, QShortcut, QMenuBar
 
 from parsec import __version__ as PARSEC_VERSION
 from parsec.core.gui.enrollment_query_widget import EnrollmentQueryWidget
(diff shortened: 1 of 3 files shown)


Ironicbay (approved):
Great change 👍

PR #2593, opened 2022-07-05
Yolo/fix/ga qa skip news fragment on main branch

FirelightFlagboy:


.github/workflows/quality-assurance.yml
```diff
@@ -28,14 +28,12 @@ jobs:
       - name: Cache Rust
         id: cache-rust
         env:
-          cache-version: 2
+          cache-version: 3
```
touilleMan, line 31: Is this related with the issue ?

.github/workflows/quality-assurance.yml
```diff
@@ -67,14 +65,19 @@ jobs:
...
       - name: Check Newsfragment
         if: |
-          ${{
-            startsWith(github.ref, 'refs/pull/')
-            && !(startsWith(github.ref_name, 'dependabot')
-              || startsWith(github.ref_name, 'yolo')
-              || startsWith(github.ref_name, 'release'))
-          }}
+          startsWith(github.ref, 'refs/pull/')
+          && !(
```
touilleMan, line 76: Given we already do the release branch check within the script, shouldn't we also migrate all those checks in the scripts ? (maybe only leaving the `startsWith(github.ref, 'refs/pull/')`)

FirelightFlagboy: we could add them to the script but I would not remove them from the conditional expression, It faster that executing the script

touilleMan: speed difference is minimal, on the other hand it's better for readability have every branch check at the same place (see my own PR to fix this issue: I only saw the check were done in the script and missed what is in this file)

.github/workflows/quality-assurance.yml
touilleMan: Can you add a comment to explain what this field is for and when it should be updated ?

I though this was related to selecting the version of the github action ^^

vxgmichel (approved):

PR #2595, opened 2022-07-05
Easier header copyright

### LINKED ISSUE #2596: Remove copyright year from source headers ?
Let's face it: updating copyright every year is a silly and inconvenient process (heck, I should have done this 5 months ago !)

However Google's open source documentation consider it is enough to only write the year of creation of the file:

https://opensource.google/documentation/reference/copyright#copyright_notice

This is confirmed by Golang project, for instance (copyright header still indicate 2016, but last modification one month old):

https://github.com/golang/go/blob/0ab71cc065c0ce70d7df8bf498723b5a1c7a89c1/src/crypto/ed25519/ed25519.go#L1

Facebook on it side seems to use the famous `2016-present` hack in the copyright:

https://github.com/facebookincubator/SocketRocket/blob/6f61b5437eaf425e073309f7252c8fa5ea908311/Configurations/SocketRocket-iOS.xcconfig#L2

MariaDB also only indicate the year of creation in their BUSL licensed code:

https://github.com/mariadb-corporation/MaxScale/blob/1231be86fee6c935f4e09e7caa6d917fb1a22222/server/core/admin.cc#L2

So I would say we should be fine by only putting the year of creation in the source code ;-)

touilleMan:
fix #2596


misc/license_headers.py
```diff
@@ -56,88 +63,128 @@ def check_header(cls, file: Path) -> bool:
...
                 return True
 
         return False
 
+    @classmethod
+    def remove_header(cls, file: Path) -> bool:
+        with open(file, "r", encoding="utf-8") as fd:
+            lines = []
+            need_rewrite = False
+            for line in fd.readlines():
```
Max-7, line 80: Should only check for a header for the first few lines, not the whole file. It would cause a problem if, for example, you were to launch the script on this very file since the function `is_possible_license_line` would itself be detected as a license header line and be removed.

Max-7 (changes requested):

misc/license_headers.py
touilleMan: > you were to launch the script on this very file [...]

There is a special case in the file filter to handle this ;-)

```python
    # Special case for ourself given we contain the license headers in the source code !
    re.compile(r"^misc/license_headers.py$"): SkipLicenser,
```

For the rest of the codebase, `is_possible_license_line` checks for `Parsec Cloud (https://parsec.cloud) Copyright (c)` which is highly unlikely to be present (I've tried it on the codebase and only headers are removed)

The thing is this `remove` command is a helper to modify the header line (like what we are going to do just before merging this PR ^^), it is not expected to be run as part of the CI or anything that will go blindly on the codebase

Max-7 (approved):

PR #2613, opened 2022-07-07
Fix/2611 pre commit GitHub action dont correctly work with the cache

### LINKED ISSUE #2611: Pre-commit github action don't correctly work with the cache
GA `pre-commit` don't correctly use the cache, it don't cache the pre-commit binary

### LINKED ISSUE #2617: rust-cache using older rustc version
when we've switched to rust 1.62 from 1.61 the cache are very often poisoned with the older version.
this break the compilation because we now require rust-1.62.

FirelightFlagboy:
closes #2611 
closes #2617


.github/workflows/quality-assurance.yml
```diff
@@ -32,32 +32,35 @@ jobs:
...
-        run: echo ${{ steps.cache-rust.outputs.cache-hit }}
-
-      - if: steps.cache-rust.outputs.cache-hit != 'true'
-        uses: actions-rs/toolchain@b2417cde72dcf67f306c0ae8e0828a81bf0b189f # pin v1.0.6
+      - uses: actions-rs/toolchain@b2417cde72dcf67f306c0ae8e0828a81bf0b189f # pin v1.0.6
         with:
           toolchain: ${{ env.rust-version }}
           components: rustfmt, clippy
           default: true
+          override: true
```
touilleMan, line 50: Override is used to bypass cache ? can you add a small comment about this

touilleMan (approved):

.github/workflows/quality-assurance.yml
FirelightFlagboy: not really, override

> Set installed toolchain as an override for the current directory
> https://github.com/actions-rs/toolchain#inputs

PR #2625, opened 2022-07-08
Split decode and execute

TimeEngineer:


FirelightFlagboy (approved):
now when doing `Cmd::decode` you can test if the decoded variant is correct :)

PR #2631, opened 2022-07-12
More mypy fixes

vxgmichel:


parsec/backend/asgi/rpc.py
```diff
@@ -1,6 +1,6 @@
 # Parsec Cloud (https://parsec.cloud) Copyright (c) BSLv1.1 (eventually AGPLv3) 2016-2021 Scille SAS
 
-from quart import current_app, Response, Blueprint, abort, request
+from quart import Response, Blueprint, abort, request, g
```
FirelightFlagboy, line 3: Why using `g` instead of `current_app` ?

touilleMan (approved):

parsec/backend/asgi/rpc.py
vxgmichel: The initial motivation was fixing mypy and it turns out that it's the way quart expects us to share objects between tasks within the same app.

PR #2638, opened 2022-07-13
rename generated artifact by libparsec python

### LINKED ISSUE #2637: `libparsec` and `libparsec_python` are creating artifact with the same name
rust crates `oxidation/libparsec` and `oxidation/libparsec_python` are creating artifact with the same name because of `libparsec_python` being configured as follow

```toml
[lib]
name = "libparsec"
# "cdylib" is necessary to produce a shared library for Python to import from.
crate-type = ["cdylib"]
```

I we were to use the lib `libparsec` for something other than binding to rust, right now we would be using the lib generated by `libparsec_python`

FirelightFlagboy:
closes #2637

### DIFF
--- a/oxidation/libparsec_python/Cargo.toml
+++ b/oxidation/libparsec_python/Cargo.toml
@@ -7,7 +7,7 @@ edition = "2021"
 name = "libparsec._libparsec"
 
 [lib]
-name = "libparsec"
+name = "libparsec_py"
 # "cdylib" is necessary to produce a shared library for Python to import from.
 crate-type = ["cdylib"]
 
--- a/oxidation/libparsec_python/Cargo.toml
+++ b/oxidation/libparsec_python/Cargo.toml
@@ -7,7 +7,6 @@ edition = "2021"
 name = "libparsec._libparsec"
 
 [lib]
-name = "libparsec_py"
 # "cdylib" is necessary to produce a shared library for Python to import from.
 crate-type = ["cdylib"]
 
--- a/oxidation/libparsec_python/Cargo.toml
+++ b/oxidation/libparsec_python/Cargo.toml
@@ -7,7 +7,7 @@ edition = "2021"
 name = "libparsec._libparsec"
 
 [lib]
-# "cdylib" is necessary to produce a shared library for Python to import from.
+# "cdylib" is necessary to produce a shared library for `maturin` to be used to generate the python module.
 crate-type = ["cdylib"]
 
 [dependencies]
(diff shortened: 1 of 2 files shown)


oxidation/libparsec_python/Cargo.toml
```diff
@@ -7,7 +7,7 @@ edition = "2021"
 name = "libparsec._libparsec"
 
 [lib]
-name = "libparsec"
+name = "libparsec_py"
```
touilleMan, line 10: Don't we need to update the import statement in python code too ?

FirelightFlagboy: `oxidation` tests pass, so 🤷

touilleMan: > `python oxidation` tests pass, so 🤷

Can you test with a build from scratch, just to be extra sure the cache is not playing us here

touilleMan: Also I'm curious: what the name of the .so file that endup in the final libparsec python package? Is it still `libparsec.so` ? (Hence maturin is doing a rename when installing the lib, which would explain why we don't need to change the python code to do the import)

FirelightFlagboy: after doing `pip install oxidation/libparsec_python` `maturin` generate the following wheel file `target/wheels/libparsec-1.0.0-cp37-abi3-linux_x86_64.whl`

touilleMan: Ok si the change you do is only relevant for the build cache which is great ^^

touilleMan (approved):

oxidation/libparsec_python/Cargo.toml
FirelightFlagboy: > > `python oxidation` tests pass, so :shrug:
> 
> Can you test with a build from scratch, just to be extra sure the cache is not playing us here

```
$ poetry shell
$ cargo clean
$ ls target
ls: cannot access 'target': No such file or directory
$ pip install oxidation/libparsec_python
[...]
Successfully installed libparsec-1.0.0
[...]
$ ls target/{release,wheels}
target/release:
build  deps  examples  incremental  liblibparsec_python.d  liblibparsec_python.so

target/wheels:
libparsec-1.0.0-cp37-abi3-linux_x86_64.whl
$ python -m pytest tests --runrust -q
[...]
1352 passed, 395 skipped, 4 xfailed in 29.46s
```

PR #2642, opened 2022-07-15
Feat/2489 maturin based pyproject

### LINKED ISSUE #2489: [Oxidation in prod] Poetry/Maturin based pyproject.toml for main parsec Python project
depends of #2371

## Overview

Once Rust extension are mandatory to run Parsec there would be no need to separate the python bindings from the regular parsec codebase.
In fact this separation makes things more complex for packaging given parsec requires libparsec (i.e. the rust extension) but must specify the relation with a relative path, which is not allowed when building a wheel (On top of that dependency means dealing version number which we also would like to avoid).

So the solution is simply to merge together parsec and libparsec into a single hybrid Python package.

## Architecture

The switch to Poetry is under way (see #2371), once it is done we should be able to add maturin to it by [taking example of this project](https://github.com/ArniDagur/python-adblock/blob/master/pyproject.toml)

## Command workflow

First use:
```
$ poetry install  # Create a new virtual env and install all dependencies within it
$ poetry shell  # Jump into the virtual env
$ maturin develop  # Build the Rust extensions and install them within the parsec source code so that they can be imported
```

First use wth a custom virtual env (letting poetry handle the venv should be the preferred way, but this might come handy):
```
$ python -m venv my_venv
$ . ./my_venv/bin/activate
$ poetry install  # Poetry detects it is already in a venv and reuse it
$ maturin develop
```

After that, switching into another branch (considering the venv is already activated):
```
$ poetry install && maturin develop
```

## Rust Release vs Debug build ?

In theory we want to have a debug build when doing maturin develop, but switch to release when installing the project with `pip install .` (or when building the wheel, which is equivalent since `pip install` build a wheel behind the scene).
So we should make sure this is the case !

FirelightFlagboy:
## How to test

```shell
poetry install
poetry run maturin develop -m oxidation/libparsec_python
poetry run pytest --runrust
```

## Todos

- [x] Revert `__init__.py` to import *unstable rust code*
- [x] Revert worflow `linux-ci` to test *unstable rust code*
- [x] Rename `IS_OXIDIZED` to `UNSTABLE_OXIDATION`
- [x] Migrate `build maturin` in `build.py`
- [x] Migrate `generate pyqt` in `build.py`
- [x] Config `cibuildwheel` to generate python wheel
- [x] Configure Maturin for `linux-ci`
- [x] Configure Maturin for `windows-ci`
- [x] Configure Maturin for `macos-ci`
- [x] Configure Maturin for `packaging-linux`
- [x] Configure Maturin for `packaging-windows`
- [x] Configure Maturin for `packaging-macos` 

## Questions

- [x] To remove the use of the option `-m <manifest_path>`, It seems that we would need to move `libparsec_python` or `parsec` code entierly to have something that respect the [Maturin project layout](https://maturin.rs/project_layout.html).
  I've tried to configure the build of maturin with `tool.maturin.manifest-path` but with no success (maybe it's a bug from `maturin`)
- [x] How to make `wheel-it` generate wheel with the pattern `parsec_cloud-2.10.0+dev.44a5606-py3-linux-x86_64.whl` if we were on `linux-x86_64` ?

closes #2489


FirelightFlagboy:
blocked by #2693

Cargo.toml
```diff
@@ -1,3 +1,15 @@
+[package]
+name = "parsec"
+version = "0.1.0"
```
touilleMan, line 3: Are we obliged to provide a version field here ?

I would be simpler not to (given the version is already handling in `parsec/_version.py`)

Alternatively we could just set `version = "0.0.0"` with a comment explaining this is a dummy version because it is mandatory

Last possibility is to modify `releaser.py` to update this, but it's much more complicated for no real benefit so we should avoid this

FirelightFlagboy: done, rust require the field `version`

So the field is set to `0.0.0` with a comment

FirelightFlagboy:
closes by #2778 supersede this PR

PR #2657, opened 2022-07-20
Migration MacOS CI to Github Action python-only part

### LINKED ISSUE #2656: Migrate Mac CI: Migrate Remaining Python test steps

FirelightFlagboy:
closes #2656

### DIFF
--- a/.github/workflows/macos-ci.yml
+++ b/.github/workflows/macos-ci.yml
@@ -43,6 +43,7 @@ jobs:
         run: >
           poetry run pytest
           -n auto
+          -vv
           -x
           --log-level=DEBUG
           --durations=10
--- a/.github/workflows/macos-ci.yml
+++ b/.github/workflows/macos-ci.yml
@@ -40,4 +40,9 @@ jobs:
         run: poetry run python misc/generate_pyqt.py
 
       - name: Test python
-        run: poetry run pytest -n auto
+        run: >
+          poetry run pytest
+          -n auto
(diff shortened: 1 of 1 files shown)


.github/workflows/macos-ci.yml
```diff
@@ -0,0 +1,48 @@
...
+        run: poetry install -E core -E backend
+
+      - name: Generate PyQT GUI data
+        run: poetry run python misc/generate_pyqt.py
+
+      - name: Test python
+        run: >
+          poetry run pytest
+          -n auto
+          --log-level=DEBUG
```
touilleMan, line 46: better to put that in a `pytest-args` env config to stay consistent with windows/linux jobs

touilleMan (commented):
What about the messages explaining why we don't do much in the macOS tests ?
 
https://github.com/Scille/parsec-cloud/blob/f040851f7c63f3cff01d3f96aad4366f953d1725/.azure-pipelines.yml#L430-L437

.github/workflows/macos-ci.yml
FirelightFlagboy: not needed for the moment, but will do latter

FirelightFlagboy:
> What about the messages explaining why we don't do much in the macOS tests ?

As you see the tests are working with `-n auto` and we're not on `azure-pipeline` so for me the comment is out-of-date

touilleMan (approved):

PR #2665, opened 2022-07-21
Feat/2661 migrate packaging linux step to GitHub action

FirelightFlagboy:


.github/workflows/package-linux.yml
```diff
@@ -0,0 +1,125 @@
...
+      - "[0-9]+.[0-9]+"
+
+env:
+  rust-version: 1.62.0
+  python-version: "3.9"
+  poetry-version: 1.1.13
+
+jobs:
+  build-wheel:
+    runs-on: ubuntu-22.04
```
vxgmichel, line 18: Maybe use 20.04 like in `linux-ci`?

vxgmichel (approved):

.github/workflows/package-linux.yml
FirelightFlagboy: We would have to rewrite the packaging anyway when we will be including rust code. 
I will let the OS as is

PR #2667, opened 2022-07-21
Add HumanHandle validation to rust implementation

### LINKED ISSUE #2666: Missing validation for rust implementation of `HumanHandle`
This could be used as a substitute to `email.utils.parseaddr` from python: https://docs.rs/email-format/0.8.0/email_format/rfc5322/types/struct.NameAddr.html

vxgmichel:
Fix #2666

### DIFF
--- a/oxidation/libparsec/crates/types/src/id.rs
+++ b/oxidation/libparsec/crates/types/src/id.rs
@@ -195,21 +195,22 @@ impl HumanHandle {
         let email = email.nfc().collect::<String>();
         let label = label.nfc().collect::<String>();
 
-        // TODO: how to check the email  easily ?
-        if email.is_empty() || email.len() >= 255 {
+        if !EmailAddress::is_valid(&email, None) {
             return Err("Invalid email address");
         }
-        if label.is_empty() || label.len() >= 255 {
+        if label.is_empty()
+            || label.len() >= 255
+            || label.chars().any(|c| match c {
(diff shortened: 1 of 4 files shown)


touilleMan (approved):
👍

TimeEngineer (approved):

oxidation/libparsec/crates/types/src/id.rs
```diff
@@ -188,24 +189,28 @@ impl std::fmt::Display for HumanHandle {
...
         let email = email.nfc().collect::<String>();
         let label = label.nfc().collect::<String>();
 
-        // TODO: how to check the email  easily ?
-        if email.is_empty() || email.len() >= 255 {
+        if !EmailAddress::is_valid(&email, None) || email.len() >= 255 {
             return Err("Invalid email address");
         }
-        if label.is_empty() || label.len() >= 255 {
+        // According to https://www.rfc-editor.org/rfc/rfc5322#section-3.2.3, these special characters are not allowed
```
FirelightFlagboy, line 201: not allowed for what ?

the following comment indicate that the label is only informative, so why the restriction ?

```rust
// Label is purely informative
pub label: String,
```

TimeEngineer: The label is equivalent to the display name when sending an email
And in test `[Test]` as label should lead to error

FirelightFlagboy: So maybe the comment is misleading here (the one for the label field)

FirelightFlagboy (approved):

PR #2668, opened 2022-07-22
 Convert notary workflow to simple script 

FirelightFlagboy:
Since we lack per orga token generation (we can only generate Personal Token) 
and we don't want to use an alt account to use is GITHUB_TOKEN.

I've remove the useless workflow to replace it with a simple script that need to be run by someone who as access to the repo & project


AntoineDupre (approved):
LGTM

Maybe as it's a script that required to be run manually, some basic documentation can be added

PR #2675, opened 2022-07-25
add windows packaging workflow

FirelightFlagboy:


.github/workflows/package-windows.yml
```diff
@@ -0,0 +1,123 @@
...
+    env:
+      # Need feature from poetry 1.2.x that aren't present in 1.1.x
+      poetry-version: 1.2.0b3
+    outputs:
+      wheel-version: ${{ steps.wheel-version.outputs.version }}
+    steps:
+      - name: Install poetry
+        shell: bash
+        run: |
+          export POETRY_HOME=${APPDATA}/.poetry
```
touilleMan, line 33: `set -eux` should be added  at the beginning of every bash script ;-)

.github/workflows/package-windows.yml
```diff
@@ -0,0 +1,123 @@
...
+            # Customize version with tag
+            sed -i "s#^\\(__version__ = \\).*\$#\\1'$VERSION'#" parsec/_version.py
+            test $(grep -E 'version = ' -o pyproject.toml | wc -l) -eq 1  # Sanity check for the next sed
+            sed -i "s#^\\(version = \\).*\$#\\1'$VERSION'#" pyproject.toml
+          fi
+
+          echo "::warning title=Wheel version::$VERSION"
+          echo "::set-output name=version::$VERSION"
+
+      - name: Create output dir
```
touilleMan, line 70: With `set -eux` there is no need for such fine separation given the executed line is printed

So we can merge `Get wheel version` -> `Debug dist dir` all together into a single `Generate Wheel` (and we would avoid the `::set-output` and have a good old env variable instead ^^)

touilleMan (changes requested):

.github/workflows/package-windows.yml
FirelightFlagboy: done

.github/workflows/package-windows.yml
touilleMan: Yup, I agree that having `Get wheel version` in it own step bring redability

however `Create output dir` + `Set file for wheel version` + `Build wheel` + `Debug dist dir` are just 4 one-liners that should be glued together (with a couple of comments if necessary)

touilleMan (approved):

PR #2685, opened 2022-07-25
Rename jobs name

FirelightFlagboy:
Rename Github Action workflow's jobs to be able to select them in the `require status check` in the configuration of this repository before allowing to merge

### DIFF
--- a/.github/workflows/macos-ci.yml
+++ b/.github/workflows/macos-ci.yml
@@ -12,7 +12,7 @@ env:
   poetry-version: 1.1.13
 
 jobs:
-  python-only:
+  macos-ci-python:
     runs-on: macos-12
     steps:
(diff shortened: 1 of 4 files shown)


TimeEngineer (approved):

.github/workflows/macos-ci.yml
```diff
@@ -12,7 +12,7 @@ env:
   poetry-version: 1.1.13
 
 jobs:
-  python-only:
+  macos-python-only:
```
touilleMan, line 15: macos-ci-python-only ?

touilleMan: And why "python-only" where other ci jobs are just "python"

FirelightFlagboy: python-only because, it only test the python code not `python & rust` code

touilleMan (approved):

PR #2691, opened 2022-07-26
Correct `Set snap build version`

FirelightFlagboy:
- The value version was not quoted, now sed will quote it.
- Added a line to debug the sed'ed file
- Trigger the workflow when it's edited

### DIFF
--- a/.github/workflows/package-linux.yml
+++ b/.github/workflows/package-linux.yml
@@ -113,7 +113,9 @@ jobs:
           cp -v ${{ github.workspace }}/dist/core-requirements.txt src/
 
       - name: Set snap build version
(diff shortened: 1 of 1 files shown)


.github/workflows/package-linux.yml
```diff
@@ -1,6 +1,9 @@
 name: package-linux
 
 on:
+  pull_request:
+    paths:
+      - .github/workflows/package-linux.yml
```
touilleMan, line 6: This will be removed before merge, right ?

touilleMan (approved):

.github/workflows/package-linux.yml
FirelightFlagboy: no

FirelightFlagboy: I want to keep this trigger, since it will be only triggered when we create a PR that edit the workflow's file

PR #2697, opened 2022-07-26
[GUI] Fixed an import error with files without extensions

### LINKED ISSUE #2696: [🐛 | Bug]: Error while importing a folder in the GUI
### Version

2.11.0rc2

Windows

### What happened ?

Error while importing a whole folder in the GUI. Seems to happen for while that don't have an extension.

### Relevant output

```
ValueError

  File "parsec\core\gui\files_widget.py", line 758, in _do_import
  File "parsec\core\gui\files_widget.py", line 859, in _import_all
  File "parsec\core\gui\files_widget.py", line 891, in _import_one
ValueError: not enough values to unpack (expected 2, got 1)
```

Max-7:
Closes #2696

### DIFF
--- a/parsec/core/gui/files_widget.py
+++ b/parsec/core/gui/files_widget.py
@@ -888,15 +888,12 @@ class FilesWidget(QWidget, Ui_FilesWidget):
         async with await trio.open_file(source, "rb") as f:
 
             # Getting the file name without extension (anything after the first dot is considered to be the extension)
-            name_splitted = dest.name.str.split(".", 1)
-            if len(name_splitted) == 1:
-                name_splitted.append("")
-            name_we, suffixes = name_splitted
+            name_we, *suffixes = dest.name.str.split(".")
             # Count starts at 2 (1 would be the file without a number)
             count = 2
             while await self.workspace_fs.exists(dest):
                 # Create the new file name by adding the count ("myfile.txt" becomes "myfile (2).txt")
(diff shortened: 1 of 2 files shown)


parsec/core/gui/files_widget.py
```diff
@@ -888,7 +888,10 @@ async def _import_one(
         async with await trio.open_file(source, "rb") as f:
 
             # Getting the file name without extension (anything after the first dot is considered to be the extension)
-            name_we, suffixes = dest.name.str.split(".", 1)
+            name_splitted = dest.name.str.split(".", 1)
+            if len(name_splitted) == 1:
+                name_splitted.append("")
```
touilleMan, line 893: This seems wrong: `foo` would get renamed as `foo (2).` instead of `foo (2)`

It would be better to do something like:
```python
suffixes: List[str]
name_we, *suffixes = dest.name.str.split(".")

[...]

# Create the new file name by adding the count ("myfile.txt" becomes "myfile (2).txt")
new_file_name = EntryName("."join([f"{name_we} ({count})", *suffixes)])
```

touilleMan (changes requested):

parsec/core/gui/files_widget.py
Max-7: Fixed! Good catch.

touilleMan (approved):

PR #2698, opened 2022-07-26
Yolo/correct trigger tags and comment

FirelightFlagboy:
- Correct tags pattern to match something link `vX.X.X[...]`
- Add comment to explain how the triggers work


touilleMan (approved):
Note we should be careful later on when adding the snap&pypi upload logic to only be triggered in case of tag

PR #2700, opened 2022-07-26
Yolo/rewrite dev docs

FirelightFlagboy:


docs/development/quickstart.rst
```diff
@@ -89,9 +89,9 @@ On linux:
 
 On Windows:
 
-.. code-block:: shell
+.. code-block:: powershell
```
touilleMan, line 92: I don't mind providing help for powershell, but cmd.exe must stay

touilleMan (changes requested):

docs/development/quickstart.rst
FirelightFlagboy: done

PR #2709, opened 2022-07-27
Added reencrypt_workspace CLI

### LINKED ISSUE #2708: Add reencrypt_workspace CLI

Max-7:
Closes #2708

### DIFF
--- a/parsec/core/cli/reencrypt_workspace.py
+++ b/parsec/core/cli/reencrypt_workspace.py
@@ -1,0 +1,46 @@
+# Parsec Cloud (https://parsec.cloud) Copyright (c) AGPLv3 2016-2021 Scille SAS
+
+import click
+
+
+from parsec.utils import trio_run
+from parsec.api.data import EntryName
+from parsec.cli_utils import cli_exception_handler, spinner
+from parsec.core import logged_core_factory
+from parsec.core.types import LocalDevice
+from parsec.core.config import CoreConfig
+from parsec.core.cli.utils import cli_command_base_options, core_config_and_device_options
+
+
+async def _reencrypt_workspace(config: CoreConfig, device: LocalDevice, name: EntryName) -> None:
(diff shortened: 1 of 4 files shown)


parsec/core/cli/reencrypt_workspace.py
```diff
@@ -0,0 +1,46 @@
+# Parsec Cloud (https://parsec.cloud) Copyright (c) AGPLv3 2016-2021 Scille SAS
+
+import click
+
+
```
touilleMan, line 5: single newline between groups of imports

parsec/core/cli/reencrypt_workspace.py
```diff
@@ -0,0 +1,46 @@
...
+async def _reencrypt_workspace(config: CoreConfig, device: LocalDevice, name: EntryName) -> None:
+    async with logged_core_factory(config, device) as core:
+        workspace = core.find_workspace_from_name(name)
+        workspace_id = workspace.id
+        workspace_fs = core.user_fs.get_workspace(workspace_id)
+        reenc_needs = await workspace_fs.get_reencryption_need()
+        if not reenc_needs.need_reencryption:
+            raise RuntimeError("The workspace does not need to be reencrypted")
```
touilleMan, line 22: Raising an error for this seems a bit to harsh, instead I'd go with a nice `click.echo` message then return (so status code of the process would be 0)

touilleMan (changes requested):

parsec/core/cli/reencrypt_workspace.py
Max-7: A bit surprised this wasn't caught by the pre-commit, but it's fixed.

parsec/core/cli/reencrypt_workspace.py
Max-7: Done.

touilleMan (approved):
:+1:

PR #2719, opened 2022-07-28
Feat/2507 implement a client factory for client connection

FirelightFlagboy:
# Questions

- [ ] Instead of adding the factory `AuthenticatedCmdsFactory`, souldn't exposing `get_http_client(,_mut)` be enough (and equivalent for `device_id`, `server_url` and `signing_key`) ?


FirelightFlagboy:
The issue is not well defined to be implemented right now

PR #2739, opened 2022-08-01
Disable azure pipelines for push and PR triggers

vxgmichel:
The azure jobs are kept for tags, while we're validating the new GH action packaging process.

Also add `package-linux-release-pypi` and `package-linux-release-snap` steps in package-linux workflow.

### DIFF
--- a/.github/workflows/package-linux.yml
+++ b/.github/workflows/package-linux.yml
@@ -132,3 +132,18 @@ jobs:
         with:
           name: ${{ runner.os }}-snap
           path: ${{ runner.temp }}/parsec*.snap
+
+  package-linux-release-pypi:
+    needs: package-linux-build-wheel
+    runs-on: ubuntu-20.04
+    if: github.event_name == 'push' && startsWith(github.event.ref, 'refs/tags/v')
+    steps:
+      - uses: actions/download-artifact@fb598a63ae348fa914e94cd0ff38f362e927b741 # pin v3.0.0
+        with:
+          name: ${{ runner.os }}-wheel
+          path: dist
+      - name: Publish wheel on PyPI
+        uses: pypa/gh-action-pypi-publish@37f50c210e3d2f9450da2cd423303d6a14a6e29f # pin v1.5.1
+        with:
+          user: __token__
+          password: ${{ secrets.PYPI_CREDENTIALS }}
--- a/.github/workflows/package-linux.yml
+++ b/.github/workflows/package-linux.yml
@@ -133,23 +133,31 @@ jobs:
           name: ${{ runner.os }}-snap
           path: ${{ runner.temp }}/parsec*.snap
 
-  package-linux-release-pypi:
-    needs: package-linux-build-wheel
+  package-linux-test-snap:
+    needs: package-linux-build-snap
     runs-on: ubuntu-20.04
-    if: github.event_name == 'push' && startsWith(github.event.ref, 'refs/tags/v')
     steps:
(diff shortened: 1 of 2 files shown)


FirelightFlagboy (approved):

.github/workflows/package-linux.yml
```diff
@@ -132,3 +132,36 @@ jobs:
...
+    runs-on: ubuntu-20.04
+    if: github.event_name == 'push' && startsWith(github.event.ref, 'refs/tags/v')
+    steps:
+      - uses: actions/download-artifact@fb598a63ae348fa914e94cd0ff38f362e927b741 # pin v3.0.0
+        with:
+          name: ${{ runner.os }}-snap
+          path: dist
+
+      - name: Install Snapcraft
+        uses: samuelmeuli/action-snapcraft@10d7d0a84d9d86098b19f872257df314b0bd8e2d # pin v1.2.0
```
touilleMan, line 162: why not using the official https://github.com/snapcore/action-build instead ?

`samuelmeuli/action-snapcraft`'s readme indicate the official is a couple of mn slower, but I'd rather pay this price and have a more supported action (`samuelmeuli/action-snapcraft` seems to use ubuntu 18.04 which will most likely cause random issues given we currently use 20.04)

touilleMan: Besides, this action onyl install snapcraft & do the login, it doesn't actually run the build command !

touilleMan: and finally, I'm not sure we gain much using this action compared to the ad-hoc stuff in azure-pipeline.yml

The action does the build in lxd which is more isolated in theory, however I guess the ubuntu-20.04 image from github action is also pretty standard so using it in --destructive is okay (at least we never had any issue in azure with this approach)

On the other hand the action makes things more opaque and using lxd is slower (cf. the couple of minute slower indication in `samuelmeuli/action-snapcraft`'r README)

.github/workflows/package-linux.yml
```diff
@@ -132,3 +132,36 @@ jobs:
         with:
           name: ${{ runner.os }}-snap
           path: ${{ runner.temp }}/parsec*.snap
+
+  package-linux-release-pypi:
+    needs: package-linux-build-wheel
+    runs-on: ubuntu-20.04
+    if: github.event_name == 'push' && startsWith(github.event.ref, 'refs/tags/v')
```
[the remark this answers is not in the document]
vxgmichel: Because this `variables[]` syntax is specific to azure.

.github/workflows/package-linux.yml
touilleMan: Too lazy to change it for now, I'll rip this off next release as soon as things start falling appart ;p

.github/workflows/package-linux.yml
```diff
@@ -132,3 +132,36 @@ jobs:
...
+          name: ${{ runner.os }}-wheel
+          path: dist
+      - name: Publish wheel on PyPI
+        uses: pypa/gh-action-pypi-publish@37f50c210e3d2f9450da2cd423303d6a14a6e29f # pin v1.5.1
+        with:
+          user: __token__
+          password: ${{ secrets.PYPI_CREDENTIALS }}
+
+  package-linux-release-snap:
+    needs: package-linux-build-snap
```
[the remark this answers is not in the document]
vxgmichel: [28043e1](https://github.com/Scille/parsec-cloud/pull/2739/commits/28043e14f0f7c775d2038b427d0764b71deba4cb)

FirelightFlagboy (approved):

touilleMan (approved):
:+1:

PR #2753, opened 2022-08-01
Better signing windows packaging

touilleMan:


.azure-pipelines.yml
```diff
@@ -474,9 +474,9 @@ jobs:
         artifact: macos_amd64_app
 
 
-  - job: s2_windows_installer_build
-    displayName: Windows installer build
-    condition: startsWith(variables['build.sourceBranch'], 'refs/tags/')
+  - job: s2_windows_installer_content_build
+    displayName: Windows installer content build
+    # condition: startsWith(variables['build.sourceBranch'], 'refs/tags/')
```
vxgmichel, line 479: Remember to revert those

packaging/win32/README.md
```diff
@@ -19,30 +14,53 @@ Note the Python version embedded inside the build will be taken from the interpr
...
 
-### 2 - Package the application
+```shell
+set PATH=C:\Program Files (x86)\Windows Kits\10\bin\$VERSION\x64;%PATH%
+set PATH=C:\Program Files (x86)\NSIS;%PATH%
+```
+
+### 2.2 - The simple way
+
+Run the `make_installe.py` Pythons script:
```
vxgmichel, line 40: `make_installer.py`

packaging/win32/README.md
```diff
@@ -19,30 +14,53 @@ Note the Python version embedded inside the build will be taken from the interpr
...
-$ "C:\Program Files (x86)\NSIS\makensis.exe" installer.nsi
+python make_installer.py --sign-mode exe
 ```
 
 This will generate a `build\parsec-<version>-<platform>-setup.exe` installer.
 
+### 2.3 - The advanced way
 
-### 3 - Sign the application
+Under the hood, `make_installe.py` does 3 things:
```
vxgmichel, line 50: `make_installer.py`

packaging/win32/make_installer.py
```diff
@@ -0,0 +1,90 @@
...
+            print("### Checking all shipped exe/dll are signed ###")
+            not_signed = []
+            for file in itertools.chain(
+                freeze_program.rglob("*.exe"), freeze_program.rglob("*.dll")
+            ):
+                if not is_signed(file):
+                    not_signed.append(file)
+                    print("Unsigned file detected:", file)
+            if not_signed:
+                raise SystemExit("Some file are not signed, aborting")
```
vxgmichel, line 82: Does that mean that `.dll` and other `*.exe` files are meant to be signed by someone else? If not, why not sign them directly?

.azure-pipelines.yml
```diff
@@ -526,8 +526,9 @@ jobs:
           cp build/parsec*-setup.exe dist
         displayName: Build NSIS setup
         workingDirectory: $(Agent.TempDirectory)/winbuild
-      - publish: $(Agent.TempDirectory)/winbuild/dist
-        artifact: win_$(vs.arch)_installer
+      # Cannot do the NSIS installer part in CI given it requires to sign `parsec.exe`
+      - publish: $(Agent.TempDirectory)/winbuild/build/parsec-*
+        artifact: win_$(vs.arch)_installer_content
```
vxgmichel, line 531: Then the `makensis.exe installer.nsi` command is not necessary anymore? Should we remove it?

.azure-pipelines.yml
```diff
@@ -474,9 +474,9 @@ jobs:
         artifact: macos_amd64_app
 
 
-  - job: s2_windows_installer_build
-    displayName: Windows installer build
-    condition: startsWith(variables['build.sourceBranch'], 'refs/tags/')
+  - job: s2_windows_installer_content_build
```
vxgmichel, line 477: The azure CI fails with this error because of the renaming:
> Job s4_wheel_pypi_upload depends on unknown job s2_windows_installer_build.

vxgmichel (changes requested):
A couple of comments and questions, a CI error but LGTM otherwise :+1:

FirelightFlagboy (changes requested):
Rename `packaging/win32` to `packaging/windows`

packaging/win32/make_installer.py
touilleMan: If I recall right, libpython is already signed by Python authority

But we don't use the `sign_mode == all` for the moment given anyway, it is for later improvement ;-)

.azure-pipelines.yml
touilleMan: done.

.azure-pipelines.yml
touilleMan: I forgot to remove the nsis stuff, done.

packaging/win32/README.md
touilleMan: done.

packaging/win32/README.md
touilleMan: done.

.github/workflows/package-windows.yml
```diff
@@ -108,14 +108,21 @@ jobs:
...
+          Copy-Item -Path build\manifest.ini -Destination dist
+          Copy-Item -Path build\install_files.nsh -Destination dist
+          Copy-Item -Path build\uninstall_files.nsh -Destination dist
+          Copy-Item -Recurse -Path build\parsec-* -Destination dist
+          Copy-Item -Path build\winfsp-* -Destination dist
+        working-directory: ${{ runner.temp }}
+
       - uses: actions/upload-artifact@3cea5372237819ed00197afe530f5a7ea3e805c8 # pin v3.1.0
         with:
           name: ${{ runner.os }}-${{ runner.arch }}-installer
```
FirelightFlagboy, line 127: Maybe rename the generated artifact ?

FirelightFlagboy (approved):

.github/workflows/package-windows.yml
touilleMan: what for ?

touilleMan: we need them with their original name so that we can locally extract the artifact archive then run `make_installer.py`

FirelightFlagboy: the artifact now don't contains the installer but the files to generate the installer
as you renamed the job to `[...]-installer-content`

touilleMan: Ho you're right ! I missed this field ;-)

FirelightFlagboy (approved):

.github/workflows/package-windows.yml
```diff
@@ -108,14 +108,22 @@ jobs:
...
+          Copy-Item -Recurse -Path build\parsec-* -Destination dist
+          Copy-Item -Path build\winfsp-* -Destination dist
+        working-directory: ${{ runner.temp }}
 
       - uses: actions/upload-artifact@3cea5372237819ed00197afe530f5a7ea3e805c8 # pin v3.1.0
         with:
-          name: ${{ runner.os }}-${{ runner.arch }}-installer
-          path: ${{ runner.temp }}/build/parsec*-setup.exe
+          name: ${{ runner.os }}-${{ runner.arch }}-installer-content
+          path: ${{ runner.temp }}/dist/
```
FirelightFlagboy, line 129: Instead of coping some files in `build/` to `dist/`, you could have listed them directly here

source [Upload using multiple paths](https://github.com/actions/upload-artifact#upload-using-multiple-paths-and-exclusions)

touilleMan: That's a good catch, too bad I've already merge it !

I'll open a new PR with this change: avoiding the copy is good given it is clunky work with powershell commands :/

PR #2764, opened 2022-08-02
Improve windows packaging

touilleMan:
Improve freeze_program.py/make_installer.py and NSIS scripts now that the output of freeze_program is generated on the CI, then fetched on a local machine to run make_installer.py

On top of that, wheel_it.py has been slightly modified so that freeze_program.py can be run with only the output of wheel_it.py (i.e. without the need for the program's sources)


FirelightFlagboy:
Nice try, try again

PR #2771, opened 2022-08-02
Add User & Device info

TimeEngineer:


TimeEngineer:
I would like to add the binding, but some change on DateTime would be apprecied :smile:

src/local_device.rs
```diff
@@ -275,3 +275,142 @@ impl LocalDevice {
...
+
+    fn __repr__(&self) -> PyResult<String> {
+        Ok(format!("{:?}", self.0))
+    }
+
+    fn __richcmp__(&self, other: &Self, op: CompareOp) -> bool {
+        match op {
+            CompareOp::Eq => self.0 == other.0,
+            CompareOp::Ne => self.0 != other.0,
+            CompareOp::Lt => self.0 < other.0,
```
touilleMan, line 387: Why do we need the full range of comparison here ? (given we only have eq/ne for UserInfo)

touilleMan (commented):
Seems like a good start, now this should remove the Python implementation of UserInfo/DeviceInfo...

src/local_device.rs
TimeEngineer: There is a sort in tests

touilleMan: Then I think we should also implement this for the UserInfo (given it should be trivial do to and it's probably interesting to have deterministic order when testing/displaying)

touilleMan (approved):

PR #2775, opened 2022-08-03
Add a script for .dmg generation for MacOS packaging

Ironicbay:
Long overdue

### DIFF
--- a/packaging/macOS/create_dmg.sh
+++ b/packaging/macOS/create_dmg.sh
@@ -11,7 +11,7 @@ Path to the app in its dist folder is needed as argument.
 if [ -z "$1" ]; then
     echo "Error: No argument given, path to app needed"
     exit 1
-elif ! [[ $1 == *"/dist/Parsec.app" ]]; then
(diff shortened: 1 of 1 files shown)


FirelightFlagboy (commented):
Instead of checking if the folder contain only 4 files, can't we just copy those 4 required files to a _tmp dir_ for building ?

That would be more explicit on which files are required and present

Ironicbay:
@FirelightFlagboy I'm using `ls -la` to check for hidden files in the folder and counting the lines, and the other 3 would be `total [...]`, `.` and `..`. It's just to make sure only `Parsec.app` is present in the folder so that we don't package anything else. These checks aren't really necessary in the first place, it's to make sure I'm not forgetting stuff when packaging.

FirelightFlagboy:
So why not just coping `Parsec.app` to a temp folder that's empty ? that way you are sure to only package `Parsec.app` and this will also check if the file is present (since we're coping it)

FirelightFlagboy (approved):

PR #2773, opened 2022-08-02
Experiment with cibuildwheel and poetry

vxgmichel:

PR #2778, opened 2022-08-04
Feat/2489 maturin based pyproject 2

### LINKED ISSUE #2489: [Oxidation in prod] Poetry/Maturin based pyproject.toml for main parsec Python project

FirelightFlagboy:
supersede #2642, supersede #2773

closes #2489

## TODOs

- [x] Remove `IS_OXIDIZE` constant in `__init__.py`
- [x] Config `linux-ci`
- [x] Config `macos-ci`
- [x] Config `windows-ci`
- [x] Use `CIBuildWheel` to create the wheel on `linux`
- [x] Use `CIBuildWheel` to create the wheel on `windows`
- [x] Use `CIBuildWheel` to create the wheel on `mac-os`


packaging/macOS/freeze_program.py
```diff
@@ -131,12 +131,17 @@ def main(
         if include_parsec_ext is not None:
             pyinstaller_venv_requirements_data += f"{include_parsec_ext}\n"
         pyinstaller_venv_requirements.write_text(pyinstaller_venv_requirements_data)
+
+        with open(pyinstaller_venv_requirements) as f:
+            for (index, line) in enumerate(f.readlines()):
+                display(f'{pyinstaller_venv_requirements}[{index:4}] = "{line}"')
+
         # Now do the actual install, note the constraint file that ensures
```
FirelightFlagboy, line 139: To remove

packaging/macOS/freeze_program.py
```diff
@@ -131,12 +131,17 @@ def main(
...
         # parsec_ext won't mess with parsec dependencies
         # TODO: `--use-deprecated=legacy-resolver` is needed due to a bug in pip
         # see: https://github.com/pypa/pip/issues/9644#issuecomment-813432613
         run(
-            f"{ pyinstaller_python } -m pip install --requirement { pyinstaller_venv_requirements } --constraint { program_constraints }  --use-deprecated=legacy-resolver"
+            f"{ pyinstaller_python } -m pip install -vvv --debug --requirement { pyinstaller_venv_requirements } --constraint { program_constraints }  --use-deprecated=legacy-resolver"
```
FirelightFlagboy, line 144: To revert

pyproject.toml
```diff
@@ -140,8 +142,37 @@ flake8 = "3.8.4"
...
+archs = ["auto64"]
+build-verbosity = 3
+before-all = "bash ci/setup-rust.sh"
+before-build = "python -m pip install pip"
+
+[tool.cibuildwheel.environment]
+PATH = "$PATH:$HOME/.cargo/bin"
+
+[tool.cibuildwheel.windows]
+before-build = "python -m pip install -U pip"
```
FirelightFlagboy, line 166: Merge both `before-build` ?

build.py
```diff
@@ -0,0 +1,64 @@
...
+def in_cibuildwheel():
+    return bool(int(os.environ.get("CIBUILDWHEEL", "0")))
+
+
+def check_venv():
+    # Both conda and venv are configured, pick conda
+    if os.environ.get("CONDA_PREFIX") and os.environ.get("VIRTUAL_ENV"):
+        del os.environ["VIRTUAL_ENV"]
+    # Build with cibuildwheel detected, set `VIRTUAL_ENV`
+    if in_cibuildwheel():
```
vxgmichel, line 35: Maybe change it to:
```suggestion
    if in_cibuildwheel() and not os.environ.get("VIRTUAL_ENV"):
```
So we don't override a configured venv, even when cibuildwheel is used?

build.py
```diff
@@ -0,0 +1,64 @@
...
+    # Debug
+    log.debug("Environment:")
+    for key, value in os.environ.items():
+        log.debug(f"- {key} = {value}")
+
+    run(f"{PYTHON_PATH} --version")
+    run(f"{PYTHON_PATH} -m pip freeze")
+    run(f"{PYTHON_PATH} misc/generate_pyqt.py")
+    check_venv()
+    release = "--release" if in_cibuildwheel() else ""
```
vxgmichel, line 54: It would be useful to be able to trigger release builds even with `poetry install`, for instance using `FORCE_MATURIN_RELEASE=1 poetry install`:
```suggestion
    release = "--release" if in_cibuildwheel() or force_maturin_release() else ""
```
with `force_maturin_release` defined as:
```python
def force_maturin_release():
    return bool(int(os.environ.get("FORCE_MATURIN_RELEASE", "0")))
```

pyproject.toml
```diff
@@ -140,8 +142,34 @@ flake8 = "3.8.4"
...
 [build-system]
-requires = ["poetry-core>=1.0.0", "setuptools"]
+requires = [
+    "poetry-core>=1.0.0",
+    "setuptools",
+    "maturin~=0.13",
+    "PyQt5~=5.15",
+    "Babel~=2.10",
+    "docutils~=0.17",
+    "pip"
```
vxgmichel, line 173: Maybe add a comment advising to update pip using `pip install -U pip` if the error `ERROR: To modify pip, please run the following command: [...]` appears

vxgmichel (approved):
I just realized we did not deal with the `--runrust` option.

We should decide whether we want to remove the `tests/rust` tests, or move them as migration tests, or something else.

Otherwise LGTM :+1: , only a few suggestions here and there :)

.github/workflows/package-macos.yml
```diff
@@ -67,43 +68,46 @@ jobs:
...
-          name: ${{ runner.os }}-wheel
+          name: ${{ runner.os }}-${{ runner.arch }}-wheel
           path: dist
 
       - name: Copy packaging script
         run: cp -Rv ${{ github.workspace }}/packaging/macOS/* ${{ runner.temp }}
 
       - name: Freeze Parsec
-        run: python3 freeze_program.py ${{ github.workspace }} --wheel-it-dir ${{ github.workspace }}/dist
+        run: ${{ steps.setup-python.outputs.python-path }} freeze_program.py ${{ github.workspace }} --wheel-it-dir ${{ github.workspace }}/dist
```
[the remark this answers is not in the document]
FirelightFlagboy: The path is instead correctly setup, and the `python` command use the python installed by `setup-python` this line was more of a sanity check

.github/workflows/package-macos.yml
```diff
@@ -20,7 +20,7 @@ env:
 
 jobs:
   package-macos-build-wheel:
-    runs-on: macos-12
+    runs-on: macos-10.15
```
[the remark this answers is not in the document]
FirelightFlagboy: done in bea44c5

.github/workflows/package-macos.yml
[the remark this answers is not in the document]
FirelightFlagboy: done in 37c6caf

build.py
FirelightFlagboy: done in ed05818

build.py
FirelightFlagboy: done in 5996000

pyproject.toml
FirelightFlagboy: done in 11bc3af

build.py
```diff
@@ -0,0 +1,68 @@
...
+        raise RuntimeError("A virtual env is required to run `maturin develop`")
+
+
+def force_maturin_release() -> bool:
+    return os.environ.get("FORCE_MATURIN_RELEASE", "0") == "1"
+
+
+def build():
+    # Debug
+    log.debug("Environment:")
```
touilleMan, line 50: Do we have a reason for dump the environment variables like this ? (I have the feeling this is some debug code useful we creating this script, but which should be removed now that the code is stable)

This might not be a good idea for instance in the CI if we expose a github secret as environment variable...

ci/setup-rust.sh
```diff
@@ -0,0 +1,14 @@
+#!bash
```
touilleMan, line 1: `ci` folder name is misleading (I was thinking it was related to github action...)

We should move this to `misc/cibuildwheel_setup_rust.sh`

ci/setup-rust.sh
```diff
@@ -0,0 +1,14 @@
...
+set -eux
+
+RUST_TOOLCHAIN=${1:-1.62}
+
+TMP_DIR=$(mktemp -d -t rust-install.XXXXXXXXXX)
```
touilleMan, line 6:
```suggestion
TMP_DIR=$(mktemp -d -t parsec-cibuildwheel-rust-install.XXXXXXXXXX)
```

ci/setup-rust.sh
```diff
@@ -0,0 +1,14 @@
...
+
+trap "rm -vrf $TMP_DIR" EXIT INT TERM
```
touilleMan, line 8: What is this for ?

I have the feeling we want to replace this script by a python one, so that nice context manager can handle cleanup-on-teardown nicely

and if this is to remove the `$TMP_DIR/rust-install.sh`, this is far too cumbersome (and you can do `curl <url> | bash` to execute the script without saving it on disk !)

ci/setup-rust.sh
```diff
@@ -0,0 +1,14 @@
+#!bash
+set -eux
```
touilleMan, line 2: Can you add a comment to explain under which context this script is going to be use

ci/setup-rust.sh
```diff
@@ -0,0 +1,14 @@
...
+
+RUST_TOOLCHAIN=${1:-1.62}
```
touilleMan, line 4: Can you add a check to avoid the curl + bash if rust toolchain is already installed.

- this is useful when offline or with poor connection
- curl + bash exec is always a security hazard, so better avoiding it when not strictly needed

touilleMan (changes requested):

parsec/__init__.py
```diff
@@ -3,13 +3,8 @@
...
-import parsec.utils  # noqa
+__import__("parsec.utils")
 
-try:
-    import libparsec  # noqa
-except ImportError:
-    IS_OXIDIZED = False
-else:
-    IS_OXIDIZED = True
+import parsec._parsec  # noqa
```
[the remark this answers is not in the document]
FirelightFlagboy: It's a sanity check, to check is the rust lib is present

ci/setup-rust.sh
FirelightFlagboy: why ?

src/lib.rs
```diff
@@ -0,0 +1,15 @@
+use pyo3::prelude::*;
+
+/// Formats the sum of two numbers as string.
+#[pyfunction]
```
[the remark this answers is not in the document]
FirelightFlagboy: This function will be remove in the later PRs

.github/workflows/linux-ci.yml
```diff
@@ -43,10 +43,6 @@ jobs:
         --side-effects-timeout=10
         -vv
         -x
-    strategy:
-      fail-fast: false
```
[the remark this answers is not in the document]
FirelightFlagboy: No more matrix 😉

.github/workflows/package-macos.yml
```diff
@@ -20,7 +20,9 @@ env:
 
 jobs:
   package-macos-build-wheel:
-    runs-on: macos-12
+    # We use an old version of `macos` because currently when we generate the wheel with cpython 3.9 and
+    # We try to install it in the later job `package-macos-build-app`, The generated wheel is considered in compatible with python 3.9 without additional hint
```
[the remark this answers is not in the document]
FirelightFlagboy: indeed

pyproject.toml
```diff
@@ -140,8 +142,38 @@ flake8 = "3.8.4"
...
+
+[tool.cibuildwheel]
+build = "cp39-{manylinux,macos,win}*"
+archs = ["auto64"]
+build-verbosity = 3
+before-all = "bash ci/setup-rust.sh"
+before-build = "python -m pip install -U pip"
+
+[tool.cibuildwheel.environment]
+PATH = "$PATH:$HOME/.cargo/bin"
```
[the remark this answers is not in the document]
FirelightFlagboy: done

pyproject.toml
```diff
@@ -140,8 +142,38 @@ flake8 = "3.8.4"
...
 [build-system]
-requires = ["poetry-core>=1.0.0", "setuptools"]
+requires = [
+    "poetry-core>=1.0.0",
+    "setuptools",
+    "maturin~=0.13",
+    "PyQt5~=5.15",
+    "Babel~=2.10",
+    "docutils~=0.17",
+    # We need pip to be installed because `maturin` require it.
```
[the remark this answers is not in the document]
FirelightFlagboy: done

pyproject.toml
```diff
@@ -140,8 +142,38 @@ flake8 = "3.8.4"
...
+before-all = "bash ci/setup-rust.sh"
+before-build = "python -m pip install -U pip"
+
+[tool.cibuildwheel.environment]
+PATH = "$PATH:$HOME/.cargo/bin"
 
 [build-system]
-requires = ["poetry-core>=1.0.0", "setuptools"]
+requires = [
+    "poetry-core>=1.0.0",
```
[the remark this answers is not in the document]
FirelightFlagboy: done

parsec/__init__.py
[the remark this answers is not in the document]
FirelightFlagboy: Changed the import method to check

packaging/wheel/wheel_it.py
```diff
@@ -59,22 +59,14 @@ def main(program_source: Path, output_dir: Path):
...
-
-    # Finally generate the wheel, note we don't use Poetry for the job given:
-    # - It is not possible to choose the output directory
-    # - And more importantly, Poetry is not PEP517 compliant and build wheel
-    #   without building binary resources (it basically only zip the source code)
-    run(f"{python} -m pip wheel {program_source} --wheel-dir {output_dir} --use-pep517 --no-deps")
+    if not skip_wheel:
+        # Finally generate the wheel, note we don't use Poetry for the job given:
+        # - It is not possible to choose the output directory
+        # - And more importantly, Poetry is not PEP517 compliant and build wheel
```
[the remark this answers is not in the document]
FirelightFlagboy: done

.github/workflows/package-macos.yml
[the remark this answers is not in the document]
FirelightFlagboy: done

.github/workflows/package-macos.yml
```diff
@@ -67,43 +70,46 @@ jobs:
...
-        run: mkdir dist
+      - name: Build wheel
+        uses: pypa/cibuildwheel@afb4329fe32181c15d34ef94bf64dad7715b44ba  # ping v2.8.1
+        with:
+          output-dir: dist
 
       - name: Set file for wheel version
         run: echo ${{ steps.wheel-version.outputs.version }} > dist/version
 
       - name: Build wheel
```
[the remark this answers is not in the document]
FirelightFlagboy: done

.github/workflows/package-linux.yml
```diff
@@ -67,21 +67,20 @@ jobs:
...
-        run: mkdir dist
+      - name: Build wheel
+        uses: pypa/cibuildwheel@afb4329fe32181c15d34ef94bf64dad7715b44ba  # ping v2.8.1
+        with:
+          output-dir: dist
 
       - name: Set file for wheel version
         run: echo ${{ steps.wheel-version.outputs.version }} > dist/version
 
       - name: Build wheel
```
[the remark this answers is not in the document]
FirelightFlagboy: done

build.py
```diff
@@ -0,0 +1,68 @@
...
+
+YELLOW_FG = "\x1b[33m"
+DEFAULT_FG = "\x1b[39m"
+
+log = logging.getLogger("build")
+logging.basicConfig(
+    level=logging.DEBUG, format=f"{YELLOW_FG}[%(levelname)s] %(message)s{DEFAULT_FG}"
+)
+
+PYTHON_PATH = sys.executable
```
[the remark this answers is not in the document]
FirelightFlagboy: done

ci/setup-rust.sh
FirelightFlagboy: `setup-rust` can be used outside of `cibuildwheel`

FirelightFlagboy: done

ci/setup-rust.sh
FirelightFlagboy: done

ci/setup-rust.sh
FirelightFlagboy: done

ci/setup-rust.sh
FirelightFlagboy: done

ci/setup-rust.sh
FirelightFlagboy: done

build.py
```diff
@@ -0,0 +1,68 @@
...
+
+import os
+import sys
+import logging
+import subprocess
+
+YELLOW_FG = "\x1b[33m"
+DEFAULT_FG = "\x1b[39m"
+
+log = logging.getLogger("build")
```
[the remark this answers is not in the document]
FirelightFlagboy: done

build.py
FirelightFlagboy: done

src/lib.rs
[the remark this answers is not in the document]
touilleMan: why not now ? this is unrelated to parsec so better cleanup now

pyproject.toml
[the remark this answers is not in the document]
touilleMan: The comment should explain *why* it we need to have this folder among the PATH

I guess this is needed because rust might not be present on the system (given `setup-rust.sh` detect and install it if needed). In this case, `$HOME/.cargo/bin` doesn't exist when cibuildwheel is started, hence the PATH environment is missing it (and `setup-rust.sh` cannot modify an environment variable for it parents process), henc why we have to force the path by hand.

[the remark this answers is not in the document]
FirelightFlagboy: Did you read the comment that I've added ? If the comment need change, open a new discussion at the start of the comment

src/lib.rs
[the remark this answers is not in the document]
FirelightFlagboy: done in 3d3f5f7

touilleMan (approved):

PR #2806, opened 2022-08-08
Improve comment in pyproject's cibuildwheel.environment

touilleMan:


pyproject.toml
```diff
@@ -159,8 +159,10 @@ before-all = "bash misc/setup-rust.sh"
 before-build = "python -m pip install -U pip"
 
 [tool.cibuildwheel.environment]
-# Need to add `Rust` installation path in the `PATH` variable
-# because `cibuildwheel` don't allow the export variable in `before-*` steps
+# As it name suggests, `misc/setup-rust.sh` (run during cibuildwheel's before-all) will
+# install Rust if it is not already available. In this case, Rust bin dir was previously
+# non-existant and hence $PATH don't know about it (and maturin will fail when calling cargo).
+# For this reason we force $PATH to contain Rust bin dir.
```
FirelightFlagboy, line 165: My two cents

- Even if rust is already installed in `$HOME/.cargo/bin`, that path isn't a standard one so we still need to add it the `PATH`
- `misc/setup-rust.sh` is mainly used by `cybuildwheel` but it can be use to setup `Rust` on a dev env

FirelightFlagboy (approved):

pyproject.toml
touilleMan: > Even if rust is already installed in $HOME/.cargo/bin, that path isn't a standard one so we still need to add it the PATH

It's the same thing with any 3rd party installer, for instance installing poetry with pipx requires to have PATH taking into account pipx's bin dir.
It's legit to consider the environment is already correctly configured. So the fact our script is working if "rust is installer in $HOME/.cargo/bin *but* $PATH hasn't been updated" is more of a unexpected side effect than a real design decision (for instance if rust has been installed in another directory there is nothing we can do to retrieve it).

> misc/setup-rust.sh is mainly used by cybuildwheel but it can be use to setup Rust on a dev env

I'm not sure how useful it is given we also have node/python3.9/poetry as project requirements, so I would say the the new dev is expected to follow `docs/development/quickstart.rst` (where a link on the official document for installing Rust is the best thing to provide)

FirelightFlagboy: :+1:

PR #2807, opened 2022-08-08
Remove step `Generate PyQT GUI data`

FirelightFlagboy:
remove the step that was generating `pyqt` data since we run the command
in the `build.py` script


touilleMan (approved):
:partying_face:

PR #2811, opened 2022-08-09
Improve handling of HTTP `server` header in backend server

touilleMan:
Currently, there is two server headers in the parsec server response:

```shell
$ curl --head http://127.0.0.1:6777/foo
HTTP/1.1 404 
content-type: text/html; charset=utf-8
content-length: 823
server: parsec/v2.11.0+dev
date: Tue, 09 Aug 2022 09:19:26 GMT
server: hypercorn-h11
```

### DIFF
--- a/parsec/backend/asgi/__init__.py
+++ b/parsec/backend/asgi/__init__.py
@@ -1,18 +1,23 @@
 # Parsec Cloud (https://parsec.cloud) Copyright (c) BUSL-1.1 (eventually AGPL-3.0) 2016-present Scille SAS
 
-from typing import Type, Optional
+from typing import Type, Optional, List, Tuple
+from pathlib import Path
+import logging
+import trio
 from quart import (
     render_template,
-    Response,
     ResponseReturnValue,
     request,
     redirect as quart_redirect,
     ctx,
 )
 from quart_trio import QuartTrio
+from hypercorn.config import Config as HyperConfig
+from hypercorn.trio import serve
 
 from parsec import __version__ as parsec_version
 from parsec.backend.app import BackendApp
+from parsec.backend.config import BackendConfig
 from parsec.backend.asgi.administration import administration_bp
 from parsec.backend.asgi.redirect import redirect_bp
 from parsec.backend.asgi.rpc import rpc_bp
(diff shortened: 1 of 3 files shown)


parsec/backend/asgi/__init__.py
```diff
@@ -84,4 +79,63 @@ async def page_not_found(e):
     return app
 
 
-__all__ = ("app_factory",)
+def _patch_server_header(backend_config: BackendConfig, hyper_config: HyperConfig) -> None:
+    # By default, Hypercorn uses `hypercorn-h11` as `server` header.
+    # This cannot be customized but only disabled, so we have to rely on function
+    # patching to provide our own custom server header.
```
vxgmichel, line 85: Why not simply add `"include_server_header": False` to the hyper config? The following example seems to work fine on my machine:
```python
import trio
from quart_trio import QuartTrio
from hypercorn.trio import serve
from hypercorn.config import Config

app = QuartTrio(__name__)

@app.route('/')
async def hello():
    return 'hello'

@app.after_request
def add_server_header(response):
    response.headers["server"] = "myserver"
    return response


config = Config()
config.include_server_header = False
trio.run(serve, app, config)
```
Header:
```bash
% curl localhost:8000 --head

HTTP/1.1 200 
content-type: text/html; charset=utf-8
content-length: 5
server: myserver
date: Tue, 09 Aug 2022 13:09:22 GMT
```

touilleMan: Yes, that was the initial solution I've implemented, however
- I didn't like the split between updating config and @app.after_request given they are in different files so it's less clear
- looking into the Hypercorn code, I wasn't clear `app.after_request` would be called in all the place `Config.response_headers` is used

See the standard path:

https://github.com/pgjones/hypercorn/blob/8cef567301a5a3f4363d1c2cad76008e58add8aa/src/hypercorn/protocol/h11.py#L118-L131

vs error path or HTTP 101 path:

https://github.com/pgjones/hypercorn/blob/8cef567301a5a3f4363d1c2cad76008e58add8aa/src/hypercorn/protocol/h11.py#L239-L248

https://github.com/pgjones/hypercorn/blob/8cef567301a5a3f4363d1c2cad76008e58add8aa/src/hypercorn/protocol/h11.py#L146-L150

vxgmichel: Oh I see.

vxgmichel (approved):
LGTM :+1:

PR #2812, opened 2022-08-09
Fix failing parsec.core import in parsec/sequester_export_reader.py when only backend deps are installed

touilleMan:


parsec/sequester_export_reader.py
```diff
@@ -52,6 +52,90 @@ class InconsistentWorkspaceError(RealmExportReaderError):
     pass
 
 
+# This is (sligthly patched) copy of `parsec.core.fs.path.FsPath`, the idea is
+# we cannot import stuff from `parsec.core` if only backend dependencies are
+# installed.
+# On top of that it's complicated to move `FsPath` out of `parsec.core` given
+# it purposefuly uses `parsec.core.fs.exceptions.FSNameTooLongError` exception
+# (that's the part we have patched out here).
+class FsPath:
```
vxgmichel, line 61: This implementation could be replaced  by `pathlib.PurePath` since all we need is a path representation without IO.

This would avoid the code duplication where most methods are not used anyway.

touilleMan: That's a great idea ! I've updated my PR with it ;-)

vxgmichel (approved):
LGTM :+1:

PR #2817, opened 2022-08-10
Improve mypy typing (including the infamous pre-commit !)

touilleMan:


FirelightFlagboy (approved):

.github/actions/use-pre-commit/action.yml
```diff
@@ -30,29 +30,23 @@ runs:
...
-      run: python -m venv ~/.pre-commit-venv
-      shell: bash
 
     - name: Install pre-commit
       if: steps.cache-pre-commit.outputs.cache-hit != 'true'
       run: |
-        source ~/.pre-commit-venv/bin/activate
-        python -m pip install pre-commit==2.19.0
+        set -eux
+        curl --proto '=https' --tlsv1.2 -sSL https://github.com/pre-commit/pre-commit/releases/download/v${{ inputs.version }}/pre-commit-${{ inputs.version }}.pyz > ~/.cache/pre-commit/pre-commit.pyz
```
FirelightFlagboy, line 38:
```diff
-      run: |
-        set -eux
-        curl --proto '=https' --tlsv1.2 -sSL https://github.com/pre-commit/pre-commit/releases/download/v${{ inputs.version }}/pre-commit-${{ inputs.version }}.pyz > ~/.cache/pre-commit/pre-commit.pyz
+      run: curl --proto '=https' --tlsv1.2 -sSL https://github.com/pre-commit/pre-commit/releases/download/v${{ inputs.version }}/pre-commit-${{ inputs.version }}.pyz > ~/.cache/pre-commit/pre-commit.pyz
```

`Github Action` show the first line of `run` since you have 1 command here (excluding the `set ...`) you can put directly `curl` and skipping `set ...`

.github/actions/use-pre-commit/action.yml
```diff
@@ -30,29 +30,23 @@ runs:
...
+        python --version
+        python ~/.cache/pre-commit/pre-commit.pyz --version
       shell: bash
 
     - name: Run pre-commit
       run: |
-        source ~/.pre-commit-venv/bin/activate
-        pre-commit run --show-diff-on-failure --color=always ${{ inputs.extra-args }}
+        set -eux
+        python ~/.cache/pre-commit/pre-commit.pyz run --show-diff-on-failure --color=always ${{ inputs.extra-args }}
```
FirelightFlagboy, line 51:
```diff
-      run: |
-        set -eux
-        python ~/.cache/pre-commit/pre-commit.pyz run --show-diff-on-failure --color=always ${{ inputs.extra-args }}
+      run: python ~/.cache/pre-commit/pre-commit.pyz run --show-diff-on-failure --color=always ${{ inputs.extra-args }}
```

`Github Action` show the first line of `run` since you have 1 command here (excluding the `set ...`) you can put directly `python ~/.cache[...]` and skipping `set ...`

FirelightFlagboy (changes requested):

.github/workflows/quality-assurance.yml
```diff
@@ -47,6 +50,7 @@ jobs:
 
       - name: Debug rust cache
         run: |
+          set -eux
           ls -la ~/.cargo/{.,bin,git,registry} target || true
```
FirelightFlagboy, line 54:
```suggestion
        run: ls -la ~/.cargo/{.,bin,git,registry} target || true
```
`set -eux` isn't needed here since `Github Action` already execute `bash -e`

.github/actions/use-pre-commit/action.yml
touilleMan: I reverted my add of `set -eux` then

However I'm style thinking we should use at least `-x` which is really convenient to see which command has made which output (and while we are at doing `set -x` why not doing the good&explicit `set -eux` ? ^^)

FirelightFlagboy: When doing multiple command in a `run` yes using `set -x` or `set -eux` could be useful.
But with one command, you already see it in the log of github action

touilleMan: I agree, the "set -eux on multi line bash script, nothing on single line script" seems like the right thing to do ;-)

.github/actions/use-pre-commit/action.yml
```diff
@@ -30,29 +30,21 @@ runs:
...
-      if: steps.cache-pre-commit.outputs.cache-hit != 'true'
-      run: python -m venv ~/.pre-commit-venv
-      shell: bash
 
     - name: Install pre-commit
       if: steps.cache-pre-commit.outputs.cache-hit != 'true'
       run: |
-        source ~/.pre-commit-venv/bin/activate
-        python -m pip install pre-commit==2.19.0
+        # Install pre-commit as a standalone .pyz archive that can be run by Python
```
FirelightFlagboy, line 37: Move this comment in the `name` field ? It feel appropriate to be placed here

mypy.ini
```diff
@@ -23,78 +23,78 @@ ignore_errors = False
...
-disallow_any_generics=True
-disallow_subclassing_any=True
-
-[mypy-parsec.api.__init__]
-ignore_errors = True
+; disallow_untyped_defs=True
+; disallow_any_unimported=True
+; disallow_any_decorated=True
+; disallow_any_generics=True
+; disallow_subclassing_any=True
```
FirelightFlagboy, line 55: Why those commented line ?

parsec/api/data/base.py
```diff
@@ -175,7 +176,7 @@ def verify_and_load(
         author_verify_key: VerifyKey,
         expected_author: Optional[DeviceID],
         expected_timestamp: Optional[DateTime] = None,
-        **kwargs,
+        **kwargs: Any,
```
FirelightFlagboy, line 179:
```suggestion
        **kwargs: Dict[str, Any],
```
`**kwargs` can be typed as a `dict` right ?

parsec/api/data/certif.py
```diff
@@ -88,10 +88,10 @@ def is_admin(self) -> bool:
     @classmethod
     def verify_and_load(
         cls,
-        *args,
+        *args: Any,
```
FirelightFlagboy, line 91:
```suggestion
        *args: List[Any],
```
`*args` can be typed as a `list` right ?

.github/actions/use-pre-commit/action.yml
touilleMan: ok

parsec/api/data/base.py
touilleMan: nop, `*args` and `**kwargs` should be typed as the type of the parameter, not the type of the entire `args/kwargs` (this is because args is always a List[Any] and kwargs as Dict[str, Any]

parsec/api/data/certif.py
touilleMan: same as above

mypy.ini
touilleMan: debug stuff I forgot to remove, done.

parsec/api/data/base.py
FirelightFlagboy: ok

parsec/api/data/certif.py
FirelightFlagboy: ok

FirelightFlagboy (approved):

.github/actions/use-pre-commit/action.yml
FirelightFlagboy: I was expecting to be put in the field `name`, to produce something like

```yaml
name: Install pre-commit as a standalone .pyz archive that can be run by Python
```

PR #2828, opened 2022-08-12
[WIP]Webhook sequester service

AntoineDupre:


AntoineDupre:
related issue https://github.com/Scille/resana-secure/issues/20

AntoineDupre:
Also fixes #2799

AntoineDupre:
also fixes #3133

touilleMan (changes requested):

oxidation/libparsec/crates/protocol/schema/authenticated_cmds/vlob_create.json
```diff
@@ -116,6 +108,38 @@
...
+                    "type": "String"
+                },
+                {
+                    "name": "service_error",
+                    "type": "String"
+                }
+            ]
+        },
+        {
+            "status": "sequester_webhook_failed",
```
touilleMan, line 129: Should we make this a more generic error ?
from my point of view the sequester can provided two kind of errors:
1) temporary failure in the sequester (for instance the webhook couldn't be reached), this error means the client should retry the sync later on.
2) hard failure in the sequester: a sequester service explicitly said this manifest is not to be synced.

For 1), a very simple implementation for this is for the server to close the client socket if it occurs (this way not change in the client and no need for a specific status \o/).

For 2) things get more complicated: the client should stop trying to sync the corresponding file (I believe this PR already implement this) AND the other sequester services that have been notified with the data should receive an erratum (given they have stored a given vlob_atom that is in fact not present in the server database !)
The simple solution to this is to consider the sequester services trust the server and replace the vlob_atom if a new one with same vlobid/version is sent to them (so there is nothing to do for webhook service given we store nothing, but for the storage service we must change the sql query a bit and add a test)

In any case we should try to avoid avoid an error specific to a kind of sequester service, this would simplify adding new type of sequester in the future.

touilleMan (changes requested):

parsec/backend/sequester.py
```diff
@@ -51,13 +52,24 @@ class SequesterServiceAlreadyEnabledError(SequesterError):
...
+
+
 @attr.s(slots=True, frozen=True, repr=False, auto_attribs=True)
 class SequesterService:
     service_id: SequesterServiceID
     service_label: str
     service_certificate: bytes
     created_on: DateTime = attr.ib(factory=DateTime.now)
     disabled_on: Optional[DateTime] = None
+    service_type: SequesterServiceType = SequesterServiceType.STORAGE
```
touilleMan, line 71: Do we need this ?
I'd rather not have a default value on this given when we build a `SequesterService` object we should entirely define it to avoid silly errors

I guess this is to avoid changes in the tests, can you have a look how much change are requested (if it's too much we can leave as this ^^)

parsec/backend/vlob.py
```diff
@@ -141,6 +181,68 @@ def __init__(
...
+
+        try:
+            await http_request(
+                url=f"{service.webhook_url}/{organization_id.str}",
+                method="POST",
+                data=sequester_data,
+            )
+        except urllib.error.HTTPError as exc:
+            if exc.code == 400:
+                body = json.loads(exc.read())
```
touilleMan, line 220: no try/except in case the body is not json ?

parsec/backend/vlob.py
```diff
@@ -141,6 +181,68 @@ def __init__(
...
+                raise VlobSequesterWebhookServiceRejectedError(
+                    service_id=service.service_id,
+                    service_label=service.service_label,
+                    error=body.get("error", ""),
```
touilleMan, line 224: Given we provide the response body as an opaque structure, we can instead directly pass `exc.read()`, this way we avoid error handling in json and bytes-to-string decoding

The drawback is we will flood the log if the body is one giant binary blob, but it seems unlikely (and we can provide something like `exc.read()[:2048]` if we really want to avoid this)

oxidation/libparsec/crates/protocol/schema/authenticated_cmds/vlob_update.json
```diff
@@ -114,6 +106,38 @@
                     "type": "List<Bytes>"
                 }
             ]
+        },
+        {
+            "status": "sequester_rejected",
```
touilleMan, line 111: same as for vlob_create

parsec/backend/cli/sequester.py
```diff
@@ -66,6 +69,9 @@ def _display_service(service: SequesterService) -> None:
     display_service_label = click.style(service.service_label, fg="green")
     click.echo(f"Service {display_service_label} (id: {display_service_id})")
     click.echo(f"\tCreated on: {service.created_on}")
+    click.echo(f"\tService type: {service.service_type}")
+    if service.service_type == SequesterServiceType.WEBHOOK:
+        click.echo(f"\tSequester service URL {service.webhook_url}")
```
touilleMan, line 74:
```suggestion
        click.echo(f"\tWebhook endpoint URL {service.webhook_url}")
```

parsec/backend/vlob.py
```diff
@@ -131,6 +141,36 @@ class VlobSequesterDisabledError(VlobError):
     pass
 
 
+class VlobSequesterWebhookServiceRejectedError(VlobError):
+    def __init__(self, service_id, service_label, error, *args, **kwargs):
+        self.service_id = service_id
+        self.service_label = service_label
+        self.error = error
+        VlobError.__init__(self, *args, **kwargs)
```
touilleMan, line 149:
```suggestion
        super().__init__(self, *args, **kwargs)
```

parsec/backend/vlob.py
```diff
@@ -141,6 +181,68 @@ def __init__(
...
 
+async def extract_sequestered_data_and_proceed_webhook(
+    services: Dict[SequesterServiceID, SequesterService],
+    organization_id: OrganizationID,
+    author: DeviceID,
+    encryption_revision: int,
+    vlob_id: VlobID,
+    timestamp: DateTime,
+    sequester_blob: Dict[SequesterServiceID, bytes],
+):
```
touilleMan, line 192: return typing

parsec/backend/vlob.py
```diff
@@ -141,6 +181,68 @@ def __init__(
...
+    ]
+    webhook_service_ids = [
+        service.service_id
+        for service in services.values()
+        if service.service_type == SequesterServiceType.WEBHOOK
+    ]
+    for webhook_service_id in webhook_service_ids:
+        service = services[webhook_service_id]
+        sequester_data = sequester_blob[webhook_service_id]
+        if not service.webhook_url:
```
touilleMan, line 207: This should never occur in theory given the service object comes from the database and has been validated when inserted.

However it's a good idea to keep this check, but I think we don't need a specific type of exception for this: just log an error (this time it's an error level log and not a warning given it shows an issue with our code) and just pretend the url was provided but didn't answer to the http request

parsec/core/core_events.py
```diff
@@ -22,6 +22,8 @@ class CoreEvent(Enum):
     FS_ENTRY_UPDATED = "fs.entry.updated"
     FS_ENTRY_FILE_CONFLICT_RESOLVED = "fs.entry.file_conflict_resolved"
     FS_WORKSPACE_CREATED = "fs.workspace.created"
+    # Webhook
+    WEBHOOK_UPLOAD_REJECTED_ERROR = "webhook.upload_rejected_error"
```
touilleMan, line 26: Generic name plz, something like `FS_ENTRY_SYNC_REFUSED_BY_SEQUESTER_SERVICE`

touilleMan (changes requested):

parsec/backend/vlob.py
```diff
@@ -309,6 +428,25 @@ async def api_vlob_update(self, client_ctx, req: VlobUpdateReq) -> VlobUpdateRep
                 sequester_services_certificates=exc.sequester_services_certificates,
             )
 
+        # TODO Change return value
```
touilleMan, line 431: todo ?

parsec/core/fs/remote_loader.py
```diff
@@ -798,6 +801,8 @@ async def upload_manifest(
             # Update our cache and retry the request
             self._sequester_services_cache = sequester_services
             return await self.upload_manifest(entry_id, manifest)
+        except VlobSequesterRejectedError as exc:
```
touilleMan, line 804: Why catching an rebuilding the same error ?

If we want to be more explicit about the fact this exception may bubble up, we can write:
```python
except VlobSequesterRejectedError as exc:
    # You didn't expect me to exist, don't you ?
    raise
```

parsec/core/fs/remote_loader.py
```diff
@@ -850,6 +855,8 @@ async def _vlob_create(
                 sequester_authority_certificate=rep.sequester_authority_certificate,
                 sequester_services_certificates=rep.sequester_services_certificates,
             )
+        elif isinstance(rep, VlobCreateRepSequesterRejected):
+            raise VlobSequesterRejectedError(id=entry_id)
```
touilleMan, line 859: no manifest provided ?

Don't we want to catch this in the caller `upload_manifest` so that we have the manifest object at hand ?

touilleMan (approved):
lgtm (finally !)

PR #2829, opened 2022-08-12
Yolo/improve workflow ci

FirelightFlagboy:
- Add mock workflow that will allow to setup required jobs to success
- Update watching path on `test-ci...` to watch for dependent local actions

### DIFF
--- a/.github/workflows/mock-test-python-ci.yml
+++ b/.github/workflows/mock-test-python-ci.yml
@@ -2,15 +2,16 @@ name: test-python-ci
 
 on:
   pull_request:
-    paths-ignore:
-      - parsec/**
-      - tests/**
-      - src/**.rs
-      - pyproject.toml
-      - poetry.lock
-      - .github/workflows/test-python-ci.yml
-      - .github/actions/setup-python-poetry/**
-      - .github/actions/setup-rust/**
+    paths:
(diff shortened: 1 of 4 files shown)


touilleMan (approved):

.github/workflows/mock-test-python-ci.yml
```diff
@@ -0,0 +1,21 @@
+name: test-python-ci
+
+on:
+  pull_request:
+    paths:
+      - "**"
+      - "!parsec/**"
+      - "!tests/**"
+      - "!src/**.rs"
```
touilleMan, line 9: This is not matching the regex in `test-python-ci.yml` (it is `src/**` there)

touilleMan: While you're fixing that, can you also add a comment in each file `# Must be kept in sync with xxx/yyy.yml !`

touilleMan (changes requested):

.github/workflows/mock-test-python-ci.yml
FirelightFlagboy: - for the matching pattern (not regex), done in c766a41
- for the comment, done in bc8a5f5

PR #2836, opened 2022-08-12
Switch strategy for building the rust extension in build.py

vxgmichel:
A proposal to switch the strategy for building the rust extension in build.py

This approach does not require pip nor a virtualenv, which makes it much more robust.

This is especially useful for fixing the CI in parsec-extensions: https://github.com/Scille/parsec-extensions/commit/1450b88f1587f407c85583e01bbc0c81ae33f38b

### DIFF
--- a/build.py
+++ b/build.py
@@ -3,6 +3,9 @@
 import os
 import sys
 import subprocess
+import tempfile
+import zipfile
+import pathlib
 
 
 def display(line: str):
@@ -13,6 +16,7 @@ def display(line: str):
     print(f"{YELLOW_FG}{line}{DEFAULT_FG}", flush=True)
 
 
+BASEDIR = pathlib.Path(__file__).parent
 PYTHON_EXECUTABLE_PATH = sys.executable
 display(f"PYTHON_EXECUTABLE_PATH={PYTHON_EXECUTABLE_PATH}")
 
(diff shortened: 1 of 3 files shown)


build.py
```diff
@@ -27,36 +27,19 @@ def in_cibuildwheel():
...
-    run(f"maturin develop {release}")
-    # An alternative to `maturin develop` is:
-    # run("maturin build -r")
-    # run("unzip -o target/wheels/parsec-*.whl parsec/_parsec.abi3.so")
-    # Note that it does not require a virtualenv, as opposed to `develop`
+    run(f"maturin build {release}")
+    if sys.platform == "win32":
+        run("unzip -o target/wheels/parsec-*.whl parsec/_parsec.pyd")
+    else:
+        run("unzip -o target/wheels/parsec-*.whl parsec/_parsec.abi3.so")
```
FirelightFlagboy, line 42: unzipping `parsec-*.whl` work on the CI, but could cause errors if we have multiple files that match the pattern ` parsec-*.whl` (e.g.: multiple version)

If we want the use a fixed version, we could set the version in:

- `cargo.toml:<package.version>`
- `pyproject.toml:<project.version>`

FirelightFlagboy (approved):

build.py
```diff
@@ -27,36 +27,19 @@ def in_cibuildwheel():
...
-    check_venv()
     release = "--release" if in_cibuildwheel() or force_maturin_release() else ""
-    run(f"maturin develop {release}")
-    # An alternative to `maturin develop` is:
-    # run("maturin build -r")
-    # run("unzip -o target/wheels/parsec-*.whl parsec/_parsec.abi3.so")
-    # Note that it does not require a virtualenv, as opposed to `develop`
+    run(f"maturin build {release}")
+    if sys.platform == "win32":
+        run("unzip -o target/wheels/parsec-*.whl parsec/_parsec.pyd")
```
touilleMan, line 40: I'd rather use python's zip library instead of the system one (otherwise we will soon discover which version of Windows/FooBarLinux is missing this command ^^)

touilleMan (changes requested):

build.py
touilleMan: implemented by afde298fd1fc4c7eadd89090fb57ffb97ead76e0

build.py
touilleMan: implemented by afde298fd1fc4c7eadd89090fb57ffb97ead76e0

touilleMan (approved):

FirelightFlagboy (approved):

FirelightFlagboy (approved):

PR #2843, opened 2022-08-17
Improve jobs names in CI with shinny emojies ;-)

touilleMan:


touilleMan:
branch renamed in #2844

PR #2849, opened 2022-08-17
Modification in .github/ force all tests in CI

touilleMan:


.github/actions/paths-filter/filters.yml
```diff
@@ -1,15 +1,18 @@
...
   - parsec/**
   - tests/**
   - pyproject.toml
   - poetry.lock
   - build.py
 rust-ext:
+  - .github/**
   - src/**/*.rs
   - Cargo.toml
   - Cargo.lock
```
FirelightFlagboy, line 18:
```suggestion
rust:
  - oxidation/**/*.rs
  - oxidation/**/*.toml
  - Cargo.toml
  - Cargo.lock
python:
  - parsec/**
  - tests/**
  - pyproject.toml
  - poetry.lock
  - build.py
rust-ext:
  - src/**/*.rs
  - Cargo.toml
  - Cargo.lock
ci:
  - .github/actions/**
  - .github/workflows/**
```
Add another field instead

FirelightFlagboy (changes requested):

.github/actions/paths-filter/filters.yml
touilleMan: I'd rather stay KISS here (on top of that, listing subfolders on `.github/` means we will have issues if we add something new in the `.github/`directly)

FirelightFlagboy: maybe, listening to the subfolder on `.github` is to much.
But I still think adding the field `ci` is better

touilleMan: >  But I still think adding the field ci is better

I don't see a direct benefit (except from the "we named it "rust" and it contains ci stuff which is not technically Rust code", hence KISS)

FirelightFlagboy (approved):

PR #2855, opened 2022-08-18
Replace BlockCreate & BlockRead

TimeEngineer:


tests/schemas/builder.py
```diff
@@ -267,7 +267,10 @@ def generate_api_protocol_specs():
                 except KeyError:
                     cmd_serializer = cmd_serializers[cmd_name]
             else:
-                cmd_serializer = cmd_serializers[cmd_name]
+                try:
+                    cmd_serializer = cmd_serializers[cmd_name]
+                except:
```
touilleMan, line 272: image

touilleMan (changes requested):

tests/schemas/builder.py
TimeEngineer: What should be the proper way ?

parsec/_parsec_pyi/protocol.pyi
```diff
@@ -0,0 +1,64 @@
...
+
+class InvitedAnyCmdReq:
+    def __repr__(self) -> str: ...
+    def dump(self) -> bytes: ...
+    def load(buf: bytes) -> BlockReadReq: ...
+
+class BlockCreateReq:
+    def __init__(self, block_id: BlockID, realm_id: RealmID, block: bytes) -> None: ...
+    def __repr__(self) -> str: ...
+    def __eq__(self, other: BlockCreateReq) -> bool: ...
```
touilleMan, line 17: In which case do we need `__eq__/__ne__` operators ? (I guess this is for tests)

parsec/_parsec_pyi/protocol.pyi
```diff
@@ -0,0 +1,64 @@
...
+    def block(self) -> bytes: ...
+    def dump(self) -> bytes: ...
+
+class BlockCreateRep:
+    def __repr__(self) -> str: ...
+    def __eq__(self, other: BlockCreateRep) -> bool: ...
+    def __ne__(self, other: BlockCreateRep) -> bool: ...
+    def is_ok(self) -> bool: ...
+    def Ok() -> BlockCreateRep: ...
+    def AlreadyExists() -> BlockCreateRep: ...
```
touilleMan, line 33: This is really cool 😎

parsec/api/protocol/block.py
```diff
@@ -1,100 +1,48 @@
...
-BlockReadRepType = Union[
-    BlockReadRepOk,
-    BlockReadRepNotFound,  # type: ignore[valid-type]
-    BlockReadRepTimeout,  # type: ignore[valid-type]
-    BlockReadRepNotAllowed,  # type: ignore[valid-type]
-    BlockReadRepInMaintenance,  # type: ignore[valid-type]
-]
+class BlockReadSerializer:
+    @staticmethod
+    def req_dumps(req) -> bytes:  # type: ignore[no-untyped-def]
```
touilleMan, line 36: Remove the type ignore and use `req: dict` instead

parsec/api/protocol/block.py
```diff
@@ -1,100 +1,48 @@
...
+        return BlockReadReq(req["block_id"]).dump()
 
-block_read_serializer = CmdSerializer.from_typed(BlockReadReq, BlockReadRep.TYPES)
+    @staticmethod
+    def rep_loads(raw: bytes) -> BlockReadRep:
```
touilleMan, line 40: We only implement `req_dumps` and `rep_loads` because the server side no longer uses this class ? (and soon we will have the backend_connection client part in Rust and this serializer would be entirely removed \o/)

touilleMan (changes requested):

parsec/_parsec_pyi/protocol.pyi
TimeEngineer: Yes for tests

parsec/core/backend_connection/cmds.py
```diff
@@ -106,16 +106,19 @@ async def _send_cmd(transport: Transport, serializer, **req) -> dict:
...
-        transport.logger.error("Invalid request data according to backend", cmd=req["cmd"], rep=rep)
-        raise BackendProtocolError("Invalid request data according to backend")
-
-    if rep["status"] == "bad_timestamp":
-        raise BackendOutOfBallparkError(rep)
-
-    # Backward compatibility with older backends (<= v2.3)
-    if rep["status"] == "invalid_certification" and "timestamp" in rep["reason"]:
-        raise BackendOutOfBallparkError(rep)
+    if isinstance(rep, Dict):
```
touilleMan, line 109: `isinstance(rep, dict)` (`Dict` is the typing thing so it's not Pythonic to use it here)

parsec/core/fs/remote_loader.py
```diff
@@ -447,21 +447,21 @@ async def load_block(self, access: BlockAccess) -> None:
...
             # Seems we lost the access to the realm
             raise FSWorkspaceNoReadAccess("Cannot load block: no read access")
-        elif rep["status"] == "in_maintenance":
+        elif rep == BlockReadRep.InMaintenance():
             raise FSWorkspaceInMaintenance(
                 "Cannot download block while the workspace in maintenance"
             )
-        elif rep["status"] != "ok":
-            raise FSError(f"Cannot download block: `{rep['status']}`")
+        elif not (rep.is_ok()):
```
touilleMan, line 459: no need for the parenthesis

parsec/core/fs/remote_loader.py
```diff
@@ -447,21 +447,21 @@ async def load_block(self, access: BlockAccess) -> None:
         # Download
         with translate_backend_cmds_errors():
             rep = await self.backend_cmds.block_read(access.id)
-        if rep["status"] == "not_found":
+        if rep == BlockReadRep.NotFound():
```
touilleMan, line 450: use `isinstance(rep, BlockReadRep.NotFound)` instead

and brace yourself, [structural matching is avaialabe in Python 3.10](https://peps.python.org/pep-0636/) ^^

parsec/api/protocol/block.py
TimeEngineer: I don't know :man_shrugging: but yes, many things can be removed IMO

tests/backend/common.py
```diff
@@ -161,7 +161,10 @@ async def _do_recv(self, ws, check_rep):
         rep = self.serializer.rep_loads(raw_rep)
 
         if check_rep:
-            assert rep["status"] == "ok"
+            try:
```
touilleMan, line 164:
```python
if isinstance(rep, dict):
    assert rep["status"] == "ok"  # Legacy rep schemas
else:
    assert rep.is_ok()  # Rust-based rep schemas

File: tests/backend/realm/test_block.py [hunk 10]
```diff
@@ -355,14 +355,14 @@ async def mock_read(organization_id, id):
 async def test_block_create_bad_msg(alice_ws, bad_msg):
     await alice_ws.send(packb({"cmd": "block_create", **bad_msg}))
     raw_rep = await alice_ws.receive()
-    rep = block_create_serializer.rep_loads(raw_rep)
-    assert rep["status"] == "bad_message"
+    with pytest.raises(ValueError):
```
touilleMan: This was an actual bug ! 😭

tests/backend/realm/test_block.py
```diff
@@ -379,11 +379,11 @@ async def test_block_read_not_found(alice_ws):
 async def test_block_read_bad_msg(alice_ws, bad_msg):
     await alice_ws.send(packb({"cmd": "block_read", **bad_msg}))
     raw_rep = await alice_ws.receive()
-    rep = block_read_serializer.rep_loads(raw_rep)
-    # Valid ID doesn't exists in database but this is ok given here we test
-    # another layer so it's not important as long as we get our
-    # `bad_message` status
-    assert rep["status"] == "bad_message"
+    with pytest.raises(ValueError):
```
touilleMan: same as above 😭

parsec/api/protocol/block.py
```diff
@@ -1,100 +1,48 @@
...
-BlockReadRepNotAllowed = cmd_rep_error_type_factory("BlockReadRepNotAllowed", "not_allowed")
-BlockReadRepInMaintenance = cmd_rep_error_type_factory(
-    "BlockReadRepInMaintenance", "in_maintenance"
-)
+    @staticmethod
+    def rep_loads(raw: bytes) -> BlockCreateRep:
+        try:
+            return BlockCreateRep.load(raw)
+        except ValueError as exc:
+            raise exc
```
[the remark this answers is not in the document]
TimeEngineer: Sorry it's due to my shortcomings in python :sweat_smile:

parsec/core/backend_connection/cmds.py
touilleMan: also this mean we don't check for those in the case of block read/create which is an issue for `rep["status"] == "invalid_msg_format"`

And it will become more of an issue when we will migrate other schema and forget to take into account this part...

Instead we'd better do a
```python
if isinstance(rep, (BlockReadRep, BlockCreateRep)):
    # New Rust rep schemas
    status = rep.status
    reason = getattr(rep, "reason", "")
else:
    # Legacy Python rep schemas
    status = rep["status]
    reason = rep.get("reason", "")

if status == "invalid_msg_format":
    [...]

return rep
```

This way we ensure this place will break everytime a new schema is migrated, hence we must check this place en make sure it behavior is expected.

parsec/api/protocol/block.py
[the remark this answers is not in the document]
TimeEngineer: The PR changes tests and now handle the `ProtocolError` instead of checking `rep["status"]`

parsec/_parsec_pyi/protocol.pyi
touilleMan: How are handled the unkown error ?

for instance if the server returns a `{"status: "unknown_command"}` ?

cf. the generic errors statuses that can raise the server no matter what is the req:

https://github.com/Scille/parsec-cloud/blob/f9639c0982f47c811711098d4532c4c87da59ae6/parsec/backend/asgi/ws.py#L176-L217

On top of that the server may be more recent than the client, in this case it might return a status that we don't understand but which is valid (adding error type is considered a minor API bump)

(I remember we say that we should have a special type representing an unknown error for this case)

touilleMan: Aaaaand we should have a test that ensure unknown status are correctly handled by those new kind of schema classes (typically add it into `tests/api/test_protocol.py`

touilleMan (changes requested):

tests/backend/realm/test_block.py
TimeEngineer: Solved previously

tests/backend/realm/test_block.py
TimeEngineer: Same

parsec/core/fs/remote_loader.py
TimeEngineer: I was searching the line, thanks !

parsec/core/fs/remote_loader.py
TimeEngineer: Are you sure that `BlockReadRep.NotFound` is a type ?

touilleMan: if it's not a type, then what `BlockReadRep.NotFound()` does ? ^^

TimeEngineer: It creates an object

parsec/_parsec_pyi/protocol.pyi
TimeEngineer: UnknownError no longer exists

parsec/api/protocol/block.py
touilleMan: dict[str, Any]

tests/backend/realm/test_block.py
```diff
@@ -356,13 +368,13 @@ async def test_block_create_bad_msg(alice_ws, bad_msg):
     await alice_ws.send(packb({"cmd": "block_create", **bad_msg}))
     raw_rep = await alice_ws.receive()
     rep = block_create_serializer.rep_loads(raw_rep)
-    assert rep["status"] == "bad_message"
+    assert isinstance(rep, BlockCreateRepUnknownStatus)
```
touilleMan, line 371: this check is not equivalent to the previous one, we must check the status is `bad_message`

so I guess we shoul do something like `rep == BlockCreateRepUnknownStatus("bad_message")`

tests/backend/realm/test_block.py
```diff
@@ -313,7 +325,7 @@ async def mock_read(organization_id, id):
 
     rep = await block_read(alice_ws, block)
     # A bad chunk result in a bad block, which should be detected by the client
-    assert rep == {"status": "ok", "block": ANY}
+    assert rep == BlockReadRepOk(BLOCK_DATA[:3] + bad_chunk[: len(BLOCK_DATA) - 3])
```
touilleMan, line 328: use `isinstance(rep, BlockReadRepOk)` instead (the `BLOCK_DATA[:3] + bad_chunk[: len(BLOCK_DATA) - 3]` is complex to read and make the reader think this is a relevant part of the test while the test is about the fact that bad data in a blockstore doesn't cause a crash in the server)

parsec/api/protocol/block.py
```diff
@@ -1,100 +1,42 @@
...
-    class SCHEMA_CLS(BaseTypedRepSchema):
-        status = fields.CheckedConstant("ok", required=True)
-        block = fields.Bytes(required=True)
-
-    block: bytes
 
+class BlockCreateSerializer:
+    @staticmethod
+    def req_dumps(req: dict[str, Any]) -> bytes:
+        return BlockCreateReq(req["block_id"], req["realm_id"], req["block"]).dump()
```
touilleMan, line 24: use `BlockCreateReq(**req)` instead

this will fail if the wrong arguments are provided, and it's generic over all the commands so you'll thank more for the next commands ^^

touilleMan: And given the serializer is all generic with this trick, we can have a common `ApiCommandSerializer` class for all commands:
```python
class ApiCommandSerializer:
    def __init__(self, req_schema, rep_schema):
        self.req_schema = req_schema
        self.rep_schema = rep_schema
    def req_dumps(self, req: dict[str, Any]) -> bytes:
        return self.req_schema(**req).dump()
    def rep_loads(self, raw: bytes) -> Any:
        return self.rep_schema.load(raw)

block_create_serializer = ApiCommandSerializer(BlockCreateReq, BlockCreateRep)
```

tests/backend/realm/test_block.py
TimeEngineer: We use `assert rep.status == "bad_message"` instead, because the only `Rep` which have the field `status` are `UnknownStatusRep `

touilleMan (approved):

PR #2866, opened 2022-08-19
Remove `_Py` classes in `manifest` python module

### LINKED ISSUE #2865: Remove `_Py*` classes in `parsec/api/data/manifest.py`
Remove `_Py*` classes in `parsec/api/data/manifest.py` like this had been done in #2838

FirelightFlagboy:
depends on #2825

closes #2865

### DIFF
--- a/oxidation/libparsec/crates/types/src/manifest.rs
+++ b/oxidation/libparsec/crates/types/src/manifest.rs
@@ -491,4 +491,68 @@ impl Manifest {
         let blob = key.decrypt(encrypted).map_err(|_| "Invalid encryption")?;
         rmp_serde::from_slice(&blob).map_err(|_| "Invalid deserialization")
     }
+
+    pub fn decrypt_verify_and_load(
+        encrypted: &[u8],
+        key: &SecretKey,
+        author_verify_key: &VerifyKey,
+        expected_author: &DeviceID,
+        expected_timestamp: DateTime,
+    ) -> Result<Self, DataError> {
+        let signed = key.decrypt(encrypted)?;
+
+        Self::verify_and_load(
+            &signed,
+            author_verify_key,
+            expected_author,
+            expected_timestamp,
+        )
+    }
+
+    pub fn verify_and_load(
+        signed: &[u8],
+        author_verify_key: &VerifyKey,
+        expected_author: &DeviceID,
+        expected_timestamp: DateTime,
+    ) -> Result<Self, DataError> {
+        let compressed = author_verify_key.verify(&signed)?;
+        let mut deserialized = Vec::new();
+
+        ZlibDecoder::new(&compressed)
+            .read_to_end(&mut deserialized)
+            .map_err(|_| DataError::Compression)?;
+
+        let obj: Self =
+            rmp_serde::from_slice(&deserialized).map_err(|_| DataError::Serialization)?;
+
+        macro_rules! internal_verify {
+            ($obj:ident, $expected_author:ident, $expected_timestamp:ident) => {
+                if $obj.author != *$expected_author {
+                    Err(DataError::UnexpectedAuthor {
+                        expected: $expected_author.clone(),
+                        got: Some($obj.author),
+                    })
+                } else if $obj.timestamp != $expected_timestamp {
+                    Err(DataError::UnexpectedTimestamp {
+                        expected: $expected_timestamp,
+                        got: $obj.timestamp,
+                    })
+                } else {
+                    Ok($obj)
+                }
+            };
+        }
+
+        match obj {
+            Manifest::File(file) => internal_verify!(file, expected_author, expected_timestamp),
+            Manifest::Folder(folder) => {
+                internal_verify!(folder, expected_author, expected_timestamp)
+            }
+            Manifest::Workspace(workspace) => {
+                internal_verify!(workspace, expected_author, expected_timestamp)
+            }
+            Manifest::User(user) => internal_verify!(user, expected_author, expected_timestamp),
+        }?;
+        Ok(obj)
+    }
 }
--- a/oxidation/libparsec/crates/types/src/manifest.rs
+++ b/oxidation/libparsec/crates/types/src/manifest.rs
@@ -51,6 +51,8 @@ macro_rules! impl_manifest_dump_load {
                 author_verify_key: &VerifyKey,
                 expected_author: &DeviceID,
                 expected_timestamp: DateTime,
+                expected_id: Option<EntryID>,
+                expected_version: Option<u32>,
             ) -> Result<Self, DataError> {
                 let signed = key.decrypt(encrypted)?;
 
@@ -59,6 +61,8 @@ macro_rules! impl_manifest_dump_load {
                     author_verify_key,
                     expected_author,
                     expected_timestamp,
+                    expected_id,
+                    expected_version,
                 )
             }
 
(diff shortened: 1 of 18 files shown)


tests/common/helpers.py
```diff
@@ -230,3 +230,11 @@ def _next_timestamp():
             raise RuntimeError("Is DateTime.now() frozen ?")
 
     return _next_timestamp
+
+
+import structlog
+
+
+@pytest.fixture
+def log() -> structlog.BoundLogger:
```
touilleMan, line 239: What is this for ?

Just use print in test instead (pytest takes care of capturing stdout when the test doesn't run with --verbose)

tests/api/test_types.py
```diff
@@ -248,3 +254,43 @@ def test_local_manifests_load_invalid_data():
         # Valid to deserialize, invalid fields
         with pytest.raises(DataError):
             cls.decrypt_and_load(valid_msgpack_but_bad_fields, key=key)
+
+
+def test_remote_manifests_load_invalid_data(alice: LocalDevice):
+    key = SecretKey.generate()
+    valid_zip_msgpack_but_bad_fields = zlib.compress(packb({"foo": 42}))
```
touilleMan, line 261: can you also add a test with
- `valid_zip_bud_bad_msgpack = zlib.compress(b"dummy")`
- `invalid_zip = b"dummy"`

those are cheap tests and cover real things ;-)

touilleMan: (and if you can also add the `invalid_zip` to `test_local_manifests_load_invalid_data` too)

tests/core/fs/workspacefs/test_workspace_fs.py
```diff
@@ -386,23 +394,43 @@ async def test_dump(alice_workspace):
...
 
-    vanilla_file_manifest_deserialize = BaseRemoteManifest._deserialize
+    from parsec.core.backend_connection.authenticated import BackendAuthenticatedCmds
 
-    def mocked_file_manifest_deserialize(*args, **kwargs):
-        return vanilla_file_manifest_deserialize(*args, **kwargs).evolve(**manifest_modifiers)
+    vanilla_vlob_read = BackendAuthenticatedCmds.vlob_read
 
-    monkeypatch.setattr(BaseRemoteManifest, "_deserialize", mocked_file_manifest_deserialize)
+    async def _vlob_read_patched(transport, encryption_revision, vlob_id, version, timestamp):
```
touilleMan, line 410: hold my beer time 😄 🍺

oxidation/libparsec_python/src/manifest.rs
```diff
@@ -395,6 +395,8 @@ impl FileManifest {
             &author_verify_key.0,
             &expected_author.0,
             expected_timestamp.0,
+            None,
```
touilleMan, line 398: Shouldn't we remove this file now ?

oxidation/libparsec/crates/types/tests/test_manifest.rs
```diff
@@ -101,6 +101,8 @@ fn serde_file_manifest(alice: &Device) {
         &alice.verify_key(),
         &alice.device_id,
         now,
+        None,
```
touilleMan, line 104: Can you add a separate tests to check that the `expected_*` fields works as expected ?

something like:

```rs
fn test_file_manifest_expected_checks() {
  let manifest = FileManifest(...);
  let encrypted = manifest.dump_sign_and_encrypt(...)

  // Invalid expects

  assert_eq!(FileManifest.decrypt_verify_and_load(
    encrypted,
    bad_author,
    expected_timestamp,
    None,
    None,
  ), Err(...))

  assert_eq!(FileManifest.decrypt_verify_and_load(
    encrypted,
    expected_author,
    bad_timestamp,
    None,
    None,
  ), Err(...))

  assert_eq!(FileManifest.decrypt_verify_and_load(
    encrypted,
    expected_author,
    expected_timestamp,
    bad_id,
    None,
  ), Err(...))

  assert_eq!(FileManifest.decrypt_verify_and_load(
    encrypted,
    expected_author,
    expected_timestamp,
    expected_id,
    bad_version,
  ), Err(...))

  // Sanity check with good expects

  FileManifest.decrypt_verify_and_load(
    encrypted,
    expected_author,
    expected_timestamp,
    None,
    None,
  ).unwrap()

  FileManifest.decrypt_verify_and_load(
    encrypted,
    expected_author,
    expected_timestamp,
    expected_id,
    expected_version,
  ).unwrap()
}
```

touilleMan (changes requested):

tests/common/helpers.py
FirelightFlagboy: done

oxidation/libparsec_python/src/manifest.rs
FirelightFlagboy: no, `manifest` is used by `local_manifest` which is used by `storage` & `file_operation`

tests/api/test_types.py
FirelightFlagboy: done

oxidation/libparsec/crates/types/tests/test_manifest.rs
FirelightFlagboy: I would add a test for `verify` that would also test `decrypt_verify_and_load`

FirelightFlagboy: done

oxidation/libparsec/crates/types/src/manifest.rs
```diff
@@ -63,18 +86,45 @@ macro_rules! impl_manifest_dump_load {
...
+                obj.verify(
+                    expected_author,
+                    expected_timestamp,
+                    expected_id,
+                    expected_version,
+                )?;
+                Ok(obj)
+            }
+
+            /// Verify the manifest against a set of expected value
```
FirelightFlagboy, line 98:
```suggestion
            /// Verify the manifest against a set of expected values
```

oxidation/libparsec/crates/types/src/manifest.rs
```diff
@@ -51,8 +51,31 @@ macro_rules! impl_manifest_dump_load {
...
+                    author_verify_key,
+                    expected_author,
+                    expected_timestamp,
+                    expected_id,
+                    expected_version,
+                )
+            }
+
+            /// Verify the signed value using the given `verify_key`
+            /// And then, it will check for the expected `author` & `timestamp`
```
FirelightFlagboy, line 70:
```suggestion
            /// And then, it will check for the expected values
```

oxidation/libparsec/crates/types/src/manifest.rs
```diff
@@ -63,18 +86,45 @@ macro_rules! impl_manifest_dump_load {
...
+            }
+
+            /// Verify the manifest against a set of expected values
+            pub fn verify(
+                &self,
+                expected_author: &DeviceID,
+                expected_timestamp: DateTime,
+                expected_id: Option<EntryID>,
+                expected_version: Option<u32>,
+            ) -> Result<&Self, DataError> {
```
touilleMan, line 105: Why returning `&Self` instead of `()` ?

(From what I've seen we always do `obj.verify([...])?;` given the verify always occurs just after object creation, so we always want the ownership on the object to return it to the our caller)

touilleMan (approved):

oxidation/libparsec/crates/types/src/manifest.rs
FirelightFlagboy: Because the block that verify the values was in `verify_and_load` and was returning `Self`, but now that block is in `verify` method.

I will change the returned value

PR #2887, opened 2022-08-22
Build web app in CI

FirelightFlagboy:
- Add job to build the Vue webapp in the CI

### DIFF
--- a/.github/actions/paths-filter/filters.yml
+++ b/.github/actions/paths-filter/filters.yml
@@ -1,17 +1,23 @@
-rust:
+shared: &shared
   - .github/**
+
+rust:
+  - *shared
   - oxidation/**/*.rs
   - oxidation/**/*.toml
   - Cargo.toml
   - Cargo.lock
+
 python:
-  - .github/**
+  - *shared
   - parsec/**
   - tests/**
   - pyproject.toml
   - poetry.lock
   - build.py
+
 rust-ext:
+  - *shared
   - .github/**
   - src/**/*.rs
   - Cargo.toml
--- a/.github/actions/paths-filter/filters.yml
+++ b/.github/actions/paths-filter/filters.yml
@@ -22,3 +22,21 @@ rust-ext:
   - src/**/*.rs
   - Cargo.toml
   - Cargo.lock
+
+client-common: &client-common
+  - oxidation/client/src/**
+  - oxidation/client/resources/**
+  - oxidation/client/public/**
+  - oxidation/client/package-lock.json
+  - oxidation/client/package.json
+  - oxidation/client/capacitor.config.ts
+  - oxidation/client/babel.config.js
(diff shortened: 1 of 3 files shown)


.github/actions/paths-filter/filters.yml
```diff
@@ -1,18 +1,41 @@
...
+
 rust-ext:
+  - *shared
   - .github/**
   - src/**/*.rs
   - Cargo.toml
   - Cargo.lock
+
+ionic-client: &ionic-client
+  - oxidation/client/src/**
```
touilleMan, line 27: Why not doing `oxidation/client/**` instead ?

touilleMan, line 27: And what about  `oxidation/bindings/neon/**` & `oxidation/bindings/common` ?

.github/actions/paths-filter/filters.yml
```diff
@@ -1,18 +1,41 @@
...
+  - oxidation/client/package-lock.json
+  - oxidation/client/package.json
+  - oxidation/client/capacitor.config.ts
+  - oxidation/client/babel.config.js
+  - oxidation/client/tsconfig.json
+
+web:
+  - *shared
+  - *ionic-client
+  - oxidation/bindings/web/**
```
touilleMan, line 39: should also add `oxidation/bindings/common/**`

.github/actions/paths-filter/filters.yml
FirelightFlagboy: This would include every client like android, electron, ios and web

FirelightFlagboy: `neon` is for electon, so it's not used for the web client, but yes I will add `bindings/common`

touilleMan: So `ionic-client` is the common part and `web` the platform specific one.

Why not call it `client-common`, `client-web`, `client-desktop`, `client-android` then ?

touilleMan: Can we use a negative match instead ?
My fear is that this config file is well hidden so there is zero chance it will be modified when new files/folders are added, hence those new files should fall into the "common has been modified" match by default

FirelightFlagboy: yes `ionic-client` is the common part

I will rename them then

FirelightFlagboy: note that the other filter for `electron` & `android` will come later

FirelightFlagboy: Sadly, `paths-filter` don't support negative match

FirelightFlagboy: done

touilleMan: 😭

touilleMan (approved):

PR #2892, opened 2022-08-23
Oxidation handle unknown status

TimeEngineer:


oxidation/libparsec/crates/serialization_format/src/parser/protocol.rs
```diff
@@ -285,6 +287,17 @@ impl Cmds {
                     #[serde(tag = "status")]
                     pub enum Rep {
                         #variants_rep
+                        #[serde(skip)]
+                        UnknownStatus {
+                            _status: String,
+                            reason: Option<String>,
+                        }
```
FirelightFlagboy, line 294:
```suggestion
                        UnknownStatus(UnknownStatus)
```

why not directly use the struct `UnknownStatus` ?

TimeEngineer: Because `status` field is already used in serialization/deserialization by the tag

oxidation/libparsec/crates/protocol/tests/test_common.rs
```diff
@@ -0,0 +1,48 @@
+// Parsec Cloud (https://parsec.cloud) Copyright (c) BUSL-1.1 (eventually AGPL-3.0) 2016-present Scille SAS
```
touilleMan, line 1: good job here 👍

oxidation/libparsec/crates/serialization_format/src/parser/protocol.rs
```diff
@@ -285,6 +287,17 @@ impl Cmds {
                     #[serde(tag = "status")]
                     pub enum Rep {
                         #variants_rep
+                        #[serde(skip)]
```
touilleMan, line 290: can you add here
```rs
// `UnknownStatus` covers the case the server returns a valid message but with
// an unknown status value (given change in error status only cause a minor bump in API version)
// Note it is meaningless to serialize a `UnknownStatus` (you created the object from scratch, you know what it is for baka !)
```

oxidation/libparsec/crates/serialization_format/src/parser/protocol.rs
```diff
@@ -293,7 +306,15 @@ impl Cmds {
                         }
 
                         pub fn load(buf: &[u8]) -> Result<Self, ::rmp_serde::decode::Error> {
-                            ::rmp_serde::from_slice(buf)
+                            Ok(if let Ok(data) = ::rmp_serde::from_slice::<Self>(buf) {
+                                data
+                            } else {
+                                let data = ::rmp_serde::from_slice::<UnknownStatus>(buf)?;
```
touilleMan, line 312: can you add:
```rs
// Due to how Serde handles variant discriminant, we cannot express unknown status as a default case in the main schema
// Instead we have this additional deserialization attempt fallback
```

touilleMan (changes requested):

touilleMan (approved):

oxidation/libparsec/crates/serialization_format/src/parser/protocol.rs
FirelightFlagboy: Well but, serde skip this variant (`#[serde(skip)]`), and we already have the value with `let data = ::rmp_serde::from_slice::<UnknownStatus>(buf)?;`

PR #2905, opened 2022-08-25
Oxidize `file_operations` module python

FirelightFlagboy:
- Replace `prepare_read` by rust impl
- Replace `prepare_write` by rust impl
- Replace `prepare_resize` by rust impl
- Replace `prepare_reshape` by rust impl
- Correct typing issue with functions from rust impl


vxgmichel (approved):
LGTM :+1:

PR #2955, opened 2022-08-26
Upload files modal test

D-Bao:
(only for testing purposes) @NicoTuxx ;)


NicoTuxx:
DO NOT MERGE, but use as inspiration when file upload will be done

PR #2963, opened 2022-08-29
Format rust with `editorconfig`

FirelightFlagboy:


FirelightFlagboy:
These changes only affect indentation inside string, we don't want to change that.
So I'm closing this PR

PR #2964, opened 2022-08-29
Format python files with `editorconfig`

FirelightFlagboy:


parsec/backend/blockstore.py
```diff
@@ -27,17 +27,17 @@ class BaseBlockStoreComponent:
 
     The key takeaways here:
     - `BlockStoreComponent.create` must be implemented in an idempotent way
-      (making the assumption multiple creates with the same orgID/ID couple
-      always comes with the same block data).
+        (making the assumption multiple creates with the same orgID/ID couple
+        always comes with the same block data).
```
vxgmichel, line 31: Why would that be good thing?

parsec/backend/cli/run.py
```diff
@@ -132,8 +132,8 @@ def handle_parse_result(self, ctx, opts, args):
     is_flag=True,
     help="""Allow organization bootstrap without prior creation.
 
-Without this flag, an organization must be created by administration (see
- `parsec core create_organization` command) before bootstrap can occur.
+Without this flag, an organization must be created by administration (see `parsec core create_organization` command)
+before bootstrap can occur.
```
vxgmichel, line 136: Why does editor config have an opinion about that?

vxgmichel (changes requested):
We already have tools for python formatting, so using too many of them might get us into conflicting changes

PR #3008, opened 2022-09-02
Add script to wipe Github Cache entries.

FirelightFlagboy:


.github/scripts/sink_cache.py
```diff
@@ -0,0 +1,235 @@
...
+from logging import basicConfig, getLogger
+
+import sys
+from os.path import dirname, abspath
+
+current_dir = dirname(abspath(__file__))
+parent_dir = dirname(dirname(current_dir))
+sys.path.insert(0, parent_dir)
+
+import misc.bench as bench
```
touilleMan, line 23: better do to `from misc import bench`

.github/scripts/sink_cache.py
```diff
@@ -0,0 +1,235 @@
...
+import json
+import logging
+import subprocess
+from typing import Any, Dict, Optional
+from logging import basicConfig, getLogger
+
+import sys
+from os.path import dirname, abspath
+
+current_dir = dirname(abspath(__file__))
```
touilleMan, line 19: pathlib is a much better module when dealing with path:
```python
from pathlib import Path
current_dir = Path(__file__).resolve().parent
```

.github/scripts/sink_cache.py
```diff
@@ -0,0 +1,235 @@
...
+parent_dir = dirname(dirname(current_dir))
+sys.path.insert(0, parent_dir)
```
touilleMan, line 21: Adding import location this way should be avoided whenever possible (it's basically a hack that lead to unexpected import dependencies)

Here you need it to get `bench.naturalsize` that is only use for a logging.info... so here just stick to KISS:
```python
f"Will remove entry {entry.key} at {ref or '<no_ref>'} (~{entry.size_in_bytes / 2**20:.3f}Mo)"
```

.github/scripts/sink_cache.py
```diff
@@ -0,0 +1,235 @@
...
+        self.version = version
+        self.last_accessed_at = last_accessed_at
+        self.created_at = created_at
+        self.size_in_bytes = size_in_bytes
+
+    def from_dict(dict: Dict[str, Any]) -> "CacheEntries":
+        # Datetime format for `2019-01-24T22:45:36.000Z``
+
+        def get_iso_datetime(raw: str) -> datetime:
+            # `fromisoformat` don't support:
```
touilleMan, line 70: Not sure why you need to handle nanosecond fallback, [according to github documentation](https://docs.github.com/en/rest/actions/cache#list-github-actions-caches-for-a-repository) they only return up to the millisecond (which is the most common for a web service)

So you should only have to do:
```python
dt = datetime.fromisoformat(d.replace("Z", "+00:00"))  # `fromisoformat` doesn't support Z suffix
```

.github/scripts/sink_cache.py
```diff
@@ -0,0 +1,235 @@
...
+class CacheEntries:
+    total_count: int
+    actions_caches: Dict[int, CacheEntry]
+
+    def __init__(self, total_count: int, actions_caches: Dict[int, CacheEntry]) -> None:
+        self.total_count = total_count
+        self.actions_caches = actions_caches
+
+    def __repr__(self) -> str:
+        return f"{CacheEntries.__name__}({self.total_count}, {repr(self.actions_caches)})"
```
touilleMan, line 112: you can do `f"{x:r}` instead of `f"{repr(x)}`

instead of using `CacheEntries.__name__`, we do `type(self).__name__` in repr. This is because otherwise a child class would have the wrong name here.

and finally you would benefit from using `@dataclasses.dataclass` given it would provide you with the repr for free ;-)

touilleMan (changes requested):

.github/scripts/sink_cache.py
```diff
@@ -0,0 +1,235 @@
...
+        "Total cache size {} ({} bytes)".format(
+            bench.naturalsize(total_removed_size), total_removed_size
+        )
+    )
+
+
+if __name__ == "__main__":
+    parser = argparse.ArgumentParser(
+        prog="sink-cache",
+        description="\n".join(
```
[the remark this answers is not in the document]
FirelightFlagboy: I know but i break the identation :(

.github/scripts/sink_cache.py
FirelightFlagboy: The actual data return a datetime with the nanosecond 🤷

.github/scripts/sink_cache.py
```diff
@@ -0,0 +1,235 @@
...
+            except subprocess.CalledProcessError as cpe:
+                log.error(f"Error while removing cache {entry.key}: {cpe}")
+            else:
+                log.debug(f"res={return_val}")
+
+
+def remove_caches_from_repository(owner: str, repo: str, dry: bool, ref: Optional[str]):
+    logger.info(f"Will remove caches from {owner}/{repo}" + (f"/{ref}" if ref is not None else ""))
+    entries = get_cache_entries(owner, repo, ref)
+    if len(entries.actions_caches) > 0:
```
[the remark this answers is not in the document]
FirelightFlagboy: done

.github/scripts/sink_cache.py
```diff
@@ -0,0 +1,235 @@
...
+        remove_cache_entries(entries, owner, repo, dry, ref)
+    else:
+        logger.info("No cache to delete !")
+    total_removed_size = reduce(
```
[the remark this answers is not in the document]
FirelightFlagboy: done

.github/scripts/sink_cache.py
```diff
@@ -0,0 +1,235 @@
...
+            ref=dict["ref"],
+            version=dict["version"],
+            last_accessed_at=last_accessed_at,
+            created_at=create_at,
+            size_in_bytes=int(dict["size_in_bytes"]),
+        )
+
+
+class CacheEntries:
+    total_count: int
```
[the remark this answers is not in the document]
FirelightFlagboy: I'll use `@dataclass` now

.github/scripts/sink_cache.py
FirelightFlagboy: the classes now use `@dataclass`

.github/scripts/sink_cache.py
FirelightFlagboy: I've rework `get_iso_datetime`

.github/scripts/sink_cache.py
FirelightFlagboy: I've removed the dependency on that

.github/scripts/sink_cache.py
FirelightFlagboy: I've remove the dependency, so I also removed this code block

.github/scripts/sink_cache.py
FirelightFlagboy: I've refactor the code to not use `naturalize`

.github/scripts/sink_cache.py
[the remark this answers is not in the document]
touilleMan: Syntax highlighting makes this ok (at least here where indentation is simple and the code is not cluttered by multiple multilines strings)

touilleMan (approved):

PR #3014, opened 2022-09-02
Introduce `TimeProvider.sleep`, your new best testing friend 🤞

touilleMan:
Currently not working when scheduling the sleep in a trio nursery:

<details>
<summary>
test_sleep_in_nursery failure
</summary>

```
________________________________________________________________________________________ test_sleep_in_nursery _________________________________________________________________________________________

runner = Runner(clock=SystemClock(offset=195593.49378819152), instruments={'_all': {}}, io_manager=<trio._core._io_windows.Wind..._autojump_threshold=inf, is_guest=False, guest_tick_scheduled=False, ki_pending=False, waiting_for_idle=SortedDict({}))
async_fn = functools.partial(<function _trio_test_runner_factory.<locals>._bootstrap_fixtures_and_run_test at 0x000001AD8245AF70>), args = (), host_uses_signal_set_wakeup_fd = False

    def unrolled_run(runner, async_fn, args, host_uses_signal_set_wakeup_fd=False):
        locals()[LOCALS_KEY_KI_PROTECTION_ENABLED] = True
        __tracebackhide__ = True

        try:
            if not host_uses_signal_set_wakeup_fd:
                runner.entry_queue.wakeup.wakeup_on_signals()

            if "before_run" in runner.instruments:
                runner.instruments.call("before_run")
            runner.clock.start_clock()
            runner.init_task = runner.spawn_impl(
                runner.init, (async_fn, args), None, "<init>", system_task=True
            )

            # You know how people talk about "event loops"? This 'while' loop right
            # here is our event loop:
            while runner.tasks:
                if runner.runq:
                    timeout = 0
                else:
                    deadline = runner.deadlines.next_deadline()
                    timeout = runner.clock.deadline_to_sleep_time(deadline)
                timeout = min(max(0, timeout), _MAX_TIMEOUT)

                idle_primed = None
                if runner.waiting_for_idle:
                    cushion, _ = runner.waiting_for_idle.keys()[0]
                    if cushion < timeout:
                        timeout = cushion
                        idle_primed = IdlePrimedTypes.WAITING_FOR_IDLE
                # We use 'elif' here because if there are tasks in
                # wait_all_tasks_blocked, then those tasks will wake up without
                # jumping the clock, so we don't need to autojump.
                elif runner.clock_autojump_threshold < timeout:
                    timeout = runner.clock_autojump_threshold
                    idle_primed = IdlePrimedTypes.AUTOJUMP_CLOCK

                if "before_io_wait" in runner.instruments:
                    runner.instruments.call("before_io_wait", timeout)

                # Driver will call io_manager.get_events(timeout) and pass it back
                # in through the yield
                events = yield timeout
                runner.io_manager.process_events(events)

                if "after_io_wait" in runner.instruments:
                    runner.instruments.call("after_io_wait", timeout)

                # Process cancellations due to deadline expiry
                now = runner.clock.current_time()
                if runner.deadlines.expire(now):
                    idle_primed = None

                # idle_primed != None means: if the IO wait hit the timeout, and
                # still nothing is happening, then we should start waking up
                # wait_all_tasks_blocked tasks or autojump the clock. But there
                # are some subtleties in defining "nothing is happening".
                #
                # 'not runner.runq' means that no tasks are currently runnable.
                # 'not events' means that the last IO wait call hit its full
                # timeout. These are very similar, and if idle_primed != None and
                # we're running in regular mode then they always go together. But,
                # in *guest* mode, they can happen independently, even when
                # idle_primed=True:
                #
                # - runner.runq=empty and events=True: the host loop adjusted a
                #   deadline and that forced an IO wakeup before the timeout expired,
                #   even though no actual tasks were scheduled.
                #
                # - runner.runq=nonempty and events=False: the IO wait hit its
                #   timeout, but then some code in the host thread rescheduled a task
                #   before we got here.
                #
                # So we need to check both.
                if idle_primed is not None and not runner.runq and not events:
                    if idle_primed is IdlePrimedTypes.WAITING_FOR_IDLE:
                        while runner.waiting_for_idle:
                            key, task = runner.waiting_for_idle.peekitem(0)
                            if key[0] == cushion:
                                del runner.waiting_for_idle[key]
                                runner.reschedule(task)
                            else:
                                break
                    else:
                        assert idle_primed is IdlePrimedTypes.AUTOJUMP_CLOCK
                        runner.clock._autojump()

                # Process all runnable tasks, but only the ones that are already
                # runnable now. Anything that becomes runnable during this cycle
                # needs to wait until the next pass. This avoids various
                # starvation issues by ensuring that there's never an unbounded
                # delay between successive checks for I/O.
                #
                # Also, we randomize the order of each batch to avoid assumptions
                # about scheduling order sneaking in. In the long run, I suspect
                # we'll either (a) use strict FIFO ordering and document that for
                # predictability/determinism, or (b) implement a more
                # sophisticated scheduler (e.g. some variant of fair queueing),
                # for better behavior under load. For now, this is the worst of
                # both worlds - but it keeps our options open. (If we do decide to
                # go all in on deterministic scheduling, then there are other
                # things that will probably need to change too, like the deadlines
                # tie-breaker and the non-deterministic ordering of
                # task._notify_queues.)
                batch = list(runner.runq)
                runner.runq.clear()
                if _ALLOW_DETERMINISTIC_SCHEDULING:
                    # We're running under Hypothesis, and pytest-trio has patched
                    # this in to make the scheduler deterministic and avoid flaky
                    # tests. It's not worth the (small) performance cost in normal
                    # operation, since we'll shuffle the list and _r is only
                    # seeded for tests.
                    batch.sort(key=lambda t: t._counter)
                    _r.shuffle(batch)
                else:
                    # 50% chance of reversing the batch, this way each task
                    # can appear before/after any other task.
                    if _r.random() < 0.5:
                        batch.reverse()
                while batch:
                    task = batch.pop()
                    GLOBAL_RUN_CONTEXT.task = task

                    if "before_task_step" in runner.instruments:
                        runner.instruments.call("before_task_step", task)

                    next_send_fn = task._next_send_fn
                    next_send = task._next_send
                    task._next_send_fn = task._next_send = None
                    final_outcome = None
                    try:
                        # We used to unwrap the Outcome object here and send/throw
                        # its contents in directly, but it turns out that .throw()
                        # is buggy, at least before CPython 3.9:
                        #   https://bugs.python.org/issue29587
                        #   https://bugs.python.org/issue29590
                        # So now we send in the Outcome object and unwrap it on the
                        # other side.
                        msg = task.context.run(next_send_fn, next_send)
                    except StopIteration as stop_iteration:
                        final_outcome = Value(stop_iteration.value)
                    except BaseException as task_exc:
                        # Store for later, removing uninteresting top frames: 1
                        # frame we always remove, because it's this function
                        # catching it, and then in addition we remove however many
                        # more Context.run adds.
                        tb = task_exc.__traceback__.tb_next
                        for _ in range(CONTEXT_RUN_TB_FRAMES):
                            tb = tb.tb_next
                        final_outcome = Error(task_exc.with_traceback(tb))
                        # Remove local refs so that e.g. cancelled coroutine locals
                        # are not kept alive by this frame until another exception
                        # comes along.
                        del tb

                    if final_outcome is not None:
                        # We can't call this directly inside the except: blocks
                        # above, because then the exceptions end up attaching
                        # themselves to other exceptions as __context__ in
                        # unwanted ways.
>                       runner.task_exited(task, final_outcome)

..\..\..\..\AppData\Local\pypoetry\Cache\virtualenvs\parsec-cloud--xPPmQVg-py3.9\lib\site-packages\trio\_core\_run.py:2205:
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

self = Runner(clock=SystemClock(offset=195593.49378819152), instruments={'_all': {}}, io_manager=<trio._core._io_windows.Wind..._autojump_threshold=inf, is_guest=False, guest_tick_scheduled=False, ki_pending=False, waiting_for_idle=SortedDict({}))
task = <Task 'None.TimeProvider.sleep' at 0x1ad8247e360>
outcome = Error(RuntimeError("Cancel scope stack corrupted: cancel scope surrounding <Task 'None.TimeProvider.sleep' at 0x1ad824...don't believe you're doing any of these things, please file a bug:\nhttps://github.com/python-trio/trio/issues/new\n"))

    def task_exited(self, task, outcome):
        if (
            task._cancel_status is not None
            and task._cancel_status.abandoned_by_misnesting
            and task._cancel_status.parent is None
        ):
            # The cancel scope surrounding this task's nursery was closed
            # before the task exited. Force the task to exit with an error,
            # since the error might not have been caught elsewhere. See the
            # comments in CancelStatus.close().
            try:
                # Raise this, rather than just constructing it, to get a
                # traceback frame included
                raise RuntimeError(
                    "Cancel scope stack corrupted: cancel scope surrounding "
                    "{!r} was closed before the task exited\n{}".format(
                        task, MISNESTING_ADVICE
                    )
                )
            except RuntimeError as new_exc:
                if isinstance(outcome, Error):
                    new_exc.__context__ = outcome.error
                outcome = Error(new_exc)

        task._activate_cancel_status(None)
        self.tasks.remove(task)
        if task is self.init_task:
            # If the init task crashed, then something is very wrong and we
            # let the error propagate. (It'll eventually be wrapped in a
            # TrioInternalError.)
            outcome.unwrap()
            # the init task should be the last task to exit. If not, then
            # something is very wrong.
            if self.tasks:  # pragma: no cover
                raise TrioInternalError
        else:
            if task is self.main_task:
                self.main_task_outcome = outcome
                outcome = Value(None)
>           task._parent_nursery._child_finished(task, outcome)

..\..\..\..\AppData\Local\pypoetry\Cache\virtualenvs\parsec-cloud--xPPmQVg-py3.9\lib\site-packages\trio\_core\_run.py:1526:
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

self = <trio.Nursery object at 0x000001AD8247D820>, task = <Task 'None.TimeProvider.sleep' at 0x1ad8247e360>
outcome = Error(RuntimeError("Cancel scope stack corrupted: cancel scope surrounding <Task 'None.TimeProvider.sleep' at 0x1ad824...don't believe you're doing any of these things, please file a bug:\nhttps://github.com/python-trio/trio/issues/new\n"))

    def _child_finished(self, task, outcome):
        self._children.remove(task)
        if isinstance(outcome, Error):
            self._add_exc(outcome.error)
>       self._check_nursery_closed()

..\..\..\..\AppData\Local\pypoetry\Cache\virtualenvs\parsec-cloud--xPPmQVg-py3.9\lib\site-packages\trio\_core\_run.py:910:
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

self = <trio.Nursery object at 0x000001AD8247D820>

    def _check_nursery_closed(self):
        if not any([self._nested_child_running, self._children, self._pending_starts]):
            self._closed = True
            if self._parent_waiting_in_aexit:
                self._parent_waiting_in_aexit = False
>               GLOBAL_RUN_CONTEXT.runner.reschedule(self._parent_task)

..\..\..\..\AppData\Local\pypoetry\Cache\virtualenvs\parsec-cloud--xPPmQVg-py3.9\lib\site-packages\trio\_core\_run.py:904:
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

self = Runner(clock=SystemClock(offset=195593.49378819152), instruments={'_all': {}}, io_manager=<trio._core._io_windows.Wind..._autojump_threshold=inf, is_guest=False, guest_tick_scheduled=False, ki_pending=False, waiting_for_idle=SortedDict({}))
task = <Task "<test 'test_sleep_in_nursery'>" at 0x1ad8247e2c0>, next_send = Value(None)

    @_public
    def reschedule(self, task, next_send=_NO_SEND):
        """Reschedule the given task with the given
        :class:`outcome.Outcome`.

        See :func:`wait_task_rescheduled` for the gory details.

        There must be exactly one call to :func:`reschedule` for every call to
        :func:`wait_task_rescheduled`. (And when counting, keep in mind that
        returning :data:`Abort.SUCCEEDED` from an abort callback is equivalent
        to calling :func:`reschedule` once.)

        Args:
          task (trio.lowlevel.Task): the task to be rescheduled. Must be blocked
              in a call to :func:`wait_task_rescheduled`.
          next_send (outcome.Outcome): the value (or error) to return (or
              raise) from :func:`wait_task_rescheduled`.

        """
        if next_send is _NO_SEND:
            next_send = Value(None)

        assert task._runner is self
        if task._next_send_fn is not None:
            breakpoint()
>       assert task._next_send_fn is None
E       AssertionError

..\..\..\..\AppData\Local\pypoetry\Cache\virtualenvs\parsec-cloud--xPPmQVg-py3.9\lib\site-packages\trio\_core\_run.py:1411: AssertionError

The above exception was the direct cause of the following exception:

runner = Runner(clock=SystemClock(offset=195593.49378819152), instruments={'_all': {}}, io_manager=<trio._core._io_windows.Wind..._autojump_threshold=inf, is_guest=False, guest_tick_scheduled=False, ki_pending=False, waiting_for_idle=SortedDict({}))
async_fn = functools.partial(<function _trio_test_runner_factory.<locals>._bootstrap_fixtures_and_run_test at 0x000001AD8245AF70>), args = (), host_uses_signal_set_wakeup_fd = False

    def unrolled_run(runner, async_fn, args, host_uses_signal_set_wakeup_fd=False):
        locals()[LOCALS_KEY_KI_PROTECTION_ENABLED] = True
        __tracebackhide__ = True

        try:
            if not host_uses_signal_set_wakeup_fd:
                runner.entry_queue.wakeup.wakeup_on_signals()

            if "before_run" in runner.instruments:
                runner.instruments.call("before_run")
            runner.clock.start_clock()
            runner.init_task = runner.spawn_impl(
                runner.init, (async_fn, args), None, "<init>", system_task=True
            )

            # You know how people talk about "event loops"? This 'while' loop right
            # here is our event loop:
            while runner.tasks:
                if runner.runq:
                    timeout = 0
                else:
                    deadline = runner.deadlines.next_deadline()
                    timeout = runner.clock.deadline_to_sleep_time(deadline)
                timeout = min(max(0, timeout), _MAX_TIMEOUT)

                idle_primed = None
                if runner.waiting_for_idle:
                    cushion, _ = runner.waiting_for_idle.keys()[0]
                    if cushion < timeout:
                        timeout = cushion
                        idle_primed = IdlePrimedTypes.WAITING_FOR_IDLE
                # We use 'elif' here because if there are tasks in
                # wait_all_tasks_blocked, then those tasks will wake up without
                # jumping the clock, so we don't need to autojump.
                elif runner.clock_autojump_threshold < timeout:
                    timeout = runner.clock_autojump_threshold
                    idle_primed = IdlePrimedTypes.AUTOJUMP_CLOCK

                if "before_io_wait" in runner.instruments:
                    runner.instruments.call("before_io_wait", timeout)

                # Driver will call io_manager.get_events(timeout) and pass it back
                # in through the yield
                events = yield timeout
                runner.io_manager.process_events(events)

                if "after_io_wait" in runner.instruments:
                    runner.instruments.call("after_io_wait", timeout)

                # Process cancellations due to deadline expiry
                now = runner.clock.current_time()
                if runner.deadlines.expire(now):
                    idle_primed = None

                # idle_primed != None means: if the IO wait hit the timeout, and
                # still nothing is happening, then we should start waking up
                # wait_all_tasks_blocked tasks or autojump the clock. But there
                # are some subtleties in defining "nothing is happening".
                #
                # 'not runner.runq' means that no tasks are currently runnable.
                # 'not events' means that the last IO wait call hit its full
                # timeout. These are very similar, and if idle_primed != None and
                # we're running in regular mode then they always go together. But,
                # in *guest* mode, they can happen independently, even when
                # idle_primed=True:
                #
                # - runner.runq=empty and events=True: the host loop adjusted a
                #   deadline and that forced an IO wakeup before the timeout expired,
                #   even though no actual tasks were scheduled.
                #
                # - runner.runq=nonempty and events=False: the IO wait hit its
                #   timeout, but then some code in the host thread rescheduled a task
                #   before we got here.
                #
                # So we need to check both.
                if idle_primed is not None and not runner.runq and not events:
                    if idle_primed is IdlePrimedTypes.WAITING_FOR_IDLE:
                        while runner.waiting_for_idle:
                            key, task = runner.waiting_for_idle.peekitem(0)
                            if key[0] == cushion:
                                del runner.waiting_for_idle[key]
                                runner.reschedule(task)
                            else:
                                break
                    else:
                        assert idle_primed is IdlePrimedTypes.AUTOJUMP_CLOCK
                        runner.clock._autojump()

                # Process all runnable tasks, but only the ones that are already
                # runnable now. Anything that becomes runnable during this cycle
                # needs to wait until the next pass. This avoids various
                # starvation issues by ensuring that there's never an unbounded
                # delay between successive checks for I/O.
                #
                # Also, we randomize the order of each batch to avoid assumptions
                # about scheduling order sneaking in. In the long run, I suspect
                # we'll either (a) use strict FIFO ordering and document that for
                # predictability/determinism, or (b) implement a more
                # sophisticated scheduler (e.g. some variant of fair queueing),
                # for better behavior under load. For now, this is the worst of
                # both worlds - but it keeps our options open. (If we do decide to
                # go all in on deterministic scheduling, then there are other
                # things that will probably need to change too, like the deadlines
                # tie-breaker and the non-deterministic ordering of
                # task._notify_queues.)
                batch = list(runner.runq)
                runner.runq.clear()
                if _ALLOW_DETERMINISTIC_SCHEDULING:
                    # We're running under Hypothesis, and pytest-trio has patched
                    # this in to make the scheduler deterministic and avoid flaky
                    # tests. It's not worth the (small) performance cost in normal
                    # operation, since we'll shuffle the list and _r is only
                    # seeded for tests.
                    batch.sort(key=lambda t: t._counter)
                    _r.shuffle(batch)
                else:
                    # 50% chance of reversing the batch, this way each task
                    # can appear before/after any other task.
                    if _r.random() < 0.5:
                        batch.reverse()
                while batch:
                    task = batch.pop()
                    GLOBAL_RUN_CONTEXT.task = task

                    if "before_task_step" in runner.instruments:
                        runner.instruments.call("before_task_step", task)

                    next_send_fn = task._next_send_fn
                    next_send = task._next_send
                    task._next_send_fn = task._next_send = None
                    final_outcome = None
                    try:
                        # We used to unwrap the Outcome object here and send/throw
                        # its contents in directly, but it turns out that .throw()
                        # is buggy, at least before CPython 3.9:
                        #   https://bugs.python.org/issue29587
                        #   https://bugs.python.org/issue29590
                        # So now we send in the Outcome object and unwrap it on the
                        # other side.
                        msg = task.context.run(next_send_fn, next_send)
                    except StopIteration as stop_iteration:
                        final_outcome = Value(stop_iteration.value)
                    except BaseException as task_exc:
                        # Store for later, removing uninteresting top frames: 1
                        # frame we always remove, because it's this function
                        # catching it, and then in addition we remove however many
                        # more Context.run adds.
                        tb = task_exc.__traceback__.tb_next
                        for _ in range(CONTEXT_RUN_TB_FRAMES):
                            tb = tb.tb_next
                        final_outcome = Error(task_exc.with_traceback(tb))
                        # Remove local refs so that e.g. cancelled coroutine locals
                        # are not kept alive by this frame until another exception
                        # comes along.
                        del tb

                    if final_outcome is not None:
                        # We can't call this directly inside the except: blocks
                        # above, because then the exceptions end up attaching
                        # themselves to other exceptions as __context__ in
                        # unwanted ways.
                        runner.task_exited(task, final_outcome)
                        # final_outcome may contain a traceback ref. It's not as
                        # crucial compared to the above, but this will allow more
                        # prompt release of resources in coroutine locals.
                        final_outcome = None
                    else:
                        task._schedule_points += 1
                        if msg is CancelShieldedCheckpoint:
                            runner.reschedule(task)
                        elif type(msg) is WaitTaskRescheduled:
                            task._cancel_points += 1
                            task._abort_func = msg.abort_func
                            # KI is "outside" all cancel scopes, so check for it
                            # before checking for regular cancellation:
                            if runner.ki_pending and task is runner.main_task:
                                task._attempt_delivery_of_pending_ki()
                            task._attempt_delivery_of_any_pending_cancel()
                        elif type(msg) is PermanentlyDetachCoroutineObject:
                            # Pretend the task just exited with the given outcome
                            runner.task_exited(task, msg.final_outcome)
                        else:
                            exc = TypeError(
                                "trio.run received unrecognized yield message {!r}. "
                                "Are you trying to use a library written for some "
                                "other framework like asyncio? That won't work "
                                "without some kind of compatibility shim.".format(msg)
                            )
                            # The foreign library probably doesn't adhere to our
                            # protocol of unwrapping whatever outcome gets sent in.
                            # Instead, we'll arrange to throw `exc` in directly,
                            # which works for at least asyncio and curio.
                            runner.reschedule(task, exc)
                            task._next_send_fn = task.coro.throw
                        # prevent long-lived reference
                        # TODO: develop test for this deletion
                        del msg

                    if "after_task_step" in runner.instruments:
                        runner.instruments.call("after_task_step", task)
                    del GLOBAL_RUN_CONTEXT.task
                    # prevent long-lived references
                    # TODO: develop test for these deletions
                    del task, next_send, next_send_fn

        except GeneratorExit:
            # The run-loop generator has been garbage collected without finishing
            warnings.warn(
                RuntimeWarning(
                    "Trio guest run got abandoned without properly finishing... "
                    "weird stuff might happen"
                )
            )
        except TrioInternalError:
            raise
        except BaseException as exc:
>           raise TrioInternalError("internal error in Trio - please file a bug!") from exc
E           trio.TrioInternalError: internal error in Trio - please file a bug!

..\..\..\..\AppData\Local\pypoetry\Cache\virtualenvs\parsec-cloud--xPPmQVg-py3.9\lib\site-packages\trio\_core\_run.py:2260: TrioInternalError
```

</details>


oxidation/libparsec/crates/platform_async/Cargo.toml
```diff
@@ -10,6 +10,11 @@ futures = "0.3.21"
 
 [dev-dependencies]
 log = "0.4.17"
+# `std::time::Instant` is not available for wasm32, crate `instant` provides a
+# fallback based on `window.performance.now()`
+instant = "0.1.12"
```
FirelightFlagboy, line 15: Shouldn't we re-export this `crate` to expose it for crates using `platform-async` ?

oxidation/libparsec/crates/platform_async/tests/eyeballs_wasm32.rs
```diff
@@ -1,144 +0,0 @@
-#![cfg(target_arch = "wasm32")]
```
FirelightFlagboy, line 1: Keep this file

oxidation/libparsec/crates/types/src/time.rs
```diff
@@ -168,8 +183,69 @@ mod time_provider {
...
+                MockedTime::RealTime => self.parent_now_or_realtime(),
+                MockedTime::FrozenTime(dt) => dt,
+                MockedTime::ShiftedTime { microseconds: us } => {
+                    self.parent_now_or_realtime().add_us(us)
+                }
+            }
+        }
+    }
+
+    /// Wrap the TimeProviderAgent into TimeProvider to hide the Arc stuff
```
FirelightFlagboy, line 245:
```suggestion
    /// Wrap the `TimeProviderAgent` into `TimeProvider` to hide the sync stuff
```

oxidation/libparsec/crates/types/src/time.rs
```diff
@@ -168,8 +183,69 @@ mod time_provider {
...
+    /// and provide the same simple API as the non-test TimeProvider
```
FirelightFlagboy, line 246:
```suggestion
    /// and provide the same simple API as the non-test `TimeProvider`
```

oxidation/libparsec/crates/types/src/time.rs
```diff
@@ -158,6 +169,10 @@ mod time_provider {
             let now = chrono::Utc::now();
             now.into()
         }
+
+        pub async fn sleep(&self, time: std::time::Duration) {
+            tokio::time::sleep(time).await
+        }
     }
 }
```
FirelightFlagboy, line 178:
```suggestion

#[cfg(not(target_arch = "wasm32"))]
```
This code will not compile on `wasm32`

FirelightFlagboy (changes requested):

oxidation/libparsec/crates/platform_async/tests/eyeballs_wasm32.rs
FirelightFlagboy: NVM

FirelightFlagboy (changes requested):
I don't see the `sleep` implementation for `wasm32`

oxidation/libparsec/crates/types/src/time.rs
touilleMan: fixed

oxidation/libparsec/crates/platform_async/Cargo.toml
touilleMan: Yes, I think that's something we want to do in the long run, but I'd rather wait a bit for that until we have an actual use (or at least until we have a better view on what features we need to export, this is to avoid a "export everything every-time you need something" reflex that may produce too much coupling between crates)

touilleMan:
> I don't see the sleep implementation for wasm32

I knew you wouldn't get fooled that easily :smile: !

You're totally right on this, yet as you can see the CI is passing.
This is because currently only a subset of the crates are compiled in wasm32. Of course this is wrong and should be corrected, but it's outside of the scope of this PR.

This PR is basically the very first actual use of async in oxidation, so we'd better go by small steps.
But don't worry, I have short-term plans to make the web platform a first-class citizen ;-)

oxidation/libparsec/crates/platform_async/Cargo.toml
FirelightFlagboy: Ok

oxidation/libparsec/crates/platform_async/src/wasm32/timer.rs
```diff
@@ -39,3 +39,9 @@ impl Future for Timer {
         self.0.as_mut().poll(cx).map(|_| ())
     }
 }
+
+pub type Sleep = Timer;
```
TimeEngineer, line 43: Why ?

oxidation/libparsec/crates/types/src/time.rs
```diff
@@ -164,18 +175,86 @@ mod time_provider {
...
+            self.time = time;
+            // Broadcast the config change given it impact everybody waiting
+            MOCK_CONFIG_HAS_CHANGED.notify_waiters();
+        }
+
+        fn parent_now_or_realtime(&self) -> DateTime {
+            match &self.parent {
+                // TODO: remove me and only rely on `chrono::Utc::now().into()` instead !
+                None => DateTime::now_legacy(),
+                Some(parent) => parent.lock().expect("Muex is poisoned").now(),
```
TimeEngineer, line 239: Mutex

oxidation/libparsec/crates/types/src/time.rs
```diff
@@ -164,18 +175,86 @@ mod time_provider {
...
+    /// Time provider system consist of a hierarchy of agents, each one able to mock it
+    /// time while taking into account it parent's mock time:
+    /// - RealTime agent will use it parent's time if it has one or the actual real time otherwise
+    /// - FrozenTime agent will always use the same time configured for it
+    /// - ShiftedTime agent work as the RealTime agent but add it configured shifted value on top
+    #[derive(Debug)]
+    struct TimeProviderAgent {
+        parent: Option<Arc<Mutex<TimeProviderAgent>>>,
+        time: MockedTime,
+        sleeping_stats: u32,
```
TimeEngineer, line 217: u64 ?

oxidation/libparsec/crates/types/src/time.rs
```diff
@@ -188,35 +267,81 @@ mod time_provider {
...
+            self.0.lock().expect("Mutex is poisoned").now()
+        }
+
+        /// When we call sleep, there is no guarantee when the starting time will be taken.
+        /// Hence it is possible this time is taken after we have call the mock_time (this
+        /// is typically the case when calling sleep from Python where the actual sleep is
+        /// scheduled on a tokio thread while directly returning a fake coroutine).
+        /// So the solution is to use a busyloop in the testing code that watch over the
+        /// number of tasks currently sleeping on our time provider.
+        pub fn sleeping_stats(&self) -> u32 {
```
TimeEngineer, line 302: u64 ?

src/runtime.rs
```diff
@@ -0,0 +1,149 @@
...
+            // the trio token to schedule a synchronous function that will do the actual schedule call
+            match trio_token.call_method1(
+                py, "run_sync_soon", (trio_reschedule_fn.as_ref(py), trio_current_task, outcome)
+            ) {
+                Ok(_) => (),
+                // We can ignore the error if the trio loop has been closed...
+                Err(err) if err.is_instance_of::<RunFinishedError>(py)  => (),
+                // ...but for any other exception there is not much we can do :'(
+                Err(err) => {
+                        panic!("Cannot call `TrioToken.run_sync_soon(trio.lowlevel.reschedule, <outcome>)`: {:?}", err);
```
TimeEngineer, line 129: indent

src/runtime.rs
```diff
@@ -0,0 +1,149 @@
...
+                }
+            }
+
```
TimeEngineer, line 132: newline

src/time.rs
```diff
@@ -39,18 +40,48 @@ impl TimeProvider {
         Ok(DateTime(self.0.now()))
     }
 
-    pub fn mock_time(&mut self, time: &PyAny) -> PyResult<()> {
+    pub fn sleeping_stats(&self) -> PyResult<u32> {
```
TimeEngineer, line 43: u64 ?

src/time.rs
```diff
@@ -39,18 +40,48 @@ impl TimeProvider {
...
-            });
+        let mock_config = match (freeze, shift, realtime) {
+            (None, None, true) => MockedTime::RealTime,
+            (Some(dt), None, false) => MockedTime::FrozenTime(dt.0),
+            (None, Some(shift_in_seconds), false) => MockedTime::ShiftedTime {
+                microseconds: (shift_in_seconds * 1e6) as i64,
+            },
+            _ => {
+                return Err(PyValueError::new_err(
+                    "Must only provide on of `freeze`, `shift` and `realtime`",
```
TimeEngineer, line 80: one

TimeEngineer (changes requested):

TimeEngineer (approved):

FirelightFlagboy (approved):

PR #3031, opened 2022-09-06
Support pickle dump for DateTime (required by PyQt signals)

touilleMan:


TimeEngineer (changes requested):
We said we use u64/i64

touilleMan:
> We said we use u64/i64

Yup, we should use u64/i64 when defining our structure... unless we have a good reason not to ;-)

Here we wrap `chrono::Datetime` that uses i32/u32, hence forcing to use u64/i64 here is useless and cumbersome (typically adding those conversion lead to adding a bug in the year type that was supposed to be signed)

src/time.rs
```diff
@@ -24,7 +24,7 @@ pub(crate) fn mock_time(time: &PyAny) -> PyResult<()> {
     Ok(())
 }
 
-#[pyclass]
+#[pyclass(module = "parsec._parsec")]
```
FirelightFlagboy, line 27: Why adding `module = "parsec._parsec"` ?

from the documentation:

| | |
| --- | --- |
| `module = "module_name"` | Python code will see the class as being defined in this module. Defaults to builtins. |

TimeEngineer:
> > We said we use u64/i64
> 
> Yup, we should use u64/i64 when defining our structure... unless we have a good reason not to ;-)
> 
> Here we wrap `chrono::Datetime` that uses i32/u32, hence forcing to use u64/i64 here is useless and cumbersome (typically adding those conversion lead to adding a bug in the year type that was supposed to be signed)

Same shit can happen for usize/isize, but we prefered to use u64/i64 in Vincent's PR :thinking:

src/time.rs
touilleMan: This is needed by pickle to know where to look when deserializing this class

FirelightFlagboy: Could you add a comment for that ?

touilleMan:
> Same shit can happen for usize/isize, but we prefered to use u64/i64 in Vincent's PR thinking

Just follow KISS then ;-)

This u64/i64 thing is really important for serialized structures (given otherwise we will fail to deserialize a msgpack document containing a big number).
Otherwise it is useful as a rule of thumb for not having to wonder between u32/u64/usize everytime we need a number type. But when the number type has already been decided for us (like what is done with `chrono::DateTime`) or if there is a good enough reason (if we where to implement DateTime from scratch I guess we could argue month/day/hour/minute/second cannot be > 60 so we may want to use `u8` here... who knows ^^)

TimeEngineer (approved):

vxgmichel:
> (required by PyQt signals)

I'm not sure pickle support is required for PyQt signals. I ran a quick test on master and it seems there's not problem with providing a `parsec._parsec.Datetime` to a `pyqtSignal(object)`. Maybe I'm missing something?

FirelightFlagboy:
Can't reproduce

Issue #2953, opened 2022-08-26
Server access log should include DeviceID in the HTTP authenticated API


touilleMan:
see https://github.com/Scille/parsec-cloud/pull/2544#discussion_r955140579

so according to [hypercorn documentation](https://pgjones.gitlab.io/hypercorn/how_to_guides/logging.html#how-to-log), it should be enough to use the {Header}i format in the access log format:

[parsec-cloud/parsec/backend/asgi/__init__.py](https://github.com/Scille/parsec-cloud/blob/890490c21d801fdc2090ec2a4ce4d3ecd159992e/parsec/backend/asgi/__init__.py#L126)

Line 126 in [890490c](https://github.com/Scille/parsec-cloud/commit/890490c21d801fdc2090ec2a4ce4d3ecd159992e)
 "access_log_format": "%(h)s %(r)s %(s)s %(b)so %(D)sus", 

Given we already have the organization id in the endpoint, we only need to provide the device id by doing something like:
"access_log_format": "%(h)s %(r)s %(s)s %(b)so %(D)sus (device_id:%({Author}i)s)",

However this is not great given we don't return the actual DeviceID but a base64 of the DeviceID :(
On top of that all request unrelated to the authenticated route will have an ugly (device_id:) in the log...

The thing is we really want to have the device id in the log given it's really useful we looking at the log to see the different queries a given client has done...

PR #3043, opened 2022-09-08
Access log enhancement

Volham22:
As mentioned in [issue 2953](https://github.com/Scille/parsec-cloud/issues/2953) it's better to have device id in logs when users are logged in.
This PR adds a custom implementation over the default Hypercorn `access logger`. A custom logger is required as the device id is base64 encoded in the HTTP author header.

### DIFF
--- a/newsfragments/2953.feature.rst
+++ b/newsfragments/2953.feature.rst
@@ -1,0 +1,1 @@
+Extended Logger to show device id when a client is authenticated
--- a/newsfragments/2953.feature.rst
+++ b/newsfragments/2953.feature.rst
@@ -1,1 +1,0 @@
-Extended Logger to show device id when a client is authenticated
--- a/parsec/backend/asgi/__init__.py
+++ b/parsec/backend/asgi/__init__.py
@@ -17,6 +17,7 @@ from hypercorn.trio import serve
 
 from parsec import __version__ as parsec_version
 from parsec.backend.app import BackendApp
+from parsec.backend.asgi.logger import ParsecLogger
 from parsec.backend.config import BackendConfig
 from parsec.backend.asgi.administration import administration_bp
 from parsec.backend.asgi.redirect import redirect_bp
@@ -45,7 +46,10 @@ def app_factory(
     backend: BackendApp, app_cls: Type[BackendQuartTrio] = BackendQuartTrio
 ) -> BackendQuartTrio:
     app = app_cls(
-        __name__, static_folder="../static", static_url_path="/static", template_folder="templates"
+        __name__,
+        static_folder="../static",
+        static_url_path="/static",
+        template_folder="templates",
     )
     app.config["MAX_CONTENT_LENGTH"] = MAX_CONTENT_LENGTH
     app.jinja_options = JINJA_ENV_CONFIG  # Overload config
@@ -97,7 +101,9 @@ def _patch_server_header(backend_config: BackendConfig, hyper_config: HyperConfi
     # ...then patch `response_headers()` to add our custom server header
     vanilla_response_headers = hyper_config.response_headers
 
-    def response_headers_with_parsec_server_header(protocol: str) -> List[Tuple[bytes, bytes]]:
+    def response_headers_with_parsec_server_header(
+        protocol: str,
+    ) -> List[Tuple[bytes, bytes]]:
         headers = vanilla_response_headers(protocol)
         headers.append(server_header_tuple)
         return headers
@@ -123,7 +129,8 @@ async def serve_backend_with_asgi(
             # Timestamp is added by the log processor configured in `parsec.logging`,
             # here we configure peer address + req line + rep status + rep body size + time
             # (e.g. "GET 88.0.12.52:54160 /foo 1.1 404 823o 12343ms")
-            "access_log_format": "%(h)s %(r)s %(s)s %(b)so %(D)sus",
+            "logger_class": ParsecLogger,
+            "access_log_format": "%(h)s %(r)s %(s)s %(b)so %(D)sus (author: %(author)s)",
             "errorlog": logging.getLogger("hypercorn.error"),
             "certfile": str(ssl_certfile) if ssl_certfile else None,
             "keyfile": str(ssl_keyfile) if ssl_certfile else None,
--- a/parsec/backend/asgi/__init__.py
+++ b/parsec/backend/asgi/__init__.py
@@ -130,7 +130,7 @@ async def serve_backend_with_asgi(
             # here we configure peer address + req line + rep status + rep body size + time
             # (e.g. "GET 88.0.12.52:54160 /foo 1.1 404 823o 12343ms")
             "logger_class": ParsecLogger,
-            "access_log_format": "%(h)s %(r)s %(s)s %(b)so %(D)sus (author: %(author)s)",
+            "access_log_format": "%(h)s %(r)s %(s)s %(b)so %(D)sus %(author)s",
             "errorlog": logging.getLogger("hypercorn.error"),
             "certfile": str(ssl_certfile) if ssl_certfile else None,
             "keyfile": str(ssl_keyfile) if ssl_certfile else None,
--- a/parsec/backend/asgi/logger.py
+++ b/parsec/backend/asgi/logger.py
@@ -33,7 +33,7 @@ class ParsecLogger(Logger):
     def _find_author_header(self, headers: Iterable[Tuple[bytes, bytes]]) -> Optional[bytes]:
         for (name, value) in headers:
             # latin1 (ISO-8859-1) is standard http charset
-            if name.decode("latin1").lower() == "author":
+            if name.decode("latin1") == "Author":
                 return value
 
         return None
--- a/parsec/backend/asgi/logger.py
+++ b/parsec/backend/asgi/logger.py
@@ -39,11 +39,13 @@ class ParsecLogger(Logger):
         return None
 
     def _try_decode_base64(self, value: Union[bytes, None]) -> Optional[str]:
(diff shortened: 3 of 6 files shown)


parsec/backend/asgi/logger.py
```diff
@@ -0,0 +1,51 @@
...
+
+        id = self._try_decode_base64(id_value)
+        access_log[b"Author".decode("latin1").lower()] = id
+
+        return access_log
+
+    def _find_author_header(self, headers: Iterable[Tuple[bytes, bytes]]) -> Optional[bytes]:
+        for (name, value) in headers:
+            # latin1 (ISO-8859-1) is standard http charset
+            if name.decode("latin1").lower() == "author":
```
FirelightFlagboy, line 36: The `author` header would be capitalize in the request

```suggestion
            if name.decode("latin1").lower() == "Author":
```

tests/backend/test_logger.py
```diff
@@ -0,0 +1,65 @@
...
+
+
+def _create_http_scope(author: bytes = bytes(), add_author: bool = True) -> HTTPScope:
+    scope = HTTPScope()
+    scope["type"] = "http"
+    scope["method"] = "GET"
+    scope["query_string"] = b"/"
+    scope["path"] = ""
+    scope["scheme"] = ""
+    scope["headers"] = [(b"Author", author)] if add_author else []
```
FirelightFlagboy, line 20:
```suggestion
def _create_http_scope(author: Optional[bytes] = None) -> HTTPScope:
    scope = HTTPScope()
    scope["type"] = "http"
    scope["method"] = "GET"
    scope["query_string"] = b"/"
    scope["path"] = ""
    scope["scheme"] = ""
    scope["headers"] = [(b"Author", author)] if author is not None else []
```
Wouldn't be simpler by using an `Optional` field ?

FirelightFlagboy (changes requested):

FirelightFlagboy (approved):
LGTM

parsec/backend/asgi/__init__.py
```diff
@@ -123,7 +129,8 @@ async def serve_backend_with_asgi(
             # Timestamp is added by the log processor configured in `parsec.logging`,
             # here we configure peer address + req line + rep status + rep body size + time
             # (e.g. "GET 88.0.12.52:54160 /foo 1.1 404 823o 12343ms")
-            "access_log_format": "%(h)s %(r)s %(s)s %(b)so %(D)sus",
+            "logger_class": ParsecLogger,
+            "access_log_format": "%(h)s %(r)s %(s)s %(b)so %(D)sus (author: %(author)s)",
```
touilleMan, line 133: would be better to avoid having `(author: anonymous or invalid)` in the log when the author is not available.

For this you can put the ` (author: %(author)s)` formatting logic within `ParsecLogger`, and only have `"access_log_format": "%(h)s %(r)s %(s)s %(b)so %(D)sus%(author)s",` here

parsec/backend/asgi/logger.py
```diff
@@ -0,0 +1,51 @@
...
+import base64
+import binascii
+
+if TYPE_CHECKING:
+    from hypercorn.config import Config
+    from hypercorn.typing import ResponseSummary, WWWScope
+
+
+class ParsecLogger(Logger):
+    def __init__(self, config: "Config") -> None:
```
touilleMan, line 15: no need to overload this `__init__` given it just calls the parent one

parsec/backend/asgi/logger.py
```diff
@@ -0,0 +1,51 @@
...
+
+    def atoms(
+        self, request: "WWWScope", response: "ResponseSummary", request_time: float
+    ) -> Mapping[str, str]:
+        """Create and return an access log atoms dictionary.
+        This can be overidden and customised if desired. It should
+        return a mapping between an access log format key and a value.
+        """
+        access_log = AccessLogAtoms(request, response, request_time)
+        id_value = self._find_author_header(request["headers"])
```
touilleMan, line 26: `_find_author_header` is pretty simple and only called here, so no need for a function, you can even make it a one liner:
```python
author_header_value = next(v for k, v in headers if k.decode("latin1").lower() == "author", None)
```

touilleMan: buuuuuut ! in fact this work is already done by AccessLogAtoms:

https://github.com/pgjones/hypercorn/blob/4645aa0e41815e2193104331c580592c6aae8dd5/src/hypercorn/logging.py#L134-L135

so we just have to do:
```python
header_author_value = access_log["authori"]
```

parsec/backend/asgi/logger.py
```diff
@@ -0,0 +1,51 @@
...
+
+        id = self._try_decode_base64(id_value)
```
touilleMan, line 28: same  as  above: just inline `_try_decode_base64` here, this will be more readable

parsec/backend/asgi/logger.py
```diff
@@ -0,0 +1,51 @@
...
+        access_log[b"Author".decode("latin1").lower()] = id
```
touilleMan, line 29: no need for the decode & lower here, you can directly do `access_log["author"] = id`

parsec/backend/asgi/logger.py
```diff
@@ -0,0 +1,51 @@
...
+        return None
+
+    def _try_decode_base64(self, value: Union[bytes, None]) -> Optional[str]:
+        DEFAULT_MESSAGE = "anonymous or invalid"
+        if value is None:
+            return DEFAULT_MESSAGE
+
+        try:
+            return base64.b64decode(value.decode(), validate=True).decode()
+        except binascii.Error:
```
touilleMan, line 48: protip: you can catch multiple exceptions at once:
```python
except (binascii.Error, UnicodeDecodeError):
    ...
```

parsec/backend/asgi/logger.py
```diff
@@ -0,0 +1,51 @@
...
+
+        return None
+
+    def _try_decode_base64(self, value: Union[bytes, None]) -> Optional[str]:
+        DEFAULT_MESSAGE = "anonymous or invalid"
+        if value is None:
+            return DEFAULT_MESSAGE
+
+        try:
+            return base64.b64decode(value.decode(), validate=True).decode()
```
touilleMan, line 47: `value.decode("latin1")` ?

newsfragments/2953.feature.rst
```diff
@@ -0,0 +1 @@
+Extended Logger to show device id when a client is authenticated
```
touilleMan, line 1: this change is reaally small, so don't include it in the newsfragment (i.e. create an empty `2953.empty.rst` instead of this file)

tests/backend/test_logger.py
```diff
@@ -0,0 +1,65 @@
+# Parsec Cloud (https://parsec.cloud) Copyright (c) AGPL-3.0 2016-present Scille SAS
+
+from typing import Mapping, Optional
+from hypercorn.config import Config as HyperConfig
```
touilleMan, line 4: we divide import in 3 category (each separated by an empty line)

1- Python std lib and 3rd party imports
2 - Parsec imports
3 - Tests imports

so here it should be:
```python

from typing import Mapping, Optional
import base64
from hypercorn.config import Config as HyperConfig
from hypercorn.typing import ResponseSummary, HTTPScope

from parsec.backend.asgi.logger import ParsecLogger
```

tests/backend/test_logger.py
```diff
@@ -0,0 +1,65 @@
...
+    assert mapped["author"] == "alice@work"
+
+
+def test_bad_base64_author():
+    import secrets
+
+    logger = ParsecLogger(HyperConfig())
+
+    # We use a sequence of 10 random bytes to simulate a bad base64 encoded author name
+    author_bytes = secrets.token_bytes(10)
```
touilleMan, line 52: This is too complicated, and we may end up with valid base64 data which makes this test is flaky ^^

So just use an invalid base64 literal, I personally uses `<dummy>` for that

touilleMan (changes requested):

parsec/backend/asgi/logger.py
Volham22: Done.

parsec/backend/asgi/logger.py
Volham22: Done

parsec/backend/asgi/logger.py
Volham22: Done.

tests/backend/test_logger.py
Volham22: Done.

parsec/backend/asgi/logger.py
Volham22: Done.

parsec/backend/asgi/logger.py
Volham22: Done.

parsec/backend/asgi/logger.py
Volham22: Done.

parsec/backend/asgi/__init__.py
Volham22: Done.

tests/backend/test_logger.py
Volham22: Done.

newsfragments/2953.feature.rst
```diff
@@ -1 +0,0 @@
-Extended Logger to show device id when a client is authenticated
```
FirelightFlagboy: Delete the file

FirelightFlagboy (changes requested):

FirelightFlagboy (approved):

touilleMan (approved):

PR #3068, opened 2022-09-13
Add job to build Unsigned APK file for Android

FirelightFlagboy:


.github/workflows/package-ci.yml
```diff
@@ -369,3 +369,56 @@ jobs:
...
+      - name: Install dependencies for Android
+        run: sdkmanager --install "ndk;${{ env.NDK_VERSION }}" "build-tools;${{ env.SDK_VERSION }}"
+
+      - name: Install dependencies for ionic project
+        run: npm clean-install
+        working-directory: oxidation/client
+
+      - name: Build ionic for Android
+        run: |
+          npx ionic build android --no-open --prod
```
touilleMan, line 411: It would be better to include this into `client/package.json` so that we could do `npm run android:prod-copy` / `npm run electron:prod-copy` etc.

touilleMan (changes requested):

Landeers (approved):

PR #3072, opened 2022-09-13
Fix invalid timestamp error in pg

TimeEngineer:


parsec/backend/postgresql/handler.py
```diff
@@ -25,6 +25,8 @@
 
 CREATE_MIGRATION_TABLE_ID = 2
 MIGRATION_FILE_PATTERN = r"^(?P<id>\d{4})_(?P<name>\w*).sql$"
+# We use int to avoid float error
+US_BETWEEN_1970_AND_2000: int = 946684800000000
```
FirelightFlagboy, line 29:
```suggestion
MICRO_SECONDS_BETWEEN_1970_AND_2000: int = 946684800000000
```
I would prefer a more explicit naming for this constant

parsec/backend/postgresql/handler.py
```diff
@@ -25,6 +25,8 @@
 
 CREATE_MIGRATION_TABLE_ID = 2
 MIGRATION_FILE_PATTERN = r"^(?P<id>\d{4})_(?P<name>\w*).sql$"
+# We use int to avoid float error
```
FirelightFlagboy, line 28:
```suggestion
# The duration between 1970 and 2000 in microseconds.
#
# > We use int to avoid float error
```

FirelightFlagboy (changes requested):
The news fragment should be a `bugfixe`

parsec/backend/postgresql/handler.py
touilleMan, line 28: should add a comment to explain why this is needed:

```
# Internally, PostgreSQL stores timestamps as 4bytes floating number representing the number of seconds since
# 2000-01-01T00:00:00Z, hence we need to do back-and-forth conversion with the regular 1970-based POSIX epoch
```

tests/backend/test_postgres_db.py
```diff
@@ -158,3 +162,32 @@ async def test_retry_policy_allow_retry(postgresql_url, asyncio_loop):
...
+
+    async with triopg.connect(postgresql_url) as conn:
+        await handle_datetime(conn, "timestamptz")
+
+        await conn.execute(
+            """
+            DROP TABLE IF EXISTS datetime;
+            CREATE TABLE IF NOT EXISTS datetime (
+                _id SERIAL PRIMARY KEY,
+                timestamp TIMESTAMPTZ
```
touilleMan, line 180: should also test `TIMESTAMP` type too

tests/backend/test_postgres_db.py
```diff
@@ -158,3 +162,32 @@ async def test_retry_policy_allow_retry(postgresql_url, asyncio_loop):
...
+            )"""
+        )
+
+        async with conn.transaction():
+            await conn.execute(
+                *Q("INSERT INTO datetime (_id, timestamp) VALUES (0, $timestamp)")(timestamp=now)
```
touilleMan, line 186: no need for `Q` here
this is useful when building complex queries, here we can just do 
```python
execute("INSERT INTO datetime (_id, timestamp) VALUES (0, $1)", now)
```

tests/backend/test_postgres_db.py
```diff
@@ -158,3 +162,32 @@ async def test_retry_policy_allow_retry(postgresql_url, asyncio_loop):
 
             # Cancel the backend nursery
             nursery.cancel_scope.cancel()
+
+
+@pytest.mark.trio
+@pytest.mark.postgresql
+async def test_datetime(postgresql_url, backend_factory):
```
touilleMan, line 169: better name: `test_rust_datetime_correctly_serialized`

tests/backend/test_postgres_db.py
```diff
@@ -158,3 +162,32 @@ async def test_retry_policy_allow_retry(postgresql_url, asyncio_loop):
...
+    now = DateTime.now()
```
touilleMan, line 170: What the test tests:
- serialization with rust datetime
- deserialization with python datetime

what should also be tested:
- serialization with python datetime
- deserialization with rust datetime

I think the simplest way to do that is to open two connections at the same time
```python
async with triopg.connect(postgresql_url) as vanilla_conn:
    async with triopg.connect(postgresql_url) as custom_conn:
        await handle_datetime(custom_conn, "timestamptz")

        await vanilla_conn.execute(...)  # Insert datetime
        await patched_conn.execute(...)  # Insert datetime

        await vanilla_conn.execute(...)  # retrieve datetime inserted by vanilla
        await vanilla_conn.execute(...)  # retrieve datetime inserted by patched
        await patched_conn.execute(...)  # retrieve datetime inserted by vanilla
        await patched_conn.execute(...)  # retrieve datetime inserted by patched
```

vxgmichel (approved):

PR #3107, opened 2022-09-16
Added a prefix to test env name so it can be more easily recognized

Max-7:
Since tmp folders created with `tests/scripts/run_testenv.py` cannot be automatically deleted, I added a little prefix to be able to recognize them.

### DIFF
--- a/tests/scripts/run_testenv.py
+++ b/tests/scripts/run_testenv.py
@@ -47,7 +47,7 @@ DEFAULT_BLOCKSTORE = "MOCKED"
 
 async def new_environment(source_file=None):
     export_lines = []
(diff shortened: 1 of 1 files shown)


tests/scripts/run_testenv.py
```diff
@@ -47,7 +47,7 @@
 
 async def new_environment(source_file=None):
     export_lines = []
-    tempdir = tempfile.mkdtemp()
+    tempdir = tempfile.mkdtemp(prefix="parsec-tmpenv-")
```
vxgmichel, line 50: What about `parsec-testenv-`?

vxgmichel (approved):

PR #3111, opened 2022-09-16
Oxidation use gen proto time

TimeEngineer:


FirelightFlagboy (changes requested):
Waiting for the packaging problem in master to be corrected

src/time.rs
```diff
@@ -236,9 +216,11 @@ impl DateTime {
 }
 
 #[pyclass]
-#[derive(PartialEq, Eq, Clone)]
+#[derive(Clone)]
 pub(crate) struct LocalDateTime(pub libparsec::types::LocalDateTime);
 
+crate::binding_utils::gen_proto!(LocalDateTime, __repr__);
```
touilleMan, line 222: no implementation of __richcmp__ to replace the derive partialeq/eq ?

FirelightFlagboy: Pyo3 don't use the trait `PartialEq` / `Eq`.
`__richcmp__` was never implemented for `LocalDateTime`

see [Pyo3 comparisons](https://pyo3.rs/v0.17.1/class/object.html#comparisons)

FirelightFlagboy (approved):

touilleMan (approved):

PR #3113, opened 2022-09-16
Oxidation use gen proto local device

TimeEngineer:


FirelightFlagboy (changes requested):
Waiting for the packaging problem in master to be corrected

touilleMan (approved):

FirelightFlagboy (approved):

PR #3121, opened 2022-09-19
Relax `test_eyeballs` expected delay timeout to 1s

FirelightFlagboy:
On the test `test_eyeballs` we check if it don't take too much time to execute. The minimal possible value for the test is `800ms` (`TIMEOUT * 2 + 300`). Previously we were adding 100ms for the overhead of async code but time to time the test will fail. Here we're adding another 100ms for the overhead and should drastically reduce the number of false positive

### DIFF
--- a/oxidation/libparsec/crates/platform_async/tests/eyeballs.rs
+++ b/oxidation/libparsec/crates/platform_async/tests/eyeballs.rs
@@ -6,25 +6,72 @@ use libparsec_platform_async::{
     JoinSet, Notify,
 };
 use std::{
+    fmt::Display,
     sync::{Arc, Mutex},
     time::Duration,
 };
 
 const TIMEOUT: u64 = 250;
+const TIMEOUT_DURATION: Duration = Duration::from_millis(TIMEOUT);
 
 #[derive(Debug)]
 struct Load {
+    id: u32,
     delay: Duration,
     result: Mode,
+    timer: Mutex<Option<Instant>>,
+    duration: Mutex<Option<Duration>>,
 }
 
 impl Load {
-    fn new(delay: Duration, result: Mode) -> Self {
-        Self { delay, result }
+    fn new(id: u32, delay: Duration, result: Mode) -> Self {
+        Self {
+            id,
+            delay,
+            result,
+            timer: Mutex::new(None),
+            duration: Mutex::new(None),
+        }
+    }
+
+    fn expected_duration(&self) -> Duration {
+        if self.delay < TIMEOUT_DURATION || self.result == Mode::Success {
+            self.delay
+        } else {
+            TIMEOUT_DURATION
+        }
+    }
+
+    async fn start(&self) -> Mode {
+        self.timer.lock().unwrap().replace(Instant::now());
+        log::debug!("[{self}] start");
+        libparsec_platform_async::sleep(self.delay).await;
+        self.stop();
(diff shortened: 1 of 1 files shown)


oxidation/libparsec/crates/platform_async/tests/eyeballs.rs
```diff
@@ -43,7 +43,11 @@ async fn test_eyeballs() -> Result<(), ()> {
     assert_eq!(Ok(2), dbg!(eyeballs(targets).await));
     let elapsed = start_time.elapsed();
     log::info!("elapsed time {}ms", elapsed.as_millis());
-    assert!(elapsed < Duration::from_millis(900));
+    // The expected time calculated following these rules:
+    // - The first 2 load will take to much time, so we're adding `TIMEOUT * 2`
+    // - The 3th load will success after 300ms, so we add `300`
+    // - We add some value to take account the possible overhead to switch between async task (here 200ms).
+    assert!(elapsed < Duration::from_millis(250 * 2 + 300 + 200));
```
touilleMan, line 50: I don't think this is the right way to fix this: `elapsed` is very dependent on how fast the machine is, and what is le the lag in polling the futures (I would be surprise if on Windows or in some specific browsers the lag is very high...)

So what we can test is:
- elapsed took less that <expected duration> (if it took less it means the code is not doing what we think, for instance the sleep is not actually sleeping)
- elapsed took a lot more that <expected duration>, of course it is hard to define "a lot more" (hence this test being flaky !) so I'm not sure what/if we should put here. BTW in Python we use this kind of check to ensure the test doesn't have a deadlock, but this is complicated to do here and the test is simple so I think it is not worth it.
- an order of resolution is the tasks, do you think we could easily capture the logs to everything occurs in the right order ?

FirelightFlagboy: 1. Yes we can do that
2. Hard to define maybe but it should take less than 2.3 sec (The duration of the first load)
3. This would complexify the test

touilleMan: > This would complexify the test

I agree, The main reason I was wondering about this caplog thing is because it's very easy to do in Pytest and seems like an interesting tool to have in Rust as well ;-)

> Hard to define maybe but it should take less than 2.3 sec (The duration of the first load)

After more thinking, I think the best thing to do for this test is to check the outcome of each job (so something like we should always have id X in succes, id Y,Z in cancelled and id W not started)

This we could add the also add the "duration expected to be more than x" check and "duration expected to be less than y", but let's say add 1s to y (considering 1s is a huge lag, I'm totally not sure it is the case, but it seems a good approximation...)

FirelightFlagboy: We could test the output log with something like [logtest](https://docs.rs/logtest/latest/logtest/).
But it seems that what we want is to check the result of each load task.
And we could use a channel for that.

1s is an good approximation for the task that take 2.3 seconds, but is not for the one that take 100ms. Maybe if we say 50% of the load task duration or at least 100ms ?

touilleMan: > Maybe if we say 50% of the load task duration or at least 100ms ?

👍

touilleMan (approved):

PR #3139, opened 2022-09-22
[GUI] Updated how workspaces are loaded

### LINKED ISSUE #3125: Separate the listing of workspaces from the fetching of user roles in the GUI
When a user logs in, the workspace list is not displayed until the `list_workspaces` job has returned. However, this job might take a very long time (several seconds on a bad connection) due the user roles being fetched from server (there is currently no caching nor storing the user roles in the local storage). 

A relatively simple fix to this issue is to use two jobs instead:
- `list_workspaces` that returns the information for all workspaces (most of the time, all of this information is in the local storage)
- `list_user_roles` that returns the user roles for all the workspaces

Then the workspaces are displayed as soon as the first job returns and workspace widgets are updated later with the content of the second job.

Another approach is to delegate the fetching of the user roles to a job specific to the different workspace widgets (called `WorkspaceButton` in the code base).

Max-7:
Closes #3125 

Removed `test_workspace_button.py` as it would now need `core` and `jobs_ctx`. It makes it too complicated to setup the tests and those tests would be too close to those in `test_workspaces.py` to be relevant anyway.


tests/core/gui/test_files.py
```diff
@@ -984,13 +984,14 @@ async def test_sort_menu(aqtbot, files_widget_testbed, monkeypatch):
...
 
 @pytest.mark.gui
-@pytest.mark.flaky(reruns=1)
 @pytest.mark.trio
 async def test_current_folder_status_menu(
     running_backend, aqtbot, files_widget_testbed, catch_file_status_widget
 ):
     f_w = files_widget_testbed.files_widget
 
+    await f_w.workspace_fs.sync()
```
vxgmichel, line 993: gg :+1:

vxgmichel (approved):
A print to remove, LGTM otherwise :+1:

PR #2903, opened 2022-08-24
Oxidation replace `WorkspaceStorage` & `WorkspaceStorageSnapshot` & `UserStorage`

FirelightFlagboy:
- Oxidize `WorkspaceStorage`.
- Oxidize `WorkspaceStorageSnapshot`.
- Oxidize `workspace_storage_non_speculative_init`
- Type some tests (the typed tests where those that failed during the migration).
- Correct some tests to work with the oxidation.
- Rename `WorkspaceStorageTimestamped` to `WorkspaceStorageSnapshot`.

## TODOs

- [x] use the change from #3224
  - [x] fix `test_database_with_invalid_pattern_resilience` in `tests/core/fs/workspacefs/test_confinement.py`
    - [x] Directly modify the database to insert an invalid regex string
- [x] use the change from #3523 
  - [x] On the rust side we need to close every sqlite connection  

closes #3836, closes #3170, closes #3169, closes #3168, closes #2898, closes #2864

PR #3141, opened 2022-09-23
Add test to check for leak of file descriptor

FirelightFlagboy:


tests/core/fs/test_fs_online_rwfile_and_sync.py
```diff
@@ -13,6 +15,47 @@
...
+            await self.restart_user_fs(self.device)
+
+        def get_root_nursery(self):
+            return nursery
+
+        @rule()
+        async def ignore(self):
+            pass
+
+    import trio
```
touilleMan, line 44: move import to the top of the file

tests/core/fs/test_fs_online_rwfile_and_sync.py
```diff
@@ -13,6 +15,47 @@
 PLAYGROUND_SIZE = BLOCK_SIZE * 10
 
 
+@pytest.mark.trio
+@pytest.mark.slow
+async def test_leak_opened_file(user_fs_online_state_machine, alice: LocalDevice):
+    LOGGER = structlog.getLogger(test_leak_opened_file.__name__)
+    import resource
+
+    resource.setrlimit(resource.RLIMIT_NOFILE, [1024, 1024])
```
touilleMan, line 24: We must undo this once the test is done, otherwise this may create unexpected side effects in other tests

```python
vanilla_rlimit = resource.getrlimit()
resource.setrlimit(my_rlimit)
try:
    # do the test here
finally:
    resourec.setrlimit(vanilla_rlimit)
```

tests/core/fs/test_fs_online_rwfile_and_sync.py
```diff
@@ -13,6 +15,47 @@
 PLAYGROUND_SIZE = BLOCK_SIZE * 10
 
 
+@pytest.mark.trio
+@pytest.mark.slow
```
touilleMan, line 19: Don't you need `@customize_fixture(real_data_storage=True)` ? Otherwise the sqlite is in memory ? (or this is no longer needed because the new rust storage doesn't use this customization ? but in that case do we need to do something as documentation/deprecation/change in another PR ?)

tests/core/fs/test_fs_online_rwfile_and_sync.py
```diff
@@ -13,6 +15,47 @@
...
+async def test_leak_opened_file(user_fs_online_state_machine, alice: LocalDevice):
+    LOGGER = structlog.getLogger(test_leak_opened_file.__name__)
+    import resource
+
+    resource.setrlimit(resource.RLIMIT_NOFILE, [1024, 1024])
+
+    class FSOnlineRwFileAndAsync(user_fs_online_state_machine):
```
touilleMan, line 26: `user_fs_online_state_machine` is an hypothesis state machine. We'd better not use it for regular test.

Here what we are doing in a loop is 1) creating a backend & a userfs (hence it underlying local storage) 2) creating a workspace.

This can be achieve very easily using the regular backend_factory/user_fs_factory fixtures.

However I wonder if this test is not too broad: checking the leak of file descriptor is only needed for component that actually interact with the outside world (remembre the `platform_...` crates in oxidation ? ^^).

So I think the test would be simpler by only testing file descriptor leakage on the storage component (for instance the lowest-level `LocalDatabase` class, considering all the other components will use it with a valid async context manager and hence we can trust they won't cause a leak unless the code is very very broken)

tests/core/fs/test_fs_online_rwfile_and_sync.py
touilleMan: Is there a way to avoid changing the limit ? It would be much better to get the current number of opened file descriptor, this way we wouldn't have to do a 10000 iteration loop, only a couple of cycle to be sure would be enough)

touilleMan (changes requested):

tests/core/fs/test_fs_online_rwfile_and_sync.py
FirelightFlagboy: 🤷 anyway this test is detecting file descriptor leaks, even with the fixture customization it currently leaks fd

tests/core/fs/test_fs_online_rwfile_and_sync.py
FirelightFlagboy: I've talked with @vxgmichel about it, I'll try to use [`psutil.Process.open_files`](https://psutil.readthedocs.io/en/latest/#psutil.Process.open_files) & [`psutil.Process.connections`](https://psutil.readthedocs.io/en/latest/#psutil.Process.connections) to compare the number of openned fd between a run.

We can't just increase the iteration loop because it would take a lot of time.
> I've already tested that before settling on `ulimit` but I've had to kill the test after ~1h30.

tests/core/fs/test_fs_online_rwfile_and_sync.py
FirelightFlagboy: You're right, currently the test is a snipped that is able to detected a fd leaks. When we known from where the leak come from, we could create a test specifically for it.

tests/core/fs/test_fs_online_rwfile_and_sync.py
```diff
@@ -13,6 +15,47 @@
...
+
+        @rule()
+        async def ignore(self):
+            pass
+
+    import trio
+
+    m = n = 100
+
+    async with trio.open_nursery() as nursery:
```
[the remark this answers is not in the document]
FirelightFlagboy: It's used on line `56`.
> It ain't much, but it's honest work

tests/core/fs/test_fs_online_rwfile_and_sync.py
```diff
@@ -13,6 +15,47 @@
 PLAYGROUND_SIZE = BLOCK_SIZE * 10
 
 
+@pytest.mark.trio
+@pytest.mark.slow
+async def test_leak_opened_file(user_fs_online_state_machine, alice: LocalDevice):
+    LOGGER = structlog.getLogger(test_leak_opened_file.__name__)
+    import resource
```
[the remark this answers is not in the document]
FirelightFlagboy: done

tests/core/fs/test_fs_online_rwfile_and_sync.py
FirelightFlagboy: done

FirelightFlagboy:
superseded by #2903

FirelightFlagboy:
Closing #2903 replace it with a better check

PR #3142, opened 2022-09-23
Links can now be created with timestamp

### LINKED ISSUE #3123: Include optional timestamp in parsec links
When a user generates a link to a file in a snapshot workspace, it would make sense to include the corresponding timestamp in the link in order for the recipient to mount the workspace at the corresponding timestamp and get this specific version of the file.

Note: this feature can then be also enabled when a user generates a link to a file in a regular workspace, in order to generate some kind of "permalink". For instance, there could be two entries in the context menu:
- Get a link to the file
- Get a permalink to the file

Volham22:
This close #3123. It adds the possibility to create links that contains timestamps to reference a file at a given time point.


TimeEngineer (changes requested):

src/addrs.rs
```diff
@@ -470,10 +471,19 @@ impl BackendOrganizationFileLinkAddr {
                 organization_id.0,
                 workspace_id.0,
                 encrypted_path,
+                encrypted_timestamp,
             ),
         ))
     }
 
+    #[getter]
+    fn encrypted_timestamp<'py>(&'py self, python: Python<'py>) -> Option<&'py PyBytes> {
```
TimeEngineer, line 480: Why `self` is bounded by `'py` lifetime ?

src/addrs.rs
```diff
@@ -470,10 +471,19 @@ impl BackendOrganizationFileLinkAddr {
...
+        self.0
+            .encrypted_timestamp()
+            .as_ref()
+            .map(|v| PyBytes::new(python, v.as_slice()))
```
TimeEngineer, line 484: &v is enough

TimeEngineer (changes requested):

parsec/core/gui/workspaces_widget.py
```diff
@@ -469,6 +480,7 @@ def on_mount_success(self, job):
         self.reset()
 
     def on_mount_error(self, job):
+        print("ERROR", job.exc)
```
TimeEngineer, line 483: We shouldn't `print` in the codebase

parsec/core/gui/workspaces_widget.py
```diff
@@ -484,6 +496,7 @@ def on_unmount_success(self, job):
         self.reset()
 
     def on_unmount_error(self, job):
+        print("UMOUNT_ERROR", job.exc)
```
TimeEngineer, line 499: Same

TimeEngineer (changes requested):

oxidation/libparsec/crates/types/src/addr.rs
```diff
@@ -632,11 +636,29 @@ impl BackendOrganizationFileLinkAddr {
             return Err("Multiple values for param `path`");
         }
 
+        let mut ts_range = pairs.filter(|(k, _)| k == "timestamp");
```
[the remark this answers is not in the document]
Volham22: I think `filter` was used here  to test if a value is given multiple times in the query string and if so return an error.

src/addrs.rs
Volham22: You're right `self` does not need to be bound to the `'py` lifetime here

oxidation/libparsec/crates/types/src/addr.rs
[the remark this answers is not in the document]
TimeEngineer: `ts_range` is used twice, one for the definition, and the other to call the method `next`, so it seems to be `find`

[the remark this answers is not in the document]
Volham22: Done.

oxidation/libparsec/crates/types/src/addr.rs
```diff
@@ -613,41 +614,34 @@ impl BackendOrganizationFileLinkAddr {
...
-            return Err("Multiple values for param `path`");
-        }
 
-        let mut ts_range = pairs.filter(|(k, _)| k == "timestamp");
-        let encrypted_timestamp = if let Some((_, value)) = ts_range.next() {
-            Some(binary_urlsafe_decode(&value).or(Err("Invalid `timestamp` param value"))?)
-        } else {
-            None
-        };
+            query_str_map.insert(k, v);
```
TimeEngineer, line 625: use insert directly:
```Rust
pub fn insert(&mut self, k: K, v: V) -> Option<V>
```
```
Inserts a key-value pair into the map.

If the map did not have this key present, None is returned.
```

TimeEngineer (changes requested):

oxidation/libparsec/crates/types/src/addr.rs
```diff
@@ -613,41 +614,34 @@ impl BackendOrganizationFileLinkAddr {
...
-            return Err("Multiple values for param `workspace_id`");
-        }
-
-        let mut path_queries = pairs.filter(|(k, _)| k == "path");
-        let encrypted_path = match path_queries.next() {
-            None => return Err("Missing mandatory `path` param"),
-            Some((_, value)) => {
-                binary_urlsafe_decode(&value).or(Err("Invalid `path` param value"))?
+        let mut query_str_map: HashMap<String, String> = HashMap::new();
+        for (key, value) in *pairs {
```
TimeEngineer, line 618: `pairs` is not used later, consume it instead of building `String`

TimeEngineer (changes requested):

oxidation/libparsec/crates/types/src/addr.rs
```diff
@@ -613,41 +614,34 @@ impl BackendOrganizationFileLinkAddr {
...
-            workspace_id,
-            encrypted_path,
-            encrypted_timestamp,
+            workspace_id: query_str_map
+                .get("workspace_id")
+                .map(|v| v.parse::<EntryID>())
+                .ok_or("no `workspace_id` key")?
+                .map_err(|_| "invalid `workspace_id`")?,
+            encrypted_path: query_str_map
+                .get("encrypted_path")
```
TimeEngineer, line 637: it's path

oxidation/libparsec/crates/types/src/addr.rs
```diff
@@ -613,40 +614,32 @@ impl BackendOrganizationFileLinkAddr {
...
             }
-        };
-        if path_queries.next().is_some() {
-            return Err("Multiple values for param `path`");
         }
 
-        let mut ts_range = pairs.filter(|(k, _)| k == "timestamp");
-        let encrypted_timestamp = if let Some((_, value)) = ts_range.next() {
-            Some(binary_urlsafe_decode(&value).or(Err("Invalid `timestamp` param value"))?)
+        let encrypted_timestamp = if let Some(ts) = query_str_map.get("timestamp") {
```
TimeEngineer, line 624: can you `map` instead ?

TimeEngineer (approved):

parsec/core/gui/file_table.py
```diff
@@ -269,7 +271,13 @@ def show_context_menu(self, pos):
...
             action.triggered.connect(self.show_status_clicked.emit)
             action = menu.addAction(_("ACTION_FILE_MENU_GET_FILE_LINK"))
-            action.triggered.connect(self.file_path_clicked.emit)
+
+            # Show the option to create a timestamped share link to the user is
+            # useless here because it has the same effect as creating a regular file link
+            if not self.has_timestamped_file:
+                action.triggered.connect(self.file_path_clicked.emit)
+                action = menu.addAction(_("ACTION_FILE_MENU_GET_FILE_LINK_TIMESTAMP"))
+            action.triggered.connect(self.file_path_timestamp_clicked.emit)
```
Max-7, line 280: I would prefer it this was a little more explicit (ie: more ifs), because I can easily see someone adding an action at the wrong place later.
```python
action = menu.addAction(_("ACTION_FILE_MENU_GET_FILE_LINK"))
action = menu.addAction("My New Action")
action.triggered.connect(self.on_my_new_action_triggered)
if not self.has_timestamped_file:
    action.triggered.connect(self.file_path_clicked.emit)
    action = menu.addAction(_("ACTION_FILE_MENU_GET_FILE_LINK_TIMESTAMP"))
action.triggered.connect(self.file_path_timestamp_clicked.emit)
```

parsec/core/gui/file_table.py
```diff
@@ -103,6 +104,7 @@ def __init__(self, *args, **kwargs):
         self.previous_selection = []
         self.setColumnCount(len(Column))
         self.config = None
+        self.has_timestamped_file = False
```
Max-7, line 107: Don't like this variable name. It's not that the workspace has a timestamped file, they all do, it's that the workspace itself is mounted at a certain timestamp. Something like `is_timestamped_workspace` would be more appropriate.

parsec/core/gui/workspaces_widget.py
```diff
@@ -244,10 +244,21 @@ def goto_file_clicked(self):
         if button is not None:
             try:
                 path = button.workspace_fs.decrypt_file_link_path(addr)
+                ts = button.workspace_fs.decrypt_timestamp(addr)
+
+                # Mount workspace at the link's creation time
+                if ts is not None:
```
Max-7, line 250: And if we have no timestamp the workspace isn't mounted?

tests/core/gui/test_main_window.py
```diff
@@ -185,70 +187,109 @@ def _folder_ready():
...
-async def test_link_file_invalid_path(aqtbot, autoclose_dialog, logged_gui_with_files):
+@pytest.mark.parametrize("timestamp", [None, "timestamp"])
+async def test_link_file_invalid_path(aqtbot, autoclose_dialog, logged_gui_with_files, timestamp):
     logged_gui, w_w, f_w = logged_gui_with_files
-    url = f_w.workspace_fs.generate_file_link("/unknown")
+    await f_w.workspace_fs.sync()
 
-    logged_gui.add_instance(url.to_url())
+    with freeze_time(DateTime.now().add(seconds=1)):
+        if timestamp == "timestamp":
```
Max-7, line 264: Probably sexier to have the parameter as `with_timestamp [True, False]` rather than having a variable named `timestamp` that holds the value `"timestamp"`

parsec/core/gui/workspaces_widget.py
Volham22: Done.

Max-7 (approved):

PR #3146, opened 2022-09-26
Update vlob tests

TimeEngineer:


oxidation/libparsec/crates/protocol/tests/test_vlob.rs
```diff
@@ -37,7 +37,37 @@ use libparsec_types::{Maybe, ReencryptionBatchEntry};
         )
     )
 )]
-#[case::with_sequester_blob(
+#[case::without(
```
FirelightFlagboy, line 40:
```suggestion
#[case::without_sequester_blob(
```

oxidation/libparsec/crates/protocol/tests/test_vlob.rs
```diff
@@ -399,7 +429,37 @@ fn serde_vlob_read_rep(#[case] raw_expected: (&[u8], authenticated_cmds::vlob_re
...
```
FirelightFlagboy, line 432:
```suggestion
#[case::without_sequester_blob(
```

parsec/api/protocol/base.py
```diff
@@ -562,3 +562,15 @@ def req_dumps(self, req: dict[str, Any]) -> bytes:
 
     def rep_loads(self, raw: bytes) -> Any:
         return self.rep_schema.load(raw)
+
+    # Temporary Used for generate_data
+    def req_loads(self, raw: bytes) -> Any:
```
FirelightFlagboy, line 567:
```suggestion
    def req_loads(self, raw: bytes) -> Union[AuthenticatedAnyCmdReq, InvitedAnyCmdReq]:
```

FirelightFlagboy (changes requested):

oxidation/libparsec/crates/protocol/tests/test_vlob.rs
```diff
@@ -37,7 +37,37 @@ use libparsec_types::{Maybe, ReencryptionBatchEntry};
...
+    (
+        // Generated from Python implementation (Parsec v2.12.1+dev)
```
touilleMan, line 42: `Generated from Python` means generated with the old python-based schemas, is this really the case in 2.12.1 ?

TimeEngineer: True

oxidation/libparsec/crates/protocol/tests/test_vlob.rs
TimeEngineer: We shouldn't have a case for every field which can be nullable

parsec/api/protocol/base.py
TimeEngineer:
```Rust
fn load(_cls: &PyType, buf: Vec<u8>, py: Python) -> PyResult<PyObject>
```
It's neither AuthenticatedAnyCmdReq nor InvitedAnyCmdReq

oxidation/libparsec/crates/protocol/tests/test_vlob.rs
FirelightFlagboy: 🤷 the difference between `full` and `without` is the field `sequester_blob` that is set to `None` or something.

oxidation/libparsec/crates/protocol/tests/test_vlob.rs
touilleMan: Any reason why we don't genrate this vlob_create from the old Python schema ?
As long as we can do it, I'd rather do it to be extra careful (remember: the rust schema are not in production and they are precisely the one we should be worried about !)

Of course if this is a new test case added because there is a new field in the schema that is not supported by the Python schema it is fine to stick with the Rust schema (as long as we also already have a testcase with the old python schema ^^).

oxidation/libparsec/crates/protocol/tests/test_vlob.rs
TimeEngineer: But I prefer if it stays ISO with others with many fields set to `None`

oxidation/libparsec/crates/protocol/tests/test_vlob.rs
TimeEngineer: You shouldn't be worried about a `None` case, I can generate from a Python implementation, but only the comment will differ and need more time (need to checkout an old branch) :thinking:

TimeEngineer: image
image

FirelightFlagboy (approved):

PR #3151, opened 2022-09-27
Improve rust/python compat for InvitationType

### LINKED ISSUE #3105: Replace old `InvitationType` with new rust `InvitationType`
During the migration of the invite protocol the `InvitationType` enum had to be moved to rust.
Hovewer it's not as easy as simply replace the old one with the new one binded from rust.
Hence, it is still necessary to completely use the new one to avoid code duplications.
See the [invite protocol migration PR](https://github.com/Scille/parsec-cloud/pull/3102).

touilleMan:
fix #3105


src/protocol/invite.rs
```diff
@@ -42,25 +42,38 @@ pub(crate) struct InvitationType(pub libparsec::types::InvitationType);
...
+
+    #[getter]
+    fn value(&self) -> &'static str {
+        match self.0 {
+            libparsec::types::InvitationType::Device => "DEVICE",
+            libparsec::types::InvitationType::User => "USER",
+        }
     }
 
     fn __richcmp__(&self, other: &Self, op: CompareOp) -> PyResult<bool> {
```
Volham22, line 79: Can this `__richcmp__` be generated by the `gen_proto` macro? I didn't know this macro existed when I wrote it.
It may look like this

```suggestion
    gen_proto!(InvitationType, __richcmp__, eq);
```

Volham22 (approved):

parsec/api/protocol/invite.py
```diff
@@ -65,13 +67,31 @@
...
-class InvitationType(Enum):
-    USER = "USER"
-    DEVICE = "DEVICE"
+InvitationTokenField: Type[fields.Field] = fields.uuid_based_field_factory(InvitationToken)
 
 
-InvitationTokenField: Type[fields.Field] = fields.uuid_based_field_factory(InvitationToken)
-InvitationTypeField: Type[fields.BaseEnumField] = fields.enum_field_factory(InvitationType)
+class InvitationTypeField(Field):
+    def _serialize(self, value, attr, obj):  # type: ignore
```
FirelightFlagboy, line 74:
```suggestion
    def _serialize(self, value: Optional[InvitationType], attr: Any, obj: Any) -> Optional[str]:
```

parsec/api/protocol/invite.py
```diff
@@ -65,13 +67,31 @@
...
+            return None
+
+        if value == InvitationType.DEVICE:
+            return "DEVICE"
+        elif value == InvitationType.USER:
+            return "USER"
+        else:
+            raise ValidationError(f"Not a InvitationType")
+
+    def _deserialize(self, value, attr, data):  # type: ignore
```
FirelightFlagboy, line 85:
```suggestion
    def _deserialize(self, value: Any, attr: Any, data: Any) -> InvitationType:
```

src/protocol/invite.rs
touilleMan: done.

FirelightFlagboy (approved):

src/protocol/invite.rs
```diff
@@ -42,36 +42,43 @@ pub(crate) struct InvitationType(pub libparsec::types::InvitationType);
...
-        Self(libparsec::types::InvitationType::Device)
+    fn device() -> &'static PyObject {
+        lazy_static::lazy_static! {
+            static ref VALUE: PyObject = {
+                Python::with_gil(|py| {
+                    InvitationType(libparsec::types::InvitationType::Device).into_py(py)
+                })
+            };
+        };
+        &VALUE
```
TimeEngineer, line 55: I would like to say it's overkill to use lazy_static there, It's like a number

touilleMan: Nop because without this lazy_static, we would return a different object each time we do `InvitationType.USER`, which means `assert InvitationType.USER is InvatitionType.USER` would fail.
This is a bad idea because it would create hard to track bug when using `is` operator instead of `==` by mistake ;-)

touilleMan: Given we have already the GIL, we could use a static variable without relying on lazy_static (i.e. we would use the python mutex to protect our own static variable), however I found using lazy_static just simpler to implement (no risk of hidden bug because we removed the `py` from the function signature for whatever reason and now the static is subject to access concurrency)

Volham22 (approved):

TimeEngineer (changes requested):

TimeEngineer (approved):

PR #3160, opened 2022-09-29
Removed warning when event queue is full

### LINKED ISSUE #3124: Remove server warning log on "dropping event (queue is full)"
`dropping event (queue is full)` occurs when the client is no longer able to fetch the events fast enough, or more likely at all because it has left without telling the server.

So in any way instead of logging this we should just close the connection. This is better given if the client had a hiccup and is in fact still connected to the server it will miss some events which should be guaranteed not to occurs.

Volham22:
This close #3124.


parsec/backend/events.py
```diff
@@ -73,7 +74,8 @@ def _on_roles_updated(
                     EventsListenRepOkRealmRolesUpdated(realm_id, role)
                 )
             except trio.WouldBlock:
-                client_ctx.logger.warning("dropping event (queue is full)")
+                client_ctx.event_bus_ctx.clear()
+                abort(409)
```
touilleMan, line 78: This won't work: `_on_role_updated` is a callback being called through the event bus by the coroutine responsible to listening the events coming from postgresql. Hence raising an abort exception doesn't work given we are not in the context of an event_listen request !

Instead we should close the cancel_scope for the client (this will close the client connection in a clean way)
https://github.com/Scille/parsec-cloud/blob/3adc3c3a4978ff353582b851e094a9a35d7f6684/parsec/backend/asgi/ws.py#L68

So what is done for the `_on_revoked` event for instance. Of course the `_on_revoked` event has directly access to this cancel_scope, while the callbacks in events.py don't.
So we need to add this cancel_scope to the client_ctx object:

```python

    if isinstance(client_ctx, AuthenticatedClientContext):
        with trio.CancelScope() as client_ctx.cancel_scope:
            with backend.event_bus.connection_context() as client_ctx.event_bus_ctx:
                [...]
```
```

Comment (touilleMan) [reply parsec/backend/events.py:78]: And we don't need to clear the event queue: it will be automatically done by garbage collector when the client connection is terminated (and triggering the cancel scope is idempotent)

**Review by touilleMan — CHANGES_REQUESTED** [review]

File: parsec/backend/client_context.py [hunk 2]
```diff
@@ -27,13 +27,21 @@
...
 
     def __init__(self, api_version: ApiVersion):
         self.api_version = api_version
         self.conn_id = uuid4().hex
+        self._cancel_scope: Optional[trio.CancelScope] = None
+
+    def set_scope(self, client_scope: trio.CancelScope) -> None:
+        self._cancel_scope = client_scope
+
+    def exit_cancel_scope(self) -> None:
```
touilleMan, line 42: Good idea have a method that does "close the client's cancel scope if any"

But I think we should not have an assert here but just do nothing if there is no cancel scope. I theory this assert is a sanity check, but I suspect the code is complex enough that we will use this method in a callback that is also used in `rpc.py` (where there is no cancel scope) and create unexpected errors.

So the idea would be to rename this method something like `close_connection_asap`, which means we will try to close the connection if possible, and otherwise let it finish.
This is good enough in practice given the connections in rpc.py don't have a cancel scope because they are short lived (the execute a single command, unlike the websocket connections that stay open and execute multiple commands)

parsec/backend/client_context.py
```diff
@@ -27,13 +27,21 @@
...
+    __slots__ = ("conn_id", "api_version", "_cancel_scope")
     TYPE: ClientType
     logger: BoundLogger
 
     def __init__(self, api_version: ApiVersion):
         self.api_version = api_version
         self.conn_id = uuid4().hex
+        self._cancel_scope: Optional[trio.CancelScope] = None
+
+    def set_scope(self, client_scope: trio.CancelScope) -> None:
```
touilleMan, line 39: remove this method and instead directly use `self.connection_cancel_scope: Optional[trio.CancelScope]`

providing a setter is often a good idea for encapsulation, but here we already directly modify client_ctx attribute:
https://github.com/Scille/parsec-cloud/blob/9f42cfb69edfe34b92c8cc2b9f80707ac07e5571/parsec/backend/asgi/ws.py#L69

So it would be strange to have two way of modifying the attribute, and we don't really need encapsulation here given how simple this object is (and we are not in Java so there is no jail time for doing this anyway :trollface: )

touilleMan (changes requested):

touilleMan (approved):
Good job !

Now do we have a test for this ?

Volham22:
> Now do we have a test for this ?

No, we don't have a test for this yet. A proper test may be to send a dozen of commands without fetching their responses and assert if the connection has been closed by the server. Is there an already existing fixture that could help to achieve this in the testsuite?

touilleMan:
I think the important thing for having a good test on this is being able to mock the size of the memory channel used for the messages (it value is 100, which means our test would need to do 101 queries in a row...)

To do that:
- create a `AUTHENTICATED_CLIENT_CHANNEL_SIZE` constant in `parsec/backend/client_context.py` and use it instead of having 100 hardcoded where the memory channel is created
- in the test you then can use the `monkeypatch` fixture to mock the value of this constant (plot twist: constant is not that constant in Python 😄 ), doing somethinrg like `monkeypatch.setattr("parsec.backend.client_context.AUTHENTICATED_CLIENT_CHANNEL_SIZE", 1)` (the good thing is monkeypatch takes care of unmocking automatically)

You should put your test in `tests/backend/test_events.py` (calling it something like `test_events_close_connection_on_backpressure`

You should take inspiration from `test_events_subscribe`, yours should be pretty similar. Typically:
- `events_subscribe(alice_ws)` first
- then `authenticated_ping(bob_ws, "foo")`
- then again `authenticated_ping(bob_ws, "foo")`
- then wait for alice_ws to be automatically closed, to check this you can do `authenticated_ping(alice_ws, "foo")` in a while True loop wrapped in a `trio.fail_after(1)` (this is needed given we don't know when the connection will actually be closed due to it asynchronous nature, we know it will be "soon enough" so a 1s time out is plenty enough)

parsec/backend/events.py
```diff
@@ -87,8 +87,10 @@ def _on_pinged(
 
             try:
                 client_ctx.send_events_channel.send_nowait(EventsListenRepOkPinged(ping))
+                print("fine")
             except trio.WouldBlock:
-                client_ctx.logger.warning("dropping event (queue is full)")
+                print("block")
```
touilleMan, line 92: debug print to remove ?

tests/backend/test_events.py
```diff
@@ -146,4 +146,29 @@ async def test_events_listen_wait_cancelled(backend_asgi_app, alice_ws):
...
+async def test_events_close_connection_on_backpressure(
+    backend_asgi_app, alice, bob, monkeypatch, backend_authenticated_ws_factory
+):
+    # The channel has a queue of size 1, meaning it will be filled after a single command
+    monkeypatch.setattr("parsec.backend.client_context.AUTHENTICATED_CLIENT_CHANNEL_SIZE", 1)
+    # We need to manually initialize the websockets using the factory here to make
+    # sure the monkeypatch happens before the memory channel creation and thus
+    # give then a size of 1
+    async with backend_authenticated_ws_factory(backend_asgi_app, alice) as alice_ws:
+        async with backend_authenticated_ws_factory(backend_asgi_app, bob) as bob_ws:
```
touilleMan, line 159: Note you can directly use `bob_ws` fixture instead of creating it here. This is because `bob_ws` is used in the test only to send commands (and hence will never use it client context's channel)

tests/backend/test_events.py
```diff
@@ -146,4 +146,29 @@ async def test_events_listen_wait_cancelled(backend_asgi_app, alice_ws):
...
+            await events_subscribe(alice_ws)
+
+            await authenticated_ping(bob_ws, "foo")
+            await authenticated_ping(bob_ws, "foo")
+
+        # Alice should be disconnected at this point and exit the loop
+        with trio.fail_after(1):
```
touilleMan, line 166: This block is in the wrong place: given it is outside of `async with backend_authenticated_ws_factory(backend_asgi_app, alice) as alice_ws`, it means `alice_ws` has been closed when leaving the async with block (you can try this test without mocking `AUTHENTICATED_CLIENT_CHANNEL_SIZE` to see the issue)

So you do this check within the async with block because we check that the connection has been closed by the server

touilleMan (changes requested):

tests/backend/test_events.py
touilleMan: nevermind I misread that ^^

touilleMan (approved):

PR #3193, opened 2022-10-03
Oxidize regex

### LINKED ISSUE #3173: Avoid converting between python and rust regex
The following conversion between python regex and rust regex is expensive and called everytime the `prevent_sync_pattern` is passed to function in the pyo3 binding:

https://github.com/Scille/parsec-cloud/blob/ad806d243a0a30672f9e8c81e1c9bf175335f202/src/binding_utils.rs#L119-L127

Instead, it would be better to use a rust regex everywhere in the code base and expose it through the binding. Python regex are created only in two places:
- `parsec.core.logged_core._get_prevent_sync_pattern`
- `parsec.core.fs.storage.manifest_storage.ManifestStorage.get_prevent_sync_pattern` 

In this last case, the method should be removed once PR #2903 is merged.

Also, we had issues with the python implementation of `fnmatch.translate` so it might make sense to oxidize `fnmatch.translate + re.compile` in a single helper, maybe using a crate like [fnmatch-regex](https://crates.io/crates/fnmatch-regex).

Volham22:
This close #3173. We're using rust regex in the python world to avoid unnecessary conversion between python and rust regex (which isn't cheap).


oxidation/libparsec/crates/types/src/regex.rs
```diff
@@ -0,0 +1,35 @@
...
+
+use std::str::FromStr;
+
+use fancy_regex::{self, RegexBuilder};
+
+pub struct Regex(pub fancy_regex::Regex);
+
+// The fnmatch_regex crate does not escape the character '$'. It is a problem
+// for us. Since we need to match strings like `$RECYCLE.BIN` we need to escape
+// it anyway.
```
FirelightFlagboy, line 11: Prefer doc-comment

oxidation/libparsec/crates/types/src/regex.rs
```diff
@@ -0,0 +1,35 @@
...
+fn escape_dollarsign(string: &str) -> String {
+    str::replace(string, "$", "\\$")
+}
```
FirelightFlagboy, line 14: So this fonction will also escape the trailing `$` that represent the `end of line`. 
This look like an edge case that need to be handled

oxidation/libparsec/crates/types/src/regex.rs
```diff
@@ -0,0 +1,35 @@
+// Parsec Cloud (https://parsec.cloud) Copyright (c) BUSL-1.1 (eventually AGPL-3.0) 2016-present Scille SAS
+
+use std::str::FromStr;
+
+use fancy_regex::{self, RegexBuilder};
```
FirelightFlagboy, line 5: Since regex are now handled in the Rust code, do we have planned to remove this crate in favor of [`regex`](https://crates.io/crates/regex) ?

oxidation/libparsec/crates/types/src/regex.rs
```diff
@@ -0,0 +1,35 @@
+// Parsec Cloud (https://parsec.cloud) Copyright (c) BUSL-1.1 (eventually AGPL-3.0) 2016-present Scille SAS
```
FirelightFlagboy, line 1: I would like to see some test for this module

parsec/core/logged_core.py
```diff
@@ -90,15 +91,15 @@ def _get_prevent_sync_pattern(prevent_sync_pattern_path: Path) -> Optional[Patte
...
+                regex.append(str(Regex.from_pattern(line)))
         regex = "|".join(regex)
     except ValueError as exc:
         logger.warning(
             "Could not parse the file containing the filename patterns to ignore", exc_info=exc
         )
         return None
     try:
-        return re.compile(regex)
+        return Regex.from_regex_str(regex)
```
FirelightFlagboy, line 102: 1. Regex are compiled 2 time, could this be merged ?
2. We could use [`RegexSet`](https://docs.rs/regex/1.6.0/regex/struct.RegexSet.html)

FirelightFlagboy (changes requested):

oxidation/libparsec/crates/types/src/regex.rs
Volham22: Yes, I've planned to try to replace `fancy_regex` by `regex` since `fnmatch-regex` generates simpler regex. This PR is still in a WIP stage

FirelightFlagboy: The replacement will be done in this PR ?

oxidation/libparsec/crates/types/src/regex.rs
FirelightFlagboy: Oh, then
WDYT of if the function was named `escape_globing_pattern` ?

oxidation/libparsec/crates/types/src/regex.rs
Volham22: Yes I'm working on it

parsec/core/logged_core.py
Volham22: I don't know if this will change anything. A regex will be built when converting the `pattern` to an actual regex by the `glob_to_regex` function.
`RegexSet` expects an iterable of string-like object which means we will build `n` regexes and then an other bigger one to make the `RegexSet` object. Do you think `RegexSet` will be faster anyway?

FirelightFlagboy: I took a more in depth view of `RegexSet` and as you said, we cannot build the set from a list of already compiled regexes but from string.
For me using `RegexSet` would convey the idea that is *Check if a string match those regexes* better.
But I do not find it pretty to build the `RegexSet` from a set of already compiled regexes that are converted to string.
As for speed I don't know which would be better.

oxidation/libparsec/crates/types/src/regex.rs
```diff
@@ -0,0 +1,67 @@
...
+        assert!(r.0.is_match("foo"));
+        assert!(r.0.is_match("fooo"));
+        assert!(!r.0.is_match("fooooooo"));
+    }
+
+    #[test]
+    fn test_bad_regex_str_creation() {
+        let r = crate::regex::Regex::from_regex_str(r"fooo][?");
+        assert!(r.is_err());
+    }
```
FirelightFlagboy, line 66: Those tests are alike, so use `rstest` and convert those tests into `rstest::case`

src/local_manifest.rs
```diff
@@ -631,29 +632,27 @@ impl LocalFolderManifest {
...
     ) -> PyResult<Self> {
         let data = data
             .into_iter()
             .map(|(en, ei)| (en.0, ei.map(|ei| ei.0)))
             .collect();
-        let prevent_sync_pattern = py_to_rs_regex(prevent_sync_pattern)?;
         Ok(Self(self.0.clone().evolve_children_and_mark_updated(
             data,
-            &prevent_sync_pattern,
+            &prevent_sync_pattern.0 .0,
```
FirelightFlagboy, line 644: Wrong search and replace ?

```suggestion
            &prevent_sync_pattern.0,
```

oxidation/libparsec_python/Cargo.toml
```diff
@@ -15,7 +15,7 @@ libparsec = { version = "0.1.0", path = "../libparsec", features = ["mock-time"]
 
 uuid = { version = "1.1.2", features = ["serde", "v4", "fast-rng"] }
 chrono = {version = "0.4.22", features = ["serde"]}
-fancy-regex = "0.10.0"
+regex = "1.6.0"
```
FirelightFlagboy, line 18: Revert that change, we don't touch that crate as it's deprecated

tests/core/fs/workspacefs/test_sync_transactions.py
```diff
@@ -192,7 +192,7 @@ def test_merge_folder_manifests_with_concurrent_remote_change(
         .to_remote(author=my_device, timestamp=timestamp)
     )
 
-    prevent_sync_pattern = re.compile(r".*\.tmp\Z")
+    prevent_sync_pattern = Regex.from_regex_str(r".*\.tmp\z")
```
FirelightFlagboy, line 195: Why the regex expression have changed ?

FirelightFlagboy (changes requested):

tests/core/fs/workspacefs/test_sync_transactions.py
Volham22: Rust's regex does not recognize capital `Z` but only `z`. The `py_to_rs_regex` function does this for us but as we don't need it anymore I've changed the regex.

src/local_manifest.rs
Volham22: This function expects a `::regex::regex` struct to we need to access the inner object twice here. Otherwise we would get an error like this
```
expected `struct regex::Regex`, found struct `libparsec::libparsec_types::regex::Regex
```

Volham22: Note that the extra space between the `0` and the `.` is added by `rustfmt`. I guess it the rust's coding style

tests/core/fs/workspacefs/test_sync_transactions.py
FirelightFlagboy: ok

src/regex.rs
```diff
@@ -0,0 +1,45 @@
...
+            Err(PyValueError::new_err(format!(
+                "Invalid regex `{}`",
+                regex_str
+            )))
+        }
+    }
+
+    fn __str__(&self) -> String {
+        self.0 .0.to_string()
+    }
```
FirelightFlagboy, line 37: impl `Display` for `libparsec::Regex`

src/local_manifest.rs
FirelightFlagboy: I don't like the `.0 .0`, Why not passing the Wrapped regex (`libparsec::Regex`) to it ?

oxidation/libparsec/crates/types/src/regex.rs
```diff
@@ -26,40 +26,27 @@ impl PartialEq for Regex {
...
-    }
-
-    #[test]
+    use rstest::rstest;
+
+    #[rstest]
+    #[case(crate::regex::Regex::from_pattern("*.rs").expect("Should be valid"), "foo.rs", "bar.py", "bar.rs")]
+    #[case(crate::regex::Regex::from_pattern("*bar*").expect("Should be valid"), "foobar.rs", "dummy", "foobarrrrr.py")]
+    #[case(crate::regex::Regex::from_pattern("$myf$le.*").expect("Should be valid"), "$myf$le.txt", "bar", "$myf$le.rs")]
+    #[case(crate::regex::Regex::from_regex_str(r"fooo?$").expect("Should be valid"), "foo", "fo", "fooo")]
```
FirelightFlagboy, line 37: To simplify reading those case, move the compilation of the regex expr in the body of the test.

TimeEngineer (approved):

oxidation/libparsec/crates/types/src/regex.rs
```diff
@@ -0,0 +1,77 @@
...
+        #[case] bad_base: &str,
+        #[case] other_case: &str,
+    ) {
+        let regex = Regex::from_regex_str(regex_str).expect("Should be valid");
+        assert!(regex.0.is_match(valid_case));
+        assert!(!regex.0.is_match(bad_base));
+        assert!(regex.0.is_match(other_case));
+    }
+
+    #[rstest]
```
FirelightFlagboy, line 72:
```suggestion
    #[test]
```

Just the `test` macro would suffice

FirelightFlagboy (approved):

PR #3206, opened 2022-10-04
Implement list_available_devices with generator

TimeEngineer:


oxidation/bindings/electron/src/index.d.ts
```diff
@@ -10,17 +10,54 @@ export type Result<T, E = Error> =
...
 
-// HelloError
-export interface HelloErrorEmptySubject {
-    tag: "EmptySubject"
+export interface AvailableDevice {
+    key_file_path: string;
+    organization_id: string;
+    device_id: string;
+    human_handle?: string | null;
+    device_label?: string | null;
```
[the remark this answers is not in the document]
TimeEngineer: `?` means it can be undefined which is different of `null`
I don't know what the front want, I would prefer the `undefined`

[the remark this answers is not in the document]
FirelightFlagboy: I don't know either but they (the persons that would work with `TypeScript`) would be required by the language to check for both the `undefined` & `null` with the current definition.

@NicoTuxx & @Landeers WDYT ?

oxidation/bindings/generator/api.py
```diff
@@ -25,35 +25,63 @@ class Ref(Generic[TypeVar("REFERENCED")]):  # noqa
 
 # A type that should be converted from/into string
 class StrBasedType:
-    pass
+    ...
```
touilleMan, line 28: nop, `StrBasedType` should not have an ellipsis: ellipsis is to indicate the body of the class/function is missing where `StrBasedType` is entirely defined here (it is a class marker)

Of course this is only cosmetic given the script doesn't look for the body, but it's more readable to stick with `pass` and only use `...` for binded functions.

oxidation/bindings/generator/api.py
```diff
@@ -25,35 +25,63 @@ class Ref(Generic[TypeVar("REFERENCED")]):  # noqa
...
 
 
 class OrganizationID(StrBasedType):
-    pass
+    ...
```
touilleMan, line 32: same as above, use pass for everything except function definitions

oxidation/bindings/generator/api.py
```diff
@@ -25,35 +25,63 @@ class Ref(Generic[TypeVar("REFERENCED")]):  # noqa
...
+    class Recovery:
+        ...
 
-class PathBuf(StrBasedType):
-    pass
+    class Smartcard:
+        ...
+
+
+class LocalDeviceError(Variant):
```
touilleMan, line 62: Do we need all those errors types ?

I even wonder if we need error at all for `list_available_devices` 🤔 

The idea is to provide a convenient API for the GUI to use, hence providing many errors is often the mark we are exposing internal details and making the API more complicated than it should be.

On top of that the API between GUI and libparsec don't have any backward compatibility requirements, so we can start simple and add errors later on when the need come ^^

touilleMan: @Max-7 @NicoTuxx Any suggestion here ?

touilleMan (changes requested):

oxidation/libparsec/crates/client_types/src/error.rs
```diff
@@ -1,18 +1,18 @@
 // Parsec Cloud (https://parsec.cloud) Copyright (c) BUSL-1.1 (eventually AGPL-3.0) 2016-present Scille SAS
 
-use std::path::PathBuf;
+use crate::FSPath;
```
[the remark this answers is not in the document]
TimeEngineer: Done

oxidation/bindings/generator/api.py
TimeEngineer: Done

oxidation/bindings/generator/api.py
TimeEngineer: Done

oxidation/bindings/generator/api.py
TimeEngineer: Done

oxidation/libparsec/crates/client_types/src/path.rs
```diff
@@ -0,0 +1,69 @@
...
+    fmt::Debug,
+    path::{Path, PathBuf},
+};
+
+#[derive(Debug, Clone, PartialEq, Eq, PartialOrd, Ord, Hash, Serialize)]
+pub struct StrPath(PathBuf);
+
+impl std::str::FromStr for StrPath {
+    // Type Err require display for generator
+    type Err = &'static str;
```
FirelightFlagboy, line 14: use [`Infallible`](https://doc.rust-lang.org/std/convert/enum.Infallible.html)

oxidation/libparsec/crates/client_types/src/path.rs
```diff
@@ -0,0 +1,69 @@
...
+
+    fn from_str(s: &str) -> Result<Self, Self::Err> {
+        Ok(s.into())
+    }
+}
+
+impl AsRef<str> for StrPath {
```
FirelightFlagboy, line 21: This is against the naming convention.
https://rust-lang.github.io/api-guidelines/naming.html#ad-hoc-conversions-follow-as_-to_-into_-conventions-c-conv

As taking the reference here is not free but expensive

oxidation/libparsec/crates/types/src/time.rs
```diff
@@ -177,7 +177,10 @@ mod time_provider {
         }
 
         pub async fn sleep(&self, time: std::time::Duration) {
-            libparsec_platform_async::native::sleep(time).await
+            #[cfg(not(target_arch = "wasm32"))]
+            libparsec_platform_async::native::sleep(time).await;
+            #[cfg(target_arch = "wasm32")]
+            libparsec_platform_async::wasm32::sleep(time).await;
```
FirelightFlagboy, line 183:
```suggestion
            libparsec_platform_async::platform::sleep(time).await;
```

FirelightFlagboy (changes requested):

oxidation/libparsec/crates/client_types/src/path.rs
TimeEngineer: Yes I know, but the generator need this trait for Path

oxidation/libparsec/crates/types/src/time.rs
TimeEngineer: Good catch

oxidation/libparsec/crates/client_types/src/path.rs
TimeEngineer:
```Shell
error[E0277]: the trait bound `std::convert::Infallible: std::convert::AsRef<str>` is not satisfied
   --> oxidation/bindings/electron/src/meths.rs:21:37
    |
21  |             Err(err) => return (cx).throw_type_error(err),
    |                                     ^^^^^^^^^^^^^^^^ the trait `std::convert::AsRef<str>` is not implemented for `std::convert::Infallible`
    |
note: required by a bound in `neon::context::Context::throw_type_error`
   --> /home/engineer/.cargo/registry/src/github.com-1ecc6299db9ec823/neon-0.10.1/src/context/mod.rs:570:28
    |
570 |     fn throw_type_error<S: AsRef<str>, T>(&mut self, msg: S) -> NeonResult<T> {
    |                            ^^^^^^^^^^ required by this bound in `neon::context::Context::throw_type_error`
    ```

File: oxidation/bindings/electron/src/index.d.ts (hunk 1 above)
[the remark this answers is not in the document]
Comment (TimeEngineer) [reply oxidation/bindings/electron/src/index.d.ts]: `undefined` is removed

File: oxidation/libparsec/crates/client_types/src/path.rs (hunk 6 above)
Comment (FirelightFlagboy) [reply oxidation/libparsec/crates/client_types/src/path.rs:14]: 🤬 fine

Comment (FirelightFlagboy) [reply oxidation/libparsec/crates/client_types/src/path.rs:14]: Can impl `FromStr` only on `wasm32`

File: oxidation/libparsec/crates/client_types/src/path.rs (hunk 7 above)
Comment (FirelightFlagboy) [reply oxidation/libparsec/crates/client_types/src/path.rs:21]: Could this implementation be lock only on `wasm32` ?

Comment (TimeEngineer) [reply oxidation/libparsec/crates/client_types/src/path.rs:24]: Done

File: oxidation/libparsec/crates/client_types/src/path.rs (hunk 6 above)
Comment (FirelightFlagboy) [reply oxidation/libparsec/crates/client_types/src/path.rs:16]:
```rust
    #[cfg(wasm32)]
    // Type Err require display for generator
    type Err = &'static str;
    #[cfg(not(wasm32))]
    type Err = Infallible;
```
Something like that then ?

TimeEngineer: The error came from electron (native), but it seems overkill no ? Also `StrPath` is used only for the client

FirelightFlagboy: Ok then

FirelightFlagboy (approved):

FirelightFlagboy (approved):

oxidation/libparsec/crates/types/src/manifest.rs
```diff
@@ -104,27 +105,27 @@ macro_rules! impl_manifest_dump_load {
                 expected_timestamp: DateTime,
                 expected_id: Option<EntryID>,
                 expected_version: Option<u32>,
-            ) -> Result<(), DataError> {
+            ) -> DataResult<()> {
                 if self.author != *expected_author {
-                    Err(DataError::UnexpectedAuthor {
+                    Err(Box::new(DataError::UnexpectedAuthor {
```
FirelightFlagboy, line 110: No way to not box the error ?

oxidation/libparsec/crates/types/src/error.rs
```diff
@@ -40,10 +40,10 @@ pub enum DataError {
     UnexpectedVersion { expected: u32, got: u32 },
 }
 
-pub type DataResult<T> = Result<T, DataError>;
+pub type DataResult<T> = Result<T, Box<DataError>>;
```
FirelightFlagboy, line 43: What's the choice to box the error ?

TimeEngineer: The size of the error, due to `UnexpectedAuthorError`

FirelightFlagboy: How 2 `DeviceID` be too big ?

FirelightFlagboy: Can't we just box the incriminated variant ?

TimeEngineer: Is it better ?

FirelightFlagboy: For me yes, it would require to do the heap allocation only for this variant

TimeEngineer: Performance isn't needed when you catch errors, we should priorize the easiest way

FirelightFlagboy: ok then

ADcorpo:
Looks good to me ! However this is not blocked by a lack of approval. We can't merge because there is an ongoing change request by @touilleMan apparently

Side note (unimportant): HumanHandle is implemented as a simple wrapper over a string, does that mean we will systematically use only one of this struct's fields ? This is something I had to implement directly in the template in my attempt

ADcorpo:
this fixes #3760

oxidation/libparsec/crates/types/src/error.rs
touilleMan: I don't understand the need for Boxing error here :/

As @TimeEngineer said, "we should priorize the easiest way": so no Box at all is the best.

My understanding is boxing error is useful to return different types that are checked at runtime:
```rust
type Result<T> = std::result::Result<T, Box<dyn error::Error>>;
```

However this is not the case here, so what are we achieving that we can't without boxing here ?

touilleMan: after some more digging:

- `String` [is literally a `Vec<u8>`](https://doc.rust-lang.org/src/alloc/string.rs.html#367-369)
- `Vec<T>` is composed of three words (number of elements in the vector, pointer of the buffer storing the elements and actual size of the buffer)
- So we endup on x86_64 with `Vec<T>`/`String` made of... 24bytes 😱  (I've checked using `std::mem::size_of` on Windows 10 64bits)
- Hence `DeviceID` weight 72bytes now that is contains 3 `String`... that quite heavy !

on the other hand, a box is 8 or 16bytes (depending if the [boxed element is a dynamically sized type](https://doc.rust-lang.org/reference/dynamically-sized-types.html))

So on paper it would be more efficient to use box...
image


We don't have actual proof this is the case.
- Using a box means we have to do a heap allocation (which is cheap, but less cheap than just using the stack).
- Rust relies heavily on compiler optimization & in-lining (hence the perf difference between dev and release profiles !), so it's very likely most function returning a `Result<T, MyBigAssErrorType>` gets optimized out and the result is only written once in it final location (hence making it similar in perf with the single write needed when initializing the box... less the heap allocation and pointer redirection during access ^^)

So in the end this use of Box [smells heavily like premature optimization](https://xkcd.com/1691/). I'd rather have simpler to use error (and not having to constantly ponder "do we need box for this new error type ?").

touilleMan: edit: this has been changed due to [clippy codesmell](https://rust-lang.github.io/rust-clippy/master/index.html#result_large_err), I'm all okay to follow clippy Rules so carry on ! ^^

oxidation/libparsec/crates/types/src/id.rs
```diff
@@ -314,14 +335,31 @@ impl_debug_from_display!(HumanHandle);
...
+    }
+}
+
+impl AsRef<str> for HumanHandle {
+    fn as_ref(&self) -> &str {
+        &self.display
+    }
+}
+
+impl FromStr for HumanHandle {
```
touilleMan, line 348: Better remove this:
- this is not used in the code
- the start/stop computation is fragile (it should be thoughtfully tested)
- we should never need to turn `foo <foo@example.com>` back to `HumanHandle`

edit: this is needed due to the bindings currently the HumanHandle object is converted back and forth into a string to be passed to javascript :(
(I'll have a look at improving this...)

touilleMan (approved):
lgtm ;-)

PR #3209, opened 2022-10-05
Oxidize trustchain error

### LINKED ISSUE #3171: Oxidize exception `TrustchainError`
# TODOs

- [ ] Oxidize the python exception `TrustchainError` that is already defined in Rust
    > Since that exception is not created in the python code base you don't need to provide a way for python to create it

https://pyo3.rs/latest/function/error_handling.html#foreign-rust-error-types

Volham22:
This close #3171


src/trustchain.rs
```diff
@@ -122,3 +126,239 @@ impl TrustchainContext {
...
+#[derive(PartialEq)]
+pub(crate) struct TrustchainError(pub(crate) libparsec::core::TrustchainError);
+
+// This object is only here to wrap our `TrustchainError` in a python exception
+// object. Without this object we can't raise or except `TrustchainError` in
+// python code
+create_exception!(_parsec, TrustchainErrorException, PyException);
+
+#[pymethods]
+impl TrustchainError {
```
FirelightFlagboy, line 140: Do we really need all of those constructor ? For me the exception was not created in the python code.

> Since that exception is not created in the python code base you don't need to provide a way for python to create it

src/trustchain.rs
```diff
@@ -122,3 +126,239 @@ impl TrustchainContext {
...
+            Err(PyAttributeError::new_err("No such attribute `got`"))
+        }
+    }
+}
+
+gen_proto!(TrustchainError, __richcmp__, eq);
+gen_proto!(TrustchainError, __str__); // Needed for python's exceptions
+gen_proto!(TrustchainError, __repr__);
+
+impl From<TrustchainError> for PyErr {
```
FirelightFlagboy, line 332: Don't we want to use `TrustchainErrorException` here ?

FirelightFlagboy (changes requested):
I don't see the exception `TrustchainError` being deleted in the python code, I'm missing something ?

src/trustchain.rs
Volham22: This impl isn't used at all. We should just remove it

Volham22:
> I don't see the exception `TrustchainError` being deleted in the python code, I'm missing something ?

Done.

src/trustchain.rs
```diff
@@ -122,3 +126,129 @@ impl TrustchainContext {
         ))
     }
 }
+
+#[pyclass]
+#[derive(PartialEq)]
```
TimeEngineer, line 131: Remove `PartialEq`

src/binding_utils.rs
```diff
@@ -135,6 +136,11 @@ pub fn rs_to_py_regex<'py>(py: Python<'py>, regex: &Regex) -> PyResult<&'py PyAn
     re.call_method1("compile", args)
 }
 
+pub fn to_trustchain_error_exception(err: libparsec::core::TrustchainError) -> PyErr {
+    let python_error = TrustchainError(err);
+    TrustchainErrorException::new_err(python_error)
+}
```
FirelightFlagboy, line 142: Using this code block into the trait `impl From<libparsec::core::TrustchainError> for PyErr` would simplify the error handling part

as we don't need to use `.map_err(to_trustchain_error_exception)` before unwrapping with `?` when we use rust code the return `libparsec::core::TrustchainError`

src/trustchain.rs
```diff
@@ -122,3 +126,129 @@ impl TrustchainContext {
...
+
+// This object is only here to wrap our `TrustchainError` in a python exception
+// object. Without this object we can't raise or except `TrustchainError` in
+// python code
+create_exception!(_parsec, TrustchainErrorException, PyException);
+
+#[pymethods]
+impl TrustchainError {
+    #[getter]
+    fn path(&self) -> PyResult<String> {
```
TimeEngineer, line 142: Can it be a `&str` instead of `String` ?

FirelightFlagboy (commented):
One more thing

src/trustchain.rs
```diff
@@ -122,3 +126,129 @@ impl TrustchainContext {
...
+        } else {
+            Err(PyAttributeError::new_err("No such attribute `exc`"))
+        }
+    }
+
+    #[getter]
+    fn user_id(&self) -> PyResult<UserID> {
+        match &self.0 {
+            libparsec::core::TrustchainError::InvalidSelfSignedUserCertificate { user_id } => {
+                Ok(ids::UserID(user_id.clone()))
```
TimeEngineer, line 177: Why do you keep `ids::` ?

src/trustchain.rs
```diff
@@ -1,16 +1,20 @@
 // Parsec Cloud (https://parsec.cloud) Copyright (c) BUSL-1.1 (eventually AGPL-3.0) 2016-present Scille SAS
 
-use pyo3::{import_exception, prelude::*, types::PyTuple};
-
 use crate::{
     api_crypto::VerifyKey,
+    binding_utils::{gen_proto, to_trustchain_error_exception},
     certif::{DeviceCertificate, RevokedUserCertificate, UserCertificate},
-    ids::{DeviceID, UserID},
+    ids::{self, DeviceID, UserID},
```
TimeEngineer, line 7: You can remove `self` then

src/binding_utils.rs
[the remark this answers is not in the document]
FirelightFlagboy: see https://github.com/Scille/parsec-cloud/pull/3209#discussion_r988014366

Volham22: I agree but shouldn't we avoid to put python related code in `libparsec::core` ? We can't implement `From<libparsec::core::TrustchainError>` in `src/trustchain.rs`

TimeEngineer: Maybe add a file `error.rs` in `src/` ?

Volham22: It will not work we would get an error like this `only traits defined in the current crate can be implemented for types defined outside of the crate` by the compiler

Volham22: I've implemented `From<TrustchainError>` trait for `PyErr` this way we can just do a `.map_err(TrustchainError)?`

FirelightFlagboy: I was thinking of something like that 

```rust
impl From<libparsec::TrustchainError> for TrustchainError {}
impl From<TrustchainError> for PyErr {}
```

src/trustchain.rs
```diff
@@ -122,3 +126,126 @@ impl TrustchainContext {
...
+    }
+
+    #[getter]
+    fn verified_timestamp(&self) -> PyResult<DateTime> {
+        if let libparsec::core::TrustchainError::SignaturePosteriorUserRevocation {
+            verified_timestamp,
+            ..
+        } = &self.0
+        {
+            Ok(time::DateTime(*verified_timestamp))
```
TimeEngineer, line 201: You can do the same for `time`

src/trustchain.rs
```diff
@@ -122,3 +126,126 @@ impl TrustchainContext {
...
+            ))
+        }
+    }
+
+    #[getter]
+    fn user_timestamp(&self) -> PyResult<DateTime> {
+        if let libparsec::core::TrustchainError::SignaturePosteriorUserRevocation {
+            user_timestamp,
+            ..
+        } = &self.0
```
TimeEngineer, line 214: Remove borrow and remove deference

src/binding_utils.rs
Volham22: Does it matter since we'll need to do at least a `map_err` to be able to use `?`?

FirelightFlagboy (approved):

FirelightFlagboy (approved):

TimeEngineer (approved):

Issue #3212, opened 2022-10-06
[🐛 | Bug]: Login fails on master


Max-7:
### Version

master

### What platforms you were running our software when the bug happened ?

Windows

### What happened ?

Login fails when trying to log to an existing organization.

### Relevant output

2022-10-06T06:56:03.699236Z [error    ] Unhandled error
Traceback (most recent call last):
  File "E:\projects\parsec-cloud\parsec\core\gui\trio_jobs.py", line 97, in __call__
    result = await self._fn(*self._args, **self._kwargs)
  File "E:\projects\parsec-cloud\parsec\core\gui\instance_widget.py", line 47, in _do_run_core
    async with logged_core_factory(config=config, device=device, event_bus=None) as core:
  File "C:\Users\maxim\AppData\Local\Programs\Python\Python39\lib\contextlib.py", line 181, in __aenter__
    return await self.gen.__anext__()
  File "E:\projects\parsec-cloud\parsec\core\logged_core.py", line 412, in logged_core_factory
    async with UserFS.run(
  File "C:\Users\maxim\AppData\Local\Programs\Python\Python39\lib\contextlib.py", line 181, in __aenter__
    return await self.gen.__anext__()
  File "E:\projects\parsec-cloud\parsec\core\fs\userfs\userfs.py", line 331, in run
    await self._load_workspace(workspace_entry.id)
  File "E:\projects\parsec-cloud\parsec\core\fs\userfs\userfs.py", line 438, in _load_workspace
    workspace = await self._instantiate_workspace(workspace_id)
  File "E:\projects\parsec-cloud\parsec\core\fs\userfs\userfs.py", line 408, in _instantiate_workspace
    local_storage = await self._workspace_storage_nursery.start(workspace_storage_task)
  File "E:\projects\parsec-cloud\parsec\service_nursery.py", line 183, in start
    return await type(nursery).start(nursery, wrap_child, name=name or async_fn)
  File "E:\projects\parsec-cloud\venv\lib\site-packages\trio\_core\_run.py", line 1042, in start
    task._eventual_parent_nursery = self
  File "E:\projects\parsec-cloud\venv\lib\site-packages\trio\_core\_run.py", line 813, in __aexit__
    raise combined_error_from_nursery
  File "E:\projects\parsec-cloud\parsec\service_nursery.py", line 181, in wrap_child
    await async_fn(*args, task_status=task_status)
[...]

PR #3214, opened 2022-10-06
Add test to check regex compat

Volham22:
When a user has an older cache some regex contains escapes sequences like `\Z` which are not supported by rust `regex` crate. Note that `fancy_regex` has the same issue here.

This is a proposal to fix the issue (#3212) by replacing the unsupported escapes sequences. Although it works, there may be better ways to handle this issue.

### DIFF
--- a/oxidation/libparsec/crates/types/src/regex.rs
+++ b/oxidation/libparsec/crates/types/src/regex.rs
@@ -20,6 +20,10 @@ impl Regex {
     pub fn from_regex_str(regex_str: &str) -> Result<Self, regex::Error> {
         Ok(Regex(regex::Regex::new(regex_str)?))
     }
+
+    pub fn is_match(&self, string: &str) -> bool {
+        self.0.is_match(string)
+    }
 }
 
 impl AsRef<str> for Regex {
@@ -56,9 +60,9 @@ mod tests {
         #[case] other_case: &str,
     ) {
         let regex = Regex::from_pattern(pattern_str).expect("Should be valid");
-        assert!(regex.0.is_match(valid_case));
-        assert!(!regex.0.is_match(bad_base));
-        assert!(regex.0.is_match(other_case));
+        assert!(regex.is_match(valid_case));
+        assert!(!regex.is_match(bad_base));
+        assert!(regex.is_match(other_case));
     }
 
     #[rstest]
@@ -70,9 +74,9 @@ mod tests {
         #[case] other_case: &str,
     ) {
         let regex = Regex::from_regex_str(regex_str).expect("Should be valid");
-        assert!(regex.0.is_match(valid_case));
-        assert!(!regex.0.is_match(bad_base));
-        assert!(regex.0.is_match(other_case));
+        assert!(regex.is_match(valid_case));
+        assert!(!regex.is_match(bad_base));
+        assert!(regex.is_match(other_case));
     }
 
     #[test]
--- a/src/regex.rs
+++ b/src/regex.rs
@@ -14,18 +14,110 @@ impl Regex {
     fn from_pattern(_cls: &PyType, pattern: &str) -> PyResult<Self> {
         libparsec::types::regex::Regex::from_pattern(pattern)
             .map(Regex)
-            .map_err(|_| PyValueError::new_err(format!("Invalid pattern `{}`", pattern)))
+            .map_err(|err| {
+                PyValueError::new_err(format!(
+                    "Failed to convert pattern to regex: `{}` with the following error `{}`",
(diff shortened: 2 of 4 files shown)


oxidation/libparsec/crates/types/src/regex.rs
```diff
@@ -11,14 +11,23 @@ fn escape_globing_pattern(string: &str) -> String {
     str::replace(string, "$", "\\$")
 }
 
+/// This function remove the `\Z` and `\\` sequences that are unsupported by the
+/// rust's regex. These esacapes can still be in the regex strings if the user
```
FirelightFlagboy, line 15:
```suggestion
/// rust's regex. These escapes can still be in the regex strings if the user
```

FirelightFlagboy (changes requested):

Volham22:
I had to move back to `fancy_regex` to ensure backward compatibility with old regex (generated by python) that might be cached.

FirelightFlagboy (approved):

oxidation/libparsec/crates/types/src/id.rs
```diff
@@ -62,7 +62,7 @@ macro_rules! new_string_based_id_type {
                         Regex::new($pattern).unwrap_or_else(|_| unreachable!());
                 }
                 // ID must respect regex AND be contained within $bytes_size bytes
-                if PATTERN.is_match(&id) && id.len() <= $bytes_size {
+                if PATTERN.is_match(&id).unwrap() && id.len() <= $bytes_size {
```
FirelightFlagboy, line 65: Replace with `unwrap_or_default()`

oxidation/libparsec/crates/types/src/regex.rs
```diff
@@ -11,14 +12,28 @@ fn escape_globing_pattern(string: &str) -> String {
...
+    pub fn from_regex_str(regex_str: &str) -> Result<Self, fancy_regex::Error> {
+        Ok(Regex(fancy_regex::Regex::new(
+            &remove_unsupported_escapes(regex_str),
+        )?))
     }
 
-    pub fn from_regex_str(regex_str: &str) -> Result<Self, regex::Error> {
-        Ok(Regex(regex::Regex::new(regex_str)?))
+    pub fn is_match(&self, string: &str) -> bool {
+        self.0.is_match(string).expect("Match failed")
```
FirelightFlagboy, line 36:
```suggestion
        self.0.is_match(string).unwrap_or_default()
```
Default to `false` but we may want to return the error instead :thinking:

Volham22: We'll drop `fancy_regex` so this result will no longer be there

vxgmichel (approved):

FirelightFlagboy (approved):

src/regex.rs
```diff
@@ -14,18 +14,87 @@ impl Regex {
...
+
+#[cfg(test)]
+mod test {
+    use pyo3::{
+        exceptions::PyValueError,
+        types::{PyModule, PyTuple},
+        Py, PyAny, PyResult, Python,
+    };
+
+    const NEW_REGEX_STR: &str = r"^.*\.tmp$|^.*\.temp$|^.*\.swp$|^.*~$|^\.fuse_hidden.*$|^\.directory$|^\.Trash-.*$|^\.nfs.*$|^\.goutputstream-.*$|^\.DS_Store$|^\.AppleDouble$|^\.LSOverride$|^\._.*$|^\.DocumentRevisions-V100$|^\.Spotlight-V100$|^\.TemporaryItems$|^\.Trashes$|^\.VolumeIcon\.icns$|^\.com\.apple\.timemachine\.donotpresent$|^\.AppleDB$|^\.AppleDesktop$|^Network Trash Folder$|^Temporary Items$|^\.apdisk$|^\.fseventsd$|^.*\..*\.sb-.*-.*$|^Thumbs\.db$|^Thumbs\.db:encryptable$|^ehthumbs\.db$|^ehthumbs_vista\.db$|^.*\.stackdump$|^[Dd]esktop\.ini$|^\$RECYCLE\.BIN$|^.*\.lnk$|^\.~.*$|^~\$.*$";
```
vxgmichel, line 51: Maybe split it over several lines?

vxgmichel: @TimeEngineer How would you do that?

FirelightFlagboy: with the `concat!` macro ;)

vxgmichel: Or maybe regenerate it from the code instead of hard coding it?

FirelightFlagboy (approved):

PR #3219, opened 2022-10-06
Oxidation replace message

TimeEngineer:


src/message.rs
```diff
@@ -0,0 +1,337 @@
...
+    ) -> PyResult<PyObject> {
+        let msg = libparsec::types::MessageContent::decrypt_verify_and_load_for(
+            ciphered,
+            &recipient_privkey.0,
+            &author_verify_key.0,
+            &expected_author.0,
+            expected_timestamp.0,
+        );
+
+        Ok(unsafe {
```
FirelightFlagboy, line 76: No way to remove the use of `unsafe` ?

TimeEngineer: Nope if we work with subclass

FirelightFlagboy: Add a comment to explain why this block is safe, I plan to enable [`clippy::undocumented_unsafe_blocks`](https://rust-lang.github.io/rust-clippy/master/#undocumented_unsafe_blocks).
Can you also limit the size of the unsafe block to code that required ?

i.e.: only put `into_new_object` & `from_owned_ptr` into the `unsafe` block and keep everything else outside

src/message.rs
```diff
@@ -0,0 +1,347 @@
...
+                let initializer = PyClassInitializer::from((PingMessageContent, Self(msg)));
+                // This block is safe because `initializer` has the correct type
+                unsafe {
+                    let ptr =
+                        initializer.into_new_object(py, PingMessageContent::type_object_raw(py))?;
+                    PyObject::from_owned_ptr(py, ptr)
+                }
+            }
+            Err(err) => return Err(PyValueError::new_err(err)),
+        })
```
FirelightFlagboy, line 119: Is it worth to use a macro here, to factorize the code ?

src/message.rs
FirelightFlagboy: :+1:

FirelightFlagboy (approved):

FirelightFlagboy (approved):

FirelightFlagboy (approved):

FirelightFlagboy (approved):

FirelightFlagboy (approved):

PR #3221, opened 2022-10-06
Replace UserProfile

TimeEngineer:

PR #3222, opened 2022-10-06
Oxidation replace pki

TimeEngineer:
depends on #3221


parsec/serde/fields.py
```diff
@@ -20,15 +20,22 @@
...
+    PkiEnrollmentSubmitPayload as _PkiEnrollmentSubmitPayload,
+)
+from parsec.crypto import (
+    HashDigest as _HashDigest,
+    PrivateKey as _PrivateKey,
+    PublicKey as _PublicKey,
+    SecretKey as _SecretKey,
+    SigningKey as _SigningKey,
+    VerifyKey as _VerifyKey,
+)
```
FirelightFlagboy, line 34: Aren't those from `parsec._parsec` ?

parsec/serde/fields.py
```diff
@@ -512,3 +519,23 @@ def _deserialize(
 
 
 SequesterEncryptionKeyDer = SequesterEncryptionKeyDerField
+
+
+class PkiEnrollmentSubmitPayloadField(Field[_PkiEnrollmentSubmitPayload]):
+    def _serialize(
+        self, value: _PkiEnrollmentSubmitPayload | None, attr: Any, obj: Any
```
Volham22, line 526: Isn't `attr` a `str`?
```
Help on function _serialize in module marshmallow.fields:

_serialize(self, value, attr, obj)
    Serializes ``value`` to a basic Python datatype. Noop by default.
    Concrete :class:`Field` classes should implement this method.

    Example: ::

        class TitleCase(Field):
            def _serialize(self, value, attr, obj):
                if not value:
                    return ''
                return unicode(value).title()

    :param value: The value to be serialized.
    :param str attr: The attribute or key on the object to be serialized.
    :param object obj: The object the value was pulled from.
    :raise ValidationError: In case of formatting or validation failure.
    :return: The serialized value
```

Volham22 (changes requested):

Volham22 (approved):

Max-7 (changes requested):
No changes needed (for the moment), it's just to block the merge until additional testing has been done.

Max-7 (approved):
LGTM

PR #3237, opened 2022-10-10
Binding invitation deleted reason

### LINKED ISSUE #3232: Fully Oxidize `InvitationDeletedReason`
The enum `InvitationDeletedReason` is still present in the python side [here](https://github.com/Scille/parsec-cloud/blob/9550dca35c380ac0b42487dc02e3620e5dfeb748/parsec/api/protocol/invite.py#L111). As it may be confusing we should remove it and only use the binded rust enum.

Volham22:
This fix #3232


touilleMan (changes requested):

Volham22:
I've also replaced all python enum usage by the new one

FirelightFlagboy (approved):

FirelightFlagboy (approved):

FirelightFlagboy (approved):

PR #3247, opened 2022-10-11
Remove UserProfile Default because not used anymore

TimeEngineer:


oxidation/libparsec/crates/protocol/tests/test_invite.rs
```diff
@@ -66,7 +66,6 @@ fn serde_invite_new_req(#[case] raw_expected: (&[u8], authenticated_cmds::AnyCmd
     (
         // Generated from Rust implementation (Parsec v2.12.1+dev)
         // Content:
-        //   email_sent: None
```
vxgmichel, line 69: Oh I missed this one, good catch :+1:

vxgmichel (approved):

PR #3283, opened 2022-10-13
mypy: Added missing types for `parsec.core.types`

### LINKED ISSUE #3284: Enable `mypy` on `core.types`

Volham22:
This adds mypy types for module `parsec.core.types`. This close #3284.

### DIFF
--- a/parsec/core/types/pki.py
+++ b/parsec/core/types/pki.py
@@ -1,7 +1,7 @@
 # Parsec Cloud (https://parsec.cloud) Copyright (c) AGPL-3.0 2016-present Scille SAS
 from __future__ import annotations
 
-from typing import Iterable, List, Optional, Dict
+from typing import Iterable, List, Optional, Dict, Union
 from pathlib import Path
 from uuid import UUID
 from parsec._parsec import DateTime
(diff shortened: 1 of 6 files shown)


FirelightFlagboy (approved):

parsec/core/types/pki.py
```diff
@@ -91,9 +91,22 @@ class SCHEMA_CLS(BaseSchema):
...
+                str,
+                Union[
+                    X509Certificate,
+                    BackendPkiEnrollmentAddr,
+                    DateTime,
+                    UUID,
+                    PkiEnrollmentSubmitPayload,
+                    bytes,
+                ],
+            ],
```
vxgmichel, line 106: Maybe `Dict[str, object]` is enough since you're using a `type: ignore` afterwards?

In general I think it makes sense to use `Dict[str, object]` to type those kind of arbitrary data structure, and then do something like this if necessary:
```python
x509_certificate  = typing.cast(X509Certificate, data["x509_certificate"])
```

vxgmichel (changes requested):

FirelightFlagboy (approved):

FirelightFlagboy (approved):

parsec/core/types/pki.py
```diff
@@ -35,17 +35,17 @@ class SCHEMA_CLS(BaseSchema):
         certificate_id = fields.String(required=True, allow_none=True)
 
         @post_load
-        def make_obj(self, data):
+        def make_obj(self, data: Dict[str, Union[str, Dict[str, str], bytes]]) -> X509Certificate:
             data.pop("type", None)
-            return X509Certificate(**data)
+            return X509Certificate(**data)  # type: ignore
```
vxgmichel, line 40: `ignore[arg-type]`?

vxgmichel (changes requested):

FirelightFlagboy (approved):

PR #3287, opened 2022-10-13
Use future annotations everywhere in parsec and tests directories

vxgmichel:
The following commands have been used:
```
find parsec -name "*.py" -exec sed -i "2i from __future__ import annotations" {} +
find tests -name "*.py" -exec sed -i "2i from __future__ import annotations" {} +
```

This allows the use of the following typing syntax:
- `A | B` (instead of `Union[A, B]`)
- `A | None`  (instead of `Optional[A]`) 
- `tuple[A]/list[A]/set[A]` (instead of `Tuple[A]/List[A]/Set[A]`)


FirelightFlagboy (approved):

touilleMan:
image

PR #3290, opened 2022-10-13
Rework jobs `CI is happy` to be always run

FirelightFlagboy:


.github/workflows/ci.yml
```diff
@@ -35,7 +35,22 @@ jobs:
       - test-web-app
       - test-wasm
     runs-on: ubuntu-latest
+    if: always()
     steps:
+      - name: Debug needs
```
touilleMan, line 40: What's that ?

(don't we risk of leaking sensitive env vars with a bad configuration on this ?)

.github/workflows/ci.yml
```diff
@@ -35,7 +35,22 @@ jobs:
...
+        env:
+          NEEDS: ${{ toJSON(needs) }}
+        run: printenv NEEDS
+
+      - run: |
+          echo "Oh No, we have jobs that have failed/cancelled/skipped :("
+          exit 42
+        if: >-
+          contains(needs.*.result, 'failure')
+          || contains(needs.*.result, 'skipped')
```
touilleMan, line 50: When we don't modify rust, the rust job is still executed (but does nothing). So this check on skipped is not affected, right ?

.github/workflows/ci.yml
FirelightFlagboy: It's a task that print the `needs` context more info [here](https://docs.github.com/en/actions/learn-github-actions/contexts#needs-context).

tl,dr: The needs context contains only the final state a jobs (if it fail or not) it's output (actually none of our jobs are configuring outputs variable)

.github/workflows/ci.yml
FirelightFlagboy: Yes, you're right. The jobs is always executed but will performed a greatly reduced amount of work

.github/workflows/ci.yml
FirelightFlagboy: done

PR #3292, opened 2022-10-13
Add mypy to backend cli

TimeEngineer:


parsec/backend/cli/sequester.py
```diff
@@ -1,18 +1,19 @@
...
-from base64 import b64encode, b64decode
-from async_generator import asynccontextmanager
-import textwrap
 import attr
 import click
-from parsec._parsec import DateTime
-from typing import Optional, Dict, List, Tuple
+import oscrypto
+import textwrap
+from oscrypto.asymmetric import PrivateKey
```
FirelightFlagboy, line 8: Do we plan to replace `oscrypto` by the Rust impl ?

FirelightFlagboy (approved):

parsec/backend/cli/sequester.py
FirelightFlagboy: @TimeEngineer up

FirelightFlagboy (approved):

parsec/backend/cli/sequester.py
TimeEngineer: Do you think we can just replace ?

FirelightFlagboy: I don't know that wasn't the question :shrug: 
More like if we want to restrain the usage of crypo to the crate `libparsec-crypto`

TimeEngineer: We should

FirelightFlagboy: I've created an issue for that #3327

PR #3309, opened 2022-10-14
Rework `build.py` on the Rust compile profile selection

FirelightFlagboy:
Allow to configure the rust build profile using the environment variable `CARGO_PROFILE`. If the environment is not present we fallback to using the value defined in `DEFAULT_CARGO_PROFILE` which will be the `release` profile.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -20,8 +20,6 @@ env:
     --side-effects-timeout=10
     -vv
     -x
-  # Compile Rust extensions in release speeds up Python tests a lot
-  FORCE_MATURIN_RELEASE: 1
 
 jobs:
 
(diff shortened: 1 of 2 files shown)


newsfragments/3308.bugfix.rst
```diff
@@ -0,0 +1,2 @@
+Ensure that the application is build in release mode from the CI without too much hustle.
```
touilleMan, line 1: No need for a newsfragment (on top of that we have always compiled in release for parsec-cloud ^^)

touilleMan (approved):

newsfragments/3308.bugfix.rst
FirelightFlagboy: done

PR #3312, opened 2022-10-14
[GUI] Display an error message when trying to share a workspace while offline

Max-7:


parsec/core/gui/workspace_sharing_widget.py
```diff
@@ -348,6 +349,7 @@ def _on_get_users_error(self, job):
             show_error(self, _("TEXT_WORKSPACE_SHARING_OFFLINE"))
         self.spinner.hide()
         self.widget_users.show()
+        self.dialog.reject()
```
vxgmichel, line 352:
```suggestion
        if self.dialog is not None:
            self.dialog.reject()
```

vxgmichel (approved):
A suggestion, LGTM otherwise

PR #3325, opened 2022-10-17
Add mypy to `core.backend_connection`

### LINKED ISSUE #3319: Enable `mypy` on `core.backend_connection`

Volham22:
This close #3319


parsec/core/backend_connection/apiv1_annonymous.py
```diff
@@ -37,22 +44,22 @@ async def apiv1_backend_anonymous_cmds_factory(
...
-    async def _destroy_transport():
+    async def _destroy_transport() -> None:
         nonlocal transport
         if transport:
             await transport.aclose()
             transport = None
 
     @asynccontextmanager
-    async def _acquire_transport(**kwargs):
+    async def _acquire_transport(**kwargs: Any) -> AsyncGenerator[Optional[Transport], Optional[Transport]]:  # type: ignore[misc]
```
vxgmichel, line 62: `object` instead of `Any`?

parsec/core/backend_connection/authenticated.py
```diff
@@ -60,7 +70,11 @@ def copy_exception(exception: BaseExceptionTypeVar) -> BaseExceptionTypeVar:
 
 
 class BackendAuthenticatedCmds:
-    def __init__(self, addr: BackendOrganizationAddr, acquire_transport):
+    def __init__(
+        self,
+        addr: BackendOrganizationAddr,
+        acquire_transport: Callable[..., _AsyncGeneratorContextManager[Transport]],
```
vxgmichel, line 76: Maybe `acquire_transport` is a good use case for `Protocol`:
```python
class AcquireTransport(Protocol):
    def __call__(
        self,
        force_fresh: bool = False,
        ignore_status: bool = False,
        allow_not_available: bool = False,
    ) -> Transport:
        ...
```
This way we can type those arguments too

parsec/core/backend_connection/authenticated.py
```diff
@@ -199,7 +219,7 @@ def __init__(
         self._status_event_sent = False
         self._cmds = BackendAuthenticatedCmds(addr, self._acquire_transport)
         self._manager_connect_cancel_scope = None
-        self._monitors_cbs: List[Callable[..., None]] = []
+        self._monitors_cbs: List[Callable[..., Any]] = []
```
vxgmichel, line 222: All monitors callbacks do not return the same type? That sounds fishy..

Also, maybe another good use case for `Protocol`

parsec/core/backend_connection/authenticated.py
```diff
@@ -242,7 +262,7 @@ async def set_status(
         # and the actual cancellation.
         await trio.lowlevel.checkpoint_if_cancelled()
         old_status, self._status = self._status, status
-        self._status_exc = status_exc
+        self._status_exc = status_exc  # type: ignore[assignment]
```
vxgmichel, line 265: Maybe typing `_status_exc` as `Optional[Exception]` would work?

parsec/core/backend_connection/authenticated.py
```diff
@@ -257,30 +277,30 @@ async def set_status(
...
+    def are_monitors_idle(self) -> bool:
         return self._monitors_idle_event.is_set()
 
-    async def wait_idle_monitors(self):
+    async def wait_idle_monitors(self) -> None:
         await self._monitors_idle_event.wait()
 
     @asynccontextmanager
-    async def run(self):
+    async def run(self) -> AsyncGenerator[None, None]:
```
vxgmichel, line 292: Use `AsyncIterator[None]`

parsec/core/backend_connection/authenticated.py
```diff
@@ -257,30 +277,30 @@ async def set_status(
...
         if self._started:
             raise RuntimeError("Already started")
-        async with trio.open_service_nursery() as nursery:
+        async with trio.open_service_nursery() as nursery:  # type: ignore[attr-defined]
```
vxgmichel, line 295: Good catch, could you import `open_service_nursery` from `parsec.utils` instead?

parsec/core/backend_connection/authenticated.py
```diff
@@ -257,30 +277,30 @@ async def set_status(
...
             nursery.start_soon(self._run_manager)
             yield
             nursery.cancel_scope.cancel()
 
-    async def _run_manager(self):
+    async def _run_manager(self) -> None:
         while True:
             try:
-                with trio.CancelScope() as self._manager_connect_cancel_scope:
+                with trio.CancelScope() as self._manager_connect_cancel_scope:  # type: ignore[assignment]
```
vxgmichel, line 303: Typing `_manager_connect_cancel_scope` is not possible?

parsec/core/backend_connection/authenticated.py
```diff
@@ -359,8 +379,8 @@ async def _manager_connect(self):
                     user_profile_outsider_allowed=rep.user_profile_outsider_allowed,
                     active_users_limit=rep.active_users_limit,
                     # Sequester introduced in APIv2.8/3.2
-                    sequester_authority=rep.sequester_authority_certificate,
-                    sequester_services=rep.sequester_services_certificates,
+                    sequester_authority=rep.sequester_authority_certificate,  # type: ignore[arg-type]
+                    sequester_services=rep.sequester_services_certificates,  # type: ignore[arg-type]
```
vxgmichel, line 383: Hmmm is this a bug in the code base?

parsec/core/backend_connection/authenticated.py
```diff
@@ -370,25 +390,30 @@ async def _manager_connect(self):
...
+                setattr(task_status, "idle", _idle)
+                setattr(task_status, "awake", _awake)
                 await monitor_cb(task_status=task_status)
 
             try:
-                async with trio.open_service_nursery() as monitors_nursery:
+                async with trio.open_service_nursery() as monitors_nursery:  # type: ignore[attr-defined]
 
-                    async with trio.open_service_nursery() as monitors_bootstrap_nursery:
+                    async with trio.open_service_nursery() as monitors_bootstrap_nursery:  # type: ignore[attr-defined]
```
vxgmichel, line 416: Same here

parsec/core/backend_connection/cmds.py
```diff
@@ -326,7 +380,11 @@ async def _send_cmd(transport: Transport, serializer, **req) -> dict:
                 raise BackendProtocolError("Invalid request data according to backend")
 
             # Backward compatibility with older backends (<= v2.3)
-            if rep.status == "invalid_certification" and "timestamp" in rep.reason:
+            if (
+                rep.status == "invalid_certification"
+                and rep.reason is not None
```
vxgmichel, line 385: Oh, so you found a bug here?

parsec/core/backend_connection/expose_cmds.py
```diff
@@ -54,4 +41,5 @@ async def wrapper(self, *args: PArgs, **kwargs: PKwargs) -> R:
             async with self.acquire_transport(force_fresh=True) as transport:
                 return await cmd(transport, *args, **kwargs)
 
+    # because of wraps mypy does not infer a proper type
```
vxgmichel, line 44: Oops, old artefact?

parsec/core/backend_connection/invited.py
```diff
@@ -2,19 +2,25 @@
...
+
+T = TypeVar("T")
 
 
 class BackendInvitedCmds:
-    def __init__(self, addr, acquire_transport):
+    def __init__(
+        self,
+        addr: BackendAddrType,
+        acquire_transport: Callable[..., _AsyncGeneratorContextManager[T]],
```
vxgmichel, line 22: Maybe use [AbstractAsyncContextManager](https://docs.python.org/3/library/contextlib.html#contextlib.AbstractAsyncContextManager)?
Also see: https://docs.python.org/3/library/typing.html#typing.AsyncContextManager

parsec/core/backend_connection/invited.py
```diff
@@ -48,22 +54,22 @@ async def backend_invited_cmds_factory(
...
-    async def _destroy_transport():
+    async def _destroy_transport() -> None:
         nonlocal transport
         if transport:
             await transport.aclose()
             transport = None
 
     @asynccontextmanager
-    async def _acquire_transport(**kwargs):
+    async def _acquire_transport(**kwargs: Any) -> AsyncGenerator[Union[BackendInvitedCmds, Optional[Transport]], BackendInvitedCmds]:  # type: ignore[misc]
```
vxgmichel, line 72: That looks fishy.. Find the right type for `transport` and add it line 54

parsec/core/backend_connection/proxy.py
```diff
@@ -47,7 +47,10 @@ def _get_proxy_from_pac(url: str, hostname: str) -> Optional[str]:
     # configured, and `get_pac` silently fails on wrong content-type by returning None)
     # Also disable WPAD protocol to retrieve PAC from DNS given it produce annoying
     # WARNING logs
-    get_pac_kwargs: dict = {"from_dns": False, "allowed_content_types": {""}}
+    get_pac_kwargs: dict[str, bool | set[str] | str] = {
+        "from_dns": False,
+        "allowed_content_types": {""},
+    }
```
vxgmichel, line 53: I typically use `dict[str, object]` in those cases, although it's fine as it is

parsec/core/backend_connection/proxy.py
```diff
@@ -70,6 +73,7 @@ def _get_proxy_from_pac(url: str, hostname: str) -> Optional[str]:
 
     try:
         proxies = pacfile.find_proxy_for_url(url, hostname)
+        assert proxies is not None, "Proxy list is None"
```
vxgmichel, line 76: Good catch! Maybe log the error and return `None` in this case, instead of failing the assertion

parsec/core/backend_connection/proxy.py
```diff
@@ -251,11 +255,12 @@ async def maybe_connect_through_proxy(
 
     conn = h11.Connection(our_role=h11.CLIENT)
 
-    async def send(event):
+    async def send(event: h11.Request) -> None:
         data = conn.send(event)
+        assert data is not None, "Connection is closed"
```
vxgmichel, line 260: @touilleMan What should we do in this case?

parsec/core/backend_connection/proxy.py
```diff
@@ -293,7 +298,7 @@ async def next_event():
                 proxy_url=proxy_url,
                 target_host=host,
                 target_url=target_url,
-                answer_status=answer.status_code,
+                answer_status=answer.status_code,  # type: ignore[attr-defined]
```
vxgmichel, line 301: Looks fishy, what's the explanation?

parsec/core/backend_connection/transport.py
```diff
@@ -52,6 +62,7 @@ async def apiv1_connect(
         BackendConnectionError
     """
     handshake = APIV1_AnonymousClientHandshake(addr.organization_id)
+    assert addr.port is not None, "Organization port is None"
```
vxgmichel, line 65: I think this is an error in the `pyi`, I'd expect `addr.port` to never be `None` (from what I see in the bindings)

parsec/core/backend_connection/transport.py
```diff
@@ -183,12 +196,12 @@ async def _do_handshake(transport: Transport, handshake: BaseClientHandshake) ->
...
     def __init__(self, connect_cb: Callable[[], Awaitable[Transport]], max_pool: int):
         self._connect_cb = connect_cb
-        self._transports = []
+        self._transports: list[Transport] = []
         self._closed = False
         self._lock = trio.Semaphore(max_pool)
 
     @asynccontextmanager
-    async def acquire(self, force_fresh: bool = False) -> AsyncContextManager[Transport]:
+    async def acquire(self, force_fresh: bool = False) -> AsyncGenerator[Transport, None]:
```
vxgmichel, line 204: Use `AsyncIterator[Transport]`

parsec/stubs/pypac/__init__.py
```diff
@@ -0,0 +1,17 @@
+# Parsec Cloud (https://parsec.cloud) Copyright (c) AGPL-3.0 2016-present Scille SAS
+
+from typing import Optional
+from pypac.api import PACFile
```
vxgmichel, line 4: I guess `PACFile` should also be typed here (for the methods we use)

parsec/core/backend_connection/apiv1_annonymous.py
vxgmichel: Also `AsyncIterator[Transport]` + remove `type: ignore`

parsec/core/backend_connection/proxy.py
Volham22: I didn't know what to do here, answer can be an instance of `h11.Event` or a `h11.Response` instance (which has a `status_code` attribute). So I decided to just ignore this error.

parsec/core/backend_connection/transport.py
Volham22: [The underlying rust struct](https://github.com/Scille/parsec-cloud/blob/7442bb3106107aa6314518cf010ca9dfb7d45261/oxidation/libparsec/crates/types/src/addr.rs#L157) takes an `Option<u16>` as port. Is there a default port number?

parsec/core/backend_connection/invited.py
Volham22: I've used the `AcquireTransport` protocol I've defined

parsec/core/backend_connection/cmds.py
Volham22: `reason` attribute is an `Optional[str]` is it an error?

parsec/core/backend_connection/authenticated.py
Volham22: It looks like there's a bug here the rust binding returns `bytes` for `sequester_authority_certificate` and `sequester_services_certificates` but the python `OrganizationConfig` object which stores these values expects `Optional[SequesterAuthorityCertificate]` and `Optional[Tuple[SequesterServiceCertificate, ...]]`. Is it an oxidation mistake?

parsec/stubs/pypac/__init__.py
```diff
@@ -0,0 +1,22 @@
...
+    timeout: int = 2,
+    allowed_content_types: Optional[str] = None,
+    sessions: Optional[requests.Session] = None,
+) -> Optional[PACFile]:
+    ...
+
+
+class PACFile:
+    def find_proxy_for_url(self, url: str, hostname: str) -> str:
+        ...
```
vxgmichel, line 22: `py`-> `pyi`

parsec/core/backend_connection/authenticated.py
```diff
@@ -257,27 +298,27 @@ async def set_status(
     def get_organization_config(self) -> OrganizationConfig:
         return self._organization_config
 
-    def register_monitor(self, monitor_cb) -> None:
+    def register_monitor(self, monitor_cb: Callable[..., Any]) -> None:
```
vxgmichel, line 301: This could too use a `Protocol`

parsec/core/backend_connection/invited.py
```diff
@@ -45,30 +52,33 @@ async def backend_invited_cmds_factory(
         BackendConnectionError
     """
     transport_lock = trio.Lock()
-    transport = None
+    transport: Optional[Transport] = None
     closed = False
 
-    async def _init_transport():
+    async def _init_transport() -> None:
```
vxgmichel, line 58: Same here

parsec/core/backend_connection/authenticated.py
vxgmichel: @touilleMan what's the proper fix for this?

vxgmichel (changes requested):

vxgmichel (approved):

PR #3330, opened 2022-10-18
Dump `poetry` install logs on failure

FirelightFlagboy:
Sometimes installing poetry fail, but we don't have any trace.

Here we will dump the log file generated during the installation,
Thus allowing use to report error if needed.

### DIFF
--- a/.github/actions/setup-python-poetry/action.yml
+++ b/.github/actions/setup-python-poetry/action.yml
@@ -21,9 +21,13 @@ runs:
       if: runner.os != 'Windows'
       run: |
         set -eux
+        set -o pipefail
         export POETRY_HOME="$HOME/.poetry"
         mkdir -p "$POETRY_HOME"
-        curl -sSL https://install.python-poetry.org | python3 - --version=${{ inputs.poetry-version }}
+        if ! curl -sSL https://install.python-poetry.org | python3 - --version=${{ inputs.poetry-version }}; then
+          tail -n +0 ${{ github.workspace }}/poetry-installer*.log
+          exit 3
+        fi
         echo "$POETRY_HOME/bin" >> "$GITHUB_PATH"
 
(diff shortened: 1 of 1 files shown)


TimeEngineer (approved):

.github/actions/setup-python-poetry/action.yml
```diff
@@ -21,9 +21,18 @@ runs:
       if: runner.os != 'Windows'
       run: |
         set -eux
+        set -o pipefail
         export POETRY_HOME="$HOME/.poetry"
         mkdir -p "$POETRY_HOME"
-        curl -sSL https://install.python-poetry.org | python3 - --version=${{ inputs.poetry-version }}
+        if ! curl -sSL https://install.python-poetry.org | python3 - --version=${{ inputs.poetry-version }}; then
+          logfiles=(${{ github.workspace }}/poetry-installer.log)
```
touilleMan, line 28: I don't understand why we do a loop while it seems we only dump `poetry-installer.log` :thinking: 

Can't we just do  `cat ${{ github.workspace }}/poetry-installer.log` instead ?

FirelightFlagboy: :open_mouth: I miss the glob part

FirelightFlagboy: all good now :)

touilleMan: still too complicated for it own good, I would have done `cat ${{ github.workspace }}/poetry-installer*.log` to avoid the tedious loop and array manipulation in shell
(or if you want to display the file name in the dump, you can use this glorious hack `tail -n +1 ${{ github.workspace }}/poetry-installer*.log`

FirelightFlagboy: Your suggestion is great, I'll use it :D

PR #3332, opened 2022-10-19
Enable mypy api data

TimeEngineer:


mypy.ini
```diff
@@ -75,6 +75,13 @@ no_implicit_reexport = False
 strict_equality = True
 disallow_any_unimported = True
 
+[mypy-parsec.api.data.*]
+ignore_errors=False
+disallow_untyped_defs=True
+disallow_any_unimported=True
+disallow_any_decorated=True
```
Volham22, line 82: Isn't it better to set `disallow_any_decorated` to `False` (at least on few problematic modules) instead of adding `type: ignore[misc]` on every decorated functions/methods?

TimeEngineer: I don't know, maybe you would like to activate in `parsec.api.*`

Volham22: I think that it's a better idea to disable this check only on as few modules as possible instead of disable it completely for `parsec.api.*`

TimeEngineer: The goal is to enable mypy for `parsec.*`, and we should not allow people create files where mypy can't check

Volham22: Yes, that why my proposal is to add something like this in `mypy.ini` to allow `any_decorated` only on this specific module.
```suggestion
[mypy-parsec.api.data.base]
disallow_any_decorated=False
```

TimeEngineer: Good idea

FirelightFlagboy (approved):

Volham22 (approved):

FirelightFlagboy (approved):

PR #3338, opened 2022-10-19
Restrain wildcard import to the prelude modules in Rust

FirelightFlagboy:
If you're using `VSCode` with `rust-analyzer` extension, you could add the following:

```json
"rust-analyzer.checkOnSave.command": "clippy",
"rust-analyzer.checkOnSave.extraArgs": [
    "--",
    "--deny=clippy::self_named_module_files",
    "--deny=clippy::wildcard_imports"
],
```


.pre-commit-config.yaml
```diff
@@ -93,6 +93,7 @@ repos:
             --,
             --deny=warnings,
             --deny=clippy::self_named_module_files,
+            --deny=clippy::wildcard_imports
```
touilleMan, line 96: so this doesn't prevent use from `pub use foo::*` reexport ?

oxidation/libparsec/crates/client_types/src/local_device.rs
```diff
@@ -1,12 +1,15 @@
...
 
 use serde::{Deserialize, Serialize};
-use serde_with::*;
+use serde_with::serde_as;
 use serialization_format::parsec_data;
 use sha2::Digest;
 
-use libparsec_crypto::*;
-use libparsec_types::*;
+use libparsec_crypto::{PrivateKey, PublicKey, SecretKey, SigningKey, VerifyKey};
```
touilleMan, line 8: didn't we say we would use `use libparsec_crypto::prelude::*` instead ?

FirelightFlagboy: Yes but `libparsec_crypto` doesn't provide a module `prelude`

FirelightFlagboy: And it was simpler to correct with rust-analyzer `expand glob import`

.pre-commit-config.yaml
FirelightFlagboy: `clippy` don't ask me to change these import, so I would say no it allow us the use of

src/protocol/cmds.rs
```diff
@@ -8,7 +8,21 @@ use pyo3::{
 
 use libparsec::protocol::{authenticated_cmds, invited_cmds};
 
-use crate::protocol::*;
+use crate::protocol::{
```
touilleMan, line 11: This is typically the kind of thing we should avoid: we want to re-expose this stuff in a simple way and not having to get error from this place next time we add a new command

src/protocol/invite.rs
```diff
@@ -32,7 +32,7 @@ import_exception!(parsec.api.protocol, ProtocolError);
 
 #[pyclass]
 #[derive(Clone)]
-pub(crate) struct InvitationType(pub libparsec::types::InvitationType);
+pub(crate) struct InvitationType(pub libparsec::types::invite::InvitationType);
```
touilleMan, line 35: why this change ?

We precisely wan simple import path, this is making things more complex !

oxidation/libparsec/crates/client_types/src/lib.rs
```diff
@@ -1,11 +1,13 @@
 // Parsec Cloud (https://parsec.cloud) Copyright (c) BUSL-1.1 (eventually AGPL-3.0) 2016-present Scille SAS
 
-mod error;
-mod local_device;
-mod local_device_file;
-mod local_manifest;
+pub mod error;
```
touilleMan, line 3: we want simple import, so exposing sub modules is making things more complexe for no reason

we should be doing something like:
```rust
mod error;
mod local_device;
mod local_device_file;
mod local_manifest;

pub mod prelude {
    pub use super::error::*;
    pub use super::local_device::*;
    pub use super::local_device_file::*;
    pub use super::local_manifest::*;
}

pub use prelude::*;
```

oxidation/libparsec/crates/client_types/tests/test_local_device_file.rs
```diff
@@ -4,7 +4,7 @@ use hex_literal::hex;
 use rstest::rstest;
 use std::collections::HashSet;
 
-use libparsec_client_types::{
+use libparsec_client_types::prelude::{
```
touilleMan, line 7: no need to use the `::prelude` if we don't do star import

touilleMan (changes requested):

src/protocol/cmds.rs
touilleMan: Can you relace this by:
```rust
use crate::protocol;
[...]
  AnyCmdReq::InviteList(x) => protocol::InviteListReq(x).into_py(py),
```

FirelightFlagboy:
Just Closing

PR #3346, opened 2022-10-19
Remove noop --timestamp param from `parsec core run` CLI command

### LINKED ISSUE #3345: `parsec core run` has a useless `--mountpount` parameter

touilleMan:
closes #3345

### DIFF
--- a/newsfragments/3345.misc.rst
+++ b/newsfragments/3345.misc.rst
@@ -1,0 +1,1 @@
+Remove noop `--timestamp`` param from `parsec core run` CLI command
--- a/newsfragments/3345.misc.rst
+++ b/newsfragments/3345.misc.rst
@@ -1,1 +1,1 @@
-Remove noop `--timestamp`` param from `parsec core run` CLI command
+Remove noop `--timestamp` param from `parsec core run` CLI command
--- a/parsec/core/cli/run.py
+++ b/parsec/core/cli/run.py
@@ -5,7 +5,6 @@ import trio
 import click
 import multiprocessing
 from pathlib import Path
-from parsec._parsec import DateTime
 from typing import Optional
 
 from parsec.utils import trio_run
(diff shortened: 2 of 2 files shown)


FirelightFlagboy (commented):
I don't see how the issue #3345 is related to this PR

The issue point a problem with the option `--mountpoint`
but here we remove the option `--timestamp`

Maybe change the title of the issue if it was miss labeled, to prevent confusion in the changelog :thinking:

touilleMan:
> The issue point a problem with the option --mountpoint
> but here we remove the option --timestamp

The issue was precisely the `--timestamp` did nothing, so there is two way to fix this 1) implement it 2) remove it
We don't have need for this feature for the moment, so the fix is 2)

FirelightFlagboy:
> The issue was precisely the `--timestamp`

Again, it's not the cause

The issue's title is

> `parsec core run` has a useless **`--mountpount`** parameter

touilleMan:
Ha ! I finally understand: I wrote `--mountpoint` by mistake in the original issue ! I've corrected it ;-)

newsfragments/3345.misc.rst
```diff
@@ -0,0 +1 @@
+Remove noop --timestamp param from `parsec core run` CLI command
```
FirelightFlagboy, line 1:
```suggestion
Remove noop `--timestamp` param from `parsec core run` CLI command
```

FirelightFlagboy (approved):
Just a minor suggestion for the fragment, but it's ok

newsfragments/3345.misc.rst
```diff
@@ -0,0 +1 @@
+Remove noop `--timestamp`` param from `parsec core run` CLI command
```
FirelightFlagboy, line 1:
```suggestion
Remove noop `--timestamp` param from `parsec core run` CLI command
```

FirelightFlagboy (changes requested):

PR #3354, opened 2022-10-20
Add SSE event support in server

### LINKED ISSUE #3057: Implement event with SSE on new HTTP RPC API
We need [`Server-Sent Events`](https://en.wikipedia.org/wiki/Server-sent_events) to be able to send events to clients from the server-side without requiring them to send HTTP request on a time basis to query them.

The received event from the server should be dispatch to the `event_bus`.

One major differences between `WebSocket` and `SSE`: `SSE` don't support binary data, so we will need to serialize them in `base64`.

On the client-side, the task that will wait for events from the server need to be run in the background (the goal is to not have to await the task).

> We will keep the handle on the session to be able to close it when the client log-out.

see #2358

touilleMan:
close #3057


FirelightFlagboy (commented):
Le dot

parsec/backend/asgi/rpc.py
```diff
@@ -1,10 +1,13 @@
 # Parsec Cloud (https://parsec.cloud) Copyright (c) BUSL-1.1 (eventually AGPL-3.0) 2016-present Scille SAS
-from __future__ import annotations
 
-from base64 import b64decode
-from typing import NoReturn, Tuple
+from __future__ import annotations
+import trio
+from base64 import b64decode, b64encode
+from typing import AsyncIterator, AsyncIterable, NoReturn, Tuple, Optional, Any
```
vxgmichel, line 6: `Optional` has been replaced everywhere in the codebase, use `X | None` instead

parsec/backend/asgi/rpc.py
```diff
@@ -268,3 +281,138 @@ async def authenticated_api(raw_organization_id: str) -> Response:
...
+        # Cancel scope will be closed automatically in case of backpressure issue
+        # Note we create it here instead of in `_make_contextmanager` to avoid concurrency
+        # issue if the cancellation kicks in between the moment we return the
+        # response object and when quart starts entering the async context manager
+        # returned by `_make_contextmanager`.
+        assert client_ctx.cancel_scope is None
+        client_ctx.cancel_scope = trio.CancelScope()
+        # Zero-sized channel to avoid backpressure issue: send is blocking until a receive occurs
+        self._body_line_sender, self._body_line_receiver = trio.open_memory_channel[bytes](0)
+        self._contextmanager = self._make_contextmanager(client_ctx, backend)
```
vxgmichel, line 314: `super()` isn't called anywhere, is that on purpose that we're not using any logic from `ResponseBody`?

parsec/backend/asgi/rpc.py
```diff
@@ -268,3 +281,138 @@ async def authenticated_api(raw_organization_id: str) -> Response:
...
+                if backend.config.sse_keepalive is None:
+                    async for event in client_ctx.receive_events_channel:
+                        body_line = b"data:" + b64encode(event.dump()) + b"\n\n"
+                        await self._body_line_sender.send(body_line)
+
+                else:
+                    while True:
+                        with trio.move_on_after(backend.config.sse_keepalive) as scope:
+                            try:
+                                print("~~~~~wait event")
```
vxgmichel, line 341: remove debug prints

parsec/backend/asgi/rpc.py
```diff
@@ -268,3 +281,138 @@ async def authenticated_api(raw_organization_id: str) -> Response:
...
+                            await self._body_line_sender.send(body_line)
+
+                        if scope.cancelled_caught:
+                            body_line = b":keepalive\n\n"
+                            await self._body_line_sender.send(body_line)
+
+            finally:
+                # Closing sender end of the channel will cause Quart stop iterating
+                # on the receiver end and hence terminate the request
+                self._body_line_sender.close()
```
vxgmichel, line 356: Any reason why we're not using an async context manager instead?
```python
async with self._body_line_sender:
    ...
```

parsec/backend/asgi/rpc.py
```diff
@@ -268,3 +281,138 @@ async def authenticated_api(raw_organization_id: str) -> Response:
...
+        assert client_ctx.cancel_scope is not None
+        with client_ctx.cancel_scope:
+            async with trio.open_nursery() as nursery:
+
+                with backend.event_bus.connection_context() as client_ctx.event_bus_ctx:
+                    await backend.events.connect_events(client_ctx)
+
+                    nursery.start_soon(_events_into_body_lines)
+
+                    yield
```
vxgmichel, line 367: I guess it would make sense to also use a context manager for `self._body_line_receiver` around the `yield` statement.

parsec/backend/asgi/rpc.py
```diff
@@ -268,3 +281,138 @@ async def authenticated_api(raw_organization_id: str) -> Response:
...
+        client_ctx.cancel_scope = trio.CancelScope()
+        # Zero-sized channel to avoid backpressure issue: send is blocking until a receive occurs
+        self._body_line_sender, self._body_line_receiver = trio.open_memory_channel[bytes](0)
+        self._contextmanager = self._make_contextmanager(client_ctx, backend)
+
+    @asynccontextmanager
+    async def _make_contextmanager(
+        self, client_ctx: AuthenticatedClientContext, backend: BackendApp
+    ) -> AsyncIterator[None]:
+        async def _events_into_body_lines() -> None:
```
vxgmichel, line 320: I'm not sure I'd call them `lines` as a single SSE can be multiline, e.g:
```
data: This is a message, it
data: has two lines.
```
or:
```
event: remove
data: 2153
```

parsec/backend/asgi/rpc.py
```diff
@@ -268,3 +281,138 @@ async def authenticated_api(raw_organization_id: str) -> Response:
...
+        response=SSEResponseIterableBody(client_ctx, backend),
+        headers={
+            "Content-Type": ACCEPT_TYPE_SSE,
+            "Cache-Control": "no-cache",
+            "Transfer-Encoding": "chunked",
+        },
+        status=200,
+    )
+    # SSE are long lasting query, so we must disable Quart DOS protection here
+    response.timeout = None
```
vxgmichel, line 417: So timeout is not an argument `current_app.response_class()`?

parsec/backend/asgi/rpc.py
```diff
@@ -268,3 +281,138 @@ async def authenticated_api(raw_organization_id: str) -> Response:
...
+        raw_organization_id=raw_organization_id,
+        backend=backend,
+        allow_missing_organization=False,
+        check_authentication=True,
+        # We don't care of Content-Type given the request has no body
+        expected_content_type=None,
+        expected_accept_type=ACCEPT_TYPE_SSE,
+    )
+    assert isinstance(user, User)
+    assert isinstance(device, Device)
```
vxgmichel, line 394:
```suggestion
    assert user is not None
    assert device is not None
```

parsec/backend/cli/run.py
```diff
@@ -273,6 +273,15 @@ def handle_parse_result(
         " --backend-addr=parsec://localhost:<port>(?no_ssl=False if ssl is not set)`"
     ),
 )
+@click.option(
+    "--sse-keepalive",
+    default=30,
+    show_default=True,
+    type=int,
```
vxgmichel, line 280: I'd use `float` if the unit is seconds

tests/common/rpc_api.py
```diff
@@ -1,21 +1,65 @@
...
+        assert self.connection.status_code is not None
+        return self.connection.status_code
+
+    async def get_next_event(self, raw: bool = False) -> bytes:
+        data = await self.connection.receive()
+        if self._status_code is None:
+            self._status_code = self.connection.status_code
+        assert data.startswith(b"data:")
+        assert data.endswith(b"\n\n")
+        raw_event = b64decode(data[len("data:") : -2])
```
vxgmichel, line 55: Apparently, a space after `:` should be ignored:
> The following stream fires two identical events:
  ```
  data:test

  data: test
  ```
>  This is because the space after the colon is ignored if present.

https://html.spec.whatwg.org/multipage/server-sent-events.html

vxgmichel (commented):
Quart provides an implementation of a `ServerSentEvent`:
```python
from dataclasses import dataclass

@dataclass
class ServerSentEvent:
    data: str
    event: str | None = None
    id: int | None = None
    retry: int | None

    def encode(self) -> bytes:
        message = f"data: {self.data}"
        if self.event is not None:
            message = f"{message}\nevent: {self.event}"
        if self.id is not None:
            message = f"{message}\nid: {self.id}"
        if self.retry is not None:
            message = f"{message}\nretry: {self.retry}"
        message = f"{message}\r\n\r\n"
        return message.encode('utf-8')
```
https://pgjones.gitlab.io/quart/how_to_guides/server_sent_events.html
Why not use it?

parsec/backend/cli/run.py
touilleMan: I wonder about int vs float here, but sub-second precision for this is overkill and using int saves us from float strangeness in parsing (passing `NaN`, `inf`, `-inf` etc.)

vxgmichel: I would not worry about that, `float` in seconds in the de-facto standard for time handling in python. Typically, `sse_keepalive` is used in `trio.move_on_after` which takes a `float`:

```python
trio.move_on_after(backend.config.sse_keepalive)
```

touilleMan:
> https://pgjones.gitlab.io/quart/how_to_guides/server_sent_events.html
> Why not use it?

This `ServerSentEvent` is only a way to encode into `bytes` an arbitrary payload.
In our code we already have event objects with `.dump()` method to do the `bytes` encoding. Of course this is not the same because we have only encoded the event where `ServerSentEvent` returns the data that should be send as HTTP, but this is 80% of the work: we only have to add the `header + ":" ` prefix and do a base64 encoding.

Long story short header + base64 seems just simpler to inline in the `_events_into_body_lines` than to introduce yet another class to handle this transformation.

parsec/backend/asgi/rpc.py
touilleMan: It's a hack in Quart:

```python
class Response(SansIOResponse):
    automatically_set_content_length = True
    default_mimetype = "text/html"
    data_body_class = DataBody
    file_body_class = FileBody
    implicit_sequence_conversion = True
    io_body_class = IOBody
    iterable_body_class = IterableBody
    json_module = json

    def __init__(
        self,
        response: Union[ResponseBody, AnyStr, Iterable, None] = None,
        status: Optional[int] = None,
        headers: Optional[Union[dict, Headers]] = None,
        mimetype: Optional[str] = None,
        content_type: Optional[str] = None,
    ) -> None:
        super().__init__(status, headers, mimetype, content_type)
        self.timeout: Any = Ellipsis
        [...]
```

parsec/backend/cli/run.py
touilleMan: I trust you on this, but my guts tell me float is just too clever for our own good :/
Anyway if float do get annoying, it would be easy (and won't break much of the compatibility) to switch to `int` here so no big deal ^^

touilleMan: On the other hand, the good part of using float here is can use `math.inf` instead of `None` to disable the keepalive, hence not needing a special case for this \o/

tests/common/rpc_api.py
touilleMan: done.

parsec/backend/asgi/rpc.py
touilleMan: :+1: done.

parsec/backend/asgi/rpc.py
touilleMan: I've renamed `body_line` into `sse_payload` (I'd rather not call that `event` given we already have the Parsec event that are also used in the code)

parsec/backend/asgi/rpc.py
touilleMan: I didn't add it the context manager on `self._body_line_receiver` so that `self._body_line_sender.send` will never raise a `BrokenResourceError`.
The idea being the cancellation propagate from producer to consumer (i.e. the receiver code is quart doing an async for loop forever)

However after closer investigation, there was an issue in the code if Quart decides (e.g. when server closes) to cancel the query. I've updated the code and added some looong comments ^^

parsec/backend/asgi/rpc.py
touilleMan: no reason, done.

parsec/backend/asgi/rpc.py
touilleMan: done.

parsec/backend/asgi/rpc.py
touilleMan: `ResponseBody` is an ABC so `super().__init__()` doesn't do anything, but adding it as you suggest prevent from thinking there is an error here ^^
done.

parsec/backend/asgi/rpc.py
touilleMan: done.

PR #3367, opened 2022-10-25
Add type `RequiredOption` & `NonRequiredOption` to rust protocol

FirelightFlagboy:
- The type `RequiredOption` mean for a field that it is always present but its value can be null.
- The type `NonRequiredOption` mean that a field can be absent or its value to be null.


vxgmichel (approved):
LGTM :+1:

PR #3369, opened 2022-10-25
Update packb since we handle datetime in pg

TimeEngineer:


FirelightFlagboy (approved):

touilleMan (approved):

parsec/serde/packing.py
```diff
@@ -27,13 +26,7 @@ def packb(data: Mapping[str, Any], exc_cls: Any = SerdePackingError) -> bytes:
...
 
     def _default(obj: Any) -> Any:
-        # TODO: we should be only testing against DateTime, but
-        # asyncpg returns datetime
-        if (
-            isinstance(obj, DateTime)
-            or isinstance(obj, LocalDateTime)
-            or isinstance(obj, datetime.datetime)
-        ):
+        if isinstance(obj, (DateTime, LocalDateTime)):
```
touilleMan, line 29: Do we need `LocalDateTime` ? I have the feeling this is only used in the GUI, and so using it here would signal a mistake has been made anyway :/

FirelightFlagboy (approved):

PR #3397, opened 2022-10-26
Use rust generate_shared_secret_key

TimeEngineer:


FirelightFlagboy (approved):

touilleMan (approved):
Can you do a manual test of an enrollment with a peer using a previous version of Parsec ?
This would further validate the algorithm (and everything around it !)  implemented in Rust is compatible with the one in Python.

PR #3399, opened 2022-10-26
Disable news fragments check

FirelightFlagboy:
- Disable the check for the news fragment bits.
- Add a pull request template to question the pr's author if the changes need a fragment.
- Update the ci to only check the news fragments if there was an edit on the folder `newsfragments`

### DIFF
--- a/.github/actions/paths-filter/action.yml
+++ b/.github/actions/paths-filter/action.yml
@@ -20,6 +20,9 @@ outputs:
   client-common:
     value: ${{ steps.changes.outputs.client-common }}
     description: True if a files in the ionic-client part was edited.
+  newsfragments:
+    value: ${{ steps.changes.outputs.newsfragments }}
+    description: True if a files in the newsfragment part was edited.
 
 runs:
   using: composite
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -72,6 +72,9 @@ jobs:
     steps:
       - uses: actions/checkout@93ea575cb5d8a053eaa0ac8fa3b40d7e05a33cc8 # pin v3.1.0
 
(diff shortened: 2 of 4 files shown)


.github/actions/paths-filter/action.yml
```diff
@@ -20,6 +20,9 @@ outputs:
   client-common:
     value: ${{ steps.changes.outputs.client-common }}
     description: True if a files in the ionic-client part was edited.
+  news-fragments:
+    value: ${{ steps.changes.outputs.news-fragments }}
```
touilleMan, line 24: should stay consistent with the name `newsfragments` without dash

.github/workflows/ci.yml
```diff
@@ -109,12 +112,7 @@ jobs:
       - name: Check Newsfragment
         if: |
           startsWith(github.ref, 'refs/pull/')
-          && !(
-            startsWith(github.head_ref, 'dependabot')
-            || startsWith(github.head_ref, 'yolo')
-            || startsWith(github.head_ref, 'release')
-            || startsWith(github.head_ref, 'revert')
-            )
+          && steps.changes.outputs.news-fragments
```
touilleMan, line 115: So we use the check_newsfragments.py to check the provided newsfragement has the right format ?

touilleMan (changes requested):

.github/workflows/ci.yml
FirelightFlagboy: yes

.github/actions/paths-filter/action.yml
FirelightFlagboy: done

touilleMan (approved):

PR #3419, opened 2022-10-27
Type `create_org_widget`

FirelightFlagboy:
Also rework `typing` for bind-ed `BackendAddr*`


parsec/core/gui/create_org_widget.py
```diff
@@ -278,10 +295,10 @@ def _on_req_success(self, job):
         self.button_validate.setText(_("ACTION_CREATE_DEVICE"))
         self.button_validate.setEnabled(self.current_widget.is_auth_valid())
 
-    def _on_authentication_changed(self, auth_method, valid):
+    def _on_authentication_changed(self, auth_method: Any, valid: bool) -> None:
```
Volham22, line 298: Isn't `auth_method` a [DeviceFileType](https://github.com/Scille/parsec-cloud/blob/c7d127d16f3c2d7c28087da19e77d38ff91a440a/parsec/core/local_device.py#L91)? It's not really a problem because `auth_method` isn't accessed here. Otherwise we can just type it as `object` as it more restrictive than `Any`.
```suggestion
    def _on_authentication_changed(self, auth_method: DeviceFileType, valid: bool) -> None:
```

Volham22 (approved):

PR #3427, opened 2022-10-28
Yolo enhance crypto impl

TimeEngineer:
# What have changed ?

Use macro instead of trait for crypto API


oxidation/libparsec/crates/crypto/src/common/secret.rs
```diff
@@ -1,72 +1,92 @@
...
+                // Filter out any unknown characters, this is typically useful to remove
+                // the `-` and whitespaces.
+                // Note we also discard possible typos from the user (for instance if he types
+                // a `8` or a `9`), but this is no big deal given 1) it should not happen
+                // because GUI should use `RECOVERY_PASSPHRASE_SYMBOLS` to prevent user
+                // from being able to provide invalid characters and 2) it will most likely
+                // lead to a bad password anyway
+                let mut b32 = passphrase
+                    .chars()
+                    .filter(|c| crate::common::RECOVERY_PASSPHRASE_SYMBOLS.contains(c))
```
Volham22, line 40: Isn't is better to use a closure like this instead of iterating in an array?
```rust
|c| c >= 'A' && c <= 'Z' || c >= '0' && c <= '9'
```

Volham22 (changes requested):

oxidation/libparsec/crates/crypto/src/common/secret.rs
TimeEngineer: match is better

Volham22:
```suggestion
    match c {
        'A' ..= 'Z' | '0' ..= '9' => true,
        _ => false
    }
```

TimeEngineer: Done

TimeEngineer: Note that '8' and '9' are not in `RECOVERY_PASSPHRASE_SYMBOLS`

Volham22 (approved):

oxidation/libparsec/crates/crypto/src/common/secret.rs
Volham22: Oh mb, I didn't noticed it

PR #3446, opened 2022-10-31
Mypy gui parsec application

### LINKED ISSUE #3444: [`core.gui`] Add mypy typing for `parsec_application`

TimeEngineer:
# What have changed ?

closes #3444

### DIFF
--- a/parsec/core/gui/parsec_application.py
+++ b/parsec/core/gui/parsec_application.py
@@ -2,29 +2,35 @@
 from __future__ import annotations
 
 import sys
+from typing import TYPE_CHECKING, Set, Tuple, cast
 
 from PyQt5.QtWidgets import QApplication
 from PyQt5.QtCore import QFile, QEvent
-from PyQt5.QtGui import QFont, QFontDatabase, QPalette, QColor
+from PyQt5.QtGui import QFont, QFontDatabase, QPalette, QColor, QFileOpenEvent, QGuiApplication
 from structlog import get_logger
+
+from parsec._parsec import DeviceID, OrganizationID
+
+if TYPE_CHECKING:
+    from parsec.core.gui.main_window import MainWindow
 
 logger = get_logger()
 
 
 class ParsecApp(QApplication):
-    connected_devices = set()
+    connected_devices: Set[Tuple[OrganizationID, DeviceID]] = set()
 
-    def __init__(self):
+    def __init__(self) -> None:
         super().__init__(["-stylesheet"])
         self.setOrganizationName("Scille")
         self.setOrganizationDomain("parsec.cloud")
         self.setApplicationName("Parsec")
-        pal = self.palette()
+        pal = cast(QGuiApplication, self).palette()
         pal.setColor(QPalette.Link, QColor(0x00, 0x92, 0xFF))
         pal.setColor(QPalette.LinkVisited, QColor(0x00, 0x70, 0xDD))
         self.setPalette(pal)
 
-    def event(self, e):
+    def event(self, e: QEvent) -> bool:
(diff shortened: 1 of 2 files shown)


parsec/core/gui/parsec_application.py
```diff
@@ -2,29 +2,32 @@
...
+    connected_devices: Set[Tuple[OrganizationID, DeviceID]] = set()
 
-    def __init__(self):
+    def __init__(self) -> None:
         super().__init__(["-stylesheet"])
         self.setOrganizationName("Scille")
         self.setOrganizationDomain("parsec.cloud")
         self.setApplicationName("Parsec")
-        pal = self.palette()
+        pal = self.palette()  # type: ignore[misc]
```
Max-7, line 25: Why the `ignore`? It should just be a `QPalette` or am I missing something?

parsec/core/gui/parsec_application.py
```diff
@@ -40,46 +43,46 @@ def event(self, e):
             # seemingly the only event we can use for this particular case,
             # ApplicationStateChange not being specific enough.
             self.get_main_window().show_top()
-        if e.type() != QEvent.FileOpen or e.url().scheme() != "parsec":
+        if e.type() != QEvent.FileOpen or e.url().scheme() != "parsec":  # type: ignore[attr-defined]
```
Max-7, line 46: `event` should probably be cast as a `QFileOpenEvent` when trying to access `url()` instead of using ignore.

parsec/core/gui/parsec_application.py
```diff
@@ -91,7 +94,7 @@ def has_active_modal(cls):
         return False
 
     @classmethod
-    def get_main_window(cls):
+    def get_main_window(cls) -> QWidget | None:
```
Max-7, line 97: Cast the return value as `MainWindow` (unless there's an circular import problem)

parsec/core/gui/parsec_application.py
```diff
@@ -40,46 +43,46 @@ def event(self, e):
...
             return super().event(e)
         try:
-            url = e.url().url()
+            url = e.url().url()  # type: ignore[attr-defined]
             mw = self.get_main_window()
             if not mw:
                 pass
             else:
-                mw.new_instance_needed.emit(url)
+                mw.new_instance_needed.emit(url)  # type: ignore[attr-defined]
```
Max-7, line 54: Having the correct type on `get_main_window()` would fix this ignore.

parsec/core/gui/snackbar_widget.py
```diff
@@ -224,7 +224,7 @@ def inform(
         snackbar = SnackbarWidget(
             msg, icon=pix, timeout=timeout, action_text=action_text, action=action, animate=animate
         )
-        main_window.snackbar_manager.add_snackbar(snackbar)
+        main_window.snackbar_manager.add_snackbar(snackbar)  # type: ignore[attr-defined]
```
Max-7, line 227: Same thing, typing `get_main_window` correctly would avoid the ignore.

Max-7 (changes requested):

parsec/core/gui/parsec_application.py
TimeEngineer: `self` doesn't have `palette` method

Max-7: But it inherits `QApplication` which inherits `QGuiApplication` which does have a `palette()` method. Is it a bug in PyQt typing, or mypy does not take inheritance into account? Is it possible to cast `self` into a `QGuiApplication` to avoid the  `ignore`?

Max-7 (approved):

PR #3473, opened 2022-11-02
Type `core.gui.navigation_bar_widget`

FirelightFlagboy:


parsec/core/gui/navigation_bar_widget.py
```diff
@@ -55,20 +55,20 @@ def __init__(self, *args, **kwargs):
         self.paths: List[EntryName] = []
         self.workspace_name: Optional[EntryName] = None
 
-    def _on_range_changed(self, min, max):
+    def _on_range_changed(self, min: int, max: int) -> None:
         self.scroll_area.horizontalScrollBar().setSliderPosition(max)
 
-    def _on_label_clicked(self, idx):
-        def _internal_label_clicked(_title):
+    def _on_label_clicked(self, idx: int) -> Callable[..., None]:
```
Volham22, line 61: Maybe we can use `object` instead of `Any` as parameter in the inner function and have something like this
```python
    def _on_label_clicked(self, idx: int) -> Callable[[object], None]:
        def _internal_label_clicked(_title: object) -> None:
            complete_path = FsPath([p for i, p in enumerate(self.paths) if i < idx])
            self.route_clicked.emit(complete_path)

        return _internal_label_clicked
```

Volham22 (changes requested):

parsec/core/gui/navigation_bar_widget.py
FirelightFlagboy: I can

PS: your comment is misplaced by one line above the incriminated line

FirelightFlagboy: done

Volham22: My bad, I miss-clicked on the wrong line 😅

Volham22: Can the return type of `_on_label_click` be `Callable[[object], None]`? Maybe I'm missing something but I think it's possible.

Volham22 (approved):

parsec/core/gui/navigation_bar_widget.py
FirelightFlagboy: I can, but since we don't use the param, i don't find the annotation useful :/

Max-7 (approved):

PR #3487, opened 2022-11-02
mypy: Added mypy typing for `remote_devices_manager`

### LINKED ISSUE #3484: Enable `mypy` on `core.remote_devices_manager`

Volham22:
# What has changed ?

This PR adds mypy typing to module `parsec.core.remote_devices_manager`. This closes #3484 

### DIFF
--- a/parsec/core/remote_devices_manager.py
+++ b/parsec/core/remote_devices_manager.py
@@ -19,7 +19,7 @@ from parsec.core.backend_connection import (
 )
 
 if TYPE_CHECKING:
-    from parsec.core.backend_connection import BackendAuthenticatedCmds, APIV1_BackendAnonymousCmds
+    from parsec.core.backend_connection import BackendAuthenticatedCmds
 
 DEFAULT_CACHE_VALIDITY = 60 * 60  # 3600 seconds, 1 hour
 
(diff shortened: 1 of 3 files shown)


parsec/core/remote_devices_manager.py
```diff
@@ -102,9 +102,7 @@ async def get_user(
             )
         return verified_user, verified_revoked_user
 
-    async def get_device(
-        self, device_id: Optional[DeviceID], no_cache: bool = False
-    ) -> DeviceCertificate:
+    async def get_device(self, device_id: DeviceID, no_cache: bool = False) -> DeviceCertificate:
```
vxgmichel, line 105: Why is `device_id` no longer optional? Are we sure it never happens?

vxgmichel (approved):
Just a question to make sure the `assert` is ok, LGTM otherwise :+1:

parsec/core/remote_devices_manager.py
Volham22: This getter calls `trustchain_ctx.get_device` which needs a `DeviceID` parameter. If this parameter is `None` this will raise an Exception because `None` can't be converted to a DeviceID object.

PR #3489, opened 2022-11-02
Type `core.gui.{trio_jobs,validators}`

FirelightFlagboy:


parsec/core/gui/trio_jobs.py
```diff
@@ -54,48 +69,55 @@ def __init__(
...
-    def is_crashed(self):
+    def is_crashed(self) -> bool:
         return self.status == "crashed"
 
-    async def __call__(self, *, task_status=trio.TASK_STATUS_IGNORED):
+    async def __call__(
+        self,
+        *,
+        # Mypy: `trio.TaskStatus` is provided by `trio_stubs`
+        task_status: trio.TaskStatus = trio.TASK_STATUS_IGNORED,  # type: ignore[name-defined]
```
TimeEngineer, line 109: Why `ignore` needed, and what's the difference with `trio_typing.TaskStatus` ?

Volham22 (approved):

parsec/core/gui/trio_jobs.py
FirelightFlagboy: I've redo the typing on that part, should be better now

PR #3506, opened 2022-11-04
mypy: Add mypy typing to `ipcinterface`

### LINKED ISSUE #3503: Enable `mypy` on `core.ipcinterface`

Volham22:
# What has changed ?

This PR adds mypy typing for `core.ipcinterface`. This closes #3503 

### DIFF
--- a/parsec/core/ipcinterface.py
+++ b/parsec/core/ipcinterface.py
@@ -2,7 +2,9 @@
 from __future__ import annotations
 
 import sys
+from typing import Any, AsyncIterator, Awaitable, Callable, Dict, Iterator, Mapping
 import trio
+from trio.abc import Stream
 from functools import partial
 from contextlib import contextmanager, asynccontextmanager
 from structlog import get_logger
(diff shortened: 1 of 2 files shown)


parsec/core/ipcinterface.py
```diff
@@ -18,10 +20,13 @@
     SerdeError,
     MsgpackSerializer,
 )
+from parsec.utils import open_service_nursery
 
 
 logger = get_logger()
 
+CommandHandlerCallback = Callable[[Dict[str, object]], Awaitable[Dict[str, Any]]]
```
FirelightFlagboy, line 28:
```suggestion
CommandHandlerCallback = Callable[[dict[str, object]], Awaitable[dict[str, Any]]]
```

using `dict` instead of `Dict` allow use to not import one element from typing (same for `list` & `tuple`)

parsec/core/ipcinterface.py
```diff
@@ -2,7 +2,9 @@
 from __future__ import annotations
 
 import sys
+from typing import Any, AsyncIterator, Awaitable, Callable, Dict, Iterator, Mapping
```
FirelightFlagboy, line 5:
```suggestion
from typing import Any, AsyncIterator, Awaitable, Callable, Iterator, Mapping
```

FirelightFlagboy (changes requested):

parsec/core/ipcinterface.py
Volham22: Done.

FirelightFlagboy (approved):

PR #3512, opened 2022-11-07
Rework rust caching in CI

FirelightFlagboy:
the same cache for rust & python jobs
- Build rust code using the same profile for rust & python jobs
- Add job to pre generate rust cache
- Update dependencies on action `actions/cache` to `v3.0.11`

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -31,6 +31,7 @@ jobs:
     name: ⭐ CI is happy ⭐
     needs:
       - check-quality-assurance
+      - rust-prebuild-matrix
       - test-python-matrix
       - test-rust-matrix
       - test-web-app
@@ -88,7 +89,8 @@ jobs:
         with:
           version: ${{ env.rust-version }}
           components: rustfmt, clippy
-          cache-key: ${{ github.job }}-ubuntu-20.04
+          cache-key: ${{ github.job }}
+          os: ubuntu-20.04
         timeout-minutes: 10
 
       - name: Debug installed tools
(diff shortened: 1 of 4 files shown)


.github/workflows/ci.yml
```diff
@@ -129,11 +131,101 @@ jobs:
...
+          set -ex
+          cargo build \
+            --locked \
+            --workspace \
+            --profile ${{ env.CARGO_PROFILE }} \
+            --all-targets
+          cargo build \
+            --locked \
+            --workspace \
+            --profile ${{ env.CARGO_PROFILE }}
```
vxgmichel, line 220: Why is it called twice?

.github/workflows/ci.yml
```diff
@@ -129,11 +131,101 @@ jobs:
...
+        timeout-minutes: 40
```
vxgmichel, line 221: 40? I would expect the build to be much quicker than that

vxgmichel (approved):
A couple of questions, LGTM otherwise :+1:

.github/workflows/ci.yml
FirelightFlagboy: When I was testing, I see the macos runner that could take up to ~30 min to build `libparsec`

.github/workflows/ci.yml
FirelightFlagboy: It's almost called twice :wink: see the first one as `--all-targets` but not the second one.

When I was testing, even tho we specify `--all-targets` (meaning to build everything (tests and libs)), after that step if I do `cargo build` It was still building/linking something.

FirelightFlagboy: see https://github.com/Scille/parsec-cloud/actions/runs/3420302223/jobs/5694945778

vxgmichel: that's weird tho, isn't it?

.github/workflows/ci.yml
vxgmichel: oO

.github/workflows/ci.yml
FirelightFlagboy: Yes It's weird

FirelightFlagboy: Anyway I've added a comment to clarify this step on why we're doing 2 `cargo build`

PR #3521, opened 2022-11-08
Send event on sqlite.OperationError and diconnect core

AntoineDupre:
# What has changed ?


vxgmichel:
Outdated with #3523

PR #3524, opened 2022-11-08
Add mypy core win registry

### LINKED ISSUE #3520: Enable `mypy` on `core.win_registry`

Volham22:
# What has changed ?

This PR adds mypy typing support to `parsec.core.win_registry. This closes #3520 


parsec/core/gui/custom_dialogs.py
```diff
@@ -80,7 +80,8 @@ def bring_process_window_to_top(
         yield
         return
 
-    def _bring_to_top(pid_target):
+    # mypy: This function always return a value or raise an exception
```
FirelightFlagboy, line 83: Left over comment from an `type: ignore` ?

parsec/core/gui/custom_dialogs.py
```diff
@@ -91,8 +92,8 @@ def _bring_to_top(pid_target):
         callback_type = ctypes.WINFUNCTYPE(win.BOOL, win.HWND, win.LPARAM)
 
         # Callback enumerating the windows
-        def enum_windows_callback(hwnd, param):
-            nonlocal success
+        def enum_windows_callback(hwnd: win.HWND, param: object) -> bool:
+            nonlocal success  # type: ignore[misc]
```
FirelightFlagboy, line 96: What are the reason of the ignore ?

parsec/core/mountpoint/manager.py
```diff
@@ -35,7 +35,7 @@
 
 RunnerType = Callable[
     [UserFS, WorkspaceFS, Path, dict[Any, Any], EventBus],
-    _AsyncGeneratorContextManager[Path],
+    _AsyncGeneratorContextManager[Union[Path, PurePath]],
```
FirelightFlagboy, line 38:
```suggestion
    _AsyncGeneratorContextManager[Path | PurePath],
```

Use `|` that way, we don't have to include `Union`

parsec/core/win_registry.py
```diff
@@ -173,22 +175,23 @@ def del_parsec_drive_icon(letter: str):
...
         return
 
     # Safe context for removing the key after usage
-    with importlib.resources.files(resources).joinpath(DRIVE_ICON_NAME) as drive_icon_path:
-        set_parsec_drive_icon(letter, drive_icon_path)
-        try:
-            yield
-        finally:
-            del_parsec_drive_icon(letter)
+    # mypy: This cast is safe because Traversable is a subset if pathlib.Path
```
FirelightFlagboy, line 185: Left over comment ?

FirelightFlagboy (changes requested):

parsec/core/win_registry.py
Volham22: Yes, I forgot to remove them but note that on Windows a `# type: ignore[return]` is needed here. However mypy complains about unused type ignore on other platforms :(

parsec/core/gui/custom_dialogs.py
Volham22: On windows mypy tells that this variable is unbound which is not the case at runtime. I'll add a comment.

parsec/core/mountpoint/manager.py
Volham22: `Union` is unfortunately mandatory here. Mypy seems to require it.
Here are the errors I get with the `|` syntax.
```
parsec/core/mountpoint/manager.py:38: error: Type expected within [...]  [misc]
parsec/core/mountpoint/manager.py:38: error: The type "Type[_AsyncGeneratorContextManager[Any]]" is not generic and not indexable  [misc]
parsec/core/mountpoint/manager.py:38: error: Unsupported left operand type for | ("Type[Path]")  [operator]
```

FirelightFlagboy (approved):

parsec/core/gui/custom_dialogs.py
TimeEngineer: This isn't the real reason, the reason is that the `nonlocal` `success` is defined in `l.108`
If you move the definition of the function between the variable definition and the function call (l.110), the error should disappear

TimeEngineer (changes requested):

FirelightFlagboy (approved):

TimeEngineer (approved):

PR #3526, opened 2022-11-08
Correct fake command group when the module `backend` was missing in the packaging

FirelightFlagboy:
When building a pre-release wheel of `parsec` `cibuildwheel` was failing in the test phase of the wheel.
The cause was a problem with a bad use of the decorator `click.group` alongside `click.command`, both seems to be exclusive.

Alongside the fix, We also choose
 
- To rename `backend_cmd` => `backend_cmd_group`
- To rename `core_cmd` => `core_cmd_group`
- Correct the definition of the fake cmd `run_gui` in `core.cli.run`

Co-Authored: @vxgmichel

### DIFF
--- a/parsec/backend/cli/__init__.py
+++ b/parsec/backend/cli/__init__.py
@@ -17,7 +17,7 @@ from parsec.backend.cli.sequester import (
 )
 
 
-__all__ = ("backend_cmd",)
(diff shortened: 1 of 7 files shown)


vxgmichel (approved):

PR #3571, opened 2022-11-14
Add the option to rename recovery device at creation

Ironicbay:
# What has changed ?

### DIFF
--- a/newsfragments/3537.misc.rst
+++ b/newsfragments/3537.misc.rst
@@ -1,0 +1,1 @@
+Recovery devices can now be renamed when created.
--- a/parsec/core/gui/device_recovery_export_widget.py
+++ b/parsec/core/gui/device_recovery_export_widget.py
@@ -69,7 +69,7 @@ class DeviceRecoveryExportPage2Widget(QWidget, Ui_DeviceRecoveryExportPage2Widge
 
     def _print_recovery_key(self) -> None:
         html = translate("TEXT_RECOVERY_HTML_EXPORT_user-organization-keyname-password").format(
-            organization=self.device.organization_id,
+            organization=self.device.organization_id.str,
             label=self.device.user_display,
             password=self.passphrase,
             keyname=PurePath(self.label_file_path.text()).name,
--- a/parsec/core/gui/device_recovery_export_widget.py
+++ b/parsec/core/gui/device_recovery_export_widget.py
@@ -19,8 +19,8 @@ from parsec.core.local_device import (
     get_recovery_device_file_name,
     save_recovery_device,
     get_available_device,
-    LocalDeviceAlreadyExistsError,
     DeviceFileType,
+    RECOVERY_DEVICE_FILE_SUFFIX,
 )
 from parsec.core.gui.trio_jobs import QtToTrioJob, JobResultError, QtToTrioJobScheduler
 from parsec.core.gui.lang import translate
@@ -93,11 +93,14 @@ class DeviceRecoveryExportPage1Widget(QWidget, Ui_DeviceRecoveryExportPage1Widge
             )
 
     def _on_select_file_clicked(self) -> None:
-        key_dir = QDialogInProcess.getExistingDirectory(
-            self, translate("TEXT_EXPORT_KEY"), str(Path.home())
+        key_file, _ = QDialogInProcess.getSaveFileName(
+            self,
+            translate("TEXT_EXPORT_KEY"),
+            str(Path.home() / get_recovery_device_file_name(self.get_selected_device())),
+            f"*{RECOVERY_DEVICE_FILE_SUFFIX}",
         )
-        if key_dir:
-            self.label_file_path.setText(key_dir)
+        if key_file:
+            self.label_file_path.setText(key_file)
         self._check_infos()
 
     def _check_infos(self) -> None:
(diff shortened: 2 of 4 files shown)


parsec/core/gui/device_recovery_export_widget.py
```diff
@@ -3,7 +3,7 @@
 from typing import Optional, Sequence
 
 from PyQt5.QtCore import pyqtSignal
-from PyQt5.QtWidgets import QWidget
+from PyQt5.QtWidgets import QWidget, QFileDialog
```
Max-7, line 6: `QFileDialog` import not needed.

parsec/core/gui/device_recovery_export_widget.py
```diff
@@ -95,11 +95,16 @@ def __init__(
...
-        key_dir = QDialogInProcess.getExistingDirectory(
-            self, translate("TEXT_EXPORT_KEY"), str(Path.home())
-        )
-        if key_dir:
-            self.label_file_path.setText(key_dir)
+        key_file = QDialogInProcess.getSaveFileName(
+            self,
+            translate("TEXT_EXPORT_KEY"),
+            str(Path.home() / get_recovery_device_file_name(self.get_selected_device())),
+            "*" + RECOVERY_DEVICE_FILE_SUFFIX,
```
Max-7, line 102: Use f-strings instead.

parsec/core/gui/device_recovery_export_widget.py
```diff
@@ -95,11 +95,16 @@ def __init__(
...
+            None,
+            QFileDialog.ShowDirsOnly,
```
Max-7, line 104: Why `ShowDirsOnly`? I feel like seeing the files too would be a good thing.

parsec/core/gui/device_recovery_export_widget.py
```diff
@@ -164,8 +169,7 @@ async def _export_recovery_device(
     ) -> tuple[LocalDevice, PurePath, str]:
         try:
             recovery_device = await generate_recovery_device(device)
-            file_name = get_recovery_device_file_name(recovery_device)
-            file_path = export_path / file_name
+            file_path = export_path
```
Max-7, line 172: Why not use `export_path` below instead of creating a new variable?

parsec/core/local_device.py
```diff
@@ -509,9 +509,6 @@ async def save_recovery_device(key_file: PurePath, device: LocalDevice, force: b
     assert key_file.suffix == RECOVERY_DEVICE_FILE_SUFFIX
     key_file = trio.Path(key_file)
 
-    if await key_file.exists() and not force:
```
Max-7, line 512: Why deleting this? There's a `force` argument if we want to force an overwrite.

tests/core/gui/test_recovery.py
```diff
@@ -40,7 +40,7 @@ def catch_export_recovery_widget(widget_catcher_factory):
 
 @pytest.mark.gui
 @pytest.mark.trio
-@pytest.mark.parametrize("kind", ["ok", "offline", "already_exists"])
```
Max-7, line 43: Nop, keep the test intact.

tests/core/gui/test_recovery.py
```diff
@@ -72,7 +72,9 @@ async def test_export_recovery_device(
     assert exp_w.current_page.combo_devices.count() == 1
     assert exp_w.current_page.combo_devices.currentData() == alice.slug
 
-    exp_w.current_page.label_file_path.setText(str(tmp_path))
+    exp_w.current_page.label_file_path.setText(
+        str(tmp_path) + "/" + get_recovery_device_file_name(alice)
```
Max-7, line 76: `str(tmp_path / get_recovery_device_file_name(alice))` (assuming that `tmp_path` is some kind of `Path`)

`+` to concatenate should probably never be used.

tests/core/gui/test_recovery.py
```diff
@@ -110,19 +112,6 @@ def _error_shown():
 
             await aqtbot.wait_until(_error_shown)
 
-    elif kind == "already_exists":
```
Max-7, line 113: I'd rather we wrote something in the file initialy, go through the recovery device creation, and check that the file has been overwritten instead of deleting this part of the test.

Max-7 (changes requested):

parsec/core/gui/device_recovery_export_widget.py
Ironicbay: Agreed, but it felt more readable this way.

parsec/core/gui/device_recovery_export_widget.py
Ironicbay: Since we're looking for a directory to save the `.psrk` file it made sense. But it makes little difference

newsfragments/3537.misc.rst
```diff
@@ -0,0 +1 @@
+Recovery devices can now be renamed when created.
```
FirelightFlagboy, line 1: Shouldn't this be a feature instead of miscellaneous ?

FirelightFlagboy (approved):

newsfragments/3537.misc.rst
FirelightFlagboy: What's the other part ?

Ironicbay: The device recovery is the feature, this change is just a miscellaneous one on it.

FirelightFlagboy (approved):

Max-7 (approved):

PR #3593, opened 2022-11-21
Several mypy tweaks

vxgmichel:


parsec/core/pki/submitter.py
```diff
@@ -236,7 +236,7 @@ async def poll(
             PkiEnrollmentInfoError
             PkiEnrollmentInfoNotFoundError
         """
-        #  Tuple[PkiEnrollmentStatus, DateTime, Optional[LocalDevice]]:
+        #  Tuple[PkiEnrollmentStatus, DateTime, LocalDevice | None]:
```
Volham22, line 239: Should we remove this comment?

Volham22 (approved):

parsec/core/pki/submitter.py
vxgmichel: I kept it in case it's useful, but we could definitely remove it

PR #3626, opened 2022-11-25
Oxidize pki anonymous commands

Volham22:
# What has changed ?


parsec/core/pki/submitter.py
```diff
@@ -118,47 +128,47 @@ async def submit(
...
+        if not isinstance(rep, PkiEnrollmentSubmitRepOk):
             raise PkiEnrollmentSubmitError(
-                f"Backend refused to create enrollment: {rep['status']}",
+                f"Backend refused to create enrollment: {type(rep)}",
                 rep,
                 self.enrollment_id,
                 self.x509_certificate,
             )
-        submitted_on = cast(DateTime, rep["submitted_on"])
+        submitted_on = cast(DateTime, rep.submitted_on)
```
TimeEngineer, line 171: cast not needed

parsec/core/pki/submitter.py
```diff
@@ -291,11 +301,11 @@ async def poll(
...
 
         else:
-            assert enrollment_status == PkiEnrollmentStatus.ACCEPTED
-            accepter_der_x509_certificate: bytes = cast(bytes, rep["accepter_der_x509_certificate"])
-            payload_signature: bytes = cast(bytes, rep["accept_payload_signature"])
-            payload: bytes = cast(bytes, rep["accept_payload"])
-            accepted_on: DateTime = cast(DateTime, rep["accepted_on"])
+            assert enrollment_status.is_accepted()
+            accepter_der_x509_certificate: bytes = cast(bytes, rep.accepter_der_x509_certificate)
+            payload_signature: bytes = cast(bytes, rep.accept_payload_signature)
```
TimeEngineer, line 306: same

tests/backend/common.py
```diff
@@ -199,7 +200,12 @@ def __init__(self, cmd, serializer, parse_args=lambda self: {}, check_rep_by_def
     async def _do_send(self, ws, req_post_processing, args, kwargs):
         req = {"cmd": self.cmd, **self.parse_args(self, *args, **kwargs)}
         if req_post_processing:
-            raw_req = packb(req_post_processing(self.serializer.req_dump(req)))
+            if not isinstance(self.serializer, ApiCommandSerializer):
+                pre_processed_req = req_post_processing(self.serializer.req_dump(req))
+                raw_req = packb(pre_processed_req)
+            else:
+                pre_processed_req = req_post_processing(req)
+                raw_req = self.serializer.req_dumps(req)
```
TimeEngineer, line 208: What have changed ?

TimeEngineer (changes requested):

parsec/backend/utils.py
```diff
@@ -54,6 +54,10 @@ async def wrapper(self: T, client_ctx: Ctx, msg: dict[str, object]) -> dict[str,
                 from parsec.api.protocol import InvitedAnyCmdReq
```
FirelightFlagboy, line 54: Why don't we use `parse._parsec.InviteAnyCmdReq` ?

parsec/backend/utils.py
```diff
@@ -54,6 +54,10 @@ async def wrapper(self: T, client_ctx: Ctx, msg: dict[str, object]) -> dict[str,
...
 
                 typed_req = InvitedAnyCmdReq.load(packb(msg))
+            elif client_ctx.TYPE == ClientType.ANONYMOUS:
+                from parsec._parsec import AnonymousAnyCmdReq
+
+                typed_req = AnonymousAnyCmdReq.load(packb(msg))
             else:
                 from parsec.api.protocol import AuthenticatedAnyCmdReq
```
FirelightFlagboy, line 62: Why don't we use `parse._parsec.AuthenticatedAnyCmdReq` ?

parsec/core/backend_connection/anonymous.py
```diff
@@ -36,11 +41,11 @@
 
 
 async def _anonymous_cmd(
-    serializer: CmdSerializer,
+    serializer: ApiCommandSerializer | CmdSerializer,
     addr: BackendPkiEnrollmentAddr | BackendOrganizationBootstrapAddr,
     organization_id: OrganizationID,
     **req: object,
-) -> dict[str, object]:
+) -> Any:
```
FirelightFlagboy, line 48: Can't this be better typed ? we don't have a enum/ generic type for anonymous cmd ?

parsec/core/pki/submitter.py
```diff
@@ -118,47 +128,47 @@ async def submit(
...
                 rep,
                 self.enrollment_id,
                 self.x509_certificate,
             )
 
-        if rep["status"] != "ok":
+        if not isinstance(rep, PkiEnrollmentSubmitRepOk):
             raise PkiEnrollmentSubmitError(
-                f"Backend refused to create enrollment: {rep['status']}",
+                f"Backend refused to create enrollment: {type(rep)}",
```
FirelightFlagboy, line 166: What the expected string here ?

On my side, I get 

```python
>>> from parsec._parsec import PkiEnrollmentAcceptRepInvalidPayloadData
>>> print(type(PkiEnrollmentAcceptRepInvalidPayloadData("foo")))
<class 'builtins.PkiEnrollmentAcceptRepInvalidPayloadData'>
```

don't we want something like ?

```python
>>> print(type(PkiEnrollmentAcceptRepInvalidPayloadData("foo")).__name__)
PkiEnrollmentAcceptRepInvalidPayloadData
```

src/protocol/cmds.rs
```diff
@@ -132,3 +133,32 @@ impl InvitedAnyCmdReq {
...
+                .dump()
+                .map_err(|e| ProtocolError::new_err(format!("encoding error: {e}")))?,
+        ))
+    }
+
+    #[classmethod]
+    fn load(_cls: &PyType, buf: Vec<u8>, py: Python) -> PyResult<PyObject> {
+        use anonymous_cmds::AnyCmdReq;
+
+        let cmd = AnyCmdReq::load(&buf)
```
FirelightFlagboy, line 157:
```suggestion
    fn load(_cls: &PyType, buf: &[u8], py: Python) -> PyResult<PyObject> {
        use anonymous_cmds::AnyCmdReq;

        let cmd = AnyCmdReq::load(buf)
```

Can we take a slice instead of coping the bytearray ?

tests/backend/common.py
```diff
@@ -199,7 +200,12 @@ def __init__(self, cmd, serializer, parse_args=lambda self: {}, check_rep_by_def
     async def _do_send(self, ws, req_post_processing, args, kwargs):
         req = {"cmd": self.cmd, **self.parse_args(self, *args, **kwargs)}
         if req_post_processing:
-            raw_req = packb(req_post_processing(self.serializer.req_dump(req)))
+            if not isinstance(self.serializer, ApiCommandSerializer):
+                pre_processed_req = req_post_processing(self.serializer.req_dump(req))
+                raw_req = packb(pre_processed_req)
+            else:
+                pre_processed_req = req_post_processing(req)
```
FirelightFlagboy, line 207: `pre_processed_req` is not used here ?

FirelightFlagboy (commented):
We should split the file `protocol.pyi` it's too big for me

tests/backend/common.py
Volham22: This is because of the `req_post_processing` callback we can't call `packb` on rust object :(

Volham22: Note that tests commands are sent as legacy `dict` https://github.com/Scille/parsec-cloud/blob/a5a95ccc7c0236c350c415add9cbaae90f5ca58c/tests/backend/common.py#L648 (as it's done for every commands)

parsec/backend/utils.py
Volham22: It seems like `parsec.api.protocol` just re-export it, I'll change to `parsec._parsec` to make it clear.

parsec/backend/utils.py
Volham22: Same thing as I answered above

parsec/core/backend_connection/anonymous.py
Volham22: Can we use an union of all the `Rep` types? I don't see a best fit here.

src/protocol/cmds.rs
TimeEngineer: Good catch

tests/backend/common.py
Volham22: That's because of `tests/backend/test_pki.py::test_pki_submit_no_email_provided` which make use of this callaback to ensure no email address is being sent to the server (which is the purpose of the test).

parsec/core/backend_connection/anonymous.py
```diff
@@ -1,11 +1,16 @@
...
+    EnrollmentID,
+    PkiEnrollmentInfoRep,
+    PkiEnrollmentSubmitRep,
+)
 from parsec._version import __version__
-from parsec.api.protocol.base import CmdSerializer
+from parsec.api.protocol.base import ApiCommandSerializer, CmdSerializer
 from parsec.crypto import VerifyKey
 from parsec.api.protocol import (
     OrganizationID,
```
FirelightFlagboy, line 16: `OrganizationID` & `VerifyKey` should be imported from `parsec._parsec`

TimeEngineer (approved):

FirelightFlagboy (approved):

PR #3631, opened 2022-11-29
Split `protocol.pyi` into submodules

Volham22:
# What has changed ?

This PR split `protocol.pyi` into smaller submodules.


parsec/_parsec_pyi/protocol/__init__.pyi
```diff
@@ -0,0 +1,391 @@
+# Parsec Cloud (https://parsec.cloud) Copyright (c) BUSL-1.1 (eventually AGPL-3.0) 2016-present Scille SAS
```
FirelightFlagboy, line 1: It's required to re-export everything in the `__init__.pyi` ?

FirelightFlagboy (approved):

parsec/_parsec_pyi/protocol/__init__.pyi
Volham22: You're right, it's alredy exported by the `import` statements. I fixed it

PR #3632, opened 2022-11-29
Rewrite the type json file to respect the type scheme

FirelightFlagboy:


oxidation/libparsec/crates/client_types/src/local_device.rs
```diff
@@ -8,8 +8,13 @@ use sha2::Digest;
 use libparsec_crypto::prelude::*;
 use libparsec_types::*;
 
+parsec_data!({
+    path = "schema/local_device.json5";
+    crates = { libparsec_client_types = "crate" }
```
TimeEngineer, line 13: I don't know if it's a good idea, you slow down the compilation time because you don't want to write `use crate as libparsec_client_types` ?

FirelightFlagboy: not really, 

because I'm now generating the struct + impl onto a versioned module `vX`,
The `use crate as libparsec_client_types` would need to be present in the said modules but this is also true for `libparsec_types`.

And the build is cached so It would not impact much :thinking:

TimeEngineer (approved):

FirelightFlagboy:
Not wanted at the time of speaking

PR #3656, opened 2022-12-01
Oxidize protocol error

### LINKED ISSUE #3653: Oxidize the python exception `ProtocolError`

Volham22:
# What has changed ?

This PR oxidize the `ProtocolError` exception. This closes #3653 


TimeEngineer (approved):

TimeEngineer:
We should rename ProtocolErrorException to ProtocolError, to keep all errors the same name convention

TimeEngineer:
You


> > We should rename ProtocolErrorException to ProtocolError, to keep the same name convention
> 
> The `ProtocolErrorException` object is just here to wrap a `ProtocolError` exception like it's done for `TrustchainError`. I don't see a way to get `ProtocolError` inheriting an exception like type. Any idea?
You can wrap with what you want, but the exception name shouldn't change in python code (because all errors have the same suffix)

Volham22:
> > > We should rename ProtocolErrorException to ProtocolError, to keep the same name convention
> > 
> > 
> > The `ProtocolErrorException` object is just here to wrap a `ProtocolError` exception like it's done for `TrustchainError`. I don't see a way to get `ProtocolError` inheriting an exception like type. Any idea?
> 
> You can wrap with what you want, but the exception name shouldn't change in python code (because all errors have the same suffix)

Then we should change the exception name introduced in #3171. It has the same naming issue

PR #3664, opened 2022-12-02
Ruff as python linter

touilleMan:
Why changing the linter ? 1) isort that just works 2) super fast 3) written in Rust :trollface:


parsec/_parsec_pyi/message.pyi
```diff
@@ -25,7 +16,7 @@ class MessageContent:
         author_verify_key: VerifyKey,
         expected_author: DeviceID,
         expected_timestamp: DateTime,
-    ) -> MessageContent: ...
```
TimeEngineer, line 28: Can we import `futures` ?

parsec/_parsec_pyi/protocol/block.pyi
```diff
@@ -15,7 +15,7 @@ class BlockCreateReq:
 class BlockCreateRep:
     def dump(self) -> bytes: ...
     @classmethod
-    def load(cls, buf: bytes) -> BlockCreateRep: ...
+    def load(cls, buf: bytes) -> "BlockCreateRep": ...
```
FirelightFlagboy, line 18:
```suggestion
    def load(cls, buf: bytes) -> BlockCreateRep: ...
```

parsec/_parsec_pyi/protocol/block.pyi
```diff
@@ -40,7 +40,7 @@ class BlockReadReq:
 class BlockReadRep:
     def dump(self) -> bytes: ...
     @classmethod
-    def load(cls, buf: bytes) -> BlockReadRep: ...
+    def load(cls, buf: bytes) -> "BlockReadRep": ...
```
FirelightFlagboy, line 43:
```suggestion
    def load(cls, buf: bytes) -> BlockReadRep: ...
```

FirelightFlagboy (changes requested):
We prefer importing 

```python
```

to not have to quote type

parsec/_parsec_pyi/message.pyi
touilleMan: I've added future in all the .pyi

parsec/_parsec_pyi/protocol/block.pyi
touilleMan: done.

parsec/_parsec_pyi/protocol/block.pyi
touilleMan: done.

vxgmichel (approved):

FirelightFlagboy (approved):

PR #3665, opened 2022-12-02
Oxidation replace core pki

TimeEngineer:
# What has changed ?

Oxidize core pki
fix #3222


Max-7 (changes requested):
Just blocking while parsec-ext is updated.

parsec/core/types/pki.py
```diff
@@ -0,0 +1,9 @@
+# Parsec Cloud (https://parsec.cloud) Copyright (c) AGPL-3.0 2016-present Scille SAS
+
+from parsec._parsec import LocalPendingEnrollment, PkiEnrollmentLocalPendingError, X509Certificate
+
+__all__ = (
+    "LocalPendingEnrollment",
+    "PkiEnrollmentLocalPendingError",
+    "X509Certificate",
+)
```
FirelightFlagboy, line 9: Do we really want to do that ? I thought that we want to import everything from `parsec._parsec` :thinking:

parsec/core/types/pki.py
```diff
@@ -1,210 +1,9 @@
...
-    type = fields.CheckedConstant("requested_device_keys", required=True)
-    private_key = fields.PrivateKey(required=True)
-    signing_key = fields.SigningKey(required=True)
-
-
-pending_device_keys_serializer = MsgpackSerializer(
-    PendingDeviceKeys,
-    validation_exc=PkiEnrollmentLocalPendingValidationError,
-    packing_exc=PkiEnrollmentLocalPendingPackingError,
+__all__ = (
```
Volham22, line 5: I'm wondering if it's a good thing to re-export rust objects like this, instead of editing related imports statements. Importing from `_parsec` makes it clear that it's not a pure python object thus it helps while debugging. What do you think?

TimeEngineer: It breaks `parsec-ext` where I can't access

Volham22 (approved):

Volham22 (approved):

Max-7 (approved):

PR #3694, opened 2022-12-02
Fix a few issues flagged by pyupgrade

DimitriPapadopoulos:
# What has changed ?

Start fixing issues flagged by [pyupgrade](https://github.com/asottile/pyupgrade).

I have left out:
* suggestions related to type checking, as they are pervasive, and I am not sure if you would be willing to consider them,
* suggestions to change immutable lists such `[gen_date() for _ in range(4)]` into tuples such as `(gen_date() for _ in range(4))`, as you might find the result less readable.

### DIFF
--- a/misc/license_headers.py
+++ b/misc/license_headers.py
@@ -229,7 +229,7 @@ LICENSERS_MAP = {
 def get_files(pathes: Iterable[Path]) -> Iterator[Path]:
     for path in pathes:
         if path.is_dir():
-            for f in chain(
+            yield from chain(
                 path.glob("**/*.py"),
                 path.glob("**/*.pyi"),
                 path.glob("**/*.sql"),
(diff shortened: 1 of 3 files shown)


touilleMan:
Thanks for the PR !

> suggestions to change immutable lists such [gen_date() for _ in range(4)] into tuples such as (gen_date() for _ in range(4)), as you might find the result less readable.

`(gen_date() for _ in range(4))` doesn't return a tuple but a generator, which would cause issue if it is used more than once:

```
>>> [x for x in range(10)]
[0, 1, 2, 3, 4, 5, 6, 7, 8, 9]
>>> (x for x in range(10))
<generator object <genexpr> at 0x7f7712128430>
>>> tuple((x for x in range(10)))
(0, 1, 2, 3, 4, 5, 6, 7, 8, 9)
```

So I think you made the good choice ignoring this, as it would require a more in depth review to make sure it does introduce a bug ^^

touilleMan (approved):
:tada: :smiley:

DimitriPapadopoulos:
By the way, would you be interested in the following [pyupgrade](https://github.com/asottile/pyupgrade) suggestions?
* lots of changes related to type checking (you might want to have a look yourself),
* printing (changing `"" %` to `f""` would fix them)

PR #3737, opened 2022-12-08
Oxidize load of ignore files pattern

Volham22:
# What has changed ?

 This PR adds a rust binding to load patterns of files for parsec to ignore.


oxidation/libparsec/crates/types/src/regex.rs
```diff
@@ -12,43 +12,110 @@ fn escape_globing_pattern(string: &str) -> String {
 }
 
 impl Regex {
+    pub fn from_patterns_file(file_content: &str) -> Result<Self, fnmatch_regex::error::Error> {
```
FirelightFlagboy, line 15:
```suggestion
    pub fn from_patterns(raw_regexes: &[str]) -> Result<Self, fnmatch_regex::error::Error> {
```
I think it would be better to pass a list of raw regexes :thinking:

Volham22: Yes, it means that we should call `splitlines` in python. Isn't it better to just take the file path as parameter and handle everything in rust?

FirelightFlagboy: You can create both `from_file(path: &Path)` and `from_raw_regexes(raw_regexes: &[str])`

Volham22: I've implemented both methods and improved error handling

oxidation/libparsec/crates/types/src/regex.rs
```diff
@@ -12,43 +30,137 @@ fn escape_globing_pattern(string: &str) -> String {
...
+            regexes
+                .push(regex::Regex::new(regex_str).map_err(|err| RegexError::ParseError { err })?)
+        }
+
+        Ok(Self(regexes))
     }
 
-    pub fn is_match(&self, string: &str) -> bool {
-        self.0.is_match(string)
+    pub fn from_pattern(pattern: &str) -> RegexResult {
```
FirelightFlagboy, line 65:
```suggestion
    /// Returns a regex from a glob pattern.
    pub fn from_glob_pattern(glob_pattern: &str) -> RegexResult {
```

oxidation/libparsec/crates/types/src/regex.rs
```diff
@@ -12,43 +30,137 @@ fn escape_globing_pattern(string: &str) -> String {
...
     }
-}
 
-impl AsRef<str> for Regex {
-    fn as_ref(&self) -> &str {
-        self.0.as_str()
+    pub fn from_regex_str(regex_str: &str) -> RegexResult {
+        Ok(Regex(vec![
+            regex::Regex::new(regex_str).map_err(|err| RegexError::ParseError { err })?
+        ]))
```
FirelightFlagboy, line 74: Use `from_raw_regexes`

oxidation/libparsec/crates/types/src/regex.rs
```diff
@@ -12,43 +30,137 @@ fn escape_globing_pattern(string: &str) -> String {
...
 }
 
 impl PartialEq for Regex {
     fn eq(&self, other: &Self) -> bool {
-        self.as_ref() == other.as_ref()
+        if self.0.len() == other.0.len() {
+            let lhs: Vec<&str> = self.0.iter().map(|r| r.as_str()).collect();
+            let rhs: Vec<&str> = other.0.iter().map(|r| r.as_str()).collect();
+
+            lhs == rhs
```
FirelightFlagboy, line 88: I don't find this comparison good:

1. We're doing allocation
2. How we should handle `[1, 2, 3]` vs `[1, 3, 2]` ? (i.e. item may not be in the same order between 2 Regex)

oxidation/libparsec/crates/types/src/regex.rs
```diff
@@ -12,43 +30,137 @@ fn escape_globing_pattern(string: &str) -> String {
...
+        assert!(regex.is_match("file.py"));
+        assert!(regex.is_match("file.rs"));
+        assert!(!regex.is_match("stories.txt"));
+    }
+
+    #[test]
+    fn load_default_pattern_file() {
+        let regex = Regex::from_file(Path::new(
+            "../../../../parsec/core/resources/default_pattern.ignore",
+        ))
```
FirelightFlagboy, line 142: use `std::include_str`

parsec/_parsec_pyi/regex.pyi
```diff
@@ -4,7 +4,9 @@ from __future__ import annotations
 
 class Regex:
     @classmethod
-    def from_pattern(cls, pattern: str) -> Regex: ...
+    def from_file(cls, pattern_file_content: str) -> Regex: ...
```
FirelightFlagboy, line 7:
```suggestion
    def from_file(cls, path: Path) -> Regex: ...
```

src/regex.rs
```diff
@@ -10,6 +12,20 @@ crate::binding_utils::gen_proto!(Regex, __str__);
 
 #[pymethods]
 impl Regex {
+    #[classmethod]
+    fn from_file(_cls: &PyType, pattern_file_content: &str) -> PyResult<Self> {
```
FirelightFlagboy, line 16:
```suggestion
    fn from_file(_cls: &PyType, Path: &Path) -> PyResult<Self> {
```

src/regex.rs
```diff
@@ -10,6 +12,20 @@ crate::binding_utils::gen_proto!(Regex, __str__);
...
+        libparsec::types::Regex::from_file(Path::new(pattern_file_content))
+            .map(Self)
+            .map_err(|err| PyValueError::new_err(err.to_string()))
+    }
+
+    #[classmethod]
+    fn from_raw_regexes(_cls: &PyType, raw_regexes: Vec<&str>) -> PyResult<Self> {
```
FirelightFlagboy, line 23:
```suggestion
    fn from_raw_regexes(_cls: &PyType, raw_regexes: &[str]) -> PyResult<Self> {
```

FirelightFlagboy (changes requested):

oxidation/libparsec/crates/types/src/regex.rs
Volham22: `from_file` method expects a path, not the content of a pattern file

parsec/_parsec_pyi/regex.pyi
Volham22: `std::path` cannot be used by `pyo3` because it does not implements `PyClass` trait.

src/regex.rs
Volham22: `std::path` cannot be used by `pyo3` because it does not implements `PyClass` trait.

src/regex.rs
Volham22: If we do so, we will get a compile error because the size of `str` can't be known at compile time. Unfortunately the type `&[&str]` does not implement the `PyClass` trait :(

parsec/_parsec_pyi/regex.pyi
Volham22: See https://github.com/Scille/parsec-cloud/pull/3737#discussion_r1044221291

oxidation/libparsec/crates/types/src/regex.rs
FirelightFlagboy: I know

```rs
let regex = Regex::from_regex_str(std::include_str!("../../../../parsec/core/resources/default_pattern.ignore"));
```

oxidation/libparsec/crates/types/src/regex.rs
Volham22: I've fix the order issue with a `HashSet`. It should also do less allocations. Do we need regex comparison outside of integration test?

parsec/_parsec_pyi/regex.pyi
FirelightFlagboy: just use `path: str` the biggest problem here is the name of the param

src/regex.rs
FirelightFlagboy: see https://github.com/Scille/parsec-cloud/pull/3737#discussion_r1044242818

parsec/_parsec_pyi/regex.pyi
```diff
@@ -4,8 +4,12 @@ from __future__ import annotations
...
 class Regex:
     @classmethod
-    def from_pattern(cls, pattern: str) -> Regex: ...
+    def from_file(cls, path: str) -> Regex: ...
+    @classmethod
+    def from_raw_regexes(cls, raw_regexes: list[str]) -> Regex: ...
     @classmethod
     def from_regex_str(cls, regex_str: str) -> Regex: ...
+    @classmethod
+    def from_glob_pattern(cls, regex_str: str) -> Regex: ...
```
FirelightFlagboy, line 13:
```suggestion
    def from_glob_pattern(cls, glob_pattern: str) -> Regex: ...
```

FirelightFlagboy (changes requested):

oxidation/libparsec/crates/types/src/regex.rs
FirelightFlagboy: I dunnon, we can disable `PartialEq` outside of test by adding `#[cfg(test)]` above the `impl PartialEq`

oxidation/libparsec/crates/types/src/regex.rs
```diff
@@ -1,8 +1,27 @@
...
 
-pub struct Regex(pub regex::Regex);
+use thiserror::Error;
+
+#[derive(Error, Debug)]
+pub enum RegexError {
+    #[error("Regex parsing err: {err}")]
+    ParseError { err: regex::Error },
+    #[error("Failed to convert shell pattern into regex: {err}")]
+    ShellPatternError { err: fnmatch_regex::error::Error },
```
FirelightFlagboy, line 17:
```suggestion
    #[error("Failed to convert glob pattern into regex: {err}")]
    GlobPatternError { err: fnmatch_regex::error::Error },
```

oxidation/libparsec/crates/types/src/regex.rs
```diff
@@ -12,43 +31,135 @@ fn escape_globing_pattern(string: &str) -> String {
...
+            if pattern_line.trim().starts_with('#') || pattern_line.trim() == "\n" {
+                continue;
+            }
+
+            let escaped_glob = escape_globing_pattern(pattern_line.trim());
+            regexes.push(
+                fnmatch_regex::glob_to_regex(&escaped_glob)
+                    .map_err(|err| RegexError::ShellPatternError { err })?,
+            );
+        }
```
FirelightFlagboy, line 50: For the challenge, if you want:

You can rewrite this block of code without using a `for-loop` but by using the `Iterator` trait

FirelightFlagboy (changes requested):
One things on the error variant

oxidation/libparsec/crates/types/src/regex.rs
Volham22: Yes, but the overload is needed for pytest, `cfg(test)` is not active when we do `maturin develop` so regex comparisons will always return `False` in python tests.

oxidation/libparsec/crates/types/src/regex.rs
Volham22: Done: https://github.com/Scille/parsec-cloud/pull/3737/commits/c2e5e1720cd3bedb1ce2551632e715a543bde89f

oxidation/libparsec/crates/types/src/regex.rs
```diff
@@ -1,8 +1,27 @@
...
+    ParseError { err: regex::Error },
+    #[error("Failed to convert glob pattern into regex: {err}")]
+    GlobPatternError { err: fnmatch_regex::error::Error },
+    #[error("Failed to open pattern file: {err}")]
+    PatternFileIOError { err: std::io::Error },
+}
+
+pub type RegexResult = Result<Regex, RegexError>;
+
+pub struct Regex(pub Vec<regex::Regex>);
```
TimeEngineer, line 24: The name isn't correct

oxidation/libparsec/crates/types/src/regex.rs
```diff
@@ -12,43 +31,131 @@ fn escape_globing_pattern(string: &str) -> String {
...
+        let file_content =
+            fs::read_to_string(file_path).map_err(|err| RegexError::PatternFileIOError { err })?;
+
+        Ok(Self(
+            file_content
+                .lines()
+                .into_iter()
+                .filter(|l| l.trim() != "\n" && !l.trim().starts_with('#'))
+                .map(|l| {
+                    fnmatch_regex::glob_to_regex(&escape_globing_pattern(l.trim()))
```
TimeEngineer, line 45: Don't `trim()` 3 times

oxidation/libparsec/crates/types/src/regex.rs
```diff
@@ -12,43 +31,131 @@ fn escape_globing_pattern(string: &str) -> String {
...
 impl PartialEq for Regex {
     fn eq(&self, other: &Self) -> bool {
-        self.as_ref() == other.as_ref()
+        if self.0.len() == other.0.len() {
+            let set: HashSet<&str> = HashSet::from_iter(self.0.iter().map(|r| r.as_str()));
+
+            other.0.iter().all(|r| set.contains(r.as_str()))
+        } else {
+            false
+        }
```
TimeEngineer, line 86: uh ? Why do you use `Vec` then ?

oxidation/libparsec/crates/types/src/regex.rs
```diff
@@ -12,43 +31,131 @@ fn escape_globing_pattern(string: &str) -> String {
...
+
+        assert!(regex.is_match("file.py"));
+        assert!(regex.is_match("file.rs"));
+        assert!(!regex.is_match("stories.txt"));
+    }
+
+    #[test]
+    fn load_default_pattern_file() {
+        let regex = Regex::from_file(Path::new(
+            "../../../../parsec/core/resources/default_pattern.ignore",
```
TimeEngineer, line 136: And windows ?

src/regex.rs
```diff
@@ -35,8 +51,8 @@ impl Regex {
     }
 
     #[getter]
-    fn pattern(&self) -> PyResult<&str> {
-        Ok(self.0.as_ref())
+    fn pattern(&self) -> String {
```
TimeEngineer, line 54: return &str

TimeEngineer (changes requested):

oxidation/libparsec/crates/types/src/regex.rs
```diff
@@ -1,8 +1,27 @@
...
+    fmt::{Display, Write},
+    fs,
+    path::Path,
+};
 
-pub struct Regex(pub regex::Regex);
+use thiserror::Error;
+
+#[derive(Error, Debug)]
+pub enum RegexError {
```
TimeEngineer, line 13: Move to `error.rs`

oxidation/libparsec/crates/types/src/regex.rs
FirelightFlagboy: The test seems to work just fine :thinking:

FirelightFlagboy: https://github.com/Scille/parsec-cloud/actions/runs/3657049957/jobs/6180500459#step:8:804

oxidation/libparsec/crates/types/src/regex.rs
Volham22: Sorry, I don't understand. What did you mean?

If your questions is why did I chose a `Vec`, I took it because `Regex` isn't hashable. This comparison is only used in a few python tests.

oxidation/libparsec/crates/types/src/regex.rs
Volham22: Do you have any suggestion? I think `Regex` is still a good fit here as all it does it matching patterns. The only difference is that it can be built from glob expressions.

Volham22: We could rename `RegexSet` but it may be confusing with `regex::RegexSet` as it isn't the same behavior.

oxidation/libparsec/crates/types/src/regex.rs
TimeEngineer: Can you add a comment then ?
Because another solution is to implement hash with his str repr

oxidation/libparsec/crates/types/src/regex.rs
```diff
@@ -12,43 +18,132 @@ fn escape_globing_pattern(string: &str) -> String {
...
-        self.0.fmt(f)
+        for (i, r) in self.0.iter().enumerate() {
+            r.fmt(f)?;
+
+            if i + 1 < self.0.len() {
+                f.write_char('|')?;
+            }
+        }
+
+        Ok(())
```
TimeEngineer, line 88: What do you prefer ?
```Rust
writeln!(f, self.0.iter().map(regex::Regex::as_str).collect::<Vec<_>>().join("|"))
```

TimeEngineer (approved):

oxidation/libparsec/crates/types/src/regex.rs
Volham22: Let's use your version it's shorter :wink:

oxidation/libparsec/crates/types/src/regex.rs
Volham22: I've added a comment

oxidation/libparsec/crates/types/src/error.rs
```diff
@@ -2,9 +2,21 @@
...
+use crate::{DateTime, DeviceID, EntryID, RealmID, Regex, UserID};
 use libparsec_crypto::CryptoError;
 
+#[derive(Error, Debug)]
+pub enum RegexError {
+    #[error("Regex parsing err: {err}")]
+    ParseError { err: regex::Error },
+    #[error("Failed to convert glob pattern into regex: {err}")]
+    GlobPatternError { err: fnmatch_regex::error::Error },
+    #[error("Failed to open pattern file: {err}")]
```
FirelightFlagboy, line 14: An IO error is not only about failure to open a file, can you change the message to be more generic ?

oxidation/libparsec/crates/types/src/regex.rs
```diff
@@ -12,43 +13,134 @@ fn escape_globing_pattern(string: &str) -> String {
...
+    #[case::trim_whitespace_and_empty_lines(
+        "   *.rs\n\n  \n\n*.py  ",
+        "trim_whitespace_and_empty_lines.tmp"
+    )]
+    #[case::ignore_comment(
+        "# This contains patterns\n## yes\n   *.rs\n\n  \n\n*.py  ",
+        "ignore_comment.tmp"
+    )]
+    fn from_pattern_file_content(#[case] file_content: &str, #[case] filename: &str) {
+        fs::write(filename, file_content).unwrap();
```
FirelightFlagboy, line 109: Can you use a temp dir ?

oxidation/libparsec/crates/types/src/regex.rs
```diff
@@ -12,43 +13,142 @@ fn escape_globing_pattern(string: &str) -> String {
...
+    #[case::ignore_comment(
+        "# This contains patterns\n## yes\n   *.rs\n\n  \n\n*.py  ",
+        "tmp-parsec-regex-tests/ignore_comment.tmp"
+    )]
+    fn from_pattern_file_content(#[case] file_content: &str, #[case] filename: &str) {
+        INIT.call_once(|| {
+            if !Path::new(TEST_DIR_PATH).exists() {
+                fs::create_dir(TEST_DIR_PATH).unwrap();
+            }
+        });
```
FirelightFlagboy, line 116: I was meaning a real temp dir like `/tmp` on linux

FirelightFlagboy: see `std::env::temp_dir()`

FirelightFlagboy: or the fixture `tmp_path` in rust

Volham22: Sorry, I thought windows has no real tmp folders so I made a directory like this. I fixed it.

oxidation/libparsec/crates/types/src/regex.rs
```diff
@@ -12,42 +13,158 @@ fn escape_globing_pattern(string: &str) -> String {
...
+
+        INIT.call_once(|| {
+            let init_tmp_dir = tmp_dir.clone();
+            if !tmp_dir.exists() {
+                fs::create_dir(init_tmp_dir).expect("Failed to create regex tmp dir");
+            }
+        });
+
+        tmp_dir
+    }
```
FirelightFlagboy, line 115: use `tests_fixture::tmp_path`

oxidation/libparsec/crates/types/src/error.rs
```diff
@@ -2,9 +2,21 @@
...
 use libparsec_crypto::CryptoError;
 
+#[derive(Error, Debug)]
+pub enum RegexError {
+    #[error("Regex parsing err: {err}")]
+    ParseError { err: regex::Error },
+    #[error("Failed to convert glob pattern into regex: {err}")]
+    GlobPatternError { err: fnmatch_regex::error::Error },
+    #[error("IO error while opening pattern file: {err}")]
+    PatternFileIOError { err: std::io::Error },
```
FirelightFlagboy, line 15:
```suggestion
    #[error("IO error on pattern file `{file}`: {err}")]
    PatternFileIOError { file: std::path::PathBuf, err: std::io::Error },
```

:thinking:

FirelightFlagboy (approved):

oxidation/libparsec/crates/types/src/error.rs
FirelightFlagboy: I'll let you judge on that proposed change

Volham22: It makes the error message better to have the file path.

FirelightFlagboy (approved):

PR #3753, opened 2022-12-09
Fix special character handling in Reply-To email header

### LINKED ISSUE #3752: Reply-To field is not properly encoded
The sending of email fails when the greeter has special characters in their name, since it is included in the `Reply-To` field

vxgmichel:
Fix #3752

### DIFF
--- a/newsfragments/3752.bugfix.rst
+++ b/newsfragments/3752.bugfix.rst
@@ -1,0 +1,1 @@
+Fix user invitation email sending error when the greeter's name contains special characters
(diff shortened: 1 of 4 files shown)


newsfragments/3752.bugfix.rst
```diff
@@ -0,0 +1 @@
+Fix the handling of special characters in the Send-To field when sending invitation email
```
touilleMan, line 1: "Fix user invitation email sending error when the greeter's name contains special characters"

touilleMan (approved):

PR #3756, opened 2022-12-12
Pre commit cspell

FirelightFlagboy:
- Add `cspell` to check for typo in the codebase using `pre-commit`
  > If you want, you can also use the `vscode` extension for it but you need to install `@cspell/dict-fr-fr` & `@cspell/dict-shell` with `npm install -g`
- Configure `cspell` with the dictionaries to used (even custom one). 
- Correct the typo found

- [x] This PR affect users


.github/workflows/package-ci.yml
```diff
@@ -345,7 +345,7 @@ jobs:
       - name: Test open app
         run: |
           set -eux
-          tar -xzvf dist/parsec-${{ env.WHEEL_VERSION }}-macos-amd64.tar.bz2
```
Volham22, line 348: I don't see a spelling mistake here. Is it a false positive or just a style we should respect

oxidation/bindings/android/libparsec/rust/src/lib.rs
```diff
@@ -27,7 +27,7 @@ pub extern "C" fn Java_com_scille_libparsec_Runtime_startRuntime(
 
 /// Interface between Web, Rust and Android
 /// Cordova bridge API for Android requests the inputs and output to be passed as a JS Object
-/// hence we have to comply with this ourself to provide the same API accross plateforms
+/// hence we have to comply with this ourself to provide the same API across plateformes
```
Volham22, line 30: no `e` needed here see (https://www.britannica.com/dictionary/platform)

> The program can be run on various PC platforms.

oxidation/client/electron/src/setup.ts
```diff
@@ -165,7 +165,7 @@ export class ElectronCapacitorApp {
     // Setup the main manu bar at the top of our window.
     Menu.setApplicationMenu(Menu.buildFromTemplate(this.AppMenuBarMenuTemplate));
 
-    /* If the splashscreen is enabled, show it first while the main window loads then dwitch it out for the main window,
+    /* If the splashscreen is enabled, show it first while the main window loads then ditch it out for the main window,
```
Volham22, line 168: What about this comment:
```
If the splash screen is enabled, show it first while the main window loads, then dismiss it to display the main window. Alternatively, you can just load the main window from the start without showing the splash screen.
```

Volham22 (approved):
Wow so much typos :exploding_head: I've made a few suggestions

.github/workflows/package-ci.yml
FirelightFlagboy: The change is indeed because of a false positive and I wasn't felling it to add to the list of ignored word

oxidation/bindings/android/libparsec/rust/src/lib.rs
FirelightFlagboy: It's platforms without any `e`, `plateformes` is the correct word in french but the text here is in english

oxidation/client/electron/src/setup.ts
FirelightFlagboy: The goal was to correct typo, not to rewrite comment ;)

FirelightFlagboy: But at least that would prevent the typo, I'll change it

oxidation/bindings/android/libparsec/rust/src/lib.rs
FirelightFlagboy: done

oxidation/client/electron/src/setup.ts
FirelightFlagboy: done

.github/workflows/package-ci.yml
FirelightFlagboy: I've disable `cspell` for this specific line and revert the change

Volham22 (approved):
:+1:

PR #3761, opened 2022-12-12
Rework ci

touilleMan:


Cargo.toml
```diff
@@ -39,13 +39,20 @@ members = [
...
-# A custom profile that is the fusion of `release` & `debug` profile.
-# We want the speed of the `release` profile but with the debug assertions of the `debug` profile.
-[profile.integration]
-inherits = "release"
-overflow-checks = true
-debug-assertions = true
-debug = true
-incremental = true
+# Rust unoptimized code is so slow that we customize dev profile to enable some
+# optimizations, otherwise Python tests take forever.
```
FirelightFlagboy, line 43: Asking the dev to use `maturin develop` with `--profile=foo` is too much work ?

.github/workflows/ci.yml
```diff
@@ -139,113 +120,25 @@ jobs:
           cat .pre-commit-config.yaml
 
       - name: Install project (needed for mypy check)
-        run: poetry install -E core -E backend
+        run: |
+          set -ex
+          poetry install -E core -E backend
+        # We skip _parsec.so compilation given Mypy only cares about _parsec.pyi
+        env:
+          POETRY_LIBPARSEC_BUILD_STRATEGY: no_build
```
FirelightFlagboy, line 128: Can you move that to the job scope ? 
That would make it simpler to find

.github/workflows/ci.yml
```diff
@@ -333,33 +226,42 @@ jobs:
...
+      - name: Use cached libparsec if Rust hasn't been modified
+        id: cache-libparsec
+        if: >-
+          steps.python-changes.outputs.run == 'true'
+          && steps.rust-changes.outputs.run != 'true'
+        uses: actions/cache@v3
+        with:
+          # Key cache contains a hash of all the files that are used to produce _parsec.so
+          # Hence if we have a cache hit we know that there is no need for a rebuild !
+          key: ${{ runner.os }}-${{ hashFiles('src/**', 'oxidation/libparsec/**', 'rust-toolchain.toml', 'Cargo.toml', 'Cargo.lock') }}-libparsec-python
```
FirelightFlagboy, line 248: This also include platform dependent crate :thinking:
`platform_web` wouldn't be use by python :wink:  

Don't use `runner.os` that would break if we update the image used, use instead `matrix.os`

.github/workflows/ci.yml
```diff
@@ -333,33 +226,42 @@ jobs:
...
-        continue-on-error: true
         run: |
-          set -x
-          poetry run python --version
-          poetry run pip --version
+          if ${{ steps.cache-libparsec.outputs.cache-hit == 'true' }}; then export POETRY_LIBPARSEC_BUILD_STRATEGY=no_build; fi
+          poetry install -E core -E backend
+        shell: bash
+        env:
+          CARGO_PROFILE: ${{ env.CARGO_PROFILE }}
```
FirelightFlagboy, line 264: no need to re-set it

FirelightFlagboy (changes requested):
Why removing WASM test ?

.github/workflows/ci.yml
```diff
@@ -434,126 +335,26 @@ jobs:
...
-            "-C", "link-arg=-undefined",
-            "-C", "link-arg=dynamic_lookup",
-          ]
-          EOF
-
-      - name: Build rust test codebase
+      - uses: actions-rust-lang/setup-rust-toolchain@685abf830623a8456904376fce2de24302d98a7f # pin v1.3.4
         if: steps.rust-changes.outputs.run == 'true'
-        shell: bash
-        run: cargo test --workspace --profile ${{ env.CARGO_PROFILE }} ${{ env.cargo-exclude-unused-crates }} --no-run
```
FirelightFlagboy, line 467: No way to keep it ?

That would be nice to not compile everything during the test

.github/workflows/ci.yml
ADcorpo: Fixed in [4e5b760](https://github.com/Scille/parsec-cloud/pull/3761/commits/4e5b7603206f48551b1165b2176c27c4d9a0b7b0)

.github/workflows/ci.yml
ADcorpo: Fixed in [3440cc3](https://github.com/Scille/parsec-cloud/pull/3761/commits/3440cc3c7cd9bdc4550fb8f02d0a1daa842bc83b)

FirelightFlagboy: done in 3440cc3

.github/workflows/ci.yml
```diff
@@ -78,12 +76,21 @@ jobs:
     name: 📊 Q&A
     # All linux jobs must run the same ubuntu version to avoid Rust caching issues !
     runs-on: ubuntu-20.04
+    # We skip _parsec.so compilation given Mypy only cares about _parsec.pyi
+    env:
+      : no_build
```
FirelightFlagboy, line 81: something is wrong I can feel it 9d9bb672d4e47dd95650893aaf16d3b25a3ed5ee

@ADcorpo

.github/workflows/ci.yml
FirelightFlagboy: done in 9d9bb67

.github/workflows/ci.yml
touilleMan: I prefer to have this here to make it obvious (this config file is big and it's easy to miss out an env variable)

.github/workflows/ci.yml
touilleMan: > This also include platform dependent crate 🤔
> platform_web wouldn't be use by python 😉

That's not a big deal for the moment. In the long run I think `platform_web`/`platform_native` will be removed given we will split the platform stuff per feature and not per platform (e.g. the `platform_storage` I was talking about the other day)

On top of that, I'd rather not play too clever on the regex: if we add a new crate we won't think about changing the regex, so we have to make sure things just work

> Don't use runner.os that would break if we update the image used, use instead matrix.os
👍

.github/workflows/ci.yml
touilleMan: nop, this is unneeded complexity:
- more code in the config file
- risk of having it wrong and doing needless compilation
- time of test vs time of compilation is already provided in the output of `cargo test`

.github/workflows/ci.yml
FirelightFlagboy: > per feature

You mean per target ?

.github/workflows/ci.yml
FirelightFlagboy: :shrug:

.github/workflows/ci.yml
touilleMan: See my comment https://github.com/Scille/parsec-cloud/pull/3761#discussion_r1050570988

I've even removed `CARGO_PROFILE` env to be extra explicit (which is good given now we have two different ci profiles !)

Cargo.toml
```diff
@@ -39,13 +39,24 @@ members = [
...
+[profile.dev]
+opt-level = 1
```
FirelightFlagboy, line 45: not required with the profile `ci-python`

.github/workflows/ci.yml
```diff
@@ -405,7 +304,6 @@ jobs:
   ##############################################################################
 
   test-rust-matrix:
-    needs: rust-prebuild-matrix
     strategy:
```
FirelightFlagboy, line 307:
```suggestion
    env:
         CARGO_PROFILE: ci-rust
    strategy:
```

Missing to use the custom rust profile for the rust job

Cargo.toml
touilleMan: Please refrain from passive-aggressive.

I agree we most likely want to use `opt-level = 0` for Rust dev and this addition probably solves as much as it breaks 😄 

We need a no-brainer way of building the project from Python, expecting user to know `maturin develop --profile=foo` not acceptable (having to use `maturin develop` instead of just `poetry install` is already annoying enough !)

I've reverted this change, we need more discussion on how to solve this

FirelightFlagboy: I'll try my best :stuck_out_tongue: 

I've found that we can configure the profile used by `maturin`

https://www.maturin.rs/metadata.html#add-maturin-build-options

we can add

```toml
[tool.maturin]
profile = "ci-python" # or even release :thinking: 
```

.github/workflows/ci.yml
touilleMan: My use of "feature" is misleading: I was referring to a component that need multiple implementation depending on the platform it is build for

For instance mountpoint on Linux/macOS sues fuse, on Windows WinFSP and Android API on Android. So we are going to put all of this in a `platform_mountpoint` crate that per-platform mostly separate codebases.

Cargo.toml
touilleMan: > I've found that we can configure the profile used by maturin in pyproject.toml

That's an interesting, however I think it will clash between release and dev build of the project (we want to use release profile for default build, and a custom python-dev profile when building with maturin develop),
I'll have a look at maturin to see if a different config can be applied between `maturn build` and `maturin develop`

touilleMan:
> Why removing WASM test ?

WASM test is basically empty for the moment, so web test (which compile in wasm but doesn't run the wasm test) is enough.

In the long run we most likely want to have a wasm job (and the 🦐 is absolutely too kawaii not to keep ^^), but it should be rethink to only being triggered when modifying the platform_* crates. This is a work for a later time (I'd like to also use this strategy for the macOS build btw).

.github/workflows/ci.yml
touilleMan: This is not needed given we don't do `poetry install` here.

On top of that this env variable is misleading given it may let you think `cargo build` will take it into account ([as far as I know this is not the case](https://doc.rust-lang.org/cargo/reference/environment-variables.html), but there is pretty similar env var such as `CARGO_PROFILE_<name>_BUILD_OVERRIDE_<key>` which make it even more misleading ^^)

In the end I find it far more readable to be explicit by avoiding global environ variable here

Cargo.toml
touilleMan: This is to have incremental build and debug info present when doing dev in local

Cargo.toml
FirelightFlagboy: providing the arg `--profile` override the configured `profile` in the `pyproject.toml`.
In the `build.py` we provide the flag `--profile`

.github/workflows/ci.yml
FirelightFlagboy: Ok :+1:

.github/workflows/ci.yml
FirelightFlagboy: So wha't the goal of having the profile `ci-rust` if we don't use it in the ci :thinking:

touilleMan: It is used in the CI to build the rust tests

FirelightFlagboy: Yes I've missed it

FirelightFlagboy (approved):

PR #3765, opened 2022-12-13
Added virtual entry infos files to winfsp mountpoints

Max-7:
# What has changed ?

This PR adds virtual files to winfsp mountpoints  so that other applications can get stats on a file without using some IPC mechanisms. It works by making the route `f"{file_path}.__parsec_entry_info__"` exists (but not shown in any listing). An open and a read on this file will return stats about the file at `file_path`.


parsec/core/mountpoint/winfsp_operations.py
```diff
@@ -33,6 +34,31 @@
...
+
+    entry_name = path.name.str.removesuffix(ENTRY_INFO_EXTENSION)
+    return path.parent / entry_name
+
+
+def update_stats_for_entry_info(stats: dict[str, Any]) -> dict[str, Any]:
+    if stats["type"] == "folder":
+        stats["type"] = "file"
+        stats["size"] = 0
+        del stats["children"]
```
vxgmichel, line 59: I don't get it, why do `folder` become `file`?

vxgmichel (approved):
Two general questions: 
- Should we do the same for linux?
- How to handle the stats of the entry info file (I'm not convinced with the approach of using the stats of the original file, I think I'd prefer building a new stat dict, and then maybe borrow some info from the original entry. Also I woudn't use zero as size, maybe `1024` or something)

LGTM otherwise :+1:

PR #3766, opened 2022-12-13
Oxidize `UserStorage`

### LINKED ISSUE #3168: Oxidize `UserStorage`
`UserStorage` depends on `ManifestStorage` so to simplify the migration.
I suggest we first migrate `ManifestStorage` (tracked in #3169) then migrate `UserStorage`

# TODOs

- [ ] Remove the dependency on the python class `ManifestStorage` (the class in implemented in Rust but not binded see #3169)
- [ ] Replace the python class `UserStorage` with the Rust one using `pyo3`
- [ ] Correct the fixture `initialize_local_user_manifest` at `tests/fixtures.py`
- [ ] Correct the fixture `initialize_local_user_manifest` at `tests/common/binder.py`
- [ ] Correct the fixture `alice_user_storage` at `tests/core/fs/storage/test_user_storage.py`
- [ ] Correct the test `test_initialization` at `tests/core/fs/storage/test_user_storage.py` 
- [ ] Correct the test `test_storage_file_tree` at `tests/core/fs/storage/test_user_storage.py`

### LINKED ISSUE #3169: Oxidize `ManifestStorage`
# TODOs

- [ ] include change from #3224
- [ ] Replace the python class by the Rust one using `pyo3`
- [ ] Correct the test `test_workspace_manifest_access_while_speculative` at `tests/core/fs/test_bootstrap.py`

### LINKED ISSUE #3170: Remove `LocalDatabase` after oxidation
Remove the python class `LocalDatabase` after oxidation of the last classes that where using it. 

# Depend on 

- #3168 
- #3169 

# TODOs

- [ ] Remove the class `LocalDatabase`
- [ ] Rewrite the suite tests `test_disk_full` at `tests/core/fs/test_disk_full.py`
- [ ] Rewrite the test `test_workspace_manifest_access_while_speculative` at `tests/core/fs/test_bootstrap.py`
- [ ] Rewrite the fixture `persistent_mockup` at `tests/conftest.py`

> Reminder the goal of the rewrite is to not have the class `LocalDatabase` defined in the python code base.

FirelightFlagboy:
closes #3170, closes #3169, closes #3168


tests/core/fs/storage/test_user_storage.py
```diff
@@ -100,10 +103,13 @@ async def test_vacuum(alice_user_storage):
 
 @pytest.mark.trio
 @customize_fixtures(real_data_storage=True)
-async def test_storage_file_tree(tmp_path, alice):
+async def test_storage_file_tree(tmp_path: Path, alice: LocalDevice):
     manifest_sqlite_db = tmp_path / alice.slug / "user_data-v1.sqlite"
 
+    assert (
```
touilleMan, line 109: this assert seems very paranoid to me: `tmp_path` fixture is guarantee to provide us an empty folder exclusive to our test, so I don't see what we are trying to prevent here

I guess the point is to stress the fact we are testing the creation of the fs file here. In this case I think it would be more understandable to add a comment, e.g.

```python
# Pristine start: DB is not created on FS...
assert not manifest_sqlite_db.exists()
async with UserStorage.run(tmp_path, alice) as aus:
    # ...and now it is !
    assert manifest_sqlite_db.is_file()
```

tests/core/fs/storage/test_user_storage.py
```diff
@@ -100,10 +103,13 @@ async def test_vacuum(alice_user_storage):
...
+        not manifest_sqlite_db.exists()
+    ), "database should not exist before the creation of user storage"
     async with UserStorage.run(tmp_path, alice) as aus:
-        assert aus.manifest_storage.path == manifest_sqlite_db
-
-    assert manifest_sqlite_db.is_file()
+        _ = aus.device
+        assert manifest_sqlite_db.exists(), "database file should exists"
+        assert manifest_sqlite_db.is_file(), "database should be a file"
```
touilleMan, line 115: No need to call `exists()` before `is_file()`: `is_file()` will raise a `FileNotFoundError` if the file doesn't exist

tests/core/fs/old_storage/manifest_storage.py
```diff
@@ -113,7 +112,7 @@ async def _create_db(self) -> None:
                 );
                 """
             )
-            # Set the default "prevent sync" pattern if it doesn't exist
+            # set the default "prevent sync" pattern if it doesn't exist
```
touilleMan, line 115: Typing renaming `Set -> set` has made some collateral damages ^^

(also see above for other comments & docstring with same change)

tests/core/fs/conftest.py
```diff
@@ -202,42 +199,45 @@ def run_as_test(cls):
...
+        )
+        from tests.core.fs.old_storage.workspace_storage import (
+            WorkspaceStorage as PyWorkspaceStorage,
+        )
+        from tests.core.fs.old_storage.workspace_storage import (
+            workspace_storage_non_speculative_init as py_wk_init,
         )
 
-        use_original_wk_storage = True
+        use_rust_impl = True
```
touilleMan, line 221: `user_fs_use_rust_impl` ? (`use_rust_impl` is to vague of a name, should be rename to something more explicit)

oxidation/libparsec/crates/core_fs/src/storage/user_storage.rs
```diff
@@ -1,30 +1,50 @@
...
-    user_manifest_id: EntryID,
+    pub device: LocalDevice,
+    pub user_manifest_id: EntryID,
     manifest_storage: ManifestStorage,
+    data_conn: SqliteConn,
 }
 
 impl UserStorage {
-    pub fn new(
+    pub fn from_db_dir(
```
touilleMan, line 22: Why passing a directory instead of just the path of the db file ?

oxidation/libparsec/crates/core_fs/src/storage/user_storage.rs
```diff
@@ -85,6 +105,10 @@ impl UserStorage {
     }
 
     pub fn set_user_manifest(&self, user_manifest: LocalUserManifest) -> FSResult<()> {
+        assert_eq!(
```
touilleMan, line 108: :+1:

touilleMan (commented):
I've seen in Cargo.toml that we have enabled [`r2d2` feature for diesel](https://docs.diesel.rs/2.0.x/diesel/r2d2/index.html), this seems odd to me that we need connection pooling for our fs storage:
- no network with sqlite so zero risk of connection breakage due to disconnection
- given we keep a cache of manifests in memory, we cannot use directly sqlite pooling but instead should handle this on a higher level

TimeEngineer (approved):

FirelightFlagboy:
> I've seen in Cargo.toml that we have enabled [r2d2 feature for diesel](https://docs.diesel.rs/2.0.x/diesel/r2d2/index.html)

With the new impl for `SqliteConn` we don't use `r2d2` pooling, I will remove it on the parent PR

oxidation/libparsec/crates/core_fs/src/storage/user_storage.rs
FirelightFlagboy: 1. When we create a new `UserStorage` with `UserStorage::run` (in python) we only have the said directory path
2. `get_{user,workspace}*_db_path` will be removed in this PR (currently not but it's a miss from my part)
    With that said the python code will not be able to provide the full path (up to the file)

This can change but would require to provide on oxidation of `get_{user,workspace}*_db_path`, this is up to you

tests/core/fs/test_bootstrap.py
```diff
@@ -50,23 +52,24 @@ async def test_workspace_manifest_access_while_speculative(
...
 
-    with freeze_time("2000-01-01"):
+    with freeze_time("2000-01-01", devices=[alice]):
         async with user_fs_factory(alice) as user_fs:
             user_fs: UserFS = user_fs
             wksp_id = await user_fs.workspace_create(EntryName("wksp"))
             # Retrieve where the database is stored
             wksp = user_fs.get_workspace(wksp_id)
-            wksp_manifest_storage_db_path = wksp.local_storage.manifest_storage.path
+            # wksp_manifest_storage_db_path = wksp.local_storage.manifest_storage.path
```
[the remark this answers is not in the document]
FirelightFlagboy: related to https://github.com/Scille/parsec-cloud/pull/3766#discussion_r1053120173, will continue on that discussion

tests/core/fs/test_bootstrap.py
```diff
@@ -50,23 +52,24 @@ async def test_workspace_manifest_access_while_speculative(
...
 
     # Manually drop the workspace manifest from database
-    async with LocalDatabase.run(path=wksp_manifest_storage_db_path) as localdb:
-        async with ManifestStorage.run(
-            device=alice, localdb=localdb, realm_id=wksp_id
-        ) as manifest_storage:
-            await manifest_storage.clear_manifest(wksp_id)
+    raise NotImplementedError("How to replace LocalDatabase & ManifestStorage ?")
+    # async with LocalDatabase.run(path=wksp_manifest_storage_db_path) as localdb:
```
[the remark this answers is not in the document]
FirelightFlagboy: Not really, it's more that with the new implementation this test need change to work with the new implementation but I would need a hand for that

tests/core/fs/storage/test_user_storage.py
```diff
@@ -100,10 +103,13 @@ async def test_vacuum(alice_user_storage):
 
 @pytest.mark.trio
 @customize_fixtures(real_data_storage=True)
-async def test_storage_file_tree(tmp_path, alice):
+async def test_storage_file_tree(tmp_path: Path, alice: LocalDevice):
     manifest_sqlite_db = tmp_path / alice.slug / "user_data-v1.sqlite"
 
+    assert (
+        not manifest_sqlite_db.exists()
+    ), "database should not exist before the creation of user storage"
```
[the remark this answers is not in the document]
FirelightFlagboy: Nice to know, I'll try to remember that

tests/core/fs/storage/test_user_storage.py
FirelightFlagboy: > I guess the point is to stress the fact we are testing the creation of the fs file here

It's the case

tests/core/fs/storage/test_user_storage.py
FirelightFlagboy: done

tests/core/fs/conftest.py
FirelightFlagboy: I renamed the variable to `use_rust_storage_impl` since it's for patching `WorkspaceStorage` & `UserStorage`

tests/core/fs/test_bootstrap.py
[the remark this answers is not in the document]
touilleMan: Add `TODO: blabla` comment then, so that we can retrieve it easily by grepping

parsec/core/fs/storage/user_storage.py
```diff
@@ -3,34 +3,21 @@
...
-
-            timestamp = device.timestamp()
-            manifest = LocalUserManifest.new_placeholder(
-                author=device.device_id,
-                id=device.user_manifest_id,
-                timestamp=timestamp,
-                speculative=False,
-            )
-            await manifest_storage.set_manifest(device.user_manifest_id, manifest)
+    return await trio.to_thread.run_sync(
```
touilleMan, line 18: Rust binding should expose an async function instead, see what has been done for `TimeProvider::sleep` (i.e. the binding function returns a `FutureIntoCoroutine`)

FirelightFlagboy: 1. `UserStorage` & co aren't async because `diesel` isn't so I can't do that 
2. We have an epic to improve `fs/storage` so we better to that in it

tests/core/fs/storage/test_user_storage.py
```diff
@@ -100,10 +103,12 @@ async def test_vacuum(alice_user_storage):
...
+async def test_storage_file_tree(tmp_path: Path, alice: LocalDevice):
     manifest_sqlite_db = tmp_path / alice.slug / "user_data-v1.sqlite"
 
+    # Pristine start: DB is not created on FS...
+    assert not manifest_sqlite_db.exists()
     async with UserStorage.run(tmp_path, alice) as aus:
-        assert aus.manifest_storage.path == manifest_sqlite_db
-
-    assert manifest_sqlite_db.is_file()
+        _ = aus.device
```
touilleMan, line 112: what is this for ?

tests/core/fs/storage/test_user_storage.py
```diff
@@ -100,10 +103,12 @@ async def test_vacuum(alice_user_storage):
...
+        # ...and now it is !
+        assert manifest_sqlite_db.is_file(), "database should be a file"
```
touilleMan, line 114: no need for the message in the test assert

parsec/core/fs/storage/user_storage.py
touilleMan: (besides I would be surprised using this to_thread for every operation involving the storage creates a noticeable performance overhead)

tests/common/freeze_time.py
```diff
@@ -2,82 +2,113 @@
...
+) -> FreezeContext:
+    restore_context: FreezeContext = dict()
+
+    for device in devices:
+        # Get device id
+        dev_id = device.device_id
+
+        # Freeze time with mockup (idempotent) & timeprovider
+        print(f"[{_freeze_devices.__name__}] device_id={dev_id}, update time provider config")
+        # Mypy don't like that we assign a function to a method type
```
touilleMan, line 33: `doesn't like`

tests/common/freeze_time.py
```diff
@@ -2,82 +2,113 @@
...
+        LocalDevice.timestamp = _timestamp_mockup  # type: ignore[assignment]
```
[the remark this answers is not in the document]
touilleMan: And I don't get why this is needed: `LocalDevice.timestamp` is implemented in the Rust binding:
```rust
    fn timestamp(&self) -> PyResult<DateTime> {
        Ok(DateTime(self.0.now()))
    }
```
which makes use:
```rust
    pub fn now(&self) -> DateTime {
        self.time_provider.now()
    }
```

So in the end time provider is used and hence `device.time_provider.mock_time(freeze=time)` should be enoguh

tests/common/freeze_time.py
```diff
@@ -98,9 +129,14 @@ def freeze_time(time: DateTime | str = None, device: LocalDevice | DeviceID | No
         mocks_stack.append(time)
         mock_time(time)
 
-        yield time
+        if len(devices) > 0 and global_freeze:
```
touilleMan, line 132: same as above

tests/common/freeze_time.py
```diff
@@ -2,82 +2,113 @@
...
+    devices = devices or list()
 
     # Get current time if not provided
     if time is None:
         time = DateTime.now()
+    elif isinstance(time, str):
+        [y, m, d] = time.split("-")
+        time = DateTime(int(y), int(m), int(d))
 
     # Freeze a single device
```
touilleMan, line 104: comment is still valid now that we handle `deviceS` ?

tests/common/freeze_time.py
```diff
@@ -2,82 +2,113 @@
...
-    if device is not None:
-        with freeze_device_time(device, time) as time:
+    if len(devices) > 0 and not global_freeze:
```
touilleMan, line 105: And instead of the early return, it would be better more readable to do have a `if ... else ...`

tests/core/fs/storage/test_user_storage.py
FirelightFlagboy: Because otherwise `aus` isn't used
I've rename `aus` to `_aus`

tests/core/fs/storage/test_user_storage.py
FirelightFlagboy: done

tests/common/freeze_time.py
FirelightFlagboy: done

tests/common/freeze_time.py
```diff
@@ -2,82 +2,113 @@
...
+    devices: list[LocalDevice], current_task: trio.lowlevel.Task | None, time: DateTime
+) -> FreezeContext:
+    restore_context: FreezeContext = dict()
+
+    for device in devices:
+        # Get device id
+        dev_id = device.device_id
+
+        # Freeze time with mockup (idempotent) & timeprovider
+        print(f"[{_freeze_devices.__name__}] device_id={dev_id}, update time provider config")
```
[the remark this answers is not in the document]
FirelightFlagboy: done

tests/core/fs/storage/test_user_storage.py
FirelightFlagboy: Otherwise `aus` isn't used, now I've renamed it to `_aus`

tests/core/fs/storage/test_user_storage.py
FirelightFlagboy: done

tests/common/freeze_time.py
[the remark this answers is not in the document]
FirelightFlagboy: done

tests/common/freeze_time.py
FirelightFlagboy: done

tests/common/freeze_time.py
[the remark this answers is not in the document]
FirelightFlagboy: I've moved `LocalDevice.timestamp = _timestamp_mockup` outside of the loop

tests/common/freeze_time.py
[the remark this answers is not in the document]
FirelightFlagboy: done

tests/common/freeze_time.py
FirelightFlagboy: done

tests/common/freeze_time.py
```diff
@@ -2,82 +2,113 @@
...
+    global_freeze: bool = False,
+) -> Iterator[DateTime]:
+    mocks_stack: list[DateTime] = []
+    devices = devices or list()
 
     # Get current time if not provided
     if time is None:
         time = DateTime.now()
+    elif isinstance(time, str):
+        [y, m, d] = time.split("-")
```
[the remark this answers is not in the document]
FirelightFlagboy: doen

tests/common/freeze_time.py
FirelightFlagboy: No, I've change it

oxidation/libparsec/crates/core_fs/src/storage/user_storage.rs
touilleMan: ok so let's stick to the simplest way then ;-)

tests/core/fs/test_bootstrap.py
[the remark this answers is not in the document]
FirelightFlagboy: done in a different way ;)

tests/core/fs/storage/test_user_storage.py
touilleMan: if you don't need aus to you can omit the `as aus` then:
```python
async with UserStorage.run(tmp_path, alice):
    ...
```

parsec/core/fs/storage/user_storage.py
touilleMan: As we discussed yesterday, we need to encapsulate the sqlite part away from the business logic (given we have to support web). So it seems the right move to have this encapsulation defining an async API, and to implement it as a simple thread-per-connection with message passing in the sqlite implementation (the web one relying on browser API, they are already async by default)

touilleMan (approved):

tests/core/fs/storage/test_user_storage.py
FirelightFlagboy: done

PR #3768, opened 2022-12-13
Fix api version serde bug

### LINKED ISSUE #3767: Compatibility broken with `ApiVersion`
Since #3743 compatibility is broken because `ApiVersion` is packed as 2 ints instead of a tuple of ints.

As a result we might get an error like this with an older backend version:
```
{'supported_api_versions': {0: ['Not a ApiVersion'], 1: ['Not a ApiVersion'], 2: ['Not a ApiVersion']}}
```

Volham22:
# What has changed ?

This PR fixes #3767 issue.


oxidation/libparsec/crates/protocol/tests/test_handshake.rs
```diff
@@ -874,5 +874,26 @@ fn test_process_result_req_bad_outcome(
     assert_eq!(err, expected)
 }
 
+// Generated from Python implementation (Parsec v2.15.0+dev)
```
TimeEngineer, line 877: Did you use the old Python or with the binding ?

Volham22: I used to `v2.15.0` tag. So I generated the hex code against the `NamedTuple` implementation in python.

TimeEngineer (approved):

FirelightFlagboy (approved):

PR #3774, opened 2022-12-15
Don't build `rust` lib on QA step when installing python

FirelightFlagboy:
Not building the `rust` binding on the QA job save us some time 
because here we do `poetry install` just to have the lint tools like `ruff` or `mypy`.


FirelightFlagboy:
supersed by #3761

touilleMan:
implemented in #3761

PR #3779, opened 2022-12-18
Faster serialization format

### LINKED ISSUE #3778: De-duplicate cmd schemas common between api versions in protocol crate
Most schemas are similar between API v1/v2/v3, it would be really interesting to have `serialization_format` detecting that given we could be able to divide by almost 3 the amount of schema to compile 😃

touilleMan:
closes #3778 

# What has changed ?

- Replace serde by miniserde for serialization_format internal needs
- Avoid duplication of similar schema

```shell
$ cargo clean -p libparsec_protocol -p libparsec_serialization_format
$ cargo build --timings
[...]
    Finished dev [unoptimized + debuginfo] target(s) in 28.74s
$ git checkout master
$ cargo build
$ cargo clean -p libparsec_protocol -p libparsec_serialization_format
$ cargo build --timings
    Finished dev [unoptimized + debuginfo] target(s) in 48.93s
```

old 
image

new
image

On a clean rebuild of libparsec (i.e. `cargo clean && cargo build`) I have:
- 2mn47s on master
- 1mn56s with this branch
This is due to the fact libparsec_protocol can start it compilation much earlier if libparsec_serialization_format doesn't depend of serde


FirelightFlagboy (changes requested):
Nop keep the dict :(

FirelightFlagboy:
I want #3632 to be merged before

oxidation/libparsec/crates/serialization_format/src/protocol.rs
```diff
@@ -0,0 +1,739 @@
...
+                let gen_req = GenCmdReq {
+                    _cmd: cmd.req.cmd.to_owned(),
+                    kind: match (&cmd.req.unit, &cmd.req.fields) {
+                        (Some(nested_type_name), None) => GenCmdReqKind::Unit {
+                            nested_type_name: nested_type_name.to_owned(),
+                        },
+                        (None, Some(fields)) => GenCmdReqKind::Composed {
+                            fields: fields.iter().filter_map(&convert_field).collect(),
+                        },
+                        (None, None) => GenCmdReqKind::Composed { fields: vec![] },
```
FirelightFlagboy, line 282: This type of struct are called unit struct 

```rust
// A unit struct called `Foo`
struct Foo;
```

oxidation/libparsec/crates/serialization_format/src/protocol.rs
```diff
@@ -0,0 +1,739 @@
...
+                            GenCmdNestedType::Struct { name, .. } => name,
+                        };
+                        allowed_extra_types.insert(name.to_owned(), name.to_owned());
+                    }
+                }
+
+                let gen_req = GenCmdReq {
+                    _cmd: cmd.req.cmd.to_owned(),
+                    kind: match (&cmd.req.unit, &cmd.req.fields) {
+                        (Some(nested_type_name), None) => GenCmdReqKind::Unit {
```
FirelightFlagboy, line 276: This type of struct is called a tuple struct

```rust
// A tuple struct called `Foo`
struct Foo(Bar)
```

oxidation/libparsec/crates/serialization_format/src/protocol.rs
```diff
@@ -0,0 +1,739 @@
...
+        #variant_name(#module_name::Req)
+    };
+
+    (variant, module)
+}
+
+impl GenCmd {
+    pub(crate) fn quote(&self) -> TokenStream {
+        let variant_name = format_ident!("{}", &snake_to_camel_case(&self.cmd));
+        let module_name = format_ident!("{}", &self.cmd);
```
FirelightFlagboy, line 447: Could you provide those values as argument ? 
That way we're sure to use the same value as in `quote_cmd` since it also create those values

oxidation/libparsec/crates/serialization_format/src/protocol.rs
```diff
@@ -0,0 +1,739 @@
...
+                            pub fn load(buf: &[u8]) -> Result<Self, ::rmp_serde::decode::Error> {
+                                ::rmp_serde::from_slice::<Self>(buf)
+                                    .or_else(|err| {
+                                        // Due to how Serde handles variant discriminant, we cannot express unknown
+                                        // status as a default case in the main schema.
+                                        // Instead we have this additional deserialization attempt fallback.
+                                        let data = ::rmp_serde::from_slice::<UnknownStatus>(buf)?;
+
+                                        match data.status.as_str() {
+                                            #(#known_rep_statuses => Err(err),)*
```
FirelightFlagboy, line 508:
```suggestion
                                            #(#known_rep_statuses)|* => Err(err)
```

Could also be written that way, which is better :shrug:

oxidation/libparsec/crates/serialization_format/src/protocol.rs
```diff
@@ -0,0 +1,739 @@
...
+        None => {
+            quote! {
+                #[::serde_with::serde_as]
+                #[derive(Debug, Clone, ::serde::Deserialize, ::serde::Serialize, PartialEq, Eq)]
+                pub enum #name {
+                    #(#variants),*
+                }
+            }
+        }
+    }
```
FirelightFlagboy, line 575: `quote` is able to expand rust option to nothing, so you could factorize that code

oxidation/libparsec/crates/serialization_format/src/protocol.rs
```diff
@@ -0,0 +1,739 @@
...
+    // present in the actual json files: it will be added just before
+    // deserialization by the parsec_protocol_familly macro !
+    items: Vec<JsonCmdFlavour>,
+}
+
+#[derive(Deserialize)]
+pub(crate) struct JsonCmdFlavour {
+    major_versions: Vec<u32>,
+    #[allow(unused)]
+    introduced_in: Option<String>,
```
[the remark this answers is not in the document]
touilleMan: I didn't know miniserde support `Deserialize` on string.

In any case, I stick to the pattern "json structs do basic desirialization, then custom code turn it into final gen structs".
This strategy has it downsides: typically it is very tedious to write a lot of different structures this way (that's why we are happy to use serde and have `Deserialize` on custom type e.g. DeviceID), however this doesn't apply here given we literally have a single schema to support.

On top of that using a `parse_api_version` function separated from miniserde allows better error messages (given miniserde  just give up without explanation if de-serialization fails)

oxidation/libparsec/crates/serialization_format/src/protocol.rs
touilleMan: `Unit` was referring to the fact this thing is defined by a `unit` field in the json.
I don't really care about the name, but I find it more understandable to name this enum after the json spec than after what is generated with it (given the generation step is much more hidden)

oxidation/libparsec/crates/serialization_format/src/protocol.rs
touilleMan: There is no need to separate this type with the one with fields (otherwise we will need to handle the case of the json containing an empty `fields` field)

However I choose `Composed` name without too much thinking (I just wanted a "not unit" name), do you think it would be better named `Struct` ? or something else ?

oxidation/libparsec/crates/serialization_format/src/protocol.rs
touilleMan: :+1: I don't know why, but I find your solution much more satisfying :smile: 

done.

oxidation/libparsec/crates/serialization_format/src/protocol.rs
touilleMan: `discriminant_field` is not an iterator so I cannot do something like `#(#[serde(tag = #discriminant_field)])*`... do you have something else in mind ?

oxidation/libparsec/crates/serialization_format/src/protocol.rs
```diff
@@ -0,0 +1,739 @@
...
+                                                    discriminant_value: variant.discriminant_value.to_owned(),
+                                                    fields: variant.fields.as_ref().map(|fields| {
+                                                        fields.iter()
+                                                        .filter_map(&convert_field)
+                                                        .collect()
+                                                    }),
+                                                }
+                                            }
+                                        ).collect()
+                                    }
```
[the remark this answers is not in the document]
touilleMan: This code use the `convert_field` mut closure so it's not easy to move it

I'd rather keep the parsing work in a single place as much as possible:
- it's easier to follow the flow
- there no need for factorization since this code is only needed in a single place
- it simplifies providing context to `panic!/assert!` (since the parent structs are available). Otherwise we would have either less context in errors (e.g. not the name of the command that contains an error in it field) or more complex code (passing context info, or returning a Result type and having to deal with it in the caller place).

oxidation/libparsec/crates/serialization_format/src/protocol.rs
```diff
@@ -0,0 +1,739 @@
...
+                                },
+                                (Some(fields), None, None) => {
+                                    GenCmdNestedType::Struct {
+                                        name: nested_type.name.to_owned(),
+                                        fields: fields
+                                            .iter()
+                                            .filter_map(&convert_field)
+                                            .collect(),
+                                    }
```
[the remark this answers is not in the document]
touilleMan: same as above

oxidation/libparsec/crates/serialization_format/src/protocol.rs
touilleMan: Even better: merge `GenCmd::quote` into `quote_cmd` !

I don't know why I put a `GenCmd::quote`: having structs without methods is good given we divide here the code between 1) json structs 2) gen structs 3) json->gen conversion 4) code generation , so having each json/gen struct implementing it own subpart of the code would make much more spaghetti code

I've also replaced `GenCmdsFamilly::quote` by a `quote_cmds_familly` to stay consistent.

As a matter of fact I also pondering removing `GenCmdsFamilly::new`, but given it already contains all the json->gen conversion, turning it into a function would only gain removing a level of indentation.


Aaand finally I've removed the `pub(crate)` on most of the structure: external module only needs to know about `JsonCmd` (need given caller code is responsible for reading the json files and hence can deserialize each json file independently for better error reporting) and `generate_protocol_cmds_familly` (which wrap `GenCmdsFamilly::new/GenCmdsFamilly::quote`).

touilleMan:
> Nop keep the dict on the custom type, you don't have to revert to array :(

miniserde doesn't support arbitrary key values, so array must be used.

From what we talked, I understand your main worry for using array is the possibility to have a key duplicated by mistake (which by definition cannot happen using a dict).
However this error will be caught during the compilation of the generated code (given a nested type would be defined multiple time with the same name). This is of course less ideal than having a clean error message at parsing time but is "good enough".

I've updated the PR to add an assert check on nested types name unicity given it was really simple to add. However it would add significant complexity to the code to do a similar check for the fields (in req/reps and within nested types), so I stick to the "good enough" fail during generated code compilation for this part.

oxidation/libparsec/crates/serialization_format/src/protocol.rs
```diff
@@ -0,0 +1,739 @@
...
+                                }
+                                _ => {
+                                    panic!("{:?}::{:?}: Enum should have `variants`&`discriminant_field`, Struct should have `fields`", &cmd.req.cmd, &nested_type.name);
```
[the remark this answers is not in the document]
touilleMan: I modified this message.

Issue #3366, opened 2022-10-25
Rework `organization_config` to use an enum for `active_users_limit` instead of checking if the field is set or `null`


FirelightFlagboy:
The command `organization_config` have a field to set the `active_users_limit`.
Currently the field is nullable and when it's value is `null` that means `no limit`.
But we could do that in a more meaningful way by using an enum

```rust
enum ActiveUsersLimits {
    NoLimit,
    LimitedTo(Size),
}
```

PR #3792, opened 2022-12-27
Rework organization config active users limit by adding an enum

Volham22:
# What has changed ?

This PR is a proposal to add an enum as described in #3366 .


parsec/backend/memory/organization.py
```diff
@@ -233,6 +235,7 @@ async def update(
                 user_profile_outsider_allowed=user_profile_outsider_allowed
             )
 
+        assert type(organization.active_users_limit) == ActiveUsersLimit
```
TimeEngineer, line 238: isinstance ?

Volham22: Done.

parsec/backend/memory/organization.py
```diff
@@ -78,6 +78,8 @@ async def create(
                 self._config.organization_initial_user_profile_outsider_allowed
             )
 
+        assert type(active_users_limit) == ActiveUsersLimit
```
TimeEngineer, line 81: isinstance ?

parsec/backend/postgresql/organization.py
```diff
@@ -272,7 +272,7 @@ async def create(
                     *_q_insert_organization(
                         organization_id=id.str,
                         bootstrap_token=bootstrap_token,
-                        active_users_limit=active_users_limit,
+                        active_users_limit=active_users_limit.to_int(),
```
TimeEngineer, line 275: Should be handle by `postgres handler`

src/protocol/organization.rs
```diff
@@ -248,11 +290,11 @@ impl OrganizationConfigRepOk {
...
 
     #[getter]
-    fn active_users_limit(_self: PyRef<'_, Self>) -> PyResult<Option<u64>> {
-        Ok(match _self.as_ref().0 {
+    fn active_users_limit(_self: PyRef<'_, Self>) -> PyResult<ActiveUsersLimit> {
+        Ok(match &_self.as_ref().0 {
             organization_config::Rep::Ok {
                 active_users_limit, ..
-            } => active_users_limit,
+            } => ActiveUsersLimit(active_users_limit.clone()),
```
TimeEngineer, line 297: Maybe make a type apart in `libparsec_types`, it is used in `OrganizationConfig` (this is my current work)
then you can derive from Copy to avoid the heap

TimeEngineer (changes requested):

parsec/backend/postgresql/organization.py
Volham22: This means that we need to set a custom type handlers for `integer`. Can this create problems with other fields?

src/protocol/organization.rs
Volham22: This would create two types for `ActiveUsersLimit`? One which implements `Copy` and the other generated by the `serialization_format` macro. Is there a way to use the same type for both?

parsec/backend/postgresql/organization.py
TimeEngineer: The handler must catch any `integer type`

src/protocol/organization.rs
TimeEngineer: I mean maybe we should don't implement ActiveUsersLimit in `serialization_format`, but we can keep your work

TimeEngineer: This is weird if a type depends on `authenticated_cmds::v2::organization_config::ActiveUsersLimit`

parsec/backend/postgresql/organization.py
TimeEngineer: Or another solution is to change to `int | None` in `python world`, and this `enum` only exists in `rust world`

TimeEngineer: But it's never good to add logic to data

src/protocol/organization.rs
Volham22: What should we do? Remove `ActiveUsersLimit` from `serialization_format`? It make sense to me if `organization_config` which has a `user_limit` field embed an `ActiveUsersLimit` type.

TimeEngineer: Yes you can embed without `serialization_format`

parsec/backend/postgresql/handler.py
```diff
@@ -184,6 +184,23 @@ async def handle_uuid(conn: triopg._triopg.TrioConnectionProxy) -> None:
     )
 
 
+async def handle_active_users_limit(conn: triopg._triopg.TrioConnectionProxy) -> None:
```
TimeEngineer, line 187: The name is misleading, it handles all `integer` type

parsec/backend/postgresql/organization.py
TimeEngineer: Don't forget to remove `.to_int()`

TimeEngineer (changes requested):

TimeEngineer (approved):

Volham22:
We need to be careful when setting a custom type handler for `integer` to convert `ActiveUsersLimit`. I noticed that `encoder` (and `decoder`) are not supposed to return `None` (it may be handled differently by `asyncpg`) so we still need to check if the value is set to `NO_LIMIT` and set it to `None` by hand when performing an update or insert.

TimeEngineer:
Can you add a test that insert `int` in raw handler and retrieve with new handler and vice versa ? Like `uuid` and `datetime`

oxidation/libparsec/crates/serialization_format/src/protocol.rs
```diff
@@ -556,11 +560,21 @@ fn quote_custom_enum(
...
-                        #variant_name {
-                            #(#variant_fields),*
+                    // The goal of untagged is to avoid fields like `{ foo: FooType }`
+                    // We want the inner value to be transparent
+                    if is_untagged {
+                        let variants_types = quote_variants_types(variant_fields);
+                        quote! {
+                            #[serde(rename = #discriminant_value)]
+                            #variant_name (#(#variants_types),*)
+                        }
```
FirelightFlagboy, line 570: This actually generate a tuple variant

parsec/backend/postgresql/organization.py
```diff
@@ -276,7 +277,9 @@ async def create(
                     *_q_insert_organization(
                         organization_id=id.str,
                         bootstrap_token=bootstrap_token,
-                        active_users_limit=active_users_limit,
+                        active_users_limit=active_users_limit
+                        if active_users_limit is not ActiveUsersLimit.NO_LIMIT
+                        else None,
```
FirelightFlagboy, line 282: Should not this case handled by `handle_interger` ?

parsec/backend/postgresql/organization.py
```diff
@@ -328,7 +331,9 @@ async def _get(
             is_expired=row["is_expired"],
             created_on=row["created_on"],
             bootstrapped_on=row["bootstrapped_on"],
-            active_users_limit=row["active_users_limit"],
+            active_users_limit=ActiveUsersLimit.LimitedTo(row["active_users_limit"])
+            if row["active_users_limit"] is not None
+            else ActiveUsersLimit.NO_LIMIT,
```
FirelightFlagboy, line 336: Expose `ActiveUserLimit::from_optional_u64` (or something like that) ?

parsec/backend/postgresql/organization.py
```diff
@@ -438,7 +443,10 @@ async def update(
         if with_is_expired:
             fields["is_expired"] = is_expired
         if with_active_users_limit:
-            fields["active_users_limit"] = active_users_limit
+            assert isinstance(active_users_limit, ActiveUsersLimit)
+            fields["active_users_limit"] = (
+                active_users_limit if active_users_limit != ActiveUsersLimit.NO_LIMIT else None
+            )
```
FirelightFlagboy, line 449: use `to_int` ?

parsec/backend/postgresql/organization.py
Volham22: > We need to be careful when setting a custom type handler for integer to convert ActiveUsersLimit. I noticed that encoder (and decoder) are not supposed to return None (it may be handled differently by asyncpg) so we still need to check if the value is set to NO_LIMIT and set it to None by hand when performing an update or insert. I've added an assert for that.

We can't do that because `NO_LIMIT` is just `NULL` in the database. If our `encoder` returns `None` it will crash in `asyncpg` at some points.

parsec/backend/postgresql/organization.py
Volham22: I wanted to make it clear that `NO_LIMIT` is `NULL` in the pgsql database. What do you think?

Volham22: Yes, a call to `to_int` would work here

oxidation/libparsec/crates/serialization_format/src/protocol.rs
Volham22: I don't know if we should remove this code. At the very beginning I made `ActiveUsersLimit` an embedded type in schemas (we changed this here)[https://github.com/Scille/parsec-cloud/pull/3792#discussion_r1059427602].
It didn't break compatibility because an enum like this were generated.

```rs
#[derive(Serialize, Deserialize)]
#[serde(untagged)]
enum MyEnum {
    SomeValue(i32),
    Other,
}
// MyEnum::SomeValue(42) Serialize to "42" and MyEnum::Other gives `null`
```

Should we keep this or just remove it?

FirelightFlagboy: I would state for keeping the minimal amount of change In a PR

src/protocol/organization.rs
```diff
@@ -175,6 +176,53 @@ impl OrganizationConfigReq {
...
+    #[pyo3(name = "LimitedTo")]
+    fn limited_to(_cls: &PyType, user_count_limit: u64) -> Self {
+        Self(types::ActiveUsersLimit::LimitedTo(user_count_limit))
+    }
+
+    #[classattr]
+    #[pyo3(name = "NO_LIMIT")]
+    fn no_limit() -> &'static PyObject {
+        lazy_static::lazy_static! {
+            static ref VALUE: PyObject = {
```
FirelightFlagboy, line 195:
```suggestion
    fn no_limit() -> &'static Self {
        lazy_static::lazy_static! {
            static ref VALUE: Self = {
```

This can't be type to `Self` ?

src/protocol/organization.rs
```diff
@@ -175,6 +176,53 @@ impl OrganizationConfigReq {
...
+                })
+            };
+        };
+
+        &VALUE
+    }
+
+    #[classmethod]
+    #[pyo3(name = "FromOptionalInt")]
+    fn from_optional_int<'py>(cls: &'py PyType, py: Python<'py>, count: Option<u64>) -> &'py PyAny {
```
FirelightFlagboy, line 208:
```suggestion
    fn from_optional_int<'py>(cls: &'py PyType, py: Python<'py>, count: Option<u64>) -> &'py Self {
```

That way it would allow a more friendly typing for it

src/protocol/organization.rs
Volham22: We can't put `Self` here unfortunately because of `lazy_static`. 
```
error[E0401]: can't use generic parameters from outer function
   --> src/protocol/organization.rs:195:31
    |
194 | /         lazy_static::lazy_static! {
195 | |             static ref VALUE: Self = {
    | |                               ^^^^
    | |                               |
    | |                               use of generic parameter from outer function
    | |                               use a type here instead
196 | |                 Python::with_gil(|py| {
197 | |                     ActiveUsersLimit(types::ActiveUsersLimit::NoLimit)
198 | |                 })
199 | |             };
200 | |         };
    | |_________- `Self` type implicitly declared here, by this `impl`
```

Hardcoding the concrete type doesn't work either.
```
error[E0277]: the trait bound `&protocol::organization::ActiveUsersLimit: IntoPyCallbackOutput<_>` is not satisfied
```
We need to return a reference of a static attribute here otherwise the code will no longer be true in python. I think that's why all static attribute in `src/enumerate.rs` returns `&PyObject` (which implements `IntoPyCallbackOutput`).
```python
a = ActiveUsersLimit.NO_LIMIT
a is ActiveUsersLimit.NO_LIMIT # True
```

FirelightFlagboy: Maybe use `ActiveUsersLimit` instead of `Self` in the `lazy_static` block :thinking:

Volham22: > Hardcoding the concrete type doesn't work either.

Unfortunately it's not working either. 

```
error[E0277]: the trait bound `&protocol::organization::ActiveUsersLimit: IntoPyCallbackOutput<_>` is not satisfied
   --> src/protocol/organization.rs:183:1
    |
183 | #[pymethods]
    | ^^^^^^^^^^^^ the trait `IntoPyCallbackOutput<_>` is not implemented for `&protocol::organization::ActiveUsersLimit`
```

FirelightFlagboy: sad :'(

PR #3829, opened 2023-01-02
Remove legacy protocol stuff

TimeEngineer:
# What has changed ?

After migration, some protocol stuff are now unused


Volham22 (approved):

parsec/core/backend_connection/cmds.py
```diff
@@ -277,161 +276,84 @@ async def _send_cmd(
         transport.logger.exception("Invalid response data", cmd=cmd, error=exc)
         raise BackendProtocolError("Invalid response data") from exc
 
-    # Legacy
-    if isinstance(rep, dict):
-        if rep["status"] == "invalid_msg_format":
+    if isinstance(
```
touilleMan, line 279: I'd rather remove this big `if isinstance`: if we create a new command, this part will break silently (i.e. `MyNewCmdRepUnknowStatus(status='invalid_msg_format')` would be skipped instead of behind handled her)

A simple way to do that is:
```python
if getattr(rep, "status", None) == "invalid_msg_format":
    ...
```
(of course this would break if `status` doesn't correspond to the actual command status value, but this is totally unlikely we will introduce such thing ^^)

touilleMan (changes requested):

parsec/core/backend_connection/cmds.py
TimeEngineer: Done

FirelightFlagboy (approved):

touilleMan (approved):
👍

Issue #3830, opened 2023-01-02
Fix the client API reporting in the backend logs


vxgmichel:
Two things to fix from #3619
- specify that it is the client **API** version that we're reporting
- actually report the client API version instead of the server API version

FirelightFlagboy:
Close because #3831 have been merge, but auto close was miss configured ;)

PR #3831, opened 2023-01-02
Fix client API version reporting in the server logs

vxgmichel:
Fix issue #3830

### DIFF
--- a/tests/backend/test_ws_handshake.py
+++ b/tests/backend/test_ws_handshake.py
@@ -159,6 +159,8 @@ async def test_api_version_in_logs_on_handshake(backend_asgi_app, backend, alice
         invitation_type=InvitationType.USER,
         token=invitation.token,
     )
+    client_api_version = ApiVersion(3, 99)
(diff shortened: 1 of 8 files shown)


tests/backend/test_ws_handshake.py
```diff
@@ -174,7 +174,7 @@ async def test_api_version_in_logs_on_handshake(backend_asgi_app, backend, alice
             assert ch.client_api_version == API_VERSION
             assert ch.backend_api_version == API_VERSION
 
-        assert f"(client version: {API_VERSION})" in caplog.text
+        assert f"(client/server API version: {API_VERSION}/{API_VERSION})" in caplog.text
```
touilleMan, line 177: It would be better to have 2 different version for client and server

touilleMan, line 177: `test_rpc_handshake.py` should also be modified

touilleMan (changes requested):

touilleMan (approved):
lgtm

FirelightFlagboy (approved):

PR #3847, opened 2023-01-05
Fix pki enrollment spinner

### LINKED ISSUE #3846: [🐛 | Bug]: Spinner is visible during pki enrollment
### Version

2.15.0

Linux

### What happened ?

Spinner should not be visible while accepting a pki enrollment request. PKI enrollment was forgotten during the recent changes to the onboarding process.

### Relevant output

_No response_

Max-7:
# What has changed ?

Hide the spinner during enrollment by PKI.

Closes #3846

### DIFF
--- a/parsec/core/gui/enrollment_widget.py
+++ b/parsec/core/gui/enrollment_widget.py
@@ -44,7 +44,7 @@ class AcceptCheckInfoWidget(QWidget, Ui_GreetUserCheckInfoWidget):
         self.setupUi(self)
 
         self.dialog: GreyedDialog[AcceptCheckInfoWidget] | None = None
-        self.label_waiting.hide()
+        self.widget_waiting.setVisible(False)
 
         self.line_edit_user_full_name.validity_changed.connect(self.check_infos)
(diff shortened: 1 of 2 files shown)


parsec/core/gui/enrollment_widget.py
```diff
@@ -44,7 +44,7 @@ def __init__(
         self.setupUi(self)
 
         self.dialog: GreyedDialog[AcceptCheckInfoWidget] | None = None
-        self.label_waiting.hide()
```
FirelightFlagboy, line 47: What was this `label_waiting` ?

FirelightFlagboy (approved):

parsec/core/gui/enrollment_widget.py
Max-7: Before we re-did the onboarding process, the spinner was contained into a QLabel named `label_waiting`. This label is now wrapped into a QWidget named `widget_waiting`. `label_waiting` still exists so the code doesn't crash, but it's not doing what it's supposed to do.

FirelightFlagboy: ok, thanks

PR #3857, opened 2023-01-06
Drop APIV2 on client side

### LINKED ISSUE #3837: Drop support for APIv2 in client
Server must keep APIv2 support for the moment, but client can remove it which would simplify the behavior of the organization bootstrap

AntoineDupre:
# What has changed ?

Remove support for API V2 in the client side, closes #3837 

- [ x ] Does this PR affect the User ?


tests/backend/test_compatibility.py
```diff
@@ -1,5 +1,4 @@
 # Parsec Cloud (https://parsec.cloud) Copyright (c) AGPL-3.0 2016-present Scille SAS
-from __future__ import annotations
```
vxgmichel, line 2: Mistake?

vxgmichel (approved):

tests/backend/test_compatibility.py
AntoineDupre: oops

PR #3858, opened 2023-01-06
first commit

NicoTuxx:
# What has changed ?


oxidation/client/src/views/HomePage.vue
```diff
@@ -113,19 +166,27 @@ async function canDismissModal(): Promise<boolean> {
...
-  position: absolute;
-  left: 0;
-  right: 0;
-  top: 50%;
-  transform: translateY(-50%);
+  height: 100%;
+  justify-content: center;
+  display: flex;
+  flex-direction: column;
+  // position: absolute;
```
TimeEngineer, line 173: Remove comment

FirelightFlagboy:
Hello, could you provide a better name for the title of the PR ?

For the icons/images, shouldn't we use `git-lfs` (that would reduce the size of the git folder) ? @touilleMan

touilleMan:
@NicoTuxx what is this `oxidation/client/src/assets/images/Logo/__MACOSX` folder ?
the `__` as prefix plus the fact it only contains hidden (i.e. starting with a dot) files make it looks like macOS internals that should not be added to the project (like the infamous `.ds_store`)

touilleMan:
And any idea why github is complaining about the files in the `__MACOSX ` folder ? are they valid file that can be read on you computer ?
image

Issue #3862, opened 2023-01-06
Check is we have serialization compatibility test between python & rust


FirelightFlagboy:
Also if tests are missing, we should add them

PR #3884, opened 2023-01-09
Add Local device file's missing schemas and tests

### LINKED ISSUE #3861: Check is python schema are alike of the rust schema ?
Also add missing scheme.

> Can we convert a `legacy_device` to a local_device ?

- `PasswordDeviceFile`: File with data encrypted with a password
- `RecoveryDeviceFile`: File with data encrypted with a pass-phrase
- `SmartcardDeviceFile`: File with data encrypted using a smartcard

Volham22:
# What has changed ?

This PR add missing `LegacyDeviceFile` schema in rust implementations and its serde tests.

This closes #3861 and #3862.


oxidation/libparsec/crates/client_types/src/local_device_file.rs
```diff
@@ -101,6 +101,24 @@ impl DeviceFile {
...
+    pub human_handle: Option<HumanHandle>,
+    pub device_label: Option<DeviceLabel>,
+}
+
+#[derive(Debug, Clone, Serialize, Deserialize, PartialEq, Eq)]
+#[serde(rename_all = "lowercase")]
+#[serde(tag = "type")]
+pub enum LegacyDeviceFile {
+    Password(LegacyDeviceFilePassword),
+}
```
FirelightFlagboy, line 120: Since we have only 1 `LegacyDeviceFileType` can't we forget the enum ?

oxidation/libparsec/crates/client_types/src/local_device_file.rs
```diff
@@ -101,6 +101,24 @@ impl DeviceFile {
     }
 }
 
+#[serde_as]
+#[derive(Debug, Clone, Serialize, Deserialize, PartialEq, Eq)]
```
FirelightFlagboy, line 105: This struct is only use to read `LegacyDevice` file that then are converted to `DeviceFile`.
We don't need to serialize the struct

oxidation/libparsec/crates/client_types/tests/test_local_device.rs
```diff
@@ -232,3 +233,138 @@ fn slug(alice: &Device) {
...
+        "c7db221e94e620eb61943fbb5ed773ec9dd42879f82d69b01c1834eda4791d81f60254d11a"
+        "9a95a388c9df32630b7e5504ef361d55741c7ced33faf5249deba7b60417ed217411166954"
+        "feb70da66d7bf1442eae01e4f71648eefd46fb037e6a57482fdea63941a9c9159807c3b04d"
+        "b6d9e195942c96c7b9ae96d1af2b23919e8841be9c85d5305b96149f58e020abb3aff26413"
+        "70ecb2aeca9084fce184f78ff56a889c9ad80e8b59e441d86c82242e70261c31cde2ba8253"
+        "5a5667cac313c1fb8cb0107e2b1a1f6224559dc9d052be275865ac6465766963655f6c6162"
+        "656caf4d792064657631206d616368696e65ac68756d616e5f68616e646c6592b1616c6963"
+        "65406578616d706c652e636f6db2416c69636579204d63416c69636546616365a473616c74"
+        "c4106f48555e77fd45429cfe26d1dcdd3a8ea474797065a870617373776f7264"
+    )[..], LegacyDeviceFile::Password(LegacyDeviceFilePassword {
```
FirelightFlagboy, line 280:
```suggestion
    )[..], 
    LegacyDeviceFile::Password(LegacyDeviceFilePassword {
```
Missing newline after `,`

oxidation/libparsec/crates/client_types/src/local_device_file.rs
FirelightFlagboy: Since you're using `Serialize` in test for the roundtrip use
```suggestion
#[derive(Debug, Clone, Deserialize, PartialEq, Eq)]
#[cfg_attr(test, derive(Serialize))]
```

FirelightFlagboy (changes requested):

oxidation/libparsec/crates/client_types/src/local_device_file.rs
Volham22: I made this enum because of the field `type` which is always password in case of legacy device file.

FirelightFlagboy: Could you add a comment for that ?

Another way would be to add a field named `type` in `LegacyDeviceFilePassword` :thinking: 
But that way serde check if `type` is set to a correct value (i.e.: `type == 'password'`)

oxidation/libparsec/crates/client_types/src/local_device_file.rs
Volham22: I did it but we can't do these tests as `integration` tests because `cfg(test)` is only enabled when tests in the current crate are being compiled.

oxidation/libparsec/crates/client_types/src/local_device_file.rs
Volham22: Done.

oxidation/libparsec/crates/client_types/src/local_device_file.rs
Volham22: Done.

oxidation/libparsec/crates/client_types/src/lib.rs
```diff
@@ -5,6 +5,7 @@ mod local_device;
 mod local_device_file;
 mod local_manifest;
 mod path;
+mod test;
```
FirelightFlagboy, line 8:
```suggestion
#[cfg(test)]
mod test;
```
We only want to include this module during test

FirelightFlagboy (changes requested):

oxidation/libparsec/crates/client_types/src/lib.rs
Volham22: Done.

oxidation/libparsec/crates/client_types/src/test/test_local_device.rs
```diff
@@ -0,0 +1,145 @@
+// Parsec Cloud (https://parsec.cloud) Copyright (c) BUSL-1.1 (eventually AGPL-3.0) 2016-present Scille SAS
+
+mod test {
```
FirelightFlagboy, line 3: Remove the module itself

FirelightFlagboy (changes requested):

FirelightFlagboy (approved):

PR #3892, opened 2023-01-11
Oxidation login

### LINKED ISSUE #3843: Login / "LoggedCore" save and load (bindings)

TimeEngineer:
# What has changed ?

- Implement a way to login and keep all rust stuff in memory

closes #3843

### DIFF
--- a/oxidation/bindings/electron/src/index.d.ts
+++ b/oxidation/bindings/electron/src/index.d.ts
@@ -37,4 +37,19 @@ export type DeviceFileType =
   | DeviceFileTypeSmartcard
 
 
+// LoggedCoreError
+export interface LoggedCoreErrorDisconnected {
+    tag: "Disconnected"
+}
+export interface LoggedCoreErrorInvalidHandle {
+    tag: "InvalidHandle"
+    handle: number;
+}
+export type LoggedCoreError =
+  | LoggedCoreErrorDisconnected
+  | LoggedCoreErrorInvalidHandle
+
+
 export function listAvailableDevices(path: string): Promise<Array<AvailableDevice>>;
+export function login(test_device_id: string): Promise<number>;
+export function loggedCoreGetTestDeviceId(handle: number): Promise<Result<string, LoggedCoreError>>;
--- a/oxidation/libparsec/crates/core/Cargo.toml
+++ b/oxidation/libparsec/crates/core/Cargo.toml
@@ -20,7 +20,9 @@ use-sodiumoxide = ["libparsec_crypto/use-sodiumoxide"]
 libparsec_crypto = { path = "../crypto", default-features = false }
 libparsec_protocol = { path = "../protocol", default-features = false }
 libparsec_types = { path = "../types", default-features = false }
+libparsec_platform_async = { path = "../platform_async" }
 
+once_cell = "1.17.0"
 rmp-serde = "1.1.1"
 serde = { version = "1.0.147", features = ["derive"] }
 serde_with = "2.1.0"
--- a/oxidation/libparsec/crates/core/src/error.rs
+++ b/oxidation/libparsec/crates/core/src/error.rs
@@ -4,7 +4,7 @@ use thiserror::Error;
 
 use libparsec_types::{DateTime, DeviceID, UserID};
 
-use crate::DeviceHandle;
+use crate::LoggedCoreHandle;
 
 #[derive(Error, Debug, PartialEq, Eq)]
 pub enum TrustchainError {
@@ -80,7 +80,7 @@ impl From<TrustchainError> for RemoteDevicesManagerError {
 #[derive(Error, Debug, PartialEq, Eq)]
 pub enum LoggedCoreError {
     #[error("The handle provided is invalid: {handle:?}")]
-    InvalidHandle { handle: DeviceHandle },
+    InvalidHandle { handle: LoggedCoreHandle },
     #[error("The device is disconnected")]
     Disconnected,
 }
--- a/oxidation/libparsec/crates/core/src/error.rs
+++ b/oxidation/libparsec/crates/core/src/error.rs
@@ -4,6 +4,8 @@ use thiserror::Error;
 
 use libparsec_types::{DateTime, DeviceID, UserID};
 
+use crate::DeviceHandle;
+
 #[derive(Error, Debug, PartialEq, Eq)]
 pub enum TrustchainError {
     #[error("{path}: Invalid certificate: {exc}")]
@@ -74,3 +76,13 @@ impl From<TrustchainError> for RemoteDevicesManagerError {
         Self::InvalidTrustchain { exc }
     }
 }
+
+#[derive(Error, Debug, PartialEq, Eq)]
+pub enum LoggedCoreError {
+    #[error("The handle provided is invalid: {handle:?}")]
+    InvalidHandle { handle: DeviceHandle },
+    #[error("The device is disconnected")]
+    Disconnected,
+}
+
+pub type LoggedCoreResult<T> = Result<T, LoggedCoreError>;
--- a/oxidation/libparsec/crates/core/src/logged_core.rs
+++ b/oxidation/libparsec/crates/core/src/logged_core.rs
@@ -1,0 +1,81 @@
+// Parsec Cloud (https://parsec.cloud) Copyright (c) BUSL-1.1 (eventually AGPL-3.0) 2016-present Scille SAS
+
+use once_cell::sync::OnceCell;
+
+use libparsec_platform_async::Mutex;
+use libparsec_types::DeviceID;
+
+use crate::{LoggedCoreError, LoggedCoreResult};
+
+static CORE: OnceCell<Mutex<Vec<Option<LoggedCore>>>> = OnceCell::new();
+
+/// i32 is because javascript number is a 64 bits float
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub struct DeviceHandle(i32);
+
+impl From<i32> for DeviceHandle {
+    fn from(value: i32) -> Self {
+        Self(value)
+    }
+}
+
+impl From<DeviceHandle> for i32 {
+    fn from(value: DeviceHandle) -> Self {
+        value.0
+    }
+}
+
+impl From<DeviceHandle> for usize {
+    fn from(value: DeviceHandle) -> Self {
+        value.0 as usize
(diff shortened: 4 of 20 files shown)


FirelightFlagboy:
> `i64` should be forbidden while javascript doesn't support it
Javascript have `bitint` type.
I've looked at the crate `wasm-bindgen` How they handle converting integer type:

- `i64` & `u64` are using `bitint`
- the other are converted to `u32`/`i32`

oxidation/libparsec/crates/client_types/src/logged_core.rs
```diff
@@ -0,0 +1,53 @@
...
+}
+
+// We don't need to hash everything !
+impl Hash for LoggedCore {
+    fn hash<H: Hasher>(&self, state: &mut H) {
+        self.test_device_id.hash(state)
+    }
+}
+
+/// 32 bits of signature seems to be enough, people don't have that much devices
```
touilleMan, line 27: I'm not super happy to have JavaScript stuff leaking into the pure rust :(

I think we should add a max check in the bindings to raise exception if the returned value is too big

This way we can still use whatever value we want here

Regarding for what value we want here, I'd rather use an unsigned value

oxidation/libparsec/crates/client_types/src/logged_core.rs
```diff
@@ -0,0 +1,53 @@
...
+/// Output: i32 is because javascript number is a 64 bits float
+pub fn login(test_device_id: DeviceID) -> i32 {
+    let core = LoggedCore { test_device_id };
+
+    // Make a signature for LoggedCore
+    let mut hasher = DefaultHasher::new();
+    core.hash(&mut hasher);
+    let key = hasher.finish() as i32;
```
touilleMan, line 35: There is no guarantee the hash is unique, and the fact we cut the hash to 32bits increase this risk.

Instead of doing a hash, use a global vector of Option<LoggedCore>.
On logging you iterate over the vector to check the device id is not already logged, then push the new LoggedCore.
This way your id is the index in the vector which is super convenient.
For log out you just replace the element in the vector by None.

This means the vector only keep growing (so you never reuse an id which is good to avoid hard to track bugs), and the impact is totally negligible on memory usage (one new element in the vector per login)

oxidation/libparsec/crates/client_types/src/logged_core.rs
```diff
@@ -0,0 +1,53 @@
...
+// We don't need to hash everything !
+impl Hash for LoggedCore {
+    fn hash<H: Hasher>(&self, state: &mut H) {
+        self.test_device_id.hash(state)
+    }
+}
+
+/// 32 bits of signature seems to be enough, people don't have that much devices
+/// Output: i32 is because javascript number is a 64 bits float
+pub fn login(test_device_id: DeviceID) -> i32 {
```
touilleMan, line 29: The function should be async, especially given it will take a mutex (mutex that itself must be async of course)

touilleMan (changes requested):

oxidation/libparsec/crates/client_types/src/logged_core.rs
FirelightFlagboy: Would it be OK, to mark every function that will interact with the `database`, `filesystem` & `network` to be marked as `async` (even if it's only for testing) ?

oxidation/bindings/electron/src/index.d.ts
```diff
@@ -38,3 +38,5 @@ export type DeviceFileType =
 
 
 export function listAvailableDevices(path: string): Promise<Array<AvailableDevice>>;
+export function login(test_device_id: string): Promise<number>;
```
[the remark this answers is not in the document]
TimeEngineer: Because `device_id` is not part of `LoggedCore`

TimeEngineer:
> > `i64` should be forbidden while javascript doesn't support it
> > Javascript have `bitint` type.
> > I've looked at the crate `wasm-bindgen` How they handle converting integer type:
> 
> * `i64` & `u64` are using `bitint`
> * the other are converted to `u32`/`i32`

`bigint` isn't primitive, so any operation related with will slow down drastically: (conversion, index, etc...)

oxidation/bindings/electron/src/index.d.ts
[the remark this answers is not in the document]
TimeEngineer: I can't handle error if there is no error for the moment :/

oxidation/libparsec/crates/client_types/src/logged_core.rs
TimeEngineer: So what's the equivalent of uint in python ?
Because I have to add in api.py

oxidation/libparsec/crates/client_types/src/logged_core.rs
TimeEngineer: Done

oxidation/bindings/electron/src/index.d.ts
```diff
@@ -37,4 +37,19 @@ export type DeviceFileType =
...
+    tag: "InvalidIndex"
+    index: number;
+}
+export type LoggedCoreError =
+  | LoggedCoreErrorDisconnected
+  | LoggedCoreErrorInvalidIndex
+
+
 export function listAvailableDevices(path: string): Promise<Array<AvailableDevice>>;
+export function login(test_device_id: string): Promise<number>;
```
FirelightFlagboy, line 54: Can this function be typed like so ?

```suggestion
export async function login(test_device_id: string): number;
```

oxidation/libparsec/crates/core/src/error.rs
```diff
@@ -74,3 +74,13 @@ impl From<TrustchainError> for RemoteDevicesManagerError {
         Self::InvalidTrustchain { exc }
     }
 }
+
+#[derive(Error, Debug, PartialEq, Eq)]
+pub enum LoggedCoreError {
+    #[error("The index provided is invalid: {index}")]
```
touilleMan, line 80: Rename `index` by `handle`: while implemented as an index, this is not relevant for the user.

"Handle" on the other hand designed an opaque id that is provided to track a resource.

oxidation/bindings/electron/src/index.d.ts
TimeEngineer: Seems `js dev` use `Promise` in `vue`

oxidation/libparsec/crates/core/Cargo.toml
```diff
@@ -21,11 +21,14 @@ libparsec_crypto = { path = "../crypto", default-features = false }
 libparsec_protocol = { path = "../protocol", default-features = false }
 libparsec_types = { path = "../types", default-features = false }
 
+futures = "0.3.5"
```
touilleMan, line 24: Move this dependenc to `platform_async` and re-expose it from there.

This is needed to ensure we use a consistent set of primitive for async operations (instead of having each crate needing this stuff to wonder between futures, futures_lite, smol etc.)

oxidation/libparsec/crates/core/src/logged_core.rs
```diff
@@ -0,0 +1,60 @@
...
+static CORE: OnceCell<Mutex<Vec<Option<LoggedCore>>>> = OnceCell::new();
+
+#[derive(Debug, Clone)]
+struct LoggedCore {
+    // TODO: Remove me when LoggedCore is implemented
+    test_device_id: DeviceID,
+}
+
+/// Output: i32 is because javascript number is a 64 bits float
+pub async fn login(test_device_id: DeviceID) -> i32 {
```
touilleMan, line 19: Returning a naked `i32` is bad for typing.
use a struct wrapping the i32 with from/to conversion instead.

oxidation/libparsec/crates/core/src/error.rs
```diff
@@ -74,3 +74,13 @@ impl From<TrustchainError> for RemoteDevicesManagerError {
         Self::InvalidTrustchain { exc }
     }
 }
+
+#[derive(Error, Debug, PartialEq, Eq)]
+pub enum LoggedCoreError {
```
touilleMan, line 79: This is fine for now, but in the long run we should probably have an error enum per function, this in order to have precisely known what error can those high-level interface return (which is a lot simpler for the GUI than to have a function that return a generic error type, then not knowing what error should be handled)

oxidation/bindings/electron/src/index.d.ts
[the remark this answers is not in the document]
touilleMan: ok, we'll modify this function signature later when it has an actual implemenation

touilleMan (changes requested):

oxidation/bindings/electron/src/index.d.ts
touilleMan: from what I understand, ts requires that an async function returns a Promise (cf. https://github.com/Scille/parsec-cloud/blob/90ad6abc6e6c4810d43a6ab75eb101ba6d7ea4fa/oxidation/client/src/plugins/libparsec/trampoline-web.ts.in#L1)

Not sure if there is a difference between `async function foo(): Promise<number>` and `function foo(): Promise<number>` though...

oxidation/libparsec/crates/client_types/src/logged_core.rs
touilleMan: > Would it be OK (more like can we), to mark every function that will interact with the database, filesystem & network to be marked as async (even if it's only for testing) ?

I think you're right on this: basically all function exposed to js should be async, so better doing this by default even if async is not stricly needed for the moment because the function body is just a mock ^^

oxidation/bindings/electron/src/index.d.ts
FirelightFlagboy: :shrug:

oxidation/libparsec/crates/core/src/error.rs
TimeEngineer: Done

oxidation/libparsec/crates/core/src/logged_core.rs
TimeEngineer: Done

oxidation/libparsec/crates/core/Cargo.toml
TimeEngineer: Done

touilleMan (approved):

PR #3895, opened 2023-01-11
Implement manifest and optional block remanence for workspaces

### LINKED ISSUE #2266: "Available when offline" workspace mode
At the moment, Parsec is lazy when it comes to download file data, meaning a specific piece of data is not going to be downloaded until the user tries to access it. The idea of the "available when offline" feature is to let the user select some specific data to be kept locally so this data is guaranteed to be available when offline.

### 1 - The feature

First, we should decide on the scope of the feature. More specifically, should the user be able to:
- Select a full workspace to be kept in offline mode
- Select a specific file or directory to be kept in offline mode

The second approach is more granular and require a greater amount of work, for a couple of reasons:
- The tree structure of file system in a workspace is dynamic
- The user interface is also more complex as it should be made clear to the user as which directory is tagged as offline or not.

For the rest of this write-up, I'll focus on the first approach, more specifically:
- The user can select a workspace to be switched to "available when offline"
- Follow a transition period during which the corresponding data is downloaded and stored locally
- The current status of the workspace is displayed to the user
- Once the transition period is over, the user can fully access every file in the workspace even when offline
- The "available when offline" doesn't prevent synchronization (up and down) but the guarantees should remain at all times
- The user can disable the "available when offline" mode to clean up space on the disk

Extra features might be added on top of this, which will not be considered in this write up:
- Warn the user of the amount of data which is required to download when switching modes
- Warn the user of the amount of data which is required to download when down-synchronizing a large files

### 2 - The implementation

From the local storage point of view, there are two ways of implementing this feature:
- either create a third database dedicated to offline mode (the first two being the cache database and the local data database)
- or alter the design of the cache database to disable the cache limit

As of now, it's difficult to tell which one is the cheapest or best solution, this should be investigated further.

Then a state machine should be implemented at workspace level to manage the transitions between the following states:
- Normal state
- Transition to "available when offline" state
- Available when offline state

In order to maintain the "available when offline" state, the workspace cannot down-synchronize manifests without performing the following task:
- Check that the manifest is a file manifest
- List the new block IDs that are part of this manifest
- Download the corresponding blocks and store them in the local storage
- And only then, acknowledge this manifest

Note: this should also be done to the transitional state, so that a single pass over all the file manifests at a given moment should be enough to guaranteed the transition to the stable state.

The synchronization system being highly asynchronous, a major challenge of this implementation is to make sure that operations described above work consistently even other operations are carried out in the meantime. Those operations being:
- User read access to the file system: seems trivial
- User write access to the file system: seems trivial
- Down-synchronization: those operations might be redundant, if two versions of a manifest are uploaded close to each-other
- Up-synchronization: this might trigger a down-synchronization which loops to the previous point

Another challenge is the implementation of a cleanup routine, since we currently do not remove blocks when they're not referenced anymore (instead we simply rely on the cache and its "remove the oldest block" policy). There are two approaches to cleaning-up those unreferenced blocks:
- Have a periodic cleanup routine, relying on a workspace sweep to list all referenced block ids. Note: Race conditions can easily be avoided by comparing the last access time of the blocks to the start time of the sweep.
- Remove the old blocks as soon as they are not referenced by comparing manifests AFTER an down-synchronization has been performed. Note: This only work under the assumption that blocks cannot be referenced more than once

A hybrid approach might also have advantages in order to clean up unreferenced blocks that might remain in the database, in the case where the application has been stopped after acknowledging a manifest but before performing the clean up for instance. The routine might also be able to identify missing blocks preemptively, in order to provide more robustness to the system.

The last item that needs to done is the exposing of those feature in the desktop GUI in an ergonomic way.

vxgmichel:
Fix #2266

With this PR a new monitor is introduced, responsible for detecting when metadata (and optionally data) is missing locally.

At the moment, the implementation works as the following:
- During the first run of the monitor, a full sweep of the workspaces is performed in order to make sure all the referenced metadata is present
- Afterward, every downsync operation registers a downsync job that is meant to be picked up by the monitor in order to download any missing information referenced by the newly acknowledged manifest.


parsec/core/gui/tr/parsec_fr.po
```diff
@@ -3185,3 +3185,48 @@ msgstr "Masquer les espaces de travail désactivés"
...
+msgstr "Un maximum de {cache_size} de données sont actuellement conservées localement pour cet espace de travail."
+
+msgid "TEXT_WORKSPACE_SIZE_local_size_total_size"
+msgstr "{local_size} / {total_size} de données sont disponibles localement."
+
+msgid "TEXT_BLOCK_REMANENCE_WARNING_remote_size"
+msgstr "Conserver les données localement pour un accès hors-ligne nécéssiterait de télécharger {remote_size} de données."
+
+msgid "TEXT_GET_REMANENCE_STATUS_FAILED"
+msgstr "Les information liées à la gestion des données dans cet espace de travail sont actuallement indisponibles."
```
Max-7, line 3214: `les informationS`, `actuEllement`

Max-7 (changes requested):

parsec/core/gui/tr/parsec_en.po
```diff
@@ -3105,3 +3105,48 @@ msgstr "Hide deactivated workspaces"
...
+msgstr "An error occured while changing the data management policy."
+
+msgid "TEXT_ENABLE_BLOCK_REMANENCE_FAILED"
+msgstr "An error occured while changing the data management policy."
+
+msgid "TEXT_REMANENCE_DIALOG_TITLE_workspace_name"
+msgstr "Data management for workspace {workspace_name}"
+
+msgid "TEXT_ENABLE_BLOCK_REMANENCE_SUCCESS"
+msgstr "The data in this workspace will be now kept locally for offline access."
```
Max-7, line 3146: `will not be kept`

parsec/core/gui/tr/parsec_fr.po
```diff
@@ -3185,3 +3185,48 @@ msgstr "Masquer les espaces de travail désactivés"
...
+msgstr "Les données de cet espace de travail sont actuellement conservées localement pour un accès hors-ligne."
+
+msgid "TEXT_BLOCK_REMANENCE_DISABLED_cache_size"
+msgstr "Un maximum de {cache_size} de données sont actuellement conservées localement pour cet espace de travail."
+
+msgid "TEXT_WORKSPACE_SIZE_local_size_total_size"
+msgstr "{local_size} / {total_size} de données sont disponibles localement."
+
+msgid "TEXT_BLOCK_REMANENCE_WARNING_remote_size"
+msgstr "Conserver les données localement pour un accès hors-ligne nécéssiterait de télécharger {remote_size} de données."
```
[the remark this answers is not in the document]
vxgmichel: 8cb29260f7c9dc35c0b347b22db0cbf02202ac5e

parsec/core/gui/tr/parsec_fr.po
vxgmichel: 8cb29260f7c9dc35c0b347b22db0cbf02202ac5e

parsec/core/gui/tr/parsec_en.po
vxgmichel: 8cb29260f7c9dc35c0b347b22db0cbf02202ac5e

Max-7 (approved):
All things related to GUI are fine, and it does work!

parsec/core/fs/storage/local_database.py
```diff
@@ -157,6 +170,15 @@ async def _create_connection(self) -> None:
...
             self._conn.execute("PRAGMA synchronous=NORMAL")
 
+            # Manage auto_vacuum
+            cursor = self._conn.execute("PRAGMA auto_vacuum")
+            rep = cursor.fetchone()
+            current_auto_vacuum = rep and rep[0]
+
+            if self.auto_vacuum != current_auto_vacuum:
+                self._conn.execute(f"PRAGMA auto_vacuum={int(self.auto_vacuum)}")
+                self._conn.execute("VACUUM")
```
vxgmichel, line 180: Add comments

vxgmichel: this should run in a thread

parsec/core/fs/storage/local_database.py
```diff
@@ -20,15 +20,25 @@ class LocalDatabase:
     # Make the trio run_sync function patchable for the tests
     run_in_thread = staticmethod(trio.to_thread.run_sync)
 
-    def __init__(self, path: Union[str, Path, trio.Path], vacuum_threshold: int | None = None):
+    def __init__(
```
touilleMan, line 23: Add a comment to explain why autovacuum and vacuum_threshold are both present

oxidation/libparsec/crates/core/src/events.rs
```diff
@@ -66,6 +68,8 @@ impl CoreEvent {
             CoreEvent::FsEntrySyncRejectedBySequesterService => {
                 "fs.entry.sync_refused_by_sequester_service"
             }
+            CoreEvent::FsBlockDownloaded => "fs.block.downloaded",
+            CoreEvent::FsBlockRemoved => "fs.block.removed",
```
touilleMan, line 72: rename to ```FsBlockCacheRemove => `fs.block.cache_removed` ```

parsec/core/fs/storage/chunk_storage.py
```diff
@@ -253,18 +266,26 @@ async def cleanup(self, cursor: Cursor | None = None) -> None:
...
-                cursor.execute,
-                """
-                DELETE FROM chunks WHERE chunk_id IN (
-                    SELECT chunk_id FROM chunks ORDER BY accessed_on ASC LIMIT ?
+            def _thread_target(cursor: Cursor) -> list[ChunkID]:
+                cursor.execute(
+                    """
+                    DELETE FROM chunks WHERE chunk_id IN (
+                        SELECT chunk_id FROM chunks ORDER BY accessed_on ASC LIMIT ?
+                    ) RETURNING chunk_id
```
touilleMan: Add a comment to indicate this cool returning is available when re-writing the code in rust (since we ship a more recent version of sqlite)

parsec/core/fs/storage/chunk_storage.py
```diff
@@ -289,3 +302,96 @@ def _thread_target(cursor: Cursor) -> list[ChunkID]:
...
+                    VALUES (0, 1)""",
+            )
+
+            # Set remanent flag
+            self._block_remanent = True
+
+        # Indicate that the state has changed
+        return True
+
+    async def disable_block_remanence(self) -> list[ChunkID] | None:
```
touilleMan, line 349: Should add a comment to explicit the return value

parsec/core/fs/storage/chunk_storage.py
```diff
@@ -289,3 +302,96 @@ def _thread_target(cursor: Cursor) -> list[ChunkID]:
...
+            )
+
+            # Set remanent flag
+            self._block_remanent = False
+
+            # Cleanup blocks and indicate that the state has changed
+            return await self.cleanup(cursor)
+
+    async def clear_unreferenced_chunks(
+        self, chunk_ids: list[ChunkID], not_accessed_after: float
```
touilleMan, line 371: Add comment to explain `not_accessed_after`, this is pretty weird to have a timestamp as float instead of DateTime

parsec/core/fs/userfs/userfs.py
```diff
@@ -430,28 +430,33 @@ async def workspace_storage_task(
...
-            get_previous_workspace_entry=get_previous_workspace_entry,
-            device=self.device,
-            local_storage=local_storage,
-            backend_cmds=self.backend_cmds,
-            event_bus=self.event_bus,
-            remote_devices_manager=self.remote_devices_manager,
-            preferred_language=self.preferred_language,
-        )
+                # Connect remanence manager events
+                with workspace.remanence_manager.manage_events():
```
vxgmichel, line 448: Instanciate before WorkspaceFS with a `run` context

tests/conftest.py
```diff
@@ -501,6 +501,22 @@ async def _reset_testbed(keep_logs=False):
     return _reset_testbed
 
 
+# Pause the remanent monitor by default
+
+
+@pytest.fixture(autouse=True)
```
vxgmichel, line 507: Investigate a change of default

parsec/core/fs/workspacefs/sync_transactions.py
```diff
@@ -45,6 +47,39 @@
...
+    added_blocks: set[BlockAccess] | None = None
+    removed_blocks: set[BlockAccess] | None = None
+    added_entries: set[EntryID] | None = None
+    removed_entries: set[EntryID] | None = None
+
+    @classmethod
+    def from_manifests(
+        cls,
+        old_manifest: AnyRemoteManifest,
+        new_manifest: AnyRemoteManifest,
```
vxgmichel, line 64: Use a TypeVar

tests/core/test_remanence_monitor.py
```diff
@@ -0,0 +1,373 @@
...
+    monkeypatch,
+):
+    # Go fast
+    alice_core.device.time_provider.mock_time(speed=1000.0)
+    alice2_core.device.time_provider.mock_time(speed=1000.0)
+    monkeypatch.setattr("parsec.utils.BALLPARK_ALWAYS_OK", True)
+
+    # Create a workspace
+    wid = await alice_core.user_fs.workspace_create(EntryName("w"))
+    await alice_core.user_fs.sync()
```
vxgmichel, line 160: alice2_core sync

parsec/core/remanence_monitor.py
```diff
@@ -0,0 +1,101 @@
...
+    monitor coroutine running in background
+    """
+
+
+async def monitor_remanent_workspaces(
+    user_fs: UserFS, event_bus: EventBus, task_status: MonitorTaskStatus
+) -> None:
+    on_going_tasks: set[tuple[EntryID, int]] = set()
+
+    def start_downsync_manager(workspace_id: EntryID) -> None:
```
vxgmichel, line 31: remanence

parsec/core/remanence_monitor.py
```diff
@@ -0,0 +1,101 @@
...
+        # Make sure the workspace is available
+        try:
+            workspace_fs = user_fs.get_workspace(workspace_id)
+        except FSWorkspaceNotFoundError:
+            return
+
+        def idle(task_id: int) -> None:
+            assert task_id in (1, 2)
```
vxgmichel, line 39: use an enum

parsec/core/fs/workspacefs/sync_transactions.py
vxgmichel: Too painful as it would require to type more stuff more strictly in sync_transactions

PR #3921, opened 2023-01-18
Allow to open database in memory for faster test

FirelightFlagboy:


oxidation/libparsec/crates/platform_local_db/src/lib.rs
```diff
@@ -55,12 +46,37 @@ impl LocalDatabase {
...
+                prefix
+                    .to_str()
+                    .expect("We generate the Path from a `&str` so UTF-8 is already checked")
+            ))
+        })?;
+    }
+    let mut connection = SqliteConnection::establish(path)?;
+    connection.batch_execute(
+        "
+                    PRAGMA journal_mode = WAL; -- better write-concurrency
```
TimeEngineer, line 73: Indentation

FirelightFlagboy: done

TimeEngineer (approved):

oxidation/libparsec/crates/platform_local_db/src/lib.rs
```diff
@@ -22,35 +22,33 @@ pub const LOCAL_DATABASE_MAX_VARIABLE_NUMBER: usize = 999;
...
+            None => {
+                let conn = new_sqlite_connection_from_path(path).await?;
+                SqliteExecutor::spawn(conn)
+            }
+        };
+        Ok(Self {
+            path: path.to_string(),
+            executor: Arc::new(RwLock::new(Some(executor))),
+        })
     }
```
FirelightFlagboy, line 43: @touilleMan We could have keep `new_in_memory`, that would prevent adding `rstest` + `tests_fixtures` in the devDeps

oxidation/libparsec/crates/platform_local_db/src/test_utils.rs
```diff
@@ -1,63 +1,128 @@
...
+                        ClosableInMemoryDB::Opened => {
+                            // The database is *currently* used, this is something we don't
+                            // do in Parsec (we have a single process running that have no
+                            // benefit at connecting multiple times to a database).
+                            // This is most likely in a bug (typically trying to start two
+                            // client with the same device).
+                            panic!(
+                                "Trying to open already opened in memory database `{}`",
+                                path
+                            )
```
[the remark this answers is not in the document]
touilleMan: That's a good point, I'll test that !

oxidation/libparsec/crates/platform_local_db/src/lib.rs
touilleMan: I think it's better not to use the in memory mock when testing this crate:
- we are testing the local database (and not using it as part of a higher level component testing) so it's better to test what is going to run in production
-  the perf impact is minimal since the test is very simple

oxidation/libparsec/crates/platform_local_db/src/test_utils.rs
```diff
@@ -1,63 +1,128 @@
...
-pub fn clear_local_db_in_memory() {
-    LOCAL_DB_IN_MEMORY
-        .lock()
-        .expect("Mutex is poisoned")
-        .clear();
+/// Note: Already opened connections remain unchanged.
+pub fn test_toggle_local_db_in_memory_mock(enabled: bool) {
+    let mut strategy = TEST_DB_STRATEGY.lock().expect("Mutex is poisoned");
+    match (enabled, &mut *strategy) {
+        (false, s @ TestDBStrategy::InMemory(_)) => {
```
TimeEngineer, line 106: You don't need `s @` if you don't match a function call

TimeEngineer (changes requested):

oxidation/libparsec/crates/platform_local_db/src/test_utils.rs
touilleMan: I need to first match that strategy contains a InMemory, then change the value of strategy, you mean there is a way to do it without using `s @`

src/runtime.rs
```diff
@@ -130,7 +131,17 @@ impl FutureIntoCoroutine {
...
 
             // Actual run of the Tokio future
-            let ret = fut.await;
+            let ret = std::panic::AssertUnwindSafe(fut).catch_unwind().await;
+            let ret = ret.unwrap_or_else(|panic_err| {
+                let msg = match panic_err.downcast::<&str>() {
+                    Ok(msg) => (*msg).to_owned(),
+                    Err(panic_err) => match panic_err.downcast::<String>() {
+                        Ok(msg) => *msg,
+                        Err(_) => "Unknown error".to_owned(),
```
FirelightFlagboy, line 140: I don't think we would fall on this step often, but providing additional information would be useful if it's the case :thinking: 

```suggestion
                        Err(_) => format!("Unknown error type {:?}", panic_err.type_id())
```

src/runtime.rs
```diff
@@ -130,7 +131,17 @@ impl FutureIntoCoroutine {
             // Here we have left the trio thread and are inside a thread provided by the Tokio runtime
 
             // Actual run of the Tokio future
-            let ret = fut.await;
+            let ret = std::panic::AssertUnwindSafe(fut).catch_unwind().await;
+            let ret = ret.unwrap_or_else(|panic_err| {
+                let msg = match panic_err.downcast::<&str>() {
+                    Ok(msg) => (*msg).to_owned(),
```
FirelightFlagboy, line 137:
```suggestion
                    Ok(msg) => msg.to_string(),
```

I'm a `to_string` guy

tests/conftest.py
```diff
@@ -435,6 +435,28 @@ def blockstore(backend_store, fixtures_customization):
     return config
 
 
+@pytest.fixture(autouse=True)
+def persistent_mockup(
+    fixtures_customization: dict[str, Any]
+) -> Generator[None, None, Callable[[], None] | None]:
+    if fixtures_customization.get("real_data_storage", False) or fixtures_customization.get(
```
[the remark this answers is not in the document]
FirelightFlagboy: No we can't.
`real_data_storage` is used by `persistent_mockup` it is used to store the data on disk during test.
`alternate_workspace_storage` is used to test the new rust impl with the old python one in a alternate manner but to do that we need to have the data stored on disk

tests/conftest.py
```diff
@@ -435,6 +436,33 @@ def blockstore(backend_store, fixtures_customization):
...
+        test_clear_local_db_in_memory_mock()
+
+        yield test_clear_local_db_in_memory_mock
+
+
+# `persistent_mockup` is autouse, hence a test only have it explicitly has dependency to be
+# able to use it clear function. Here we just re-expose this function with a better name ;-)
+@pytest.fixture
+def clear_persistent_mockup(persistent_mockup):
+    return persistent_mockup or (lambda: None)
```
FirelightFlagboy, line 463: Would not it be simple to just rename `persistent_mock` to `clear_persistent_mock` ?
(and make it always return a Closure)

tests/core/fs/conftest.py
```diff
@@ -145,6 +145,7 @@ def user_fs_offline_state_machine(
     monkeypatch: pytest.MonkeyPatch,
     user_fs_factory: UserFsFactory,
     clear_database_dir: Callable[[bool], None],
+    clear_persistent_mockup: Callable[[], None] | None,
```
FirelightFlagboy, line 148:
```suggestion
    clear_persistent_mockup: Callable[[], None],
```
With your fixture, it always return a closure now ;)

oxidation/libparsec/crates/platform_local_db/src/test_utils.rs
```diff
@@ -0,0 +1,63 @@
...
+
+    Some(res.map(|executor| LocalDatabase {
+        executor: Arc::new(RwLock::new(Some(executor))),
+    }))
+}
+
+/// Enable or not to have [LocalDatabase] in memory.
+///
+/// Note: Already openned connection would remain unchanged.
+pub fn toggle_local_db_in_memory(enabled: bool) {
```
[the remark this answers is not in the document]
FirelightFlagboy: I've done it

oxidation/libparsec/crates/platform_local_db/src/test_utils.rs
```diff
@@ -0,0 +1,63 @@
...
+use tokio::sync::RwLock;
+
+use diesel::{Connection, SqliteConnection};
+
+use crate::{executor::SqliteExecutor, DatabaseError, DatabaseResult, LocalDatabase};
+
+static OPEN_CONNECTION_IN_MEMORY: AtomicBool = AtomicBool::new(false);
+
+lazy_static::lazy_static! {
+    static ref LOCAL_DB_IN_MEMORY: Arc<Mutex<HashMap<String, Arc<SqliteExecutor>>>> =
```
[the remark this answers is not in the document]
FirelightFlagboy: You've done it

tests/conftest.py
```diff
@@ -468,11 +490,18 @@ def _clear_database_dir(allow_missing_path: bool):
...
+    caplog,
+    persistent_mockup: Callable[[], None] | None,
+    clear_database_dir: Callable[[bool], None],
+):
     async def _reset_testbed(keep_logs=False):
         if request.config.getoption("--postgresql"):
             await trio_asyncio.aio_as_trio(asyncio_reset_postgresql_testbed)
         clear_database_dir(True)
+        if persistent_mockup is not None:
+            persistent_mockup()
```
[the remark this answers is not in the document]
FirelightFlagboy: https://github.com/Scille/parsec-cloud/pull/3921#discussion_r1082207106

src/local_db.rs
```diff
@@ -0,0 +1,18 @@
...
+
+pub(crate) fn add_mod(_py: Python, m: &PyModule) -> PyResult<()> {
+    m.add_function(wrap_pyfunction!(toggle_local_db_in_memory, m)?)?;
+    m.add_function(wrap_pyfunction!(clear_local_db_in_memory, m)?)?;
+
+    Ok(())
+}
+
+#[pyfunction]
+pub(crate) fn toggle_local_db_in_memory(enabled: bool) {
```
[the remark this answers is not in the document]
FirelightFlagboy: You've done it

tests/conftest.py
touilleMan: yes that was my initial intent, but I didn't like the fact the name of the function didn't match at all it body. So introducing this dummy fixture make it more readable.

src/runtime.rs
touilleMan: 👍

src/runtime.rs
touilleMan: That's a very good idea !
Having a weak clue is much much better than not having any clue here ^^

oxidation/libparsec/crates/platform_local_db/src/test_utils.rs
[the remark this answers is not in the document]
touilleMan: So:
- panic in a synchronous function with python as caller raise an exception (but it's a `BaseException` so it's not 100% perfect anyway, see #3929)
- panic in a thread (e.g. in the executor) is printed on stderr and returned as error when joining the thread. This is good enough (we only should improve the logging of the crash see #3930 )
- panic in an asynchronous function with python as caller was broken, but I've fixed it so it also raises an exception into python \o/

oxidation/libparsec/crates/platform_local_db/src/executor.rs
```diff
@@ -43,7 +48,14 @@ impl SqliteExecutor {
...
         let wrapped_job = move |conn: &mut SqliteConnection| {
             let res = job(conn);
-            tx.send(res).unwrap();
+            // If send fails it means the caller's future has been dropped
+            // (hence dropping `rx`). In theory there is nothing wrong about
+            // it, however we log it anyway given the caller's unexpected drop
+            // may also be the sign of a bug...
+            if tx.send(res).is_err() {
+                // TODO: replace this by a proper warning log !
+                eprintln!("Caller has left");
```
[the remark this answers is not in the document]
touilleMan: yes, but we should discuss this in another PR

oxidation/libparsec/crates/platform_local_db/src/executor.rs
```diff
@@ -60,11 +72,21 @@ impl SqliteExecutor {
...
 impl Drop for SqliteExecutor {
     fn drop(&mut self) {
-        if let Err(e) = self.job_sender.send(ExecOrStop::Stop) {
-            eprintln!("Cannot send message to the background executor (Could just mean the executor is closed): {e}")
+        if self.job_sender.send(ExecOrStop::Stop).is_err() {
+            // If the send fails, this is most likely because the executor
+            // thread is already stopped. This might be because:
+            // - the thread has panicked
+            // - a bug in the code has cause another `ExecOrStop::Stop` to be
+            //   send to the thead before
```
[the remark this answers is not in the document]
touilleMan: so you agree if `ExecOrStop::Stop` is send twice it's a bug in our code ? or there is something else ? 🤔

tests/core/fs/conftest.py
touilleMan: 👍

oxidation/libparsec/crates/platform_local_db/src/executor.rs
```diff
@@ -27,7 +27,12 @@ impl SqliteExecutor {
             job_receiver,
             connection,
         };
-        let handle = std::thread::spawn(move || background_executor.serve());
+        // TODO: currently if the thread panic the error is printed to stderr,
+        // we should instead have a proper panic handler that log an error
+        let handle = std::thread::Builder::new()
+            .name("SqliteExecutor".to_string())
+            .spawn(move || background_executor.serve())
+            .expect("Cannot start thread");
```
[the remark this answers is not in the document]
touilleMan: I don't think this is needed given `thread::spawn` does an expect:
https://doc.rust-lang.org/src/std/thread/mod.rs.html#702-709

Failing to create a thread means the OS is very angry at us and we are basically doomed, so doing an expect here seems the right thing to do given otherwise the caller will have to handle a weird corner case error that it will most likely be dealt with with a panic anyway ;-)

oxidation/libparsec/crates/platform_local_db/src/test_utils.rs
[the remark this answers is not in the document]
FirelightFlagboy: :+1:

oxidation/libparsec/crates/platform_local_db/src/executor.rs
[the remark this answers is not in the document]
FirelightFlagboy: In #3930 then

oxidation/libparsec/crates/platform_local_db/src/executor.rs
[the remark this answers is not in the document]
FirelightFlagboy: If `Stop` is sent twice it's a flaw, meaning we would have `drop(executor)` multiple and `rust` won't allow to compile a code like that

oxidation/libparsec/crates/platform_local_db/src/executor.rs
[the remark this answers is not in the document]
FirelightFlagboy: Ok that

tests/conftest.py
FirelightFlagboy: Ok so

oxidation/libparsec/crates/platform_local_db/src/executor.rs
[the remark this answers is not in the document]
touilleMan: > If Stop is sent twice it's a flaw

Yes, that is my point saying in the comment this case is a bug

> meaning we would have drop(executor) multiple

You are making assumption on how bug could work.
This could also be that for whatever reason `ExecOrStop::Stop` is send where `ExecOrStop::Exec` was supposed to be send.

oxidation/libparsec/crates/platform_local_db/src/executor.rs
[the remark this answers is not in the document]
touilleMan: yes

oxidation/libparsec/crates/platform_local_db/src/test_utils.rs
TimeEngineer: it's `strategy` itself

tests/common/core.py
```diff
@@ -110,11 +122,48 @@ async def adam_core(
...
         core_config.data_base_dir, bob, initial_user_manifest=initial_user_manifest
     )
     async with core_factory(bob) as core:
         yield core
+
+
+@pytest.fixture
+def global_core_monitors_freeze(monkeypatch):
+    # In theory the `LoggedCore` (unlike `UserFS` alone) will run monitors in the
+    # background. However this an issue in some tests given it cause flakiness due
```
FirelightFlagboy, line 142:
```suggestion
    # background. However this is an issue in some tests given it cause flakiness due
```

PR #3931, opened 2023-01-19
Fix CI packaging

TimeEngineer:
# What has changed ?

The CI packaging is broken, because the `default-feature` isn't disabled for `libparsec_platform_device_loader` which depends on `libparsec_crypto`

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -373,13 +373,9 @@ jobs:
         shell: bash
         run: echo OPENSSL_SRC_PERL=C:/Strawberry/perl/bin/perl >> $GITHUB_ENV
 
-      - name: Additional Check
+      - name: Additional Check (for packaging)
         if: startsWith(matrix.os, 'ubuntu')
-        run: |
-          set -ex
-          rustup target add wasm32-unknown-unknown
-          cargo check --no-default-features --features use-sodiumoxide
-          cargo check --target wasm32-unknown-unknown
+        run: cargo check --no-default-features --features use-sodiumoxide
 
       - name: Test rust codebase
         if: steps.rust-changes.outputs.run == 'true'
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -124,13 +124,6 @@ jobs:
         env:
           POETRY_LIBPARSEC_BUILD_STRATEGY: no_build
 
-      - name: Additional Check
-        run: |
-          set -ex
-          rustup target add wasm32-unknown-unknown
-          cargo check --no-default-features --features use-sodiumoxide
-          cargo check --target wasm32-unknown-unknown
-
       # Clippy basically compile the project, hence it's faster to run it in
       # the test-rust-matrix job where compilation cache is reused !
       - uses: ./.github/actions/use-pre-commit
@@ -379,6 +372,14 @@ jobs:
         if: startsWith(matrix.os, 'windows')
         shell: bash
         run: echo OPENSSL_SRC_PERL=C:/Strawberry/perl/bin/perl >> $GITHUB_ENV
+
+      - name: Additional Check
+        if: startsWith(matrix.os, 'ubuntu')
+        run: |
+          set -ex
+          rustup target add wasm32-unknown-unknown
+          cargo check --no-default-features --features use-sodiumoxide
+          cargo check --target wasm32-unknown-unknown
 
       - name: Test rust codebase
         if: steps.rust-changes.outputs.run == 'true'
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -375,7 +375,7 @@ jobs:
 
       - name: Additional Check (for packaging)
         if: startsWith(matrix.os, 'ubuntu')
-        run: cargo check --no-default-features --features use-sodiumoxide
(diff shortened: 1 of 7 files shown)


.github/workflows/ci.yml
```diff
@@ -124,6 +124,13 @@ jobs:
         env:
           POETRY_LIBPARSEC_BUILD_STRATEGY: no_build
 
+      - name: Additional Check
+        run: |
+          set -ex
+          rustup target add wasm32-unknown-unknown
+          cargo check --no-default-features --features use-sodiumoxide
+          cargo check --target wasm32-unknown-unknown
```
FirelightFlagboy, line 132: cargo check could use the build artifact to be faster.
So move it to `test-rust-matrix` like for the task `pre-commit clippy`

FirelightFlagboy (commented):
Another way would be to remove completely the `features` that aren't compatible between desktop and web from `default-features` :thinking:

Volham22 (approved):

.github/workflows/ci.yml
TimeEngineer: Is it better ?

build.py
```diff
@@ -64,14 +64,11 @@ def build():
     # native module and discard the rest !
 
     maturin_build_profile = "--profile=" + os.environ.get("CARGO_PROFILE", DEFAULT_CARGO_PROFILE)
-    features = (
-        "--no-default-features --features use-sodiumoxide"
-        if maturin_build_profile == "--profile=release"
-        else ""
-    )
+    maturin_build_features = os.environ.get("CARGO_FEATURES", "")
```
touilleMan, line 67: This environ variable can be use to inject any parameter, hence we should rename it as `CARGO_EXTRA_ARGS`

.github/workflows/ci.yml
```diff
@@ -373,6 +373,14 @@ jobs:
         shell: bash
         run: echo OPENSSL_SRC_PERL=C:/Strawberry/perl/bin/perl >> $GITHUB_ENV
 
+      - name: Additional Check
```
touilleMan, line 376: This check is quiet costly: we re-compile the entire project and dependencies (and we add pressure on the cache)

Instead what we should do:
- modify `.github/actions/paths-filter` to add a new `rust-platform` entry this match `oxidation/libparsec/crates/platform_*/**`[^1]
- add this `rust-platform` to the `client-web` filter

Now `🌐 Web tests` will be triggered any time the platform code is modified ;-)

[^1]: we can also be a bit more cunning in filtering and have instead a `rust-platform-web` that match on `oxidation/libparsec/crates/platform_*/src/web/**`, this way we avoid triggering the web build if changing the native part

.github/workflows/ci.yml
```diff
@@ -373,6 +373,14 @@ jobs:
...
+        if: startsWith(matrix.os, 'ubuntu')
+        run: |
+          set -ex
+          rustup target add wasm32-unknown-unknown
+          cargo check --no-default-features --features use-sodiumoxide
```
touilleMan, line 381: What the need for that ? I would say check is basically the first half of compilation, so this line is redundant with when we do `cargo test --profile ci-rust --package libparsec_crypto --no-default-features --features use-sodiumoxide`

touilleMan (changes requested):

.github/workflows/ci.yml
TimeEngineer: Here we compile all with sodiumoxide, that's how we can detect if the package will work

build.py
TimeEngineer: Done

.github/workflows/ci.yml
FirelightFlagboy: Yes because we could use the cache

FirelightFlagboy: But the important part is to measure

.github/workflows/ci.yml
TimeEngineer: Done

FirelightFlagboy (approved):

.github/workflows/ci.yml
```diff
@@ -373,6 +373,10 @@ jobs:
         shell: bash
         run: echo OPENSSL_SRC_PERL=C:/Strawberry/perl/bin/perl >> $GITHUB_ENV
 
+      - name: Additional Check (for packaging)
+        if: startsWith(matrix.os, 'ubuntu')
+        run: cargo check --no-default-features --features use-sodiumoxide
```
touilleMan, line 378: add `--profile ci-rust`

build.py
```diff
@@ -64,14 +64,11 @@ def build():
     # native module and discard the rest !
 
     maturin_build_profile = "--profile=" + os.environ.get("CARGO_PROFILE", DEFAULT_CARGO_PROFILE)
-    features = (
-        "--no-default-features --features use-sodiumoxide"
-        if maturin_build_profile == "--profile=release"
-        else ""
-    )
+    maturin_build_features = os.environ.get("CARGO_EXTRA_ARGS", "")
```
touilleMan, line 67: rename variable to match env var

touilleMan (approved):

.github/workflows/ci.yml
TimeEngineer: It takes 3mins to check but we win 2mins in rust test (so 1min remains)
And it was approximately ~ 2mins

PR #3932, opened 2023-01-20
Better distinction between file and folder in GUI

### LINKED ISSUE #3855: Better distinction between file and folder in the QT application
Some dialogs and menu entries do not discriminate between file and folders, using the term `file` everywhere. 

This includes:
- Rename file
- Delete file
- File history
- etc.

This is confusing and should be fixed.

Ironicbay:
# What has changed ?

This PR adds more distinctions between files and folders in contextual menus and titles.

Closes #3855 

"élément" feels fitting in french, but not quite as much in english. I didn't find another generic term for "file or folder".

### DIFF
--- a/newsfragments/3855.misc.rst
+++ b/newsfragments/3855.misc.rst
@@ -1,0 +1,1 @@
+Added better distinction between file and folder actions in contextual menus.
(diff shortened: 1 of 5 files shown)


FirelightFlagboy (changes requested):
You forgot to add a news fragment since the change here affect the user

FirelightFlagboy (approved):

PR #3933, opened 2023-01-20
Make `load` devices functions async

Volham22:
# What has changed ?

This PR remove some unnecessary calls to `clone` in local_device_file.


FirelightFlagboy (approved):

TimeEngineer:
Can you make these function async, to justify more these changes ?

touilleMan:
So the solution here is to move the clone responsibility from `save_device_with_password` body to `save_device_with_password` caller

I guess this is okay when we change the password of a device (given we first load the device with the old password, then save it with the new, hence producing then consuming a device object)

However when we create a new device (during organization bootstrap and user/device enrollment) we want save the device then login right away. In this case we have to clone the device object :(

So in the end I don't like this solution given we move the burden onto the user instead of keeping it hidden.

On top of that performance is no big deal here, so cloning is okay. I think we should just remove the `// TODO: Can we avoid these copy?` and move along ^^

Volham22:
> Can you make these functions async ?
Yes, I'll do it.

oxidation/libparsec/crates/platform_device_loader/src/native/local_device.rs
```diff
@@ -21,6 +21,9 @@ pub fn load_device_with_password_from_path(
     }
 }
 
-pub fn load_device_with_password(key: &str, password: &str) -> LocalDeviceResult<LocalDevice> {
+pub async fn load_device_with_password(
+    key: &str,
+    password: &str,
+) -> LocalDeviceResult<LocalDevice> {
     load_device_with_password_from_path(Path::new(key), password)
```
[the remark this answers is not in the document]
Volham22: Done.

oxidation/libparsec/crates/platform_device_loader/src/native/local_device.rs
```diff
@@ -7,7 +7,7 @@ use libparsec_client_types::{DeviceFile, LocalDevice, LocalDeviceResult};
 use crate::load_device_file;
 use crate::load_device_with_password_core;
 
-pub fn load_device_with_password_from_path(
+pub async fn load_device_with_password_from_path(
```
FirelightFlagboy, line 10: The call to `load_device_file` should be made async

FirelightFlagboy (changes requested):

oxidation/libparsec/crates/platform_device_loader/src/native/local_device.rs
Volham22: Done.

FirelightFlagboy (approved):

TimeEngineer (approved):

TimeEngineer (changes requested):
Need to fix tests

touilleMan:
We decide against merging this PR: adding async is a cost and we'd rather avoid it if possible.
Instead we should split the APIs functions to keep the part requiring asyncness separated. Of course this is not always possible, but for  instance for the login API:

```rust
// Step 1
fn get_device_file_from_fs(path: Path) -> Result<Vec<u8>, ...>
async fn get_device_file_from_server(server_addr: BackendAddr, auth_login: String, auth_password: String)-> Result<Vec<u8>, ...>

// Step 2
fn open_device_with_password(file_device: Vec<u8>,  password: String) -> Result<LocalDevice, ...>
fn open_device_with_smartcard(file_device: Vec<u8>) -> Result<LocalDevice, ...>

// Step3
async fn start_client(device: LocalDevice) -> Result<...>

// All done ! now we have a logged_core running \o/
```

PR #3934, opened 2023-01-20
Do not use Vec<u8> as argument type in the binding

### LINKED ISSUE #3925: Do not use `Vec<u8>` as argument type in the binding
See the corresponding issue on pyo3 for more information: https://github.com/PyO3/pyo3/issues/2888

Instead we should take a `PyObject` as argument and perform an explicit conversion:
```rust
fn extract_to_vec_u8(object: PyObject, py: Python) -> PyResult<Vec<u8>> {
    Ok(match object.extract::<&PyByteArray>(py) {
        Ok(x) => x.to_vec(),
        Err(_) => object.extract::<&PyBytes>(py)?.as_bytes().to_vec(),
    })
}
```

Note: this performs a copy than can usually be avoided but it requires an `unsafe` block in the case of a `bytearray`. There's already a couple of such blocks in the codebase, e.g:
https://github.com/Scille/parsec-cloud/blob/9fbd8ef5c06541d2b04de317c6ba5d75021fdc63/oxidation/libparsec_python/src/api_crypto.rs#L29-L35

vxgmichel:
Fix #3925

### DIFF
--- a/newsfragments/3925.misc.rst
+++ b/newsfragments/3925.misc.rst
@@ -1,0 +1,1 @@
+Fix an underlying issue that caused an overall performance degradation
--- a/src/addrs.rs
+++ b/src/addrs.rs
@@ -10,6 +10,7 @@ use std::str::FromStr;
 
 use crate::{
     api_crypto::VerifyKey,
+    binding_utils::BytesWrapper,
     enumerate::InvitationType,
     ids::{EntryID, InvitationToken, OrganizationID},
 };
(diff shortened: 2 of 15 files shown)


vxgmichel:
It should be exhaustive:
```shell
± git grep 'Vec<u8>' src
src/binding_utils.rs:impl From<BytesWrapper<'_>> for Vec<u8> {
src/binding_utils.rs:    type ResultType = Vec<u8>;
src/protocol/mod.rs:pub(crate) type Bytes = Vec<u8>;
src/protocol/mod.rs:pub(crate) type ListOfBytes = Vec<Vec<u8>>;
```

vxgmichel:
It should be exhaustive:
```shell
± git grep 'Vec<u8>' src
src/binding_utils.rs:impl From<BytesWrapper<'_>> for Vec<u8> {
src/binding_utils.rs:    type ResultType = Vec<u8>;
src/protocol/mod.rs:pub(crate) type Bytes = Vec<u8>;
src/protocol/mod.rs:pub(crate) type ListOfBytes = Vec<Vec<u8>>;
```

FirelightFlagboy (approved):

PR #3943, opened 2023-01-23
Fix available device key file path type

### LINKED ISSUE #3942: AvailableDevice.key_file_path returns incorrect type
Returns a `str` instead of a `Path`.

```python
>>> from parsec._parsec import list_available_devices
>>> type(list_available_devices("/tmp/parsec-testenv-v9lgjm12/config/parsec")[0].key_file_path)
<class 'str'>

```

Volham22:
# What has changed ?

This closes #3942 . When we return a `PathBuf` with `pyo3` the default conversion gives a python string. We want this conversion to yield a `pathlib.Path` because it makes path manipulation easier in python.

### DIFF
--- a/src/binding_utils.rs
+++ b/src/binding_utils.rs
@@ -1,13 +1,35 @@
 // Parsec Cloud (https://parsec.cloud) Copyright (c) BUSL-1.1 (eventually AGPL-3.0) 2016-present Scille SAS
 
 use pyo3::{
-    exceptions::PyNotImplementedError, pyclass::CompareOp, types::PyByteArray, types::PyBytes,
-    types::PyFrozenSet, FromPyObject, PyAny, PyResult,
+    exceptions::PyNotImplementedError,
+    pyclass::CompareOp,
+    types::{PyByteArray, PyBytes, PyFrozenSet, PyTuple},
+    FromPyObject, IntoPy, PyAny, PyObject, PyResult,
 };
 use std::{
     collections::{hash_map::DefaultHasher, HashSet},
     hash::{Hash, Hasher},
 };
+
+#[derive(FromPyObject)]
+pub(crate) struct PathWrapper(pub std::path::PathBuf);
+
+impl IntoPy<PyObject> for PathWrapper {
+    fn into_py(self, py: pyo3::Python<'_>) -> pyo3::PyObject {
+        // Pathlib is part of the standard library
+        let pathlib_module = py
+            .import("pathlib")
+            .expect("import `pathlib` module failed.");
+        let path_ctor = pathlib_module
+            .getattr("Path")
+            .expect("can't get `Path` from `pathlib`.");
+
+        path_ctor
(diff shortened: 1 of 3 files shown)


src/binding_utils.rs
```diff
@@ -1,14 +1,34 @@
...
+
+impl IntoPy<PyObject> for PathWrapper {
+    fn into_py(self, py: pyo3::Python<'_>) -> pyo3::PyObject {
+        // Pathlib is part of the standard library
+        let pathlib_module = py.import("pathlib").unwrap();
+        let path_ctor = pathlib_module.getattr("Path").unwrap();
+
+        path_ctor
+            .call1(PyTuple::new(py, [self.0.to_str().unwrap()]))
+            .unwrap()
```
FirelightFlagboy, line 27: Use `expect`

FirelightFlagboy (changes requested):

src/binding_utils.rs
Volham22: Done.

src/binding_utils.rs
```diff
@@ -1,14 +1,38 @@
...
+        let pathlib_module = py
+            .import("pathlib")
+            .expect("import `pathlib` module failed");
+        let path_ctor = pathlib_module
+            .getattr("Path")
+            .expect("can't get `Path` from `pathlib");
+
+        path_ctor
+            .call1(PyTuple::new(py, [self.0.to_str().unwrap()]))
+            .unwrap()
```
FirelightFlagboy, line 31: Missing this unwrap

FirelightFlagboy (changes requested):

src/binding_utils.rs
```diff
@@ -1,14 +1,38 @@
 // Parsec Cloud (https://parsec.cloud) Copyright (c) BUSL-1.1 (eventually AGPL-3.0) 2016-present Scille SAS
 
 use pyo3::{
-    exceptions::PyNotImplementedError, pyclass::CompareOp, types::PyByteArray, types::PyBytes,
-    types::PyFrozenSet, FromPyObject, PyAny, PyResult,
+    exceptions::PyNotImplementedError,
+    pyclass::CompareOp,
+    types::PyByteArray,
+    types::PyFrozenSet,
+    types::{PyBytes, PyTuple},
```
TimeEngineer, line 8: group import

src/binding_utils.rs
```diff
@@ -1,14 +1,38 @@
...
+        // Pathlib is part of the standard library
+        let pathlib_module = py
+            .import("pathlib")
+            .expect("import `pathlib` module failed");
+        let path_ctor = pathlib_module
+            .getattr("Path")
+            .expect("can't get `Path` from `pathlib");
+
+        path_ctor
+            .call1(PyTuple::new(py, [self.0.to_str().unwrap()]))
```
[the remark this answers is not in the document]
Volham22: Yes it works.

FirelightFlagboy (approved):

PR #3944, opened 2023-01-23
Storage module impro perf

FirelightFlagboy:
- Improve test env script => add cleanup function
- Improve `Vec<u8>` conversion in `set_chunk`
- Improve `set_manifest` => allow to do operation in cache
- Improve `get_manifest` => allow to do operation in cache
- Improve `load_file_descriptor` => allow to do operation in cache

### DIFF
--- a/src/storage/workspace_storage.rs
+++ b/src/storage/workspace_storage.rs
@@ -143,23 +143,28 @@ impl WorkspaceStorage {
         removed_ids: Option<HashSet<ChunkID>>,
     ) -> PyResult<FutureIntoCoroutine> {
         let manifest = manifest_from_py_object(py, manifest)?;
-        let ws = self.0.clone();
-
-        let fut = FutureIntoCoroutine::from(async move {
-            ws.set_manifest(
-                entry_id.0,
-                manifest,
-                cache_only,
-                false,
-                removed_ids.map(|x| {
-                    x.into_iter()
-                        .map(|id| libparsec::core_fs::ChunkOrBlockID::ChunkID(id.0))
-                        .collect()
-                }),
-            )
-            .await
-            .map_err(fs_to_python_error)
+        let removed_ids = removed_ids.map(|x| {
+            x.into_iter()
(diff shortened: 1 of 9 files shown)


src/storage/workspace_storage.rs
```diff
@@ -266,16 +262,24 @@ impl WorkspaceStorage {
...
+    fn set_chunk(
+        &self,
+        py: Python,
+        chunk_id: ChunkID,
+        block: PyObject,
+    ) -> PyResult<FutureIntoCoroutine> {
+        let block = match block.extract::<&PyByteArray>(py) {
+            Ok(x) => x.to_vec(),
+            Err(_) => block.extract::<&PyBytes>(py)?.as_bytes().to_vec(),
+        };
```
vxgmichel, line 274: This should be done using `BytesWrapper` et `unwrap_bytes!` instead

tests/scripts/run_testenv.sh
```diff
@@ -41,6 +41,21 @@ $SCRIPT_DIR/run_testenv.py --source-file $SOURCE_FILE $@ || return $?
 source $SOURCE_FILE
 
 # Clean up
-rm $SOURCE_FILE
-unset SOURCE_FILE
-unset SCRIPT_DIR
```
vxgmichel, line 46: If those variables stay in the current shell we should find less conflicting names for them

vxgmichel (approved):

src/storage/workspace_storage.rs
FirelightFlagboy: Would be done when I rebase the base branch with master

tests/scripts/run_testenv.sh
FirelightFlagboy: They may stay in the current shell if the user don't invoke `cleanup` after the test are finished.
I've renamed them by adding the prefix `PARSEC_TESTENV_`

TimeEngineer (approved):

PR #3947, opened 2023-01-23
[GUI] Maximize window by default

### LINKED ISSUE #3946: [🐛 | Bug]: GUI isn't maximized by default
### Version

latest

Linux

### What happened ?

When running the GUI for the first time (no geometry saved in config), the window should take the whole screen.

### Relevant output

_No response_

Max-7:
# What has changed ?

When opening the GUI for the first time, the window should take all the available space.

Closes #3946

### DIFF
--- a/parsec/core/gui/main_window.py
+++ b/parsec/core/gui/main_window.py
@@ -581,7 +581,7 @@ class MainWindow(QMainWindow, Ui_MainWindow):
 
     def show_window(self, skip_dialogs: bool = False) -> None:
         try:
-            if self.config.gui_geometry:
+            if self.config.gui_geometry is not None:
                 self.restoreGeometry(self.config.gui_geometry)
             else:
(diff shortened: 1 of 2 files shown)


parsec/core/gui/main_window.py
```diff
@@ -581,9 +581,9 @@ def on_config_updated(self, event: CoreEvent, **kwargs: object) -> None:
 
     def show_window(self, skip_dialogs: bool = False) -> None:
         try:
-            if self.config.gui_geometry is not None and not self.restoreGeometry(
-                self.config.gui_geometry
-            ):
+            if self.config.gui_geometry:
```
vxgmichel, line 584:
```suggestion
            if self.config.gui_geometry is not None:
```

vxgmichel (approved):
LGTM :+1: 

A minor suggestion

PR #3954, opened 2023-01-25
Update maturin & patchelf

FirelightFlagboy:
- Update `maturin` used in build system to 0.14 (from 0.13)
- Install `patchelf` on Linux platform
- Install extra for `maturin` to use `patchelf`

### DIFF
--- a/build.py
+++ b/build.py
@@ -33,6 +33,8 @@ def run(cmd: str, **kwargs) -> subprocess.CompletedProcess:
 def build():
     run(f"{PYTHON_EXECUTABLE_PATH} --version")
     run(f"{PYTHON_EXECUTABLE_PATH} misc/generate_pyqt.py")
+    if sys.platform == "linux":
+        run("patchelf --version")
 
     if sys.platform == "win32":
         libparsec_path = "parsec/_parsec.cp39-win_amd64.pyd"
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -141,7 +141,10 @@ black = "^22.6.0"
 mypy = "^0.982"
 
 setuptools = "^63.1.0"
-maturin = "~=0.14"
+
+maturin = "^0.14.10"
+patchelf = { version = "^0.17.2.0", markers = "platform_system=='Linux'" }
+
 cibuildwheel = "2.10.0"
 editorconfig-checker = "2.4.0"
 types-requests = "^2.28"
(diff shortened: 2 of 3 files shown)


TimeEngineer (approved):

pyproject.toml
```diff
@@ -141,7 +141,8 @@ black = "^22.6.0"
 mypy = "^0.982"
 
 setuptools = "^63.1.0"
-maturin = "~=0.14"
+maturin = {extras = ["pathelf"], version = "^0.14.10"}
```
Volham22, line 144: There's a missing `c` at `patchelf`

Volham22 (changes requested):

pyproject.toml
FirelightFlagboy: done, strange `poetry` don't say anything :thinking:

Volham22 (approved):

pyproject.toml
```diff
@@ -141,7 +141,10 @@ black = "^22.6.0"
 mypy = "^0.982"
 
 setuptools = "^63.1.0"
-maturin = "~=0.14"
+
+maturin = "^0.14.10"
```
touilleMan, line 145: - Version number is different from the `requires` part
- `maturin[patchelf]` only present in requires, why ?

build.py
```diff
@@ -33,6 +33,8 @@ def run(cmd: str, **kwargs) -> subprocess.CompletedProcess:
 def build():
     run(f"{PYTHON_EXECUTABLE_PATH} --version")
     run(f"{PYTHON_EXECUTABLE_PATH} misc/generate_pyqt.py")
+    if sys.platform == "linux":
```
touilleMan, line 36: Why ? We need a comment to understand - what is achieved here and why it is needed !

touilleMan: Also we don't need that for macOS ?

FirelightFlagboy: 1. `patchelf --version` print the version of `patchelf` is for debug purpose, to be sure it use the correct version (not the system one)
2. `elf` is linux specific or darwin it's `mach-o` file

pyproject.toml
FirelightFlagboy: 1. `poetry` and `python` don't resolve the same way there dependencies version

    a. for `poetry`: `^1.2.3` mean `>=1.2.3,<2.0.0`
    b. for `python`: `~=1.2.3` mean `>=1.2.3.0,==1.2.*`

    maybe I should rewrite those version requirement in `build-system` to be more clear 
2. Because, that would mean `windows` & `darwin` to install / build `patchelf` both don't support `elf` file format.
    What I would have wanted is a conditional extra (i.e.: to have the extra `patchelf` only on linux)
    Here It work because I install `patchelf` alongside `maturin` on linux and it seems to be the only dependencies listed for the extra `maturin[patchelf]`

PR #3956, opened 2023-01-25
Extract extra lib required by libparsec

FirelightFlagboy:
libparsec can require extra shared libs to work but the current
workflow for `ci` testing don't work because it don't extract those shared lib
and cause the main lib to not find those.

### DIFF
--- a/build.py
+++ b/build.py
@@ -88,6 +88,13 @@ def build():
         with zipfile.ZipFile(wheel_path) as wheel:
             wheel.extract(libparsec_path, path=BASEDIR)
 
+            if build_profile == "ci" and sys.platform == "linux":
+                zipinfo = wheel.infolist()
+                for extra_lib in filter(
+                    lambda member: member.filename.startswith("parsec.libs/"), zipinfo
+                ):
+                    wheel.extract(extra_lib, path=BASEDIR)
+
 
 if __name__ == "__main__":
     display("Launching poetry build.py script")


build.py
```diff
@@ -88,6 +88,13 @@ def build():
         with zipfile.ZipFile(wheel_path) as wheel:
             wheel.extract(libparsec_path, path=BASEDIR)
 
+            if build_profile == "ci" and sys.platform == "linux":
+                zipinfo = wheel.infolist()
+                for extra_lib in filter(
```
touilleMan, line 93: Use a `if` in the for loopi instead of the filter, it is more pythonic and readable

```python
for item in zipinfo:
  if item.filename.startswith("parsec.libs/"):
    wheel.extract(item, path=BASEDIR)
```

build.py
```diff
@@ -88,6 +88,13 @@ def build():
         with zipfile.ZipFile(wheel_path) as wheel:
             wheel.extract(libparsec_path, path=BASEDIR)
 
+            if build_profile == "ci" and sys.platform == "linux":
```
touilleMan, line 91: @Volham22 @FirelightFlagboy Please add comment, and have the reflex to wonder about comments in when reviewing a PR 🙏 

This is a typical piece of code that will require efforts to understand why it is needed (why only on linux ? why only on ci ? what `parsec.libs` is for ? what is generated by ? answering those questions means ones has to do tests by building on his machine and on the CI)

touilleMan (changes requested):

build.py
touilleMan: Besides I find is really weird to only have this check in CI mode (it seems like a very ad-hoc change, why don't we have the same issue in dev mode while maturin basically does the same thing for ci and dev... so I would guess it is an issue with `poetry install -E backend -E core` and not `maturin develop`)

FirelightFlagboy:
This PR was to show another hack to solve the problem on missing lib on linux.
It was solved in another way

PR #3970, opened 2023-01-30
Fix fstring usage for tr files

Ironicbay:
# What has changed ?

### DIFF
--- a/parsec/core/gui/file_table.py
+++ b/parsec/core/gui/file_table.py
@@ -294,15 +294,24 @@ class FileTable(QTableWidget):
 
             # Show the option to create a timestamped share link to the user is
             # useless here because it has the same effect as creating a regular file link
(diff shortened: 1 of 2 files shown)


FirelightFlagboy (commented):
Does this need a *news fragment* ?
It seems it's mostly a rewrite to make the code more clear on where to use `folder` or `file` text.
Does it correct a bug ?

FirelightFlagboy:
And what does `fstring` mean ?

Max-7:
> Does this need a _news fragment_ ? It seems it's mostly a rewrite to make the code more clear on where to use `folder` or `file` text. Does it correct a bug ?

The script that can be used to extract translation strings does not execute the code and extract the strings "as is". Strings that are added manually do work correctly though, so this just fixes the script. No impact on the user whatsoever.

Max-7 (approved):

PR #4003, opened 2023-02-02
[GUI] Include user invitation in the query string

### LINKED ISSUE #3967: Show pending users when filtering is active
When filtering users, the pending invitation are not shown even if the text matches the email address. This is confusing to the user.

Max-7:
# What has changed ?

User invitations were ignored by the query string and not displayed even if the email matched what was searched.

Closes #3967

### DIFF
--- a/parsec/core/gui/users_widget.py
+++ b/parsec/core/gui/users_widget.py
@@ -212,18 +212,19 @@ async def _do_list_users_and_invitations(
 ) -> tuple[int, list[UserInfo], list[InviteListItem]]:
     try:
         invitations = [] if omit_invitation else await core.list_invitations()
+        # Filtering invitation using the pattern if present
+        invitations = [
+            inv
+            for inv in invitations
+            if inv.type == InvitationType.USER and (pattern is None or pattern in inv.claimer_email)
+        ]
         users, total = await core.find_humans(
             page=page, per_page=USERS_PER_PAGE, query=pattern, omit_revoked=omit_revoked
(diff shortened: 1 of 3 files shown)


FirelightFlagboy (approved):

parsec/core/gui/users_widget.py
```diff
@@ -211,17 +211,20 @@ async def _do_list_users_and_invitations(
...
+            page=page, per_page=USERS_PER_PAGE, query=pattern, omit_revoked=omit_revoked
+        )
+        return (
+            total,
+            users,
+            [
+                inv
+                for inv in invitations
+                if inv.type == InvitationType.USER
+                and (pattern is not None and pattern in inv.claimer_email)
```
vxgmichel, line 225: What if pattern is None? Nothing is included then?

vxgmichel:
```suggestion
                and (pattern is None or pattern in inv.claimer_email)
```

vxgmichel: still I would do the filtering line 214 instead

vxgmichel (changes requested):

parsec/core/gui/users_widget.py
Max-7: I'm dumb. The funny thing is, it did work properly because the pattern is always provided (never actually None) as an empty string. It's fixed and I moved the filtering to line 214.

PR #4010, opened 2023-02-02
[GUI] Show unmounted workspace by default and save the state

### LINKED ISSUE #3969: UX issue when disabling a workspace
There's a UX issue when disabling a workspace: the default state for the "hide deactivated workspaces" tick box is ticked so when a user clicks on the slider to deactivate the workspace, the corresponding workspace disappear instantly and is replaced by the workspace after it in the list (if any). That can be very confusing, especially since it's the default behavior.

Max-7:
# What has changed ?

No longer hides unmounted workspaces by default, and the state of show/hide unmounted workspaces is saved in config.

Closes #3969

### DIFF
--- a/newsfragments/3969.bugfix.rst
+++ b/newsfragments/3969.bugfix.rst
@@ -1,0 +1,1 @@
+Unmounted workspaces are no longer hidden by default and the state is saved when the application exits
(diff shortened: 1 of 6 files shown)


FirelightFlagboy (approved):

vxgmichel (approved):
LGTM :+1:

PR #4012, opened 2023-02-02
Improve storage closing

FirelightFlagboy:


oxidation/libparsec/crates/core_fs/src/storage/chunk_storage.rs
```diff
@@ -485,8 +487,8 @@ mod tests {
         );
 
         let local_chunk_ids = chunk_storage.get_local_chunk_ids(&chunk_ids).await.unwrap();
-        let set0: HashSet<_, RandomState> = HashSet::from_iter(local_chunk_ids.iter());
-        let set1: HashSet<_, RandomState> = HashSet::from_iter(chunk_ids.iter());
+        let set0: HashSet<_> = HashSet::from_iter(local_chunk_ids.iter());
```
TimeEngineer, line 490: Remove type

oxidation/libparsec/crates/core_fs/src/storage/chunk_storage.rs
```diff
@@ -485,8 +487,8 @@ mod tests {
...
+        let set1: HashSet<_> = HashSet::from_iter(chunk_ids.iter());
```
TimeEngineer, line 491: same

oxidation/libparsec/crates/core_fs/src/storage/chunk_storage.rs
FirelightFlagboy: done

oxidation/libparsec/crates/core_fs/src/storage/chunk_storage.rs
FirelightFlagboy: done

Note: You could have a one comment for the 2 lines by selecting them

oxidation/libparsec/crates/core_fs/src/storage/chunk_storage.rs
```diff
@@ -485,8 +489,8 @@ mod tests {
         );
 
         let local_chunk_ids = chunk_storage.get_local_chunk_ids(&chunk_ids).await.unwrap();
-        let set0: HashSet<_, RandomState> = HashSet::from_iter(local_chunk_ids.iter());
-        let set1: HashSet<_, RandomState> = HashSet::from_iter(chunk_ids.iter());
+        let set0 = HashSet::<&ChunkID>::from_iter(local_chunk_ids.iter());
+        let set1 = HashSet::<&ChunkID>::from_iter(chunk_ids.iter());
```
TimeEngineer, line 493: isn't `type` infered by `local_chunk_ids` ?

FirelightFlagboy: Apparently no, I was getting error that `set0` & `set1` need to be typed

src/storage/workspace_storage.rs
```diff
@@ -311,13 +356,13 @@ impl WorkspaceStorage {
...
     }
 
-    fn get_need_sync_entries(&self) -> FutureIntoCoroutine {
-        let ws = self.0.clone();
+    fn get_need_sync_entries(&self) -> PyResult<FutureIntoCoroutine> {
+        let ws = self.get_storage()?;
 
-        FutureIntoCoroutine::from(async move {
+        Ok(FutureIntoCoroutine::from(async move {
             ws.get_need_sync_entries()
```
TimeEngineer, line 366: Why don't you call `get_storage` inside the coroutine, the scope seems the same for your guard and the second `Arc` seems not needed ? :thinking:

oxidation/libparsec/crates/core_fs/src/storage/chunk_storage.rs
TimeEngineer: Maybe you could consider
```rust
let set0 = local_chunk_ids.iter().collect::<HashSet<_>>();
let set1 = chunk_ids.iter().collect::<HashSet<_>>();
```
And I tried, only 1 of both is required, but it's strange :eyes:

src/storage/workspace_storage.rs
FirelightFlagboy: 1. Doing that require to clone `Self` is it something better ?
2. That way it prevent to send a job to `tokio` if the workspace storage is dropped, but this case should not append (that mean we use the `WorkspaceStorage` outside the context manager)

oxidation/libparsec/crates/core_fs/src/storage/chunk_storage.rs
FirelightFlagboy: You prefer the collect form ?

TimeEngineer: `collect` avoid the use of `FromIter` `trait` and `2` characters less than `from_iter`

src/storage/workspace_storage.rs
TimeEngineer: You clone self in `get_storage`

oxidation/libparsec/crates/core_fs/src/storage/chunk_storage.rs
FirelightFlagboy: done

FirelightFlagboy: There's a problem with `trait FromIterator` ?

src/storage/workspace_storage.rs
FirelightFlagboy: On another thought, this isn't possible because (to remove the second `Arc`):
1. I don't (and clippy) want to hold a `MutexGuard` between await point.
2. `WorkspaceStorage` (pure rust) isn't clonable and I don't want to make it so because it have to manage a connection database and we better have only 1 instance managing that behind a `Arc`.

oxidation/libparsec/crates/core_fs/src/storage/chunk_storage.rs
TimeEngineer: Nope

PR #4047, opened 2023-02-09
Correct wrong definition of `LocalDeviceError`

### LINKED ISSUE #4046: :bug: Wrong use of exception in `parsec/core/local_device.py`
## Intro

In `parsec/core/local_device.py` we define `LocalDeviceError`
But in `parsec._parsec` we also define `LocalDeviceError` and use it in the rust code.
That create problem when we do the following:

```python
try:
    # ...
except LocalDeviceError:
    pass
```

or an actual code example: https://github.com/Scille/parsec-cloud/blob/0e2f2d3d780a5528e38687de399b86834e87a09e/parsec/core/local_device.py#L79-L84

Because we `except` for the wrong exception (It have the same name but still)

## Mitigation

Use `parsec._parsec.LocalDeviceExc`

## Correction

1. Rename `parsec._parsec.LocalDeviceExc` to `parsec._parsec.LocalDeviceExc`
2. Replace `LocalDeviceError` definition in `parsec.core.local_device` by the `parsec._parsec` one

FirelightFlagboy:
We define `LocalDeviceError` 2 time, 1 in `parsec.core.local_device` and the other in the rust lib. Before the 2 aren't the same the rust lib raise the said exception in python we wouldn't be able to catch it with `LocalDeviceError` that is defined in the python code.

closes #4046


tests/core/test_recovery.py
```diff
@@ -95,6 +97,38 @@ async def test_recovery_ok(tmp_path, user_fs_factory, running_backend, bob):
...
+    file_name = get_recovery_device_file_name(recovery_device)
+
+    file_path = tmp_path / file_name
+    passphrase = await save_recovery_device(file_path, recovery_device, force=False)
+
+    # 2) Try to load the recovery device file with the wrong passphrase.
+    wrong_passphrase = "-".join([chunk[::-1] for chunk in passphrase.split("-")])
+    assert wrong_passphrase != passphrase
+
+    with pytest.raises(LocalDeviceError):
```
Max-7, line 128: `LocalDeviceError` is too broad, it should be a `LocalDeviceCryptoError` specifically.

Max-7 (changes requested):

tests/core/test_recovery.py
FirelightFlagboy: Would be done in another PR see #4048

Max-7 (approved):

PR #4049, opened 2023-02-09
Fix `FutureIntoCoroutine` runtime binding

TimeEngineer:
# What has changed ?

### DIFF
--- a/src/runtime.rs
+++ b/src/runtime.rs
@@ -1,17 +1,16 @@
 // Parsec Cloud (https://parsec.cloud) Copyright (c) BUSL-1.1 (eventually AGPL-3.0) 2016-present Scille SAS
 
+use futures::FutureExt;
 use pyo3::{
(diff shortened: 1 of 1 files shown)


src/runtime.rs
```diff
@@ -1,17 +1,18 @@
 // Parsec Cloud (https://parsec.cloud) Copyright (c) BUSL-1.1 (eventually AGPL-3.0) 2016-present Scille SAS
 
+#![allow(dead_code)]
+
```
FirelightFlagboy, line 4: Remove the code that ins't used

FirelightFlagboy (changes requested):

src/runtime.rs
TimeEngineer: Ok

TimeEngineer: Done

FirelightFlagboy: Depends on how fast you review the said PR

FirelightFlagboy (approved):

PR #4067, opened 2023-02-16
[parsec] Improve typing

FirelightFlagboy:
- Replace `BaseWorkspaceStorage` with `AnyWorkspaceStorage`
    > This change will make more sense when we merge #2903 
- Replace `Tuple` by `tuple`, `Dict` by `dict`, etc
- Move import mostly to `parsec._parsec`

### DIFF
--- a/parsec/core/fs/workspacefs/file_operations.py
+++ b/parsec/core/fs/workspacefs/file_operations.py
@@ -1,7 +1,6 @@
 # Parsec Cloud (https://parsec.cloud) Copyright (c) AGPL-3.0 2016-present Scille SAS
 from __future__ import annotations
 
-# Imports
 from parsec._parsec import prepare_read, prepare_reshape, prepare_resize, prepare_write
 
 __all__ = [
--- a/parsec/core/fs/workspacefs/workspacefs.py
+++ b/parsec/core/fs/workspacefs/workspacefs.py
@@ -4,16 +4,31 @@ from __future__ import annotations
 from collections import defaultdict
 from contextlib import asynccontextmanager
 from pathlib import Path
-from typing import TYPE_CHECKING, AsyncIterator, Awaitable, Callable, Dict, List, Tuple, cast
+from typing import TYPE_CHECKING, AsyncIterator, Awaitable, Callable, cast
 
 import attr
 import structlog
 import trio
 
-from parsec._parsec import CoreEvent, DateTime, FileManifest, RealmStatusRepOk, Regex
-from parsec.api.data import AnyRemoteManifest, BlockAccess
-from parsec.api.data import FileManifest as RemoteFileManifest
-from parsec.api.protocol import MaintenanceType, RealmID, UserID
+from parsec._parsec import (
+    BlockAccess,
+    CoreEvent,
+    DateTime,
+    EntryID,
+    EntryName,
+    LocalDevice,
+    LocalFileManifest,
+    LocalFolderManifest,
+    LocalWorkspaceManifest,
+    MaintenanceType,
+    RealmID,
+    RealmStatusRepOk,
+    Regex,
+    UserID,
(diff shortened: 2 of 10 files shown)


parsec/core/fs/workspacefs/workspacefs.py
```diff
@@ -4,16 +4,32 @@
...
+    LocalFolderManifest,
+    LocalWorkspaceManifest,
+    MaintenanceType,
+    RealmID,
+    RealmStatusRepOk,
+    Regex,
+    UserID,
+    WorkspaceEntry,
+)
+from parsec._parsec import FileManifest as RemoteFileManifest
```
Max-7, line 31: Dupplicate. I think you can use the alias in the import tuple, eg
```
from parsec._parsec import (
    EntryID,
    EntryName,
    ...
    FileManifest as RemoteFileManifest
)
```

parsec/core/fs/workspacefs/workspacefs.py
```diff
@@ -252,7 +261,9 @@ def is_read_only(self) -> bool:
     def is_revoked(self) -> bool:
         return self.get_workspace_entry().role is None
 
-    async def path_info(self, path: AnyPath) -> Dict[str, object]:
+    # Information
```
Max-7, line 264: Useless comment, the function name is explicit enough.

parsec/core/fs/workspacefs/workspacefs.py
```diff
@@ -4,16 +4,32 @@
...
-from parsec.api.data import AnyRemoteManifest, BlockAccess
-from parsec.api.data import FileManifest as RemoteFileManifest
-from parsec.api.protocol import MaintenanceType, RealmID, UserID
+from parsec._parsec import (
+    BlockAccess,
+    CoreEvent,
+    DateTime,
+    EntryID,
+    EntryName,
+    FileManifest,
```
Max-7, line 19: `FileManifest` was previously imported from `parsec.api.data` as `RemoteFileManifest`, and from `parsec._parsec` as `FileManifest`. What's the difference between these two? Is the alias useless? Do we have two different classes with the same name?

parsec/core/fs/workspacefs/file_operations.py
```diff
@@ -2,7 +2,18 @@
...
+    Chunk,
+    ChunkID,
+    prepare_read,
+    prepare_reshape,
+    prepare_resize,
+    prepare_write,
+)
+
+Chunks = tuple[Chunk, ...]
+ChunkIDSet = set[ChunkID]
```
Max-7, line 15: Are those used somewhere? I'm not a big fan of the alias on simple things like lists, it makes it less explicit. If we take something like `Chunks` and I see this type in the codebase, now I have to wonder weither it's a class on its own or a collection, and what type of collection.

Max-7 (changes requested):

parsec/core/fs/workspacefs/workspacefs.py
FirelightFlagboy: Ruff force to have a new line when we write an alias :/

parsec/core/fs/workspacefs/workspacefs.py
FirelightFlagboy: `FileManifest` from `parsec.api.data` is a re-export of `parsec._parsec`

parsec/core/fs/workspacefs/workspacefs.py
FirelightFlagboy: done

parsec/core/fs/workspacefs/file_operations.py
FirelightFlagboy: Those alias aren't used, I've removed them

Max-7 (approved):

PR #4075, opened 2023-02-21
[ci] Print current python gui test before exec

FirelightFlagboy:
Sometime some python gui tests can blocks on linux (or other platform).
Setting the option `numprocesses` to `1` make `pytest` print the next test that it will be executing, allowing the dev to have more clue on which test is blocking.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -313,11 +313,13 @@ jobs:
           && startsWith(matrix.os, 'ubuntu-')
         run: poetry run pip install pytest-xvfb
 
-      - name: (🐧🏁 Not macOS) GUI tests
+      - name: (🐧🏁 Not MacOS) GUI tests
         if: >-
           steps.python-changes.outputs.run == 'true'
           && startsWith(matrix.os, 'macos-') != true
-        run: poetry run pytest ${{ env.pytest-base-args }} tests --runmountpoint --runslow --rungui -m gui
+        # Note: using `--numprocesses 1` force pytest to print the test that it will be executing
+        # This is usefull to pinpoint exactly which test in blocking.
+        run: poetry run pytest ${{ env.pytest-base-args }} tests --runmountpoint --runslow --rungui -m gui --numprocesses 1
         timeout-minutes: 10
 
       - name: (🐧 Linux) PostgreSQL tests


.github/workflows/ci.yml
```diff
@@ -311,11 +311,13 @@ jobs:
...
 
-      - name: (🐧🏁 Not macOS) GUI tests
+      - name: (🐧🏁 Not MacOS) GUI tests
         if: >-
           steps.python-changes.outputs.run == 'true'
           && startsWith(matrix.os, 'macos-') != true
-        run: poetry run pytest ${{ env.pytest-base-args }} tests --runmountpoint --runslow --rungui -m gui
+        # Note: using `--numprocesses 1` force pytest to print the test that it will be executing
+        # This is usefull to pinpoint exactly which test in blocking.
+        run: poetry run pytest ${{ env.pytest-base-args }} tests --runmountpoint --runslow --rungui -m gui --numprocesses 1
```
touilleMan, line 320: passing `--vebose` makes pytest providing the name of the tests being run:
```
$ py.test tests/backend -n auto -v
tests/backend/test_access.py::test_invited_has_limited_access 
tests/backend/test_events.py::test_sse_events_subscribe 
[gw0] [  0%] PASSED tests/backend/test_access.py::test_invited_has_limited_access 
tests/backend/test_access.py::test_authenticated_has_limited_access 
[gw1] [  0%] PASSED tests/backend/test_events.py::test_sse_events_subscribe 
tests/backend/test_events.py::test_sse_events_bad_auth 
[...]
```

touilleMan (changes requested):

.github/workflows/ci.yml
FirelightFlagboy: It's already the case, we provide `env.pytest-base-args`

which is defined with the following value

```yaml
pytest-base-args: >-
    --log-level=DEBUG
    --durations=10
    --side-effects-timeout=10
    -vv
    -x
```

touilleMan: So the issue is *even* with `-vv` we don't get the information of the current test being executed when running in parallel ?

That seems weird to me (I even think to recall we have an exercise about a similar problem :trollface:  ), the next test to be executed is printed when running in parallel (it is less readable given multiple workers start/finish execution in concurrency, but should be feasible to check in theory)

Another solution would be to use [pytest-timeout](https://github.com/pytest-dev/pytest-timeout) to configure a generic timeout in the configuration (typically something like 5mn or 10mn). The only trick is to make sure it works well with the pytest parallel option (given the timeout will occurs in a worker, the main pytest should detect this an close itself in a correct fashion)

FirelightFlagboy:
Merged in #2903

Issue #4071, opened 2023-02-20
Web test for `date` are failing on master


FirelightFlagboy:
See the jobs that fail:

- https://github.com/Scille/parsec-cloud/actions/runs/4221214110/jobs/7328245125
- https://github.com/Scille/parsec-cloud/actions/runs/4222048202/jobs/7330074290

```
FAIL tests/unit/common/date.spec.ts
  ● common.date.ts EN › formats date as time elapsed since date in English

    expect(received).toBe(expected) // Object.is equality

    Expected: "Thursday, November 19, 1998 at 8:00 AM"
    Received: "Thursday, November 19, 1998 at 8:00 AM"

      43 |     jest.useRealTimers();
      44 |     // Way way in the past, should display the full date
    > 45 |     expect(formatTimeSince(now, t, d)).toBe('Thursday, November 19, 1998 at 8:00 AM');
         |                                        ^
      46 |
      47 |     jest.useFakeTimers();
      48 |     // 3 days before

      at Object.<anonymous> (tests/unit/common/date.spec.ts:45:40)
```

Using `cat -e` seems to show invalid spacing character being used :thinking: 

```diff
- Thursday, November 19, 1998 at 8:00 AM
+ Thursday, November 19, 1998 at 8:00M-bM-^@M-/AM
```

Landeers:
See https://github.com/nodejs/node/issues/46123

PR #4079, opened 2023-02-22
Prevent using node > 18.13.0

FirelightFlagboy:
Currently our unit test started to fail because of a change introduced in node-18.13.0 (nodejs/node#46123).
So using a version before that change would correct the test that fail in #4071.

Changes
-------

- Change node-version in `ci.yml` to `18.12.0`
- Use github action `actions/setup-node` in `Web tests`

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -10,7 +10,9 @@ on:
 env:
   python-version: "3.9"
   rust-version: 1.66.1
-  node-version: 18.13.0
+  # We use the version 18.12 because the version >= 18.13 have some breaking changes on how they format the date.
+  # That would break our unit test if we don't update them.
+  node-version: 18.12.0
   poetry-version: 1.2.2
   # TODO: We stick to PostgreSQL 12 for the moment given later versions are
   # much slower (postgresql tests runs in ~9mn on 12 vs ~36mn on 14 !)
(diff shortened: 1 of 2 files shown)


oxidation/client/package.json
```diff
@@ -79,7 +79,7 @@
         "wasm-pack": "^0.10.3"
     },
     "engines": {
-        "node": ">=16.13.0 <19.0.0"
+        "node": ">=16.13.0 <18.13.0"
```
touilleMan, line 82: Must add a comment to explain why this is needed !
(and ideally when upgrade will be possible)

touilleMan (changes requested):

oxidation/client/package.json
FirelightFlagboy: I can't add a comment in json :(

touilleMan (approved):
add a `"//": "<comment here>"` comment hack, then :rocket:

.github/workflows/ci.yml
```diff
@@ -10,7 +10,7 @@ on:
 env:
   python-version: "3.9"
   rust-version: 1.66.1
-  node-version: 18.13.0
+  node-version: 18.12.0
```
FirelightFlagboy, line 13: Add comment why we use the version 18.12.0

PR #4081, opened 2023-02-23
[pytest] WIP `test_link_file_too_far_in_the_past`

FirelightFlagboy:
This test will generate a like too far in the past (before the creation of the workspace). We need to ensure that the app don't crash in this condition and the displayed error to the user is _somewhat_ clear.

@Max-7 We need your help to finish this test.


FirelightFlagboy:
Create an issue instead #4597

PR #4086, opened 2023-02-23
`Regex` are clonable

FirelightFlagboy:


TimeEngineer (approved):

src/regex.rs
```diff
@@ -4,6 +4,7 @@ use std::path::Path;
 
 use pyo3::{exceptions::PyValueError, pyclass, pymethods, types::PyType, PyResult};
 
+#[derive(Clone)]
 #[pyclass]
 pub(crate) struct Regex(pub libparsec::types::Regex);
```
TimeEngineer, line 9: I prefer `pyclass` first then `derive`, but ok

FirelightFlagboy: :sweat_smile:

FirelightFlagboy: done ;)

PR #4092, opened 2023-02-24
Configure `CodeQL` to check Java & Kotlin code

FirelightFlagboy:


touilleMan:
CI is failing due to a kotlin dependency being to recent for codeql

I think the issue is we don't pin the dependencies
- we just specify `org.jetbrains.kotlin.android` must be 1.8.0
- `org.jetbrains.kotlin.android` probably (I didn't check it own code so not 100% sure) depends on `org.jetbrains.kotlin.jvm`
- a new [version 1.8.10 of `org.jetbrains.kotlin.jvm` has been released](https://plugins.gradle.org/plugin/org.jetbrains.kotlin.jvm), so this is the one being used
- CodeQL complains ``org.jetbrains.kotlin.jvm` is too recent

See https://docs.gradle.org/current/userguide/dependency_locking.html for a solution

On top of there is a warning in the node install because the wrong version of node is used. I think we should first focus on solving the Gradle pinning issue, then try to entirely disable the node stuff (so npm install and also compilation of the web project). It's possible the Android project won't be affected by this (or it is affected in a way we could touch a couple of files to simulate the build of the js project).

PR #4098, opened 2023-02-27
Remove the use uuid in chunk_storage

TimeEngineer:
# What has changed ?


touilleMan (approved):

touilleMan (commented):
do we still need `uuid` as dependency in `core_fs`'s Cargo.toml ?

TimeEngineer:
It's used in `error.rs` with LocalMissError where it can be `EntryID` or `ChunkID`

PR #4146, opened 2023-03-02
Add `rel="noopener noreferrer"` to anchor in html template

FirelightFlagboy:
Add relationship `noopener` & `noreferrer` for anchor elements in HTML template used by the parsec backend.

- [noopener](https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/rel#noopener)
- [noreferrer](https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/rel#noreferrer)

For security reason, it's adviced to add those element to and html anchor when `target="_blank"`.


Max-7 (approved):

Max-7:
> Should I create a newsfragment for that ?

Nah, not necessary. It shouldn't affect the user at all.

PR #4148, opened 2023-03-02
Cache `run_sync_soon` in runtime.rs

TimeEngineer:
# What has changed ?


FirelightFlagboy (commented):
The change are ok, but how do we measure the performance increase/decrease ?

TimeEngineer:
It seems there is no difference as I expected

FirelightFlagboy:
How did you measure ?

TimeEngineer:
I check test time, because binding has downgraded

Issue #4126, opened 2023-03-01
Move Development docs to a single file in GitHub


mmmarcos:
Development docs are not referenced from the official documentation and therefore not displayed anywhere: 
https://github.com/Scille/parsec-cloud/tree/master/docs/development

If they contain up-to-date information, it is proposed to move contents to a single file (i.e. `HACKING.md`) that can be read directly on GitHub. Using Markdown is probably better because `.rst` files are automatically used to generate translations (`.pot`, `.po` files).

Q: There is a [Release Cheat-Sheet section](https://github.com/Scille/parsec-cloud/blob/master/docs/development/release.rst). Is it OK to leave this information here ? Otherwise, where should we move it ?

touilleMan:
see also the [CONTRIBUTING.md](https://github.com/Scille/parsec-cloud/blob/master/CONTRIBUTING.md) that should be also moved then

FirelightFlagboy:
@touilleMan From <https://opensource.guide/how-to-contribute/#anatomy-of-an-open-source-project>

It's recommended to leave `CONTRIBUTING.md` in the root folder

PR #4171, opened 2023-03-07
Add a docs dependencies group to pyprojecttoml

FirelightFlagboy:
Update `docutils` to `0.18.1` the last version supported by `sphinx-rtd-theme`.

### DIFF
--- a/poetry.lock
+++ b/poetry.lock
@@ -13,6 +13,18 @@ files = [
 ]
 
 [[package]]
+name = "alabaster"
+version = "0.7.13"
+description = "A configurable sidebar-enabled Sphinx theme"
+category = "dev"
+optional = false
+python-versions = ">=3.6"
+files = [
+    {file = "alabaster-0.7.13-py3-none-any.whl", hash = "sha256:1ee19aca801bbabb5ba3f5f258e4422dfa86f82f3e9cefb0859b283cdd7f62a3"},
+    {file = "alabaster-0.7.13.tar.gz", hash = "sha256:a27a4a084d5e690e16e01e03ad2b2e552c61a65469419b907243193de1a84ae2"},
+]
+
+[[package]]
 name = "altgraph"
 version = "0.17.3"
 description = "Python graph (network) package"
@@ -706,7 +718,7 @@ name = "charset-normalizer"
 version = "2.1.1"
 description = "The Real First Universal Charset Detector. Open, modern and actively maintained alternative to Chardet."
 category = "main"
-optional = true
+optional = false
 python-versions = ">=3.6.0"
 files = [
     {file = "charset-normalizer-2.1.1.tar.gz", hash = "sha256:5a3d016c7c547f69d6f81fb0db9449ce888b418b5b9952cc5e6e66843e9dd845"},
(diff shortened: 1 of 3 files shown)


mmmarcos:
Dependencies added to pyproject.toml LGTM :+1: 

IMO: 
1. This PR should not modify user docs (it is outside the PR scope) unless is a really small fix
2. Even for small fixes in the docs, `.po` files should not be re-generated : if I recall well, we agreed to update translations only on release, otherwise each change in the docs will require `.po` regeneration and translation. We should therefore leave `.po` files unchaged until next release. 
3. About Sphinx warnings: they are OK since we agreed to remove development docs away from user documentation. They will be there until we migrate those files (`quickstart`, `function_architecture`, ...) to a specific file outside user docs (see https://github.com/Scille/parsec-cloud/issues/4126).
4. Idem for `development` sub-directory.

poetry.lock
```diff
@@ -1162,12 +1174,24 @@ files = [
...
+    {file = "imagesize-1.4.1.tar.gz", hash = "sha256:69150444affb9cb0d5cc5a92b3676f0b2fb7cd9ae39e947a5e11a36b4497cd4a"},
+]
+
 [[package]]
 name = "importlib-metadata"
 version = "5.0.0"
 description = "Read metadata from Python packages"
 category = "main"
-optional = true
+optional = false
```
touilleMan, line 1194: There is multiple `optional = False` changes across this file, I think this is an error (poetry is handling group wrong ?)

For instance coming from master, if I checkout on this branch and start a regular install, I should not be installing any new packages, but this is not what happen:
```
$ POETRY_LIBPARSEC_BUILD_STRATEGY=no_build poetry install -E backend -E core
Installing dependencies from lock file

Package operations: 14 installs, 1 update, 0 removals

  • Installing alabaster (0.7.13)
  • Updating docutils (0.17.1 -> 0.18.1)
  • Installing imagesize (1.4.1)
  • Installing pygments (2.14.0)
  • Installing snowballstemmer (2.2.0)
  • Installing sphinxcontrib-applehelp (1.0.4)
  • Installing sphinxcontrib-devhelp (1.0.2)
  • Installing sphinxcontrib-htmlhelp (2.0.1)
  • Installing sphinxcontrib-jsmath (1.0.1)
  • Installing sphinxcontrib-qthelp (1.0.3)
  • Installing sphinxcontrib-serializinghtml (1.1.5)
[...]
```

touilleMan (changes requested):

poetry.lock
touilleMan: This is most likely because this group is not declared optional, see https://python-poetry.org/docs/master/managing-dependencies/#optional-groups

```ini
[tool.poetry.group.docs]
optional = true

[tool.poetry.group.docs.dependencies]
mkdocs = "*"
```

FirelightFlagboy: :open_mouth: I will look into that

touilleMan:
A modification in the `make.py i` command should be done to install this docs group by default (this is because `./make.py i` is designed as a no-brainer for developer, hence having it taking a bit more time by installing less useful dependencies is okay if it save headache when said developer wants to work on docs stuff)

poetry.lock
FirelightFlagboy: I've revert the change and regenerate the `poetry.lock` with

```shell
poetry lock --no-update
```

It result in the same lock file even with 

```toml
tool.poetry.group.docs.optional = true
```

FirelightFlagboy: - `charset-normalizer` is required by `requests`
- `importlib-metadata` is required by `sphinx`
- `jinja2` required by `quart` (optional) & `sphinx`

FirelightFlagboy: I've tried `poetry install -E core` and I don't have `jinja2` installed so it seems to be ok ?

touilleMan: So it seems pip install works as expected (i.e. it never installs sphinx&co when `docs` group is not specified):
```shell
$ POETRY_LIBPARSEC_BUILD_STRATEGY=no_build pip install -e .[core,backend]
$ pip freeze > pip_install_new.log
$ diff pip_install_old.log pip_install_new.log  # pip_install_old.log is generated from master
0a1
66c67
< -e git+ssh://git@github.com/Scille/parsec-cloud.git@b1c8817b6d1258b165d4300d9646a9ead9e88d33#egg=UNKNOWN
---
> -e git+ssh://git@github.com/Scille/parsec-cloud.git@7407dd006d45ddf7244b79c769ff45634a079ce7#egg=UNKNOWN
```
```shell
$ POETRY_LIBPARSEC_BUILD_STRATEGY=no_build poetry install -E backend -E core
$ pip freeze > poetry_install_new.log
$ diff poetry_install_old.log poetry_install_new.log  # poetry_install_old.log is generated from master
0a1
25c26
< docutils==0.17.1
---
> docutils==0.18.1
```

touilleMan (approved):

touilleMan (approved):

PR #4198, opened 2023-03-09
Fix file saving in mountpoints on MacOS from XCode and Office

### LINKED ISSUE #2330: [🐛 | Bug]: File saving not working as intended with Xcode
### Version

2.9.0

MacOS

### What happened ?

Can't save a file through XCode when editing it directly from a Parsec workspace, getting message `The document xxx could not be saved`.

For the file `filename.extension`, the file `filename-.extension` is created in the workspace, and the original get deleted. Its content is the same as the original file. So three issues:

- Original file is deleted, and with it its creation date and history
- New file has to be renamed to replace original
- Changes can't be saved, and auto-save messes up editing with pop-up messages telling it can't save

This could be related to #2211, and interestingly hasn't been fixed by it.

### Relevant output

_No response_

Ironicbay:
# What has changed ?

File saving on mountpoints from a few softwares was broken on MacOS. When saving changes made on a `file.md`, the steps are as follows:
1. XCode copies `file.md` into `file~.md`, without the current changes (backup file)
2. A temporary folder is created, inside of which the updated file is stored: `(A Document Being Saved By Xcode)/file.md`
3. The original file `file.md` is deleted
4. XCode tries to move `(A Document Being Saved By Xcode)/file.md` -> `file.md`
5. If step 4 is successful, the backup `file~.md` is deleted

With the current code, step 4 failed in fuse operations, being a `rename` operation from one folder to another; meaning the original file would disappear for the user, leaving only the copy `file~.md`.

A same-but-different issue happened on Word, for which as specific case has been made.

Closes #2330


FirelightFlagboy:
parsec core was removed so this PR cannot be merged

FirelightFlagboy:
Re-open that PR request for the 2.16 release

FirelightFlagboy:
Could you do a rebase

mmmarcos (approved):
LGTM. A couple of suggestions though.

mmmarcos (approved):
:+1:

PR #4202, opened 2023-03-09
Try unstable test with `-n auto`

TimeEngineer:
# What has changed ?


parsec/__init__.py
```diff
@@ -1,12 +1,15 @@
...
 
+from os import environ
+
 from parsec._version import __version__
 
 FEATURE_FLAGS = {
     # TODO: remove me once client connection oxidation is done
     # Activates RsBackendAuthenticatedCmds binding
-    "UNSTABLE_OXIDIZED_CLIENT_CONNECTION": False,
+    "UNSTABLE_OXIDIZED_CLIENT_CONNECTION": environ.get("UNSTABLE_OXIDIZED_CLIENT_CONNECTION")
```
touilleMan, line 11: Considering
```shell
export UNSTABLE_OXIDIZED_CLIENT_CONNECTION=
export UNSTABLE_OXIDIZED_CLIENT_CONNECTION=false
```
First will result in `False`
Second will result in `True`

This might be misleading, so better do this instead:
```suggestion
    "UNSTABLE_OXIDIZED_CLIENT_CONNECTION": environ.get("UNSTABLE_OXIDIZED_CLIENT_CONNECTION", "false").lower() == "true"
```

FirelightFlagboy:
Superseded by #4515

PR #4212, opened 2023-03-13
Setup gradle to use a lock file

FirelightFlagboy:
Changes
-------

- Setup gradle to use a lock files.
- Update the binding README concerning the android part.


.github/workflows/package-ci.yml
```diff
@@ -111,7 +111,7 @@ jobs:
         with:
           name: ${{ runner.os }}-${{ runner.arch }}-wheel
           path: dist/
-          if-no-file-found: error
+          if-no-files-found: error
```
touilleMan, line 114: Can you extract that in a separate PR for fast merging ?

oxidation/bindings/README.md
```diff
@@ -19,24 +33,77 @@ npm run build  # Generate index.node (basically a .so that node can load)
...
+  aarch64-linux-android \
+  x86_64-linux-android \
+```
+
+> The `bindings/android` directory is a valid Android project to generate a libparsec AAR.
+> However the AAR doesn't have to be built when using `../client/android` project (given it depends explicitly on the `bindings/android/libparsec`).
+
+#### Install Java development kit to build the application
+
+Having installed `openjdk-8` in the step [Install Android tools](#install-android-tools) wont be enough to build the android apps.
```
touilleMan, line 87: So we need two different version of openjdk ?

I can run just fine android tools with the default jre on ubuntu (i.e. openjdk 11)

so I would say we just need a recent enough version of openjdk, as we currently have no configuration in Gradle to specify a specific version

touilleMan (commented):
There is no hash signature in the lock files ? :'(

Have you had a look at https://docs.gradle.org/current/userguide/dependency_verification.html#sub:enabling-verification (the checksum verification part, we don't care about signature verification)

.github/workflows/package-ci.yml
FirelightFlagboy: Merged in #4214, would be remove when a rebase the branch

oxidation/bindings/README.md
FirelightFlagboy: Strange, mine don't like it :thinking: I will test more that one

FirelightFlagboy:
@touilleMan I've added verification metadata for gradle.

With the metadata verification & locking dependency, it seems we're duplicating information, Should we only keep the metadata one ?

touilleMan:
> would require a manual action on each PR made by dependabot for android

I think that's acceptable enough ;-)

oxidation/bindings/README.md
```diff
@@ -19,24 +33,134 @@ npm run build  # Generate index.node (basically a .so that node can load)
...
+
+```shell
+bash ./gradle --write-verification-metadata sha512 help
+```
+
+> It you want to update the verification metadata for a specific module use: `-p <module-path>`
+
+A more detailed documentation can be found here [Gradle - dependency verifycation](https://docs.gradle.org/current/userguide/dependency_verification.html#sub:enabling-verification)
+
+A more detailed documentation can be found here [Gradle - locking dependencies](https://docs.gradle.org/current/userguide/dependency_locking.html)
```
touilleMan, line 157: duplication of line 145 ?

FirelightFlagboy:
Ok

One last thing

Do we keep the locking of dependencies & the verification metadata ?

https://github.com/Scille/parsec-cloud/pull/4212#issuecomment-1472095385

oxidation/bindings/README.md
```diff
@@ -19,24 +33,134 @@ npm run build  # Generate index.node (basically a .so that node can load)
...
-*Note: The `bindings/android` directory is a valid Android project to generate a libparsec AAR.
-However the AAR doesn't have to be built when using `../client/android` project (given it depends explicitly on the `bindings/android/libparsec`).*
+> It you want to update the dependencies lock for a specific module use: `-p <module-path>`
+
+A more detailed documentation can be found here [Gradle - locking dependencies](https://docs.gradle.org/current/userguide/dependency_locking.html)
+
+### Update Gradle verification metadata
+
+```shell
+bash ./gradle --write-verification-metadata sha512 help
```
touilleMan, line 150: any reason for using sha512 instead of 🌟 sha256 🌟 ? (the Gradle documentation uses sha256 in it example, sha256 is also used in poetry.lock/package.lock/Cargo.lock)

oxidation/bindings/README.md
```diff
@@ -19,24 +33,134 @@ npm run build  # Generate index.node (basically a .so that node can load)
...
+
+#### Install Java development kit to build the application
+
+Install the latest LTS of `openjdk`:
+
+- Install the latest java LTS development kit (currently `JDK-17`).
+- If need to be, update your `JAVA_HOME` to point to the installed jdk version folder.
+
+### Update the Gradle dependencies locks
+
```
touilleMan, line 138: A small word about the difference between dependency lock and verification metadata would be useful here:
```
In most languages pinning dependency version and storing each dependency checksum is a single step (e.g. see `Cargo.lock`), but with Gradle this is divided into two steps:

- dependency locking  consists of pinning the list of dependencies (including transitive ones) use by the project
- verification metadata contains the list of dependency checksums
```

touilleMan (approved):

oxidation/bindings/README.md
FirelightFlagboy: Yes

oxidation/bindings/README.md
FirelightFlagboy: done

oxidation/bindings/README.md
FirelightFlagboy: I've the opportunity to use a more robust checksum algorithm, I take it.

touilleMan: If sha256 was broken we wouldn't use it at all, but that's not the case 😅 
I don't think the theoretical improved robustness is worth the ~2x increase in size of `verification-metadata.xml` and (more importantly) the cognitive load of having the reader wonder why in this very peculiar case we use a different algorithm than everywhere in the codebase and than the official gradle documentation

FirelightFlagboy: done

PR #4218, opened 2023-03-13
Oxidation log

### LINKED ISSUE #4219: Oxidation log in tests

TimeEngineer:
# What has changed ?

closes #4219

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -453,6 +453,8 @@ jobs:
           # By default, `libparsec_crypto` uses Rust Crypto
           cargo test --profile ci-rust --package libparsec_crypto --features use-sodiumoxide
         timeout-minutes: 30
+        env:
+          RUST_LOG: debug
 
       # Clippy basically compile the project, hence it's faster to run it in
       # the test-rust-matrix job where compilation cache is reused !
--- a/oxidation/libparsec/crates/platform_local_db/src/executor.rs
+++ b/oxidation/libparsec/crates/platform_local_db/src/executor.rs
@@ -67,8 +67,7 @@ impl SqliteExecutor {
             // it, however we log it anyway given the caller's unexpected drop
             // may also be the sign of a bug...
             if tx.send(res).is_err() {
-                // TODO: replace this by a proper warning log !
-                eprintln!("Caller has left");
+                log::warn!("Caller has left");
             }
         };
         let wrapped_job = Box::new(wrapped_job);
@@ -200,6 +199,7 @@ impl BackgroundSqliteExecutor {
                 Operation::Job(job) => job(&mut connection),
             }
         }
+        log::info!("BackgroundSqliteExecutor finished all his jobs");
     }
 }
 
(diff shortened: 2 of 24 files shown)


.github/workflows/ci.yml
```diff
@@ -445,6 +445,8 @@ jobs:
           # By default, `libparsec_crypto` uses Rust Crypto
           cargo test --profile ci-rust --package libparsec_crypto --features use-sodiumoxide
         timeout-minutes: 30
+        env:
+          RUST_LOG: trace
```
FirelightFlagboy, line 449: Should we use the trace level ? Won't it produce too much noise ?

FirelightFlagboy (commented):
Can you split this PR into 2 ?

- One that only add the log support
- One that add the log support in test

TimeEngineer:
The problem is that if I split, we would not have the println in the tests, so log are not used

.github/workflows/ci.yml
TimeEngineer: I set debug then

FirelightFlagboy (approved):

FirelightFlagboy:
@TimeEngineer Could you try if we have the log output when a rust test timeout ?

.github/workflows/ci.yml
```diff
@@ -441,10 +441,12 @@ jobs:
         shell: bash
         run: |
           set -ex
-          cargo test --profile ci-rust --workspace ${{ env.cargo-exclude-unused-crates }}
+          cargo test --profile ci-rust --workspace ${{ env.cargo-exclude-unused-crates }} -- --nocapture
```
FirelightFlagboy, line 444: There is a way to only show the output log for failing tests only (included the tests that time out) ?

FirelightFlagboy (changes requested):
I've looked at the tests output `--no-capture` isn't something we want when running all tests

.github/workflows/ci.yml
TimeEngineer: I don't find

FirelightFlagboy: The question can be rewritten as:

What append when a test fail without using `--nocapture` ?

TimeEngineer: It logs nothing in case of timeout

oxidation/libparsec/crates/platform_local_db/src/executor.rs
```diff
@@ -200,17 +199,22 @@ impl BackgroundSqliteExecutor {
...
     use crate::DatabaseError;
     use diesel::{connection::SimpleConnection, Connection, SqliteConnection};
+    use rstest::rstest;
+    use std::time::Duration;
 
     use super::SqliteExecutor;
 
-    #[tokio::test]
+    #[rstest]
+    #[timeout(Duration::new(60, 0))]
```
FirelightFlagboy, line 216:
```suggestion
    #[timeout(Duration::new(30, 0))]
```

`30sec` to `open -> close -> re-open` a file should be plenty enough

FirelightFlagboy (changes requested):

touilleMan (approved):
Should investigate a way to avoid using test-log decorator for each function (typically configuring once the log in the main of the tests)

oxidation/libparsec/crates/platform_local_db/src/executor.rs
TimeEngineer: Done in another PR

PR #4220, opened 2023-03-13
Finish oxidation of `parsec.crypto` to remove last need for `PyNaCl`

FirelightFlagboy:
The goal was to remove the dependencies to `PyNaCl`

For that we have to re-defined `CryptoError` & remplace the mock `mock_crypto`.

Other changes
-------------

- The file `parsec/crypto.py` was removed.
- The imports for `CryptoError` & co were udpated.


oxidation/libparsec/crates/crypto/src/rustcrypto/secret.rs
```diff
@@ -101,9 +101,19 @@ impl SecretKey {
     pub fn from_password(password: &str, salt: &[u8]) -> Self {
         let mut key = [0; KEY_SIZE];
 
-        ARGON2
-            .hash_password_into(password.as_bytes(), salt, &mut key)
-            .expect("Invalid salt");
+        if cfg!(feature = "test-unsecure-but-fast-secretkey-from-password") {
```
touilleMan, line 104: Should add a comment to explain 1) why this is needed 2) what algorithm we use instead

oxidation/libparsec/crates/crypto/src/sodiumoxide/secret.rs
```diff
@@ -91,16 +91,27 @@ impl SecretKey {
...
-        derive_key(
-            &mut key,
-            password.as_bytes(),
-            &salt,
-            OPSLIMIT_INTERACTIVE,
-            MEMLIMIT_INTERACTIVE,
-        )
-        .expect("Can't fail");
+
+        if cfg!(feature = "test-unsecure-but-fast-secretkey-from-password") {
```
touilleMan, line 95: same as above

touilleMan (changes requested):
Do we have an idea of the performance gain with fast-but-unsafe crypto in the tests ?

FirelightFlagboy:
> Do we have an idea of the performance gain with fast-but-unsafe crypto in the tests ?

On my workstation I go from ~10sec for basic test to ~8sec :shru

oxidation/libparsec/crates/crypto/src/rustcrypto/secret.rs
FirelightFlagboy: The feature name do not suffice ?

If no, would this suffice ?

```suggestion
   // During test we want to skip the `argon2` algorithm for hashing the password
   // Because it take sometime.
   // Force we replace argon with a very basic algorithm that copy the `password + salt` to the key first bytes.
   if cfg!(feature = "test-unsecure-but-fast-secretkey-from-password") {
```

oxidation/generate_test_data/protocol/pki.py
```diff
@@ -4,8 +4,8 @@
 
 from utils import *
 
+from uuid import uuid4
```
TimeEngineer, line 7: Don't import `uuid4`, use `EnrollmentID` instead

oxidation/libparsec/crates/crypto/src/rustcrypto/secret.rs
```diff
@@ -101,9 +101,19 @@ impl SecretKey {
...
+            let password_end = KEY_SIZE.min(password.len());
+
+            key[..password_end].copy_from_slice(&password.as_bytes()[..password_end]);
+
+            let salt_end = (KEY_SIZE - password_end).min(salt.len());
+
+            key[password_end..password_end + salt_end].copy_from_slice(&salt[..salt_end]);
```
TimeEngineer, line 111: `password_end + salt_end` seems equal to `KEY_SIZE`:
if `password.len() >= KEY_SIZE`, then `password_end == KEY_SIZE`, so `salt_end == 0` so the sum is equal to `KEY_SIZE`
if `password.len() < KEY_SIZE`, then `salt_end == KEY_SIZE - password_end`, so the sum is also equal to `KEY_SIZE`

TimeEngineer (changes requested):

oxidation/libparsec/crates/crypto/src/rustcrypto/secret.rs
FirelightFlagboy: > `password_end + salt_end` seems equal to `KEY_SIZE`: 
> if `password.len() >= KEY_SIZE`, then `password_end == `KEY_SIZE`, 

That the role of `password_end = min(KEY_SIZE, password.len())`

> so `salt_end == 0` so the sum is equal to `KEY_SIZE` 

Also the case with `salt_end = min(KEY_SIZE - password_end, salt.len())`

> if `password.len() < KEY_SIZE`, then `salt_end == KEY_SIZE - password_end`, so the sum is also equal to `KEY_SIZE`

-----

I'm don't understand this discussion, what do you want from it ?

1. Is it only to deconstruct the logic behind ?
2. Is it to propose / request a change ?

oxidation/libparsec/crates/crypto/src/rustcrypto/secret.rs
touilleMan: The feature name is good, but it is much better with your comment ;-)

(especially given the replacement algorithm is totally insecure and hence it's better for the reader to confirm this is intended)

touilleMan (approved):

oxidation/libparsec/crates/crypto/src/sodiumoxide/secret.rs
FirelightFlagboy:
```suggestion
        // During test we want to skip the `argon2` algorithm for hashing the password
        // Because it take sometime.
        // For that we replace argon with a very basic algorithm that copy the `password + salt` to the first key bytes.
        if cfg!(feature = "test-unsecure-but-fast-secretkey-from-password") {
```

oxidation/libparsec/crates/crypto/src/rustcrypto/secret.rs
TimeEngineer: My bad, forget about salt.len(), you are right

oxidation/libparsec/crates/crypto/src/rustcrypto/secret.rs
```diff
@@ -101,6 +101,9 @@ impl SecretKey {
     pub fn from_password(password: &str, salt: &[u8]) -> Self {
         let mut key = [0; KEY_SIZE];
 
+        // During test we want to skip the `argon2` algorithm for hashing the password
+        // Because it take sometime.
```
TimeEngineer, line 105: it takes time

oxidation/generate_test_data/protocol/pki.py
FirelightFlagboy: done

oxidation/libparsec/crates/crypto/src/rustcrypto/secret.rs
FirelightFlagboy: done

TimeEngineer (approved):

PR #4221, opened 2023-03-13
Define constant for custom HTTP code in `backend.rpc`

FirelightFlagboy:
I've replace the custom header value by using an enum to give those values a name.

### DIFF
--- a/parsec/backend/asgi/rpc.py
+++ b/parsec/backend/asgi/rpc.py
@@ -4,6 +4,7 @@ from __future__ import annotations
 
 from base64 import b64decode, b64encode
 from contextlib import asynccontextmanager
+from enum import Enum
 from typing import Any, AsyncIterable, AsyncIterator, NoReturn
 
 import trio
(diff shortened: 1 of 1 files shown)


parsec/backend/asgi/rpc.py
```diff
@@ -84,14 +85,27 @@ def _rpc_msgpack_rep(data: dict[str, object], api_version: ApiVersion) -> Respon
...
 
+class CustomHttpStatus(Enum):
+    BadAuthenticationInfo = 401
+    OrganizationOrInvitationInvalidOrNotFound = 404
+    BadAcceptType = 406
+    InvitationAlreadyUsedOrDeleted = 410
+    BadContentTypeOrInvalidBodyOrUnknownCommand = 415
+    UnsupportedApiVersion = 422
+    OrganizationExpired = 460
+    UserRevoked = 461
```
FirelightFlagboy, line 96: Should I use constant instead of an enum ?

TimeEngineer (approved):

parsec/backend/asgi/rpc.py
touilleMan: I think enum like this is fine :-)

touilleMan (approved):

mmmarcos:
I know this has been merged already (sorry I am late!), but just for my general culture: why do we call it "custom" HTTP codes? To me it seems that we are just assigning specific meanings to standard (registered) codes, which is probably OK. So `401` is still `Unauthorized`, `404` is `Not Found` and so on. Only `460` and `461` are custom (unregistred) codes.

If needed we might even use docstrings to specify meaning:
```python
class CustomHttpStatus(Enum):
    ...
    OrganizationExpired = 460
    """The Organization has expired"""

    UserRevoked = 461
    """The user has been revoked"""
```

Anyway, no need to change anything as this has already been merged :)

PR #4223, opened 2023-03-14
Rework paths filter used in the CI

FirelightFlagboy:
Rework the paths filter used in the CI that is used to determined which jobs's tasks (python, rust, web) should be run.

This change is more strict than the previous implementation

I've also simplified the filters, the ci only need to use the `*-jobs` corresponding to `{python,rust,web}-jobs` on its side.

Some paths aren't watched because we don't (currently) have tests for them (except the basic `Q&A`)


mmmarcos (approved):
LGTM. It's much cleaner now :+1:

PR #4226, opened 2023-03-14
Don't save python lib `libparsec` when on Github merge queue

### LINKED ISSUE #4153: CI only restore cache for queue branch
Currently the CI is saving cache entry created in the queue branch.
But those cache entry are used only one time.
To resolve that issue, we should only `save` the cache when we are not in a queue branch.

See

- [restore-only](https://github.com/actions/cache/blob/main/restore/action.yml)
- [save-only](https://github.com/actions/cache/blob/main/save/action.yml)

FirelightFlagboy:
Saving the python lib `libparsec` generated by the binding is useless on the merge queue

Because each branch are one time use so the cache isn't reused.

To correct that, I split the `cache` system into 2 phase `restore` & `save`:

- `restore` will always tries to restore the cache
- `save` will only save the cache when we aren't on Github merge queue and the rust code isn't modified

closes #4153

### DIFF
--- a/.github/filters/ci.yml
+++ b/.github/filters/ci.yml
@@ -26,12 +26,6 @@ rust-python-binding: &rust-python-binding src
 
 rust-web-binding: &rust-web-binding oxidation/bindings/web
 
-python: &python
-  - parsec
-  - tests
-  - make.py
-  - build.py
-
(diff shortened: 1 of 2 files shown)


mmmarcos (approved):
LGTM

PR #4228, opened 2023-03-14
Fix parsing error in BackendAddr with invalid hostname

touilleMan:
# What has changed ?


oxidation/libparsec/crates/types/src/addr.rs
```diff
@@ -16,6 +16,15 @@ const PARSEC_SCHEME: &str = "parsec";
...
 const PARSEC_NO_SSL_DEFAULT_PORT: u16 = 80;
 
+// Url has a special way to parse http/https schemes. This is because those kind
+// of url have special needs (for instance host cannot be empty).
+// The issue is we want our custom parsec scheme to work in a similar fashion, but
+// we cannot just tell Url "apply to parsec:// whatever rules you use for http://".
+// So instead we have to rely on a hack and manually replace our scheme by http before
+// handing it to Url :'(
+// see. https://github.com/servo/rust-url/issues/763
+struct ParsecUrlAsHTTPScheme<'a>(pub &'a Url);
```
FirelightFlagboy, line 26: Use doc-comment

```suggestion
/// Url has a special way to parse http/https schemes. This is because those kind
/// of url have special needs (for instance host cannot be empty).
/// The issue is we want our custom parsec scheme to work in a similar fashion, but
/// we cannot just tell Url "apply to parsec:// whatever rules you use for http://".
/// So instead we have to rely on a hack and manually replace our scheme by http before
/// handing it to Url :'(
/// see. https://github.com/servo/rust-url/issues/763
struct ParsecUrlAsHTTPScheme<'a>(pub &'a Url);
```

oxidation/libparsec/crates/types/src/addr.rs
```diff
@@ -26,7 +35,7 @@ macro_rules! impl_common_stuff {
             pub fn to_http_redirection_url(&self) -> Url {
                 let mut url = self.base.to_http_url(None);
                 url.path_segments_mut()
-                    .unwrap_or_else(|()| unreachable!())
+                    .expect("url is not a cannot-be-a-base")
```
FirelightFlagboy, line 38: You use double negation can you rewrite this message ?

oxidation/libparsec/crates/types/src/addr.rs
```diff
@@ -99,11 +94,47 @@ macro_rules! impl_common_stuff {
...
-                // Note `Url::parse` takes care of percent-encoding for query params
-                let parsed = Url::parse(url).map_err(|_| "Invalid URL")?;
-                let pairs = parsed.query_pairs();
+                lazy_static! {
+                    static ref SCHEME_WITH_SLASHES: String = format!("{}://", PARSEC_SCHEME);
+                }
+                lazy_static! {
+                    static ref SCHEME_ERROR_MSG: String =
+                        format!("Must start with `{}://`", PARSEC_SCHEME);
+                }
```
FirelightFlagboy, line 103: Use constant value instead (use `concat!`)

https://doc.rust-lang.org/std/macro.concat.html

oxidation/libparsec/crates/types/src/addr.rs
```diff
@@ -157,42 +188,27 @@ struct BaseBackendAddr {
...
 
         let default_port = if use_ssl {
             PARSEC_SSL_DEFAULT_PORT
         } else {
             PARSEC_NO_SSL_DEFAULT_PORT
         };
         let port = parsed
+            .0
             .port()
             .and_then(|p| if p == default_port { None } else { Some(p) });
```
FirelightFlagboy, line 213: Could this logic moved to `ParsecUrlAsHTTPScheme` ?

Could be logical that `ParsecUrlAsHTTPScheme` provides methods to get the `hostname`, `port` & `use_ssl`

oxidation/libparsec/crates/types/src/addr.rs
```diff
@@ -319,10 +332,13 @@ impl BackendAddr {
         self.base.to_http_url(path)
     }
 
-    fn _from_url(parsed: &Url, pairs: &url::form_urlencoded::Parse) -> Result<Self, AddrError> {
-        let base = BaseBackendAddr::from_url(parsed, pairs)?;
+    fn _from_url(
+        parsed: &ParsecUrlAsHTTPScheme,
+        _pairs: &url::form_urlencoded::Parse,
```
FirelightFlagboy, line 337: If you don't need it, remove it ;)

FirelightFlagboy (changes requested):

oxidation/libparsec/crates/types/src/addr.rs
touilleMan: no, because `cannot-be-a-base` is the specific term used in the URL RFC (https://docs.rs/reqwest/latest/reqwest/struct.Url.html#method.cannot_be_a_base)

oxidation/libparsec/crates/types/src/addr.rs
touilleMan: This doesn't work because concat only allow literal (so cannot use `PARSEC_SCHEME`)

oxidation/libparsec/crates/types/src/addr.rs
touilleMan: it is needed by the other `_from_url` (given they all got called by the same common macro)

FirelightFlagboy: Seems like a trait would be nice ;)

oxidation/libparsec/crates/types/src/addr.rs
touilleMan: I'd rather have it here and keep `ParsecUrlAsHTTPScheme` as simple as possible: it is just a simple wrapper to indicate the Url object is a bit hacky, but in a perfect world it would be needed and a parsec:// url would be passed here

touilleMan:
@FirelightFlagboy can you rebase the branch for me plz ?

FirelightFlagboy:
> @FirelightFlagboy can you rebase the branch for me plz ?

WIP

oxidation/libparsec/crates/types/src/addr.rs
```diff
@@ -433,7 +449,7 @@ impl BackendOrganizationAddr {
 
     fn _to_url(&self, mut url: Url) -> Url {
         url.path_segments_mut()
-            .unwrap_or_else(|()| unreachable!())
+            .expect("url is not a cannot-be-a-base")
```
FirelightFlagboy, line 452:
```suggestion
            .expect("url should be a cannot-be-a-base")
```

oxidation/libparsec/crates/types/src/addr.rs
```diff
@@ -825,7 +850,7 @@ impl BackendInvitationAddr {
...
```
FirelightFlagboy, line 853:
```suggestion
            .expect("url should be a cannot-be-a-base")
```

oxidation/libparsec/crates/types/src/addr.rs
```diff
@@ -920,7 +948,7 @@ impl BackendPkiEnrollmentAddr {
...
```
FirelightFlagboy, line 951:
```suggestion
            .expect("url should be a cannot-be-a-base")
```

oxidation/libparsec/crates/types/src/addr.rs
touilleMan: expect means the message describe why we expect it to be ok (cf. https://doc.rust-lang.org/std/result/enum.Result.html#recommended-message-style)

So here we expect the url not to be a cannot-be-a-base, otherwise [`Url::path_segments` returns `None`](https://docs.rs/url/latest/url/struct.Url.html#method.path_segments)

touilleMan: I've changed the wording to make it more explicit what is expected

oxidation/libparsec/crates/types/src/addr.rs
touilleMan: could be, but this code is "fine enough" given we won't touch it much now that it is stable

FirelightFlagboy (approved):

FirelightFlagboy (approved):

PR #4233, opened 2023-03-15
Replace `string error` with an enum in `libparsec_types::addr`

### LINKED ISSUE #4230: Replace str error in `libparsec_types::addr`
When reviewing #4228, I see that we are using string as error message in `libparsec_types::addr` (`oxidation/libparsec/crates/types/src/addr.rs`).

We should use an enum instead as it makes it easier to compare an error when testing and we can also provide more context

FirelightFlagboy:
Using an enum instead of string errors:

- Reduce the amount of duplicate string found in the codebase.
- Reduce the amount of place where we could have typo.

closes #4230

### DIFF
--- a/oxidation/libparsec/crates/types/src/addr.rs
+++ b/oxidation/libparsec/crates/types/src/addr.rs
@@ -46,7 +46,8 @@ macro_rules! impl_common_stuff {
                 let url = &url_with_forced_custom_scheme;
 
                 // Note `Url::parse` takes care of percent-encoding for query params
-                let mut parsed = Url::parse(url).map_err(|_| "Invalid URL")?;
+                let mut parsed =
+                    Url::parse(url).map_err(|e| AddrError::InvalidUrl(url.to_string(), e))?;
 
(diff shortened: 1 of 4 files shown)


oxidation/libparsec/crates/types/src/addr.rs
```diff
@@ -147,7 +150,38 @@ macro_rules! impl_common_stuff {
     };
 }
 
-type AddrError = &'static str;
+/// The enum constains the list of possible errors when working with url / addr.
+#[derive(Debug, thiserror::Error, PartialEq, Eq)]
+pub enum AddrError {
+    /// We failed to parse the url
+    #[error("Can't parse the raw url `{0}`: {1}")]
```
mmmarcos, line 157:
```suggestion
    #[error("Cannot parse raw url `{0}`: {1}")]
```

mmmarcos (approved):
LGTM. Just a minor suggestion to use the non-contracted forms (does not, cannot, ...) which are less informal.

PR #4234, opened 2023-03-15
Don't save python lib `libparsec` when on Github merge queue

### LINKED ISSUE #4153: CI only restore cache for queue branch

FirelightFlagboy:
Saving the python lib `libparsec` generated by the binding is useless on the merge queue

Because each branch are one time use so the cache isn't reused.

To correct that, I split the `cache` system into 2 phase `restore` & `save`:

- `restore` will always tries to restore the cache
- `save` will only save the cache when we aren't on Github merge queue and the rust code isn't modified

depends on #4223
closes #4153

> This is a re-open of the PR #4226 because it depends on the PR #4223

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -280,12 +280,10 @@ jobs:
       # libparsec is slow to compile, so we save it in cache and skip the
       # compilation entirely if the Rust code hasn't changed !
 
-      - name: Use cached libparsec if Rust hasn't been modified
+      - name: Restore libparsec if Rust hasn't been modified
         id: cache-libparsec
-        if: >-
-          steps.python-changes.outputs.run == 'true'
-          && steps.rust-changes.outputs.run != 'true'
-        uses: actions/cache@v3
+        if: steps.python-changes.outputs.run == 'true'
+        uses: actions/cache/restore@88522ab9f39a2ea568f7027eddc7d8d8bc9d59c8 # pin v3.3.1
         with:
           # Key cache contains a hash of all the files that are used to produce _parsec.so
           # Hence if we have a cache hit we know that there is no need for a rebuild !
@@ -326,6 +324,26 @@ jobs:
           if ${{ steps.cache-libparsec.outputs.cache-hit == 'true' }}; then export POETRY_LIBPARSEC_BUILD_STRATEGY=no_build; fi
           python make.py python-ci-install
         timeout-minutes: 20
+
+      # We only save the libparsec lib when:
+      # - We are not in a github queue branch (they're a one time use so caching won't help)
+      # - The rust code isn't modifed.
+      # - We haven't add a cache it.
+      - name: Save cached libparsec to be reused on later call
+        if: >-
+          steps.python-changes.outputs.run == 'true'
+          && steps.changes.outputs.rust-changes != 'true'
+          && steps.cache-libparsec.outputs.cache-hit != 'true'
+          && contains(github.ref, 'gh-readonly-queue') != 'true'
+        uses: actions/cache/save@88522ab9f39a2ea568f7027eddc7d8d8bc9d59c8 # pin v3.3.1
+        with:
+          # Key cache contains a hash of all the files that are used to produce _parsec.so
+          # Hence if we have a cache hit we know that there is no need for a rebuild !
+          key: ${{ matrix.os }}-${{ hashFiles('src/**', 'oxidation/libparsec/**', 'rust-toolchain.toml', 'Cargo.toml', 'Cargo.lock') }}-libparsec-python
+          path: |
+            parsec/_parsec.*.pyd
+            parsec/_parsec.*.so
+        timeout-minutes: 2
 
       - name: Basic tests
         if: steps.python-changes.outputs.run == 'true'
--- a/.github/filters/ci.yml
+++ b/.github/filters/ci.yml
@@ -26,12 +26,6 @@ rust-python-binding: &rust-python-binding src
 
 rust-web-binding: &rust-web-binding oxidation/bindings/web
 
-python: &python
-  - parsec
-  - tests
-  - make.py
-  - build.py
-
 python-dependencies-project: &python-dependencies-project
   - pyproject.toml
   - poetry.lock
@@ -50,6 +44,18 @@ web: &web
   - oxidation/client/*.js
   - oxidation/client/*.ts
 
+python-changes: &python-changes
+  - *python-dependencies-project
+  - parsec
+  - tests
+  - make.py
+  - build.py
+
+rust-changes: &rust-changes
+  - *rust-dependencies-workspace
+  - *rust-libparsec
+  - *rust-toolchain
+
 # The python jobs need to be run when:
 # - The ci workflow has changed
 # - The action `setup-python-poetry` has changed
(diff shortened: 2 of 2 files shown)


.github/workflows/ci.yml
```diff
@@ -327,6 +325,26 @@ jobs:
           python make.py python-ci-install
         timeout-minutes: 20
 
+      # We only save the libparsec lib when:
+      # - We are not in a github queue branch (they're a one time use so caching won't help)
+      # - The rust code isn't modifed.
+      # - We haven't add a cache it.
```
touilleMan, line 331: "We haven't already cached it" ?

.github/workflows/ci.yml
```diff
@@ -327,6 +325,26 @@ jobs:
           python make.py python-ci-install
         timeout-minutes: 20
 
+      # We only save the libparsec lib when:
+      # - We are not in a github queue branch (they're a one time use so caching won't help)
+      # - The rust code isn't modifed.
```
touilleMan, line 330: isn't it the opposite: we want  to save libparsec in cache because the rust code has been modified (and hence libparsec has been rebuilt from the sources) ?

.github/workflows/ci.yml
```diff
@@ -327,6 +325,26 @@ jobs:
...
+      # - We haven't add a cache it.
+      - name: Save cached libparsec to be reused on later call
+        if: >-
+          steps.python-changes.outputs.run == 'true'
```
touilleMan, line 334: Can't we just use `steps.cache-libparsec.outputs.cache-hit != 'true' && contains(github.ref, 'gh-readonly-queue') != 'true'` ?

i.e. we want to save the cache if it wasn't already present and if we are not in a github queue branch

(the idea is to let `cache-libparsec` job determine whether or not libparsec should be built, then not duplicate this check here to avoid inconsistency bugs)

touilleMan (changes requested):

.github/workflows/ci.yml
```diff
@@ -327,6 +325,26 @@ jobs:
...
+          && steps.changes.outputs.rust-changes != 'true'
+          && steps.cache-libparsec.outputs.cache-hit != 'true'
+          && contains(github.ref, 'gh-readonly-queue') != 'true'
```
mmmarcos, line 337: Wouldn't it be clearer to use `== 'false'` instead of `!= 'true'`?
```suggestion
          && steps.changes.outputs.rust-changes == 'false'
          && steps.cache-libparsec.outputs.cache-hit == 'false'
          && contains(github.ref, 'gh-readonly-queue') == 'false'
```

.github/workflows/ci.yml
FirelightFlagboy: https://github.com/Scille/parsec-cloud/blob/2fc9e47d7dd8cbf2e0611ad1988d48a370d08467/.github/workflows/ci.yml#L280-L281

I've a comment made by you that said otherwise ;)

Here is the following condition to meet:

- On a PR, if we change the rust code, we could not reuse the pre-compiled lib.
- On the merge queue, we never want to save the compiled lib (the branch is a one-time use)
- On master, we want to save the pre-compile lib for later PR

.github/workflows/ci.yml
FirelightFlagboy: Use a suggestion ;)

````markdown
```suggestion
foo bar
```
````

File: .github/workflows/ci.yml (hunk 4 above)
Comment (FirelightFlagboy) [reply .github/workflows/ci.yml:337]: Yes, it would but it's more robust when the condition return `''` instead of `'false'`

File: .github/workflows/ci.yml (hunk 3 above)
Comment (FirelightFlagboy) [reply .github/workflows/ci.yml:334]: Almost perfect :p

Just missing the part for the PR request ;)

```txt
steps.cache-libparsec.outputs.cache-hit != 'true' # We have a cache hit, so we don't need to save again
&& contains(github.ref, 'gh-readonly-queue') != 'true' # We are in a queue branch so we don't save it
&& (contains(github.ref, 'refs/pull') == 'true' && steps.changes.outputs.rust-changes == 'true') != 'true' # We are in a PR and the rust code have changed
```

FirelightFlagboy: WDYT ?

.github/workflows/ci.yml
FirelightFlagboy: Continue the discussion on https://github.com/Scille/parsec-cloud/pull/4234#discussion_r1138424381

.github/workflows/ci.yml
touilleMan: we indeed want to keep in cache the modified libparsec in the branch (because it makes working on the branch much more convenient, and it libparsec is only ~10mo so it's very light on the cache)

> Should we have a condition on the PR ?

I don't understand what you mean by "condition on the PR" is this a github action specific term ?

My point is only:
- if we had a cache hit there is no point in saving the cache
- if we didn't have a cache hit it is either because 1) we didn't try to setup the cache (i.e. because we don't need to test the python code) or 2) because of actual cache miss (i.e. the rust code has changed)

I forgot about cache miss option 1) in my previous comment.

So thing would be more like:
```
# Something has changed among Python code, libparsec Python bindings or libparsec Rust codeSomething needed to generate the Python wheel has changed...
steps.changes.need-run-python-tests == 'true'
# ...and we had to rebuild libparsec
&& steps.cache-libparsec.outputs.cache-hit != 'true'
# ...and we are not in the merge queue branch (where caching is useless given the branch is short lived)
&& contains(github.ref, 'gh-readonly-queue') != 'true'
```

.github/workflows/ci.yml
touilleMan: indeed, the issue is it's very easy to break string comparison. So `!= 'true'` is less readable but save us from debugging headaches with the CI ^^

.github/workflows/ci.yml
FirelightFlagboy: :+1:

FirelightFlagboy: done

.github/workflows/ci.yml
FirelightFlagboy: done

.github/workflows/ci.yml
```diff
@@ -327,6 +325,24 @@ jobs:
...
+      - name: Save cached libparsec to be reused on later call
+        if: >-
+          steps.python-changes.outputs.run == 'true'
+          && steps.cache-libparsec.outputs.cache-hit != 'true'
+          && contains(github.ref, 'gh-readonly-queue') != 'true'
+        uses: actions/cache/save@88522ab9f39a2ea568f7027eddc7d8d8bc9d59c8 # pin v3.3.1
+        with:
+          # Key cache contains a hash of all the files that are used to produce _parsec.so
+          # Hence if we have a cache hit we know that there is no need for a rebuild !
+          key: ${{ matrix.os }}-${{ hashFiles('src/**', 'oxidation/libparsec/**', 'rust-toolchain.toml', 'Cargo.toml', 'Cargo.lock') }}-libparsec-python
```
touilleMan, line 340: can we create a step dedicated to generate this key ?

This has two advantages:
- avoid producing an invalid value if one of the input file gets modified (or something is added in the directory) during the build
- (bonus) save a bit of time

touilleMan (approved):

.github/workflows/ci.yml
FirelightFlagboy: done

PR #4238, opened 2023-03-16
Update gen web rust template to convert rust error to string

FirelightFlagboy:
This was required by the change made in #4233 where we change the error type for AddrError from a string to an enum. But `TypeError` expect a `string` in the `TypeError::new` (would be nice if it allows `impl Display` instead :thinking:)

Other change
------------

Rework generator script to format generated rust code

> Before exiting the script, the generator will call `cargo-fmt` on the modified packages.

### DIFF
--- a/.github/filters/ci.yml
+++ b/.github/filters/ci.yml
@@ -1,6 +1,6 @@
 ci-workflow: &ci-workflow .github/workflows/ci.yml
 
-rust-libparsec: &rust-libparsec oxidation/libparsec
+rust-libparsec: &rust-libparsec oxidation/libparsec/**
 
 rust-dependencies-workspace: &rust-dependencies-workspace
   - Cargo.toml
@@ -8,7 +8,7 @@ rust-dependencies-workspace: &rust-dependencies-workspace
 
 rust-toolchain: &rust-toolchain rust-toolchain.toml
 
-rust-python-binding: &rust-python-binding src
+rust-python-binding: &rust-python-binding src/**
 
 # TODO: We currently don't test the electron application
 # So we don't have to watch the electron binding (likewise for the client code related to electron)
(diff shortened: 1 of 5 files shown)


.github/filters/ci.yml
```diff
@@ -1,14 +1,14 @@
 ci-workflow: &ci-workflow .github/workflows/ci.yml
 
-rust-libparsec: &rust-libparsec oxidation/libparsec
+rust-libparsec: &rust-libparsec oxidation/libparsec/**
```
touilleMan, line 3: What is this for ? it wasn't working properly before ?

.github/filters/ci.yml
```diff
@@ -43,10 +43,10 @@ new-client-dependencies-project: &new-client-dependencies-project
...
 web: &web
-  - oxidation/client/public
-  - oxidation/client/resources
-  - oxidation/client/src
-  - oxidation/client/tests
+  - oxidation/client/public/**
+  - oxidation/client/resources/**
+  - oxidation/client/src/**
+  - oxidation/client/tests/**
   - oxidation/client/*.js
```
touilleMan, line 50: Is it normal we only take into consideration .js/.ts in the root folder ? I'd say we want everything (especially given `ionic.config.json` exists ^^)

touilleMan (approved):

.github/filters/ci.yml
FirelightFlagboy: We don't want everything given we store the code for `android`, `ios` & `electron` (changing the code for one of those don't require to run the web tests)

.github/filters/ci.yml
FirelightFlagboy: The `/**` was to watch the change recursively on the folder `oxidation/libparsec`

.github/filters/ci.yml
FirelightFlagboy: I've changed the pattern for `oxidation/client/*.{ts,js,json}`

Issue #4176, opened 2023-03-07
[🐛 | Bug]: [RUST] `fail_reopen_database` test fails with timeout


TimeEngineer:
### Version

2.15.0

### What platforms you were running our software when the bug happened ?

Linux

### What happened ?

Test fail: https://github.com/Scille/parsec-cloud/actions/runs/4355599502/jobs/7612437395

TimeEngineer:
Another fail: https://github.com/Scille/parsec-cloud/actions/runs/4354487782/jobs/7609848850

FirelightFlagboy:
https://github.com/Scille/parsec-cloud/actions/runs/4362308265/jobs/7627050345

FirelightFlagboy:
Require #3939

FirelightFlagboy:
https://github.com/Scille/parsec-cloud/actions/runs/4382950634/jobs/7672606483

FirelightFlagboy:
Resolved in #4254

PR #4241, opened 2023-03-16
Add timeout to `fail_reopen_database`

FirelightFlagboy:
This timeout would help analyse the flakyness of the test on the CI

see #4176

### DIFF
--- a/oxidation/libparsec/crates/platform_local_db/src/executor.rs
+++ b/oxidation/libparsec/crates/platform_local_db/src/executor.rs
@@ -207,9 +207,14 @@ impl BackgroundSqliteExecutor {
 mod tests {
     use crate::DatabaseError;
     use diesel::{connection::SimpleConnection, Connection, SqliteConnection};
+    use rstest::rstest;
+    use std::time::Duration;
 
     use super::SqliteExecutor;
 
+    #[rstest]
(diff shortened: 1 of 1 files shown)


oxidation/libparsec/crates/platform_local_db/src/executor.rs
```diff
@@ -207,9 +207,13 @@ impl BackgroundSqliteExecutor {
 mod tests {
     use crate::DatabaseError;
     use diesel::{connection::SimpleConnection, Connection, SqliteConnection};
+    use rstest::rstest;
+    use std::time::Duration;
 
     use super::SqliteExecutor;
 
+    #[rstest]
+    #[timeout(Duration::from_secs(30))]
```
touilleMan, line 216: Add a TODO comment to explain why this is need (and that it's a temporary thing)

FirelightFlagboy: I don't think we would remove the `timeout` 

```suggestion
    // FIXME: We need a timeout because the test seems to block on the CI, see #4176 for more information
    #[timeout(Duration::from_secs(30))]
```

FirelightFlagboy: done

touilleMan (approved):

PR #4244, opened 2023-03-16
Add PartialEq & Eq for CommandError & InviteError

TimeEngineer:
# What has changed ?


FirelightFlagboy (approved):
Given this should be used only in test, should we use ?


```rust
// For CommandError
#[cfg(any(test, feature = "test-utils"))]
// For InvitationError
#[cfg_attr(any(test, feature = "test-utils"), derive(PartialEq, Eq)]
```

TimeEngineer:
> Given this should be used only in test, should we use ?
> 
> ```rust
> // For CommandError
> #[cfg(any(test, feature = "test-utils"))]
> // For InvitationError
> #[cfg_attr(any(test, feature = "test-utils"), derive(PartialEq, Eq)]
> ```

Yes it seems a good idea, need an another PR to change every errors

oxidation/libparsec/crates/client_connection/src/error.rs
```diff
@@ -67,6 +67,43 @@ pub enum CommandError {
...
+            (Self::MissingApiVersion, Self::MissingApiVersion) => true,
+            (Self::MissingSupportedApiVersions, Self::MissingSupportedApiVersions) => true,
+            // For the moment, NoResponseError are the same
+            (Self::NoResponse(..), Self::NoResponse(..)) => true,
+            (Self::RevokedUser, Self::RevokedUser) => true,
+            // For the moment, SerializationError are the same
+            (Self::Serialization(..), Self::Serialization(..)) => true,
+            (
+                Self::UnsupportedApiVersion {
+                    api_version: v0, ..
```
touilleMan, line 93: v0 vs v1 is misleading (it seems you are refering to API version v0 vs v1, where it left version vs right version)

replace it by `right_api_version/left_api_version` ?

oxidation/libparsec/crates/client_connection/src/error.rs
```diff
@@ -67,6 +67,43 @@ pub enum CommandError {
     WrongApiVersion(String),
 }
 
+impl PartialEq for CommandError {
```
touilleMan, line 70: Add `// Custom equality to skip comparison of some fields`

TimeEngineer: Done

oxidation/libparsec/crates/client_connection/src/error.rs
TimeEngineer: Done

touilleMan:
> Given this should be used only in test, should we use ?

That's cheap to add so I'd say go for it (and add a comment of course ^^)

FirelightFlagboy (approved):

PR #4245, opened 2023-03-16
Improve electron template for string convertion

FirelightFlagboy:
Rework the electron template when generating the code that do string conversion in rust. This allow a more configurable code that gives more hints to the rust compiler that need it.

> This also correct the error when converted `AddrError` to `TypeError`

### DIFF
--- a/oxidation/bindings/generator/generate.py
+++ b/oxidation/bindings/generator/generate.py
@@ -77,6 +77,7 @@ env.globals["raise"] = _raise_helper
 class BaseTypeInUse:
     # Name of the type familly (e.g. `bytes`, `struct`, `variant`), to be defined in subclass
     kind: str
+    namespace: str = "libparsec"
 
     @staticmethod
     def parse(param: str) -> "BaseTypeInUse":
@@ -116,6 +117,9 @@ class BaseTypeInUse:
         elif isinstance(param, type) and issubclass(param, StrBasedType):
             return StrBasedTypeInUse(
                 name=param.__name__,
+                type=getattr(param, "type", param.__name__),
+                namespace=getattr(param, "namespace", "libparsec"),
+                from_str_error=getattr(param, "from_str_error", "_"),
                 custom_from_rs_string=getattr(param, "custom_from_rs_string", None),
                 custom_to_rs_string=getattr(param, "custom_to_rs_string", None),
             )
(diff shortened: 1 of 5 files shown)


oxidation/bindings/generator/generate.py
```diff
@@ -175,7 +179,11 @@ class RefTypeInUse(BaseTypeInUse):
 class StrBasedTypeInUse(BaseTypeInUse):
     kind = "str_based"
     name: str
+    type: str
+    namespace: str
 
+    # The error returned when trying to parse the value from a string.
+    from_str_error: str
```
touilleMan, line 186: I don't see the need for this given we already specified in `custom_from/to_rs_string` the return type is AsRef<str>

oxidation/bindings/generator/templates/binding_electron_meths.rs.j2
```diff
@@ -101,15 +101,18 @@ JsFunction
 {%- elif type.kind == "str" -%}
 {{ js_val }}.value({{ mut_cx_ref }})
 {%- elif type.kind == "str_based" -%}
+{
 {% if type.custom_from_rs_string -%}
-{let custom_from_rs_string = {{ type.custom_from_rs_string }};
-match custom_from_rs_string({{ js_val }}.value({{ mut_cx_ref }}))
+    let custom_from_rs_string = {{ type.custom_from_rs_string }};
+    let value: Result<{{ type.namespace }}::{{ type.type }}, {{ type.from_str_error }}> = custom_from_rs_string({{ js_val }}.value({{ mut_cx_ref }}));
```
touilleMan, line 107: 👍 this should make `custom_from_rs_string` easier to debug !

oxidation/bindings/generator/generate.py
touilleMan: We could however add typing info in the `custom_from/to_rs_string`:

```rust
    let custom_from_rs_string: Fn(String) -> Result<{{ type.rs_type }}, AsRef<str>> = {{ type.custom_from_rs_string }};
    let value: Result<{{ type.rs_type }}, _> = custom_from_rs_string({{ js_val }}.value({{ mut_cx_ref }}));
```

touilleMan (changes requested):

FirelightFlagboy:
Superseded by #4247

PR #4247, opened 2023-03-17
Simpler web electron error convertion

FirelightFlagboy:
- Use `AsRef<str>` to support more error type for `TypeError` and `ctx.throw_type_error`
- Remove extra space added when generating the typescript definition files from the jinja template

### DIFF
--- a/oxidation/bindings/electron/src/meths.rs
+++ b/oxidation/bindings/electron/src/meths.rs
@@ -177,8 +177,9 @@ fn struct_clientconfig_js_to_rs<'a>(
     let preferred_org_creation_backend_addr = {
         let js_val: Handle<JsString> = obj.get(cx, "preferredOrgCreationBackendAddr")?;
         {
-            let custom_from_rs_string =
-                |s: String| -> Result<_, _> { libparsec::BackendAddr::from_any(&s) };
+            let custom_from_rs_string = |s: String| -> Result<_, String> {
+                libparsec::BackendAddr::from_any(&s).map_err(|e| e.to_string())
+            };
             match custom_from_rs_string(js_val.value(cx)) {
                 Ok(val) => val,
                 Err(err) => return cx.throw_type_error(err),
@@ -791,8 +792,9 @@ fn test_new_testbed(mut cx: FunctionContext) -> JsResult<JsPromise> {
     let test_server = match cx.argument_opt(1) {
         Some(v) => match v.downcast::<JsString, _>(&mut cx) {
             Ok(js_val) => Some({
-                let custom_from_rs_string =
-                    |s: String| -> Result<_, _> { libparsec::BackendAddr::from_any(&s) };
+                let custom_from_rs_string = |s: String| -> Result<_, String> {
+                    libparsec::BackendAddr::from_any(&s).map_err(|e| e.to_string())
+                };
                 match custom_from_rs_string(js_val.value(&mut cx)) {
                     Ok(val) => val,
                     Err(err) => return cx.throw_type_error(err),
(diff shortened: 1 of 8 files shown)


FirelightFlagboy:
https://github.com/Scille/parsec-cloud/actions/runs/4449507257

touilleMan (approved):
image

oxidation/bindings/electron/src/meths.rs
```diff
@@ -21,7 +21,7 @@ fn struct_availabledevice_js_to_rs<'a>(
                 |s: String| -> Result<_, &'static str> { Ok(std::path::PathBuf::from(s)) };
             match custom_from_rs_string(js_val.value(cx)) {
                 Ok(val) => val,
-                Err(err) => return cx.throw_type_error(err),
+                Err(err) => return cx.throw_type_error(AsRef::<str>::as_ref(&err)),
```
TimeEngineer, line 24: `AsRef::<str>::as_ref` is not needed, because `throw_type_error` take `AsRef<str>`  as argument

TimeEngineer (changes requested):

oxidation/bindings/generator/templates/binding_web_meths.rs.j2
```diff
@@ -58,7 +58,7 @@ let {{ param_name }} =
 {
     let custom_from_rs_string = {{ type.custom_from_rs_string }};
     custom_from_rs_string({{ param_name }})
-        .map_err(|e| TypeError::new(&e.to_string()))
+        .map_err(|e| TypeError::new(AsRef::<str>::as_ref(&e)))
```
TimeEngineer, line 61: Why don't you call `.as_ref()`, simply ? `TypeError` needs `&str` so you are sure that `as_ref` must be from `AsRef<str>`

FirelightFlagboy: done

oxidation/bindings/electron/src/meths.rs
FirelightFlagboy: done

oxidation/bindings/generator/templates/binding_electron_meths.rs.j2
```diff
@@ -101,15 +101,18 @@ JsFunction
...
-match custom_from_rs_string({{ js_val }}.value({{ mut_cx_ref }}))
-{%- else -%}
-{match {{ js_val }}.value({{ mut_cx_ref }}).parse()
-{%- endif %} {
-    Ok(val) => val,
-    Err(err) => return cx.throw_type_error(err),
-}}
+{
+    {% if type.custom_from_rs_string -%}
+    let custom_from_rs_string = {{ type.custom_from_rs_string }};
```
touilleMan, line 106: see https://github.com/Scille/parsec-cloud/pull/4245#discussion_r1141785242

FirelightFlagboy: The comment in question:

> We could however add typing info in the `custom_from/to_rs_string`:
> 
> ```rust
>     let custom_from_rs_string: Fn(String) -> Result<{{ type.rs_type }}, AsRef<str>> = {{ type.custom_from_rs_string }};
>     let value: Result<{{ type.rs_type }}, _> = custom_from_rs_string({{ js_val }}.value({{ mut_cx_ref }}));
> ```


1. For `Fn(String) -> Result<{{ type.rs_type }}, AsRef<str>>`: Would not work => `AsRef<str>` is a trait not a concrete type (spoiler using `impl` won't work)
2. Could do

FirelightFlagboy: We already type the return value on the `custom_{from,to}_rs_string`.
Is this really useful ?

FirelightFlagboy: This PR is blocking the progress of other, so I merge it.
For you recommendation I've opened the issue #4251

touilleMan: > We already type the return value on the custom_{from,to}_rs_string.

The idea is to help when we write the string containing the custom type, it makes it obvious when the signature of the function should be (instead of relying on copy/paste from other strings)

This is just a nice-to-have, as we don't add custom strings that often ;-)

PR #4253, opened 2023-03-20
Improve logging for test `platform_local_db::fail_reopen_database`

FirelightFlagboy:
Add logging when:

- Starting the background executor
- The `full_vacuum` failed
- Change the log level to `debug` for when the executor is exiting
- Add logs to the various steps in the tests


touilleMan (changes requested):
Wait for #4258

PR #4254, opened 2023-03-20
Fix deadlock in BackgroundSqliteExecutor 

touilleMan:
# What has changed ?


touilleMan:
> Can you do a benchmark ?

I don't think this have any impact on performances: congestion is small, and the task has no choice but to wait for the result anyway.

Do you have something in mind ?

FirelightFlagboy:
The classic `Trust me, I known what I'm doing` ? :thinking:

touilleMan:
> The classic Trust me, I known what I'm doing ? 🤔

can you do the bench then ? I'm busy with #4258 right now ;-)

PR #4260, opened 2023-03-21
Run `pre-commit` on all files on master & merge queue

FirelightFlagboy:
This will save sometime on the CI by only running `pre-commit` and the modified files in a PR

From `~5m` to `~1m`

### DIFF
--- a/.github/actions/use-pre-commit/action.yml
+++ b/.github/actions/use-pre-commit/action.yml
@@ -5,7 +5,11 @@ inputs:
   extra-args:
     description: options to pass to pre-commit run-precommit
     required: false
-    default: --all-files
+    default: ''
+  changed-files:
+    description: List of changed files to run `pre-commit` on, if empty will run on all files
+    required: false
+    default: ''
   config-file:
     description: where is located the pre-commit config file
     required: false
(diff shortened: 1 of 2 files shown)


.github/actions/use-pre-commit/action.yml
```diff
@@ -45,6 +49,37 @@ runs:
...
+        echo "args=$ARGS" >> $GITHUB_OUTPUT
+      env:
+        ALL_FILES: ${{ steps.changes.outputs.all_files }}
+        ARGS: "${{ inputs.extra-args }}"
+      shell: bash
+
     - name: Run pre-commit
-      run: python ~/.cache/pre-commit/pre-commit.pyz run --show-diff-on-failure --color=always ${{ inputs.extra-args }}
+      run: >-
+        python ~/.cache/pre-commit/pre-commit.pyz
```
touilleMan, line 81: `>-` is hard to read (it seems like `run --show-diff-on-failure` is a separate command)
you should add indent to improve this:
```
python ~/.cache/pre-commit/pre-commit.pyz
    run --show-diff-on-failure
    --color=always
```

but even like that I'm not a great fan, I'd rather use a `|` with backslashes to stay within bash expectations:
```
run: |
  python ~/.cache/pre-commit/pre-commit.pyz \
    run --show-diff-on-failure \
    --color=always
```

touilleMan (approved):

PR #4273, opened 2023-03-23
Use 'opt-level = 1' for all non libparsec_* dependencies in dev

touilleMan:
# What has changed ?

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -37,11 +37,15 @@ tokio = { version = "1.25", features = ["rt", "rt-multi-thread"] }
 lazy_static = "1.4.0"
 futures = "0.3.21"
 
-# Rust unoptimized code is so slow that we customize dev profile to enable some
-# optimizations, otherwise Python tests take forever.
+# Rust unoptimized code is really slow with crypto algorithm and regex compilation,
+# hence we optimize our 3rd party dependencies.
+[profile.dev.package."*"]
+opt-level = 1
+
 [profile.dev-python]
 inherits = "dev"
-opt-level = 1
+# We don't optimize our own crates here given it comes with a high impact on
+# when doing single change recompilation.
 
 # Custom profiles for the CI. The idea here is to try to save compilation time
 # and artifacts size (to keep cache efficient).


Cargo.toml
```diff
@@ -37,11 +37,13 @@ tokio = { version = "1.25", features = ["rt", "rt-multi-thread"] }
...
-# Rust unoptimized code is so slow that we customize dev profile to enable some
-# optimizations, otherwise Python tests take forever.
+# Rust unoptimized code is really slow with crypto algorithm and regex compilation,
+# hence we optimize our 3rd party dependencies.
+[profile.dev.package."*"]
+opt-level = 1
+
 [profile.dev-python]
 inherits = "dev"
-opt-level = 1
```
touilleMan, line 44: not sure if we should remove this: the difference is without this the libparsec_* crates won't be optimized
@FirelightFlagboy or @TimeEngineer can you do a small bench to see the difference in compilation time & running the python tests

Cargo.toml
```diff
@@ -37,11 +37,13 @@ tokio = { version = "1.25", features = ["rt", "rt-multi-thread"] }
 lazy_static = "1.4.0"
 futures = "0.3.21"
 
-# Rust unoptimized code is so slow that we customize dev profile to enable some
-# optimizations, otherwise Python tests take forever.
+# Rust unoptimized code is really slow with crypto algorithm and regex compilation,
+# hence we optimize our 3rd party dependencies.
+[profile.dev.package."*"]
+opt-level = 1
```
TimeEngineer, line 43: why `opt-1`, not `opt-2` ?

FirelightFlagboy: The difference between `opt-1` and `opt-2` was 20s of increase compilation time for `opt-2` for a gain of `200ms` for executing the test (on @touilleMan computer)

Cargo.toml
FirelightFlagboy: ## Compilation time

### With your current change

```shell
$ cargo clean
$ python make.py r
[...]
    Finished dev-python [unoptimized + debuginfo] target(s) in 1m 14s
```

### With `profile.dev-python.opt-level = 1`

```shell
$ sed '/profile.dev-python/a opt-level = 1' Cargo.toml
$ cargo clean
$ python make.py r
[...]
    Finished dev-python [optimized + debuginfo] target(s) in 2m 01s
```

## Test time

### With your current change

```shell
$ pytest -n 4 tests
[...]
1208 passed, 386 skipped, 1 xfailed in 14.23s
```

### With `profile.dev-python.opt-level = 1`

```shell
$ pytest -n 4
[...]
1208 passed, 386 skipped, 1 xfailed in 11.95s
```

FirelightFlagboy (approved):

PR #4307, opened 2023-03-27
Fix CI saving `libparsec-python` during the merge queue

### LINKED ISSUE #4304: [CI] Ci is saving the artifact `libparsec-python` on merge queue branch
The CI is saving the artifact `libparsec-python on the merge queue branches.
This step is unneeded because the merge queue branch are a one time uses.

FirelightFlagboy:
During the merge queue, the python jobs where saving the `libparsec-python` artifact but this is a waste of resources given the merge queue branch isn't reused.

The problem was `contains` directly return a boolean and not a boolean string (`'true'` or `'false'`) The previous expression was checking `true != 'true'` which result in `true`

The fix was to remove the `string` comparaison and directly inverting (using the not `!` operator) the result of `contains(..)`

Closes #4304

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -356,7 +356,7 @@ jobs:
         if: >-
           steps.should-run-python-jobs.outputs.run == 'true'
           && steps.cache-libparsec.outputs.cache-hit != 'true'
-          && contains(github.ref, 'gh-readonly-queue') != 'true'
+          && !contains(github.ref, 'gh-readonly-queue')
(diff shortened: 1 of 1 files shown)


mmmarcos (approved):
LGTM. Maybe we can add a note/comment above : `NOTE: the "contains" below returns a boolean (and not a boolean string like 'true')`

mmmarcos:
> The other usage of `contains` don't have a comment for that. I prefer to limit the amount of comment that refers to something from the documentation that could change in the future.

Good point. Also I missed the fact that you had already explained it in the commit message :+1:

PR #4346, opened 2023-03-30
RFC invisible minimal sync

touilleMan:
# What has changed ?

### DIFF
--- a/docs/rfcs/xxxx-invisible-minimal-sync.md
+++ b/docs/rfcs/xxxx-invisible-minimal-sync.md
@@ -1,0 +1,91 @@
+# Invisible minimal sync
+
+## 0 - Basic considerations
+
+Considering the following :
+
+1. Alice creates a file in location `/foo/big.txt`
+2. Alice write a big amount of data into `/foo/big.txt`
+3. Alice create another file in location `/foo/small.txt`
+4. Bob wants to access to `/foo/small.txt` asap
+
+Here the simplest way to synchronize the data is:
+
+1. Sync `/foo/small.txt`'s file manifest
+2. Sync `/foo/big.txt`'s file manifest
+3. Sync `/foo`'s folder manifest
+
+However this is bad because `/foo`'s changes cannot be accessed by Bob as long as
+`/foo/big.txt` is not done with synchronization.
+
+For that reason Parsec currently has slightly more complexe way of doing synchronization:
+
+1. Sync `/foo/small.txt`'s file manifest
+2. Sync a minimal version `/foo/big.txt` (i.e. file manifest corresponding to an empty file)
+3. Sync `/foo`'s folder manifest
+4. Sync of `/foo/big.txt` actual data as version 2 of it file manifest
+
+## 1 - The issues with minimal sync
+
+The minimal sync approach is great to reduce synchronization congestion, however it creates
+an empty files manifest that cannot be differentiate from an actual legitimate empty file
+(e.g. by doing a `touch foo.txt`).
+
+### 1.1 - Issue in prod
+
+This empty file is especially bad for non-text documents because the software that should
+open them is not expecting the file to be empty, hence bad error message that make Parsec look broken :'(
+
+### 1.2 - Issue with tests
+
+[See this issue](https://github.com/Scille/parsec-cloud/issues/4312)
+
+Condiering the following test code:
+
+```python
+    # Create a shared workspace
+    wid = await create_shared_workspace(EntryName("w"), bob_user_fs, alice_core)
+    alice_workspace = alice_core.user_fs.get_workspace(wid)
+
+    # Alice creates a non-empty files and sync it
+    await alice_workspace.write_bytes("/test.txt", b"hello world")
+    await alice_core.wait_idle_monitors()
+
+    info = await alice_workspace.path_info("/test.txt")
+    assert info["base_version"] == 2  # or is it 3 ????
+```
+
+The issue here is we have two manifests to synchronize: `/` workspace manifest and
+`/test.txt` file manifest.
+So the sync monitor may choose to start synchronization either by the file manifest or
+the workspace manifest. If it starts by the workspace manifest, a minimal sync must be
+done first on the file manifest !
+
+So in the end, once fully synchronized the file manifest may be at version 2 or 3...
+
+## Not a solution: ignore version 1 of the manifest
+
+Given minimal sync may or may not occur, we cannot just ignore the version 1 of the manifest.
+On top of that an empty file can be legitime (e.g. `touch foo.txt`) so we also cannot filter
+on "version 1 and empty".
+
+## Solution 1: `minimal` tag in file/folder manifest
+
+TODO: description
+TODO: Backward compatibility impact
+
+## Solution 2: allow unknown id in file manifest entry
+
+Folder manifest can have be crafted by a malicious user, hence the "entry ID in the folder
+manifest's children should always reference a valid vlob containing a file/folder manifest"
+cannot be enforced.
+
+So no matter what we have to deal with the fact an entry ID is invalid:
+
+- no vlob have this entry ID
+- a vlob exists but it contains the wrong type of manifest (e.g. workspace manifest)
+- a vlob exists but it contains garbage data
+- a vlob exists and used to contains valid data, but not it latest version
+
+TODO: Current handling of invalid entry ID
+TODO: Backward compatibility impact


docs/rfcs/xxxx-invisible-minimal-sync.md
FirelightFlagboy, line 1: Would be nice if you add a small introduction before the steps `0 - Basic considation`

docs/rfcs/xxxx-invisible-minimal-sync.md
```diff
@@ -0,0 +1,91 @@
+# Invisible minimal sync
+
+## 0 - Basic considerations
+
+Considering the following :
```
mmmarcos, line 5:
```suggestion
Considering the following situation:
```

docs/rfcs/xxxx-invisible-minimal-sync.md
```diff
@@ -0,0 +1,91 @@
...
+
+1. Alice creates a file in location `/foo/big.txt`
+2. Alice write a big amount of data into `/foo/big.txt`
+3. Alice create another file in location `/foo/small.txt`
```
mmmarcos, line 9:
```suggestion
2. Alice writes a big amount of data into `/foo/big.txt`
3. Alice creates another file in location `/foo/small.txt`
```

docs/rfcs/xxxx-invisible-minimal-sync.md
```diff
@@ -0,0 +1,91 @@
...
+4. Bob wants to access to `/foo/small.txt` asap
+
+Here the simplest way to synchronize the data is:
+
+1. Sync `/foo/small.txt`'s file manifest
+2. Sync `/foo/big.txt`'s file manifest
+3. Sync `/foo`'s folder manifest
+
+However this is bad because `/foo`'s changes cannot be accessed by Bob as long as
+`/foo/big.txt` is not done with synchronization.
```
mmmarcos, line 19:
```suggestion
However this is not ideal because `/foo`'s changes (included in `/foo`'s folder manifest) cannot be accessed by Bob as long as `/foo/big.txt` is not done with synchronization.
```

docs/rfcs/xxxx-invisible-minimal-sync.md
```diff
@@ -0,0 +1,91 @@
...
+
+For that reason Parsec currently has slightly more complexe way of doing synchronization:
```
mmmarcos, line 21:
```suggestion
For this reason Parsec currently has a slightly more complex way of doing the synchronization:
```

docs/rfcs/xxxx-invisible-minimal-sync.md
```diff
@@ -0,0 +1,91 @@
...
+
+1. Sync `/foo/small.txt`'s file manifest
+2. Sync a minimal version `/foo/big.txt` (i.e. file manifest corresponding to an empty file)
+3. Sync `/foo`'s folder manifest
+4. Sync of `/foo/big.txt` actual data as version 2 of it file manifest
```
mmmarcos, line 26:
```suggestion
2. Sync a minimal version of `/foo/big.txt` (i.e. a file manifest corresponding to an empty file) _as version 1_
3. Sync `/foo`'s folder manifest
4. Sync `/foo/big.txt` actual data as _version 2_ of it file manifest
```

docs/rfcs/xxxx-invisible-minimal-sync.md
```diff
@@ -0,0 +1,91 @@
...
+
+## 1 - The issues with minimal sync
+
+The minimal sync approach is great to reduce synchronization congestion, however it creates
+an empty files manifest that cannot be differentiate from an actual legitimate empty file
+(e.g. by doing a `touch foo.txt`).
```
mmmarcos, line 32:
```suggestion
## 1 - Issues with minimal sync

The minimal sync approach is a great way to reduce synchronization congestion, however it creates
an empty files manifest which cannot be differentiated from an actual, legitimate, empty file
(e.g. created by `touch foo.txt`).
```

docs/rfcs/xxxx-invisible-minimal-sync.md
```diff
@@ -0,0 +1,91 @@
...
+
+### 1.1 - Issue in prod
+
+This empty file is especially bad for non-text documents because the software that should
+open them is not expecting the file to be empty, hence bad error message that make Parsec look broken :'(
```
mmmarcos, line 37:
```suggestion
This empty file is especially inconvenient for non-text documents because the software that
open them may not be expecting the file to be empty, resulting in an error message that makes Parsec look broken :'(
```

docs/rfcs/xxxx-invisible-minimal-sync.md
```diff
@@ -0,0 +1,91 @@
...
+
+### 1.2 - Issue with tests
+
+[See this issue](https://github.com/Scille/parsec-cloud/issues/4312)
+
+Condiering the following test code:
```
mmmarcos, line 43:
```suggestion
Considering the following test code:
```

docs/rfcs/xxxx-invisible-minimal-sync.md
```diff
@@ -0,0 +1,91 @@
...
+```
+
+The issue here is we have two manifests to synchronize: `/` workspace manifest and
+`/test.txt` file manifest.
+So the sync monitor may choose to start synchronization either by the file manifest or
+the workspace manifest. If it starts by the workspace manifest, a minimal sync must be
+done first on the file manifest !
+
+So in the end, once fully synchronized the file manifest may be at version 2 or 3...
+
```
mmmarcos, line 65:
```suggestion
There are two manifests to synchronize, the `/` workspace manifest and 
the `/test.txt` file manifest, so the sync monitor may choose to start 
synchronization either by the file manifest or the workspace manifest.
The issue here is that if it starts by the workspace manifest, then a minimal sync must be
performed first on the file manifest! 

So in the end, once fully synchronized, the file manifest may be at version 2 or 3...

```

docs/rfcs/xxxx-invisible-minimal-sync.md
```diff
@@ -0,0 +1,91 @@
...
+## Not a solution: ignore version 1 of the manifest
+
+Given minimal sync may or may not occur, we cannot just ignore the version 1 of the manifest.
+On top of that an empty file can be legitime (e.g. `touch foo.txt`) so we also cannot filter
```
mmmarcos, line 69:
```suggestion
Besides, an empty file can be legitime (e.g. `touch foo.txt`) so we also cannot simply filter
```

docs/rfcs/xxxx-invisible-minimal-sync.md
```diff
@@ -0,0 +1,91 @@
...
+## Solution 1: `minimal` tag in file/folder manifest
+
+TODO: description
+TODO: Backward compatibility impact
+
+## Solution 2: allow unknown id in file manifest entry
+
+Folder manifest can have be crafted by a malicious user, hence the "entry ID in the folder
+manifest's children should always reference a valid vlob containing a file/folder manifest"
+cannot be enforced.
```
mmmarcos, line 81:
```suggestion
Folder manifest may have been crafted by a malicious user, hence the "entry ID in the folder
manifest's children should always reference a valid vlob containing a file/folder manifest"
cannot be enforced.
```

docs/rfcs/xxxx-invisible-minimal-sync.md
```diff
@@ -0,0 +1,91 @@
...
+
+So no matter what we have to deal with the fact an entry ID is invalid:
```
mmmarcos, line 83:
```suggestion
So no matter what, we have to deal with the fact an entry ID is invalid:
```

mmmarcos (changes requested):
Some rephrasing and typo fixing

mmmarcos:
The document describes the problem related to file _creation_, but what about file _update_? Similar problems should occur when trying to update (most of) the contents of a single large file. If other files are added/removed at the same time, changes in the directory may not be visible to Bob until synchronization is done on the large file. Am I missing something?

This leads me to ask: the `version` field in the file manifest is used only for the minimal sync mechanism? or it also serves other purposes? 

There is already a `need_sync` field, maybe it can be generalized to enable more complex workflows: 

```markdown
## Solution 3: use a `sync_status` tag in file/folder manifest

The `need_sync` boolean tag may be generalized into a `sync_status`. 

By choosing the right values, the `sync_status` may enable more complex workflows:
- `need_sync`
- `sync_minimal_ongoing`
- `sync_minimal_done`
- `sync_ongoing`
- `sync_done`
- ...

TODO: Backward compatibility impact
```

touilleMan:
> If other files are added/removed at the same time, changes in the directory may not be visible to Bob until synchronization is done on the large file. Am I missing something?

The path of the file and it content are stored in separate entities:
- the path is split accross the file's parents folder manifests up to the workspace storage
- the content is only present in the file manifest

hence:
- moving the file ==> doing a change in a folder manifest
- modifying the file ==> doing a change in it file manifest

So we can both move and modify a file and each operation can be synced independently. However there is one exception: given we want to guarantee that an entry ID in a folder manifest point to a valid manifest, we must ensure the file manifest has been synchronized before being able to synchronize the folder manifest. Hence the "minimal sync" operation that sync an empty file manifest to meet folder manifest sync requirements as fast as possible.

> This leads me to ask: the version field in the file manifest is used only for the minimal sync mechanism? or it also serves other purposes?

The version field is very important given we want to make sure the client won't overwrite a modification made by someone else (on top of that the version is signed, which prevent the server from trying to re-order the versions of a manifest to trick the client)

> There is already a need_sync field, maybe it can be generalized to enable more complex workflows:

`need_sync` flag is only for local manifest (i.e. manifest present in the client that get modified when the user interact with the workspace), the local manifest gets turned into a proper remote manifest (i.e. what we call a file/folder/workspace manifest) which doesn't contain the `need_sync` flag (because a remote manifest has been by definition stored into the server !)

So I don't think adding more step to `need_sync` brings a benefit here (in local we already know the file manifest has not been synchronized yet)

mmmarcos:
This seem outdated. We'll reopen if needed.

PR #4370, opened 2023-04-03
Oxidation merge types

TimeEngineer:
# What has changed ?

This PR was discussed with @touilleMan, the idea is to simplify `libparsec_types` and `libparsec_client_types`, because most of time we use both crates together (excepted in `libparsec_protocol` and `libparsec_client_connection`)


oxidation/libparsec/crates/types/tests/mod.rs
```diff
@@ -1,5 +1,8 @@
 // Parsec Cloud (https://parsec.cloud) Copyright (c) BUSL-1.1 (eventually AGPL-3.0) 2016-present Scille SAS
 
+// Functions using rstest parametrize ignores `#[warn(clippy::too_many_arguments)]`
+// decorator, so we must do global ignore instead :(
+#![allow(clippy::too_many_arguments)]
```
FirelightFlagboy, line 5: Maybe it's the indication that we need to define/use a struct :shrug:

TimeEngineer: Yes but out of scope

FirelightFlagboy: Open an issue

TimeEngineer: https://github.com/Scille/parsec-cloud/issues/4371

PR #4372, opened 2023-04-03
Increase Retrieve Rust cache timeout (MacOS timeout sometimes)

TimeEngineer:
# What has changed ?

Why 5 minutes ? Because it's 5 minutes in Rust tests


TimeEngineer:
Why 5 ? Because it's 5 minutes in Rust tests

PR #4376, opened 2023-04-04
Split CI workflow into sub-workflows

FirelightFlagboy:
I have splitted `ci.yml` into sub-workflows (`ci-{web,rust,python}.yml`) to reduce it's size.

This change don't change how the CI workflows run, since it will call each sub-workflows when triggered. But now a specific workflow can be manually called (We can now run the `rust` workflow without running the `python` & `web` workflows).

On top of that, I've updated the filters in a way that each part (`web`, `python` & `rust`) watch also for their corresponding workflows (and not the global workflow => `ci.yml`). Thus improving the CI load given if we modify the `python` workflow the `rust` & `web` workflows wont run (They will skip each steps)


.github/filters/ci.yml
```diff
@@ -73,7 +71,7 @@ rust-changes: &rust-changes
 # > Note: We don't use `rust-changes` because it include all the change made to the rust code
 # > including change made to the ionic part.
 python-jobs:
-  - *ci-workflow
+  - .github/workflows/ci-python.yml
```
touilleMan, line 74: What about the `ci-workflow.yml`'s own dependencies ?
For instance if something in `.github/actions/` changes the build must be triggered

touilleMan: given there is not much things in `.github/actions` and it doesn't changes often, better play it safe and add the whole directory instead of just what we need ;-)

.github/workflows/ci.yml
FirelightFlagboy, line 1: Remove unused env variable

FirelightFlagboy: All the list env variable is used by the job `Q&A` ;)

.github/filters/ci.yml
FirelightFlagboy: Should be good now

PR #4378, opened 2023-04-04
Test fuzz core setup hardening

FirelightFlagboy:
This PR add typing to `test_fuzz_core` and also harden the setup of the test

- Comment the `trio.sleep`
- Wait for monitors to idle
- Use `fail_after` for `workspace_create`

This PR is related to #4237

### DIFF
--- a/tests/fuzzing/test_fuzz_core.py
+++ b/tests/fuzzing/test_fuzz_core.py
@@ -12,23 +12,24 @@ import trio
 
 from parsec.api.data import EntryName
 from parsec.api.protocol import UserID
-from parsec.core.fs import FSError
+from parsec.core.fs import FSError, WorkspaceFS
+from parsec.core.logged_core import LoggedCore
 from parsec.core.types import WorkspaceRole
 
 FUZZ_PARALLELISM = 10
 FUZZ_TIME = 10.0
 
 
-def generate_name():
+def generate_name() -> str:
     return "".join([choice(ascii_lowercase) for x in range(4)])
 
 
 @attr.s
 class FSState:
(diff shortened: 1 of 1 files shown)


tests/fuzzing/test_fuzz_core.py
```diff
@@ -233,9 +235,14 @@ async def _fuzzer_cmd(id, core, workspace, fs_state):
...
-async def test_fuzz_core(request, running_backend, alice_core):
-    await trio.sleep(0.1)  # Somehow fixes the test
-    wid = await alice_core.user_fs.workspace_create(EntryName("w"))
+async def test_fuzz_core(request, running_backend, alice_core: LoggedCore):
+    # FIXME: Does this "power nap" is really required ?
+    # await trio.sleep(0.1)  # Somehow fixes the test
+    with trio.fail_after(1):
+        await alice_core.wait_idle_monitors()
+    with trio.fail_after(1):
+        wid = await alice_core.user_fs.workspace_create(EntryName("w"))
```
touilleMan, line 244: This line will trigger the sync monitor, so shouldn't we wait for the idle monitors after that?

FirelightFlagboy: Maybe, but I find it strange that we directly use `user_fs` instead of `LoggedCore` to create the workspace.

I could add a `wait_idle_monitors` after the workspace create.

FirelightFlagboy: I've tried adding a `wait_idle_monitors` after the workspace creation.

It result in `TooSlowError`.

FirelightFlagboy:
This PR is deprecated, the fuzzing test was removed

PR #4380, opened 2023-04-04
Support `--version` flag almost everywhere

### LINKED ISSUE #4363: Support `parsec --version`
I would like to be able to use the flag `--version` to display the version of parsec.

## What I got

Currently it result in the following message:

```
Usage: python -m parsec.cli core gui [OPTIONS] [URL]
Try 'python -m parsec.cli core gui --help' for help.

Error: No such option: --version
```

## What I expect

```
$ parsec --version
2.15.0
```

FirelightFlagboy:
Currently, the `--version` is only handled on the root CLI

```
python -m parsec.cli --version
```

But the installed `parsec` app is aliased like that

```
python -m parsec.cli core gui
```

So previously when adding the flag `--version` result in the error described in #4363.

To fix that, I've added `click.version` in the function `debug_config_options` So everywhere this function is used to add the `--debug` flag, we also get the `--version` flag.

I've added more case for the test `test_version` to show it's working with sub-command

Closes #4363

### DIFF
--- a/newsfragments/4363.feature.rst
+++ b/newsfragments/4363.feature.rst
@@ -1,0 +1,2 @@
+Handle the CLI flag ``--version`` on parsec sub-command.
+You can now type ``parsec --version`` to get the current version of parsec in a terminal.
--- a/parsec/backend/cli/__init__.py
+++ b/parsec/backend/cli/__init__.py
@@ -15,11 +15,13 @@ from parsec.backend.cli.sequester import (
     list_services,
     update_service,
 )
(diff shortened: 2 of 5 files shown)


steevenakintilo:
The code is working but can we get this option on the command:

parsec backend --version ?

FirelightFlagboy:
Yes we could add `--version` to `parsec core` & `parsec backend`

Max-7 (approved):

Ironicbay (approved):

PR #4387, opened 2023-04-06
Log support in python-rust code

FirelightFlagboy:
Add support to be able to log message in rust using the python logger


FirelightFlagboy:
Out of date, need to start from the begining

PR #4401, opened 2023-04-11
Break cyclic dependencies in `libparsec_protocol`

FirelightFlagboy:
The crate `libparsec_protocol` depends on `libparsec_serialization_format` which it self depends on `libparsec_protocol::Request` (in dev-dependencies for testing purpose).

To resolve that, I've create an intermediate crate `libparsec_miniprotocol` that will be used by the crate `protocol` and `serialization_format`.

The Gain
--------

When compiling `libparsec_protocol[test]` the gain is not obvious:

```
master: 11.1s
this branch: 10.4s
```

But when compiling `libparsec_serialization_format[test]` the gain is clear:

```
master: 12.5s
this branch: 2.7s
```

This gain is caused because `libparsec_protocol` take a lot of time in `code-gen` (=> generating code) and `serialization_format` only required `protocol` for `Request`

> Those result is obtain with
>
> ```bash
> cargo clean -p <crate>
> cargo build -timings -p <crate> --tests
> ```

Others Changes
--------------

- Updated tests to use `libparsec_miniprotocol` when needed.
- Rework how we expose `libparsec::protocol`.
- Update `serialization_format` to use `libparsec_miniprotocol` when generating code.

### DIFF
--- a/oxidation/libparsec/crates/serialization_format/src/utils.rs
+++ b/oxidation/libparsec/crates/serialization_format/src/utils.rs
@@ -116,7 +116,7 @@ pub(crate) fn _inspect_type(ty: &Type, types: &HashMap<String, String>) -> Strin
                 "PkiEnrollmentSubmitPayload" => "PkiEnrollmentSubmitPayload",
                 "X509Certificate" => "libparsec_types::X509Certificate",
                 // Used only in protocol
-                "IntegerBetween1And100" => "crate::IntegerBetween1And100",
+                "IntegerBetween1And100" => "libparsec_miniprotocol::IntegerBetween1And100",
                 ident if types.get(ident).is_some() => {
                     types.get(ident).unwrap_or_else(|| unreachable!())
                 }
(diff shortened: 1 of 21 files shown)


oxidation/libparsec/crates/serialization_format/src/utils.rs
```diff
@@ -116,7 +116,7 @@ pub(crate) fn _inspect_type(ty: &Type, types: &HashMap<String, String>) -> Strin
                 "PkiEnrollmentSubmitPayload" => "PkiEnrollmentSubmitPayload",
                 "X509Certificate" => "libparsec_types::X509Certificate",
                 // Used only in protocol
-                "IntegerBetween1And100" => "crate::IntegerBetween1And100",
+                "IntegerBetween1And100" => "libparsec_miniprotocol::IntegerBetween1And100",
```
TimeEngineer, line 119: According to the comment, the type is used only in `protocol` is it really needed to move ? Or the comment is outdated ?

TimeEngineer (approved):
I'm not very fan that we have tests in `proc_macro` crate because of cyclic dependencies.
So if we want to keep tests in `serialization_format` I guess we need this sub crate, so LGTM.

PR #4403, opened 2023-04-12
Provide light install for the testbed requirement

FirelightFlagboy:
I've added a new install type `testbed-server` (alias `ts`) that install the python dependencies required for the testbed server
and compile the rust lib with the appropriate flags enabled.

I've use the term `light install` because I disable the install of the `pyqt` dependencies in the `build-system` (and in other places).

> I don't find this solution optimal because it modify the `pyproject.toml` in place.
> I would like to be able to specify a specific `pyproject.toml` to use instead or even better
> Being able to specify which dependencies is required for the build-system depending on:
>
> - The enabled `extra` (`PyQt5; extra=='core'`)
> - Or env variable (`PyQt5; POETRY_PYQT_BUILD_STRATEGY!='no_build'`)

Other changes
-------------

I've changed of `Cmd` work when provided with a `list[str]` previously it consider that list being a list of commands to execute,
now It's the list of argument of the given Command.

> This was required for the use of the `sed` cmd with the space in a argument of the command.

### DIFF
--- a/make.py
+++ b/make.py
@@ -14,6 +14,7 @@ from typing import Optional, Union
 PYTHON_RELEASE_CARGO_FLAGS = "--profile=release --features libparsec/use-sodiumoxide"
 PYTHON_DEV_CARGO_FLAGS = "--profile=dev-python --features test-utils"
 PYTHON_CI_CARGO_FLAGS = "--profile=ci-python --features test-utils"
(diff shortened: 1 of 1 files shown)


FirelightFlagboy:
Now we don't have to install pyqt5, so this Pr is deprecated

PR #4409, opened 2023-04-13
Update Python to `3.11`

steevenakintilo:
The version of python on parsec went from 3.9.6 to 3.11.0

The goal of that is to make the code easier to use for everybody


FirelightFlagboy:
Hi, you have included a commit from the main branch, you need to remove it.

Use `git rebase -i` for that

pyproject.toml
```diff
@@ -48,8 +48,8 @@ msgpack = "^1.0"
 wsproto = "^1.2"
 h11 = ">=0.13,<0.15"
 # Can use marshmallow or the toasted flavour as you like ;-)
-# marshmallow = "2.14.0"
-toastedmarshmallow = "0.2.6" # updating causes consistency problem in test
+marshmallow = "2.16.2"
+# toastedmarshmallow = "0.2.6" # updating causes consistency problem in test
```
FirelightFlagboy, line 52: Wasn't because `toastedmarshmallow` don't support python-3.11 because it import

```python
from collections import Mapping
```

instead of 

```python
from collections.abc import Mapping
```

?

steevenakintilo:
> Hi, you have included a commit from the main branch, you need to remove it.
> 
> Use `git rebase -i` for that

thanks , I've done it

FirelightFlagboy (changes requested):
You also need to change the python version used in `codeql` workflow

pyproject.toml
steevenakintilo: Based on what I've found on internet many people fixed their problem by doing
from collections.abc import Mapping
So I guess that could be the Issue there but I'm not 100% sure about it

FirelightFlagboy: The stacktrace you've provided seems to indicate that

parsec/core/logged_core.py
```diff
@@ -89,7 +89,9 @@ def get_prevent_sync_pattern(prevent_sync_pattern_path: Path | None = None) -> R
         pattern = _get_prevent_sync_pattern(prevent_sync_pattern_path)
     # Default to the pattern from the ignore file in the core resources
     if pattern is None:
-        with importlib.resources.files(core_resources).joinpath("default_pattern.ignore") as path:  # type: ignore[attr-defined]
+        path = Path(importlib.resources.files(core_resources)).joinpath("default_pattern.ignore")
+        # Do something with the file
+        with open(path) as file:
```
vxgmichel, line 94: Why is this change needed?

parsec/utils.py
```diff
@@ -225,13 +227,12 @@ def collapse_exception_group(exception_group: BaseExceptionGroup[E]) -> Exceptio
         # Craft the specific class, inheriting from BaseExceptionGroup
         cls: Any = type(name, (type(pick), BaseExceptionGroup), attrs)
         # Instantiate the instance
-        result = cls()
+        result = cls(exception_group.message, exception_group.exceptions)
```
vxgmichel, line 230: This is probably more robust:
```suggestion
        result = BaseExceptionGroup.__new__(cls, exception_group.message, exception_group.exceptions)
```

tests/test_nursery.py
```diff
@@ -22,6 +21,10 @@ async def test_open_service_nursery_exception_group_collapse(caplog):
     async def _raise(exc):
         raise exc
 
+    assert issubclass(ExceptionGroup, Exception)
+    assert issubclass(BaseExceptionGroup, BaseException)
+    assert issubclass(ExceptionGroup, BaseExceptionGroup)
+
```
vxgmichel, line 27: I don't think it makes sense to test things that are outside our code base

tests/test_nursery.py
```diff
@@ -39,8 +42,10 @@ async def _raise(exc):
 
     caplog.assert_occurred_once("[warning  ] A BaseExceptionGroup has been detected [parsec.utils]")
 
+    assert ctx.errisinstance(ZeroDivisionError)  # cspell: disable-line
     exception = ctx.value
-    assert isinstance(exception, ZeroDivisionError)
```
vxgmichel, line 43: Why this change? This assert still works right?

vxgmichel (approved):
LGTM :+1:

vxgmichel:
BTW the CI fails because we're not providing winfspy wheels for python 3.11, I'll fix that

tests/test_nursery.py
vxgmichel: sure, but why?

mmmarcos:
Soon to be replaced by #5691 (See https://github.com/Scille/parsec-cloud/pull/5691/files#diff-cbb8e7dd5c29affc67d7d19ffb2f7bf0ff7316718cf3034537f5c86b3e62280f)

PR #4439, opened 2023-04-20
Correct CODEOWNERS assigned folder

FirelightFlagboy:
With the previous definition the main problem was with the `src` folder. The system will look for every `src` folder and will assign the related team.

The is problematic when we modify a file in `oxidation/client/src`, this will assign the python team to it.

Other Changes
-------------

- Correct that some folder is shared between multiple team (e.g. `/src` folder).
- Correct @touilleMan not correctly assign to `.github/CODEOWNERS` (the cause was with the `.github/` match of the ops team).
- Re-order the file this way:
  - JS
  - JS + Rust
  - Rust
  - Rust + Python
  - Python
  - OPS
  - CODEOWNERS

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -2,22 +2,8 @@
 # These parts are assigned to a specific team whose members are proficient in them.
 # If you think you belongs to one of these teams because you're proficient in one of those parts, ask @touilleMan.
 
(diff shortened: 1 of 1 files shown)


touilleMan (approved):
:+1:

PR #4441, opened 2023-04-20
Don't require to install `poetry` & `python` in Q&A.

FirelightFlagboy:
Installing `poetry` & `python` in `Q&A` was only required when we change python code.

It's more efficient to do that in `ci-python` since this workflow should be run when actual python code is changed (note, it is also run when we modify `rust` code).

Other Changes
-------------

- Action `use-pre-commit`:
  - Return the install path of `pre-commit` in the outputs variable `install-path`.
  - Skip `Run pre-commit` step when `install-only != true`.

### DIFF
--- a/.github/actions/use-pre-commit/action.yml
+++ b/.github/actions/use-pre-commit/action.yml
@@ -18,11 +18,22 @@ inputs:
     description: pre-commit version to use
     required: false
     default: 2.21.0
+  install-only:
+    description: Only install pre-commit without running it.
+    required: false
+    default: 'false'
+  install-dir:
+    description: Path where to install pre-commit.
+    required: false
+    default: ~/.cache/pre-commit
 
 outputs:
   cache-hit:
     description: A boolean value to indicate an exact match was found for the key.
     value: ${{ steps.cache-pre-commit.outputs.cache-hit }}
+  install-path:
+    description: A path where pre-commit.pyz is installed
+    value: ${{ inputs.install-dir }}/pre-commit.pyz
 
 runs:
   using: composite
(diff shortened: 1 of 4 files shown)


Ironicbay (approved):

.github/actions/use-pre-commit/action.yml
```diff
@@ -18,11 +18,22 @@ inputs:
     description: pre-commit version to use
     required: false
     default: 2.21.0
+  install-only:
+    description: Only install pre-commit without running it.
+    required: false
+    default: "false"
```
mmmarcos, line 24: Use single quotes to be consistent with other strings?
```suggestion
    default: 'false'
```

mmmarcos (approved):
Just a minor suggestion, otherwise LGTM :+1:

.github/actions/use-pre-commit/action.yml
FirelightFlagboy: Maybe we should enforce that in `yamllint` configuration ?

mmmarcos: Why not... will it impact a lot of yaml files? 

Btw, someone seem to prefer double quotes :smile: 
https://github.com/Scille/parsec-cloud/blob/c009a160434cb1edae8806caa8388389d6f08147/.yamllint.yml#L77-L78

FirelightFlagboy: Yes that's me :sweat_smile: 
Yes it will impact some yaml file but that will be in a specific PR to simplify that

PR #4450, opened 2023-04-24
Remove `env` variables in `ci.yml`

FirelightFlagboy:
Remove the variables defined in `env` since they're now unused.


touilleMan (approved):
:+1:

PR #4451, opened 2023-04-25
Change `yamllint` configuration to only allow for double-quote style in yaml files

FirelightFlagboy:
This will ensure a consistent style across our yaml files. 
I choose double-quote since it's the same style enforced in the python files (but remain open to discussion).


Ironicbay (approved):

mmmarcos (approved):
:+1:

PR #4444, opened 2023-04-21
Re-enable web unit test

FirelightFlagboy:
Remove `wasm-pack` from `devDependencies` since it don't work.
Instead install it from the wasm-pack repo.

Other Changes
-------------

- Correct unused warning when compiling `platform_http_proxy` on `wasm32`.
- In `bindings/web/scripts/build.js`:
  - Provide `process.env` to sub command.
  - Now the script will directly invoque `wasm-pack`.
  - Ensure rust target `wasm32-unknown-unknown` is installed.
- Remove `ci` build type for rust web since `wasm-pack` don't support custom profile.
- Cache rust artifact when build the web bindings.

Closes #4447

PR #4453, opened 2023-04-25
Add updated workflow to build webapp 

### LINKED ISSUE #4419: Update package-ci to work with the new build env for the web application

FirelightFlagboy:
Other Changes
-------------

- Correct `run: command not found`

  I discover that the script `web:release` was executing `run lint` after `vite build`.
  I suppose the command here was meant to be `npm run lint` but running the lint at that moment seems odd.

Depends on #4444
Closes #4419

### DIFF
--- a/oxidation/client/package.json
+++ b/oxidation/client/package.json
@@ -11,7 +11,7 @@
         "test:unit": "cypress run --component",
         "test:e2e": "cypress run --e2e",
         "web:open": "vite --open",
-        "web:release": "vite build && run lint",
+        "web:release": "npm run lint && vite build",
         "native:build": "node ./scripts/vite_build_for_native.cjs",
         "android:copy": "npm run native:build && npx cap copy android",
         "android:copy:release": "npm run native:build && npx cap copy android",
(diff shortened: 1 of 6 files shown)


Max-7 (approved):

oxidation/client/package.json
```diff
@@ -11,7 +11,7 @@
         "test:unit": "cypress run --component",
         "test:e2e": "cypress run --e2e",
         "web:open": "vite --open",
-        "web:release": "vite build && run lint",
+        "web:release": "vite build",
```
touilleMan, line 14: lint should stay (damn json format that prevented us from adding a comment !!!! ><'')
The issue is `vite build` doesn't do all the check on the typescript code (it's not called "vite" for nothing), so we also have to run the lint after to ensure the code is correct (and hence it compilation has produced a valid output)

touilleMan (changes requested):

oxidation/client/package.json
touilleMan: see https://vitejs.dev/guide/features.html#typescript

> Transpile Only
> Note that Vite only performs transpilation on .ts files and does NOT perform type checking. It assumes type checking is taken care of by your IDE and build process.

FirelightFlagboy: done

PR #4463, opened 2023-04-26
Config `pre-commit` to validate snapcraft file

FirelightFlagboy:


.github/workflows/package-ci.yml
```diff
@@ -157,7 +157,7 @@ jobs:
 
       - name: Set snap build version
         run: |
-          sed -i.back 's/__PARSEC_VERSION__/"${{ env.WHEEL_VERSION }}"/' ${{ runner.temp }}/snap/snapcraft.yaml
+          sed -i.back 's/^version: [a-zA-Z0-9]\([a-zA-Z0-9:.+~-]*[a-zA-Z0-9+~]\)\?/version: ${{ env.WHEEL_VERSION }}/' ${{ runner.temp }}/snap/snapcraft.yaml
```
touilleMan, line 160: is there a way to add a `set -eux` to this part and have sed return a non zero result if it replaced nothing ?
(my fear is that the more complex regex may silently not match the content at one point (typically because we added a strange final character in the version name)

touilleMan: the `$` is missing in the regex, hence `version: 0.0.0-rc1#dummy` leaves the `#dummy` part untouched !

touilleMan: what about changing the regex to `^version: .*$` ? we don't really care what the version field use to contain (and it's not this script's job to do sanity check here), and this has the benefit of being very easy to read and ensure the match it done no matter what

touilleMan: and finally, is there a way to tell sed to only replace the first occurrence of this ? this is not strictly needed, but I worry if there is another `version: ..` field later on in the configuration (to specify a module version for instance). In theory the match shouldn't work given the indentation would be different, but it's safer not to play with that

touilleMan (changes requested):

.github/workflows/package-ci.yml
FirelightFlagboy: done

touilleMan: seems good to me, but can you add a small comment `# Sed with a range specifier to only match from line 2 until we find the pattern` and `# Use grep as sanity check to ensure sed worked as expected`

touilleMan (approved):

PR #4466, opened 2023-04-27
Fix shamir RFC

TimeEngineer:
# What has changed ?

`self` is a rust keyword which isn't allowed as raw identifier, see: https://github.com/rust-lang/rust/blob/1.31.1/src/libsyntax_pos/symbol.rs#L455-L463

### DIFF
--- a/docs/rfcs/1000-shamir-based-recovery.md
+++ b/docs/rfcs/1000-shamir-based-recovery.md
@@ -107,20 +107,20 @@ Authenticated API:
         },
         "reps": [
             {
-                "status": "ok",
+                "status": "ok"
             },
             {
                 // Certificate is not signed by the authenticated user, or timestamp is invalid
-                "status": "invalid_certification",
+                "status": "invalid_certification"
             },
             {
                 // Cannot deserialize data into the expected certificate, or inconsistency
                 // between  certificates and/or threshold
-                "status": "invalid_data",
+                "status": "invalid_data"
             },
             {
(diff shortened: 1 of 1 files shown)


docs/rfcs/1000-shamir-based-recovery.md
```diff
@@ -267,21 +267,21 @@ Authenticated API:
...
         "reps": [
             {
                 "status": "ok",
-                "other_fields": [
+                "fields": [
                     {
                         // `None` means no configuration
                         // Otherwise, contains a `ShamirRecoveryBriefCertificate`
-                        "name": "self",
+                        "name": "my",
```
FirelightFlagboy, line 279: `info` or `my_info` ?

TimeEngineer: I can also set to `self_info`

docs/rfcs/1000-shamir-based-recovery.md
```diff
@@ -430,7 +430,7 @@ Invited API, we reuse the `invite_info` command:
                             },
                             {
                                 "name": "recipients",
-                                "type": "Vec<ShamirRecoveryRecipient>",
+                                "type": "Vec<ShamirRecoveryRecipient>"
```
touilleMan, line 433: Not a `List<...>` here ?

touilleMan (approved):

docs/rfcs/1000-shamir-based-recovery.md
TimeEngineer: Sure

PR #4467, opened 2023-04-28
Update snapcraft to use any python version

### LINKED ISSUE #4452: Configure snapcraft to allow to use for any python version
Currently with `snapcraft` we are forced to use the version provided by `snapcraft` it would be more simpler if we do not have such constraint

I've found this discussion that propose some steps to bundle our own python version:

<https://forum.snapcraft.io/t/build-a-snap-with-any-version-of-python-i-want/10420>

We may need to build python from source:

<https://www.python.org/ftp/python/3.11.3>

FirelightFlagboy:
Closes #4452


FirelightFlagboy:
At the commit f862e36, I've tested the generated snap against the following isos:

- Ubuntu-18
- Ubuntu-20
- Ubuntu-22
- Rocky-9

The tests where at follow:

- Able to show parsec version.
- Able to open parsec
- Able to be invited in an organization
- Able to list the files inside a workspace
- The mountpoint `Parsec/<workspace>` is present
- Able to read from the mountpoint
- Able to write from the mountpoint
- Able to rename a file from the GUI

packaging/snap/bin/parsec-setup.sh
```diff
@@ -7,14 +7,26 @@
...
+export PYTHON=${SNAP}/bin/python3.9
 export PYTHONPATH=${SNAP}/site-packages
 
 # We can't rely on `ctypes.util.find_library` in classic snap environment,
 # so we use the `*_LIBRARY_PATH` environment variables.
 export FUSE_LIBRARY_PATH=${SNAP}/lib/x86_64-linux-gnu/libfuse.so.2
-export SSL_LIBRARY_PATH=${SNAP}/usr/lib/x86_64-linux-gnu/libssl.so.1.1
-export CRYPTO_LIBRARY_PATH=${SNAP}/usr/lib/x86_64-linux-gnu/libcrypto.so.1.1
+# We directly use the provide ssl lib by `core20`.
+export SSL_LIBRARY_PATH=/snap/core20/current/lib/x86_64-linux-gnu/libssl.so.1.1
```
touilleMan, line 17: Why not the $SNAP env var ? Doesn't it bring issues if the snap folder is not at the default location?

FirelightFlagboy: The `SNAP` var only point to our snap env, so maybe using relative path could work :thinking: 

Since now we check if the lib file exist that would prevent running parsec with an invalid lib path

FirelightFlagboy:
I've made another test using `opensuse-tumbleweed`

touilleMan (approved):

packaging/snap/snap/snapcraft.yaml
```diff
@@ -70,166 +126,178 @@ parts:
...
+    stage:
+      - $certs
+      - $site-packages
+      - $stage-lib
+    prime:
+      - $certs
+      - $site-packages
+      - $stage-lib
+    # Why so much packages ?
+    # Well because we use a python that come wit the bare minimum to save space.
```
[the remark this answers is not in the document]
FirelightFlagboy: done

PR #4507, opened 2023-05-05
Fix Cypress e2e tests

### LINKED ISSUE #4393: Fix Cypress E2E tests with Vite

Max-7:
# What has changed ?

Not sure at all about the CI stuff, trying to bluff my way through it.

Closes #4393


.github/workflows/ci-web.yml
```diff
@@ -147,10 +147,9 @@ jobs:
...
         # Currently disable because of #4393
-        if: false && steps.should-run-web-jobs.outputs.run == 'true'
-        # Without `VUE_CLI_TEST` set, the command hangs forever because
-        # it fails to stop the server :'(
-        # (see https://github.com/vuejs/vue-cli/issues/3401)
-        run: VUE_CLI_TEST=1 npm run test:e2e:headless
+        if: steps.should-run-web-jobs.outputs.run == 'true'
+        run: npm run test:e2e:headless
+        env:
+          TESTBED_SERVER_URL: parsec://127.0.0.1:6770?no_ssl=true
```
FirelightFlagboy, line 153: I don't see when the testbed server is started

FirelightFlagboy (approved):

oxidation/client/package.json
```diff
@@ -5,11 +5,14 @@
     "type": "module",
     "scripts": {
         "postinstall": "cd electron && npm install",
-        "dev": "vite",
+        "dev": "vite --port 8080",
```
FirelightFlagboy, line 8: The vite server port is also configured in `vite.config.ts`

Max-7: Yep but it is...weird. If it set the option `--port 8080`, the server starts on port 8080 and everything seems to work. If I set the port in vite.config, it starts on port 8081 because for some reason, it detects the port 8080 as being already in use. I can force `strictPort: true` in vite.config, but then the server doesn't start because it is not able to bind on port 8080.

NicoTuxx (approved):

Issue #4527, opened 2023-05-11
Create custom dropdown


fabienSvtr:
Link of the component on Figma :
[Dropdown component](https://www.figma.com/file/XW3SNWHV8BhPsPJpsOCV2h/Application?type=design&node-id=22336-44899&t=dWpryLUZoFvzv1Fy-4)

NicoTuxx:
linked to #4603

PR #4528, opened 2023-05-11
Settings modal homepage

fabienSvtr:
# What has changed ?

**New settings modal**
This modal contains General and Advanced settings tabs.
It is only accessible when the user is on the home page (logged out).

> Visual issue remaining:
> - All select input must be redesign (#4527)

closes#3964


oxidation/client/src/components/SettingsModal.vue
```diff
@@ -0,0 +1,344 @@
...
+            v-if="showTOS != 'advanced'"
+            class="settings-general"
+          >
+            <ion-list class="settings-list">
+              <!-- synchro wifi -->
+              <settings-option
+                :title="$t('SettingsPage.synchroWifiOnly')"
+                :description="$t('SettingsPage.synchroWifiOnlyDescription')"
+                v-model="config.synchroWifiOnly"
+              />
```
FirelightFlagboy, line 80: Does this setting is only relevant on `smartphone` ?

Max-7: The name is probably a bit misleading, I should have more to do with metered connection rather than wifi.

oxidation/client/src/components/SettingsModal.vue
```diff
@@ -0,0 +1,344 @@
...
+import { onMounted, onUnmounted } from '@vue/runtime-core';
+import { toggleDarkMode } from '@/states/darkMode';
+import { Config, StorageManager } from '@/services/storageManager';
+import { storageManagerKey } from '@/main';
+import SettingsOption from './SettingsOption.vue';
+
+const { locale } = useI18n();
+const storageManager = inject(storageManagerKey)!;
+const config = ref<Config>(structuredClone(StorageManager.DEFAULT_CONFIG));
+const showTOS = ref('general');
```
Max-7, line 188: What does `TOS` stand for? In my head it's `Terms of Service`, which is a bit confusing in this case ^^

Max-7, line 188: Instead of using hard coded string, we could use an enum here. Something like

```typescript
enum SettingsPage {
  General = 'General',
  Advanced = 'Avanced'
}

const showTOS = ref(SettingsPage.General);

// Check what page we're on
if (showTOS == SettingsPage.General) {
  ...
}
```

oxidation/client/src/components/SettingsModal.vue
```diff
@@ -0,0 +1,344 @@
...
+
+function closeModal(): Promise<boolean> {
+  return modalController.dismiss(null, 'cancel');
+}
+
+const configUnwatch = watch(config, async (_, oldConfig) => {
+  if (JSON.stringify(toRaw(oldConfig)) !== JSON.stringify(StorageManager.DEFAULT_CONFIG)) {
```
Max-7, line 195: Add a comment to explain that the first time this function is called is when the config is retrieved from the local storage and there is no point in storing what we just retrieved, so we add this `if` to skip this case.

oxidation/client/src/components/SettingsModal.vue
```diff
@@ -0,0 +1,344 @@
...
+async function changeLang(selectedLang: string): Promise<void> {
+  config.value.locale = selectedLang;
+  locale.value = selectedLang;
+  await storageManager.storeConfig(toRaw(config.value));
+}
+
+async function changeTheme(selectedTheme: string): Promise<void> {
+  config.value.theme = selectedTheme;
+  toggleDarkMode(selectedTheme);
+  await storageManager.storeConfig(toRaw(config.value));
```
Max-7, line 210: Remove the call to storageManager here. The function `changeTheme` is only useful to force the changes of theme globally so that the whole application gets its new appearance. Since we also update `config.value.theme`, the watcher will be called and will do the `storeConfig` for us.

oxidation/client/src/components/SettingsModal.vue
```diff
@@ -0,0 +1,344 @@
...
+  if (JSON.stringify(toRaw(oldConfig)) !== JSON.stringify(StorageManager.DEFAULT_CONFIG)) {
+    await storageManager.storeConfig(toRaw(config.value));
+  }
+  console.log(config.value);
+}, { deep: true });
+
+async function changeLang(selectedLang: string): Promise<void> {
+  config.value.locale = selectedLang;
+  locale.value = selectedLang;
+  await storageManager.storeConfig(toRaw(config.value));
```
Max-7, line 204: Remove the call to `storeConfig` here (same as `changeTheme`)

oxidation/client/src/components/SettingsModal.vue
```diff
@@ -0,0 +1,344 @@
...
+            <ion-list class="settings-list">
+              <!-- send error report -->
+              <settings-option
+                :title="$t('SettingsPage.enableTelemetry')"
+                :description="$t('SettingsPage.enableTelemetryDescription')"
+                v-model="config.enableTelemetry"
+              />
+              <!-- minimise in status bar -->
+              <settings-option
+                :title="$t('SettingsPage.minimizeToSystemTray')"
```
Max-7, line 134: This option is only relevant if we're running in Electron.

oxidation/client/src/components/SettingsModal.vue
```diff
@@ -0,0 +1,344 @@
...
+                :description="$t('SettingsPage.minimizeToSystemTrayDescription')"
+                v-model="config.minimizeToTray"
+              />
+              <!-- display unsync files -->
+              <settings-option
+                :title="$t('SettingsPage.unsyncFiles')"
```
Max-7, line 140: `SettingsPage.unsyncFiles` and `SettingsPage.unsyncFilesDescription` seem to be missing from both translation files.

Max-7 (changes requested):

oxidation/client/src/views/SettingsPage.vue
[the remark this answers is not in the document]
fabienSvtr: The settings page we will be used it when the user is logged in. It will be an issue to do.

oxidation/client/src/components/SettingsModal.vue
```diff
@@ -0,0 +1,344 @@
...
+                  <ion-select-option value="system">
+                    {{ $t('SettingsPage.theme.system') }}
+                  </ion-select-option>
+                </ion-select>
+              </ion-item>
+            </ion-list>
+          </div>
+          <!-- advanced -->
+          <div
+            v-else
```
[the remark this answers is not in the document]
fabienSvtr: Sure!  This should be interesting if we need to add a settings tab in the future (e.g. notifications)

oxidation/client/src/components/SettingsModal.vue
fabienSvtr: As discussed with @Max-7, we can't handle this setting for now. So I think we should add the locales for better understanding and hide it for now.

oxidation/client/src/components/SettingsModal.vue
fabienSvtr: It was for `Tabs of Settings` but yes, it might be confusing. Replacing by `setting-tab` sound better?

oxidation/client/src/components/SettingsModal.vue
fabienSvtr: replaced by `meteredConnection` and the translation files are added

PR #4664, opened 2023-06-06
Update mypy to `1.3.0` to support `python-3.11`

FirelightFlagboy:
This required updating `trio_typing` and `qtrio` (because it require `trio_typing` internally).

### DIFF
--- a/parsec/core/gui/custom_dialogs.py
+++ b/parsec/core/gui/custom_dialogs.py
@@ -7,7 +7,7 @@ import sys
 import time
 from concurrent.futures import ThreadPoolExecutor
 from contextlib import contextmanager
-from typing import Callable, Collection, Generic, Iterator, TypeVar, cast
+from typing import Any, Callable, Collection, Generic, Iterator, Type, TypeVar, cast
 
 from PyQt5.QtCore import QEventLoop, Qt, pyqtSignal
 from PyQt5.QtGui import (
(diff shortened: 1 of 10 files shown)


parsec/core/gui/custom_dialogs.py
```diff
@@ -324,23 +335,67 @@ def _exec_method(
...
+    # getOpenFileName = classmethod(
+    #     generate_dialog_method(QFileDialog, QFileDialog.getOpenFileName, ("", ""))
+    # )
+    @classmethod
+    @copy_args_types(QFileDialog.getOpenFileName)
+    def getOpenFileName(cls, *args: Any, **kwargs: Any) -> tuple[str, str]:
+        try:
+            return cls._exec_method(QFileDialog, QFileDialog.getOpenFileName, *args, **kwargs)
+        except cls.DialogCancelled:
+            return ("", "")
```
FirelightFlagboy, line 347: I needed to change how those `classmethod` where generated because I was getting the error

```txt
Argument 1 to "__get__" of "classmethod" has incompatible type "None"
```
From `mypy`, when we where calling those methods like: 

https://github.com/Scille/parsec-cloud/blob/0d56def78ed028a0f44024454fc67adabf7420a8/parsec/core/gui/files_widget.py#L829-L831

I couldn't find why I was getting that error

Maybe @vxgmichel you have a idea on that ?

mmmarcos (approved):
LGTM

PR #4538, opened 2023-05-15
WIP: Add support for multi-workspace mount points on Windows

mmmarcos:
## Describe your changes

This PR adds support for multi-workspace mount points on Windows. 

More info: https://github.com/Scille/resana-secure/issues/137#issuecomment-1545467985

## Checklist

Before you submit this pull request, please make sure to:

- [ ] Keep changes in the pull request as small as possible
- [ ] Ensure the commit history is sanitized
- [x] Give a meaningful title to your PR
- [x] Describe your changes
- [x] Link any related issue in the description
- [x] Link any dependent pull request in the description

If this PR adds a new feature or fixes a bug:

- [ ] I have added or updated relevant tests

If this PR affects the user:

- [ ] I have created a news fragment
- [ ] I have updated the user documentation

PR #4665, opened 2023-06-06
Add single mountpoint mode for Windows

### LINKED ISSUE #4668: Add support for single mountpoint on Windows
In order to avoid running out of drive letters when mounting workspaces on Windows, it was proposed to have a single mountpoint containing all the workspaces inside. This could be done either with:

1. a single drive-based mountpoint (`P:\\`)
2. a single directory-based mountpoint (`C:\\Users\User\Parsec`)

All workspaces will be mounted inside this single mountpoint:
`P:\\<workspace1>`
`P:\\<workspace2>`
or
`C:\\Users\User\Parsec\<workspace1>`
`C:\\Users\User\Parsec\<workspace2>`

This was initially proposed for the Resana-Secure project: https://github.com/Scille/resana-secure/issues/137

mmmarcos:
## Describe your changes

This PR adds support for single mountpoint mode on Windows: when enabled via the `single_mountpoint_enabled` option, workspaces are mounted into the `base_mountpoint_path` directory instead of using drive letters. 

It is a simpler and safer alternative solution to https://github.com/Scille/parsec-cloud/pull/4538 for #4668.

Closes #4668

## Checklist

- [x] Describe your changes

### DIFF
--- a/parsec/core/config.py
+++ b/parsec/core/config.py
@@ -60,6 +60,7 @@ class CoreConfig:
     invitation_token_size: int = 8
 
     mountpoint_enabled: bool = False
+    mountpoint_in_directory: bool = False
     disabled_workspaces: FrozenSet[EntryID] = frozenset()
 
     sentry_dsn: str | None = None
@@ -99,6 +100,7 @@ def config_factory(
     mountpoint_base_dir: Path | None = None,
     prevent_sync_pattern_path: Path | None = None,
     mountpoint_enabled: bool = False,
+    mountpoint_in_directory: bool = False,
     disabled_workspaces: FrozenSet[EntryID] = frozenset(),
     backend_max_cooldown: int = 30,
     backend_connection_keepalive: int | None = 29,
@@ -145,6 +147,7 @@ def config_factory(
         mountpoint_base_dir=mountpoint_base_dir,
         prevent_sync_pattern_path=prevent_sync_pattern_path,
         mountpoint_enabled=mountpoint_enabled,
+        mountpoint_in_directory=mountpoint_in_directory,
         disabled_workspaces=disabled_workspaces,
         backend_max_cooldown=backend_max_cooldown,
         backend_connection_keepalive=backend_connection_keepalive,
@@ -175,7 +178,7 @@ def config_factory(
     core_config.data_base_dir.mkdir(mode=0o700, parents=True, exist_ok=True)
 
     # Mountpoint base directory is not used on windows
-    if sys.platform != "win32":
+    if sys.platform != "win32" or core_config.mountpoint_in_directory:
         core_config.mountpoint_base_dir.mkdir(mode=0o700, parents=True, exist_ok=True)
 
     return core_config
--- a/parsec/core/mountpoint/manager.py
+++ b/parsec/core/mountpoint/manager.py
@@ -382,6 +382,7 @@ async def mountpoint_manager_factory(
     mount_on_workspace_shared: bool = False,
     unmount_on_workspace_revoked: bool = False,
     exclude_from_mount_all: frozenset[EntryID] = frozenset(),
+    mountpoint_in_directory: bool = False,
 ) -> AsyncGenerator[MountpointManager, Any]:
     config = {"debug": debug}
 
@@ -390,6 +391,11 @@ async def mountpoint_manager_factory(
     # Now is a good time to perform some cleanup in the registry
     if sys.platform == "win32":
         cleanup_parsec_drive_icons()
+
+        # Enable mountpoint in directory mode only on Windows (also, note that
+        # the config dict is *unpacked* on Linux in fuse_mountpoint_runner)
+        config["mountpoint_in_directory"] = mountpoint_in_directory
+
     elif sys.platform == "darwin":
         await cleanup_macos_mountpoint_folder(base_mountpoint_path)
 
--- a/parsec/core/mountpoint/winfsp_runner.py
+++ b/parsec/core/mountpoint/winfsp_runner.py
@@ -5,6 +5,7 @@ import math
 import unicodedata
 from contextlib import asynccontextmanager
 from functools import partial
+from itertools import count
 from pathlib import PurePath
 from typing import Any, AsyncIterator
 from zlib import adler32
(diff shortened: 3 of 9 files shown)


mmmarcos:
**User Documentation**: this PR adds the option `single_mountpoint_enabled`. Are configuration options documented somewhere?

mmmarcos:
**Tests**: In principle, there are no new tests required. We should only enable `single_mountpoint_enabled` and run all existing mountpoint tests. However it seems that a memory FS is used for the tests, so I am wondering if changing the option will have any effect at all :thinking:

parsec/core/win_registry.py
```diff
@@ -187,7 +187,7 @@ def _get_drive_icon_path(device: LocalDevice) -> Traversable:
 @contextmanager
 def parsec_drive_icon_context(letter: str, device: LocalDevice) -> Iterator[None]:
     # Winreg is not available for some reasons
-    if sys.platform != "win32" or not try_winreg():
+    if sys.platform != "win32" or not try_winreg() or not letter:
```
mmmarcos, line 190: The drive icon context should not be use in single mountpoint mode (since no drive letter is used). The `parsec_drive_icon_context` is called with `letter=""`.

An alternative solution could be to use [contextlib.nullcontext](https://docs.python.org/3.7/library/contextlib.html#contextlib.nullcontext) in the caller: 
```python
if single_mountpoint_enabled:
    parsec_drive_icon_context = contextlib.nullcontext(drive_letter, device=user_fs.device)

with parsec_drive_icon_context(drive_letter, device=user_fs.device):
    ...
```

mmmarcos:
> **User Documentation**: this PR adds the option `single_mountpoint_enabled`. Are configuration options documented somewhere?

The answer is no. Plus, the added option is not configurable via the CLI so we do not need to change user documentation nor the news fragment.

FirelightFlagboy:
@mmmarcos you need to remove all the commit before `3c6eaf5e86b88802343ed5210be3d6ff3895aa44`

mmmarcos:
> @mmmarcos you need to remove all the commit before [3c6eaf5](https://github.com/Scille/parsec-cloud/commit/3c6eaf5e86b88802343ed5210be3d6ff3895aa44) (5 commit to remove so)

Done

parsec/core/mountpoint/manager.py
```diff
@@ -390,6 +391,11 @@ async def mountpoint_manager_factory(
     # Now is a good time to perform some cleanup in the registry
     if sys.platform == "win32":
         cleanup_parsec_drive_icons()
+
+        # Enable single mountpoint mode only on Windows (also, the config dict
+        # is unpacked for FUSE on Linux)
+        config["single_mountpoint_enabled"]: single_mountpoint_enabled
```
vxgmichel, line 397: Did you mean?
```suggestion
        config["single_mountpoint_enabled"] = single_mountpoint_enabled
```

parsec/core/config.py
```diff
@@ -60,6 +60,7 @@ class CoreConfig:
     invitation_token_size: int = 8
 
     mountpoint_enabled: bool = False
+    single_mountpoint_enabled: bool = False
```
vxgmichel, line 63: Maybe:
```suggestion
    mountpoint_in_directory: bool = False
```

vxgmichel (approved):
LGTM :+1: 

A couple of suggestions though

parsec/core/mountpoint/manager.py
mmmarcos: yep

parsec/core/config.py
mmmarcos: Agree. I'll rename it widely.

mmmarcos: Done

mmmarcos:
A potential issue highlighted by @vxgmichel is that with directory mounts, the directory may persist after a crash or forced shutdown. In this case it's a good idea to have a logic to detect and clean up these directories at mount time. Otherwise, a `Workspace X` directory artifact would cause the workspace to end up being mounted on `Workspace X (1)`, which is confusing for the user.

Basically, if the directory exists but its contents can't be accessed, then it's an artifact and should be deleted.

parsec/core/mountpoint/winfsp_runner.py
```diff
@@ -95,39 +95,66 @@ def _generate_volume_serial_number(device: LocalDevice, workspace_id: EntryID) -
     )
 
 
-# Mypy reports a missing return statement, however it's not possible to exit this
-# function without raising an exception or successfully return
-async def _bootstrap_mountpoint(base_mountpoint_path: PurePath, workspace_fs: WorkspaceFS) -> PurePath:  # type: ignore[return]
+async def cleanup_artefacts(base_mountpoint_path: trio.Path) -> None:
```
mmmarcos, line 98: TIL: "artefact" is the British English version of "artifact" :+1:

parsec/core/mountpoint/winfsp_runner.py
```diff
@@ -95,39 +95,66 @@ def _generate_volume_serial_number(device: LocalDevice, workspace_id: EntryID) -
...
+        # Artefacts from previous run can remain listed in the directory
+        # (even though `path.exists()`) returns `False`
+        # In this case, `.unlink()` still works and fixes the issue
+        try:
+            await path.unlink()
+        except OSError:
+            # No artefact was present
+            pass
+        else:
+            # An aretefact has been cleaned up
```
mmmarcos, line 116:
```suggestion
            # An artefact has been cleaned up
```

parsec/core/win_registry.py
```diff
@@ -211,3 +229,67 @@ def cleanup_parsec_drive_icons() -> None:
...
+        winreg_write_user_dword(rf"{key}\ShellFolder", "FolderValueFlags", 0x28)
+
+    key = rf"Software\Microsoft\Windows\CurrentVersion\Explorer\Desktop\NameSpace\{appguid}"
+    winreg.SetValue(hkcu, key, winreg.REG_SZ, appname)
+
+    key = rf"Software\Microsoft\Windows\CurrentVersion\Explorer\HideDesktopIcons\NewStartPanel"
+    winreg_write_user_dword(key, rf"{appguid}", 0x1)
+
+
+def remove_parsec_mountpoint_directory_to_quick_access(appguid: str | None = None) -> None:
```
Max-7, line 280: Should be `remove_parsec_mountpoint_directory_from_quick_access`

Max-7 (changes requested):

parsec/core/win_registry.py
mmmarcos: done

Max-7 (approved):

PR #4704, opened 2023-06-12
Specify shared dependencies in `workspace.dependencies`

FirelightFlagboy:
Move the multiple (more than 1) definition of a rust dependencies in the root `Cargo.toml`.

Using `workspace.dependencies` provide the following benefit:

- Prevent conflict when 2 actors update one of those dependencies in parallel.

  The potential conflict is simpler to resolve, only `/Cargo.toml` is modified.

- Ensure all sub-crates are using the same version.
- Reduce that amount of modified files by `dependabot`.

Here is the list of dependencies that I've moved to `workspace.dependencies`:

- `regex`
- `paste`
- `uuid`
- `tokio`
- `lazy_static`
- `futures`
- `rmp-serde`
- `serde`
- `log`
- `wasm-bindgen`
- `wasm-bindgen-futures`
- `js-sys`
- `console_log`
- `wasm-bindgen-test`
- `reqwest`
- `url`
- `chrono`
- `thiserror`
- `bytes`
- `once_cell`
- `anyhow`
- `pretty_assertions`
- `serde_test`
- `hex-literal`
- `rstest`
- `glob`
- `sha2`
- `web-sys`
- `instant`
- `serde_json`
- `itertools`
- `diesel`
- `libsqlite3-sys`
- `rand`
- `serde_with`
- `quote`
- `proc-macro2`
- `syn`
- `serde_bytes`

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -38,6 +38,119 @@ default-members = [
     "libparsec/crates/types",
 ]
 
+# All dependencies should be specified here instead of just in the crate(s) requiring them:
+# - It ensures each given dependency version is consistent accross all crates
+# - It simplifies updating dependency version (only a single place to update)
+#
+# Note the dependencies are marked `default-features = false`, this is because otherwise
+# the default features are always provided (see https://github.com/rust-lang/cargo/issues/8366).
+# This means the dependencies must always specify all features they use instead of relying
+# on default features... Let's say it's a way to reduce bloat by making the developer
+# check the default feature and making sure they are actually needed ;-)
+[workspace.dependencies]
+libparsec = { path = "libparsec", default-features = false }
+libparsec_client_high_level_api = { path = "libparsec/crates/client_high_level_api", default-features = false }
+libparsec_protocol = { path = "libparsec/crates/protocol", default-features = false }
+libparsec_crypto = { path = "libparsec/crates/crypto", default-features = false }
+libparsec_types = { path = "libparsec/crates/types", default-features = false }
+libparsec_platform_device_loader = { path = "libparsec/crates/platform_device_loader", default-features = false }
+libparsec_platform_storage = { path = "libparsec/crates/platform_storage", default-features = false }
+libparsec_serialization_format = { path = "libparsec/crates/serialization_format", default-features = false }
+libparsec_testbed = { path = "libparsec/crates/testbed", default-features = false }
+libparsec_tests_fixtures = { path = "libparsec/crates/tests_fixtures", default-features = false }
+libparsec_client_connection = { path = "libparsec/crates/client_connection", default-features = false }
+libparsec_platform_async = { path = "libparsec/crates/platform_async", default-features = false }
+libparsec_platform_http_proxy = { path = "libparsec/crates/platform_http_proxy", default-features = false }
+libparsec_client = { path = "libparsec/crates/client", default-features = false }
+libparsec_tests_macros = { path = "libparsec/crates/tests_macros", default-features = false }
+
+# Third parties
+android_logger = { version = "0.13", default-features = false }
+anyhow = { version = "1.0.71", default-features = false }
+argon2 = { version = "0.5.0", default-features = false }
+base32 = { version = "0.4.0", default-features = false }
+base64 = { version = "0.21.2", default-features = false }
+blake2 = { version = "0.10.6", default-features = false }
+bytes = { version = "1.4.0", default-features = false }
+chrono = { version = "0.4.26", default-features = false }
+console_error_panic_hook = { version = "0.1.6", default-features = false }
+console_log = { version = "1.0.0", default-features = false }
+crc32fast = { version = "1.3.2", default-features = false }
+crypto_box = { version = "0.8.2", default-features = false }
+data-encoding = { version = "2.4.0", default-features = false }
+diesel = { version = "2.0.3", default-features = false }
+digest = { version = "0.10.6", default-features = false }
+ed25519-dalek = { version = "1.0.1", default-features = false }
+email-address-parser = { version = "2.0.0", default-features = false }
+env_logger = { version = "0.10.0", default-features = false }
+flate2 = { version = "1.0.26", default-features = false }
+flume = { version = "0.10.14", default-features = false }
+fnmatch-regex = { version = "0.2.0", default-features = false }
+futures-lite = { version = "1.12.0", default-features = false }
+futures = { version = "0.3.28", default-features = false }
+generic-array = { version = "0.14.7", default-features = false }
+# Two versions of `getrandom` needed in `libparsec_crypto`
+getrandom_01 = { package = "getrandom", version = "0.1.16", default-features = false }
+getrandom_02 = { package = "getrandom", version = "0.2.3", default-features = false }
+glob = { version = "0.3.0", default-features = false }
+hex-literal = { version = "0.4.1", default-features = false }
+hex = { version = "0.4", default-features = false }
+http-body = { version = "0.4.5", default-features = false }
+hyper = { version = "0.14.26", default-features = false }
+instant = { version = "0.1.12", default-features = false }
+itertools = { version = "0.10.5", default-features = false }
+jni = { version = "0.21.1", default-features = false }
+js-sys = { version = "0.3.59", default-features = false }
+lazy_static = { version = "1.4.0", default-features = false }
+libsodium-sys = { version = "0.2.7", default-features = false }
+libsqlite3-sys = { version = "0.25.2", default-features = false }
+log = { version = "0.4.17", default-features = false }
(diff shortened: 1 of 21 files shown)


Cargo.toml
```diff
@@ -40,6 +40,47 @@ default-members = [
     "libparsec/crates/types",
 ]
 
+[workspace.dependencies]
```
touilleMan, line 43: [According to documentation](https://doc.rust-lang.org/cargo/reference/workspaces.html#the-dependencies-table), the dependencies in the workspace cannot be marked as optional.

Does this means the crates are downloaded/compiled no matter what ? Can this pose an issue in case of a crate that doesn't support a given platform ?

My guess is there is no issue there, but I'd rather rely on actual proof ^^
I theory web/android related crates can compile on non-web platform (otherwise rust-analyzer would not work with them), so this may not be an issue anyway...

Cargo.toml
```diff
@@ -40,6 +40,47 @@ default-members = [
...
+anyhow = { version = "1.0.71", default-features = false, features = ["std"] }
```
touilleMan, line 44: What is the rational for setting `default-feature` and `features` here or in the crate ?

(I would say we want to only list dependency version here, so every one of them should be in `default-feature = false`)

touilleMan (changes requested):

Cargo.toml
FirelightFlagboy: For me the rational would be:

- Always set `default-features` to `false` and list the `features` you actually rely on
  
  I haven't done it everywhere, because we surely discuss about that.

  The goal of `default-features` is to be explicit on the features we use / rely on.
  If a dependency add a feature in the future it won't be enabled by default.

- `features` in `workspace.dependencies` are for the `features` that are required across all the workspace members.

touilleMan: >  Always set default-features to false and list the features you actually rely on
> [...]
> If a dependency add a feature in the future it won't be enabled by default.

I think this is overkill given modifying default features [is considered a breaking change](https://doc.rust-lang.org/cargo/reference/semver.html#major-removing-a-feature-from-a-feature-list-if-that-changes-functionality-or-public-items)

> features in workspace.dependencies are for the features that are required across all the workspace members.

After grepping the code base for dependencies with features:
```
2x tokio = { version = "1.28", features = ["rt-multi-thread"] }
tokio = { version = "1.28.0", features = ["rt", "time", "sync", "macros"] }
tokio = { version = "1.28.1", features = ["rt", "time", "sync", "macros"] }
tokio = { version = "1.28.1", features = ["full"] }
tokio = { version = "1.28.1", features = ["fs", "sync"] }
tokio = { version = "1.28.0", features = ["fs", "sync"] }
tokio = { version = "1.28.0", features = ["fs", "sync"] }
tokio = { version = "1.28", features = ["rt", "rt-multi-thread"] }

4x serde = { version = "1.0.147", features = ["derive"] }

3x libsqlite3-sys = { version = "0.25.2", features = ["bundled"] }

3x bytes = { version = "1.4.0", features = ["serde"] }

uuid = { version = "1.2.2", features = ["v4", "fast-rng"] }
2x uuid = { version = "1.2.2", features = ["serde", "v4", "fast-rng"] }

2x pretty_assertions = { version = "1.3.0", features = ["unstable"] }

reqwest = { version = "0.11.17", features = ["native-tls-vendored"] }
reqwest = { version = "0.11.17", default-features = false }

neon = { version = "0.10", default-features = false, features = ["napi-6", "channel-api", "promise-api"] }
hyper = { version = "0.14.26", features = ["server"] }
openssl = { version = "0.10.48", optional = true, features = ["vendored"] }
zeroize = { version = "1.6.0", optional = true, features = ["alloc"]}
crypto_box = { version = "0.8.2", features = ["std"] }  # Optional ruscrypto dep
generic-array = { version = "0.14.7", features = ["serde"] }  # Optional ruscrypto dep
getrandom_01 = { package = "getrandom", version = "0.1.16", features = ["wasm-bindgen"] }  # Optional rustcrypto dep
getrandom_02 = { package = "getrandom", version = "0.2.3", features = ["js"] }  # Optional rustcrypto dep
web-sys = { version = "0.3.60", features = ["Window"] }
web-sys = { version = "0.3.60", features = ["Window", "Storage"] }
instant = { version = "0.1.12", features = ["wasm-bindgen"] }
diesel = { version = "2.0.3", features = ["sqlite", "returning_clauses_for_sqlite_3_35"], default-features = false }
anyhow = { version="1.0.71", features = ["backtrace"] }
pyo3 = { version = "0.17.3", features = ["multiple-pymethods", "extension-module"] }
```

I don't think there is a clear way to define feature required across the workspace.
Just like there is no dependency  needed across the workspace, but some crate need some dependency to achieve some goal (hence also needing some features is such dependency).

In the end I'd rather keep things simple so that adding a new dependency is obvious enough (==> workspace depencies specify version, features are specified on per-crate dependencies).

FirelightFlagboy: > workspace depencies specify version, features are specified on per-crate dependencies).

Some dependencies always use the same set of `features` (see `serde` with `derive`, `web-sys` with `Windows`)

touilleMan: > Some dependencies always use the same set of features (see serde with derive, web-sys with Windows)

Yes, but there is no hard reason for that, it just happen to be the case. So I don't see the gain in merging the features: in fact this would make harder to spot when a feature is no longer needed.

Cargo.toml
touilleMan: :+1:

Cargo.toml
```diff
@@ -38,6 +38,48 @@ default-members = [
     "libparsec/crates/types",
 ]
 
+[workspace.dependencies]
+anyhow = { version = "1.0.71" }
```
touilleMan, line 42: we need `default-feature = false` here, otherwise using `default-feature = false` in a sub `cargo.toml` would be silently ignored :/

Cargo.toml
FirelightFlagboy: done

bindings/android/libparsec/rust/Cargo.toml
```diff
@@ -8,8 +8,8 @@ crate-type = ["staticlib", "cdylib"]
 
 [dependencies]
 libparsec = { path = "../../../../libparsec" }
-lazy_static = "1.4.0"
+lazy_static = { workspace = true }
 jni = "0.21.1"
```
touilleMan, line 12: I think we should also move the version of the dependencies that are needed once in the workspace.

This is because it is likely with time we will endup with this dependency needed at multiple places (see for instance `base64` which seems really common and is only in `libparsec_client_connection` so far)

We want a no brainer when adding a new dependency, so:
- add the dependency in the crate it is needed and you're done (i.e. the current way to do)
- add the dependency in teh crate it is needed, and ensure it is listed in the workspace

However  the way to do with this PR is: add the depedency in the crate, then search in the whole project if the dependency is also used somewhere, if that's the case ensure it is listed in the workspace, else put the version in the crate. This is complex and we can be pretty sure the dev will go the easy way and just put the dependency with version in the crate (especially given some other dependency are declared this way)

So I'd say we want to have every version declared in the workspace

FirelightFlagboy: Done, all dependencies are now define in the root `Cargo.toml`, same goes for our own lib as I find it simpler to handle the path on the root `Cargo.toml`

bindings/electron/Cargo.toml
```diff
@@ -11,7 +11,7 @@ test-utils = ["libparsec/test-utils"]
...
 
 [dependencies]
-libparsec = { path = "../../libparsec" }
-lazy_static = "1.4.0"
-tokio = { version = "1.28", features = ["rt-multi-thread"] }
-neon = { version = "0.10", default-features = false, features = ["napi-6", "channel-api", "promise-api"] }
+libparsec = { workspace = true }
+lazy_static = { workspace = true }
+tokio = { workspace = true, features = ["rt-multi-thread"] }
+neon = { workspace = true, default-feature = false, features = ["napi-6", "channel-api", "promise-api"] }
```
FirelightFlagboy, line 17: @touilleMan `default-features` need an `s`

touilleMan: :+1:

libparsec/crates/platform_http_proxy/Cargo.toml
```diff
@@ -6,4 +6,4 @@ edition = "2021"
 # See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
 
 [dependencies]
-reqwest = { version = "0.11.17", default-features = false }
+reqwest = { workspace = true, features = ["default-tls"] }
```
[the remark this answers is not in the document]
FirelightFlagboy: https://github.com/Scille/parsec-cloud/issues/4846

PR #4726, opened 2023-06-15
[Dependabot] Change interval to weekly

### LINKED ISSUE #4721: Dependabot don't have enough time to update some dependencies
We take for example the PR #4717 that update `serde` from `1.0.147` to `1.0.164`.

The `1.0.147` was release the `2022-10-21` since then `dependabot` hasn't updated that dependencies until #4717.

My assumption is:

- We have configured `dependabot` with a limit of 5 open PR for each type of manager (`cargo`, `npm`, ...) and to run only at the start of the month. 
- Once the limit reach `dependabot` stop and wait for the already open PR to be merged before opening another.
- If we merge the next day (the 2nd of each month), `dependabot` wont provide the queued PR since the condition `at the start of the month` isn't meet.

Suggested corrections:

- Increase the PR limit ?
- Increase `dependabot` frequency ?
  > Having a increased frequence would make `dependabot` create more PR more frequently, but the delta to review is decreased.
- Use `renovatebot` to provide grouped PR ?
  > It's a tool like `dependabot` but with more feature, notably we can group update into the same PR, so we limit the CI exhaustion

FirelightFlagboy:
This will increase the update frequency, and reduce the update backlog. I also assign each ecosystem a different weekday to open update PRs.

Closes #4721

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -9,8 +9,9 @@ updates:
     pull-request-branch-name:
       separator: "-"
     schedule:
-      interval: monthly
-      time: "00:30"
+      interval: weekly
+      day: monday
+      time: "04:20"
       timezone: Europe/Paris
     open-pull-requests-limit: 5
     rebase-strategy: disabled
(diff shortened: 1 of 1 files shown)


mmmarcos (approved):
LGTM

Ironicbay (approved):

PR #4731, opened 2023-06-15
Update quickstart documentation

### LINKED ISSUE #4729: Update `Quickstart` documentation

FirelightFlagboy:
Closes #4729

### DIFF
--- a/docs/development/quickstart.md
+++ b/docs/development/quickstart.md
@@ -1,15 +1,13 @@
 # Hacking Quickstart
 
-- [Hacking Quickstart](#hacking-quickstart)
-  - [Get the sources](#get-the-sources)
-  - [Shared requirements](#shared-requirements)
-  - [Hacking the Python code (front + back)](#hacking-the-python-code-front--back)
-    - [Run the Python tests](#run-the-python-tests)
-    - [Run Parsec in local](#run-parsec-in-local)
-  - [Hacking the Web client (front)](#hacking-the-web-client-front)
-    - [Start working on the web client](#start-working-on-the-web-client)
-  - [Hacking the Rust code](#hacking-the-rust-code)
-  - [If you are working on Mac](#if-you-are-working-on-mac)
+- [Get the sources](#get-the-sources)
+- [Shared requirements](#shared-requirements)
+- [Hacking the Python server](#hacking-the-python-server)
+  - [Run the Python tests](#run-the-python-tests)
+- [Hacking the Web client (front)](#hacking-the-web-client-front)
+  - [Start working on the web client](#start-working-on-the-web-client)
+- [Hacking the Rust code](#hacking-the-rust-code)
+- [If you are working on Mac](#if-you-are-working-on-mac)
 
 ## Get the sources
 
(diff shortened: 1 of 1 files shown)


docs/development/quickstart.md
```diff
@@ -175,80 +148,66 @@ Happy Hacking 🐍
...
-You can use the ``run_testenv`` scripts to easily create a development environment:
+## Hacking the Web client (front)
 
-On Linux and MacOS:
+In addition to the [shared requirements](#shared-requirements), you need to install [`Node 18.12.0`](https://nodejs.org/en/download/releases).
 
-```shell
-. ./tests/scripts/run_testenv.sh
-```
+> You could use [`nvm`](https://github.com/nvm-sh/nvm) to manage multiple node version
```
mmmarcos, line 172:
```suggestion
> You can use [`nvm`](https://github.com/nvm-sh/nvm) to manage multiple node version
```

docs/development/quickstart.md
```diff
@@ -175,80 +148,66 @@ Happy Hacking 🐍
...
+> or [`nvm-windows`](https://github.com/coreybutler/nvm-windows) for `Windows`
 
-On Windows:
+<!-- TODO: Currently the web client via electron don't provide mountpoint so fuse isn't required
```
mmmarcos, line 175:
```suggestion
<!-- TODO: Currently the web client via electron doesn't provide mountpoint so fuse isn't required
```

mmmarcos (approved):
LGTM, thanks for these updated instructions!

docs/development/quickstart.md
```diff
@@ -302,9 +261,11 @@ that will provide some mocked data.
 1. Make sure you have the latest change for the python testbed server
 
    ```shell
-   POETRY_LIBPARSEC_BUILD_PROFILE=ci poetry install --extra 'backend'
+   POETRY_LIBPARSEC_BUILD_PROFILE=ci poetry install
```
touilleMan, line 264:
```suggestion
./make.py i
```

Not sure why the `ci` profile is specified instead of using the `dev` mode

FirelightFlagboy: The CI profile provide a slightly optimized lib

docs/development/quickstart.md
```diff
@@ -175,80 +148,66 @@ Happy Hacking 🐍
...
-   > You'll need to remove `PyQt5` from the build-system requirement in `pyproject.toml`
-   > see [PR-4403](https://github.com/Scille/parsec-cloud/pull/4403)
+     - On Linux: you need to have installed `libfuse2` -->
 
-2. Initialize a poetry env
+1. Initialize a poetry env
 
    ```shell
-   POETRY_LIBPARSEC_BUILD_PROFILE=ci poetry install --extra 'backend'
+   pushd server
```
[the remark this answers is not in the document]
FirelightFlagboy: I've simplified that step by linking it to the previous one

docs/development/quickstart.md
touilleMan: This is not worth the added complexity, having `./make.py i` as the single command to always use is very convenient for newcomers and people not versed into Rust

We should add complexity that only once we've reach a actual performance issue ^^

FirelightFlagboy: Note: this dicussion is an outdated change, did you review the latest change ?

touilleMan: image

I think I did, yes

FirelightFlagboy: So this is still a single command

touilleMan: - we don't need (until further proof is given) the additional perf
- we don't need the additional cognitive load of having to deal with `./make.py python-ci-install` vs `./make.py i`
- we don't need the additional compile time and disk usage when dev switches between `./make.py python-ci-install` vs `./make.py i` (because the dev is not really sure what is the difference, and might be given one or the other command to run depending on who is helping him)
- so just KISS please :pray: 

btw can you be more precise when replying (what have changed/what is the rational/what should we do, instead of short and broad reply that let the reader puzzled).
For instance
> Note: this dicussion is an outdated change, did you review the latest change ?

can be rephrased as 
> I've updated the code there, `./make.py python-ci-install` is still used but I've added a comment to explain the rational. I believe this is worth the added complexity over simply using `./make.py i` because [...]

FirelightFlagboy: I should have close that discussion, that way you would have reopen a new one and I would have understand the clear intent of the previous comment (<https://github.com/Scille/parsec-cloud/pull/4731#discussion_r1241940786>)

FirelightFlagboy: I'm now using `make.py i`

touilleMan (approved):

Issue #4762, opened 2023-06-26
Automatic user revoke


mmmarcos:
An entity deploying Parsec on-premise needs to be able to automatically revoke a user when it leaves. The idea is to avoid a manual task that may be forgotten by the administrator(s), specially in a context of a high turnover rate.

mmmarcos:
Not planned. It was replaced by the ability to (automatically) freeze users: #5811

PR #4763, opened 2023-06-26
Add RFC for automatic user revoke

mmmarcos:
This PR adds RFC for #4762

### DIFF
--- a/docs/rfcs/1002-automatic-user-revoke.md
+++ b/docs/rfcs/1002-automatic-user-revoke.md
@@ -1,0 +1,57 @@
+# Automatic user revoke
+
+## Abstract
+
+This document describes design and implementation considerations to enable
+automatic user revoke based on an external service.
+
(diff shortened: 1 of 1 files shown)


docs/rfcs/xxxx-automatic-user-revoke.md
FirelightFlagboy, line 1: The filename should be `1002-...`

FirelightFlagboy (commented):
The begin looks greats :+1: 

On top of that, I have that thing in mind:

> We could provide bots (i.e.: Automated user) to perform specific _users action_

That would solve the problem with revocation, but not re-encrypt

mmmarcos:
Closing this PR as not planned. 

The [RFC 1007](https://github.com/Scille/parsec-cloud/blob/master/docs/rfcs/1007-frozen-users.md) was proposed as an alternative solution and is implemented in https://github.com/Scille/parsec-cloud/pull/5842.

PR #4771, opened 2023-06-27
Generate SBOM for generated Package

### LINKED ISSUE #4770: Generate SBOM for Parsec
Generate a `SBOM`[^SBOM] for each generated artifact.

To generate the `SBOM`, I suggest the tool [`syft`](https://github.com/anchore/syft) because it's the simplest tool I've tested.

The SBOM generate should likely be done on the CI in the workflows `package-*.yml`.

Doing it for each package allow to take into account the transitive dependencies and aditional tools used by those (`syft` would have more entries if it index the `node_modules` folders, for example).

Each `SBOM` should have a unique and meaningfull name (e.g.: For the server, it could be `parsec-server.[...]`) and uploaded as an artifact alongside the generate package.
This allow the workflow `release.py` to include them in a release :smile:

**Why not using the SBOM provided by Github ?**

Currently the SBOM provided by Github only reflect the dependency graph for the main branch.
A release will have a different dependency graph, so this document isn't adapted for a release.

[^SBOM]: Software Bills Of Materials

FirelightFlagboy:
Closes #4770

### DIFF
--- a/.github/workflows/package-server.yml
+++ b/.github/workflows/package-server.yml
@@ -49,10 +49,13 @@ jobs:
       matrix:
         include:
           - name: 🐧 Linux
+            platform: linux
             os: ubuntu-22.04
           - name: 🍎 macOS
+            platform: macos
             os: macos-12
           - name: 🏁 Windows
+            platform: windows
             os: windows-2022
     name: "${{ matrix.name }}: 📦 Packaging (build Wheel)"
     runs-on: ${{ matrix.os }}
@@ -141,9 +144,18 @@ jobs:
       - name: Generate requirements & constraints infos
         run: python server/packaging/wheel/wheel_it.py ./server --output dist --skip-wheel
 
+      # Install syft
+      - uses: taiki-e/install-action@486baeb8af63bc3d9ec3ec66db5af6ba0ca78774 # pin v2.11.6
+        with:
+          tool: syft@0.84.0
+
+      - name: Generate SBOM
+        run: syft packages --config=.syft.yaml --output=spdx-json=dist/Parsec-Wheel-${{ matrix.platform }}.spdx.json .
+
       - uses: actions/upload-artifact@0b7f8abb1508181956e8e162db84b466c27e18ce # pin v3.1.2
         with:
           name: ${{ runner.os }}-${{ runner.arch }}-wheel
-          path: dist/
+          path: |
+            dist/
           if-no-files-found: error
         timeout-minutes: 5
--- a/.github/workflows/package-webapp.yml
+++ b/.github/workflows/package-webapp.yml
@@ -7,6 +7,13 @@ on:
   workflow_call:
   workflow_dispatch:
 
+# Set `concurrency` to prevent this workflow from being run on code that is not up-to-date on a PR (e.g. when making many push quickly on a PR).
+# This behavior is only intended for a PR and not for merge commits on the main branch. Having the workflow run on each merge commit can be useful to spot regressions missed by previous checks.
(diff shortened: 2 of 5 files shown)


.github/workflows/package-webapp.yml
```diff
@@ -7,6 +7,15 @@ on:
   workflow_call:
   workflow_dispatch:
 
+# We set `concurrency` to prevent having this workflow being run on code that is not up-to-date on a PR (a user make multiple push in a quick manner).
+# But on the main branch, we don't want that behavior.
+# Having the workflow run on each merge commit is something we would like, that could help us where a regression was made and missed by previous checks.
+#
+# For that we use `head_ref` that is only defined on `pull-request` and fallback to `run_id` (this is a counter, so it's value is unique between workflow call).
```
mmmarcos, line 14: Just a rewording suggestion
```suggestion
# Set `concurrency` to prevent this workflow from being run on code that is not up-to-date on a PR (e.g. when making many push quickly on a PR). 
# This behavior is only intended for a PR and not for merge commits on the main branch. Having the workflow run on each merge commit can be useful to spot regressions missed by previous checks.
# To distinguish between these cases, we use `head_ref` that is only defined on `pull-request` and fallback to `run_id` (this is a counter, so it's value is unique between workflow call).
```

.syft.yaml
```diff
@@ -0,0 +1,15 @@
+# Config for syft-0.84.0
+quiet: false
+
+check-for-app-update: false
+
+exclude:
+  - ./.git
+  # We don't ignore `target` & `node_modules` folders because they could containe additional dependencies not listed in the lock files.
```
mmmarcos, line 8:
```suggestion
  # We don't ignore `target` & `node_modules` directories because they could contain additional dependencies not listed in the lock files.
```

newsfragments/4770.doc.rst
```diff
@@ -0,0 +1,2 @@
+Add SBOM (Software Bills Of Materials) generation when packaging the softwares.
+This provide the list of dependencies used to build the software.
```
mmmarcos, line 2:
```suggestion
Add SBOM (Software Bills Of Materials) generation on software packaging.
This provides the list of dependencies used to build the software.
```

mmmarcos (approved):
LGTM. Just a couple of suggestions.

.github/workflows/package-server.yml
```diff
@@ -141,9 +144,18 @@ jobs:
       - name: Generate requirements & constraints infos
         run: python server/packaging/wheel/wheel_it.py ./server --output dist --skip-wheel
 
+      # Install syft
+      - uses: taiki-e/install-action@486baeb8af63bc3d9ec3ec66db5af6ba0ca78774 # pin v2.11.6
+        with:
+          tool: syft@0.84.0
+
+      - name: Generate SBOM
+        run: syft packages --config=.syft.yaml --output=spdx-json=dist/Parsec-Wheel-${{ matrix.platform }}.spdx.json .
```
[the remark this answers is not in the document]
FirelightFlagboy: It is uploaded, the generated SBOM is put in the `dist/` folder that is then uploaded as a whole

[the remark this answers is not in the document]
FirelightFlagboy: > why it is generated in dist/ instead of ${{ runner.temp }} like the other SBOM below?

I depend on how we build the artifact, some here build in the temp folder some other are not (like here). the goal is to execute `syft` where the generated artifacts lays

mmmarcos (approved):

PR #4789, opened 2023-07-03
[Dependabot] Configure dependencies group udpate

FirelightFlagboy:
Dependabot has a public beta for [Grouped version updates](https://github.blog/changelog/2023-06-30-grouped-version-updates-for-dependabot-public-beta/)

This allow dependabot to group in a single PR some dependencies will be updated.

The documentation about dependabot group can be found here: <https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file#groups>


.github/dependabot.yml
```diff
@@ -167,4 +203,8 @@ updates:
       time: "04:20"
       timezone: Europe/Paris
     open-pull-requests-limit: 5
+    groups:
```
Ironicbay, line 206: Might want to insert it before `open-pull-requests-limit` for consistency ^^

FirelightFlagboy: Indeed, I've just changed that

PR #4802, opened 2023-07-03
Replace unwrap_or_else(|_| unreachable!()) with expect(...)

### LINKED ISSUE #4542: [:boxing_glove: :crab: :1st_place_medal: Final oxidation] Remove `unwrap_or_else(|_| unreachable!())`
see e.g. https://github.com/Scille/parsec-cloud/blob/27918409d5f3643782084316e3f0990695f500cf/oxidation/libparsec/crates/types/src/invite.rs#L90

`libparsec_types` crate still uses `unwrap_or_else(|_| unreachable!())` instead of the `expect(...)` format, should be corrected.

mmmarcos:
## Describe your changes

This PR replaces all `unwrap_or_else(|_| unreachable!())` with `expect(...)` in the `types` crate.

Error messages use the "error as precondition" style as defined in https://doc.rust-lang.org/std/error/index.html#common-message-styles

Closes #4542 

## Checklist

- [x] Describe your changes
   
### DIFF
--- a/libparsec/crates/types/src/certif.rs
+++ b/libparsec/crates/types/src/certif.rs
@@ -48,10 +48,11 @@ fn dump<T>(obj: &T) -> Vec<u8>
 where
     T: Serialize,
 {
-    let serialized = ::rmp_serde::to_vec_named(obj).unwrap_or_else(|_| unreachable!());
+    let serialized = ::rmp_serde::to_vec_named(obj).expect("object should be serializable");
     let mut e = ZlibEncoder::new(Vec::new(), flate2::Compression::default());
-    e.write_all(&serialized).unwrap_or_else(|_| unreachable!());
-    e.finish().unwrap_or_else(|_| unreachable!())
+    e.write_all(&serialized)
+        .and_then(|_| e.finish())
+        .expect("in-memory buffer should not fail")
 }
 
 fn check_author_allow_root(
--- a/libparsec/crates/types/src/ext_types.rs
+++ b/libparsec/crates/types/src/ext_types.rs
@@ -85,7 +85,7 @@ impl<'de> serde::de::Visitor<'de> for DateTimeExtVisitor {
         let ts = f64::from_be_bytes(
             data[..F64_SIZE]
                 .try_into()
-                .unwrap_or_else(|_| unreachable!()),
+                .expect("data.len() should be equal to F64_SIZE"),
         );
 
         Ok(Self::Value::from_f64_with_us_precision(ts))
--- a/libparsec/crates/types/src/invite.rs
+++ b/libparsec/crates/types/src/invite.rs
@@ -86,8 +86,8 @@ impl FromStr for SASCode {
 
     fn from_str(s: &str) -> Result<Self, Self::Err> {
         lazy_static! {
-            static ref PATTERN: Regex =
-                Regex::new(SAS_CODE_PATTERN).unwrap_or_else(|_| unreachable!());
+            static ref PATTERN: Regex = Regex::new(SAS_CODE_PATTERN)
+                .expect("`SAS_CODE_PATTERN` should be a valid regular expression");
         }
         if PATTERN.is_match(s) {
             Ok(Self(s.to_string()))
@@ -145,7 +145,8 @@ impl SASCode {
         sas_codes.push(SASCode(self.to_string()));
         while sas_codes.len() < size {
             let num = rand::thread_rng().gen_range(0..=SAS_CODE_MASK);
-            let candidate = SASCode::try_from(num as u32).unwrap_or_else(|_| unreachable!());
+            let candidate = SASCode::try_from(num as u32)
+                .expect("SASCode number should not exceed `SAS_CODE_BITS` bit long");
             if &candidate != self {
                 sas_codes.push(candidate);
             }
@@ -174,9 +175,11 @@ impl SASCode {
         let greeter_part_as_int = ((hmac_as_int >> SAS_CODE_BITS) & SAS_CODE_MASK as u64) as u32;
 
         // Big endian number extracted from bits [0, 20[
-        let claimer_sas = SASCode::try_from(claimer_part_as_int).unwrap_or_else(|_| unreachable!());
+        let claimer_sas = SASCode::try_from(claimer_part_as_int)
+            .expect("SASCode number should not exceed `SAS_CODE_BITS` bit long");
         // Big endian number extracted from bits [20, 40[
-        let greeter_sas = SASCode::try_from(greeter_part_as_int).unwrap_or_else(|_| unreachable!());
+        let greeter_sas = SASCode::try_from(greeter_part_as_int)
+            .expect("SASCode number should not exceed `SAS_CODE_BITS` bit long");
 
         (claimer_sas, greeter_sas)
     }
@@ -191,12 +194,14 @@ macro_rules! impl_dump_and_encrypt {
         impl $name {
             pub fn dump_and_encrypt(&self, key: &::libparsec_crypto::SecretKey) -> Vec<u8> {
                 let serialized =
-                    ::rmp_serde::to_vec_named(&self).unwrap_or_else(|_| unreachable!());
+                    ::rmp_serde::to_vec_named(&self).expect("object should be serializable");
                 let mut e =
                     ::flate2::write::ZlibEncoder::new(Vec::new(), flate2::Compression::default());
                 use std::io::Write;
-                e.write_all(&serialized).unwrap_or_else(|_| unreachable!());
-                let compressed = e.finish().unwrap_or_else(|_| unreachable!());
+                let compressed = e
+                    .write_all(&serialized)
+                    .and_then(|_| e.finish())
+                    .expect("in-memory buffer should not fail");
                 key.encrypt(&compressed)
             }
         }
--- a/libparsec/crates/types/src/local_manifest.rs
+++ b/libparsec/crates/types/src/local_manifest.rs
@@ -22,7 +22,7 @@ macro_rules! impl_local_manifest_dump_load {
         impl $name {
             pub fn dump_and_encrypt(&self, key: &SecretKey) -> Vec<u8> {
                 let serialized =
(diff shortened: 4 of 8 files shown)


libparsec/crates/types/src/certif.rs
```diff
@@ -48,10 +48,12 @@ fn dump<T>(obj: &T) -> Vec<u8>
 where
     T: Serialize,
 {
-    let serialized = ::rmp_serde::to_vec_named(obj).unwrap_or_else(|_| unreachable!());
+    let serialized = ::rmp_serde::to_vec_named(obj).expect("object should be serializable");
     let mut e = ZlibEncoder::new(Vec::new(), flate2::Compression::default());
-    e.write_all(&serialized).unwrap_or_else(|_| unreachable!());
-    e.finish().unwrap_or_else(|_| unreachable!())
+    e.write_all(&serialized)
+        .expect("serialized object should be writable to ZlibEncoder");
```
FirelightFlagboy, line 54: An object that impl `Read` will always be writable to `ZlibEncoder` (`object -> read -> zlib -> write -> output`), here the possible error is not about that.

The possible error that can raise here is for the used writer. 

A writer could be anything that impl `Write`, it could be a file, a socket or in this case a `Vec` (thus writing in memory).

Writing to a file or a socket require to pass information to the kernel and with really bad luck can fail (socket is closed, file has no space left to write the data, ...)

In the end, the suggested error message here is not correct.
It should be something like

```suggestion
        .expect("The data is written in memory and should not fail");
```

TimeEngineer (approved):
LGTM

libparsec/crates/types/src/certif.rs
```diff
@@ -48,10 +48,12 @@ fn dump<T>(obj: &T) -> Vec<u8>
...
+    e.finish()
+        .expect("flushing ZlibEncoder output stream should not generate I/O error")
```
FirelightFlagboy, line 56: Could be more precise here, we won't have `I/O` error because the buffer is in memory (but we could still get an OOM but that something else :wink:)

libparsec/crates/types/src/ext_types.rs
```diff
@@ -85,7 +85,7 @@ impl<'de> serde::de::Visitor<'de> for DateTimeExtVisitor {
         let ts = f64::from_be_bytes(
             data[..F64_SIZE]
                 .try_into()
-                .unwrap_or_else(|_| unreachable!()),
+                .expect("DateTimeExtVisitor `data` should be convertible to a [u8; 8] byte array"),
```
FirelightFlagboy, line 88: The real reason is we already check the boundary on line `79`.

Here you're trying to convert a `&[u8]` to `[u8; 8]` that conversion can fail if `data.len() != 8`, that check if effectively already done above

libparsec/crates/types/src/invite.rs
```diff
@@ -191,12 +194,16 @@ macro_rules! impl_dump_and_encrypt {
         impl $name {
             pub fn dump_and_encrypt(&self, key: &::libparsec_crypto::SecretKey) -> Vec<u8> {
                 let serialized =
-                    ::rmp_serde::to_vec_named(&self).unwrap_or_else(|_| unreachable!());
+                    ::rmp_serde::to_vec_named(&self).expect("object should be serializable");
+                //Serialization can fail if T’s implementation of Serialize decides to fail.
```
FirelightFlagboy, line 198: Is this comment that useful, for me you repeat yourself with what is above

touilleMan (approved):

libparsec/crates/types/src/invite.rs
```diff
@@ -191,12 +194,16 @@ macro_rules! impl_dump_and_encrypt {
...
                 let mut e =
                     ::flate2::write::ZlibEncoder::new(Vec::new(), flate2::Compression::default());
                 use std::io::Write;
-                e.write_all(&serialized).unwrap_or_else(|_| unreachable!());
-                let compressed = e.finish().unwrap_or_else(|_| unreachable!());
+                e.write_all(&serialized)
+                    .expect("serialized object should be writable to ZlibEncoder");
+                let compressed = e
+                    .finish()
+                    .expect("flushing ZlibEncoder output stream should not generate I/O error");
```
FirelightFlagboy, line 206: Same reflection concerning in memory write

libparsec/crates/types/src/local_manifest.rs
```diff
@@ -87,7 +87,8 @@ impl Chunk {
             stop,
             raw_offset: start,
             // TODO: what to do with overflow
-            raw_size: NonZeroU64::try_from(stop.get() - start).unwrap_or_else(|_| unreachable!()),
+            raw_size: NonZeroU64::try_from(stop.get() - start)
+                .expect("Chunk `raw_size` integer should not be zero"),
             access: None,
         }
     }
```
FirelightFlagboy, line 94:
```suggestion
    }
 
```

Minor suggestion to improve the spacing between method

libparsec/crates/types/src/local_manifest.rs
```diff
@@ -100,7 +101,7 @@ impl Chunk {
             // TODO: what to do with overflow
             stop: (block_access.offset + block_access.size.get())
                 .try_into()
-                .unwrap_or_else(|_| unreachable!()),
+                .expect("Chunk `stop` integer should not be zero"),
```
FirelightFlagboy, line 104: On top of the `block_access.size` is `NonZeroU64` so the predicate is already checked

libparsec/crates/types/src/manifest.rs
```diff
@@ -30,10 +30,13 @@ macro_rules! impl_manifest_dump_load {
...
-                    ::rmp_serde::to_vec_named(&self).unwrap_or_else(|_| unreachable!());
+                    ::rmp_serde::to_vec_named(&self).expect("object should be serializable");
                 let mut e = ZlibEncoder::new(Vec::new(), Compression::default());
-                e.write_all(&serialized).unwrap_or_else(|_| unreachable!());
-                let compressed = e.finish().unwrap_or_else(|_| unreachable!());
+                e.write_all(&serialized)
+                    .expect("serialized object should be writable to ZlibEncoder");
+                let compressed = e
+                    .finish()
+                    .expect("flushing ZlibEncoder output stream should not generate I/O error");
```
FirelightFlagboy, line 39: Same reflector for the in memory write

libparsec/crates/types/src/message.rs
```diff
@@ -113,10 +113,14 @@ impl MessageContent {
...
+        let serialized =
+            rmp_serde::to_vec_named(&self).expect("MessageContent should be serializable");
         let mut e = ZlibEncoder::new(Vec::new(), flate2::Compression::default());
-        e.write_all(&serialized).unwrap_or_else(|_| unreachable!());
-        let compressed = e.finish().unwrap_or_else(|_| unreachable!());
+        e.write_all(&serialized)
+            .expect("serialized MessageContent should be writable to ZlibEncoder");
+        let compressed = e
+            .finish()
+            .expect("flushing ZlibEncoder output stream should not generate I/O error");
```
FirelightFlagboy, line 123: Same reflection for the in memory write

FirelightFlagboy (changes requested):

libparsec/crates/types/src/certif.rs
mmmarcos: OK, thanks for the explanation

libparsec/crates/types/src/certif.rs
mmmarcos: Are you suggesting something like `"in-memory buffer should not fail"` in both cases here?

FirelightFlagboy: yes

libparsec/crates/types/src/certif.rs
```diff
@@ -48,10 +48,11 @@ fn dump<T>(obj: &T) -> Vec<u8>
...
     T: Serialize,
 {
-    let serialized = ::rmp_serde::to_vec_named(obj).unwrap_or_else(|_| unreachable!());
+    let serialized = ::rmp_serde::to_vec_named(obj).expect("object should be serializable");
     let mut e = ZlibEncoder::new(Vec::new(), flate2::Compression::default());
-    e.write_all(&serialized).unwrap_or_else(|_| unreachable!());
-    e.finish().unwrap_or_else(|_| unreachable!())
+    e.write_all(&serialized)
+        .expect("in-memory buffer should not fail");
+    e.finish().expect("in-memory buffer should not fail")
```
FirelightFlagboy, line 55:
```suggestion
    e.write_all(&serialized)
        .and_then(|_| e.finish())
        .expect("in-memory buffer should not fail");
```

Should be possible to write it that way, that remove the duplicate expect

libparsec/crates/types/src/local_manifest.rs
FirelightFlagboy: @mmmarcos up

mmmarcos: I updated the message to be a little more precise about the error

libparsec/crates/types/src/local_manifest.rs
```diff
@@ -100,7 +102,9 @@ impl Chunk {
             // TODO: what to do with overflow
             stop: (block_access.offset + block_access.size.get())
                 .try_into()
-                .unwrap_or_else(|_| unreachable!()),
+                .expect(
+                    "Chunk stop should be NonZeroU64 since bloc_access.size is already NonZeroU6",
```
FirelightFlagboy, line 106:
```suggestion
                    "Chunk stop should be NonZeroU64 since bloc_access.size is already NonZeroU64",
```

FirelightFlagboy (commented):
Almost good :+1:

libparsec/crates/types/src/local_manifest.rs
FirelightFlagboy: I didn't see the comment above, maybe you published them only recently  :thinking:

FirelightFlagboy (approved):

PR #4803, opened 2023-07-03
[MS] Added pages and tests for join org modal

### LINKED ISSUE #4759: Create join organisation modal (claim user)
https://www.figma.com/file/XW3SNWHV8BhPsPJpsOCV2h/Application?type=design&node-id=21558-24460&mode=design&t=65Ig2v5O0sDtdymg-0

Max-7:
Closes #4759

Added the pages, the logic and the tests for the join org modal (claim user).

It looks ugly but that's the point of #4748.

A lot of the logic depends upon the invitation process which is not yet available. I've done my best to mock it so that all we will have to do is replace a few function calls.

- [X] Describe your changes
   
### DIFF
--- a/client/src/views/JoinOrganizationModal.vue
+++ b/client/src/views/JoinOrganizationModal.vue
@@ -1,0 +1,495 @@
+<!-- Parsec Cloud (https://parsec.cloud) Copyright (c) BUSL-1.1 (eventually AGPL-3.0) 2016-present Scille SAS -->
+
+<template>
+  <!-- the modal must be decomposed in view and components -->
+  <!-- View: ionHeader; ionButton(closeBtn); ionFooter -->
+  <!-- Components: 1 part = 1 component -->
+
+  <ion-page class="modal">
+    <ion-buttons
+      slot="end"
+      class="closeBtn-container"
+    >
+      <ion-button
+        slot="icon-only"
+        @click="cancelModal()"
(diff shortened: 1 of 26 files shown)


client/src/views/JoinOrganizationModal.vue
```diff
@@ -0,0 +1,495 @@
...
+      <!-- part 5 (get password)-->
+      <div
+        v-show="pageStep === JoinOrganizationStep.GetPassword"
+        class="step"
+        id="get-password"
+      >
+        <choose-password
+          ref="passwordPage"
+        />
+      </div>
```
fabienSvtr, line 113: The page becomes very slow when you paste a long password like this: parsec://parsec.cloud/Test?action=claim_user&token=47265123969c4d6584c2bc15960cf212
After a while, the situation returns to normal, but if you try to add characters at the end, they are not displayed for a few seconds. 

https://github.com/Scille/parsec-cloud/assets/81075511/5895207f-1844-4a05-89d9-dd4bad9dbe22

Max-7: Caused by zxcvbn with English and French dictionnaries called multiple times every time the password changed. I removed the dictionnaries so that it would be faster, and added a variable to cache the password strength so it wouldn't be called multiple times.

fabienSvtr (approved):

PR #4819, opened 2023-07-05
Update minimal used poetry version to `1.5.1`

FirelightFlagboy:
The previous version was `1.3.2`.

Closes #4808

PR #4823, opened 2023-07-05
Constraint `trio` version to `~0.22,<0.22.1`

FirelightFlagboy:


mmmarcos (approved):
Is this PR dependent on #4819? I am asking because the `"category"` field in `poetry.lock` is removed and this seems to be a change introduced in Poetry `1.5.0` https://github.com/python-poetry/poetry/blob/master/CHANGELOG.md#changed-1

Otherwise LGTM.

FirelightFlagboy:
> Is this PR dependent on https://github.com/Scille/parsec-cloud/pull/4819?

No it isn't, It's because I have `poetry-1.5.1` installed and cause the format of the `poetry.lock` file to change

Issue #4545, opened 2023-05-15
[:boxing_glove: :crab: :1st_place_medal: Final oxidation] Rework `libparsec_protocol` tests


touilleMan:
In a similar fashion to the server tests (#4543), libparsec tests should be splitted on a per `api-major-version`/`command`/`req-or-rep` base.

The idea is to auto-generate the test layout for all API versions: for a given command, its `req` and its `rep` flavors must have a dedicated test.

The goal of the auto-generation is to ensure that a test is added when introducing changes to the API \o/

```rust
// tests/serialization.rs
// Proc macro defined in `libparsec_serialization_format` that crawls the json schemas
// and generate a layout of tests
generate_serialization_tests!()
// Expands to:
// mod v4 {
//   use libparsec_protocol::authenticated_cmds::v4 as authenticated_cmds;
//   use libparsec_protocol::invited_cmds::v4 as invited_cmds;
//   use libparsec_protocol::anonymous_cmds::v4 as anonymous_cmds;
//   #[path="./v4/block_create.rs"]
//   mod block_create;
//   #[parsec_test]
//   fn block_create_rep() {
//       block_create::rep()
//   }
//   #[parsec_test]
//   fn block_create_req_ok() {
//       block_create::req_ok()
//   }
//   #[parsec_test]
//   fn block_create_req_unknown_status() {
//       block_create::req_unknown_status()
//   }
//   [...]
// }
// mod v3 {
//   use libparsec_protocol::authenticated_cmds::v3 as authenticated_cmds;
//   use libparsec_protocol::invited_cmds::v3 as invited_cmds;
//   use libparsec_protocol::anonymous_cmds::v3 as anonymous_cmds;
//   #[path="./v3/block_create.rs"]
//   mod block_create;
//   #[parsec_test]
//   fn block_create_rep() {
//       block_create::rep()
//   }
//   #[parsec_test]
//   fn block_create_req_ok() {
//       block_create::req_ok()
//   }
//   #[parsec_test]
//   fn block_create_req_unknown_status() {
//       block_create::req_unknown_status()
//   }
//   [...]
// }

// ./v4/block_create.rs
use super::authenticated_cmds;

#[path="../v3/block_create.rs"]
mod compat;

// For this command, the req schema is similar with previous major API....
pub use compat::req;

// ... but the rep has totally changed

fn rep_ok() {
// Do something
}

fn rep_unknown_status() {
// Do something
}
```

PR #4845, opened 2023-07-10
Drop support for API v2

### LINKED ISSUE #4546: [:boxing_glove: :crab: :snake: :1st_place_medal: Final oxidation] Remove APIv2
APIv2 should no longer be needed.
By removing APIv2, we can also remove from the json schemas the introduced in
minor version fields that are <3.0. This is good given the protocol generator
doesn't factorize commands that have this field set.
In other words a command that is `majors_versions: [2, 3, 4]` and have a field
marked `introduced_in_minor: 2.1` will produce three separate structures, while
a single structure shared between v3 and v4 will be produced if v2 is removed.

mmmarcos:
## Describe your changes

This PR drops support for API v2 as described in #4546.

Specifically:
- `libparsec/crates/protocol/schemas`
  - Removed version 2 from the `major_versions` array value in schemas
  - Removed the "introduced_in" field whenever its value was `2.x`
- `libparsec/crates/protocol/tests`
  - Updated tests to use v3 instead of v2 commands
  - Removed "legacy" tests that are no longer required (e.g. [bad_timestamp_legacy](https://github.com/Scille/parsec-cloud/pull/4845/files#diff-ccbe3173b57c3cbb0e56e5c03d9be7ec64ec035c5ffaabbe214efec286791ba6) where `Maybe::Absent` fields can no longer be absent in v3 onwards) 
- `server/parsec/backend`
  - Removed `apiv2_` command methods from components
  - Removed some `apiv3_` command methods that otherwise generated a fail in the following assert:  https://github.com/Scille/parsec-cloud/blob/90182b311be6710859710c42b4bcb995a4896dd2/server/parsec/backend/utils.py#L106-L107 It seemed as if they were redundant with the latest implementation, but that was not always the case (an explanation is welcome ^^)
    - @touilleMan explanation (thanks!): The `apiv3_xxx` methods are required if the `Req`/`Rep` types are different in APIv3 vs APIv4. Here, some `Req`/`Rep` used to be different but it is no longer the case after removing APIv2 because this also removed some `introduced_in` fields. The heuristic on this field is really simple: a schema with this field is never shared between APIs (i.e. if `"introduced_in" = "2.1"`, the schema in APIv3 and APIv4 could be shared but they are not)

Closes #4546

## Checklist

- [x] Describe your changes

If this PR affects the user:


mmmarcos:
> @mmmarcos you have used a `merge` to update the branch with the latest change of `origin/master` this is risky, I suggest that you redo the update using a `rebase` instead of a `merge` ;)

Yep, wrong (default) action ^^

server/parsec/backend/realm.py
```diff
@@ -121,32 +121,6 @@ def evolve(self, **kwargs: Any) -> RealmGrantedRole:
...
-    ) -> authenticated_cmds.v2.realm_create.Rep:
-        # `realm_create` command is similar between APIv2 and v4+ from the server
-        # point of view.
-        # (from client point of view, server may return `bad_timestamp` response
-        # with some fields missing)
-        # TODO: proper api req/rep conversion
-        return await self.api_realm_create(client_ctx, req)  # type: ignore[return-value, arg-type]
-
-    @api
-    async def apiv3_realm_create(
```
touilleMan, line 138: APIv3 should stay !

server/parsec/backend/realm.py
```diff
@@ -259,32 +233,6 @@ async def apiv2v3_realm_get_role_certificates(
...
-    ) -> authenticated_cmds.v2.realm_update_roles.Rep:
-        # `realm_update_roles` command is similar between APIv2 and v4+ from the server
-        # point of view.
-        # (from client point of view, server may return `bad_timestamp` response
-        # with some fields missing)
-        # TODO: proper api req/rep conversion
-        return await self.api_realm_update_roles(client_ctx, req)  # type: ignore[return-value, arg-type]
-
-    @api
-    async def apiv3_realm_update_roles(
```
touilleMan, line 276: APIv3 should stay !

server/parsec/backend/realm.py
```diff
@@ -381,32 +329,6 @@ async def api_realm_update_roles(
...
-    ) -> authenticated_cmds.v2.realm_start_reencryption_maintenance.Rep:
-        # `realm_start_reencryption_maintenance` command is similar between APIv2 and v4+ from the server
-        # point of view.
-        # (from client point of view, server may return `bad_timestamp` response
-        # with some fields missing)
-        # TODO: proper api req/rep conversion
-        return await self.api_realm_start_reencryption_maintenance(client_ctx, req)  # type: ignore[return-value, arg-type]
-
-    @api
-    async def apiv3_realm_start_reencryption_maintenance(
```
touilleMan, line 398: APIv3 should stay !

server/parsec/backend/organization.py
```diff
@@ -98,26 +98,6 @@ def __init__(self, webhooks: WebhooksComponent, config: BackendConfig):
...
-        client_ctx: AuthenticatedClientContext,
-        req: authenticated_cmds.v2.organization_config.Req,
-    ) -> authenticated_cmds.v2.organization_config.Rep:
-        # `organization_config` command is similar between APIv2 and v4+ from the server
-        # point of view.
-        # (from client point of view, server may return `ok` response with some fields missing)
-        return await self.api_authenticated_organization_config(client_ctx, req)  # type: ignore
-
-    @api
-    async def apiv3_authenticated_organization_config(
```
touilleMan, line 113: APIv3 should stay !

server/parsec/backend/invite.py
```diff
@@ -289,55 +289,6 @@ def _on_status_changed(
...
-            assert isinstance(v2_unit, authenticated_cmds.v2.invite_new.UserOrDeviceDevice)
-            unit = authenticated_cmds.latest.invite_new.UserOrDeviceDevice(
-                send_email=v2_unit.send_email,
-            )
-        # TODO: proper api req/rep conversion
-        latest_req = authenticated_cmds.latest.invite_new.Req(unit)
-        return await self.api_invite_new(client_ctx, latest_req)  # type: ignore[return-value]
-
-    @api
-    async def apiv3_invite_new(
```
touilleMan, line 319: APIv3 should stay !

FirelightFlagboy (commented):
I'm not sure about deleting `apiv3_*` and keeping `api_*`

> Not all `apiv3_*` where deleted :thinking:

libparsec/crates/protocol/tests/block.rs
```diff
@@ -2,7 +2,7 @@
 
 use hex_literal::hex;
 
-use libparsec_protocol::authenticated_cmds::v2 as authenticated_cmds;
+use libparsec_protocol::authenticated_cmds::v3 as authenticated_cmds;
```
touilleMan, line 5: use `latest` instead of a specific version of the api, this will save us modifications when we will introduce new

(as a matter of fact, we want to test all version but this is another story: https://github.com/Scille/parsec-cloud/issues/4545 ^^)

libparsec/crates/protocol/tests/invite.rs
```diff
@@ -61,20 +61,6 @@ fn serde_invite_new_req(#[case] raw: &[u8], #[case] expected: authenticated_cmds
 }
 
 #[parsec_test]
-#[case::ok_without_email_sent(
```
touilleMan, line 64: If this is no longer valid, it would be good to add it as test of invalid data (with a comment explaining this used to be valid in APIv2)

libparsec/crates/protocol/tests/organization.rs
```diff
@@ -134,23 +134,6 @@ fn serde_organization_config_req() {
 }
 
 #[parsec_test]
-#[case::ok_with_absent_sequester(
```
touilleMan, line 137: Add as invalid data test

touilleMan (changes requested):

libparsec/crates/protocol/tests/invite.rs
FirelightFlagboy: Does adding `#[should_panic]` be enough ?

mmmarcos: I am not sure this is possible with the way the tests are written (or maybe I do not understand what you mean by "test of invalid data" :thinking:)

If I restore that test, the following will not compile because `email_sent` cannot be absent in v3/v4, it needs to be of type `InvitationEmailSentStatus`:
```rust
    authenticated_cmds::invite_new::Rep::Ok {
        token: InvitationToken::from_hex("d864b93ded264aae9ae583fd3d40c45a").unwrap(),
        email_sent: Maybe::Absent, // <-- Compile Error
    }
```

I can use a specific status (such as  `InvitationEmailSentStatus::NOT_AVAILABLE` or `InvitationEmailSentStatus::BAD_RECIPIENT`) but then I wouldn't be testing "invalid data", just a different status value. 

I suppose I can also restore v2 API schemas just to use them in tests, but it seem overkill isn't it?

FirelightFlagboy: You could add `#[should_panic]` (because with the invalid bytes the `Rep::load(bytes).unwrap()` will panic) 
or you create a new test that does `unwrap_err` instead at the `Rep::load(bytes)`

libparsec/crates/protocol/tests/block.rs
mmmarcos: If I am not mistaken with `latest` we will be testing api v4, so the tests will also need to be updated whenever there is a difference between v3 and v4. Is this what you are suggesting?

server/parsec/backend/invite.py
mmmarcos: Yep! I removed it because otherwise there was an error when generating the schemas (see my comment in the PR description). I will re-add them and try to investigate more about that error.

libparsec/crates/protocol/tests/invite.rs
mmmarcos: I'm not sure how the `#[should_panic]` would help if the code is not _compiling_ 🫤
From what I understand, I need to specify a concrete a status or use a different `Rep` (because the one coming from `authenticated_cmds::latest::invite_new::Rep` does require an email_sent status)

FirelightFlagboy: You could use any value for the expected value, since it would fail at the `Rep::load` (would be safer if we used a `expect` with a message)

server/parsec/backend/invite.py
touilleMan: Ho yes you're right !

The `apiv3_xxx` method is required if the req/rep types are different in APIv3 than in APIv4, however some req/rep used to be different but this is no longer the case: since removing APIv2 remove some introduced_in fields.

This is because the heuristic on this field is really simple there: a schema with this field is never shared between APIs (i.e. if `introduced_in==2.1`, the schema in APIv3 and APIv4 could be shared but they are not)

libparsec/crates/protocol/tests/block.rs
touilleMan: yes, APIv4 is considered the base (so all tests are against it), and additional tests should be provided to test specific stuff on APIv3

And in the long term, we want instead to have a proper set of test of each API family (with a way to factorize the tests, given a new API most likely introduce few breaking changes and we don't want to duplicate 1000s of line of tests ^^)

server/parsec/backend/realm.py
mmmarcos: See https://github.com/Scille/parsec-cloud/pull/4845/files#r1276297985

server/parsec/backend/realm.py
mmmarcos: See https://github.com/Scille/parsec-cloud/pull/4845/files#r1276297985

server/parsec/backend/realm.py
mmmarcos: See https://github.com/Scille/parsec-cloud/pull/4845/files#r1276297985

libparsec/crates/protocol/tests/invite.rs
mmmarcos: Done

libparsec/crates/protocol/tests/vlob.rs
```diff
@@ -179,18 +179,28 @@ fn serde_vlob_create_req(#[case] raw: &[u8], #[case] expected: authenticated_cmd
         strictly_greater_than: "2000-1-2T01:00:00Z".parse().unwrap(),
     }
 )]
+#[should_panic(expected = "missing field `backend_timestamp`")]
 #[case::bad_timestamp_legacy(
-    // Generated from Python implementation (Parsec v2.8.1+dev)
+    // Generated from Python implementation (Parsec v2.11.1+dev)
```
FirelightFlagboy, line 184: Why the version have changed here ?

libparsec/crates/protocol/tests/vlob.rs
```diff
@@ -561,18 +571,28 @@ fn serde_vlob_update_req(#[case] raw: &[u8], #[case] expected: authenticated_cmd
...
```
FirelightFlagboy, line 576: Ditto

libparsec/crates/protocol/tests/block.rs
mmmarcos: Actually there are some significant changes between v3 and v4, including some semantic(?) changes. 

For example the `wait` parameter in `authenticated_cmds/events_listen` does no longer exists because : https://github.com/Scille/parsec-cloud/blob/cd6951efc537fabe65568207c8599e7b461dfed3/libparsec/crates/protocol/schema/authenticated_cmds/events_listen.json5#L8

I propose:
- To keep `v3` in tests to achieve the main goal of this PR (to drop API v2)
- Open another issue to "migrate" protocol tests from `v3` to `latest`.

If you are OK with this, then this PR can be re-reviewed and (hopefully!) merged.

libparsec/crates/protocol/tests/vlob.rs
mmmarcos: Because copy-paste is not always a friend! I'll fix that ^^

mmmarcos: Done

libparsec/crates/protocol/tests/vlob.rs
mmmarcos: Done

libparsec/crates/protocol/tests/block.rs
touilleMan: yeah I'm good with that ^^

touilleMan (approved):

FirelightFlagboy (approved):

PR #4891, opened 2023-07-24
Use `python-version-file` to load the python version in `pyproject.toml`

### LINKED ISSUE #4869: Configure GA `actions/setup-python` to use the python version specified in `server/pyproject.toml`
~Adding this field simplify the usage of the github action `actions/setup-python` that can now read the `pyproject.toml` file in the search for that field to install the expected python version.~

The action `actions/setup-python` can now read a pyproject file to search for the required python version in the field `object.requires-python` or `tool.poetry.dependencies.python`.

We already have the field `tool.poetry.dependencies.python`, so we just need to update the configuration of the action to use the pyproject file instead of the provided python version.

FirelightFlagboy:
Closes #4869

### DIFF
--- a/.github/actions/setup-python-poetry/action.yml
+++ b/.github/actions/setup-python-poetry/action.yml
@@ -2,10 +2,6 @@ name: setup-python-poetry
 description: Install Python & Poetry
 
 inputs:
-  python-version:
-    description: Python version to install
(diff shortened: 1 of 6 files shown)


mmmarcos (approved):
LGTM

PR #4916, opened 2023-07-27
Add more groups to python manager

FirelightFlagboy:


.github/dependabot.yml
```diff
@@ -82,9 +82,33 @@ updates:
...
           - "*"
+        exclude-patterns:
+          - sphinx*
+          - pytest-*
+          - hypothesis
+          - ruff
+          - black
+          - mypy
+          - maturin
+          - editorconfig-checker
```
mmmarcos, line 111: It would be nice to avoid duplication. 
Isn't there a syntax to re-use `pytest`, `pytools`, ... in the `pydeps` exclude-patterns?
If not, maybe if we completely remove the `pydeps` group, the deps that do not match the previous groups will be "grouped" together anyway? :thinking:

FirelightFlagboy: > Isn't there a syntax to re-use [...] `exclude-patterns`

Yes, YAML syntax support anchor to avoid repetition (https://support.atlassian.com/bitbucket-cloud/docs/yaml-anchors/) sadly the parser used by `dependabot-core` don't support that :cry: dependabot/dependabot-core#1582

> If not, [...] if we completely remove the `pydeps` group, the deps that do not match the previous groups will be "grouped" together anyway?

No, the deps that doesn't match a group pattern will have a dedicated PR for it

mmmarcos: Then maybe list the `exclude-patterns` in the same order just so it is easier to check if a pattern is missing (btw the `setuptools` pattern is missing?)

```suggestion
      pydeps:
        patterns:
          - "*"
        exclude-patterns:
          - pytest-*
          - hypothesis
          - ruff
          - black
          - mypy
          - maturin
          - editorconfig-checker
          - sphinx*
```

FirelightFlagboy: Done, that's better indeed

mmmarcos (approved):
:+1:

PR #4919, opened 2023-07-27
Greedy HTTP Proxy config validation

touilleMan:


libparsec/crates/client_connection/src/anonymous_cmds.rs
```diff
@@ -36,11 +36,11 @@ impl AnonymousCmds {
     pub fn new(
         config_dir: &Path,
         addr: BackendAnonymousAddr,
-        config: ProxyConfig,
-    ) -> reqwest::Result<Self> {
+        proxy: ProxyConfig,
+    ) -> anyhow::Result<Self> {
```
FirelightFlagboy, line 40: `reqwest::Error` is the only possible error, using `anyhow` here seems overkill.
Wouldn't it be better to let the caller handle that ?

libparsec/crates/platform_http_proxy/src/lib.rs
```diff
@@ -63,50 +56,56 @@ mod native {
 
     impl ProxyConfig {
         /// Set the http proxy to use, will overwrite the previous configuration.
-        pub(crate) fn with_http_proxy<S: AsRef<str>>(mut self, proxy: S) -> Self {
-            self.http_proxy.replace(proxy.as_ref().to_string());
-            self
+        pub(crate) fn with_http_proxy(mut self, proxy: String) -> anyhow::Result<Self> {
```
FirelightFlagboy, line 59: Why requiring an allocated String ? `impl AsRef<str>` or `impl IntoProxyScheme` would be better IMO

libparsec/crates/client_connection/src/anonymous_cmds.rs
touilleMan: The rational here is there is two kinds of errors:
- The specific ones we care about: this is the functional errors that are defined on a per-function basis (a given function has multiple expected errors outcomes, hence we create a dedicated error type for the function listing each one of them)
- The unexpected ones: for instance accessing the file system may fail in a multiple of way (that depend on the platform for extra fun ^^)

In the second case, case we don't care about the specific error, we just want to represent a generic error with a human-readable representation of it initial reason.

Here we are the the second case: the only possible error is a `reqwest::Error` however it's not relevant given we are not going to do any processing depending on it actual type. If we had such type then we would create our own `AnonymousCmdsNewError` type to specify the different outcomes we care about.

PR #4504, opened 2023-05-02
Implementation of shamir-based shared recovery device 

TimeEngineer:
# What has changed ?

See: https://github.com/Scille/parsec-cloud/blob/master/docs/rfcs/1000-shamir-based-recovery.md
- `invite_list` changes are not indicated in the RFC, I chose to add `ShamirRecovery` in `InviteListItem`
- I used `sharks` crate for the `shamir secret sharing` algorithm
- I bumped `parsec API` to ~~`v4`~~
- I implemented a set of tests in `test_shamir.py`

- [ ] Does this PR affect the User ?

depends on #4466

PR #4921, opened 2023-07-27
Add missing exception BackendInvitationShamirRecoveryNotSetup

vxgmichel:
Fixup for #4504


src/backend_connection/error.rs
```diff
@@ -22,6 +22,13 @@ pyo3::create_exception!(
     BackendInvitationOnExistingMember,
     BackendConnectionError
 );
+// TODO: `LoggedCore.new_user_invitation` raise `BackendInvitationOnExistingMember`
```
touilleMan, line 25: is there something to do ? or is this for the future vxgmichel

>  These are the tales of those times. Forget the power of Rust and Python, for so much has been forgotten, never to be relearned. Forget the promise of progress and understanding, for in the grim dark future there is only PHP4.
> --- future vxgmichel

touilleMan (approved):

src/backend_connection/error.rs
vxgmichel: Only a sad copy/paste from there: https://github.com/Scille/parsec-cloud/pull/4921/files/5351af53a47825df933cede7b6b142a84545e41f#diff-b2c3a9e82f595b0292acf758e26ffac1bace35ad68c9c81da51afc29525d7356R18-R19

vxgmichel: I forgot to rename it accordingly tho

PR #4927, opened 2023-07-27
Minor fixes

touilleMan:


libparsec/crates/types/src/local_device_file.rs
```diff
@@ -153,15 +153,14 @@ impl DeviceFileType {
     }
 }
 
-#[derive(Debug, Clone, PartialEq, Eq, Hash, Serialize)]
+#[derive(Debug, Clone, PartialEq, Eq, Hash)]
```
FirelightFlagboy, line 156: What is the reason for the `Serialize` to be remove, we no longer want `AvailableDevice` to be serializable ?

touilleMan: `AvailableDevice` is never serialized (it is only provided to the GUI to list the devices detected on disk)

FirelightFlagboy: thanks

FirelightFlagboy (approved):

touilleMan:
included in #4942

PR #4929, opened 2023-07-27
Decouple get_shamir_recovery_share_data from ShamirRecoveryGreetInitialCtx

vxgmichel:
.. and add prelude attributes to all shamir recovery claim ctx.

Fixup #4504


touilleMan (approved):
:+1:

PR #5008, opened 2023-08-16
Debug winfsp tests in CI

vxgmichel:
Just debugging stuff, don't mind me


.github/workflows/ci-python.yml
```diff
@@ -127,14 +127,24 @@ jobs:
         if: startsWith(matrix.os, 'windows-')
         shell: bash
         run: |
+          set -eux
           choco install -y --limit-output winfsp --version=${{ matrix.winfsp-version }}
           mkdir winfsp-test
           curl -L https://github.com/billziss-gh/winfsp/releases/download/v1.8/winfsp-tests-1.8.20304.zip -o winfsp-test/winfsp-tests.zip
           unzip winfsp-test/winfsp-tests.zip -d winfsp-test
```
FirelightFlagboy, line 134: Use The same path everywhere & use the temp directory instead.

```suggestion
          mkdir -vp ${{ runner.temp }}/winfsp-test
          curl -L https://github.com/billziss-gh/winfsp/releases/download/v1.8/winfsp-tests-1.8.20304.zip -o ${{ runner.temp }}/winfsp-test/winfsp-tests.zip
          unzip ${{ runner.temp }}/winfsp-test/winfsp-tests.zip -d ${{ runner.temp }}/winfsp-test
```

.github/workflows/ci-python.yml
```diff
@@ -127,14 +127,24 @@ jobs:
...
 
+      - name: (🏁 Windows) Check that winfsp-tests is available
+        if: startsWith(matrix.os, 'windows-')
+        shell: python
+        run: |
+          from distutils.spawn import find_executable
+          path = find_executable("winfsp-tests-x86")
+          if not path:
+            raise RuntimeError("winfsp-tests-x86 not found")
+          print(path)
```
FirelightFlagboy, line 146: Why not removing the check at https://github.com/Scille/parsec-cloud/blob/1898409da00fb4879afaa453cd417dc770487921/tests/core/mountpoint/test_winfsp_tests.py#L15-L18 ?

vxgmichel: because it's ok to run the tests without `winfsp-tests`, locally for instance

.github/workflows/ci-python.yml
vxgmichel: is `runner.temp` really persistent between jobs? is it really worth making this change?

FirelightFlagboy: `runner.temp` goal is not to be persistent between job (the directory is cleared).
Using `github.workspace` or `runner.temp` is the same, the installed `winfst-tests` isn't keep

.github/workflows/ci-python.yml
FirelightFlagboy: Why not using another markers for tests that require winfsp tests ?

TimeEngineer (approved):

.github/workflows/ci-python.yml
vxgmichel: Because the winfsp-tests executable is necessary to get the list of the test cases.

If the executable is not available, there are no tests to collect hence no tests to mark.

.github/workflows/ci-python.yml
vxgmichel: `github.workspace` is persistent, otherwise our winfsp-tests wouldn't run. 
The whole point of this is task to make the `winfsp-tests` executable available for the futures tasks, in particular the mounpoint tests.

PR #5021, opened 2023-08-17
Add `winfsp-tests` & `mountpoint` on `windows`

TimeEngineer:
## Checklist

- [ ] Describe your changes
   
If this PR is related to GUI:

If this PR affects the user:

### DIFF
--- a/.cspell/custom-words.txt
+++ b/.cspell/custom-words.txt
@@ -133,6 +133,7 @@ excrepr
 executemany
 executescript
 farn
+fdata
 fetchrow
 fetchval
 fghij
@@ -161,6 +162,8 @@ geoclue
 geometryloaders
 geoservices
 getcapacitor
+getfileattr
+getfileinfo
 getplugin
 GetRandom
 gloo
@@ -334,6 +337,7 @@ nosetests
 notbase
 NOTIFS
 notr
+notraverse
 notuptodate
 nplurals
 npmkeep
@@ -434,10 +438,12 @@ qrichtext
 qtbot
 qthelp
 qtrio
+querydir
 quickstart
 Quotify
 randombytes
 randr
+readonlydir
 recoverin
 reenc
 Reencrypt
@@ -494,10 +500,12 @@ Serde
 serialnmea
 sessionfinish
 sessionstart
+setfileinfo
 setlocal
 setmainfont
 setmonofont
 setsansfont
+setsecurity
 shakeplugin
 shortid
 SignatureBytes
--- a/.github/workflows/ci-rust.yml
+++ b/.github/workflows/ci-rust.yml
@@ -17,6 +17,8 @@ env:
   cargo-nextest-version: 0.9.53
   CARGO_CI_FLAGS: --locked --profile ci-rust
   CARGO_NEXTEST_CI_FLAGS: --locked --cargo-profile ci-rust
+  WINFSP_VERSION: 2.0.23075
+  WINFSP_TEST_EXE: C:/Program Files (x86)/WinFsp/bin/winfsp-tests-x64.exe
 
 permissions:
   contents: read
@@ -132,7 +134,7 @@ jobs:
             os: windows-2022
     name: "${{ matrix.name }}: 🦀 Rust tests"
     # Just a fail-safe timeout, see the fine grain per-task timeout instead
-    timeout-minutes: 30
+    timeout-minutes: 60
     runs-on: ${{ matrix.os }}
     steps:
       - uses: actions/checkout@3df4ab11eba7bda6032a0b82a6bb43b11571feac # pin v4.0.0
@@ -163,12 +165,20 @@ jobs:
         run: echo OPENSSL_SRC_PERL=C:/Strawberry/perl/bin/perl >> $GITHUB_ENV
         timeout-minutes: 1
 
+      - name: Install winfsp
+        if: startsWith(matrix.os, 'windows')
+        shell: bash
+        run: |
+          set -eux
+          choco install winfsp -y --version=${{ env.WINFSP_VERSION }}
+          curl -L https://github.com/winfsp/winfsp/releases/download/v2.0/winfsp-tests-${{ env.WINFSP_VERSION }}.zip -o D:/a/_temp/winfsp-tests.zip
+          unzip D:/a/_temp/winfsp-tests.zip -d D:/a/_temp/
+          mv 'D:/a/_temp/winfsp-tests-x64.exe' 'C:/Program Files (x86)/WinFsp/bin/'
+
       # Install cargo nextest command
       - uses: taiki-e/install-action@b89cfc4cdbf6930a4276f4c1d662f3217ef3edae # pin v2.18.6
         with:
           tool: nextest@${{ env.cargo-nextest-version }}
-
-      # Don't run platform agnostic tests here (only ran on Linux)
 
       - name: Test Rust codebase (${{ matrix.name }} specific)
         shell: bash
@@ -181,7 +191,7 @@ jobs:
           cargo nextest run ${{ env.CARGO_NEXTEST_CI_FLAGS }} --package libparsec_crypto --features use-sodiumoxide
           NON_BINDINGS_CRATES=`python3 misc/libparsec_crates_flags.py agnostic platform`
           cargo check ${{ env.CARGO_CI_FLAGS }} $NON_BINDINGS_CRATES --features use-sodiumoxide
-        timeout-minutes: 30
+        timeout-minutes: 50 # It can be very slow if cache has missed
         env:
           RUST_LOG: debug
           TESTBED_SERVER: http://localhost:6777
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -7,6 +7,7 @@ members = [
     "libparsec/crates/crypto",
     "libparsec/crates/platform_async",
     "libparsec/crates/platform_device_loader",
+    "libparsec/crates/platform_mountpoint",
     "libparsec/crates/platform_storage",
     "libparsec/crates/protocol",
     "libparsec/crates/serialization_format",
@@ -28,6 +29,7 @@ default-members = [
     "libparsec/crates/crypto",
     "libparsec/crates/platform_async",
     "libparsec/crates/platform_device_loader",
+    "libparsec/crates/platform_mountpoint",
     "libparsec/crates/platform_storage",
     "libparsec/crates/protocol",
     "libparsec/crates/serialization_format",
@@ -67,6 +69,7 @@ libparsec_protocol = { path = "libparsec/crates/protocol", default-features = fa
 libparsec_crypto = { path = "libparsec/crates/crypto", default-features = false }
 libparsec_types = { path = "libparsec/crates/types", default-features = false }
 libparsec_platform_device_loader = { path = "libparsec/crates/platform_device_loader", default-features = false }
+libparsec_platform_mountpoint = { path = "libparsec/crates/platform_mountpoint", default-features = false }
 libparsec_platform_storage = { path = "libparsec/crates/platform_storage", default-features = false }
 libparsec_serialization_format = { path = "libparsec/crates/serialization_format", default-features = false }
 libparsec_testbed = { path = "libparsec/crates/testbed", default-features = false }
@@ -163,6 +166,8 @@ wasm-bindgen-futures = { version = "0.4.37", default-features = false }
 wasm-bindgen-test = { version = "0.3.37", default-features = false }
 wasm-bindgen = { version = "0.2.87", default-features = false }
 web-sys = { version = "0.3.64", default-features = false }
+winfsp_build = { version = "0.1.1", default-features = false }
+winfsp_wrs = { git = "https://github.com/Scille/winfsp_wrs.git", rev = "debed94", default-features = false }
 # Using a non-stable version is exceptionally allowed since no significant changes
 # were made from the last stable version to this major pre-release version.
 # TODO: bump to a stable version.
--- a/libparsec/crates/platform_mountpoint/build.rs
+++ b/libparsec/crates/platform_mountpoint/build.rs
@@ -1,0 +1,14 @@
+// Parsec Cloud (https://parsec.cloud) Copyright (c) BUSL-1.1 2016-present Scille SAS
+
+fn main() {
+    #[cfg(target_os = "windows")]
+    // WinFSP is provided as a DLL that our main binary must loaded.
+    // By default Windows look for DLL in 1) the executable directory and 2) the Windows system folders.
+    // However this doesn't work for WinFSP given it is distributed as a separate program
+    // (and hence end up install in an arbitrary directory we must retrieve at the runtime
+    // by querying the Windows Registry) .
+    // Long story short, we are here here informing the linker the WinFSP DLL must be lazy
+    // loaded ("delayload" option in MSVC) so that we will have time to first configure the
+    // lookup directory.
+    winfsp_build::build();
+}
--- a/libparsec/crates/platform_mountpoint/src/error.rs
+++ b/libparsec/crates/platform_mountpoint/src/error.rs
@@ -1,0 +1,19 @@
+// Parsec Cloud (https://parsec.cloud) Copyright (c) BUSL-1.1 2016-present Scille SAS
+
+use thiserror::Error;
+
+#[derive(Debug, Error)]
+pub(crate) enum MountpointError {
+    #[error("Access denied")]
+    AccessDenied,
+    #[error("Dir is not empty")]
+    DirNotEmpty,
+    #[error("End of file")]
+    EndOfFile,
+    #[error("Name collision")]
+    NameCollision,
+    #[error("Not found")]
+    NotFound,
+}
+
+pub(crate) type MountpointResult<T> = Result<T, MountpointError>;
--- a/libparsec/crates/platform_mountpoint/src/lib.rs
+++ b/libparsec/crates/platform_mountpoint/src/lib.rs
@@ -1,0 +1,99 @@
+// Parsec Cloud (https://parsec.cloud) Copyright (c) BUSL-1.1 2016-present Scille SAS
+
+#![doc = include_str!("../README.md")]
+
+mod error;
+mod memfs;
+#[cfg(target_os = "windows")]
+mod windows;
+
+use chrono::{DateTime, Utc};
+use std::{collections::HashMap, path::Path};
+#[cfg(target_os = "windows")]
+pub use windows::mount;
+
+use libparsec_types::{EntryID, EntryName, FileDescriptor};
+
+pub(crate) use error::{MountpointError, MountpointResult};
+pub use memfs::MountpointManager;
+
+// TODO: remove me once ParsecEntryInfo is used
+#[allow(dead_code)]
+#[derive(Debug, Clone, Copy, PartialEq)]
+pub(crate) enum EntryInfoType {
+    Dir,
+    File,
+    /// Special Parsec type for icon overlay handler
+    ParsecEntryInfo,
+}
+
+/// - Normal: write normally from offset to offset + data.len().
+/// - Constrained: write from offset without exceeding the file size limit.
+/// - StartEOF: write starting at end of file.
+// Remove me once fuse use WriteMode
+#[allow(dead_code)]
+pub(crate) enum WriteMode {
+    Normal,
+    Constrained,
+    StartEOF,
+}
+
+// TODO: remove me once fuse is implemented
+#[allow(dead_code)]
+#[derive(Debug, Clone)]
+pub(crate) struct EntryInfo {
+    id: EntryID,
+    ty: EntryInfoType,
+    created: DateTime<Utc>,
+    updated: DateTime<Utc>,
+    size: u64,
+    need_sync: bool,
+    children: HashMap<EntryName, EntryID>,
+}
+
+pub(crate) trait MountpointInterface {
+    // Rights check
+    fn check_read_rights(&self, path: &Path) -> MountpointResult<()>;
+    fn check_write_rights(&self, path: &Path) -> MountpointResult<()>;
+
+    // Entry transactions
+
+    fn entry_info(&self, path: &Path) -> MountpointResult<EntryInfo>;
+    fn entry_rename(
+        &self,
+        source: &Path,
+        destination: &Path,
+        overwrite: bool,
+    ) -> MountpointResult<()>;
+
+    // Directory transactions
+
+    fn dir_create(&self, path: &Path) -> MountpointResult<()>;
+    fn dir_delete(&self, path: &Path) -> MountpointResult<()>;
+
+    // File transactions
+
+    fn file_create(&self, path: &Path, open: bool) -> MountpointResult<FileDescriptor>;
+    fn file_open(&self, path: &Path, write_mode: bool) -> MountpointResult<Option<FileDescriptor>>;
+    fn file_delete(&self, path: &Path) -> MountpointResult<()>;
+    fn file_resize(&self, path: &Path, len: u64) -> EntryID;
+
+    // File descriptor transactions
+
+    fn fd_close(&self, fd: FileDescriptor);
+    fn fd_read(
+        &self,
+        fd: FileDescriptor,
+        buffer: &mut [u8],
+        offset: u64,
+    ) -> MountpointResult<usize>;
+    fn fd_write(
+        &self,
+        fd: FileDescriptor,
+        data: &[u8],
+        offset: u64,
+        mode: WriteMode,
+    ) -> MountpointResult<usize>;
+    fn fd_resize(&self, fd: FileDescriptor, len: u64, truncate_only: bool) -> MountpointResult<()>;
+    fn fd_flush(&self, fd: FileDescriptor);
+}
--- a/libparsec/crates/platform_mountpoint/src/windows/mod.rs
+++ b/libparsec/crates/platform_mountpoint/src/windows/mod.rs
@@ -1,0 +1,495 @@
+// Parsec Cloud (https://parsec.cloud) Copyright (c) BUSL-1.1 2016-present Scille SAS
+
+mod error;
+
+use once_cell::sync::{Lazy, OnceCell};
+use std::{
+    ops::Deref,
+    path::{Path, PathBuf},
+    sync::{Arc, Mutex},
+};
+
+use winfsp_wrs::{
+    filetime_from_utc, filetime_now, u16cstr, CleanupFlags, CreateFileInfo, CreateOptions,
+    FileAccessRights, FileAttributes, FileInfo, FileSystem, FileSystemContext, PSecurityDescriptor,
+    Params, SecurityDescriptor, U16CStr, U16CString, VolumeInfo, VolumeParams,
+    WriteMode as WinWriteMode, NTSTATUS, STATUS_DIRECTORY_NOT_EMPTY, STATUS_NOT_IMPLEMENTED,
+    STATUS_RESOURCEMANAGER_READ_ONLY,
+};
+
+use libparsec_types::prelude::*;
+
+use crate::{EntryInfo, EntryInfoType, MountpointInterface, MountpointManager, WriteMode};
+
+/// we currently don't support arbitrary security descriptor and instead use only this one
+/// https://docs.microsoft.com/fr-fr/windows/desktop/SecAuthZ/security-descriptor-string-format
+static SECURITY_DESCRIPTOR: Lazy<SecurityDescriptor> = Lazy::new(|| {
+    SecurityDescriptor::from_wstr(u16cstr!("O:BAG:BAD:P(A;;FA;;;SY)(A;;FA;;;BA)(A;;FA;;;WD)"))
+        .expect("unreachable, valid SecurityDescriptor")
+});
+
+static VOLUME: OnceCell<Arc<Mutex<VolumeInfo>>> = OnceCell::new();
+
+/// Used for the virtual file to retrieve entry info
+/// This is used as a way for external applications to get information
+/// on a file in a Parsec mountpoint by opening and reading the
+/// virtual file f"{file_path}{ENTRY_INFO_EXTENSION}".
+const ENTRY_INFO_EXTENSION: &str = "__parsec_entry_info__";
+
+const VOLUME_LABEL: &U16CStr = u16cstr!("parsec");
+
+/// The sector size is what is set in VolumeParams via the SectorSize
+/// property when the file system is first created. (And of course the
+/// "sector size" is misnamed for non-disk file systems, but it determines
+/// the granularity at which non-buffered I/O is done.)
+/// see: https://github.com/winfsp/winfsp/issues/240#issuecomment-518629301
+const SECTOR_SIZE: u16 = 512;
+
+fn is_entry_info_path(path: &Path) -> bool {
+    path.extension()
+        .map(|ext| ext == ENTRY_INFO_EXTENSION)
+        .unwrap_or_default()
+}
+
+impl From<&EntryInfo> for FileInfo {
+    fn from(value: &EntryInfo) -> Self {
+        let created = filetime_from_utc(value.created);
+        let updated = filetime_from_utc(value.updated);
+
+        match value.ty {
+            EntryInfoType::File => *FileInfo::default()
+                .set_file_attributes(value.ty.into())
+                .set_creation_time(created)
+                .set_last_access_time(updated)
+                .set_last_write_time(updated)
+                .set_change_time(updated)
+                .set_index_number(value.id.as_u128() as u64)
+                .set_file_size(value.size)
+                // AllocationSize is the size actually occupied on the storage medium, this is
+                // meaningless for Parsec (however WinFSP requires FileSize <= AllocationSize)
+                .set_allocation_size(value.size),
+            EntryInfoType::Dir => *FileInfo::default()
+                .set_file_attributes(value.ty.into())
+                .set_creation_time(created)
+                .set_last_access_time(updated)
+                .set_last_write_time(updated)
+                .set_change_time(updated)
+                .set_index_number(value.id.as_u128() as u64),
+            EntryInfoType::ParsecEntryInfo => *FileInfo::default()
+                .set_file_attributes(value.ty.into())
+                // TODO: should we use 0 value instead ?
+                // Arbitrary non-zero size
+                .set_file_size(1024)
+                .set_time(filetime_now()),
+        }
+    }
+}
+
+impl From<EntryInfoType> for FileAttributes {
+    fn from(value: EntryInfoType) -> Self {
+        FileAttributes::not_content_indexed()
+            | match value {
+                EntryInfoType::File => FileAttributes::archive(),
+                EntryInfoType::Dir => FileAttributes::directory(),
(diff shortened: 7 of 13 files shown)


.github/workflows/ci-rust.yml
```diff
@@ -146,13 +147,30 @@ jobs:
...
 
+      - name: Install winfsp
+        if: startsWith(matrix.os, 'windows')
+        shell: bash
+        run: |
+          set -eux
+          choco install winfsp -y --version=${{ env.WINFSP_VERSION }}
+          mkdir -vp ${{ runner.temp }}
+          curl -L https://github.com/winfsp/winfsp/releases/download/v2.0/winfsp-tests-${{ env.WINFSP_VERSION }}.zip -o ${{ runner.temp }}/winfsp-tests.zip
+          unzip ${{ runner.temp }}/winfsp-tests.zip -d ${{ runner.temp }}/winfsp-tests.exe
```
FirelightFlagboy, line 158: `-d` option is to specify an output directory here, `winfsp-test.exe` will be a directory (not a file ;) )

```suggestion
          unzip ${{ runner.temp }}/winfsp-tests.zip -d ${{ runner.temp }}/winfsp-tests
```

.github/workflows/ci-rust.yml
```diff
@@ -146,13 +147,30 @@ jobs:
         run: echo OPENSSL_SRC_PERL=C:/Strawberry/perl/bin/perl >> $GITHUB_ENV
         timeout-minutes: 1
 
+      - name: Install winfsp
+        if: startsWith(matrix.os, 'windows')
+        shell: bash
+        run: |
+          set -eux
+          choco install winfsp -y --version=${{ env.WINFSP_VERSION }}
+          mkdir -vp ${{ runner.temp }}
```
FirelightFlagboy, line 156:
```suggestion
```

The temporary folder is already present, no need to create it

.github/workflows/ci-rust.yml
```diff
@@ -146,13 +147,30 @@ jobs:
...
           tool: nextest@${{ env.cargo-nextest-version }}
 
       # Don't run platform agnostic tests here (only runned on Linux)
 
+      - name: Test with winfsp-tests
+        if: startsWith(matrix.os, 'windows')
+        shell: bash
+        run: cargo nextest run ${{ env.CARGO_NEXTEST_CI_FLAGS }} --package winfsp_tests
+        env:
+          WINFSP_TEST_EXE: ${{ runner.temp }}/winfsp-tests.exe
```
FirelightFlagboy, line 172: According to the comment about the `unzip -d ..` that value need to be changed

```suggestion
          WINFSP_TEST_EXE: ${{ runner.temp }}/winfsp-tests/winfsp-tests.exe
```

TimeEngineer: Ty

.github/workflows/ci-rust.yml
TimeEngineer: Ty

.github/workflows/ci-rust.yml
TimeEngineer: Ty

.cspell/custom-words.txt
```diff
@@ -50,6 +50,7 @@ clearkey
 CLSID
 Cmds
 complexify
+configurator
```
FirelightFlagboy, line 53: Could be removed

.cspell/custom-words.txt
```diff
@@ -84,10 +85,12 @@ duckpy
 dvipdfmx
 egldeviceintegrations
 EINVAL
+engineerd
```
FirelightFlagboy, line 88: Could be removed

Cargo.toml
```diff
@@ -143,6 +147,7 @@ wasm-bindgen-futures = { version = "0.4.37", default-features = false }
 wasm-bindgen-test = { version = "0.3.37", default-features = false }
 wasm-bindgen = { version = "0.2.87", default-features = false }
 web-sys = { version = "0.3.64", default-features = false }
+winfsp_wrs = { git = "https://github.com/Scille/winfsp_wrs.git", default-features = false }
```
FirelightFlagboy, line 150: Specify a revision or a tag

libparsec/crates/platform_mountpoint/src/error.rs
```diff
@@ -0,0 +1,19 @@
+// Parsec Cloud (https://parsec.cloud) Copyright (c) BUSL-1.1 (eventually AGPL-3.0) 2016-present Scille SAS
+
+use thiserror::Error;
+
+#[derive(Debug, Error)]
+pub(crate) enum MountpointError {
```
FirelightFlagboy, line 6: Should the variants have more context ? I feel likes it's missing the path that generate the error :thinking:

.cspell/custom-words.txt
```diff
@@ -84,10 +85,12 @@ duckpy
...
 enmascarado
 excrepr
 executemany
 farn
+fdata
```
FirelightFlagboy, line 93: Could be removed

libparsec/crates/platform_mountpoint/src/error.rs
TimeEngineer: `Windows` doesn't handle any path

FirelightFlagboy: How that possible ? to open a file I need a path

TimeEngineer: It uses `GetLastError` function

.github/workflows/ci-rust.yml
```diff
@@ -146,13 +147,31 @@ jobs:
...
+          ls ${{ runner.temp }}
+          unzip ${{ runner.temp }}/winfsp-tests.zip -d ${{ runner.temp }}
+          ls ${{ runner.temp }}
+
       # Install cargo nextest command
       - uses: taiki-e/install-action@7412a5eb82f4686390763b0f92637284e91b6627 # pin v2.15.2
         with:
           tool: nextest@${{ env.cargo-nextest-version }}
 
       # Don't run platform agnostic tests here (only runned on Linux)
```
FirelightFlagboy, line 166: This comment is false :sweat:

.github/workflows/ci-rust.yml
```diff
@@ -146,13 +147,31 @@ jobs:
         run: echo OPENSSL_SRC_PERL=C:/Strawberry/perl/bin/perl >> $GITHUB_ENV
         timeout-minutes: 1
 
+      - name: Install winfsp
+        if: startsWith(matrix.os, 'windows')
+        shell: bash
+        run: |
+          set -eux
+          choco install winfsp -y --version=${{ env.WINFSP_VERSION }}
+          curl -L https://github.com/winfsp/winfsp/releases/download/v2.0/winfsp-tests-${{ env.WINFSP_VERSION }}.zip -o ${{ runner.temp }}
```
FirelightFlagboy, line 156: `-o` or `--output` is the output file, here the value is a directory, you should add the filename

.cspell/custom-words.txt
FirelightFlagboy: Sorry, it was used :sweat_smile:

libparsec/crates/platform_mountpoint/src/error.rs
TimeEngineer: It is called automatically for the user

.github/workflows/ci-rust.yml
TimeEngineer: Something goes wrong, because D:\a\_temp exists, but ${{ runner.temp }} is D:\a_temp :/ @FirelightFlagboy

.cspell/custom-words.txt
TimeEngineer: How should I resolve `cspell` ?

touilleMan (changes requested):

libparsec/crates/platform_mountpoint/src/windows/mod.rs
```diff
@@ -0,0 +1,473 @@
...
+            Context::Dir(_) => unreachable!(),
+        })
+    }
+
+    fn write(
+        &self,
+        file_context: &Self::FileContext,
+        buffer: &[u8],
+        offset: u64,
+        write_to_end_of_file: bool,
```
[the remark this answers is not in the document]
TimeEngineer: Done

libparsec/crates/platform_mountpoint/src/windows/mod.rs
```diff
@@ -0,0 +1,473 @@
...
+    path: PathBuf,
+    fd: FileDescriptor,
+}
+
+#[derive(Debug, Clone)]
+pub struct Dir {
+    path: PathBuf,
+}
+
+#[derive(Debug, Clone)]
```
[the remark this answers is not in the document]
TimeEngineer: Done

libparsec/crates/platform_mountpoint/src/windows/mod.rs
```diff
@@ -0,0 +1,473 @@
...
+            return Ok(Arc::new(Mutex::new(Context::Info(ParsecEntryInfo {
+                path,
+                encoded: format!("{{\"need_sync\":{need_sync}}}"),
+            }))));
+        }
+
+        if let Some(fd) = fd {
+            Ok(Arc::new(Mutex::new(Context::File(File { path, fd }))))
+        } else {
+            Ok(Arc::new(Mutex::new(Context::Dir(Dir { path }))))
```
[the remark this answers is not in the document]
TimeEngineer: Done

libparsec/crates/platform_mountpoint/src/windows/mod.rs
```diff
@@ -0,0 +1,473 @@
...
+    fn read(&self, buffer: &mut [u8], offset: usize) -> usize {
+        let end_offset = std::cmp::min(self.encoded.len(), offset + buffer.len());
+        let data = &self.encoded.as_bytes()[offset..end_offset];
+        buffer[..data.len()].copy_from_slice(data);
+        data.len()
+    }
+}
+
+impl FileSystemContext for MountpointManager {
+    type FileContext = Arc<Mutex<Context>>;
```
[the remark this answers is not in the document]
TimeEngineer: It worked because behind the wrapper, I use Box before to pass to `winfsp`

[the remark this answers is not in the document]
touilleMan: So after more reading on the WinFSP API, a little change on what winfsp-wrs should expose:

From [winfsp documentation](https://winfsp.dev/doc/WinFsp-API-winfsp.h/):
> Most of the operations accept a FileContext parameter. This parameter has different meanings depending on the value of the FSP_FSCTL_VOLUME_PARAMS flags UmFileContextIsUserContext2 and UmFileContextIsFullContext.
>
> There are three cases to consider:
>
>    When both of these flags are unset (default), the FileContext parameter represents the file node. The file node is a void pointer (or an integer that can fit in a pointer) that is used to uniquely identify an open file. Opening the same file name should always yield the same file node value for as long as the file with that name remains open anywhere in the system.
>
>    When the UmFileContextIsUserContext2 is set, the FileContext parameter represents the file descriptor. The file descriptor is a void pointer (or an integer that can fit in a pointer) that is used to identify an open instance of a file. Opening the same file name may yield a different file descriptor.
>
>    When the UmFileContextIsFullContext is set, the FileContext parameter is a pointer to a FSP_FSCTL_TRANSACT_FULL_CONTEXT. This allows a user mode file system to access the low-level UserContext and UserContext2 values. The UserContext is used to store the file node and the UserContext2 is used to store the file descriptor for an open file.

So long story short:
- winfsp-wrs has to deal with FileContext being either A) a pointer-sized scalar (i.e. a `u32` that correspond to a handle) or B) an actual pointer on data (in this case the data must not allow mutability given winfsp may call multiple operations at the same time with the same FileContext)
- winfsp-wrs may call close while another read/write operation is still ongoing. This means the data under the FileContext should not be freed in the close operation (typically this means if the FileContext contains a pointer on data, those data should be under a reference counter so that they get freed only when the read/write operation finishes)
- we need conversion operation to turn a FileContext pointer into the concrete type used in the implementation

In practice what we should have is:
```rust
use std::sync::Arc;

trait FileContextKind {
    // Write the data into winfsp's `PVOID *PFileContext`
    fn write(self, out: *mut libc::c_void);

    // write_ptr|scalar/full_context should be used depending of
    // UmFileContextIsUserContext2 and UmFileContextIsFullContext
    fn write_ptr<T>(ptr: *const T, out: *mut libc::c_void) {
        todo!()
    }
    fn write_scalar(val: usize, out: *mut libc::c_void) {
        todo!()
    }
//    fn write_full_context(context: FSP_FSCTL_TRANSACT_FULL_CONTEXT, out: *mut libc::c_void) {
//        todo!()
//    }

    // Retreive the data from winfsp's `FileContext`
    // If the data are Arc, the refcount should be increased
    // This is called in winfsp-wrs before calling FileSystemContext::read/write etc.
    fn access(raw: usize) -> Self;
    // Retreive the data from winfsp's `FileContext`
    // If the data are Arc, the refcount shouldn't be increased so that
    // they are freed once all operations on it have finished.
    // This is called in winfsp-wrs before calling FileSystemContext::close
    fn access_for_close(raw: usize) -> Self;
}

trait FileSystemContext<T> {
    type FileContext: FileContextKind;
    fn open(&self) -> Result<Self::FileContext, usize>;
    // winfsp-wrs does a `unsafe { &*ptr }` to convert the `*mut T` into a `&mut T`
    // (this is safe given the pointer is only freed when passing it to close)
    fn read(&self, file_context: Self::FileContext) -> Result<usize, usize>;
    // close must give back ownership of the pointer given it must now be freed
    fn close(&self, file_context: Self::FileContext);
}

// Example of implementation using an Arc

enum FileOrDir {
 File,
 Dir,
}

impl FileContextKind for Arc<FileOrDir> {
    fn write(self, out: *mut libc::c_void) {
      Self::write_ptr(Arc::into_raw(self), out);
    }
    fn access(raw: usize) -> Self {
        Self::access_for_close(raw).clone()
    }
    fn access_for_close(raw: usize) -> Self {
        unsafe{ Arc::<FileOrDir>::from_raw(raw as *const FileOrDir) }
    }
}


// Example using a scalar

impl FileContextKind for usize {
    fn write(self, out: *mut libc::c_void) {
      Self::write_scalar(self, out);
    }
    fn access(raw: usize) -> Self {
        raw
    }
    fn access_for_close(raw: usize) -> Self {
        raw
    }
}

```

mmmarcos (commented):
@TimeEngineer Consider adding a `README.md` to document the crate. You should be able to include it in `src/lib.rs` (see https://github.com/Scille/parsec-cloud/pull/5058/files)

Cargo.toml
```diff
@@ -158,6 +161,8 @@ wasm-bindgen-futures = { version = "0.4.37", default-features = false }
 wasm-bindgen-test = { version = "0.3.37", default-features = false }
 wasm-bindgen = { version = "0.2.87", default-features = false }
 web-sys = { version = "0.3.64", default-features = false }
+winfsp_build = { version = "0.1.1", default-features = false }
```
touilleMan, line 164: would have been better to call it `winfsp_wrs_build`

(but you can leave it if it's too complicated to change the name now that a crate has been published)

libparsec/crates/platform_mountpoint/src/lib.rs
```diff
@@ -0,0 +1,100 @@
...
+pub(crate) use error::{MountpointError, MountpointResult};
+pub use memfs::MountpointManager;
+
+// TODO: remove me once ParsecEntryInfo is used
+#[allow(dead_code)]
+#[derive(Debug, Clone, Copy, PartialEq)]
+pub(crate) enum EntryInfoType {
+    Dir,
+    File,
+    ParsecEntryInfo,
```
touilleMan, line 24: Can you add a comment to explain what is this ?

libparsec/crates/platform_mountpoint/src/windows/mod.rs
```diff
@@ -0,0 +1,480 @@
...
+        .set_file_system_name(&mountpoint)
+        .set_prefix(u16cstr!(""));
+
+    let params = Params {
+        volume_params,
+        ..Default::default()
+    };
+
+    FileSystem::new(params, Some(&mountpoint), MountpointManager::default())
+        .expect("Failed to init FileSystem ")
```
touilleMan, line 479: The idea in platform crates is to return a `anyhow::Error` when something wrong occurred and the caller can't realistically be expecting it (e.g. in platform storage we return a `anyhow::Error` if the sqlite db fails)

libparsec/crates/platform_mountpoint/src/windows/mod.rs
```diff
@@ -0,0 +1,480 @@
...
+            return Err(STATUS_DIRECTORY_NOT_EMPTY);
+        }
+
+        Ok(())
+    }
+}
+
+pub fn mount(mountpoint: &Path) -> FileSystem<MountpointManager> {
+    let mountpoint =
+        U16CString::from_os_str(mountpoint.as_os_str()).expect("Unreachable, valid OsStr");
```
touilleMan, line 456: what if `mountpoint` contains utf8 characters ? utf8 and utf16 are not equivalent and hence require conversion between the two (which are not done here)

For instance, considering `0xFF` is not valid in utf8:
```python
>>> b"\xff".decode("utf8")
  File "<stdin>", line 1, in <module>
UnicodeDecodeError: 'utf-8' codec can't decode byte 0xff in position 0: invalid start byte
>>> b"\x00\xff".decode("utf16")
'\uff00'
>>> b"\x00\xff".decode("utf8")
  File "<stdin>", line 1, in <module>
UnicodeDecodeError: 'utf-8' codec can't decode byte 0xff in position 1: invalid start byte
```

So in this example we can see that the buffer `0x00 0xff` will work with `U16CString::from_os_str` but will cause an error with `U8CString::from_os_str`

So I guess you will have to do conversion between utf8 and utf16 here (but fortunatly it seems [Rust has you covered here ^^](https://doc.rust-lang.org/std/primitive.str.html#method.encode_utf16) )

libparsec/crates/platform_mountpoint/src/lib.rs
```diff
@@ -0,0 +1,100 @@
...
+    id: EntryID,
+    ty: EntryInfoType,
+    created: DateTime<Utc>,
+    updated: DateTime<Utc>,
+    size: u64,
+    need_sync: bool,
+    children: HashMap<EntryName, EntryID>,
+}
+
+pub(crate) trait MountpointInterface {
```
touilleMan, line 51: No longer needed right ? (it is only used in the memfs version)

libparsec/crates/platform_mountpoint/src/memfs/mod.rs
```diff
@@ -0,0 +1,439 @@
+// Parsec Cloud (https://parsec.cloud) Copyright (c) BUSL-1.1 2016-present Scille SAS
```
touilleMan, line 1: We don't need this here (it should be in winfsp-wrs as example instead)

On top of that this uses the `MountpointInterface` which is no longer needed, so better to remove this code entirely than try to fix it !

touilleMan (changes requested):

libparsec/crates/platform_mountpoint/src/windows/mod.rs
TimeEngineer: What ? o_O

Cargo.toml
TimeEngineer: We can't change the name, but yank

libparsec/crates/platform_mountpoint/src/lib.rs
TimeEngineer: Done

libparsec/crates/platform_mountpoint/src/lib.rs
touilleMan: plot twist: still needed !

libparsec/crates/platform_mountpoint/src/lib.rs
```diff
@@ -0,0 +1,103 @@
+// Parsec Cloud (https://parsec.cloud) Copyright (c) BUSL-1.1 2016-present Scille SAS
+
+#![doc = include_str!("../README.md")]
+
+mod error;
+mod memfs;
```
touilleMan, line 6: hide it behind a feature ? or move it into the tests folder ?

libparsec/crates/platform_mountpoint/src/windows/mod.rs
touilleMan: `mountpoint.as_os_str()` doesn't do any conversion (i.e. it only expose the underlying bytes buffer)

Similarly `U16CString::from_os_str` doesn't do any conversion (i.e. it takes the bytes buffer and use it as-is)

Also see  https://docs.rs/widestring/latest/x86_64-pc-windows-msvc/widestring/ustring/struct.U16String.html#method.from_os_str

> Since [OsStr](https://doc.rust-lang.org/nightly/std/ffi/os_str/struct.OsStr.html) makes no guarantees that it is valid data, there is no guarantee that the resulting [U16String](https://docs.rs/widestring/latest/x86_64-pc-windows-msvc/widestring/ustring/struct.U16String.html) will be valid UTF-16.

So what you are doing is equivalent to:
```c
char *my_uft8_string = "'I \xe2\x9d\xa4\xef\xb8\x8f Rust"; // UTF8 of "I ❤️ Rust"
char *my_utf16_string = my_utf8_string;
```
In this example it is trivial to understand `my_utf8_string` and `my_utf16_string` point to the exact same data, which cannot be both utf8 and utf16 !
Here, the "I love Rust" in UTF16 is `\xff\xfeI\x00 \x00d'\x0f\xfe \x00R\x00u\x00s\x00t\x00`, hence doing `my_utf16_string = my_utf8_string` is wrong because then `my_utf16_string` would not contains valid UTF16 data.
Instead we should do something like `my_utf16_string = utf8_to_utf16(my_utf8_string)`.

touilleMan: I think I'm wrong here: I was under the assumption that `Path` was always UTF8, but this is wrong: `Path` is a wrapper over `OSStr` and hence has an encoding which is dependent of the platform (hence UTF16 on Windows)

So doing `U16CString::from_os_str(mountpoint.as_os_str())` is always ok given the `OSStr` has been obtained from a `Path`

(on top of that `Path` has to deal with FFI so it cannot guarantee that it contains valid UTF16 data)

so sorry for the noise, everything is fine here ^^

libparsec/crates/platform_mountpoint/src/lib.rs
TimeEngineer: `MountpointManager` is used by `windows`, no other `MountpointManager` is implemented to replace it

libparsec/crates/platform_mountpoint/src/error.rs
FirelightFlagboy: I'm not sure to understand why `GetLastError` can't be called when creating `MountpointError` to get basic information (at least the path) about the error

Cargo.toml
TimeEngineer: which tag do you want ?

libparsec/crates/platform_mountpoint/src/error.rs
TimeEngineer: I can add the path in the error for debug purpose if you want

TimeEngineer: We can also enable logger to know in which context we are

Cargo.toml
touilleMan: I leave the decision up to you then ;-)

libparsec/crates/platform_mountpoint/src/error.rs
FirelightFlagboy: The user of this crate perform an action that result in an error, It would have partial information about it with the enum (currently only the cause) and to get additional information it would need to call `GetLastError` is it right ?

libparsec/crates/platform_mountpoint/src/windows/mod.rs
```diff
@@ -0,0 +1,481 @@
...
+use crate::{EntryInfo, EntryInfoType, MountpointInterface, MountpointManager, WriteMode};
+
+static SD: Lazy<SecurityDescriptor> = Lazy::new(|| {
+    SecurityDescriptor::from_wstr(u16cstr!("O:BAG:BAD:P(A;;FA;;;SY)(A;;FA;;;BA)(A;;FA;;;WD)"))
+        .expect("unreachable, valid SecurityDescriptor")
+});
+
+static VOLUME: OnceCell<Arc<Mutex<VolumeInfo>>> = OnceCell::new();
+
+const ENTRY_INFO_EXTENSION: &str = "__parsec_entry_info__";
```
touilleMan, line 32: Add a small comment to explain what this is for ?

libparsec/crates/platform_mountpoint/src/windows/mod.rs
```diff
@@ -0,0 +1,481 @@
...
+const MAX_FILE_NODES: u64 = 1024;
```
touilleMan, line 33: I don't understand those MAX_* constants...

1024 nodes means maximum 1024 files/folders ?
16 * 1024 * 1024 means maximum 16mo for a given file ?

are we back to the 1980's :trollface:  ?

libparsec/crates/platform_mountpoint/src/windows/mod.rs
```diff
@@ -0,0 +1,481 @@
...
+const MAX_FILE_SIZE: u64 = 16 * 1024 * 1024;
+const ALLOCATION_UNIT: u64 = 4096;
+const FILE_NODES: u64 = 1;
```
touilleMan, line 36: I also don't understand this... form it name it seems to be the current number of nodes, but it is a constant...

and is it only used in `get_volume_info` so I have the feeling this const is not needed at all

libparsec/crates/platform_mountpoint/src/windows/mod.rs
```diff
@@ -0,0 +1,481 @@
...
+    SecurityDescriptor::from_wstr(u16cstr!("O:BAG:BAD:P(A;;FA;;;SY)(A;;FA;;;BA)(A;;FA;;;WD)"))
+        .expect("unreachable, valid SecurityDescriptor")
+});
+
+static VOLUME: OnceCell<Arc<Mutex<VolumeInfo>>> = OnceCell::new();
+
+const ENTRY_INFO_EXTENSION: &str = "__parsec_entry_info__";
+const MAX_FILE_NODES: u64 = 1024;
+const MAX_FILE_SIZE: u64 = 16 * 1024 * 1024;
+const ALLOCATION_UNIT: u64 = 4096;
```
touilleMan, line 35: Add a small comment about why this value is used ? (I've seen this post https://github.com/winfsp/winfsp/issues/50#issuecomment-278764580 saying `written in increments of the NTFS allocation unit (usually 4096 on modern systems)` so I guess there is a reason for this value ^^)

libparsec/crates/platform_mountpoint/src/windows/mod.rs
```diff
@@ -0,0 +1,481 @@
...
+            EntryInfoType::File => *FileInfo::default()
+                .set_file_attributes(value.ty.into())
+                .set_creation_time(created)
+                .set_last_access_time(updated)
+                .set_last_write_time(updated)
+                .set_change_time(updated)
+                .set_index_number(value.id.as_u128() as u64)
+                .set_file_size(value.size)
+                .set_allocation_size(
+                    ((value.size + ALLOCATION_UNIT - 1) / ALLOCATION_UNIT) * ALLOCATION_UNIT,
```
touilleMan, line 60: From [winfsp tutorial](https://winfsp.dev/doc/WinFsp-Tutorial/):

> Files in a Windows file system can have two sizes: an "EndOfFile" size or FileSize and an AllocationSize. The FileSize is the number of bytes contained in a file. The AllocationSize is a concept that **many file systems can safely ignore** (or not expose to the kernel): it is the actual number of bytes that a file occupies on its storage medium.
> 
> [...]
> Regardless of whether a file system exposes AllocationSize it must obey the following rule: **it must always be that FileSize <= AllocationSize**.

In Parsec we never allocate more space for the file data, so this allocation size is meaningless.
I think we should just do `set_allocation_size(value.size)` and provide a comment stating 
```rust
// AllocationSize is the size actually occupied on the storage medium, this is
// meaningless for Parsec (however WinFSP requires FileSize <= AllocationSize)
```

libparsec/crates/platform_mountpoint/src/windows/mod.rs
```diff
@@ -0,0 +1,481 @@
...
+                .set_file_attributes(value.ty.into())
+                .set_creation_time(created)
+                .set_last_access_time(updated)
+                .set_last_write_time(updated)
+                .set_change_time(updated)
+                .set_index_number(value.id.as_u128() as u64),
+            EntryInfoType::ParsecEntryInfo => *FileInfo::default()
+                .set_file_attributes(value.ty.into())
+                // Arbitrary non-zero size
+                .set_file_size(1024)
```
touilleMan, line 72: Why not a zero size ? having an empty file is something very common :/

libparsec/crates/platform_mountpoint/src/windows/mod.rs
```diff
@@ -0,0 +1,481 @@
...
+    }
+}
+
+pub fn mount(mountpoint: &Path) -> anyhow::Result<FileSystem<MountpointManager>> {
+    let mountpoint =
+        U16CString::from_os_str(mountpoint.as_os_str()).expect("Unreachable, valid OsStr");
+    let mut volume_params = VolumeParams::default();
+
+    volume_params
+        .set_sector_size(512)
```
touilleMan, line 461: From this comment https://github.com/winfsp/winfsp/issues/240#issuecomment-518629301

> The sector size is what is set in VolumeParams via the [SectorSize](https://github.com/billziss-gh/winfsp/blob/v1.5B2/src/dotnet/FileSystemHost.cs#L84) property when the file system is first created. (And of course the "sector size" is misnamed for non-disk file systems, but it determines the granularity at which non-buffered I/O is done.)

I think it would be good to add this comment (and the link) here, and also explain how we've choosen `512` as value (I guess we have just taken what is in winfsp example as it is a good default, which is totally a fine explanation :+1: )

touilleMan (changes requested):

libparsec/crates/platform_mountpoint/src/windows/mod.rs
TimeEngineer: From `2.16` comment

libparsec/crates/platform_mountpoint/src/memfs/mod.rs
```diff
@@ -0,0 +1,437 @@
...
```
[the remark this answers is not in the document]
TimeEngineer: Done

libparsec/crates/platform_mountpoint/src/memfs/mod.rs
```diff
@@ -0,0 +1,437 @@
...
+
```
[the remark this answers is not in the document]
TimeEngineer: Done

libparsec/crates/platform_mountpoint/src/windows/mod.rs
TimeEngineer: Done

libparsec/crates/platform_mountpoint/src/windows/mod.rs
TimeEngineer: I removed and set `total_size` and `free_size` to `1TB` as it was

libparsec/crates/platform_mountpoint/src/windows/mod.rs
TimeEngineer: Done

libparsec/crates/platform_mountpoint/src/windows/mod.rs
TimeEngineer: Done

libparsec/crates/platform_mountpoint/src/windows/mod.rs
TimeEngineer: There is one test which will fail then in `winfsp_tests`

libparsec/crates/platform_mountpoint/src/windows/mod.rs
TimeEngineer: Done

libparsec/crates/platform_mountpoint/src/windows/mod.rs
TimeEngineer: nvm

libparsec/crates/platform_mountpoint/src/windows/mod.rs
```diff
@@ -0,0 +1,495 @@
...
+
+use winfsp_wrs::{
+    filetime_from_utc, filetime_now, u16cstr, CleanupFlags, CreateFileInfo, CreateOptions,
+    FileAccessRights, FileAttributes, FileInfo, FileSystem, FileSystemContext, PSecurityDescriptor,
+    Params, SecurityDescriptor, U16CStr, U16CString, VolumeInfo, VolumeParams,
+    WriteMode as WinWriteMode, NTSTATUS, STATUS_DIRECTORY_NOT_EMPTY, STATUS_NOT_IMPLEMENTED,
+    STATUS_RESOURCEMANAGER_READ_ONLY,
+};
+
+use libparsec_types::FileDescriptor;
```
touilleMan, line 21:
```suggestion
use libparsec_types::prelude::*;
```

and then Bam ! You also got `anyhow` exposed for free too ^^

libparsec/crates/platform_mountpoint/src/windows/mod.rs
touilleMan: Can you add a `// TODO: should we use 0 value instead ?`, we will then try to tweak this once the FS is working fine ^^

libparsec/crates/platform_mountpoint/build.rs
```diff
@@ -0,0 +1,6 @@
+// Parsec Cloud (https://parsec.cloud) Copyright (c) BUSL-1.1 2016-present Scille SAS
+
+fn main() {
+    #[cfg(target_os = "windows")]
```
touilleMan, line 4: Add comment:
```rust
// WinFSP is provided as a DLL that our main binary must loaded.
// By default Windows look for DLL in 1) the executable directory and 2) the Windows system folders.
// However this doesn't work for WinFSP given it is distributed as a separate program
// (and hence end up install in an arbitrary directory we must retrieve at the runtime
// by querying the Windows Registry) .
// Long story short, we are here here informing the linker the WinFSP DLL must be lazy
// loaded ("delayload" option in MSVC) so that we will have time to first configure the
// lookup directory.
```

(this comment is pretty sweet, you should also put it in https://github.com/Scille/winfsp_wrs/blob/main/winfsp_build/src/lib.rs ^^)

touilleMan (approved):

libparsec/crates/platform_mountpoint/build.rs
TimeEngineer: Done

libparsec/crates/platform_mountpoint/src/windows/mod.rs
TimeEngineer: Done

PR #5025, opened 2023-08-17
Bump rust version to 1.71.1

TimeEngineer:
## Checklist

- [ ] Describe your changes
   
If this PR is related to GUI:

If this PR affects the user:

### DIFF
--- a/bindings/generator/templates/binding_electron_meths.rs.j2
+++ b/bindings/generator/templates/binding_electron_meths.rs.j2
@@ -138,12 +138,16 @@ JsFunction
 {
     {% if type.custom_from_rs_bytes -%}
     let custom_from_rs_bytes = {{ type.custom_from_rs_bytes }};
+    #[allow(clippy::unnecessary_mut_passed)]
     match custom_from_rs_bytes({{ js_val }}.as_slice({{ mut_cx_ref }}))
     {%- else -%}
+    #[allow(clippy::unnecessary_mut_passed)]
     match {{ js_val }}.as_slice({{ mut_cx_ref }}).try_into()
     {%- endif %}
     {
         Ok(val) => val,
+        // err can't infer type in some case, because of the previous `try_into`
+        #[allow(clippy::useless_format)]
         Err(err) => return cx.throw_type_error(format!("{}", err)),
     }
 }
@@ -259,6 +263,7 @@ match {{ rs_value }} {
 }
 {%- elif type.kind == "none" -%}
 {
+    #[allow(clippy::let_unit_value)]
     let _ = {{ rs_value }};
     JsNull::new({{ mut_cx_ref }})
 }
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,7 +1,7 @@
 exclude: ^misc/bench.py$
 
 default_language_version:
-  rust: 1.70.0
+  rust: 1.71.1
   node: 18.12.0
 
 repos:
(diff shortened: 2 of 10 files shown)


FirelightFlagboy:
Maybe we should wait for 1.72.0 that should be release at the and of this month ?

TimeEngineer:
> Maybe we should wait for 1.72.0 that should be release at the end of this month ?

Maybe

FirelightFlagboy (approved):

bindings/generator/templates/binding_electron_meths.rs.j2
```diff
@@ -259,6 +265,7 @@ match {{ rs_value }} {
 }
 {%- elif type.kind == "none" -%}
 {
+    #[allow(clippy::let_unit_value)]
```
FirelightFlagboy, line 268: Missing fixme comment ?

TimeEngineer: There is no fix possible, or litterally a big change in jinja, we have to test before if the result is `None`

FirelightFlagboy (approved):
The code gen seems outdate with the comment you've added for `let_unit_value`

bindings/generator/templates/binding_electron_meths.rs.j2
```diff
@@ -259,6 +265,10 @@ match {{ rs_value }} {
 }
 {%- elif type.kind == "none" -%}
 {
+    // TODO: good luck to remove this
```
touilleMan, line 268: I don't think we want to fix this, clippy "Why is this bad?" says about it:

> A unit value cannot usefully be used anywhere. So binding one is kind of pointless.

In our case, the let binding totally has a point !

So I'd rather remove entirely your todo comment (I don't think an explanation comment is needed here, but feel free to add one if you want ^^)

bindings/generator/templates/binding_electron_meths.rs.j2
```diff
@@ -138,12 +138,18 @@ JsFunction
 {
     {% if type.custom_from_rs_bytes -%}
     let custom_from_rs_bytes = {{ type.custom_from_rs_bytes }};
+    // TODO: Some references don't need mutability
```
touilleMan, line 141: I wouldn't put those comments as TODO: fixing those will only make clippy happy, but won't improve the codebase in performance or readability (most likely it will worsen the readability !).

So adding a `#[allow(clippy::...)]` on a precise part of the code is the right approach here

bindings/generator/templates/binding_electron_meths.rs.j2
```diff
@@ -138,12 +138,18 @@ JsFunction
...
+    #[allow(clippy::unnecessary_mut_passed)]
     match custom_from_rs_bytes({{ js_val }}.as_slice({{ mut_cx_ref }}))
     {%- else -%}
+    // TODO: Some references don't need mutability
+    #[allow(clippy::unnecessary_mut_passed)]
     match {{ js_val }}.as_slice({{ mut_cx_ref }}).try_into()
     {%- endif %}
     {
         Ok(val) => val,
+        // TODO: err can't infer type in some case
```
touilleMan, line 151: Are you sure of the comment ?
I'd say the issue is something like: `format!("{}", err)` is useless in the case `err` is already a `String`
(I'd guess it's a bit more complex: `throw_type_error` want't a trait X, `format!()` returns a `String` that impl trait X, but this is useless if `err` already impl trait X)

touilleMan (changes requested):
No changes needed in `bindings_web_meths.rs.j2` ?

bindings/generator/templates/binding_electron_meths.rs.j2
TimeEngineer: The problem was an error (CryptoError) which can't infer type because of the previous `.try_into` which is needed for `.to_string`

touilleMan (approved):
:+1:

PR #5026, opened 2023-08-18
Add `wasm32-unknown-unknown` to the list of target for rust

FirelightFlagboy:
> Note that the host target is implicitly added


FirelightFlagboy:
https://github.com/Scille/parsec-cloud/actions/runs/5899906995/job/16003181022

TimeEngineer:
Why we don't add `windows`, `unix` and `darwin` ?

FirelightFlagboy:
> Why we don't add `windows`, `unix` and `darwin` ?

Read the description, if you need more information: https://rust-lang.github.io/rustup/overrides.html#targets

mmmarcos (approved):

touilleMan (approved):
Reading this, I had the same wondering than @TimeEngineer about why not adding windows, unix and darwin 

Can you add a comment such as `# Note native target is always implicit added` ?

PR #5044, opened 2023-08-21
Define cargo manifest values in the root cargo.toml

### LINKED ISSUE #5030: Add `publish = false` to our crates
This prevent publishing by mistake on crate.io

But the main reason for that, is it creates a clean distinction between our crates and the 3rd party ones.
For instance, `cargo deny` can be configured to ignore non published crates (and without this, it would fail given our crates are BUSL which is not allowed), see #5031

### LINKED ISSUE #5043: Rust: define package table to share configuration
https://doc.rust-lang.org/cargo/reference/workspaces.html#the-package-table

FirelightFlagboy:
Closes #5030, Closes #5043

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -36,6 +36,21 @@ default-members = [
     "libparsec/crates/types",
 ]
 
+# Define values to be inherited by workspace members.
+#
+# For a member to inherit a value, it need to set `{key}.workspace = true`.
+#
+# https://doc.rust-lang.org/cargo/reference/workspaces.html#the-package-table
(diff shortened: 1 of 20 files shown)


touilleMan (commented):
:+1:

touilleMan (approved):

mmmarcos (approved):
LGTM. 

I guess there is no way to avoid duplication of this block of code between `Cargo.toml` files? (something like an include or similar):
```
publish.workspace = true
authors.workspace = true
edition.workspace = true
...
```

PR #5052, opened 2023-08-22
Make bindings/generator mypy compatible

FirelightFlagboy:


bindings/generator/api/common.py
```diff
@@ -37,7 +41,7 @@ class VariantItemUnit:
 
 
 class VariantItemTuple:
-    def __init__(self, *items):
+    def __init__(self, *items: Any) -> None:
```
touilleMan, line 44: are you sure `__init__`  is required to have a return type defined ?

mmmarcos (approved):
LGTM

bindings/generator/api/common.py
FirelightFlagboy: Mypy don't seems to require the return type defined in that case, I've removed it.

bindings/generator/generate.py
```diff
@@ -348,10 +384,14 @@ class ParsingPlaceholder:
                         is_struct=True,
                         struct=StructSpec(
                             name=variant_val_name,
-                            attributes={
-                                k: BaseTypeInUse.parse(v)
-                                for k, v in getattr(variant_val_type, "__annotations__", {}).items()
-                            },
+                            attributes=OrderedDict(
```
touilleMan, line 387: Needed ? Regular dict are ordered since python 3.6

FirelightFlagboy: If I revert the `OrderedDict` like it was previously I get the following error:

```
bindings/generator/generate.py:387: error: Argument "attributes" to "StructSpec" has incompatible type "Dict[Any, BaseTypeInUse]"; expected "OrderedDict[str, BaseTypeInUse]"  [arg-type]
```

> The error specify `Dict[Any, BaseTypeInUse]` vs `OrderedDict[str, BaseTypeInUse]`.
> I've tried to make the key str with `str(k)` I still get a similar error:
>
> ```
> bindings/generator/generate.py:387: error: Argument "attributes" to "StructSpec" has incompatible type "Dict[str, BaseTypeInUse]"; expected "OrderedDict[str, BaseTypeInUse]
> ```

Maybe I should update the definition of `StructSpec` to use a `dict` instead of an `OrderedDict` ?

touilleMan: Ha ! I forgot `StructSpec` used an `OrderedDict` :/

I guess both approach are ok then:
- using `OrderedDict` stress the fact the attribute retain order 
- using `dict` makes is simpler in the code

I'd say let stick to `OrderedDict` to make it more explicit :nerd_face:

FirelightFlagboy: I will use `OrderedDict` with the simpler notation then :+1:

FirelightFlagboy: done

PR #5078, opened 2023-08-28
Rework libparsec_crates_flags to categorize crates by type

FirelightFlagboy:
Having a specific steps for each test suite improve the CI because it allow to set tiddier timeout and in case of failure provide a shorter stack trace.

### DIFF
--- a/.cspell/custom-words.txt
+++ b/.cspell/custom-words.txt
@@ -243,6 +243,7 @@ Nanos
 napi
 newsfragment
 newsfragments
+nextest
 Niño
 nmspc
 nocapture
--- a/.github/workflows/ci-rust.yml
+++ b/.github/workflows/ci-rust.yml
@@ -35,6 +35,8 @@ jobs:
     # testbed template is introduced).
     # In such case, the container url should be updated from the, see:
     # https://github.com/Scille/parsec-cloud/pkgs/container/parsec-cloud%2Fparsec-testbed-server
+    env:
+      TESTBED_SERVER: http://localhost:6777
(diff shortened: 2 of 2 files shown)


misc/libparsec_crates_flags.py
```diff
@@ -6,6 +6,11 @@
 if __name__ == "__main__":
     parser = argparse.ArgumentParser()
     parser.add_argument("what", nargs="+", choices=["all", "agnostic", "platform", "bindings"])
+    parser.add_argument(
```
touilleMan, line 9: Stay KISS, just call the script multiple time with a different argument instead

.cspell/cspell.config.yml
```diff
@@ -133,3 +133,5 @@ import:
   - "@cspell/dict-bash/cspell-ext.json"
   - "@cspell/dict-shell/cspell-ext.json"
   - "@cspell/dict-win32/cspell-ext.json"
+enableFiletypes:
```
touilleMan, line 136: Is this only needed for actions files ?

[the documentation uses json format as example](https://streetsidesoftware.com/vscode-spell-checker/docs/configuration/#cspellenablefiletypes), I don't understand why we didn't need it before (and why now we don't have to also set all the other types we want checked) :thinking:

FirelightFlagboy: I now known understand the culprit: 

it's `allowCompoundWords == true` with the current setting, it allow `accross` because it's a combination of `acc+ross` :exploding_head:

PR #5085, opened 2023-08-30
Refactor `ConnectionMonitor`

FirelightFlagboy:


libparsec/crates/client/src/connection_monitor.rs
```diff
@@ -153,6 +106,72 @@ impl ConnectionMonitor {
...
+        }
+        err => {
+            let event = EventIncompatibleServer(IncompatibleServerReason::Unexpected(err.into()));
+            event_bus.send(&event);
+            ControlFlow::Break(())
+        }
+    }
+}
+
+fn set_closed_state(state: &mut ConnectionState, event_bus: &EventBus) {
```
touilleMan, line 149: `set_closed_state` and `set_opened_state` are pretty small and used only in a single place, so I'd rather have them inlined instead

touilleMan (approved):

libparsec/crates/client/src/connection_monitor.rs
FirelightFlagboy: done

PR #5094, opened 2023-08-31
Refactor validation of testbed argument

FirelightFlagboy:


libparsec/crates/tests_macros/src/lib.rs
```diff
@@ -23,24 +28,8 @@ pub fn parsec_test(attr: TokenStream, item: TokenStream) -> TokenStream {
...
-                            Type::Path(p)
-                                if p.path.segments.last().expect("Incomplete path").ident
-                                    == "TestbedEnv" => {}
-                            _ => panic!("The last argument must be: `env: &TestbedEnv`"),
-                        },
-                        _ => panic!("The last argument must be: `env: &TestbedEnv`"),
-                    }
-                }
-                _ => panic!("Missing argument: `env: &TestbedEnv`"),
+            if let ControlFlow::Break(value) = validate_testbed_argument(&mut sig) {
```
touilleMan, line 31: Using a good'old `Result` seems more logical here since we stop because the there is errors !

touilleMan (changes requested):

libparsec/crates/tests_macros/src/lib.rs
```diff
@@ -97,6 +86,57 @@ pub fn parsec_test(attr: TokenStream, item: TokenStream) -> TokenStream {
...
 }
 
+fn validate_testbed_argument(sig: &mut syn::Signature) -> ControlFlow<TokenStream> {
+    match sig.inputs.pop().map(Pair::into_value) {
+        Some(FnArg::Typed(typed)) => {
+            validate_testbed_argument_name(&typed)?;
+            validate_testbed_argument_type(&typed)?;
+        }
+        _ => {
+            return ControlFlow::Break(compile_error(
```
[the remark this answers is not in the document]
FirelightFlagboy: The function `compile_error` provide better error message.

Instead of having the message `your macro has panic'ed` with the actual error message somewhere, that function make the actual error message to be the error message and provide the `span` (the piece of code that produced the message)

libparsec/crates/tests_macros/src/lib.rs
FirelightFlagboy: It was more logical for me to use ControlFlow to control the flow of the function

libparsec/crates/tests_macros/src/lib.rs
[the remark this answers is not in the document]
FirelightFlagboy: I've renamed it to `generate_compile_error`

libparsec/crates/tests_macros/src/lib.rs
```diff
@@ -1,8 +1,12 @@
 // Parsec Cloud (https://parsec.cloud) Copyright (c) BUSL-1.1 2016-present Scille SAS
 
 use proc_macro::TokenStream;
-use quote::quote;
-use syn::{parse_macro_input, punctuated::Pair, FnArg, ItemFn, LitStr, Pat, Type, TypeReference};
+use proc_macro2::Span;
```
touilleMan, line 4: Why not proc_macro  (https://doc.rust-lang.org/proc_macro/struct.Span.html) instead of proc_macro2 to stay consistent?

touilleMan (approved):

PR #5105, opened 2023-09-01
Add `get_os` for ionic bindings

FirelightFlagboy:


bindings/electron/src/meths.rs
```diff
@@ -5194,6 +5194,15 @@ fn test_drop_testbed(mut cx: FunctionContext) -> JsResult<JsPromise> {
...
 }
 
+// get_os
+fn get_os(mut cx: FunctionContext) -> JsResult<JsPromise> {
+    let ret = libparsec::get_os();
+    let js_ret = JsString::try_new(&mut cx, ret).or_throw(&mut cx)?;
+    let (deferred, promise) = cx.promise();
+    deferred.resolve(&mut cx, js_ret);
+    Ok(promise)
+}
```
FirelightFlagboy, line 5204: @touilleMan Generating a promise for that seems a little overkill

bindings/generator/api/os.py
```diff
@@ -0,0 +1,4 @@
+# Parsec Cloud (https://parsec.cloud) Copyright (c) BUSL-1.1 2016-present Scille SAS
+
+def get_os() -> str:
```
touilleMan, line 3: what about a variant instead of a string ?

libparsec/src/os.rs
```diff
@@ -0,0 +1,5 @@
+// Parsec Cloud (https://parsec.cloud) Copyright (c) BUSL-1.1 2016-present Scille SAS
+
+pub const fn get_os() -> &'static str {
+    std::env::consts::OS
```
touilleMan, line 4: if we want an enum instead, we can use conditional compilation `#[cfg(target_os = "macos")]`

touilleMan (changes requested):

bindings/generator/api/os.py
FirelightFlagboy: We can use a variant instead, would you prefer that ?

> By the way, I find it strange to ask that question when you request change at the same time.

bindings/electron/src/meths.rs
touilleMan: I'd rather avoid adding sync method support in the bindings in order to avoid the added complexity


Most methods in the API are async anyway, and the remaining ones should work fine in async (given everything is async in javascript, there is not colored function issue there)

bindings/generator/api/os.py
touilleMan: Yes I'd rather have a variant 👍

FirelightFlagboy: done

bindings/generator/api/os.py
```diff
@@ -0,0 +1,21 @@
+# Parsec Cloud (https://parsec.cloud) Copyright (c) BUSL-1.1 2016-present Scille SAS
+
+from .common import Variant
+
+
+class OSes(Variant):
```
touilleMan, line 6: `OS` ? Most of the time we don't call a variant plural (e.g. `UserProfile`, `HTTPError`)

touilleMan (approved):

bindings/generator/api/os.py
FirelightFlagboy: done

PR #5129, opened 2023-09-06
Force personal workspace to mount as letter on windows

Landeers:
## Describe your changes

The aim of this change is to deliver the personal workspace as quickly as possible, while avoiding the use of the Windows registry.
As the personal workspace is unique per user and per organization, there is no risk of reaching letter saturation on Window.


mmmarcos (approved):
LGTM

PR #5141, opened 2023-09-07
Provide typescript enum for basic rust enum (flat enum) 

FirelightFlagboy:
The conversion between `typescript enum <-> rust enum` is made using by strings.
The enums needed to be defined in `common.ts` because if it where
defined in `definitions.d.ts` we would not be able to use it as a value.

Other changes
-------------

- Use `type` instead of `str` for `BaseTypeInUse::parse`

### DIFF
--- a/bindings/generator/api/client.py
+++ b/bindings/generator/api/client.py
@@ -13,17 +13,18 @@ from .common import (
     Result,
     UserID,
     Variant,
-    VariantItemUnit,
+    Enum,
+    EnumItemUnit,
 )
 from .config import ClientConfig
 from .events import OnClientEventCallback
 
 
-class RealmRole(Variant):
-    Owner = VariantItemUnit
-    Manager = VariantItemUnit
-    Contributor = VariantItemUnit
-    Reader = VariantItemUnit
+class RealmRole(Enum):
+    Owner = EnumItemUnit
+    Manager = EnumItemUnit
+    Contributor = EnumItemUnit
+    Reader = EnumItemUnit
 
 
 class DeviceAccessStrategy(Variant):
--- a/bindings/generator/api/common.py
+++ b/bindings/generator/api/common.py
@@ -17,6 +17,14 @@ class Result(Generic[OK, ERR]):
     pass
 
 
+class Enum:
+    pass
+
+
+class EnumItemUnit:
+    pass
+
+
 # e.g.
 #
 #       enum Foo {
@@ -195,7 +203,7 @@ class InvitationToken(StrBasedType):
     )
 
 
-class UserProfile(Variant):
-    Admin = VariantItemUnit
-    Standard = VariantItemUnit
-    Outsider = VariantItemUnit
+class UserProfile(Enum):
+    Admin = EnumItemUnit
+    Standard = EnumItemUnit
+    Outsider = EnumItemUnit
--- a/bindings/generator/generate.py
+++ b/bindings/generator/generate.py
@@ -10,7 +10,7 @@ from importlib import import_module
 from inspect import isclass, iscoroutinefunction, isfunction, signature
 from pathlib import Path
 from types import ModuleType
-from typing import Any, List, Tuple, Union, get_args
+from typing import Any, List, Tuple, Union, Callable, get_args
 
 from jinja2 import Environment, FileSystemLoader, StrictUndefined
 
@@ -31,6 +31,7 @@ META_TYPES = [
     "ErrorVariant",
     "Structure",
     "OnClientEventCallback",
+    "Enum",
 ]
 
 
@@ -82,6 +83,10 @@ class OnClientEventCallback:
     ...
 
 
+class Enum:
+    ...
+
+
 env = Environment(
     loader=FileSystemLoader(BASEDIR / "templates"),
     autoescape=False,
@@ -93,7 +98,14 @@ env = Environment(
 )
 
 
-def snake_case_to_camel_case(s: str) -> str:
+def _test_change_case_function(input: str, expected: str, func: Callable[[str], str]) -> None:
+    camel_case = func(input)
+    assert (
+        camel_case == expected
+    ), f"expected `{expected}` but got `{camel_case}` (input = `{input}`, func = {func.__name__})"
+
+
+def snake_to_camel_case(s: str) -> str:
     camel = ""
     next_is_uppercase = False
     for c in s.lower():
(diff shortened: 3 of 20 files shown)


bindings/generator/generate.py
```diff
@@ -446,6 +488,13 @@ class ParsingPlaceholder:
             placeholder.__dict__ = struct.__dict__
             placeholder.__class__ = struct.__class__
             structs.append(placeholder)  # type: ignore[arg-type]
+        elif isclass(item) and issubclass(item, Enum):
+            placeholder = TYPES_DB[item]
+            annotations = getattr(item, "__annotations__", {})
+            enum = EnumInUse(member_names=item._member_names_, name=item.__name__)
+            placeholder.__dict__ = enum.__dict__
+            placeholder.__class__ = enum.__class__
```
touilleMan, line 496: `TYPES_DB`'s type should be updated to avoid lint error here

bindings/generator/generate.py
```diff
@@ -446,6 +488,13 @@ class ParsingPlaceholder:
...
+            enums.append(placeholder)
```
touilleMan, line 497: should add a `# type: ignore[arg-type]` here

bindings/generator/generate.py
```diff
@@ -513,12 +563,20 @@ def generate_client(api_specs: ApiSpecs) -> list[str]:
     return []
 
 
+def generate_client_common(api_specs: ApiSpecs) -> list[str]:
```
touilleMan, line 566: instead of having yet another .ts file, we can just rename `definition.d.ts` into `definition.ts`

This way we can put enum along with the typing stuff:
```ts
export const enum OS {
    Linux = 'OSLinux',
    MacOs = 'OSMacOs',
    Windows = 'OSWindows',
    Android = 'OSAndroid',
}

export type Result<T, E = Error> =
  | { ok: true; value: T }
  | { ok: false; error: E }

[...]
```

bindings/generator/templates/binding_electron_meths.rs.j2
```diff
@@ -479,7 +488,33 @@ fn {{ variant_rs_to_js_function_name(variant) }}<'a>(
 }
 {% endmacro -%}
 
+{%- macro enum_js_to_rs_function_name(type) %}enum_{{ type.name | pascal2snake }}_js_to_rs{% endmacro -%}
```
touilleMan, line 491: We currently use `type.name | lower` for `struct_js_to_rs_function_name/struct_rs_to_js_function_name/variant_js_to_rs_function_name/variant_rs_to_js_function_name`

If you really think it is worth it to add `pascal2snake` instead of just lowercasing, then we should also correct those functions

touilleMan (changes requested):
While you're here, can you also fix:

## 1)

Add a `export * from '@/plugins/libparsec/definitions';` in `client/src/plugins/libparsec/index.ts`, this way we can do `import { AvailableDevice } from '@/plugins/libparsec';` instead of `import { AvailableDevice } from '@/plugins/libparsec/definitions';`

## 2)

Add `OS.Web` choice instead of panicking when running on web.
Having `get_os` panicking when called on web is too error prone: it means a dedicated check must be run by the caller before every call :/

And if you're picky and don't consider web an OS, you can rename this os -> platform :trollface:

FirelightFlagboy:
1. I can't do `import * from ...`, but I can do `import * as foo from ...`, nevertheless It's not related to this PR, I've created the issue #5168
2. not related to this PR, I've created the issue #5167 
3. Not related to this PR, I've created the issue #5166

bindings/generator/api/client.py
```diff
@@ -13,17 +13,17 @@
...
+from enum import IntEnum
 
 
-class RealmRole(Variant):
-    Owner = VariantItemUnit
-    Manager = VariantItemUnit
-    Contributor = VariantItemUnit
-    Reader = VariantItemUnit
+class RealmRole(IntEnum):
+    Owner = 0
```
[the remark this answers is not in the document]
FirelightFlagboy: Using the alternative form:

```python
OS = Enum(
    "OS",
    ["Linux", "MacOs", "Windows", "Android"],
)
```

result in:

```
Variable "bindings.generator.api.os.OS" is not valid as a type
```

when `Enum` is defined like so:

```python
class Enum(enum.Enum):
  pass
```

Otherwise, directly using `enum.Enum` don't produce that error.

bindings/generator/generate.py
FirelightFlagboy: I don't understand, other types don't update `TYPES_DB` (struct, variant).
Can you provide an snippet of what is expected ?

bindings/generator/api/client.py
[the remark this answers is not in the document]
FirelightFlagboy: done

bindings/generator/api/client.py
```diff
@@ -13,17 +13,17 @@
     Result,
     UserID,
     Variant,
-    VariantItemUnit,
 )
 from .config import ClientConfig
 from .events import OnClientEventCallback
+from enum import IntEnum
```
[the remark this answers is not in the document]
FirelightFlagboy: done

bindings/generator/generate.py
FirelightFlagboy: done

bindings/generator/templates/binding_electron_meths.rs.j2
FirelightFlagboy: done

bindings/generator/generate.py

bindings/generator/api/common.py
```diff
@@ -17,6 +17,14 @@ class Result(Generic[OK, ERR]):
     pass
 
 
+class Enum:
+    pass
+
+
+class EnumUnit:
```
touilleMan, line 24:
```suggestion
class EnumItemUnit:
```

(to be consistent with `VariantItemUnit`)

bindings/generator/templates/client_plugin_definitions.ts.j2
```diff
@@ -32,6 +32,8 @@ string
 Uint8Array
 {%- elif t.kind == "OnClientEventCallback" -%}
 (event: ClientEvent) => void
+{% elif t.kind == "enum"%}
```
touilleMan, line 35:
```suggestion
{%- elif t.kind == "enum" -%}
```

the `-` are needed to strip the newlines (otherwise the `,` end up on the next line)

bindings/generator/generate.py
touilleMan: I was refering to `TYPES_DB: dict[type, OpaqueSpec | StructSpec | VariantSpec | EnumSpec]` which was missing `EnumSpec`, this is all good now ;-)

touilleMan (approved):

bindings/generator/api/common.py
FirelightFlagboy: done

PR #5164, opened 2023-09-11
Use function to generate `generate_{type}` function

FirelightFlagboy:


bindings/generator/generate.py
```diff
@@ -505,42 +505,44 @@ class ParsingPlaceholder:
...
+        # Don't use `write_text` given it outputs \r\n for newlines on Windows
+        dest.write_bytes(_template.render(api=api_specs).encode("utf8"))
+        return modified_crate
 
-def generate_electron(api_specs: ApiSpecs) -> list[str]:
-    template = env.get_template("binding_electron_meths.rs.j2")
-    output = (BASEDIR / "../electron/src/meths.rs").resolve()
-    print(f"Generating {output}")
-    output.write_bytes(template.render(api=api_specs).encode("utf8"))
+    return _generate
```
mmmarcos, line 520: IMO this is overcomplex and makes the code less easy to understand. Always prioritize readability first.

If the goal is to avoid duplication, why not simply abstract the common code into a function receiving the necessary parameters?

```python
generate_target(api_specs, template, dest, modified_crates):
    ...
```

bindings/generator/generate.py
```diff
@@ -505,42 +505,44 @@ class ParsingPlaceholder:
...
+generate_electron_client = gen_generate_fn(
+    template="binding_electron_index.d.ts.j2", dest=BASEDIR / "../electron/src/index.d.ts"
+)
 
-    return ["libparsec_bindings_web"]
+generate_web = gen_generate_fn(
+    template="binding_web_meths.rs.j2",
+    dest=BASEDIR / "../web/src/meths.rs",
+    modified_crate="libparsec_bindings_web",
+)
```
mmmarcos, line 542: By using a `generate_target` function, these variables should not be needed anymore.

bindings/generator/generate.py
```diff
@@ -505,42 +505,44 @@ class ParsingPlaceholder:
...
 
 
-def generate(what: str, api_specs: ApiSpecs) -> list[str]:
+def generate(what: str, api_specs: ApiSpecs) -> str | None:
```
mmmarcos, line 545: In this function, we can replace the `generate_*` function calls with `generate_target` function. Eventually, the concrete arguments (`client_plugin_definitions.d.ts.j2`, and so on) can be defined somewhere in the file as constants.

bindings/generator/generate.py
```diff
@@ -580,15 +582,17 @@ def generate(what: str, api_specs: ApiSpecs) -> list[str]:
...
+    rust_modified_crates = []
 
     if "all" in args.what:
         args.what = TEMPLATE_CHOICES
 
     for what in args.what:
-        rust_modified_packages.extend(generate(what, api_specs))
+        possible_modified_crate = generate(what, api_specs)
+        if possible_modified_crate:
+            rust_modified_crates.append(possible_modified_crate)
```
mmmarcos, line 593: Why the change here? the `extend` method already manages empty lists

mmmarcos (changes requested):

bindings/generator/generate.py
FirelightFlagboy: Yes but `generate` return an `str | None` not a `list[str]`

bindings/generator/generate.py

bindings/generator/generate.py
FirelightFlagboy: done

FirelightFlagboy: The constant value is planned for a later PR (next to this one)

bindings/generator/generate.py

mmmarcos (approved):
:+1:

Max-7 (approved):

PR #5174, opened 2023-09-11
Merge `TEMPLATE_CHOICES` with `generate` dispatch

FirelightFlagboy:


bindings/generator/generate.py
```diff
@@ -572,54 +572,38 @@ def generate_target(
...
+    ),
+    "web": lambda api_spec: generate_target(
+        api_spec,
+        template="binding_web_meths.rs.j2",
+        dest=BASEDIR / "../web/src/meths.rs",
+        modified_crate="libparsec_bindings_web",
+    ),
+    # TODO: Android isn't ready yet!
+    # "android": lambda _spec: raise NotImplementedError("Android isn't ready yet!")
+}
```
mmmarcos, line 600: The original code seems already fine for me, I do not think this change adds much readability.

If the goal is to avoid the multiple calls to `generate_target` it should be sufficient to move the function call to the bottom of the `generate` function. Something like:
```python
def generate(what: str, api_specs: ApiSpecs) -> str | None:
    if what == "client":
        template="client_plugin_definitions.ts.j2",
        dest=BASEDIR / "../../client/src/plugins/libparsec/definitions.ts",
        modified_crate = ""
    elif what == "electron":
        template="binding_electron_meths.rs.j2",
        dest=BASEDIR / "../electron/src/meths.rs",
        modified_crate="libparsec_bindings_electron",
    elif what == "electron_client":
        template="binding_electron_index.d.ts.j2",
        dest=BASEDIR / "../electron/src/index.d.ts",
        modified_crate = ""
    elif what == "web":
        template="binding_web_meths.rs.j2",
        dest=BASEDIR / "../web/src/meths.rs",
        modified_crate="libparsec_bindings_web",
    elif what == "android":
        # TODO !
        raise NotImplementedError("Android target isn't ready yet")
    else:
        raise ValueError(f"Unknown target `{what}`")
    
    return generate_target(api_specs, template, dest, modified_crate)
```
If the goal is to avoid the `if`/`elif`/`else`, the values (template, dest, ...) could also be stored with a dataclass (e.g. TargetConf) inside a dictionary. But again, I don't think the change is really needed.

FirelightFlagboy: The goal is to limit the `if / else` but also to manage both `generate` & `TEMPLATE_CHOICES`

With that change we don't have to maintain the list of key at 2 places.

touilleMan: I agree with @mmmarcos , putting lambda in a global dict is less readable than the simple if/elif/else based existing solution.
If we really want to avoid errors due to strings used as test value, we should instead define a couple of global constants:
```python
# Or use Enum
TEMPLATE_CLIENT = "client"
TEMPLATE_ELECTRON = "electron"
TEMPLATE_ELECTRON_CLIENT = "electron_client
TEMPLATE_WEB = "web"
TEMPLATE_CHOICES =(
  TEMPLATE_CLIENT,
  TEMPLATE_ELECTRON,
  TEMPLATE_ELECTRON_CLIENT,
  TEMPLATE_WEB,
)
```

FirelightFlagboy:
```suggestion
TEMPLATE_CHOICES = {
    "client": partial(
        generate_target,
        template="client_plugin_definitions.ts.j2",
        dest=BASEDIR / "../../client/src/plugins/libparsec/definitions.ts",
    ),
    "electron": partial(
        generate_target,
        template="binding_electron_meths.rs.j2",
        dest=BASEDIR / "../electron/src/meths.rs",
        modified_crate="libparsec_bindings_electron",
    ),
    "electron_client": partial(
        generate_target,
        template="binding_electron_index.d.ts.j2",
        dest=BASEDIR / "../electron/src/index.d.ts",
    ),
    "web": partial(
        generate_target,
        template="binding_web_meths.rs.j2",
        dest=BASEDIR / "../web/src/meths.rs",
        modified_crate="libparsec_bindings_web",
    ),
    # TODO: Android isn't ready yet!
    # "android": lambda _spec: raise NotImplementedError("Android isn't ready yet!")
}
```

What about using `partial` ?

touilleMan: still less readable, just go the KISS way ;-)

PR #5188, opened 2023-09-12
Fix invalid use of `matrix.os`

FirelightFlagboy:


.github/workflows/ci-python.yml
```diff
@@ -102,7 +102,7 @@ jobs:
       # Hence if we have a cache hit we know that there is no need for a rebuild !
       - name: Setup cache-key
         id: cache-key
-        run: echo "key=${{ matrix.os }}-${{ hashFiles('server/src/**', 'server/Cargo.toml', 'libparsec/**', 'rust-toolchain.toml', 'Cargo.toml', 'Cargo.lock') }}-libparsec-python-${{ env.python-version }}" >> $GITHUB_OUTPUT
+        run: echo "key=ubuntu-20.04-${{ hashFiles('server/src/**', 'server/Cargo.toml', 'libparsec/**', 'rust-toolchain.toml', 'Cargo.toml', 'Cargo.lock') }}-libparsec-python-${{ env.python-version }}" >> $GITHUB_OUTPUT
```
mmmarcos, line 105: Isn't `matrix.os` also used in other workflows? Why it is invalid here?

FirelightFlagboy: Simply because we don't define a matrix (`strategy.matrix`) in this job

https://docs.github.com/en/actions/using-jobs/using-a-matrix-for-your-jobs

mmmarcos (approved):

PR #5195, opened 2023-09-13
Format python code in generator

FirelightFlagboy:


mmmarcos (approved):
:+1:

bindings/generator/generate.py
```diff
@@ -139,7 +139,7 @@ def peek_next(index: int) -> str:
     for i, c in enumerate(s):
         if c.isupper():
             next_lower = peek_next(i).islower()
-            if next_lower or previous_lower == True:
+            if next_lower or previous_lower is True:
```
touilleMan, line 142:
```suggestion
            if next_lower or previous_lower:
```

`next_lower` and `previous_lower` are both boolean, so it's odd to have different comparison methods (an it is more pythonic to not have the `is True` part if our variable don't have to deal with `None`)

touilleMan (approved):

PR #5197, opened 2023-09-13
Run `wasm-pack test` only when rust platform crates are modified

FirelightFlagboy:


.github/workflows/ci-rust.yml
```diff
@@ -2,7 +2,19 @@ name: CI Rust
 
 on:
   workflow_call:
+    inputs:
+      run-wasm-tests:
+        description: Run wasm test (mostly in platform crates)
```
mmmarcos, line 7: I think the "mostly" comment is not needed because we are describing the input, which is either run wasm tests or not.

.github/workflows/ci-rust.yml
```diff
@@ -2,7 +2,19 @@ name: CI Rust
...
+        required: true
+        default: false
+        type: boolean
   workflow_dispatch:
+    inputs:
+      run-wasm-tests:
+        description: Run wasm test (mostly in platform crates)
```
mmmarcos, line 14: Idem

mmmarcos (approved):
LGTM

.github/workflows/ci-rust.yml
FirelightFlagboy: Done

.github/workflows/ci-rust.yml
FirelightFlagboy: Done

PR #5202, opened 2023-09-13
Ms notification modals

### LINKED ISSUE #4675: Create Notif/Alert/Error Modal

NicoTuxx:
closes #4675 

## Checklist

- [ ] Describe your changes
   
If this PR is related to GUI:

If this PR affects the user:

### DIFF
--- a/client/src/components/core/ms-modal/MsModal.vue
+++ b/client/src/components/core/ms-modal/MsModal.vue
@@ -1,81 +1,111 @@
 <!-- Parsec Cloud (https://parsec.cloud) Copyright (c) BUSL-1.1 2016-present Scille SAS -->
 
 <template>
-  <ion-page class="modal">
-    <ion-buttons
-      slot="end"
-      class="closeBtn-container"
-      v-if="closeButtonEnabled"
+  <ion-page
+    class="modal"
+    :class="theme"
+  >
+    <div
+      @keyup.enter="$emit('onEnterKeyup')"
+      tabindex="0"
+      ref="modal"
     >
-      <ion-button
-        slot="icon-only"
-        @click="cancel()"
-        class="closeBtn"
+      <ion-buttons
+        slot="end"
+        class="closeBtn-container"
+        v-if="closeButtonEnabled"
       >
-        <ion-icon
-          :icon="close"
-          size="large"
-          class="closeBtn__icon"
-        />
-      </ion-button>
-    </ion-buttons>
-    <div class="ms-modal">
-      <ion-header class="ms-modal-header">
-        <ion-title class="ms-modal-header__title title-h2">
-          {{ title }}
-        </ion-title>
-        <ion-text
-          class="ms-modal-header__text body"
-          v-if="subtitle"
+        <ion-button
+          slot="icon-only"
+          @click="cancel()"
+          class="closeBtn"
         >
-          {{ subtitle }}
-        </ion-text>
-      </ion-header>
-      <div
-        class="ms-modal-content inner-content"
-      >
-        <slot />
-      </div>
-      <ion-footer
-        class="ms-modal-footer"
-        v-if="cancelButton || confirmButton"
-      >
-        <ion-toolbar class="ms-modal-toolbar">
-          <ion-buttons
-            slot="primary"
-            class="ms-modal-footer-buttons"
+          <ion-icon
+            :icon="close"
+            size="large"
+            class="closeBtn__icon"
+          />
+        </ion-button>
+      </ion-buttons>
+      <div class="ms-modal">
+        <ion-header class="ms-modal-header">
+          <div
+            class="ms-modal-header__title-container"
+            v-if="title"
           >
-            <ion-button
-              v-if="cancelButton"
-              fill="clear"
-              size="default"
-              id="cancel-button"
-              @click="cancelButton.onClick ? cancelButton.onClick() : cancel()"
-              :disabled="cancelButton.disabled"
+            <ion-icon
+              v-if="theme || titleIcon"
+              :icon="getTitleIcon()"
+              size="large"
+              class="ms-modal-header__title-icon"
+            />
+            <ion-title class="ms-modal-header__title title-h2">
+              {{ title }}
+            </ion-title>
+          </div>
+          <template v-if="subtitle">
+            <ms-informative-text
+              v-if="theme"
+              :theme="theme ?? MsTheme.Basic"
             >
-              {{ cancelButton.label }}
-            </ion-button>
-            <ion-button
-              v-if="confirmButton"
-              fill="solid"
-              size="default"
-              id="next-button"
-              type="submit"
-              @click="confirmButton.onClick ? confirmButton.onClick() : confirm()"
-              :disabled="confirmButton.disabled"
+              {{ subtitle }}
+            </ms-informative-text>
+            <ion-text
+              class="ms-alert-modal-header__text body"
+              v-else
             >
-              {{ confirmButton.label }}
-            </ion-button>
-          </ion-buttons>
-        </ion-toolbar>
-      </ion-footer>
+              {{ subtitle }}
+            </ion-text>
+          </template>
+        </ion-header>
+        <div
+          class="ms-modal-content inner-content"
+        >
+          <slot />
+        </div>
+        <ion-footer
+          class="ms-modal-footer"
+          v-if="cancelButton || confirmButton"
+        >
+          <ion-toolbar class="ms-modal-toolbar">
+            <ion-buttons
+              slot="primary"
+              class="ms-modal-footer-buttons"
+            >
+              <ion-button
+                v-if="cancelButton"
+                fill="clear"
+                size="default"
+                id="cancel-button"
+                @click="cancelButton.onClick ? cancelButton.onClick() : cancel()"
+                :disabled="cancelButton.disabled"
+              >
+                {{ cancelButton.label }}
+              </ion-button>
+              <ion-button
+                v-if="confirmButton"
+                fill="solid"
+                size="default"
+                id="next-button"
+                type="submit"
+                @click="confirmButton.onClick ? confirmButton.onClick() : confirm()"
+                :disabled="confirmButton.disabled"
+              >
+                {{ confirmButton.label }}
+              </ion-button>
+            </ion-buttons>
+          </ion-toolbar>
+        </ion-footer>
+      </div>
     </div>
   </ion-page>
 </template>
 
 <script lang="ts">
 export interface MsModalConfig {
-  title: string,
+  title?: string,
+  titleIcon?: string,
+  theme?: MsTheme,
   subtitle?: string,
   closeButtonEnabled: boolean,
   cancelButton?: {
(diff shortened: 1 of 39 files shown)


client/src/components/core/ms-modal/MsModal.vue
```diff
@@ -74,8 +82,17 @@
...
+  Success = 'modal-success',
+  Warning = 'modal-warning',
+  Error = 'modal-error',
+}
+
 export interface MsModalConfig {
-  title: string,
+  title?: string,
+  titleIcon?: string,
+  theme?: MsModalTheme,
```
Max-7, line 95: I feel like `theme` and `titleIcon` are too similar and clash with one another a bit (like you can have a warning icon with a success theme). I would be in favor of dropping the icon, making the theme mandatory, and letting the component choose what icon it wants to display depending on the theme. It gives us a more reliable API, which means less options for devs like me to screw up the UI. Of course it also means that you could not customise the modal, but I think it's a good thing to be more restrictive.

client/src/locales/fr-FR.json
```diff
@@ -52,6 +52,10 @@
             "joinByLink": "Rejoindre par lien",
             "joinByQRcode": "Rejoindre par QR code"
         },
+        "loginNotification": {
+            "title": "Connexion échouée !",
```
Max-7, line 56: I would change that by `Impossible de se connecter`. It feels less like a straight-up translation from English.

client/src/components/core/ms-modal/MsModal.vue
Max-7, line 1: Why put that in MsModal? The notification modals should be a specialized component using ms-modal instead.

client/src/services/notificationCenter.ts
```diff
@@ -10,84 +14,79 @@ export enum NotificationLevel {
...
-    const notif: Notification = {
-      id: uuid4(),
-      message: message,
-      level: level,
-      read: false,
-      time: DateTime.now(),
-      data: data,
-    };
-    if (trace) {
+  async showSnackbar(notif: Notification): Promise<void> {
```
Max-7, line 70: Keep a `addToList` option. Maybe I want my notification to create a snack and to be in the list of notifications as well. For example, something important like `Hey you joined a new organization!` could be in both. Granted, I could call `showSnackbar()` and then `addToList()` but it means that I cannot create the notification in place.
IE I would have to do
`const notif = new Notification(); showSnackbar(notif); addToList(notif);`
instead of
`showSnackBar(new Notification(), addToList=true);`

Also, there's probably very few cases where you would need to call `addToList` directly, because it would mean not warning the user directly that something as happened, so the prefered way to add a notif to the list should be with `showSnackbar`.

client/src/services/notificationCenter.ts
```diff
@@ -10,84 +14,79 @@ export enum NotificationLevel {
...
-      title: title,
-      message: message,
-      level: level,
-      read: false,
-      time: DateTime.now(),
-      data: data,
-    };
-    if (trace) {
-      this.trace(notif);
+  getModalTheme(notificationLevel: NotificationLevel): MsModalTheme {
```
Max-7, line 76: Is there a way to make methods private or hidden? I usually just put an `_` at the start of its name like in Python. It's just to have an indication that you shouldn't call this method outside of this class (or at least that there's no point in doing so).

client/src/views/home/HomePage.vue
```diff
@@ -352,7 +355,14 @@ async function login(device: AvailableDevice, password: string): Promise<void> {
   if (result.ok) {
     router.push({ name: 'workspaces', params: {handle: result.value}});
   } else {
-    console.log('Could not login', result.error);
+    const notif = new Notification(
```
Max-7, line 358: Is there a way to have default values for arguments? Or named arguments?
I have no idea what `false` and `{}` do in that function call. I would like to be able to do either
`new Notification('message', 'title', NotificationLevel.Error)`
and/or
`new Notification('message', title='title', level=NotificationLevel.Error)`

Max-7 (changes requested):
Can't comment a line that has not been modified so I'm commenting here:
https://github.com/Scille/parsec-cloud/pull/5202/files#diff-a5b18019b56c603e62103b01318a1d2596a4b760e0369e245f2684bdbd1bdb19R352 this should be moved when the login has succeeded, just before the `router.push('workspaces')`, so the page does not go back to the device list when an incorrect password has been entered.

And two more things:
- would it be possible to close the modal with `Enter`? It is possible with `Escape` but I feel like `Enter` should also count.
- is it possible to put the focus back on the password input when the modal is closed?

client/src/services/notificationCenter.ts
NicoTuxx: Yes, by writing private before the function name. the '_' prefix is also commin in JS

client/src/views/home/HomePage.vue
NicoTuxx: Add default values for data and trace

client/src/services/notificationCenter.ts
NicoTuxx: Move trace as a parameter of showSnackbar and showModal

client/src/components/core/ms-modal/MsModal.vue
NicoTuxx: Done, some code could be common between the 2 files, but the same applies to the CSS.
To keep focus of the issue, I'll keep it simple and make 2 files with their code each.

Maybe it's even better to keep them totally distincts (MsModal and MsAlertModal)

client/src/services/notificationCenter.ts
```diff
@@ -10,92 +14,123 @@ export enum NotificationLevel {
...
+  title?: string;
+  constructor({
+    message,
+    level,
+    data,
+    title,
+  }: {
+    message: string,
+    level: NotificationLevel,
+    data: object,
```
Max-7, line 33: `data` should be optional (or have a default value of an empty objet). In the vast majority of cases it won't be used, I just added it in case we have specific information to provide to whatever is using the notification.

client/src/services/notificationCenter.ts
```diff
@@ -10,92 +14,123 @@ export enum NotificationLevel {
...
+      this._trace(notification);
     }
-    if (addToList) {
-      this.notifications.push(notif);
+
+    const alertModalConfig: MsAlertModalConfig = {
+      theme: this._getModalTheme(notification.level),
+      subtitle: notification.message,
+      onConfirm: () => {
+        this.markAsRead(notification.id);
```
Max-7, line 74: See my remark in MsAlertModal about onConfirm, it would mean that the dismissing is not done here, which is a good thing, it should be left to the modal to decide. It would also mean that the behavior stays consistent between having an `onConfirm` parameter and not having one.

client/src/views/home/DeviceJoinOrganizationModal.vue
```diff
@@ -89,9 +89,10 @@
           class="step"
         >
           <ms-informative-text
-            :icon="checkmarkCircle"
-            :text="$t('ClaimDeviceModal.subtitles.done')"
-          />
+            :theme="MsInformativeTextTheme.Success"
```
Max-7, line 92: Are we sure about that? It's not what is planned in the mock-ups.

client/tests/component/specs/testMsChoosePasswordInput.spec.ts
```diff
@@ -19,14 +19,14 @@ describe('Choose password', () => {
     expect(await wrapper.vm.areFieldsCorrect()).to.be.false;
 
     const ionInputs = wrapper.findAllComponents(IonInput);
-    ionInputs[0].vm.$emit('ionInput', {target: {value: 'P@ssw0rd'}});
-    expect(wrapper.vm.password).to.equal('P@ssw0rd');
+    ionInputs[0].vm.$emit('ionInput', {target: {value: 'P@ssw0rd.'}});
```
Max-7, line 22: Useless updates, we're just testing the component here.

client/tests/e2e/specs/test_home_page.ts
```diff
@@ -30,12 +30,12 @@ describe('Check organization list', () => {
...
+    cy.get('#ms-password-input').find('input').type('P@ssw0rd.');
     cy.get('.login-button-container > ion-button').should('not.have.class', 'button-disabled');
     cy.get('.login-button-container > ion-button').click();
     cy.get('@configPath').then(($elem) => {
       const orgId = ($elem as unknown as string).split('/').slice(-1)[0];
-      cy.get('@consoleLog').should('have.been.calledWith', `Log in to ${orgId} with password "P@ssw0rd"`);
+      cy.get('@consoleLog').should('have.been.calledWith', `Log in to ${orgId} with password 'P@ssw0rd.'`);
     });
     cy.contains('My workspaces');
   });
```
Max-7, line 41: We should add a test to check what happens when we use the wrong password.

Max-7 (changes requested):

client/src/components/core/ms-text/MsInformativeText.vue
```diff
@@ -1,44 +1,110 @@
...
       class="subtitles-normal container-textinfo__text"
     >
-      {{ text }}
+      <slot />
     </ion-text>
   </div>
 </template>
 
+<script lang="ts">
+export enum MsInformativeTextTheme {
```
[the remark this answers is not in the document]
NicoTuxx: done with ms-types

client/src/views/home/UserJoinOrganizationModal.vue
```diff
@@ -343,12 +340,12 @@ async function nextStep(): Promise<void> {
...
-    notificationCenter.showSnackbar(
-      t('JoinOrganization.successMessage'),
-      NotificationLevel.Success,
-      false,
-      true,
-    );
+    const notification = new Notification({
+      message: t('JoinOrganization.successMessage'),
+      level: NotificationLevel.Success,
+      data: {},
```
Max-7, line 346: `data` can be removed here.

client/tests/e2e/specs/test_home_page.ts
```diff
@@ -26,16 +26,30 @@ describe('Check organization list', () => {
...
+    cy.get('.login-button-container > ion-button').should('have.class', 'button-disabled');
+    cy.get('#ms-password-input').find('input').invoke('attr', 'type').should('eq', 'password');
+    cy.get('#ms-password-input').find('input').type('Wr0ngP@ssw0rd.');
+    cy.get('.login-button-container > ion-button').should('not.have.class', 'button-disabled');
+    cy.get('.login-button-container > ion-button').click();
+    cy.get('@configPath').then(($elem) => {
+      const orgId = ($elem as unknown as string).split('/').slice(-1)[0];
+      cy.get('@consoleLog').should('have.been.calledWith', `Log in to ${orgId} with password 'Wr0ngP@ssw0rd.'`);
+    });
+    cy.contains('Could not login!');
```
Max-7, line 40: Should probably use a stricter condition: is the modal opened, is it danger-themed, ... I think `cy.contains` checks if it's in the DOM, not if it's visible to the user.
If it's going to be replaced by a toast then nevermind for now.

Max-7 (approved):

client/tests/e2e/specs/test_home_page.ts
NicoTuxx: it will be replaced yeah

PR #5214, opened 2023-09-15
Ms notification toasts

NicoTuxx:
## Checklist

- [ ] Describe your changes
   
If this PR is related to GUI:

If this PR affects the user:

### DIFF
--- a/client/src/services/notificationCenter.ts
+++ b/client/src/services/notificationCenter.ts
@@ -3,9 +3,15 @@
 import { DateTime } from 'luxon';
 import { v4 as uuid4 } from 'uuid';
 import MsAlertModal, { MsAlertModalConfig } from '@/components/core/ms-modal/MsAlertModal.vue';
-import { modalController } from '@ionic/vue';
+import { modalController, toastController } from '@ionic/vue';
 import { ComposerTranslation } from 'vue-i18n';
 import { MsModalResult, MsTheme } from '@/components/core/ms-types';
+import {
+  informationCircle,
+  checkmarkCircle,
+  warning,
+  closeCircle,
+} from 'ionicons/icons';
 
 export enum NotificationLevel {
   Info = 'INFO',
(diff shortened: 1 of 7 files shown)


client/src/services/notificationCenter.ts
```diff
@@ -43,6 +49,15 @@ export class Notification {
   }
 }
 
+export interface NotificationToastConfig {
```
Max-7, line 52: Not much point to this interface. It seems it's only used once when calling `_createAndPresentToast`, might as well just pass arguments to the function, no need for an interface.

client/src/services/notificationCenter.ts
```diff
@@ -43,6 +49,15 @@ export class Notification {
...
+  title?: string,
+  icon: string,
+  message: string,
+  theme: MsTheme,
+}
+
+export const DEFAULT_NOTIFICATION_DURATION = 10000;
```
Max-7, line 59: Shouldn't be exported.

client/src/services/notificationCenter.ts
```diff
@@ -92,7 +107,7 @@ export class NotificationCenter {
     return modal.onWillDismiss();
   }
 
-  async showSnackbar({
+  async showToast({
```
Max-7, line 110: I'm guessing the `{}` are there to be able to set `addToList` or `trace` to true, like `showToast({notification, trace: true})` instead of doing `showToast(notification, false, true)` because typescript does not have real named arguments so you can't just do `showToast(notification, trace=true)`. Is that it?

Not being able to just do `showToast(notification)` kinda sucks, wouldn't it be better to have
`showToast(notification, {addToList = false, trace = false}`? I'm guessing this would work (not sure but I don't see why not) and would allow us to just do `showToast(notification)`, or `showToast(notification, {trace: true})` in case we want to use the other parameters.

Max-7 (changes requested):

client/src/services/notificationCenter.ts
NicoTuxx: export was useless.
However the interface was helping to type the function easily. Now we're forced to: 
```
    const toastConfig: any = {
      theme: this._getMsTheme(notification.level),
      icon: this._getToastIcon(notification.level),
      message: notification.message,
    };
    if (notification.title) {
      toastConfig.title = notification.title;
    }
```

Max-7 (approved):

PR #5231, opened 2023-09-18
Simplify `Restore libparsec ...`

FirelightFlagboy:
- Correct `permissions`.
- Skip voiding possible `ls` errors those wouldn't be captured by the `$(..)` since it only capture stdout.
- Rewrite error message to have a title and simplify listing of incriminating file inside `./server/parsec.libs`.

### DIFF
--- a/.github/workflows/ci-python.yml
+++ b/.github/workflows/ci-python.yml
@@ -37,7 +37,7 @@ env:
   # much slower (postgresql tests runs in ~9mn on 12 vs ~36mn on 14 !)
   postgresql-version: 12
 
-  permissions:
+permissions:
   contents: read
 
 jobs:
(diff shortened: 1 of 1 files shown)


.github/workflows/ci-python.yml
```diff
@@ -156,9 +156,9 @@ jobs:
           python make.py python-ci-install
 
           # Make sure POETRY_LIBPARSEC_BUNDLE_EXTRA_SHARED_LIBRARIES=false worked
-          if [ -n "$( ls -A ./server/parsec.libs 2>/dev/null )" ];
+          if [ -n "$(ls -A ./server/parsec.libs)" ];
```
mmmarcos, line 159: What `ls` errors can occur here? I guess probably a non-existing directory which can be check explicitly:
```suggestion
          if [ -d "./server/parsec.libs" ] && [ -n "$(ls -A ./server/parsec.libs)" ];
```

mmmarcos (changes requested):

mmmarcos (approved):
:+1:

PR #5235, opened 2023-09-18
Don't use regex for OrganizationID/UserID/DeviceName validation to speedup tests

touilleMan:
Regex compilation is done once but is very costly. When testing this is done for every test (since nextest runs each test in it own process).
Given the regex pattern is simple, we can instead to the check by hand ;-)

Bench:
```
# master
$ cargo nextest run
[...]
     Summary [   4.271s] 1380 tests run: 1380 passed, 1 skipped
# This PR
$ cargo nextest run
[...]
Summary [   3.159s] 1380 tests run: 1380 passed, 1 skipped
```

This is still a WIP though: we use `\w` pattern rule [which corresponds to `word character (\p{Alphabetic} + \p{M} + \d + \p{Pc} + \p{Join_Control})`](https://docs.rs/regex/1.9.5/regex/#perl-character-classes-unicode-friendly)
So we should take extra care that this is correctly handled (and add more tests about that !)

### DIFF
--- a/libparsec/crates/types/src/id.rs
+++ b/libparsec/crates/types/src/id.rs
@@ -1,7 +1,6 @@
 // Parsec Cloud (https://parsec.cloud) Copyright (c) BUSL-1.1 2016-present Scille SAS
 
 use email_address_parser::EmailAddress;
-use regex::Regex;
 use serde::{Deserialize, Serialize};
 use serde_with::{DeserializeFromStr, SerializeDisplay};
 use std::convert::TryFrom;
@@ -145,7 +144,7 @@ macro_rules! new_uuid_type {
 }
 
 macro_rules! new_string_based_id_type {
-    (pub $name:ident, $bytes_size:expr, $pattern:expr) => {
+    (pub $name:ident, $match_fn:expr) => {
         #[derive(
             Clone, SerializeDisplay, DeserializeFromStr, PartialEq, Eq, Hash, PartialOrd, Ord,
         )]
@@ -154,23 +153,6 @@ macro_rules! new_string_based_id_type {
         impl Default for $name {
             fn default() -> Self {
                 Self(uuid::Uuid::new_v4().as_simple().to_string())
-            }
-        }
-
-        impl $name {
-            const BYTES_SIZE: usize = $bytes_size;
-
-            fn pattern() -> &'static Regex {
-                lazy_static! {
-                    static ref PATTERN: Regex = Regex::new($pattern)
-                        .expect("`pattern` should be a valid regular expression");
-                }
-                &PATTERN
-            }
-
-            /// Validate that the input is contained in [`Self::BYTES_SIZE`] and match the regex provided by [`Self::pattern`].
-            pub fn is_valid(input: &str) -> bool {
-                input.len() <= Self::BYTES_SIZE && Self::pattern().is_match(input)
             }
         }
 
@@ -196,7 +178,7 @@ macro_rules! new_string_based_id_type {
             fn try_from(s: &str) -> Result<Self, Self::Error> {
                 let id: String = s.nfc().collect();
 
-                if Self::is_valid(&id) {
+                if $match_fn(&id) {
                     Ok(Self(id))
                 } else {
                     Err(concat!("Invalid ", stringify!($name)))
@@ -242,17 +224,28 @@ impl From<ChunkID> for BlockID {
     }
 }
 
+// Equivalent to pattern "^[\w\-]{1,32}$" on 32 bytes
+#[inline]
+fn match_legacy_id(id: &str) -> bool {
+    // `str::len` returns the number of bytes
+    let bytes_size = id.len();
+
+    let is_valid_char = |c: char| c == '-' || regex_syntax::is_word_character(c);
+
+    (1..=32).contains(&bytes_size) && id.chars().all(is_valid_char)
+}
+
 /*
  * OrganizationID
  */
 
-new_string_based_id_type!(pub OrganizationID, 32, r"^[\w\-]{1,32}$");
+new_string_based_id_type!(pub OrganizationID, match_legacy_id);
 
 /*
  * UserID
  */
 
-new_string_based_id_type!(pub UserID, 32, r"^[\w\-]{1,32}$");
+new_string_based_id_type!(pub UserID, match_legacy_id);
 
 impl UserID {
     pub fn to_device_id(&self, device_name: DeviceName) -> DeviceID {
@@ -264,13 +257,22 @@ impl UserID {
  * DeviceName
  */
 
-new_string_based_id_type!(pub DeviceName, 32, r"^[\w\-]{1,32}$");
+new_string_based_id_type!(pub DeviceName, match_legacy_id);
 
 /*
  * DeviceLabel
 */
 
-new_string_based_id_type!(pub DeviceLabel, 255, r"^.+$");
+// Equivalent to pattern r"^.+$" (i.e. at least 1 character) on 255 bytes
+#[inline]
+fn match_device_label(id: &str) -> bool {
+    // `str::len` returns the number of bytes
+    let bytes_size = id.len();
+
+    (1..=255).contains(&bytes_size)
+}
+
+new_string_based_id_type!(pub DeviceLabel, match_device_label);
 impl_from_maybe!(Option<DeviceLabel>);
 
 /*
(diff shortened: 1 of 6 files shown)


libparsec/crates/types/src/id.rs
```diff
@@ -314,7 +325,7 @@ impl TryFrom<&str> for DeviceID {
     fn try_from(s: &str) -> Result<Self, Self::Error> {
         const ERR: &str = "Invalid DeviceID";
         let (raw_user_id, raw_device_name) = s.split_once('@').ok_or(ERR)?;
-        let user_id = raw_user_id.parse().map_err(|_| ERR)?;
+        let user_id = raw_user_id.parse().map_err(|_| dbg!(ERR))?;
```
TimeEngineer, line 328: why dbg ?

FirelightFlagboy: That's something I forget to remove :sweat:

touilleMan:
Seems great to me, can you also add some tests ?

libparsec/crates/types/tests/unit/id.rs
```diff
@@ -38,3 +38,20 @@ fn human_handle(#[case] email: &str, #[case] label: &str, #[case] is_ok: bool) {
         is_ok
     );
 }
+
+#[rstest]
+#[case::alphanum("01239AbcdDEFA", true)]
+#[case::with_underscore("hello_World", true)]
+#[case::with_hyphen("hello-World", true)]
+// cspell:disable-next-line
+#[case::to_long("aaaabbbbccccddddeeeeffffgggghhhhiiiijjjjj", false)]
```
touilleMan, line 47:
```suggestion
#[case::too_long("aaaabbbbccccddddeeeeffffgggghhhhiiiijjjjj", false)]
```

touilleMan, line 47: add also `#[case::empty("", false)]`

libparsec/crates/types/tests/unit/id.rs
```diff
@@ -38,3 +38,20 @@ fn human_handle(#[case] email: &str, #[case] label: &str, #[case] is_ok: bool) {
...
+// cspell:disable-next-line
+#[case::with_unicode_1("ªµºÖØøˆˠˬˮ\u{301}ͶͽͿΆΊ", true)]
+#[case::with_unicode_2("ΌΎΣҁ\u{484}Աՙֈ\u{5af}\u{5bf}\u{5c1}\u{5c4}\u{5c7}", true)]
+#[case::with_unicode_3("ת\u{5ef}\u{610}٩ۓە", true)]
+#[case::with_unicode_4("\u{6df}\u{6ea}ۿܐݍ߀ߺ\u{7fd}ࠀࡀ", true)]
+// cspell:disable-next-line
+#[case::hello("𝔅𝔬𝔫𝔧𝔬𝔲𝔯", true)]
+fn test_match_legacy_id(#[case] input: &str, #[case] result: bool) {
```
touilleMan, line 55: I'd rather merge this with `libparsec/crates/types/tests/id.rs's `organization_id_user_id_and_device_name`/`bad_organization_id_user_id_and_device_name` tests as it is better to test the actual components we care about that they internals

libparsec/crates/types/tests/unit/id.rs
touilleMan: and also, this is 41 characters long (so 41bytes), would be better to have one test failing with 33bytes and another one passing with 32bytes

FirelightFlagboy: done

FirelightFlagboy: done

libparsec/crates/types/tests/unit/id.rs
FirelightFlagboy: done

libparsec/crates/types/tests/id.rs
```diff
@@ -15,31 +15,43 @@ fn device_label_bad_size() {
...
-#[case("f")]
-#[case("f-o-o")]
-#[case("f_o_o")]
-#[case(&"x".repeat(32))]
-#[case("三国")]
+#[case::text_32_long("This_one_is_exactly_32_char_long")]
+#[case::alphanum("01239AbcdDEFA")]
+#[case::with_underscore("hello_World")]
+#[case::with_hyphen("hello-World")]
+#[case::text_32_long("This_one_is_exactly_32_char_long")]
```
touilleMan, line 22: duplication of line 18 ?

libparsec/crates/types/tests/id.rs
```diff
@@ -15,31 +15,43 @@ fn device_label_bad_size() {
...
+#[case::with_unicode_2("ΌΎΣҁ\u{484}Աՙֈ\u{5af}\u{5bf}\u{5c1}\u{5c4}\u{5c7}")]
+#[case::with_unicode_3("ת\u{5ef}\u{610}٩ۓە")]
+#[case::with_unicode_4("\u{6df}\u{6ea}ۿܐݍ߀ߺ\u{7fd}ࠀࡀ")]
+// cspell:disable-next-line
+#[case::hello("𝔅𝔬𝔫𝔧𝔬𝔲𝔯")]
 fn organization_id_user_id_and_device_name(#[case] raw: &str) {
-    let organization_id = OrganizationID::from_str(raw).unwrap();
+    use unicode_normalization::UnicodeNormalization;
+
+    let raw = raw.nfc().collect::<String>();
```
touilleMan, line 33: I'd rather do a sanity check on the test input that convert them (this way the test stop if something is unexpected instead of running with unexpected data)

```suggestion
    // Sanity check to avoid debugging visually same but different stuff
    let nfc_raw = raw.nfc().collect::<String>();
    p_assert_eq!(nfc_raw, raw);
```

libparsec/crates/types/tests/id.rs
FirelightFlagboy: Yes, removed

libparsec/crates/types/tests/id.rs
FirelightFlagboy: If I've done that, it is because the `raw != nfc_raw`

touilleMan: if `raw != nfc_raw` that means our test is broken

Typically is `ΌΎΣҁ\u{484}Աՙֈ\u{5af}\u{5bf}\u{5c1}\u{5c4}\u{5c7}` NFC ? We don't know. Yet this is the string we will look it for debugging. So we will have to run the nfc conversion to then have the correct string value.

So I think it is better to provided-valid-value-and-assert-they-are-what-we-expect than to provide-potentially-invalid-value-then-convert-it-into-something-valid

touilleMan: so you know, by adding `p_assert_eq!(raw.as_bytes(), raw.nfc().collect::<String>().as_bytes());` the test fail. So indeed some of the test data are not nfc :stuck_out_tongue:

vxgmichel:
Before merging let's consider this alternative fix:
```rust
diff --git a/libparsec/crates/types/src/id.rs b/libparsec/crates/types/src/id.rs
index bd7d85b2d..9b58c57b3 100644
--- a/libparsec/crates/types/src/id.rs
+++ b/libparsec/crates/types/src/id.rs
@@ -246,13 +246,13 @@ impl From<ChunkID> for BlockID {
  * OrganizationID
  */
 
-new_string_based_id_type!(pub OrganizationID, 32, r"^[\w\-]{1,32}$");
+new_string_based_id_type!(pub OrganizationID, 32, r"^[\w\-]+$");
 
 /*
  * UserID
  */
 
-new_string_based_id_type!(pub UserID, 32, r"^[\w\-]{1,32}$");
+new_string_based_id_type!(pub UserID, 32, r"^[\w\-]+$");
 
 impl UserID {
     pub fn to_device_id(&self, device_name: DeviceName) -> DeviceID {
@@ -264,7 +264,7 @@ impl UserID {
  * DeviceName
  */
 
-new_string_based_id_type!(pub DeviceName, 32, r"^[\w\-]{1,32}$");
+new_string_based_id_type!(pub DeviceName, 32, r"^[\w\-]+$");
 
 /*
  * DeviceLabel
```

FirelightFlagboy:
@vxgmichel does changing the regex improve the compilation of it ? can you add some metric ?

libparsec/crates/types/tests/id.rs
FirelightFlagboy: done

FirelightFlagboy (approved):

PR #5237, opened 2023-09-18
A couple of updates for workspace archiving and deletion

vxgmichel:
A fixup for #5061


oxidation/libparsec/crates/types/src/organization.rs
```diff
@@ -21,6 +21,7 @@ pub struct OrganizationConfig {
     pub sequester_authority: Option<Vec<u8>>,
     // TODO: Should this be `SequesterServiceCertificate` instead of bytes?
     pub sequester_services: Option<Vec<Vec<u8>>>,
+    pub minimum_archiving_period: u64,
```
touilleMan, line 24: Use [`libparsec_types::time::Duration`](https://docs.rs/chrono/latest/chrono/struct.Duration.html) instead

The is no Python binding to `Duration` though, but I think it's okay to use `int`  in Python.
The conversion to `i64` [is pretty trivial btw](https://docs.rs/chrono/latest/chrono/struct.Duration.html#method.num_seconds)

parsec/core/backend_connection/authenticated.py
```diff
@@ -65,6 +65,15 @@
 DESYNC_RETRY_TIME = 10  # seconds
 
 
+# On the client side, the default minimum archiving period is simply 0.
+# It's used in two cases:
+# - The server is too old to support workspace archiving
+# - The client is offline.
+# Note that this value is different from the initial minimum archiving period
+# set when creating a new organization, which is 30 days.
+DEFAULT_MINIMUM_ARCHIVING_PERIOD = 0
```
touilleMan, line 74:
```suggestion
DEFAULT_MINIMUM_ARCHIVING_PERIOD = 0  # seconds
```

parsec/_parsec_pyi/organization.pyi
```diff
@@ -48,3 +49,5 @@ class OrganizationConfig:
     def sequester_services(self) -> tuple[bytes, ...] | None: ...
     @property
     def is_sequestered_organization(self) -> bool: ...
+    @property
+    def minimum_archiving_period(self) -> int: ...
```
touilleMan, line 53: Add a doctstring to specify this is in seconds ?
(the other solution is to rename to something like `minimum_archiving_period_secs`)

oxidation/libparsec/crates/types/src/organization.rs
vxgmichel: Do we really want this? 

> This also allows for the negative duration;

parsec/core/backend_connection/authenticated.py
vxgmichel: 073986e100f7a1256a42456419ce2992f6004d18

oxidation/libparsec/crates/types/src/organization.rs
touilleMan: I was wondering, but negative duration doesn't seem like a big issue given we only do `now - archiving_start > minimum_archiving_period`

The good point about using a `Duration` type is we are much more explicit about what this type represent, with a `u64` we always wonder if it's day ? seconds ? milliseconds ? etc.

The other solution is to add the unit in the name of the variable, but I find it is just better to use `Duration` instead (unless in the Python bindings were we definitely don't want to bother with more binding work for this ^^).

vxgmichel: [fb3447e](https://github.com/Scille/parsec-cloud/pull/5237/commits/fb3447e5d81652455a04a0649162772c97abe03e)

touilleMan (approved):

PR #5247, opened 2023-09-19
Add validation functions to bindings

FirelightFlagboy:


bindings/electron/src/index.d.ts
```diff
@@ -959,3 +959,39 @@ export function testGetTestbedBootstrapOrganizationAddr(
...
+export function validateEmail(
+    raw: string
+): Promise<boolean>
+export function validateDeviceLabel(
+    raw: string
+): Promise<boolean>
+export function validateInvitationToken(
+    raw: string
+): Promise<boolean>
+export function validateBackendAddr(
```
Max-7, line 980: Backend addresses can already be validated using `parseBackendAddr` (with a bit more precision even, being able to tell between UserInvitation and DeviceInvitation), so there's not really any need for these functions. The frontend already uses `parseBackendAddr`, and this just adds code that will not be used and will have to be maintained.

Max-7 (approved):
Looks good from an API standpoint. I would remove the functions to validate backend addresses just to reduce the API size.

bindings/electron/src/index.d.ts
FirelightFlagboy: The functions that check `BackendAddr*` have been removed

TimeEngineer (approved):

libparsec/src/validation.rs
```diff
@@ -0,0 +1,31 @@
+// Parsec Cloud (https://parsec.cloud) Copyright (c) BUSL-1.1 2016-present Scille SAS
+
+use std::str::FromStr;
+
+/// Validate a label without doing a unicode normalization.
+pub fn validate_entry_name(raw: &str) -> bool {
+    libparsec_types::EntryName::is_valid(raw).is_ok()
```
touilleMan, line 7: why not just `libparsec_types::EntryName::from_str(raw).is_ok()` ?

FirelightFlagboy: I wanted to not have an allocation being made for validating an input, that's why I've added some `is_valid` that only take an `&str` and don't do the unicode normalization

touilleMan: well shame the binding system is going to make an allocation anyway when converting the string from JS to Rust :trollface:

FirelightFlagboy: I can change to use `from_str` if prefered in another PR, WDYT ?

touilleMan: Don't bother that's not a big deal, just keep it in mind next time you're tempted with premature optimization ^^

PR #5250, opened 2023-09-20
Migrate `message_get` tests

FirelightFlagboy:


TimeEngineer:
Allow `unwrap` in your tests

libparsec/crates/protocol/tests/authenticated_cmds/v4/message_get.rs
```diff
@@ -1,15 +1,77 @@
...
+
+    let authenticated_cmds::AnyCmdReq::MessageGet(req2) = data else {
+        unreachable!()
+    };
+
+    let raw2 = req2.dump().unwrap();
+
+    let data2 = authenticated_cmds::AnyCmdReq::load(&raw2).unwrap();
+
+    assert_eq!(data2, expected);
```
mmmarcos, line 32: Consider re-using the `req` test from v3 (seems identical). Here's an example:
https://github.com/Scille/parsec-cloud/blob/0566f6e68cb953ee0385dcda64fef203cc05fe3e/libparsec/crates/protocol/tests/authenticated_cmds/v4/vlob_read.rs#L7-L13

FirelightFlagboy: They're indeed identical, but I can't reuse-it because it would cause a compilation error for the `rep_ok` which Response don't share the same fields

libparsec/crates/protocol/tests/authenticated_cmds/v3/message_get.rs
```diff
@@ -1,15 +1,74 @@
...
+
+    let authenticated_cmds::AnyCmdReq::MessageGet(req2) = data else {
+        unreachable!()
+    };
+
+    let raw2 = req2.dump().unwrap();
+
+    let data2 = authenticated_cmds::AnyCmdReq::load(&raw2).unwrap();
+
+    assert_eq!(data2, expected);
```
TimeEngineer, line 32: use `p_assert_eq!` instead from `libparsec_test_lite`

TimeEngineer: @touilleMan will tell you to use: ```Rust
use libparsec_tests_lite::prelude::*
```

libparsec/crates/protocol/tests/authenticated_cmds/v4/message_get.rs
mmmarcos: Ouch... this seems like something we'll probably come across quite often. 

I suppose it is because of the `use super::authenticated` in `v3/message_get.rs`. Maybe there is a way to force the `compat` module to use the `Rep` from `v3` :thinking:

FirelightFlagboy: It's because of the `use super::authenticated`

libparsec/crates/protocol/tests/authenticated_cmds/v3/message_get.rs
FirelightFlagboy: done

TimeEngineer (approved):

PR #5281, opened 2023-09-21
Sort generated element by name

FirelightFlagboy:
Having the element sorted help reviewing the generated files.

### DIFF
--- a/bindings/generator/generate.py
+++ b/bindings/generator/generate.py
@@ -532,30 +532,30 @@ def generate_api_specs(api_module: ModuleType) -> ApiSpecs:
             )
 
     return ApiSpecs(
-        str_based_types=[
+        str_based_types=sorted(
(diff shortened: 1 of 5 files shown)


bindings/generator/generate.py
```diff
@@ -532,30 +532,38 @@ class ParsingPlaceholder:
...
-            item.__name__
-            for item in api_items.values()
-            if isinstance(item, type) and issubclass(item, U32BasedType)
-        ],
-        enums=enums,
-        variants=variants,
-        structs=structs,
-        meths=meths,
+        str_based_types=sorted(
+            [
```
touilleMan, line 536: you can omit `[`  in a call

touilleMan (approved):

bindings/generator/generate.py

PR #5169, opened 2023-09-11
CLI: invite

TimeEngineer:
## Describe your changes

This is a fast prototype of invite CLI

`<URL>` is like: `parsec://127.0.0.1:41905`
`<TOKEN>` is like: `4e45cc21e7604af196173ff6c9184a1f`

```Rust
cargo run --bin cli_greet -- --url <URL>
cargo run --bin cli_claim -- --url <URL> --token <TOKEN>
```

## Checklist

Before you submit this pull request, please make sure to:

- [ ] Keep changes in the pull request as small as possible
   
- [ ] Ensure the commit history is sanitized
   
- [ ] Give a meaningful title to your PR
- [ ] Describe your changes
   
- [ ] Link any related issue in the description
   
- [ ] Link any dependent pull request in the description

If this PR adds a new feature or fixes a bug:

- [ ] I have added or updated relevant tests

If this PR modifies installation or updates packages:

- [ ] I have checked that installation works on all platforms (Windows, Linux, Mac ...)
    

- [ ] I have updated related documentations (installation commands, fix to specific problems ...)

If this PR is related to GUI:

- [ ] I have checked that GUIs affected by my changes are working properly.

If this PR affects the user:

- [ ] I have created a news fragment
   
- [ ] I have updated the user documentation

PR #5282, opened 2023-09-21
[MS] Added greet user with bindings

### LINKED ISSUE #5149: Bind greet user

Max-7:
Closes #5149 

Used #5169 for testing. Errors are not correctly handled in the vast majority of cases. It will be easier to handle them properly when testing with claim user directly though the UI.

- [X] Describe your changes
   
- [X] Link any dependent pull request in the description

### DIFF
--- a/client/src/views/users/GreetUserModal.vue
+++ b/client/src/views/users/GreetUserModal.vue
@@ -62,7 +62,7 @@
           class="step"
         >
           <sas-code-provide
-            :code="'ABCD'"
+            :code="greeter.hostSASCode"
           />
         </div>
 
@@ -72,7 +72,7 @@
           class="step"
         >
           <sas-code-choice
-            :choices="['ABCD', 'EFGH', 'IJKL', 'MNOP']"
+            :choices="greeter.SASCodeChoices"
             @select="selectGuestSas"
           />
         </div>
@@ -93,9 +93,9 @@
           class="step user-info-page"
         >
           <user-information
-            :default-email="invitation.claimerEmail"
-            :default-name="'Name Entered By The Guest'"
-            :default-device="'Device-given-by-the-guest'"
+            :default-email="''"
+            :default-name="''"
+            :default-device="''"
             :email-enabled="false"
             ref="guestInfoPage"
           />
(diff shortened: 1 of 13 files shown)


client/src/views/users/GreetUserModal.vue
```diff
@@ -123,21 +123,21 @@
...
-          v-if="userInfo"
+          v-if="guestInfoPage"
           class="step final-step"
         >
           <div class="final-step__name">
             <user-avatar-name
-              :user-name="userInfo.name"
-              :user-avatar="userInfo.name"
+              :user-name="guestInfoPage?.fullName"
+              :user-avatar="guestInfoPage?.fullName"
```
NicoTuxx, line 132: the same information is passed twice, there should be only one props and the user-avatar-name should use it for both on its own then.

client/src/views/users/GreetUserModal.vue
```diff
@@ -274,8 +267,24 @@ function getNextButtonText(): string {
...
+    if (result.ok) {
+      await nextStep();
+    } else {
+      console.log('Signify trust failed', result.error);
+      pageStep.value = GreetUserStep.WaitForGuest;
+    }
+  } else {
+    console.log('Incorrect code entered');
+    pageStep.value = GreetUserStep.WaitForGuest;
+  }
```
NicoTuxx, line 287: no abort on greeter side?

client/tests/e2e/specs/test_greet_user_modal.ts
```diff
@@ -15,7 +15,7 @@ describe('Greet user into an organization', () => {
     cy.dropTestbed();
   });
 
-  it('Open greet user modal', () => {
+  it.skip('Open greet user modal', () => {
```
NicoTuxx, line 18: why skip it?

NicoTuxx (changes requested):

client/src/views/users/GreetUserModal.vue
Max-7: Wanted to avoid changing UserAvatarName. It has been designed to be able to take something else as the user avatar, but in the meantime, we use the user name.

client/src/views/users/GreetUserModal.vue
Max-7: Slightly updated to restart the process in this case, but this will have to be better handled during the complete tests.

client/tests/e2e/specs/test_greet_user_modal.ts
Max-7: So the tests would run a little faster while debugging, forgot to remove it. Good catch!

PR #5288, opened 2023-09-21
Ms claim device

NicoTuxx:
## Checklist

- [ ] Describe your changes
   
If this PR is related to GUI:

If this PR affects the user:


client/src/components/users/UserInformation.vue
```diff
@@ -61,6 +72,10 @@ const props = defineProps({
     type: Boolean,
     default: true,
   },
+  activeFields: {
```
Max-7, line 75: Not a big fan of those updates. The API is a bit convoluted, it doesn't make much sense, you can get weird cases like no active fields at all.
If someone wants an input to get a device name, their first thought shouldn't be `I'll use the UserInformation component and deactivate name and email`, it should be `I'll create an input`.

client/src/parsec/classes.ts
Max-7, line 1: Rename the file to something like `claim.ts`. Since those files are not meant to be imported directly, we may even add `claim_user.ts`, `claim_device.ts`. And yeah we have a `functions.ts` file, and I intend to rename it too.

client/src/parsec/classes.ts
```diff
@@ -0,0 +1,469 @@
+// Parsec Cloud (https://parsec.cloud) Copyright (c) BUSL-1.1 2016-present Scille SAS
+
+import {
+  DeviceClaimFinalizeInfo,
+  DeviceClaimInProgress1Info,
+  DeviceClaimInProgress2Info,
+  DeviceClaimInProgress3Info,
+  libparsec,
+} from '@/plugins/libparsec';
```
Max-7, line 9: Only `libparsec` should be imported from `plugins/libparsec`, other types should be imported from `parsec/types`. This ensures that they are all exported.

client/src/parsec/classes.ts
```diff
@@ -0,0 +1,469 @@
...
+  UserClaimInProgress1Info,
+  UserClaimInProgress2Info,
+  UserClaimFinalizeInfo,
+  SASCode,
+  HumanHandle,
+  ClaimerRetrieveInfoError,
+  DeviceFileType,
+} from '@/parsec/types';
+
+import { getClientConfig } from '@/parsec';
```
Max-7, line 29: Import from `parsec/functions`. I'm a bit worried about circular imports otherwise.

client/src/parsec/classes.ts
```diff
@@ -0,0 +1,469 @@
...
+    this._assertState(true, true);
+
+    if (window.isDesktop()) {
+      const clientConfig = getClientConfig();
+      const result = await libparsec.claimerRetrieveInfo(clientConfig, eventCallback, invitationLink);
+      if (result.ok) {
+        this.handle = result.value.handle;
+        this.greeter = result.value.greeterHumanHandle;
+      }
+      return result as Result<UserOrDeviceClaimInitialInfoUser, ClaimerRetrieveInfoError>;
```
Max-7, line 318: The return value is incorrect. `claimerRetrieveInfo` returns a `UserOrDeviceClaimInitialInfoDevice`, your return value is a `UserOrDeviceClaimInitialInfoUser`. Shouldn't need to cast.

client/src/parsec/classes.ts
```diff
@@ -0,0 +1,469 @@
...
+    } else {
+      await _wait(this.MOCK_HOST_WAITING_TIME);
+      this.handle = DEFAULT_HANDLE;
+      this.greeter = {
+        email: 'gale@waterdeep.faerun',
+        // cspell:disable-next-line
+        label: 'Gale Dekarios',
+      };
+      return new Promise<Result<UserOrDeviceClaimInitialInfoUser, ClaimerRetrieveInfoError>>((resolve, _reject) => {
+        resolve({ok: true, value: {
```
Max-7, line 328: This returns a `UserOrDeviceClaimInitialInfoDevice`, tag should be `Device`, no need for any `claimerEmail`.

client/src/parsec/classes.ts
```diff
@@ -0,0 +1,469 @@
...
+      this.canceller = null;
+      return result;
+    } else {
+      this.SASCodeChoices = ['5MNO', '6PQR', '7STU', '8VWX'];
+      this.correctSASCode = '7STU';
+      return new Promise<Result<DeviceClaimInProgress1Info, ClaimInProgressError>>((resolve, _reject) => {
+        resolve({ok: true, value: {
+          handle: DEFAULT_HANDLE,
+          greeterSas: '7STU',
+          greeterSasChoices: ['5MNO', '6PQR', '7STU', '8VWX'],
```
Max-7, line 364: Might as well use the variables, this avoid declaring the values twice (and avoid potential bugs if it's changed later).

client/src/parsec/classes.ts
```diff
@@ -0,0 +1,469 @@
...
+      await _wait(this.MOCK_HOST_WAITING_TIME);
+      return new Promise<Result<DeviceClaimInProgress3Info, ClaimInProgressError>>((resolve, _reject) => {
+        resolve({ok: true, value: {
+          handle: DEFAULT_HANDLE,
+        }});
+      });
+    }
+  }
+
+  async doClaim(deviceLabel: string):
```
Max-7, line 416: It makes the API a little clearer to type it as `DeviceLabel` (even though it is just a string underneath).

client/src/parsec/classes.ts
```diff
@@ -0,0 +1,469 @@
...
+    if (window.isDesktop()) {
+      // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
+      const result = await libparsec.claimerDeviceFinalizeSaveLocalDevice(this.handle!, {tag: 'Password', password: password});
+      if (result.ok) {
+        this.device = result.value;
+      }
+      this.handle = null;
+      return result;
+    } else {
+      return new Promise<Result<AvailableDevice, ClaimInProgressError>>((resolve, _reject) => {
```
Max-7, line 452: Should also reset the handle in this case.

client/src/parsec/functions.ts
```diff
@@ -59,7 +59,7 @@ export async function listAvailableDevices(): Promise<Array<AvailableDevice>> {
   return await libparsec.listAvailableDevices(window.getConfigDir());
 }
 
-function getClientConfig(): ClientConfig {
+export function getClientConfig(): ClientConfig {
```
Max-7, line 62: This function should be put into another file, something like `internals.ts` that isn't exported by the index.ts. This way the `parsec` module can do `import { getClientConfig } from '@/parsec/internals';` but a dev cannot do `import { getClientConfig } from '@/parsec';`. This is also true for the `_wait` function, and for the `MOCK_HOST_WAITING_TIME` variable that could be changed into `MOCK_WAITING_TIME`.

This can be done later after this PR has been merged.

client/src/views/home/DeviceJoinOrganizationModal.vue
```diff
@@ -226,8 +222,26 @@ const titles = new Map<DeviceJoinOrganizationStep, Title>([[
   }],
 ]);
 
-function selectHostSas(_code: string | null): void {
-  nextStep();
+async function selectHostSas(selectedCode: string | null): Promise<void> {
+  if (!selectedCode) {
```
Max-7, line 226: Could probably be shortened with
```
if (selectedCode && selectedCode === claimer.value.correctSASCode) {
  // await signifyTrust();
} else {
  if (!selectedCode) {
    console.log('Not selected');
  } else {
    console.log('Incorrect');
  }
  await abort();
  pageStep.value = Start;
}
```

Max-7 (changes requested):

client/src/parsec/classes.ts
NicoTuxx: Perfect I was going to propose something like this for a future issue.

client/src/parsec/classes.ts
NicoTuxx: automatic imports, my bad will fix

client/src/parsec/claim_device.ts
```diff
@@ -0,0 +1,243 @@
...
+  DeviceLabel,
+  DeviceClaimFinalizeInfo,
+  DeviceClaimInProgress1Info,
+  DeviceClaimInProgress2Info,
+  DeviceClaimInProgress3Info,
+  AvailableDevice,
+  ClientEvent,
+  Handle,
+  Result,
+  DEFAULT_HANDLE,
```
Max-7, line 17: DEFAULT_HANDLE could be moved to `internals` too, but we can leave it to another PR.

client/src/views/home/DeviceJoinOrganizationModal.vue
```diff
@@ -170,36 +185,20 @@ enum DeviceJoinOrganizationStep {
...
 }
 
-const OTHER_USER_WAITING_TIME = 300;
-
 const { t } = useI18n();
 
 const pageStep = ref(DeviceJoinOrganizationStep.Information);
+const deviceName = ref();
 const passwordPage = ref();
+const fieldsUpdated = ref(false);
```
Max-7, line 193: Useless variable.

Max-7 (approved):

PR #5353, opened 2023-09-25
Make `docs` have its own `pyproject.toml` file to simplify dependencies specification.

FirelightFlagboy:
Also update the Docker file accordingly


docs/pyproject.toml
```diff
@@ -0,0 +1,19 @@
+[tool.poetry]
+name = "parsec-docs"
+version = "0.0.0"
+description = "The parsec-cloud documentations"
```
mmmarcos, line 4:
```suggestion
description = "Parsec documentation"
```

docs/pyproject.toml
```diff
@@ -0,0 +1,19 @@
...
+authors = ["Scille SAS <contact@scille.fr>"]
+license = "BUSL-1.1"
+readme = "../README.md"
+packages = []
+
+[tool.poetry.dependencies]
+python = "^3.9"
```
mmmarcos, line 11: Should this file also be added to the `/misc/version_updater.py` script? (it contains a Python version)

server/pyproject.toml
mmmarcos, line 1: Since this is moved to `/docs/pyproject.toml`, does dependabot configuration needs to be updated?

make.py
```diff
@@ -160,7 +160,7 @@ def run(self, cwd: Path, extra_cmd_args: Iterable[str]) -> None:
         # (note only the first `maturin develop` is impacted, as all maturin
         # develop uses the same default virtualenv)
         Cmd(
-            cmd="poetry install --with=docs,testbed-server",
+            cmd="poetry install --with=testbed-server",
```
mmmarcos, line 163: No need to install docs in `make.py`? btw, maybe we should create an issue to add a `$ ./make.py docs` option.

FirelightFlagboy: Both `pyproject.toml` with poetry will not share the same `virtualenv`.

IMO very few persons need the docs dependencies so maybe a `make.py docs` that do a `poetry -C docs install` could do the trick

server/pyproject.toml
FirelightFlagboy: Done, I've choose to make dependabot to update those dependencies in a monthly basis since they're not as important

mmmarcos (approved):
LGTM

make.py
FirelightFlagboy: I've created the issue #5356 to track that

Issue #4941, opened 2023-07-31
`libparsec`'s `panic` handle strategy (especially for async function !)


touilleMan:
## Electron

Currently a panic in `libparsec` result in the js GUI only receiving a very broad ```Error: `neon::types::Deferred` was dropped without being settled
Ok @ index-375ed170.js:1```:

![image](https://github.com/Scille/parsec-cloud/assets/3187637/bbf13b40-9698-4fa2-83f1-88525bf6bfda)

This is because calling a `libparsec` function result in a job dispatch in a different thread and issue a promise. The panic then occurs in the rust thread while running the job, and the promise is then cancel in a dirty way (hence the message about `neon::types::Deferred`)

## Web

No sure if there is an issue here given there is no separate thread the rust code runs on. However we should do some testing to ensure the behavior is sane.

## Logging & Sentry support

In any way, in case of a panic we want to:
- return a correct error message so that the GUI can display something
- send an error log on the libparsec event bus. The GUI code should be listening to it, and hence print a `console.log` (and, for electron, write the log in a file)
- have Sentry pic the panic and send a crash repport

PR #5366, opened 2023-09-26
Report `async panic` for `electron`

TimeEngineer:
## Describe your changes

Currently, when rust panics in `async` binding, it will show: `Uncaught (in promise) Error: neon::types::Deferred was dropped without being settled`, this PR adds a backtrace to the message to know where it crashes. (`RUST_BACKTRACE` env must be set)
See: #4941
The question now is: what to do next ?

## Checklist

- [x] Describe your changes
   
If this PR is related to GUI:

If this PR affects the user:


bindings/electron/src/lib.rs
```diff
@@ -10,6 +10,37 @@ lazy_static::lazy_static! {
...
+        }
+    };
+}
+
+macro_rules! handle_backtrace {
+    ($rx: ident, $cx: ident) => {
+        let _ = $rx.recv();
+
+        if let Some(backtrace) = crate::BACKTRACE.lock().expect("Mutex is poisoned").take() {
+            return $cx.throw_type_error(format!("Program panicked:\n{backtrace}"));
```
FirelightFlagboy, line 36:
```suggestion
            return $cx.throw_error(format!("Program panicked:\n{backtrace}"));
```

We don't know for sure that panic are caused by a type error

bindings/electron/src/lib.rs
```diff
@@ -10,6 +10,37 @@ lazy_static::lazy_static! {
...
+macro_rules! capture_backtrace {
+    ($tx: ident, $($tt: tt)*) => {
+        {
+            std::panic::set_hook(Box::new(|_| {
+                *crate::BACKTRACE.lock().expect("Mutex is poisoned") = Some(std::backtrace::Backtrace::capture());
+            }));
+
+            let ret = $($tt)*;
+
+            let _ = $tx.send(());
```
FirelightFlagboy, line 24: Instead of using the channel for synchronization why not sending the backtrace in ?

TimeEngineer: And in the case there is no error ?

TimeEngineer: If send is success, there is no error

bindings/electron/src/lib.rs
TimeEngineer: True, I didn't search for others `errors`

bindings/electron/src/lib.rs
FirelightFlagboy: I'm not sure to understand, did you mean that we send a message in the channel only when there is a backtrace to be send ?

Max-7 (approved):
It does work, I do get the stacktrace. Can't say much about the code though.

bindings/electron/src/lib.rs
TimeEngineer: I mean why do you want to send the backtrace in ?

TimeEngineer: If the code reachs the `send` statement, the function hasn't panicked, so why do you want to send the backtrace ?

bindings/electron/src/lib.rs
```diff
@@ -10,6 +10,37 @@ lazy_static::lazy_static! {
...
+
+            ret
```
FirelightFlagboy, line 26: After `set_hook` you need to call `take_hook` to reset to the default hook handler

TimeEngineer: I don't think we want to reset the default hook :thinking: Do you have a case ?

FirelightFlagboy: If we panic outside the your system, we would lose the backtrace

bindings/electron/src/lib.rs
TimeEngineer: How ?

bindings/electron/src/lib.rs
TimeEngineer: Done

bindings/electron/src/lib.rs
TimeEngineer: I mean `tx` can't be in the hook

FirelightFlagboy: `tx` can't be move into the closure ?

TimeEngineer: The hook has `static` lifetime, `tx` don't live enough

FirelightFlagboy: [`Sender`](https://doc.rust-lang.org/std/sync/mpsc/struct.Sender.html) don't have a lifetime constrain, I'm missing something ?

TimeEngineer: `tx` lives in his current scope (when it has been created) since we don't return it.

FirelightFlagboy: So I've tried on my side and rust complain about `Sender<Option<Backtrace>>` not implementing `Send` even tho we have:

- `impl<T: Send> Send for Sender<T>`
- `impl<T: Send> Send for Option<T>`
- `impl Send for Bracktrace`

PR #5368, opened 2023-09-27
Bindings list users

touilleMan:


libparsec/crates/client/src/certificates_ops/mod.rs
```diff
@@ -2,31 +2,63 @@
...
 
 use crate::{event_bus::EventBus, ClientConfig};
 
+pub struct UserInfo {
+    pub id: UserID,
+    pub human_handle: Option<HumanHandle>,
+    pub current_profile: UserProfile,
+    pub created_on: DateTime,
+    // `None` if signed by root verify key (i.e. the user that bootstrapped the organization)
+    pub created_by: Option<DeviceID>,
```
FirelightFlagboy, line 28: The first user don't sign itself ? that would remove the need of an option

PR #5372, opened 2023-09-27
[MS] Use bindings to list users

Max-7:
   
- [X] Describe your changes


client/src/components/users/UserListItem.vue
```diff
@@ -54,7 +54,8 @@
       <ion-label
         class="user-join-label cell"
       >
-        {{ timeSince(user.joined, '--', 'short') }}
+        {{ user.createdOn }}
+        <!-- {{ timeSince(user, '--', 'short') }} -->
```
Max-7, line 58: Will fix when dates are correctly deserialized.

client/src/parsec/functions.ts
```diff
@@ -331,3 +333,89 @@ export async function isValidDeviceName(name: string): Promise<boolean> {
...
 }
+
+export async function listUsers(skipRevoked = true): Promise<Result<Array<UserInfo>, ClientListUsersError>> {
+  const handle = getParsecHandle();
+
+  if (handle !== null && window.isDesktop()) {
+    const result = await libparsec.clientListUsers(handle, skipRevoked);
+    if (result.ok) {
+      result.value.map((item) => {
+        item.createdOn = DateTime.fromMillis(item.createdOn as any as number);
```
Max-7, line 344: This is done because the bindings returns a number, but typed as a luxon DateTime. It means the object can be used by the UI as a DateTime later on, we just have to force a conversion first.

client/src/parsec/functions.ts
```diff
@@ -3,8 +3,6 @@
 import {
   libparsec,
   InviteListItem,
```
NicoTuxx, line 5: should not this be imported by /types too?

client/src/views/users/ActiveUsersPage.vue
```diff
@@ -189,13 +188,6 @@ const indeterminateState = computed({
   set: (_val) => _val,
 });
 
-const filteredUsers = computed(() => {
-  const revokedUsers = userList.value.filter((user) => {
-    return user.revoked === false;
-  });
-  return revokedUsers;
-});
```
NicoTuxx, line 197: why removed filteredUsers from this page, but not from RevokedUsers?

NicoTuxx (changes requested):

client/src/views/users/ActiveUsersPage.vue
Max-7: Here it can be filtered directly using the API. the function `listUsers` has a boolean parameter to include or not revoked users, so we don't need any filter later on.
In the case of the revoked user page, we don't have anything in the API to list only the revoked users, so we need to filter the list.

client/src/parsec/functions.ts
Max-7: There's no reason to import/export it in types.ts. `InviteListItem` will never be used by someone who uses the API, they will only use `InviteListItemUser` or `InviteListItemDevice`. Since it's only used internally, I'd rather not export it at all.

NicoTuxx (approved):

PR #5378, opened 2023-09-28
Use `BigInt` for 64-bits integer

FirelightFlagboy:
To support 64-bits integer in the Javascript binding, I discover that we need to convert the rust type `i64` & `u64` to `BigInt` to not lose information.

Blocked by:

- [x] #5379

### DIFF
--- a/bindings/generator/templates/binding_web_meths.rs.j2
+++ b/bindings/generator/templates/binding_web_meths.rs.j2
@@ -169,22 +169,12 @@ if {{ js_val }}.is_null() {
     }
     v as u32
 }
-{%- elif type.kind == "i64_based" -%}
-{
-    let v = {{ js_val }}.dyn_into::<Number>().map_err(|_| TypeError::new("Not a number"))?.value_of();
-    if v < (i64::MIN as f64) || (i64::MAX as f64) < v {
-        return Err(JsValue::from(TypeError::new("Not an i64 number")));
-    }
-    v as i64
-}
-{%- elif type.kind == "u64_based" -%}
-{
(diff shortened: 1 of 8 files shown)


bindings/generator/templates/binding_web_meths.rs.j2
```diff
@@ -171,15 +171,15 @@ if {{ js_val }}.is_null() {
...
     if v < (i64::MIN as f64) || (i64::MAX as f64) < v {
         return Err(JsValue::from(TypeError::new("Not an i64 number")));
     }
     v as i64
 }
 {%- elif type.kind == "u64_based" -%}
 {
-    let v = {{ js_val }}.dyn_into::<Number>().map_err(|_| TypeError::new("Not a number"))?.value_of();
+    let v = {{ js_val }}.dyn_into::<BigInt>().map_err(|_| TypeError::new("Not a number"))?.value_of();
     if v < (u64::MIN as f64) || (u64::MAX as f64) < v {
```
TimeEngineer, line 183: Why `v` still `f64` ? :/

FirelightFlagboy: I didn't go further because I was blocked by electron (cf #5379)

TimeEngineer: if `value_of` converts to `f64` you will lose the information too

TimeEngineer: > I didn't go further because I was blocked by electron (cf #5379)

Ok :+1:

FirelightFlagboy: I did a quick change for web only ;)

PR #5380, opened 2023-09-28
Add well known user other than root

### LINKED ISSUE #5370: [🐛 | Bug]: `parsec-backend` can't start because of `getpwuid(): uid not found 1234`
### Version

2.16

Linux

### What happened ?

The docker container is configured to execute the command with the arbitrary user with id `1234:1234`

https://github.com/Scille/parsec-cloud/blob/9cdaec0075dea51108c648dc86dc843a4163a5fa/packaging/server/server.dockerfile#L36

This raise an `KeyError` exception in the container because the python code will use `getpwuid` with the current user id that isn't a real user. 

#### Workaround

Overwrite the user value, for example in a docker compose file:

```yaml
services:
  parsec-backend:
    image: Scille/parsec-server:2.16
    user: 0:0 # Line added to use a well know user inside the container (i.e. `root`)
```

### Relevant output

_No response_

FirelightFlagboy:
Fix #5370

### DIFF
--- a/server/packaging/server/server.dockerfile
+++ b/server/packaging/server/server.dockerfile
@@ -34,7 +34,10 @@ FROM python:3.9-slim
 LABEL org.opencontainers.image.source=https://github.com/Scille/parsec-cloud
 LABEL org.opencontainers.image.description="Run the Parsec backend server."
 
-USER 1234:1234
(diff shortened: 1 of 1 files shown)


touilleMan (approved):
:partying_face:

PR #5387, opened 2023-09-29
Add tests for `CertificatesOps`

### LINKED ISSUE #5176: Full tests of CertificateOps
The new [CertificateOps](https://github.com/Scille/parsec-cloud/tree/master/libparsec/crates/client/src/certificates_ops) has no tests for the moment, we should remediate this :smile: 

## Tests (non exhaustive !):

- Start failure (unit test: a bad path for the native storage is the simplest way to create the failure)
- Server error when doing `poll_server_for_new_certificates`
- Invalid certificate coming from the server
  - unreadable data
  - valid data but invalid certificate index
  - self-signed certificate that shouldn't be (e.g. user revoke certificate)
  - already existing certificate (e.g. user certificate for an already existing user)
  - Outsider cannot share one of it realm to somebody else
  - etc. 
- valid certificate integration (integrate a single certificate or a burst of multiple certificates at the same time)
- validate message & manifest (both with valid or invalid data)

## `CertificatesCachedStorage`

On top of that I think it would be good to test the `CertificatesCachedStorage` (not to be mixed with `CertificatesStorage` that is defined in the `libparsec_platform_storage` crate !).
`CertificatesCachedStorage` is an internal structure, but it is important and non-trivial so it would be good to test it. The good thing is it should be rather simple to test by just following the methods it exposes (basically only `set_foo_manifest`/`get_foo_manifest`) . And special attention should be payed to the cache system to ensure no inconsistency occurs there !

## Where to put the tests ?

Currently our approach is to write tests by default as integration tests, and only put them as unit tests if they require some internal stuff.
For testing this component, I think we should put everything as unit tests:
- `CertificatesCachedStorage` is internal so no way around to test it
- It simplify testing `CertificatesOps`, because it allows us to access it `CertificatesCachedStorage` field to simply get the internal state (`CertificatesOps` only store state in `CertificatesCachedStorage`, and it's good to ensure the state hasn't changed after having tried to integrate an invalid certificate)

The other solution is to expose everything that needs to be tested, but it is pretty ugly given it forces us to expose far too much things :(

(As a matter of fact I'm starting to wonder if we shouldn't just put all our tests as unit to avoid altogether this artificial unit/integration visibility split that is of no use for us...)

## Pro tip !

Note testbed is able to simulate the server response without having to start an actual server, [see this](https://github.com/Scille/parsec-cloud/blob/66727228812296aebd6714fbb63b31cf5c5249b6/libparsec/crates/client/tests/user_ops_tests/sync.rs#L44-L54) and [that for multiple requests](https://github.com/Scille/parsec-cloud/blob/66727228812296aebd6714fbb63b31cf5c5249b6/libparsec/crates/client/tests/user_ops_tests/sync.rs#L97-L120)

- [x] #5441
- [x] #5442
- [x] #5541
- [x] #5673
- [x] #5674

TimeEngineer:
## Describe your changes

closes #5176 

## Checklist

- [ ] Describe your changes
   
If this PR is related to GUI:

If this PR affects the user:


libparsec/crates/client/tests/unit/certificates_ops/store.rs
```diff
@@ -0,0 +1,190 @@
...
+
+use std::sync::Arc;
+
+use libparsec_platform_storage::certificates::GetCertificateError;
+use libparsec_tests_fixtures::prelude::*;
+use libparsec_types::prelude::*;
+
+use super::utils::certificates_store_factory;
+
+#[parsec_test(testbed = "minimal", with_server)]
```
touilleMan, line 11: It's better to avoid relying on the testbed server as much as possible (given it makes the test much slower)

It should be easy to do that for all tests involving the certificate ops:
- only the poll function communicates with the server
- the poll function is triggered only by the validate functions

so:
- you can test the `add_certificates_batch` function with absolutely no server
- for testing the validation function, you should use the testbed customize function to specify the certificate ops storage should be already populated for device alice@dev1 (see what is done for the minimal_client_sync template)
- for testing the poll command, it is better anyway to use the testbed request mock (again for speed, but also to ensure what command is needed to do this action)

libparsec/crates/client/tests/unit/certificates_ops/store.rs
```diff
@@ -0,0 +1,190 @@
...
+async fn add_device_certificate(alice_device_certif: &DeviceCertificate, env: &TestbedEnv) {
```
touilleMan, line 12: Regarding testing of the store: I think it's okay to have a couple of test on this internal component (given testing a small part of bigger system helps narrow down errors) just like you did here.

However don't try to go too far on this: `CertificateOps` is considered a public component ("public" meaning it is used by unrelated systems such as user/workspace ops and `libparsec/src`), so it api shouldn't change much and hence is a good fit for test

On the other hand `store.rs` is internal stuff and subject to change (guess what, I've already entirely rewritten it last week ^^)

So we want our test to use the more stable `CertificateOps` api so that we can do our refactoring with confidence.

touilleMan: Aaaand I realize `add_certificates_batch` is not part of the `CertificateOps` api

In theory this is is not needed given only `poll_server_for_new_certificates` needs to call it.

But I wonder if we shouldn't expose it there anyway (as `pub(crate)`, maybe with a `#[test]`). The idea is to make it explicit this function is considered stable api and we heavily rely on it for our tests.

I'm not sure how useful it is though: once you've written a ton of tests that relies on this `certfificates_ops::add::add_certificates_batch` function, it becomes pretty obvious this function should not be modified

It can also be convenient to do `certificate_ops.add_certificates_batch(...)` in your tests instead of the more exotic `certfificates_ops::add::add_certificates_batch(certificate_ops)`

what do you think ?

libparsec/crates/client/tests/unit/certificates_ops/store.rs
TimeEngineer: Outdated

libparsec/crates/client/tests/unit/certificates_ops/store.rs
TimeEngineer: Outdated

libparsec/crates/types/src/certif.rs
```diff
@@ -673,6 +673,8 @@ impl UnsecureSequesterServiceCertificate {
 impl SequesterServiceCertificate {
     pub fn unsecure_load(signed: Bytes) -> DataResult<UnsecureSequesterServiceCertificate> {
         let (_, compressed) =
+            // TODO: It should be SequesterVerifyKeyDer instead, but the signature
```
touilleMan, line 676: Also add a comment for `AnyCertificate::unsecure_load`

libparsec/crates/crypto/src/rustcrypto/sequester.rs
```diff
@@ -297,6 +297,17 @@ impl SequesterVerifyKeyDer {
         self.0.as_ref().n().bits() / 8
     }
 
+    pub fn unsecure_unwrap(
```
touilleMan, line 300: remove this

libparsec/crates/client/tests/unit/certificates_ops/store.rs
```diff
@@ -0,0 +1,393 @@
...
+        .add_next_certificate(1, &service_certif, &service_signed)
+        .await
+        .unwrap();
+
+    // Check that certificate is present
+    let got = write.get_any_certificate(1).await.unwrap();
+
+    p_assert_eq!(got, service_certif);
+
+    store.stop().await;
```
touilleMan, line 235: retart the store and try to access the certificate (this ensure we don't rely on a cache but on data in persistent database)

touilleMan (approved):

libparsec/crates/crypto/src/rustcrypto/sequester.rs
TimeEngineer: Done

libparsec/crates/client/tests/unit/certificates_ops/store.rs
```diff
@@ -0,0 +1,437 @@
...
+    let got = write.get_any_certificate(1).await.unwrap();
+
+    p_assert_eq!(got, authority_certif);
+}
+
+// TODO: CertificatesStore doesn't work with SequesterServiceCertificate for
+// the moment, because it is signed with SequesterSigningKeyDer and it try to
+// unsecure_load with VerifyKey, which is wrong. But SequesterVerifyKey need to
+// know the size of the key, because the signature size depends on.
+#[ignore]
```
FirelightFlagboy, line 238: Replace with a should_panic ?

TimeEngineer: No it shouldn't panic, it should pass but this is a bug

FirelightFlagboy: :open_mouth:

PR #5392, opened 2023-09-29
Format correctly `check_realm_role_certificate_consistency`

TimeEngineer:


touilleMan (approved):
the diff is hard to read, I guess the only change is the match turned into a if/else, right ?

TimeEngineer:
> the diff is hard to read, I guess the only change is the match turned into a if/else, right ?

Yes

PR #5407, opened 2023-10-02
Implement SasCode::from_str without Regex to avoid costly initial regex build

touilleMan:


vxgmichel:
I timed the compilation of the regex to be about:
- 13 microseconds in release
- 50 microseconds in WASM
- 200 microseconds in unoptimized
- 400 microseconds in WASM unoptimized

Given the regex is cached using `lazy_static!`, I'm -1 on these kind of error-prone optimizations.

vxgmichel:
Regex are only costly when they use `\w`. I vote for closing this.

libparsec/crates/types/src/invite.rs
```diff
@@ -85,15 +83,15 @@ impl FromStr for SASCode {
...
+            return Err("Invalid SAS code");
         }
-        if PATTERN.is_match(s) {
-            Ok(Self(s.to_string()))
-        } else {
-            Err("Invalid SAS code")
+        for c in s.as_bytes() {
+            if !SAS_CODE_CHARS.contains(c) {
+                return Err("Invalid SAS code");
+            }
```
FirelightFlagboy, line 92: Use iter::any

libparsec/crates/types/src/invite.rs
```diff
@@ -65,7 +64,6 @@ impl ToString for InvitationType {
 
 // SAS code is composed of 4 hexadecimal characters
```
FirelightFlagboy, line 65: The comment seems out of place now, I think it should be move to a doc-comment of the struct `SASCode`

libparsec/crates/types/src/invite.rs
```diff
@@ -65,7 +64,6 @@ impl ToString for InvitationType {
...
 const SAS_CODE_CHARS: &[u8; 32] = b"ABCDEFGHJKLMNPQRSTUVWXYZ23456789";
```
FirelightFlagboy, line 66: Should we add a comment explaining why we don't have `I` (uppercase `i`) and `1` ?

I though we had a comment like that but I don't remember where :thinking: maybe it was lost :shrug:

touilleMan: I added a comment about that  :+1:

libparsec/crates/types/src/invite.rs
touilleMan: I moved the comment to make it a doc-comment

PR #5412, opened 2023-10-03
Update python dependencies before 2.16 release

vxgmichel:


vxgmichel:
Not happening for the moment: https://github.com/pgjones/quart-trio/issues/11

PR #5413, opened 2023-10-03
[MS] Handle case when there are no devices on the machine

Max-7:
Not pretty but it will do for now.

- [X] Describe your changes

### DIFF
--- a/client/src/views/home/HomePage.vue
+++ b/client/src/views/home/HomePage.vue
@@ -62,6 +62,7 @@
               v-if="showOrganizationList"
               v-model="orgSearchString"
               id="ms-search-input"
+              v-show="deviceList.length >= 2"
             />
             <ion-button
               @click="isPopoverOpen = !isPopoverOpen; openPopover($event)"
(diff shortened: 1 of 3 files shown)


client/src/views/home/HomePage.vue
```diff
@@ -62,6 +62,7 @@
               v-if="showOrganizationList"
               v-model="orgSearchString"
               id="ms-search-input"
+              v-show="deviceList.length > 1"
```
NicoTuxx, line 65: Should it not match the filter condition?

client/src/views/home/HomePage.vue
```diff
@@ -93,54 +94,64 @@
...
+                    {{ $t('HomePage.noDevices') }}
+                  </ion-card-title>
+                  {{ $t('HomePage.howToAddDevices') }}
+                </ion-card-content>
+              </div>
+              <div v-if="deviceList.length > 0">
+                <ion-card-content class="organization-container">
+                  <ion-card-title class="organization-filter">
+                    <!-- No use in showing the sort/filter options for less than 2 devices -->
+                    <template v-if="deviceList.length > 2">
```
NicoTuxx, line 109: Here the condition does not match the comment.
Also it should be the same condition as the search one.

NicoTuxx (changes requested):

NicoTuxx (approved):

PR #5418, opened 2023-10-04
Fix docker build for parsec 2.16

vxgmichel:
Fix issue #5406

Also update the poetry lockfile.

PR #5437, opened 2023-10-06
Fix docker build 

vxgmichel:
Port of PR #5418

### DIFF
--- a/server/packaging/server/server.dockerfile
+++ b/server/packaging/server/server.dockerfile
@@ -6,7 +6,7 @@
 
 FROM python:3.9 AS builder
 
-WORKDIR /work
+WORKDIR /server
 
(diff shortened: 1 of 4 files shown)


touilleMan (approved):

server/packaging/server/server.dockerfile
```diff
@@ -36,12 +36,21 @@ LABEL org.opencontainers.image.description="Run the Parsec backend server."
 
 RUN groupadd --gid=1234 parsec && useradd --home-dir=/home/parsec --create-home --uid=1234 --gid=1234 parsec
 
+# Create parsec user and group
```
FirelightFlagboy, line 39: We created the group & the user at the line 37, here we change the default user with `USER`

PR #5458, opened 2023-10-09
Fix `SequesterAuthority` keys in test

TimeEngineer:
When we do
```rust
builder.bootstrap_organization("alice")
   .and_set_sequestered_organization()
```

it raises the following panic:
```shell
thread 'certificates_ops::tests::store::add_sequester_authority_certificate' panicked at 'called `Result::unwrap()` on an `Err` value: SequesterPrivateKeyDer("PKCS#8 ASN.1 error: unexpected ASN.1 DER tag: expected SEQUENCE, got INTEGER at DER byte 5")', libparsec/crates/testbed/src/template/build.rs:334:81
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```


FirelightFlagboy (commented):
Hello @TimeEngineer can you provide more context about what you've fixed in the body of the pull-request or commit ?

touilleMan (approved):

touilleMan:
> Hello @TimeEngineer can you provide more context about what you've fixed in the body of the pull-request or commit ?

@FirelightFlagboy The sample key seems to be corrupted, so we replace it with a valid one

PR #5465, opened 2023-10-10
Replace `base64` by `data-encoding` in libparsec_client_connection

touilleMan:


FirelightFlagboy:
Is it related to https://mastodon.social/@rust_discussions/111177778310492873 ?

If that's the case can you make the comment more clear about it

touilleMan:
> Is it related to https://mastodon.social/@rust_discussions/111177778310492873 ?

Partially: since we already use `data-encoding` for base32, I figured out we might as well use it for base64 to have a ergonomic and unified API on this encoding stuff

(okay I considered myself personally insulted by the quote "Rust is not the language for you if you don't like traits" :trollface: )

Cargo.toml
FirelightFlagboy, line 1: We still has `base64` & `base32` listed as dependencies

touilleMan: Good catch !

`base32` was still used by `libparsec_crypto`, I've fixed that (it is especially good to remove `base32` crate as it hasn't had update in 5years ^^)

`base64` is required by 3rd party libraries so there not much we can do here

FirelightFlagboy (approved):

PR #5472, opened 2023-10-10
[MS] Use file API to list/rename/delete

### LINKED ISSUE #5433: Create a new TextInputModal
Simple modal with only an input, a title, a text, and a validator, to replace `CreateUserInvitationModal`, `CreateWorkspaceModal`, and it will also be used for files.

Max-7:
Use the bindings to navigate into folders and create/rename/delete files and folders.

A lot of refactoring, some functions have been moved to `workspace.ts` and `validation.ts`. Route uses a `workspaceHandle` instead of a `workspaceId` now.

Also added TextInputModal (closes #5433).

- [X] Describe your changes


client/src/locales/en-US.json
```diff
@@ -306,7 +310,30 @@
             "maxSize": "Maximale size: 25Mb",
             "or": "or",
             "browse": "Browse"
-        }
+        },
+        "errors": {
+            "deleteFailed": "Failed to delete `{name}`",
+            "renameFailed": "Failed to rename `{name}`",
+            "createFolderFailed": "Failed to create new folder `{name}`",
+            "listFailed": "Failed to list folder `{path}`"
```
NicoTuxx, line 318:
```suggestion
            "listFailed": "Failed to list the content of the folder `{path}`"
```

client/src/views/files/FoldersPage.vue
```diff
@@ -207,59 +207,121 @@ import { routerNavigateTo } from '@/router';
 import { FileAction } from '@/views/files/FileContextMenu.vue';
 import MsActionBar from '@/components/core/ms-action-bar/MsActionBar.vue';
 import FileUploadModal from '@/views/files/FileUploadModal.vue';
-
+import { NotificationCenter, Notification, NotificationKey, NotificationLevel } from '@/services/notificationCenter';
+import * as parsec from '@/parsec';
+import { join as pathJoin } from '@/common/path';
+import { useI18n } from 'vue-i18n';
+import { getText } from '@/components/core/ms-modal/MsTextInputModal.vue';
```
NicoTuxx, line 214: The name is really not clear.
Should atleast ne something like `getTextInputModal` or `showTextInputModal`

NicoTuxx (changes requested):

client/src/components/core/ms-modal/MsModal.vue
```diff
@@ -150,7 +150,7 @@ import MsInformativeText from '@/components/core/ms-text/MsInformativeText.vue';
 const modal: Ref<HTMLDivElement | null> = ref(null);
 const props = defineProps<MsModalConfig>();
 defineEmits<{
-  (e: 'onEnterKeyup', value: string): void
+  (e: 'onEnterKeyup'): void
```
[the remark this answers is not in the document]
Max-7: Except it's emitted here https://github.com/Scille/parsec-cloud/pull/5472/files/c165630223ee9ddb80ad680b074d334678d1411f#diff-708af6618a0947288171f43e64d128e3a958f60565733ec02cdbf2d87139b64dR9 without passing it a value.
`onEnterKeyup` from a ms-input does take a value, but from a the modal it doesn't (what would be the value anyway?). If I remember correctly I even had a warning from VSCode about it.

client/src/views/files/FoldersPage.vue
Max-7: The function does not return a text input modal, and while it does show a text input modal, its role is to get a string from the user. I do agree that `getText` is kinda meh but I didn't really find something better. `getTextFromUser` maybe but it sounds weird too.

client/src/components/core/ms-modal/MsModal.vue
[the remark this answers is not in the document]
NicoTuxx: My bad, did not remember it was for modals too.
Keep it like this

client/src/views/files/FoldersPage.vue
NicoTuxx: On the technical side this is a createAndPresentModal function like there are others in the project.
On the UX side its role is to getTextInputfromUser.
Should be based on one or the other.

Max-7: Renamed to `getInputTextFromUser`

NicoTuxx (approved):

PR #5473, opened 2023-10-10
Fix runtime default path config

touilleMan:


libparsec/src/config.rs
```diff
@@ -31,3 +31,53 @@ impl From<ClientConfig> for libparsec_client::ClientConfig {
...
+pub fn get_default_config_dir() -> PathBuf {
+    #[cfg(target_arch = "wasm32")]
+    {
+        PathBuf::from("")
+    }
+    #[cfg(not(target_arch = "wasm32"))]
+    {
+        // TODO: Windows path differs from Parsec v2 here:
+        // - previous: `%APPDATA%\parsec\config`
+        // - current: `%APPDATA%\Roaming\parsec-v3-alpha`
```
Max-7, line 62: `%APPDATA%`, contrary to what the name suggests, points to `C:\Users\<user>\APPDATA\Roaming`

libparsec/src/config.rs
```diff
@@ -31,3 +31,53 @@ impl From<ClientConfig> for libparsec_client::ClientConfig {
...
+pub fn get_default_data_base_dir() -> PathBuf {
+    #[cfg(target_arch = "wasm32")]
+    {
+        PathBuf::from("")
+    }
+    #[cfg(not(target_arch = "wasm32"))]
+    {
+        // TODO: Windows path differs from Parsec v2 here:
+        // - previous: `%APPDATA%\parsec\data`
+        // - current: `%APPDATA%\Roaming\parsec-v3-alpha`
```
Max-7, line 44: Same here, `%APPDATA%` already points to `\Roaming`

Max-7 (approved):

libparsec/src/config.rs
FirelightFlagboy:
```suggestion
        // - current: `%APPDATA%\parsec-v3-alpha`
```

libparsec/src/config.rs
FirelightFlagboy:
```suggestion
        // - current: `%APPDATA%\parsec-v3-alpha`
```

bindings/electron/src/meths.rs
```diff
@@ -6521,6 +6521,66 @@ fn client_stop(mut cx: FunctionContext) -> JsResult<JsPromise> {
...
+        };
+        match custom_to_rs_string(ret) {
+            Ok(ok) => ok,
+            Err(err) => return cx.throw_type_error(err),
+        }
+    })
+    .or_throw(&mut cx)?;
+    let (deferred, promise) = cx.promise();
+    deferred.resolve(&mut cx, js_ret);
+    Ok(promise)
```
mmmarcos, line 6541: This code seems identical in the 3 functions, maybe there is a way to extract it into a function :
```rust
fn from_dir(path: std::path::PathBuf, mut cx: FunctionContext) -> JsResult<JsPromise>
```

touilleMan: this code is autogenerated (and differs between electron, web and android in there future), so I'd rather stay KISS there :smiley:

libparsec/src/config.rs
touilleMan: I based this comment on Parsec v2.16 code: https://github.com/Scille/parsec-cloud/blob/75da5c338ba837d25e09a75b732dbc59de44b551/parsec/core/config.py#L23-L44

Have you experienced the config/data folder to be in another directory with Parsec v2 ?

FirelightFlagboy: [`dirs::data_dir`](https://docs.rs/dirs/5.0.1/dirs/fn.data_dir.html) correspond to `<user>\AppData\Roaming` (i.e. `%APPDATA%`)

FirelightFlagboy (approved):

PR #5481, opened 2023-10-11
[MS] Use MsTextInputModal for invitation link and user email

Max-7:
- [X] Describe your changes


client/src/components/core/ms-modal/MsTextInputModal.vue
```diff
@@ -3,6 +3,7 @@
 <template>
   <ms-modal
     :title="$props.title"
+    :subtitle="$props.subtitle"
```
NicoTuxx, line 6: why do you need to use $props, it should work with only subtitle.

NicoTuxx (dismissed):

client/src/components/core/ms-modal/MsTextInputModal.vue
Max-7: Yeah it does but I like to be explicit. I don't know what happens if I create a variable in my component named `subtitle`. This way it's explicit, I know where the variable is from, and I avoid potential conflicts.

NicoTuxx: .

NicoTuxx (changes requested):

client/src/components/core/ms-modal/MsTextInputModal.vue
Max-7: Removed the `$props`, see #5488

NicoTuxx (approved):

PR #5498, opened 2023-10-12
feat: add toastManager

NicoTuxx:


client/src/services/notificationCenter.ts
```diff
@@ -114,46 +110,17 @@ export class NotificationCenter {
       this._trace(notification);
     }
 
-    const result = await this._createAndPresentToast({
+    const result = await this.toastManager.createAndPresent({
```
Max-7, line 113: I would change this to
```typescript
    this.toastManager.createAndPresent({
      theme: this._getMsTheme(notification.level),
      message: notification.message,
      title: notification.title,
      duration: DEFAULT_NOTIFICATION_DURATION,
    }).then(async (result) => {
      if (result && result.role === 'confirm') {
        this.markAsRead(notification.id);
      }
    });
```
Just to avoid it being blocking in case someone calls it with `await notificationManager.showToast()`, as it could create unexpected behaviour. I don't think there's ever a case where we want to ensure that the toast has been closed before continuing.

Max-7 (changes requested):

Max-7 (approved):

PR #5501, opened 2023-10-13
Provide enum for tag in variants

FirelightFlagboy:
I choose the same format for the enum tag value as for simple enum meaning the value of the enum is `{{ enum.name }}{{ enum.variant.name }}`


Max-7 (changes requested):
Way too many subjective changes in style that are not enforced by the linter configuration. Keep the PR to what it's supposed to do : provide enums for tags.

FirelightFlagboy:
@Max-7 can you pin point the files that need to be reset ?

Max-7:
> @Max-7 can you pin point the files that need to be reset ?

Every change that is a style update (added or removed spaces, new lines, ...). Keep the PR to replacing tags by enums.

If you want to make changes in how the code is formatted, it should be discussed beforehand and enforced with the linter.

Max-7 (approved):

PR #5510, opened 2023-10-14
Migrate the Glossary from `wiki` to `docs`

### LINKED ISSUE #3852: Create a glossary to improve consistency in translations
We should write and maintain a glossary that will serve as a reference for the different concepts that appear in the application. In particular we should settle on consistent and specific wordings for user translations. A possible template could look like this:
```
Noun: English/French
Verb: English/French
Definition: ...
Not to be confused with: ...
Synonyms to avoid: ...
```
Example:
```
Noun: invitation/invitation
Verb: to invite/inviter
Definition: getting a new user into a given organization
Not to be confused with:
 - the on-boarding process: this is specifically used for the part where both the host and the new user are connected simultaneously to exchange information in order to finalize the invitation
 - adding a device: invitation is specifically used for users, the term "adding" should be used instead for devices
Synonyms to avoid: enroll, enlist, on-board, greet, join
```

A list of concept that might go in the glossary:
- Organization/Organisation
- User/Utilisateur
- Device/Appareil
- Invitation/Invitation
- Joining/Rejoindre
- On-boarding/Accueil
- Workspace/Espace de travail
- Snapshots/Vue instantanée (timestamped workspaces)
- File/Fichier
- Folder/Dossier
- Administrator/Administrateur
- Standard user/Utilisateur standard
- Outsider/Externe
- Owner/Propriétaire
- Manager/Gestionnaire
- Contributor/Contributeur
- Reader/Lecteur

List of entries that already need a fix in the translations:
- ~enrôlement~ : procédure d’accueil
- ~greet~: onboard
- ~donner votre avis~: nous contacter
- ~directory~: folder
- ~voir l'état~: détails

Current version in the wiki https://github.com/Scille/parsec-cloud/wiki/Glossary

### LINKED ISSUE #5509: Glossary

FirelightFlagboy:
The glossary page is archived here: <https://github.com/Scille/parsec-cloud.old-wiki/blob/master/Glossary.md>

I choose to rely on an html table instead of markdown one because:

- I wanted to not have ultra wide row due to long text that need to fit in a single line. So the long text could be split on multiple line but still being display in the same cell.

- It also allows me to include complex element in a cell. Some cell use html list

Closes #3852, Closes #5509

### DIFF
--- a/docs/development/glossary.md
+++ b/docs/development/glossary.md
@@ -1,0 +1,231 @@
+<!-- Parsec Cloud (https://parsec.cloud) Copyright (c) BUSL-1.1 2016-present Scille SAS -->
+
+# Glossary
+
+<!-- markdownlint-disable no-inline-html -->
+<table>
+  <tr>
+    <th>Noun (EN / FR)</th>
+    <th>Verb (EN / FR)</th>
+    <th>Definition</th>
+    <th>Not to be confused with</th>
+    <th>Synonyms to avoid</th>
+  </tr>
+  <tr>
+    <td>User / Utilisateur</td>
+    <td></td>
+    <td>A Parsec User. They can be standard or outsider.</td>
+    <td></td>
+    <td></td>
+  </tr>
+  <tr>
+    <td>Standard user / Utilisateur standard
+    <td></td>
+    <td>
+      They can have any role in a workspace (Owner, Manager, Contributor, Reader).
+      Create workspaces in their organizations.
+      Add other devices to their organizations.
+      This is set during the onboarding process.
+    </td>
+    <td>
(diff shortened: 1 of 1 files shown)


FirelightFlagboy:
you can visually compare both documentation using the following links:

- (old) https://github.com/Scille/parsec-cloud.old-wiki/blob/master/Glossary.md
- (new) https://github.com/Scille/parsec-cloud/blob/5509-glossary/docs/development/glossary.md

mmmarcos (approved):
I can spot some improvements, but I guess the goal of the PR is just to move information away from the wiki. In that case this PR is OK :+1:

PR #5530, opened 2023-10-16
Remove outdated command for `x25519-dalek`

FirelightFlagboy:
That comment was when we where using the release-candidate of that crate.
That was changed in the commit 2b563996652c6fb49f0693aa675984a6a97d7ae2


touilleMan (approved):
\o/

PR #5539, opened 2023-10-17
Add `Delay` struct that is generated by `sleep`

FirelightFlagboy:
That still allow to use sleep in the old fashion way:

```rust
sleep(ONE_SECOND).await;
```

But now that it return an object, it allow something like:

```rust
let delay = sleep(ONE_SECOND).await;

// Other stuff

delay.await
```

### DIFF
--- a/libparsec/crates/platform_async/src/native/mod.rs
+++ b/libparsec/crates/platform_async/src/native/mod.rs
@@ -1,5 +1,7 @@
 // Parsec Cloud (https://parsec.cloud) Copyright (c) BUSL-1.1 2016-present Scille SAS
 
+use std::pin::Pin;
+
 pub mod oneshot {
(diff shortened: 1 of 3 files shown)


libparsec/crates/platform_async/src/native/mod.rs
```diff
@@ -16,9 +18,26 @@ pub mod watch {
     };
 }
 
+pub struct Delay {
+    sleep: Pin<Box<tokio::time::Sleep>>,
```
TimeEngineer, line 22: Why do you need a field ?

touilleMan:
do we have a use case for this ? In the Python codebase we only used `await sleep(DURATION)` and never felt limited by this

FirelightFlagboy:
Was initially made for #5597 but it appear that it isn't needed.
So closing

PR #5561, opened 2023-10-19
Fix confusing numbering of versions in the file history widget

vxgmichel:
The versions displayed in the file history widget are confusing since they correspond to the vlob version and not the actual file version. This PR fixes this by enumerating the vlob ids and displaying the version as `<vlob_id_counter>.<vlob_version>`:

Before:
broken-versions

After:
fixed_versions

This fix is enough as the version history will likely change in parsec v3.

### DIFF
--- a/newsfragments/5561.bugfix.rst
+++ b/newsfragments/5561.bugfix.rst
@@ -1,0 +1,1 @@
+Fix confusing numbering of versions in the file history display.


newsfragments/5561.bugfix.rst
FirelightFlagboy, line 1: Could you add the previous and new format ?

parsec/core/gui/file_history_widget.py
```diff
@@ -167,9 +167,15 @@ def on_get_version_success(
         if download_limit_reached:
             self.button_load_more_entries.setVisible(False)
         self.versions_job = None
+        version_counter = 0
+        current_id: EntryID | None = None
         for author, version in versions_list:
+            if version.id != current_id:
+                version_counter += 1
+                current_id = version.id
+            version_string = f"{version_counter}.{version.version}"
```
FirelightFlagboy, line 176: Why not the other way around ?
Why we have the counter first and not the version, I'm confused

newsfragments/5561.bugfix.rst
vxgmichel: Sorry I merged it before I saw your message. This is just a quick fix to avoid confusion anyway, it will change with parsec v3

parsec/core/gui/file_history_widget.py
vxgmichel: `version.version` here represents the version of the vlob.
But when a file is replaced by another one, a new entry is created so the versioning for the new entry starts back at 1.

PR #5566, opened 2023-10-19
Add `sentry`

### LINKED ISSUE #5440: Investigate Sentry in Electron & Web
Try sentry in `rust/js` to see how it renders stacktrace on Sentry

TimeEngineer:
closes #5440 
It looks like that in `sentry`:
image

### DIFF
--- a/bindings/electron/src/lib.rs
+++ b/bindings/electron/src/lib.rs
@@ -1,6 +1,8 @@
 // Parsec Cloud (https://parsec.cloud) Copyright (c) BUSL-1.1 2016-present Scille SAS
 
 use neon::prelude::*;
+use once_cell::sync::OnceCell;
+use sentry::{ClientInitGuard, ClientOptions};
 use std::sync::Mutex;
 
 mod meths;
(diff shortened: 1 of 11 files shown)


bindings/electron/src/lib.rs
```diff
@@ -10,6 +12,22 @@ lazy_static::lazy_static! {
...
+        SENTRY.get_or_init(|| {
+            sentry::init((
+                sentry_url,
+                ClientOptions {
+                    release: sentry::release_name!(),
+                    ..Default::default()
+                },
+            ))
+        });
+    }
```
FirelightFlagboy, line 28: Should we add a log indicating that sentry was or wasn't configured ?

bindings/electron/src/meths.rs
FirelightFlagboy, line 1: Can't the electron just call `init_sentry` a single time ?

bindings/web/src/lib.rs
```diff
@@ -2,12 +2,30 @@
...
 #[allow(unused_imports)]
 use wasm_bindgen::prelude::*;
 
 mod meths;
 pub use meths::*;
 
+static SENTRY: OnceCell<ClientInitGuard> = OnceCell::new();
+
+fn init_sentry() {
+    if let Ok(sentry_url) = std::env::var("SENTRY_URL") {
```
FirelightFlagboy, line 16: Env variable in web ? how does that work ?

bindings/web/src/meths.rs
FirelightFlagboy, line 1: Like `electron` can't the web warmup the binding by calling `init_sentry` a single time ?

FirelightFlagboy (commented):
So we choose to have 2 sentry-client running at the same time ?

TimeEngineer:
> So we choose to have 2 sentry-client running at the same time ? (on javascript & rust)

We haven't decided yet

bindings/electron/src/meths.rs
TimeEngineer: The problem is that we don't have a main thread accessible in `rust`

bindings/electron/src/lib.rs
TimeEngineer: @touilleMan

bindings/web/src/lib.rs
FirelightFlagboy: Still, how does that work ?

bindings/electron/src/meths.rs
TimeEngineer: See: https://docs.sentry.io/platforms/rust/#async-main-function

mmmarcos:
This work is included in #7814

PR #5587, opened 2023-10-23
[Backport] Correct dependencies installation on RTD

FirelightFlagboy:
Configure RTD to install python dependencies trough `pip install docs`.
For that I need to correct the `pyproject.toml` to make it work with
`pip install ...`, for that:

- I used the correct path for the README (damned .rst).
- Having to specify at least something to include in the package
  `parsec-docs`


mmmarcos (approved):
LGTM

Issue #5511, opened 2023-10-14
SSE Client: Test the `last-event-id` mechanism works when connection disconnects


FirelightFlagboy:
The client is connected with the client, if it lose connection it should send a `last-event-id` header when reconnecting.

If events has occurred since then, the server push the events the client has missed.

FirelightFlagboy:
The documentation about how a user-agent should behave when reestablishng the connection: <https://html.spec.whatwg.org/multipage/server-sent-events.html#reestablish-the-connection>.

We need to force the user-agent to enter in this State. 

As we can't access the inner mecanism of the SSE user-agent (the socket), the disruption need to be made on the server-side.

Issue #4550, opened 2023-05-15
Additional tests for SSE (client side)


touilleMan:
Basic tests for SSE already exists: https://github.com/Scille/parsec-cloud/blob/76f06a1ec77fd794f50fc140a1f5fa594727a0fa/oxidation/libparsec/crates/client_connection/tests/cmds.rs#L66-L161

Additional tests we want:

- [ ] SSE Client: Test the `missed_events` mechanism 

  The server can send a special SSE message `missed_events` that should trigger a [special SSE response type in the client](https://github.com/Scille/parsec-cloud/blob/76f06a1ec77fd794f50fc140a1f5fa594727a0fa/oxidation/libparsec/crates/client_connection/src/authenticated_cmds.rs#LL80C5-L80C17).

- [x] #5511

  The client is connected with the client, if it loses connection it should send a `last-event-id` header when reconnecting. 
  If events has occurred since then, the server push the events the client has missed.

However those tests are a bit tedious to write:

- We want to test errors behavior, so the testbed server cannot be used.
- We want to test the whole SSE machinery, including the 3rd party libs that implement SSE protocol.
  So we cannot use the testbed mock system.

So there is two solutions to implemented those tests:

1. Modify the testbed server to be able to notify it when it should voluntarily close the client connection to simulate loss of connection.
2. Start our own dummy HTTP server and serve SSE payloads 

   We used to have an [implementation of such code](https://github.com/Scille/parsec-cloud/blob/27918409d5f3643782084316e3f0990695f500cf/oxidation/libparsec/crates/client_connection/tests/auth_cmds.rs#L148-L164), 
   but it got replaced by the use of the testbed server. Guess it's time for it to return \o/.

Solution `2` seems the cleaner and less painful way to go (given for testing the `last-event-id` system, we want).

PR #5597, opened 2023-10-23
Add SSE client test for `last-event-id` mechanism

FirelightFlagboy:
Depends on:

- [x] #5635

### DIFF
--- a/libparsec/crates/client_connection/Cargo.toml
+++ b/libparsec/crates/client_connection/Cargo.toml
@@ -51,3 +51,7 @@ reqwest = { workspace = true, features = ["native-tls"] }
 libparsec_tests_lite = { workspace = true }
 # Note `libparsec_tests_fixtures` enables our `test-with-testbed` feature
 libparsec_tests_fixtures = { workspace = true, features = ["test-with-client-connection-testbed"] }
+
+[target.'cfg(not(target_arch = "wasm32"))'.dev-dependencies]
+# Needed by proxy server used for testing SSE
+tokio = { workspace = true, features = ["net"] }
--- a/libparsec/crates/client_connection/src/authenticated_cmds/mod.rs
+++ b/libparsec/crates/client_connection/src/authenticated_cmds/mod.rs
@@ -19,7 +19,7 @@ use libparsec_types::prelude::*;
 use crate::testbed::{get_send_hook, SendHookConfig};
 use crate::{
     error::{ConnectionError, ConnectionResult},
-    SSEConnectionError, SSEEventID, SSEStream, API_VERSION_HEADER_NAME, PARSEC_CONTENT_TYPE,
+    SSEConnectionError, SSEStream, API_VERSION_HEADER_NAME, PARSEC_CONTENT_TYPE,
 };
 
 use self::sse::EVENT_STREAM_CONTENT_TYPE;
@@ -148,7 +148,7 @@ impl AuthenticatedCmds {
 // SSE method impls
 
 impl AuthenticatedCmds {
-    fn sse_request_builder<T>(&self) -> RequestBuilder
+    fn sse_request_builder<T>(&self, last_event_id: Option<&str>) -> RequestBuilder
     where
         T: ProtocolRequest<API_LATEST_MAJOR_VERSION> + Debug + 'static,
     {
@@ -162,7 +162,7 @@ impl AuthenticatedCmds {
             url
         };
 
-        let mut content_headers = HeaderMap::with_capacity(4);
+        let mut content_headers = HeaderMap::with_capacity(5);
 
         let api_version = api_version_major_to_full(T::API_MAJOR_VERSION);
         let api_version_header_value =
@@ -171,23 +171,40 @@ impl AuthenticatedCmds {
         // No Content-Type as this request is a GET
         content_headers.insert(CONTENT_LENGTH, HeaderValue::from_static("0"));
         content_headers.insert(ACCEPT, HeaderValue::from_static(EVENT_STREAM_CONTENT_TYPE));
+
         let authorization_header_value = HeaderValue::from_str(
             &generate_authorization_header_value(&self.device.device_id, &self.device.signing_key),
         )
         .expect("always valid");
         content_headers.insert(AUTHORIZATION, authorization_header_value);
 
+        if let Some(last_event_id) = last_event_id {
+            // Last event ID is passed as a utf8 string, hence it is possible
+            // (in theory at least) that it cannot be encoded as a header value
+            // (e.g. if it contains NULL bytes).
+            //
+            // In such unlikely case we cannot just ignore the event id and not
+            // send a Last-Event-ID header as it would silently make us miss events.
+            //
+            // Instead we default to a dummy header value (i.e. empty string, so
+            // the server will be forced to send us a `missed_events` event message
+            // (just like if we had provided a very old but valid Last-Event-ID).
+            let value =
+                HeaderValue::from_str(last_event_id).unwrap_or(HeaderValue::from_static(""));
+            content_headers.insert("Last-Event-ID", value);
+        }
+
         self.client.get(url).headers(content_headers)
     }
 
     pub async fn start_sse<T>(
         &self,
-        last_event_id: Option<&SSEEventID>,
+        last_event_id: Option<String>,
     ) -> Result<SSEStream<T>, SSEConnectionError>
     where
         T: ProtocolRequest<API_LATEST_MAJOR_VERSION> + Debug + 'static,
     {
-        let request_builder = self.sse_request_builder::<T>();
+        let request_builder = self.sse_request_builder::<T>(last_event_id.as_deref());
 
         #[cfg(feature = "test-with-testbed")]
         let response = self
@@ -219,10 +236,9 @@ impl AuthenticatedCmds {
             return Err(SSEConnectionError::InvalidContentType(content_type.clone()));
         }
 
-        // log::trace!("Will listen for SSE event at {}", response.url());
         let mut sse_stream = response.bytes_stream().eventsource();
         if let Some(last_event_id) = last_event_id {
-            sse_stream.set_last_event_id(last_event_id.to_str());
+            sse_stream.set_last_event_id(last_event_id);
         }
         let event_source = Box::pin(sse_stream);
 
--- a/.github/workflows/ci-rust.yml
+++ b/.github/workflows/ci-rust.yml
@@ -63,7 +63,7 @@ jobs:
       TESTBED_SERVER: http://localhost:6777
     services:
       parsec-testbed-server:
-        image: ghcr.io/scille/parsec-cloud/parsec-testbed-server:2.16.0-a.0-dev-20240223.sha.21ad97f
+        image: ghcr.io/scille/parsec-cloud/parsec-testbed-server:2.16.0-a.0-dev-20240225.sha.4321df1
         ports:
           - 6777:6777
     steps:
--- a/.github/workflows/ci-web.yml
+++ b/.github/workflows/ci-web.yml
@@ -36,7 +36,7 @@ jobs:
     # https://github.com/Scille/parsec-cloud/pkgs/container/parsec-cloud%2Fparsec-testbed-server
     services:
       parsec-testbed-server:
-        image: ghcr.io/scille/parsec-cloud/parsec-testbed-server:2.16.0-a.0-dev-20240223.sha.21ad97f
+        image: ghcr.io/scille/parsec-cloud/parsec-testbed-server:2.16.0-a.0-dev-20240225.sha.4321df1
         ports:
           - 6777:6777
     steps:
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1580,6 +1580,7 @@ dependencies = [
  "reqwest",
  "reqwest-eventsource",
  "thiserror",
+ "tokio",
 ]
 
 [[package]]
--- a/libparsec/crates/client/src/monitors/connection.rs
+++ b/libparsec/crates/client/src/monitors/connection.rs
@@ -3,7 +3,7 @@
 use std::{collections::HashMap, ops::ControlFlow, sync::Arc};
 
 use libparsec_client_connection::{
-    AuthenticatedCmds, ConnectionError, RateLimiter, SSEEventID, SSEResponseOrMissedEvents,
+    AuthenticatedCmds, ConnectionError, RateLimiter, SSEResponseOrMissedEvents,
 };
 use libparsec_platform_async::{pretend_future_is_send_on_web, stream::StreamExt};
 use libparsec_platform_storage::certificates::PerTopicLastTimestamps;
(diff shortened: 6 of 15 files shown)


libparsec/crates/client_connection/src/authenticated_cmds/mod.rs
```diff
@@ -206,6 +206,13 @@ impl AuthenticatedCmds {
         // No Content-Type as this request is a GET
         content_headers.insert(CONTENT_LENGTH, HeaderValue::from_static("0"));
         content_headers.insert(ACCEPT, HeaderValue::from_static(EVENT_STREAM_CONTENT_TYPE));
+        if let Some(id) = last_event_id {
+            content_headers.insert(
+                "last-event-id",
+                HeaderValue::from_str(id)
+                    .expect("Invalid char in `id` to be converted into a header value"),
```
TimeEngineer, line 213: Is it safe ?

FirelightFlagboy: panic-ing is safe, yes :nerd_face:

libparsec/crates/client_connection/tests/proxy/mod.rs
```diff
@@ -0,0 +1,166 @@
...
+                            break
+                        },
+                        None => {},
+                    }
+                }
+            }
+        }
+        log::info!("Proxy stopped");
+        Ok(())
+    }
```
TimeEngineer, line 155: Can you space a bit ?

libparsec/crates/client_connection/tests/sse.rs
```diff
@@ -0,0 +1,114 @@
+#![cfg(not(target_arch = "wasm32"))]
+// Parsec Cloud (https://parsec.cloud) Copyright (c) BUSL-1.1 2016-present Scille SAS
```
TimeEngineer, line 2: Header comes first

libparsec/crates/client_connection/tests/proxy/mod.rs
FirelightFlagboy: Does the loop block need spaces or the entire fn block ?

libparsec/crates/platform_http_proxy/src/lib.rs
```diff
@@ -56,15 +56,15 @@ mod native {
...
+        pub fn with_http_proxy(mut self, proxy: String) -> anyhow::Result<Self> {
             let proxy = Proxy::http(proxy)
                 .map_err(|e| anyhow::anyhow!("Invalid HTTP proxy configuration: {}", e))?;
             self.http_proxy.replace(proxy);
             Ok(self)
         }
 
         /// Set the https proxy to use, will overwrite the previous configuration.
-        pub(crate) fn with_https_proxy(mut self, proxy: String) -> anyhow::Result<Self> {
+        pub fn with_https_proxy(mut self, proxy: String) -> anyhow::Result<Self> {
```
TimeEngineer, line 67: Same

libparsec/crates/client_connection/tests/sse.rs
```diff
@@ -0,0 +1,114 @@
...
+
+use futures_util::StreamExt;
+use libparsec_client_connection::{AuthenticatedCmds, ConnectionError, SSEResponseOrMissedEvents};
+use libparsec_platform_http_proxy::ProxyConfig;
+use libparsec_protocol::authenticated_cmds::latest as authenticated_cmds;
+use libparsec_testbed::TestbedEnv;
+use libparsec_tests_lite::{p_assert_eq, p_assert_matches, parsec_test};
+
+#[parsec_test(testbed = "coolorg", with_server)]
+#[timeout(std::time::Duration::from_secs(5))]
```
TimeEngineer, line 14: Why this timeout ? is the test flacky ?

libparsec/crates/client_connection/tests/proxy/mod.rs
TimeEngineer: If you can space between log :thinking:

libparsec/crates/client_connection/tests/sse.rs
FirelightFlagboy: done

libparsec/crates/client_connection/tests/sse.rs
FirelightFlagboy: > is the test flacky ?

I don't know :shrug:, only time will tell.

> Why this timeout ?

During development of the test, I've found out that if something goes wrong there will be no progress (i.e. the test will not finish).
This timeout is here to prevent the test for being blocked for too long (by aborting it).

libparsec/crates/client_connection/src/authenticated_cmds/mod.rs
FirelightFlagboy: I've choose to use a panic because I was lazy to create a new variant in `ConnectionError` for an invalid id provided by the server.

Should I add a variant nevertheless for that ?

libparsec/crates/client_connection/tests/sse.rs
touilleMan: I think it's a good idea to have a timeout on tests.
What do you think of nextest's timeout system ? (https://nexte.st/book/slow-tests.html#terminating-tests-after-a-timeout)

In think it would be better to handle this with it so that we don't have to bother with marking each test with the timeout marker.
However from my quick look at this, it seems we must add a `.config/cargo.toml` file to configure the timeout, instead of just passing the config on the command line :cry:  (which would be very convenient given we want very small timeout value when debugging on local machine, and much larger value when running on the CI)

What do you think ?

libparsec/crates/client_connection/src/authenticated_cmds/mod.rs
touilleMan: > I've choose to use a panic because I was lazy to create a new variant in ConnectionError for an invalid id provided by the server.

As a generic rule, it shouldn't be possible for a peer to make the application panic. Otherwise this typically opens the door to denial of service attacks.

In the current case, we should do
- validate `last-event-id` as soon as possible, then pass it around as dedicated `SSEEventID` type (typically something that just wrap `String` and implement a `FromStr` that do the validation and a `AsRef<str>` to access it).
- if validation fail, we end up with a `ConnectionError::InvalidLastEventID`, so that it can be handle like a regular error
- in `sse_request_builder` we take as parameter the `SSEEventID`, and hence can have a `.expect("Already validated")` ;-)

touilleMan (changes requested):

libparsec/crates/client_connection/tests/proxy/mod.rs
```diff
@@ -0,0 +1,169 @@
+// Parsec Cloud (https://parsec.cloud) Copyright (c) BUSL-1.1 2016-present Scille SAS
```
touilleMan, line 1: Avoid folder with single `mod.rs`, it's more redable to just have `proxy.rs` instead

libparsec/crates/client_connection/tests/proxy/mod.rs
```diff
@@ -0,0 +1,169 @@
...
+    }
+}
+
+enum Event {
+    Close,
+    Disconnect,
+}
+
+pub async fn spawn(backend_addr: BackendAddr) -> io::Result<ProxyHandle> {
+    let port = rand::thread_rng().gen_range(10_000..u16::MAX);
```
touilleMan, line 50: Specify a port at 0 in `TcpListener::bind` and get back to allocated port with `TcpListener::local_addr` instead

libparsec/crates/platform_http_proxy/src/lib.rs
```diff
@@ -56,15 +56,15 @@ mod native {
 
     impl ProxyConfig {
         /// Set the http proxy to use, will overwrite the previous configuration.
-        pub(crate) fn with_http_proxy(mut self, proxy: String) -> anyhow::Result<Self> {
+        pub fn with_http_proxy(mut self, proxy: String) -> anyhow::Result<Self> {
```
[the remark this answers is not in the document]
touilleMan: If we expose those method, we must implement them in wasm, which is currently not possible given in wasm the proxy config is always a noop...

From what I understand, you use the proxy object in the SSE tests to tell the client to go to the test proxy server.
However this test proxy server is already a TCP proxy, so it should be fine to just modify `alice_device`'s organization addr to use the TCP proxy domain/port instead of the testbed one.

libparsec/crates/client_connection/Cargo.toml
```diff
@@ -43,8 +43,16 @@ log = { workspace = true }
...
 [dev-dependencies]
 libparsec_tests_lite = { workspace = true }
 # Note `libparsec_tests_fixtures` enables our `test-with-testbed` feature
 libparsec_tests_fixtures = { workspace = true, features = ["test-with-client-connection-testbed"] }
+# FIXME: `rstest` is unstable when renamed. https://github.com/la10736/rstest/issues/221
+rstest = { workspace = true }
+futures-util = { workspace = true }
+rand = { workspace = true }
+anyhow = { workspace = true }
+log = { workspace = true }
```
touilleMan, line 57: use print instead of log:
- in test you don't need fancy level of output: it's either the test pass and the whole output is discarded, or the test fail and we want all logs
- using log is error prone (nothing will show up if the correct level is not configured, which is not trivial to do given nothing explains you should use the `LOG_LEVEL` environ variable)
- using print force to think a bit more of the logs that should stay (with logs it is tempting to have plenty of logs of low level "just in case" which in the end are never used, or are set at a higher level and clutter everything)
- using log is probably not going to work in the long run: I plan to use the event bus to pass the logs in the application (this has the benefit of 1. being testable and 2. simplify passing logs to the GUI for Sentry)

libparsec/crates/client_connection/tests/sse.rs
```diff
@@ -0,0 +1,114 @@
+// Parsec Cloud (https://parsec.cloud) Copyright (c) BUSL-1.1 2016-present Scille SAS
+#![cfg(not(target_arch = "wasm32"))]
+
+mod proxy;
+
+use futures_util::StreamExt;
```
touilleMan, line 6: can you use `libparsec_platform_async::stream::StreamExt` (i.e. `futures_lite::stream::StreamExt`) instead ?

libparsec/crates/client_connection/tests/proxy/mod.rs
FirelightFlagboy: I would need to disable `autotest`, that would work ?

libparsec/crates/client_connection/tests/sse.rs
FirelightFlagboy: done

libparsec/crates/client_connection/tests/proxy/mod.rs
FirelightFlagboy: done

libparsec/crates/client_connection/tests/sse.rs
FirelightFlagboy: > What do you think of nextest's timeout system ?

That could work.

---

> it seems we must add a `.config/cargo.toml`

In the first paragraph of your link:

> add this to `.config/nextest.toml`

So no `.config/cargo.toml` involved :smile: 

---

> What do you think ?

A issue is needed

FirelightFlagboy: I've created the issue #5625

FirelightFlagboy: In the mean time, I've removed the rstest timeout

libparsec/crates/client_connection/Cargo.toml
FirelightFlagboy: done

libparsec/crates/platform_http_proxy/src/lib.rs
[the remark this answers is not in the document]
FirelightFlagboy: I tried to mutate `alice` with the following snippet:

```rust
    let mut alice = env.local_device("alice@dev1");
    // ...

    let alice_proxy = proxy::spawn(env.server_addr.clone()).await.unwrap();

    std::sync::Arc::get_mut(&mut alice)
        .expect("We don't have the exclusive copy of `alice`")
        .organization_addr = BackendOrganizationAddr::new(
        alice_proxy.backend_addr(),
        bob.organization_id().clone(),
        bob.root_verify_key().clone(),
    );
```

Sadly this snippet panic at the ``.expect("We don't have the exclusive copy of `alice`")``'s line.

Help wanted

libparsec/crates/client_connection/src/authenticated_cmds/mod.rs
FirelightFlagboy: done

libparsec/crates/client_connection/tests/proxy/mod.rs
touilleMan: Ha ! you're right !

In that case just leave this as-is ;-)

(also note in the future we will most likely move all tests in `tests/unit` which will, among other things, allow us to rename this correctly ^^)

libparsec/crates/platform_http_proxy/src/lib.rs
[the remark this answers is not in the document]
touilleMan: you don't need to modify the device returned by `env.local_device("alice@dev1");` given the device object is not used internally by the testbed

Basically you could be creating your own `LocalDevice` object (think for example of a test where we try to load the device from a file and then start the client from this).

Hence you can use `Arc::make_mut(&mut alice)` to create a new alice device with the patched addr, then create the authenticated cmds with this device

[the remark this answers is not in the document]
FirelightFlagboy: OK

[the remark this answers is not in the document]
FirelightFlagboy: done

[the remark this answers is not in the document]
TimeEngineer: Why do you need in public ?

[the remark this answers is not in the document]
FirelightFlagboy: That modification was reverted, strange that github don't indicate it.

libparsec/crates/platform_http_proxy/src/lib.rs
FirelightFlagboy: done

mmmarcos (approved):
LGTM

TimeEngineer (approved):

libparsec/crates/client_connection/Cargo.toml
```diff
@@ -44,8 +44,13 @@ log = { workspace = true }
 [target.'cfg(not(target_arch = "wasm32"))'.dependencies]
 # TODO: support SSE for Web
 reqwest-eventsource = { workspace = true }
+tokio = { workspace = true, features = ["net"] }
```
touilleMan, line 47: only used in test, so dev dependency ?

mmmarcos:

mmmarcos:
@touilleMan @FirelightFlagboy this PR is linked to #5511 which is part of #4550 (currently on the V3 Roadmap). 
Is the PR still relevant ? should we resolve conflicts in order to merge it ?

touilleMan (approved):

PR #5616, opened 2023-10-24
Simplify cargo-clippy step

FirelightFlagboy:


.github/workflows/ci-rust.yml
```diff
@@ -135,9 +135,8 @@ jobs:
 
       # Clippy basically compile the project, hence it's faster to run it in
       # the test-rust-matrix job where compilation cache is reused !
-      - uses: ./.github/actions/use-pre-commit
-        with:
-          extra-args: clippy --verbose
+      - name: Check rust lint with cargo-clippy
+        run: cargo clippy --workspace
```
touilleMan, line 139: I guess this change modify the clippy arguments then:
```
- id: clippy
  name: clippy
  entry: cargo clippy
  language: system
  types: [ rust ]
  pass_filenames: false
  args:
    [
      --workspace,
      --tests,
      --bins,
      --lib,
      --exclude=libparsec_bindings_android,
      --exclude=libparsec_bindings_web,
      --exclude=libparsec_bindings_electron,
      --,
      --deny=warnings,
      --deny=clippy::undocumented_unsafe_blocks,
    ]
```

Is this okay ?

touilleMan (changes requested):

.github/workflows/ci-rust.yml
FirelightFlagboy: You are right, with this change clippy won't not run with the expected argument.

I'll wait for rust-lang/cargo#12115 to land on 1.75

FirelightFlagboy:
https://github.com/Scille/parsec-cloud/pull/5616#discussion_r1369810982

PR #5641, opened 2023-10-26
Move `@sentry/electron` to `client/electron`

FirelightFlagboy:


TimeEngineer:
It does not work, it builds correctly but nothing happen on `Sentry`

PR #5651, opened 2023-10-26
Ms select to dropdown

### LINKED ISSUE #4603: Split MsSelect component into two new components: MsDropdown, MsSortBy

NicoTuxx:
closes #4603


client/src/components/core/ms-types.ts
```diff
@@ -12,3 +12,35 @@ export enum MsModalResult {
...
+
+export interface MsChangeEvent {
+  option: MsOption,
+}
+
+export interface MsSorterLabels {
+  asc: string,
+  desc: string,
+}
+export interface MsDropdownChangeEvent extends MsChangeEvent {
```
Max-7, line 39: No need to reimplement `option` here.

```typescript
interface MsChangeEvent {
  option: MsOption,
}

export type MsDropdownChangeEvent = MsChangeEvent;

export interface MsSortedChangeEvent extends MsChangeEvent {
  sortByAsc: boolean,
}

```

client/src/components/core/ms-types.ts
```diff
@@ -12,3 +12,35 @@ export enum MsModalResult {
...
+export function getMsOptionByKey(options: MsOption[], key: any): MsOption | undefined {
+  return options.find((option) => {
+    return option.key === key;
+  });
+}
+
+export type MsDropdownOption = MsOption;
+export type MsSorterOption = MsOption;
+
+export interface MsChangeEvent {
```
Max-7, line 31: Could remove the export here, noone is supposed to use `MsChangeEvent` directly.

client/src/components/core/ms-types.ts
```diff
@@ -12,3 +12,35 @@ export enum MsModalResult {
   Cancel = 'cancel',
   Confirm = 'confirm',
 }
+
+export interface MsOption {
```
Max-7, line 16: Could remove the export, not supposed to use `MsOption` directly.

Max-7 (changes requested):

Max-7 (approved):

PR #5670, opened 2023-10-30
Add `User-Agent` header

FirelightFlagboy:
The format of the user agent is:

```
Parsec-Client/{version}; {platform}
```

Changes Summary
---------------

- Add a new crate `libparsec_constant`.
- Move the content of `libparsec/src/platform.rs` to `libparsec_constant` and re-export stuff here.
- Add `get_client_user_agent` in `libparsec_constant`.
- Use `libparsec_constant::get_client_user_agent` in `AuthenticatedCmds`, `InvitedCmds` and `AnonymousCmds`.
- `misc/version_updater.py` handle `libparsec_constant::CLIENT_VERSION`.

Current Blockers
----------------

- [x] seanmonstar/reqwest#2016


libparsec/crates/constant/tests/tests.rs
FirelightFlagboy, line 1: I've chosen to not use `libparsec_tests` as they're very basic and I don't think `p_assert_eq` would add much value here.

touilleMan:
This look too complicated to me

What about instead:
- don't move `libparsec/src/platform.rs`
 - keep the `get_client_user_agent` in `libparsec_client_connection`
 - `get_client_user_agent` uses a constant `PARSEC_VERSION` string obtained with a [`std::include_str!("../../version")`](https://doc.rust-lang.org/std/macro.include_str.html)
- `./libparsec/version` file only contains the version (e.g. `2.16.0-a.0+dev`, so no parsing is needed when using it from rust code) and any component in `libparsec` that need the version should use it (This way `version_updater.py` still have a single place to track)

Regarding the test, I'm not sure how to make them useful: the code to generate the client header is trivial, so we mostly test that we didn't forget to update the version... by using a code where we could also forget it (and it's very likely if we forget in one place we forget everywhere given the code to test and the tests are written in the same way and  language, at the same time by the same person ^^)

So my guess is to not bother to add tests on this

libparsec/crates/client_connection/src/constant.rs
```diff
@@ -0,0 +1,4 @@
+// Parsec Cloud (https://parsec.cloud) Copyright (c) BUSL-1.1 2016-present Scille SAS
+
+pub(crate) const CLIENT_USER_AGENT: &str =
+    concat!("Parsec-Client/", std::include_str!("../../../../version"));
```
mmmarcos, line 4: Why not putting this in [lib.rs](https://github.com/Scille/parsec-cloud/blob/998ef300224e63f14d5a805b075c9dfc25e81fd8/libparsec/crates/client_connection/src/lib.rs)? (That would be consistent with other constants like `PARSEC_CONTENT_TYPE` and `API_VERSION_HEADER_NAME`)

version
```diff
@@ -0,0 +1 @@
+2.16.0-a.0+dev
```
mmmarcos, line 1: Maybe move this file to the root of `client_connection` crate? Unless we already know it would be required by another crate (but it can also be moved later if/when needed)

libparsec/crates/client_connection/src/constant.rs
FirelightFlagboy: done

version
FirelightFlagboy: done

touilleMan (approved):

mmmarcos:
@FirelightFlagboy The linked pr is merged, is there another blocker to merge this?

FirelightFlagboy:
> is there another blocker to merge this?

Yes, I'm waiting for seanmonstar to produce a new release of `reqwest`.

FirelightFlagboy:
Require to have [`reqwest>v0.11.23`](https://github.com/seanmonstar/reqwest/releases/tag/v0.11.23)

mmmarcos:
> Require to have [`reqwest>v0.11.23`](https://github.com/seanmonstar/reqwest/releases/tag/v0.11.23)

@FirelightFlagboy It seems to have been released. Should we rebase & merge the PR linked to this issue? 
cc @touilleMan 


BTW, I think there is no need for the `;` in `Parsec-Client/{version}; {platform}`. Browsers seem to use ` ` (space):
https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/User-Agent

touilleMan:
I've added some tests (and discovered in the process the code was broken because pre-commit automatically added a trailing newline in `libparsec/version` :smile: )

> BTW, I think there is no need for the ; in Parsec-Client/{version}; {platform}. Browsers seem to use  (space):

Done.

PR #5679, opened 2023-10-31
[MS] file list not selectable

### LINKED ISSUE #5564: Last file in file list mode is not selectable
last-file-not-selectable

fabienSvtr:
Fixed bug: item behind gradient background (footer) not selectable

closes #5564

### DIFF
--- a/client/src/views/workspaces/WorkspacesPage.vue
+++ b/client/src/views/workspaces/WorkspacesPage.vue
@@ -83,8 +83,14 @@
           </ion-item>
         </div>
       </div>
-      <div class="workspaces-footer title-h5">
-        {{ $t('WorkspacesPage.itemCount', { count: workspaceList.length }, workspaceList.length) }}
+      <div class="workspaces-footer">
+        <div class="folder-footer__container">
+          <ion-text
+            class="text title-h5"
+          >
+            {{ $t('WorkspacesPage.itemCount', { count: workspaceList.length }, workspaceList.length) }}
+          </ion-text>
+        </div>
       </div>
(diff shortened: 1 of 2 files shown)


Max-7 (changes requested):
Few things:
- this only applies to files and users, not workspaces.
- in grid mode the counter becomes hidden behind the counter
- the item is still partially hidden behind the counter, is that the expected behavior?

A few screenshots to show some edge cases
workspace-behind

item-behind

behind

fabienSvtr:
> Few things:
> 
> * this only applies to files and users, not workspaces.
> * in grid mode the counter becomes hidden behind the counter
> * the item is still partially hidden behind the counter, is that the expected behavior?
> 
> A few screenshots to show some edge cases workspace-behind
> 
> item-behind
> 
> behind

Workspace footer was missing a `div`, the footer UI is now the same for workspaces, files and users.

client/src/views/workspaces/WorkspacesPage.vue
```diff
@@ -84,7 +84,9 @@
         </div>
       </div>
       <div class="workspaces-footer title-h5">
-        {{ $t('WorkspacesPage.itemCount', { count: workspaceList.length }, workspaceList.length) }}
+        <div class="folder-footer__container">
```
Max-7, line 87: Maybe not your fault, but it seems that the counter on the workspace page is black, whereas the counter on other pages (users or files) is gray.

Max-7 (changes requested):

Max-7 (approved):

PR #5710, opened 2023-11-02
feat: split ms-informative-text in 2 components (ms-report-text)

NicoTuxx:


client/src/components/core/ms-text/MsInformativeText.vue
```diff
@@ -22,80 +21,33 @@
...
-  closeCircle,
 } from 'ionicons/icons';
-import { MsTheme } from '@/components/core/ms-types';
 
 const props = defineProps<{
   icon?: string,
-  theme?: MsTheme,
 }>();
 
 function getIcon(): string {
```
Max-7, line 30: May as well remove the function and put the code directly into the template.

client/src/components/core/ms-text/MsInformativeText.vue
```diff
@@ -3,7 +3,6 @@
 <template>
   <div
     class="container-textinfo"
-    :class="props.theme ?? MsTheme.Basic"
   >
     <ion-icon
       :icon="getIcon()"
```
Max-7, line 8: Might as well remove the function call and put the code directly into the template.

PR #5715, opened 2023-11-03
[MS] Prevent electron from opening a second window

### LINKED ISSUE #5563: Restrain Electron GUI to only one instance

Max-7:
Closes #5563 

Tested only on Windows (since it requires that the package is built and installed, it's difficult to fully test otherwise). I don't see any reason why it wouldn't work on other platforms, it may not just be the best behavior for the OS (minimize/hide, tray icon, ...).

- [X] Describe your changes

### DIFF
--- a/client/electron/src/index.ts
+++ b/client/electron/src/index.ts
@@ -69,7 +69,6 @@ if (!lock) {
         myCapacitorApp.getMainWindow().webContents.send('open-link', lastArg);
       }
     }
-
   });
 }
 
--- a/client/electron/src/index.ts
+++ b/client/electron/src/index.ts
@@ -50,6 +50,29 @@ if (electronIsDev) {
   // autoUpdater.checkForUpdatesAndNotify();
 })();
 
+const lock = app.requestSingleInstanceLock();
+
+if (!lock) {
+  app.quit();
+} else {
+  app.on('second-instance', (_event, commandLine, _workingDirectory) => {
+    if (!myCapacitorApp.isMainWindowVisible()) {
+      myCapacitorApp.showMainWindow();
+    } else {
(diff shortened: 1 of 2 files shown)


client/electron/src/index.ts
```diff
@@ -50,6 +50,29 @@ if (electronIsDev) {
...
+    } else {
+      myCapacitorApp.getMainWindow().focus();
+    }
+
+    if (commandLine.length > 0) {
+      const lastArg = commandLine.at(-1);
+      // We're only interested in potential Parsec links
+      if (lastArg.startsWith('parsec://')) {
+        myCapacitorApp.getMainWindow().webContents.send('open-link', lastArg);
+      }
```
FirelightFlagboy, line 70: Electron provide a command line parser, I recommend to use it since an user can provide additional flag to tweak the electron behavior

https://www.electronjs.org/docs/latest/api/command-line-switches

Max-7: I don't see anything in this doc that is any help, at least not if we use the same format as V2 (where the link is given as a solo argument). Also keep in mind that this will never be called directly by anyone (outside of testing), it will be called by the `parsec://` protocol registration.

NicoTuxx: @Max-7 conclusion looks goot to me atm.

client/electron/src/index.ts
```diff
@@ -50,6 +50,29 @@ if (electronIsDev) {
...
+    }
+
```
NicoTuxx, line 72:
```suggestion
```

NicoTuxx (changes requested):

NicoTuxx (approved):

PR #5742, opened 2023-11-07
Use `FsPath` for `mountpoint`

TimeEngineer:


libparsec/crates/platform_mountpoint/src/memfs.rs
```diff
@@ -355,44 +333,35 @@ impl MountpointInterface for MemFS {
...
         let mut entries = self.entries.lock().expect("Mutex is poisoned");
 
-        if entries.keys().any(|entry| entry.parent() == Some(path)) {
+        if entries.keys().any(|entry| &entry.parent() == path) {
             return Err(MountpointError::DirNotEmpty);
         }
 
         let (parent, _) = entries
-            .get_mut(path.parent().expect("Can't be root"))
+            .get_mut(&path.clone().parent())
```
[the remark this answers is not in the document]
TimeEngineer: Done

libparsec/crates/platform_mountpoint/src/windows/mod.rs
```diff
@@ -510,3 +508,18 @@ pub fn mount<T: MountpointInterface>(
         .map(|fs| FileSystemMounted { fs })
         .map_err(|status| anyhow::anyhow!("Failed to init FileSystem {status}"))
 }
+
+fn os_path_to_fs_path(path: &U16CStr) -> Result<FsPath, NTSTATUS> {
+    path.to_os_string()
+        .to_str()
+        .map(|s| {
+            s.replace('\\', "/")
+                .split('/')
```
FirelightFlagboy, line 517:
```suggestion
            s.split('\\')
```

directly split by `\` ?

TimeEngineer: If your `path` contains '/' instead of `\`, it will fails, there you aggregate, it's not the same thing

TimeEngineer: I agree that currently `winfsp` will returns path only with `\`, but it's not my job to know precisely which separator the `OS` will use

libparsec/crates/platform_mountpoint/src/unix/inode.rs
```diff
@@ -2,26 +2,28 @@
...
+    paths_indexed_by_inode: Vec<FsPath>,
     stack_unused_inodes: Vec<usize>,
 }
 
 impl Default for PathsStore {
     fn default() -> Self {
         Self {
-            paths_indexed_by_inode: vec![PathBuf::from(""), PathBuf::from("/")],
+            paths_indexed_by_inode: vec![
+                "/".parse().expect("unreachable"),
```
touilleMan, line 24: Add a comment: `// Inode 0 is error prone, so we fill it with a dummy value`

libparsec/crates/platform_mountpoint/src/windows/mod.rs
```diff
@@ -239,10 +237,10 @@ impl<T: MountpointInterface> FileSystemContext for FileSystemWrapper<T> {
         // NOTE: The "." and ".." directories should ONLY be included
         // if the queried directory is not root
 
-        if fc.path() != Path::new("/") && marker.is_none() {
+        if fc.path() != &"/".parse().expect("unreachable") && marker.is_none() {
```
TimeEngineer, line 240: use is_root

touilleMan (approved):

PR #5691, opened 2023-11-02
[WIP] APIv4 server

touilleMan:

Issue #4607, opened 2023-05-26
Support proxy pac in `libparsec_platform_http_proxy` crate


touilleMan:
The issue with proxy pac is that it is an url to a javascript code.

So the idea would be to have the high level api (used by the GUI) define a function to register a callback when proxy pac config need to be executed. Given the new GUI is in javascript it's fairly simple for it to provide such callback, and it save us to embed another javascript runtime like we currently do in Python !

PR #5749, opened 2023-11-08
`Windows` register support for `libparsec_platform_http_proxy`

### LINKED ISSUE #4608: Add Windows register support for `libparsec_platform_http_proxy`
Currently we only rely on environment variable to configure the http proxy.

We want to also use the default proxy configuration provided by Windows.

TimeEngineer:
closes #4608


libparsec/crates/platform_http_proxy/src/lib.rs
```diff
@@ -1,13 +1,27 @@
...
 #[derive(Default, Debug, Clone)]
 pub struct ProxyConfig {
     #[cfg(not(target_arch = "wasm32"))]
     http_proxy: Option<reqwest::Proxy>,
     #[cfg(not(target_arch = "wasm32"))]
     https_proxy: Option<reqwest::Proxy>,
+    #[cfg(target_os = "windows")]
+    proxy_filter: Vec<String>,
+    #[cfg(target_os = "windows")]
+    proxy_script_url: String,
```
FirelightFlagboy, line 24: Convert the value directly to `request::Proxy`

libparsec/crates/platform_http_proxy/src/windows.rs
```diff
@@ -0,0 +1,77 @@
...
+        self.proxy_script_url = url;
+        self
+    }
+
+    fn with_proxy_filter(mut self, filter: String) -> Self {
+        self.proxy_filter = filter.split(';').map(|x| x.to_string()).collect();
+        self
+    }
+
+    pub fn with_register(self) -> anyhow::Result<Self> {
```
FirelightFlagboy, line 33: Should be called by `with_env_internal`

FirelightFlagboy, line 33: Should be internal to the crate

FirelightFlagboy (changes requested):

libparsec/crates/platform_http_proxy/src/lib.rs
FirelightFlagboy: `script_url` should be handled with the issue #4607, can you extract the change related to it ?

TimeEngineer: The problem is that I have to let some `bytes` unused in the `register`, also the issue is about to support the `windows register`, not only the `proxy` server

libparsec/crates/platform_http_proxy/src/windows.rs
TimeEngineer: tests using `with_env_internal` will fail on `windows` because we don't set `windows register` which need admin `privilege`

libparsec/crates/platform_http_proxy/src/lib.rs
TimeEngineer: > Convert the values directly to `request::Proxy` (for `proxy_filter`)

According to the doc: https://docs.rs/reqwest/latest/reqwest/struct.Proxy.html#method.no_proxy

TimeEngineer: Why do we have 2 proxies ? (`http` and `https` ?)

FirelightFlagboy: > Why do we have 2 proxies ? (http and https ?)

I took example of curl that handle `[scheme]_proxy` [curl env var](https://curl.se/libcurl/c/libcurl-env.html), some other softwares also handle those type env variables (`wget`, `python.request` if I'm not mistaken because it surely use `libcurl` internally)

 
---

> > Convert the values directly to `request::Proxy` (for `proxy_filter`)
>
> According to the doc: https://docs.rs/reqwest/latest/reqwest/struct.Proxy.html#method.no_proxy

We currently don't handle `no-proxy` so we can ignore that for the moment (or provide support for the `NO_PROXY` env var)

---

> The problem is that I have to let some bytes unused in the register

We do you mean about `unused bytes in the register` ?

---

> the issue is about to support the windows register, not only the proxy server

Since we have another issue about supporting proxy pac, I would have expected that handling this case to be done on that issue.

TimeEngineer: > We do you mean about `unused bytes in the register` ?

I mean this: https://github.com/Scille/parsec-cloud/pull/5749/files#diff-36d928dd01fae253e0c4f4b61a386f48de34e1ccdd266e3c0be6858bb5bf55f4R66-R73

libparsec/crates/platform_http_proxy/src/wasm.rs
```diff
@@ -0,0 +1,18 @@
+// Parsec Cloud (https://parsec.cloud) Copyright (c) BUSL-1.1 2016-present Scille SAS
```
touilleMan, line 1: Rename this in `web.rs`

TimeEngineer: Done

libparsec/crates/platform_http_proxy/src/windows.rs
```diff
@@ -0,0 +1,68 @@
...
+};
+
+use libparsec_types::anyhow;
+
+use crate::ProxyConfig;
+
+impl ProxyConfig {
+    pub(crate) fn with_register(self) -> anyhow::Result<Self> {
+        // Safety: Hope all is safe
+        unsafe {
```
FirelightFlagboy, line 19: Should we use multiple smaller `unsafe` block ?

libparsec/crates/platform_http_proxy/src/windows.rs
```diff
@@ -0,0 +1,68 @@
...
+    System::Memory::GlobalFree,
+};
+
+use libparsec_types::anyhow;
+
+use crate::ProxyConfig;
+
+impl ProxyConfig {
+    pub(crate) fn with_register(self) -> anyhow::Result<Self> {
+        // Safety: Hope all is safe
```
FirelightFlagboy, line 18: The comment is not helpful at all

FirelightFlagboy (changes requested):

libparsec/crates/platform_http_proxy/src/native/windows.rs
```diff
@@ -0,0 +1,80 @@
...
+                    "WinHttpGetProxyForUrl failed with code: {}",
+                    GetLastError()
+                ));
+            }
+        }
+
+        // Proxy Auto Config enabled
+        if !proxy_config.lpszAutoConfigUrl.is_null() {
+            // Safety: According to Windows doc, we should free it
+            // https://learn.microsoft.com/en-us/windows/win32/api/winhttp/nf-winhttp-winhttpgetieproxyconfigforcurrentuser#remarks
```
FirelightFlagboy, line 40: That is not a reason why it's safe, what's important here is we don't free something that we don't own or something that will be used after (Use after free vuln).

libparsec/crates/platform_http_proxy/src/native/windows.rs
```diff
@@ -0,0 +1,80 @@
...
+            unsafe {
+                GlobalFree(proxy_config.lpszProxy as isize);
+            }
+        }
+
+        // Proxy Bypass enabled
+        // TODO: We don't handle ProxyBypass for the moment
+        if !proxy_config.lpszProxyBypass.is_null() {
+            // Safety: According to Windows doc, we should free it
+            // https://learn.microsoft.com/en-us/windows/win32/api/winhttp/nf-winhttp-winhttpgetieproxyconfigforcurrentuser#remarks
```
FirelightFlagboy, line 72: CF previous comment about use after free.

libparsec/crates/platform_http_proxy/src/native/windows.rs
```diff
@@ -0,0 +1,80 @@
...
+        let mut cfg = self;
+
+        let mut proxy_config = WINHTTP_CURRENT_USER_IE_PROXY_CONFIG {
+            fAutoDetect: 0,
+            lpszAutoConfigUrl: std::ptr::null_mut(),
+            lpszProxy: std::ptr::null_mut(),
+            lpszProxyBypass: std::ptr::null_mut(),
+        };
+
+        // Safety: Nothing unsafe here
```
FirelightFlagboy, line 27: Improve the comment, why is it safe here ? (hint we initialize the struct `proxy_config` at the previous step with sensible value)

TimeEngineer: which sensible value ?

libparsec/crates/platform_http_proxy/src/native/windows.rs
TimeEngineer: It's safe because the `Windows API` allocated the data

libparsec/crates/platform_http_proxy/src/native/windows.rs
```diff
@@ -0,0 +1,80 @@
...
+            } else if proxy.starts_with("http") {
+                cfg.with_http_proxy(proxy)
+            } else {
+                return Err(anyhow::anyhow!(
+                    "Proxy server doesn't start with http or https"
+                ));
+            }?;
+
+            // Safety: According to Windows doc, we should free it
+            // https://learn.microsoft.com/en-us/windows/win32/api/winhttp/nf-winhttp-winhttpgetieproxyconfigforcurrentuser#remarks
```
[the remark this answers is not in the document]
TimeEngineer: Not on `windows`, for example when I used the `winfsp api`, the application can crash due to leak

libparsec/crates/platform_http_proxy/src/native/windows.rs
FirelightFlagboy: `proxy_config` being initialize with null ptr for pointer data instead of a random value.

libparsec/crates/platform_http_proxy/src/native/windows.rs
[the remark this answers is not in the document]
FirelightFlagboy: You free it, but you also need to unsure that the pointer is used after freeing it.

libparsec/crates/platform_http_proxy/src/native/windows.rs
TimeEngineer: We don't care about the parameters, because this is the output

TimeEngineer: The only thing we should ensure is that the `pointer` provided is a `WINHTTP_CURRENT_USER_IE_PROXY_CONFIG` for the correct `size` but I don't cast anything (no other `unsafe block`) and `rust` provides the check on `type`

libparsec/crates/platform_http_proxy/src/native/windows.rs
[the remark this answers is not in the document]
TimeEngineer: Do you want I call explicitely the drop ?

libparsec/crates/platform_http_proxy/src/native/windows.rs
FirelightFlagboy: The last comment would make a nice safety comment ;)

vxgmichel (approved):

libparsec/crates/platform_http_proxy/src/native/windows.rs
TimeEngineer: Done

libparsec/crates/platform_http_proxy/src/native/windows.rs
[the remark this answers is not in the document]
TimeEngineer: This is because I don't use it, that I call the `free`, maybe you want I put at the end of the function ? But it will need more `if` statement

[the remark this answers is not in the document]
FirelightFlagboy: As an afterthought, the discussion wasn't clear from the start :sweat_smile: .

What I want is a better safety comment that state 

> We free a pointer that was allocated by windows and we don't use it after.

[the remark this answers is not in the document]
TimeEngineer: Done

libparsec/crates/platform_http_proxy/src/native/windows.rs
```diff
@@ -0,0 +1,87 @@
...
+            // According to Windows doc, we should free it.
+            // https://learn.microsoft.com/en-us/windows/win32/api/winhttp/nf-winhttp-winhttpgetieproxyconfigforcurrentuser#remarks
+            unsafe {
+                GlobalFree(proxy_config.lpszAutoConfigUrl as isize);
+            }
+        }
+
+        // Proxy enabled
+        if !proxy_config.lpszProxy.is_null() {
+            // Safety: According to Windows doc, this parameter is a nul terminated widestring
```
FirelightFlagboy, line 52:
```suggestion
            // Safety: The ptr is allocated by windows and is not null.
            // According to Windows doc, this parameter is a nul terminated widestring
```

It's unsafe because we need to unsure we provided a correct ptr to `from_ptr_str`

TimeEngineer: This comment is useless because of the previous `if`

FirelightFlagboy: The goal of the `SAFETY` comment is to document why the `unsafe` block below it is safe.

In this case it is because the `ptr` was allocated by windows and we check that it's not `null`.

I don't find that useless

TimeEngineer: If I say that this parameter is a nul terminated widestring, I assume the pointer is not null

mmmarcos:

libparsec/crates/platform_http_proxy/Cargo.toml
```diff
@@ -14,4 +14,8 @@ workspace = true
 [dependencies]
 libparsec_types = { workspace = true }
 
-reqwest = { workspace = true }
+reqwest = { workspace = true, features = ["default-tls"] }
```
touilleMan, line 17: why this feature here ? it should be up to `libparsec_client_connection` to choose which tls flavor to rely on

FirelightFlagboy (approved):

libparsec/crates/platform_http_proxy/src/windows.rs
```diff
@@ -0,0 +1,77 @@
...
+    }
+
+    pub fn with_register(self) -> anyhow::Result<Self> {
+        let mut cfg = self;
+
+        let hkcu = RegKey::predef(HKEY_CURRENT_USER);
+        let proxy_register_path = hkcu.open_subkey(PROXY_REGISTER_PATH)?;
+        let proxy_register_value = proxy_register_path.get_raw_value(PROXY_KEY)?;
+
+        let (signature, others) = proxy_register_value.bytes.split_at(12);
```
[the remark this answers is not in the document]
FirelightFlagboy: That code no longer exist

libparsec/crates/platform_http_proxy/Cargo.toml
TimeEngineer: Done

PR #5751, opened 2023-11-09
Bump rust to 1.72.1

FirelightFlagboy:


libparsec/crates/testbed/src/template/events.rs
```diff
@@ -85,6 +85,7 @@ macro_rules! impl_certificates_meth_for_single_certificate {
             ) -> impl Iterator<Item = TestbedTemplateEventCertificate> + 'c {
                 std::iter::once(()).map(move |_| {
                     let mut guard = self.cache.lock().expect("Mutex is poisoned");
+                    #[allow(clippy::redundant_closure_call)]
```
TimeEngineer, line 88: Can we fix it ?

FirelightFlagboy: Instead of

```rust
#[allow(clippy::redundant_closure_call)]
let populate = || $populate(self, template);
```

I can do

```rust
let populate_fn = $populate;
let populate = || populate_fn(self, template);
```

FirelightFlagboy: FYI, the warning is raise because the expension would look like something:

```rust
let populate || (|foo: u32, bar: String| format!("{bar}-{foo}"))(42,"anwser".into());
```

Cargo.toml
```diff
@@ -39,6 +39,7 @@ default-members = [
     "libparsec/crates/tests_macros",
     "libparsec/crates/types",
 ]
+resolver = "2"
```
mmmarcos, line 42: According to the [docs](https://doc.rust-lang.org/edition-guide/rust-2021/default-cargo-resolver.html), using `edition = 2021` implies `resolver = 2`, so this line is redundant

FirelightFlagboy: Still in the doc:

> If you are using a [virtual workspace](https://doc.rust-lang.org/cargo/reference/workspaces.html#virtual-workspace), 
> you will still need to explicitly set the [resolver field](https://doc.rust-lang.org/cargo/reference/resolver.html#resolver-versions) in the `[workspace]` definition if you want to opt-in to the new resolver.

mmmarcos: So maybe we can add a comment in case someone has the same doubt? 
```toml
# Set the resolver explicitly becase we are in a [workspace] definition
# See: https://doc.rust-lang.org/cargo/reference/workspaces.html#virtual-workspace
resolver = "2"
```

TimeEngineer: It will warns you anyway if you forget it

mmmarcos: Really? well in that case, no need for the comment I guess

libparsec/crates/testbed/src/template/events.rs
TimeEngineer: Can you fix it ?

FirelightFlagboy: Yes, I just didn't have time :sweat_smile:

FirelightFlagboy: done

mmmarcos (approved):

Max-7 (approved):

TimeEngineer (approved):

PR #5758, opened 2023-11-09
[MS] Added prettier to format code

Max-7:
Adds `prettier` to format the code (and ensures the code is formatted in CI).

ESLint is not really made to format the code, its role is to check for errors. It is possible to add formatting rules, and we do have some, but it's bothersome and it's easy to forget some cases.

There's also a VS extension, I haven't looked at it yet.

- [X] Describe your changes


.pre-commit-config.yaml
```diff
@@ -201,6 +201,18 @@ repos:
...
+    rev: v3.0.3
+    hooks:
+      - id: prettier
+        name: prettier
+        entry: prettier
+        files: ^client/
+        types_or: [ vue, ts, tsx, javascript, jsx, html ]
+        args: [ --fix ]
+        stages:
+          - manual
```
FirelightFlagboy, line 214: That is nice to have added a configuration for `prettier` on `pre-commit`, but it will likely be forgotten (since dependabot will update `package.json` and not `pre-commit-config.yaml`).

I would recommend either:

- Doing something like `ruff` or `mypy` on `pre-commit` (i.e.: running the command `npx prettier . --fix`).
- Handling the version of the hook using `version_udpater` script.

.pre-commit-config.yaml
```diff
@@ -201,6 +201,18 @@ repos:
           - eslint-plugin-promise@6.1.1
           - eslint-plugin-vue@8.7.1
 
+  - repo: https://github.com/pre-commit/mirrors-prettier
+    rev: v3.0.3
+    hooks:
+      - id: prettier
+        name: prettier
+        entry: prettier
+        files: ^client/
```
FirelightFlagboy, line 210: We have js & ts files in `bindings`, I recommend to include those.

client/.prettierignore
FirelightFlagboy, line 1: Can you add comment about why we ignore some file/folder ?

I was thinking of:

- `electron`: It contain ts files
- `scripts`: Simpler to just format everything using the same style (cf `rustfmt`)
- `android`: just stating that folder contain only `java` files is enough

client/.prettierrc
FirelightFlagboy, line 1: I would recommend renaming this file to `.prettierrc.json5` (that would provide syntax highlight on github and allow for comment in the file)

Or convert it to yaml (`.prettierrc.yml`)

source: https://prettier.io/docs/en/configuration

client/src/locales/en-US.json
FirelightFlagboy, line 1: spacing conflict with editor config (`2` vs `4`)

client/src/locales/fr-FR.json
FirelightFlagboy, line 1: spacing conflict with editor config (`2` vs `4`)

client/src/plugins/libparsec/definitions.ts
FirelightFlagboy, line 1: This file is auto-generate, the template should be updated alongside of it.

FirelightFlagboy (commented):
`scss` files aren't configured on `editorconfig`, can you fix that ?

(You just need to add the pattern alongside the other `css`, `vue` pattern on L31)

client/tsconfig.eslint.json
FirelightFlagboy, line 1: conflict with spacing (`2` vs `4`)

client/tsconfig.json
FirelightFlagboy, line 1: conflict with spacing (`2` vs `4`)

.pre-commit-config.yaml
Max-7: I think it's all boilerplate. And it would mean that
- either I add configuration related to the bindings in `client/.prettierignore` which is weird
- put the prettier config directly into parsec-cloud's root folder.
Both can be done but I'd rather use prettier only for the client for now and change it later if we really need it.

client/.prettierignore
Max-7: I included a bit more folders and added a few comments.

client/ionic.config.json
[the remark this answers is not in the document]
Max-7: I excluded json files from prettier in the client.

client/src/common/mocks.ts
```diff
@@ -9,105 +9,114 @@ export function getAppVersion(): string {
...
+  version: string;
+  date: DateTime;
+  features: Change[];
+  fixes: Change[];
+  misc: Change[];
 }
 
 // Added multiple values for testing
 // can be removed before release
 export function getChanges(): VersionChange[] {
```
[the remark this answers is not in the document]
Max-7: No, but I'm not sure we really know how we want to handle the changelog for now.

client/src/plugins/libparsec/definitions.ts
Max-7: I'm just excluding `src/plugins` from prettier.

client/.prettierignore
```diff
@@ -0,0 +1,18 @@
...
+scripts
+# Node files
+node_modules
+package.json
+package-lock.json
+# Leave JSON alone
+**/*.json
+**/*.json5
+# Auto-generated
+src/plugins
```
FirelightFlagboy, line 18:
```suggestion
src/plugins/libparsec/definitions.ts
```

Only `client/src/plugins/libparsec/definitions.ts` is auto generated

client/.prettierrc.json5
FirelightFlagboy, line 1: Spacing conflict (`2` vs `4`)

FirelightFlagboy, line 1: For json files and alike, you could configure an [override](https://prettier.io/docs/en/configuration#configuration-overrides) ?

.github/workflows/ci-web.yml
```diff
@@ -63,6 +63,11 @@ jobs:
         working-directory: client
         timeout-minutes: 5
 
+      - name: Check prettier
+        run: npx prettier . --check
```
FirelightFlagboy, line 67: This is different than `pre-commit run prettier --all-files`, the later command would include the javascript files under `bindings`.

Max-7: Yeah, I removed `bindings` from the pre-commit. Let's keep it simple with only the client folder for now.

FirelightFlagboy: Will likely be forgotten when we add bindings.

Max-7: Yeah but it's all auto-generated stuff or boilerplate that will never change. I would have to move prettier config from the client folder to the root folder, and if prettier finds faults with the generated code, I would also have to change the template. I can open an issue if we really want prettier on the code in bindings, but honestly, it's probably just wasted time.

.github/workflows/ci-web.yml
```diff
@@ -63,6 +63,11 @@ jobs:
...
+        working-directory: client
+        timeout-minutes: 5
```
FirelightFlagboy, line 69: We have a step `Check lint` that run the command `npm run lint` should `prettier` run at that step ?

client/src/common/mocks.ts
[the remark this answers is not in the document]
FirelightFlagboy: I've create the issue #5761 to trace that.

.github/workflows/ci-web.yml
Max-7: Does it change anything? If it all the same, we can keep them appart, they're not exactly the same thing.

client/src/common/asyncComputed.ts
```diff
@@ -4,13 +4,13 @@
...
 export interface AsyncComputedOptions {
-  lazy?: boolean
-  evaluating?: Ref<boolean>
-  shallow?: boolean
-  onError?: (e: unknown) => void
+  lazy?: boolean;
+  evaluating?: Ref<boolean>;
+  shallow?: boolean;
+  onError?: (e: unknown) => void;
 }
```
NicoTuxx, line 14: Can we update prettier rules to avoid this horendous format style for interfaces?

.github/workflows/ci-web.yml
NicoTuxx: I would rather keep the 2 distinct commands; `npm run lint`, already execute 2 distinct tools, it's enough.
And the roles are not identical:
- ESLint is here to watch syntax rules and check if we apply them.
- prettier is more a re-write tool, that will basically help us to automatically change wrong syntax.

NicoTuxx (changes requested):

client/src/common/asyncComputed.ts
Max-7: I haven't seen anything for that. I kinda like it this way but I'm a C and C++ guy, I like `;` after my attributes.
Prettier configuration is very minimal, which is a good thing, we don't want too many options.

NicoTuxx (changes requested):
We should add and use:
 - https://github.com/prettier/prettier-eslint

client/src/common/asyncComputed.ts
NicoTuxx: LGTM

NicoTuxx (approved):

.github/workflows/ci-web.yml
FirelightFlagboy, line 69: Move that step to be near `check lint` step

Max-7: What do you mean? It's right under the link step.

FirelightFlagboy: I thought the step wasn't following `check lint`, I should have double check

FirelightFlagboy (approved):

PR #5768, opened 2023-11-10
Normalize values to avoid overflow issues with the remanence widget progress bar

### LINKED ISSUE #5759: [🐛 | Bug]: Exception raised from remanence management widget
### Parsec version tested on:

2.16.0

### Platforms tested on:

Linux

### Bug description:

The value used in progress bar is not bounded: https://github.com/Scille/parsec-cloud/blob/releases/2.16/parsec/core/gui/remanence_management_widget.py#L111

This problem has already been seen (and fixed) before: https://github.com/Scille/parsec-cloud/blob/releases/2.16/parsec/core/gui/loading_widget.py#L17-L20

### Relevant output:

```
OverflowError: argument 1 overflowed: value must be in the range -2147483648 to 2147483647
  File "parsec\core\gui\workspaces_widget.py", line 343, in _on_manage_remanence_clicked
  File "parsec\core\gui\remanence_management_widget.py", line 190, in show_modal
  File "parsec\core\gui\remanence_management_widget.py", line 40, in __init__
  File "parsec\core\gui\remanence_management_widget.py", line 111, in _update_text
```

vxgmichel:
Fix #5759 

### DIFF
--- a/parsec/core/gui/remanence_management_widget.py
+++ b/parsec/core/gui/remanence_management_widget.py
@@ -19,6 +19,14 @@ from parsec.core.gui.ui.remanence_management_widget import Ui_RemanenceManagemen
 from parsec.core.logged_core import LoggedCore
 
 
+def normalize_progress(value: int, total: int, maxi: int = 0x7FFFFFFF) -> tuple[int, int]:
+    if total <= maxi:
+        return value, total
+    assert total != 0
+    normalized_value = int(round(value * maxi / total))
+    return normalized_value, maxi
+
+
 class RemanenceManagementWidget(QWidget, Ui_RemanenceManagementWidget):
     def __init__(self, jobs_ctx: QtToTrioJobScheduler, core: LoggedCore, workspace_fs: WorkspaceFS):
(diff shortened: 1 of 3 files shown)


FirelightFlagboy (approved):

parsec/core/gui/remanence_management_widget.py
```diff
@@ -19,6 +19,16 @@
...
 
 
+def normalize_progress(value: int, total: int, maxi: int = 0x7FFFFFFF) -> tuple[int, int]:
+    if total <= maxi:
+        return value, total
+    assert total != 0
+    normalized_total = min(total, maxi)
+    factor = normalized_total / total
+    normalized_value = int(round(value * factor))
+    return normalized_value, normalized_total
```
mmmarcos, line 29: You shouldn't need `normalized_total` here because of the check in line 23 (`total` is always greater than `maxi` here) 
```suggestion
    factor = maxi / total
    normalized_value = int(round(value * factor))
    return normalized_value, maxi
```

PR #5769, opened 2023-11-10
Unify French translation of "device" in the docs

mmmarcos:
The French docs translates "devices" sometimes as "_appareils_" and sometimes as "_terminaux_". 

This PR unifies the translation to "_appareils_" (rationale: only "_appareils_" is used in the GUI)


FirelightFlagboy (approved):
The rational is that we have a glossary that state which word to use and not to use (https://github.com/Scille/parsec-cloud/blob/master/docs/development/glossary.md)

PR #5787, opened 2023-11-14
[Backport] Prevent concurrent run of `publish` workflow

FirelightFlagboy:


.github/workflows/publish.yml
```diff
@@ -12,6 +12,13 @@ on:
       - published
       - edited
 
+# We set `concurrency` to prevent having this workflow being run on code that is not up-to-date on a PR (a user make multiple push in a quick manner).
```
mmmarcos, line 15:
```suggestion
# Prevent concurrent run of `publish` workflow
# We  set `concurrency` to prevent having this workflow being run on code that is not up-to-date on a PR (a user make multiple push in a quick manner).
```

.github/workflows/publish.yml
```diff
@@ -12,6 +12,13 @@ on:
...
+# But on the main branch, we don't want that behavior.
+# Having the workflow run on each merge commit is something we would like, that could help us where a regression was made and missed by previous checks.
```
mmmarcos, line 17:
```suggestion
# This does not apply to the `master` branch where we want the workflow to run on each merge commit to help us identify regressions missed by previous checks.
```

mmmarcos (approved):
LGTM. I would just change comment a little bit.

.github/workflows/publish.yml
FirelightFlagboy: I've simplified the comment

.github/workflows/publish.yml
FirelightFlagboy: I've simplified the comment

mmmarcos (approved):

PR #5792, opened 2023-11-14
Fix `check_newsfragments` should compare against base

FirelightFlagboy:
The script `check_newsfragments` use `origin/master` as the base target (where the pull is targeted) but `origin/master` is not always our target.

For example, if we work on a release branch or on the current `dev/v2` branch, that check don't work if `origin/master` already has the same newsfragment

Other Changes
-------------

- The script no longuer check if it need to check if it's on a branch that should be skipped (e.g.: `master`)

  The check is redundant with github actions and require to maintain 2 source of truth.

- Simplify the condition to run the `check_newsfragments`:

  - Use `event_name == 'pull_request'` instead of looking for the prefix `ref/pull`.
  - Remove the exception for `yolo` prefixed branch as we are no longer required to add a newsfragment.

### DIFF
--- a/.github/scripts/check_newsfragments.py
+++ b/.github/scripts/check_newsfragments.py
@@ -6,30 +6,23 @@ PRs must contain a newsfragment that reference an opened issue
 
 import argparse
 import json
-import re
 from concurrent.futures import ThreadPoolExecutor
 from pathlib import Path
 from subprocess import run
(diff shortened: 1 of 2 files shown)


vxgmichel (approved):

mmmarcos (approved):
LGTM

PR #5797, opened 2023-11-14
`Windows` `pac` support for `libparsec_platform_http_proxy`

### LINKED ISSUE #4607: Support proxy pac in `libparsec_platform_http_proxy` crate

TimeEngineer:
Depends on #5749 

closes #4607


mmmarcos:

FirelightFlagboy (approved):

PR #5802, opened 2023-11-15
Configure nextest timeout for slow tests

FirelightFlagboy:
Add nextest configuration:

- Change the default configuration for `slow-timeout`

  The default display a message `SLOW {TEST}` every `60s` without terminating it.
  Now the message is displayed every `10s` and stoppind the test at the
  3rd warning.

- Add profile `ci` to be used in the CI:
  - Laxer `slow-timeout`

    Stop the test after 5 slow message (instead of 3).

  - Show test failure outputs immediatly and at the end of the run
    (default to `immediate` which work well when `fail-fast` is enabled).
  - Disable fail fast (enabled in `default` profile).

Full nextest configuration options can be found here: <https://nexte.st/book/configuration.html>

### DIFF
--- a/.config/nextest.toml
+++ b/.config/nextest.toml
@@ -1,0 +1,14 @@
+[profile.default]
+# https://nexte.st/book/slow-tests.html
+slow-timeout = { period = "10s", terminate-after = 3 } # Timeout a test after 30s (3 * 10s)
+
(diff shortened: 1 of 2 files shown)


mmmarcos (approved):
👍

PR #5808, opened 2023-11-15
Add `cli` & `list-devices`

TimeEngineer:


libparsec/Cargo.toml
```diff
@@ -36,3 +36,8 @@ libparsec_serialization_format = { workspace = true }
 libparsec_testbed = { workspace = true, optional = true }
 
 dirs = { workspace = true }
+
+# CLI dependencies
```
FirelightFlagboy, line 40: Should we have a separate bin crate `libparsec_cli` ?

Currently `libparsec` is used as a point of centralization of crate to import and is used notably for wasm32 bindings which won't support tokio.

libparsec/src/lib.rs
```diff
@@ -2,6 +2,7 @@
 
 mod addr;
 mod cancel;
+pub mod cli;
```
FirelightFlagboy, line 5: We probably don't want the cli to be included in the lib :thinking:

libparsec/src/cli/mod.rs
```diff
@@ -0,0 +1,9 @@
+// Parsec Cloud (https://parsec.cloud) Copyright (c) BUSL-1.1 2016-present Scille SAS
+
+mod list_devices;
+
+pub use list_devices::list_devices;
+
+const GREEN: &str = "\x1B[92m";
+const RESET: &str = "\x1B[39m";
+const YELLOW: &str = "\x1B[33m";
```
FirelightFlagboy, line 9: For writing a tui, I would recommend using one of:

- [tui](https://crates.io/crates/tui): Use `crossterm` as backend but can use `termion`. 
- [crossterm](https://crates.io/crates/tui): Support all platform (Window, Unix).
- [termion](https://crates.io/crates/termion): Seems to support only Unix.

libparsec/src/cli/list_devices.rs
```diff
@@ -0,0 +1,20 @@
+// Parsec Cloud (https://parsec.cloud) Copyright (c) BUSL-1.1 2016-present Scille SAS
+
+use std::path::Path;
+
+use crate::{cli::*, list_available_devices};
+
+pub async fn list_devices(config_dir: &Path) {
+    let devices = list_available_devices(config_dir).await;
+    let n = devices.len();
+    let slug_len = 2 + (n + 1).ilog2() as usize;
```
FirelightFlagboy, line 10: Why we limit the len of the slug by the number of devices ?

libparsec/src/bin/cli.rs
```diff
@@ -0,0 +1,30 @@
...
+#[derive(Parser)]
+struct Arg {
+    #[command(subcommand)]
+    command: Command,
+}
+
+#[derive(Subcommand)]
+enum Command {
+    /// List all devices
+    ListDevices { config_dir: Option<PathBuf> },
```
FirelightFlagboy, line 18: use a struct for the argument of `ListDevices`, that allow to define the struct on `cli::list_devices`

libparsec/Cargo.toml
TimeEngineer: I can configure for `not(wasm32)` this is not a problem

libparsec/src/lib.rs
TimeEngineer: Why ?

libparsec/src/cli/mod.rs
TimeEngineer: It depends if I need more. For the moment it's only `3` lines

libparsec/src/cli/list_devices.rs
TimeEngineer: I forgot the comment there

libparsec/src/bin/cli.rs
TimeEngineer: Done

libparsec/src/cli/list_devices.rs
TimeEngineer: Done

libparsec/Cargo.toml
FirelightFlagboy: Yes with `not(wasm32)` that would solve that issue.

Now I think about the electron binding, with feature normalization that may cause problem in the future (especially with `tokio`)

libparsec/src/lib.rs
FirelightFlagboy: The lib is not an executable so the CLI function would not be used.

FirelightFlagboy (commented):
Would be nice if we had tests for the CLI

libparsec/Cargo.toml
TimeEngineer: This can be also done under a feature

libparsec/src/lib.rs
TimeEngineer: This is not the `cli`, but more like `cli_utils`

libparsec/src/cli/list_devices.rs
```diff
@@ -0,0 +1,22 @@
...
+    config_dir: Option<PathBuf>,
+}
+
+pub async fn list_devices(list_devices: ListDevices) {
+    let config_dir = list_devices.config_dir.unwrap_or(get_default_config_dir());
+    let devices = list_available_devices(&config_dir).await;
+    let n = devices.len();
+
+    println!("Found {GREEN}{n}{RESET} device(s) in {YELLOW}{config_dir:?}{RESET}:");
+    format_devices(&devices)
```
mmmarcos, line 21: Shouldn't we have an if/else in case of no device (`n == 0`)? 
So we can print `"No devices found in ..."`

mmmarcos (approved):
LGTM. 

There's still this comment though https://github.com/Scille/parsec-cloud/pull/5808#discussion_r1397476693

parsec_cli/src/list_devices.rs
```diff
@@ -0,0 +1,28 @@
...
+
+use libparsec::{get_default_config_dir, list_available_devices};
+
+use crate::utils::*;
+
+#[derive(Args)]
+pub struct ListDevices {
+    /// Parsec config directory
+    #[arg(short, long)]
+    config_dir: Option<PathBuf>,
```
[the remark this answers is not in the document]
TimeEngineer: First one

FirelightFlagboy (approved):

parsec_cli/Cargo.toml
```diff
@@ -0,0 +1,20 @@
+[package]
+name = "parsec_cli"
+publish.workspace = true
+authors.workspace = true
+edition.workspace = true
+homepage.workspace = true
+license.workspace = true
+version.workspace = true
```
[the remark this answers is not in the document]
TimeEngineer: I think the idea is to release a first version when all is ready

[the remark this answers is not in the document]
FirelightFlagboy: I've create the issue #5896 to discuss about that

FirelightFlagboy (approved):

PR #5810, opened 2023-11-16
Further fix to `check_newsfragments` script

FirelightFlagboy:
- Improve logging of newsfragment (display the `git log` command that will be executed).
- Prefix `base` argument with `origin/`.
- Use repeat on base argument to prevent iterating over each char of the string.
- Allow closed issue/PR.
- Fetch `base_ref` before executing the script.

### DIFF
--- a/.github/scripts/check_newsfragments.py
+++ b/.github/scripts/check_newsfragments.py
@@ -1,5 +1,7 @@
 # Parsec Cloud (https://parsec.cloud) Copyright (c) BUSL-1.1 2016-present Scille SAS
 
+# cspell:words oneline
+
 """
 PRs must contain a newsfragment that reference an opened issue
 """
@@ -7,6 +9,7 @@ PRs must contain a newsfragment that reference an opened issue
 import argparse
 import json
 from concurrent.futures import ThreadPoolExecutor
+from itertools import repeat
 from pathlib import Path
 from subprocess import run
 from typing import Optional
@@ -18,11 +21,13 @@ VALID_TYPE = ["feature", "bugfix", "doc", "removal", "api", "misc", "empty"]
 
 def check_newsfragment(fragment: Path, base: str) -> Optional[bool]:
     fragment_name = fragment.name
+
     # If file never existed in `base`, consider as a new newsfragment
     # Cannot just git diff against `base` branch here given newsfragments
     # removed in master will be considered as new items in our branch
     # --exit-code makes the command exit with 0 if there are changes
-    cmd_args = ["git", "log", base, "--exit-code", "--", str(fragment)]
+    cmd_args = ["git", "log", base, "--exit-code", "--oneline", "--", str(fragment)]
+    print(f"[{fragment_name}] Checking news fragment on base {base} with `{' '.join(cmd_args)}`")
     ret = run(cmd_args, capture_output=True)
 
     if ret.returncode == 0:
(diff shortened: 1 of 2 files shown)


.github/scripts/check_newsfragments.py
```diff
@@ -78,7 +76,11 @@ def check_newsfragment(fragment: Path, base: str) -> Optional[bool]:
         ret = list(
             filter(
                 lambda value: value is not None,
-                pool.map(check_newsfragment, Path("newsfragments").glob("*.rst"), args.base),
+                pool.map(
+                    check_newsfragment,
+                    Path("newsfragments").glob("*.rst"),
+                    repeat("origin/" + args.base),
```
mmmarcos, line 82: Wouldn't be better to include the `"origin/"` in the `args.base` when invoking the script? 

I don't think we'll need to change it, but that way we avoir the hard-coded string and we already specify "origin" in https://github.com/Scille/parsec-cloud/pull/5810/files#diff-b803fcb7f17ed9235f1e5cb1fcd2f5d3b2838429d4368ae4c57ce4436577f03fR162

.github/scripts/check_newsfragments.py
```diff
@@ -50,19 +55,12 @@ def check_newsfragment(fragment: Path, base: str) -> Optional[bool]:
...
-            if "pull_request" in data:
-                print(
-                    f"[{fragment_name}] fragment ID correspond to a pull request instead of an issue !"
-                )
-            else:
-                if data["state"] == "open":
-                    return True
-                else:
-                    print(f"[{fragment_name}] fragment ID correspond to a closed issue !")
-        return False
```
mmmarcos, line 65: No need to check if it's PR or closed issue?

.github/scripts/check_newsfragments.py
```diff
@@ -50,19 +55,12 @@ def check_newsfragment(fragment: Path, base: str) -> Optional[bool]:
             print(
                 f"[{fragment_name}] fragment ID doesn't correspond to an issue ! (On <{url}> HTTP error {exc.code} {exc.reason})"
             )
+            return False
         else:
+            # Sanity check is we can deserialize the response.
```
mmmarcos, line 60: The sanity check IS the deserialization? If so, I'd rephrase it like this:
```suggestion
            # Sanity check: try to deserialize the response.
```

.github/scripts/check_newsfragments.py
```diff
@@ -50,19 +55,12 @@ def check_newsfragment(fragment: Path, base: str) -> Optional[bool]:
             print(
                 f"[{fragment_name}] fragment ID doesn't correspond to an issue ! (On <{url}> HTTP error {exc.code} {exc.reason})"
```
mmmarcos, line 56:
```suggestion
                f"[{fragment_name}] fragment ID doesn't correspond to an issue! (On <{url}> HTTP error {exc.code} {exc.reason})"
```

.github/scripts/check_newsfragments.py
FirelightFlagboy: I've chosen to add the prefix in the script because, it's require for `git log` so everytime you use that script, you need to specify `origin/`.
I find it simpler that way.

CF: the failing run that use `base` without the prefix https://github.com/Scille/parsec-cloud/actions/runs/6888409905/job/18737350209

.github/scripts/check_newsfragments.py
FirelightFlagboy: That check is problematic when you do a backport that include a newsfragment for an issue.

> That linked issue is likely already closed by the first PR that targeted the main branch.

IMO, I don't find that check relevant: 

1. Here we want to document an issue that as been resolve and link it in the changelog so it as nothing to do if the issue is open or not.
2. If later on we want to document additional issues that as been close in the changelog, that's not possible with that check.

mmmarcos (approved):

PR #5812, opened 2023-11-16
5750 functions to manipulate paths in bindings

FirelightFlagboy:


libparsec/crates/types/src/fs_path.rs
```diff
@@ -248,6 +248,22 @@ impl FsPath {
...
+            match part.0.as_str() {
+                "." | "" => continue,
+                ".." => {
+                    new_parts.pop();
+                }
+                _ => new_parts.push(part.clone()),
+            }
+        }
+        Self { parts: new_parts }
+    }
```
TimeEngineer, line 266: Maybe we don't need to allocate a new `path`, it is better if we have the normalization. WDYT ?

```suggestion
    /// Normalize the path
    pub fn normalize(self) -> Self {
        let mut new_parts = Vec::with_capacity(self.parts.len());

        for part in &self.parts {
            match part.0.as_str() {
                "." | "" => continue,
                ".." => {
                    new_parts.pop();
                }
                _ => new_parts.push(part),
            }
        }
        Self { parts: new_parts }
    }
```

libparsec/src/path.rs
```diff
@@ -0,0 +1,23 @@
+// Parsec Cloud (https://parsec.cloud) Copyright (c) BUSL-1.1 2016-present Scille SAS
+
+use libparsec_types::{EntryName, FsPath};
+
+pub fn path_join(root: &FsPath, leaf: &EntryName) -> FsPath {
```
TimeEngineer, line 5: A `directory` can be assimilate as a tree, not `path`
```suggestion
pub fn path_join(parent: &FsPath, child: &EntryName) -> FsPath {
```

bindings/generator/api/path.py
```diff
@@ -0,0 +1,24 @@
+# Parsec Cloud (https://parsec.cloud) Copyright (c) BUSL-1.1 2016-present Scille SAS
+
+from .common import Ref, FsPath, EntryName
+from typing import Optional
+
+
+def path_join(root: Ref[FsPath], leaf: Ref[EntryName]) -> FsPath:
```
TimeEngineer, line 7: Same

bindings/generator/api/__init__.py
```diff
@@ -1,13 +1,15 @@
 # Parsec Cloud (https://parsec.cloud) Copyright (c) BUSL-1.1 2016-present Scille SAS
+# ruff: noqa: F403
```
TimeEngineer, line 2: Nice :+1:

bindings/generator/api/path.py
FirelightFlagboy:
```suggestion
def path_join(parent: Ref[FsPath], child: Ref[EntryName]) -> FsPath:
```

libparsec/crates/types/src/fs_path.rs
FirelightFlagboy: I've taken inspiration of [`Path::canonicalize`](https://doc.rust-lang.org/std/path/struct.Path.html#method.canonicalize) where it take a ref.

With your suggestion I would need to clone `FsPath` before using it :shrug:

Max-7 (approved):

mmmarcos (commented):
Why so many changes to `definitions.ts`?

Max-7:
> Why so many changes to `definitions.ts`?

It's my own fault I think. I updated the `definitions.ts` when I added `prettier`, but later excluded it from the files to update (added it to the `prettierignore` file) without reverting the original file. So what happened was:
- updated definitions.ts with prettier
- excluded the file from prettier update
- forgot to revert definitions.ts
- definitions.ts gets updated in this PR and goes back to its original format

mmmarcos (approved):
LGTM

libparsec/crates/types/src/fs_path.rs
TimeEngineer: cloning the `vec` is faster than cloning each element `1` by `1`, and why do you want to keep the old one ? In my opinion, most of the time you would like to drop the older one :thinking:

FirelightFlagboy: You never know :shrug:,

I've applied your suggestion yesterday.

TimeEngineer (approved):

PR #5815, opened 2023-11-17
Bump rust to 1.74

FirelightFlagboy:
- Use new `lint` table.

  https://doc.rust-lang.org/stable/cargo/reference/manifest.html#the-lints-section
  https://doc.rust-lang.org/stable/cargo/reference/workspaces.html#the-lints-table

- Fix const ref without an explicit static lifetime.
- Remove unnecessary `let` binding for downcast.
- Fix the borrowed expression implements the required traits for `std::fs::read_dir(path)`.
- Fix redundant redifinition of a binding.
- Ignore warning about usage of `skip(0)`.
- Ignore unwrap used in example `memfs`.

### DIFF
--- a/libparsec/crates/platform_mountpoint/examples/memfs.rs
+++ b/libparsec/crates/platform_mountpoint/examples/memfs.rs
@@ -1,4 +1,5 @@
 // Parsec Cloud (https://parsec.cloud) Copyright (c) BUSL-1.1 2016-present Scille SAS
+#![allow(clippy::unwrap_used)]
 
 fn main() {
     #[cfg(not(target_os = "macos"))]
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,7 +1,7 @@
 exclude: ^misc/bench.py$
 
 default_language_version:
-  rust: 1.72.1
+  rust: 1.74.0
   node: 18.12.0
 
 repos:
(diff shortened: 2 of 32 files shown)


libparsec/crates/platform_mountpoint/examples/memfs.rs
```diff
@@ -16,10 +16,13 @@ fn main() {
         let arg = args.nth(1).expect("Please provide mountpoint path");
         let path = Path::new(&arg);
 
+        #[allow(clippy::unwrap_used)]
```
TimeEngineer, line 19: Can you use global allow ?

TimeEngineer:
> Fix usage of skip(0) use 1 to skip the first element

What do you mean ? You fixed with a `clippy allow`

TimeEngineer:
> Fix & without an explicit lifetime.

This is for const only ? so maybe you mean without explicit static lifetime ?

libparsec/crates/platform_mountpoint/examples/memfs.rs
FirelightFlagboy: done

FirelightFlagboy:
@TimeEngineer I've updated the commit message (I often forget to update the PR description)

TimeEngineer (approved):

Max-7 (approved):

mmmarcos (approved):

PR #5816, opened 2023-11-17
[MS] Export recovery device page

Ironicbay:
## Checklist

- [ ] Describe your changes
   
If this PR is related to GUI:

### DIFF
--- a/client/src/components/core/ms-modal/MsPasswordInputModal.vue
+++ b/client/src/components/core/ms-modal/MsPasswordInputModal.vue
@@ -5,11 +5,6 @@
     :title="title"
     :subtitle="subtitle"
     :close-button="{ visible: true }"
-    :cancel-button="{
-      label: $t('PasswordInputModal.cancel'),
-      disabled: false,
-      onClick: cancel,
-    }"
     :confirm-button="{
       label: okButtonText || $t('PasswordInputModal.ok'),
       disabled: password.length === 0,
--- a/client/src/components/core/ms-modal/MsPasswordInputModal.vue
+++ b/client/src/components/core/ms-modal/MsPasswordInputModal.vue
@@ -1,0 +1,82 @@
+<!-- Parsec Cloud (https://parsec.cloud) Copyright (c) BUSL-1.1 2016-present Scille SAS -->
+
+<template>
+  <ms-modal
+    :title="title"
+    :subtitle="subtitle"
+    :close-button="{ visible: true }"
+    :cancel-button="{
+      label: $t('PasswordInputModal.cancel'),
+      disabled: false,
+      onClick: cancel,
+    }"
+    :confirm-button="{
+      label: okButtonText || $t('PasswordInputModal.ok'),
+      disabled: password.length === 0,
+      onClick: confirm,
+    }"
+  >
+    <ms-password-input
+      :label="inputLabel || ''"
+      v-model="password"
+      @on-enter-keyup="confirm()"
+    />
+  </ms-modal>
+</template>
+
+<script lang="ts">
+import MsPasswordInputModal from '@/components/core/ms-modal/MsPasswordInputModal.vue';
+
+export interface GetPasswordOptions {
+  title: string;
+  subtitle?: string;
+  inputLabel?: string;
+  okButtonText?: string;
+}
+
+export async function getPasswordFromUser(options: GetPasswordOptions): Promise<string | null> {
+  const modal = await modalController.create({
+    component: MsPasswordInputModal,
+    canDismiss: true,
+    cssClass: 'text-input-modal',
+    componentProps: {
+      title: options.title,
+      subtitle: options.subtitle,
+      inputLabel: options.inputLabel,
+      okButtonText: options.okButtonText,
+    },
+  });
+  await modal.present();
+  const result = await modal.onWillDismiss();
+  await modal.dismiss();
+  return result.role === MsModalResult.Confirm ? result.data : null;
+}
+</script>
+
+<script setup lang="ts">
+import { modalController } from '@ionic/vue';
+import { ref } from 'vue';
+import MsModal from '@/components/core/ms-modal/MsModal.vue';
+import MsPasswordInput from '@/components/core/ms-input/MsPasswordInput.vue';
+import { MsModalResult } from '@/components/core/ms-types';
+
+defineProps<GetPasswordOptions>();
+
+const password = ref('');
+
+async function confirm(): Promise<boolean> {
+  if (password.value.length === 0) {
+    return false;
+  }
+  return await modalController.dismiss(password.value, MsModalResult.Confirm);
+}
+
+function cancel(): Promise<boolean> {
+  return modalController.dismiss(null, MsModalResult.Cancel);
+}
+</script>
+
+<style scoped lang="scss">
+.password-input-modal {
+}
+</style>
--- a/client/src/locales/fr-FR.json
+++ b/client/src/locales/fr-FR.json
@@ -535,6 +535,35 @@
         "passwordCriteria": "Assurez-vous qu'il correspond à un niveau fort : évitez les répétitions, les mots trop communs et les suites.",
         "noMatch": "Ne correspondent pas"
     },
+    "PasswordInputModal": {
+        "passwordNeeded": "Mot de passe requis",
+        "enterPassword": "Veuillez saisir le mot de passe {org} pour continuer.",
+        "password": "Mot de passe",
+        "validate": "Valider"
+    },
+    "RecoveryDevicePage": {
+        "titles":{
+            "recoveryFile": "Fichier de récupération",
+            "secretKey": "Clé secrète"
+        },
+        "subtitles":{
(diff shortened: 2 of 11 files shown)


client/src/components/core/ms-modal/MsPasswordInputModal.vue
```diff
@@ -0,0 +1,77 @@
...
+  inputLabel?: string;
+  okButtonText?: string;
+}
+
+export async function getPasswordFromUser(options: GetPasswordOptions): Promise<string | null> {
+  const modal = await modalController.create({
+    component: MsPasswordInputModal,
+    canDismiss: true,
+    cssClass: 'text-input-modal',
+    componentProps: {
```
Max-7, line 37: I know I wrote the code, but you should be able to do `componentProps: options`, we don't need to redeclare every prop one by one.

client/src/locales/fr-FR.json
```diff
@@ -534,6 +535,35 @@
         "passwordCriteria": "Assurez-vous qu'il correspond à un niveau fort : évitez les répétitions, les mots trop communs et les suites.",
         "noMatch": "Ne correspondent pas"
     },
+    "PasswordInputModal": {
+        "passwordNeeded": "Mot de passe requis",
+        "enterPassword": "Veuillez saisir le mot de passe {org} pour continuer.",
```
Max-7, line 540: `{org}` is never set.

client/src/views/devices/ExportRecoveryDevicePage.vue
```diff
@@ -0,0 +1,204 @@
...
+            <ms-informative-text>
+              {{ $t('RecoveryDevicePage.subtitles.twoFilesToKeep') }}
+            </ms-informative-text>
+
+            <div>
+              <div class="block">
+                {{ $t('RecoveryDevicePage.titles.recoveryFile') }}
+              </div>
+              <div class="block">
+                {{ $t('RecoveryDevicePage.titles.secretKey') }}
```
Max-7, line 21: Should be `ExportRecoveryDevicePage` or something like that, since we'll have another page for the import, it may be confusing.

client/src/views/devices/ExportRecoveryDevicePage.vue
```diff
@@ -0,0 +1,204 @@
...
+let code = '';
+let file = '';
+const downloadLink = ref();
+const recoveryKeyDownloaded = ref(false);
+const recoveryFileDownloaded = ref(false);
+const notificationManager: NotificationManager = inject(NotificationKey)!;
+
+async function exportDevice(): Promise<void> {
+  const password = await getPasswordFromUser({
+    title: t('PasswordInputModal.passwordNeeded'),
```
Max-7, line 126: If these texts are generic and don't apply only in this case, might as well put them as default in `MsPasswordInputModal` and not have them as arguments.

client/src/views/devices/ExportRecoveryDevicePage.vue
```diff
@@ -0,0 +1,204 @@
...
+    subtitle: t('PasswordInputModal.enterPassword'),
+    inputLabel: t('PasswordInputModal.password'),
+    okButtonText: t('PasswordInputModal.validate'),
+  });
+  if (!password) {
+    return;
+  }
+  const result = await exportRecoveryDevice(password);
+  if (!result.ok) {
+    // TODO: showToast();
```
Max-7, line 136: Missing the error handling here. As we've said, you may be able to add an error tag to `exportRecoveryDevice()`, so the errors returned could be `WrongAuthentication` if the password is not `P@ssw0rd.` and `Internal` for other cases, and have two different errors

client/src/views/devices/ExportRecoveryDevicePage.vue
```diff
@@ -0,0 +1,204 @@
...
+    console.log('ERROR !');
+    return;
+  }
+  code = result.value.code;
+  file = result.value.file;
+  state.value = ExportDevicePageState.Download;
+}
+
+async function downloadKey(): Promise<void> {
+  download(code, 'Recovery_code.txt');
```
Max-7, line 147: Probably better to change the file name here, something closer to what V2 is doing https://github.com/Scille/parsec-cloud/blob/7d8cce2fad2573d04e52bb0971c3e0b449c3f051/parsec/core/local_device.py#L77
We may even translate the file name, have something like `"Parsec_Code_Récupération_${OrgID}"` in French, `"Parsec_Recovery_Code_${OrgID}"` in English (those are just example).

`OrgID` can be retrieved with `getClientInfo()`, in `parsec/login.ts`.

client/src/views/devices/ExportRecoveryDevicePage.vue
```diff
@@ -0,0 +1,204 @@
...
+  notificationManager.showToast(
+    new Notification({
+      message: t('RecoveryDevicePage.toasts.keyDownloadOk'),
+      level: NotificationLevel.Success,
+    }),
+  );
+}
+
+async function downloadFile(): Promise<void> {
+  download(file, 'Recovery_device.data');
```
Max-7, line 158: Same here, better to change the name.

client/src/components/core/ms-modal/MsPasswordInputModal.vue
```diff
@@ -0,0 +1,77 @@
...
+const password = ref('');
+
+async function confirm(): Promise<boolean> {
+  if (password.value.length === 0) {
+    return false;
+  }
+  return await modalController.dismiss(password.value, MsModalResult.Confirm);
+}
+
+function cancel(): Promise<boolean> {
```
Max-7, line 69: cancel() function is not needed.

Max-7 (changes requested):

client/src/components/core/ms-modal/MsPasswordInputModal.vue
```diff
@@ -0,0 +1,76 @@
...
+
+export interface GetPasswordOptions {
+  title: string;
+  subtitle?: string;
+  inputLabel?: string;
+  okButtonText?: string;
+}
+
+export async function getPasswordFromUser(t: ComposerTranslation): Promise<string | null> {
+  const clientInfo = await getClientInfo();
```
Max-7, line 35: Probably a bit of a misunderstanding here. `getPasswordFromUser` should take a `GetPasswordOptions`, not a `ComposerTranslation`. Title, subtitles and all should be decided by the caller. All I was trying to say was that you could do
```typescript
const modal = await modalController.create({
  ...
  componentProps: options,
});
```

instead of
```typescript
const modal = await modalController.create({
  ...
  componentProps: {
    title: options.title,
    subtitle: options.subtitle,
    ...
  }
});
```

client/src/views/devices/ExportRecoveryDevicePage.vue
```diff
@@ -0,0 +1,208 @@
...
+                  <ion-icon
+                    :icon="checkmarkCircle"
+                    class="checked"
+                  />
+                  {{ $t('ExportRecoveryDevicePage.subtitles.fileDownloaded') }}
+                </div>
+              </div>
+              <div class="block">
+                {{ $t('ExportRecoveryDevicePage.titles.recoveryKey') }}
+                <div v-if="!recoveryKeyDownloaded">
```
Max-7, line 55: `v-show` are probably better in those cases, mainly because it's easier to test.

`v-if` can be useful in cases where you have an undefined object for example:
```typescript
const obj: Ref<object | null> = ref(null);
```
```vue
<div v-if="obj">
  <!-- You have a guarantee that within this div tag, obj will never be null -->
  {{ obj.attribute }}
</div>
```

For testing, it's easier to have `cy.get('element').should('not.be.visible')` rather than testing whether the tag exists or not.

client/src/views/devices/ExportRecoveryDevicePage.vue
```diff
@@ -0,0 +1,208 @@
...
+          </div>
+          <div v-if="state === ExportDevicePageState.Download">
+            <ms-informative-text>
+              {{ $t('ExportRecoveryDevicePage.subtitles.keepFilesSeparate') }}
+            </ms-informative-text>
+
+            <div>
+              <div class="block">
+                {{ $t('ExportRecoveryDevicePage.titles.recoveryFile') }}
+                <div v-if="!recoveryFileDownloaded">
```
Max-7, line 38: Same here, `v-show` is better.

client/src/views/devices/ExportRecoveryDevicePage.vue
```diff
@@ -0,0 +1,208 @@
...
+  notificationManager.showToast(
+    new Notification({
+      message: t('ExportRecoveryDevicePage.toasts.keyDownloadOk'),
+      level: NotificationLevel.Success,
+    }),
+  );
+}
+
+async function downloadRecoveryFile(): Promise<void> {
+  const clientInfo = await getClientInfo();
```
Max-7, line 161: A bit of a waste to call `getClientInfo()` multiple times, it's better to do it once in the `onMounted()` function and store the result.

Max-7 (changes requested):

NicoTuxx (approved):

Issue #5811, opened 2023-11-16
API User freeze/unfreeze


vxgmichel:
Add server administration route to freeze and unfreeze users.

Frozen users won't be able to connect the parsec servers.

This allows for automated processes to temporarily block users while waiting for an organization administrator to actually (and cryptographically) revoke the corresponding user

FirelightFlagboy:
Fixed in #5842

PR #5817, opened 2023-11-17
Add RFC 1007 - Frozen users

vxgmichel:
Issue #5811

### DIFF
--- a/docs/rfcs/1007-frozen-users.md
+++ b/docs/rfcs/1007-frozen-users.md
@@ -1,70 +1,114 @@
 <!-- Parsec Cloud (https://parsec.cloud) Copyright (c) BUSL-1.1 2016-present Scille SAS -->
 
-Frozen users
-============
+# Frozen users
 
 Frozen users are users that are not allowed to connect to the parsec server.
 
 This feature is not part of the parsec revocation system, but a way for system administrators to temporarily block users from connecting to the server while waiting for organization administrators to actually revoke them. It is exposed as an HTTP route that requires administration token. This way, it is possible for existing directory services to automatically freeze users that have been removed from the directory.
 
-HTTP `frozen_users` route
--------------------------
+## HTTP `users` route
 
-This route is made available as `/administration/organizations/<raw_organization_id>/frozen_users` and requires an administration token.
+This route is made available as `/administration/organizations/<raw_organization_id>/users` and requires an administration token.
 
 ### `GET` method
 
-The `GET` method is used to list frozen status for all users, identified by their email addresses.
+The `GET` method is used to list information for all users, including parsec ID, user name, user email and frozen status.
 
 The body of the `GET` response is a JSON structure with the following format:
 
-```json
+```jsonc
 [
     {
+        "user_id": "<parsec_user_id_1>",
+        "user_name": "<user_name_1>",
         "user_email": "<user_email_1>",
-        "frozen": true
+        "frozen": true // or false
     },
-    [...]
+    // [...]
     {
+        "user_id": "<parsec_user_id_N>",
+        "user_name": "<user_name_N>",
         "user_email": "<user_email_N>",
-        "frozen": false
+        "frozen": true // or false
     }
 ]
 ```
 
 The list contains all users of the organization, including those that are not frozen.
 
-However, it does not contain the email addresses of users that have been revoked.
+However, it does not contain the users that have been revoked.
+
+## HTTP `users/frozen` route
+
+This route is made available as `/administration/organizations/<raw_organization_id>/users/frozen` and requires an administration token.
 
 ### `PATCH` method
 
-The `PATCH` method is used to modify this status for a given set of users.
+The `PATCH` method is used to modify the `frozen` status for a given user.
 
-The body of the `PATCH` request must be a JSON structure with the following format:
+The body of the `PATCH` request must be a JSON structure with either one of following formats:
 
-```json
+```jsonc
 {
     "user_email": "<user_email>",
-    "frozen": true
+    "frozen": true // or false
 }
 ```
 
-### Error handling
+Or:
+
+```jsonc
+{
+    "user_id": "<parsec_user_id>",
+    "frozen": true // or false
+}
+```
+
+This way, a user can either be identified by its email or by its parsec ID.
+
+The body of the `PATCH` response is a JSON structure summarizing the updated information for the selected user:
+
+```jsonc
+{
+    "user_id": "<parsec_user_id>",
+    "user_name": "<user_name>",
+    "user_email": "<user_email>",
+    "frozen": true // or false
+}
+```
+
+### User identification: email address vs parsec ID
+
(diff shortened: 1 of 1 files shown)


docs/rfcs/1007-frozen-users.md
```diff
@@ -0,0 +1,89 @@
...
+The `PATCH` method is used to modify this status for a given set of users.
+
+The body of the `PATCH` request must be a JSON structure with the following format:
+
+```json
+{
+    "user_email": "<user_email>",
+    "frozen": true
+}
+```
```
mmmarcos, line 50: Considering GET returns an array of users, I think it should be possible (and useful!) to also PATCH an array of users.

The only extra thing to consider is what happen if one or more users are not found (e.g. invalid email). According to the [IETF RFC 5789](https://www.rfc-editor.org/rfc/rfc5789.html#section-2):

> The server MUST apply the entire set of changes atomically and never
>    provide (e.g., in response to a GET during this operation) a
>    partially modified representation.  If the entire patch document
>    cannot be successfully applied, then the server MUST NOT apply any of
>    the changes.

docs/rfcs/1007-frozen-users.md
```diff
@@ -0,0 +1,89 @@
...
+
+### Error handling
+
+The following standard errors are handled the same way as for the other administration routes:
+- Organization not found: `404` with JSON body `{"error": "not_found"}`
+- Invalid administration token: `403` with JSON body `{"error": "not_allowed"}`
+- Wrong request format: `400` with JSON body `{"error": "bad_data"}`
+
+On top of it, an extra error is handled when the `PATCH` request contains a user that does not exist in the organization:
+- User not found: `404` with JSON body `{"error": "user_not_found"}`
```
mmmarcos, line 60: :+1: this should also be the case if we accept to PATCH multiple users and there is at least one that is not found.

mmmarcos: Actually, if we enable `PATCH`ing multiple users, it makes sense to return `404` + the list of users not found :thinking:  

```json
{
    "error": "user_not_found"
    "not_found": ["<user_email_1>", "<user_email_2>", "<user_email_N>"]
}
```

docs/rfcs/1007-frozen-users.md
vxgmichel: We discussed it @touilleMan and agreed that was not necessary for the moment, I added an extra section in the RFC about this future improvement. Is that ok with you?

docs/rfcs/1007-frozen-users.md
```diff
@@ -0,0 +1,89 @@
+<!-- Parsec Cloud (https://parsec.cloud) Copyright (c) BUSL-1.1 2016-present Scille SAS -->
+
+Frozen users
+============
+
+Frozen users are users that are not allowed to connect to the parsec server.
```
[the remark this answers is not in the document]
vxgmichel: I don't know.. The first version, although a bit clumsy, looks more like a definition than your suggestion

docs/rfcs/1007-frozen-users.md
```diff
@@ -0,0 +1,133 @@
...
+    "user_email": "<user_email>",
+    "frozen": true // or false
+}
+```
+
+### User identification: email address vs parsec ID
+
+There is a subtle difference between the two ways to identify a user. At any given time, an email address can be used to uniquely identify an non-revoked user from a given organization. In contrast, a parsec user ID identifies uniquely any user from all organizations in the parsec server, including revoked users. This means that over time, an email address can identify different parsec users with different parsec IDs, even from the same organization.
+
+The frozen status configured by the `PATCH` method is specifically associated with the parsec user ID, regardless of the identification method used in the request body. This has a the following consequence: if a user is revoked and then a new user is created with the same email address, the frozen status will **not** be applied to the new user.
```
mmmarcos, line 84: I understand the differences between email address and parsec ID, but I am not sure I follow the logic here... 

Regarding the _identification_ of the user to be frozen : 
The requirement is to _freeze a user from a specific organization_. The organization is already identified from the endpoint (`/administration/organizations/<raw_organization_id>/users/frozen`) therefore the email address seem sufficient to uniquely _identify_ the user to freeze (and the identification it via parsec ID in the PATCH request seem redundant).  

Regarding the "attaching" of the frozen status to the email vs parsec ID:
I am not convinced that we want to allow the new user (with same email) to `not` be frozen... and in any case, this could be implemented during revocation (i.e. clearing frozen status for the corresponding email address).

docs/rfcs/1007-frozen-users.md
```diff
@@ -0,0 +1,89 @@
...
+This feature is not part of the parsec revocation system, but a way for system administrators to temporarily block users from connecting to the server while waiting for organization administrators to actually revoke them. It is exposed as an HTTP route that requires administration token. This way, it is possible for existing directory services to automatically freeze users that have been removed from the directory.
+
+HTTP `frozen_users` route
+-------------------------
+
+This route is made available as `/administration/organizations/<raw_organization_id>/frozen_users` and requires an administration token.
+
+### `GET` method
+
+The `GET` method is used to list frozen status for all users, identified by their email addresses.
```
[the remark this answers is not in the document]
mmmarcos: The project requiring this feature will have ~2000 users in the organization. We can leave it out of scope if it simplifies implementation, but we'll probably need it sooner or later IMO

docs/rfcs/1007-frozen-users.md
mmmarcos: As you said, this is not really necessary for the moment :+1:

docs/rfcs/1007-frozen-users.md
FirelightFlagboy, line 1: The file should be named `1006-frozon-users`

docs/rfcs/1007-frozen-users.md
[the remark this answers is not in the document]
vxgmichel: 2000 users * 150 bytes is about 300KB, which is probably fine as long as the directory does not constantly poll this route.
We discussed it with @touilleMan and he says we'll have to implement some paging mechanism for this route anyway so I think we can think about a more generic approach later.

docs/rfcs/1007-frozen-users.md
vxgmichel: Nope: https://github.com/Scille/parsec-cloud/blob/a6c336be8503c2320582b9e494ce075d232d84b3/docs/rfcs/1005-realm-key-rotation.md?plain=1#L526

docs/rfcs/1007-frozen-users.md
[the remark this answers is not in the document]
vxgmichel: 150 bytes times 2000 users is about 300KB which I think is fine as long as the directory service does not poll this route constantly. @touilleMan said we'll have to implement some paging mechanism at some point so we can maybe think of a more generic approach then.

docs/rfcs/1007-frozen-users.md
FirelightFlagboy: Good catch :+1:

FirelightFlagboy: https://github.com/Scille/parsec-cloud/issues/5836

touilleMan (approved):

PR #5834, opened 2023-11-20
[MS] Fixed file/user list item background and options button

Max-7:
Two very small changes related to the file list item:
1. correctly displays a background when selected
2. hide the options button when item not hover for files and users

- [X] Describe your changes

### DIFF
--- a/client/src/components/files/FileCard.vue
+++ b/client/src/components/files/FileCard.vue
@@ -4,10 +4,18 @@
   <div
     class="card"
     @click="$emit('click', $event, file)"
+    @mouseenter="isHovered = true"
+    @mouseleave="isHovered = false"
   >
     <div
       class="card-option"
-      @click.stop="$emit('menuClick', $event, file)"
+      v-show="isHovered || menuOpened"
+      @click.stop="
+        menuOpened = true;
+        $emit('menuClick', $event, file, () => {
+          menuOpened = false;
+        });
+      "
     >
       <ion-icon :icon="ellipsisHorizontal" />
     </div>
@@ -39,9 +47,12 @@
 <script setup lang="ts">
 import { folder, document, ellipsisHorizontal, cloudDone, cloudOffline } from 'ionicons/icons';
 import { IonAvatar, IonIcon, IonText, IonTitle } from '@ionic/vue';
-import { inject } from 'vue';
+import { inject, ref } from 'vue';
 import { FormattersKey, Formatters } from '@/common/injectionKeys';
 import { EntryStat } from '@/parsec';
+
+const isHovered = ref(false);
+const menuOpened = ref(false);
 
 const props = defineProps<{
   file: EntryStat;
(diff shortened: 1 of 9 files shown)


fabienSvtr:
In `FileListItem.vue` we should remove the line 161 `:focus` because it creates a bug when we maintain left click and drag on a not selected item.

fabienSvtr (approved):

client/src/components/files/FileCard.vue
```diff
@@ -4,10 +4,18 @@
...
     class="card"
     @click="$emit('click', $event, file)"
+    @mouseenter="isHovered = true"
+    @mouseleave="isHovered = false"
   >
     <div
       class="card-option"
-      @click.stop="$emit('menuClick', $event, file)"
+      v-show="isHovered || menuOpened"
+      @click.stop="
```
NicoTuxx, line 13: all the following code makes reading the template laborious, should be moved to a `onSomethingClick` function.

client/src/components/files/FileListItem.vue
```diff
@@ -72,8 +73,14 @@
     <div class="file-options ion-item-child-clickable">
       <ion-button
         fill="clear"
+        v-show="isHovered || menuOpened"
         class="options-button"
-        @click.stop="$emit('menuClick', $event, file)"
+        @click.stop="
+          menuOpened = true;
```
NicoTuxx, line 79: same as previous comment.

client/src/components/users/UserCard.vue
```diff
@@ -21,8 +21,13 @@
     </div>
     <div
       class="card-option"
-      v-show="showOptions"
-      @click.stop="$emit('menuClick', $event, user)"
+      v-show="isHovered || menuOpened"
+      @click.stop="
+        menuOpened = true;
```
NicoTuxx, line 26: same here too.

client/src/components/users/UserListItem.vue
```diff
@@ -55,10 +55,15 @@
     <!-- options -->
     <div class="user-options ion-item-child-clickable">
       <ion-button
-        v-show="showOptions"
+        v-show="(isHovered || menuOpened) && !hideOptions"
         fill="clear"
         class="options-button"
-        @click.stop="$emit('menuClick', $event, user)"
+        @click.stop="
+          menuOpened = true;
```
NicoTuxx, line 62: same.

NicoTuxx (changes requested):

NicoTuxx (changes requested):
A weird behavior is happening when selecting elements and then clicking on an other one.
It only seems to happen when we click on the element himself, not his options button.
It does not happen when we click outside of the list too.
https://github.com/Scille/parsec-cloud/assets/19772607/46f8aa08-e1d6-48c0-b1d0-cb0e3c80d827

Max-7:
> A weird behavior is happening when selecting elements and then clicking on an other one. It only seems to happen when we click on the element himself, not his options button. It does not happen when we click outside of the list too. https://github.com/Scille/parsec-cloud/assets/19772607/46f8aa08-e1d6-48c0-b1d0-cb0e3c80d827

It's because of the mocks. You're travelling inside the folder `Dir1` but since the mocks always returns the same files/folders, it feels like the selection is just removed. If you look at the header, you'll see that you are now inside `/Dir1` instead of `/`.

NicoTuxx (approved):

PR #5838, opened 2023-11-21
Use postgres v14.10 in on-premise installation documentation

FirelightFlagboy:
- Use `version_updater` script to do that.
- Change `hide_patch_version` to only use the 1st 2 values of the split (i.e.: the MAJOR and the minor)


mmmarcos (approved):
👍 Maybe we can consider using a lib to parse version strings (not right now, but for later)

PR #5840, opened 2023-11-21
Change postgres version to v14 in ci-python

FirelightFlagboy:
Note that the tests that use the database are disabled because that don't work at the moment.

Depends on #5838


mmmarcos (approved):
👍

PR #5937, opened 2023-12-05
[MS] Moved input function to a separate file

### LINKED ISSUE #5847: Move input functions to a specific file
Right now, functions that use components to get a user input (askQuestion, getTextInput, ...) are located in the same file as the component, which creates some weird import situations (a file importing itself).

Those functions should be moved to a specific file.

Non exhaustive list:
- askQuestion
- getTextInputFromUser
- getPasswordFromUser (#5816)
- selectFolder (#5814 )

Max-7:
Closes #5847 

- [X] Describe your changes


client/src/common/inputs.ts
NicoTuxx, line 1: this file should not be in `src/common`:
- we only have 'technical' and 'logical' typescript file in this folder, basically tools or utils, but nothing directly generating UIs
- we already have files that do similar things (notificationManager, toastManager...), it could be under `src/services`, being an `inputManager.ts`
- if you don't like the previous idea, maybe it should only be a `utils.ts` file under `src/components` like `ms-types.ts`

client/src/components/core/ms-modal/MsTextInputModal.vue
```diff
@@ -25,51 +25,17 @@
...
-      defaultValue: options.defaultValue,
-    },
-  });
-  await modal.present();
-  const result = await modal.onWillDismiss();
-  await modal.dismiss();
-  return result.role === MsModalResult.Confirm ? result.data : null;
-}
-</script>
+<script lang="ts"></script>
```
NicoTuxx, line 28: useless line, can be removed

NicoTuxx (changes requested):

NicoTuxx (approved):

PR #5942, opened 2023-12-05
Cli: `tests`

TimeEngineer:


FirelightFlagboy:
Can you rebase, the PR is outdated with the change we have merged before

.github/workflows/ci-rust.yml
```diff
@@ -89,6 +89,13 @@ jobs:
         with:
           tool: nextest@0.9.54, wasm-pack@0.11.0, cargo-deny@0.14.3
 
+      - name: Test cli
+        shell: bash
+        run: |
+          set -ex
+          cargo nextest run ${{ env.CARGO_NEXTEST_CI_FLAGS }} --package parsec_cli --features testenv
```
FirelightFlagboy, line 96:
```suggestion
        run: cargo nextest run ${{ env.CARGO_NEXTEST_CI_FLAGS }} --package parsec_cli --features testenv
```

.github/workflows/ci-rust.yml
```diff
@@ -224,3 +231,10 @@ jobs:
         env:
           RUST_LOG: debug
           TESTBED_SERVER: http://localhost:6777
+
+      - name: Test cli (${{ matrix.name }} specific)
+        shell: bash
+        run: |
+          set -ex
+          cargo nextest run ${{ env.CARGO_NEXTEST_CI_FLAGS }} --package parsec_cli --features testenv
```
FirelightFlagboy, line 239:
```suggestion
        run: cargo nextest run ${{ env.CARGO_NEXTEST_CI_FLAGS }} --package parsec_cli --features testenv
```

cli/src/run_testenv.rs
```diff
@@ -44,6 +45,11 @@ pub struct RunTestenv {
     empty: bool,
 }
 
+pub fn process_id_to_port(pid: u32) -> u16 {
+    // The first ports are reserved, so let's take an available port
+    pid as u16 % AVAILABLE_PORT_COUNT + RESERVED_PORT_OFFSET
+}
```
mmmarcos, line 51: Where do this come from? There is no relation between pid and port :thinking: 
I did not found something similar in the v2 branch.

Also: modulo should be the last operation, otherwise the return port may be greater than the available port count.

TimeEngineer: It comes from me to have an available port for the testbed server.
Note that there is a `RESERVED_PORT_OFFSET` which means that the port from 1 to 1024 are reserved.
The modulo guarantees that you can't overflow.

.github/workflows/ci-rust.yml
```diff
@@ -89,6 +90,22 @@ jobs:
...
+        with:
+          poetry-version: ${{ env.poetry-version }}
+          project-path: ./server
+        timeout-minutes: 10
+
+      - name: Test cli
+        shell: bash
+        run: |
+          set -ex
+          python make.py python-dev-install
```
FirelightFlagboy, line 104: Install the python dependencies in a dedicated step with a dedicated timeout.

.github/workflows/ci-rust.yml
```diff
@@ -89,6 +90,22 @@ jobs:
...
+          cargo build --profile ci-rust --package parsec_cli --features testenv
```
FirelightFlagboy, line 105: Does that step is still required ?

.github/workflows/ci-rust.yml
```diff
@@ -224,3 +241,19 @@ jobs:
...
+        with:
+          poetry-version: ${{ env.poetry-version }}
+          project-path: ./server
+        timeout-minutes: 10
+
+      - name: Test cli
+        shell: bash
+        run: |
+          set -ex
+          python make.py python-dev-install
```
FirelightFlagboy, line 256: Same as above

.github/workflows/ci-rust.yml
```diff
@@ -224,3 +241,19 @@ jobs:
...
+          cargo build --profile ci-rust --package parsec_cli --features testenv
```
FirelightFlagboy, line 257: Same question here as above

FirelightFlagboy (changes requested):

.github/workflows/ci-rust.yml
TimeEngineer: Yes, because assert_cmd doesn't build the binary, it uses only

.github/workflows/ci-rust.yml
TimeEngineer: Done

.github/workflows/ci-rust.yml
FirelightFlagboy: Can you do a dedicated step like for the python dependencies then ?

cli/src/run_testenv.rs
TimeEngineer: Yes it avoids to store the port value, we are not using a constant value, because we run tests in multiple threads, so we will have multiple servers. I use the PID to retrieve this port in test, if you have another idea you are welcome.

.github/workflows/ci-rust.yml
```diff
@@ -89,6 +90,26 @@ jobs:
...
+      - name: Install python dependencies
+        run: python make.py python-dev-install
+        timeout-minutes: 5
+
+      - name: Build CLI binary
+        run: cargo build --profile ci-rust --package parsec_cli --features testenv
+        timeout-minutes: 5
+
+      - name: Test CLI
+        shell: bash
```
FirelightFlagboy, line 109: Still required to specify the shell ?

.github/workflows/ci-rust.yml
```diff
@@ -224,3 +245,23 @@ jobs:
...
```
FirelightFlagboy, line 265: Still required to specify the shell ?

FirelightFlagboy (approved):

cli/src/run_testenv.rs
TimeEngineer: Done

mmmarcos (approved):
LGTM

PR #5948, opened 2023-12-05
[MS] Fix all existing popover placement

### LINKED ISSUE #5890: Fix popover placement problems

NicoTuxx:
closes #5890 
@fabienSvtr please take a look at the last commit, we have CSS managing border, border radius and overflow on this class: `popover-side-bottom`. This is causing a lot of problems like you can see, I've used a lot of :not to clean the placement mess due to this global override.
The CSS here is not meant to remain like this, but I would like your opinion on the subject, do we remove this override? Do we prefer case by case? or override on ion-popover maybe?
On another note, I must specify that I like the popovers without the border as much (maybe even more) as they are supposed to be, it is another subject, but maybe we could keep them like they are in the PR.

### DIFF
--- a/client/src/theme/components/popovers.scss
+++ b/client/src/theme/components/popovers.scss
@@ -1,58 +1,41 @@
 /* Parsec Cloud (https://parsec.cloud) Copyright (c) BUSL-1.1 2016-present Scille SAS */
 
 /* **** Parsec popover **** */
-
-ion-popover {
-  --width: 300px;
+:root {
+  --popover-border-radius: var(--parsec-radius-8);
+  --popover-default-width: 20rem;
+  --popover-option-width: 18rem;
 }
 
-.popover-side-bottom:not(.tooltip-popover) {
-  --width: 250px;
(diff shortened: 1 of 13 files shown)


client/src/theme/components/popovers.scss
fabienSvtr, line 1: To answer your questions about this popover CSS file, I think we need to remove all the `:not()` that aren't used because the `border` has been removed from all the popovers except the `dropdown outline` which hasn't been done yet (see the image).
image

In addition, in order to improve the overall management of popovers, I suggest that you put a `:root` at the beginning of the file, which would allow you to manage global parameters such as the default size and the border radius (which must be set in 2 places on the popover to work correctly). And then I realised that a lot of the parameters were redundant or useless, so I deleted them. I've made a commit so that you can see the changes for yourself, and we can delete it if you don't want it.

fabienSvtr (commented):
The global commit is working, see my comments for `popovers.css` file to improve the code.

Max-7 (approved):

PR #5950, opened 2023-12-06
Added reencrypt all button to workspace page

### LINKED ISSUE #5939: [🚀 | Feature request]: Bulk re-encryption of workspaces

When a user is revoked, the workspaces in which it was participating must be re-encrypted by users having an `Owner` role on those workspaces.

Currently, re-encryption is performed on each workspace by clicking, one by one, on the red padlock.

This is a relatively frequent operation in a high turnover context so the request is to add a function enabling bulk re-encryption of all workspaces for which the `user` performing the action has the `Owner` role.

It is proposed to add a global action to re-encrypt all workspaces owned by a user.

This action will be displayed in the workspace list via a button or similar (e.g. next to "Create a new workspace" ). The user will be asked to confirm before re-encryption.

_No response_

Max-7:
Closes #5939 

- [X] Describe your changes


mmmarcos (commented):
Some comments:

* In the docs, the term "re-encryption" is written with a hyphen ([Workspace-re-encryption](https://docs.parsec.cloud/en/latest/userguide/revoke_user.html#workspace-re-encryption)). Probably both are OK, but better to change it here that everywhere in the docs.
* I'm not sure about the wording: using "all" (as in "re-encrypt **all** workspaces") is misleading since not all workspaces are going to be re-encrypted. Maybe we should stick with "Re-encrypt workspaces" and having the dialog describe what would actually be re-encrypted. 
* We should update the [Workspace re-encryption](https://docs.parsec.cloud/en/latest/userguide/revoke_user.html#workspace-re-encryption) to document this feature.

vxgmichel (approved):
Ran a few tests, works fine :+1:

Max-7:
> Some comments:
> 
> * In the docs, the term "re-encryption" is written with a hyphen ([Workspace-re-encryption](https://docs.parsec.cloud/en/latest/userguide/revoke_user.html#workspace-re-encryption)). Probably both are OK, but better to change it here than everywhere in the docs.
> * I'm not sure about the wording: "re-encrypt **all** workspaces" is misleading since not _all_ workspaces are going to be re-encrypted. Maybe we should stick with "Re-encrypt workspaces" and having the dialog describe what would actually be re-encrypted.
> * We should document this feature in the docs ([Workspace re-encryption](https://docs.parsec.cloud/en/latest/userguide/revoke_user.html#workspace-re-encryption)).

- Changed `reencryption` to `re-encryption`
- Changed `all workspaces` to `multiple workspaces`.
- Added the documentation. This caused diffs in the whole translation file.

mmmarcos (approved):

PR #5954, opened 2023-12-06
Ms better image management

### LINKED ISSUE #5901: Improve icon management
- [ ] create MsIcon to manage ion-icon and custom icons
- [ ] replace custom icon by svg file and MsIcon
- [ ] remove the comment about: `Make it possible to add an icon in HTML without having to import it into TS` => not possible and we'll stay as it is, less `magic` usage of ion-icon

NicoTuxx:
closes #5901 
- create MsImage to manage custom assets (with extensions, fallback system)
- re-organize assets images folder:
  - background is meant for css background images
  - images root is meant for ms-image resources
  - rename files to match a pattern and be more explicit
- delete unused images
- add specific management for SVG, for more info read comment below


Max-7 (commented):
Don't forget to check on Electron if everything is working properly.

fabienSvtr (commented):
**3 comments:**
* Change the `MsImages` name (to avoid confusion in imports)
* Maybe add locales for alternative text
* `imgRessourceMap`: Can we provide a function to avoid adding files manually (with the `enum`)?

--
Otherwise, the way to use the `MsImages` is very simple and the `--fill-color` is fine (as we discussed yesterday) so I agree with this new way to add images.

NicoTuxx:
problem with merge provoked multiple code-owners to be required, closing and re-opening.

PR #5956, opened 2023-12-07
[MS] Better core components imports

### LINKED ISSUE #5955: Improve core components import management
## About exporting directly in the same file in another <script> block:

- Errors can be encountered due to how Vue and JavaScript handle module exports. In JavaScript, only values (like objects, functions, classes, etc.) can be exported and imported between modules.
- TypeScript interfaces, on the other hand, are not values and do not exist at runtime, so they cannot be exported or imported in the same way.
- Most of the time, it's working because TypeScript is able to find the interface in the global scope, not because it's being imported as a value from the `MsExample.vue` module.
- To ensure consistent behavior and avoid confusion, I would recommend moving our enums and interfaces to a separate TypeScript file and importing them from there.
- This way, we can be sure that the interfaces are always available when we need them, and we won't run into issues with the Vue `<script setup>` syntax.

## Benefits from exporting .vue components from index.ts:

- Avoid name confusion since we can name the import whatever we want
- Bring more control on what we want to use outside of components/core (ex: we don't export MsQuestionModal since we prefer to use askQuestion utils function)
- Improve significantly the readability of files that use these imports
- Make our imports very similar to ionic ones, easier to understand / read

## New usage:

- Outside of components/core:
    ```typescript
    import { ... } from '@components/core';
    ```

- Inside of components/core:
    ```typescript
    import { ... } from '@components/core/ms-utils';
    import { ... } from '@components/core/ms-types';
    import MsExample from '@components/core/ms-example/MsExample.vue';
    ```

- `ms-utils.ts` is meant for functions
- `ms-types.ts` is meant for enums, interfaces, maps (basically types and constants)

NicoTuxx:
depends on #5975  (need to rebase after #5975  will be merged)
closes #5955 

## About exporting directly in the same file in another <script> block:

- Errors can be encountered due to how Vue and JavaScript handle module exports. In JavaScript, only values (like objects, functions, classes, etc.) can be exported and imported between modules.
- TypeScript interfaces, on the other hand, are not values and do not exist at runtime, so they cannot be exported or imported in the same way.
- Most of the time, it's working because TypeScript is able to find the interface in the global scope, not because it's being imported as a value from the `MsExample.vue` module.
- To ensure consistent behavior and avoid confusion, I would recommend moving our enums and interfaces to a separate TypeScript file and importing them from there.
- This way, we can be sure that the interfaces are always available when we need them, and we won't run into issues with the Vue `<script setup>` syntax.

## Benefits from exporting .vue components from index.ts:

- Avoid name confusion since we can name the import whatever we want
- Bring more control on what we want to use outside of components/core (ex: we don't export MsQuestionModal since we prefer to use askQuestion utils function)
- Improve significantly the readability of files that use these imports
- Make our imports very similar to ionic ones, easier to understand / read

## New usage:

- Outside of components/core:
    ```typescript
    import { ... } from '@components/core';
    ```

- Inside of components/core:
    ```typescript
    import { ... } from '@components/core/ms-utils';
    import { ... } from '@components/core/ms-types';
    import MsExample from '@components/core/ms-example/MsExample.vue';
    ```

- `ms-utils.ts` is meant for functions
- `ms-types.ts` is meant for enums, interfaces, maps (basically types and constants)

### DIFF
--- a/client/src/components/core/ms-dropdown/index.ts
+++ b/client/src/components/core/ms-dropdown/index.ts
@@ -1,0 +1,5 @@
+// Parsec Cloud (https://parsec.cloud) Copyright (c) BUSL-1.1 2016-present Scille SAS
+import MsDropdown from '@/components/core/ms-dropdown/MsDropdown.vue';
+
+export * from '@/components/core/ms-dropdown/types';
+export { MsDropdown };
--- a/client/src/components/core/index.ts
+++ b/client/src/components/core/index.ts
@@ -1,0 +1,34 @@
+// Parsec Cloud (https://parsec.cloud) Copyright (c) BUSL-1.1 2016-present Scille SAS
+
+// ** ms-types ** //
+export * from '@/components/core/ms-types';
+
+// ** ms-action-bar ** //
+export * from '@/components/core/ms-action-bar';
+
+// ** ms-dropdown ** //
+export * from '@/components/core/ms-dropdown';
+
+// ** ms-image ** //
+export * from '@/components/core/ms-image';
+
+// ** ms-input ** //
+export * from '@/components/core/ms-input';
+
+// ** ms-modal ** //
+export * from '@/components/core/ms-modal';
+
+// ** ms-sorter ** //
+export * from '@/components/core/ms-sorter';
+
+// ** ms-spinner ** //
+export * from '@/components/core/ms-spinner';
+
+// ** ms-stepper ** //
+export * from '@/components/core/ms-stepper';
+
+// ** ms-text ** //
+export * from '@/components/core/ms-text';
+
+// ** ms-toggle ** //
+export * from '@/components/core/ms-toggle';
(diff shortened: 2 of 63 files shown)


client/src/components/core/ms-dropdown/index.ts
```diff
@@ -0,0 +1,4 @@
+// Parsec Cloud (https://parsec.cloud) Copyright (c) BUSL-1.1 2016-present Scille SAS
+import MsDropdown from '@/components/core/ms-dropdown/MsDropdown.vue';
+
+export { MsDropdown };
```
Max-7, line 4: Probably needs to export the `MsDropdownChangeEvent` too. It may not be explicitly required but it could be used to type a variable (something like `async def myFunction(event: MsDropdownChangeEvent): Promise<void>`

Max-7 (changes requested):

Max-7 (approved):

client/src/components/core/ms-dropdown/index.ts
NicoTuxx: missed it. Done!

PR #5958, opened 2023-12-07
Added bulk reassignment of workspaces roles

### LINKED ISSUE #5938: [🚀 | Feature request]: Bulk re-assignment of workspace rights

The need is motivated by a high turnover context. Generally, when a user leaves an organization, it is necessary to reassign its rights to another user.

The user role (`Reader`, `Contributor`, `Manager`, `Owner`) define the rights to perform actions within a specific workspace. A user can have different roles in different workspaces. However, roles held by a user are currently non-transferable. A new user will therefore have to be invited to each workspace with the same roles, one by one, manually.

The request is to add a function enabling the bulk reassignment of all workspace rights held by a user. This action would be carried out by a user via the GUI and will be applied on all the workspaces for which it has an `Owner` or `Manager` role (i.e. will not apply on workspaces where it has `Reader` or `Contributor` role).

Role reassignment functionality will be accessible from the Parsec GUI (e.g. button or dialog). The user will be asked to confirm before the role is reassigned.

The following use case is proposed:

1. `Alice` has the role of `Manager` or `Owner` in a number of workspaces.
2. `Bob` has the role of `Reader` or `Contributor` or `Manager` in these workspaces.
3. `Bob` leaves the organization and is replaced by `Carol` (existing Parsec user)
4. `Alice` transfers all `Bob`'s roles to `Carol` in a single action via the Parsec interface.
5. `Carol` obtains the same roles as `Bob` on workspaces shared with `Alice`
6. `Carol` doesn't get a role on workspaces where `Alice` has a `Reader` or `Contributor` role (these roles don't allow her to re-assign roles).

NOTE: The `Manager` role does not allow `Alice` to assign the `Owner` role to another user. In this case, the options are: 
* (1) Allow `Alice` as `Manager` to reassign `Bob`'s `Owner` role to `Carol`
* (2)  `Alice` reassigns a **demoted** role to `Manager` (allowing `Carol` to at least access the workspace).
* (3)  No role transfer

Option (2) seems a good compromise between security and usability. This is the chosen option.

_No response_

Max-7:
Closes #5938

1. Can only assign roles on workspaces where the client is Manager or Owner
2. When assigning to an Outsider, Manager and Owner roles are converted to Contributor
3. If the client is a Manager, and the role to assign is Manager or Owner, the role assigned becomes Contributor (a Manager cannot set roles higher than Contributor)
4. If the user has a higher role on the workspace than the one we want to assign, do nothing

- [x] Describe your changes


FirelightFlagboy (approved):
I validate the file `misc/generate_pyqt.py`

docs/userguide/share_data.rst
```diff
@@ -65,3 +65,34 @@ Data copied from file explorer also appear in the Parsec client. In addition, th
...
+
+.. image:: screens/bulk_assign_roles_summary.png
+    :align: center
+    :alt: Bulk assignment summary
+
+.. note::
+
+    1. You can only assign roles on workspaces where you are an Owner or Manager
+    2. Outsider cannot have a role higher than Contributor
+    3. Assigned roles can be downgraded if your own role on the workspace is not high enough (if you're a Manager and you're trying to assign an Owner role, it will be downgraded to Contributor)
```
mmmarcos, line 96: In the spec we've decided it should be downgraded to `Manager`. 
Is there a reason why you've chosen `Contributor`?

Also, I would re-phrase it a little bit:
```suggestion
    3. You cannot assign roles higher than your own. If you have the Manager role in a workspace and you're trying to assign the Owner role, it will be downgraded to Contributor).
```

docs/userguide/share_data.rst
```diff
@@ -65,3 +65,34 @@ Data copied from file explorer also appear in the Parsec client. In addition, th
...
+    4. If the user already has a higher role on the workspace than the one you're trying to assign, do nothing
```
mmmarcos, line 97:
```suggestion
    4. If the user already has a higher role on the workspace than the one you're trying to assign, no change will be made to its role
```

docs/userguide/share_data.rst
```diff
@@ -65,3 +65,34 @@ Data copied from file explorer also appear in the Parsec client. In addition, th
...
+    5. The option is only available if you have workspace in common with the source user and you have a role high enough
```
mmmarcos, line 98: I would move this to the top (as first point)
```suggestion
    5. The option is only available if you have workspaces in common with the source user and you have a role high enough
```

mmmarcos (changes requested):
Only some suggestions regarding the docs, otherwise LGTM :+1: 

Also, can you regenerate the images to replace the "Toto" user and its mail? It may seem like a detail, but it makes the docs look unprofessional

docs/userguide/share_data.rst
Max-7: `Managers` can not set other managers, only readers and contributor. Only Owners can assign other Owners.

docs/userguide/screens/bulk_assign_roles_select_user.png
mmmarcos, line 1: The context menu item that opens this dialog is named "Assign workspace roles to another user". 
However, once opened, it is not immediately clear what the dialog is about. 

Even if that is explained in the following step, I think we can improve it a little bit:

* Title: `Assign Bob workspace roles to:` (with the username emphasized? :thinking:)
* Description: `Select a user to see the roles that will be assigned.`

docs/userguide/share_data.rst
```diff
@@ -65,3 +65,34 @@ Data copied from file explorer also appear in the Parsec client. In addition, th
     Parsec client can work while offline (however only data present locally
     are available), synchronization will occur automatically as soon as the
     connection with the server is established.
+
+Bulk assign workspace roles to users
+------------------------------------
+
+Assigning the same roles to multiple users on multiple workspaces can be a bit annoying. Parsec provides a way to bulk assign the roles of one user to another.
```
mmmarcos, line 72: I'd use "tedious" instead of "annoying":
```suggestion
Manually assigning the same workspace roles to multiple users can be a bit tedious and error prone. Parsec provides a way to bulk assign the roles from one user to another.
```
In French:
```
L'assignation manuelle des mêmes rôles d'espace de travail à plusieurs utilisateurs peut s'avérer fastidieuse et source d'erreurs. Parsec propose un moyen d'attribuer en bloc les rôles d'un utilisateur à un autre.
```

docs/userguide/share_data.rst
```diff
@@ -65,3 +65,34 @@ Data copied from file explorer also appear in the Parsec client. In addition, th
...
+    :align: center
+    :alt: Start a bulk assignement
+
+2. Search the user you want to assign the roles to:
+
+.. image:: screens/bulk_assign_roles_select_user.png
+    :align: center
+    :alt: Select the user to assign roles to
+
+3. You're shown a summary of what roles are going to be assigned on each workspace. Click on `Assign roles` to complete the operation.
```
mmmarcos, line 86:
```suggestion
3. You're shown a summary of which roles are going to be assigned on each workspace. Click on `Assign roles` to complete the operation.
```

docs/userguide/share_data.rst
```diff
@@ -65,3 +65,34 @@ Data copied from file explorer also appear in the Parsec client. In addition, th
...
+
+.. image:: screens/bulk_assign_roles_summary.png
+    :align: center
+    :alt: Bulk assignment summary
+
+.. note::
+
+    1. You can only assign roles on workspaces where you are an Owner or Manager
+    2. Outsider cannot have a role higher than Contributor
+    3. You cannot assign roles higher than your own. If you have the Manager role in a workspace and you're trying to assign the Owner role, it will be downgraded to Contributor).
```
mmmarcos, line 96:
```suggestion
    1. You can only assign roles on a workspace where you have an Owner or Manager role
    2. An Outsider user cannot have a role higher than Contributor
    3. You cannot assign roles higher than your own. If you have the Manager role in a workspace and you're trying to assign the Owner role, it will be downgraded to Contributor.
```

mmmarcos (approved):
A few more suggestions regarding the docs and screenshots.

Otherwise LGTM :+1:

PR #5973, opened 2023-12-11
Add `empty` to `run-testenv`

TimeEngineer:


cli/src/run_testenv.rs
```diff
@@ -39,6 +39,9 @@ pub struct RunTestenv {
     /// When this process stops, the server will be automatically killed
     #[arg(long)]
     main_process_id: u32,
+    /// Skip initialization
+    #[arg(short, long)]
+    empty: Option<bool>,
```
FirelightFlagboy, line 44: You can configure `clap` to use a default value
 
```suggestion
    /// Skip initialization
    #[arg(short, long, default_value_t)]
    empty: bool,
```

The code above will set `empty` to `false` when the value wasn't provided in the CLI

https://docs.rs/clap/latest/clap/_derive/index.html#arg-attributes

TimeEngineer: Nice

vxgmichel (changes requested):
I would not expect a server to be started when the `--empty` option is passed, but here's what I got:
```python
± source ./cli/run_testenv.sh --empty
Running server with the process id: 67969 on port: 65247
```
IMO an empty env is more similar to a new user installing parsec on their machine getting invited with a link. That means that the corresponding server is already running another environment (typically another shell running `run_testenv.sh` without `--empty`)

Also I wouldn't make `empty` an argument to `run_local_organization` as we are not running an empty organization but we're creating an empty env.

vxgmichel (approved):
LGTM :+1:

PR #5986, opened 2023-12-13
Harden notary script

FirelightFlagboy:
- Set and use `SCRIPTDIR` to be agnostic of the current working directory (i.e.: being able to execute it outside of `parsec-cloud` repo)
- Rework `Debug tool version` block to oneline tools' version instead of relying on debug output (`set -x`).

  This allow to run the script with `-x` without it being disable by the `set +x`.
- Allow to provide additional args to `Github-CLI` via the optional var `GH_ADDITIONAL_ARGS` (useful to set `--repo=<Owner>/<Repo>`)

### DIFF
--- a/.github/scripts/notary/script.sh
+++ b/.github/scripts/notary/script.sh
@@ -1,5 +1,6 @@
-#!bash
-source .github/scripts/notary/lib.sh
+#!/usr/bin/env bash
+SCRIPTDIR=${SCRIPTDIR:="$(dirname $(realpath -s "$0"))"}
+source "$SCRIPTDIR"/lib.sh
 
 if [ $# -ne 2 ]; then
     echo "usage: $0 <ORGA> <PROJECT_NUMBER>" 1>&2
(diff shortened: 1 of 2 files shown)


.github/scripts/notary/script.sh
```diff
@@ -1,5 +1,6 @@
 #!bash
-source .github/scripts/notary/lib.sh
+SCRIPTDIR=${SCRIPTDIR:=$(dirname $(realpath -s "$0"))}
+source $SCRIPTDIR/lib.sh
```
mmmarcos, line 3:
```suggestion
source "$SCRIPTDIR"/lib.sh
```

mmmarcos (changes requested):
Two more comments:

1. We should fix the files shebang: they should use an absolute path, or better: 
```bash
#!/usr/bin/env bash
```

2. There seems to be an error on line `script.sh`: the `$?` in line 62 refers to the return value of `echo` on line 61 (not the return value of the `add_item_to_project` function!)
https://github.com/Scille/parsec-cloud/blob/2055668e5f074839066beaff3a9d450a39b47ac6/.github/scripts/notary/script.sh#L59-L66
The return value should be stored just after the command:
```sh
    add_item_to_project $PROJECT_ID $ID
    RC=$?
    echo
```

mmmarcos (commented):
The `script.sh` is OK. There's still the shebang of `lib.sh` to be fixed.

mmmarcos (approved):

PR #6019, opened 2023-12-19
[MS] Update grid style

### LINKED ISSUE #5981: [🦋 Final Touch] Update grid item

fabienSvtr:
## Describe your changes

- Update style to match with mockup 
- Unified code between file and user card list

image
image

## Checklist

- [ ] Describe your changes
   
closes #5981

### DIFF
--- a/client/src/components/files/FileCard.vue
+++ b/client/src/components/files/FileCard.vue
@@ -1,12 +1,23 @@
 <!-- Parsec Cloud (https://parsec.cloud) Copyright (c) BUSL-1.1 2016-present Scille SAS -->
 
 <template>
-  <div
-    class="card"
+  <ion-item
+    class="file-card-item ion-no-padding"
+    :class="{ selected: isSelected }"
     @click="$emit('click', $event, file)"
     @mouseenter="isHovered = true"
     @mouseleave="isHovered = false"
   >
+    <div class="card-checkbox">
+      <ion-checkbox
+        aria-label=""
+        class="checkbox"
+        v-model="isSelected"
+        v-show="isSelected || isHovered || showCheckbox"
+        @click.stop
+        @ion-change="$emit('select', file, isSelected)"
+      />
+    </div>
     <div
       class="card-option"
       v-show="isHovered || menuOpened"
@@ -15,9 +26,9 @@
       <ion-icon :icon="ellipsisHorizontal" />
     </div>
     <div class="card-content">
-      <ion-avatar class="card-content-icons">
+      <ion-avatar class="card-content__icons">
         <ion-icon
-          class="card-content-icons__item"
+          class="icon-item"
           :icon="file.isFile() ? document : folder"
         />
         <ion-icon
@@ -27,7 +38,7 @@
         />
       </ion-avatar>
 
-      <ion-title class="card-content__title body-lg">
+      <ion-title class="card-content__title body">
         {{ file.name }}
       </ion-title>
 
@@ -36,27 +47,35 @@
         <span>{{ timeSince(file.updated, '--', 'short') }}</span>
       </ion-text>
     </div>
-  </div>
+  </ion-item>
 </template>
 
 <script setup lang="ts">
 import { Formatters, FormattersKey } from '@/common/injectionKeys';
 import { EntryStat } from '@/parsec';
-import { IonAvatar, IonIcon, IonText, IonTitle } from '@ionic/vue';
+import { IonAvatar, IonCheckbox, IonIcon, IonItem, IonText, IonTitle } from '@ionic/vue';
 import { cloudDone, cloudOffline, document, ellipsisHorizontal, folder } from 'ionicons/icons';
 import { inject, ref } from 'vue';
 
 const isHovered = ref(false);
 const menuOpened = ref(false);
+const isSelected = ref(false);
 
 const props = defineProps<{
   file: EntryStat;
+  showCheckbox: boolean;
 }>();
 
 const emits = defineEmits<{
   (e: 'click', event: Event, file: EntryStat): void;
   (e: 'menuClick', event: Event, file: EntryStat, onFinished: () => void): void;
+  (e: 'select', file: EntryStat, selected: boolean): void;
 }>();
+
+defineExpose({
+  isSelected,
+  props,
+});
 
 // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
 const { timeSince } = inject(FormattersKey)! as Formatters;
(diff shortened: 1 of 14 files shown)


client/src/components/files/FileCard.vue
```diff
@@ -36,28 +47,35 @@
...
     </div>
-  </div>
+  </ion-item>
 </template>
 
 <script setup lang="ts">
 import { Formatters, FormattersKey } from '@/common/injectionKeys';
 import { EntryStat } from '@/parsec';
-import { IonAvatar, IonIcon, IonText, IonTitle } from '@ionic/vue';
+import { IonAvatar, IonCheckbox, IonIcon, IonText, IonTitle } from '@ionic/vue';
```
Max-7, line 56: Missing `IonItem` import. Always keep you dev tools open in your browser :p

client/src/components/files/FileCard.vue
```diff
@@ -1,12 +1,23 @@
 <!-- Parsec Cloud (https://parsec.cloud) Copyright (c) BUSL-1.1 2016-present Scille SAS -->
 
 <template>
-  <div
-    class="card"
+  <ion-item
+    class="card-file-item ion-no-padding"
+    :class="{ selected: isSelected }"
```
Max-7, line 6: The select for `FileCard` seems to be displayed correctly, and the action buttons change depending on the selection, but none of them work.

client/src/views/files/FoldersPage.vue
```diff
@@ -148,17 +148,19 @@
...
             class="folder-grid-item"
             v-for="child in children"
             :key="child.id"
           >
             <file-card
               :file="child"
+              :show-checkbox="selectedFilesCount > 0"
               @click="onFileClick"
               @menu-click="openFileContextMenu"
+              ref="fileItemRefs"
```
Max-7, line 161: This doesn't work, it's going to reference only one element. For `file-list-item` it works because the `v-for` is directly on the element, here the `v-for` is on the parent div. This is probably the reason why the selection is not working for files in grid display.

Max-7 (changes requested):

client/src/components/files/FileCard.vue
fabienSvtr: Sure, done it ^^

client/src/components/files/FileCard.vue
fabienSvtr: After investigating, a props was missing, but now it works.

client/src/views/files/FoldersPage.vue
fabienSvtr: I update this with a `file-list-item-refs` and `file-grid-item-refs`.

client/src/views/files/FoldersPage.vue
```diff
@@ -343,18 +345,30 @@ async function importFiles(): Promise<void> {
...
   allFilesSelected.value = checked;
 }
 
 function getSelectedEntries(): parsec.EntryStat[] {
-  if (!fileItemRefs.value) {
-    return [];
+  const selectedEntries: parsec.EntryStat[] = [];
+
+  if (displayView.value === DisplayState.List) {
+    for (const item of fileListItemRefs.value) {
```
Max-7, line 359: The first `for` is useless,
```typescript
if (displayView.value === DisplayState.List) {
  return fileListItemRefs.value.filter((item) => item.isSelected).map((item) => item.props.file);
}
```
is enough.

client/src/views/files/FoldersPage.vue
```diff
@@ -343,18 +345,30 @@ async function importFiles(): Promise<void> {
...
+      if (item.isSelected) {
+        return fileListItemRefs.value.filter((item) => item.isSelected).map((item) => item.props.file);
+      }
+    }
+  } else {
+    for (const item of fileGridItemRefs.value) {
```
Max-7, line 365: Same here, no need for the for.

client/src/views/files/FoldersPage.vue
```diff
@@ -343,18 +345,30 @@ async function importFiles(): Promise<void> {
...
     item.isSelected = checked;
     item.showCheckbox = checked;
   }
   allFilesSelected.value = checked;
 }
 
 function getSelectedEntries(): parsec.EntryStat[] {
-  if (!fileItemRefs.value) {
-    return [];
+  const selectedEntries: parsec.EntryStat[] = [];
```
Max-7, line 356: Useless variable.

client/src/views/files/FoldersPage.vue
```diff
@@ -226,11 +225,14 @@ const workspaceId = computed(() => {
 });
 const folderInfo: Ref<parsec.EntryStatFolder | null> = ref(null);
 const children: Ref<Array<parsec.EntryStat>> = ref([]);
+const displayView = ref(DisplayState.List);
 const selectedFilesCount = computed(() => {
-  const count = fileItemRefs.value.filter((item) => item.isSelected).length;
-  return count;
+  if (displayView.value === DisplayState.List) {
```
Max-7, line 230: Use the function `getSelectedEntries()`

client/src/views/files/FoldersPage.vue
```diff
@@ -689,7 +703,7 @@ async function openFileContextMenu(event: Event, file: parsec.EntryStat, onFinis
 }
 
 async function resetSelection(): Promise<void> {
-  fileItemRefs.value = [];
+  fileListItemRefs.value = [];
```
Max-7, line 706: Probably a mistake to clear this list. I even think this function is useless. It's just equivalent of calling `selectAllFiles(false)`

Max-7 (commented):
Missing tests to check grid selection for both users and files.

client/src/views/files/FoldersPage.vue
fabienSvtr: After reflexion, we don't need the `resetSelection()` function. In future, selected files will remain selected while the user chooses the card or list view.

client/src/views/files/FoldersPage.vue
```diff
@@ -343,18 +340,19 @@ async function importFiles(): Promise<void> {
 }
 
 function selectAllFiles(checked: boolean): void {
-  for (const item of fileItemRefs.value || []) {
+  for (const item of fileListItemRefs.value || []) {
```
Max-7, line 343: Should check what display mode we're in and set the values for either `fileListItemRefs` or `fileGridItemRefs`. I know, you can't select all files in GridMode, but this function might be used to reset the selection at one time and it should handle GridMode too.

client/tests/e2e/specs/test_active_users_page.ts
```diff
@@ -91,70 +91,85 @@ describe('Check active users page', () => {
...
-    cy.checkToastMessage('success', '2 users have been revoked and can no longer access this organization!');
-    // Ultimately, they probably should not appear anymore
-    cy.get('@userItems').eq(1).find('ion-checkbox').should('not.have.class', 'checkbox-checked');
-    cy.get('@userItems').eq(2).find('ion-checkbox').should('not.have.class', 'checkbox-checked');
+    cy.get('@userItems').eq(2).realHover().find('.checkbox').click();
+    cy.get('#activate-users-ms-action-bar').find('#button-revoke-user').contains('Revoke this user');
+    cy.get('#activate-users-ms-action-bar').find('#button-common-workspaces').contains('See common workspaces');
   });
+
+  // it('Tests context menu', () => {
```
Max-7, line 109: Did you forget to uncomment those?

client/tests/e2e/specs/test_folders_page.ts
```diff
@@ -130,6 +150,20 @@ describe('Check folders page', () => {
     cy.get('.question-modal').find('#next-button').click();
     cy.get('.question-modal').should('not.exist');
     cy.get('@consoleLog').should('have.been.called.with', /File File_[a-z_]+ deleted/);
+
+    // grid view
+    cy.get('#folders-ms-action-bar').find('#grid-view').as('gridButton');
```
Max-7, line 155: Move it to another test.

Max-7 (changes requested):

client/src/views/files/FoldersPage.vue
fabienSvtr: Ok so I added a `if` condition to check that

client/tests/e2e/specs/test_active_users_page.ts
fabienSvtr: Yes, I was to accelerate the test, I uncommented them.

client/tests/e2e/specs/test_folders_page.ts
fabienSvtr: Done

client/src/views/files/FoldersPage.vue
```diff
@@ -345,18 +340,21 @@ async function importFiles(): Promise<void> {
 }
 
 function selectAllFiles(checked: boolean): void {
-  for (const item of fileItemRefs.value || []) {
-    item.isSelected = checked;
-    item.showCheckbox = checked;
+  if (displayView.value === DisplayState.List) {
```
Max-7, line 343: I meant handle the Grid case by selecting/unselecting all items too.

You may be able to do something like
```typescript
for (const item of displayView.value === DisplayState.List ? fileListItemRefs.value : fileGridItemRefs.value) {
      item.isSelected = checked;
      item.showCheckbox = checked;
}
allFilesSelected.value = checked;
```

client/tests/e2e/specs/test_folders_page.ts
Max-7: Naaah. I just meant add another test in the same file to test deletion in grid mode. Something like

```typescript
it('Test delete one file list mode', () => {

});

it('Test delete on file grid mode', () => {

});
```

client/tests/e2e/specs/test_folders_page_grid.ts
Max-7, line 1: Don't need this file, everything can be in the `test_folders_page.ts`.

Max-7 (changes requested):

client/src/views/files/FoldersPage.vue
fabienSvtr: I understand, but in the grid case we don't have a button to check everything, so why do we need to add a grid case in the condition?

Max-7: > I know, you can't select all files in GridMode, but this function might be used to reset the selection at one time and it should handle GridMode too.

The function is called `selectAllFiles`, not `selectAllFilesInListMode`, and it should handle both cases to avoid bugs further down the line where someone is calling this function to reset the selection for example, and forgets to check the result in grid mode.

client/src/views/files/FoldersPage.vue
```diff
@@ -345,18 +340,19 @@ async function importFiles(): Promise<void> {
 }
 
 function selectAllFiles(checked: boolean): void {
-  for (const item of fileItemRefs.value || []) {
+  for (const item of fileListItemRefs.value || fileGridItemRefs.value) {
```
Max-7, line 343: Better to use `displayView` to choose the list to iterate on. I'm not exactly sure what happens with `refs` when they're shown/hidden.

Max-7 (changes requested):

Max-7 (approved):

PR #5842, opened 2023-11-22
Implement RFC 1007 - Frozen users for parsec v2

vxgmichel:
Close #5811
Related #5817

Remaining tasks
- [x] Add news fragment
- [x] Disconnect client when frozen
- [x] Add GUI error message
- [x] Add server administration documentation

PR #6020, opened 2023-12-19
Add CLI commands for frozen users administration API

vxgmichel:
Add the missing CLI from #5842

### DIFF
--- a/parsec/core/cli/freeze_user.py
+++ b/parsec/core/cli/freeze_user.py
@@ -1,0 +1,133 @@
+# Parsec Cloud (https://parsec.cloud) Copyright (c) AGPL-3.0 2016-present Scille SAS
+from __future__ import annotations
+
+from typing import TypedDict, cast
+
+import click
+
+from parsec.api.protocol import OrganizationID, UserID
+from parsec.api.rest import (
+    freeze_user_rep_serializer,
+    freeze_user_req_serializer,
+    user_list_rep_serializer,
+)
+from parsec.cli_utils import cli_exception_handler
+from parsec.core.backend_connection.transport import http_request
+from parsec.core.cli.utils import cli_command_base_options
+from parsec.core.types import BackendAddr
+from parsec.utils import trio_run
+
+
+class User(TypedDict):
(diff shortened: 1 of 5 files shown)


parsec/core/cli/freeze_user.py
```diff
@@ -0,0 +1,133 @@
...
+    **kwargs: object,
+) -> None:
+    """List users of an organization using the administration API.
+
+    The following information is returned for each user:
+
+    \b
+    • USER_NAME <USER_EMAIL>
+      - Parsec ID: USER_ID
+      - Status: Frozen/Not frozen
```
mmmarcos, line 79: I wonder if this might not be annoying in case you want to `grep` the output :thinking: (typically for large organizations)

Maybe we can add an option (similar to `git log --oneline` or `git rev-parse --short`) to have a one-line-per-user output:
```
$ parsec core list_users --oneline -B "parsec://$server" -T $administration_token $organization
Alice <alice@example.com> (67ee640058aa45ca9281717c866baa06) Not frozen
Bob <bob@example.com> (0d22530361484b86b28e5b2c3a089772) Frozen
David <david@example.com> (1f1b81052fa947babdbacd6b147b3622) Not Frozen

$ parsec core list_users --oneline -B "parsec://$server" -T $administration_token $organization | grep "Frozen"
Bob <bob@example.com> (0d22530361484b86b28e5b2c3a089772) Frozen
```

IDK what would be the better format though:
- maybe Parsec ID could be first column, because same length? 
- maybe we can use `\t` for separator?
- maybe there is a `click` option to format output as a table?

mmmarcos (commented):
LGTM. Just a single comment regarding the output.

parsec/core/cli/freeze_user.py
mmmarcos: I'll answer myself: actually we can `grep` lines before (`-B`) and after (`-A`). So something like `grep -i -B 2 frozen` should be enough to filter by frozen status.

mmmarcos (approved):

PR #6025, opened 2023-12-19
[MS] fix HomePage transitions

### LINKED ISSUE #6003: Re-enable animation in homepage between devices list and login
Turned of in #5983

NicoTuxx:
closes #6003

### DIFF
--- a/client/src/transitions/SlideHorizontal.vue
+++ b/client/src/transitions/SlideHorizontal.vue
@@ -1,38 +1,54 @@
 <!-- Parsec Cloud (https://parsec.cloud) Copyright (c) BUSL-1.1 2016-present Scille SAS -->
 <template>
-  <transition :name="!props.reverseDirection ? 'slide-left' : 'slide-right'">
+  <transition :name="getName()">
     <slot />
   </transition>
 </template>
 
 <script setup lang="ts">
+import { Position } from '@/transitions/types';
 import { defineProps } from 'vue';
 
 const props = defineProps<{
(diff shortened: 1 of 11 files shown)


fabienSvtr (approved):

client/src/transitions/SlideHorizontal.vue
```diff
@@ -1,38 +1,54 @@
...
   </transition>
 </template>
 
 <script setup lang="ts">
+import { Position } from '@/transitions/types';
 import { defineProps } from 'vue';
 
 const props = defineProps<{
-  reverseDirection?: boolean;
+  from: Position;
```
Max-7, line 13: I'm not a big fan of having two props to define the transition behavior, because it delegates the configuration to the caller.

Which means you get a lot of logic in the caller:
```html
<slide-horizontal
  :from="state === HomePageState.OrganizationList ? Position.Left : Position.Right"
  :to="state === HomePageState.OrganizationList ? Position.Right : Position.Left"
>
```
and it can lead to weird things:
```html
<slide-horizontal
  :from="Position.Left"
  :to="Position.Left"
>
```
Why not define instead a `TransitionDirection` as `LeftToRight` and `RightToLeft`? It reduces the amount of logic, it's easier to understand, and it's less prone to errors. It does require a bit more logic in the component itself but that's ok.

Max-7 (changes requested):

Max-7 (approved):

PR #6036, opened 2023-12-22
Refactor run-testenv

### LINKED ISSUE #6034: [CLI testenv] Allow to use already running testbed server
Currently the sub-command `run-testenv` start a testbed server everytime meaning it require to have the python dependencies installed with the rust-extension compiled.

I find it not satisfactory in the CI because in `ci-rust` we already have a testbed server running without needing to compile it.

## Proposal

Like the rust tests that use the testbed server, `run-testenv` could skip starting the testbed server if we set the env variable `TESTBED_SERVER` and use the URL contains in it.

FirelightFlagboy:
- Split `run_testenv` function into smaller functions.
- Add logic to skip starting the testbed server if the `TESTBED_SERVER` environment variable is set using that URL instead.
- CI: Don't install python dependencies on the rust workflow

Closes #6034

### DIFF
--- a/cli/Cargo.toml
+++ b/cli/Cargo.toml
@@ -11,7 +11,7 @@ repository.workspace = true
 # See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
 
 [features]
-testenv = []
+testenv = ["dep:url"]
 
 [dependencies]
 libparsec = { workspace = true }
@@ -25,6 +25,7 @@ serde_json = { workspace = true }
 terminal-spinners = { workspace = true }
 tokio = { workspace = true, features = ["rt-multi-thread", "macros"] }
 uuid = { workspace = true }
+url = { workspace = true, optional = true}
 
 [dev-dependencies]
 libparsec = { workspace = true, features = ["cli-tests"] }
@@ -32,3 +33,4 @@ libparsec = { workspace = true, features = ["cli-tests"] }
 assert_cmd = { workspace = true }
 predicates = { workspace = true, features = ["regex"] }
 rstest = { workspace = true }
+uuid = { workspace = true, features = ["v6", "std", "rng"] }
--- a/cli/src/run_testenv.rs
+++ b/cli/src/run_testenv.rs
@@ -39,7 +39,7 @@ pub struct RunTestenv {
     /// Main process id.
     /// When this process stops, the server will be automatically killed
     #[arg(long)]
-    main_process_id: u32,
+    main_process_id: Option<u32>,
     /// Skip initialization
     #[arg(short, long, default_value_t)]
     empty: bool,
@@ -209,9 +209,8 @@ async fn register_new_user(
 pub async fn initialize_test_organization(
     client_config: ClientConfig,
     addr: BackendAddr,
+    organization_id: OrganizationID,
 ) -> anyhow::Result<[Arc<LocalDevice>; 3]> {
-    let organization_id: OrganizationID = "Org".parse().expect("Unreachable");
-
     // Create organization
     let bootstrap_token =
         create_organization_req(&organization_id, &addr, DEFAULT_ADMINISTRATION_TOKEN).await?;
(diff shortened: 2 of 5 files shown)


TimeEngineer:
Can you split into smaller PR ?

TimeEngineer:
What's the difference between set your own `TestbedServer` and `--empty` flag ?

FirelightFlagboy:
> What's the difference between set your own `TestbedServer` and `--empty` flag ?

I you don't pass `empty`, we still have the code-block that initialize the server with random data

cli/src/run_testenv.rs
```diff
@@ -278,111 +280,70 @@ pub async fn initialize_test_organization(
     Ok([alice_device, other_alice_device, bob_device])
 }
 
+pub enum TestenvConfig {
+    ServerAlreadyRunning(String),
```
TimeEngineer, line 284: Why String ?

FirelightFlagboy: That value come from `std::env::var` which directly return a string

cli/src/tests.rs
```diff
@@ -7,44 +7,58 @@ use std::{
...
-    new_environment(tmp_dir, source_file, stop_after_process, false).await?;
-
-    initialize_test_organization(
-        ClientConfig::default(),
-        BackendAddr::new("127.0.0.1".into(), Some(port), false),
-    )
-    .await
+    config: TestenvConfig,
+) -> anyhow::Result<(BackendAddr, [Arc<LocalDevice>; 3])> {
+    println!("Hello");
```
TimeEngineer, line 53: Remove

cli/src/run_testenv.rs
TimeEngineer: Isn't it better to be parsed before ?

cli/src/run_testenv.rs
```diff
@@ -278,111 +277,70 @@ pub async fn initialize_test_organization(
...
```
mmmarcos, line 281: I'd call it something like 
```suggestion
    ConnectToServer(String),
```

cli/Cargo.toml
```diff
@@ -11,7 +11,7 @@ repository.workspace = true
 # See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
 
 [features]
-testenv = []
+testenv = ["dep:url"]
```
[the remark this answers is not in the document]
TimeEngineer: That means if you use the `testenv` `feature`, you will have `url` as `dependency`

cli/src/tests.rs
FirelightFlagboy: doen

cli/src/run_testenv.rs
FirelightFlagboy: done

cli/src/tests.rs
```diff
@@ -7,44 +7,56 @@ use std::{
...
-        BackendAddr::new("127.0.0.1".into(), Some(port), false),
-    )
-    .await
+    config: TestenvConfig,
+) -> anyhow::Result<(BackendAddr, [Arc<LocalDevice>; 3])> {
+    let url = new_environment(tmp_dir, source_file, config, false)
+        .await?
+        .unwrap();
+
+    println!("Initializing test organization to {}", url);
```
TimeEngineer, line 56:
```suggestion
    println!("Initializing test organization to {url}");
```

cli/src/tests.rs
```diff
@@ -81,18 +93,19 @@ async fn list_devices(tmp_path: TmpPath) {
...
+        .unwrap();
+    set_env(&tmp_path_str, &url);
 
     Command::cargo_bin("parsec_cli")
         .unwrap()
         .args([
             "create-organization",
             "--organization-id",
-            "Org1",
+            unique_org_id().to_string().as_str(),
```
TimeEngineer, line 108:
```suggestion
            &unique_org_id().to_string(),
```

mmmarcos (approved):
LGTM

PR #6053, opened 2023-12-28
Improve glossary

mmmarcos:
- Move from table to sections to support linkable terms
- Organize terms into categories for easier read
- Fix and improve some descriptions (e.g. Administrator) and remove non-relevant information
- Move (relevant) translations to a different file

### DIFF
--- a/docs/development/glossary.md
+++ b/docs/development/glossary.md
@@ -2,6 +2,30 @@
 
 # Glossary
 
+- [Glossary](#glossary)
+  - [General terms](#general-terms)
+    - [Organization](#organization)
+    - [Workspace](#workspace)
+    - [User](#user)
+    - [Device](#device)
+  - [User Profile](#user-profile)
+    - [Standard](#standard)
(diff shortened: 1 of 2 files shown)


FirelightFlagboy (commented):
I'm not sure about splitting the glossary and the translation.
I find it useful to have the translation along side the term and it's definition.

With how the glossary is now written, we could use a Table Of Content (TOC) ?

mmmarcos:
> I'm not sure about splitting the glossary and the translation. I find it useful to have the translation along side the term and it's definition.

The thing is that 1) it is not necessary to note every term, only those who may cause doubt during translation, and 2) we are considering moving the glossary to the user docs, and the translations will not be needed there. Also, having it in the same file would not be well adapted if we decide to add other languages later.

> With how the glossary is now written, we could use a Table Of Content (TOC) ?

Done :+1:

FirelightFlagboy (approved):

PR #6068, opened 2024-01-02
Cli: list-invitations

### LINKED ISSUE #6067: CLI `list-invitations`

TimeEngineer:
closes #6067

### DIFF
--- a/cli/src/list_invitations.rs
+++ b/cli/src/list_invitations.rs
@@ -1,0 +1,75 @@
+// Parsec Cloud (https://parsec.cloud) Copyright (c) BUSL-1.1 2016-present Scille SAS
+
+use clap::Args;
+use std::path::PathBuf;
+
+use libparsec::{
+    authenticated_cmds::latest::invite_list::{self, InviteListItem, InviteListRep},
+    get_default_config_dir, InvitationStatus,
+};
+
+use crate::utils::*;
+
+#[derive(Args)]
+pub struct ListInvitations {
+    /// Parsec config directory
+    #[arg(short, long, default_value_os_t = get_default_config_dir())]
+    config_dir: PathBuf,
+    /// Device slughash
+    #[arg(short, long)]
+    device: Option<String>,
+}
+
+pub async fn list_invitations(list_invitations: ListInvitations) -> anyhow::Result<()> {
+    load_device_and_run(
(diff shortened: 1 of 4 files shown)


cli/src/list_invitations.rs
```diff
@@ -0,0 +1,75 @@
...
+                        "Server error while listing invitations: {rep:?}"
+                    ));
+                }
+            };
+
+            handle.done();
+
+            if invitations.is_empty() {
+                println!("No invitation.");
+            }
```
FirelightFlagboy, line 45:
```suggestion
                println!("No invitation.");
                return Ok(());
            }
```

I would prefer to have an early return or have an `if .. else ..` block

cli/src/list_invitations.rs
```diff
@@ -0,0 +1,75 @@
...
+                        (token, status, "device".into())
+                    }
+                };
+
+                let token = token.as_simple();
+
+                let display_status = match status {
+                    InvitationStatus::Idle => format!("{YELLOW}idle{RESET}"),
+                    InvitationStatus::Ready => format!("{GREEN}idle{RESET}"),
+                    InvitationStatus::Deleted => format!("{RED}idle{RESET}"),
```
FirelightFlagboy, line 65: Is it expected that all text is `idle` ?

cli/src/tests.rs
```diff
@@ -253,3 +253,58 @@ async fn cancel_invitation(tmp_path: TmpPath) {
             "Deleting invitation\nInvitation deleted",
         ));
 }
+
+#[rstest::rstest]
+#[tokio::test]
+async fn list_invitations(tmp_path: TmpPath) {
```
FirelightFlagboy, line 259: Is it possible to have the other cases tested (`Deleted` & `Ready`) ?

cli/src/list_invitations.rs
TimeEngineer: No lol

cli/src/tests.rs
TimeEngineer: I don't know how to reproduce them

TimeEngineer: If you delete the `invitation`, it will be absent instead of the status `deleted`

FirelightFlagboy (approved):

PR #6073, opened 2024-01-03
Cli: import-recovery-device

### LINKED ISSUE #6072: CLI: `import-recovery-device`

TimeEngineer:
closes #6072

### DIFF
--- a/cli/src/import_recovery_device.rs
+++ b/cli/src/import_recovery_device.rs
@@ -1,0 +1,51 @@
+// Parsec Cloud (https://parsec.cloud) Copyright (c) BUSL-1.1 2016-present Scille SAS
+
+use clap::Args;
+use std::path::{Path, PathBuf};
+
+use libparsec::{get_default_config_dir, load_recovery_device, DeviceAccessStrategy};
+
+use crate::utils::*;
+
+#[derive(Args)]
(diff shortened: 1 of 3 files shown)


FirelightFlagboy (commented):
Looks good but can't approve since we have a commit the first commit that already in a different PR.

Maybe you could rebase this one directly with `master` ? it seems doable

cli/src/import_recovery_device.rs
```diff
@@ -0,0 +1,51 @@
...
+
+    let password = choose_password()?;
+
+    let key_file = libparsec::get_default_key_file(&import_recovery_device.config_dir, &device);
+
+    let access = DeviceAccessStrategy::Password { key_file, password };
+
+    let handle = start_spinner("Saving new device");
+
+    libparsec::save_device(Path::new(""), &access, &device).await?;
```
FirelightFlagboy, line 44: Should we use `import_recovery_device.config_dir` instead of `Path::new("")` ?
The arg isn't used outside of testbed feature but :shrug:

FirelightFlagboy (approved):

cli/src/import_recovery_device.rs
TimeEngineer: Maybe open an issue about that ?

PR #6076, opened 2024-01-03
[MS] Refactored how translations are being done accross the app

Max-7:
Done as a follow-up to https://github.com/Scille/parsec-cloud/commit/11d7ee29b6f148c4995aa6359cfdc7b00fc938c0.

The main idea is that `vue-i18n` can only be used while inside a component, but we have translations that are done outside of components (filesize, dates, validators reason to reject an input, ...). We usually got around by passing the `t` from `vue-i18n` around, but it causes multiple problems:
1. Some things are independant and it's very annoying to give them the `t` (validators for example)
2. It makes the API a bit more complicated (`$d` can't format luxon dates, you have to pass a `ComposerTranslation` around without really knowing what it is, ...). Even the setup with `import { useI18n} from 'vue-i18n'; const { t } = useI18n();` is a bit weird.
3. It make the tests unrealiable and more complicated to write since we had to mock those calls

i18n init is now done separately, meaning we can save the i18n object ourselves, and we can init the translation from the tests (which was not possible before because we would have to init the whole app). The unit/component tests can now use the real translation strings. And translations can be done outside of a component.

- [x] Describe your changes


client/tests/unit/specs/testValidators.spec.ts
```diff
@@ -7,13 +7,13 @@ const VALID_TOKEN = 'a'.repeat(32);
...
-        },
-      };
-    });
+    // vi.mock('@/services/translation', () => {
+    //   return {
+    //     getI18n: (): any => {
+    //       return { global: { t: (key: string) => key } };
+    //     },
+    //   };
+    // });
```
NicoTuxx, line 16: do we need to keep this comment?

NicoTuxx (approved):

PR #6081, opened 2024-01-03
Cli stats server

### LINKED ISSUE #6080: CLI: `stats_server`

TimeEngineer:
closes #6080

### DIFF
--- a/cli/src/stats_server.rs
+++ b/cli/src/stats_server.rs
@@ -1,0 +1,90 @@
+// Parsec Cloud (https://parsec.cloud) Copyright (c) BUSL-1.1 2016-present Scille SAS
+
+use clap::Args;
+use reqwest::{Client, Response};
+use serde_json::Value;
+
+use libparsec::{BackendAddr, DateTime};
+
+#[derive(Args)]
+pub struct StatsServer {
+    /// Server address (e.g: parsec://127.0.0.1:6770?no_ssl=true)
+    #[arg(short, long)]
+    addr: BackendAddr,
+    /// Administration token
+    #[arg(short, long)]
+    token: String,
+    /// Output format (json/csv)
+    #[arg(short, long, default_value_t = Format::Json)]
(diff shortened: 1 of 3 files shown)


cli/src/stats_server.rs
```diff
@@ -0,0 +1,94 @@
...
+    addr: BackendAddr,
+    /// Administration token
+    #[arg(short, long)]
+    token: String,
+    /// Output format (json/csv)
+    #[arg(short, long, default_value_t = Format::Json)]
+    format: Format,
+    /// Ignore everything after this date (e.g: 2024-01-01T00:00:00-00:00)
+    #[arg(long)]
+    at: Option<DateTime>,
```
FirelightFlagboy, line 22: I would prefer `start_date` as the doc-comment say it ignore everything before that date.

`at` could confuse the user that the stat will only be for a specific date

cli/src/stats_server.rs
```diff
@@ -0,0 +1,94 @@
...
+impl std::fmt::Display for Format {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        write!(
+            f,
+            "{}",
+            match self {
+                Self::Json => "json",
+                Self::Csv => "csv",
+            }
+        )
```
FirelightFlagboy, line 52:
```suggestion
        f.write_str(match self {
            Self::Json => "json",
            Self::Csv => "csv",
        })
```

cli/src/stats_server.rs
TimeEngineer: Not before, after

TimeEngineer: Maybe you prefer `ignore-after` ?

FirelightFlagboy: Oops, `end-date` then

TimeEngineer: Done

PR #6095, opened 2024-01-05
CLI: create-workspace

### LINKED ISSUE #6094: `create_workspace`

TimeEngineer:
closes #6094


cli/src/utils.rs
```diff
@@ -108,6 +109,87 @@ where
     Ok(())
 }
 
+pub async fn load_client_and_run<F, Fut>(
```
FirelightFlagboy, line 112: This function is very similar to `load_device_and_run`, can it be possible to shared identical code between the two ?

TimeEngineer: Done

PR #6101, opened 2024-01-05
Cli list users

TimeEngineer:
It seems it is needed to retrieve the users id somewhere (for example for sharing workspace)

### DIFF
--- a/cli/src/list_users.rs
+++ b/cli/src/list_users.rs
@@ -1,0 +1,56 @@
+// Parsec Cloud (https://parsec.cloud) Copyright (c) BUSL-1.1 2016-present Scille SAS
+
+use clap::Args;
+use std::path::PathBuf;
+
+use libparsec::get_default_config_dir;
+
+use crate::utils::*;
+
+#[derive(Args)]
+pub struct ListUsers {
+    /// Parsec config directory
+    #[arg(short, long, default_value_os_t = get_default_config_dir())]
+    config_dir: PathBuf,
+    /// Device slughash
+    #[arg(short, long)]
+    device: Option<String>,
+    /// Skip revoked users
+    #[arg(short, long, default_value_t)]
+    skip_revoked: bool,
+}
+
+pub async fn list_users(list_users: ListUsers) -> anyhow::Result<()> {
+    let ListUsers {
+        config_dir,
+        device,
+        skip_revoked,
+    } = list_users;
+
+    load_client_and_run(config_dir, device, |client| async move {
+        client.user_ops.sync().await?;
+        let users = client.certificates_ops.list_users(skip_revoked, None, None).await?;
+
+        if users.is_empty() {
+            println!("No users found");
+        } else {
+            let n = users.len();
+            println!("Found {GREEN}{n}{RESET} user(s):");
+            for info in users {
+                let id = info.id;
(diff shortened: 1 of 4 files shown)


cli/src/list_users.rs
```diff
@@ -0,0 +1,55 @@
...
+    device: Option<String>,
+    /// Skip revoked users
+    #[arg(short, long, default_value_t)]
+    skip_revoked: bool,
+    /// Result offset
+    #[arg(short, long)]
+    offset: Option<usize>,
+    /// Result limit
+    #[arg(short, long)]
+    limit: Option<usize>,
```
FirelightFlagboy, line 26: I'm not sure if it's useful to provide that to the user.

What was your plan with this ?
How the user would have used those options ?

FirelightFlagboy (commented):
I don't see tests for this sub-command, is it something you have forgotten ?

cli/src/list_users.rs
TimeEngineer: Idk, it was present with `human_find`

FirelightFlagboy: Then how the user will use those options ?

TimeEngineer: When he needs to paginate :man_shrugging:

TimeEngineer: Do you suggest to remove it ?

FirelightFlagboy: I think it would be best to do so

cli/src/tests.rs
```diff
@@ -558,6 +558,43 @@ async fn create_workspace(tmp_path: TmpPath) {
...
+    Command::cargo_bin("parsec_cli")
+        .unwrap()
+        .args([
+            "create-workspace",
+            "--device",
+            &alice.slughash(),
+            "--name",
+            "new-workspace",
+        ])
+        .unwrap();
```
FirelightFlagboy, line 587: Why creating a workspace _magically_ added 3 users ?

cli/src/tests.rs
```diff
@@ -558,6 +558,43 @@ async fn create_workspace(tmp_path: TmpPath) {
...
+
+    Command::cargo_bin("parsec_cli")
+        .unwrap()
+        .args(["list-users", "--device", &alice.slughash()])
+        .assert()
+        .stdout(predicates::str::contains(format!(
+            "Found {GREEN}3{RESET} user(s)",
+        )));
```
FirelightFlagboy, line 595: Could you check the user list ?

TimeEngineer: I can't know the users id

cli/src/tests.rs
TimeEngineer: I have to investigate I guess

cli/src/tests.rs
FirelightFlagboy: Is it possible to only check the username (handle) ?

TimeEngineer: I prefer to know that there are all users :thinking:

cli/src/tests.rs
TimeEngineer: I found that a `sync` was missing, and the `create_workspace` `sync` it

FirelightFlagboy (approved):

PR #6120, opened 2024-01-09
CLI: enhance list-workspaces by adding self role

TimeEngineer:


FirelightFlagboy (approved):

FirelightFlagboy:
No test need to be udpated ?

TimeEngineer:
> No test need to be udpated ? that feels strange since you've added more information on the output

The test checks only that the workspace has been created, not the full line, because we don't know the id, but I can add the role

PR #6128, opened 2024-01-10
Add comment for `UserOps`' `sync` method

TimeEngineer:


libparsec/crates/client/src/user_ops/mod.rs
```diff
@@ -82,14 +82,15 @@ impl UserOps {
...
 
-    pub async fn sync(&self) -> Result<(), SyncError> {
-        sync::sync(self).await
-    }
-
     /*
      * Public interface
      */
 
+    // TODO: It has been moved in the public interface, because some of the commands in the CLI didn't sync automatically
```
FirelightFlagboy, line 89: What should be done ? `sync` should not be publicly exposed ?

TimeEngineer: I don't know but I have changed something, that's why I mention it

TimeEngineer: It's better to know no ?

TimeEngineer: According to @touilleMan it's something low level that we don't want to expose, but it should be fixed if we want to keep private.

FirelightFlagboy: Then can you update the comment to state that the method is currently expose but it's not something we want ?

TimeEngineer: Sure

TimeEngineer: Done

PR #6165, opened 2024-01-15
[MS] Fixes animation for toast progress bar

### LINKED ISSUE #6163: [🐛 | Bug]: Toast progress bar doesn't take the duration into account
### Parsec version tested on:

3.0.0

### Platforms tested on:

Linux

### Bug description:

The duration used is the default duration, but this duration can be set individually for each toast, meaning the progress bar will either be too fast or too slow.

### Relevant output:

_No response_

Max-7:
Closes #6163 

It would probably be easier if toast where our own component, we could probably just bind the duration with Vue. Sadly, they're created by the toastController.

The idea behind this fix is to assigned a unique id to each toast, and dynamically add a `<style>` tag in the `<head>` of the page that contains the CSS with the appropriate duration. The style is removed once the toast is closed.

- [X] Describe your changes

### DIFF
--- a/client/src/services/toastManager.ts
+++ b/client/src/services/toastManager.ts
@@ -5,7 +5,7 @@ import { translate } from '@/services/translation';
 import { toastController } from '@ionic/vue';
 import { checkmark, closeCircle, information, warning } from 'ionicons/icons';
 
-const DEFAULT_TOAST_DURATION = 3000;
+const DEFAULT_TOAST_DURATION = 5000;
 
 export class ToastManager {
   toasts: HTMLIonToastElement[];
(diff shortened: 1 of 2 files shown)


NicoTuxx:
Maybe use a ref instead of IDs? (you can look at how we modify css values in the sidebarMenu component for an example)

Max-7:
So...
Tried to get the `toast` variable and add style directly to it to avoid adding/removing the style in the header. But since it's a shadow tree, it seems that I cannot update the element style because it's a `::after` selector (mix of shadow tree being hard to access + read only + CSS only selector).

```typescript
`#${toastId} {
  &.active {
    &::part(container) {
      &::after {
        width: 0%;
        transition: width ${duration}s ease-in-out;
      }
    }
  }
}`;
```
this does get applied to the toast but for some reason the transition doesn't work. If you remove the `width: 0%;` it does change the does, which shows that the style is indeed applied, but the transition does nothing.

Max-7:
Instead of all the black magic, I've instead added a `ms-toast-duration` variable in CSS. The caveat is that we can only ever show one toast at a time because the variable will apply to all toasts, but it does work correctly.

I've also fixed a bug where the animation of a toast would play at its creation instead of when it's shown.

PR #6171, opened 2024-01-16
[MS] Fixed homepage with no devices/one device

### LINKED ISSUE #6170: Fixed new homepage bugs when only one device
With only one device, the homepage goes straight to the login page, which means it's impossible to create or join an org.

Also, the sort button is hidden but it's in the same block as the join/create org button, meaning that if you have only one device, you can't create or join an org.

The search input should also be hidden.

Max-7:
Closes #6170 

This fixes multiple issues on the home page:
1. Create/join buttons when there are no devices work properly
2. Made it possible to create/join an org with only one device
3. Displays the create/join button in all cases

To test, in `OrganizationListPage` in the `refreshDeviceList`, you can add after the `deviceList.value = await listAvailableDevices();`:

`deviceList.value = [];` to check with no devices
`deviceList.value = [deviceList.value[0]];` to check with one device

- [X] Describe your changes

### DIFF
--- a/client/src/views/home/HomePageButtons.vue
+++ b/client/src/views/home/HomePageButtons.vue
@@ -4,7 +4,7 @@
   <ion-list class="container">
     <ion-item
       class="container__item"
-      @click="onClick(HomePageAction.CreateOrganization)"
+      @click="clicked(HomePageAction.CreateOrganization)"
     >
       <ion-icon
         :icon="addCircle"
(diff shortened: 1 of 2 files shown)


fabienSvtr (commented):
LGTM 

Suggestion: From one organization, the user can see the search bar.
`<ms-search-input `
  `:label="$t('HomePage.organizationList.search')"`
  `v-model="searchQuery"`
  `id="ms-search-input"`
  `v-show="deviceList.length > 1"` // remove this part
 `/>`

fabienSvtr (approved):

PR #6177, opened 2024-01-16
Make write access to the Windows registry more robust

### LINKED ISSUE #6142: [🐛 | Bug]: missing registry key
### Parsec version tested on:

2.16.3

### Platforms tested on:

Windows

### Bug description:

`add_parsec_mountpoint_directory_to_quick_access` function writes registry entry but keys can be missing which results in an error that makes it impossible to launch the client.
<https://github.com/Scille/parsec-cloud/blob/v2.16.3/parsec/core/win_registry.py#L276>

`winreg_write_user_dword` function should create registry key before insert value.

### Relevant output:

image

vxgmichel:
Fix #6142

### DIFF
--- a/parsec/core/win_registry.py
+++ b/parsec/core/win_registry.py
@@ -61,7 +61,7 @@ def try_winreg() -> ModuleType | None:
         return None
 
 
-def winreg_read_user_dword(key: str, name: str) -> object:
+def winreg_read_user_dword(key: str, name: str) -> int:
     winreg = get_winreg()
     hkcu = winreg.HKEY_CURRENT_USER
     with winreg.OpenKey(hkcu, key, access=winreg.KEY_READ) as key_obj:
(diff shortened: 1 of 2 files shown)


parsec/core/win_registry.py
```diff
@@ -74,22 +74,25 @@ def winreg_read_user_dword(key: str, name: str) -> object:
...
 def winreg_delete_user_dword(key: str, name: str) -> None:
     winreg = get_winreg()
     hkcu = winreg.HKEY_CURRENT_USER
-    with winreg.OpenKey(hkcu, key, access=winreg.KEY_SET_VALUE) as key_obj:
-        winreg.DeleteValue(key_obj, name)
+    try:
+        with winreg.OpenKey(hkcu, key, access=winreg.KEY_SET_VALUE) as key_obj:
+            winreg.DeleteValue(key_obj, name)
+    except OSError:
+        pass
```
mmmarcos, line 95: No need to log as info or warning to know when a key could not be deleted and remains in the registry?

mmmarcos (approved):
LGTM

parsec/core/win_registry.py
vxgmichel: Right I wanted to catch `FileNotFound` errors when the key does not exist (hence nothing to delete). But this `except OSError` is probably too much of a catchall

Issue #5626, opened 2023-10-25
Handle multi-organization


Max-7:
Be able to log in into multiple organization at the same time. Be able to switch between orgs.

Max-7:
No idea how to handle this. Since the app is stateless (it gets its information from the URL and calls the bindings), my first idea was to store the whole router state, and create a new one for the new org. Something like:

```typescript
let orgStates = [];

function switchToLoginPage() {
    orgStates.push({clientHandle: router.clientHandle, history: router.history, currentRoute: router.currentRoute});
    router.clearHistory();
    router.replace('home');
}

function getLoggedInOrgs() {
    return orgStates.map(async (state) => state.clientHandle);
}

function switchToOrg(clientHandle) {
    const state = orgStates.find((state) => state.clientHandle === clientHandle);

    if (state) {
        orgStates.push({clientHandle: router.clientHandle, history: router.history, currentRoute: router.currentRoute});

        orgStates.remove(state);
        router.history = state.history;
        router.clientHandle = state.clientHandle;
        router.replace(currentRoute);
    }
}
```
But this doesn't seem possible with vue-router. The history cannot be overwritten or cleared, which is a big problem because we don't want to be able to navigate between orgs with the history.

Max-7:
Create the branch https://github.com/Scille/parsec-cloud/tree/ms-multi-org, but what's on it doesn't work.

The idea was to use multiple apps and mount/unmount them depending on the state. Use a `LoginApp` for the main page, and then a `LoggedInApp` for each connected device, and mount/unmount each app when logging in/logging out or switching org. Each app has its own router, meaning the history between them should be preserved.
Sadly, Vue does not handle unmounting well, or rather, you can't re-mount an app you unmounted, you have to create a new one from scratch, which is frankly useless.

I then tried by mounting each app once on its own element in ths DOM and switch the element visibility depending on which app is active. But this also fails, ionic elements like popovers or modals don't seem to like it at all.

tl;dr; I've tried a few things to get the multi-org to work but it doesn't and I'm all out of ideas.

PR #6181, opened 2024-01-17
[MS] Refactored how routing is done throughout the app

Max-7:
Done in preparation for #5626 (routing can be a pain for multi-org).

The initial goal was just to look at routing from a multi-org point of view to see what could be changed. I ended up being a bit carried away and redid how routing is done throughout the whole app.

1. Updated how initialization is done
2. Use enums for routes, a bit less ugly than having raw strings
3. Fixed a potential bug with the init in Cypress where the app could be a bit too slow to attach the `nextStageHook` function to the window.
4. Added functions to handle navigation and checks, so we don't have to import vue-router and do things by hand everytime

- [X] Describe your changes


client/tests/e2e/specs/test_home_page.ts
```diff
@@ -37,15 +37,16 @@ describe('Check organization list', () => {
...
+  // it('Go to login page and enter password', () => {
+  //   cy.get('.organization-list-row__col').should('have.length', 5);
+  //   cy.contains('Boby McBobFace').click();
+  //   cy.get('.login-button').should('have.class', 'button-disabled');
+  //   cy.get('#ms-password-input').find('input').invoke('attr', 'type').should('eq', 'password');
+  //   cy.get('#ms-password-input').find('input').type('P@ssw0rd.');
+  //   cy.get('.login-button').should('not.have.class', 'button-disabled');
+  //   cy.get('.login-button').click();
+  //   cy.get('.topbar-left__breadcrumb').contains('My workspaces');
+  // });
```
NicoTuxx, line 49: why is this commented?

NicoTuxx (approved):

PR #6188, opened 2024-01-19
Add `show_envvar` to display env variables in help

### LINKED ISSUE #6173: Add documentation for PARSEC_SPONTANEOUS_ORGANIZATION_BOOTSTRAP
The ability for regular users to create organizations within a Parsec Server instance is configured via the `PARSEC_SPONTANEOUS_ORGANIZATION_BOOTSTRAP` variable during deployment. 

This is not well documented in [Parsec Server configuration](https://docs.parsec.cloud/en/latest/adminguide/hosting/index.html#parsec-server-configuration) and it is particularly important for self-hosted deployments.

```
# Allow bootstrapping organizations without prior creation.
PARSEC_SPONTANEOUS_ORGANIZATION_BOOTSTRAP=false
```
We should add some more detail to the variable description: briefly explain what does it means to bootstrap an organization (link to the relevant section in the docs?), which users/profiles are allowed to do it, what is the default behavior (if variable is not defined), etc..

Also, maybe add a Warning notice suggesting to set the value to `false` for on-premise deployments.

mmmarcos:
Also, update Administrator Guide:
- Remove `PARSEC_SPONTANEOUS_ORGANIZATION_BOOTSTRAP` from docs since it has the expected default value (False)
- Add note explaining how to get the full list of available env var

Closes #6173

### DIFF
--- a/docs/adminguide/hosting/parsec.env
+++ b/docs/adminguide/hosting/parsec.env
@@ -3,26 +3,24 @@ PARSEC_HOST=0.0.0.0
 
 # The SSL key file.
 PARSEC_SSL_KEYFILE=/run/secrets/parsec-pem-key
+
 # The SSL certificate file.
 PARSEC_SSL_CERTFILE=/run/secrets/parsec-pem-crt
+
 # Enforce HTTPS by redirecting HTTP request.
 PARSEC_FORWARD_PROTO_ENFORCE_HTTPS=X-Forwarded-Proto:https
 
 # The granularity of Error log outputs.
 PARSEC_LOG_LEVEL=WARNING
+
 # The log formatting to use (`CONSOLE` or `JSON`).
 PARSEC_LOG_FORMAT=CONSOLE
+
 # The log file to write to (default to `stderr`).
 # PARSEC_LOG_FILE
 
-# The URL to reach Parsec server
+# The URL to reach Parsec server.
 PARSEC_BACKEND_ADDR=parsec://127.0.0.1:6777
-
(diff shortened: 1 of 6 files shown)


FirelightFlagboy (approved):
LGTM

docs/adminguide/hosting/parsec.env
```diff
@@ -3,27 +3,25 @@ PARSEC_HOST=0.0.0.0
...
+
 # The log file to write to (default to `stderr`).
 # PARSEC_LOG_FILE
 
-# The URL to reach Parsec server
+# The URL to reach Parsec server.
 PARSEC_BACKEND_ADDR=parsec://127.0.0.1:6777
 
-# Allow organization bootstrap without prior creation.
-PARSEC_SPONTANEOUS_ORGANIZATION_BOOTSTRAP=false
```
vxgmichel, line 22: I think it's still useful to have those variables documented in the template.

Maybe:
```
# Uncomment to allow organization bootstrap without prior creation.
# PARSEC_SPONTANEOUS_ORGANIZATION_BOOTSTRAP=true
```

vxgmichel (approved):
One comment, LGTM otherwise

docs/adminguide/hosting/parsec.env
mmmarcos: I agree with you. For the moment, I will merge this PR as is and will open another issue to document all variables in this file.

mmmarcos: Done #6198

PR #6219, opened 2024-01-23
[MS] Implement keyboard shortcut manager

### LINKED ISSUE #6184: Implement application shortcuts
Depends on #6183

### LINKED ISSUE #6288: Keyboard hotkeys manager

Ironicbay:
Closes #6288

### DIFF
--- a/client/src/services/hotkeyManager.ts
+++ b/client/src/services/hotkeyManager.ts
@@ -1,0 +1,114 @@
+// Parsec Cloud (https://parsec.cloud) Copyright (c) BUSL-1.1 2016-present Scille SAS
+
+import { isDesktop, isLinux, isMacOS, isMobile, isWeb, isWindows } from '@/parsec/environment';
+
+export const HotkeyManagerKey = 'hotkey';
+
+export enum Modifiers {
+  None = 0,
+  Ctrl = 1 << 0,
+  Alt = 1 << 1,
+  Shift = 1 << 2,
+}
+
+export enum Platforms {
+  Windows = 1 << 0,
+  MacOS = 1 << 1,
+  Linux = 1 << 2,
+  Web = 1 << 3,
+  Mobile = 1 << 4,
+  Desktop = Windows | MacOS | Linux,
+  All = Desktop | Web | Mobile,
+}
+
+export interface Hotkey {
+  key: string;
+  modifiers: number;
+  platforms: number;
+  command: () => Promise<void>;
+}
+
+class Hotkeys {
+  currentHotkeys: Hotkey[];
+  id: number;
+  enabled: boolean;
+
+  constructor(id: number) {
+    this.currentHotkeys = [];
+    this.id = id;
+    this.enabled = true;
+  }
+
+  add(key: string, modifiers: number, platforms: number, command: () => Promise<void>): void {
+    this.currentHotkeys.push({ key: key, modifiers: modifiers, platforms: platforms, command: command });
+  }
+
+  enable(): void {
+    this.enabled = true;
+  }
+
+  disable(): void {
+    this.enabled = false;
+  }
+}
+
+export class HotkeyManager {
+  currentHotkeys: Hotkeys[];
+  index: number;
+
+  constructor() {
+    this.currentHotkeys = [];
+    this.index = 0;
+    window.addEventListener('keydown', async (event: KeyboardEvent): Promise<void> => await this.onKeyPress(event, this.currentHotkeys));
+  }
+
+  newHotkeys(): Hotkeys {
+    const newKeys = new Hotkeys(this.index);
+    this.index++;
+    this.currentHotkeys.push(newKeys);
+    return newKeys;
+  }
+
+  unregister(toRemove: Hotkeys): void {
+    this.currentHotkeys = this.currentHotkeys.filter((hotkeys) => hotkeys.id !== toRemove.id);
+  }
+
+  private async onKeyPress(event: KeyboardEvent, hotKeys: Array<Hotkeys>): Promise<void> {
+    if (!isDesktop() && !isWeb()) {
+      return;
+    }
+    for (const keyPresses of hotKeys) {
+      for (const keyPress of keyPresses.currentHotkeys) {
+        if (!this.doPlatformsMatch(keyPress.platforms)) {
+          continue;
+        }
+        if (event.key.toLowerCase() === keyPress.key) {
+          if (!this.doModifiersMatch(event, keyPress.modifiers)) {
+            continue;
+          }
+          event.preventDefault();
+          if (keyPresses.enabled) {
+            await keyPress.command();
+          }
+        }
+      }
+    }
+  }
+
+  private doModifiersMatch(event: KeyboardEvent, modifiers: number): boolean {
+    const ctrlKey = isMacOS() ? event.metaKey : isDesktop() ? event.ctrlKey : event.ctrlKey || event.metaKey;
+    let eventMods = 0;
+    eventMods |= ctrlKey ? Modifiers.Ctrl : Modifiers.None;
+    eventMods |= event.altKey ? Modifiers.Alt : Modifiers.None;
+    eventMods |= event.shiftKey ? Modifiers.Shift : Modifiers.None;
+    return eventMods === modifiers;
+  }
+
+  private doPlatformsMatch(platforms: number): boolean {
+    return ((platforms & Platforms.Windows && isWindows()) ||
+      (platforms & Platforms.Linux && isLinux()) ||
+      (platforms & Platforms.MacOS && isMacOS()) ||
+      (platforms & Platforms.Web && isWeb()) ||
+      (platforms & Platforms.Mobile && isMobile())) as boolean;
+  }
+}
--- a/client/src/views/home/HomePage.vue
+++ b/client/src/views/home/HomePage.vue
@@ -211,6 +211,7 @@ async function openSettingsModal(): Promise<void> {
   });
   await modal.present();
   await modal.onWillDismiss();
+  await modal.dismiss();
 }
 
 async function onJoinOrganizationClicked(): Promise<void> {
--- a/client/src/main.ts
+++ b/client/src/main.ts
@@ -31,6 +31,7 @@ import { Validity, claimLinkValidator, fileLinkValidator } from '@/common/valida
 import { Answer, askQuestion } from '@/components/core';
 import { isElectron } from '@/parsec';
 import { Platform, libparsec } from '@/plugins/libparsec';
+import { HotkeyManager, HotkeyManagerKey } from '@/services/hotkeyManager';
 import { ImportManager, ImportManagerKey } from '@/services/importManager';
 import { Notification, NotificationKey, NotificationLevel, NotificationManager } from '@/services/notificationManager';
 import { initTranslations } from '@/services/translation';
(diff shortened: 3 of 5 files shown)


client/src/parsec/keyboard_events.ts
```diff
@@ -0,0 +1,17 @@
+// Parsec Cloud (https://parsec.cloud) Copyright (c) BUSL-1.1 2016-present Scille SAS
```
Max-7, line 1: `parsec/` is more for things that are somewhat related to parsec (bindings, devices, login, files, ...). This should be moved to `src/services` for example.

client/src/views/home/HomePage.vue
```diff
@@ -97,10 +97,24 @@ const routeWatchCancel = watchRoute(async () => {
 
 onMounted(async () => {
   storedDeviceDataDict.value = await storageManager.retrieveDevicesData();
+  document.addEventListener('keydown', function(event) {
```
Max-7, line 100: The API is not super easy to understand and can be improved.

Few things to note:
- the component shouldn't do the `document.addEventListener`, I don't want to look up the API for that thing when I want to add a shortcut.
- removing the event listener is just as bad, I don't want to repeat the same code
- the `onKeyPress` parameter is hard to understand (`combinations: Array<[string, () => Promise<void>, boolean?]>`)
- can callback functions take parameters?
- if I understand correctly, the third boolean parameter is the shift key, how about ctrl or alt?

Max-7 (changes requested):

client/src/components/core/ms-input/MsPasswordInput.vue
```diff
@@ -38,9 +38,23 @@
...
 import { IonCol, IonGrid, IonIcon, IonInput, IonItem, IonRow, IonText } from '@ionic/vue';
 import { eye, eyeOff } from 'ionicons/icons';
-import { ref } from 'vue';
+import { inject, onMounted, onUnmounted, ref } from 'vue';
+
+const hotkeyManager: HotkeyManager = inject(HotkeyManagerKey)!;
+const hotkeys = hotkeyManager.newHotkeys();
+
+onMounted(async () => {
+  await hotkeys.add('/', Modifiers.Ctrl, Platforms.Desktop | Platforms.Web, async () => {
```
Max-7, line 50: We should be careful about what components can register hotkeys. For example, what does this do in the change password page? Three components toggle the visibility at the same time, which is not what we want.

Registering the shortcut on the parent component can be a pain. In this case, I think checking that the component is in focus before toggling the visibility should be enough.

client/src/components/core/ms-input/MsSearchInput.vue
```diff
@@ -37,6 +38,7 @@ defineProps<{
 }>();
 
 const searchRef = ref('');
+const inputRef = ref();
```
Max-7, line 41: Not used.

client/src/services/hotkeyManager.ts
```diff
@@ -0,0 +1,126 @@
...
+  Web = 1 << 2,
+  Mobile = 1 << 3,
+  Desktop = Windows | MacOS | Linux,
+  All = Desktop | Web | Mobile,
+}
+
+export interface Hotkey {
+  key: string;
+  modifiers: Modifiers;
+  platforms: Platforms;
```
Max-7, line 26: `platforms` should be a number. For example, `Linux | MacOS` would take the value `1 + 2` which is not a valid `Platforms` type.

client/src/services/hotkeyManager.ts
```diff
@@ -0,0 +1,126 @@
...
+export const HotkeyManagerKey = 'hotkey';
+
+export enum Modifiers {
+  Ctrl = 1 << 0,
+  Alt = 1 << 1,
+  Shift = 1 << 2,
+}
+
+export enum Platforms {
+  Windows = 0,
```
Max-7, line 14: Should start at `1 << 0` (or `1`). It's impossible to check if `Platforms.Windows` is set otherwise. If you want to add a 0 value, it should be something like `NoPlatforms`. I don't know whether it will be used though.

client/src/services/hotkeyManager.ts
```diff
@@ -0,0 +1,126 @@
...
+class Hotkeys {
+  currentHotkeys: Hotkey[];
+  id: number;
+
+  constructor(id: number) {
+    this.currentHotkeys = [];
+    this.id = id;
+  }
+
+  async add(key: string, modifiers: Modifiers, platforms: Platforms, command: () => Promise<void>): Promise<void> {
```
Max-7, line 39: `platforms` should be a `number`

client/src/services/hotkeyManager.ts
```diff
@@ -0,0 +1,126 @@
...
+        return !ctrlKey && !event.shiftKey && event.altKey;
+      case Modifiers.Ctrl + Modifiers.Shift:
+        return ctrlKey && event.shiftKey && !event.altKey;
+      default:
+        return false;
+    }
+  }
+
+  private async checkPlatforms(platforms: Platforms): Promise<boolean> {
+    switch (platforms) {
```
Max-7, line 105: It's a bitmask https://en.wikipedia.org/wiki/Mask_(computing)#:~:text=In%20computer%20science%2C%20a%20mask,in%20a%20single%20bitwise%20operation.

client/src/services/hotkeyManager.ts
```diff
@@ -0,0 +1,126 @@
...
+      }
+    }
+  }
+
+  private async checkModifiers(event: KeyboardEvent, modifiers?: number): Promise<boolean> {
+    if (!modifiers) {
+      return !event.shiftKey && !event.ctrlKey && !event.metaKey && !event.altKey;
+    }
+    const ctrlKey = isMacOS() ? event.metaKey : isDesktop() ? event.ctrlKey : event.ctrlKey || event.metaKey;
+    switch (modifiers) {
```
Max-7, line 90: Bitmask https://en.wikipedia.org/wiki/Mask_(computing)#:~:text=In%20computer%20science%2C%20a%20mask,in%20a%20single%20bitwise%20operation

client/src/views/home/HomePage.vue
```diff
@@ -84,35 +86,47 @@ enum HomePageState {
...
 });
 
-onUnmounted(() => {
+onUnmounted(async () => {
   routeWatchCancel();
 });
 
 async function openCreateOrganizationModal(): Promise<void> {
-  const modal = await modalController.create({
+  if (currentModal.value) {
```
Max-7, line 113: I don't think this is the right way of handling these cases. It adds way too much code with way too much states. I think a better way of doing it would be to disable/reenable the shortcuts. 

```typescript
async function openCreateOrganizationModal(): Promise<void> {
  shortcuts.disable();
  const modal = await modalController.create({
    component: CreateOrganizationModal,
    canDismiss: true,
    cssClass: 'create-organization-modal',
    backdropDismiss: false,
  });
  await modal.present();
  const { data, role } = await modal.onWillDismiss();
  await modal.dismiss();
  shotcuts.enable();
```
and just add some code in the HotkeyManager to check whether a shortcut group is enabled or not.

and if some shortcuts have to remain valid, we could always create multiple groups and only disable one of them:
```typescript
const modalShortcuts = hotkeyManager.newHotkeys();
const otherShortcuts = hotkeyManager.newHotkeys();

modalShortcuts.add(blabla, openSettings);
modalShortcuts.add(blabla, openJoinOrg);
otherShortcuts.add(blabla, doSomething);

modalShortcuts.disable();
// open modal
modalShortcuts.enable();
```

Max-7 (changes requested):

client/src/components/core/ms-input/MsSearchInput.vue
Ironicbay: It's used [there](https://github.com/Scille/parsec-cloud/pull/6219/files#diff-6b65ea0ed0a55a527b277ba701426036e55d958d41bc0de1d15d42cd3e29f02fR180)

client/src/services/hotkeyManager.ts
```diff
@@ -0,0 +1,117 @@
...
+    const ctrlKey = isMacOS() ? event.metaKey : isDesktop() ? event.ctrlKey : event.ctrlKey || event.metaKey;
+    return ((modifiers & Modifiers.Ctrl && !ctrlKey) ||
+      (modifiers & Modifiers.Shift && !event.shiftKey) ||
+      (modifiers & Modifiers.Alt && !event.altKey) ||
+      (ctrlKey && !(modifiers & Modifiers.Ctrl)) ||
+      (event.shiftKey && !(modifiers & Modifiers.Shift)) ||
+      (event.altKey && !(modifiers & Modifiers.Alt))) as boolean;
+  }
+
+  private checkPlatforms(platforms: number): boolean {
```
Max-7, line 109: `checkPlatforms` and `checkModifiers` are weird names for functions returning booleans. In fact, I thought they were doing the opposite of what the comments say (return true if we're on the right platforms / if it has the right modifiers).

client/src/services/hotkeyManager.ts
```diff
@@ -0,0 +1,117 @@
...
+  id: number;
+  enabled: boolean;
+
+  constructor(id: number) {
+    this.currentHotkeys = [];
+    this.id = id;
+    this.enabled = true;
+  }
+
+  async add(key: string, modifiers: number, platforms: number, command: () => Promise<void>): Promise<void> {
```
Max-7, line 42: Functions in this class do not need to be async.

client/src/services/hotkeyManager.ts
```diff
@@ -0,0 +1,117 @@
...
+  }
+
+  newHotkeys = (): Hotkeys => {
+    const newKeys = new Hotkeys(this.index);
+    this.index++;
+    this.currentHotkeys.push(newKeys);
+    return newKeys;
+  };
+
+  async unregister(toRemove: Hotkeys): Promise<void> {
```
Max-7, line 72: Does not need to be async.

client/src/services/hotkeyManager.ts
```diff
@@ -0,0 +1,117 @@
+// Parsec Cloud (https://parsec.cloud) Copyright (c) BUSL-1.1 2016-present Scille SAS
+
+import { isDesktop, isLinux, isMacOS, isMobile, isWeb, isWindows } from '@/parsec/environment';
+
+export const HotkeyManagerKey = 'hotkey';
+
+export enum Modifiers {
+  None = 1 << 0,
```
Max-7, line 8: `None` should be 0.

client/src/services/hotkeyManager.ts
```diff
@@ -0,0 +1,117 @@
...
+          }
+        }
+      }
+    }
+  }
+
+  private checkModifiers(event: KeyboardEvent, modifiers: number): boolean {
+    // Returns true if needed modifiers aren't pressed or if not needed modifiers are pressed
+    const ctrlKey = isMacOS() ? event.metaKey : isDesktop() ? event.ctrlKey : event.ctrlKey || event.metaKey;
+    return ((modifiers & Modifiers.Ctrl && !ctrlKey) ||
```
Max-7, line 101: I think
```typescript
let eventMods = 0;
eventMods |= ctrlKey ? Modifiers.Ctrl : Modifiers.None;
eventMods |= event.altKey ? Modifiers.Alt : Modifiers.None;
eventMods |= event.shiftKey ? Modifiers.Shift : Modifiers.None;

return eventMods === modifiers;
```
would be easier to read and maintain.

Max-7 (changes requested):

Max-7 (approved):

PR #6229, opened 2024-01-26
Update `ruff` to `0.1.14` and remove `black`

FirelightFlagboy:
Format python code using `ruff` instead of `black`


touilleMan:
In branch `dev/v3`, black has already been replaced by ruff: https://github.com/Scille/parsec-cloud/blob/41831b9ad8529b43d099bca82e720005a8fa8923/.pre-commit-config.yaml#L148-L158

(besides this PR wil clash heavily with server code from `dev/v3`)

PR #6298, opened 2024-01-31
Add user freeze feature

touilleMan:
- Implementation only in memory
- [Documentation is missing](https://github.com/Scille/parsec-cloud/pull/5842/files#diff-d4f43e12c074bcf616dead86bed45702a8b68331df85a4dc7804c7e352780833)
- No client support

### DIFF
--- a/server/parsec/asgi/administration.py
+++ b/server/parsec/asgi/administration.py
@@ -22,6 +22,7 @@ from parsec._parsec import (
     BootstrapToken,
     DateTime,
     OrganizationID,
+    UserID,
     UserProfile,
 )
 from parsec.components.organization import (
@@ -33,6 +34,7 @@ from parsec.components.organization import (
     OrganizationUpdateBadOutcome,
     Unset,
 )
+from parsec.components.user import UserFreezeUserBadOutcome, UserInfo, UserListUsersBadOutcome
 from parsec.events import OrganizationIDField
 
 if TYPE_CHECKING:
@@ -48,6 +50,18 @@ def check_administration_auth(
 ) -> None:
     if request.app.state.backend.config.administration_token != credentials.credentials:
         raise HTTPException(status_code=403, detail="Bad authorization token")
+
+
+# This function is a workaround for FastAPI's broken custom type in query parameters
+# (see https://github.com/tiangolo/fastapi/issues/10259)
+def parse_organization_id_or_die(raw_organization_id: str) -> OrganizationID:
+    try:
+        return OrganizationID(raw_organization_id)
+    except ValueError:
+        raise HTTPException(
+            status_code=404,
+            detail="Invalid organization ID",
+        )
 
 
 class CreateOrganizationIn(BaseModel):
@@ -122,13 +136,7 @@ async def administration_get_organization(
 ) -> GetOrganizationOut:
     backend: Backend = request.app.state.backend
 
-    try:
-        organization_id = OrganizationID(raw_organization_id)
-    except ValueError:
-        raise HTTPException(
-            status_code=404,
-            detail="Invalid organization ID",
-        )
+    organization_id = parse_organization_id_or_die(raw_organization_id)
 
     # Check whether the organization actually exists
     outcome = await backend.organization.get(id=organization_id)
@@ -186,13 +194,7 @@ async def administration_patch_organization(
 ) -> PatchOrganizationOut:
     backend: Backend = request.app.state.backend
 
-    try:
-        organization_id = OrganizationID(raw_organization_id)
-    except ValueError:
-        raise HTTPException(
-            status_code=404,
-            detail="Invalid organization ID",
-        )
+    organization_id = parse_organization_id_or_die(raw_organization_id)
 
     outcome = await backend.organization.update(
         id=organization_id,
@@ -219,13 +221,7 @@ async def administration_organization_stat(
 ) -> Response:
     backend: Backend = request.app.state.backend
 
-    try:
-        organization_id = OrganizationID(raw_organization_id)
-    except ValueError:
-        raise HTTPException(
-            status_code=404,
-            detail="Invalid organization ID",
-        )
+    organization_id = parse_organization_id_or_die(raw_organization_id)
 
     outcome = await backend.organization.organization_stats(organization_id)
     match outcome:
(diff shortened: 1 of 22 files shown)


server/parsec/events.py
```diff
@@ -338,9 +338,17 @@ class EventOrganizationExpired(BaseModel):
...
+class EventUserRevokedOrFrozen(BaseModel):
     model_config = ConfigDict(arbitrary_types_allowed=True, strict=True)
-    type: Literal["USER_REVOKED"] = "USER_REVOKED"
+    type: Literal["USER_REVOKED_OR_FROZEN"] = "USER_REVOKED_OR_FROZEN"
+    event_id: UUID = Field(default_factory=uuid4)
+    organization_id: OrganizationIDField
+    user_id: UserID
+
+
+class EventUserUnFrozen(BaseModel):
```
vxgmichel, line 349:
```suggestion
class EventUserUnfrozen(BaseModel):
```
Weird casing

server/parsec/components/memory/user.py
```diff
@@ -593,3 +597,74 @@ async def test_dump_current_users(
...
+                except KeyError:
+                    return UserFreezeUserBadOutcome.USER_NOT_FOUND
+            case (None, str() as user_email):
+                for user in org.users.values():
+                    if user.cooked.human_handle.email == user_email:
+                        break
+                else:
+                    return UserFreezeUserBadOutcome.USER_NOT_FOUND
+            case _:
+                return UserFreezeUserBadOutcome.BOTH_USER_ID_AND_EMAIL
```
vxgmichel, line 648: No exhaustive match here?

server/parsec/events.py
```diff
@@ -338,9 +338,17 @@ class EventOrganizationExpired(BaseModel):
     organization_id: OrganizationIDField
 
 
-class EventUserRevoked(BaseModel):
+class EventUserRevokedOrFrozen(BaseModel):
```
vxgmichel, line 341: Why merge both events?
Defining two separate events seems more convenient in the long run

server/parsec/components/auth.py
```diff
@@ -70,12 +73,24 @@ class AuthenticatedAuthInfo:
...
+    def _on_event(self, event: Event) -> None:
+        match event:
+            # Revocation and freezing/unfreezing affect the authentication process,
+            # so we clear the cache when such events occur.
+            case EventUserUnFrozen() | EventUserRevokedOrFrozen():
+                self._device_cache = {
+                    (org_id, device_id): v
+                    for ((org_id, device_id), v) in self._device_cache.items()
+                    if org_id != event.organization_id or device_id.user_id != event.user_id
+                }
```
vxgmichel, line 93: Why not simply remove the corresponding entry?
```suggestion
                self._device_cache.pop((event.organization_id, event.user_id), None)
```

vxgmichel (approved):

server/parsec/components/memory/user.py
touilleMan: Exhaustive matching on tuple is not supported (see https://github.com/python/mypy/issues/16650)

server/parsec/events.py
touilleMan: This is to stay KISS: this event is only needed to indicate a user should no longer be allowed.

It is unlikely we need to differentiate further, and even if that's the case it's better to add a field to this event instead of splitting it into two the events (as this would open the door of bugs where user freeze is overlooked and only user revoked event is handled)

server/parsec/components/auth.py
touilleMan: `self._device_cache` contains keys of `tuple[OrganizationID, DeviceID]`, but the event provide `OrganizationID` and `UserID`

We filter the devices in the cache on the `UserID` then

server/parsec/asgi/administration.py
```diff
@@ -357,3 +359,111 @@ async def administration_server_stats(
...
+
+    return JSONResponse(
+        status_code=200,
+        content={
+            "users": [
+                {
+                    "user_id": user.user_id.str,
+                    "user_email": user.human_handle.email,
+                    "user_name": user.human_handle.label,
+                    "frozen": user.frozen,
```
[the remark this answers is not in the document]
touilleMan: KISS: serializer schema is more complicated than just `JSONResponse` (typically custom types such as `UserID` has to be cooked with the field validator system which is cumbersome)

We might want to switch to the serialization schema in the future if we use [fastapi documentation generation](https://fastapi.tiangolo.com/features/#automatic-docs) (btw documention generation is currently broken due to...  custom types in `server/parsec/events.py`'s serializers ¯|_(ツ)_/¯)

server/parsec/asgi/administration.py
```diff
@@ -357,3 +359,111 @@ async def administration_server_stats(
...
+    request: Request,
+) -> Response:
+    backend: Backend = request.app.state.backend
+
+    try:
+        organization_id = OrganizationID(raw_organization_id)
+    except ValueError:
+        raise HTTPException(
+            status_code=404,
+            detail="Invalid organization ID",
```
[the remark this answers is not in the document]
touilleMan: What do you think would be better ?

Should we keep those readable messages (and update RFC accordingly) or switch to the error_code ?

My guess is the status code is enough (the only issue I see is 404 is for both organization or user not found), and switching to error_code requires to change output in other administration routes to stay consistent.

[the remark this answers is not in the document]
vxgmichel: I would just update the RFC, it has a dedicated section for v3 errors

server/parsec/events.py
vxgmichel: Don't let cspell get to you, we're the ones in charge

server/parsec/components/memory/user.py
vxgmichel: That's sad. Maybe you can still write it with:
```python
case _: 
    assert False  # Can't use assert_never here due to https://github.com/python/mypy/issues/16650
```

server/parsec/events.py
vxgmichel: I hear your point about the bug but it still feels contrived to me. IMO different events should be different classes, which seems pretty simple if KISS is your concern. I won't fight over it though

server/parsec/components/auth.py
vxgmichel: Erh sorry I read too fast

server/parsec/asgi/administration.py
```diff
@@ -357,3 +359,111 @@ async def administration_server_stats(
...
+
+class UserFreezeIn(BaseModel):
+    model_config = ConfigDict(arbitrary_types_allowed=True, strict=True)
+    frozen: bool
+    user_email: str | None = None
+    user_id: UserID | None = None
+
+    @field_validator("user_id", mode="plain")
+    @classmethod
+    def validate_active_users_limit(cls, v: Any) -> UserID | None:
```
mmmarcos, line 413: The name of this validator does not seem right, how it is validating _active user limit_?

server/parsec/asgi/administration.py
```diff
@@ -357,3 +359,111 @@ async def administration_server_stats(
...
+        match v:
+            case UserID():
+                return v
+            case None:
+                return None
+            case raw:
+                return UserID(raw)
```
mmmarcos, line 420: BTW [according to the docs](https://docs.pydantic.dev/latest/concepts/validators/#field-validators):
> _validators should either return the parsed value or raise a ValueError or AssertionError (assert statements may be used)._
Isn't there an error case here?

server/parsec/asgi/administration.py
```diff
@@ -357,3 +359,111 @@ async def administration_server_stats(
...
+    request: Request,
+) -> Response:
+    backend: Backend = request.app.state.backend
+
+    try:
+        organization_id = OrganizationID(raw_organization_id)
+    except ValueError:
+        raise HTTPException(
+            status_code=404,
+            detail="Invalid organization ID",
```
mmmarcos, line 437: This seems a rather common code construct (used 5 times in this PR), maybe it should be moved to a function?
Or maybe adding some reusable decorator logic?

server/parsec/events.py
mmmarcos: Just for my general culture: are these events internal to the server? or they are meant to be consumed by some other component?

mmmarcos (commented):
LGTM overall. Some minor comments.

* Are you planning to add the docs from #5842 to this PR?
* By "No client support" you mean the CLI?

server/parsec/events.py
touilleMan: > are these events internal to the server? 

Those events are internal to parsec server, so we can change them whenever we want with no backward compatibility whatsoever ;-)


> IMO different events should be different classes

I agree in theory, but here I see it as a very high risk of forgetting the user freeze given how niche this feature is

BTW this is not that different than the events we send when a new certificate is created: the event structure is defined according to what it is going to be consumed for, not what it comes from (hence we don't have one event per action creating a certificate, but an event if a common certif has been added, another one for realm certifs and another one for shamir)

This approach works fine here given our events are basically always single purposed (it wouldn't be the right approach if we had a lot of frequently changing logic reacting to various combination of events)

server/parsec/asgi/administration.py
```diff
@@ -357,3 +359,111 @@ async def administration_server_stats(
...
+        )
+
+    outcome = await backend.user.list_users(organization_id)
+    match outcome:
+        case list() as users:
+            pass
+        case UserListUsersBadOutcome.ORGANIZATION_NOT_FOUND:
+            raise HTTPException(status_code=404, detail="Organization not found")
+        case unknown:
+            assert_never(unknown)
```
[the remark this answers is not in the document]
touilleMan: "Outcome" comes from the [outcome library](https://github.com/python-trio/outcome), so an outcome wraps the result of a function.

> Usually this things are named result or response

I think `response` is to be used when doing http request.

`result` would have been possible name, but it lacks the fact both ok and error are contained. Given in Python you expect a function to return the ok part and raise an exception for the errors, a dedicated name like outcome is good for readability.

> Also suffix BadOutcome for enums which has the downside of using 2 words instead something like just Error.

This is a feature ! ^^
You have `FooError` which is an exception and `FooBadOutcome` which is an enum of possible errors.

server/tests/administration/test_freeze_user.py
```diff
@@ -0,0 +1,174 @@
...
+
+        with pytest.raises(StopAsyncIteration):
+            # Loop given the server might have send us some events before the freeze
+            while True:
+                await alice_sse.next_event()
+
+        # ...and we cannot reconnect !
+
+    rep = await coolorg.alice.raw_sse_connection()
+    assert rep.status_code == 462
```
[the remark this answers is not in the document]
touilleMan: The test has to make the status code is 462, if we `CustomHttpStatus.UserFrozen.value` we wouldn't detect if `CustomHttpStatus.UserFrozen.value` is the broken part !

server/parsec/asgi/administration.py
touilleMan: `UserID(raw)` raises `ValueError` if `raw` is invalid

server/parsec/asgi/rpc.py
```diff
@@ -226,6 +228,8 @@ def _handshake_abort(status_code: int, api_version: ApiVersion) -> NoReturn:
             detail = "Organization expired"
         case 461:
             detail = "User revoked"
+        case 462:
```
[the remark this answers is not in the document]
touilleMan: Great idea, done :+1:

server/parsec/asgi/administration.py
touilleMan: good catch ! done.

server/parsec/asgi/administration.py
touilleMan: I agree, in theory we should be able to have `organization_id: OrganizationID` in the function signature and let fastapi do the conversion, [but this is broken](https://github.com/tiangolo/fastapi/issues/10259) :'( 

Anyway I've factorized this code, it's a bit better this way :+1:

server/parsec/asgi/administration.py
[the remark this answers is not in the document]
touilleMan: I've updated the RFC to specify the error body is only for human consumption and hence has no guaranteed format

PR #6300, opened 2024-02-01
[MS] Add setFocus() expose to input components

### LINKED ISSUE #6299: Add `setFocus()` expose to input components
- MsInput
- MsPasswordInput
- MsSearchInput

Also use it on the main page to set the focus to the `MsPasswordInput`, and remove the `autofocus` prop.

Ironicbay:
Closes #6299


Max-7:
Didn't manager to find a better way than to set a timeout. I think it happens specifically when animations occur, we can't set the focus immediately. It's somewhat wobbly, but worst case scenario is the focus is not set.

fabienSvtr (approved):

PR #6265, opened 2024-01-29
Fix build wheel in `parsec-server`

mmmarcos:
Update `Cargo.lock` with: `cargo update --offline`

Issue #5971, opened 2023-12-11
[server-wheel] Fail because of outdated `Cargo.lock`


FirelightFlagboy:
Currently the package jobs fail with the following error when generating the wheel with `cibuildwheel`:

```
error: the lock file /Users/runner/work/parsec-cloud/parsec-cloud/Cargo.lock needs to be updated but --locked was passed to prevent this
If you want to try to generate the lock file without accessing the network, remove the --locked flag and use --offline instead.
```

FirelightFlagboy:
Tested on edff01a5e95a90128262902e8b050b3325682518

`Cargo.lock` isn't update (`cargo update --locked` fail)

When I try to update it with `cargo update --offline`, I get:

```
error: failed to select a version for `once_cell`.
    ... required by package `openssl v0.10.61`
    ... which satisfies dependency `openssl = "^0.10.61"` of package `libparsec_crypto v2.16.0-a.0+dev (/[..]/parsec-cloud/libparsec/crates/crypto)`
    ... which satisfies path dependency `libparsec_crypto` of package `parsec v2.16.0-a.0+dev (/[..]/parsec-cloud/server)`
versions that meet the requirements `^1.5.2` are: 1.18.0, 1.17.1, 1.17.0

all possible versions conflict with previously selected packages.

  previously selected package `once_cell v1.19.0`
    ... which satisfies dependency `once_cell = "^1.19.0"` of package `libparsec_platform_mountpoint v2.16.0-a.0+dev (/[..]/parsec-cloud/libparsec/crates/platform_mountpoint)`

failed to select a version for `once_cell` which could resolve this conflict
```

PR #6340, opened 2024-02-01
Update server build system deps

touilleMan:


touilleMan:
Not sure why the CI fails... 

The error is `    error: the lock file /project/Cargo.lock needs to be updated but --locked was passed to prevent this`, however it runs fine on my machine (and changing maturin version shouldn't impact the Rust dependencies lock...)

@FirelightFlagboy is it possible it's a cache issue ?

mmmarcos:
> Not sure why the CI fails...
> 
> The error is ` error: the lock file /project/Cargo.lock needs to be updated but --locked was passed to prevent this`, however it runs fine on my machine (and changing maturin version shouldn't impact the Rust dependencies lock...)
> 
> @FirelightFlagboy is it possible it's a cache issue ?

We've discussed it yesterday. We are unable to reproduce it locally: `cibuildwheel` works fine locally, even without updating `Cargo.lock` at all.

I opened #6265 to try to fix it. I bumped `cibuildwheel` to `v2.16.4` (which is the version currently pinned in `.github/workflows/package-server.yml`). Still no luck 😕

FirelightFlagboy:
@touilleMan you can take a look at the issue #5971

FirelightFlagboy (approved):

PR #6349, opened 2024-02-02
[MS] Distinct information and notification managers

### LINKED ISSUE #6235: Update  NotifManager by InformationManager
Replace the `title` by a `message`

NicoTuxx:
closes #6235

### DIFF
--- a/client/src/services/informationManager.ts
+++ b/client/src/services/informationManager.ts
@@ -1,0 +1,128 @@
+// Parsec Cloud (https://parsec.cloud) Copyright (c) BUSL-1.1 2016-present Scille SAS
+
+import { MsAlertModal, MsAlertModalConfig, MsReportTheme } from '@/components/core';
+import { NotificationManager } from '@/services/notificationManager';
+import { ToastManager } from '@/services/toastManager';
+import { modalController } from '@ionic/vue';
+import { v4 as uuid4 } from 'uuid';
+
+export const InformationKey = 'information';
+
+export enum InformationLevel {
+  Info = 'INFO',
+  Success = 'SUCCESS',
+  Warning = 'WARNING',
+  Error = 'ERROR',
+}
+
+export interface InformationOptions {
+  message: string;
+  level: InformationLevel;
+  data?: object;
+}
+
+export class Information {
+  id: string;
+  message: string;
+  level: InformationLevel;
+  data?: object;
+  constructor({ message, level, data = {} }: InformationOptions) {
+    this.id = uuid4();
+    this.message = message;
+    this.level = level;
+    this.data = data;
+  }
+
+  get theme(): MsReportTheme {
+    switch (this.level) {
+      case InformationLevel.Info:
+        return MsReportTheme.Info;
+      case InformationLevel.Success:
+        return MsReportTheme.Success;
+      case InformationLevel.Warning:
+        return MsReportTheme.Warning;
+      case InformationLevel.Error:
+        return MsReportTheme.Error;
+    }
+  }
+}
+
+export enum PresentationMode {
+  Console = 1 << 0,
+  Notification = 1 << 1,
+  Toast = 1 << 2,
+  Modal = 1 << 3,
+}
+
+export class InformationManager {
+  toastManager: ToastManager;
+  notificationManager: NotificationManager;
+
+  constructor() {
+    this.toastManager = new ToastManager();
+    this.notificationManager = new NotificationManager();
+  }
+
+  private async showModal(information: Information): Promise<void> {
+    const alertModalConfig: MsAlertModalConfig = {
+      theme: information.theme,
+      message: information.message,
+    };
+
+    await this._createAndPresentModal(alertModalConfig);
+  }
+
+  private async _createAndPresentModal(modalConfig: MsAlertModalConfig): Promise<any> {
+    const top = await modalController.getTop();
+    if (top) {
+      top.classList.add('overlapped-modal');
+    }
+    const modal = await modalController.create({
+      component: MsAlertModal,
+      canDismiss: true,
+      cssClass: 'information-modal',
+      componentProps: modalConfig,
+    });
+    await modal.present();
+    const result = await modal.onWillDismiss();
+    await modal.dismiss();
+
+    if (top) {
+      top.classList.remove('overlapped-modal');
+    }
+    return result;
+  }
+
+  private async showToast(information: Information): Promise<void> {
+    this.toastManager.createAndPresent({
+      theme: information.theme,
+      message: information.message,
+    });
+  }
+
+  private sendNotification(information: Information): void {
+    this.notificationManager.add(information);
+  }
+
+  async present(information: Information, presentationMode: PresentationMode): Promise<void> {
+    if (presentationMode & PresentationMode.Modal && presentationMode & PresentationMode.Toast) {
+      console.error('Error: forbidden usage, toasts and modals are not meant to be used at the same time.');
+    }
+    if (presentationMode & PresentationMode.Console) {
+      this._trace(information);
+    }
+    if (presentationMode & PresentationMode.Notification) {
+      this.sendNotification(information);
+    }
+    if (presentationMode & PresentationMode.Toast) {
+      await this.showToast(information);
+    }
+    if (presentationMode & PresentationMode.Modal) {
+      await this.showModal(information);
+    }
+  }
+
+  private _trace(information: Information): void {
+    console.trace(`[${information.level}] ${information.message}`);
+  }
+}
--- a/client/src/services/notificationManager.ts
+++ b/client/src/services/notificationManager.ts
@@ -1,37 +1,19 @@
 // Parsec Cloud (https://parsec.cloud) Copyright (c) BUSL-1.1 2016-present Scille SAS
 
-import { MsAlertModal, MsAlertModalConfig, MsModalResult, MsReportTheme } from '@/components/core';
-import { ToastManager } from '@/services/toastManager';
-import { modalController } from '@ionic/vue';
+import { Information } from '@/services/informationManager';
 import { DateTime } from 'luxon';
-import { v4 as uuid4 } from 'uuid';
 import { Ref, computed, ref } from 'vue';
 
 export const NotificationKey = 'notification';
 
-export enum NotificationLevel {
-  Info = 'INFO',
-  Success = 'SUCCESS',
-  Warning = 'WARNING',
-  Error = 'ERROR',
-}
-
 export class Notification {
-  id: string;
+  information: Information;
   time: DateTime;
   read: boolean;
-  title: string;
-  message: string;
-  level: NotificationLevel;
-  data?: object;
-  constructor({ title, level, data = {}, message }: { title: string; message: string; level: NotificationLevel; data?: object }) {
-    this.id = uuid4();
+  constructor(information: Information) {
+    this.information = information;
     this.time = DateTime.now();
     this.read = false;
-    this.message = message;
-    this.level = level;
-    this.data = data;
-    this.title = title;
   }
 }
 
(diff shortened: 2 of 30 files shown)


client/src/services/notificationManager.ts
```diff
@@ -40,105 +22,26 @@ export interface NotificationOptions {
...
-      case NotificationLevel.Info:
-        return MsReportTheme.Info;
-      case NotificationLevel.Success:
-        return MsReportTheme.Success;
-      case NotificationLevel.Warning:
-        return MsReportTheme.Warning;
-      case NotificationLevel.Error:
-        return MsReportTheme.Error;
-    }
+  create(information: Information): Notification {
```
Max-7, line 33: Is there a point in returning the object? Why not just call this method `addNotification()` or something like that and avoid exposing an object that is only used internally?

client/src/services/informationManager.ts
```diff
@@ -0,0 +1,155 @@
...
+export const InformationKey = 'information';
+
+export enum InformationLevel {
+  Info = 'INFO',
+  Success = 'SUCCESS',
+  Warning = 'WARNING',
+  Error = 'ERROR',
+}
+
+export interface IInformation {
```
Max-7, line 18: What's the point of this interface?

client/src/services/informationManager.ts
```diff
@@ -0,0 +1,155 @@
...
+        duration: options?.duration,
+      })
+      .then(async () => {});
+  }
+
+  sendNotification(information: Information): void {
+    this.notificationManager.create(information);
+  }
+
+  async present(
```
Max-7, line 120: When is this method used?

client/src/services/informationManager.ts
```diff
@@ -0,0 +1,155 @@
...
+  }
+
+  async showToast(information: Information, options?: InformationToastOptions): Promise<void> {
+    this.toastManager
+      .createAndPresent({
+        theme: information.theme,
+        message: information.message,
+        duration: options?.duration,
+      })
+      .then(async () => {});
```
Max-7, line 113: Is there a point to this `then`?

client/src/services/informationManager.ts
```diff
@@ -0,0 +1,155 @@
...
+
+export interface InformationOptions {
+  trace?: boolean;
+}
+
+export enum InformationPresentationMode {
+  Toast,
+  Modal,
+  Notification,
+  ToastAndNotification,
```
Max-7, line 58: Bitmasks are perfect to avoid this.

client/src/services/informationManager.ts
```diff
@@ -0,0 +1,155 @@
...
+    presentationMode: InformationPresentationMode = InformationPresentationMode.Toast,
+    options: InformationOptions,
+  ): Promise<void> {
+    if (options && options.trace) {
+      this._trace(information);
+    }
+    switch(presentationMode) {
+      case InformationPresentationMode.Toast:
+        return await this.showToast(information, options);
+        break;
```
Max-7, line 131: Return makes the `break` redundant, I'm surprised the linter doesn't cry.

client/src/services/informationManager.ts
```diff
@@ -0,0 +1,155 @@
...
+
+export enum InformationPresentationMode {
+  Toast,
+  Modal,
+  Notification,
+  ToastAndNotification,
+  ModalAndNotification,
+}
+
+export interface InformationToastOptions extends InformationOptions {
```
Max-7, line 62: Not sure about this interface. It is used in `showToast()`, but in `present()` you display a toast using `InformationOptions` only.

Max-7 (changes requested):

client/src/services/notificationManager.ts
NicoTuxx: At the moment, I'm not sure we need to return it yh.
However, the `create` name is the more logical looking at what is doing the function. It takes an Information and create a Notification.
Instead, `addNotification` is quite disturbing on multiple points:
 - it creates it, not only 'add' it
 - repeating notification on a notificationManager seems useless: `notificationManager.addNotification()` we already know it's a notification thanks to the manager name.
 Maybe only `add` could be a compromise even if I don't see the point, since `create` is perfect match here?

client/src/services/informationManager.ts
NicoTuxx: Previously I was exporting it instead of the class, not anymore, however it's more readable to have the interface instead of the inline interface declaration in the already consistent class declaration.

client/src/services/informationManager.ts
NicoTuxx: As explained on slack, it will be the main use for real use cases not implemented atm.

client/src/services/informationManager.ts
NicoTuxx: There was a markAsRead executed in here before, since we don't care about this for toast anymore, it's safe to say that we can delete this part.

client/src/services/informationManager.ts
NicoTuxx: I see the point after our last discussion about bitmasking, but I don't see anything we would want to "avoid" here.
It's clear, simple and control which cases we want to accept.
Can you detail more what's the matter here?

client/src/services/informationManager.ts
NicoTuxx: copy/paste my bad 🥷 
good catch

client/src/services/informationManager.ts
NicoTuxx: normally since it extends the `InformationOptions` it should work anyway, or maybe I need to update to:
`    options: InformationOptions | InformationToastOptions,`

client/src/services/notificationManager.ts
Max-7: `create` to me takes a more factory connotation: I'm creating an object and returning it.

Here we indeed create it, but we're also adding it to a list, and returning it seems pointless. But I'm not going to die on this hill, it doesn't matter much.

client/src/services/informationManager.ts
Max-7: I meant that the goal of an interface is to abstract something we can have multiple of.
```typescript
interface Vehicle { wheels: number; }

class Car extends Vehicle { this.wheels = 4; }
class Bicycle extends Vehicle { this.wheels = 2; }

function doSomething(v: Vehicle) {}
```
Here it seems that only the `Information` class uses it, it's never used anywhere else.

client/src/services/notificationManager.ts
NicoTuxx: That's okay to me, `add` seems what we want if we as you said don't return it yh.

client/src/services/informationManager.ts
Max-7: Bitmask would make defining `ToastAndNotification` useless, and would facilitate the `switch` that checks this value.

```typescript
enum PresentationMode {
  // Using bit shifting because it makes it more confusing
  // and you look more like you know what you're doing, but
  // 1, 2 and 4 are acceptable values.
  Toast = 1 << 0,
  Modal = 1 << 1,
  Notification = 1 << 2,
}

function show(mode: number) {
  if (mode & PresentationMode.Toast) {
    showToast();
  }
  if (mode & PresentationMode.Modal) {
    showModal();
  }
  if (mode & PresentationMode.Notification) {
    addToNotification();
  }
}

show(PresentationMode.Toast | PresentationMode.Notification);
```
It makes it way easier to add or remove modes. For example, we currently have a trace option to print in the console, but it could become a mode too:

```typescript
enum PresentationMode {
  ...
  Console = 1 << 3,
}

function show(mode: number) {
  ...
  if (mode & PresentationMode.Console) {
    console.log(...);
  }
}

show(PresentationMode.Toast | PresentationMode.Modal | PresentationMode.Console);
```

client/src/services/informationManager.ts
Max-7: It shouldn't in a normal object language because `InformationOptions` does not have a `duration` attribute. Updating it is probably better for now, we'll check later if we find something better.

client/src/services/informationManager.ts
NicoTuxx: interface is not only meant to abstract things, in TS its first purpose is to type data-structured types, here we were already using an interface without making it explicit:
```
export class Information {
  id: string;
  message: string;
  level: InformationLevel;
  data?: object;
  constructor({ message, level, data = {} }: {
  message: string;
  level: InformationLevel;
  data?: object
}) {
    this.id = uuid4();
    this.message = message;
    this.level = level;
    this.data = data;
  }
}
```
the part after the `:` in the constructor is an interface even without declaring it, I just found more readable and simple to maintain with the interface declared separetely.

Max-7: But it adds complexity needlessly. `IInformation` is not used anywhere else, and just defines a few attributes that are redeclared anyway on `Information`.

In your example, I don't see the difference with
```typescript
export class Information {
  id: string;
  message: string;
  level: InformationLevel;
  data?: object;

  constructor(message: string, level: InformationLevel, data = {}) {
    this.id = uuid4();
    this.message = message;
    this.level = level;
    this.data = data;
  }
}
```
but this seems easier to read.

client/src/services/informationManager.ts
NicoTuxx: Hmm yh seems good to me, will replace.

client/src/services/informationManager.ts
NicoTuxx: We agreed on this format, it's easier to use and offer the possibility to create Notifications without caring about parameters order.

client/src/services/informationManager.ts
```diff
@@ -0,0 +1,137 @@
...
+      console.error('Error: forbidden usage, toasts and modals are not meant to be used at the same time.');
+    }
+    if (presentationMode & InformationPresentationMode.Console) {
+      this._trace(information);
+    }
+    if (presentationMode & InformationPresentationMode.Notification) {
+      this.sendNotification(information);
+    }
+    if (presentationMode & InformationPresentationMode.Toast) {
+      return await this.showToast(information);
```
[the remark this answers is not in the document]
NicoTuxx: yup

client/src/services/informationManager.ts
```diff
@@ -0,0 +1,137 @@
...
+    }
+    if (presentationMode & InformationPresentationMode.Modal) {
+      return await this.showModal(information);
```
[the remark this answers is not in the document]
NicoTuxx: yup

client/src/services/informationManager.ts
```diff
@@ -0,0 +1,137 @@
...
+    }
+    return new Promise(() => {});
```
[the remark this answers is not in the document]
NicoTuxx: yup

client/src/services/informationManager.ts
```diff
@@ -0,0 +1,132 @@
...
+      theme: information.theme,
+      message: information.message,
+    });
+  }
+
+  private sendNotification(information: Information): void {
+    this.notificationManager.add(information);
+  }
+
+  async present(information: Information, presentationMode: InformationPresentationMode): Promise<void> {
```
Max-7, line 108: `presentationMode` should by typed as a `number` (it's a combination of values from InformationPresentationMode)

client/src/services/informationManager.ts
```diff
@@ -0,0 +1,132 @@
...
+        return MsReportTheme.Success;
+      case InformationLevel.Warning:
+        return MsReportTheme.Warning;
+      case InformationLevel.Error:
+        return MsReportTheme.Error;
+    }
+  }
+}
+
+export enum InformationPresentationMode {
```
Max-7, line 50: Ins't the name a bit long? Since we're importing it from `informationManager`, we could probably shorten it to just `PresentationMode`.

.cspell/custom-words.txt
```diff
@@ -98,6 +98,7 @@ corofn
 Couldnt
 crtime
 cryptolockers
+cspellisannoying
```
Max-7, line 101: Please remove (even though I understand).

Max-7 (approved):

PR #6377, opened 2024-02-07
Update docs/Dockerfile to support Python 3.12

mmmarcos:
Dockerfile was broken since the upgrade to Python 3.12.

Also, update README.md to add more details about how to build the docs with Docker.

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -1,26 +1,81 @@
 <!-- Parsec Cloud (https://parsec.cloud) Copyright (c) BUSL-1.1 2016-present Scille SAS -->
 
-# How to generate the documentation website
+# Parsec documentation
+
+This directory contains the sources for Parsec docs.
+
+## Introduction
+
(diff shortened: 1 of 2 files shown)


docs/README.md
```diff
@@ -1,26 +1,81 @@
...
-## Update translation
+        ```bash
+        docker run --rm -it -v $(pwd):/data parsec-docs
+        ```
 
-see <http://www.sphinx-doc.org/en/master/usage/advanced/intl.html>
+        > **Note**
+        > The current working directory (which must be the local path to parsec-cloud repo)
+        > is mounted in the "/data" in the container. This means that builds started from
+        > the container will be available in your local repo under "_build" directory.
```
FirelightFlagboy, line 43:
```suggestion
        > The current working directory (which must be the local path to parsec-cloud repo)
        > is mounted in the "/data" in the container. This means that builds started from
        > the container will be available in your local repo under `_build` directory.
```

FirelightFlagboy (approved):

PR #6378, opened 2024-02-08
Run post update hook to fix Cargo lock when updating parsec version

### LINKED ISSUE #5971: [server-wheel] Fail because of outdated `Cargo.lock`

FirelightFlagboy:
When updating the parsec version, the Cargo.lock file needs to be updated.
This is done by running `cargo generate-lockfile --direct-minimal-versions` (nightly feature for now).

> `cargo generate-lockfile` is the simplest way I've found to update the lockfile without having the compile or check the entire project.

Fix #5971

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -125,7 +125,7 @@ jobs:
   quality-assurance:
     name: 📊 Q&A
     # All linux jobs must run the same ubuntu version to avoid Rust caching issues !
-    runs-on: ubuntu-20.04
+    runs-on: ubuntu-22.04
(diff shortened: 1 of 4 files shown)


FirelightFlagboy:
https://github.com/Scille/parsec-cloud/actions/runs/7826984428

PR #6421, opened 2024-02-14
Add tests: `multiple` for `add_certificates_batch`

TimeEngineer:


libparsec/crates/client/tests/unit/certif/add_device_certificate.rs
```diff
@@ -27,6 +27,27 @@ async fn ok(env: &TestbedEnv) {
...
+            &env.get_common_certificates_signed(),
+            &[],
+            &[],
+            &Default::default(),
+        )
+        .await
+        .unwrap();
+
+    p_assert_matches!(switch, MaybeRedactedSwitch::NoSwitch);
+}
```
FirelightFlagboy, line 49: I don't see where this test uses multiple certs, how multiple certs are at play here ?

libparsec/crates/client/tests/unit/certif/add_user_certificate.rs
```diff
@@ -27,6 +27,27 @@ async fn ok(env: &TestbedEnv) {
...
+            &env.get_common_certificates_signed(),
+            &[],
+            &[],
+            &Default::default(),
+        )
+        .await
+        .unwrap();
+
+    p_assert_matches!(switch, MaybeRedactedSwitch::NoSwitch)
+}
```
FirelightFlagboy, line 49: How there is multiple certs at play here ?

libparsec/crates/client/tests/unit/certif/add_device_certificate.rs
TimeEngineer: Because there is already `alice`'s device in `env`

libparsec/crates/client/tests/unit/certif/add_user_certificate.rs
TimeEngineer: Because there is also `alice` user in `env`

libparsec/crates/client/tests/unit/certif/add_device_certificate.rs
FirelightFlagboy: You mean because on line 33 we create a device alice and only line 35 we create another device alice ?

TimeEngineer: No I mean there is already a device in the testbed `minimal`, `l.35` shows that there is one, but it's `LocalDevice`

libparsec/crates/client/tests/unit/certif/add_device_certificate.rs
```diff
@@ -27,6 +27,27 @@ async fn ok(env: &TestbedEnv) {
     p_assert_matches!(switch, MaybeRedactedSwitch::NoSwitch);
 }
 
+#[parsec_test(testbed = "minimal")]
+async fn multiple(env: &TestbedEnv) {
```
touilleMan, line 31: I think the name is not clear enough

The `ok` test ensure we can add all the certificates (i.e. the one we are interested into, and the ones it depends on) in an empty certif ops
The `multiple` test ensure we can add multiple certificates of the same type

What about:
- rename `ok` tests to `ok_from_scratch`
- add a call to `builder.certificates_storage_fetch_certificates()` so that the `multiple` test only insert the certificates we are interested into (and not the certificates they depend on)
- rename `multiple` tests to `ok_only_2_additional_certifs` (I'm okay with a better name ^^)

libparsec/crates/client/tests/unit/certif/add_device_certificate.rs
```diff
@@ -27,6 +27,27 @@ async fn ok(env: &TestbedEnv) {
...
+    let env = env.customize(|builder| {
+        builder.new_device("alice");
+    });
+    let alice = env.local_device("alice@dev1");
+    let ops = certificates_ops_factory(&env, &alice).await;
+
+    let switch = ops
+        .add_certificates_batch(
+            &env.get_common_certificates_signed(),
```
touilleMan, line 40: Given the test is about adding multiple certificates, it would be good to have an assert about the number of certificates:
```rust
let to_add_certifs = env.get_common_certificates_signed();
p_assert_eq!(to_add_certif.len(), 2);  // Sanity check
...
let switch = ops
        .add_certificates_batch(
            &to_add_certifs,
...
```

libparsec/crates/client/tests/unit/certif/add_device_certificate.rs
touilleMan: would be good to add a comment in the customize to explain which already existing certificates we are going to use in the integration

libparsec/crates/client/tests/unit/certif/add_user_certificate.rs
touilleMan: add a comment then ;-)

touilleMan (changes requested):

libparsec/crates/client/tests/unit/certif/add_device_certificate.rs
TimeEngineer: Done

libparsec/crates/client/tests/unit/certif/add_user_certificate.rs
TimeEngineer: Done

libparsec/crates/client/tests/unit/certif/add_device_certificate.rs
TimeEngineer: It can be done in another PR to change the tests name, because this PR adds some tests (not all), so there are others which need to be changed.

I agree that the 2nd point is cleaner, but that doesn't bring more and the test would be longer

libparsec/crates/client/tests/unit/certif/add_device_certificate.rs
TimeEngineer: Don't forget that there is also all the required certificates, and even if you want to fetch and add your own certificates, the `env` will create `alice user & device`

touilleMan: yes, my example is broken here, I should have done something like:

```rust
let all_common_certifs = env.get_common_certificates_signed();
let to_add_certifs = all_common_certifs[all_common_certifs.len() - 2..];
```
Or even better: have the builder return the list of certificates that have been created since the `builder.certificates_storage_fetch_certificates()`

libparsec/crates/client/tests/unit/certif/add_device_certificate.rs
touilleMan: > I agree that the 2nd point is cleaner,

What point are you talking about ?

TimeEngineer: > add a call to builder.certificates_storage_fetch_certificates() so that the multiple test only insert the certificates we are interested into (and not the certificates they depend on)

touilleMan: > I agree that the 2nd point is cleaner, but that doesn't bring more and the test would be longer

It is an important test though: we are testing the certificate ops is able to go fetch the requested data from it local storage.
If we add all certificates at once, we are in the case the certificates are only fetched from the cache.

So I really think we should do that to have a good coverage !

TimeEngineer: What about testing without the cache then ?

TimeEngineer: In which situation is it different ? Can we ensure that for every certificate: adding certificate is the same as fetching at a moment ?

libparsec/crates/client/tests/unit/certif/add_device_certificate.rs
TimeEngineer: Need to add this sanity check for others tests then

TimeEngineer: Of course the `len` is `2` if you keep only `2` :joy:
```Rust
let to_add_certifs = all_common_certifs[all_common_certifs.len() - 2..];
```

libparsec/crates/client/tests/unit/certif/add_device_certificate.rs
touilleMan: > What about testing with/without the cache then ?

That's precisely the point of using `builder.certificates_storage_fetch_certificates()`

touilleMan: > In which situation is it different ?

Every time the cache is involved, there is two possible outcome: cache hit or cache miss. Hence both must be tested.

> Can we ensure that for every certificate: adding certificate is the same as fetching at a moment ?

That's the point of those tests.

TimeEngineer: Done

touilleMan (approved):
Don't forget the subsequent PR that fix the naming in the tests ;-)

Issue #6393, opened 2024-02-09
PostgreSQL implementation


mmmarcos:
The PostgreSQL implementation of APIv4 is missing.

## Progress

- #6432 
- #6605
- #7060

## Related issues

- ~#4820~ (postponed)
- #4552 (done in #7060 with a different approach)
- #5801 (done in #7060)

mmmarcos:
@touilleMan can you update this issue to add detail on what it needs to be done?

vxgmichel:
# Current status

## Done

#### SQL data model updated

Including future tables for:
- shamir recovery
- workspace archiving

#### All main modules migrated

- `auth`
- `block`
- `event`
- `invite`
- `organization` (with user freezing)
- `ping`
- `realm`
- `user`
- `vlob`

#### All tests running with `--postgresql`

 - Except `PKI` tests that have been explicitly marked as `xfail`

#### Other changes:

- Have the tests re-use template organization by implementing organization duplication
- Change invitation `greeter` of type `UserID` to `created_by` of type `DeviceID`

## Remaining

#### Add `--postgresql` to CI
#### Cleanup dead code
#### Factorize code duplication
#### Use a systematic approach to concurrency and invariant checks

#### Sort, cleanup and uniformise SQL queries

#### Treat every `# TODO` added during development

- Look for `author: UserID` vs `author: DeviceID`
- What to do about `retry_on_unique_violation` decorators?
- double check in block create
- un-share has no `RealmUnshareStoreBadOutcome.USER_REVOKED`
- etc. (use `grep` to be exhaustive)

## Postponed

#### PKI invites
#### Sequester functionalities
#### Shamir recovery
#### Workspace archiving and deletion

vxgmichel:
6 tasks created:
- [ ] https://github.com/Scille/parsec-cloud/issues/7116
- [ ] https://github.com/Scille/parsec-cloud/issues/7117
- [ ] https://github.com/Scille/parsec-cloud/issues/7118
- [ ] https://github.com/Scille/parsec-cloud/issues/7119
- [ ] https://github.com/Scille/parsec-cloud/issues/7120
- [ ] https://github.com/Scille/parsec-cloud/issues/7121

mmmarcos:
This is completed by #6432, #6605 and #7060. 

Remaining tasks are tracked by new issues and will be discussed to assign a priority.

PR #6432, opened 2024-02-15
A first batch of changes for postgre server implementation

vxgmichel:
See #6393


touilleMan (approved):
I swear I read everything :trollface:

PR #6455, opened 2024-02-19
Binding: Add `change_access`

### LINKED ISSUE #5824: Change device authentication
Change password/authentication of the current device.

Something like `(clientHandle: Handle, currentAuth: DeviceAccessStrategy, newAuth: DeviceSaveStrategy) -> ??`

TimeEngineer:
closes #5824

### DIFF
--- a/client/src/plugins/libparsec/definitions.ts
+++ b/client/src/plugins/libparsec/definitions.ts
@@ -436,6 +436,36 @@ export type ClientCancelInvitationError =
   | ClientCancelInvitationErrorNotFound
   | ClientCancelInvitationErrorOffline
 
+// ClientChangeAuthentificationError
+export enum ClientChangeAuthentificationErrorTag {
+    DecryptionFailed = 'ClientChangeAuthentificationErrorDecryptionFailed',
+    Internal = 'ClientChangeAuthentificationErrorInternal',
+    InvalidData = 'ClientChangeAuthentificationErrorInvalidData',
+    InvalidPath = 'ClientChangeAuthentificationErrorInvalidPath',
+}
+
+export interface ClientChangeAuthentificationErrorDecryptionFailed {
+    tag: ClientChangeAuthentificationErrorTag.DecryptionFailed
+    error: string
+}
+export interface ClientChangeAuthentificationErrorInternal {
+    tag: ClientChangeAuthentificationErrorTag.Internal
+    error: string
+}
+export interface ClientChangeAuthentificationErrorInvalidData {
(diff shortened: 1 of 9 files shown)


Max-7 (approved):

client/src/plugins/libparsec/definitions.ts
```diff
@@ -1361,6 +1373,10 @@ export interface LibParsecPlugin {
         client: Handle,
         token: InvitationToken
     ): Promise<Result<null, ClientCancelInvitationError>>
+    clientChangeAccess(
+        current_access: DeviceAccessStrategy,
+        new_access: DeviceAccessStrategy
```
Max-7, line 1378: Shouldn't this be a `DeviceSaveStrategy` instead? The `keyFile` shouldn't change.

client/src/plugins/libparsec/definitions.ts
```diff
@@ -436,6 +436,18 @@ export type ClientCancelInvitationError =
   | ClientCancelInvitationErrorNotFound
   | ClientCancelInvitationErrorOffline
 
+// ClientChangeAccessError
+export enum ClientChangeAccessErrorTag {
+    ChangeAccess = 'ClientChangeAccessErrorChangeAccess',
```
Max-7, line 441: It's lacking in details. `clientStart` can have all these errors
```
ClientStartErrorInternal
 ClientStartErrorLoadDeviceDecryptionFailed
ClientStartErrorLoadDeviceInvalidData
ClientStartErrorLoadDeviceInvalidPath
```
which probably means that `clientChangeAccess` should at least have those because you get the same problems (finding the device key file, decrypting it, deserializing it).

Max-7 (changes requested):

touilleMan (commented):
LGTM

`client_change_access` seems too vague to me though 
(`libparsec_platform_device_loader::change_access` has a context that helps understanding it is about the device file, but `client_change_access` doesn't have such... reading this PR I was first thinking this was about changing file attributes access :/)

what about renaming it to ~`client_device_loader_change_access`~ `client_change_authentication` ?

client/src/plugins/libparsec/definitions.ts
TimeEngineer: Done

client/src/plugins/libparsec/definitions.ts
TimeEngineer: Done

Max-7 (approved):

PR #6463, opened 2024-02-19
Add missing tests for block_create

### LINKED ISSUE #6462: Server test block_create

mmmarcos:
Closes #6462

### DIFF
--- a/server/tests/api_v4/authenticated/test_block_create.py
+++ b/server/tests/api_v4/authenticated/test_block_create.py
@@ -42,6 +42,9 @@ async def test_authenticated_block_create_realm_not_found(
     assert rep == authenticated_cmds.v4.block_create.RepRealmNotFound()
 
     # Block should not exist
+    # This check is not strictly needed since the block existence
+    # is already checked below with the test_dump_blocks call.
+    # Anyway, let's leave it as an extra cheap check.
     content = await backend.block.read_as_user(
         organization_id=coolorg.organization_id,
         author=coolorg.alice.device_id.user_id,
@@ -109,6 +112,9 @@ async def test_authenticated_block_create_store_unavailable(
     assert rep == authenticated_cmds.v4.block_create.RepStoreUnavailable()
 
     # Block should not exist
+    # This check is not strictly needed since the block existence
+    # is already checked below with the test_dump_blocks call.
+    # Anyway, let's leave it as an extra cheap check.
     content = await backend.block.read_as_user(
         organization_id=coolorg.organization_id,
         author=coolorg.alice.device_id.user_id,
--- a/server/tests/api_v4/authenticated/test_block_create.py
+++ b/server/tests/api_v4/authenticated/test_block_create.py
@@ -2,7 +2,11 @@
 
 from unittest.mock import ANY
 
-from parsec._parsec import BlockID, DateTime, authenticated_cmds
+import pytest
+
+from parsec._parsec import BlockID, DateTime, VlobID, authenticated_cmds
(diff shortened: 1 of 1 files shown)


server/tests/api_v4/authenticated/test_block_create.py
```diff
@@ -27,6 +31,28 @@ async def test_authenticated_block_create_ok(coolorg: CoolorgRpcClients, backend
...
+) -> None:
+    block_id = BlockID.new()
+    block = b"<block content>"
+
+    # Try to create block with an invalid realm ID
+    rep = await coolorg.alice.block_create(block_id, VlobID.new(), block)
+    assert rep == authenticated_cmds.v4.block_create.RepRealmNotFound()
+
+    # Block should not exist
+    content = await backend.block.read_as_user(
```
touilleMan, line 45: No need for this part given `test_dump_blocks` already shows us the block doesn't exist

Note: `read_as_user` is present for the `_ok`  test case. This is not strictly required, but works as a cheap check to ensure the block can be actually retrieved (while testing the block cannot be retrieved when it is not present in the first place seems very paranoid).

In the end I'm fine with either of those solutions:
- do nothing and keep `read_as_user` check in all tests
- considering `test_dump_blocks` is already enough and remove `read_as_user` in all tests 
- keep `read_as_user` only in the `_ok` test to balance test readability

your call :+1:

server/tests/api_v4/authenticated/test_block_create.py
```diff
@@ -27,6 +31,28 @@ async def test_authenticated_block_create_ok(coolorg: CoolorgRpcClients, backend
...
+        organization_id=coolorg.organization_id,
+        author=coolorg.alice.device_id.user_id,
+        block_id=block_id,
+    )
+    assert content != BlockReadBadOutcome.BLOCK_NOT_FOUND.value
```
touilleMan, line 50: This seems strange to me: `read_as_user` is supposed to return a `BlockReadBadOutcome` enum, and we expect it to be abound the block not being found...

```suggestion
    assert content = BlockReadBadOutcome.BLOCK_NOT_FOUND
```

server/tests/api_v4/authenticated/test_block_create.py
```diff
@@ -63,4 +89,32 @@ async def test_authenticated_block_create_not_allowed(
...
+    rep = await coolorg.alice.block_create(block_id, coolorg.wksp1_id, block)
+    assert rep == authenticated_cmds.v4.block_create.RepStoreUnavailable()
+
+    # Block should not exist
+    content = await backend.block.read_as_user(
+        organization_id=coolorg.organization_id,
+        author=coolorg.alice.device_id.user_id,
+        block_id=block_id,
+    )
+    assert content != BlockReadBadOutcome.BLOCK_NOT_FOUND.value
```
touilleMan, line 117: same as above
```suggestion
    assert content = BlockReadBadOutcome.BLOCK_NOT_FOUND
```

touilleMan (changes requested):

server/tests/api_v4/authenticated/test_block_create.py
mmmarcos: You're right. Actually, if ever `read_as_user` check fails, but not `test_dump_blocks`, that would mean there is a bug in `read_as_user`, not really in `block_create`.

Anyway, maybe it is better to leave it as a "cheap check". I added some comments based on your review.
WDYT?

touilleMan: > WDYT

:+1:

Issue #6476, opened 2024-02-21
Allow fast resize operations


vxgmichel:
At the moment, the `fd_resize` transaction is very costly for large files since chunks of data filled up zeroes are created, ciphered, and stored in the sqlite database. This is an issue since this is the first operation Windows do when creating a new file. If the file is large, this operations might take up to several minutes and disrupt the file explorer as the manifest is locked during the whole operation.

A possible solution that does not affect the data model is to flag the chunk objects that correspond to an all-zeros data range. This way, such chunks don't have to be read from/written to the local database. 

If we're willing to break compatibility (i.e if we introduce this change during the first v3 release), we can simply skip the all-zeroes blocks when uploading the blocks and generating the remote file manifest. When such a manifest is loaded by another client, all-zeroes blocks are automatically created to fill up the gaps.

If we're not willing to break compatibility, a single all-zeroes chunk might be created once and referenced several times in the local manifest during the resize operation. This way, the system keeps working just like before except that only a single all-zeroes block is uploaded during the synchronization.

In either case, such a solution can easily be tested on v2 and deployed on v3.

PR #6483, opened 2024-02-22
Another POC for faster fd_resize operation

vxgmichel:
Another approach for #6476 

Related: #6481


vxgmichel:
Outdated by #6508

PR #6508, opened 2024-02-24
Yet another POC for faster fd_resize operation

vxgmichel:
A POC for #6476 

Related:
- #6481 
- #6483


vxgmichel:
Implemented in parsec v3 in https://github.com/Scille/parsec-cloud/pull/6577

PR #6521, opened 2024-02-26
Fix `change_authentification`

TimeEngineer:


FirelightFlagboy (commented):
Can you add a test ?

FirelightFlagboy (commented):
I would prefer that the param `with_template` to be renamed to `load_from_testbed_template`

TimeEngineer:
> I would prefer that the param `with_template` to be renamed to `load_from_testbed_template`

It's not necessary coming from `testbed_template`: It's search first from the `cache`, then if it doesn't find, it will retrieve from the `testbed template`

FirelightFlagboy:
I would like that the parameter have at least the indication that it's only used by the testbed

TimeEngineer:
> I would like that the parameter have at least the indication that it's only used by the testbed

I changed for `with_testbed_template`

FirelightFlagboy (approved):

Issue #6546, opened 2024-02-27
Create nighly builds for early user testing 


mmmarcos:
Client and Server nightly builds. 

1 build/day should be enough, no need to store previous builds.

FirelightFlagboy:
Superseded by #6588

PR #6555, opened 2024-02-28
Rfc nightly build

FirelightFlagboy:
## Describe your changes

Add RFC to discuss then implement nightly build into parsec.

The rendered file can be viewed here: <https://github.com/Scille/parsec-cloud/blob/rfc-nightly-build/docs/rfcs/1008-nightly-build.md>

The way it's written allows to implement it in part independently

Related to #6546

### DIFF
--- a/docs/rfcs/1008-nightly-build.md
+++ b/docs/rfcs/1008-nightly-build.md
@@ -1,0 +1,160 @@
+<!-- Parsec Cloud (https://parsec.cloud) Copyright (c) BUSL-1.1 2016-present Scille SAS -->
+
+# Nightly build
+
+This RFC proposes adding nightly builds to Parsec.
+
+- [Goals](#goals)
+- [What we already have](#what-we-already-have)
+- [What is missing](#what-is-missing)
+  - [For the client part](#for-the-client-part)
+    - [Missing version argument in the workflow `package-ionic-app.yml`](#missing-version-argument-in-the-workflow-package-ionic-appyml)
+    - [Include the client part in the releaser workflow](#include-the-client-part-in-the-releaser-workflow)
+  - [Overwriting nightly release \& automatic release](#overwriting-nightly-release--automatic-release)
+  - [Snapcraft channel with tracks](#snapcraft-channel-with-tracks)
+  - [Schedule the workflow](#schedule-the-workflow)
+  - [Slack notification (optional)](#slack-notification-optional)
+  - [Improve the unique dev format](#improve-the-unique-dev-format)
+- [F.A.Q](#faq)
+  - [Why do we need to release the nightly build? Or why we can't directly use the artifact link of a workflow?](#why-do-we-need-to-release-the-nightly-build-or-why-we-cant-directly-use-the-artifact-link-of-a-workflow)
+  - [Why do we use the `dev` identifier for a nightly build ?](#why-do-we-use-the-dev-identifier-for-a-nightly-build-)
+
+## Goals
+
+- Periodically test the release pipeline.
+- Have a recent build software at the ready (client and server).
+
+## What we already have
+
+We already have a workflow to create a release and build its artifacts[^releaser-workflow]. It is triggered by pushing a tag with the format `v[0-9]+.[0-9]+.[0-9]+*`.
+
+Currently, the workflow builds the server's Python wheel for Linux, Windows & macOS.
+
+[^releaser-workflow]: https://github.com/Scille/parsec-cloud/blob/05e4deba964c6b68b6fcbca4d458c235c43f6cff/.github/workflows/releaser.yml
+
+The script `releaser.py` can generate unique version that can be used for versioning and tagging nightly build.
+
+## What is missing
+
+### Client packaging
+
+The following changes are required for the [`package-ionic-app.yml`] workflow:
+
+- it needs to take a version argument (like [`package-server.yml`]).
+- it needs to be added to the [`releaser.yml`] workflow.
+
+[`package-server.yml`]: https://github.com/Scille/parsec-cloud/blob/05e4deba964c6b68b6fcbca4d458c235c43f6cff/.github/workflows/package-server.yml
+[`package-ionic-app.yml`]: https://github.com/Scille/parsec-cloud/blob/05e4deba964c6b68b6fcbca4d458c235c43f6cff/.github/workflows/package-server.yml
+[`releaser.yml`]: https://github.com/Scille/parsec-cloud/blob/05e4deba964c6b68b6fcbca4d458c235c43f6cff/.github/workflows/releaser.yml
+
+### Overwriting nightly release & automatic release
+
+The [`releaser.yml`] workflow creates a [GitHub release] in draft mode. Creating a `draft-release` for _every_ nightly build (everyday so) is problematic because we might be overwhelmed by the number of draft releases.
+
+We could [override the assets] of the previous nightly release (instead of recreating a new one). The problem with this approach is that GitHub sort the releases by creation date: nightly release will be put behind newer releases.
+
+So the idea is to _remove_ the previous nigthly release and recreate a new one. In order to do so, the API to [delete a release] requires a `release_id`.
+
+[override the assets]: https://stackoverflow.com/questions/62934246/github-update-overwrite-existing-asset-of-a-release
+[GitHub release]: https://docs.github.com/en/repositories/releasing-projects-on-github/about-releases
+
+[delete a release]: https://docs.github.com/en/rest/releases/releases?apiVersion=2022-11-28#delete-a-release
+
+The `release_id` could be obtained by [getting a release using its tag name] but a `draft-release` isn't identified by a tag name (the more you know).
+
+[getting a release using its tag name]: https://docs.github.com/en/rest/releases/releases?apiVersion=2022-11-28#get-a-release-by-tag-name
+
+Another possibility is to [list all releases] to find the one to be removed. This would allow to update the workflow in order to directly create the release (see [Why do we need to release the nightly build?](#why-do-we-need-to-release-the-nightly-build-or-why-we-cant-directly-use-the-artifact-link-of-a-workflow)).
+
+[list all releases]: https://docs.github.com/en/rest/releases/releases?apiVersion=2022-11-28#list-releases
+
+### Snapcraft channel with tracks
+
+Currently we only use the `stable` & `edge` from the default `latest` track. We could introduce a new [track] in order to include nightly builds, for example:
+
+- `latest`: We already have this one since it's the default track (when no track is specified)
+- `v{MAJOR}`: A track per major version (in the case we need to maintain multiple version at the same time)
+- `nightly`: A track for the nightly builds
+
+  > The name `nightly` (Firefox style) was chosen by the parsec-dev team based on a poll. Alternatives that were considered:
+  > - `canary` (Chrome style)
+  > - `insider` (VS-Code style)
+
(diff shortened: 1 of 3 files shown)


docs/rfcs/1008-nightly-build.md
```diff
@@ -0,0 +1,206 @@
...
+
+Slack provide GitHub action for that: <https://github.com/slackapi/slack-github-action>
+
+### Improve the unique dev format
+
+Currently, `releaser.py` generate a dev version by appending the `dev` part at the end of the current version before the local part.
+
+The dev part is formatted that way `dev.{number}`. The `{number}` value is calculated from the current date with the following formula:
+
+$f(year, month, day) = year * 366 + month * 31 + day$
```
mmmarcos, line 146: Where do this formula come from?

It is not clear to me why is this an _improvement_ (as in the section title)? Is there a particular problem/issue with the current formula (besides the mistake on year factor)? 
If the answer is no, is probably better to open an issue to "fix" the dev format, proposing the alternatives that you listed here.

docs/rfcs/1008-nightly-build.md
```diff
@@ -0,0 +1,206 @@
...
+
+We need to have a version that is compatible with [semver] and [pep440], that indicate that it's a nightly build.
+
+On [semver], we could use any verb like `nightly`, so the choice come from [pep440].
+
+Pep440 only allows the verb `dev` at the end of the version before the local identifiers ([Specify a development release in pep440])
+
+[semver]: https://semver.org/
+[pep440]: https://peps.python.org/pep-0440/
+[Specify a development release in pep440]: https://peps.python.org/pep-0440/#developmental-releases
```
mmmarcos, line 206: There is a note in PEP440 saying that the reference spec is in https://packaging.python.org/en/latest/specifications/version-specifiers/#developmental-releases

docs/rfcs/1008-nightly-build.md
```diff
@@ -0,0 +1,206 @@
...
+- You could not easily know if the workflow summary you're using is the latest available.
+- The artifacts may contain data that is not useful for you.
+
+### Why do we use the `dev` identifier for a nightly build ?
+
+We need to have a version that is compatible with [semver] and [pep440], that indicate that it's a nightly build.
+
+On [semver], we could use any verb like `nightly`, so the choice come from [pep440].
+
+Pep440 only allows the verb `dev` at the end of the version before the local identifiers ([Specify a development release in pep440])
```
mmmarcos, line 202: I don't know if it makes sense, but how about adding `nigthly` to the local identifiers?

FirelightFlagboy: The local identifier isn't used when comparing 2 version together

docs/rfcs/1008-nightly-build.md
```diff
@@ -0,0 +1,206 @@
...
+
+> We could also [list all releases] to find the one we look for.
+
+[list all releases]: https://docs.github.com/en/rest/releases/releases?apiVersion=2022-11-28#list-releases
+
+So we could update the workflow to directly create the release.
+
+> At this step, you may be asking yourself [Why do we need to release the nightly build ?](#why-do-we-need-to-release-the-nightly-build-or-why-we-cant-directly-use-the-artifact-link-of-a-workflow)
+
+### Snapcraft channel with tracks
```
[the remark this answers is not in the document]
FirelightFlagboy: You mean that we should push the nightly build into latest ?

docs/rfcs/1008-nightly-build.md
```diff
@@ -0,0 +1,206 @@
...
+[GitHub release]: https://docs.github.com/en/repositories/releasing-projects-on-github/about-releases
+
+This is problematic if the workflow create a `draft-release` for each nightly build (everyday so).
+
+> We would be overwhelmed by the amount of draft releases that would accumulate.
+
+To circumvent that we could make the workflow to remove the previous nightly release.
+But we have some limitation, the API to [delete a release] require a `release_id`.
+
+[delete a release]: https://docs.github.com/en/rest/releases/releases?apiVersion=2022-11-28#delete-a-release
```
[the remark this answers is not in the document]
FirelightFlagboy: We could do that. I've choose that approach of re-creating the release because GitHub sort the releases by creation date. that mean the nightly release will be put behind older releases.

docs/rfcs/1008-nightly-build.md
FirelightFlagboy: They're virtually the same :shrug:

docs/rfcs/1008-nightly-build.md
FirelightFlagboy: > Where do this formula come from?

From my hat :tophat:

> It is not clear to me why is this an improvement ...

For me some formulas are simpler to use

> Is there a particular problem/issue with the current formula (besides the mistake on year factor)?

The problem is that value overlap between December and January:

```math
\begin{align}
f(y, m, d) = y * 366 + m * 31 + d \\
f(0, 12, 26) = f(1, 1, 1) = 398 \\
f(0, 12, 31) = f(1, 1, 6) = 403
\end{align}
```

docs/rfcs/1008-nightly-build.md
mmmarcos: :man_shrugging:

docs/rfcs/1008-nightly-build.md
mmmarcos: > > Where do this formula come from?
> 
> From my hat 🎩

So indeed, better try to find a more standard way. I am not sure that we really need the date information, so maybe just an incremental index.

> > Is there a particular problem/issue with the current formula (besides the mistake on year factor)?
> 
> The problem is that value overlap between December and January:

IMO this is acceptable for the moment. I suggest dealing with this subject in a separate issue (so we can discuss there what is the best naming scheme).

docs/rfcs/1008-nightly-build.md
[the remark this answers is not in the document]
mmmarcos: You mean it will be put behind _newer_ releases? That could be inconvenient later, but I think it's no big deal for the moment. Right now, there are no other release planned on the `v2.x` or `v3.x` branch and we need a nightly release _of the upcoming v3.0_. We could improve the nigthly workflow later, based on this RFC.
WDYT?

docs/rfcs/1008-nightly-build.md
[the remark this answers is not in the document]
mmmarcos: Resolving the conversation as this is no longer relevant (see https://github.com/Scille/parsec-cloud/pull/6590)

docs/rfcs/1008-nightly-build.md
mmmarcos: Resolving the conversation as this is no longer relevant (see https://github.com/Scille/parsec-cloud/pull/6592)

docs/rfcs/1008-nightly-build.md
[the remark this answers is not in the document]
mmmarcos: Maybe we can put the two alternatives (update vs delete+create) in the RFC, just to document them so we can merge the PR.

docs/rfcs/1008-nightly-build.md
```diff
@@ -0,0 +1,208 @@
...
+
+[`package-server.yml`]: https://github.com/Scille/parsec-cloud/blob/05e4deba964c6b68b6fcbca4d458c235c43f6cff/.github/workflows/package-server.yml
+
+#### Include the client part in the releaser workflow
+
+We need to add the client to the workflow [`releaser.yml`].
+Thankfully we already have the workflow for that ([`package-ionic-app.yml`]).
+
+[`package-ionic-app.yml`]: https://github.com/Scille/parsec-cloud/blob/05e4deba964c6b68b6fcbca4d458c235c43f6cff/.github/workflows/package-server.yml
+[`releaser.yml`]: https://github.com/Scille/parsec-cloud/blob/05e4deba964c6b68b6fcbca4d458c235c43f6cff/.github/workflows/releaser.yml
```
mmmarcos, line 59: This is quite verbose, I would summarize it:

```suggestion
### Client packaging

The following changes are required for the [`package-ionic-app.yml`] workflow:

- it needs to take a version argument (like [`package-server.yml`]).
- it needs to be added to the [`releaser.yml`] workflow.

[`package-server.yml`]: https://github.com/Scille/parsec-cloud/blob/05e4deba964c6b68b6fcbca4d458c235c43f6cff/.github/workflows/package-server.yml
[`package-ionic-app.yml`]: https://github.com/Scille/parsec-cloud/blob/05e4deba964c6b68b6fcbca4d458c235c43f6cff/.github/workflows/package-server.yml
[`releaser.yml`]: https://github.com/Scille/parsec-cloud/blob/05e4deba964c6b68b6fcbca4d458c235c43f6cff/.github/workflows/releaser.yml
```

docs/rfcs/1008-nightly-build.md
```diff
@@ -0,0 +1,208 @@
...
+Currently, the [`releaser.yml`] workflow create a [GitHub release] in draft mode.
+
+[GitHub release]: https://docs.github.com/en/repositories/releasing-projects-on-github/about-releases
+
+This is problematic if the workflow create a `draft-release` for each nightly build (everyday so).
+
+> We would be overwhelmed by the amount of draft releases that would accumulate.
+
+To circumvent that we could make the workflow to remove the previous nightly release.
+But we have some limitation, the API to [delete a release] require a `release_id`.
```
mmmarcos, line 72:
```suggestion
The [`releaser.yml`] workflow creates a [GitHub release] in draft mode. Creating a `draft-release` for _every_ nightly build (everyday so) is problematic because we might be overwhelmed by the number of draft releases.

We could [override the assets] of the previous nightly release (instead of recreating a new one). The problem with this approach is that GitHub sort the releases by creation date: nightly release will be put behind newer releases.

So the idea is to _remove_ the previous nigthly release and recreate a new one. In order to do so, the API to [delete a release] requires a `release_id`.

[override the assets]: https://stackoverflow.com/questions/62934246/github-update-overwrite-existing-asset-of-a-release
[GitHub release]: https://docs.github.com/en/repositories/releasing-projects-on-github/about-releases
```

docs/rfcs/1008-nightly-build.md
```diff
@@ -0,0 +1,208 @@
...
+- The corrected formulas using $372$:
+
+  For `2024-02-27 15:50:28+01:00`, the value is calculated with:
+
+  $$f'(year, month, day) = year * 372 + month * 31 + day$$
+  $$f'(2024, 2, 27) = 753017$$
+
+  That would give `dev.753017`
+
+- A date (`dev.YYYYMMDD`): For `2024-02-27 15:50:28+01:00` That gives `dev.20240227`
```
mmmarcos, line 183:
```suggestion
Considering the previous example date, the formula could be changed to:

- A _timestamp_: `dev.1709045428`
- A _timestamp_ without the time info ($floor(timestamp / 24 * 3600)$ : `dev.19780` 
- The current formula with fixed value ($year * 372 + month * 31 + day$): `dev.753017`
- A date format (`dev.YYYYMMDD`): `dev.20240227`
```

mmmarcos (commented):
LGTM overall. Some suggestions to summarize the different sections a little bit.

docs/rfcs/1008-nightly-build.md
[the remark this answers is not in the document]
FirelightFlagboy: For tracking purpose, we have a paragraph that mention `override the assets`

mmmarcos (approved):
:+1:

PR #6557, opened 2024-02-28
A POC to add keyring and biometrics as an authent method to access the devices

vxgmichel:
A demo for parsec v2


vxgmichel:
Outdated by #9681

PR #6560, opened 2024-02-28
Revoke user

touilleMan:


Max-7 (approved):

libparsec/crates/client/src/certif/user_revoke.rs
```diff
@@ -0,0 +1,170 @@
...
+
+pub(super) async fn revoke_user(
+    ops: &CertifOps,
+    user: UserID,
+) -> Result<CertificateBasedActionOutcome, CertifRevokeUserError> {
+    if *ops.device.device_id.user_id() == user {
+        return Err(CertifRevokeUserError::UserIsSelf);
+    }
+
+    // Loop is needed to deal with server requiring greater timestamp
```
TimeEngineer, line 69: Maybe we can use a generic function which handle greater timestamp
fn foo<T>(...) -> T {
    loop { ... }
}

TimeEngineer (approved):
LGTM for `rust` part
Need to change a test because the `json` has been modified

server/tests/api_v4/authenticated/test_user_revoke.py
```diff
@@ -75,3 +75,37 @@ async def test_disconnect_sse(
...
+    coolorg: CoolorgRpcClients, backend: Backend
+) -> None:
+    t1 = DateTime.now()
+    certif1 = RevokedUserCertificate(
+        author=coolorg.alice.device_id,
+        timestamp=t1,
+        user_id=coolorg.bob.device_id.user_id,
+    )
+
+    outcome = await backend.user.revoke_user(
```
mmmarcos, line 90: This test is testing the rpc client, isn't it ? so why not using `coolorg.alice.user_revoke` twice?

mmmarcos (approved):
LGTM for the `server` part.
Just a minor comment on test user already revoked

libparsec/crates/client/src/certif/user_revoke.rs
touilleMan: I'd rather keep it simple given how fast Rust + async becomes out of hand

Typically your `foo` method has to be asynchronous and capture the lifetime of the closure's parameter, I had to do that for certificate store and it involve unsafe stuff (in this case they it was worth it because it bring guarantee on how data should be locked, while here it's just a convenient helper)

https://github.com/Scille/parsec-cloud/blob/ea898fa300fc2292b78a13d81450fe7ad34b1d5f/libparsec/crates/client/src/certif/store.rs#L158-L229

server/tests/api_v4/authenticated/test_user_revoke.py
mmmarcos: It felt weird at first... it seemed to me that is more interesting to test the (lack of) _idempotence_ of the `user_revoke` (rpc) call rather than modifying state + testing that the function behaves as expected. Calling `coolorg.alice.user_revoke` twice would also tests that the function has set the correct state (revoke certificate) in the first call. 

I guess you could say that this is already tested in the "ok" test. 
So, in the end both seem more or less equivalent :+1:

Issue #6559, opened 2024-02-28
Add user filtering


Max-7:
https://www.figma.com/file/XW3SNWHV8BhPsPJpsOCV2h/Application?type=design&node-id=29131-40930&mode=dev

PR #6567, opened 2024-02-29
[MS] Added sort to UsersPage

### LINKED ISSUE #6451: Sort-by item on `UserPage`
https://www.figma.com/file/XW3SNWHV8BhPsPJpsOCV2h/Application?node-id=25612%3A118643&mode=dev

See what's done for files

Ironicbay:
Closes #6451 and #6559

## Checklist

- [ ] Describe your changes
   
If this PR is related to GUI:


Max-7 (changes requested):

client/tests/e2e/specs/test_users_page.ts
```diff
@@ -20,35 +20,98 @@ describe('Check active users page', () => {
     cy.get('#activate-users-ms-action-bar').find('#list-view').should('have.attr', 'disabled');
     cy.get('#activate-users-ms-action-bar').find('#button-invite-user').contains('Invite a user');
     cy.get('.user-list-header').find('ion-checkbox').should('not.have.class', 'checkbox-checked');
-    cy.get('.users-container').find('.user-list-item').as('userItems').should('have.length', 4);
+    cy.get('.users-container').find('.user-list-item').as('userItems').should('have.length', 13);
```
Max-7, line 23: You're going way too far with the tests. Having 3-4 users should be enough, there's no need to have 12 of them that are mostly similar and checking that each invidual checkbox is checked or not. These tests are already a chore to maintain, having a lot of copy/paste makes it even more of a struggle.

Just keep the sort test, check that the first and last of the list are what we expect, and it's more than enough.

client/src/components/users/types.ts
```diff
@@ -51,9 +61,42 @@ export class UserCollection {
     return this.users.filter((user) => user.isSelected && user.isRevoked()).length;
   }
 
-  sort(_property: SortProperty, _ascending: boolean): void {
-    this.users.sort((_item1, _item2) => {
-      return 1;
+  sort(property: SortProperty, ascending: boolean): void {
```
Max-7, line 64: `SortProperty` also contains `Email` but it's not used in the code. Either add the ability to sort on it or remove it from the enum.

This function could probably benefit from using switch/cases instead of ifs, the linter should be able to warn us if you forget to cover a case (like here with the `Email`).

Max-7 (changes requested):

PR #6576, opened 2024-03-01
[MS] UsersPage filtering

### LINKED ISSUE #6559: Add user filtering

Ironicbay:
Closes #6559 

## Checklist

- [ ] Describe your changes
   
If this PR is related to GUI:

### DIFF
--- a/client/src/components/users/types.ts
+++ b/client/src/components/users/types.ts
@@ -23,37 +23,83 @@ function compareUserProfiles(profile1: UserProfile, profile2: UserProfile): numb
   return WEIGHTS.get(profile2)! - WEIGHTS.get(profile1)!;
 }
 
+export interface UserFilterLabels {
+  statusActive?: boolean;
+  statusRevoked?: boolean;
+  profileAdmin?: boolean;
+  profileStandard?: boolean;
+  profileOutsider?: boolean;
+}
+
+export interface UserFilterChangeEvent {
+  filters: UserFilterLabels;
+}
+
 export class UserCollection {
   users: Array<UserModel>;
+  filters: UserFilterLabels;
 
   constructor() {
     this.users = [];
+    this.filters = {
+      statusActive: true,
+      statusRevoked: true,
+      profileAdmin: true,
+      profileStandard: true,
+      profileOutsider: true,
+    };
   }
 
   hasSelected(): boolean {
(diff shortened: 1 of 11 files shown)


client/src/components/core/ms-filter/types.ts
```diff
@@ -0,0 +1,13 @@
+// Parsec Cloud (https://parsec.cloud) Copyright (c) BUSL-1.1 2016-present Scille SAS
+
+export interface MsFilterLabels {
```
Max-7, line 3: I would mark them as optional (`statusActive?: boolean`) and consider undefined as false, which makes it easier to call the function
```typescript
getVisibleUsers({statusActive: true, profileStandard: true});
```
instead of having to list them all.

client/src/components/users/types.ts
```diff
@@ -39,6 +40,21 @@ export class UserCollection {
     return this.users.length;
   }
 
+  getVisibleUsers(filters: MsFilterLabels): Array<UserModel> {
```
Max-7, line 43: Be mindful of selection: selecting all will select hidden users, and selecting a user and then hiding them will not reset the selection. I think simply adding a function will not be enough.

client/src/components/core/ms-filter/MsFilterPopover.vue
```diff
@@ -0,0 +1,96 @@
...
+
+<template>
+  <ion-list class="filter-container">
+    {{ $t('UsersPage.filter.status') }}
+    <div>
+      <user-status-tag :revoked="false" />
+      <ion-checkbox
+        aria-label=""
+        class="checkbox"
+        :checked="props.filters.statusActive"
```
Max-7, line 11: You can replace it all with

```vue
<ion-checkbox
  aria-label=""
  class="checkbox"
  v-model="filters.statusActive"
/>
```
The v-model will automatically bind the value of the checkbox to our variable, which means that the value will be updated without having to dismiss the popover.

Max-7 (changes requested):

Max-7 (approved):

NicoTuxx (approved):

PR #6577, opened 2024-03-01
Use proptest for file operations

vxgmichel:
And revert file_operations to v2 logic


libparsec/crates/client/tests/unit/workspace/file_operations_stateful.rs
```diff
@@ -0,0 +1,211 @@
...
+            .flat_map(|b| b.iter().map(|c| c.id))
+            .collect();
+        let storage_ids: HashSet<_> = state.storage.0.keys().cloned().collect();
+        assert_eq!(manifest_ids, storage_ids);
+    }
+}
+
+prop_state_machine! {
+    #![proptest_config(Config {
+        // Turn failure persistence off for demonstration. This means that no
```
touilleMan, line 187: should still be turned off ? the comment seems odd

libparsec/crates/client/tests/unit/workspace/file_operations_stateful.rs
```diff
@@ -0,0 +1,211 @@
...
+        // regression file will be captured.
+        failure_persistence: None,
+        // Enable verbose mode to make the state machine test print the
+        // transitions for each case.
+        verbose: 0,
+        .. Config::default()
+    })]
+
+    // NOTE: The `#[test]` attribute is commented out in here so we can run it
```
touilleMan: wrong comment

libparsec/crates/client/src/workspace/transactions/fd_read.rs
```diff
@@ -80,7 +81,7 @@ pub async fn fd_read(
         match found {
             Some(chunk_data) => {
                 chunk
-                    .copy_between_start_and_stop(chunk_data, buf)
+                    .copy_between_start_and_stop(chunk_data, offset, buf, &mut buf_size)
```
touilleMan, line 84: `offset` is passed as parameter in the function, so I would expect it to be only used for the first chunk and not the subsequent ones :/

libparsec/crates/types/src/local_manifest.rs
```diff
@@ -108,11 +108,23 @@ impl Chunk {
     pub fn copy_between_start_and_stop(
         &self,
         chunk_data: &[u8],
+        offset: u64,
```
touilleMan, line 111: maybe rename to `offset_after_start` given having a `offset` param in a function called `copy_between_start_and_stop` is a bit puzzling...

libparsec/crates/types/src/local_manifest.rs
```diff
@@ -278,36 +293,49 @@ impl LocalFileManifest {
     pub fn assert_integrity(&self) {
```
touilleMan, line 293: Can you create an issue about this not having the right type ?
(this is rather important given it means any bad data simply crashes the application...)

libparsec/crates/client/src/workspace/transactions/file_operations.rs
```diff
@@ -272,122 +267,127 @@ fn prepare_truncate(
...
 
-// Prepare reshape
-
 pub(crate) struct ReshapeBlockOperation<'a> {
     manifest_chunks: &'a mut Vec<Chunk>,
     reshaped_chunk: Chunk,
 }
 
 impl ReshapeBlockOperation<'_> {
+    pub fn new(chunks: &mut Vec<Chunk>) -> Option<ReshapeBlockOperation> {
```
touilleMan, line 369: Rename `new` as `try_reshape` ? (Given `new` is expected to resurn `Self`, having a method with `try` or `maybe` emphasizes the Option in the return)

libparsec/crates/client/src/workspace/transactions/file_operations.rs
```diff
@@ -396,6 +396,22 @@ impl ReshapeBlockOperation<'_> {
...
 
+    pub fn is_pseudo_block(&self) -> bool {
+        self.manifest_chunks.len() == 1 && self.manifest_chunks[0].is_pseudo_block()
+    }
+
+    pub fn write_back(&self) -> bool {
+        !self.is_pseudo_block()
+    }
+
+    pub fn cleanup_ids(&self) -> HashSet<ChunkID> {
```
touilleMan, line 407: The input data is a Vec, it is used by the caller as a Vec (i.e. for doing a loop)... would you consider using a Vec here :point_right: :point_left: (˶˃ᆺ˂˶)

libparsec/crates/client/src/workspace/transactions/fd_read.rs
vxgmichel: I tried to keep those changes minimal. The reference logic for building data is demonstrated in the test `Storage`:
```rust
    fn read_chunk(&self, chunk: &Chunk) -> &[u8] {
        let data = self.read_chunk_data(chunk.id);
        let start = (chunk.start - chunk.raw_offset) as usize;
        let stop = (chunk.stop.get() - chunk.raw_offset) as usize;
        data.get(start..stop).unwrap()
    }

    fn build_data(&self, chunks: &[Chunk], size: u64, offset: u64) -> Vec<u8> {
        let start = offset as usize;
        let mut result: Vec<u8> = vec![0; size as usize];
        for chunk in chunks {
            let data = self.read_chunk(chunk);
            result[chunk.start as usize - start..chunk.stop.get() as usize - start]
                .copy_from_slice(data);
        }
        result
    }

    // File operations

    pub fn read(&self, manifest: &LocalFileManifest, size: u64, offset: u64) -> Vec<u8> {
        let (written_size, chunks) = prepare_read(manifest, size, offset);
        let data = self.build_data(&chunks.collect::<Vec<_>>(), written_size, offset);
        p_assert_eq!(data.len() as u64, written_size);
        data
```

The TLDR is that the `written_size` returned by `prepare_read` should be used to allocate a `Vec<u8>` filled with zero. Then, each chunk correspond to a slice of data that should be copied into that buffer.

The issue with `fd_read` is that the buffer is a `std::io::Write`, hence the complicated logic.

I'll let you change it as you see fit.

libparsec/crates/types/src/local_manifest.rs
vxgmichel: See this comment: https://github.com/Scille/parsec-cloud/pull/6577#discussion_r1510899522

libparsec/crates/client/src/workspace/transactions/file_operations.rs
vxgmichel: And I already updated the caller:
```rust
            for to_remove_chunk_id in reshape.cleanup_ids() {
                ...
```
I don't see the issue here.

vxgmichel: However, the hashset here is due to a legacy interface, I'll turn it into a vec.

touilleMan (approved):

libparsec/crates/client/src/workspace/transactions/file_operations.rs
vxgmichel: Nevermind, there might be duplicates so it's good to remove them.

libparsec/crates/types/src/local_manifest.rs
vxgmichel: https://github.com/Scille/parsec-cloud/issues/6599

PR #6600, opened 2024-03-04
Bump v3 alpha

FirelightFlagboy:


Cargo.lock
```diff
@@ -29,12 +29,12 @@ dependencies = [
 
 [[package]]
 name = "ahash"
-version = "0.8.7"
+version = "0.8.11"
```
touilleMan, line 32: Why the changes in dependencies ? this PR should only change parsec version...

FirelightFlagboy: The change in version is caused by `cargo generate-lockfile -Z direct-minimal-versions`

That command is run in `misc.version_updater.refresh_cargo_lock`. It's the cleanest way I've found to update the lock file easily

FirelightFlagboy: The goal of this command is to update the `Cargo.lock` file with the updated package version of parsec

mmmarcos: What about something like `cargo update parsec --precise 1.2.3`?

FirelightFlagboy: The command does nothing on my side

mmmarcos: I tried `cargo update libparsec` on your branch (after discarding changes to `Cargo.lock` from this PR) and it only updates `libparsec` and `syn` in `Cargo.lock` :thinking:

FirelightFlagboy: `cargo update` first goal is to update the dependencies.
To generate/update a lock file we should use `generate-lockfile` but currently I update the dependencies.

Cargo have an issue to improve `generate-lockfile` to not update the dependencies (https://github.com/rust-lang/cargo/issues/4269) But it don't have much traction

mmmarcos (approved):
LGTM

PR #6601, opened 2024-03-04
Remove `with_testbed_template` argument

TimeEngineer:


touilleMan (approved):
Is there the related tests ? (i.e. test that does a save device or change auth method, then check the old auth can no longer be used)

PR #6617, opened 2024-03-06
Add web `storage`

### LINKED ISSUE #4903: Web support for CertificatesStorage
Currently [`CertificatesStorage` is only available on native](https://github.com/Scille/parsec-cloud/blob/master/libparsec/crates/platform_storage/src/native/certificates.rs)

We need it to be available on web for testing the GUI

`CertificateStorage` only store cache information (i.e. removing the database won't result in data loss).
So a first implementation could be memory only if [`IndexedDB`](https://developer.mozilla.org/en-US/docs/Web/API/IndexedDB_API) proves more complicated than expected

Depends on #4902 for tests support

TimeEngineer:
Downgrade `uuid` version because of dependencies conflict (`indexed_db_futures`)
closes #4903

### DIFF
--- a/libparsec/crates/platform_storage/src/web/certificates.rs
+++ b/libparsec/crates/platform_storage/src/web/certificates.rs
@@ -1,20 +1,52 @@
-#![allow(unused_variables)]
 // Parsec Cloud (https://parsec.cloud) Copyright (c) BUSL-1.1 2016-present Scille SAS
 
-use std::{marker::PhantomData, path::Path};
+#![allow(unused_variables)]
 
+use std::collections::HashMap;
+use std::path::Path;
+use std::sync::Arc;
+
+use indexed_db_futures::prelude::IdbTransaction;
+use indexed_db_futures::IdbDatabase;
 use libparsec_types::prelude::*;
 
 use crate::certificates::{GetCertificateError, GetCertificateQuery, PerTopicLastTimestamps, UpTo};
+use crate::web::model::{Certificate, CertificateFilter};
+use crate::web::DB_VERSION;
+
+const COMMON_CERTIFICATES: [&str; 4] = [
+    "user_certificate",
+    "device_certificate",
+    "user_update_certificate",
+    "revoked_user_certificate",
+];
+
+const SEQUESTER_CERTIFICATES: [&str; 3] = [
+    "sequester_authority_certificate",
+    "sequester_service_certificate",
+    "sequester_revoked_service_certificate",
+];
+
+const REALM_CERTIFICATES: [&str; 4] = [
+    "realm_role_certificate",
+    "realm_name_certificate",
+    "realm_key_rotation_certificate",
+    "realm_archiving_certificate",
+];
+
+const SHAMIR_RECOVERY_CERTIFICATES: [&str; 2] = [
+    "shamir_recovery_share_certificate",
+    "shamir_recovery_brief_certificate",
+];
 
 #[derive(Debug)]
 pub(crate) struct PlatformCertificatesStorageForUpdateGuard<'a> {
-    _marker: PhantomData<&'a ()>,
+    transaction: IdbTransaction<'a>,
 }
 
 impl<'a> PlatformCertificatesStorageForUpdateGuard<'a> {
     pub async fn commit(self) -> anyhow::Result<()> {
-        todo!()
+        super::db::commit(self.transaction).await
     }
 
     pub async fn get_certificate_encrypted(
@@ -22,7 +54,58 @@ impl<'a> PlatformCertificatesStorageForUpdateGuard<'a> {
         query: GetCertificateQuery,
         up_to: UpTo,
     ) -> Result<(DateTime, Vec<u8>), GetCertificateError> {
-        todo!()
+        let certifs = Certificate::get_values(
+            &self.transaction,
+            CertificateFilter {
+                certificate_type: query.certificate_type.into(),
+                filter1: query.filter1.clone(),
+                filter2: query.filter2.clone(),
+            },
+        )
+        .await?;
+
+        let maybe_certif_timestamp = certifs
+            .get(0)
+            .map(|certif| (certif.certificate_timestamp, certif.certificate.clone()));
+
+        let certifs = if let UpTo::Timestamp(up_to) = up_to {
+            certifs
+                .into_iter()
+                .filter(|certif| certif.certificate_timestamp <= up_to.get_f64_with_us_precision())
+                .collect()
+        } else {
+            certifs
+        };
+
+        if let Some(certif) = certifs.into_iter().max_by(|x, y| {
+            x.certificate_timestamp
+                .partial_cmp(&y.certificate_timestamp)
+                .expect("Timestamp should not be undefined")
+        }) {
+            return Ok((
+                DateTime::from_f64_with_us_precision(certif.certificate_timestamp),
+                certif.certificate.to_vec(),
+            ));
+        }
+
+        let up_to = if let UpTo::Timestamp(up_to) = up_to {
+            up_to
+        } else {
+            return Err(GetCertificateError::NonExisting);
+        };
+
+        // Determine if the result is an actual success or a ExistButTooRecent error
+        if let Some((certif_timestamp, certif)) = maybe_certif_timestamp {
+            let certificate_timestamp = DateTime::from_f64_with_us_precision(certif_timestamp);
+
+            if certificate_timestamp > up_to {
+                return Err(GetCertificateError::ExistButTooRecent {
+                    certificate_timestamp,
+                });
+            }
+        }
+
+        Err(GetCertificateError::NonExisting)
     }
 
     /// Certificates are returned ordered by timestamp in increasing order (i.e. oldest first)
@@ -33,11 +116,49 @@ impl<'a> PlatformCertificatesStorageForUpdateGuard<'a> {
         offset: Option<u32>,
         limit: Option<u32>,
     ) -> anyhow::Result<Vec<(DateTime, Vec<u8>)>> {
-        todo!()
+        let certifs = Certificate::get_values(
+            &self.transaction,
+            CertificateFilter {
+                certificate_type: query.certificate_type.into(),
+                filter1: query.filter1,
+                filter2: query.filter2,
+            },
+        )
+        .await?;
+
+        let mut certifs = if let UpTo::Timestamp(up_to) = up_to {
+            certifs
+                .into_iter()
+                .filter(|certif| certif.certificate_timestamp <= up_to.get_f64_with_us_precision())
+                .collect()
+        } else {
+            certifs
+        };
+
+        certifs.sort_by(|x, y| {
+            x.certificate_timestamp
+                .partial_cmp(&y.certificate_timestamp)
+                .expect("Timestamp should not be undefined")
+        });
+
+        let offset = offset.unwrap_or_default() as usize;
+        let limit = limit.unwrap_or(certifs.len() as u32) as usize;
+
+        Ok(certifs
+            .into_iter()
+            .skip(offset)
+            .take(limit)
+            .map(|x| {
+                (
+                    DateTime::from_f64_with_us_precision(x.certificate_timestamp),
+                    x.certificate.to_vec(),
+                )
+            })
+            .collect())
     }
 
     pub async fn forget_all_certificates(&mut self) -> anyhow::Result<()> {
-        todo!()
+        Certificate::clear(&self.transaction).await
     }
 
     pub async fn add_certificate(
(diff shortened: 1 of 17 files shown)


libparsec/crates/platform_storage/src/web/db.rs
```diff
@@ -0,0 +1,126 @@
...
+{
+    let tx = conn
+        .transaction_on_one(store)
+        .map_err(|e| anyhow::anyhow!("{e:?}"))?;
+
+    let store = tx
+        .object_store(store)
+        .map_err(|e| anyhow::anyhow!("{e:?}"))?;
+
+    let indexed_store = store.index(index).map_err(|e| anyhow::anyhow!("{e:?}"))?;
```
FirelightFlagboy, line 25: Would it be possible to factorize those steps shared with `get_values` ?

TimeEngineer: Yes

libparsec/crates/platform_storage/src/web/model.rs
```diff
@@ -0,0 +1,195 @@
...
+        {
+            let store = evt.db().create_object_store_with_params(
+                Certificate::STORE,
+                IdbObjectStoreParameters::default().auto_increment(true),
+            )?;
+            store.create_index(
+                Certificate::INDEX_CERTIFICATE_TYPE,
+                &IdbKeyPath::str(Certificate::INDEX_CERTIFICATE_TYPE),
+            )?;
+        }
```
FirelightFlagboy, line 179: Can you split and do a check per store ?

something like

```rust
create_certificate_store(evt);
create_remanence_store(evt);
```

libparsec/crates/platform_storage/src/web/model.rs
```diff
@@ -0,0 +1,195 @@
...
+    pub(super) async fn insert(self, conn: &IdbDatabase) -> anyhow::Result<()> {
+        super::db::insert(conn, Self::STORE, self).await
+    }
+}
+
+pub(super) async fn initialize_model_if_needed(
+    mut db_req: OpenDbRequest,
+) -> anyhow::Result<IdbDatabase> {
+    db_req.set_on_upgrade_needed(Some(|evt: &IdbVersionChangeEvent| {
+        if let None = evt.db().object_store_names().find(|n| n == Vlob::STORE) {
```
FirelightFlagboy, line 110:
```suggestion
        if evt.db().object_store_names().find(|n| n == Vlob::STORE).is_none() {
```

We don't use the value so you can directly use `.is_none()`

TimeEngineer: yes of course, I took this from the crate example :sweat_smile:

libparsec/crates/platform_storage/src/web/model.rs
TimeEngineer: I will make methods

libparsec/crates/platform_storage/src/web/user.rs
```diff
@@ -2,76 +2,122 @@
...
     pub(crate) async fn no_populate_start(
         _data_base_dir: &Path,
-        _device: &LocalDevice,
+        device: &LocalDevice,
     ) -> anyhow::Result<Self> {
-        todo!();
+        // 1) Open the database
+
+        let db_req =
+            IdbDatabase::open_u32("database_name", 1).map_err(|e| anyhow::anyhow!("{e:?}"))?;
```
FirelightFlagboy, line 33: the `"datebase_name"` is expected ? fells like a placeholder

TimeEngineer: I don't have a name for this, I'm open to suggestion :smile:

FirelightFlagboy, line 33: `1` Should likely be a constant value representing the current database version

FirelightFlagboy, line 33: You likely need to setup an handle on for the event [`IdbVersionChangeEvent`](https://docs.rs/indexed_db_futures/latest/indexed_db_futures/prelude/struct.IdbVersionChangeEvent.html)

If configured that way, you rewrite `initialize_model_if_needed` to be a migration from `v0 (uninitialized) -> v1 (initialized)`.
That would also simplify the said function as you won't need to check if the model exist.

libparsec/crates/platform_storage/src/web/db.rs
TimeEngineer: I think this is not possible without macro due to borrowed data :(

libparsec/crates/platform_storage/src/web/model.rs
TimeEngineer: Done

libparsec/crates/platform_storage/src/web/db.rs
FirelightFlagboy: The borrowed data is `store` that does not live long enough ? a macro could be used since it would be small

libparsec/crates/platform_storage/src/web/user.rs
FirelightFlagboy: `versionchange` is send when the event `upgradeneeded` is send or when db is deleted.

<https://developer.mozilla.org/en-US/docs/Web/API/IDBDatabase/versionchange_event>

 But having a handle on [`upgradeneeded`](https://developer.mozilla.org/en-US/docs/Web/API/IDBOpenDBRequest/upgradeneeded_event) seems to be what we need

libparsec/crates/platform_storage/src/web/db.rs
TimeEngineer: Yes

FirelightFlagboy: If it's too hard we can skip the factorization :sweat_smile:

libparsec/crates/platform_storage/src/web/certificates.rs
```diff
@@ -1,28 +1,121 @@
...
+        // 2) BUT ! if 1) fails we want the first result in order
+        //    to correctly return NonExisting vs ExistButTooRecent
+        //
+        // So to do that we act in three steps:
+        //  1) do the SQL query with the `up_to` filter
+        //  2) if the previous query returned nothing, do another SQL query without `up_to`
+        //  3) compare the datetime of the result with the on provided by `up_to`: if it's
+        //     too high it's a ExistButTooRecent error otherwise it's a valid result
+
+        // Step 1: do the regular SQL query
```
FirelightFlagboy, line 68: The comment state about SQL, but we are currently using IDB + manual filtering / sorting

libparsec/crates/platform_storage/src/web/certificates.rs
```diff
@@ -1,28 +1,121 @@
...
+                DateTime::from_f64_with_us_precision(certif.certificate_timestamp),
+                certif.certificate.to_vec(),
+            ));
+        }
+
+        let up_to = if let UpTo::Timestamp(up_to) = up_to {
+            up_to
+        } else {
+            return Err(GetCertificateError::NonExisting);
+        };
```
FirelightFlagboy, line 107:
```suggestion
        if !matches!(up_to, UpTo::Timestamp(_)) {
            return Err(GetCertificateError::NonExisting);
        };
```

I don't see `up_to` being used after, maybe `matches` is better here ?

libparsec/crates/platform_storage/src/web/certificates.rs
```diff
@@ -48,11 +179,121 @@ impl<'a> PlatformCertificatesStorageForUpdateGuard<'a> {
...
+                        filter2: None,
+                    },
+                )
+                .await?
+                .into_iter()
+                .map(|certif| DateTime::from_f64_with_us_precision(certif.certificate_timestamp)),
+            );
+        }
+
+        let common_last_timestamp = common_certifs.into_iter().max_by(|x, y| x.cmp(y));
```
FirelightFlagboy, line 217: Why the DateTime are stored in Vec, where we only use the greater one ?

libparsec/crates/platform_storage/src/web/certificates.rs
```diff
@@ -62,36 +303,69 @@ impl<'a> PlatformCertificatesStorageForUpdateGuard<'a> {
...
+
+        // 3) All done !
+
+        let storage = Self { conn };
+        Ok(storage)
     }
 
     pub async fn stop(self) -> anyhow::Result<()> {
-        todo!()
+        Ok(())
```
FirelightFlagboy, line 344: Why we don't call [`idb.close()`](https://developer.mozilla.org/fr/docs/Web/API/IDBDatabase/close) here ?

libparsec/crates/platform_storage/src/web/mod.rs
```diff
@@ -1,5 +1,11 @@
...
 
+use libparsec_types::VersionInt;
+
 pub mod certificates;
+mod db;
+mod model;
 pub mod user;
 pub mod workspace;
+
+const DB_VERSION: VersionInt = 1;
```
FirelightFlagboy, line 11: I don't need to need to use `VersionInt` for `DB_VERSION`, that value is internal to the web implementation.
I would suggest to keep it simple and use `u32` directly :wink:

libparsec/crates/platform_storage/src/web/workspace.rs
```diff
@@ -4,117 +4,479 @@
...
+        let storage = Self {
+            conn,
+            cache_max_blocks,
+        };
+        Ok(storage)
     }
 
     pub async fn stop(self) -> anyhow::Result<()> {
-        todo!()
+        Ok(())
```
FirelightFlagboy, line 66: Should `idb.close()` be called here ?

libparsec/crates/platform_storage/src/web/model.rs
FirelightFlagboy, line 1: Currently we only have to handle the db version change from 0 to 1, but should start actively checking the old/new version of the event to start establish a pattern ?

Maybe that should be done in a new PR WDYT ?

libparsec/crates/platform_storage/src/web/user.rs
```diff
@@ -2,76 +2,149 @@
...
+
+        Ok(Self {
+            conn,
+            realm_id: device.user_realm_id,
+        })
     }
 
     pub async fn stop(&self) -> anyhow::Result<()> {
-        todo!();
+        Ok(())
```
FirelightFlagboy, line 57: Should call `idb.close()` ?

libparsec/crates/platform_storage/src/web/certificates.rs
TimeEngineer: Sorry

libparsec/crates/platform_storage/src/web/certificates.rs
TimeEngineer: This is because I missed something:
```Rust
if certificate_timestamp > up_to {
    Err(GetCertificateError::ExistButTooRecent {
        certificate_timestamp,
    })
}
```

TimeEngineer: Corrected

libparsec/crates/platform_storage/src/web/certificates.rs
TimeEngineer: Done

libparsec/crates/platform_storage/src/web/certificates.rs
TimeEngineer: How do you prefer to find the greater one ?

libparsec/crates/platform_storage/src/web/certificates.rs
TimeEngineer: I didn't even know check, they don't use it on their example :grimacing:

TimeEngineer: Done

libparsec/crates/platform_storage/src/web/model.rs
TimeEngineer: Yes I prefer

libparsec/crates/platform_storage/src/web/mod.rs
TimeEngineer: :+1:

libparsec/crates/platform_storage/src/web/certificates.rs
FirelightFlagboy: I was think of something like:

```suggestion
        let mut common_last_timestamp = None;
        
        for ty in COMMON_CERTIFICATES {
            common_last_timestamp = Certificate::get_values(
                    &self.transaction,
                    CertificateFilter {
                        certificate_type: ty.into(),
                        filter1: None,
                        filter2: None,
                    },
                )
                .await?
                .into_iter()
                .map(|certif| DateTime::from_f64_with_us_precision(certif.certificate_timestamp))
                .chain(common_last_timestamp)
                .max_by(|x, y| x.cmp(y));
        }
```

That convert the value to `DateTime` like the current solution but it chain the previous `common_last_timestamp`, then the resulting iterator is used with `max_by`

libparsec/crates/platform_storage/src/web/certificates.rs
```diff
@@ -1,28 +1,111 @@
...
+                DateTime::from_f64_with_us_precision(certif.certificate_timestamp),
+                certif.certificate.to_vec(),
+            ));
+        }
+
+        let up_to = if let UpTo::Timestamp(up_to) = up_to {
+            up_to
+        } else {
+            return Err(GetCertificateError::NonExisting);
+        };
```
FirelightFlagboy, line 95:
```suggestion
        let UpTo::Timestamp(up_to) = up_to else {
            return Err(GetCertificateError::NonExisting);
        };
```

libparsec/crates/platform_storage/src/web/model.rs
FirelightFlagboy: I've create the issue #6959 for that

libparsec/crates/platform_storage/src/web/certificates.rs
TimeEngineer: Oh ok :+1: yes it is better

libparsec/crates/platform_storage/src/web/certificates.rs
TimeEngineer: :+1:

libparsec/crates/platform_storage/src/web/certificates.rs
```diff
@@ -48,11 +167,114 @@ impl<'a> PlatformCertificatesStorageForUpdateGuard<'a> {
...
+            .into_iter()
+            .map(|x| {
+                x.0.ok_or(anyhow::anyhow!("Missing realm id"))
+                    .and_then(|id| {
+                        VlobID::from_hex(&id)
+                            .map(|id| (id, x.1))
+                            .map_err(|e| anyhow::anyhow!(e))
+                    })
+            })
+            .collect::<Result<HashMap<_, _>, _>>()?;
```
FirelightFlagboy, line 254: Why do you need to sort the `realm_certifs` if you put them in a `HashMap` ?

TimeEngineer: Because It need the last timestamp for each realm

FirelightFlagboy: Why do you need the most recent entry to be inserted last ?

From what I know `HashMap` don't provide any guaranty about the iteration order, what I'm missing ?

TimeEngineer: The order is guaranted by the sort as you mention

FirelightFlagboy: Now I understand

> the last inserted is the most recent

You need an `Hasmap<VlobID, DateTime>` where when you have the same `VlobID` you only want to keep the most recent date.

FirelightFlagboy (approved):

PR #6620, opened 2024-03-06
Add tests for user_update

### LINKED ISSUE #6619: Server test user_update

mmmarcos:
Closes #6619

### DIFF
--- a/server/tests/api_v4/authenticated/test_user_update.py
+++ b/server/tests/api_v4/authenticated/test_user_update.py
@@ -1,6 +1,15 @@
 # Parsec Cloud (https://parsec.cloud) Copyright (c) BUSL-1.1 2016-present Scille SAS
 
-from parsec._parsec import DateTime, UserProfile, UserUpdateCertificate, authenticated_cmds
+import pytest
+
+from parsec._parsec import (
+    DateTime,
+    DeviceID,
+    RevokedUserCertificate,
+    UserProfile,
+    UserUpdateCertificate,
+    authenticated_cmds,
+)
 from parsec.events import EventUserUpdated
 from tests.common import Backend, CoolorgRpcClients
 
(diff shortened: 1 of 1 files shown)


server/tests/api_v4/authenticated/test_user_update.py
```diff
@@ -33,3 +42,212 @@ async def test_authenticated_user_update_ok(coolorg: CoolorgRpcClients, backend:
...
+            ).dump_and_sign(coolorg.alice.signing_key)
+        case "user_update_certificate_author_update_itself":
+            certif = UserUpdateCertificate(
+                author=coolorg.alice.device_id,
+                timestamp=now,
+                user_id=coolorg.alice.device_id.user_id,
+                new_profile=UserProfile.STANDARD,
+            ).dump_and_sign(coolorg.alice.signing_key)
+        case unknown:
+            assert False, unknown
```
FirelightFlagboy, line 181: Instead of a match, you could use a fixture for each action kind, and request the result with `certif = request.getfixture(kind)`.

I'm not sure if that would be better :thinking:

FirelightFlagboy, line 181: Raise an exception instead ;)

server/tests/api_v4/authenticated/test_user_update.py
```diff
@@ -33,3 +42,212 @@ async def test_authenticated_user_update_ok(coolorg: CoolorgRpcClients, backend:
...
+    kind: str,
+) -> None:
+    now = DateTime.now()
+    match kind:
+        case "same_timestamp":
+            user_update_timestamp = now
+        case "previous_timestamp":
+            user_update_timestamp = now.subtract(seconds=1)
+        case unknown:
+            assert False, unknown
```
FirelightFlagboy, line 221: You could directly put those value in the `parametrize`

mmmarcos: Can you provide a code example?

server/tests/api_v4/authenticated/test_user_update.py
mmmarcos: I'm not sure either. I think is important to be consistent with the rest of the test code base (which uses `match` on a `kind` paramter).

mmmarcos: Done

server/tests/api_v4/authenticated/test_user_update.py
FirelightFlagboy:
```python
@pytest.mark.parametrize(
  "timestamp_offset", 
  (
    pytest.param(0, id="same_timestamp"), 
    pytest.param(42, id="previous_timestamp")
  )
```

In the function you directly use

```python
user_update_timestamp = now.subtract(seconds=timestamp_offset)
```

mmmarcos: Worked like a charm. Thanks!

Issue #6566, opened 2024-02-29
Improve keyring strategy


vxgmichel:
The naive approach implemented in #6564 and #6557 identifies keyring entries using `parsec` as service and the device slug hash as user.

In order to limit the number of entries in the keyring and allow easier testing in an isolated environment, the following approach is suggested instead:

1. Add a `keyring_user` field in the `DeviceFileKeyring` structure. This makes the loading process more robust in the long run (especially if we decide to change approach overtime). Optionally, a `keyring_service` field might also be added (although the service probably won't change overtime)

2. Update the saving logic so that the `keyring_user` is taken from a dedicated config file:
  - If this file does not exist:
    - Generate a keyring user 
    - Generate a key
    - Add the key to the keyring
    - Save the keyring user to the dedicated config file
    - Save the device key file using the generated key and keyring user
  - If the config file does exist but the corresponding keyring entry does not:
    - Generate a new keyring user 
    - Generate a new key
    - Add the new key to the keyring
    - Overwrite the dedicated config file with the new keyring user
    - Save the device key file using the generated key and keyring user
  - If both the config file and the corresponding keyring entry exist:
    - Get the key from the keyring
    - Save the device key file using this key

This way, keyring entries are never deleted or modified.

3. Modify the testenv scripts to create this file with a hardcoded keyring user to avoid the creation of new keyring entries everytime an isolated environment is created

vxgmichel:
Implemented in #6625

PR #6625, opened 2024-03-06
Improve keyring strategy

TimeEngineer:
Treats #6566
Missing:
- CLI

### DIFF
--- a/libparsec/crates/platform_device_loader/src/native/mod.rs
+++ b/libparsec/crates/platform_device_loader/src/native/mod.rs
@@ -6,6 +6,7 @@ use std::{
     path::{Path, PathBuf},
     sync::Arc,
 };
+use uuid::Uuid;
 use zeroize::Zeroize;
 
 use libparsec_types::prelude::*;
@@ -16,6 +17,12 @@ use crate::{
 };
 
 const KEYRING_SERVICE: &str = "parsec";
+
+fn get_default_data_dir() -> PathBuf {
+    dirs::data_dir()
+        .expect("Could not determine base data directory")
+        .join("parsec-v3-alpha")
+}
 
 impl From<keyring::Error> for LoadDeviceError {
     fn from(value: keyring::Error) -> Self {
@@ -227,7 +234,7 @@ pub async fn load_device(
                 DeviceFile::load(&content).map_err(|_| LoadDeviceError::InvalidData)?;
 
             if let DeviceFile::Keyring(x) = device_file {
-                let entry = KeyringEntry::new(KEYRING_SERVICE, &slughash(&x.slug))?;
+                let entry = KeyringEntry::new(&x.keyring_service, &x.keyring_user)?;
 
                 let passphrase = entry.get_password()?.into();
 
@@ -288,9 +295,7 @@ pub async fn load_device(
     Ok(Arc::new(device))
 }
 
-fn save_device_file(key_file: &PathBuf, file_content: &DeviceFile) -> Result<(), SaveDeviceError> {
-    let file_content = file_content.dump();
-
+fn save_content(key_file: &PathBuf, file_content: &[u8]) -> Result<(), SaveDeviceError> {
     if let Some(parent) = key_file.parent() {
         std::fs::create_dir_all(parent).map_err(|e| SaveDeviceError::InvalidPath(e.into()))?;
     }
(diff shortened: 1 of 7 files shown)


libparsec/crates/platform_device_loader/src/native/mod.rs
```diff
@@ -322,17 +327,52 @@ fn save_device_file(key_file: &PathBuf, file_content: &DeviceFile) -> Result<(),
...
+                let keyring_user = std::fs::read_to_string(&keyring_user_path).map_err(|_| {
+                    SaveDeviceError::Internal(anyhow::anyhow!("Cannot read keyring user file"))
+                })?;
+
+                let entry = KeyringEntry::new(KEYRING_SERVICE, &keyring_user)?;
+
+                match entry.get_password() {
+                    Ok(passphrase) => (
+                        SecretKey::from_recovery_passphrase(passphrase.into())
+                            .map_err(|_| SaveDeviceError::DecryptionFailed)?,
```
vxgmichel, line 368: This is operation is not about decryption, as we're just trying to convert the passphrase into the secret key.

In this specific case, I would generate a new keyring user as well.

libparsec/crates/platform_device_loader/src/native/mod.rs
```diff
@@ -322,17 +327,52 @@ fn save_device_file(key_file: &PathBuf, file_content: &DeviceFile) -> Result<(),
...
-            let entry = KeyringEntry::new(KEYRING_SERVICE, &device.slughash())?;
-
-            let (passphrase, key) = SecretKey::generate_recovery_passphrase();
-
-            entry.set_password(&passphrase)?;
+            let keyring_user_path = get_default_config_dir().join("keyring_user.txt");
+
+            let (key, keyring_user) = if keyring_user_path.exists() {
+                let keyring_user = std::fs::read_to_string(&keyring_user_path).map_err(|_| {
+                    SaveDeviceError::Internal(anyhow::anyhow!("Cannot read keyring user file"))
```
vxgmichel, line 360: I would treat this error as for non-existing file, since there's a slight chance we might be able to overwrite this file.

libparsec/crates/platform_device_loader/src/native/mod.rs
```diff
@@ -322,17 +327,52 @@ fn save_device_file(key_file: &PathBuf, file_content: &DeviceFile) -> Result<(),
...
+                })?;
+
+                let entry = KeyringEntry::new(KEYRING_SERVICE, &keyring_user)?;
```
vxgmichel, line 363: What if the keyring user is empty? In that case `new` might fail but we still want to generate a new keyring user.

libparsec/crates/platform_device_loader/src/native/mod.rs
```diff
@@ -17,6 +18,12 @@ use crate::{
 
 const KEYRING_SERVICE: &str = "parsec";
 
+fn get_default_config_dir() -> PathBuf {
+    dirs::data_dir()
```
vxgmichel, line 22: Should that be `dirs::config_dir()`?

vxgmichel (changes requested):
A couple of suggestions, LGTM otherwise :+1:

libparsec/crates/platform_device_loader/src/native/mod.rs
TimeEngineer: No, it fails only if keyring is not available on the platform (it creates a new entry)

libparsec/crates/platform_device_loader/src/native/mod.rs
TimeEngineer: Oh yes

libparsec/crates/platform_device_loader/src/native/mod.rs
TimeEngineer: Done

libparsec/crates/platform_device_loader/src/native/mod.rs
TimeEngineer: This error can't happen since I check if the file exists before

libparsec/crates/platform_device_loader/src/native/mod.rs
TimeEngineer: Oh you mean empty string, so it will fail for MacOS, I add an additional check

libparsec/crates/platform_device_loader/src/native/mod.rs
FirelightFlagboy: The error can append ;)

https://en.wikipedia.org/wiki/Time-of-check_to_time-of-use

Better directly call `read_to_string` and handle the error in case the file doesn't exist

TimeEngineer: Oh it's even better :+1:

TimeEngineer: Done

libparsec/crates/platform_device_loader/src/native/mod.rs
```diff
@@ -322,17 +327,50 @@ fn save_device_file(key_file: &PathBuf, file_content: &DeviceFile) -> Result<(),
...
     device: &LocalDevice,
 ) -> Result<(), SaveDeviceError> {
     match access {
         DeviceAccessStrategy::Keyring { key_file } => {
-            let entry = KeyringEntry::new(KEYRING_SERVICE, &device.slughash())?;
-
-            let (passphrase, key) = SecretKey::generate_recovery_passphrase();
-
-            entry.set_password(&passphrase)?;
+            let keyring_user_path = get_default_config_dir().join("keyring_user.txt");
```
FirelightFlagboy, line 356: I'm thinking if it would be better to store the file under `data_dir` :thinking:

I don't think that something that will be configured.

FirelightFlagboy (approved):

libparsec/crates/platform_device_loader/src/native/mod.rs
vxgmichel: > I'm thinking if it would be better to store the file under data_dir 🤔

I would keep it close to the device files. In v2, device files are stored in the config directory, but I don't know about v3. What do you think @FirelightFlagboy ?

vxgmichel (approved):
LGTM :+1: 

Still we should answer those questions before merging:
- Should `keyring_service` be added to the device key file?
- Should `keyring_user.txt` be in the data or config dir?

TimeEngineer:
Why `keyring_service` must be discussed in this PR ?

libparsec/crates/platform_device_loader/src/native/mod.rs
FirelightFlagboy: Storing the device in the data dir would be nice :nerd_face:

FirelightFlagboy:
> Should `keyring_service` be added to the device key file?

I think adding a `keyring_service` entry in the device future-proof the keyring system in case we change the service name for something other than `parsec`.

TimeEngineer:
All done !

PR #6626, opened 2024-03-07
Add schedule trigger to build a nightly release a 22h30 every day

FirelightFlagboy:
I needed to update the dependent workflows to use the commit SHA from the newly created nightly commit. The schedule job was put in `releaser` workflow because Github Actions aren't trigger by actions originating from itself.

> If I have a workflow the respond to push to `master`,
> and I have another workflow that push a commit to `master` authenticated using `GH_TOKEN`.
> the first one will not be triggered by the action of second one.


FirelightFlagboy:
Currently the setup would not work as we need to authorized `github-action[bot]` permission for the branch `releases/nightly` & tag `nightly`

mmmarcos (approved):
LGTM

PR #6636, opened 2024-03-07
Add unmount-by-fusermount test in platform_mountpoint

touilleMan:


TimeEngineer (approved):
LGTM, but maybe check that when you unmount, you won't retrieve files/folders

PR #6648, opened 2024-03-08
Workspace info bindings

touilleMan:


Max-7 (approved):

bindings/electron/src/index.d.ts
```diff
@@ -230,8 +239,8 @@ export interface UserInfo {
 
 export interface WorkspaceInfo {
     id: string
-    name: string
-    selfCurrentRole: RealmRole
+    currentName: string
```
mmmarcos, line 242: The `current` prefix seems redundant: even if the name changes, it will always be the _current_ name isn't it?

PR #6666, opened 2024-03-11
Event bindings

touilleMan:
Bound events:
- `Ping`: for test onyl, just ignore it
- `Offline` & `Online`
- `ServerConfigChanged` always triggered once on first SSE connection with the server (given before the connection, the client starts with a "good enough" default server config)
- `InvitationChanged`
- `TooMuchDriftWithServerClock`
- `IncompatibleServer` 

Main event missing are `ExpiredOrganization` & `RevokedSelfUser` for errors, and `WorkspaceSelfRoleChanged` (on sharing/unsharing) & `WorkspaceEntryChanged` (when data change)

Additional missing events are the ones to inform about the synchronization progress (both for inbound and outbound sync)

### DIFF
--- a/libparsec/crates/client/src/monitors/server_config.rs
+++ b/libparsec/crates/client/src/monitors/server_config.rs
@@ -1,0 +1,62 @@
+// Parsec Cloud (https://parsec.cloud) Copyright (c) BUSL-1.1 2016-present Scille SAS
+
+use std::{future::Future, sync::Arc};
+
+use libparsec_platform_async::{pretend_future_is_send_on_web, sleep};
+
+use super::Monitor;
+use crate::{
+    event_bus::{EventBus, EventServerConfigChanged, EventServerConfigNotified},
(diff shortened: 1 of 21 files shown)


libparsec/crates/client/src/monitors/server_config.rs
```diff
@@ -0,0 +1,62 @@
...
+                event_bus.send(&EventServerConfigChanged);
+            }
+        }
+    });
+
+    async move {
+        // Our coroutine is only here to keep the event connection alive.
+        let _events_connection_lifetime = events_connection_lifetime;
+        // We want to sleep forever, but we can only do it for `Duration::MAX` (which is
+        // very, very long but not infinite). Hence the loop, which is mostly theorical.
```
TimeEngineer, line 57:
```suggestion
        // very, very long but not infinite). Hence the loop, which is mostly theoretical.
```

PR #6676, opened 2024-03-11
[fix #6535]: Add WinFSP installer to the Electron build

### LINKED ISSUE #6535: Add WinFSP to installer
To include `WinFSP` in the electron installer for Windows, we need to use a custom `nsi` script provide with [`nsis.script`](https://www.electron.build/configuration/nsis.html).

We could be inspired by what we have done previously [`installer.nsi`](https://github.com/Scille/parsec-cloud/blob/v2.17.0/packaging/windows/installer.nsi)

But we would need to setup the env beforehand (e.g.: downloading `winfsp setup.exe` that was done in [`freeze_program.py`](https://github.com/Scille/parsec-cloud/blob/v2.17.0/packaging/windows/freeze_program.py))

Fortunatly, `electron-builder` provide an API that could be used inside a `js` script: [electron-build-api](https://www.electron.build/api/electron-builder)

FirelightFlagboy:
- Use the electron-builder hook `boforePack` to download the WinFSP.
- Add a custom NSIS script to install WinFSP if it's not already installed.

### DIFF
--- a/.cspell/custom-words.txt
+++ b/.cspell/custom-words.txt
@@ -31,6 +31,11 @@ asyncpg
 athrow
 authurl
 archivings
+oname
+PLUGINSDIR
+macroend
+msiexec
+insertmacro
 archs
 autobuild
 autojump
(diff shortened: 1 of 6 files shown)


FirelightFlagboy:
https://github.com/Scille/parsec-cloud/actions/runs/8235618963

Max-7 (approved):

PR #6742, opened 2024-03-14
Harden pyRight configuration

FirelightFlagboy:
Each commit enable a new rule with it's required change


vxgmichel (commented):
There seems to be a changed of default here.

Before this PR, `match` statements didn't have to be exhaustive.

Exhaustiveness could then be enforced using:
```python
                case unknown:
                    assert_never(unknown)
```

After this PR, exhaustiveness is enforced and can by bypassed with a catch-all:

```python
             case _:
                pass
```

Did I get that right @FirelightFlagboy ?

Any opinion about that @touilleMan ?

FirelightFlagboy:
@vxgmichel `pyright` is now configured with:

- `reportMatchNotExhaustive` which require match to catch all possible case.

  That correspond to the pattern

  ```python
  match foo:
    case _:
      # code
  ```

- `reportUnnecessaryComparison` which signal when we are doing useless comparison

   For `if` branch:

   ```python
   foo = True
   if foo == False: # unnecessary
     pass
   ```

   For `match`:

   ```python
   foo: bool
   match foo:
     case True:
        pass
     case False:
        pass
     case _: # impossible, all case are tested
        pass
   ```

vxgmichel (approved):

PR #6743, opened 2024-03-14
Sunset mypy

FirelightFlagboy:
We have remplaced `mypy` by `pyRight` (or `pylance` in VSCode) for type checking. The `CI` was already checking for `pyRight` error only.


server/parsec/__main__.py
```diff
@@ -21,9 +21,9 @@ def patched_cli_main(args: Sequence[str] | None = None, **kwargs: Any) -> Any:
     return vanilla_cli_main(args=args_list[:], **kwargs)
 
 
-# Mypy: here we want to use the patched main instead of the provided one by `BaseCommand`
+# [pyright] here we want to use the patched main instead of the provided one by `BaseCommand`
```
mmmarcos, line 24:
```suggestion
# [pyright] here we want to use the patched main instead of the one provided by `BaseCommand`
```

server/parsec/cli/__init__.py
```diff
@@ -68,6 +68,6 @@ def patched_cli_main(args: Sequence[str] | None = None, **kwargs: Any) -> Any:
     return vanilla_cli_main(args=args_list[:], **kwargs)
 
 
-# Mypy: here we want to use the patched main instead of the provided one by `BaseCommand`
+# [pyright] here we want to use the patched main instead of the provided one by `BaseCommand`
```
mmmarcos, line 71:
```suggestion
# [pyright] here we want to use the patched main instead of the one provided by `BaseCommand`
```

mmmarcos (approved):
LGTM

PR #6811, opened 2024-03-17
[MS] Remove file links handling

Max-7:
- [X] Describe your changes

### DIFF
--- a/client/src/main.ts
+++ b/client/src/main.ts
@@ -27,7 +27,7 @@ import { Config, StorageManager, StorageManagerKey } from '@/services/storageMan
 import { isPlatform } from '@ionic/vue';
 
 /* Theme variables */
-import { Validity, claimLinkValidator, fileLinkValidator } from '@/common/validators';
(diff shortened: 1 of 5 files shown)


client/src/main.ts
```diff
@@ -150,8 +150,9 @@ async function setupApp(): Promise<void> {
     window.electronAPI.receive('open-link', async (link: string) => {
       if ((await claimLinkValidator(link)).validity === Validity.Valid) {
         await handleJoinLink(link);
-      } else if ((await fileLinkValidator(link)).validity === Validity.Valid) {
-        await handleFileLink(link, informationManager, t);
+        // Commented until we can handle file links
```
FirelightFlagboy, line 153:
```suggestion
        // FIXME: Commented until we can handle file links
```
For that type of comment, I recommend using the prefix `FIXME:` or `TODO:` some extension / IDE highlight those comment and we can easily search for them using `grep`.

If you can also add a link to and issue that fix the link issue, that would be greeter

PR #6852, opened 2024-03-19
Replace usage of `sha1` by `sha256`

FirelightFlagboy:


libparsec/crates/crypto/src/rustcrypto/sequester.rs
```diff
@@ -150,8 +149,7 @@ impl SequesterPublicKeyDer {
     //   <algorithm name>:<encrypted secret key with RSA key><encrypted data with secret key>
     pub fn encrypt(&self, data: &[u8]) -> Vec<u8> {
         let mut rng = rand_08::thread_rng();
-        // No choice but to use SHA1 here: this is the default in PKCS#1 OAEP standard
```
FirelightFlagboy, line 153: From what I read `PKCS#1` don't limit to using `sha1`

- https://crypto.stackexchange.com/questions/70443/oaep-security-pkcs1-v1-5-vs-sha1-vs-sha256
- https://www.inf.pucrs.br/%7Ecalazans/graduate/TPVLSI_I/RSA-oaep_spec.pdf
- https://www.cryptsoft.com/pkcs11doc/v220/group__SEC__12__1__8__PKCS____1__RSA__OAEP.html
- https://datatracker.ietf.org/doc/html/rfc8017#section-7.1

FirelightFlagboy: Yes `RSA/OAEP` can be configured with another hash algorithm than the default SHA1

FirelightFlagboy:
Windows don't provide documented way to identify a certificate with a SHA256, only with SHA1 ([CERT_SHA1_HASH_PROP_ID](https://learn.microsoft.com/en-us/windows/win32/api/wincrypt/nf-wincrypt-certgetcertificatecontextproperty#cert_sha1_hash_prop_id), tough some arbitrary value provide the SHA256 blob).

So this isn't something that can be done currently

PR #6876, opened 2024-03-20
Use ubuntu 22.04 instead of 20.04 for the client package workflow

FirelightFlagboy:
The more recent version use libssl 3.0 by default instead of 1.1.


mmmarcos (approved):
:raised_hands:

PR #6884, opened 2024-03-20
Extract certificate used in test to specific files

FirelightFlagboy:
Having the certificate in files formats allow to use external tools to work on them (e.g.: openssl), that simplify updating the certs as we just have a single command to use without having a `copy/paste` involved.


touilleMan (commented):
I'd rather not merge this:
- I don't see the gain of having a file with the key given it is not used as is elsewhere
- have the keys in a file means `sequester.rs` no longer has the sole ownership of those keys (i.e. they could be used/modified by something else), hence it's weird to have comments about how the keys got generated
- similarly, modifying the key become more tricky given we now have to make sure it is not used by something else

The only pros is `sequester.rs` becomes 150 lines smaller, which I think is not worth it

PR #6885, opened 2024-03-20
Explicitly configure sequester to use SHA1 for RSA OAEP

FirelightFlagboy:
Rework the code to directly use `PKey` instead of `Rsa` for the internal data.

Using `PKey` allow to configure a `{De,En}crypter` with the given key.
Then we can configure explicitly the padding using in the new instance


touilleMan:
superseded by #6886

PR #6910, opened 2024-03-21
Snap rollback

FirelightFlagboy:


FirelightFlagboy:
The rollback wasn't enougth to make it work on @tleblond laptop, closing

PR #6924, opened 2024-03-23
Keep native node shared library out of the asar archive on electron packaging

touilleMan:


FirelightFlagboy (approved):

touilleMan:
@FirelightFlagboy Can you assume the merge of this PR ?

I've tested this on tweak on Linux without issues, but I didn't run it on Windows (or even in Linux with snap)

I guess we can just merge it and see if it works in the next nightly, or manually build & test on it Windows/snap before that. In any way I'd rather leave you with the choice given I'm not involve with the nightly release cycle ^^

FirelightFlagboy:
That PR won't solve the issue with `libarsec/index.node` using system lib instead of the snap provided one:

```shell
$ ldd /snap/parsec-v3/current/app/resources/app.asar.unpacked/build/src/libparsec/index.node 
	linux-vdso.so.1 (0x00007ffeb9b9e000)
	libfuse3.so.3 => /usr/lib/libfuse3.so.3 (0x00007eed46d6d000)
	libssl.so.3 => /usr/lib/libssl.so.3 (0x00007eed46c8d000)
	libcrypto.so.3 => /usr/lib/libcrypto.so.3 (0x00007eed46600000)
	libgcc_s.so.1 => /usr/lib/libgcc_s.so.1 (0x00007eed46c68000)
	libm.so.6 => /usr/lib/libm.so.6 (0x00007eed46b7c000)
	libc.so.6 => /usr/lib/libc.so.6 (0x00007eed4641e000)
	/usr/lib64/ld-linux-x86-64.so.2 (0x00007eed47ebc000)
```

> Same result if I use the wrapper script provided by `electron-builder`

But that simplify a lot the patching done in #6923.

For windows, I don't think that would cause a problem or at least it would be easy to spot and make an exception for it in the script `package.js`

PR #6948, opened 2024-03-26
Fix pyright error when style-only is true

FirelightFlagboy:
pyrigth would fail on the CI because it cannot find the module `parsec._parsec`. That module is the rust python extension, so the try to restore it from the cache else we fail.


Max-7 (dismissed):
Failed rebase?

FirelightFlagboy:
No, just wrong base ;)

FirelightFlagboy:
Superseded by #6947

PR #6949, opened 2024-03-26
Fix electron-snap skipped because `version` is ignored on release

FirelightFlagboy:
Forgot to use the same condition has the other jobs

### DIFF
--- a/.github/workflows/package-client.yml
+++ b/.github/workflows/package-client.yml
@@ -5,6 +5,7 @@ on:
     paths:
       - .github/workflows/package-client.yml
       - ./client/electron/package.js
+      - ./client/electron/snap
   workflow_call:
     inputs:
       version:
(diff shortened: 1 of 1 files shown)


.github/workflows/package-client.yml
```diff
@@ -110,6 +111,7 @@ jobs:
   electron-snap:
     needs: version
     runs-on: ubuntu-22.04
+    if: ${{ inputs.version_patch_run_id != '' && always() || success() }}
```
touilleMan, line 114: Can you add a comment to explain the reason for always and success (I guess it's to always run when version patch is set, otherwise to run if the build is good, but why is that so ? Why setting a custom version bypass the success check ? And what success is referred to here ? )

I encountered this kind of if in other places in the codebase but I couldn't understand it

Similarly, I try to debug the reason the build was skip but had absolutely no clue for the reason, my guess is there is some condition `electeon-snap` implicitly depends on...
If that's the case we'd ideally rather replace this by something explicit, or if that's not possible at least document this (this should be done in a separate PR, I guess there is multiple jobs we want to document with this)

touilleMan (approved):

PR #6950, opened 2024-03-26
Use version from package.json in the client app

FirelightFlagboy:


client/src/common/mocks.ts
```diff
@@ -4,9 +4,7 @@
 
 import { DateTime } from 'luxon';
 
-export function getAppVersion(): string {
-  return '3.0.0a';
-}
+export const APP_VERSION = __APP_VERSION__;
```
Max-7, line 7: Is there a reason why you didn't keep the function and make it return `__APP_VERSION__`? It keeps changes to a minimum and is slightly more maintenable.

Max-7 (changes requested):

client/src/common/mocks.ts
FirelightFlagboy: Could APP_VERSION not be a constant value ?

Max-7: It could, I just don't see what it brings to the table.
My reasoning is that you made a change that impact other files (you changed an API) when there was no need to do so: it creates more changes than what's stricly necessary.

I'm being pedantic, I'll approve the PR, I just wanted to know if there was a specific reason to change this.

Max-7 (approved):

Issue #6862, opened 2024-03-19
Fix incorrect use of translations in the app


Max-7:
A translated string should never be passed as a prop to a component, it should always be the translation key. Otherwise the component cannot update properly.

Replace
```vue
// Component MyComp
<template>
  {{ s }}
</template>
<script setup lang="ts">
const props = defineProps<{
  s: string;
}>();
</script>

// Use this component
<my-comp
  s="$t('translationKey')"
>
```
by
```vue
// Component MyComp
<template>
  {{ $t(s) }}
</template>
<script setup lang="ts">
const props = defineProps<{
  s: string;
}>();
</script>

// Use this component
<my-comp
  s="'translationKey'"
>
```

PR #6958, opened 2024-03-26
[MS] Fix incorrect use of translations in the app

### LINKED ISSUE #6786: [MS] When english is selected locale, theme names are in french
image

Ironicbay:
Closes #6786 #6862

### DIFF
--- a/client/src/common/date.ts
+++ b/client/src/common/date.ts
@@ -1,6 +1,6 @@
 // Parsec Cloud (https://parsec.cloud) Copyright (c) BUSL-1.1 2016-present Scille SAS
 
-import { formatDate, translate } from '@/services/translation';
+import { formatDate, msTranslate } from '@/services/translation';
 import { DateTime } from 'luxon';
 
 export function formatTimeSince(
@@ -19,17 +19,17 @@ export function formatTimeSince(
   if (!diff || (diff.years && diff.years > 0) || (diff.months && diff.months > 0) || (diff.days && diff.days > 6)) {
     return formatDate(date, format);
   } else if (roundDays) {
-    return translate('common.date.lastLoginDays', { days: diff.days || 0 }, diff.days || 0);
+    return msTranslate({ key: 'common.date.lastLoginDays', data: { days: diff.days || 0 }, count: diff.days || 0 });
   } else if (diff.days && diff.days > 0) {
-    return translate('common.date.lastLoginDays', { days: diff.days }, diff.days);
+    return msTranslate({ key: 'common.date.lastLoginDays', data: { days: diff.days }, count: diff.days });
   } else if (diff.hours && diff.hours > 0) {
-    return translate('common.date.lastLoginHours', { hours: diff.hours }, diff.hours);
+    return msTranslate({ key: 'common.date.lastLoginHours', data: { hours: diff.hours }, count: diff.hours });
   } else if (diff.minutes && diff.minutes > 0) {
-    return translate('common.date.lastLoginMinutes', { minutes: diff.minutes }, diff.minutes);
+    return msTranslate({ key: 'common.date.lastLoginMinutes', data: { minutes: diff.minutes }, count: diff.minutes });
   } else if (diff.seconds && diff.seconds > 30) {
-    return translate('common.date.lessThanAMinute');
+    return msTranslate('common.date.lessThanAMinute');
   } else {
-    return translate('common.date.fewSeconds');
+    return msTranslate('common.date.fewSeconds');
   }
 }
 
--- a/client/src/common/date.ts
+++ b/client/src/common/date.ts
@@ -1,6 +1,6 @@
 // Parsec Cloud (https://parsec.cloud) Copyright (c) BUSL-1.1 2016-present Scille SAS
 
-import { formatDate, msTranslate } from '@/services/translation';
+import { formatDate, Translatable } from '@/services/translation';
 import { DateTime } from 'luxon';
 
 export function formatTimeSince(
@@ -8,7 +8,7 @@ export function formatTimeSince(
   defaultValue = '',
   format: 'long' | 'short' = 'long',
   roundDays = false,
-): string {
+): Translatable {
   if (!date) {
     return defaultValue;
   }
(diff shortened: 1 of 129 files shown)


client/src/components/core/ms-dropdown/MsDropdown.vue
```diff
@@ -22,7 +22,7 @@
         slot="end"
         :icon="getIcon()"
       />
-      <span class="input-text">{{ labelRef }}</span>
+      <span class="input-text">{{ $t(labelRef || '') }}</span>
```
Max-7, line 25: Empty keys create warnings
`[intlify] Not found '' key in 'fr-FR' locale messages.`

client/src/views/devices/GreetDeviceModal.vue
```diff
@@ -311,9 +311,9 @@ async function updateCanGoForward(): Promise<void> {
 
 function getNextButtonText(): string {
   if (pageStep.value === GreetDeviceStep.WaitForGuest) {
-    return translate('DevicesPage.greet.actions.start');
+    return 'DevicesPage.greet.actions.start';
   } else if (pageStep.value === GreetDeviceStep.Summary) {
-    return translate('DevicesPage.greet.actions.finish');
+    return 'DevicesPage.greet.actions.finish';
   }
   return '';
```
Max-7, line 318: Same here, careful about empty strings being used as key.

client/src/views/header/HeaderPage.vue
```diff
@@ -267,21 +267,21 @@ onUnmounted(async () => {
...
-      return translate('HeaderPage.titles.about');
+      return 'HeaderPage.titles.about';
     case Routes.ContactDetails:
-      return translate('HeaderPage.titles.myContactDetails');
+      return 'HeaderPage.titles.myContactDetails';
     case Routes.RecoveryExport:
-      return translate('HeaderPage.titles.recoveryExport');
+      return 'HeaderPage.titles.recoveryExport';
     case null:
       return '';
```
Max-7, line 286: Same here.

client/src/views/home/CreateOrganizationModal.vue
```diff
@@ -28,13 +28,13 @@
           v-if="titles.get(pageStep)?.title !== ''"
           class="modal-header__title title-h2"
         >
-          {{ titles.get(pageStep)?.title }}
+          {{ $t(titles.get(pageStep)!.title) }}
```
Max-7, line 31: Probably will lead to problems if you try to give `undefined` to `$t()`

client/src/views/home/CreateOrganizationModal.vue
```diff
@@ -28,13 +28,13 @@
...
         </ion-title>
         <ion-text
           v-if="titles.get(pageStep)?.subtitle !== ''"
           class="modal-header__text body"
         >
-          {{ titles.get(pageStep)?.subtitle }}
+          {{ $t(titles.get(pageStep)!.subtitle || '') }}
```
Max-7, line 37: Same thing, no empty strings.

Max-7 (changes requested):

client/src/views/files/FoldersPage.vue
```diff
@@ -95,23 +95,23 @@
...
+                  count: folders.entriesCount() + files.entriesCount(),
+                })
               }}
             </ion-text>
             <ion-text
               class="body item-selected"
               v-if="selectedFilesCount > 0"
             >
-              {{ $t('FoldersPage.itemSelectedCount', { count: selectedFilesCount }, selectedFilesCount) }}
+              {{ $msTranslate({ key: 'FoldersPage.itemSelectedCount', data: { count: selectedFilesCount }, count: selectedFilesCount }) }}
```
NicoTuxx, line 109: why selectedFilesCount is present 2 times?
Is count really useful? Is data not enough?

client/src/services/translation.ts
```diff
@@ -2,12 +2,30 @@
...
 import { DateTime } from 'luxon';
+import { App } from 'vue';
 import { createI18n } from 'vue-i18n';
 
 let i18n: any | null = null;
 
+export interface TranslationData {
+  key: string;
+  data?: object;
+  count?: number;
```
NicoTuxx, line 15: what is `count`? Could it not be passed in data as any other named parameter?

NicoTuxx (changes requested):

client/src/services/translation.ts
```diff
@@ -66,10 +84,14 @@ export function initTranslations(locale?: Locale): any {
   return i18n;
 }
 
-export function translate(key: string, attrs?: object, count?: number): string {
+export function msTranslate(content: Translatable): string {
```
Max-7, line 87: You can let it be named `translate()`, it shouldn't be exported anymore, everything should be done either with `$msTranslate` (most of the time), or if it can't be done in the template, with inject:
```typescript
const msTranslate = inject('msTranslate');

msTranslate('a.b.c');
```

client/src/services/translation.ts
```diff
@@ -2,12 +2,30 @@
...
+}
+
+export type Translatable = string | TranslationData | undefined;
+
+export const TranslationPlugin = {
+  install: (app: App<any>): void => {
+    app.config.globalProperties.$msTranslate = (translatable: Translatable): string => {
+      return msTranslate(translatable);
+    };
+    app.provide('msTranslate', msTranslate);
```
Max-7, line 25: Better to define a key that can be used when injecting later rather than a raw string.

client/src/services/translation.ts
```diff
@@ -2,12 +2,30 @@
...
 
 let i18n: any | null = null;
 
+export interface TranslationData {
+  key: string;
+  data?: object;
+  count?: number;
+}
+
+export type Translatable = string | TranslationData | undefined;
```
Max-7, line 18: `Translatable` shouldn't include `undefined`.

client/src/views/about/ChangesModal.vue
```diff
@@ -38,7 +38,7 @@
                   {{ change.description }}
                 </ion-text>
                 <a
-                  :href="$t('app.projectSources') + '/issues/' + change.issue"
+                  :href="$msTranslate('app.projectSources') + '/issues/' + change.issue"
```
Max-7, line 41: Didn't know this was a thing. This should have it's own key in the translation files,
`$msTranslate('app.projectIssues', { issueId: change.issue });`

Granted, it's not currently used.

client/src/views/about/ChangesModal.vue
```diff
@@ -68,7 +68,7 @@
...
```
Max-7, line 71: Same here, should have its own string.

NicoTuxx (approved):

client/src/common/date.ts
```diff
@@ -19,17 +19,17 @@ export function formatTimeSince(
   if (!diff || (diff.years && diff.years > 0) || (diff.months && diff.months > 0) || (diff.days && diff.days > 6)) {
     return formatDate(date, format);
   } else if (roundDays) {
-    return translate('common.date.lastLoginDays', { days: diff.days || 0 }, diff.days || 0);
+    return msTranslate({ key: 'common.date.lastLoginDays', data: { days: diff.days || 0 }, count: diff.days || 0 });
```
Max-7, line 22: All the functions in `common/` should not call `msTranslate()`, but instead return a `Translatable`.

client/src/common/file.ts
```diff
@@ -78,11 +78,11 @@ function size(bytes: number, system: [number, string][]): string {
 
 export function formatFileSize(bytesize: number): string {
   const SYSTEM: [number, string][] = [
-    [Math.pow(1024, 0), translate('common.filesize.bytes')],
-    [Math.pow(1024, 1), translate('common.filesize.kilobytes')],
-    [Math.pow(1024, 2), translate('common.filesize.megabytes')],
-    [Math.pow(1024, 3), translate('common.filesize.gigabytes')],
-    [Math.pow(1024, 4), translate('common.filesize.terabytes')],
+    [Math.pow(1024, 0), msTranslate('common.filesize.bytes')],
```
Max-7, line 81: Should be `const SYSTEM: [number, Translatable][]` instead, we're not translating directly.

client/src/common/validators.ts
```diff
@@ -34,7 +34,7 @@ export const emailValidator: IValidator = async function (value: string) {
   }
   return (await isValidEmail(value))
     ? { validity: Validity.Valid }
-    : { validity: Validity.Invalid, reason: translate('validators.userInfo.email') };
+    : { validity: Validity.Invalid, reason: msTranslate('validators.userInfo.email') };
```
Max-7, line 37: Same here, reason should be a `Translatable`.

Max-7 (changes requested):
The whole idea behind re-doing the translation is that all the texts that are displayed to the user should be `Translatable` and not `string`, and that at no point we should have to use either `msTranslate` or `inject('msTranslate')` and that the translation should only be done at the last moment, in the template.

Everytime a translation is not done in the template, there should be a justification for it.

client/src/vite-env.d.ts
```diff
@@ -8,6 +8,9 @@ declare module '*.vue' {
   // eslint-disable-next-line @typescript-eslint/ban-types
   const component: DefineComponent<{}, {}, any>;
   export default component;
+  interface ComponentCustomProperties {
+    $tr: { tr: (Translatable) => string };
```
Ironicbay, line 12: `$tr` or `$msTranslate` ?

NicoTuxx (approved):

PR #6962, opened 2024-03-27
[MS] Fix invitation button wrongly remaining in multi org

### LINKED ISSUE #6788: Multi-org: Pending invitations button is present if invitations are pending on other org but not current one
image

Ironicbay:
Closes #6788

### DIFF
--- a/client/src/components/header/InvitationsButton.vue
+++ b/client/src/components/header/InvitationsButton.vue
@@ -18,7 +18,7 @@
 <script setup lang="ts">
 import { Answer, MsModalResult, askQuestion } from '@/components/core';
 import { ClientCancelInvitationErrorTag, UserInvitation, cancelInvitation, listUserInvitations } from '@/parsec';
-import { Routes, navigateTo } from '@/router';
+import { Routes, navigateTo, watchOrganizationSwitch } from '@/router';
 import { EventData, EventDistributor, EventDistributorKey, Events } from '@/services/eventDistributor';
 import { Information, InformationLevel, InformationManager, InformationManagerKey, PresentationMode } from '@/services/informationManager';
 import { translate } from '@/services/translation';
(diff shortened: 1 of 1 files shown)


client/src/views/header/HeaderPage.vue
```diff
@@ -60,7 +60,7 @@
           class="topbar-right"
         >
           <div class="topbar-right-button">
-            <invitations-button />
+            <invitations-button :key="$route.fullPath" />
```
Max-7, line 63: It is quite a smart solution that I completly forgot existed and we could probably make use of it in other instances. In this specific instance though, I have two problems with it:
- We don't want to use `$route` or anything injected from vue-router. We have our own router management that builds upon vue-router to handle multi-org, I'm afraid we may encounted some problem if we use both interchangeably. In cases where you need the route, `getCurrentRoute()` in `@/router` should work for that. In this specific case, you could probably even write `:key="organizationKey"` where `import { organizationKey } from '@/router`, which means it would not be updated everytime the route changes, but only when the connection handle changes (which would be super sexy if not for my second point).
- I don't want to rely on the parent component having to tell the child component to update. It's legitimate in some cases, when the state of the parent changes and we want the state of the children to follow, but in this case, a global state changes and the child should be aware of it by itself. In other words, if we use the `invitations-button` in some other view, I don't want the other view to have to worry about the button not updating.

I would instead use `watchOrganizationSwitch` from `@/router` in the `InvitationsButton` component to update. Don't forget to cancel the watch too.
I do agree that it's less sexy and requires a bit more work but it would make the component update by itself instead of relying on the parent setting a specific prop.

I really do like your solution though and in other contexts it may prove very useful.

Max-7 (changes requested):

Max-7 (approved):

PR #6963, opened 2024-03-27
Add test for missed events in SSE

FirelightFlagboy:


server/tests/test_sse.py
```diff
@@ -32,3 +35,50 @@ async def test_events_listen_auth_then_not_allowed(
 # TODO: - test close on user revoked
 # TODO: - test close on backpressure (too many events pilling up)
 # TODO: - test bad accept type
+
+
+async def test_missed_events(minimalorg: MinimalorgRpcClients, backend: Backend) -> None:
+    from parsec._parsec import ActiveUsersLimit, authenticated_cmds
```
touilleMan, line 41: Move import at the top of the file ?

server/tests/test_sse.py
```diff
@@ -32,3 +35,50 @@ async def test_events_listen_auth_then_not_allowed(
...
+
+    backend.config.sse_events_cache_size = 2
+    backend.events._last_events_cache = deque(maxlen=backend.config.sse_events_cache_size)  # pyright: ignore[reportPrivateUsage]
```
touilleMan, line 44: This seems rather fragile, but I guess we don't have much choice (the alternative would be to avoid using `minimalorg` & `backend`, but this add too much complexity to the test...)

server/tests/test_sse.py
```diff
@@ -32,3 +35,50 @@ async def test_events_listen_auth_then_not_allowed(
...
+
+    first_event = EventPinged(
+        organization_id=minimalorg.organization_id,
+        ping="event0",
+    )
+    backend.event_bus._dispatch_incoming_event(first_event)  # pyright: ignore[reportPrivateUsage]
```
touilleMan, line 50: would be better to wrap this use of internal:
```
dispatch_event_with_no_delay = end.event_bus._dispatch_incoming_event
```

the benefit is twofold:
- we only have the pyright ignore in a single place
- we can add a comment explaining we need to use this internal to get the guarantee the event has been processed once the function returns, which is not the case with the regular `EventBus.send` where the event goes through the LISTEN/NOTIFY dispatch system of Postgres.

server/tests/test_sse.py
FirelightFlagboy: done

server/tests/test_sse.py
FirelightFlagboy: done

touilleMan (approved):

PR #6982, opened 2024-03-28
A PoC demonstrating the use of keepalive messages in the msgpack stream

vxgmichel:
This allows blocking HTTP requests to stay alive.


vxgmichel:
Outdated

PR #6984, opened 2024-03-29
[fix #6504]: Rework `InvitationToken` & `BootstrapToken` to not use UuidV4

### LINKED ISSUE #6504: Don't use UUIDv4 for token
`BootstrapToken` and `InvitationToken` are implemented on top of UUIDv4:
https://github.com/Scille/parsec-cloud/blob/efc9505e10e8ff882d365d97cb255060cf9132db/libparsec/crates/types/src/id.rs#L213-L214

However according to [UUID's RFC](https://www.rfc-editor.org/rfc/rfc4122)

> [6](https://www.rfc-editor.org/rfc/rfc4122#section-6).  Security Considerations
>
>   Do not assume that UUIDs are hard to guess; they should not be used
>   as security capabilities (identifiers whose mere possession grants
>   access), for example.  A predictable random number source will
>   exacerbate the situation.

FirelightFlagboy:


libparsec/crates/types/src/token.rs
```diff
@@ -0,0 +1,152 @@
...
+/// This is used to ensure that the tokens are not too small during deserialization.
+const TOKEN_MINIMAL_SIZE: usize = 16;
+
+#[derive(thiserror::Error, Debug)]
+pub enum TokenDecodeError {
+    #[error("The token is not a valid hex string")]
+    InvalidHex,
+    #[error(
+        "The token is too small, got {got} bytes, expected at least {} bytes",
+        TOKEN_MINIMAL_SIZE
```
TimeEngineer, line 15:
```suggestion
        "The token is too small, got {got} bytes, expected at least {TOKEN_MINIMAL_SIZE} bytes"
```

libparsec/crates/types/tests/unit/token.rs
```diff
@@ -0,0 +1,39 @@
...
+            use super::*;
+
+            #[test]
+            fn test_default_random() {
+                let token = $name::default();
+                assert_eq!(token.as_bytes().len(), TOKEN_SIZE);
+            }
+
+            #[test]
+            fn test_eq() {
```
TimeEngineer, line 17: can we test `ne` ?

cli/src/cancel_invitation.rs
```diff
@@ -19,7 +19,7 @@ pub struct CancelInvitation {
     #[arg(short, long)]
     device: Option<String>,
     /// Invitation token
-    #[arg(short, long, value_parser = InvitationToken::from_hex)]
+    #[arg(short, long, value_parser = crate::clap_utils::token_from_hex::<InvitationToken>)]
```
touilleMan, line 22: Implement a `InvitationToken::from_hex` instead of relying on the `FromHex` trait:
- this avoid the `token_from_hex` helper
- this also avoid having to expose `FromHex` trait in `libparsec_types::prelude`

libparsec/crates/types/tests/unit/token.rs
```diff
@@ -0,0 +1,39 @@
+// Parsec Cloud (https://parsec.cloud) Copyright (c) BUSL-1.1 2016-present Scille SAS
+
+use super::*;
+
+macro_rules! base_token_tests {
```
touilleMan, line 5: :+1:

libparsec/crates/types/src/token.rs
```diff
@@ -0,0 +1,152 @@
+// Parsec Cloud (https://parsec.cloud) Copyright (c) BUSL-1.1 2016-present Scille SAS
+
+/// The size of the generated tokens.
+const TOKEN_SIZE: usize = 16;
+/// The minimal size of the generated tokens.
+/// This is used to ensure that the tokens are not too small during deserialization.
+const TOKEN_MINIMAL_SIZE: usize = 16;
```
touilleMan, line 7: Do we need this ?
I would say the token must be of a given size, so if the input data are longer than `TOKEN_SIZE` we just consider it as invalid

libparsec/crates/types/src/token.rs
```diff
@@ -0,0 +1,152 @@
...
+
+#[derive(thiserror::Error, Debug)]
+pub enum TokenDecodeError {
+    #[error("The token is not a valid hex string")]
+    InvalidHex,
+    #[error(
```
touilleMan, line 13: I we remove `TOKEN_MINIMAL_SIZE`, this error should be renamed `InvalidSize`

cli/src/cancel_invitation.rs
FirelightFlagboy: done

libparsec/crates/types/src/token.rs
FirelightFlagboy: The discussion will continue in https://github.com/Scille/parsec-cloud/pull/6984#discussion_r1547370876

libparsec/crates/types/src/token.rs
FirelightFlagboy: The use case was if we increase the `TOKEN_SIZE` later down the road, the previous generated token are still valid if they're `>= TOKEN_MINIMAL_SIZE`.

If it's OK to render the previous tokens invalid when we change `TOKEN_SIZE` then we don't need that.

FirelightFlagboy: I've changed the code to not use `TOKEN_MINIMAL_SIZE`, now the code will return `InvalidSize` if the provided hex string isn't `TOKEN_SIZE`.

TimeEngineer (approved):

PR #7016, opened 2024-04-05
Add mountpoint for macos

### LINKED ISSUE #6144: Implement mountpoint for macOS
Parsec packages are already available for macOS (although not yet signed). The only thing missing seems to be file system integration (mountpoints).

Implementation should probably be based on `macFUSE` because already used for V2, although it may be useful to do some research about other APIs (see comments below).

TimeEngineer:
closes #6144

This is a start on the implementation of `mountpoint` for `MacOS`, it should be possible to use `fuser` for `macOS`, it uses `macfuse`(`osxfuse` is the old name).
`fuser` clearly shows that it supports `macOS` but it is untested (because you can find `#[cfg(target_os = "macos")]` in its code and see: https://github.com/cberner/fuser?tab=readme-ov-file#macos-untested where it describes how to install `macfuse`)

What I have done in this PR:
- Shorcut the `CI` to test `macOS platform` only
- Printed where the problem comes from (It's when we call `Session::new`)
- When I installed `macfuse`, `pkg-config --list-all` returns `fuse` in its list, so the reason is not that `fuser` can't find the library: you can see in `build.rs`: https://github.com/cberner/fuser/blob/master/build.rs#L11 it searches for `fuse` (it could be `fuse4x` according to https://github.com/fusepy/fusepy/blob/master/fuse.py#L90-L91 which is missing in the `build.rs`)
- According to an issue on `fuser`, the problem comes from an infinite loop when you `mount`: https://github.com/cberner/fuser/issues/211 this is also how it appears on the `CI`: the test `timeout`
- I have tried to tweak `fuser`'s version, because `fuse-rs` which was the old repository of `fuser` seemed to maintain `MacOS` (see: https://github.com/zargony/fuse-rs?tab=readme-ov-file#macos) I have also found a crate which use an old version (which is not compatible with our version of `fuser` for `linux`): https://github.com/mgree/ffs/blob/main/Cargo.toml#L30
- I tried to debug on the CI, but it seems too hard to find how the bug comes from, so I asked to debug on MacOS device.

Remote machine: (with `sw_vers`)                   
ProductName:            macOS
ProductVersion:         14.4.1
BuildVersion:           23E224

- Note that the CI uses macos-12

- I installed `macfuse` with `brew install --cask macfuse`
- When I run the example `hello.rs` of `fuser`, it throws the following error message: `mount_macfuse: the file system is not available (1)` with: `fuse_mount_compat25`

- I tried `/usr/bin/sudo /usr/bin/kmutil load -p /Library/Filesystems/macfuse.fs/Contents/Extensions/14/macfuse.kext` on the remote machine which throw `Error Domain=KMErrorDomain Code=27 "Extension with identifiers io.macfuse.filesystems.macfuse not approved to load. Please approve using System Preferences." UserInfo={NSLocalizedDescription=Extension with identifiers io.macfuse.filesystems.macfuse not approved to load. Please approve using System Preferences.}`
- I found https://github.com/osxfuse/osxfuse/issues/1000 which is a similar issue, but I don't know how to access gui interface from terminal :(

Utils: https://github.com/macfuse/macfuse/wiki/Getting-Started

### DIFF
--- a/.github/workflows/ci-rust.yml
+++ b/.github/workflows/ci-rust.yml
@@ -49,135 +49,135 @@ permissions:
 jobs:
   # Cannot factorize the jobs with a matrix since we use a service container that is
   # only available on linux (see https://github.com/orgs/community/discussions/25578)
-  test-rust-linux:
-    name: "🐧 Linux: 🦀 Rust tests"
-    # Just a fail-safe timeout, see the fine grain per-task timeout instead
-    timeout-minutes: 30
-    runs-on: ubuntu-20.04
-    # Testbed server comes as a Docker image, so it will eventually goes out of sync
-    # with the tests (typically a new API is introduced in the Parsec server, or a new
-    # testbed template is introduced).
-    # In such case, the container url should be updated from the, see:
-    # https://github.com/Scille/parsec-cloud/pkgs/container/parsec-cloud%2Fparsec-testbed-server
-    env:
-      TESTBED_SERVER: http://localhost:6777
-    services:
-      parsec-testbed-server:
-        image: ghcr.io/scille/parsec-cloud/parsec-testbed-server:3.0.0-b.6.dev.19811.e388245
-        ports:
-          - 6777:6777
-    steps:
-      - uses: actions/checkout@9bb56186c3b09b4f86b1c65136769dd318469633 # pin v4.1.2
-        timeout-minutes: 5
+  # test-rust-linux:
+  #   name: "🐧 Linux: 🦀 Rust tests"
+  #   # Just a fail-safe timeout, see the fine grain per-task timeout instead
+  #   timeout-minutes: 30
+  #   runs-on: ubuntu-20.04
+  #   # Testbed server comes as a Docker image, so it will eventually goes out of sync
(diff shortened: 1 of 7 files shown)


.github/workflows/ci-rust.yml
```diff
@@ -231,6 +231,15 @@ jobs:
           unzip D:/a/_temp/winfsp-tests.zip -d D:/a/_temp/
           mv 'D:/a/_temp/winfsp-tests-x64.exe' 'C:/Program Files (x86)/WinFsp/bin/'
 
+      - name: Install macfuse
+        if: startsWith(matrix.os, 'macos')
+        shell: bash
+        run: |
+          set -eux
+          brew install --cask macfuse
+          mv /Library/Filesystems/macfuse.fs /Library/Filesystems/fuse.fs
```
FirelightFlagboy, line 240:
```suggestion
          sudo ln -sv /Library/Filesystems/macfuse.fs /Library/Filesystems/fuse.fs
```

Make a symlink instead of moving the lib.

We also need `superuser` permission to do that

TimeEngineer: Nvm we don't need this

PR #7032, opened 2024-04-08
 [fix #7029]: Add test sse keepalive using a real server 

### LINKED ISSUE #7029: Test server sse keepalive

FirelightFlagboy:


server/tests/common/backend.py
```diff
@@ -68,6 +69,35 @@ def app(backend: Backend, monkeypatch: pytest.MonkeyPatch) -> Iterator[AsgiApp]:
...
+            s.bind((host, 0))
+            return s.getsockname()[1]
+
+    HOST = "localhost"
+    PORT = get_local_port(HOST)
+
+    proc = Process(
+        target=serve_parsec_asgi_app,
+        kwargs={
+            "app": asgi_app,
```
touilleMan, line 91: should be the `app` passed as argument

server/tests/common/backend.py
```diff
@@ -68,6 +69,35 @@ def app(backend: Backend, monkeypatch: pytest.MonkeyPatch) -> Iterator[AsgiApp]:
     asgi_app.state.backend = backend
     yield asgi_app
 
+@contextmanager
+def run_asgi_server(app: AsgiApp) -> Iterator[tuple[str, int]]:
+    from multiprocessing import Process
```
touilleMan, line 74: Add this comment to explain why we have to rely on multiprocessing:
```
# Uvicorn is not designed to be run programmatically like we are going to do here
# (it contains calls to `sys.exit`, captures signal, and spawn tasks in the event loop).
# So instead we have to spawn it in it own sub-process...
#
# Note this comes with is own issues given it means the our test process gets
# forked and hence the `app`/`backend` objects diverge between the two processes
# (i.e. cannot send a request to the server then use the `backend` object to check
# how the server internal state has changed).
```

server/tests/common/backend.py
```diff
@@ -68,6 +69,35 @@ def app(backend: Backend, monkeypatch: pytest.MonkeyPatch) -> Iterator[AsgiApp]:
...
+
+    from parsec.asgi import serve_parsec_asgi_app
+
+    def get_local_port(host: str) -> int:
```
touilleMan, line 78: Add a TODO comment:
```
# TODO: This is a hack to get a unique-enough port for the server.
#              It would be cleaner to pass port=0, then parse the sub-process
#              stdout to retrieve the actual port in use.
```

server/tests/common/backend.py
```diff
@@ -68,6 +69,35 @@ def app(backend: Backend, monkeypatch: pytest.MonkeyPatch) -> Iterator[AsgiApp]:
     asgi_app.state.backend = backend
     yield asgi_app
 
+@contextmanager
+def run_asgi_server(app: AsgiApp) -> Iterator[tuple[str, int]]:
```
touilleMan, line 73: I'd rather move this code in the `test_sse` file.
This is because this function is not good for arbitrary use (see my comment about the process fork making `app` & `backend` variables no longer usable)

As matter of fact I think we should add this info in the function name given it's a very unexpected behavior (something like `fork_process_to_run_asgi_server`)

server/tests/common/backend.py
FirelightFlagboy: Oups :sweat_smile:

FirelightFlagboy: done

server/tests/common/backend.py
FirelightFlagboy: done

server/tests/common/backend.py
FirelightFlagboy: done

server/tests/common/backend.py
FirelightFlagboy: done

PR #7034, opened 2024-04-09
[fix #7033]: Add test server sse empty `Last-Event-ID`

### LINKED ISSUE #7033: Test server sse empty `Last-Event-ID` should return `event:missed_events`

FirelightFlagboy:


server/tests/test_sse.py
```diff
@@ -86,6 +85,35 @@ async def test_missed_events(minimalorg: MinimalorgRpcClients, backend: Backend)
...
+                active_users_limit=ActiveUsersLimit.NO_LIMIT,
+                user_profile_outsider_allowed=True,
+            )
+        )
+
+        # Backend should inform that some events were missed and we could not retrieve them
+        event = await alice_sse.next_event()
+        assert event == authenticated_cmds.v4.events_listen.RepOk(
+            authenticated_cmds.v4.events_listen.APIEventPinged(ping="recent_event")
+        )
```
mmmarcos, line 114: I don't understand the comment. For me, the backend does not inform of "missed" events (nor that they could not be retrieved). It simply returns events without applying any filtering on event id. 

If that's the case, I would update the comment to something like this:
```suggestion
        # Backend should sent events without applying any filtering on event id 
        event = await alice_sse.next_event()
        assert event == authenticated_cmds.v4.events_listen.RepOk(
            authenticated_cmds.v4.events_listen.APIEventPinged(ping="recent_event")
        )
```

server/tests/test_sse.py
```diff
@@ -86,6 +85,35 @@ async def test_missed_events(minimalorg: MinimalorgRpcClients, backend: Backend)
         )
 
 
+async def test_empty_last_event_id(minimalorg: MinimalorgRpcClients, backend: Backend) -> None:
+    """
+    When the client provide `Last-Event-ID` with an empty value,
+    the server should ignore it and keep processing as if no `Last-Event-ID` was provided.
```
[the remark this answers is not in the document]
mmmarcos: BTW: header is sometimes referred as `Last-Event-ID` and sometimes as `Last-Event-Id` in our code base. This is not much of an issue because header field names are case-insensitive, but it would be nice to use the same case everywhere.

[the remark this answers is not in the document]
FirelightFlagboy: The spec write `Last-Event-ID` with `ID` in full caps: https://html.spec.whatwg.org/multipage/server-sent-events.html#the-last-event-id-header

server/tests/test_sse.py
FirelightFlagboy: It's a wrong copy/paste :sweat_smile: forgot about the comment

FirelightFlagboy:
```suggestion
        # Backend still send new events without processing `Last-Event-ID`
        event = await alice_sse.next_event()
        assert event == authenticated_cmds.v4.events_listen.RepOk(
            authenticated_cmds.v4.events_listen.APIEventPinged(ping="recent_event")
        )
```

mmmarcos (approved):

PR #7042, opened 2024-04-10
[fix #7041]: Test server sse close connection on backpressure

### LINKED ISSUE #7041: Test server sse close on backpressure (too many events stacking)

FirelightFlagboy:


server/tests/test_sse.py
```diff
@@ -85,6 +86,43 @@ async def test_missed_events(minimalorg: MinimalorgRpcClients, backend: Backend)
...
+        event = await alice_sse.next_event()
+        assert event == authenticated_cmds.v4.events_listen.RepOk(
+            authenticated_cmds.v4.events_listen.APIEventServerConfig(
+                active_users_limit=ActiveUsersLimit.NO_LIMIT,
+                user_profile_outsider_allowed=True,
+            )
+        )
+
+        # Patch the backend to simulate backpressure
+        # We simulate that sending an event to the client will block
```
mmmarcos, line 106: I'm not sure that testing the server behavior when facing a blocking client is _equivalent_ to testing backpressure of events stacking up on the server. The client may still be consuming events at a slower rate without blocking, so I guess there is somewhere a "backpressure detection" monitoring the stack of events? BTW: Where is this implemented? (I couldn't find any reference in the code base)

mmmarcos: OK, so [`send_nowait`](https://anyio.readthedocs.io/en/stable/api.html#anyio.streams.memory.MemoryObjectSendStream.send_nowait) raises `anyio.WouldBlock` if the buffer is full and there are no tasks waiting to receive :+1: 

What do you think about (also?) testing the behavior by filling up the buffer with `PER_CLIENT_MAX_BUFFER_EVENTS` events? 
https://github.com/Scille/parsec-cloud/blob/dbf6225b204a07001481ff0312cdc94866dc1820/server/parsec/components/events.py#L30

FirelightFlagboy: It does not show, but I already tried first you solution by sending `4 * PER_CLIENT_MAX_BUFFER_EVENTS` without luck, I was benchmarking my RAM + kernel in that case :sweat_smile:

mmmarcos:
```suggestion
        # Patch the backend to simulate backpressure
        # We simulate that sending an event to the client will block (this is 
        # the expected behavior if the buffer is full and there are no tasks
        # waiting to receive)
```

mmmarcos (approved):
LGTM. Just a minor suggestion.

FirelightFlagboy:
Wrong manip, a new PR is openned #7081

PR #7043, opened 2024-04-10
pyright report missing argument type for generic

FirelightFlagboy:


server/parsec/cli/testbed.py
```diff
@@ -40,24 +40,24 @@ class UnknownTemplateError(Exception):
     pass
 
 
+type Template = tuple[OrganizationID, int, testbed.TestbedTemplateContent]
```
vxgmichel, line 43: Use `TypeAlias` for the moment, `type` is still a bit too fresh for our IDEs

server/parsec/components/postgresql/__init__.py
```diff
@@ -1,6 +1,11 @@
 # Parsec Cloud (https://parsec.cloud) Copyright (c) BUSL-1.1 2016-present Scille SAS
 from __future__ import annotations
 
+import asyncpg
+
+type PgConn = asyncpg.Connection[asyncpg.Record]
+type PgPool = asyncpg.Pool[asyncpg.Record]
```
vxgmichel, line 7: Same here. 

I would call them `Connection` and `Pool` but I guess it's a matter of taste?

server/tests/administration/test_create_organization.py
```diff
@@ -71,6 +72,11 @@ async def test_bad_method(
     assert response.status_code == 405, response.content
 
 
+class CreateOrganizationParams(TypedDict):
+    active_user_limit: NotRequired[int | None]
+    user_profile_outsider_allowed: NotRequired[bool]
+
+
```
vxgmichel, line 79: I think `reportMissingTypeArgument` could also be disabled for the tests, although having it enabled is not a big deal either.

server/parsec/cli/testbed.py
mmmarcos, line 43:
```suggestion
type TestbedTemplate = tuple[OrganizationID, int, testbed.TestbedTemplateContent]
```

server/parsec/components/postgresql/__init__.py
mmmarcos, line 7: Can it be interesting to add module name to type name? 
I'm asking because our components are also prefixed with `PG...` (i.e. `PGUserComponent`) 
 
```suggestion
type AsyncpgConnection = asyncpg.Connection[asyncpg.Record]
type AsyncpgPool = asyncpg.Pool[asyncpg.Record]
```

server/parsec/api.py
```diff
@@ -36,7 +36,7 @@
 }
 
 
-def api(fn: ApiFnWithSelf) -> ApiFnWithSelf:
+def api(fn: ApiFnWithSelf) -> ApiFnWithSelf:  # pyright: ignore[reportMissingTypeArgument] see TODO for Req/Rep above
```
vxgmichel, line 39: I wonder how hard it would be to fix that

vxgmichel (approved):

server/tests/common/backend.py
```diff
@@ -100,14 +100,17 @@ def app(backend: Backend, monkeypatch: pytest.MonkeyPatch) -> Iterator[AsgiApp]:
     yield asgi_app
 
 
+type LoadedTemplate = dict[str, Template]
```
mmmarcos, line 103: It seems to me that there is nothing special about "loaded". I'd just add `Dict` suffix (if `Template` is renamed in `TestbedTemplate`):
```suggestion
type TestbedTemplateDict = dict[str, TestbedTemplate]
```

mmmarcos (commented):
LGTM overall. Just some suggestions about type names.

server/tests/common/backend.py
FirelightFlagboy: done

server/parsec/components/postgresql/__init__.py
vxgmichel: I like that :+1:

server/parsec/cli/testbed.py
FirelightFlagboy: done

server/parsec/api.py
FirelightFlagboy: :shrug: That for another PR

server/parsec/components/postgresql/__init__.py
FirelightFlagboy: done

PR #7045, opened 2024-04-10
[MS] Add workspace contextual menu in sidebar

### LINKED ISSUE #6966: Add option menu (popover) on workspaces in Sidebar
Image

Ironicbay:
Closes #6966 

## Checklist

- [x] Describe your changes
   
If this PR is related to GUI:

### DIFF
--- a/client/src/components/workspaces/WorkspaceCard.vue
+++ b/client/src/components/workspaces/WorkspaceCard.vue
@@ -3,11 +3,11 @@
 <template>
   <div
     class="card"
-    @click="$emit('click', $event, workspace)"
+    @click="$emit('click', workspace, $event)"
   >
     <div
       class="card-option"
-      @click.stop="$emit('menuClick', $event, workspace)"
+      @click.stop="$emit('menuClick', workspace, $event)"
     >
       <ion-icon :icon="ellipsisHorizontal" />
     </div>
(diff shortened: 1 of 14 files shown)


client/src/components/workspaces/utils.ts
```diff
@@ -46,3 +63,164 @@ export function canChangeRole(
...
+      clientRole: workspace.currentSelfRole,
+    },
+  });
+  await popover.present();
+
+  const { data } = await popover.onDidDismiss();
+  if (data !== undefined) {
+    switch (data.action) {
+      case WorkspaceAction.Share:
+        return workspaceShareClick(new Event('ignored'), workspace, informationManager);
```
Max-7, line 110: Missing an await.

client/src/components/workspaces/WorkspaceCard.vue
```diff
@@ -3,11 +3,11 @@
 <template>
   <div
     class="card"
-    @click="$emit('click', $event, workspace)"
+    @click="$emit('click', workspace)"
```
Max-7, line 6: The `$event` should still be present here.

Don't see the use of this component only in this particular context. This component should be usable in different ways. What if we want to open a context menu on click? We need the event to know where to position it.

All the emits should take an optional `Event` parameter, and if we can provide it, we do.

This component provides an API, it shouldn't be strongly coupled with the use we make of it. Instead, it should be as generic as possible.

This comment stands for other similar emits.

client/src/components/workspaces/utils.ts
```diff
@@ -46,3 +63,161 @@ export function canChangeRole(
...
+        await copyLinkToClipboard(workspace, informationManager);
+        break;
+      case WorkspaceAction.OpenInExplorer:
+        await openWorkspace(workspace, informationManager);
+        break;
+      case WorkspaceAction.Rename:
+        await openRenameWorkspaceModal(workspace, informationManager);
+        break;
+      default:
+        break;
```
Max-7, line 118: Add a little warning log, we should never reach the default.

Max-7 (changes requested):

Max-7 (commented):
Also would benefit from a few tests. No need to go too deep, but at least check that the button appears and opens the context menu and the context menu contains the expected actions.

NicoTuxx (approved):

Max-7 (approved):

PR #7054, opened 2024-04-11
[MS] Add mountpoint to Windows quick access

### LINKED ISSUE #6968: Add sidebar pointing to mountpoint dirs on Windows
Here
Screenshot 2024-03-27 150447

Max-7:
Closes #6968 

- [X] Describe your changes


client/electron/src/setup.ts
```diff
@@ -56,6 +57,8 @@ export class ElectronCapacitorApp {
   private customScheme: string;
   private config: object;
   public forceClose: boolean;
+  private APP_GUID = '2f56a772-db54-4a32-b264-28c42970f684';
```
FirelightFlagboy, line 60: I think you also need to set `options.nsis.guid` in `package.js` to that value

client/electron/package.js
```diff
@@ -92,6 +92,13 @@ const options = {
 
   win: {
     target: 'nsis',
+    extraResources: [
+      {
+        from: 'node_modules/regedit/vbs',
+        to: 'vbs',
+        filter: ['**/*'],
+      },
+    ],
```
FirelightFlagboy, line 101: Note that modules listed under `dependencies` in `package.json` are included in the application

(that also mean we include regedit, on Linux / MacOS)

FirelightFlagboy: I would also note that you are including `vbs` 2 time

client/electron/src/setup.ts
```diff
@@ -65,13 +68,24 @@ export class ElectronCapacitorApp {
...
       }),
     ];
 
+    log.initialize();
+    log.info('Starting app');
+    Object.assign(console, log.functions);
+
+    if (process.platform === 'win32') {
+      const reg = require('regedit');
+      const vbsDirectory = join(app.getAppPath(), '../vbs');
```
FirelightFlagboy, line 84:
```suggestion
      const vbsDirectory = join(app.getAppPath(), '../node_modules/regedit/vbs');
```

I think that should work :thinking:

NicoTuxx (approved):

client/electron/package.js
Max-7: I don't know if there's an easy way to add a dependency for one OS only. Since I import the package only on Windows it's probably not much of a problem.

PR #7085, opened 2024-04-16
Uses Argon2id in MODERATE mode for password derivation

touilleMan:


libparsec/crates/crypto/src/sodiumoxide/secret.rs
```diff
@@ -4,9 +4,7 @@ use serde::Deserialize;
 use serde_bytes::Bytes;
 use sodiumoxide::{
     crypto::{
-        pwhash::argon2i13::{
-            derive_key, Salt, MEMLIMIT_INTERACTIVE, OPSLIMIT_INTERACTIVE, SALTBYTES,
-        },
+        pwhash::argon2id13::{derive_key, Salt, MEMLIMIT_MODERATE, OPSLIMIT_MODERATE, SALTBYTES},
```
FirelightFlagboy, line 7: I would suggest to define your own custom values, instead of depending on the one used by `sodiumoxide`.

In fact we already do that for `rustcrypto`, that would simplify synchronizing the value between both implementation.

FirelightFlagboy (approved):
That values used by `sodiumoxide` are above the suggested values by `OWASP` :+1: (<https://cheatsheetseries.owasp.org/cheatsheets/Password_Storage_Cheat_Sheet.html>)

touilleMan:
> Those values used by sodiumoxide are above the suggested values by OWASP 👍 (https://cheatsheetseries.owasp.org/cheatsheets/Password_Storage_Cheat_Sheet.html)

The page you link is about password hashing, while we are doing key derivation (it is confusing given argon2 is capable of doing both ^^)

libparsec/crates/crypto/src/sodiumoxide/secret.rs
touilleMan: I agree, besides this configuration should be serialized along with the salt and encrypted data (this way we can change this configuration as needed, without breaking compatibility)

FirelightFlagboy:
For me it's the same thing between password hashing and key derivation, it's just the role you give to the resulting blob output that differ

PR #7091, opened 2024-04-17
Add RFC for formatting hexstring

FirelightFlagboy:
This RFC discus about how we should format hexstring that span multiple lines. Currently the codebase is not consistent on that matter.

The [rendered RFC](https://github.com/Scille/parsec-cloud/blob/rfc-hexstring-format/docs/rfcs/1009-hexstring-format.md)

> Better to see code example


vxgmichel (commented):
Shouldn't this be the job of `cargo fmt` though? I wouldn't mind having an explicit rule in the configuration.

FirelightFlagboy:
I would hope, but as you can see, we have lines that are >900 columns wide.
So it's clear that `cargo-fmt` isn't doing something here :sweat_smile:

vxgmichel (approved):
LGTM :+1:

PR #7107, opened 2024-04-19
Workspaceops rework read folder

### LINKED ISSUE #5820: Bindings: Move an entry
Move a file/folder to another location in the same workspace.

Something like `(clientHandle: Handle, workspaceHandle: Handle, source: FsPath, destination: FsPath) -> ??`
`source` and `destination` are full path to a file or folder.

If possible, I'd like a version that does as much thing as possible, meaning that for
`move('/a/b/c.txt', '/e/f/g.mp3')`
it:
- `/e/f/` is created if the path does not exist
- `c.txt` is removed
- `g.mp3` contains the data from `c.txt`

If `source` is a folder, move the folder and its content.

We may need to add a `force: boolean` option to replace the dest file if it already exists.

### LINKED ISSUE #6491: Handle file conflict on inbound sync in workspace ops
see https://github.com/Scille/parsec-cloud/blob/c40ad7d297aa05b6b0e2d1171b1c0ce70c9e01ff/libparsec/crates/client/src/workspace/transactions/inbound_sync.rs#L316-L325

### LINKED ISSUE #6914: `libparsec/crates/client/src/workspace/transactions/outbound_sync.rs:251:17`: not yet implemented

touilleMan:
- [x] API change
- [x] Fuse changes
- [x] WinFSP changes
 
Closes #5820
Closes #6491
Closes #6914


Max-7 (approved):

libparsec/crates/client/src/workspace/merge.rs
```diff
@@ -1,150 +1,315 @@
...
+    if remote_manifest.author == *local_author {
+        let mut new_local = local_manifest.to_owned();
+        new_local.base = remote_manifest;
+        return MergeLocalFileManifestOutcome::Merged(new_local);
+    }
+
+    // 4) Merge data and ensure the sync is still needed
+
+    // Destruct local and remote manifests to ensure this code with fail to compile
+    // whenever a new field is introduced.
```
AureliaDolo, line 61: I don't quite get why you would want that ? Is it to have a real human person check that the new field does not interact with the merging process, or to adapt the merging process ? 

Maybe annotating LocalFileManifest with [non_exhaustive](https://doc.rust-lang.org/reference/attributes/type_system.html#the-non_exhaustive-attribute) could do the trick

libparsec/crates/client/src/workspace/store/block_access.rs
```diff
@@ -0,0 +1 @@
+// Parsec Cloud (https://parsec.cloud) Copyright (c) BUSL-1.1 2016-present Scille SAS
```
AureliaDolo, line 1: This file is mostly empty

libparsec/crates/client/src/workspace/store/cache.rs
```diff
@@ -0,0 +1,411 @@
...
+pub(super) use manifest_hash_map::ManifestsHashMap;
+
+#[derive(Debug)]
+pub(super) struct CurrentViewCache {
+    /// `manifests` contains a cache on the database:
+    /// - the cache may be cleared at any given time (i.e. inserting an entry in the cache
+    ///   doesn't guarantee it will be available later on).
+    /// - if the cache is present, it always correspond to the latest value (so the cache
+    ///   should always be preferred over data coming from the database).
+    /// - the root manifest is guaranteed to be always present (even after a clear !).
```
AureliaDolo, line 161: It's just an idea and I don't have a clear idea of how it's used yet, but we could keep the root manifest in a separated field, hence avoiding some special cases.

libparsec/crates/client/src/workspace/store/per_manifest_update_lock.rs
```diff
@@ -0,0 +1,172 @@
...
+    NeedWait(EventListener),
+}
+
+/// ⚠️ This lock doesn't release on drop ⚠️
+///
+/// Instead you must manually call `PerManifestUpdateLock::release` once you are done.
+#[derive(Debug)] // Note this should not be made `Clone` !
+pub(super) struct ManifestUpdateLockGuard {
+    entry_id: VlobID,
+    #[allow(unused)]
```
AureliaDolo, line 51: This field seem to be used below.

libparsec/crates/client/src/workspace/store/reparent_updater.rs
```diff
@@ -0,0 +1,203 @@
...
+/// responsible to drop the update guards, while `ReparentingUpdater` can still be
+/// destructured (which is done in `update_manifests` to populate the cache with
+/// it manifests).
+struct ReparentingUpdaterGuards<'a> {
+    store: &'a super::WorkspaceStore,
+    update_guards: Option<(
+        ManifestUpdateLockGuard,
+        ManifestUpdateLockGuard,
+        ManifestUpdateLockGuard,
+        Option<ManifestUpdateLockGuard>,
```
AureliaDolo, line 71: Are these guards the guards of each manifest in ReparentingUpdater ?

libparsec/crates/client/src/workspace/store/block_access.rs
touilleMan: good catch ! this is a dummy file (not even including in the mod.rs)

libparsec/crates/client/src/workspace/merge.rs
touilleMan: Consider the following:

```rust
struct Data {
  updated: DateTime,
}

fn merge(local: Data, remote: Data) -> Data {
   remote.updated = max(local.updated, remote.updated);
   remote
}
```

Then we introduce a new `created` field

```rust
struct Data {
  created: DateTime,
  updated: DateTime,
}
```

The code will compile fine event if we haven't updated `merge` (typically we should have added a line like `remote.created = min(local.created, remote.created)`).

This is an issue given this `merge` function is easy to forget to update:
- it is in a different file
- its tests will most likely just ignore the new field

On the other hand, if the `merge` function is implemented with exhaustive de-structuring, it will fail to compile any time a new field is added. Hence forcing the developer to have a look at this part of the code.

This is a pattern we also use in the [CRC code of the testbed](https://github.com/Scille/parsec-cloud/blob/30f3d2d238759bce33eb3c104055b542c268d4d4/libparsec/crates/testbed/src/template/crc_hash.rs#L316-L322). Nobody knows about this piece of code and nobody cares... but if a structure changes and the CRC starts being wrong, we will have very hard time understanding why the tests start becoming randomly wrong (as client and testbed server code may end up with different data with the same CRC...)

Regarding `non_exhaustive`, its point is to forbid exhausting destructuring, which is the opposite of what we want here ;-)

libparsec/crates/client/src/workspace/store/per_manifest_update_lock.rs
touilleMan: This `is_released` boolean has only one purpose: allowing the `Drop` to ensure the the guard has been released (the release cannot be done in the `Drop` given it involves having access to a specific third party object)

However the `Drop` code has been commented for the moment: the issue anytime an arbitrary panic occurs, the guard is not released, so the `Drop` in turn panic and shadow the initial panic !

So for the moment this is disabled... but we must do something about it !
I've created any issue about that: https://github.com/Scille/parsec-cloud/issues/7367

libparsec/crates/client/src/workspace/store/reparent_updater.rs
touilleMan: yes, the `ManifestUpdateLockGuard` are low-level guard that are not released on drop

Given they must be handled with extra care, we have the Updater structures that wrap them and allow better ergonomics

libparsec/crates/client/src/workspace/merge.rs
AureliaDolo: Thanks, so we could put that behind a debug_assert to make sure that it won't be present during runtime ?

PR #7145, opened 2024-04-23
[MS] Modified app hide & close behavior on MacOS to be more native

Ironicbay:
## Checklist

- [x] Describe your changes
   
If this PR is related to GUI:

### DIFF
--- a/client/electron/src/setup.ts
+++ b/client/electron/src/setup.ts
@@ -59,6 +59,7 @@ export class ElectronCapacitorApp {
   private customScheme: string;
   private config: object;
   public forceClose: boolean;
+  public macOSForceQuit: boolean = false;
   private APP_GUID = '2f56a772-db54-4a32-b264-28c42970f684';
   private winRegistry: WinRegistry | null = null;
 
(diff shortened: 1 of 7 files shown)


client/src/main.ts
```diff
@@ -176,6 +176,14 @@ async function setupApp(): Promise<void> {
         window.electronAPI.closeApp();
       }
     });
+    window.electronAPI.receive('parsec-force-close-request', async () => {
```
Max-7, line 179: You can just pass a `force: boolean = false` parameter to `parsec-close-request` instead, and not ask any question if this parameter is set.

Max-7 (changes requested):

client/electron/src/setup.ts
```diff
@@ -256,25 +254,41 @@ export class ElectronCapacitorApp {
...
-      const tray = this.isTrayEnabled();
-      if (tray) {
-        this.hideMainWindow();
+      if (process.platform === 'darwin') {
+        if (!this.macOSForceQuit) {
+          // Red X clicked
+          if (this.MainWindow.isFullScreen()) {
+            // If fullscreen, exit fullscreen first then hide window with a one-time listener
+            this.MainWindow.setFullScreen(false);
+            this.MainWindow.once('leave-full-screen', () => {
```
Max-7, line 263: Shouldn't it be declared prior to exiting fullscreen? I don't know the internal workings but I wouldn't be surprised if the event was fired right when the `setFullScreen(false)` is executed, and thus not triggering the `once('leave-full-screen')` event. And even if it seems to work right now, it's probably better to move it before just to be sure.

Max-7 (changes requested):

client/electron/src/setup.ts
```diff
@@ -201,17 +201,30 @@ export class ElectronCapacitorApp {
...
   }
 
+  setAppMenu(): void {
+    if (process.platform === 'darwin') {
+      const menu = new Menu();
+      menu.append(new MenuItem({
+        label: app.name,
+        submenu: [
+          {
+            label: 'Quit',
```
Max-7, line 211: `Quit` should be translated depending on the locale (and updated when the locale changes). See what's done in `updateConfig()` for the tray.

Max-7 (changes requested):

Max-7 (approved):

PR #7187, opened 2024-04-25
Simplify usage of megashark-lib local build

NicoTuxx:


client/package.json
```diff
@@ -30,7 +30,10 @@
         "electron:release": "npm run electron:copy && cd electron && npm run electron:release",
         "electron:debug": "npm run electron:copy && cd electron && npm run electron:start",
         "prettier:check": "prettier . --check",
-        "prettier:fix": "prettier . --write"
+        "prettier:fix": "prettier . --write",
+        "lib:update": "rm -rf node_modules/megashark-lib && rm package-lock.json && npm install",
```
Max-7, line 34: I would name them `megashark:*` instead. `lib` sounds very generic, you get the impression that you're updating a bit of everything.

Max-7, line 34: This regenerates the `package-lock.json`. While I understand that it's supposed to be temporary, I would rather not have the command at all. The package-lock is only supposed to be updated manually when we want to explicitly update the lib, it should not be used as a dev tool (especially since the link works well!)

Max-7 (changes requested):

Max-7 (approved):

PR #7199, opened 2024-04-26
Configure auto-update for electron

FirelightFlagboy:
It require to configure `electron-builder` to use a custom publisher.

> We do not use the `publish` fonction of `electron-builder`, we just
use it to build the app files.
> We also needed the change the default GitHub publisher that was
renaming the artifact name in the channel file.

The artifact generated by `electron-builder` were renamed to contained
the full version + arch info.

> That allow use to include the `channel` file generated by
`electron-builder` in the release.

Configure electron auto-update
------------------------------

Provide a custom provider that is based on the default GitHub provider.

> That allow to have custom `channel` file.

`Parsec` is now packaged in appimage

> That will allow for linux dev to test the auto update feature.


client/electron/src/setup.ts
```diff
@@ -134,10 +138,12 @@ export class ElectronCapacitorApp {
 
   // Will evolve with auto-update
```
Max-7, line 139: Can remove this comment.

client/electron/src/setup.ts
```diff
@@ -382,8 +388,9 @@ export class ElectronCapacitorApp {
         this.MainWindow.show();
       }
       setTimeout(() => {
-        if (electronIsDev) {
+        if (electronIsDev || true) {
```
Max-7, line 391: Remove the `|| true`

Max-7 (approved):

PR #7229, opened 2024-04-29
[MS] Updated file details modal + PW tests

Max-7:
Fixed file details modal:
1. created and updated appear correctly
2. the copy path option handles error
3. the copy path option copy the full path on the system

Also added playwright tests for file details modal (and removed the cypress ones)

- [X] Describe your changes

### DIFF
--- a/client/src/parsec/workspace.ts
+++ b/client/src/parsec/workspace.ts
@@ -466,5 +466,5 @@ export async function getSystemPath(
   if (!needsMocks()) {
     return await libparsec.mountpointToOsPath(infoResult.value.mountpoints[0][0], entryPath);
   }
-  return { ok: true, value: '/home' };
+  return { ok: true, value: `/home${entryPath}` };
 }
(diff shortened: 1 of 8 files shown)


client/src/parsec/workspace.ts
```diff
@@ -466,5 +466,5 @@ export async function getSystemPath(
   if (!needsMocks()) {
     return await libparsec.mountpointToOsPath(infoResult.value.mountpoints[0][0], entryPath);
   }
-  return { ok: true, value: '/home' };
+  return { ok: true, value: `/home${entryPath}` };
```
NicoTuxx, line 469: is entryPath always starting by a '/'?

NicoTuxx (approved):
LGTM

client/src/parsec/workspace.ts
Max-7: It doesn't matter much since this is just for the mocks, but yeah, it's a `FsPath` so it should always start with a `/` if we intend to give it to the bindings.

Issue #7231, opened 2024-04-29
`package-lock.json` is missing its dependency pin info


touilleMan:
Last commit has removed them by error:

https://github.com/Scille/parsec-cloud/commit/8e3bd5ca7e220abfa591f531ddcf4874937a9b2c#diff-925cb809ce77c7f7f7f8bbd7594593438fc006dc56bf432a624c47722a784d3dL83-L84

> We talk about the `integrity` field that was removed for all dependencies

PR #7236, opened 2024-04-30
Fix missing dependencies pin info in package-lock.json

Ironicbay:
Closes https://github.com/Scille/parsec-cloud/issues/7231


FirelightFlagboy (approved):
Integrity is back :smile:

touilleMan:
image

client/package-lock.json
```diff
@@ -1085,92 +1511,244 @@
             }
         },
         "node_modules/@swc/core": {
-            "version": "1.5.0",
+            "version": "1.4.17",
```
touilleMan, line 1514: It would be better to fix this issue by going back to the last version of `package-lock.json` containing the pinned versions, then re-doing whatever change introduced the removal of the pins in the first place

I have the feeling the numerous change in version in this PR are due to the fact we fixed the issue by re-pinning the version from scratch (hence losing the old pins and there precise versions)

Ironicbay:
Addressed among other things in https://github.com/Scille/parsec-cloud/pull/7202

PR #7248, opened 2024-04-30
Add test for `fd_write/fd_read`

TimeEngineer:


TimeEngineer:
`fd_read` takes an `empty` `buffer` in argument

PR #7249, opened 2024-04-30
Add File transactions stateful test

### LINKED ISSUE #6538: Port Hypothesis test: file read/write
see https://github.com/Scille/parsec-cloud/blob/cb5cce5972e8e8228069762bbf2c0e8becd23c29/tests/core/fs/workspacefs/test_file_transactions.py#L215-L297

TimeEngineer:
closes #6538 

- `AsyncStateMachineTest` is there to allow `async` in `proptest`, this is not the greatest implementation but good enough for the moment.

- `TestbedScope` couldn't be implemented in the macro, because the `discriminant_dir` of the `env` is needed by `WorkspaceOps`.

Observation:
`ReferenceState` initialize once for each case and use `clone` to save the state and uses it to shrink in order to find the minimal reproductible fail, so it's not viable for `I/O` that's why the `oracle` is implemented in `SUT` (System under test).

Explanation:
The `ReferenceState` is run alone then both `ReferenceState` and `SUT` are run to test the `SUT` behavior,
So if something is written in file, it will be kept in the first run and the state will be different when both are run together.
Ask @vxgmichel for more information

### DIFF
--- a/libparsec/crates/client/tests/unit/workspace/file_operations_stateful.rs
+++ b/libparsec/crates/client/tests/unit/workspace/file_operations_stateful.rs
@@ -185,9 +185,7 @@ impl StateMachineTest for FileOperationStateMachine {
 
 prop_state_machine! {
     #![proptest_config(Config {
-        // Enable verbose mode to make the state machine test print the
-        // transitions for each case.
-        verbose: 0,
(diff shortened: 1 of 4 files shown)


vxgmichel (approved):
LGTM :+1:

PR #7265, opened 2024-05-03
[MS] Added pw test for rename workspace

Ironicbay:


client/tests/pw/e2e/workspaces_actions.spec.ts
```diff
@@ -0,0 +1,14 @@
...
+import { expect } from '@tests/pw/helpers/assertions';
+import { msTest } from '@tests/pw/helpers/fixtures';
+import { clearThenFillInputModal } from '@tests/pw/helpers/utils';
+
+msTest('Rename a workspace', async ({ connected }) => {
+  const firstWorkspace = connected.locator('.workspaces-container-grid').locator('.workspaces-grid-item').nth(1);
+  await firstWorkspace.locator('.card-option').click();
+  const popover = connected.locator('.popover-viewport');
+  await popover.getByRole('group').nth(0).click();
+  await clearThenFillInputModal(connected, 'New Workspace Name');
```
Max-7, line 12: A `fill()` directly doesn't work with the pre-selection? Do we really need to clear the input first before entering text?
Anyway, just add an argument with a default value to the existing function, no need to add an entire function for that.

client/tests/pw/helpers/utils.ts
```diff
@@ -45,6 +45,11 @@ export async function fillIonInput(ionInput: Locator, text: string): Promise<voi
   await input.blur();
 }
 
+export async function clearIonInput(ionInput: Locator): Promise<void> {
```
Max-7, line 48: An input can be cleared by using `fill('')`, we don't really need another function. Or if you really feel like we need it, it could just be defined as
`export const clearIonInput = async (ionInput: Locator) => await fillIonInput(ionInput, '');`

Max-7 (changes requested):

Max-7 (approved):

PR #7275, opened 2024-05-06
[MS] Adds workspace role bulk assignment

### LINKED ISSUE #7083: [🚀 | Feature request]: Port: Bulk re-assignment of workspace rights
#5938 should be ported to V3. In principle, this can be fully implemented in the client side.

Max-7:
closes #7083 

https://www.figma.com/file/XW3SNWHV8BhPsPJpsOCV2h/Parsec-App?type=design&node-id=30161-85608&mode=design&t=dt6374A9svgzPN4L-0

Needs a little makeover.

- [X] Describe your changes

### DIFF
--- a/client/src/components/users/UserAvatarName.vue
+++ b/client/src/components/users/UserAvatarName.vue
@@ -81,7 +81,7 @@ defineProps<{
   }
 
   &:not(.main-cell) .person-name {
-    color: var(--parsec-color-light-secondary-grey);
+    color: var(--parsec-color-light-secondary-text);
     text-overflow: ellipsis;
     white-space: nowrap;
     overflow: hidden;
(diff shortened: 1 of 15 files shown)


mmmarcos:
If this is related to #7083, can you link the issue?

Ironicbay (changes requested):
- The option should not be available on revoked users
- Text does not display properly in the popover
- I would suggest adding the option in the UsersPage action bar when a user is selected
Screenshot 2024-05-31 at 13 13 49

Max-7:
> * The option should not be available on revoked users
It should. Revoked users keep their role so you may want to assign the role of a user you've revoked to someone else.

> * Text does not display properly in the popover
What do you mean?

> * I would suggest adding the option in the UsersPage action bar when a user is selected
Yeah but we can't select revoked users so it may seem a bit weird to be able to do the action from one place but not the other.

Ironicbay (approved):

PR #7292, opened 2024-05-13
Enable electron sandbox when not in dev

FirelightFlagboy:


Max-7 (commented):
Apparently, `sandboxing` is disabled when `nodeIntegration` is set to true anyway so it shouldn't change anything (see https://www.electronjs.org/docs/latest/tutorial/sandbox#disabling-the-sandbox-for-a-single-process)

FirelightFlagboy:
This does not solve the issue that @AureliaDolo encountered, I'm closing the PR

PR #7293, opened 2024-05-13
Update wasm-pack version in the developpement readme.

AureliaDolo:


FirelightFlagboy (changes requested):
The version need to be changed in 

https://github.com/Scille/parsec-cloud/blob/0361ef72b6a82254328a852b70e6b1dbefa3fac6/misc/version_updater.py#L175

Then you run 

```
python misc/version_updater.py --tool wasm-pack
```

mmmarcos (approved):
LGTM

FirelightFlagboy (approved):

PR #7294, opened 2024-05-14
Detect when a glog pattern produce no match

FirelightFlagboy:
The file `docs/development/README.md` was not monitored by the script since it was renamed from `quickstart.md`. The script will now check that the glob pattern produce at least one match and will log an error if it does not.

### DIFF
--- a/misc/version_updater.py
+++ b/misc/version_updater.py
@@ -255,7 +255,7 @@ FILES_WITH_VERSION_INFO: dict[Path, dict[Tool, RawRegexes]] = {
         Tool.License: [JSON_LICENSE_FIELD],
         Tool.Parsec: [JSON_VERSION_FIELD],
     },
-    ROOT_DIR / "docs/development/quickstart.md": {
+    ROOT_DIR / "docs/development/README.md": {
         Tool.Rust: [
             ReplaceRegex(r"Rust v[0-9.]+", "Rust v{version}"),
             ReplaceRegex(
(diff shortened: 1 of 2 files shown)


misc/version_updater.py
```diff
@@ -467,14 +467,18 @@ def check_tool(tool: Tool, update: bool) -> VersionUpdateResult:
...
+            for file in matched_files:
+                file = Path(file)
+                if update:
+                    res = update_tool_version(file, regexes, version)
+                    update_res.errors += res.errors
+                    update_res.updated.update(res.updated)
+                else:
+                    update_res.errors += check_tool_version(file, regexes, version).errors
+        else:
+            update_res.errors.append(f"No match with regex `{filename}`")
```
mmmarcos, line 481: A couple of suggestions: 
- Use "error first" pattern (treat "error" cases first to remove indentation levels) 
- Rename `filename` in `filepath_regex`
- Use `map` in the inner loop to get a list of `Path` objects

```suggestion
        matched_files = glob.glob(str(filepath_regex), recursive=True)
        if not matched_files:
            update_res.errors.append(f"No match with regex `{filepath_regex}`")
            continue

        # Check or update matched files
        for filepath in map(Path, matched_files):
            if update:
                res = update_tool_version(filepath, regexes, version)
                update_res.errors += res.errors
                update_res.updated.update(res.updated)
            else:
                update_res.errors += check_tool_version(filepath, regexes, version).errors
```

FirelightFlagboy: See #7299

PR #7299, opened 2024-05-14
[fix #7294]: Apply suggestions for `version_updater`

FirelightFlagboy:
Those suggestion were proposed by @mmmarcos in the PR #7294.

It also include some typing improvements.


misc/version_updater.py
```diff
@@ -462,23 +462,23 @@ def check_tool(tool: Tool, update: bool) -> VersionUpdateResult:
...
-                if update:
-                    res = update_tool_version(file, regexes, version)
-                    update_res.errors += res.errors
-                    update_res.updated.update(res.updated)
-                else:
-                    update_res.errors += check_tool_version(file, regexes, version).errors
-        else:
-            update_res.errors.append(f"No match with regex `{filename}`")
+        matched_files = glob.glob(str(glob_path), recursive=True)
+        if len(matched_files) == 0:
```
mmmarcos, line 471: If glob returns an empty list in case of no match: 
```suggestion
        if not matched_files:
```

mmmarcos (approved):

PR #7302, opened 2024-05-14
Don't create an updater if the configuration isn't available

FirelightFlagboy:
If we're in dev mode, we likely don't have the publish configuration file. 
In that case, we should not create the updater and consider no update will be available.

### DIFF
--- a/client/electron/src/updater.ts
+++ b/client/electron/src/updater.ts
@@ -47,10 +47,28 @@ class CustomGithubProvider extends GitHubProvider {
   }
 }
 
-const publishOption: CustomPublishOptions | CustomGitHubOptions = {
-  ...require('../assets/publishConfig.json'),
-  updateProvider: CustomGithubProvider,
-};
+function loadPublishOption(): (CustomPublishOptions & CustomGitHubOptions) | undefined {
+  let data = undefined;
+  try {
+    data = require('../assets/publishConfig.json');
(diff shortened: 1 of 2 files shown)


client/electron/src/updater.ts
```diff
@@ -47,10 +48,29 @@ class CustomGithubProvider extends GitHubProvider {
...
-};
+function loadPublishOption(): (CustomPublishOptions & CustomGitHubOptions) | undefined {
+  let rawData = undefined;
+  try {
+    rawData = readFileSync('../assets/publishConfig.json', 'utf8');
+  } catch {
+    console.log('Failed to load publish config file');
+    return undefined;
+  }
+  const data = JSON.parse(rawData);
```
Max-7, line 59: Can't you use import for that? `loadPublishOption` would need to be async but you should be able to do
```typescript
try {
  config = await import('../assets/publishConfig.json');
} catch {
  ...
}
```

Max-7 (changes requested):

client/electron/src/updater.ts
FirelightFlagboy: `loadPublishOption` can't be made public because that would require the constructor of `ElectronCapacitorApp` to be async which isn't possible (I can't simply put await in the constructor)

Max-7 (approved):

PR #7308, opened 2024-05-15
Disable snap in release

FirelightFlagboy:


touilleMan (approved):
Ideally we should add a `# TODO: snap is temporary disabled` comment to ensure we won't miss something when re-enabling

PR #7312, opened 2024-05-15
fix: adapt settings and config to lib upgrade

NicoTuxx:


client/src/views/settings/SettingsModal.vue
```diff
@@ -145,16 +145,7 @@ const storageManager: StorageManager = inject(StorageManagerKey)!;
...
-    label: 'SettingsModal.language.values.enUS',
-  },
-  {
-    key: 'fr-FR',
-    label: 'SettingsModal.language.values.frFR',
-  },
-]);
+const languageOptions: MsOptions = new MsOptions(LocaleOptions);
 
 const themeOptions: MsOptions = new MsOptions([
```
Max-7, line 150: Should use `Theme.System` as key.

client/src/views/settings/SettingsModal.vue
```diff
@@ -196,9 +187,6 @@ async function changeLang(lang: Locale): Promise<void> {
 }
 
 async function changeTheme(selectedTheme: Theme | 'system'): Promise<void> {
```
Max-7, line 189: Can remove `| 'theme` from the typing.

Max-7 (changes requested):

Max-7 (approved):

Issue #7119, opened 2024-04-19
[Sever v3] Remove code duplication for each authenticated route


vxgmichel:
Those blocks are repeated for each authenticated route:
```python
        match await self.organization._get(conn, organization_id):
            case OrganizationGetBadOutcome.ORGANIZATION_NOT_FOUND:
                return RealmShareStoreBadOutcome.ORGANIZATION_NOT_FOUND
            case Organization() as org:
                pass

        if org.is_expired:
            return RealmShareStoreBadOutcome.ORGANIZATION_EXPIRED

        match await self.user._check_device(conn, organization_id, author):
            case CheckDeviceBadOutcome.DEVICE_NOT_FOUND:
                return RealmShareStoreBadOutcome.AUTHOR_NOT_FOUND
            case CheckDeviceBadOutcome.USER_NOT_FOUND:
                return RealmShareStoreBadOutcome.AUTHOR_NOT_FOUND
            case CheckDeviceBadOutcome.USER_REVOKED:
                return RealmShareStoreBadOutcome.AUTHOR_REVOKED
            case (UserProfile(), DateTime()):
                pass
```

That's a lot of duplication that can be factorized away.

PR #7324, opened 2024-05-16
Port: Shamir-based shared recovery device

### LINKED ISSUE #6090: [🚀 | Feature request]: Port: Shamir-based shared recovery device
Shamir-based shared recovery device is described in [RFC 1000](https://github.com/Scille/parsec-cloud/blob/master/docs/rfcs/1000-shamir-based-recovery.md). 

It was implemented for the V2 branch in the following PRs:
- #4504 
- #4929
- #5946

It must be ported to the V3 branch (`master`).

NOTE: V2 implementation was CLI-based, so the GUI part (#8671) is out of scope for the moment.

## Task List

- [x] #7362
- [x] ~#7360~ (edit: it's already in get_certificates) 
- [x] #7357
- [x] #8206
- [x] #7359
- [x] #7363
- [x] #8841
- [x] #7358
- [x] #8672
- [x] #7364
- [x] #7361
- [x] #9180
- [x] #8990

The GUI would not be implemented right now, so we can leave the following issue for later: 

- #8998

### Setup

TODO: Describe how to local setup for testing this feature (testbed ? local server ? ...)

### Tests

TODO: List some test cases to be performed when reviewing the feature.

### LINKED ISSUE #7357: Shamir port: server side route create_shared_recovery_device
Related #6090 

- [x] command schema in server/parsec/crates/schema/authenticated_cmds
- [x] run gen_protocol_typing.py
- [x] implementation in server/parsec/components, leaving the potgre implementation for #7358
- [ ] libparsec tests
- [ ] server tests

AureliaDolo:
Related #6090 

Fix #7357


AureliaDolo:
I have to split the checklist into issues and I'll implement the postrgre version once every route has been tested with the memory impl

libparsec/crates/protocol/schema/authenticated_cmds/shamir_create_shared_recovery_device.json5
```diff
@@ -0,0 +1,71 @@
...
+                "status": "invalid_certification"
+            },
+            {
+                // Cannot deserialize data into the expected certificate, or inconsistency
+                // between  certificates and/or threshold
+                "status": "invalid_data"
+            },
+            {
+                // Future evolution 1: Shamir recovery has already been setup, should ask your admin to reset it first !
+                "status": "already_set"
```
AureliaDolo, line 31: For now we overwrite the shamir setup

server/parsec/components/shamir.py
```diff
@@ -0,0 +1,102 @@
...
+        raise NotImplementedError
+
+
+def verify_certificates(
+    setup: authenticated_cmds.latest.shamir_recovery_setup.ShamirRecoverySetup,
+    author: DeviceID,
+    author_verify_key: VerifyKey,
+) -> (
+    tuple[ShamirRecoveryBriefCertificate, dict[UserID, bytes]]
+    | authenticated_cmds.latest.shamir_recovery_setup.RepInvalidData
```
AureliaDolo, line 70: return custom error

server/tests/api_v4/authenticated/test_shamir_recovery_setup.py
```diff
@@ -0,0 +1,62 @@
...
+async def test_authenticated_shamir_recovery_setup_already_set(
+    coolorg: CoolorgRpcClients,
+) -> None:
+    pass
+
+
+async def test_authenticated_shamir_recovery_setup_invalid_certification(
+    coolorg: CoolorgRpcClients,
+) -> None:
+    pass
```
AureliaDolo, line 50: TODO

libparsec/crates/testbed/src/template/crc_hash.rs
```diff
@@ -173,6 +173,7 @@ macro_rules! impl_crc_hash_for_uuid_based {
 }
 
 impl_crc_hash_for_uuid_based!(VlobID);
+impl_crc_hash_for_uuid_based!(ShamirRevealToken);
```
AureliaDolo, line 176: I'm not sure about that. I did that because it was in the 2.17 version

libparsec/crates/types/src/id.rs
```diff
@@ -211,6 +211,8 @@ new_uuid_type!(pub BlockID);
 new_uuid_type!(pub ChunkID);
 new_uuid_type!(pub SequesterServiceID);
 new_uuid_type!(pub EnrollmentID);
+new_uuid_type!(pub ShamirRevealToken);
```
FirelightFlagboy, line 214: Should be defined in `token.rs`

FirelightFlagboy (changes requested):

libparsec/crates/types/src/id.rs
```diff
@@ -211,6 +211,7 @@ new_uuid_type!(pub BlockID);
 new_uuid_type!(pub ChunkID);
 new_uuid_type!(pub SequesterServiceID);
 new_uuid_type!(pub EnrollmentID);
+
 impl_from_maybe!(std::collections::HashSet<VlobID>);
```
FirelightFlagboy, line 215: Revert this change

libparsec/crates/testbed/src/template/crc_hash.rs
FirelightFlagboy: See how it's done for `InvitationToken` (it's defined few line below it)

server/tests/api_v4/authenticated/test_shamir_recovery_setup.py
FirelightFlagboy: For those tests, do the following:

1. add the `xfail` marker (`@pytest.mark.xfail(reason="TODO")`)
2. In the body of the test add `raise NotImplementedError`

server/tests/api_v4/authenticated/test_shamir_recovery_setup.py
```diff
@@ -0,0 +1,63 @@
...
+
+
+async def test_authenticated_shamir_recovery_setup_invalid_data(
+    coolorg: CoolorgRpcClients, with_postgresql: bool
+) -> None:
+    if with_postgresql:
+        pytest.skip("TODO: postgre not implemented yet")
+    setup = authenticated_cmds.v4.shamir_recovery_setup.ShamirRecoverySetup(
+        bytes("abc", "utf-8"), ShamirRevealToken(), bytes("ijk", "utf-8"), [bytes("lmn", "utf-8")]
```
FirelightFlagboy:
```suggestion
        b"abc", ShamirRevealToken(), b"ijk", [b"lmn"]
```
Simpler notation, it's was done in the first test function on this file.

A specific reason to use `bytes` instead ?

server/parsec/components/postgresql/shamir.py
```diff
@@ -0,0 +1,8 @@
+# Parsec Cloud (https://parsec.cloud) Copyright (c) BUSL-1.1 2016-present Scille SAS
+
+from parsec.components.shamir import BaseShamirComponent
+
+
+class PGShamirComponent(BaseShamirComponent):
+    # TODO
```
FirelightFlagboy, line 7:
```suggestion
    # TODO: https://github.com/Scille/parsec-cloud/issues/7358
```

We have a specific issue to track that, When we can, it's nice to reference it in the `TODO` & co

server/parsec/_parsec_pyi/ids.pyi
```diff
@@ -214,3 +214,20 @@ class BootstrapToken:
     def int(self) -> int: ...
     @property
     def hyphenated(self) -> str: ...
+
+class ShamirRevealToken:
```
AureliaDolo, line 218: Is there a reason for tokens to be defined here with ids ?

FirelightFlagboy: `InvitationToken` is also defined in this file, we curently don't have a specific file for token types

server/tests/api_v4/authenticated/test_shamir_recovery_setup.py
AureliaDolo: I forgot about this notation ^^ I'm a bit rusty in python

server/parsec/_parsec_pyi/ids.pyi
FirelightFlagboy: cf: https://github.com/Scille/parsec-cloud/issues/7388

server/tests/api_v4/authenticated/test_shamir_recovery_setup.py
```diff
@@ -0,0 +1,91 @@
...
+        b"def",
+        ShamirRevealToken.new(),
+        brief.dump_and_sign(coolorg.alice.signing_key),
+        [share.dump_and_sign(coolorg.alice.signing_key)],
+    )
+    rep = await coolorg.alice.shamir_recovery_setup(setup)
+    assert rep == authenticated_cmds.v4.shamir_recovery_setup.RepAlreadySet()
+
+
+@pytest.mark.xfail(reason="TODO")
```
AureliaDolo, line 72: I wasn't able to find the equivalent in v2.17.0, so I don't know when to return the InvalidCertification errror

libparsec/crates/protocol/schema/authenticated_cmds/shamir_create_shared_recovery_device.json5
```diff
@@ -0,0 +1,71 @@
...
+                }
+            ]
+        },
+        "reps": [
+            {
+                "status": "ok"
+            },
+            {
+                // Certificate is not signed by the authenticated user, or timestamp is invalid
+                "status": "invalid_certification"
```
FirelightFlagboy, line 22:
```suggestion
                "status": "invalid_certificate"
```

For consistency: https://github.com/search?q=repo%3AScille%2Fparsec-cloud+invalid_certificate+path%3A*.json5&type=code

libparsec/crates/protocol/tests/authenticated_cmds/v4/shamir_recovery_setup.rs
FirelightFlagboy, line 1: Those tests are missing comment about the data contained in the hexstring.

Some examples:

- https://github.com/Scille/parsec-cloud/blob/bebe03bb7d2ec1b826275554f589ff514633b7b2/libparsec/crates/protocol/tests/authenticated_cmds/v4/block_read.rs#L15-L22
- https://github.com/Scille/parsec-cloud/blob/bebe03bb7d2ec1b826275554f589ff514633b7b2/libparsec/crates/protocol/tests/authenticated_cmds/v4/invite_1_greeter_wait_peer.rs#L15-L25

server/parsec/components/memory/datamodel.py
```diff
@@ -395,3 +399,24 @@ class MemoryBlock:
...
+    reveal_token: ShamirRevealToken
+    # The Shamir recovery setup provided as a `ShamirRecoveryBriefCertificate`.
+    # It contains the threshold for the quorum and the shares recipients.
+    # This field has a certain level of duplication with the "shares" below,
+    # but they are used for different things (we provide the encrypted share
+    # data only when needed)
+    brief: ShamirRecoveryBriefCertificate
+    # The shares provided as a `ShamirRecoveryShareCertificate` since
+    # each share is aimed at a specific recipient.
+    shares: dict[UserID, bytes]
```
FirelightFlagboy, line 422: @vxgmichel will we provide "weight" to user share ?

In simpler term it's how many shares a user get, example:

I create 5 shares with a minimum of 3 shares,
I give 2 shares to Alice and 2 shares to Bob and the last one to John.

To reconstruct the secret, Alice provide its 2 shares and either John or Bob need to provide one.

server/parsec/components/shamir.py
```diff
@@ -0,0 +1,106 @@
...
+            ):
+                case None:
+                    return authenticated_cmds.latest.shamir_recovery_setup.RepOk()
+                case authenticated_cmds.latest.shamir_recovery_setup.Rep() as error:
+                    return error
+
+    # async def test_dump_current_shamir(
+    #     self, organization_id: OrganizationID
+    # ) -> dict[UserID, ShamirDump]:
+    #     raise NotImplementedError
```
FirelightFlagboy, line 47: Dead code ? can it be removed ?

server/tests/api_v4/authenticated/test_shamir_recovery_setup.py
FirelightFlagboy: That error is when the provided certificate is invalid, the certificate is provided as `bytes` but it should be validated by the server (or another client)

We have a similar example:

https://github.com/Scille/parsec-cloud/blob/771699834edb838f0e1d6fc8b72f65a20da802ef/server/tests/api_v4/authenticated/test_realm_create.py#L82

server/tests/common/rpc.py
```diff
@@ -31,12 +31,25 @@ async def ping(self, ping: str) -> anonymous_cmds.latest.ping.Rep:
         raw_rep = await self._do_request(req.dump())
         return anonymous_cmds.latest.ping.Rep.load(raw_rep)
 
-    async def pki_enrollment_info(
-        self, enrollment_id: EnrollmentID
-    ) -> anonymous_cmds.latest.pki_enrollment_info.Rep:
-        req = anonymous_cmds.latest.pki_enrollment_info.Req(enrollment_id=enrollment_id)
+    async def pki_enrollment_submit(
```
FirelightFlagboy, line 34: A specific reason why those functions `pki_enrollment_info` and `pki_enrollment_submit` were inverted in position ?

FirelightFlagboy: NVM: It's because it's auto-generated

https://github.com/Scille/parsec-cloud/blob/4687a9109461f2b667d25c7df668a96d3dcae49b/server/tests/common/rpc.py#L3

Looks like someone forgot to sort the method before

server/parsec/components/memory/datamodel.py
vxgmichel: @FirelightFlagboy  Yes, the shares are weighted. As explained in the comment, each value in the dict is a serialized `ShamirRecoveryShareCertificate`. This share certificate contains a `ciphered_share` field that the user can decrypt and deserialize into a `ShamirRecoveryShareData`. This structure contains a `weighted_share` field which is a list of `ShamirShare`. The length of this list is the weight of the share.

server/tests/common/rpc.py
FirelightFlagboy, line 1: With the latest change on master (2199b48930efceb6d37be5025cd9020fb3d10775) regenerate this file.

This require a rebase, you will have a conflict but it will be solve by executing the command `python misc/gen_protocol_typings.py`

libparsec/crates/protocol/schema/authenticated_cmds/shamir_create_shared_recovery_device.json5
```diff
@@ -0,0 +1,71 @@
...
+    {
+        "major_versions": [
+            4
+        ],
+        "req": {
+            "cmd": "shamir_recovery_setup",
+            "fields": [
+                {
+                    "name": "setup",
+                    // Set to `None` to clear previous Shamir recovery setup
```
mmmarcos, line 11:
```suggestion
                    // Set to `None` to clear previous Shamir recovery setup
                    "name": "setup",
```

libparsec/crates/protocol/schema/authenticated_cmds/shamir_create_shared_recovery_device.json5
mmmarcos, line 31: Create commands (`block_create`, `device_create`, `realm_create`, `user_create`) have a status named `{entity}_already_exists` (i.e. `user_alread_exists`). Maybe we should use the same logic here. 
That said, I'm not sure what the entity should be:
```suggestion
                "status": "shamir_already_exists"
```
```suggestion
                "status": "shamir_setup_already_exists"
```

It can also be just `already_exists` although it may become ambiguous if we add more arguments to the request.

libparsec/crates/protocol/schema/authenticated_cmds/shamir_create_shared_recovery_device.json5
```diff
@@ -0,0 +1,71 @@
...
+            }
+        ],
+        "nested_types": [
+            {
+                "name": "ShamirRecoverySetup",
+                "fields": [
+                    {
+                        // The actual data we want to recover.
+                        // It is encrypted with `data_key` that is itself split into shares.
+                        // This should contains a serialized `LocalDevice`
```
mmmarcos, line 41:
```suggestion
                        // It is encrypted with `data_key` which is itself split into shares.
                        // This should contain a serialized `LocalDevice`
```

libparsec/crates/protocol/schema/authenticated_cmds/shamir_create_shared_recovery_device.json5
```diff
@@ -0,0 +1,71 @@
...
+                        "name": "ciphered_data",
+                        "type": "Bytes"
+                    },
+                    {
+                        // The token the claimer should provide to get access to `ciphered_data`.
+                        // This token is split into shares, hence it acts as a proof the claimer
+                        // asking for the `ciphered_data` had it identity confirmed by the recipients.
```
mmmarcos, line 48:
```suggestion
                        // asking for the `ciphered_data` had its identity confirmed by the recipients.
```

server/parsec/components/memory/datamodel.py
```diff
@@ -68,6 +70,8 @@ class MemoryOrganization:
     vlobs: dict[VlobID, list[MemoryVlobAtom]] = field(default_factory=dict)
     blocks: dict[BlockID, MemoryBlock] = field(default_factory=dict)
     block_store: dict[BlockID, bytes] = field(default_factory=dict, repr=False)
+    # The user id is the author of the shamir recovery process
+    shamir_setup: dict[UserID, MemoryShamirSetup] = field(default_factory=dict)
```
vxgmichel, line 74: Contrary to v2, shamir certificates need to be kept even when they're outdated by a new setup, due to those certificates being part of the shamir topic.

It's probably enough to turn `shamir_setup` into a `dict[UserID, List[MemoryShamirSetup]]` for the moment, although the postgresql tables will probably be trickier than that.

Also, this doesn't take shamir removal into account, so this attribute might eventually become a `dict[UserID, List[MemoryShamirSetup | MemoryShamirRemoval]]`

server/parsec/components/memory/shamir.py
```diff
@@ -0,0 +1,47 @@
...
+        super().__init__()
+        self._data = data
+        self._event_bus = event_bus
+
+    async def remove_recovery_setup(
+        self,
+        organization_id: OrganizationID,
+        author: UserID,
+    ) -> None:
+        self._data.organizations[organization_id].shamir_setup.pop(author)
```
vxgmichel, line 27: Maybe add a `TODO` to indicate this route will change once the `ShamirRecoveryRemovalCertificate` is added.

server/parsec/components/memory/shamir.py
```diff
@@ -0,0 +1,47 @@
...
+
+    async def add_recovery_setup(
```
mmmarcos, line 29: Some sanity_check is probably required before calling `verify_certificates`: the organization exists and is not expired? author exists and is not revoked? 
See for example https://github.com/Scille/parsec-cloud/blob/53642307a06c6eddf2e1de42226d750cacfe7dc5/server/parsec/components/memory/realm.py#L79-L92

BTW there is an ongoing discussion about these checks (error-prone, duplicated code): #7119

server/parsec/components/shamir.py
```diff
@@ -0,0 +1,106 @@
...
+        raise NotImplementedError
+
+    async def add_recovery_setup(
+        self,
+        organization_id: OrganizationID,
+        author: UserID,
+        device: DeviceID,
+        author_verify_key: VerifyKey,
+        setup: authenticated_cmds.latest.shamir_recovery_setup.ShamirRecoverySetup,
+    ) -> None | authenticated_cmds.latest.shamir_recovery_setup.Rep:
```
vxgmichel, line 63: The return type should be `None | ShamirAddRecoverySetupBadOutcome`

libparsec/crates/protocol/schema/authenticated_cmds/shamir_create_shared_recovery_device.json5
```diff
@@ -0,0 +1,71 @@
...
+            {
+                "status": "ok"
+            },
+            {
+                // Certificate is not signed by the authenticated user, or timestamp is invalid
+                "status": "invalid_certification"
+            },
+            {
+                // Cannot deserialize data into the expected certificate, or inconsistency
+                // between  certificates and/or threshold
```
mmmarcos, line 26:
```suggestion
                // between certificates and/or threshold
```

server/parsec/components/memory/shamir.py
```diff
@@ -0,0 +1,47 @@
...
+            case (brief, shares):
+                if self._data.organizations[organization_id].shamir_setup.get(author) is None:
+                    self._data.organizations[organization_id].shamir_setup[author] = (
+                        MemoryShamirSetup(setup.ciphered_data, setup.reveal_token, brief, shares)
+                    )
+                else:
+                    return authenticated_cmds.latest.shamir_recovery_setup.RepAlreadySet()
+
+            case VerifyCertificatesBadOutcome.INVALID_DATA:
+                return authenticated_cmds.latest.shamir_recovery_setup.RepInvalidData()
```
vxgmichel, line 47: Contrary to v2, this method needs to include all the required checks:
- organization exists and not expired
- author exists and not revoked and not frozen
- all recipients exists and not revoked
- check that certificate timestamps are strictly increasing in the shamir topic

See methods in other components for reference.

vxgmichel (changes requested):

libparsec/crates/protocol/schema/authenticated_cmds/shamir_create_shared_recovery_device.json5
```diff
@@ -0,0 +1,71 @@
...
+                "status": "invalid_data"
```
mmmarcos, line 27: Why not expose specific status? i.e. `invalid_recipient`, `recipient_not_in_brief`, `recipient_already_has_share`, etc.
It seems that this is how it was defined in the RFC, but I ask just out of curiosity.
cc @vxgmichel

mmmarcos (commented):
Just some comments. Thanks for all those docstrings and code comments! :+1:

libparsec/crates/protocol/tests/authenticated_cmds/v4/shamir_recovery_setup.rs
```diff
@@ -0,0 +1,117 @@
...
+
+    // Also test serialization round trip
+    let raw2 = data.dump().unwrap();
+    let data2 = authenticated_cmds::shamir_recovery_setup::Rep::load(&raw2).unwrap();
+    p_assert_eq!(data2, expected);
+}
+
+pub fn req() {
+    let empty_req = authenticated_cmds::shamir_recovery_setup::Req { setup: None };
+    let expected = authenticated_cmds::AnyCmdReq::ShamirRecoverySetup(empty_req.clone());
```
FirelightFlagboy, line 72: **nitpick**: Move the `expected` after `let data = ...` (line 78)

libparsec/crates/protocol/tests/authenticated_cmds/v4/shamir_recovery_setup.rs
```diff
@@ -0,0 +1,117 @@
...
+    let raw = hex!("82a3636d64b57368616d69725f7265636f766572795f7365747570a57365747570c0");
+    let data = authenticated_cmds::AnyCmdReq::load(&raw).unwrap();
+    p_assert_eq!(data, expected);
+
+    // Also test serialization round trip
+    let raw2 = empty_req.dump().unwrap();
+    let data2 = authenticated_cmds::AnyCmdReq::load(&raw2).unwrap();
+    p_assert_eq!(data2, expected);
+
+    let req = authenticated_cmds::shamir_recovery_setup::Req {
```
FirelightFlagboy, line 86: **nitpick**: Move the `expected` after `let data = ...` (line 110)

libparsec/crates/protocol/schema/authenticated_cmds/shamir_create_shared_recovery_device.json5
vxgmichel: The RFC doesn't provide more specific status as far as I can tell.

More status could be useful for debug, although I would not expect the client code to behave differently depending on those status (it's probably a bug if the client sent invalid data).

So I don't really have an opinion about it.

server/parsec/components/memory/shamir.py
```diff
@@ -0,0 +1,89 @@
...
+                # check that certificate timestamps are strictly increasing in the shamir topic
+                match self._data.organizations[organization_id].last_shamir_certificate:
+                    case None:
+                        # no previous shamir setup
+                        pass
+                    case v if v < brief.timestamp:
+                        # previous shamir happened strictly before
+                        pass
+                    case _:
+                        return ShamirAddRecoverySetupBadOutcome.INVALID_CERTIFICATE
```
vxgmichel, line 79: This should be `RequireGreaterTimestamp`

server/parsec/components/memory/shamir.py
```diff
@@ -0,0 +1,89 @@
...
+
+                if self._data.organizations[organization_id].shamir_setup.get(author) is None:
+                    self._data.organizations[organization_id].shamir_setup[author] = (
+                        MemoryShamirSetup(setup.ciphered_data, setup.reveal_token, brief, shares)
+                    )
+                else:
+                    return ShamirAddRecoverySetupBadOutcome.ALREADY_SET
+
+            case VerifyCertificatesBadOutcome.INVALID_DATA:
+                return ShamirAddRecoverySetupBadOutcome.INVALID_CERTIFICATE
```
vxgmichel, line 89: Minor style comments: in the other methods, we tend to have a `match` to validate the certificate and then perform some logic once it's validated outside of the match case. This makes the flow of the method more obvious to the reader:
```python
        match verify_certificates(setup, device, author_verify_key):
            case (brief, shares):
                pass
            case VerifyCertificatesBadOutcome.INVALID_DATA:
                return ShamirAddRecoverySetupBadOutcome.INVALID_CERTIFICATE
        # Then do something with `brief` and `shares`
```

server/parsec/components/shamir.py
```diff
@@ -0,0 +1,130 @@
...
+    author: DeviceID,
+    author_verify_key: VerifyKey,
+) -> tuple[ShamirRecoveryBriefCertificate, dict[UserID, bytes]] | VerifyCertificatesBadOutcome:
+    share_certificates: dict[UserID, bytes] = {}
+    try:
+        brief_certificate = ShamirRecoveryBriefCertificate.verify_and_load(
+            setup.brief, author_verify_key, expected_author=author
+        )
+    except ValueError:
+        return VerifyCertificatesBadOutcome.INVALID_DATA
```
vxgmichel, line 106: The other methods try to stick to a consistent naming. Here's what certificate verification looks like for `realm_create`:
```python
class RealmCreateValidateBadOutcome(BadOutcomeEnum):
    INVALID_CERTIFICATE = auto()
    TIMESTAMP_MISMATCH = auto()
    INVALID_ROLE = auto()
    USER_ID_MISMATCH = auto()


def realm_create_validate(
    now: DateTime,
    expected_author: DeviceID,
    author_verify_key: VerifyKey,
    realm_role_certificate: bytes,
) -> RealmRoleCertificate | TimestampOutOfBallpark | RealmCreateValidateBadOutcome:
```

That means the following changes:
- rename `VerifyCertificatesBadOutcome` to `ShamirAddRecoverySetupValidateBadOutcome`
- rename `verify_certificates` to `shamir_add_recovery_setup_validate`
- add and manage `TimestampOutOfBallpark` inside the validate method

libparsec/crates/protocol/schema/authenticated_cmds/shamir_create_shared_recovery_device.json5
```diff
@@ -0,0 +1,71 @@
+[
+    {
+        "major_versions": [
+            4
+        ],
+        "req": {
+            "cmd": "shamir_recovery_setup",
```
vxgmichel, line 7: File name (`shamir_create_shared_recovery_device`) should match the command name

server/parsec/components/shamir.py
```diff
@@ -0,0 +1,130 @@
...
+    authenticated_cmds,
+)
+from parsec.api import api
+from parsec.client_context import AuthenticatedClientContext
+from parsec.types import BadOutcomeEnum
+
+
+class BaseShamirComponent:
+    @api
+    async def create_shared_recovery_device(
```
vxgmichel, line 23: Should be called `api_<name of the command>`, so `api_shamir_recovery_setup` for consistency with other routes

server/parsec/components/memory/shamir.py
```diff
@@ -0,0 +1,89 @@
...
+        self._data = data
+        self._event_bus = event_bus
+
+    async def remove_recovery_setup(
+        self,
+        organization_id: OrganizationID,
+        author: UserID,
+    ) -> None:
+        # TODO update after https://github.com/Scille/parsec-cloud/issues/7364
+        self._data.organizations[organization_id].shamir_setup.pop(author)
```
vxgmichel, line 29: This method doesn't perform any of the expected checks.

Maybe merge it with `add_recovery_setup` (and rename to `shamir_recovery_setup`) to factorize the checks.

You could also duplicate the checks, they're already duplicated all around the code base anyway.

vxgmichel (changes requested):
A bit of naming to change, LGTM otherwise :+1:

server/parsec/components/memory/shamir.py
AureliaDolo: and nor brief nor shares are scoped because python :snake:

vxgmichel (changes requested):
Ooops I missed a couple of things about error handling in my first reviews :)

libparsec/crates/protocol/tests/authenticated_cmds/v4/shamir_recovery_setup.rs
```diff
@@ -0,0 +1,236 @@
...
+
+fn encode_and_format_with_70_width<T: AsRef<[u8]>>(data: T) -> String {
+    hex::encode(data)
+        .as_bytes()
+        .chunks(70)
+        .map(std::str::from_utf8)
+        .collect::<Result<Vec<&str>, _>>()
+        .unwrap()
+        .join(" \\\n")
+}
```
AureliaDolo, line 36: This could also be used in other tests

mmmarcos: Indeed! Most, if not all, tests follow the same pattern. Can you create an issue for that?

server/tests/api_v4/authenticated/test_shamir_recovery_setup.py
```diff
@@ -0,0 +1,409 @@
...
+
+    # TODO dump
+
+
+async def test_authenticated_shamir_recovery_setup_shamir_setup_already_exists(
+    coolorg: CoolorgRpcClients, with_postgresql: bool
+) -> None:
+    # Setup previous shamir
+    if with_postgresql:
+        pytest.skip("TODO: postgre not implemented yet")
```
mmmarcos, line 54: I'd use `xfail` which usually means "this should fail because something is not implemented _yet_". `skip` is more for permanent failures (e.g. "this would _always_ fail on Windows platform").

Also, note that you can specify a condition with [`skipif(condition, reason)`](https://docs.pytest.org/en/6.2.x/skipping.html#id1) and [`xfail(condition, reason)`](https://docs.pytest.org/en/6.2.x/skipping.html#condition-parameter) 

So:
```suggestion
    pytest.xfail(with_postgresql, "TODO: postgre not implemented yet")
```

mmmarcos (approved):
Great job for this first PR! :100: :clap: 
I left some comments, but nothing major.

libparsec/crates/protocol/tests/authenticated_cmds/v4/shamir_recovery_setup.rs
```diff
@@ -0,0 +1,236 @@
...
+
+pub fn rep_shamir_setup_already_exists() {
+    // Generated from Rust implementation (Parsec v3.0.0-b.6+dev 2024-06-06)
+    // Content:
+    //   "status": "shamir_setup_already_exist"
+    test_de_serialization!(
+        "81a6737461747573bb7368616d69725f73657475705f616c72656164795f6578697374 \
+73",
```
FirelightFlagboy, line 44:
```suggestion
        "81a6737461747573bb7368616d69725f73657475705f616c72656164795f6578697374"
        "73",
```

Invalid per: https://github.com/Scille/parsec-cloud/blob/master/docs/rfcs/1009-hexstring-format.md

libparsec/crates/protocol/tests/authenticated_cmds/v4/shamir_recovery_setup.rs
```diff
@@ -0,0 +1,236 @@
...
+        authenticated_cmds::shamir_recovery_setup::Rep::load
+    );
+}
+pub fn rep_share_recipient_not_in_brief() {
+    // Generated from Rust implementation (Parsec v3.0.0-b.6+dev 2024-06-06)
+    // Content:
+    //   "status": "share_recipient_not_in_brief"
+    test_de_serialization!(
+        "81a6737461747573bc73686172655f726563697069656e745f6e6f745f696e5f627269 \
+        6566",
```
FirelightFlagboy, line 86:
```suggestion
        "81a6737461747573bc73686172655f726563697069656e745f6e6f745f696e5f627269"
        "6566",
```

libparsec/crates/protocol/tests/authenticated_cmds/v4/shamir_recovery_setup.rs
```diff
@@ -0,0 +1,236 @@
...
+        authenticated_cmds::shamir_recovery_setup::Rep::load
+    );
+}
+pub fn rep_duplicate_share_for_recipient() {
+    // Generated from Rust implementation (Parsec v3.0.0-b.6+dev 2024-06-06)
+    // Content:
+    //   "status": "duplicate_share_for_recipient"
+    test_de_serialization!(
+        "81a6737461747573bd6475706c69636174655f73686172655f666f725f726563697069 \
+656e74",
```
FirelightFlagboy, line 97:
```suggestion
        "81a6737461747573bd6475706c69636174655f73686172655f666f725f726563697069"
	"656e74",
```

libparsec/crates/protocol/tests/authenticated_cmds/v4/shamir_recovery_setup.rs
```diff
@@ -0,0 +1,236 @@
...
+        authenticated_cmds::shamir_recovery_setup::Rep::load
+    );
+}
+pub fn rep_author_included_as_recipient() {
+    // Generated from Rust implementation (Parsec v3.0.0-b.6+dev 2024-06-06)
+    // Content:
+    //   "status": "author_included_as_recipient"
+    test_de_serialization!(
+        "81a6737461747573bc617574686f725f696e636c756465645f61735f72656369706965 \
+6e74",
```
FirelightFlagboy, line 108:
```suggestion
        "81a6737461747573bc617574686f725f696e636c756465645f61735f72656369706965"
	"6e74",
```

libparsec/crates/protocol/tests/authenticated_cmds/v4/shamir_recovery_setup.rs
```diff
@@ -0,0 +1,236 @@
...
+        authenticated_cmds::shamir_recovery_setup::Rep::load
+    );
+}
+pub fn rep_missing_share_for_recipient() {
+    // Generated from Rust implementation (Parsec v3.0.0-b.6+dev 2024-06-06)
+    // Content:
+    //   "status": "missing_share_for_recipient"
+    test_de_serialization!(
+        "81a6737461747573bb6d697373696e675f73686172655f666f725f726563697069656e \
+74",
```
FirelightFlagboy, line 119:
```suggestion
        "81a6737461747573bb6d697373696e675f73686172655f666f725f726563697069656e"
	"74",
```

libparsec/crates/protocol/tests/authenticated_cmds/v4/shamir_recovery_setup.rs
```diff
@@ -0,0 +1,236 @@
...
+        authenticated_cmds::shamir_recovery_setup::Rep::load
+    );
+}
+pub fn rep_threshold_greater_than_total_shares() {
+    // Generated from Rust implementation (Parsec v3.0.0-b.6+dev 2024-06-06)
+    // Content:
+    //   "status": "threshold_greater_than_total_shares"
+    test_de_serialization!(
+        "81a6737461747573d9237468726573686f6c645f677265617465725f7468616e5f746f \
+74616c5f736861726573",
```
FirelightFlagboy, line 130:
```suggestion
        "81a6737461747573d9237468726573686f6c645f677265617465725f7468616e5f746f"
	"74616c5f736861726573",
```

libparsec/crates/protocol/tests/authenticated_cmds/v4/shamir_recovery_setup.rs
```diff
@@ -0,0 +1,236 @@
...
+        ballpark_client_late_offset: 32.,
+        client_timestamp: dt,
+        server_timestamp: dt,
+    };
+    test_de_serialization!(
+        "85a6737461747573b974696d657374616d705f6f75745f6f665f62616c6c7061726bbc \
+62616c6c7061726b5f636c69656e745f6561726c795f6f6666736574cb404000000000 \
+0000bb62616c6c7061726b5f636c69656e745f6c6174655f6f6666736574cb40400000 \
+00000000b0636c69656e745f74696d657374616d70d70141d26580b4800000b0736572 \
+7665725f74696d657374616d70d70141d26580b4800000",
```
FirelightFlagboy, line 156:
```suggestion
        "85a6737461747573b974696d657374616d705f6f75745f6f665f62616c6c7061726bbc"
	"62616c6c7061726b5f636c69656e745f6561726c795f6f6666736574cb404000000000"
	"0000bb62616c6c7061726b5f636c69656e745f6c6174655f6f6666736574cb40400000"
	"00000000b0636c69656e745f74696d657374616d70d70141d26580b4800000b0736572"
	"7665725f74696d657374616d70d70141d26580b4800000",
```

libparsec/crates/protocol/tests/authenticated_cmds/v4/shamir_recovery_setup.rs
```diff
@@ -0,0 +1,236 @@
...
+    // Content:
+    //   "status": "require_greater_timestamp"
+    //   "strictly_greater_than": "2009-02-13T23:31:30Z"
+    //
+    let req = authenticated_cmds::shamir_recovery_setup::Rep::RequireGreaterTimestamp {
+        strictly_greater_than: DateTime::from_f64_with_us_precision(1234567890.),
+    };
+    test_de_serialization!(
+        "82a6737461747573b9726571756972655f677265617465725f74696d657374616d70b5 \
+        7374726963746c795f677265617465725f7468616ed70141d26580b4800000",
```
FirelightFlagboy, line 172:
```suggestion
        "82a6737461747573b9726571756972655f677265617465725f74696d657374616d70b5"
        "7374726963746c795f677265617465725f7468616ed70141d26580b4800000",
```

vxgmichel, line 172: It seems like some payloads are indented and some are not.

server/parsec/components/shamir.py
```diff
@@ -0,0 +1,195 @@
...
+            setup.brief, author_verify_key, expected_author=author
+        )
+    except ValueError:
+        return ShamirAddRecoverySetupValidateBadOutcome.BRIEF_INVALID_DATA
+
+    match timestamps_in_the_ballpark(brief_certificate.timestamp, DateTime.now()):
+        case TimestampOutOfBallpark() as error:
+            return error
+        case _:
+            pass
```
vxgmichel, line 169: The timestamp for share certificates should also be checked.

We might also want to make sure that all those certificates share the same timestamp. That would probably help with the topic stuff.

vxgmichel (approved):
A couple of minor comments, LGTM otherwise :+1:

libparsec/crates/protocol/tests/authenticated_cmds/v4/shamir_recovery_setup.rs
AureliaDolo: The strings are 70 col wide. The issue I have is that the kind of input accepted by the hex macro is not the way multi line strings are formatted. So I can try to make the test_de_serialization macro take the same kind of input as hex! for $encoded, but I'm not sure it's worth it. What do you think ?

The goal of test_de_serialization is that when you use it and the test fails, it outputs you the expected hex nicely broken down.

libparsec/crates/protocol/tests/authenticated_cmds/v4/shamir_recovery_setup.rs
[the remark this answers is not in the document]
FirelightFlagboy: That's my point on my recent review, I prefer the example you have provided.
It's easier to work with it in a multi-cursor context + the characters are aligned on each lines

libparsec/crates/protocol/tests/authenticated_cmds/v4/shamir_recovery_setup.rs
AureliaDolo: Thats a known issue that the formatting inside macros is not ideal (see [here](https://github.com/rust-lang/rustfmt/issues/8)). I think I've seen a few fmt unstable options that could help with that.

Anyway, I could have been more careful ^^

server/parsec/components/shamir.py
AureliaDolo: Nice catch!

AureliaDolo: So if we check that each share timestamp is the same as the brief timestamp, it should be enough ?

libparsec/crates/protocol/tests/authenticated_cmds/v4/shamir_recovery_setup.rs
```diff
@@ -0,0 +1,236 @@
...
+            "Expected hex is:\n\"{}\"",
+            encode_and_format_with_70_width($expected.dump().unwrap())
+        );
+
+        // Also test serialization round trip
+        let raw2 = data.dump().unwrap();
+        let data2 = $load(&raw2).unwrap();
+        p_assert_eq!(data2, $expected);
+    };
+}
```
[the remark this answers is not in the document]
AureliaDolo: I was trying to write de/serialization ^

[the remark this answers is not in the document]
vxgmichel: Oh I see. As far as I can tell, "deserialization" is usually written as a single word.

server/parsec/components/shamir.py
vxgmichel: Yes, although you might have to yet another error status for that.

libparsec/crates/protocol/tests/authenticated_cmds/v4/shamir_recovery_setup.rs
AureliaDolo: See #7549

libparsec/crates/protocol/tests/authenticated_cmds/v4/shamir_recovery_setup.rs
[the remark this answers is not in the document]
AureliaDolo: @vxgmichel it is for deserialization and serialization

@mmmarcos thanks for the suggestion, I'll take it

vxgmichel (approved):
LGTM :+1:

FirelightFlagboy (approved):

touilleMan (changes requested):

libparsec/crates/protocol/Cargo.toml
```diff
@@ -25,6 +25,7 @@ futures = { workspace = true, features = ["std", "async-await", "executor"] }
 bytes = { workspace = true, features = ["serde"] }
 
 [dev-dependencies]
+hex = { workspace = true }
```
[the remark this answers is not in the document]
AureliaDolo: I'm using hex::encode in `encode_and_format_with_70_width`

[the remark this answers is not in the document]
touilleMan: Ah I missed that !

server/parsec/components/shamir.py
```diff
@@ -0,0 +1,195 @@
...
+                        )
+                    )
+
+    async def remove_recovery_setup(
+        self,
+        organization_id: OrganizationID,
+        author: UserID,
+    ) -> (
+        None
+        | ShamirAddRecoverySetupStoreBadOutcome
```
touilleMan, line 100: Return type differs here and in the memory implementation, I guess you should remove the errors that are not actually used from here

touilleMan: Also you should introduce a `ShamirRemoveRecoverySetupStoreBadOutcome` instead of using `ShamirAddRecoverySetupStoreBadOutcome` in `remove_recovery_setup` (to stays consistent with the rest of the code base)

Alternatively, you could introduce a special error (named something like `ShamirAddOrRemoveRecoverySetupStoreBadOutcome`) to be used by both. The point being to have an error with a name making it clear it is used in multiple places (and hence care must be taken if another error case is added to this error type)

touilleMan (approved):
:+1:

PR #7339, opened 2024-05-17
Fix not a git repo to update nightly tag

FirelightFlagboy:
Checkout the repository on the last job if we need to update the nightly tag.

### DIFF
--- a/.github/workflows/releaser.yml
+++ b/.github/workflows/releaser.yml
@@ -160,6 +160,13 @@ jobs:
           GH_ARGS: --repo=${{ github.server_url}}/${{ github.repository }}
           GH_TOKEN: ${{ github.token }}
 
+      - name: Checkout the repository to update the nightly tag
+        if: env.NIGHTLY_RELEASE == 'true'
+        uses: actions/checkout@44c2b7a8a4ea60a981eaca3cf939b5f4305c123b # pin v4.1.5
+        with:
+          ref: ${{ needs.version.outputs.commit_sha }}
+        timeout-minutes: 5
+
       - name: Update the nightly tag
         if: env.NIGHTLY_RELEASE == 'true'
         run: |


.github/workflows/releaser.yml
```diff
@@ -160,6 +160,13 @@ jobs:
           GH_ARGS: --repo=${{ github.server_url}}/${{ github.repository }}
           GH_TOKEN: ${{ github.token }}
 
+      - name: Checkout the repository to update the nightly tag
```
touilleMan, line 163: I'd rather merge this with the next item given they work closely together (and they are both behind the same `if env.NIGHTLY_RELEASE`)

.github/workflows/releaser.yml
```diff
@@ -160,6 +160,13 @@ jobs:
...
+        if: env.NIGHTLY_RELEASE == 'true'
+        uses: actions/checkout@44c2b7a8a4ea60a981eaca3cf939b5f4305c123b # pin v4.1.5
```
touilleMan, line 165: I don't understand how the next command (line 173&174) worked before this PR since the repo wasn't checked out and the command does a `git tag` & `git push` :thinking:

.github/workflows/releaser.yml
FirelightFlagboy: The action checkout also manage the authentication to the github repo using the provided github token.
Using the action is trivial, so I will keep it that way

.github/workflows/releaser.yml
FirelightFlagboy: Before this PR those lines don't work as we were not in a git repository.

.github/workflows/releaser.yml
touilleMan: Yeah that makes sense :+1:

touilleMan (approved):

PR #7384, opened 2024-05-22
Handle in client all custom HTTP error codes from server

touilleMan:


libparsec/crates/client_connection/src/authenticated_cmds/sse.rs
```diff
@@ -120,6 +120,7 @@ pub(crate) fn handle_sse_error(error: EventStreamError<reqwest::Error>) -> Conne
     }
 }
 
+// TODO: This custom error is always converted into `ConnectionError`, remove it ?
```
mmmarcos, line 123: Maybe create an issue?

mmmarcos (approved):
LGTM

PR #7389, opened 2024-05-23
[MS] Updated import manager to file operation manager

Max-7:
The idea is to re-use the import manager and menu for other file operations like copy and move. This is the first step, it doesn't include copy or move yet, it shouldn't change anything in terms of UX/UI.

- [X] Keep changes in the pull request as small as possible
- [X] Ensure the commit history is sanitized
- [X] Give a meaningful title to your PR
- [X] Describe your changes

PR #7395, opened 2024-05-24
Implementing entry move operation

### LINKED ISSUE #7379: GUI: move an entry
depends on #5820

Max-7:
Closes #7379
Supersedes #7389

So. When moving entries, we'll check if entries of the same name exist in the destination. If some do, we ask the user if they'd like to replace or skip them.
In some weird cases where the entries were not present when the user started the move but were added later (by a sync for example), we rename the file, adding `(n)` with `1 <= n <= 9` at the end of the file name.

We make use of the import manager that was renamed for the occasion to `fileOperationsManager`. The idea is to treat move and copy as we do imports: letting the manager know what operations to perform, and letting it do the orchestration. Move and copy are also shown in the `FileOperationMenu` (formely `FileImportMenu`).

I've noticed some files disapearing sometimes (only the display, it can be fixed by navigating out and in the current folder again) while performing a move. I'm attributing this to the 10s refresh since I've not noticed it happenning when the refresh was turned off. I'm not wasting too much time investigating right now, I'm betting on the fact that it will disapear once we have proper events.

Good luck reviewing this.

- [X] Describe your changes
- [X] Link any dependent pull request in the description


client/src/views/files/FileOperationMenu.vue
```diff
@@ -372,3 +409,4 @@ async function onImportEvent(state: ImportState, importData?: ImportData, stateD
   }
 }
 </style>
+@/services/fileOperationManager
```
NicoTuxx, line 412: remove this

NicoTuxx (changes requested):

client/src/locales/fr-FR.json
```diff
@@ -455,7 +452,10 @@
         "deleteMultipleNo": "Conserver les éléments",
         "copySelectFolderTitle": "Copier un élément | Copier {count} éléments",
         "moveSelectFolderTitle": "Déplacer un élément | Déplacer {count} éléments",
-        "moveSuccess": "L'élément a été déplacé dans le dossier choisi. | Tous les éléments ont été déplacés dans le dossier choisi.",
+        "moveAlreadyExistTitle": "Remplacer ou ignorer l'élément | Remplacer ou ignorer les éléments",
+        "moveAlreadyExistQuestion": "Un élément nommé `{file}` existe déjà dans le dossier de destination. Que souhaitez-vous faire ? | Plusieurs éléments du même nom existe déjà dans le dossier de destination. Que souhaitez-vous faire ?",
```
Ironicbay, line 456:
```suggestion
        "moveAlreadyExistQuestion": "Un élément nommé `{file}` existe déjà dans le dossier de destination. Que souhaitez-vous faire ? | Plusieurs éléments du même nom existent déjà dans le dossier de destination. Que souhaitez-vous faire ?",
```

Ironicbay (changes requested):
FoldersPage breaks when importing files: only the currently importing files are shown, then a long empty screen before the list refreshes. Functionally it seems fine though. 

https://github.com/Scille/parsec-cloud/assets/41511029/4eb9a918-14b4-4567-a4e0-8817a1595e98

Ironicbay (approved):

NicoTuxx (approved):

PR #7413, opened 2024-05-27
Improve reproducibility of `gen_protocol_typings` script by sorting some values

### LINKED ISSUE #7411: Python RPC fixture file is not sorted
The file `server/tests/common/rpc.py` is auto-generated but its content is not sorted.

That cause a lot of change where function/method just change in position.

It would be best to have those sorted during the generation.
             
_Originally posted by @FirelightFlagboy in https://github.com/Scille/parsec-cloud/pull/7324#discussion_r1616083887_

FirelightFlagboy:
Closes #7411


mmmarcos (approved):
`misc/gen_protocol_typings.py` LGTM

PR #7418, opened 2024-05-28
[MS] Log libparsec calls on failure

Max-7:
The result of all calls to libparsec gets logged if it fails.

- [X] Describe your changes


NicoTuxx (approved):

NicoTuxx:
Maybe decorators would be interesting in this case

PR #7419, opened 2024-05-28
Fix certif ops validation

touilleMan:


libparsec/crates/client/src/certif/store.rs
```diff
@@ -673,6 +684,9 @@ macro_rules! impl_read_methods {
             )
         }
 
+        // TODO: Provide a `get_last_user_realms_roles` that, for a given user, returns
+        ///      the last valid realm role certificates for each realm he is part of.
```
mmmarcos, line 688:
```suggestion
        // TODO: Provide a `get_user_last_realm_roles` that, for a given user, returns
        //       the last valid realm role certificate for each of its realms.
```

libparsec/crates/client/src/certif/store.rs
```diff
@@ -708,6 +729,9 @@ macro_rules! impl_read_methods {
             )
         }
 
+        // TODO: Provide a `get_last_realm_roles` that, for a given realm, returns
+        ///      the last valid realm role certificates for each user part of it.
```
mmmarcos, line 733:
```suggestion
        // TODO: Provide a `get_realm_last_user_roles` that, for a given realm, returns
        //       the last valid realm role certificate for each of its users.
```

mmmarcos (approved):
LGTM, just some suggestions about code comments

libparsec/crates/client/src/certif/add.rs
```diff
@@ -1349,23 +1349,25 @@ async fn check_realm_role_certificate_consistency(
         // 2.b) The realm already exists, so certificate cannot be self-signed and
         // author must have a sufficient role.
 
-        let author_current_role = match realm_current_roles
+        let author_current_role = match realm_roles
             .iter()
+            .rev()
```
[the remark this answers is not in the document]
touilleMan: :+1:

libparsec/crates/client/src/certif/store.rs
```diff
@@ -655,6 +659,13 @@ macro_rules! impl_read_methods {
         }
 
         #[allow(unused)]
+        /// Return all certificates involving the user.
+        ///
+        /// ⚠️ This method returns *all* certificates, including the ones that have been
+        /// superseded ! (i.e. you should iter from the last to the first certificate to
+        /// get the current role for a given realm).
```
vxgmichel, line 666: Maybe this logic should be provided as a helper method. I can see that the following logic appears several times:
```rs
realm_roles
            .iter()
            .rev()            
            .find(|role| &role.user_id == cooked.author.user_id())
            .and_then(|role| role.role)
```

vxgmichel (approved):
A suggestion, LGTM otherwise :+1:

libparsec/crates/client/src/certif/store.rs
touilleMan: :+1:

libparsec/crates/client/src/certif/store.rs
touilleMan: :+1:

libparsec/crates/client/src/certif/store.rs
touilleMan: yes, I wanted to provide a `store.get_realm_last_user_roles()` for this purpose, but that was a bit tricky to correctly implement (it needs to use group by in the query to the sqlite database). So I went the KISS way instead.

I will add an issue about this improvement

PR #7431, opened 2024-05-29
Add proper integrity checking for file manifest content

### LINKED ISSUE #6599: Proper error handling for file manifest integrity
At the moment, file manifest integrity is asserted using `LocalFileManifest::assert_integrity`.

There should be proper error handling and a dedicated error type for that

vxgmichel:
Close #6599

This PR add the following checks:
1. Check remote file manifest content integrity in `CertificateOps::validate_child_manifest`
2. ~Check content integrity after remote file manifest conversion in `outbound_sync_file`~ (just a sanity check, panicking if something goes wrong)
3. Check content integrity after deserializing local file manifest in `LocalFileManifest::decrypt_and_load`

Remaining:
- [x] Add tests for check 1
- [x] ~Add tests for check 2~
- [x] Add tests for check 3

A couple questions to the reviewers:

1. Is it worth it to move this code:
    ```rust
                // Check that the chunk is internally consistent
                if chunk.start >= chunk.stop.get()
                    || chunk.raw_offset > chunk.start
                    || chunk.stop.get() > chunk.raw_offset + chunk.raw_size.get()
                {
                    return Err(DataError::InvalidFileContent);
                }
    ```
    to a `Chunk::check_content_integrity` method?

2. Is it worth it to be more precise than the added `DataError::InvalidFileContent`?
3. Should the following (existing) check:
    https://github.com/Scille/parsec-cloud/blob/3954287f2bce276440b1a69978446444eb1230b1/libparsec/crates/client/src/certif/manifest_validate.rs#L339-L354
    be moved in `ChildManifest::check_content_integrity` along with `FileManifest::check_content_integrity`?


libparsec/crates/client/tests/unit/certif/validate_child_manifest.rs
```diff
@@ -172,7 +174,7 @@ async fn corrupted(#[values("dummy", "parent_pointing_on_itself")] kind: &str, e
                 timestamp: now,
                 id: child_id,
                 parent: child_id,
-                version: 1,
```
touilleMan, line 175: Remote manifest version is not supposed to be zero:
- A remote manifest is synchronized in the server, so it is accepted only if it version is > 0
- We don't enforce this in the type though given we use remote manifest within the local manifest where it can have `version == 0` if the local manifest is a placeholder.

Here the value is expected to be a valid one (given the error kind `parent_pointing_on_itself` doesn't mention anything about an invalid version), so I'd rather stick to `1`

libparsec/crates/client/tests/unit/certif/validate_child_manifest.rs
```diff
@@ -196,6 +198,165 @@ async fn corrupted(#[values("dummy", "parent_pointing_on_itself")] kind: &str, e
...
+            child_id,
+            &alice.device_id,
+            0,
+            now,
+            &encrypted,
+        )
+        .await
+        .unwrap_err();
+
+    match kind {
```
touilleMan, line 216: Small comment on style: in the tests I try to have a single `match kind` statement so that it can always have the same pattern:
- every case is matched
- unknown guard
- every expected value is generated there and returned from the match (so here we would have an `let expected_data_error: DataError = match kind { ... };` 

My main fear is to open the door to non exhaustive match (typically when somewhere in the code we need to do something for kind A and B but not for C and D, then we start have `match kind { ... }` without the unknown guard, which can then be copied from one test to another by new dev...)

libparsec/crates/client/tests/unit/workspace/file_operations_stateful.rs
```diff
@@ -160,7 +160,10 @@ impl StateMachineTest for FileOperationStateMachine {
         ref_state: &<Self::Reference as ReferenceStateMachine>::State,
     ) {
         // 1. Manifest integrity
-        state.manifest.assert_integrity();
+        state
+            .manifest
+            .check_content_integrity()
+            .expect("Manifest integrity");
```
touilleMan, line 166: I'd stick with `unwrap()` here given we only use `.unwrap()` in the tests so it is obvious this is a test assert  (while using `.expect()` can lead to thing the test itself is broken)

libparsec/crates/types/src/error.rs
```diff
@@ -75,6 +75,12 @@ pub enum DataError {
 
     #[error("Invalid version: expected `{expected}`, got `{got}`")]
     UnexpectedVersion { expected: u32, got: u32 },
+
+    #[error("Invalid file content")]
+    InvalidFileContent,
+
+    #[error("Invalid file content")]
```
touilleMan, line 82: Error message to change ?

libparsec/crates/types/src/error.rs
```diff
@@ -75,6 +75,12 @@ pub enum DataError {
 
     #[error("Invalid version: expected `{expected}`, got `{got}`")]
     UnexpectedVersion { expected: u32, got: u32 },
+
+    #[error("Invalid file content")]
+    InvalidFileContent,
```
touilleMan, line 80: `InvalidFileContent` seems very broad... but still restricted to file (so not *that* broad)

What about `InconsistentContent` ?

Or, if we want to have a specific error for file (but I'd rather not since we didn't have errors for specific manifests so far), `InconsistentFileContent` ? (given invalid is used everywhere we don't know much about the cause of the error, while inconsistent show that different data within the manifest are in conflict)

libparsec/crates/types/src/local_manifest.rs
```diff
@@ -373,7 +415,10 @@ impl LocalFileManifest {
             size: self.size,
             blocksize: self.blocksize,
             blocks,
-        })
+        };
+        // The content integrity check is also done on the remote manifest, just in case
+        manifest.check_content_integrity()?;
```
touilleMan, line 420: use `.expect(...)` since this is a sanity check ?

libparsec/crates/types/src/local_manifest.rs
```diff
@@ -336,33 +370,41 @@ impl LocalFileManifest {
...
 
-    pub fn to_remote(
-        &self,
-        author: DeviceID,
-        timestamp: DateTime,
-    ) -> Result<FileManifest, &'static str> {
-        self.assert_integrity();
+    pub fn to_remote(&self, author: DeviceID, timestamp: DateTime) -> DataResult<FileManifest> {
+        // Make sure we don't upload an invalid manifest
+        self.check_content_integrity()?;
```
touilleMan, line 385: use `.expect(...)` since this is a sanity check ?

libparsec/crates/types/src/local_manifest.rs
```diff
@@ -336,33 +370,41 @@ impl LocalFileManifest {
...
         manifest
     }
 
-    pub fn to_remote(
-        &self,
-        author: DeviceID,
-        timestamp: DateTime,
-    ) -> Result<FileManifest, &'static str> {
-        self.assert_integrity();
+    pub fn to_remote(&self, author: DeviceID, timestamp: DateTime) -> DataResult<FileManifest> {
```
touilleMan, line 383: I think we should return another error type than `DataError` here:
- the local manifest we work on is expected to be valid (hence if it integrity is wrong we should just panic)
- same thing for the remote manifest we have just generated
- in the end, the only error possible is if the local manifest is not reshaped

On top of that it would be good to avoid adding to `DataError` such a specific error (given the errors in `DataError` are themselves rather broad and related check done during deserialization, while here we are not doing deserialization...)

```rust
pub enum LocalFileManifestToRemoteError {
  #[error("Local file manifest needs reshape")]
  NeedReshape,
}
```

touilleMan (changes requested):

libparsec/crates/client/tests/unit/certif/validate_child_manifest.rs
vxgmichel: The problem is that this test wasn't actually testing that the internal error was actually corresponding to `parent_pointing_to_itself` (more specifically `DataError::Serialization`, which is a bit weird as well). So I improve the test and got the following error:
```
Diff < left / right > :
<InvalidManifest(
<    Corrupted {
<        realm: VlobID(
<            f0000000-0000-0000-0000-000000000002,
<        ),
<        vlob: VlobID(
<            a14d0099-ca88-447c-afd6-30383e1270ea,
<        ),
<        version: 0,
<        author: DeviceID(
<            "alice@dev1",
<        ),
<        timestamp: DateTime(
<            "2020-01-01T00:00:00Z",
<        ),
<        error: UnexpectedVersion {
<            expected: 0,
<            got: 1,
<        },
<    },
<)
>CertifValidateManifestError::InvalidManifest(boxed) if
>matches!(& * boxed, InvalidManifestError :: Corrupted { error, .. } if * *
>    error == DataError :: Serialization)
```
Since it said `UnexpectedVersion, expected 0 got 1` made the change without thinking about it. Isn't there something fishy here?

touilleMan:
> Is it worth it to move this code: [...] to a Chunk::check_content_integrity method?

I'd say yes, as it seems consistent with the logic brought by this PR (and it make it easier to test)
But I don't have a strong opinion on this, so go as you feel best ;-)

> Is it worth it to be more precise than the added DataError::InvalidFileContent?

Adding a `hint: String` field to this type could be a good idea (see what is done for `InvalidCertificateError` for instance.
I think it is worth it given how helpful it can be if we detect a bug !

> Should the following (existing) check: [...]  be moved in ChildManifest::check_content_integrity along with FileManifest::check_content_integrity?

I'd say yes given it brings consistency. On top of that it show we need to rename `InvalidFileContent` -> `InconsistentContent` given this error type would be use for both `check_content_integrity`

vxgmichel:
TODO (either in this PR or open an issue):

Reshape in `fd_flush` does not use `manifest.new_chunks` to build the new data to turn into a block:
https://github.com/Scille/parsec-cloud/blob/5da5a48b635e1c02e7486a0411a729540e1f83c8/libparsec/crates/client/src/workspace/transactions/fd_flush.rs#L128-L155

Compare with what's done in `fd_read`:
https://github.com/Scille/parsec-cloud/blob/5da5a48b635e1c02e7486a0411a729540e1f83c8/libparsec/crates/client/src/workspace/transactions/fd_read.rs#L82-L103

Also, the corresponding test should be added.

libparsec/crates/types/src/certif.rs
```diff
@@ -1001,13 +1001,32 @@ impl_unsecure_dump!(ShamirRecoveryBriefCertificate);
...
+    }
+
     pub fn verify_and_load(
         signed: &[u8],
         author_verify_key: &VerifyKey,
         expected_author: DeviceID,
     ) -> DataResult<Self> {
         let r = verify_and_load::<Self>(signed, author_verify_key)?;
 
+        r.check_data_integrity()?;
```
touilleMan, line 1028: I find it logical to have this check here.

But then there is no need to also have it in the `add_certificates_batch`. This is especially true given `add_certificates_batch` receive the certificates as `Bytes` and hence deserializes them itself (vs if `ShamirRecoveryBriefCertificate` object were provided to it and it then had to trust the caller on the fact those objects hasn't been manually modified after creation/deserialization)

On top of that, there is the remote manifests that now have a `check_data_integrity` method, but which is not called in their  `verify_and_load` method...

So my guess:
- Having a *public* `check_data_integrity` method is useful to plug on object that get modified across they lifetime (so only local manifest).
- All types should call their own `check_data_integrity` when deserializing (i.e. the "verify_and_load" means "verify the signature, verify the format by deserializing, and verify the integrity")
- We *may* want to add an assert on a call to `check_data_integrity` before each serialization as a sanity check.
- `FolderManifest` should expose `FolderManifest::verify_and_load_for_root` and `FolderManifest::verify_and_load_for_child` instead of a single `FolderManifest::verify_and_load`

touilleMan (approved):

libparsec/crates/types/src/certif.rs
```diff
@@ -70,6 +59,34 @@ pub enum UnsecureSkipValidationReason {
...
+        impl $name {
+            pub fn base_verify_and_load(
+                signed: &[u8],
+                author_verify_key: &VerifyKey,
+            ) -> DataResult<Self> {
+                let serialized = author_verify_key
+                    .verify(signed)
+                    .map_err(|_| DataError::Signature)?;
+
+                let result: Self = format_vx_load(serialized)?;
```
touilleMan, line 73: Should settle to use `format_vx_load` (like it is here) or ` $crate::serialization::format_vx_load` (like it is in `base_unsecure_load` ?

given this macro is exposed with `pub(super) use`, I'd rather use `$crate::...` is it won't break if the caller module remove its `use serialzation::format_vx_load`

libparsec/crates/types/src/certif.rs
```diff
@@ -235,8 +251,13 @@ pub struct UserCertificate {
 impl_unsecure_load!(UserCertificate -> CertificateSignerOwned);
 impl_unsecure_dump!(UserCertificate);
 impl_dump_and_sign!(UserCertificate);
+impl_base_load!(UserCertificate);
 
 impl UserCertificate {
+    pub fn check_data_integrity(&self) -> DataResult<()> {
```
touilleMan, line 257: I think this method should be private: right now it is unclear that this method is used internally by `impl_base_load` (and hence that there is no need to call it manually)

If possible, I'd say we should remove this method entirely and replace it by an inline in the macro call (this way it's very obvious when this code is used):
```rust
// Note the `check_data_integrity:` that helps document what this closure is for
impl_base_load!(RevokedUserCertificate, check_data_integrity: |data: RevokedUserCertificate| -> DataResult<()> { Ok(()) });
```

Note `macro_rules` support this by simply passing the closure as an `$expr`:
```rust
macro_rules! foo {
   (function: $function:expr) => {
       $function(42)
   };
}

fn main() {
    let ret = foo!(function: |i| i * 2);
    assert_eq!(ret, 84);
}
```

If this is not possible (typically the `check_data_integrity` is used after the deserialization step because the object gets modified during its lifetime), then it's still worth it to have an explicit link between the method and the macro:
```rust
impl RevokedUserCertificate {
    /// This object can be modified during its lifetime, hence we provide this method to run
    /// sanity checks once a modification occurs.
    pub fn check_data_integrity(&self) -> DataResult<()> {
        ...
    }
}
impl_base_load!(RevokedUserCertificate, check_data_integrity: RevokedUserCertificate::check_data_integrity);

File: libparsec/crates/types/src/certif.rs [hunk 12]
```diff
@@ -1151,11 +1225,29 @@ pub enum UnsecureAnyCertificate {
     SequesterAuthority(UnsecureSequesterAuthorityCertificate),
 }
 
+impl_base_load!(AnyCertificate);
+
 impl AnyCertificate {
+    fn check_data_integrity(&self) -> DataResult<()> {
+        match self {
+            AnyCertificate::User(c) => c.check_data_integrity(),
```
touilleMan, line 1233: If this is the reason `check_data_integrity` is declared as `pub(super)`, not that private fields can be accessed from within the module the structure belong to:

```
struct Foo{
    a: usize  // Not pub
}

impl Foo {
    fn b(self) -> usize { self.a * 2 }  // Also not pub
}

fn main() {
    let f = Foo { a: 1 };
    assert_eq!(f.b(), 2);
}
```

vxgmichel: [6731fa0](https://github.com/Scille/parsec-cloud/pull/7431/commits/6731fa051bb816f0bb76b1ff21d27139208df024)

libparsec/crates/types/src/certif.rs
vxgmichel: [6731fa0](https://github.com/Scille/parsec-cloud/pull/7431/commits/6731fa051bb816f0bb76b1ff21d27139208df024)

libparsec/crates/types/src/certif.rs
vxgmichel: [6731fa0](https://github.com/Scille/parsec-cloud/pull/7431/commits/6731fa051bb816f0bb76b1ff21d27139208df024)

vxgmichel: > If possible, I'd say we should remove this method entirely and replace it by an inline in the macro call (this way it's very obvious when this code is used): [...]

Isn't it less readable to have the integrity logic defined in the macro rather than in the struct impl?

> then it's still worth it to have an explicit link between the method and the macro

Is it tho? I can do it sure but I don't see the benefit.

vxgmichel: Here's how it looks: e2c33cc83c1c1cb57fee88a028697086f86adc25

touilleMan: > Isn't it less readable to have the integrity logic defined in the macro rather than in the struct impl?

Yes, it gets less readable pretty fast when the integrity logic gets bigger.
Hence the `impl_base_load!(RevokedUserCertificate, check_data_integrity: RevokedUserCertificate::check_data_integrity);` where there is a dedicated `check_data_integrity` method

> Is it tho? I can do it sure but I don't see the benefit.

If the function is private, the reader understand it is an implementation detail.
However if the function is public, the reader assume it is for external usage. The fact the function is used in a macro further hide the fact it is an internal detail.
On top of that we have to consider the fact a lot of other data structures have a similar `check_data_integrity`, which is often to be used outside of deserialization (e.g. `LocaFilelManifest`).

So in the end I'd say:
- If the function is private and the macro define in the same module,  no need to make the link between the two explict.
- Avoid `pub` as much as possible for this method, use `pub(crate)` or `pub(super)` if it is used from another module in the crate. This way we won't be puzzled about this method when used from `libparsec_client`.
- Add a comment to explain why it is not `pub`.

I would add something like:
```rust
/// This structure represent immutable data (as it is created once, signed, and never updated).
/// Hence this `check_data_integrity` is only used during deserialization (and also as sanity check
/// right before serialization) and not exposed publicly.
pub fn check_data_integrity(&self) -> DataResult<()> {
...
}
```

And for structures that can be modified:
```rust
/// This structure represent mutable data (it gets loaded from disk, updated, then stored back modified)
/// Hence this `check_data_integrity`'s main goal is during deserialization.
/// However it is also useful as sanity check:
/// - Right before serialization
/// - After any modification (hence the need for this method to be public)
pub fn check_data_integrity(&self) -> DataResult<()> {
...
}
```

touilleMan (approved):

PR #7442, opened 2024-05-31
Fix workspace share

touilleMan:


libparsec/crates/client/src/client/mod.rs
```diff
@@ -343,6 +346,10 @@ impl Client {
         workspace_rename::rename_workspace(self, realm_id, new_name).await
     }
 
+    /// Share the workspace with another user, this function requires to be online.
+    ///
+    /// If the workspace is not bootstrapped, this function will try to bootstrap
+    /// it (but this may fail the current user is not an OWNER role).
```
mmmarcos, line 352:
```suggestion
    /// it (this may fail if the current user does not have the OWNER role).
```

libparsec/crates/client/src/client/workspace_rename.rs
```diff
@@ -66,6 +66,10 @@ pub async fn rename_workspace(
         .map_err(|e| match e {
             CertifBootstrapWorkspaceError::Offline => ClientRenameWorkspaceError::Offline,
             CertifBootstrapWorkspaceError::Stopped => ClientRenameWorkspaceError::Stopped,
+            // Note if our author is not allowed, it means the workspace has been shared
+            // with us... and hence the bootstrap has been done already !
+            // However this also means there is no point in trying to do the rename
+            // given it requires the same kind of authorizations.
```
mmmarcos, line 72:
```suggestion
            // Note that in case of author not allowed, it means that the workspace
            // has already been bootstrapped (since it is being shared with us).
            // However this also means there is no point in trying to rename it
            // since this will require the same kind of authorizations.
```

libparsec/crates/client/tests/unit/client/share_workspace.rs
```diff
@@ -1,188 +1,401 @@
...
+use std::collections::HashMap;
+
+use libparsec_tests_fixtures::prelude::*;
+use libparsec_types::prelude::*;
+
+use super::utils::client_factory;
+use crate::{ClientShareWorkspaceError, WorkspaceUserAccessInfo};
+
+#[parsec_test(testbed = "coolorg", with_server)]
+async fn ok_with_author_having_changed_it_role(
```
mmmarcos, line 12:
```suggestion
async fn ok_with_author_having_changed_its_role(
```

mmmarcos (approved):
LGTM, some minor comments

libparsec/crates/client/src/client/mod.rs
touilleMan: :+1:

libparsec/crates/client/src/client/workspace_rename.rs
touilleMan: :+1:

PR #7443, opened 2024-05-31
Fix server logger.info_with_debug_extra to ensure no extra debug info are named `event

touilleMan:


server/parsec/logging.py
```diff
@@ -79,6 +79,7 @@ def info_with_debug_extra(
         def info_with_debug_extra(
             self, event: str, debug_extra: dict[str, Any], **kwargs: Any
         ) -> None:
+            assert "event" not in debug_extra
```
vxgmichel, line 82: Maybe rename `event` to `_event` to avoid crashing?

vxgmichel (approved):
A single comment, LGTM otherwise :+1:

server/parsec/logging.py
touilleMan: :+1:

PR #7445, opened 2024-05-31
Remove Parsec addr from scheme

touilleMan:
- Prevent compatibility issue with data if we change the scheme.
- Simplify the codebase


libparsec/crates/types/tests/unit/pki.rs
```diff
@@ -374,6 +226,7 @@ fn serde_local_pending_enrollment(
     #[case] generate_expected: Box<dyn FnOnce(&Device) -> LocalPendingEnrollment>,
 ) {
     let expected = generate_expected(alice);
+    println!("{:?}", expected.dump());
```
mmmarcos, line 229: debug print?

FirelightFlagboy (approved):

mmmarcos (approved):
LGTM

libparsec/crates/types/tests/unit/pki.rs
touilleMan: removed :+1:

PR #7372, opened 2024-05-22
Rework DateTime to use integer instead of float \o/

touilleMan:
depends on #7107

TODO:
- [ ] protocol tests
- [ ] types tests
- [ ] bindings
- [ ] web storage

PR #7514, opened 2024-06-05
Trigger the wasm tests more frequently

FirelightFlagboy:
The wasm tests will be run when the toolchain is changed or the dependencies is updated


touilleMan (approved):
wait for #7372  to be merged before merging this one (to avoid caching build on ubuntu 20.04 which will cause issue)

PR #7516, opened 2024-06-05
Harden cache entry with runner version

FirelightFlagboy:
Add runner version to:

- Rust cache entry
- libparsec cache entry

### DIFF
--- a/.github/actions/system-info/action.yml
+++ b/.github/actions/system-info/action.yml
@@ -1,0 +1,24 @@
+name: system-info
+description: Get system information
+
+outputs:
+  platform:
+    description: The platform name (e.g. linux, win32, darwin)
+    value: ${{ steps.system-info.outputs.platform }}
+  id:
+    description: The system ID (e.g. ubuntu-22.04, windows-10, ...)
+    value: ${{ steps.system-info.outputs.id }}
+  os:
+    description: The OS name (e.g. ubuntu, windows, macos)
+    value: ${{ steps.system-info.outputs.os }}
+  release:
+    description: The release version of the OS (e.g. 22.04, 10, 12.7.5, ...)
+    value: ${{ steps.system-info.outputs.release }}
+
+runs:
+  using: composite
+  steps:
+    - name: Get system info
+      id: system-info
+      shell: bash
+      run: python $GITHUB_ACTION_PATH/system-info.py | tee $GITHUB_OUTPUT
--- a/.github/actions/system-info/system-info.py
+++ b/.github/actions/system-info/system-info.py
@@ -1,0 +1,33 @@
+# Parsec Cloud (https://parsec.cloud) Copyright (c) BUSL-1.1 2016-present Scille SAS
+
+import platform
+import sys
+
+print(f"{sys.platform=}", file=sys.stderr)
+
+if sys.platform == "linux":
+    platform_info = platform.freedesktop_os_release()
+    # VERSION_ID is not available in all platforms
+    release = platform_info.get("VERSION_ID", "unknown")
+    os = platform_info["ID"]
+elif sys.platform == "darwin":
+    platform_info = platform.mac_ver()
+    release = platform_info[0]
+    os = "macos"
+elif sys.platform == "win32":
+    platform_info = platform.win32_ver()
+    release = platform_info[0]
+    os = "windows"
+else:
+    raise ValueError(f"Unsupported platform {sys.platform}")
+
+print(f"{platform_info=}", file=sys.stderr)
+print(f"{os=} {release=}", file=sys.stderr)
+
+id = f"{os}-{release}"
+print(f"{id=}", file=sys.stderr)
+
+print(f"platform={sys.platform}")
+print(f"{id=}")
+print(f"{os=}")
+print(f"{release=}")
--- a/.github/workflows/ci-python.yml
+++ b/.github/workflows/ci-python.yml
@@ -51,6 +51,11 @@ jobs:
       - uses: actions/checkout@692973e3d937129bcbf40652eb9f2f61becf3332 # pin v4.1.7
         timeout-minutes: 5
 
+      - name: Retrieve runner specs
+        id: runner-specs
+        uses: ./.github/actions/system-info
+        timeout-minutes: 1
+
       - name: Set apt mirror
         if: (!inputs.style-only)
         # GitHub Actions apt proxy is super unstable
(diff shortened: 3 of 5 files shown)


.github/workflows/ci-python.yml
```diff
@@ -48,6 +48,11 @@ jobs:
     # 20.04 is required to install PostgreSQL 12
     runs-on: ubuntu-20.04
     steps:
+      - name: Retrieve runner specs
+        id: runner-specs
+        uses: kenchan0130/actions-system-info@b98fe44bc35efcc60a9828cf4a04783a8b370bde # pin v1.3.0
```
touilleMan, line 53: Can we avoid using this external action ?
My issue is that the repos is not very popular (only 6 stars) and has a [lot of dependencies](https://github.com/kenchan0130/actions-system-info/blob/master/package-lock.json) so I'm not confident in adding it to our codebase...

My guess is we can achieve something similar by doing;
```
name: Retrieve runner specs
id: runner-specs
run: |
  python -c '
    import platform
    # allowed values are https://docs.github.com/en/actions/learn-github-actions/contexts#runner-context
    match "${runner.os}":
        case "Linux":
            r = platform.freedesktop_os_release()
            # e.g. `ubuntu-22.04`
            os_id = f"{r["ID"]}-{r["VERSION_ID"]}"
        case "macOS":
            # e.g. `macos-13.6.7`
            os_id = "macos-" + platform.mac_ver()[0]
        case "Windows":
            # e.g. `windows-10`
            os_id = "windows-" + platform.win32_ver()[0]
        case unknown:
            raise ValueError(f"Unknown platform {unknown!r}")
    print("::set-output name=os_id::" + os_id)
  '
```

touilleMan (dismissed):

.github/actions/system-info/system-info.py
```diff
@@ -0,0 +1,32 @@
...
+print(f"{platform_info=}", file=sys.stderr)
+print(f"{os=} {release=}", file=sys.stderr)
+
+id = f"{os}-{release}"
+print(f"{id=}", file=sys.stderr)
+
+print(f"platform={sys.platform}")
+print(f"id={id}")
+print(f"os={os}")
+print(f"release={release}")
```
mmmarcos, line 32:
```suggestion
print(f"{id=}")
print(f"{os=}")
print(f"{release=}")
```

.github/actions/system-info/action.yml
```diff
@@ -0,0 +1,24 @@
...
+    value: ${{ steps.system-info.outputs.platform }}
+  id:
+    description: The system ID
+    value: ${{ steps.system-info.outputs.id }}
+  os:
+    description: The OS name
+    value: ${{ steps.system-info.outputs.os }}
+  release:
+    description: The release version of the OS
+    value: ${{ steps.system-info.outputs.release }}
```
mmmarcos, line 16: A couple of comments:
1. only `os` and `release` seem to be used in workflows. Maybe `platform` and `id` could be added later when necessary.
2. `platform` (`linux`, `darwin`, `win32`) and `os` (`<distro>`, `macos`, `windows`) seem somehow mixed and redundant. Unless we want different caches for different distros (we only use ubuntu in the CI isnt' it?), why not using only `platform` for cache?

.github/actions/system-info/system-info.py
```diff
@@ -0,0 +1,32 @@
+# Parsec Cloud (https://parsec.cloud) Copyright (c) BUSL-1.1 2016-present Scille SAS
+
+import platform
+import sys
+
+print(f"{sys.platform=}", file=sys.stderr)
+
+if sys.platform == "linux":
+    platform_info = platform.freedesktop_os_release()
+    release = platform_info["VERSION_ID"]
```
mmmarcos, line 10: According to the docs `VERSION_ID` is optional. Only `NAME`, `ID`, and `PRETTY_NAME` are always present.
I guess this shouldn't be an issue if the CI runs only on Ubuntu, but maybe we should add a comment like:
```suggestion
    release = platform_info["VERSION_ID"]  # not available in all platforms
```
Or maybe something like:
```suggestion
    release = platform_info.get("VERSION_ID", "unknown")
```

BTW, on my config:
```
>>> print(json.dumps(platform.freedesktop_os_release(), indent=4))
{
    "NAME": "Debian GNU/Linux",
    "ID": "debian",
    "PRETTY_NAME": "Debian GNU/Linux trixie/sid",
    "VERSION_CODENAME": "trixie",
    "HOME_URL": "https://www.debian.org/",
    "SUPPORT_URL": "https://www.debian.org/support",
    "BUG_REPORT_URL": "https://bugs.debian.org/"
}
```

.github/actions/system-info/system-info.py
```diff
@@ -0,0 +1,32 @@
+# Parsec Cloud (https://parsec.cloud) Copyright (c) BUSL-1.1 2016-present Scille SAS
+
+import platform
+import sys
+
+print(f"{sys.platform=}", file=sys.stderr)
```
mmmarcos, line 6: Just curious: why do we need to print to `sys.stderr`?

.github/actions/system-info/system-info.py
```diff
@@ -0,0 +1,32 @@
...
+    platform_info = platform.win32_ver()
+    release = platform_info[0]
+    os = "windows"
+else:
+    raise ValueError(f"Unsupported platform {sys.platform}")
+
+print(f"{platform_info=}", file=sys.stderr)
+print(f"{os=} {release=}", file=sys.stderr)
+
+id = f"{os}-{release}"
```
mmmarcos, line 26: This gives consistent output for windows and macos, but not for linux?
- `windows-<release>` 
- `macos-<release>`
- `<distro>-<version>`

.github/actions/system-info/action.yml
```diff
@@ -0,0 +1,24 @@
+name: system-info
+description: Get system information
+
+outputs:
+  platform:
+    description: The platform name
```
mmmarcos, line 6: It would be nice to have examples of possible values in the descriptions, like
```suggestion
    description: The platform name ("linux", "windows", "macos")
```

.github/actions/system-info/system-info.py
FirelightFlagboy: To not "poison" `$GITHUB_OUTPUT`

.github/actions/system-info/system-info.py
FirelightFlagboy: That's exected, you only have one flavor of `windows` & `macos` but multiple flavor for `linux`

.github/actions/system-info/action.yml
FirelightFlagboy: 1. Having those extra output cost nothing
2. I prefer to use `os` has you could run github action locally

.github/actions/system-info/system-info.py
FirelightFlagboy: done

.github/actions/system-info/action.yml
FirelightFlagboy: done

mmmarcos (approved):

PR #7527, opened 2024-06-05
Change data & config dir of parsec to remove the suffix `-alpha`

FirelightFlagboy:


cli/src/tests.rs
```diff
@@ -875,7 +875,7 @@ async fn remove_device(tmp_path: TmpPath) {
 
     let alice_device_file = tmp_path
         .join("config")
-        .join("parsec-v3-alpha")
+        .join("parsec3/libparsec")
```
Max-7, line 878: Probably doesn't change much but you could also separate those:
```
.join("config")
.join("parsec3")
.join("libparsec")
...
```

libparsec/crates/platform_device_loader/src/lib.rs
```diff
@@ -26,6 +26,65 @@ pub const ARGON2ID_DEFAULT_PARALLELISM: u32 = 1;
...
+    #[cfg(target_arch = "wasm32")]
+    {
+        PathBuf::from("")
+    }
+    #[cfg(not(target_arch = "wasm32"))]
+    {
+        let mut path = if let Ok(config_dir) = std::env::var(PARSEC_CONFIG_DIR) {
+            PathBuf::from(config_dir)
+        } else {
+            // TODO: check MacOS path correspond to Parsec v2
```
Max-7, line 61: Is this comment still relevant?

Max-7 (approved):

libparsec/crates/platform_device_loader/src/lib.rs
FirelightFlagboy: nop

cli/src/tests.rs
FirelightFlagboy: Just do everything on one `join` expect for that last value which is dynamic

PR #7573, opened 2024-06-10
Declare local storage's SQLite tables as STRICT (to avoid error in table typing)

### LINKED ISSUE #6529: Declare SQlite tables as STRICT
see https://www.sqlite.org/stricttables.html

touilleMan:
closes #6529

This should be mostly compatible with old way of declaring flexible tables, however it's always better to avoid exotic corner cases so merging it fast (or discarding early !) is advised ;-)


vxgmichel (approved):
LGTM :+1:

FirelightFlagboy (approved):

touilleMan:
I've done a bit more testing before merging this: the "non-strict" way of declaring table seems fully compatible with the new strict way (given the fields with type BOOL/BOOLEAN stored INTEGER, and we never have a field that store different types of data).

PR #7587, opened 2024-06-13
Document how to create a api route.

AureliaDolo:


AureliaDolo:
I'm thinking about adding somewhere the full diff of what you have when you follow the doc with the given example, but I don't know how useful will it be, where to put it and how readable it is.

In my idea, if you want to add a route, you could apply the patch and then see where you need to add stuff.

docs/development/how_to_add_a_route.md
```diff
@@ -0,0 +1,146 @@
...
+    ) -> None | DummyBadOutcome:
+        print("There are ",len(self._data.organizations), " organizations here.")
+```
+
+
+## Server tests
+
+In import the new test file in `server/tests/api_v4/xxxxx/init.py`. It must be named `test_{new command name}`
+
+Start a new testbed
```
mmmarcos, line 139:
```suggestion
Start a new testbed (TODO instructions)
```

docs/development/how_to_add_a_route.md
```diff
@@ -0,0 +1,146 @@
...
+        self,
+        # parameters found in client_ctx and req
+    ) -> None | DummyBadOutcome:
+        print("There are ",len(self._data.organizations), " organizations here.")
+```
+
+
+## Server tests
+
+In import the new test file in `server/tests/api_v4/xxxxx/init.py`. It must be named `test_{new command name}`
```
mmmarcos, line 137:
```suggestion
Add the new test file `server/tests/api_v4/xxxxx/dummy.py`. It should contain a test function for each response status (see other command's tests as a guide on how to write them).

Import the new test file in `server/tests/api_v4/xxxxx/__init__.py`. It must be named `test_{new command name}` (e.g. `test_dummy`).   
```

docs/development/how_to_add_a_route.md
```diff
@@ -0,0 +1,146 @@
...
+
+Start a new testbed
+
+`test_each_cmd_req_rep_has_dedicated_test` will fail if a test per output are missing
```
mmmarcos, line 141:
```suggestion
`test_each_cmd_req_rep_has_dedicated_test` will fail if a test function is missing
```

docs/development/how_to_add_a_route.md
```diff
@@ -0,0 +1,146 @@
...
+
+
+## Client side
```
mmmarcos, line 144:
```suggestion
## Client-side implementation
```

mmmarcos (commented):
Thanks for these docs! Here are some comments.

vxgmichel (approved):
LGTM :+1:

docs/development/how_to_add_a_route.md
```diff
@@ -0,0 +1,175 @@
+<!-- Parsec Cloud (https://parsec.cloud) Copyright (c) BUSL-1.1 2016-present Scille SAS -->
+
+The goal of this document is to cover all the necessary steps to add an API route to Parsec.
```
touilleMan, line 3: Instead of a"API route", I'd would talk about "API RPC command" given we have `invitated_cmds`/`authenticated_cmds`/`anonymous_cmds` modules that contains the schemas of those RPC commands (and "API route" may seem to refer to the administration REST API which is a totally unrelated component).

touilleMan: so I guess this file should be renamed to something like `how_to_add_a_rpc_cmd.md`

docs/development/how_to_add_a_route.md
```diff
@@ -0,0 +1,175 @@
...
+
+In summary:
+1. Add the schema describing the new API route
+2. Generate binding code and serialization tests
+3. Implement server-side code (In-memory & PostgreSQL) and tests
+4. Implement client-side code and tests
+
+> Note that you will need a working development environment. See (./README.md).
+## Schema generation
```
touilleMan, line 12: There is multiple markdown syntax error in the file, here is a patch to fix them

```
$ git apply how_to_add_a_route.md.patch.txt
```


[how_to_add_a_route.md.patch.txt](https://github.com/user-attachments/files/15903059/how_to_add_a_route.md.patch.txt)

docs/development/how_to_add_a_route.md
```diff
@@ -0,0 +1,175 @@
...
+
+Let's add a new `dummy` route to `invited_cmds`.
+
+> API routes are also called "commands" in Parsec. There are three command families: `anonymous_cmds` (authentication not required), `authenticated_cmds` (authentication required) and `invited_cmds` (invitation required).
```
touilleMan, line 16: I think you can move this comment at the beginning of the file (hence this would explain the naming of the file) and also add a line about the unrelated administration REST API:

```suggestion
> Client-to-server RPC API routes are also called "commands" in Parsec. There are three command families: `anonymous_cmds` (authentication not required), `authenticated_cmds` (authentication required) and `invited_cmds` (invitation required).
>
> Note there is another unrelated API for administration purpose that uses REST instead of RPC (as it is expected to be used by admin using curl), this documentation doesn't cover it.
```

docs/development/how_to_add_a_route.md
```diff
@@ -0,0 +1,175 @@
...
+
+```
+For a full description of the accepted format see [the parsing/generating script](../../misc/gen_protocol_typings.py).
+
+> Note that the files are .json5. This is to allow for comments inside these files. But the parser we are using only supports .json files and we're removing the comments ourselves. So don't use any other specific feature of the json5 format (trailing commas, I'm looking at you 👀).
+
+Then from the root directory, run the following scripts to generate binding code.
+
+```shell
+python make.py r
```
touilleMan, line 60: This is the wrong command, it should be `python misc/gen_protocol_typings.py`

docs/development/how_to_add_a_route.md
```diff
@@ -0,0 +1,175 @@
...
+```
+
+This will rebuild the bindings (use option `i` to install) and modify the following files:
```
touilleMan, line 63: with `make.py r` we only rebuild the bindings

docs/development/how_to_add_a_route.md
```diff
@@ -0,0 +1,175 @@
...
+    - `server/parsec/_parsec_pyi/protocol/__init__.pyi`
+    - `server/parsec/_parsec_pyi/protocol/xxxxxx_cmds/v4/__init__.pyi`
+    - `server/tests/common/rpc.py`
+
+```shell
+python misc/gen_protocol_typings.py
+```
+
+This will generate the .pyi files and some helpers for the tests. The generated code is in `/target` directory.
```
touilleMan, line 72: I'd would remove the `The generated code is in `/target` directory` given we never look into this folder when working with python (once `./make.py r` is run, the python module is automatically copied to `server/parsec/_parsec.cpython-312-x86_64-linux-gnu.so`

touilleMan (changes requested):

docs/development/how_to_add_a_route.md
AureliaDolo: Ok so first we should run gen_protocol_typing and then make.py r ?

docs/development/how_to_add_a_rpc_cmd.md
```diff
@@ -0,0 +1,174 @@
...
+### Server tests
+
+Add the new test file `server/tests/api_v4/xxxxx/dummy.py`. It should contain a test function for each response status (see other command's tests as a guide on how to write them).
+
+Import the new test file in `server/tests/api_v4/xxxxx/__init__.py`. It must be named `test_{new command name}` (e.g. `test_dummy`).
+
+Start a new testbed :
+
+```shell
+python make.py rts
```
FirelightFlagboy, line 165:
```suggestion
python make.py run-testbed-server
```

I would prefer using the long name, I find it less "cryptic"

docs/development/how_to_add_a_rpc_cmd.md
```diff
@@ -0,0 +1,174 @@
...
+```
+
+For a full description of the accepted format see [the parsing/generating script](../../misc/gen_protocol_typings.py).
+
+> Note that the files are .json5. This is to allow for comments inside these files. But the parser we are using only supports .json files and we're removing the comments ourselves. So don't use any other specific feature of the json5 format (trailing commas, I'm looking at you 👀).
+
+Then from the root directory, run the following scripts to generate binding code.
+
+```shell
+python make.py r
```
FirelightFlagboy, line 65:
```suggestion
python make.py python-dev-rebuild
```

I would prefer using the long name, I find it less "cryptic"

docs/development/how_to_add_a_route.md
touilleMan: they are orthogonal:

- `gen_protocol_typing.py` regenerates some python files that gets commited to the project
- `make.py r` (re)build `server/parsec/_parsec.cpython-312-x86_64-linux-gnu.so`

Modifying Python file doesn't invalidate `server/parsec/_parsec.cpython-312-x86_64-linux-gnu.so` so there is no need to run `make.py r` after `gen_protocol_typing.py`.

What is misleading is the reason for running `gen_protocol_typing.py` is because the protocol schema have been updated... which is also a reason to run `make.py r` ! (so you could say `make.py r` and `gen_protocol_typing.py` are cousin as they are triggered by the same reason, but don't have a parenting relationship between them.

AureliaDolo: Ok so if stuff is changed everywhere, we need to run both in either order ?

touilleMan: > Ok so if stuff is changed everywhere, we need to run both in either order ?

yup ! ^^

mmmarcos (approved):

touilleMan (approved):

Issue #7547, opened 2024-06-07
Error when launching newly installed app


Ironicbay:
```Unhandled Promise Rejection
Error: ZIP file not provided: [
  {
    "url": "https://github.com/Scille/parsec-cloud/releases/download/v3.0.0-b.7/Parsec_3.0.0-b.7_mac_x86_64.dmg",
    "info": {
      "url": "Parsec_3.0.0-b.7_mac_x86_64.dmg",
      "sha512": "/wR94/yTeqUi4CA3H30ObSQ7HCJXp2sfmQvhNO1Z/pF7X0XFf13nj+/0Sy5zWSLckEaMzAZkNMVxtrbpnemMew==",
      "size": 114761227
    }
  }
]
    at newError (/Applications/parsec.app/Contents/Resources/app.asar/node_modules/builder-util-runtime/out/index.js:47:19)
    at MacUpdater.doDownloadUpdate (/Applications/parsec.app/Contents/Resources/app.asar/node_modules/electron-updater/out/MacUpdater.js:67:55)
    at MacUpdater.downloadUpdate (/Applications/parsec.app/Contents/Resources/app.asar/node_modules/electron-updater/out/AppUpdater.js:393:25)
    at MacUpdater.doCheckForUpdates (/Applications/parsec.app/Contents/Resources/app.asar/node_modules/electron-updater/out/AppUpdater.js:357:55)
    at process.processTicksAndRejections (node:internal/process/task_queues:95:5)
    at async AppUpdater.checkForUpdates (/Applications/parsec.app/Contents/Resources/app.asar/build/src/updater.js:358:13)
    at async ElectronCapacitorApp.checkForUpdates (/Applications/parsec.app/Contents/Resources/app.asar/build/src/setup.js:157:9)```

PR #7589, opened 2024-06-14
Fix the error message on MacOS about a missing ZIP file

Ironicbay:
Closes https://github.com/Scille/parsec-cloud/issues/7547


FirelightFlagboy (commented):
You need to update the `package-client` and `release` workflow to upload the new artifact

FirelightFlagboy (approved):

Issue #7585, opened 2024-06-13
[🐛 | Bug]: Files in the mountpoint seem unacessible during synchronization


mmmarcos:
### Parsec version tested on:

3.0.0-b.11

### Platforms tested on:

Linux

### Bug description:

Reported by Thierry: 

1. Copy a directory with files from a mounted workspace in Parsec V2 to a mounted workspace in Parsec V3
2. While copying, files starting to appear with 0 bytes in Parsec V3 mountpoint leading to an error when trying to open.
3. After "some time" files seem to be synchronized and can be open normally.

Is there a way to not display the files if they are not synchronized/accessible?

touilleMan:
After more investigation, the issue comes from two things:
- Reading data from Parsec v2 when the data is not is cache is much slower than reading data from from a regular on-disk filesystem (and on top of that the performance is heavily dependent on the internet connection).
- When the file is under copy, it displayed size (e.g. when doing a `ls -l` on the parent directory) only grows when a flush occurs. A flush can occur if explicitly issued by the application doing the copy (this is not the case here), or automatically if the amount of data written since the last flush is bigger than the blocksize of the file (i.e. 512ko).

The conjunction of those means the file can appear with a size of 0  for quiet some time (while in fact data are being transferred, just at a slow pace that makes reaching the 512ko threshold).

On top of that, opening the file lead to an error... just because the file has not finished to be copied yet !

So in the end I think there is not much to do here: having a really slow copy speed is a corner case, and the whole issue here comes from trying to open a file that is part of an import of multiple files still in progress.

PR #7606, opened 2024-06-17
Add troubleshooting info for #7585

mmmarcos:
Also add missing translations for Troubleshooting section.

Related to #7585

### DIFF
--- a/docs/userguide/troubleshooting.rst
+++ b/docs/userguide/troubleshooting.rst
@@ -5,8 +5,9 @@
 Troubleshooting
 ===============
 
-Excel/Word or another MS Office software says the file name is too long
------------------------------------------------------------------------
+
+MS Excel/Word or another MS Office software says the file name is too long
+--------------------------------------------------------------------------
 
 .. note::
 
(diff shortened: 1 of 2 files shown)


docs/userguide/troubleshooting.rst
```diff
@@ -18,19 +19,37 @@ Excel/Word or another MS Office software says the file name is too long
...
+1. Go to Parsec configuration directory
 
-**On Windows**, in the explorer's address bar, go to ``%APPDATA%/parsec**`` (or ``C:\Users\<YourUser>\AppData\Roaming``, AppData is a hidden folder).
-**On Linux**, go to ``~/.config/parsec``.
+   - **On Windows**, open the file explorer and put ``%APPDATA%/parsec**`` in the address bar (or go to ``C:\Users\<YourUser>\AppData\Roaming``, note that AppData is a hidden folder).
+   - **On Linux**, open ``~/.config/parsec``.
 
-- Remove unwanted devices in config/devices
-- Remove unwanted devices in data
+2. Remove unwanted devices in config/devices
```
FirelightFlagboy, line 50: The path has changed since #7527.

Should be `libparsec/config/devices` :thinking:

docs/userguide/troubleshooting.rst
```diff
@@ -18,19 +19,37 @@ Excel/Word or another MS Office software says the file name is too long
...
-This usually occurs when a user has tried Parsec beta versions in which test databases could be reset at any time, leaving old devices still registered on their computer.
+You can still remove these devices manually:
 
-A user can still erase these devices manually though:
+1. Go to Parsec configuration directory
 
-**On Windows**, in the explorer's address bar, go to ``%APPDATA%/parsec**`` (or ``C:\Users\<YourUser>\AppData\Roaming``, AppData is a hidden folder).
-**On Linux**, go to ``~/.config/parsec``.
+   - **On Windows**, open the file explorer and put ``%APPDATA%/parsec**`` in the address bar (or go to ``C:\Users\<YourUser>\AppData\Roaming``, note that AppData is a hidden folder).
+   - **On Linux**, open ``~/.config/parsec``.
```
FirelightFlagboy, line 48: The path has changed since #7527, should be `parsec3`.

```suggestion
   - **On Windows**, open the file explorer and put ``%APPDATA%/parsec3`` in the address bar (or go to ``C:\Users\<YourUser>\AppData\Roaming``, note that AppData is a hidden folder).
   - **On Linux**, open ``~/.config/parsec3``.
```

docs/userguide/troubleshooting.rst
```diff
@@ -18,19 +19,37 @@ Excel/Word or another MS Office software says the file name is too long
...
 
-- Remove unwanted devices in config/devices
-- Remove unwanted devices in data
+2. Remove unwanted devices in config/devices
+3. Remove unwanted devices in data
```
FirelightFlagboy, line 51: We store the device file in a single directly, it should be the config dir

FirelightFlagboy (changes requested):

docs/userguide/troubleshooting.rst
mmmarcos: Notice this is inside a section "Specific to Parsec V2" that I've just added. 

Besides, I'm not sure if that is still an issue in V3 since the concept of devices is more or less hidden to the user and I think only the more recent device is displayed.

docs/userguide/troubleshooting.rst
mmmarcos: I don't understand your comment: are you suggesting to change something?

BTW, I think data is stored in `~/.local/share/parsec` (instead of `~./config/parsec`) which is not mentioned in the article.

IMHO this is written like this since V2 and users should not have too many devices. I'd just leave it like it is for the moment and just remove it when V3 becomes public. What do you think?

FirelightFlagboy: I've just checked, for parsec v2:

- We store the devices in `~/.config/parsec/devices`
- And in `~/.local/share/parsec` the workspaces

My comment is more about that the user won't find the devices in `~/.local/share/parsec`

FirelightFlagboy: > IMHO this is written like this since V2 and users should not have too many devices. I'd just leave it like it is for the moment and just remove it when V3 becomes public. What do you think?

IMHO, I don't see any complains for clients so good enough :shrug:

PR #7607, opened 2024-06-17
Download APT mirrors for specific country

FirelightFlagboy:


.github/workflows/ci-python.yml
```diff
@@ -56,11 +56,14 @@ jobs:
...
-          set -e -o pipefail
+          set -eux -o pipefail
+          COUNTRY=$(curl ipinfo.io | jq -r .country)
+          MIRROR_FILE=$(mktemp)
+          curl "http://mirrors.ubuntu.com/$COUNTRY.txt" > $MIRROR_FILE
           (
             # make sure there is a `\t` between URL and `priority:*` attributes
             printf 'http://azure.archive.ubuntu.com/ubuntu	priority:1\n';
-            curl http://mirrors.ubuntu.com/mirrors.txt | grep https
+            grep -e '^https' $MIRROR_FILE
```
touilleMan, line 66: why not 

```suggestion
            curl "http://mirrors.ubuntu.com/$COUNTRY.txt" | grep -e '^https'
```

?

touilleMan (approved):

.github/workflows/ci-python.yml
FirelightFlagboy: Typically before, it failed because the downloaded mirror file doesn't contained any https endpoint but that wasn't really clear since everything append at the same state

PR #7609, opened 2024-06-17
Shamir deletion specification - server side

### LINKED ISSUE #7364: [Shamir port] Certificate deletion
Related #6090 

When a shamir setup is overwritten/deleted a deletion certificate must be produced and sent to the author and the share recipients.

AureliaDolo:
Related #7364

### DIFF
--- a/docs/rfcs/1000-shamir-based-recovery.md
+++ b/docs/rfcs/1000-shamir-based-recovery.md
@@ -87,7 +87,7 @@ To use the Shamir recovery:
    secret `SK` and decrypt `SK(alice@shamir1)`.
 6) Alice uses `alice@shamir1` to re-create a new device (the recovered device).
 
-## 1 - Create & update a Shamir recovery setup
+## 1 - Create a Shamir recovery setup
 
 Authenticated API:
 
@@ -102,8 +102,7 @@ Authenticated API:
             "fields": [
                 {
                     "name": "setup",
-                    // Set to `None` to clear previous Shamir recovery setup
-                    "type": "RequiredOption<ShamirRecoverySetup>"
+                    "type": "ShamirRecoverySetup"
                 }
             ]
         },
@@ -228,6 +227,8 @@ Consistency between `brief` and `shares` must be checked by the Parsec server:
 > **_Future evolution 2:_** Check the Shamir recovery setup against some
 > organization-specific rules (e.g. the number of shares, the recipient's
 > profiles, max number of share per recipient, etc.). See "Bonus" section below.
+
+To update the setup, the previous setup must be deleted first.
 
 And the related certificates:
 
@@ -704,6 +705,121 @@ The claimer gets access to `reveal_token` and `data_key`, it can then retrieve `
 Then `ciphered_data` can be decrypted with `data_key`. From then on, the recovery works
 just like the recovery device system (see `parsec core import_recovery_device` CLI).
 
+## 4 - Delete a shamir setup
+
+Authenticated API:
+
+```json5
+[
+    {
+        "major_versions": [
+            4
+        ],
+        "req": {
+            "cmd": "shamir_recovery_deletion",
+            "fields": [
+                {
+                    "name": "certificate",
+                    // Contains a ShamirRecoveryDeletionCertificate
+                    "type": "bytes"
+                }
+            ]
+        },
+        "reps": [
+            {
+                "status": "ok"
+            },
+            {
+                // Cannot deserialize data into the expected certificate
+                "status": "invalid_data"
+            },
+            {
+                // Given timestamp does not match previous timestamps
+                "status": "incoherent_previous_timestamp"
+            }
+            {
+                // Nothing to delete
+                "status": "no_setup_to_delete"
+            },
+            {
+                // Returned if the timestamp in the certificate is too far away compared
+                // to server clock.
+                "status": "timestamp_out_of_ballpark",
+                "fields": [
+                    {
+                        "name": "ballpark_client_early_offset",
+                        "type": "Float"
+                    },
+                    {
+                        "name": "ballpark_client_late_offset",
+                        "type": "Float"
+                    },
+                    {
+                        "name": "server_timestamp",
+                        "type": "DateTime"
+                    },
+                    {
+                        "name": "client_timestamp",
+                        "type": "DateTime"
+                    }
+                ]
+            },
+            {
+                // Returned if another certificate or vlob in the server has a timestamp
+                // posterior or equal to our current one.
+                "status": "require_greater_timestamp",
+                "fields": [
+                    {
+                        "name": "strictly_greater_than",
+                        "type": "DateTime"
+                    }
+                ]
+            }
+        ],
+    }
+]
+```
+The deletion certificate
+
+```json5
+{
+    "label": "ShamirRecoveryDeletionCertificate",
+    "type": "shamir_recovery_deletion_certificate",
+    "other_fields": [
+        {
+            "name": "author",
+            "type": "DeviceID"
+        },
+        {
+            "name": "timestamp",
+            "type": "DateTime"
+        },
+        {
+            "name": "previous_timestamp",
+            "type": "DateTime"
+        },
+        {
+            "name": "share_recipients",
+            "type": "List<UserId>"
+        }
+    ]
+}
+```
+
+The certificate needs to include the previous certificate timestamp in the deletion certificateto link both certificates together.
+
+> Future evolution: link the certificates with 1/ the brief signature 2/ a certificate id. This kind of link can be done between the share and brief certificate too.
+
+Who need witch certificate ?
+
+| certificate | author | share recipient |
+|-------------|--------|-----------------|
+| brief       | x      |                 |
+| share       |        | x               |
+| deletion    | x      | x               |
+
+
+
 ## Bonus
 
 ### **Future evolution 1**: Only Admin can clear Shamir recovery
@@ -803,3 +919,16 @@ On the other hand, approach 1) allow things like "recovery requires Alice, Bob a
 requires Alice and Bob, or Adam alone". However given we cannot trust the server
 on such precise configuration, a new certificate type must be introduced which is cumbersome :(
 Also we should be able to provide approach 2) as part of approach 1).
+
+### **Future evolution 3**: notify shamir author when shamir secret become irrecoverable
+
+If Alice has setup a shamir with Bob, Mallory and John, each of then having one share and a threshold of two.
+Then Bob and Mallory leave the organization. So this setup becomes unusable.
+
+The goal would be to have a notification to prompt the user to setup a new shamir.
+Depending on the configuration, if any of the share recipient is deleted a warning could be sent first
+event if the shamir could still be used.
+
+Two propositions could mitigate that:
+- each time a user is deleted, if they were a share recipient a notification could be sent to users on the other end of this shamir setup.
+- at each connection, check if share recipients are still valid.
--- a/docs/rfcs/1000-shamir-based-recovery.md
+++ b/docs/rfcs/1000-shamir-based-recovery.md
@@ -139,7 +139,7 @@ Authenticated API:
                 "status": "missing_share_for_recipient"
             },
             {
-                // Future evolution 1: Shamir recovery has already been setup, should ask your admin to reset it first !
+                // Shamir recovery has already been setup
                 "status": "shamir_setup_already_exists"
             },
             {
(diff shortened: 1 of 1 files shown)


docs/rfcs/1000-shamir-based-recovery.md
```diff
@@ -730,6 +731,121 @@ The claimer gets access to `reveal_token` and `data_key`, it can then retrieve `
...
+
+The certificate needs to include the previous certificate timestamp in the deletion certificateto link both certificates together.
+
+> Future evolution: link the certificates with 1/ the brief signature 2/ a certificate id. This kind of link can be done between the share and brief certificate too.
+
+Who need witch certificate ?
+
+| certificate | author | share recipient |
+|-------------|--------|-----------------|
+| brief       | x      |                 |
```
AureliaDolo, line 843:
```suggestion
| brief       | x      | x               |
```

The shamir_recovery_others_list provides both certificates. I'm not exactly sure why though

docs/rfcs/1000-shamir-based-recovery.md
```diff
@@ -730,6 +731,121 @@ The claimer gets access to `reveal_token` and `data_key`, it can then retrieve `
...
+            {
+                "status": "ok"
+            },
+            {
+                // Cannot deserialize data into the expected certificate
+                "status": "invalid_data"
+            },
+            {
+                // Given timestamp does not match previous timestamps
+                "status": "incoherent_previous_timestamp"
```
vxgmichel, line 764: We tend to use the synonym `inconsistent` in the code base:
```
± git grep inconsistent | wc -l
15
[vinmic:~/repos/parsec-cloud-v3] [base] master+*
± git grep incoherent | wc -l  
4
```

docs/rfcs/1000-shamir-based-recovery.md
```diff
@@ -730,6 +731,121 @@ The claimer gets access to `reveal_token` and `data_key`, it can then retrieve `
...
+            "name": "timestamp",
+            "type": "DateTime"
+        },
+        {
+            "name": "previous_timestamp",
+            "type": "DateTime"
+        },
+        {
+            "name": "share_recipients",
+            "type": "List<UserId>"
```
vxgmichel, line 829: There is no "share recipients mismatch" status in `shamir_recovery_deletion`, is that right?

vxgmichel, line 829: Maybe acknowledge that `previous_timestamp` and `share_recipients` were not strictly required here, as this information is already available by browsing the shamir topic. If we do include this information, it'd be good to have some kind of justification for it (clarity, or something else, I don't remember what the conclusions were about this topic).

vxgmichel (approved):
A couple of comments but LGTM otherwise :+1: 

@touilleMan should review and validate this before it's merged

docs/rfcs/1000-shamir-based-recovery.md
vxgmichel: The 4 occurrences of `incoherent` were from your previous PR, sorry for not catching that earlier.

docs/rfcs/1000-shamir-based-recovery.md
AureliaDolo: Yes that was what I mean by the line below ^^

It was as a mean to link to the previous certificate. But I saw that @touilleMan is working on a better solution here #7577

AureliaDolo: Yes indeed, we can remove that.

docs/rfcs/1000-shamir-based-recovery.md
AureliaDolo: Nice catch :fishing_pole_and_fish: 

So here you have the correction #7621

docs/rfcs/1000-shamir-based-recovery.md
```diff
@@ -730,6 +731,121 @@ The claimer gets access to `reveal_token` and `data_key`, it can then retrieve `
...
+                // Cannot deserialize data into the expected certificate
+                "status": "invalid_data"
+            },
+            {
+                // Given timestamp does not match previous timestamps
+                "status": "incoherent_previous_timestamp"
+            }
+            {
+                // Nothing to delete
+                "status": "no_setup_to_delete"
```
touilleMan, line 764:
```suggestion
                "status": "shamir_setup_not_found"
```

to stay consistent with e.g. [`user_not_found`](https://github.com/Scille/parsec-cloud/blob/048b4c6528cbb1b78b6daa9d567f356c6307462f/libparsec/crates/protocol/schema/authenticated_cmds/user_revoke.json5#L23) and [`realm_not_found`](https://github.com/Scille/parsec-cloud/blob/048b4c6528cbb1b78b6daa9d567f356c6307462f/libparsec/crates/protocol/schema/authenticated_cmds/realm_share.json5#L33
And `shamir_setup` instead of just `shamir` given it's how it is specified in `shamir_recovery_setup`:
https://github.com/Scille/parsec-cloud/blob/048b4c6528cbb1b78b6daa9d567f356c6307462f/libparsec/crates/protocol/schema/authenticated_cmds/shamir_recovery_setup.json5#L54

I'm also okay to stick with just `setup`, so the vocabulary would be "a `ShamirRecoveryBrief` describes a setup, and has been added in the server with the `shamir_recovery_setup` command", but in this case we have to correct the `shamir_recovery_setup` command and make sure we are consistent everywhere (for instance I didn't check the comments in the code)

docs/rfcs/1000-shamir-based-recovery.md
```diff
@@ -726,6 +727,121 @@ The claimer gets access to `reveal_token` and `data_key`, it can then retrieve `
...
+Authenticated API:
+
+```json5
+[
+    {
+        "major_versions": [
+            4
+        ],
+        "req": {
+            "cmd": "shamir_recovery_deletion",
```
touilleMan, line 741:
```suggestion
            "cmd": "shamir_recovery_delete_setup",
```

The command name should contain a verb

(btw I guess `shamir_recovery_setup` is a bit weird on this regard given `setup` is both the verb and the noun, since after that we refer to the configuration set by this command as the `shamir setup`)

docs/rfcs/1000-shamir-based-recovery.md
AureliaDolo: After thinking about it a bit more, I'd rather keep the share recipients in the certificate, even if we could retrieve the information, because it's a certificate we went them to get.

docs/rfcs/1000-shamir-based-recovery.md
```diff
@@ -233,6 +232,8 @@ Consistency between `brief` and `shares` must be checked by the Parsec server:
 > organization-specific rules (e.g. the number of shares, the recipient's
 > profiles, max number of share per recipient, etc.). See "Bonus" section below.
 
+To update the setup, the previous setup must be deleted first.
```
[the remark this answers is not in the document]
touilleMan: On top of that, the `shamir_setup_already_exists` should have a `DateTime` field `last_shamir_certificate_timestamp`.

This is important given the Parsec client is expected to eagerly download all certificates, hence it should have been aware of the already existing shamir setup. And if that's not the case it is because the client is lagging behind (typically the certificate is very new), and in this case the client polls the server for certificates... but there is great chances this is already under way (the certificates ops start polling as soon as it gets the "new certificates" events from the server), hence this `last_shamir_certificate_timestamp` that allows to detect if the poll is really needed or if it has been concurrently achieved.

See for instance what is done for `user_revoke` command:
https://github.com/Scille/parsec-cloud/blob/048b4c6528cbb1b78b6daa9d567f356c6307462f/libparsec/crates/protocol/schema/authenticated_cmds/user_revoke.json5#L26-L31
and in the client (note the `RemoteIdempotent` status that means the server rejected our certificate, but we are missing some certificates and once we fetch them we will discover our current operation has already been achieved):
https://github.com/Scille/parsec-cloud/blob/048b4c6528cbb1b78b6daa9d567f356c6307462f/libparsec/crates/client/src/client/user_revoke.rs#L15-L26

docs/rfcs/1000-shamir-based-recovery.md
```diff
@@ -726,6 +727,121 @@ The claimer gets access to `reveal_token` and `data_key`, it can then retrieve `
...
+            {
+                "status": "ok"
+            },
+            {
+                // Cannot deserialize data into the expected certificate
+                "status": "invalid_data"
+            },
+            {
+                // Given timestamp does not match previous timestamps
+                "status": "incoherent_previous_timestamp"
```
touilleMan, line 760: I would merge `incoherent_previous_timestamp` into `no_setup_to_delete`: having both error makes sens if we consider the server first look for the only certificate that can be deleted, then check that its timestamp is what the client has sent.

However another way to see things is to consider the timestamp is used to uniquely identify a setup to delete (see another suggestion I did below, where I suggest to use user ID + timestamp as unique identifier of shamir setup). In this case if the unique identifier leads to nothing it's normal to return a `setup_not_found` error.

What makes me feel the balance tips toward the second solution is this comment: `// Future evolution 1: Shamir recovery has already been setup, should ask your admin to reset it first !`. So it seems interesting to later allow admin to remove other people's shamir setup. With the "user ID + timestamp is the unique identifier of the setup" there is nothing to change to the API to support this usecase.

docs/rfcs/1000-shamir-based-recovery.md
touilleMan: On top of that I would add a `shamir_setup_already_deleted` (with a `DateTime` field `last_shamir_recovery_certificate`) to cover the case the shamir setup exists but can no longer be deleted (see the comment about `shamir_setup_already_exists` for the reason for the field `last_shamir_recovery_certificate`)

docs/rfcs/1000-shamir-based-recovery.md
```diff
@@ -726,6 +727,121 @@ The claimer gets access to `reveal_token` and `data_key`, it can then retrieve `
...
+            "name": "share_recipients",
+            "type": "List<UserId>"
+        }
+    ]
+}
+```
+
+The certificate needs to include the previous certificate timestamp in the deletion certificateto link both certificates together.
+
+> Future evolution: link the certificates with 1/ the brief signature 2/ a certificate id. This kind of link can be done between the share and brief certificate too.
```
touilleMan, line 833: This is no longer needed if we have the user ID + timestamp (and is complex to add later on anyway given we have to support backward compatibility and hence not all certificates will have the ID...)

So I would suggest to just remove this comment

touilleMan (changes requested):

docs/rfcs/1000-shamir-based-recovery.md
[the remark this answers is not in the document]
AureliaDolo: > On top of that, the shamir_setup_already_exists should have a DateTime field last_shamir_certificate_timestamp. 

Done in #7622 :tada:

docs/rfcs/1000-shamir-based-recovery.md
vxgmichel: Sorry for not being clear, I meant that we probably need to add a "share recipients mismatch" status to `shamir_recovery_deletion` to return when this field doesn't match what is in the previous certificate.

docs/rfcs/1000-shamir-based-recovery.md
```diff
@@ -730,6 +731,121 @@ The claimer gets access to `reveal_token` and `data_key`, it can then retrieve `
...
+        {
+            "name": "author",
+            "type": "DeviceID"
+        },
+        {
+            "name": "timestamp",
+            "type": "DateTime"
+        },
+        {
+            "name": "previous_timestamp",
```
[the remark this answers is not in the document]
AureliaDolo: >     // (i.e. a user can only remove its own Shamir recovery)

And that means if we want a configuration where only admins can delete shamir certificates, it should be quite easy

docs/rfcs/1000-shamir-based-recovery.md
AureliaDolo: Maybe `shamir_recovery_setup_setup` ?

docs/rfcs/1000-shamir-based-recovery.md
```diff
@@ -739,30 +739,136 @@ Authenticated API, `organization_config`:
...
-                "status": "ok",
+                "status": "ok"
+            },
+            {
+                // Cannot deserialize data into the expected certificate
+                "status": "invalid_data"
+            },
+             {
+                // Share recipients lists are incoherent
+                "status": "incoherent_date"
```
AureliaDolo, line 761: Share recipients are the only not identifying data that we have in the deletion certificate to check. Maybe we could include more data from the brief to ensure coherence. I'm not sure how much it's needed though

touilleMan (approved):

Issue #7360, opened 2024-05-21
[Shamir port] server side list_shared_recovery_devices


AureliaDolo:
Related #6090 

- [ ] command schema in server/parsec/crates/schema/authenticated_cmds
- [ ] run gen_protocol_typing.py
- [ ] implementation in server/parsec/components, leaving the potgre implementation for #7358
- [ ] libparsec tests
- [ ] server tests

PR #7617, opened 2024-06-19
[Shamir] get_certificates returns shamir certificates.

### LINKED ISSUE #7362: [Shamir port] server side shared_recovery_device_info
Related #6090 

- [x] command schema in server/parsec/crates/schema/authenticated_cmds
- [x] run gen_protocol_typing.py
- [x] implementation in server/parsec/components, leaving the potgre implementation for #7358
- [x] libparsec tests
- [x] server tests

AureliaDolo:
Fix #7362 and #7360

In v2, certificates were queried when needed. In v3 all certificates are pulled regularly, so here we don't implement  shared_recovery_device_info and list_shared_recovery_devices routes but we add the shamir certificates to the get_certificates.

### DIFF
--- a/docs/rfcs/1000-shamir-based-recovery.md
+++ b/docs/rfcs/1000-shamir-based-recovery.md
@@ -329,14 +329,11 @@ This secret contains both:
 
 ## 2 - Get the Shamir recovery setup
 
-This enables two use cases:
+The shamir related certificates are retrieved by the `get_certificates` route, with other certificates depending on the timestamp.
+If the user has authored a shamir setup, they will get the corresponding brief.
+If they are a share recipient, they will get their share and the associated brief.
 
-1) A User retrieving its own Shamir recovery setup. Useful for displaying it to
-   the user and to ensure the setup is still valid (i.e. no recipient has been
-   revoked)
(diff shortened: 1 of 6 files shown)


vxgmichel (approved):
LGTM :+1:

docs/rfcs/1000-shamir-based-recovery.md
```diff
@@ -329,14 +329,11 @@ This secret contains both:
...
+If the user has authored a shamir setup, they will get the corresponding brief.
+If they are a share recipient, they will get their share and the associated brief.
 
-1) A User retrieving its own Shamir recovery setup. Useful for displaying it to
-   the user and to ensure the setup is still valid (i.e. no recipient has been
-   revoked)
-2) A user retrieving all the Shamir recovery information they are a part of.
-
-Authenticated API:
+> TODO: what happens with shamir deletion certificates.
```
touilleMan, line 336: I don't think shamir deletion certificate are a special case here: they get provided just like brief and share ones (ordered by timestamp) and the client's certificates ops will take care of them (typically by checking a corresponding brief certificate exists before integrating them, which is already what is done for share certificate btw)

touilleMan (approved):

PR #7622, opened 2024-06-21
Add timestamp to shamir already set error.

AureliaDolo:


libparsec/crates/protocol/schema/authenticated_cmds/shamir_recovery_setup.json5
```diff
@@ -51,7 +51,14 @@
...
                 // Future evolution 1: Shamir recovery has already been setup, should ask your admin to reset it first !
-                "status": "shamir_setup_already_exists"
+                "status": "shamir_setup_already_exists",
+                "fields": [
+                    {
+                        "name": "last_shamir_certificate_timestamp",
+                        "type": "DateTime"
+                    }
+
+                ]
```
mmmarcos, line 61: Is this implemented? I'm asking because the comment before says "Future evolution 1..."

server/parsec/components/shamir.py
```diff
@@ -129,7 +135,11 @@ class ShamirAddOrDeleteRecoverySetupStoreBadOutcome(BadOutcomeEnum):
     AUTHOR_NOT_FOUND = auto()
     AUTHOR_REVOKED = auto()
     INVALID_RECIPIENT = auto()
-    ALREADY_SET = auto()
+
+
+@dataclass(slots=True)
+class ShamirAlreadySet(BadOutcome):
```
mmmarcos, line 141: For consistency with the status, shouldn't it be named `ShamirSetupAlreadyExists` ?

server/tests/api_v4/authenticated/test_shamir_recovery_setup.py
```diff
@@ -108,7 +108,7 @@ async def test_authenticated_shamir_recovery_setup_shamir_setup_already_exists(
         [share.dump_and_sign(coolorg.alice.signing_key)],
     )
     rep = await coolorg.alice.shamir_recovery_setup(setup)
-    assert rep == authenticated_cmds.v4.shamir_recovery_setup.RepShamirSetupAlreadyExists()
+    assert isinstance(rep, authenticated_cmds.v4.shamir_recovery_setup.RepShamirSetupAlreadyExists)
```
mmmarcos, line 111: You should probably check the timestamp too. Something like:
```suggestion
    assert rep == authenticated_cmds.v4.shamir_recovery_setup.RepShamirSetupAlreadyExists(
        last_shamir_certificate_timestamp=dt
    )
```

libparsec/crates/protocol/schema/authenticated_cmds/shamir_recovery_setup.json5
AureliaDolo: Yes it is implemented, I'm removing the future evolution comment [here](https://github.com/Scille/parsec-cloud/pull/7609#discussion_r1647292959)

server/tests/api_v4/authenticated/test_shamir_recovery_setup.py
AureliaDolo: dt is the timestamp of the new certificate we're trying to add, it's not the timestamp of the already present setup.

For now `setup_shamir_for_coolorg` return nothing, but there is a commit somewhere when it returns the brief and share certificates to we will be able to use that.

mmmarcos: OK, in that case the last two options seem fine. Just follow what you think is the shortest path :)

vxgmichel (approved):
LGTM :+1:

server/tests/api_v4/authenticated/test_shamir_recovery_setup.py
AureliaDolo: I'll go for the third option

mmmarcos (approved):

touilleMan (approved):

PR #7636, opened 2024-06-24
Document `libparsec_client`'s events

touilleMan:


libparsec/crates/client/src/event_bus.rs
```diff
@@ -199,25 +199,131 @@ impl ToString for IncompatibleServerReason {
...
+    // reqwest library (see `libparsec_client_connection` implementation), hence we have no
+    // guarantee on the number of physical TCP connections being used to handle RPC and SSE.
+    //
+    // However what we know is 1) RPC and SSE connect to the same server and 2) server
+    // and client both support HTTP/2. Hence we can expect all connections to be multiplexed
+    // over a single physical TCP connection.
+    //
+    // With this in mind, the choice has been to have the connection-related events only
+    // fired by the connection monitor: if a RPC command fails due to server disconnection,
+    // it is very likely the SSE connection will also fail right away.
```
AureliaDolo, line 224: This could be moved in a module's doc comment or in the docs folder, otherwise we may never read it again ^^

AureliaDolo (approved):
Nice explanation :page_with_curl:

libparsec/crates/client/src/event_bus.rs
touilleMan: Totally, I'm thinking of adding this directly in `libparsec/README.md` so that we describe here the architecture of `libparsec` (the goal of the main crate, the layout of the sub-crates and the high-level view of how `libparsec_client` works)

PR #7641, opened 2024-06-26
Use test and testing feature flag instead of unused when needed.

AureliaDolo:
The idea is to hide behind compilation flags the functions that are used only in tests.

### DIFF
--- a/libparsec/crates/platform_storage/Cargo.toml
+++ b/libparsec/crates/platform_storage/Cargo.toml
@@ -41,7 +41,7 @@ serde-wasm-bindgen = { workspace = true }
 libparsec_tests_lite = { workspace = true }
 # Note `libparsec_tests_fixtures` enables our `test-with-testbed` feature
 libparsec_tests_fixtures = { workspace = true, default-features = false, features = ["test-with-platform-storage-testbed"] }
-libparsec_platform_storage = {path = ".",features = ["expose-test-methods"]}
+libparsec_platform_storage = { path = ".", features = ["expose-test-methods"] }
 
 [target.'cfg(target_arch = "wasm32")'.dev-dependencies]
 wasm-bindgen-test = { workspace = true }
--- a/libparsec/crates/platform_storage/Cargo.toml
+++ b/libparsec/crates/platform_storage/Cargo.toml
@@ -13,6 +13,8 @@ workspace = true
 
 [features]
 test-with-testbed = ["libparsec_testbed"]
+# to access methods that must only be used during tests
+expose-test-methods = []
 
 [dependencies]
 libparsec_types = { workspace = true }
@@ -39,6 +41,7 @@ serde-wasm-bindgen = { workspace = true }
 libparsec_tests_lite = { workspace = true }
 # Note `libparsec_tests_fixtures` enables our `test-with-testbed` feature
 libparsec_tests_fixtures = { workspace = true, default-features = false, features = ["test-with-platform-storage-testbed"] }
+libparsec_platform_storage = {path = ".",features = ["expose-test-methods"]}
 
 [target.'cfg(target_arch = "wasm32")'.dev-dependencies]
 wasm-bindgen-test = { workspace = true }
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1918,6 +1918,7 @@ dependencies = [
 name = "libparsec_client"
 version = "3.0.0-b.11+dev"
 dependencies = [
+ "libparsec_client",
  "libparsec_client_connection",
  "libparsec_platform_async",
  "libparsec_platform_device_loader",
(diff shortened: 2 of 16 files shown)


AureliaDolo:
Also there is a lot of `#[allow(unused)]`  and `#[allow(dead_code)]` without specifying why so maybe it could be worth having a closer look at that :eyes:

AureliaDolo:
So I kinda fell into a rabbit hole, some "only used in tests" functions are indeed not used.

libparsec/crates/platform_storage/Cargo.toml
```diff
@@ -39,6 +41,8 @@ serde-wasm-bindgen = { workspace = true }
 libparsec_tests_lite = { workspace = true }
 # Note `libparsec_tests_fixtures` enables our `test-with-testbed` feature
 libparsec_tests_fixtures = { workspace = true, default-features = false, features = ["test-with-platform-storage-testbed"] }
+libparsec_platform_storage = {path = ".",features = ["testing"]}
```
vxgmichel, line 44: It's not formatted like the line above.

vxgmichel (commented):
I don't understand why we sometimes have:
```rust
    #[cfg(any(test, feature = "testing"))]
    #[allow(unused)]
```
or only this:
```rust
#[cfg(any(test, feature = "testing"))]
```
or even that:
```rust
#[cfg(feature = "testing")]
```

Could you provide a bit more info about that?

AureliaDolo:
@vxgmichel 

plateform_storage is a dependency of the client. And some `debug_dump` are used in the client so we cannot just use `#[cfg(test)]` there. 

So to detail on the cases you mentionned:

```rust
    #[cfg(any(test, feature = "testing"))]
    #[allow(unused)]
```

This is because the function is *really* never used; I choose to add both flags in case it is used in the future

```rust
#[cfg(any(test, feature = "testing"))]
```

This is when the function is used both in its crate tests and in another crate's tests

```rust
#[cfg(feature = "testing")]
```
And this is when the function is only used in another crate's tests.

Could you provide a bit more info about that?

vxgmichel:
> This is because the function is _really_ never used; I choose to add both flags in case it is used in the future

Then maybe use `#[allow(unused)]` only and replace it with `#[cfg(...)]` once it's used?

> This is when the function is used both in its crate tests and in another crate's tests

Is there a way to merge `test` and `testing` somehow? I find it a bit confusing. Or maybe improve the wording.

vxgmichel (approved):
See my previous comment, I don't have a strong opinion though.

LGTM anyway :+1:

AureliaDolo:
> Then maybe use `#[allow(unused)]` only and replace it with `#[cfg(...)]` once it's used?

If we do this, there won't be any warning if someone uses it not in tests

> Is there a way to merge test and testing somehow? I find it a bit confusing. Or maybe improve the wording.

I don't think so. The `test` attribute is part of the language whereas `testing` is a feature that's defined in the project and we could choose to expose anything behind it. Maybe `test_interface` ?

vxgmichel:
> Maybe `test_interface` ?

Or `expose-test-methods`? I'll let you choose what you find less confusing.

FirelightFlagboy (approved):
LGTM,

concerning the naming of the features, we could keep `testing` or I would go with the followings:

- `expose-test-methods`
- `expose-debug-methods`

Since rust features should be additive:

- just `test-methods`
- just `debug-methods`

> I understand that as _add {tests,debug} methods_

libparsec/crates/platform_storage/Cargo.toml
```diff
@@ -39,6 +41,8 @@ serde-wasm-bindgen = { workspace = true }
 libparsec_tests_lite = { workspace = true }
 # Note `libparsec_tests_fixtures` enables our `test-with-testbed` feature
 libparsec_tests_fixtures = { workspace = true, default-features = false, features = ["test-with-platform-storage-testbed"] }
+libparsec_platform_storage = {path = ".",features = ["expose-test-methods"]}
 
 [target.'cfg(target_arch = "wasm32")'.dev-dependencies]
 wasm-bindgen-test = { workspace = true }
+libparsec_platform_storage = {path = ".",features = ["expose-test-methods"]}
```
FirelightFlagboy, line 48: The first include in `dev-dependencies` is not enough ?

AureliaDolo: Indeed

libparsec/crates/platform_storage/Cargo.toml
```diff
@@ -39,6 +41,7 @@ serde-wasm-bindgen = { workspace = true }
 libparsec_tests_lite = { workspace = true }
 # Note `libparsec_tests_fixtures` enables our `test-with-testbed` feature
 libparsec_tests_fixtures = { workspace = true, default-features = false, features = ["test-with-platform-storage-testbed"] }
+libparsec_platform_storage = {path = ".",features = ["expose-test-methods"]}
```
FirelightFlagboy, line 44: Missing space around brace and after comma:

```suggestion
libparsec_platform_storage = { path = ".", features = ["expose-test-methods"] }
```

FirelightFlagboy (approved):
LGTM with a minor change

PR #7620, opened 2024-06-19
Server SQL queries refactoring

mmmarcos:
Tasklist:

- [x] Check that consistent naming is used
- [x] Comment unused queries (to be removed later?)
- [x] Add docstrings and comments
- [x] Move `q_user_can_[read|write]_vlob` to block.py (not used elsewhere)

To be done later in another PR:
- Check that selected columns are actually used by the code
- See if some queries overlap
- Remove duplicated queries
- Extract & store internal ids for organization, user and device during the checks common to all authenticated commands

Related #7122

PR #7660, opened 2024-07-03
Refactoring server SQL queries, part 2

### LINKED ISSUE #7122: [Server v3] Sort, cleanup and uniformise SQL queries
A cleanup of the SQL queries could be useful, in particular:
- check that consistent naming is used
- check that selected columns are actually used by the code
- see if some queries overlap

Some queries are also duplicated in different component modules. Instead, one component should probably expose a method to be used by the other components.

Another improvement would be to replace all the occurrences of such pattern:
```sql
FROM realm_topic
WHERE realm = { q_realm_internal_id(organization_id="$organization_id", realm_id="$realm_id") }
```
by:
```sql
FROM realm_topic
JOIN realm ON realm_topic.realm = realm._id
JOIN organization ON realm.organization = organization._id
WHERE realm.realm_id = $realm_id
AND organization.organization_id = $organization_id
```
At the moment both approaches are used, so it's a bit confusing to navigate between queries.

Another suggestion: internal ids for organization, user and device could be extracted during the checks common to all authenticated commands (see #7119 ). The same could also performed for the internal realm id for the realm commands. This would simplify a lot of SQL queries like the example above. There was already an attempt at implementing this here:
https://github.com/Scille/parsec-cloud/blob/11efa76470e7e173f1f60153f05370566e67eb92/server/parsec/components/auth.py#L46-L67

mmmarcos:
This PR is a sequel to #7620.

- [x] Add doc about how to write SQL queries
    - It contains a summary of the discussions with @vxgmichel in #7122 
- [x] Check that selected columns are actually used by the code
- [x] See if some queries overlap
- [x] Remove duplicated queries

Closes #7122

### DIFF
--- a/docs/development/how-to-write-sql-queries.md
+++ b/docs/development/how-to-write-sql-queries.md
@@ -17,8 +17,7 @@ For convenience, queries are defined at the top of each component file in order
 to be close to the methods using them (as you often go back and forth between
 the two).
 
-Components might expose methods to other components in order to access tables
-specific tables.
+Components might expose methods to other components in order to access specific tables.
 
 ## Naming convention
 
--- a/docs/development/how-to-write-sql-queries.md
+++ b/docs/development/how-to-write-sql-queries.md
@@ -1,0 +1,84 @@
+<!-- Parsec Cloud (https://parsec.cloud) Copyright (c) BUSL-1.1 2016-present Scille SAS -->
+
+# How to write SQL queries
+
+The goal of this document is to describe best practices and conventions for
+writing SQL queries in Parsec.
+
+## Components and queries
+
+PostgreSQL components are implemented in`./server/parsec/components/postgresql/`.
+
+Each component is responsible for a specific set of tables. So to find out
+in which component you should write your query, ask yourself which component
+is responsible for the set of tables you will be requesting.
+
+For convenience, queries are defined at the top of each component file in order
+to be close to the methods using them (as you often go back and forth between
+the two).
+
+Components might expose methods to other components in order to access tables
+specific tables.
+
+## Naming convention
+
+Query names start with `_` to indicate that they should be private to the
+component file. The usual convention is to name them like:
+
+- `_q_<action>_<data>` (i.e. `_q_revoke_user`)
+- `_q_<action>_<data>_for_<parameter>` (i.e. `_q_get_user_id_for_device`)
+
+## The `Q` class
+
+`Q` instances are constructed from SQL query strings containing variables
+(such as `$user_id`). These variables are replaced by [positional parameters](https://www.postgresql.org/docs/current/sql-expressions.html#SQL-EXPRESSIONS-PARAMETERS-POSITIONAL)
+(such as `$1`).
+
+When the Q object is called, the values are supplied after the query.
+
+```python
+_q_get_user = Q("SELECT name FROM user WHERE user_id = $user_id")
+print(_q_get_user.sql)           # SELECT name FROM user WHERE user_id = $1
+print(_q_get_user(user_id=42)) # ['SELECT name FROM user WHERE user_id = $1', 42]
+```
+
+## `_make_q_` functions
+
+When two queries differ only slightly, a function can be used to generate them.
+
+```python
+def _make_q_lock_common_topic(for_update: bool = False, for_share=False) -> Q:
+    assert for_update ^ for_share
+    share_or_update = "SHARE" if for_share else "UPDATE"
+    return Q(f"""
+      SELECT last_timestamp
+      FROM common_topic
+      JOIN organization ON common_topic.organization = organization._id
+      WHERE organization_id = $organization_id
+      FOR {share_or_update}
+    """)
+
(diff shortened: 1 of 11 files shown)


docs/development/how-to-write-sql-queries.md
```diff
@@ -0,0 +1,84 @@
...
+Each component is responsible for a specific set of tables. So to find out
+in which component you should write your query, ask yourself which component
+is responsible for the set of tables you will be requesting.
+
+For convenience, queries are defined at the top of each component file in order
+to be close to the methods using them (as you often go back and forth between
+the two).
+
+Components might expose methods to other components in order to access tables
+specific tables.
```
AureliaDolo, line 21:
```suggestion
Components might expose methods to other components in order to access specific tables.
```

server/parsec/components/postgresql/user.py
```diff
@@ -487,18 +471,18 @@ async def _create_user(
         if result != "INSERT 0 1":
             assert False, f"Insertion error: {result}"
 
-        # Finally make sure there is only one non-revoked user with this human handle
+        # Finally make sure there is only one active user with this human handle
```
AureliaDolo, line 474: Just to be sure, active means nor revoked, nor frozen  ?

mmmarcos: Users are either active or revoked. Only active users are taken into account when invoicing customers.

A frozen user is "temporarily" prevented from authenticating to Parsec server, but is still considered an active user. The frozen status requires an action from an admin: either revoke or unfreeze the user.

server/parsec/components/postgresql/realm.py
```diff
@@ -100,7 +79,18 @@ def _make_q_lock_realm(for_update: bool = False, for_share=False) -> Q:
...
+            SELECT realm_user_role.role
+            FROM realm_user_role
+            WHERE
+                realm_user_role.realm = realm._id
+                AND realm_user_role.user_ = { q_user_internal_id(organization="realm.organization", user_id="$user_id") }
+            ORDER BY certified_on DESC
+            LIMIT 1
+        ),
+        NULL
+    ) AS role,
```
vxgmichel, line 93: Maybe a `JOIN` would be more readable than a subquery? Or maybe it's too tricky to write due to `COALESCE`?

vxgmichel (approved):
LGTM !

And nice guide as well, clear and to the point :+1:

server/parsec/components/postgresql/realm.py
mmmarcos: Maybe... actually I think I skipped this query because I wasn't sure about it. 

`COALESCE` it supposed to returns the first non-`NULL` element, so:
1. if the 1st element (the Select) is `NULL`, check next element
2. the 2nd element is `NULL`, check next element
3. No more elements, return `NULL` 

So this query
```sql
COALESCE(
    (
        SELECT --...
    ),
    NULL
) AS role,
```
is equivalent to:
```sql
COALESCE(
    (
        SELECT --...
    )
) AS role,
```
which is equivalent to:
```sql
    (
        SELECT --...
    ) AS role,
```

mmmarcos: I've left the subquery but removed the `COALESCE`

docs/development/how-to-write-sql-queries.md
```diff
@@ -0,0 +1,83 @@
...
+
+In Parsec, you'll sometimes see code like this:
+
+```sql
+AND COALESCE(revoked_on > some_parameter, TRUE)
+```
+
+This takes advantage of the fact the following fact that ordinary comparison
+operators yield null, not true or false, when either input is null. So the
+previous example will return `TRUE` if `revoked_on` is null.
```
FirelightFlagboy, line 83:
```suggestion
previous example will return `TRUE` if either `revoked_on` or `some_parameter are null.
```

I understand that `COALESCE` will return the value of the first expression that is not null.
But is it true that when doing a comparaison `a > b` if `a` is null then the result is also null ?
And does that apply if `b` is null ?

mmmarcos: Yep. Acording to the docs: 

> Ordinary comparison operators yield null (signifying “unknown”), not true or false, when either input is null. For example, 7 = NULL yields null, as does 7 <> NULL. When this behavior is not suitable, use the IS [ NOT ] DISTINCT FROM predicates
 https://www.postgresql.org/docs/current/functions-comparison.html

FirelightFlagboy: If I understand that correctly both `5 > NULL` and `NULL > 5` return NULL

mmmarcos: Yes, in the lines you commented I put only `revoked_on` because I was referring to that concrete example. 
You prefer to specify that works with either input ?

FirelightFlagboy: Yes, I would prefer or specify if the firsts expressions are null

mmmarcos: Done

FirelightFlagboy (approved):

PR #7662, opened 2024-07-03
RFC 1011 - Non blocking invite transport

vxgmichel:


libparsec/crates/protocol/schema/authenticated_cmds/invite_greeter_step.json5
```diff
@@ -0,0 +1,207 @@
...
+                        ]
+                    },
+                    {
+                        "name": "WaitPeerAcknowledgment",
+                        "discriminant_value": "8"
+                    }
+                ]
+            },
+            {
+                // ClaimerStep should be identical to the one in invite_claimer_step.json5
```
AureliaDolo, line 139: Could it be defined in types instead of being a nested type?

docs/rfcs/1011-non-blocking-invite-transport.md
```diff
@@ -0,0 +1,915 @@
...
+
+The second change is related to the concept of greeters. In Parsec v2, only the administrator that created an invitation is able to greet the invited user. This is something we want to change to open the way to easier invitations, where any administrator can greet a user that has been invited. This may be configurable of course, but the transport should allow for multiple greeters per invitation. Note that this is already the case for shamir invitations, where the greeters are the shamir recipients. For this reason, the new design can draw inspiration from what was done during the shamir implementation. More precisely, each invitation will have its corresponding set of greeters which is different for each invitation type:
+- for user invitation, this would be the administrators
+- for device invitation, this would be the user alone
```
AureliaDolo, line 25: Would it make sense if an administrator could greets an another user's device ? (ex: if in an organization only one kind of device is authorized and it must be admin approved)

docs/rfcs/1011-non-blocking-invite-transport.md
```diff
@@ -0,0 +1,915 @@
...
+
+Mutable properties:
+- attempts ids: list of attempts
+   * Invariant: most recent attempt is active, all the rest is cancelled
+
+Getters:
+- get active attempt
+
+Methods:
+- cancel active attempt (and create a new one)
```
AureliaDolo, line 113: Is the creation of a new attempt optional ?

docs/rfcs/1011-non-blocking-invite-transport.md
```diff
@@ -0,0 +1,915 @@
...
+- the invitation exists
+- the invitation is pending (i.e. not completed or cancelled)
+- the author is allowed (i.e. it is part of the greeters)
+- the attempt exists
+- the attempt is active (i.e. joined and not cancelled)
+
+Then, the active attempt is cancelled with the provided reason and a new active attempt is created. The cancellation reason covers several cases:
+- the user manually cancelled the attempt
+- the hashed nonce didn't match the provided nonce
+- the SAS code communicated to the user was invalid
```
AureliaDolo, line 269: So if the user misclicks, they must start over from the beginning ? Could there be a way to just go back to the previous state ?

AureliaDolo (commented):
Nice work :tada:

docs/rfcs/1011-non-blocking-invite-transport.md
vxgmichel: During a device invitation, the private part of the existing parsec user key is sent to the new device. An administrator doesn't know this key, so it wouldn't be able to invite a device for another user.

> (ex: if in an organization only one kind of device is authorized and it must be admin approved)

Maybe something for another RFC :)

docs/rfcs/1011-non-blocking-invite-transport.md
vxgmichel: My idea was to have the following invariant:
- there always exactly one active attempt by channel and it is the most recent
In order to respect the invariant, cancelling the active attempt implies the creation of a new one, atomically.
In practice, the creation of the active attempt could also be done lazily when needed.

But that's a proposal, there might be another way to handle this.

docs/rfcs/1011-non-blocking-invite-transport.md
vxgmichel: This is up to the client logic, but there has to be a reason defined so that the client is able to communicate this information to the server (i.e "I'm cancelling this attempt because the SAS code didn't match, indicating there might be a MitM in the channel").

In practice, we only display 4 code, so recovering on misclicks might lead to weaker SAS code verification (e.g imagine an impatient user just trying the different code until they find the right one).

libparsec/crates/protocol/schema/authenticated_cmds/invite_greeter_step.json5
vxgmichel: I agree that they should, but I'm not sure they can. So far, the `"variants"` key is only used in `"nested_types"`, so I don't think the macros support the definition an enum from the json schema (without using `nested_types`).

Another approach would be to use a nested type from another command in a command definition, but macros would need to be adapted as well. My thoughts were to duplicate those `ClaimerStep` and `GreeterStep` definition, add a comment about this, a refactor this by updating the macros later.

@touilleMan could you share your thoughts about this?

docs/rfcs/1011-non-blocking-invite-transport.md
FirelightFlagboy: With the current proposal, created a new active attempts could not be done lazily.
As you said it would break your Invariant that always the most recent is activate.

But I think that would be doable if we had 2 fields:

```typescript
interface Attempts {
	activate_attempt: ID | Null;
	cancelled_attempts: ID[];
}
```

docs/rfcs/1011-non-blocking-invite-transport.md
```diff
@@ -0,0 +1,1226 @@
...
+| 5     | `signify_trust`    | acknowledgment | `wait_peer_trust`   |                |
+| 6     | `get_payload`      |                | `send_payload`      | payload        |
+| 7     | `send_payload`     | payload        | `get_payload`       |                |
+| 8     | `wait_peer_ack`    |                | `acknowledge`       | acknowledgment |
+
+Those steps work very similarly to Parsec v2. There are a couple of differences though:
+- Steps are not longer counted as `1`, `2a`, `2b`, `3a`, `3b`, and `4` like in version 2
+- Three different exchanges were performed in steps `2a` and `2b`. Now those exchanges are expanded in steps `1`, `2` and `3`.
+- Step `4` in version 2 could be used for several exchanges, until the greeter marked an exchange with `last=True`. Instead, those exchanges are expanded in steps `6` and `7`
+- In version 2, the last step was used to complete the invite. In version 3 this is now done with a dedicated `invite_complete` command, so we add a last step (index `8`) for the claimer to acknowledge that it's been able to successfully process the greeter payload.
```
FirelightFlagboy, line 1037: We don't need a dedicated step in the table above, because by using the `invite_complete` command we mutate the step of the invitation to `complete`, Am I right?

docs/rfcs/1011-non-blocking-invite-transport.md
```diff
@@ -0,0 +1,1226 @@
...
+
+The `GreeterStep` nested data type is then define as such:
+```json5
+            {
+                "name": "GreeterStep",
+                "discriminant_field": "step",
+                "variants": [
+                    {
+                        "name": "WaitPeer",
+                        "discriminant_value": "0",
```
FirelightFlagboy, line 1047: Note that `discriminant_value` are used as literal (as string not integer).
Maybe we could use the `name` as discriminant instead of `int` ?

docs/rfcs/1011-non-blocking-invite-transport.md
```diff
@@ -0,0 +1,1226 @@
...
+```
+
+## Typical workflow
+
+Typical workflow
+----------------
+
+Here's how a typical workflow for a user invitation can play out:
+
+```python
```
FirelightFlagboy, line 1181: Are you interested by a Sequence Diagram ? you could do that using the mermaid syntax

https://mermaid.js.org/syntax/sequenceDiagram.html

docs/rfcs/1011-non-blocking-invite-transport.md
vxgmichel: > With the current proposal, created a new active attempts could not be done lazily.

How so? IMO, as long as methods like `get_active_attempt` always return an active attempt, the invariant is respected. That means that `get_active_attempt` should be able to create the active attempt if it does not exist. That's what I meant by lazy creation. But this is an implementation detail, and we could also choose to atomically create a new attempt when the previous one is cancelled. Whether the first solution or the second solution is implemented it does not matter, it should be transparent for the caller that the invariant is respected in both cases.

> But I think that would be doable if we had 2 fields:

I think I shouldn't have put `attempts ids` as a mutable property of the `channel` because in practice, we'll have a postgre table called `attempts` for storing all attempts id and their internal state. That means that methods like `get_active_attempt` or `cancel_active_attempt` are merely requests over `attempts` table. The invariant is an invariant to maintain in the `attempts` table, not in the internal state of the `channel`.

So all things considered, a `channel` have no internal state meaning it doesn't have a corresponding table either. However, I think it's still useful that the concept exists to properly name things in the code.

docs/rfcs/1011-non-blocking-invite-transport.md
```diff
@@ -0,0 +1,1226 @@
...
+
+### Invitation
+
+An invitation is used to verify the identity of a user (called the claimer) on a specific device. The users allowed to verify the identity are called greeters. The invitation can be of 3 types:
+- User invitation
+- Device invitation
+- Shared recovery (shamir) invitation
+
+An invitation is identified by:
+- organization id
```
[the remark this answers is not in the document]
vxgmichel: Yes, identification is immutable by definition

docs/rfcs/1011-non-blocking-invite-transport.md
vxgmichel: Yes because a completed attempt is different from a completed invitation (especially for the shamir):

See in the `invite_complete` section:

> In the previous implementation, this was done automatically when the last exchange was performed. However, this is confusing in some cases:
> - Maybe the last exchange worked but the reply didn't reach one of the peers. In this case, the peer cannot retry the exchange since the invite is now completed.
> - In a shared recovery (shamir) invitation, the invitation must not complete after a successful attempt since there might be more exchanges to complete with other recipients.

docs/rfcs/1011-non-blocking-invite-transport.md
vxgmichel: Yes I'm not fully decided on those definitions. However I think it's useful to have the number of step as constant somewhere.

docs/rfcs/1011-non-blocking-invite-transport.md
vxgmichel: Right I'll try that :+1:

docs/rfcs/1011-non-blocking-invite-transport.md
```diff
@@ -0,0 +1,1226 @@
...
+```
+
+
+#### `invite_claimer_start_attempt` command
+
+This is the new invited command introduced for the claimer to start an attempt and get the corresponding attempt ID.
+
+In order for the claimer to start an attempt, a couple of checks have to be performed:
+- the author exists
+- the author is allowed (i.e. it is part of the greeters)
```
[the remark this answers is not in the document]
vxgmichel: Good catch :+1:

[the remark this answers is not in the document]
vxgmichel: 9dba2c1709716f303a0b4502bad25d0ac6852294

docs/rfcs/1011-non-blocking-invite-transport.md
FirelightFlagboy: We would come back eventually to the hybrid method with both the name + index :sweat_smile:

docs/rfcs/1011-non-blocking-invite-transport.md
```diff
@@ -0,0 +1,1226 @@
...
+            }
+        ]
+    }
+]
+```
+
+
+#### `invite_greeter_cancel_attempt` command
+
+Two new commands are introduced to cancel an attempt, which registers a cancellation reason for the given attempt. For the greeter, an authenticated command is exposed. After using `invite_greeter_cancel_attempt`, the greeter can start a new attempt by using the `invite_greeter_start_attempt` command.
```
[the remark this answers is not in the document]
vxgmichel: Ah I should have been more explicit about the difference between an active attempt and a joined attempt.

There's always exactly one active attempt, but this attempt might be in an idle state (i.e. no peer has started it). Then a first peer calls `start_attempt`, effectively joining the attempt, and the attempt id is returned to the peer. The attempt is now in an half joined state. Then the second peer starts the attempt as well, getting the same attempt id. The attempt is now in a fully joined state.

Feel free to propose a new wording or semantics if you find the `cancelled/active` distinction or `idle/joined` distinction unclear :)

docs/rfcs/1011-non-blocking-invite-transport.md
```diff
@@ -0,0 +1,1231 @@
...
+            }
+        ]
+    }
+]
+```
+
+
+#### `invite_greeter_step` command
+
+This authenticated command, along with `invite_greeter_step`, is the main route used to establish a secure channel between both peers.
```
mmmarcos, line 602:
```suggestion
This authenticated command, along with `invite_claimer_step`, is the main route used to establish a secure channel between both peers.
```

mmmarcos (commented):
Kudos for this RFC! :100:

docs/rfcs/1011-non-blocking-invite-transport.md
```diff
@@ -0,0 +1,1231 @@
...
+
+## Design
+
+### New commands
+
+7 commands are added in total, 4 authenticated commands and 3 invited commands.
+
+#### `invite_greeter_start_attempt` command
+
+Two new commands are introduced to start an attempt, which returns an attempt ID used in the subsequent queries. This attempt ID is the same for the greeter and the claimer. For the greeter, an authenticated command is exposed.
```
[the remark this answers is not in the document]
vxgmichel: `authenticated`, because the greeter has to already be part of the system

docs/rfcs/1011-non-blocking-invite-transport.md
```diff
@@ -0,0 +1,1231 @@
...
+            }
+        ]
+    }
+]
+```
+
+
+#### `invite_greeter_cancel_attempt` command
+
+Two new commands are introduced to cancel an attempt, which registers a cancellation reason for the given attempt. For the greeter, an authenticated command is exposed. After using `invite_greeter_cancel_attempt`, the greeter can start a new attempt by using the `invite_greeter_start_attempt` command.
```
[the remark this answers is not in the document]
vxgmichel: `authenticated`, because the greeter has to already be part of the system

docs/rfcs/1011-non-blocking-invite-transport.md
```diff
@@ -0,0 +1,1231 @@
...
+
+The second change is related to the concept of greeters. In Parsec v2, only the administrator that created an invitation is able to greet the invited user. This is something we want to change to open the way to easier invitations, where any administrator can greet a user that has been invited. This may be configurable of course, but the transport should allow for multiple greeters per invitation. Note that this is already the case for shamir invitations, where the greeters are the shamir recipients. For this reason, the new design can draw inspiration from what was done during the shamir implementation. More precisely, each invitation will have its corresponding set of greeters which is different for each invitation type:
+- for user invitation, this would be the administrators
+- for device invitation, this would be the user alone
+- for shamir invitation, this would be the shamir recipients
+
+> [!NOTE]
+> This set may evolve overtime as users can be revoked and user profiles can change.
```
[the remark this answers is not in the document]
vxgmichel: 14206e6270bd461622d80fec478d13715fed2100

docs/rfcs/1011-non-blocking-invite-transport.md
```diff
@@ -0,0 +1,1231 @@
...
+It's possible for the active attempt to be in a state where the greeter has already joined. Usually, previous attempts are expected to be cancelled by one of the peers, but one can think of many cases where this cancellation didn't occur (e.g. the greeter device shutdown, or another device has an ongoing attempt running that hasn't been cancelled by the user).
+
+In this case, the `invite_greeter_start_attempt` command cancels the active attempts automatically (using a dedicated reason), and joins the new active attempt. This means that the last device calling `start_attempt` always has the priority.
+
+If no cancellation of the active attempt is required, the greeter simply joins it and the command returns the corresponding attempt ID.
+
+Pseudo code:
+
+```python
+invite_greeter_start_attempt(
```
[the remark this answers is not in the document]
vxgmichel: 14206e6270bd461622d80fec478d13715fed2100

docs/rfcs/1011-non-blocking-invite-transport.md
```diff
@@ -0,0 +1,1231 @@
...
+Methods:
+- set claimer data
+- set greeter data
+
+
+## Design
+
+### New commands
+
+7 commands are added in total, 4 authenticated commands and 3 invited commands.
```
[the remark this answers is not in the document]
vxgmichel: 14206e6270bd461622d80fec478d13715fed2100

docs/rfcs/1011-non-blocking-invite-transport.md
```diff
@@ -0,0 +1,1231 @@
...
+```
+
+
+#### `invite_claimer_start_attempt` command
+
+This is the new invited command introduced for the claimer to start an attempt and get the corresponding attempt ID.
+
+In order for the claimer to start an attempt, a couple of checks have to be performed:
+- the greeter exists
+- the greeter is allowed (i.e. it is part of the greeters)
```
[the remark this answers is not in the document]
vxgmichel: 14206e6270bd461622d80fec478d13715fed2100

docs/rfcs/1011-non-blocking-invite-transport.md
```diff
@@ -0,0 +1,1231 @@
...
+
+> [!NOTE]
+> The state of the invitation doesn't have to be checked since it is already checked by the `/invited` HTTP route.
+
+Otherwise this command has the same logic as `invite_greeter_start_attempt`.
+
+Pseudo code:
```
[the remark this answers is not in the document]
vxgmichel: 14206e6270bd461622d80fec478d13715fed2100

docs/rfcs/1011-non-blocking-invite-transport.md
```diff
@@ -0,0 +1,1231 @@
...
+                "status": "attempt_not_found"
+            },
+            {
+                // The author did not join the attempt
+                // This should not happen, since joining is required to get the attempt ID
+                "status": "attempt_not_joined"
+            },
+            {
+                // The attempt has been already cancelled
+                "status": "attempt_already_cancelled",
```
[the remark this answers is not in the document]
vxgmichel: `already` is there when needed to avoid confusion (e.g `invitiation_already_completed` is more clear than `invitation_completed` when calling `invite_complete`)

docs/rfcs/1011-non-blocking-invite-transport.md
```diff
@@ -0,0 +1,1231 @@
...
+- the hashed nonce didn't match the provided nonce
+- the SAS code communicated to the user was invalid
+- the payload could not be deciphered
+- the payload could not be deserialized
+- the payload contained invalid information
+- the attempt has been automatically cancelled by a new `start_attempt` command
+
+It is also registered that the greeter cancelled the attempt, so that this information can be provided in subsequent `invitation_cancelled` replies, along with the corresponding reason and timestamp. This way, the front-end application has all the information to properly communicate to the user what happened during the attempt.
+
+Pseudo code:
```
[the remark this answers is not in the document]
vxgmichel: 14206e6270bd461622d80fec478d13715fed2100

docs/rfcs/1011-non-blocking-invite-transport.md
```diff
@@ -0,0 +1,1231 @@
...
+
+```python
+invite_greeter_cancel_attempt(
```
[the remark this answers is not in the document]
vxgmichel: 14206e6270bd461622d80fec478d13715fed2100

docs/rfcs/1011-non-blocking-invite-transport.md
```diff
@@ -0,0 +1,1231 @@
...
+
+> [!NOTE]
+> The state of the invitation doesn't have to be checked since it is already checked by the `/invited` HTTP route.
+
+Otherwise this command has the same logic as `invite_greeter_cancel_attempt`.
+
+Pseudo-code:
+
+```python
+invite_claimer_cancel_attempt(
```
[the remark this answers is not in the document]
vxgmichel: 14206e6270bd461622d80fec478d13715fed2100

docs/rfcs/1011-non-blocking-invite-transport.md
vxgmichel: 14206e6270bd461622d80fec478d13715fed2100

docs/rfcs/1011-non-blocking-invite-transport.md
```diff
@@ -0,0 +1,1231 @@
...
+- a `not_ready` status otherwise
+
+In the second case, the author can poll the command with the same parameters until it gets the claimer step data.
+
+If the submitted step differs from the one that has already been registered, a dedicated `step_mismatch` status is returned.
+
+Pseudo-code:
+
+```python
+invite_greeter_step(
```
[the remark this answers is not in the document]
vxgmichel: 14206e6270bd461622d80fec478d13715fed2100

docs/rfcs/1011-non-blocking-invite-transport.md
```diff
@@ -0,0 +1,1231 @@
...
+                ]
+            }
+        ]
+    }
+]
+```
+
+#### `invite_claimer_step` command
+
+This invited command, along with `invite_claimer_step`, is the main route used to establish a secure channel between both peers.
```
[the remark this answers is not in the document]
vxgmichel: 14206e6270bd461622d80fec478d13715fed2100

docs/rfcs/1011-non-blocking-invite-transport.md
```diff
@@ -0,0 +1,1231 @@
...
+
+> [!NOTE]
+> The state of the invitation doesn't have to be checked since it is already checked by the `/invited` HTTP route.
+
+Otherwise this command has the same logic as `invite_greeter_start_attempt`.
+
+Pseudo code:
+
+```python
+invite_claimer_start_attempt(
```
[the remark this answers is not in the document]
vxgmichel: 14206e6270bd461622d80fec478d13715fed2100

docs/rfcs/1011-non-blocking-invite-transport.md
```diff
@@ -0,0 +1,1226 @@
...
+It is also registered that the greeter cancelled the attempt, so that this information can be provided in subsequent `invitation_cancelled` replies, along with the corresponding reason and timestamp. This way, the front-end application has all the information to properly communicate to the user what happened during the attempt.
+
+Pseudo code:
+
+```python
+invite_greeter_cancel_attempt(
+    token: InvitationToken,
+    attempt: AttemptID,
+    reason: CancelledAttemptReason,
+
```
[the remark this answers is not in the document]
FirelightFlagboy: @vxgmichel the extra line is still here, it's expected ?

docs/rfcs/1011-non-blocking-invite-transport.md
```diff
@@ -0,0 +1,1238 @@
...
+The second change is related to the concept of greeters. In Parsec v2, only the administrator that created an invitation is able to greet the invited user. This is something we want to change to open the way to easier invitations, where any administrator can greet a user that has been invited. This may be configurable of course, but the transport should allow for multiple greeters per invitation. Note that this is already the case for shamir invitations, where the greeters are the shamir recipients. For this reason, the new design can draw inspiration from what was done during the shamir implementation. More precisely, each invitation will have its corresponding set of greeters which is different for each invitation type:
```
touilleMan, line 23:
```suggestion
The second change is related to the concept of greeters. In Parsec v2, only the administrator that created an invitation is able to greet the invited user. This is something we want to change to open the way to easier invitations, where any administrator can greet a user that has been invited. This may be configurable of course, but the transport should allow for multiple greeters per invitation. Note that this is already the case for Shamir invitations, where the greeters are the Shamir recipients. For this reason, the new design can draw inspiration from what was done during the Shamir implementation. More precisely, each invitation will have its corresponding set of greeters which is different for each invitation type:
```

docs/rfcs/1011-non-blocking-invite-transport.md
```diff
@@ -0,0 +1,1238 @@
...
+- for user invitation, this would be the administrators
+- for device invitation, this would be the user alone
+- for shamir invitation, this would be the shamir recipients
+
+> [!NOTE]
+> This set may evolve over time as users can be revoked and user profiles can change.
```
touilleMan, line 29: the "over time" is a bit misleading here: at fist I thought you were referring to a potential evolution in Parsec

On top of that the set for a given device invitation and shamir invitation are set in stone, only the set for user invitation can change when user profile is updated.

docs/rfcs/1011-non-blocking-invite-transport.md
```diff
@@ -0,0 +1,1238 @@
...
+  * for shamir invitation: all recipients
+- get channel for greeter
+
+Methods:
+- create invitation
+- cancel invitation
+- complete invitation
+
+
+### Channel
```
touilleMan, line 97: Channel is a very broad word (though some would say it's still better than "conduit" :trollface: )

I think it's okay as long as we are always referring to it as "invitation channel" (e.g. having a `invitation_channel` table in the database)

On top of that, the attempt part look more like a channel to me (as data flow through a channel, which is what is done in the attempts)... especially given the very first description of the attempt describes it as "trying to establish a secure communication channel" !

So given the "Channel" thing is about having meeting point between the claimer and a given greeter, why not calling it `appointment` ? (by analogy to e.g. "My wife have an appointment to the l'Haÿ-les-Roses préfecture for her visa, I hope it won't take too long...")

docs/rfcs/1011-non-blocking-invite-transport.md
```diff
@@ -0,0 +1,1238 @@
...
+                    "type": "CancelledAttemptReason"
+                }
+            ]
+        },
+        "reps": [
+            {
+                "status": "ok",
+                "fields": [
+                    {
+                        "name": "attempt",
```
touilleMan, line 431: The attemp ID has been provided in the request, so there is no need to return it !

docs/rfcs/1011-non-blocking-invite-transport.md
```diff
@@ -0,0 +1,1238 @@
...
+                "status": "author_not_allowed"
+            },
+            {
+                // The attempt id doesn't correspond to any existing attempt
+                "status": "attempt_not_found"
+            },
+            {
+                // The author did not join the attempt
+                // This should not happen, since joining is required to get the attempt ID
+                "status": "attempt_not_joined"
```
touilleMan, line 459: Given this should not happen, I would be more practical to just merge this error into `attempt_not_found`:
```json5
{
                // No attempt joined by the user has this attempt ID
                "status": "attempt_not_found"
}
```

Just like for the merge of `greeter_not_found`&`greeter_not_allowed` I was talking above, this merge simplify the code and prevent probing attack (which are more theoretical than anything given our IDs are UUID4 and the routes have authentication, but still ! ^^)

docs/rfcs/1011-non-blocking-invite-transport.md
```diff
@@ -0,0 +1,1238 @@
...
+[
+    {
+        "major_versions": [
+            4
+        ],
+        "req": {
+            "cmd": "invite_greeter_cancel_attempt",
+            "fields": [
+                {
+                    "name": "token",
```
touilleMan, line 413: We should remove `token` and only rely on `attempt` here: the attempt ID should be a unique identifier, hence there is no need for the couple token+attempID

On top of that, removing the `token` field save us from handling the weird edge cases where token and attemptID are both valid but point on unrelated things.

docs/rfcs/1011-non-blocking-invite-transport.md
```diff
@@ -0,0 +1,1238 @@
...
+```json5
+[
+    {
+        "major_versions": [
+            4
+        ],
+        "req": {
+            "cmd": "invite_claimer_cancel_attempt",
+            "fields": [
+                {
```
touilleMan, line 529: Same as above: token+greeter+attemptID makes rooms for inconsistency, we should only have a `attempt` field instead

docs/rfcs/1011-non-blocking-invite-transport.md
```diff
@@ -0,0 +1,1238 @@
...
+            //     // The invitation has already been completed
+            //     "status": "invitation_completed"
+            // },
+            // {
+            //     // The invitation has been cancelled
+            //     "status": "invitation_cancelled"
+            // },
+            {
+                // The provided greeter ID doesn't correspond to any existing greeter
+                "status": "greeter_not_found"
```
touilleMan, line 568: with no `greeter` field we don't need `greeter_not_found` & `greeter_not_allowed` \o/

docs/rfcs/1011-non-blocking-invite-transport.md
```diff
@@ -0,0 +1,1238 @@
...
+                "fields": [
+                    {
+                        "name": "attempt",
+                        "type": "AttemptID"
+                    }
+                ]
+            },
+            {
+                // The invitation token doesn't correspond to any existing invitation
+                "status": "invitation_not_found"
```
touilleMan, line 438: without `token` field, we don't need `invitation_not_found` \o/

docs/rfcs/1011-non-blocking-invite-transport.md
```diff
@@ -0,0 +1,1238 @@
...
+            {
+                // The provided greeter ID doesn't correspond to any existing greeter
+                "status": "greeter_not_found"
+            },
+            {
+                // The greeter is not part of the allowed greeters for this invitation
+                "status": "greeter_not_allowed"
+            },
+            {
+                // The attempt id doesn't correspond to any existing attempt
```
touilleMan, line 575:
```suggestion
                 // No attempt joined by the greeter corresponds to this attempt ID
```

(the change is here to show that using existing attempt ID that is unrelated to this greeter will end up with this error)

docs/rfcs/1011-non-blocking-invite-transport.md
```diff
@@ -0,0 +1,1238 @@
...
+    - [`invite_claimer_step`](#invite_claimer_step-command)
+- 4 new authenticated commands
+    - [`invite_greeter_start_attempt`](#invite_greeter_start_attempt-command)
+    - [`invite_greeter_cancel_attempt`](#invite_greeter_cancel_attempt-command)
+    - [`invite_greeter_step`](#invite_greeter_step-command)
+    - [`invite_complete`](#invite_complete-command)
+
+#### `invite_greeter_start_attempt` command
+
+Two new commands are introduced to start an attempt, which returns an attempt ID used in the subsequent queries. This attempt ID is the same for the greeter and the claimer. For the greeter, an authenticated command is exposed.
```
touilleMan, line 186:
```suggestion
Two new commands are introduced to start an attempt, which returns an attempt ID used in the subsequent queries. This attempt ID is the same for the greeter and the claimer and is a UUID that uniquely identify the attempt. For the greeter, an authenticated command is exposed.
```

See my remarks below, I think it's much better to only use the attempt ID to identify the attempt that rely on the couple token+attempt ID.

On top of that I don't think it's interesting to have the attempt ID being an index (so first attempt has ID 0, next one 1 etc.) as we don't care of the number of previous attempts made, and handle counter is always more complicated to implement than plain simple UUID.

docs/rfcs/1011-non-blocking-invite-transport.md
touilleMan: I don't mind adding diagram, but I think the current pseudo code snippet is very convenient to read in its current form ;-)

docs/rfcs/1011-non-blocking-invite-transport.md
```diff
@@ -0,0 +1,1238 @@
...
+
+Here's how a typical workflow for a user invitation can play out:
+
+```python
+# An admin creates a new invitation
+invite_new_user(...) -> token
+
+# Later the user receives the invitation and starts polling
+invite_claimer_start_attempt(token, greeter) -> attempt ID
+invite_claimer_step(token, greeter, attempt, claimer_step_0) -> NotReady
```
touilleMan, line 1199:
```suggestion
invite_claimer_step(attempt, claimer_step_0) -> NotReady
```

docs/rfcs/1011-non-blocking-invite-transport.md
```diff
@@ -0,0 +1,1238 @@
...
+
+# Later a greeter connects
+invite_greeter_start_attempt(token) -> attempt ID
+invite_greeter_step(token, attempt, greeter_step_0) -> claimer_step_0
```
touilleMan, line 1203:
```suggestion
invite_greeter_step(attempt, greeter_step_0) -> claimer_step_0
```

docs/rfcs/1011-non-blocking-invite-transport.md
touilleMan: I also think it would be better to have name as `Step0WaitPeer`, otherwise it's too cumbersome to work with those types, especially given there is 8 of them !

docs/rfcs/1011-non-blocking-invite-transport.md
```diff
@@ -0,0 +1,1238 @@
...
+            },
+            {
+                "status": "invitation_not_found"
+            },
+            {
+                "status": "invitation_already_completed"
+            },
+            {
+                "status": "invitation_cancelled"
+            }
```
touilleMan, line 996: There should also be a `author_not_allowed` status

docs/rfcs/1011-non-blocking-invite-transport.md
```diff
@@ -0,0 +1,1238 @@
...
+```
+
+
+#### `invite_greeter_step` command
+
+This authenticated command, along with `invite_claimer_step`, is the main route used to establish a secure channel between both peers.
+
+It is used by the greeter to submit the data for a given step. The step is identified by:
+- the author of the command (being the greeter)
+- the invitation token
```
touilleMan, line 613:
```suggestion
- the attempt (itself also identifying the invitation)
```

docs/rfcs/1011-non-blocking-invite-transport.md
```diff
@@ -0,0 +1,1238 @@
...
+- the type of the submitted step data
+
+For the step data to be accepted, it has to pass the following the checks:
```
touilleMan, line 616:
```suggestion
For the step data to be accepted, it has to pass the following checks:
```

docs/rfcs/1011-non-blocking-invite-transport.md
```diff
@@ -0,0 +1,1238 @@
...
+[
+    {
+        "major_versions": [
+            4
+        ],
+        "req": {
+            "cmd": "invite_greeter_step",
+            "fields": [
+                {
+                    "name": "token",
```
touilleMan, line 665: remove `token`, only use `attempt`

docs/rfcs/1011-non-blocking-invite-transport.md
```diff
@@ -0,0 +1,1238 @@
...
+                    }
+                ]
+            },
+            {
+                // The claimer has not submitted its step yet
+                "status": "not_ready"
+            },
+            {
+                // The invitation token doesn't correspond to any existing invitation
+                "status": "invitation_not_found"
```
touilleMan, line 694: remove and only rely on `attempt_not_found`

docs/rfcs/1011-non-blocking-invite-transport.md
```diff
@@ -0,0 +1,1238 @@
...
+]
+```
+
+#### `invite_claimer_step` command
+
+This invited command, along with `invite_greeter_step`, is the main route used to establish a secure channel between both peers.
+
+It is used by the claimer to submit the data for a given step. The step is identified by:
+- the provided greeter
+- the invitation token
```
touilleMan, line 773:
```suggestion
- the attempt (itself also identifying the invitation)
```

docs/rfcs/1011-non-blocking-invite-transport.md
```diff
@@ -0,0 +1,1238 @@
...
+[
+    {
+        "major_versions": [
+            4
+        ],
+        "req": {
+            "cmd": "invite_claimer_step",
+            "fields": [
+                {
+                    "name": "token",
```
touilleMan, line 828: remove `token` and `greeter`, only rely on `attempt`

docs/rfcs/1011-non-blocking-invite-transport.md
```diff
@@ -0,0 +1,1238 @@
...
+            //     // The invitation has already been completed
+            //     "status": "invitation_completed"
+            // },
+            // {
+            //     // The invitation has been cancelled
+            //     "status": "invitation_cancelled"
+            // },
+            {
+                // The provided greeter ID doesn't correspond to any existing greeter
+                "status": "greeter_not_found"
```
touilleMan, line 876: remove `greeter_not_found`/`greeter_not_allowed`

touilleMan (dismissed):

docs/rfcs/1011-non-blocking-invite-transport.md
vxgmichel: > See my remarks below, I think it's much better to only use the attempt ID to identify the attempt that rely on the couple token+attempt ID.

I don't agree it's better, I would argue it's worse. Your main argument is that this makes room for inconsistency, as pointed in this comment:

> Same as above: token+greeter+attemptID makes rooms for inconsistency, we should only have a attempt field instead

However the attempt is always part of a specific invitation and a specific channel, which means we do need to make sure that those are valid in order to deal with the attempt. More specifically, we need to make sure that:
- the invitation is not cancelled
- the invitation is not completed
- the greeter is allowed (not revoked)

The corresponding pseudo-code is this:
```python
invitation = organization.get_invitation(invitation_token)
assert invitation.is_active()
channel = invitation.get_channel(greeter_id)
assert channel.is_active()
attempt = channel.get_attempt(attempt_id)
assert attempt.is_active()
```

I don't see room for inconsistency: the attempt is simply unique in the context of a specific channel. The fact that it is identified by an UUID and then unique globally is merely an implementation detail. It's actually already the same for invitation token and organization ID: invitation token are expected to be unique in a given organization, although they are usually unique globally.

In my opinion, there is nothing to gain by having the attempt ID global to an organization: it serves no purpose other than removing the invitation token from the commands, although the server implementation would still have to get the corresponding invitation in order to perform the proper checks.

@mmmarcos @FirelightFlagboy @AureliaDolo Do you have an opinion about it?

docs/rfcs/1011-non-blocking-invite-transport.md
```diff
@@ -0,0 +1,1238 @@
...
+            // },
+            // {
+            //     // The invitation has already been completed
+            //     "status": "invitation_completed"
+            // },
+            // {
+            //     // The invitation has been cancelled
+            //     "status": "invitation_cancelled"
+            // },
+            {
```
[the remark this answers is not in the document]
vxgmichel: I think it's always useful to differentiate the cases and let the client logic choose whether to handle those status differently or not. This approach reduces the coupling between the server and the client in general. Plus, factorizing this kind of logic makes it harder to add more information in the future. For instance consider the following status:
- `greeter_not_found`
- `greeter_revoked`

If factorized in a single status, then adding information like the timestamp of the greeter revocation makes it clumsy with something like:
```
// Only present if the greeter has been revoked
greeter_revoked_timestamp: Option<DateTime>
```

Adding status in server and handling them in the client is very easy, so the rule of thumb should one status for each possible error IMO.

docs/rfcs/1011-non-blocking-invite-transport.md
vxgmichel: As explained in the earlier comment, there is no weird edge cases to handle if implemented correctly: an attempt id is only looked up in the context of the corresponding invitation channel.

docs/rfcs/1011-non-blocking-invite-transport.md
vxgmichel: As explained in the comment before, I don't think merging status simplify the code, quite the contrary.

Although in that case, an argument could be made that this specific case happens only in the case of a bug and should be handled as such. However, this is not exactly an internal bug to the server (the attempt id could be shared through some other mean) which is why I made it a separate status. I don't have a strong opinion about dealing with it as an internal bug though.

Also I don't really see how the comment about probing attack is relevant here, in this specific context.

docs/rfcs/1011-non-blocking-invite-transport.md
touilleMan: > However the attempt is always part of a specific invitation and a specific channel, which means we do need to make sure that those are valid in order to deal with the attempt. More specifically, we need to make sure that:
>
>    the invitation is not cancelled
>    the invitation is not completed
>    the greeter is allowed (not revoked)

I agree

> The corresponding pseudo-code is this:

I think it's more practical to use pseud-SQL to illustrate here.

With UUID for attempt ID and channel ID:
```sql
SELECT
    invitation.is_active as invitation_is_active,
    channel.is_active as channel_is_active,
    attempt.is_active as attempt_is_active
FROM
    attempt
    LEFT JOIN channel ON attempt.channel = channel._id
    LEFT JOIN invitation ON channel.invitation = invitation._id
WHERE attempt.attempt_id = $?;
```

VS with counters for attempt ID and channel ID:
```sql
SELECT
    invitation.is_active as invitation_is_active,
    channel.is_active as channel_is_active,
    attempt.is_active as attempt_is_active
FROM
    attempt
    LEFT JOIN channel ON attempt.channel = channel._id
    LEFT JOIN invitation ON channel.invitation = invitation._id
WHERE
    invitation.invitation_id = $? AND
    channel.channel_id = $? AND
    attempt.attempt_id = $?
```

This is just needlessly more complicated.

On top of that we have to consider the creation of a new channel & attempt: with UUID it's trivial, but with counter-based IDs we need to first request the database to get the next value or to compute the value in the request then use a returning clause to get back the value in our code.

Finally while having a counter for the attempt makes sens given it represent the attempts a client has done, the rational is much weaker for the channel (where the counter represent the order in which greeters has accepted to do the invitation).

Of course we could avoid using counter-based ID and rely on UUID-but-not-globally-unique (or using the internal ID for the row in SQL) for attempt ID & channel ID, but at this point we are just pretending the ID is not globally unique ;-)

> I don't see room for inconsistency

You are correct, I was assuming channel ID & attempt ID where UUID AND where provided (along with invitation ID)  in the request, in which case only attempt ID is used for the query, then invitation ID & channel ID must be checked against the query output (hence the possible inconsistency).

> The fact that it is identified by a UUID and that it's unique globally is merely an implementation detail
>
> In my opinion, there is nothing to gain by having the attempt ID global to an organization
>
> On the contrary, keeping proper scoping for the different IDs makes it explicit that the attempt command only makes sense in the context of a valid invitation and channel

I think those three parts summary well the situation ^^
I am in favor of using UUID for pragmatic reasons:
- it simplifies the implementation (sql queries + errors handling)
- the only thing we sacrifice is being able to order channels/attempt but we don't need it (I would even say adding order brings implicit rules to the database, for instance doing `SELECT .. FROM attempt WHERE is_active = TRUE` is implicitly expected to return the attempt with the higher counter ID)

In the end the question is how much do we value "different IDs makes it explicit that the attempt command only makes sense in the context of a valid invitation and channel" ? If you think it is of high important then go with counter, if not got with UUID (and if you need outside opinion:  I'm strongly in favor of UUID, especially given it's cheep to add a comment next in the invitation commands to explicit the relashionship between attempt/channel/invitation).

docs/rfcs/1011-non-blocking-invite-transport.md
touilleMan: > Also I don't really see how the comment about probing attack is relevant here, in this specific context.

If the attempt ID is a UUID, then this status allows to probe for valid attempt ID. I mentioned it because it is often considered as a dangerous design, but in our case I don't thing it's relevant (so I only see it as "it makes me uncomfortable, even if it should be safe" ^^)

If the attempt ID is not a UUID, then the error means the peer has joined a new attempt, and we use the new attempt ID before having ourselves joined. In this case no probing is possible whatsoever ;-)

> an argument could be made that this specific case happens only in the case of a bug and should be handled as such. However, this is not exactly an internal bug to the server

This situation is triggered by client-provided data, hence it's not an internal bug from the server point of view.

> which is why I made it a separate status

My point for merging `attempt_not_found` and `attempt_not_joined` is that they follow the same pattern: at this point the client is supposed to know the attempt he can use.
For instance if we consider attempt ID is a counter, and in our special example the current valid attempt ID is 10:
- if the client picks attempt ID < 10 he will get a `attempt_already_cancelled` (that's fine given the client knows about those IDs)
- if the client picks attempt ID == 10 he will get a `attempt_not_joined`
- if the client picks attempt ID > 10 he will get a `attempt_not_found`
I don't see the point on providing a dedicated "congratulation, you picked the right number !" error instead of just considering the attempt the client hasn't joined are not visible

docs/rfcs/1011-non-blocking-invite-transport.md
touilleMan: If the attempt ID is not a unique identifier you are correct ;-)

docs/rfcs/1011-non-blocking-invite-transport.md
FirelightFlagboy: **Pavé Cesar !**

If I understand the discussion correctly, I would resume it to the following question:

> Should we use an UUID or a counter for `attempt_id` ?

## My observations

### Either UUID or counter

- The clients/server would still need to check the validity of attempt/channel/invitation.

### With UUID

- Could be used to identify the couple `attempt/channel/invitation` with it's sole value.
- Prevent to sort the attempts.

### With counter

- Require additional information to be provided to identify the couple `attempt/channel/invitation`.
- Could sort the attempts.

## My Opinions

### Regarding the information to be provided to identify the couple `attempt/channel/invitation`

Requiring to pass `attempt_id`, `channel_id` and `invitation_id`: 

I assume would increase the bookkeeping of those value during the invitation process.
But that information that we already have on hand and would ensure we are referencing the correct invitation process.

Identity the couple with only `attempt_id`:

That only possible with an ID with a high entropy (like UUID).

### Validating the couple

Either way, we would have to use the all the ids either by using the ones provided or by fetching them in the database.

From the 2 examples provided by @touilleMan, I'm fine with both.

## Conclusion

If I want simplicity, I would choose the UUID for the `attempt_id` given it would identify the couple `attempt/channel/invitation` with a single value.

If I want to be stricter, I would require use additional ids alongside `attempt_id`, like:

- `attempt_id + channel_id`
- `attempt_id + invitation_id`
- `attempt_id + channel_id + invitation_id`

> [!NOTE]
> With the stricter way, using an UUID or a counter is an implementation details, 
> given we would define the invitation with an additional id that is an UUID.

> I would be in a more peace of mind state with the stricter approach.

## Side dish: What about ULID over UUID ?

It as the advantage of both a counter and an UUID:

- It's unique (but it's partialy generated with a timestamp) with an entropy of 128bit.
- Is sortable.
- Is compatible with UUID.

<https://github.com/ulid/spec>

docs/rfcs/1011-non-blocking-invite-transport.md
```diff
@@ -0,0 +1,1238 @@
...
+
+# Non-blocking invite transport
+
+## Overview
+
+This RFC describes the new invite transport model, adapted to work with non-blocking HTTP routes and providing a better semantic for handling attempts at building a secure channel between peers.
+
+## Background & Motivation
+
+The invite system is based on the creation of a secure channel between two peers, a claimer and a greeter. This is done using the Short Authentication String algorithm and requires a couple of messages to exchange by the peers, along with some acknowledgments.
```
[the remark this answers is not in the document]
vxgmichel: d36e8a3440bf5a5cb645e168de63108650bc36bf

docs/rfcs/1011-non-blocking-invite-transport.md
vxgmichel: d36e8a3440bf5a5cb645e168de63108650bc36bf

docs/rfcs/1011-non-blocking-invite-transport.md
vxgmichel: d36e8a3440bf5a5cb645e168de63108650bc36bf

docs/rfcs/1011-non-blocking-invite-transport.md
vxgmichel: d36e8a3440bf5a5cb645e168de63108650bc36bf

docs/rfcs/1011-non-blocking-invite-transport.md
vxgmichel: d36e8a3440bf5a5cb645e168de63108650bc36bf

docs/rfcs/1011-non-blocking-invite-transport.md
```diff
@@ -0,0 +1,1238 @@
...
+- the attempt exists
+- the attempt is active (i.e. joined and not cancelled)
+
+Then, the active attempt is cancelled with the provided reason and a new active attempt is created. The cancellation reason covers several cases:
+- the user manually cancelled the attempt
+- the hashed nonce didn't match the provided nonce
+- the SAS code communicated to the user was invalid
+- the payload could not be deciphered
+- the payload could not be deserialized
+- the payload contained invalid information
```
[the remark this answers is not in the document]
vxgmichel: b99a01bc93149b33f694e8a28713d7fb1fcdf5e4

docs/rfcs/1011-non-blocking-invite-transport.md
mmmarcos: I don't see an actual "debate" regarding UUIDs here:

- Identifying an attempt with a UUID seems like the right implementation choice. The RFC does not state the contrary since it does not make any assumptions on that (except maybe for the choice of fields in the requests, see my answer below).
- There is maybe a misunderstanding regarding phrase like ("An attempt is identified by: ..."). Those will definitely be required fields (foreign keys) and may form and alternate key in the attempt schema. However, the schema will most probably be implemented with a UUID surrogate key (`attempt_id`) as this is a common practice with many benefits over natural/composite keys.

On the other hand, when discussing what is used in the requests (see https://github.com/Scille/parsec-cloud/pull/7662/files#r1680380598), we are not talking about an "implementation detail" anymore, but an API design choice. 
IMO, a request parameter should be mandatory if it is a _sine qua non_ condition to fulfill the request (and optional if it allows to tune the response). I don't think this is the case for the invitation `token` parameter if the goal is to retrieve a specific `attempt` so it feels weird to having it in the request (as one would not put `name` or `email` together with a `user_id` to retrieve a specific user). So unless the invitation `token` serves another purpose, it seems redundant to me and this would be the main reason to avoid it (regardless if this makes one response status more or less to manage, which I think is not the essential to the discussion).

In any case, having retrieved the right attempt does not avoid the need to perform the validation steps described by @vxgmichel in this RFC.

docs/rfcs/1011-non-blocking-invite-transport.md
vxgmichel: ccf6d568d950206e8b632bc2469462e7910f790c

docs/rfcs/1011-non-blocking-invite-transport.md
vxgmichel: ccf6d568d950206e8b632bc2469462e7910f790c

docs/rfcs/1011-non-blocking-invite-transport.md
vxgmichel: ccf6d568d950206e8b632bc2469462e7910f790c

docs/rfcs/1011-non-blocking-invite-transport.md
vxgmichel: ccf6d568d950206e8b632bc2469462e7910f790c

docs/rfcs/1011-non-blocking-invite-transport.md
vxgmichel: ccf6d568d950206e8b632bc2469462e7910f790c

docs/rfcs/1011-non-blocking-invite-transport.md
vxgmichel: ccf6d568d950206e8b632bc2469462e7910f790c

docs/rfcs/1011-non-blocking-invite-transport.md
vxgmichel: ccf6d568d950206e8b632bc2469462e7910f790c

docs/rfcs/1011-non-blocking-invite-transport.md
vxgmichel: ccf6d568d950206e8b632bc2469462e7910f790c

docs/rfcs/1011-non-blocking-invite-transport.md
vxgmichel: ccf6d568d950206e8b632bc2469462e7910f790c

docs/rfcs/1011-non-blocking-invite-transport.md
vxgmichel: ccf6d568d950206e8b632bc2469462e7910f790c

docs/rfcs/1011-non-blocking-invite-transport.md
vxgmichel: Per @mmmarcos decision we keep those errors separated

docs/rfcs/1011-non-blocking-invite-transport.md
vxgmichel: ccf6d568d950206e8b632bc2469462e7910f790c

docs/rfcs/1011-non-blocking-invite-transport.md
vxgmichel: ccf6d568d950206e8b632bc2469462e7910f790c

docs/rfcs/1011-non-blocking-invite-transport.md
vxgmichel: ccf6d568d950206e8b632bc2469462e7910f790c

docs/rfcs/1011-non-blocking-invite-transport.md
vxgmichel: Per @mmmarcos decision, we keep those errors separated

docs/rfcs/1011-non-blocking-invite-transport.md
vxgmichel: 45c5cbdd13c257e7879a45fb02b69cb8614a5995

mmmarcos (approved):

PR #7671, opened 2024-07-05
Fix payloads in libparsec types unit tests

### LINKED ISSUE #7669: Fix tests involving payload dump in libparsec_types

mmmarcos:
This PR fixes all impacted payloads in `libparsec_types` unit tests.

Only the following are still to be fixed. They involve _invalid_ payloads that are expected to fail with BadSerialization at some step ("msgpack+validation", "zstd"). But I don't know how to generate these payloads since there is no `expected` value:
- `libparsec/crates/platform_device_loader/tests/list.rs::list_devices`
- `libparsec/crates/types/tests/unit/manifest.rs::dump_load`
- `libparsec/crates/types/tests/unit/manifest.rs::invalid_load`
- `libparsec/crates/types/tests/unit/manifest.rs::serde_file_manifest_invalid_blocksize`

Closes #7669

### DIFF
--- a/libparsec/crates/types/tests/unit/pki.rs
+++ b/libparsec/crates/types/tests/unit/pki.rs
@@ -8,24 +8,24 @@ use crate::fixtures::{alice, Device};
 use crate::prelude::*;
 
 #[test]
-#[ignore = "TODO: scheme has changed, must regenerate the dump"]
 fn serde_pki_enrollment_answer_payload() {
-    // Generated from Python implementation (Parsec v2.13.0+dev)
+    // Generated from Parsec v3.0.0-b.11+dev
     // Content:
     //   type: "pki_enrollment_answer_payload"
-    //   device_id: "alice@dev1"
+    //   user_id: ext(2, a11cec00100000000000000000000000)
+    //   device_id: ext(2, de10a11cec0010000000000000000000)
     //   device_label: "My dev1 machine"
     //   human_handle: ["alice@example.com", "Alicey McAliceFace"]
     //   profile: "ADMIN"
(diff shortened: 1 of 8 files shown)


libparsec/crates/types/tests/unit/pki.rs
```diff
@@ -53,31 +53,31 @@ fn serde_pki_enrollment_answer_payload() {
...
         verify_key: VerifyKey::try_from(hex!("845415cd821748005054dba8d456ac18ad3e71acdf980e19d6a925191362c9f9")).unwrap()
     }
 )]
-#[ignore = "TODO: scheme has changed, must regenerate the dump"]
 fn serde_pki_enrollment_submit_payload(
     #[case] raw: &[u8],
     #[case] expected: PkiEnrollmentSubmitPayload,
 ) {
+    println!("***expected: {:?}", expected.dump());
+
```
AureliaDolo, line 80: there is a print here

AureliaDolo (approved):
LGTM :+1:

AureliaDolo:
For the invalid payload, I think you can take a valid output and truncate it ?

vxgmichel (approved):
LGTM :+1:

mmmarcos:
> For the invalid payload, I think you can take a valid output and truncate it ?

I guess that might help with "zstd" step since it will fail in the decompression. 

For the "msgpack+validation" step, a payload needs to be constructed in a way that it can be correctly decompressed, but fail during deserialization. I don't fully understand the code, so I am not sure how to do this, I will create an issue.

libparsec/crates/types/tests/unit/pki.rs
mmmarcos: Done

PR #7698, opened 2024-07-10
Rename chunk to chunk view when applicable

vxgmichel:
And add an explanation in the docstrings:
```rust
/*
 * ChunkView
 */

/// Represents the content of a file manifest over a specific range of addresses.
///
/// It is called a `ChunkView` because it is implemented as a view over an
/// underlying chunk of data. Example:
///
/// File addressing:            raw offset            raw offset + raw size
/// Underlying chunk data: |--------|abcdefghijklmnopqrstuvwxyz|-------------|
/// Specific chunk view:   |--------------|ghijklmnopqrstu|------------------|
/// File addressing:                    start           stop
///

```

### DIFF
--- a/libparsec/crates/testbed/src/template/crc_hash.rs
+++ b/libparsec/crates/testbed/src/template/crc_hash.rs
@@ -466,11 +466,11 @@ impl CrcHash for BlockAccess {
         digest.crc_hash(hasher);
     }
 }
-impl CrcHash for Chunk {
+impl CrcHash for ChunkView {
     fn crc_hash(&self, hasher: &mut crc32fast::Hasher) {
         hasher.update(b"Chunk");
 
-        let Chunk {
+        let ChunkView {
             id,
             start,
             stop,
--- a/libparsec/crates/testbed/src/template/crc_hash.rs
+++ b/libparsec/crates/testbed/src/template/crc_hash.rs
@@ -468,8 +468,7 @@ impl CrcHash for BlockAccess {
 }
 impl CrcHash for ChunkView {
     fn crc_hash(&self, hasher: &mut crc32fast::Hasher) {
-        // TODO: change to `ChunkView``
-        hasher.update(b"Chunk");
+        hasher.update(b"ChunkView");
 
         let ChunkView {
             id,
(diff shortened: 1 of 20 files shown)


libparsec/crates/testbed/src/template/crc_hash.rs
```diff
@@ -466,11 +466,12 @@ impl CrcHash for BlockAccess {
         digest.crc_hash(hasher);
     }
 }
-impl CrcHash for Chunk {
+impl CrcHash for ChunkView {
     fn crc_hash(&self, hasher: &mut crc32fast::Hasher) {
+        // TODO: change to `ChunkView``
         hasher.update(b"Chunk");
```
FirelightFlagboy, line 472: Do it

libparsec/crates/types/tests/unit/local_manifest.rs
FirelightFlagboy, line 1: It is expected ?

libparsec/crates/testbed/src/template/crc_hash.rs
vxgmichel: That would break the testbed server tho, what do you think?

libparsec/crates/types/tests/unit/local_manifest.rs
vxgmichel: definitely not, I have no idea what happened oO
good catch :+1:

libparsec/crates/testbed/src/template/crc_hash.rs
FirelightFlagboy: Not a problem, we could rebuild it

libparsec/crates/types/tests/unit/local_manifest.rs
vxgmichel: fixed

libparsec/crates/testbed/src/template/crc_hash.rs
vxgmichel: [1a7577b](https://github.com/Scille/parsec-cloud/pull/7698/commits/1a7577b6c81c43428ebdaa3dd74675c832eb2073)

FirelightFlagboy (approved):

Issue #7642, opened 2024-06-26
App signing process in macOS for V3


mmmarcos:
We need to document the method currently used for V2 releases and make sure it still work for V3 to distribute Parsec.

According to [App code signing process in macOS](https://support.apple.com/guide/security/app-code-signing-process-sec3ad8e6e53/web):

> In macOS 10.15, all apps distributed outside the App Store must be signed by the developer using an **Apple-issued Developer ID certificate** (combined with a private key) and notarized by Apple to run under the default Gatekeeper settings. Apps developed in-house should also be signed with an Apple-issued Developer ID so that users can validate their integrity.

More info:
- [Distribute outside the Mac App Store (macOS)](https://help.apple.com/xcode/mac/current/#/dev033e997ca)
- [Sign a Mac Installer Package with a Developer ID certificate](https://help.apple.com/xcode/mac/current/#/deve51ce7c3d)

Ironicbay:
Process used so far for v2: https://github.com/Scille/parsec-cloud/tree/dev/v2/packaging/macOS#readme

Ironicbay:
Notarization fails with the following log accessible:

```{
  "logFormatVersion": 1,
  "jobId": "977bb38d-a8c5-435d-b3f7-336586b8cd5d",
  "status": "Invalid",
  "statusSummary": "Archive contains critical validation errors",
  "statusCode": 4000,
  "archiveFilename": "parsec.zip",
  "uploadDate": "2024-07-11T13:57:35.848Z",
  "sha256": "a7f28642df87b729919f1625c1c2d9ab04879fa1b390d046a9840a6189817d0f",
  "ticketContents": null,
  "issues": [
    {
      "severity": "error",
      "code": null,
      "path": "parsec.zip/parsec.app/Contents/Resources/app.asar.unpacked/node_modules/fsevents/fsevents.node",
      "message": "The binary is not signed.",
      "docUrl": "https://developer.apple.com/documentation/security/notarizing_macos_software_before_distribution/resolving_common_notarization_issues#3087721",
      "architecture": "x86_64"
    },
    {
      "severity": "error",
      "code": null,
      "path": "parsec.zip/parsec.app/Contents/Resources/app.asar.unpacked/node_modules/fsevents/fsevents.node",
      "message": "The signature does not include a secure timestamp.",
      "docUrl": "https://developer.apple.com/documentation/security/notarizing_macos_software_before_distribution/resolving_common_notarization_issues#3087733",
[...]

PR #7723, opened 2024-07-15
Fix MacOS app signing & notarization

Ironicbay:
Closes https://github.com/Scille/parsec-cloud/issues/7642

- [x] Enable hardened runtime for code signing to enable manual notarization (same state as v2)
- [x] Enable automatic notarization in electron-builder

### DIFF
--- a/client/electron/macOS/entitlements.plist
+++ b/client/electron/macOS/entitlements.plist
@@ -1,0 +1,10 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!DOCTYPE plist PUBLIC "-//Apple//DTD PLIST 1.0//EN" "http://www.apple.com/DTDs/PropertyList-1.0.dtd">
+<plist version="1.0">
+  <dict>
+    <key>com.apple.security.cs.allow-unsigned-executable-memory</key>
+    <true/>
+    <key>com.apple.security.cs.allow-jit</key>
+    <true/>
+  </dict>
+</plist>
--- a/client/electron/package.js
+++ b/client/electron/package.js
@@ -138,9 +138,14 @@ const options = {
     guid: '2f56a772-db54-4a32-b264-28c42970f684',
   },
 
+  afterSign: OPTS.sign === false || OPTS.platform !== 'darwin' ? undefined : 'electron-builder-notarize',
+
   mac: {
     target: 'dmg',
     category: 'public.app-category.productivity',
+    hardenedRuntime: true,
+    entitlements: './macOS/entitlements.plist',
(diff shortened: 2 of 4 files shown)


client/electron/macOS/entitlements.plist
```diff
@@ -0,0 +1,13 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!DOCTYPE plist PUBLIC "-//Apple//DTD PLIST 1.0//EN" "http://www.apple.com/DTDs/PropertyList-1.0.dtd">
+<plist version="1.0">
+<dict>
+	<!-- These are required for binaries built by PyInstaller -->
```
FirelightFlagboy, line 5: Outdated comment

client/electron/package.js
```diff
@@ -138,9 +138,14 @@ const options = {
     guid: '2f56a772-db54-4a32-b264-28c42970f684',
   },
 
+  afterSign: 'electron-builder-notarize',
```
FirelightFlagboy, line 141: Should be set to `undefined` if `OPTS.sign` is false (to allow to build an app that is not signed)

client/package.json
```diff
@@ -49,6 +49,7 @@
         "@ionic/vue-router": "^7.6.4",
         "@types/luxon": "^3.3.4",
         "axios": "^1.7.2",
+        "electron-builder-notarize": "^1.5.2",
```
Max-7, line 52: Should be a dev dependency. And I see `@types/luxon` in this list, it should also be a dev dependency.

Max-7 (changes requested):

client/electron/macOS/entitlements.plist
```diff
@@ -0,0 +1,13 @@
...
+	<key>com.apple.security.cs.allow-jit</key>
+	<true/>
+	<key>com.apple.security.cs.allow-unsigned-executable-memory</key>
+	<true/>
+	<key>com.apple.security.cs.disable-library-validation</key>
+	<true/>
```
FirelightFlagboy, line 11: Could we remove that setting ?

The tutorial under: https://kilianvalkhof.com/2019/electron/notarizing-your-electron-application/

Don't use it

FirelightFlagboy (changes requested):

client/electron/package.js
FirelightFlagboy, line 141: Should only be set if on darwin platform

FirelightFlagboy (approved):

Max-7 (approved):

PR #7725, opened 2024-07-15
Workspace keys rotation

### LINKED ISSUE #6526: Implement workspace rotate key
https://github.com/Scille/parsec-cloud/blob/d1862e5c3296d0e65db6dda9bde3b8dc96e44b24/libparsec/crates/client/src/client/workspace_rotate_key.rs#L27-L32

# Key rotation

The key rotation system should be implemented in three different layers:
1 - CertificateOps
2 - Client
3 - Monitor

## 1 - CertificateOps

This is the low level layer, its goal is to expose specific commands involving certificates (e.g. `CertificateOps::share_realm`, `CertificateOps::poll_server_for_new_certificates`).

In our case, we want:

- `CertificateOps::rotate_workspace_key`
- `CertificateOps::get_workspaces_needs`

## 1.1 - `CertificateOps::rotate_workspace_key`

This method should:

- Create a new `RealmRoleCertificate`
- Upload the certificate in the server
- If the server rejects it with a `bad_key_index` status, this means we are not aware
  of a concurrent key rotation. In this case we should return a `CertificateBasedActionOutcome::RemoteIdempotent`
  outcome.

In theory, the code should be very similar to what is done in `/libparsec/crates/client/src/certif/user_revoke.rs`.

## 1.2 - `CertificateOps::get_workspaces_needs`

The workspace has two type of needs (i.e. operations that should occur without user interaction):

- Unshare workspace with a newly revoked user
- Do key rotation whenever a unshare occurs

This method should search in the certificate store for all the workspaces our user is OWNER.
Then for each workspace, get the last key rotation date and look for any unshare or any user
that is member of the workspace and has been revoked since then.

This method should return a Vec of structure analogue to:

```rust
pub enum WorkspaceNeeds{
    Nothing,
    KeyRotationOnly,
    /// A user that is part of the workspace has been revoked, this creates a cascade
    /// of needs: first the worskpace needs to be unshared, which is turn means a key
    /// rotation is needed.
    UnshareThenKeyRotation {
        revoked_users: Vec<UserID>
    }
}

CertificateOps::get_workspaces_needs(...) -> Vec<(VlobID, WorkspaceNeeds)> {
  ...
}
```

*Notes:*

- Certificate store relies on the certificate storage (which is defined in
  `libparsec_platform_storage` and handle the storage in the SQlite database).
- Certificate storage query system is poorly fitted for the kind of query we need
  (i.e. "get the unshare or user revoke certificates involving the workspace XXX
  since the timestamp of it last key rotation certificate"). So for the moment
  we should just fetch 1) last key rotation, 2) all real roles and 3) all user
  revocation certificates, then from there do the logic manually in code.

## 2 - Client

This is the high level layer, typically client methods gets exposed in the
libparsec bindings (i.e. they are accessible from the GUI).

In our case, we want `Client::process_workspaces_needs`.

Note a `Client::rotate_workspace_key` stub already exists, it should be removed since
`Client::process_workspaces_needs` replaces it.

### 2 - `Client::process_workspaces_needs`

This method should:

- retrieve the need for key rotation in all the workspaces the user OWNER, this
  should be done with the `CertificateOps::get_workspaces_needs` method
- do unshare of revoked users if any
- do key rotation
- On success, refresh the certif ops certificates by calling `CertificateOps::poll_server_for_new_certificates`.

In theory, the code should be very similar to what is done in `/libparsec/crates/client/src/client/workspace_rename.rs`.

*Notes:*

- We don't put this method in `WorkspaceOps`: this is because workspace ops object
  represents a started workspace, but we want to be able to do the key rotation without
  starting it (given it's up to the user to choose which workspace he wants to start,
  but key rotation should occur automatically when a user is no longer shared). And,
  on top of that, we want to be able to handle all workspaces at once.

- There is a footgun to be aware of in the implementation: we should only use the
  certificate ops to get informations about what workspace we are part of.

  Indeed, it could be tempting to use the `Client::list_workspaces` however this method relies on the lecal
  user manifest that itself is updated whenever a new certificate arrives (in a very similar
  fashion as what we are doing here).

  So issue is we would have a new workspace event triggering two monitors (the one refreshing
  the local user manifest and the one doing key rotation), while one of them depend on the output of the other...

## 3 - Monitor

A simple way to describe what is a monitor is to imagine a stuboorn user that clicks
refresh on the GUI every time he sees a new notification.

So the monitor:

- Listen to specific events
- Call workspace ops methods in response
- Is expected to be very simple, given all the heavy lifting is done in workspace ops's
  methods (as it simplify testing)

In our case, we want a monitor `/libparec/crates/client/src/monitors/workspaces_needs.rs` that:

- Listen to `EventMissedServerEvents` and `EventNewCertificates`
- If any of the event is triggered, the monitor calls the `Client::process_workspaces_needs` method.

In theory, the code should be similar to `/libparsec/crates/client/src/monitors/workspaces_refresh_list.rs`.

*Notes:*

- The `EventNewCertificates` is very broad, in the future we may want to split
  it so that getting a new certificate about something totally unrelated (e.g. a new
  shamir recovery has been configured) doesn't trigger the check for key rotation need.

# Related issues 

- [ ] #6322
- [ ] #6527

touilleMan:
closes #6526


libparsec/crates/client/src/certif/realms_needs.rs
```diff
@@ -0,0 +1,115 @@
...
+
+                        match &last_key_rotation {
+                            Some(last_key_rotation)
+                                if last_key_rotation.timestamp < role.timestamp =>
+                            {
+                                needs = RealmNeeds::KeyRotationOnly { current_key_index };
+                            }
+                            None => {
+                                needs = RealmNeeds::KeyRotationOnly { current_key_index };
+                            }
```
AureliaDolo, line 81: nit: You could maybe merge these two branches

libparsec/crates/client/src/certif/store.rs
```diff
@@ -33,6 +33,42 @@ use crate::certif::CertifPollServerError;
...
+    /// (i.e. we haven't send the initial realm role certificate).
+    LocalOnly,
+    /// Step 1
+    ///
+    /// The realm exists on the server (i.e. initial realm role certificate has been uploaded).
+    /// Notes:
+    /// - The user realm doesn't go beyond this state.
+    /// - For a workspace realm, this is the first step of the workspace bootstrap.
+    CreatedInServer,
+    /// Step 2
```
AureliaDolo, line 58:
```suggestion
    /// Step 2: for workspace only
```

AureliaDolo: And maybe we could add that to next step's doc for consistency, but the name WorkspaceBootstapped is clear enough

libparsec/crates/client/src/certif/realms_needs.rs
```diff
@@ -0,0 +1,115 @@
...
+    /// A user that is part of the workspace has been revoked, this creates a cascade
+    /// of needs: first the workspace needs to be unshared, which is turn means a key
+    /// rotation is needed.
+    UnshareThenKeyRotation {
+        current_key_index: Option<IndexInt>,
+        revoked_users: Vec<UserID>,
+    },
+}
+
+/// Note realm existence is not checked (and `RealmNeeds::Nothing` is returned if that's not the case)
```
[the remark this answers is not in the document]
touilleMan: No, given this function is only used in `Client::process_workspaces_needs` where we first get the list of our workspaces.

However from this function point of view there is no guarantee the provided realm_id is a valid one, and if that's not the case returning `RealmNeeds::Nothing` might be a bit surprising (given the case was unlikely to occur in reality, I didn't want to make the codebase more complex here), hence the comment.

libparsec/crates/client/src/certif/realms_needs.rs
touilleMan: I can't because there is the if clause that is only for the first branch
```rust
enum X {
  A(u32),
  B(u32),
  C
}

match X::A(1) {
  X::A(x) | X::B(x) if x == 1 => (),  // This is allowed
  X::C | X::B(x) if x == 1 => (), // C doesn't bind x, this cannot compile
   _ => (),
}
```

See [this playground](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=8be3edeae62afef17af63388d6fa8b5e)

libparsec/crates/client/src/certif/store.rs
touilleMan: :+1: done.

AureliaDolo (approved):
LGTM

PR #7730, opened 2024-07-16
 [Confinement point] Apply merge folder logic from v2

### LINKED ISSUE #7705: Update sync monitor to manage the syncing for no-longer-confined entries
- Plenty of TODOs in `merge.rs`

  https://github.com/Scille/parsec-cloud/blob/fc76ec52b7ac92370d3ed24cce3cbd14c50b5e9b/libparsec/crates/client/src/workspace/merge.rs#L212-L215

- But the manifest methods are implemented already:
  
  https://github.com/Scille/parsec-cloud/blob/fc76ec52b7ac92370d3ed24cce3cbd14c50b5e9b/libparsec/crates/types/src/local_manifest.rs#L799-L810

FirelightFlagboy:
Closes #7705

Depends #7729

### DIFF
--- a/libparsec/crates/client/src/workspace/merge.rs
+++ b/libparsec/crates/client/src/workspace/merge.rs
@@ -147,29 +147,65 @@ pub(super) fn merge_local_file_manifest(
 pub(super) fn merge_local_folder_manifest(
     local_author: DeviceID,
     timestamp: DateTime,
+    prevent_sync_pattern: Option<&Regex>,
     local: &LocalFolderManifest,
     remote: FolderManifest,
 ) -> MergeLocalFolderManifestOutcome {
-    // 0) Sanity checks, caller is responsible to handle them properly !
-    debug_assert_eq!(local.base.id, remote.id);
+    // Destruct local and remote manifests to ensure this code with fail to compile whenever a new field is introduced.
+    let LocalFolderManifest {
+        base:
+            FolderManifest {
+                id: local_base_id,
+                version: local_base_version,
+                children: local_base_children,
+                parent: local_base_parent,
+                // Ignored, we don't merge data that change on each sync
+                author: _,
+                // `created` should never change, so in theory we should have
+                // `local.base.created == remote.base.created`, but there is no strict
+                // guarantee (e.g. remote manifest may have been uploaded by a buggy client)
+                // so we have no choice but to accept whatever value remote provides.
+                created: _,
+                // Ignored, we don't merge data that change on each sync
+                timestamp: _,
+                // Ignored, we don't merge data that change on each sync
+                updated: _,
+            },
+        children: local_children,
+        parent: local_parent,
+        need_sync: local_need_sync,
+        speculative: local_speculative,
+        // Ignored, that field is merged in `from_remote_with_local_context`
+        remote_confinement_points: _,
+        // Ignored, that field is merged in `from_remote_with_local_context`
+        local_confinement_points: _,
+        // Ignored, we don't merge data that change on each sync
+        updated: _,
+    } = local;
 
-    // TODO
-    // Start by re-applying pattern (idempotent)
-    // if force_apply_pattern and isinstance(
-    //     local_manifest, (LocalFolderManifest, LocalWorkspaceManifest)
-    // ):
+    // 0) Sanity checks, caller is responsible to handle them properly !
+    debug_assert_eq!(local_base_id, &remote.id);
+
+    // TODO: Allow to force re-applying the prevent sync pattern (idempotent)
+    // if force_apply_pattern {
     //     local_manifest = local_manifest.apply_prevent_sync_pattern(prevent_sync_pattern, timestamp)
+    // }
 
     // 1) Shortcut in case the remote is outdated
-    if remote.version <= local.base.version {
+    if remote.version <= *local_base_version {
         return MergeLocalFolderManifestOutcome::NoChange;
     }
 
+    let local_from_remote = LocalFolderManifest::from_remote_with_local_context(
+        remote,
+        prevent_sync_pattern,
+        local,
+        timestamp,
+    );
+
     // 2) Shortcut in case only the remote has changed
-    if !local.need_sync {
-        return MergeLocalFolderManifestOutcome::Merged(Arc::new(
-            LocalFolderManifest::from_remote(remote, None),
-        ));
+    if !local_need_sync {
+        return MergeLocalFolderManifestOutcome::Merged(Arc::new(local_from_remote));
     }
 
     // Both the remote and the local have changed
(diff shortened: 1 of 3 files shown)


libparsec/crates/client/src/workspace/merge.rs
```diff
@@ -263,27 +236,16 @@ pub(super) fn merge_local_folder_manifest(
...
-        // TODO: confinement points support not implemented yet !
-        local_confinement_points: HashSet::new(),
-        // TODO: confinement points support not implemented yet !
-        remote_confinement_points: HashSet::new(),
-        speculative: false,
+    local_from_remote.speculative = false;
+    local_from_remote.need_sync =
+        merged_children != local_from_remote.children || merged_parent != local_from_remote.parent;
+    if local_from_remote.need_sync {
+        local_from_remote.updated = timestamp;
```
touilleMan, line 243: The original code was
```rust
let updated = if merged_need_sync {
        timestamp
    } else {
        remote.updated
    };
```
I guess this has changed the behavior, why ?
(and we are missing a test here if the behavior has changed without us noticing !)

libparsec/crates/client/src/workspace/merge.rs
```diff
@@ -263,27 +236,16 @@ pub(super) fn merge_local_folder_manifest(
...
-        parent: merged_parent,
-        children: merged_children,
-        updated,
-        base: remote,
-        // TODO: confinement points support not implemented yet !
-        local_confinement_points: HashSet::new(),
-        // TODO: confinement points support not implemented yet !
-        remote_confinement_points: HashSet::new(),
-        speculative: false,
+    local_from_remote.speculative = false;
```
touilleMan, line 239: I'd rather stick with the `merged need_sync` field: the fact we re-cycle `local_from_remote` to create `merged_manifest` instead of building a `LocalFolderManifest` from scratch is an implementation detail, but it makes the code more complex to read.

So I'd rather keep this behavior in a single place:
```rust
    let merged_need_sync = merged_children != remote.children || merged_parent != remote.parent;
    let merged_updated = if merged_need_sync {
        timestamp
    } else {
        remote.updated
    };
    let merged_base = local_from_remote.base;
    // TODO: Add a comment here to explain why we got the local confinement points from what seems to be the remote data ! 
    let merged_local_confinement_points = local_from_remote.local_confinement_points:,
    let merged_remote_confinement_points = local_from_remote.remote_confinement_points:,
    let merged_manifest = LocalFolderManifest {
       need_sync: merged_need_sync,
        parent: merged_parent,
        children: merged_children,
        updated: merged_updated,
        base: merged_base,
        local_confinement_points: merged_local_confinement_points:,
        remote_confinement_points: merged_remote_confinement_points:,
        speculative: false,
    };
```

libparsec/crates/client/src/workspace/merge.rs
```diff
@@ -165,11 +166,16 @@ pub(super) fn merge_local_folder_manifest(
         return MergeLocalFolderManifestOutcome::NoChange;
     }
 
+    let mut local_from_remote = LocalFolderManifest::from_remote_with_local_context(
```
touilleMan, line 169: This is not related to the PR, but I find this method poorly designed: we are both applying remote sync pattern (which is normal since we are dealing with remote data) AND restoring local sync pattern (which by definition reference files that are in local only...)

I'm wondering how this behave in corner cases (e.g. if a new file in the remote manifest match the local prevent sync pattern, and a file with the same name is also in local)

So my guess here is `LocalFolderManifest::from_remote_with_local_context` shouldn't exist in the first place: it is the job of `merge_local_folder_manifest` to provide a new local folder manifest when given a remote and a local.

touilleMan (changes requested):

libparsec/crates/client/src/workspace/merge.rs
FirelightFlagboy: Can you be more precise on why you think so ?

For me it's the same behavior, where we update `updated`, `children` and `parent` only if `children` or `parent` are different from the merged result and the remote.

libparsec/crates/client/src/workspace/merge.rs
```diff
@@ -196,54 +202,21 @@ pub(super) fn merge_local_folder_manifest(
...
+    if local_from_remote.base.author == local_author && !local.speculative {
         let mut new_local = local.to_owned();
-        new_local.base = remote;
+        new_local.base = local_from_remote.base;
         return MergeLocalFolderManifestOutcome::Merged(Arc::new(new_local));
     }
 
     // 4) Merge data and ensure the sync is still needed
 
-    // Destruct local and remote manifests to ensure this code with fail to compile
```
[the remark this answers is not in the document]
FirelightFlagboy: done

libparsec/crates/client/src/workspace/merge.rs
FirelightFlagboy: I've created an issue for that #7738

libparsec/crates/client/src/workspace/merge.rs
```diff
@@ -147,11 +147,34 @@ pub(super) fn merge_local_file_manifest(
 pub(super) fn merge_local_folder_manifest(
     local_author: DeviceID,
     timestamp: DateTime,
+    prevent_sync_pattern: Option<&Regex>,
     local: &LocalFolderManifest,
     remote: FolderManifest,
 ) -> MergeLocalFolderManifestOutcome {
+    // Destruct local and remote manifests to ensure this code with fail to compile
```
touilleMan, line 154: Comment is broken (it talks about remote that is no longer destructed, and the 2nd line is missing !)

libparsec/crates/client/src/workspace/merge.rs
```diff
@@ -147,11 +147,34 @@ pub(super) fn merge_local_file_manifest(
...
+    let LocalFolderManifest {
+        base:
+            FolderManifest {
+                id: base_id,
```
touilleMan, line 158: the use of `local_` prefix instead of just `base_xxx` is important since we also have a base provided by `local_from_remote.base`

libparsec/crates/client/src/workspace/merge.rs
```diff
@@ -263,27 +255,24 @@ pub(super) fn merge_local_folder_manifest(
...
-        base: remote,
-        // TODO: confinement points support not implemented yet !
-        local_confinement_points: HashSet::new(),
-        // TODO: confinement points support not implemented yet !
-        remote_confinement_points: HashSet::new(),
+        parent: merged_parent,
+        need_sync: merge_need_sync,
+        updated: merge_update,
         speculative: false,
+        ..local_from_remote
```
touilleMan, line 272: Again, sorry to insist by I'd rather have an explicit code about what goes in `merged_manifest` instead of `..local_from remote` which tends to implicitly resolve newly added field to whatever is in `local_from_remote`

touilleMan (changes requested):

libparsec/crates/client/src/workspace/merge.rs
FirelightFlagboy: For me it still deconstruct a remote manifest since I deconstruct the base manifest :sweat_smile:

libparsec/crates/client/src/workspace/merge.rs
touilleMan: > Can you be more precise on why you think so ?

the previous code:
```rust
let updated = if merged_need_sync {
        timestamp
    } else {
        remote.updated
    };
```
the new code:
```rust
    let local_from_remote = LocalFolderManifest::from_remote_with_local_context(
        remote,
        prevent_sync_pattern,
        local,
        timestamp,
    );
[...]
    if local_from_remote.need_sync {
       local_from_remote.updated = timestamp
   }
```

I guess `local_from_remote.updated` get the same value than `remote.update` when building `local_from_remote` with `LocalFolderManifest::from_remote_with_local_context`', however `timestamp` is also passed to this function...

In fact it's even worse than that when reading the function: `LocalFolderManifest::restore_local_confinement_points` uses `timestamp` to update `updated` field ONLY IF `local_confinement_points` or `remote_confinement_points` are set... which means in your current code, `merged_manifest.updated` get set with `timestamp` if it is not `need_sync` but has confinement points (while the expected behavior is to have `merged_manifest.updated` to get set with `remote.updated`)

In that regard, the first version of the code is much cleaner given you don't have to think about how `LocalFolderManifest::from_remote_with_local_context` works to understand the code, and the full reasoning on what will be the `updated` value is in a single place.

libparsec/crates/client/src/workspace/merge.rs
touilleMan: > For me it still deconstruct a remote manifest since I deconstruct the base manifest 😅

You're absolutely right ! I didn't see it :smile: 

(however there is still the second line of the comment to restore)

libparsec/crates/client/src/workspace/merge.rs
FirelightFlagboy: done

libparsec/crates/client/src/workspace/merge.rs
FirelightFlagboy: done

libparsec/crates/client/src/workspace/merge.rs
FirelightFlagboy: done

libparsec/crates/client/src/workspace/merge.rs
FirelightFlagboy: Ok

touilleMan: Sorry for the never-ending comment, but I think your current code is still affected:
```rust
 let merge_update = if merge_need_sync {
        timestamp
    } else {
        local_from_remote.updated
    };
```

As I've said, `remote.updated` is not strictly equivalent to `local_from_remote.updated`

My guess (I didn't test the code) is if `local_confinement_points` or `remote_confinement_points` are set but `need_sync == false`, then `merge_update == local_from_remote.updated == timestamp` (while the correct behavior would be `merged_update == remote.updated`

FirelightFlagboy: Given the indication of `remote.updated`, I should use `local_from_remote.base.updated` ?

FirelightFlagboy: Done

touilleMan (approved):

PR #7735, opened 2024-07-17
Rework debug dump for workspace storage

FirelightFlagboy:
Depends on #7730

### DIFF
--- a/libparsec/crates/platform_storage/src/native/workspace.rs
+++ b/libparsec/crates/platform_storage/src/native/workspace.rs
@@ -12,6 +12,8 @@ use std::path::Path;
 
 use libparsec_types::prelude::*;
 
+#[cfg(any(test, feature = "expose-test-methods"))]
+use crate::workspace::{DebugBlock, DebugChunk, DebugDump, DebugVlob};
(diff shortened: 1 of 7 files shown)


libparsec/crates/platform_storage/src/native/workspace.rs
```diff
@@ -304,10 +306,9 @@ impl PlatformWorkspaceStorage {
     }
 
     /// Only used for debugging tests
-    #[allow(unused)]
-    pub async fn debug_dump(&mut self) -> anyhow::Result<String> {
+    #[cfg(test)]
```
touilleMan, line 309: It's important to have `debug_dump` exposed to the other crates, as it is a very useful when debugging tests !

touilleMan (changes requested):

libparsec/crates/platform_storage/src/native/workspace.rs
FirelightFlagboy: I'm in phase with what you're saying, but I'm waiting for #7641 to be merged

FirelightFlagboy: done

touilleMan (approved):

PR #7739, opened 2024-07-18
Add `prevent-sync-pattern` route to platform storage

### LINKED ISSUE #7733: [Confinement point] Restore the dedicate routes in the workspace storage (that have somehow been removed from v3):
How previous workspace storage used that api:

https://github.com/Scille/parsec-cloud/blob/d397c035d1369e498a48dea26837d93b7291dfe4/oxidation/libparsec/crates/core_fs/src/storage/workspace_storage.rs#L512-L535

How it was implemented on the data storage:

https://github.com/Scille/parsec-cloud/blob/d397c035d1369e498a48dea26837d93b7291dfe4/oxidation/libparsec/crates/platform_storage/src/sqlite/data.rs#L262-L313

FirelightFlagboy:
Closes #7733

### DIFF
--- a/libparsec/crates/platform_storage/src/workspace.rs
+++ b/libparsec/crates/platform_storage/src/workspace.rs
@@ -215,6 +215,23 @@ impl WorkspaceStorage {
     pub async fn debug_dump(&mut self) -> anyhow::Result<DebugDump> {
         self.platform.debug_dump().await
     }
+
+    pub async fn set_prevent_sync_pattern(&mut self, pattern: &Regex) -> anyhow::Result<bool> {
+        self.platform.set_prevent_sync_pattern(pattern).await
+    }
+
+    pub async fn get_prevent_sync_pattern(&mut self) -> anyhow::Result<(Regex, bool)> {
+        self.platform.get_prevent_sync_pattern().await
+    }
+
+    pub async fn mark_prevent_sync_pattern_fully_applied(
+        &mut self,
+        pattern: &Regex,
+    ) -> anyhow::Result<bool> {
+        self.platform
+            .mark_prevent_sync_pattern_fully_applied(pattern)
+            .await
+    }
 }
 
 #[cfg(test)]
--- a/libparsec/crates/platform_storage/src/native/workspace.rs
+++ b/libparsec/crates/platform_storage/src/native/workspace.rs
@@ -407,6 +407,122 @@ impl PlatformWorkspaceStorage {
             blocks,
         })
     }
+
+    pub async fn set_prevent_sync_pattern(&mut self, pattern: &Regex) -> anyhow::Result<bool> {
+        let pattern = pattern.to_string();
+
+        let mut transaction = self.conn.begin().await?;
+
+        let fully_applied = db_update_prevent_sync_pattern(&mut transaction, &pattern).await?;
+
+        transaction.commit().await?;
+
+        Ok(fully_applied)
+    }
+
+    pub async fn get_prevent_sync_pattern(&mut self) -> anyhow::Result<(Regex, bool)> {
+        db_get_prevent_sync_pattern(&mut self.conn)
+            .await
+            .and_then(|(pattern, fully_applied)| {
+                Regex::from_regex_str(&pattern)
+                    .map(|re| (re, fully_applied))
+                    .map_err(anyhow::Error::from)
+            })
+    }
+
+    pub async fn mark_prevent_sync_pattern_fully_applied(
+        &mut self,
+        pattern: &Regex,
+    ) -> anyhow::Result<bool> {
+        let pattern = pattern.to_string();
+
+        let mut transaction = self.conn.begin().await?;
+
+        let fully_applied =
+            db_set_prevent_sync_pattern_as_fully_applied(&mut transaction, &pattern).await?;
+
+        transaction.commit().await?;
+
+        Ok(fully_applied)
+    }
+}
+
+async fn db_update_prevent_sync_pattern<T, E>(mut trans: T, pattern: &str) -> anyhow::Result<bool>
+where
+    T: AsMut<E>,
+    for<'c> &'c mut E: sqlx::Executor<'c, Database = sqlx::Sqlite>,
+{
+    let res = sqlx::query(concat!(
+        "UPDATE prevent_sync_pattern",
+        " SET pattern = ?1, fully_applied = false",
+        " WHERE _id = 0 AND pattern != ?1",
+    ))
+    .bind(pattern)
+    .execute(trans.as_mut())
+    .await?;
+
+    if res.rows_affected() == 1 {
+        Ok(false)
+    } else {
+        // No row was updated, the pattern was already set.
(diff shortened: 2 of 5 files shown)


FirelightFlagboy:
Depends on #7735

libparsec/crates/platform_storage/src/workspace.rs
```diff
@@ -217,6 +217,26 @@ impl WorkspaceStorage {
     }
 }
 
+// Prevent sync pattern manipulation
+impl WorkspaceStorage {
```
touilleMan, line 221: For consistency, remove the dedicated `impl WorkspaceStorage` given all the other methods are already in a single impl

libparsec/crates/platform_storage/src/native/workspace/mod.rs
```diff
@@ -3,6 +3,7 @@
 // Certificate storage relies on the upper layer to do the actual certification
 // validation work and takes care of handling concurrency issues.
 // Hence no unique violation should occur under normal circumstances here.
+mod sync_pattern;
```
touilleMan, line 6: I'd rather keep everything in the same file for the moment: having multiple impl of the same structure in different file is error prone (see what is done for `CertificatesOps`/`WorkspaceOps` etc. to split the code while having a single impl)

With your changes added, the file will be ~1000lines which in the high limit, but I think it's still okay given the code there is very simple (basically impl PlatformWorkspaceStorage exposing methods, and a bunch of `db_` function to implement the queries, with each method being very isolated from each other).

libparsec/crates/platform_storage/src/native/workspace/sync_pattern.rs
```diff
@@ -0,0 +1,126 @@
...
+async fn db_update_prevent_sync_pattern<T, E>(mut trans: T, pattern: &str) -> anyhow::Result<bool>
+where
+    T: AsMut<E>,
+    for<'c> &'c mut E: sqlx::Executor<'c, Database = sqlx::Sqlite>,
+{
+    let row = sqlx::query(concat!(
+        "UPDATE prevent_sync_pattern",
+        " SET pattern = ?1, fully_applied = false",
+        " WHERE _id = 0 AND pattern != ?1",
+        " RETURNING fully_applied",
```
touilleMan, line 57: the `RETURNING` seems useless here given if a row is returned, then the `fully_applied = false` has been set

libparsec/crates/platform_storage/tests/unit/workspace/mod.rs
```diff
@@ -3,6 +3,7 @@
 // `allow-unwrap-in-test` don't behave as expected, see:
 // https://github.com/rust-lang/rust-clippy/issues/11119
 #![allow(clippy::unwrap_used)]
+mod sync_pattern;
```
touilleMan, line 6: Please stop doing this kind of move, while convenient at development time, it breaks the logic in the code hierarchy (why should we have everything in a `mod.rs`, then a specific sup-part in it own .rs file...)

Instead what you can do:
- solution1: follow the layout logic of the code
- solution2: properly re-split the code (typically asking on slack first if you are unsure how to split it)
- solution3: communicate on slack or during meeting about your pain-point with the codebase ;-)

Right now, the test file is ~1000lines long and you are adding ~100 more lines. I think it's just simpler to add your lines in the file, considering there is no much (if any) we are planning to add to the platform storage (so it's not worth refactoring the tests at the moment).

libparsec/crates/platform_storage/src/native/workspace/sync_pattern.rs
```diff
@@ -0,0 +1,126 @@
...
+) -> anyhow::Result<bool>
+where
+    T: AsMut<E>,
+    for<'c> &'c mut E: sqlx::Executor<'c, Database = sqlx::Sqlite>,
+{
+    let row = sqlx::query(concat!(
+        "UPDATE prevent_sync_pattern",
+        " SET fully_applied = true",
+        " WHERE _id = 0 AND pattern = ?1",
+        " RETURNING fully_applied",
```
touilleMan, line 114: Same as above: `RETURNING` seems useless here

libparsec/crates/platform_storage/src/native/workspace/sync_pattern.rs
```diff
@@ -0,0 +1,126 @@
...
+    ))
+    .bind(pattern)
+    .fetch_optional(trans.as_mut())
+    .await?;
+
+    if let Some(row) = row {
+        row.try_get::<bool, _>(0).map_err(anyhow::Error::from)
+    } else {
+        // No row was updated, the pattern a different pattern is set in the database.
```
touilleMan, line 123:
```suggestion
        // No row was updated, a different pattern is set in the database.
```

libparsec/crates/platform_storage/src/native/workspace/sync_pattern.rs
```diff
@@ -0,0 +1,126 @@
...
+    let pattern = row.try_get::<String, _>(0)?;
+    let fully_applied = row.try_get::<bool, _>(1)?;
+
+    Ok((pattern, fully_applied))
+}
+
+async fn db_set_prevent_sync_pattern_as_fully_applied<T, E>(
+    mut trans: T,
+    pattern: &str,
+) -> anyhow::Result<bool>
```
touilleMan, line 105: The return value seems confusing to me:
- if the pattern is the same in database, it returns `True`
- if the pattern is not the same than in database, it  returns whatever `fully_applied` is in database

So if the caller gets a `true`, he doesn't know if *his* pattern is the one in database and fully applied.

So instead of a boolean I would introduce a dedicate error type:
```rust
pub enum SetPreventSyncPatternAsFullyAppliedError {
    PatternMismatch { in_db_pattern_fully_applied: bool },  // I'm not even sure returning the fully applied value from the mismatch is useful...
   Internal(...),
}
```

touilleMan (changes requested):

libparsec/crates/platform_storage/src/native/workspace/sync_pattern.rs
FirelightFlagboy: I keep the idea on the side (I've created the issue #7755).
I'm also a little confused about that behavior, I'll try to look for it once I see the bigger picture.

libparsec/crates/platform_storage/src/workspace.rs
FirelightFlagboy: done

libparsec/crates/platform_storage/src/native/workspace/mod.rs
FirelightFlagboy: done

libparsec/crates/platform_storage/tests/unit/workspace/mod.rs
FirelightFlagboy: done

touilleMan (approved):
:+1:

PR #7756, opened 2024-07-22
Add additional tests for merge_local_folder_manifest

FirelightFlagboy:


libparsec/crates/types/src/local_manifest.rs
```diff
@@ -821,13 +830,35 @@ impl LocalFolderManifest {
         self
     }
 
+    // TODO: FINISH DOC !!!
+
+    /// Restore in the current manifest all the entries that were confined (according to
+    /// the given `prevent_sync_pattern`) in the `other` manifest.
+    ///
+    /// Note `other` is expected to be up to date with `prevent_sync_pattern` (i.e.
+    /// `apply_prevent_sync_pattern()` should have been called on it with this pattern).
```
vxgmichel, line 839: No it's not expected to be up-to-date: it's even used in `apply_prevent_sync_pattern` to restore previously confined entries that are no longer confined. This is why a timestamp is required: new entries might get added

libparsec/crates/types/src/local_manifest.rs
```diff
@@ -821,13 +830,35 @@ impl LocalFolderManifest {
...
+    /// - `base.children`: `{"a.tmp": 1, "c.tmp": 3}`
+    ///
+    /// Then the resulting manifest would have an empty `remote_confinement points` and
+    /// `{"a.tmp": 1, "b.txt": 2}` as `children`.
     fn restore_local_confinement_points(
         self,
         other: &Self,
         prevent_sync_pattern: Option<&Regex>,
+        // TODO: why `timestamp` ? if we are only restoring confinement points then
+        //       the `updated` field shouldn't change !
```
vxgmichel, line 858: it's explained in the comment above: the `updated` field might change if the `other` manifest used a different filter.

libparsec/crates/types/src/local_manifest.rs
```diff
@@ -857,6 +888,15 @@ impl LocalFolderManifest {
...
 
+    /// Apply the prevent sync pattern *on the local children* (i.e. not on
+    /// `base.children` as one could expect) to remove from it all the confined
+    /// entries and collect them into `remote_confinement_points`.
+    ///
+    /// This method is expected to be run on a local manifest that have just been created
+    /// from a remote manifest. This way it local children are currently a mere copy
+    /// of it remote children (i.e. `base.children`) and it's this method's goal to
+    /// filter out the entries that should be remotely confined (i.e. entries that
+    /// should be kept in `base.children`, but not appear in local children).
```
vxgmichel, line 899: That's not exactly true: it's also called when applying a new pattern, after restoring the remote confinement points.

But now that I checked the implementation, I think there's an issue: the new remote confinement points should be the entries that appear both in `children` AND `base.children` and that have a remote name that matches the `prevent_sync_pattern`.

So I would do 3 things:
- fix the implementation (and see how it reflects on the tests)
- update the docstring to explain that this method filters out entries from `children` that have a corresponding remote name that matches the `prevent_sync_pattern`
- update the docstring to note that this method is expected to run on a local manifest that doesn't have remote confinement point (i.e either a local manifest that has just been created from a remote manifest, or a local manifest that had their remote confinement points restored).

vxgmichel (commented):
Now that I read the comments that you've written, it seems to me that most of the confusion stems from the fact that the `LocalFolderManifest` type has transitional states:
- Local manifest without filtering of remote entries nor locally confined entries
- Local manifest with filtering of remote entries but no locally confined entries

Those states are not meant to be used for long, as they are steps in a transformation pipeline. They are not sent over the network, nor stored in the local database.

A better choice for structuring this logic would have been to use two (or more) dedicated types for those transitional states. Then the transformation would have made more sense in the following cycle:
```
RemoteManifest ------> WithRemoteContext ------> RemoteManifest
                       v              ^
                RemoteConfined    LocalRemoved
                       v              ^
LocalManifest -------> WithLocalContext -------> LocalManifest                                         
```

libparsec/crates/client/src/workspace/merge.rs
```diff
@@ -186,31 +186,68 @@ pub(super) fn merge_local_folder_manifest(
...
+    // we must do a proper children merge and update the manifest accordingly.
+
+    let mut merge_in_progress =
+        LocalFolderManifest::from_remote_with_restored_local_confinement_points(
+            remote,
+            prevent_sync_pattern,
+            local,
+            timestamp,
+        );
+    let remote = &merge_in_progress.base;
```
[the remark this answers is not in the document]
touilleMan: Given this code is about merging a local and a remote, I think it's clearer to refer as `remote` instead of `merge_in_progress.base` when doing things like:
```rust
    let merged_parent = merge_parent(*local_base_parent, *local_parent, remote.parent);
```

libparsec/crates/client/src/workspace/merge.rs
```diff
@@ -186,31 +186,68 @@ pub(super) fn merge_local_folder_manifest(
...
+
+    // 2) Create a local manifest uniquely from the remote, this would give us the new
+    // local manifest if we hadn't already an existing local (i.e. if we were fetching
+    // this manifest for the very first time... which is obviously not the case here
+    // since we are doing a merge !)
+
+    // 3) Confinement points is a special case: given any entry that match the prevent
+    // sync pattern is kept appart, there is no possibility of conflict here (i.e.
+    // confined local entries stay in `local.children`, confined remote entries stay in
+    // `local.base.children`).
```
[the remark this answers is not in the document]
touilleMan: yup, old comment kept around "just in case" ^^

libparsec/crates/client/src/workspace/merge.rs
```diff
@@ -186,31 +186,68 @@ pub(super) fn merge_local_folder_manifest(
...
+            prevent_sync_pattern,
+            local,
+            timestamp,
+        );
+    let remote = &merge_in_progress.base;
+
+    // 2) Create a local manifest uniquely from the remote, this would give us the new
+    // local manifest if we hadn't already an existing local (i.e. if we were fetching
+    // this manifest for the very first time... which is obviously not the case here
+    // since we are doing a merge !)
```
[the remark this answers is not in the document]
touilleMan: same as above: old outdated comment that should be removed

libparsec/crates/client/src/workspace/merge.rs
```diff
@@ -232,21 +269,51 @@ pub(super) fn merge_local_folder_manifest(
...
+            speculative: merged_speculative,
+        } = &mut merge_in_progress;
+
+        *merged_speculative = false;
+        *merged_parent = *local_parent;
+        *merged_need_sync = *local_need_sync;
+        *merged_updated = local.updated;
+        *merged_children = local.children.to_owned();
+        *merged_local_confinement_points = local.local_confinement_points.to_owned();
+        *merged_remote_confinement_points = local.remote_confinement_points.to_owned();
```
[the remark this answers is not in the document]
touilleMan: then `merge_in_process.local_confinement_points` should also be used ?
(I guess we really don't want to have local and remote confinement points out of sync)

libparsec/crates/types/src/local_manifest.rs
```diff
@@ -821,13 +830,35 @@ impl LocalFolderManifest {
...
+    ///
+    /// This method can be seen as the opposite of `filter_remote_entries`.
+    ///
+    /// For example, considering a manifest with:
+    /// - `remote_confinement_points`: [1]
+    /// - `children`: `{"b.txt": 2}`
+    /// - `base.children`: `{"a.tmp": 1, "c.tmp": 3}`
+    ///
+    /// Then the resulting manifest would have an empty `remote_confinement points` and
+    /// `{"a.tmp": 1, "b.txt": 2}` as `children`.
```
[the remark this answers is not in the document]
touilleMan: brainchild of a drunken lost soul , to be sure !

libparsec/crates/types/src/local_manifest.rs
touilleMan: :+1:

libparsec/crates/types/src/local_manifest.rs
touilleMan: :+1: 
I've updated the comment:
```rust
    /// - `other` doesn't need to be up to date with `prevent_sync_pattern` (i.e.
    ///   `apply_prevent_sync_pattern()` may have been called on it with a different
```

libparsec/crates/client/src/workspace/merge.rs
```diff
@@ -186,31 +186,68 @@ pub(super) fn merge_local_folder_manifest(
...
-        timestamp,
-    );
+    // TODO: do we need this step ?
+    // // 2) Allow to force re-applying the prevent sync pattern (idempotent)
+    // //
+    // // The prevent sync pattern can change at any time, which in turn triggers
+    // // a full refresh of all local manifest to update their confinement points.
+    // // Here we make sure the local manifest doesn't have its confinement points
+    // // outdated, which would cause inconsistency in the next step.
+    // let local = local.apply_prevent_sync_pattern(prevent_sync_pattern, timestamp);
```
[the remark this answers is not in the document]
touilleMan: Out of curiosity: given `LocalFolderManifest::restore_local_confinement_points` is able to work with mismatching `prevent_sync_pattern`, what is the issue here if `local` has a mismatching `prevent_sync_pattern` ?

libparsec/crates/types/src/local_manifest.rs
touilleMan: On top of that, if `prevent_sync_pattern` can mismatch then I guess this sentence is wrong:
```rust
    /// Restore in the current manifest all the entries that were confined (according to
    /// the given `prevent_sync_pattern`) in the `other` manifest.
```

The current manifest gets filled with confined entries from `other` (whatever the prevent sync pattern used to generate the confined entries list in `other`).

And given `self.remote_confinement_points` is also used, I guess the comment can be turned into:
```rust
     /// Restore in the current manifest all the entries that:
    /// - Were confined in `other` (according to `other`'s `local_confinement_points`).
    /// - Were in `other` but are now part of our remote confinement points.
    ///
    /// On top of that this method also apply `prevent_sync_pattern` on the restored
    /// entries.
    ///
    /// This method can be seen as the opposite of `filter_local_confinement_points`.
    ///
    /// Notes:
    /// - This method must be called only after `filter_remote_entries`, otherwise
    ///   our `remote_confinement_points` field may be out-of-sync with
    ///   `prevent_sync_pattern` and lead to inconsistent results.
    /// - `other` doesn't need to be up to date with `prevent_sync_pattern` (i.e.
    ///   `apply_prevent_sync_pattern()` may have been called on it with a different
    ///   prevent sync pattern).
    /// - `timestamp` is only used to update the `updated` field if a previously
    ///   confined entry is now not confined (and hence should now be synchronized !).
    ///
    ///
    /// For example, considering `.tmp` as prevent sync pattern and a manifest with:
    /// - `remote_confinement_points`: [1, 2]
    /// - `children`: `{"3.txt": 3}`
    /// - `base.children`: `{"1.tmp": 1, "2.tmp": 2, "3.txt": 3}`
    ///
    /// And an `other` manifest with:
    /// - `local_confinement_points`: [4]
    /// - `children`: `{"3.txt": 3, "2-renamed.txt": 2, "a.tmp": 4, "b.txt": 5}`
    ///
    /// Then the resulting manifest would have:
    /// - `base.children`: `{"1.tmp": 1, "2.tmp": 2, "3.txt": 3}` (no changes)
    /// - `remote_confinement_points`: [1, 2] (no changes)
    /// - `local_confinement_points`: []
    /// - `children`: `{"3.txt": 3, "2-renamed.txt": 2, "a.tmp": 4}`
    ///
    /// Note that `"b.txt": 5` is not restored as it is a local change that has nothing
    /// to do with confinement. Hence the output of this method is not to be considered
    /// as a merge with `other` (for that see `libparsec_client`'s `merge_local_folder_manifest`).
```

libparsec/crates/types/src/local_manifest.rs
touilleMan: I've updated the comment:
```rust
    /// Apply the prevent sync pattern *on the local children* (i.e. not on
    /// `base.children` as one could expect) to remove from it all the confined
    /// entries and collect them into `remote_confinement_points`.
    ///
    /// This method is expected to be run on a local manifest where it's local children
    /// are currently a mere copy of it remote children (i.e. `base.children`).
    /// This occurs in two places:
    /// - When creating a new local manifest from a remote manifest.
    /// - When applying a new prevent sync pattern, in which case `filter_local_confinement_points`
    ///   and `restore_remote_confinement_points` has just been previously called to revert
    ///   the local children according to the local&remote confinement points.
    ///
    /// Once in this state, it is the method's goal to filter out the entries that
    /// should be remotely confined (i.e. entries that should be kept in `base.children`,
    /// but not appear in local children).
```

concerning the change in the implementation, I'm totally puzzled on what is the reasoning here...

I've considered that when this method is called we should have `self.children == self.base.children`, however you say we should check for both `self.children` and `self.base.children` which would indicate my assumption is wrong.
Can you give an example where calling `filter_local_confinement_points` then `restore_remote_confinement_points` doesn't lead to a local children equal to it remote (I suspect this has to do with renaming fields) ? I think it would be good to explain this case in the documentation.

libparsec/crates/client/src/workspace/merge.rs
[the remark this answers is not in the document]
vxgmichel: This is the special case I was explaining in an earlier conversation in slack, and maybe it should be dealt with with a bit more care. But the main idea is: 
- `local_confinement_points` relates to the entries in `children`, so we should modify neither of them in this case.
- `remote_confinement_points` relates to the entries in `base.children`, so we should modify both in this case.

But again, this should be validated with more tests since this was most probably already broken in v2.

libparsec/crates/types/src/local_manifest.rs
vxgmichel: I think `a.tmp` would end up in the local confinement points:
```
    /// Then the resulting manifest would have:
    /// - `base.children`: `{"1.tmp": 1, "2.tmp": 2, "3.txt": 3}` (no changes)
    /// - `remote_confinement_points`: [1, 2] (no changes)
    /// - `local_confinement_points`: [4]
    /// - `children`: `{"3.txt": 3, "2-renamed.txt": 2, "a.tmp": 4}`
```

libparsec/crates/types/src/local_manifest.rs
vxgmichel: > Can you give an example

Imagine this local file manifest with an old `prevent_sync_pattern` that matches nothing:
```
LocalFolderManifest
- base.children: {}
- children: {"a.tmp": 1}
- local_confinement_points: []
- remote_confinement_points: []
```

Note that a `a.tmp` has been created locally but it hasn't been synced yet.

Then we call `apply_prevent_sync_pattern` on it with the `*.tmp` filter. Remember this implementation:
```rust
        let result = self.clone();
        result
            .filter_local_confinement_points()
            .restore_remote_confinement_points()
            .filter_remote_entries(prevent_sync_pattern)
            .restore_local_confinement_points(self, prevent_sync_pattern, timestamp)
```

So after `filter_local_confinement_points` and `restore_remote_confinement_points` the manifest hasn't changed (cause it has no local nor remote confinement points). But after `filter_remote_entries(prevent_sync_pattern)`, with the current implementation, the manifest becomes:
```
LocalFolderManifest
- base.children: {}
- children: {}
- local_confinement_points: []
- remote_confinement_points: [1]
```
This breaks my assumption that all entries in `remote_confinement_points` should appear in `base.children`. This is surprising but doesn't lead to a bug, due to the implementation of `restore_local_confinement_points` which will restore `a.tmp` due to `1` being in `self.remote_confinement_points`. The end manifest is:
```
LocalFolderManifest
- base.children: {}
- children: {"a.tmp": 1}
- local_confinement_points: [1]
- remote_confinement_points: [1]
```
One might think that `1` being part of the `remote_confinement_points` is an issue, but it will be simply cleared by the next call to `restore_remote_confinement_points` (because the logic is: restore any child present in `base.children` if its id is present in `remote_confinement_points`, then clear `remote_confinement_points`)

vxgmichel:
I've been thinking a bit more about the structure proposed in https://github.com/Scille/parsec-cloud/pull/7756#pullrequestreview-2197086671:
```
RemoteManifest ------> WithRemoteContext ------> RemoteManifest
                       v              ^
                RemoteConfined    LocalRemoved
                       v              ^
LocalManifest -------> WithLocalContext -------> LocalManifest                                         
```
I noticed 2 things:
- `RemoteConfined` and `LocalRemoved` are only there as a step to illustrate the 4 existing methods but don't really serve a purpose as a class. They can be abstracted away by merging the 4 methods into 2.
- `WithLocalContext` doesn't seem to have a meaningful differences with `LocalManifest`

By renaming `WithRemoteContext` by `UnconfinedLocalManifest` we end up with
```
RemoteManifest -> UnconfinedLocalManifest -> RemoteManifest
                      v           ^
                      LocalManifest
```                                      

In this case, `UnconfinedLocalManifest` has the same fields as `LocalManifest`, except for the `local_confinement_points` and `remote_confinement_points` fields which are absent.

It has 4 methods:
- `to_remote(self) -> RemoteManifest`
- `from_remote(remote: RemoteManifest) -> Self`
- `apply_confinement(self, existing_local_manifest: Option<LocalManifest>, prevent_sync_pattern: Pattern) -> LocalManifest`
- `remove_confinement(local_manifest: LocalManifest) -> Self`

This way, the full confinement logic is maintained in `apply_confinement` and `remove_confinement`.  More precisely:
- `apply_confinement` is equivalent to the the old `filter_remote_entries + restore_local_confinement_points`
- `remove_confinement` is equivalent to the the old `filter_local_confinement_points + restore_remote_confinement_points`

With this design, the semantics of the fields `remote/local_confinement_points` of `LocalManifest` don't change over time, and more robust invariants can be defined and enforced.

Note: the merging of `filter_remote_entries` and `restore_local_confinement_points` together into `apply_confinement` would allow to deal more explicitly with the subtleties explained in this comment: https://github.com/Scille/parsec-cloud/pull/7756#discussion_r1694001074

I'll create a new issue that can be scheduled after the public release of parsec v3.

libparsec/crates/client/src/workspace/merge.rs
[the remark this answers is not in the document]
touilleMan: I agree with `assert_eq!(remote, merge_in_progress.base)`, however this code is about merging `remote` and `local`.

So I think it's more readable to see:
```rust
    let merge_need_sync = merged_children != remote.children || merged_parent != remote.parent;
```
given it's simply obvious we are comparing the merged data with the remote data.

While doing things like `merged_children != merge_in_progress.base.children` is more convoluted given you have one more conversion to do in your head.

libparsec/crates/client/src/workspace/merge.rs
```diff
@@ -23,35 +23,122 @@ pub(super) enum MergeLocalFolderManifestOutcome {
...
+
 pub(super) fn merge_local_file_manifest(
     local_author: DeviceID,
     timestamp: DateTime,
-    local_manifest: &LocalFileManifest,
-    remote_manifest: FileManifest,
+    local: &LocalFileManifest,
+    remote: FileManifest,
 ) -> MergeLocalFileManifestOutcome {
     // 0) Sanity checks, caller is responsible to handle them properly !
```
AureliaDolo, line 121: It would be nice to add that to the function doc

libparsec/crates/client/src/workspace/merge.rs
```diff
@@ -193,21 +284,52 @@ pub(super) fn merge_local_folder_manifest(
...
+
+    // 4) Shortcut in case only the remote has changed
     if !local_need_sync {
-        return MergeLocalFolderManifestOutcome::Merged(Arc::new(local_from_remote));
+        // Note a weird corner case here:
+        // If the prevent sync pattern has changed but the local manifest hasn't had
+        // time to be updated yet, then it's possible to end up with
+        // `local_need_sync == false`, but `merge_in_progress.need_sync == true` !
+
+        // TODO: determine if that's an issue !!!
```
AureliaDolo, line 326: Is there a github issue ?

libparsec/crates/types/src/local_manifest.rs
```diff
@@ -924,7 +1007,17 @@ impl LocalFolderManifest {
         .filter_remote_entries(prevent_sync_pattern)
     }
 
-    pub fn from_remote_with_local_context(
+    /// Create a `LocalFolderManifest` from the provided `FolderManifest`, then
+    /// apply to the new local folder manifest the entry that were locally confined
+    /// in the provided `local_manifest`.
+    ///
+    /// The result shouldn't be considered a merge of remote and local, but a intermedia
```
AureliaDolo, line 1014:
```suggestion
    /// The result shouldn't be considered a merge of remote and local, but an intermediate
```

AureliaDolo (approved):

libparsec/crates/types/src/local_manifest.rs
touilleMan: :+1: done.

libparsec/crates/client/src/workspace/merge.rs
touilleMan: yes https://github.com/Scille/parsec-cloud/issues/7924

libparsec/crates/client/src/workspace/merge.rs
touilleMan: I've added a docstring similar to the one for `merge_local_folder_manifest`

AureliaDolo (approved):

PR #7758, opened 2024-07-23
[MS] Added the skeleton of the customer area

Max-7:
Adds a few pages to handle the customer login to the BMS and the customer area.

- [X] Describe your changes

### DIFF
--- a/client/src/router/types.ts
+++ b/client/src/router/types.ts
@@ -17,6 +17,8 @@ export enum Routes {
   MyProfile = 'myProfile',
   RecoveryExport = 'recoveryExport',
   Loading = 'loading',
+  ClientAreaLogin = 'clientLogin',
+  ClientArea = 'clientArea',
 }
 
 const routes: Array<RouteRecordRaw> = [
@@ -33,6 +35,16 @@ const routes: Array<RouteRecordRaw> = [
     path: `/${Routes.Loading}`,
     name: Routes.Loading,
     component: () => import('@/views/layouts/LoadingLayout.vue'),
+  },
+  {
+    path: `/${Routes.ClientAreaLogin}`,
+    name: Routes.ClientAreaLogin,
+    component: () => import('@/views/client-area/AuthenticationPage.vue'),
+  },
+  {
+    path: `/${Routes.ClientArea}`,
+    name: Routes.ClientArea,
+    component: () => import('@/views/client-area/ClientAreaPage.vue'),
   },
   {
     // ConnectedLayout ensure that every children components are provided
--- a/client/src/services/bms/api.ts
+++ b/client/src/services/bms/api.ts
@@ -1,6 +1,17 @@
 // Parsec Cloud (https://parsec.cloud) Copyright (c) BUSL-1.1 2016-present Scille SAS
 
-import { AuthenticationToken, BmsError, BmsResponse, CreateOrganizationQueryData, DataType, LoginQueryData } from '@/services/bms/types';
+import {
+  AuthenticationToken,
+  BmsError,
+  BmsResponse,
+  CreateOrganizationQueryData,
+  DataType,
+  InvoicesQueryData,
+  ListOrganizationsQueryData,
+  LoginQueryData,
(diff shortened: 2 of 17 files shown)


client/src/services/bms/connection.ts
NicoTuxx, line 1: Would not authentication be more appropriate here?

client/src/router/types.ts
NicoTuxx, line 1: Seems more appropriate to rename 'Customer' to 'Client':
- we sells solutions not products
- in the BMS it is called Client, would make sense to keep the same

client/src/services/bms/connection.ts
```diff
@@ -0,0 +1,63 @@
...
+
+  get(): BmsConnection {
+    if (!this.connection) {
+      this.connection = new BmsConnection();
+    }
+    return this.connection;
+  }
+}
+
+export const BmsConnectionInstance = new ConnectionInstance();
```
NicoTuxx, line 63: Is there a reason to do this way instead of inject/provide or ways we already do in the project?

NicoTuxx (changes requested):
I would put all views in a `clientArea` folder to separate it more from the parsec views.
You can keep the bms and customers (renamed to clients) subfolders if you want.

client/src/router/types.ts
Max-7: Meh. I started with `client` and pivoted to `customer`, just because it's usually a less confusing word in computing.

client/src/services/bms/connection.ts
Max-7: Yeah, I suspect this is going to be used outside of a component. The more I look at it, the less I see why we bothered with inject/provide. It can be useful in some cases, like for the multi-org when we want to switch the instance being injected instead of having one global instance, but in that case we don't care: we just want one solo instance.

client/src/services/bms/api.ts
```diff
@@ -149,21 +133,85 @@ async function createOrganization(token: AuthenticationToken, query: CreateOrgan
...
+        headers: { Authorization: `Bearer ${token}` },
+        validateStatus: (status) => status === 200,
+      });
+    return {
+      status: axiosResponse.status,
+      isError: false,
+      data: {
+        type: DataType.OrganizationStatus,
+        activeUsersLimit: axiosResponse.data.active_users_limit ?? undefined,
+        isBootstrap: axiosResponse.data.is_bootstrapped,
```
NicoTuxx, line 202: Why not keep `isBootstrapped`?

client/src/services/bms/api.ts
```diff
@@ -51,16 +61,25 @@ function parseBmsErrors(data: object): Array<BmsError> {
   return [];
 }
 
-async function login(query: LoginQueryData): Promise<BmsResponse> {
+async function wrapQuery(queryFunc: () => Promise<BmsResponse>): Promise<BmsResponse> {
```
NicoTuxx, line 65: A lot better than before, but would be maybe even more better with decorators, not necessary now, but to keep in mind in the future.

client/src/services/bms/authentication.ts
NicoTuxx, line 1: the change of `connection` to `àuthentication` needs to be applied to te whole file, not only the filename.

client/src/views/client-area/AuthenticationPage.vue
```diff
@@ -0,0 +1,44 @@
...
+
+async function onLoginSuccess(_token: AuthenticationToken, _personalInformation: PersonalInformationResultData): Promise<void> {
+  await goToClientArea();
+}
+
+async function goToClientArea(): Promise<void> {
+  loginInProgress.value = true;
+  setTimeout(async () => {
+    await navigateTo(Routes.ClientArea, { replace: true });
+  }, 500);
```
NicoTuxx, line 40: Why do you need the timeout?

NicoTuxx (changes requested):

NicoTuxx (approved):

PR #7761, opened 2024-07-23
[RFC-1011] In-memory server side implementation of the non-blocking invite transport

### LINKED ISSUE #7018: Update the internal invitation logic for better network failure resilience
**Update: [RFC 1011](https://github.com/Scille/parsec-cloud/blob/master/docs/rfcs/1011-non-blocking-invite-transport.md) describes design and implementation details to address this issue.**

## Task list

- [x] https://github.com/Scille/parsec-cloud/pull/7662
- [x] #7761
- [x] #8073
- [x] #8080
- [x] #8200
- [x] #8233

----

A couples of issues have been identified with the current invitation API:

- The nonce exchange is convoluted: the claimer has two routes:
	- `2a` performing the exchange `2_1` 
	- `2b` performing the exchanges `2_2` and `2_3`
	 
  while the greeter has two routes:
	- `2a` performing the exchanges `2_1` and `2_2`
	- `2b` performing the exchange `2_3`
- Performing blocking HTTP requests is an issue since it leads to gateway timeout
- The invitation process gets restarted after a connection loss. The invitation process requiring a lot of user actions, it has to be resilient to network failures
- When a user cancels an invitation due to incorrect SAS code, the other peer is simply notified that the invitation process needs to be restarted. No explanation can be provided since there is no way to know the reason behind the reset.

This issues can be addressed independently. For instance:
- A refactoring can be applied to turn the `2a` et `2b` routes into `2a`, `2b` and `2c` so that each corresponds to a specific exchange.
- https://github.com/Scille/parsec-cloud/pull/6982 is a proof of concept to add a keep-alive message in the reply body in order to avoid gateways timeout
- https://github.com/Scille/parsec-cloud/pull/6971 adds more resilience to network for failure by allowing commands to be retried.
- an optional `reason` field might be added to the `*_wait_peer` commands, that can be propagated to the `enrollment_wrong_state` error received by the other peer.

However, there's one issue that is not fixable without deeper changes. Consider the following scenario:
- Alice connects to the server to perform step `n`, submits her payload and block waiting for Bob
- Bob connects to the server to perform step `n`, submits his payload.
- Both peers are connected simultaneously, the current step advances to `n+1` and each requests return with the other peer's payload. 
- Now a network failure prevents one of the reply to reach either one of the peer, say Alice.
- Alice might want to retry the last step in order to get Bob's payload. However, the system has already advanced to the next step, and Bob's payload is no longer part of the conduit. This is because only the current payloads are stored in the database, not the previous ones.
- The server simply replies to Alice that she's in the wrong state, so here only option at this point is to restart the invitation process.

For this reason, it might be advised to go for a deeper solution addressing all the points mentioned above all at once. Here's what this solution might look like.

- Invitations can have several invitation conduits (specific to a greeter+claimer couple)
- Each invitation conduit can have several invitation attempts (where only the last one is active)
- Each attempts has several exchanges identified by their corresponding step, an integer counting from zero.
- Each exchange stores the corresponding claimer payload and the greeter payload, once they've arrived.

This way, two generic non-blocking operations can be defined:
- `greeter_exchange(step: int, payload: bytes) -> bytes | NotReady`
- `claimer_exchange(step: int, payload: bytes) -> bytes | NotReady`

These operations both do the following:
 - check that the previous step is completed, i.e both payloads have been submitted for `step-1`
 - register the corresponding payload at the current step if it's not already there (indempotency)
 - return the other peer payload if it has already been registered
 - return `NotReady` otherwise

Also, if `step` is zero  and the current attempt has already completed step 0, a new attempt is created, effectively resetting the invitation conduit.

Those two internal operations are then wrapped into one or several routes and exposed by the server. This route might embed specific logic to validate the data format of the payload, or it might be fully generic, letting the other peer validate the data (something the peer has to do anyway). To sum it up:

```markdown
|                 | Generic data format | Specific data format         |
|-----------------|---------------------|------------------------------|
| Single route    | Easier to implement | More checks                  |
| Multiple routes | Doesn't make sense  | Similar to existing solution |
```

Note that the route for step 0 will still have to be specialized since it has to allow for the peer to optionally provide the reason that caused them to reset the previous invite.

Alternative solution: have a route specifically used to cancel an ongoing attempt. In this case, a new attempt is only created if the new one has been explicitly cancelled. 

One major change with the previous routes is that they are no longer blocking. They simply return `NotReady` when the other peer has not submitted its payload yet. This means that the client has to explicitly poll the route until the status is ready. Since this route is only used for invitation, having  a delay like 1 second configured between each poll is most likely enough and not too heavy on the server. In the future, event-based solution can easily be implemented (e.g sending an SSE whenever a change occurs for a given invitation, indicating that the peer has to poll again).

Note that this solution is very resilient to network failures since polling is at the core of this solution anyway: a network error is dealt with the same logic as a `NotReady` result: try again later.

## What's the amount of work for this solution:

- The server is heavily impacted, but the amount of logic is also greatly reduced as the new approach is simpler than the previous exchange mechanism where the server had to make sure that both peers where connected simultaneously.
- The API routes might or might not be affected depending on where the chosen solution fits in the table above (e.g it's possible to keep the routes mostly as they where if we go for the multiple routes approach).
- The client core is slightly affected. The routes might have changed and the polling logic has to be added. However, the main logic of the invitation remains the same, and the correspond data structures too.
- The front-end is almost not affected, as the exposed core API should remain the same. Only the support for the reset reasons is added.

For reference, 7 commands are added in total:
- 3 new invited commands
    - `invite_claimer_start_greeting_attempt`
    - `invite_claimer_cancel_greeting_attempt`
    - `invite_claimer_step`
- 4 new authenticated commands
    - `invite_greeter_start_greeting_attempt`
    - `invite_greeter_cancel_greeting_attempt`
    - `invite_greeter_step`
    - `invite_complete`

The following types are added
- GreetingAttemptID
- CancelledGreetingAttemptReason
- ClaimerStep (defined as nested types in related commands)
- GreeterStep (defined as nested types in related commands)
- GreeterOrClaimer

### LINKED ISSUE #7890: Invite transport: `authenticated_cmds::invite_greeter_start_greeting_attempt`

AureliaDolo:
Fix #7890 

Implemented as described in [RFC 1011](https://github.com/Scille/parsec-cloud/pull/7662)


docs/development/how-to-add-a-type-with-bindings.md
```diff
@@ -0,0 +1,10 @@
+<!-- Parsec Cloud (https://parsec.cloud) Copyright (c) BUSL-1.1 2016-present Scille SAS -->
+
+# How to add a type with its python bindings
+
+1. Define it in the `libparsec/crates/types`
+2. Implement (or derive) Debug, Copy, Clone, Hash, Serialize, Deserialize, PartialEq, Eq, FromStr, ToString
+3. Use #[serde(rename_all = "SCREAMING_SNAKE_CASE")] (because python naming conventions are different)
```
FirelightFlagboy, line 7:
```suggestion
3. Use `#[serde(rename_all = "SCREAMING_SNAKE_CASE")]` (because python naming conventions are different)
```

docs/development/how-to-add-a-type-with-bindings.md
```diff
@@ -0,0 +1,10 @@
+<!-- Parsec Cloud (https://parsec.cloud) Copyright (c) BUSL-1.1 2016-present Scille SAS -->
+
+# How to add a type with its python bindings
+
+1. Define it in the `libparsec/crates/types`
+2. Implement (or derive) Debug, Copy, Clone, Hash, Serialize, Deserialize, PartialEq, Eq, FromStr, ToString
```
FirelightFlagboy, line 6:
```suggestion
2. Implement (or derive):
   - `Debug`
   - `Copy` (Optional)
   - `Clone`
   - `Hash`
   - `Serialize`
   - `Deserialize`
   - `PartialEq`
   - `Eq`
   - `FromStr`
   - `ToString`
```

FirelightFlagboy, line 6: We do you get that list ? seems strange to require implementing `Hash`, `Copy`, `FromStr` for all types

libparsec/crates/client/src/workspace/tests/unit/workspace/file_operations_stateful.txt
FirelightFlagboy, line 1: Why that file exist ?

libparsec/crates/client/src/workspace/tests/unit/workspace/file_transactions_stateful.txt
FirelightFlagboy, line 1: Why that file exist ?

libparsec/crates/protocol/schema/authenticated_cmds/invite_greeter_step.json5
```diff
@@ -0,0 +1,216 @@
...
+        ],
+        "nested_types": [
+            {
+                // GreeterStep should be identical to the one in invite_claimer_step.json5
+                "name": "GreeterStep",
+                "discriminant_field": "step",
+                "variants": [
+                    {
+                        "name": "WaitPeer",
+                        "discriminant_value": "0",
```
FirelightFlagboy, line 93: Use proper value instead of fake number ? the discriminant value is always a string, so better to use a value that is more human readable :thinking:

libparsec/crates/protocol/schema/invited_cmds/invite_claimer_step.json5
```diff
@@ -0,0 +1,223 @@
...
+        ],
+        "nested_types": [
+            {
+                // GreeterStep should be identical to the one in invite_greeter_step.json5
+                "name": "GreeterStep",
+                "discriminant_field": "step",
+                "variants": [
+                    {
+                        "name": "WaitPeer",
+                        "discriminant_value": "0",
```
FirelightFlagboy, line 100: Use proper value instead of fake number ? the discriminant value is always a string, so better to use a value that is more human readable :thinking:

libparsec/crates/protocol/tests/authenticated_cmds/v4/invite_greeter_cancel_greeting_attempt.rs
```diff
@@ -0,0 +1,130 @@
...
+    // Content:
+    //   status: 'greeting_attempt_already_cancelled'
+    //   origin: 'GREETER'
+    //   reason: 'MANUALLY_CANCELLED'
+    //   timestamp: ext(1, 946774800000000) i.e. 2000-01-02T02:00:00Z
+    let raw: &[u8] = hex!(
+    "84a6737461747573d9226772656574696e675f617474656d70745f616c72656164795f"
+    "63616e63656c6c6564a66f726967696ea747524545544552a6726561736f6eb24d414e"
+    "55414c4c595f43414e43454c4c4544a974696d657374616d70d70100035d162fa2e400"
+    )
```
FirelightFlagboy, line 79:
```suggestion
    let raw: &[u8] = hex!(
        "84a6737461747573d9226772656574696e675f617474656d70745f616c72656164795f"
        "63616e63656c6c6564a66f726967696ea747524545544552a6726561736f6eb24d414e"
        "55414c4c595f43414e43454c4c4544a974696d657374616d70d70100035d162fa2e400"
    )
```

Indentation seems off, `rustfmt` did not format the tests ?

server/parsec/_parsec_pyi/protocol/__init__.pyi
```diff
@@ -23,4 +23,4 @@ class ActiveUsersLimit:
     def __lt__(self, other: object) -> bool: ...
     def __ne__(self, other: object) -> bool: ...
 
-__all__ = ["ActiveUsersLimit", "authenticated_cmds", "anonymous_cmds", "invited_cmds"]
+__all__ = ["ActiveUsersLimit", "invited_cmds", "authenticated_cmds", "anonymous_cmds"]
```
FirelightFlagboy, line 26: No longer ASCII sorted ?

libparsec/crates/protocol/tests/authenticated_cmds/v4/invite_greeter_cancel_greeting_attempt.rs
AureliaDolo: Rustfmt doesn't always interact well with macros :grimacing:

server/tests/test_greeting_attempt.py
```diff
@@ -0,0 +1,254 @@
...
+    """
+    The load and dump of the step data into the server database is not protected by the usual JSON schema
+    since this process is internal to the server. Instead, this logic is implemented in the `process_greeter_step`
+    and `process_claimer_step` functions.
+
+    This test contains snapshot data to test against change in this logic. Such a change would require a database
+    migration, so make sure to have a migration plan if you end up changing the snapshot data.
+    """
+    # Greeter data
+    # spell-checker: disable
```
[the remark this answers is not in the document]
vxgmichel: There's about ten lines to disable, so I'll keep the `disable/enable` commands

libparsec/crates/protocol/tests/authenticated_cmds/v4/invite_complete.rs
```diff
@@ -0,0 +1,109 @@
...
+    //   token: 0xd864b93ded264aae9ae583fd3d40c45a
+    let raw: &[u8] = hex!(
+    "82a3636d64af696e766974655f636f6d706c657465a5746f6b656ec410d864b93ded26"
+    "4aae9ae583fd3d40c45a"
+    )
+    .as_ref();
+    let req = authenticated_cmds::invite_complete::Req {
+        token: InvitationToken::from_hex("d864b93ded264aae9ae583fd3d40c45a").unwrap(),
+    };
+    println!("***expected: {:?}", req.dump().unwrap());
```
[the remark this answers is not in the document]
vxgmichel: Is there a value in removing them? it's unlikely we'll change the schemas anytime soon but I find those quite convenient, for instance when copy-pasting for new tests.

libparsec/crates/protocol/schema/authenticated_cmds/invite_greeter_step.json5
vxgmichel: @touilleMan I don't have a strong opinion about this, what do you think?

As far as I understand, the discriminant value is only used by serde to discriminate when loading bytes, do does it need to be human readable? Plus, it documents the order of the steps so I found numbering them useful.

server/parsec/_parsec_pyi/protocol/__init__.pyi
vxgmichel: 0554fc46defbcc3e3628f4e61f35dcea9acb8a22

libparsec/crates/protocol/schema/invited_cmds/invite_claimer_step.json5
vxgmichel: See above.

libparsec/crates/types/src/invite.rs
```diff
@@ -71,6 +71,92 @@ pub enum InvitationStatus {
...
+
+#[derive(Debug, Copy, Clone, Hash, Serialize, Deserialize, PartialEq, Eq)]
+#[serde(rename_all = "UPPERCASE")]
+pub enum GreeterOrClaimer {
+    Greeter,
+    Claimer,
+}
+
+impl FromStr for GreeterOrClaimer {
+    type Err = &'static str;
```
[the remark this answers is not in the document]
AureliaDolo: So it should be done for RealmRole too ?

https://github.com/Scille/parsec-cloud/blob/982976345de5f7610ca362dd93ec42aa693457c4/libparsec/crates/types/src/manifest.rs#L209-L210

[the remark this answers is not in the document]
FirelightFlagboy: Yes, but not in this PR ;) cf: #8005

server/parsec/components/invite.py
```diff
@@ -276,6 +281,253 @@ class InviteAsInvitedInfoBadOutcome(BadOutcomeEnum):
...
+@dataclass
+class GreetingAttemptCancelledBadOutcome(BadOutcome):
+    origin: GreeterOrClaimer
+    reason: CancelledGreetingAttemptReason
+    timestamp: DateTime
+
+
+@dataclass
+class NotReady:
+    pass
```
AureliaDolo, line 297: Here notReady doesn't need any data ?

server/parsec/components/memory/datamodel.py
```diff
@@ -249,10 +253,149 @@ class MemoryInvitation:
...
+    # TODO: remove the old fields once the new system is fully deployed
+    greeting_sessions: dict[UserID, MemoryGreetingSession] = field(default_factory=dict)
+
     @property
     def is_deleted(self) -> bool:
         return self.deleted_on is not None
 
+    @property
+    def is_completed(self) -> bool:
+        return self.deleted_reason == MemoryInvitationDeletedReason.FINISHED
```
AureliaDolo, line 266: Do we keep old invitations for ever ?

server/parsec/components/memory/datamodel.py
```diff
@@ -249,10 +253,149 @@ class MemoryInvitation:
...
+            greeter_id=self.greeter_id,
+        )
+        org.greeting_attempts[attempt.greeting_attempt] = attempt
+        self.greeting_attempts.append(attempt.greeting_attempt)
+        return attempt
+
+    def new_attempt_for_greeter(
+        self, org: MemoryOrganization, now: DateTime
+    ) -> MemoryGreetingAttempt:
+        while True:
```
AureliaDolo, line 309: Should we add a limit for the number of retries, just in case ?

server/parsec/components/memory/datamodel.py
```diff
@@ -249,10 +253,149 @@ class MemoryInvitation:
...
+    # Mutable properties
+    claimer_joined: None | DateTime = None
+    greeter_joined: None | DateTime = None
+    cancelled_reason: None | tuple[GreeterOrClaimer, CancelledGreetingAttemptReason, DateTime] = (
+        None
+    )
+    greeter_steps: list[bytes] = field(default_factory=list)
+    claimer_steps: list[bytes] = field(default_factory=list)
+
+    class StepOutcome(Enum):
```
AureliaDolo, line 341: Why is stepOutcome defined inside of MemoryGreetingAttempt ?

Could it be shared with the postgre impl ?

server/parsec/components/invite.py
vxgmichel: Yes, it's just a status. It could have been part of the `*BadOutcome` enums but it's not really a "bad" outcome, it's just a status communicating that the other peer hasn't performed this step yet.

server/parsec/components/memory/datamodel.py
vxgmichel: At the moment yes. In general, almost no information is deleted from the server.

server/parsec/components/memory/datamodel.py
vxgmichel: In practice there could be at most one retry. I'll change the implementation to make it clearer.

server/parsec/components/memory/datamodel.py
vxgmichel: Because it's a return status for two methods of `MemoryGreetingAttempt`. This allows stuff like this without an extra import:
```python
        match attempt.greeter_step(step_index, greeter_data):
            case attempt.StepOutcome.MISMATCH:
                [...]
```
It's a matter of taste though, and I don't think it's worth sharing with the postgre impl.

server/parsec/components/memory/datamodel.py
vxgmichel: f79a54daf25763a1aa4140c5de40aa0190047647

server/parsec/components/memory/datamodel.py
AureliaDolo: Ok got it ^

libparsec/crates/protocol/schema/authenticated_cmds/invite_greeter_step.json5
vxgmichel: e53026bc1b0f71e1c57ed01f31d961201056d4d3

vxgmichel: Fixed to have the discriminant value match the name, and include the step number in the name

libparsec/crates/protocol/tests/authenticated_cmds/v4/invite_complete.rs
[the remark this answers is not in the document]
vxgmichel: https://github.com/Scille/parsec-cloud/commit/30c41f5ce02d512c02eaddaf20de8cfaa92ebcfc
is the proof that it was useful in the end ^^

docs/development/how-to-add-a-type-with-bindings.md
FirelightFlagboy: @vxgmichel I did receive any reply, why did you close that ?

server/parsec/components/memory/datamodel.py
FirelightFlagboy: But we don't have a production server ATM, so that don't pose any issues

mmmarcos (approved):
LGTM. 

@AureliaDolo @vxgmichel please update issue #7890 to match what has been done in this PR. Also please update issue #7018 to describe what remains to be done (there is already a Task List but it should be updated based on how you are planning to tackle remaining tasks).

docs/development/how-to-add-a-type-with-bindings.md
vxgmichel: Ooops I thought it was solved by @AureliaDolo since the snippet as marked as `Outdated`

AureliaDolo: Yes, I got carried away by the outdated

docs/development/how-to-add-a-type-with-bindings.md
```diff
@@ -0,0 +1,20 @@
+<!-- Parsec Cloud (https://parsec.cloud) Copyright (c) BUSL-1.1 2016-present Scille SAS -->
+
+# How to add a type with its python bindings
+
+1. Define it in the `libparsec/crates/types`
+2. Implement (or derive):
```
AureliaDolo, line 6:
```suggestion
2. Implement (or derive) the needed traits:
```

@FirelightFlagboy is that ok ?

docs/development/how-to-add-a-type-with-bindings.md
AureliaDolo: I did as it was for InvitationType, should have checked with other impl

docs/development/how-to-add-a-type-with-bindings.md
FirelightFlagboy: I'm OK, if the list below is updated as some traits are only required for certain capabilities

PR #7767, opened 2024-07-24
Fix non-linux WASM support for ZSTD

touilleMan:
Replace `use-zstd-in-serialization-format` feature by `libparsec_zstd/use-pure-rust-but-dirty-zstd`

Real C zstd implementation is always use for production, however it is cumbersome to
compile for WASM when on Windows/MacOS.

For this reason the new crate `libparsec_zstd` provides two implementations of zstd:

- The real C one (enabled by `libparsec_zstd/use-real-zstd`)
- A pure Rust one: easy to compile but with minimal support for compression (enabled
  by `libparsec_zstd/use-pure-rust-but-dirty-zstd`)

`libparsec_zstd/use-real-zstd` is automatically used when building non-WASM target ( 
see `libparsec_types`'s `Cargo.toml`).
When building WASM, it's up to the `make.py` script (that already handle other features
base on target platform and release/dev/ci build) to decide if
`libparsec_zstd/use-pure-rust-but-dirty-zstd` should be used instead based on the host OS.


make.py
```diff
@@ -15,6 +15,29 @@
...
+    "ฅ(・㉨・˶)ฅ",
+    "(ﾐꆤ ﻌ ꆤﾐ)∫",
+    "(ﾐㆁ㉨ㆁﾐ)",
+    "ฅ(=ච ω ච=)",
+]
+
+
+class LazyStr(str):
+    def __str__(self):
+        return self
```
FirelightFlagboy, line 38: It don't see the class used, it's expected ?

libparsec/crates/types/src/serialization.rs
```diff
@@ -79,18 +51,6 @@ where
...
-        // only wasm client which is reasonable in a dev scenario).
-        Some(&FORMAT_VFF_VERSION_BYTE) => {
-            rmp_serde::from_slice(&raw[1..]).map_err(|_| DataError::BadSerialization {
-                format: Some(FORMAT_VFF_VERSION_BYTE),
-                step: "msgpack+validation",
-            })
-        }
-
         Some(_) | None => Err(DataError::BadSerialization {
             format: None,
```
FirelightFlagboy, line 55: I would expect that `format` to be set to the invalid value

libparsec/crates/client/Cargo.toml
```diff
@@ -31,12 +31,5 @@ proptest-state-machine = { workspace = true }
 libparsec_platform_storage = { workspace = true, features = ["expose-test-methods"] }
 libparsec_client = {path = ".",features = ["expose-test-methods"]}
 
-[target.'cfg(any(not(target_arch = "wasm32"), target_os = "linux"))'.dependencies]
-# TODO: `use-zstd-in-serialization-format` is a temporary feature to disable the
-#       use of zstd, as its compilation with WASM is buggy on some host (e.g.
-#       MacOS/Windows/Ubuntu 20.04).
-#       TL;DR: ALWAYS ENABLE `use-zstd-in-serialization-format` FOR PRODUCTION !!!
```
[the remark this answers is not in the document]
touilleMan: I guess you are talking a bout the new `use-real-zstd` feature

I don't think `default` feature is what we need here: we need to enable this feature under certain (most) circumstances, and keep it disabled while enabling the other feature under other circumstances.
Given `default` is just a convenient way to enable a predefined set of features, we would end up with the same logic we currently have, only to know where to set `default` (which then would be less explicit that actually setting the feature we want in the first place)

make.py
touilleMan: removed.

libparsec/crates/types/src/serialization.rs
touilleMan: The step that has failed is marked to be `format detection`, then the `format` has to be set to `None` to indicate we are not in any format (i.e. this field is not to indicate what is the format value that is not recognized, I consider there is no need for such information, just like if zstd or msgpack fail de-serialization the specific error message is not provided)

AureliaDolo:
It would be nice to have a readme at crate level (and included as doc at top level) (like [here]()) that details what the feature does, where to use it, where to not use it etc.

touilleMan:
> It would be nice to have a readme at crate level (and included as doc at top level) (like [here](https://github.com/Scille/parsec-cloud/blob/master/libparsec/crates/serialization_format/src/lib.rs)) that details what the feature does, where to use it, where to not use it etc.

Done \o/

libparsec/crates/zstd/README.md
```diff
@@ -0,0 +1,61 @@
...
+```shell
+cargo build -p libparsec_whatever
+```
+
+Compiling with Zstd support provided by the pure-Rust dirty implementation (i.e.
+only do that for dev).
+
+```shell
+RUSTFLAGS='--cfg use_pure_rust_but_dirty_zstd cargo build -p libparsec_whatever
+```
```
FirelightFlagboy, line 36: Can we provide a `build.rs` to set the `cfg` automatically ?

libparsec/crates/zstd/src/dirty.rs
```diff
@@ -0,0 +1,157 @@
...
+    decoder.read_to_end(&mut result)?;
+
+    Ok(result)
+}
+
+/// Compress all data from the given source as if using an `Encoder`.
+///
+/// Compressed data will be appended to `destination`.
+///
+/// A level of `0` uses zstd's default (currently `3`).
```
AureliaDolo, line 22: You mean currently 3 in parsec overall ?

AureliaDolo (approved):

libparsec/crates/zstd/README.md
```diff
@@ -0,0 +1,61 @@
...
+under no circumstances end up being used in production !
+
+So using a feature would require more configuration in multiple places, and may
+lead to the wrong implementation being used by mistake while being hard to
+detect (this has already happened when generating test data !).
+
+Finally features cannot be enabled for transitive dependencies, this is an issue
+when using `wasm-pack test` on `libparsec_platform_storage` (given a feature must
+be passed to `libparsec_zstd`, but only `libparsec_types` depends on it so it
+cannot be done from `libparsec_platform_storage`).
```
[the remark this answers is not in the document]
touilleMan: Yup (that was my initial solution as a matter of fact), but then it works until we want to build another crate than doesn't directly depends on `libparsec_types`...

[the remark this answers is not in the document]
AureliaDolo: Yeah you'd have to re expose the feature again ...

libparsec/crates/zstd/src/dirty.rs
touilleMan: I've copied this comment from zstd's documentation :eyes:

touilleMan: I've updated the comment to:
```rust
/// Compress all data from the given source as if using an `Encoder`.
///
/// Compressed data will be appended to `destination`.
///
/// Note `level` is simply ignored here given this implementation of compressor
/// doesn't do any actual compression !
```

PR #7774, opened 2024-07-24
Cache prevent sync pattern in generic workspace storage

FirelightFlagboy:


FirelightFlagboy:
Depends on #7739

libparsec/crates/platform_storage/src/workspace.rs
```diff
@@ -217,20 +224,35 @@ impl WorkspaceStorage {
     }
 
     pub async fn set_prevent_sync_pattern(&mut self, pattern: &Regex) -> anyhow::Result<bool> {
-        self.platform.set_prevent_sync_pattern(pattern).await
+        // TODO: Should we skip the next call if the pattern is the same?
```
FirelightFlagboy, line 227: Help needed

libparsec/crates/platform_storage/src/workspace.rs
```diff
@@ -217,20 +224,35 @@ impl WorkspaceStorage {
...
+            self.prevent_sync_pattern_applied,
+        )
     }
 
     pub async fn mark_prevent_sync_pattern_fully_applied(
         &mut self,
         pattern: &Regex,
     ) -> anyhow::Result<bool> {
-        self.platform
+        // TODO: Should we skip the next call if the pattern is different?
```
FirelightFlagboy, line 247: Help needed

libparsec/crates/platform_storage/src/workspace.rs
```diff
@@ -217,20 +224,35 @@ impl WorkspaceStorage {
...
+        let applied = self
+            .platform
             .mark_prevent_sync_pattern_fully_applied(pattern)
-            .await
+            .await?;
```
FirelightFlagboy, line 251: Help needed

libparsec/crates/platform_storage/src/workspace.rs
```diff
@@ -217,20 +224,35 @@ impl WorkspaceStorage {
     }
 
     pub async fn set_prevent_sync_pattern(&mut self, pattern: &Regex) -> anyhow::Result<bool> {
-        self.platform.set_prevent_sync_pattern(pattern).await
+        // TODO: Should we skip the next call if the pattern is the same?
+        let applied = self.platform.set_prevent_sync_pattern(pattern).await?;
```
FirelightFlagboy, line 228: Help needed

libparsec/crates/platform_storage/src/workspace.rs
```diff
@@ -20,6 +20,8 @@ pub const TYPICAL_BLOCK_SIZE: u64 = 512 * 1024; // 512 KB
 #[derive(Debug)]
 pub struct WorkspaceStorage {
     platform: PlatformWorkspaceStorage,
+    prevent_sync_pattern: Regex,
+    prevent_sync_pattern_applied: bool,
```
touilleMan, line 24:
```suggestion
    prevent_sync_pattern_fully_applied: bool,
```

libparsec/crates/platform_storage/src/workspace.rs
```diff
@@ -217,20 +224,35 @@ impl WorkspaceStorage {
...
+
+        self.prevent_sync_pattern = pattern.clone();
+        self.prevent_sync_pattern_applied = applied;
+
+        Ok(applied)
     }
 
-    pub async fn get_prevent_sync_pattern(&mut self) -> anyhow::Result<(Regex, bool)> {
-        self.platform.get_prevent_sync_pattern().await
+    pub fn get_prevent_sync_pattern(&self) -> (&Regex, bool) {
```
touilleMan, line 236: Using cache here is great :+1:

libparsec/crates/platform_storage/src/workspace.rs
```diff
@@ -217,20 +224,35 @@ impl WorkspaceStorage {
     }
 
     pub async fn set_prevent_sync_pattern(&mut self, pattern: &Regex) -> anyhow::Result<bool> {
```
touilleMan, line 226:
```suggestion
    pub async fn set_prevent_sync_pattern(&mut self, pattern: Regex) -> anyhow::Result<bool> {
```
My own rule of thumb on passing by reference vs by ownership:
- if the ownership is always needed in the body of the function (which is the case here, cf `self.prevent_sync_pattern = pattern.clone();`) then it should be passed by ownership.
- if the ownership is needed in corner case (typically when building errors), then passing by reference likely the good approach

libparsec/crates/platform_storage/src/workspace.rs
touilleMan, line 247: I think it would be good to:
- rename `pattern`  -> `expected_pattern`
- return a dedicated error type:
```rust
enum MarkPreventSyncPatternFullyApplied {
    PatternHasChanged,
    Internal(anyhow::Error)
}
```

This way we can let the caller decide what to do, and on our side we do the comparison between `expected_pattern` and `self.prevent_sync_pattern` as the very first check so that we never modify anything if the pattern is unexpected

libparsec/crates/platform_storage/src/workspace.rs
```diff
@@ -217,20 +224,35 @@ impl WorkspaceStorage {
...
+        )
     }
 
     pub async fn mark_prevent_sync_pattern_fully_applied(
         &mut self,
         pattern: &Regex,
     ) -> anyhow::Result<bool> {
-        self.platform
+        // TODO: Should we skip the next call if the pattern is different?
+        let applied = self
```
touilleMan, line 248: Returning `applied` and using it to set `self.prevent_sync_pattern_applied` seems a bad idea to me:

- If the pattern is different but fully applied, then we get `true` as result.
- Obviously `self.platform.mark_prevent_sync_pattern_fully_applied` is about setting `prevent_sync_pattern_applied` in the database to `true`, so why the method should ever return `false` ?

In the implementation of `platform.mark_prevent_sync_pattern_fully_applied` there is this:
https://github.com/Scille/parsec-cloud/blob/cc8383e117390d13fd915bb013f087f84835ba48/libparsec/crates/platform_storage/src/native/workspace.rs#L520-L525

I think returning the information if the is fully applied in the case the pattern is not the one we are expecting is the wrong approach here (if the pattern is not the one we are expecting, then something wrong is going on since we shouldn't be handling this kind concurrency at the database level, cf. the sync pattern cache field)

So I would change the signature of `self.platform.mark_prevent_sync_pattern_fully_applied` to also return the `MarkPreventSyncPatternFullyAppliedError` that allows to return the fact the provided `expected_pattern` is no the one in database.
And if this error occurs, then a panic should be raised (since the normal operation is to have this error being handled when checking the cache value of the prevent sync pattern)

libparsec/crates/platform_storage/tests/unit/workspace.rs
```diff
@@ -1081,9 +1081,9 @@ async fn set_prevent_sync_pattern(env: &TestbedEnv) {
     let res = workspace.set_prevent_sync_pattern(&regex).await.unwrap();
 
     assert_eq!(res, false);
-    let (got_regex, bool) = workspace.get_prevent_sync_pattern().await.unwrap();
+    let (got_regex, bool) = workspace.get_prevent_sync_pattern();
 
-    assert_eq!(got_regex, regex);
+    assert_eq!(got_regex, &regex);
     assert_eq!(bool, false);
```
touilleMan, line 1087: Should also restart the workspace to ensure the prevent sync pattern is not only in cache

```rust
// Restart the workspace to ensure the new prevent sync pattern is not just in cache
workspace.stop().await.unwrap();
let workspace = start_workspace(env).await;
let (got_regex, bool) = workspace.get_prevent_sync_pattern();
assert_eq!(got_regex, &regex);
assert_eq!(bool, false);
```

touilleMan (changes requested):

libparsec/crates/platform_storage/src/workspace.rs
FirelightFlagboy: done

libparsec/crates/platform_storage/src/workspace.rs
FirelightFlagboy: done

libparsec/crates/platform_storage/tests/unit/workspace.rs
FirelightFlagboy: done

libparsec/crates/platform_storage/src/workspace.rs
FirelightFlagboy: done

libparsec/crates/platform_storage/src/workspace.rs
FirelightFlagboy: done

touilleMan (approved):

FirelightFlagboy:
Will be closed by #7843

FirelightFlagboy:
Superseded by #7843

PR #7778, opened 2024-07-25
Add `apply_prevent_sync_pattern` to `WorkspaceStore`

FirelightFlagboy:
The goal of this PR is to port the method `WorkspaceFS.apply_prevent_sync_pattern` that was used in `v2`:

https://github.com/Scille/parsec-cloud/blob/d397c035d1369e498a48dea26837d93b7291dfe4/parsec/core/fs/workspacefs/workspacefs.py#L1075-L1103

### DIFF
--- a/libparsec/crates/client/src/workspace/store/mod.rs
+++ b/libparsec/crates/client/src/workspace/store/mod.rs
@@ -10,12 +10,13 @@ mod resolve_path;
 mod sync_updater;
 
 use std::{
+    future::Future,
     path::Path,
     sync::{Arc, Mutex},
 };
 
 use libparsec_client_connection::AuthenticatedCmds;
-use libparsec_platform_async::lock::Mutex as AsyncMutex;
+use libparsec_platform_async::{future::TryFutureExt, lock::Mutex as AsyncMutex};
 use libparsec_platform_storage::workspace::{UpdateManifestData, WorkspaceStorage};
 use libparsec_types::prelude::*;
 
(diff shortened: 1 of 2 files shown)


libparsec/crates/client/src/workspace/store/mod.rs
```diff
@@ -562,4 +573,65 @@ impl WorkspaceStore {
...
+        Ok(())
+    }
+}
+
+fn recursive_apply_prevent_sync_pattern(
+    storage: &mut WorkspaceStorage,
+    entry_id: VlobID,
+    pattern: &Regex,
+    device: &LocalDevice,
+) -> Box<dyn Future<Output = Result<(), ApplyPreventSyncPatternError>>> {
```
FirelightFlagboy, line 606: TODO Handle RAM cache + DB cache.

I should use `CurrentViewCache`.

- Should I first try to update a manifest in the cache then update the database ?
- Only update the database and indicate the cache to load the data from the DB ?

libparsec/crates/client/src/workspace/store/mod.rs
```diff
@@ -562,4 +573,65 @@ impl WorkspaceStore {
...
+                // Fallback to workspace.
+                LocalWorkspaceManifest::decrypt_and_load(&manifest, &device.local_symkey)
+                    .map(|workspace| workspace.0.children)
+            })
+        else {
+            // TODO: Should it handle `DataError` in case the decryption failed or something else ?
+            return Ok(()); // Not a folder or workspace
+        };
+
+        // TODO: Apply prevent sync pattern
```
FirelightFlagboy, line 629: I'm currently blocked on that TODO, previously we used a `SyncTransactions.apply_prevent_sync_pattern` to apply the pattern but it don't seems that we use this pattern.

I suppose I should do something like `for_update_sync_local_only`, where I call `manifest.apply_prevent_sync_pattern`:

https://github.com/Scille/parsec-cloud/blob/7bb7e0e1b95ea017b7368cf05af72890748ccbb9/libparsec/crates/client/src/workspace/store/sync_updater.rs#L58-L61

FirelightFlagboy:
Superseded by #7843

PR #7781, opened 2024-07-25
Call `apply_prevent_sync_pattern` on `WorkspaceStore` creation

### LINKED ISSUE #7780: Call `apply_prevent_sync_pattern` method during workspace initialization
https://github.com/Scille/parsec-cloud/blob/d397c035d1369e498a48dea26837d93b7291dfe4/parsec/core/fs/workspacefs/workspacefs.py#L223

FirelightFlagboy:
Closes #7780

### DIFF
--- a/libparsec/crates/client/src/workspace/store/mod.rs
+++ b/libparsec/crates/client/src/workspace/store/mod.rs
@@ -185,14 +185,16 @@ impl WorkspaceStore {
 
         // 3) All set !
 
-        Ok(Self {
+        let workspace_store = Self {
             realm_id,
             device,
             cmds,
             certificates_ops,
             current_view_cache: Mutex::new(CurrentViewCache::new(Arc::new(root_manifest.into()))),
             storage: AsyncMutex::new(Some(storage)),
-        })
+        };
+        workspace_store.apply_prevent_sync_pattern().await?;
+        Ok(workspace_store)
     }
 
     pub async fn stop(&self) -> anyhow::Result<()> {


FirelightFlagboy:
Depends on #7778

libparsec/crates/client/src/workspace/store/mod.rs
```diff
@@ -185,14 +185,16 @@ impl WorkspaceStore {
...
+        let workspace_store = Self {
             realm_id,
             device,
             cmds,
             certificates_ops,
             current_view_cache: Mutex::new(CurrentViewCache::new(Arc::new(root_manifest.into()))),
             storage: AsyncMutex::new(Some(storage)),
-        })
+        };
+        workspace_store.apply_prevent_sync_pattern().await?;
```
touilleMan, line 196: I guess this operation can take a long time in case the sync pattern is not currently fully applied and the workspace is big enough (I guess `recursive_apply_prevent_sync_pattern` doesn't need to download manifest that are not yet available, otherwise it's even worse performance wise)

The issue is obviously this slow down the startup of the workspace. Another approach would be to introduce a dedicated monitor that would:
- on startup check the pattern is fully applied
- listen on an event fired whenever the pattern change (so there should be a `workspace.set_new_prevent_sync_pattern()` that doesn't do the apply but only set the value and fire the event.
- is responsible to run `workspace_store.apply_prevent_sync_pattern()` when needed (`workspace_store.apply_prevent_sync_pattern()` should probably have a lock to ensure it is never run more than once concurrently)
- on top of that, the monitor could abort any pending `workspace_store.apply_prevent_sync_pattern()` if a new event is fired

This way the long operation is run in a dedicated coroutine and doesn't slow down anything else \o/

touilleMan: Of course those changes are not vital for the moment but are good improvement, so I think it's good to talk about it now and, if we are ok, create an issue to track it

touilleMan (approved):

libparsec/crates/client/src/workspace/store/mod.rs
FirelightFlagboy: Continuing the discussion in #7799

FirelightFlagboy:
Superseded by #7843

PR #7782, opened 2024-07-25
Allow to force apply the prevent sync pattern during merge of local folder manifest

### LINKED ISSUE #7762: Enable `force_apply_pattern` when needed
https://github.com/Scille/parsec-cloud/blob/fc76ec52b7ac92370d3ed24cce3cbd14c50b5e9b/libparsec/crates/client/src/workspace/merge.rs#L239-L242

FirelightFlagboy:
Closes #7762

### DIFF
--- a/libparsec/crates/client/src/workspace/merge.rs
+++ b/libparsec/crates/client/src/workspace/merge.rs
@@ -150,7 +150,19 @@ pub(super) fn merge_local_folder_manifest(
     prevent_sync_pattern: Option<&Regex>,
     local: &LocalFolderManifest,
     remote: FolderManifest,
+    force_apply_pattern: bool,
 ) -> MergeLocalFolderManifestOutcome {
+    // 0) Sanity checks, caller is responsible to handle them properly !
+    debug_assert_eq!(local.base.id, remote.id);
+
+    let cleaned_up_local_manifest = if force_apply_pattern {
+        Some(local.apply_prevent_sync_pattern(prevent_sync_pattern, timestamp))
+    } else {
+        None
+    };
+
+    let local_manifest = cleaned_up_local_manifest.as_ref().unwrap_or(local);
+
     // Destruct local and remote manifests to ensure this code with fail to compile whenever a new field is introduced.
     let LocalFolderManifest {
         base:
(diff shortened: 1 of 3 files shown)


libparsec/crates/client/src/workspace/merge.rs
```diff
@@ -150,7 +150,19 @@ pub(super) fn merge_local_folder_manifest(
     prevent_sync_pattern: Option<&Regex>,
```
FirelightFlagboy, line 150: Should I remove the option ?

For me I can since I ways provide a value during the `inbound_sync`

FirelightFlagboy:
Depends on #7781

libparsec/crates/client/src/workspace/merge.rs
touilleMan: Yes, the `prevent_sync_pattern` should now always be present

(the `Option` was added in the codebase as a way to disable prevent sync pattern so we could skip it implementation, so it's totally useless once you job is done here ^^)

libparsec/crates/client/src/workspace/merge.rs
```diff
@@ -150,7 +150,19 @@ pub(super) fn merge_local_folder_manifest(
...
     local: &LocalFolderManifest,
     remote: FolderManifest,
+    force_apply_pattern: bool,
 ) -> MergeLocalFolderManifestOutcome {
+    // 0) Sanity checks, caller is responsible to handle them properly !
+    debug_assert_eq!(local.base.id, remote.id);
+
+    let cleaned_up_local_manifest = if force_apply_pattern {
```
touilleMan, line 158:
```suggestion
    let local_with_pattern_applied = if force_apply_pattern {
```

libparsec/crates/client/src/workspace/merge.rs
```diff
@@ -150,7 +150,19 @@ pub(super) fn merge_local_folder_manifest(
...
+        Some(local.apply_prevent_sync_pattern(prevent_sync_pattern, timestamp))
+    } else {
+        None
+    };
+
+    let local_manifest = cleaned_up_local_manifest.as_ref().unwrap_or(local);
```
touilleMan, line 164:
```suggestion
    let local = cleaned_up_local_manifest.as_ref().unwrap_or(local);
```
we have `remote`, `local` & `merge_in_progress` so I think it's more consistent to avoid using `_manifest` suffix here and just stick with `local`

libparsec/crates/client/src/workspace/merge.rs
```diff
@@ -150,7 +150,19 @@ pub(super) fn merge_local_folder_manifest(
     prevent_sync_pattern: Option<&Regex>,
     local: &LocalFolderManifest,
     remote: FolderManifest,
+    force_apply_pattern: bool,
```
touilleMan, line 153:
```suggestion
    pattern_already_applied_to_local: bool,
```
I'd rather rename this option to make it more explicit about when it is needed

FirelightFlagboy:
```suggestion
    pattern_not_applied_to_local: bool,
```

Should be the inverse, the option should be set to true if the pattern is not fully applied.
Otherwise if should inverse it's use in the if WDYT?

FirelightFlagboy:
With the PR #7843 we no longer need to force the application of the prevent sync pattern when it's not fully applied, since the pattern will always be fully applied

PR #7811, opened 2024-07-26
Fix clippy issue in platform_storage test

mmmarcos:


AureliaDolo (approved):
FYI if DebugDump i meant to evolve, you could mark it as [non exhaustive](https://doc.rust-lang.org/reference/attributes/type_system.html#the-non_exhaustive-attribute) and keep the default here. Here is an [example](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=df057dc5aa99414fbaf5fa3b28c7ee76)

FirelightFlagboy (approved):

PR #7832, opened 2024-07-29
Update indexdb to 0.5.0

FirelightFlagboy:
This update will be required to use the cursor feature


AureliaDolo (approved):
There's less than two hours between the two version releases :tada:

touilleMan (approved):

PR #7835, opened 2024-07-30
Add method to list manifest in database

FirelightFlagboy:


libparsec/crates/platform_storage/src/web/workspace.rs
```diff
@@ -140,6 +143,16 @@ impl PlatformWorkspaceStorage {
...
     }
 
+    pub async fn list_manifests(
+        &mut self,
+        page: u32,
+        limit: u32,
+    ) -> anyhow::Result<Vec<RawEncryptedManifest>> {
+        Vlob::list(&self.conn, page, limit)
+            .await
+            .map(|vlobs| vlobs.into_iter().map(|vlob| vlob.blob.to_vec()).collect())
```
touilleMan, line 153:
```suggestion
.map(|vlobs| vlobs.into_iter().map(|vlob| Vec::<u8>::from(vlob.blob)).collect())
```
This should allows us to do zero copy given the `Bytes` object is only referenced once here (see https://github.com/tokio-rs/bytes/issues/711#issuecomment-2150558312)

touilleMan: And as a side note: is there a reason to use `Bytes` types in the web model ?

I'm under the impression we are always passing `Vec<u8>` types (given the caller is passing data that have been serialized for the only purpose of being stored in the database) and returning a `Vec<u8>` is also okay (again, the caller will only do de-serialization of those data then drop the buffer).

So should we create an issue about converting those `Bytes` into `Vec<u8>` ?

touilleMan (changes requested):

libparsec/crates/platform_storage/src/workspace.rs
```diff
@@ -69,6 +69,8 @@ pub struct DebugBlock {
     pub accessed_on: String,
 }
 
+pub type RawEncryptedManifest = Vec<u8>;
```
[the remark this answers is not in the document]
FirelightFlagboy: Done, moved in #7842

FirelightFlagboy:
Depends on #7842

libparsec/crates/platform_storage/src/workspace.rs
```diff
@@ -133,10 +135,21 @@ impl WorkspaceStorage {
...
+    pub async fn get_manifest(
+        &mut self,
+        entry_id: VlobID,
+    ) -> anyhow::Result<Option<RawEncryptedManifest>> {
         self.platform.get_manifest(entry_id).await
     }
 
+    pub async fn list_manifests(
+        &mut self,
+        page: u32,
```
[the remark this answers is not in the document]
FirelightFlagboy: done

libparsec/crates/platform_storage/Cargo.toml
```diff
@@ -24,6 +24,7 @@ libparsec_testbed = { workspace = true, optional = true }
 log = { workspace = true }
 paste = { workspace = true }
 lazy_static = { workspace = true, optional = true }
+thiserror = { workspace = true }
```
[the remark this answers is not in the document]
FirelightFlagboy: Wasn't used, removed now

libparsec/crates/platform_storage/src/web/workspace.rs
FirelightFlagboy: done

libparsec/crates/platform_storage/src/native/workspace.rs
```diff
@@ -641,6 +650,30 @@ async fn db_get_manifest(
...
+) -> anyhow::Result<Vec<RawEncryptedManifest>> {
+    let rows = sqlx::query("SELECT blob FROM vlobs LIMIT ?1 OFFSET ?2")
+        .bind(limit)
+        .bind(offset)
+        .fetch_many(executor);
+
+    let res: anyhow::Result<Vec<RawEncryptedManifest>> = rows
+        .filter_map(|row_res| async {
+            match row_res {
+                Ok(sqlx::Either::Left(_res)) => None,
```
touilleMan, line 666: This `sqlx::Either` is pretty unclear, I had to do some hard digging into sqlx & sqlx-sqlite to properly understand it...

So I suggest adding a comment:
```rust
// `fetch_many` returns a single "left" element for the query result (containing, among others,
// the number of changes), then (what we are actually interested into here)  one "right"
// elements for each row in the result.
```

touilleMan (approved):

libparsec/crates/platform_storage/src/native/workspace.rs
FirelightFlagboy: done

PR #7839, opened 2024-07-30
Fix headers in PO files

mmmarcos:


touilleMan (approved):
We should update `misc/version_updater.py` script to take into account those files

mmmarcos:
> We should update `misc/version_updater.py` script to take into account those files

Actually, `version_updater.py` already updates `docs/conf.py` which is used by Sphinx to generate POT files (templates). The problem: when `sphinx-intl` updates existing PO files, it only updates messages being changed, but not the PO headers:
- Headers in POT file have the right version, but they are temporary files
- Headers in PO file does not change on update
  - In the usual translation workflow, PO files (including headers) are responsibility of the translator.  

Anyway, that doesn't mean we can't update PO headers via `version_updater.py`. I'll create an issue.

PR #7843, opened 2024-07-31
Apply prevent sync pattern on startup of a workspace

FirelightFlagboy:


FirelightFlagboy:
Depends on #7835

libparsec/crates/client/src/workspace/store/prevent_sync_pattern.rs
```diff
@@ -0,0 +1,117 @@
...
+pub(super) async fn ensure_prevent_sync_pattern_applied_to_wksp(
+    storage: &mut WorkspaceStorage,
+    device: Arc<LocalDevice>,
+) -> Result<(), ApplyPreventSyncPatternError> {
+    const PAGE_SIZE: u32 = 1000;
+
+    let (pattern, fully_applied) = storage
+        .get_prevent_sync_pattern()
+        .await
+        .map_err(ApplyPreventSyncPatternError::Internal)?;
```
FirelightFlagboy, line 26: Should we retrieve the pattern from the DB, or add it as a parameters of the function ?

libparsec/crates/client/src/workspace/store/prevent_sync_pattern.rs
```diff
@@ -0,0 +1,117 @@
...
+            .map_err(ApplyPreventSyncPatternError::Internal)?;
+
+        let mut updated_manifest = Vec::with_capacity(manifests.len());
+        for encoded_manifest in &manifests {
+            if let Some(folder) = decode_folder(encoded_manifest, &device.local_symkey)? {
+                let new_folder =
+                    folder.apply_prevent_sync_pattern(Some(&pattern), device.time_provider.now());
+                // TODO: Is it ok to just check the updated field (DateTime) to see if the manifest has changed,
+                // or better to do a full eq ?
+                if new_folder.updated != folder.updated {
```
FirelightFlagboy, line 46: ?

libparsec/crates/client/src/workspace/store/prevent_sync_pattern.rs
```diff
@@ -0,0 +1,117 @@
...
+            } else if let Some(workspace) =
+                decode_workspace(&encoded_manifest, &device.local_symkey)?
+            {
+                let new_workspace = LocalWorkspaceManifest(
+                    workspace
+                        .apply_prevent_sync_pattern(Some(&pattern), device.time_provider.now()),
+                );
+                // TODO: Is it ok to just check the updated field (DateTime) to see if the manifest has changed,
+                // or better to do a full eq ?
+                if new_workspace.updated != workspace.updated {
```
FirelightFlagboy, line 63: ?

libparsec/crates/client/src/workspace/store/prevent_sync_pattern.rs
```diff
@@ -0,0 +1,118 @@
...
+    }
+
+    let mut offset = 0;
+    loop {
+        let manifests = storage
+            .list_manifests(offset, PAGE_SIZE)
+            .await
+            .map_err(ApplyPreventSyncPatternError::Internal)?;
+
+        let mut updated_manifest = Vec::with_capacity(manifests.len());
```
touilleMan, line 40: if you want to optimize this code you should:
- don't set a capacity to vec
- move the vec creation outside of the loop (and add a clear at the end of the loop obviously ^^)

This way we avoid doing one allocation per loop iteration, and we most likely avoid having to do a 1000 elements allocation (considering file manifests are skipped, and among folders there is few with confinement, it is likely we only have a couple of manifests to modified for each batch of 1000 elements).

libparsec/crates/client/src/workspace/store/prevent_sync_pattern.rs
touilleMan: For now you should go with a full eq (this way play safe and avoid weird bugs !)

Then, you should open an issue to indicate we should modify `apply_prevent_sync_pattern` to make it returns a `Option<Self>`  so that we know if the manifest has been actually modified

libparsec/crates/client/src/workspace/store/prevent_sync_pattern.rs
touilleMan: The pattern should be passed to this function (this way we have a single way to set the pattern and ensure it is applied)

libparsec/crates/client/src/workspace/store/prevent_sync_pattern.rs
```diff
@@ -0,0 +1,118 @@
+// Parsec Cloud (https://parsec.cloud) Copyright (c) BUSL-1.1 2016-present Scille SAS
+use std::sync::Arc;
+
+use libparsec_platform_storage::workspace::{UpdateManifestData, WorkspaceStorage};
+use libparsec_types::{
+    prelude::{anyhow, thiserror},
```
touilleMan, line 6: `use libparsec_types::prelude::*`

libparsec/crates/client/src/workspace/store/prevent_sync_pattern.rs
```diff
@@ -0,0 +1,118 @@
...
+                // or better to do a full eq ?
+                if new_folder.updated != folder.updated {
+                    updated_manifest.push(UpdateManifestData {
+                        entry_id: new_folder.base.id,
+                        encrypted: new_folder.dump_and_encrypt(&device.local_symkey),
+                        need_sync: new_folder.need_sync,
+                        base_version: new_folder.base.version,
+                    })
+                }
+            } else if let Some(workspace) =
```
touilleMan, line 55: LocalWorkspaceManifest is a sub set of LocalFolderManifest (i.e. it deserialize a LocalFolderManifest, then apply a specific sanity check to ensure parent == id)

So I think this part is useless

libparsec/crates/client/src/workspace/store/prevent_sync_pattern.rs
```diff
@@ -0,0 +1,118 @@
...
+}
+
+fn decode_folder(
+    encoded: &[u8],
+    symkey: &SecretKey,
+) -> Result<Option<LocalFolderManifest>, ApplyPreventSyncPatternError> {
+    match LocalFolderManifest::decrypt_and_load(encoded, symkey) {
+        Ok(manifest) => Ok(Some(manifest)),
+        Err(DataError::Decryption) => Err(ApplyPreventSyncPatternError::DecryptionError),
+        // We consider that BadSerialization indicate that the manifest is not of folder type.
```
touilleMan, line 101: :+1:

touilleMan (changes requested):

libparsec/crates/client/src/workspace/store/prevent_sync_pattern.rs
FirelightFlagboy: done

libparsec/crates/client/src/workspace/store/prevent_sync_pattern.rs
FirelightFlagboy: Resolved with https://github.com/Scille/parsec-cloud/pull/7843#discussion_r1700195177

libparsec/crates/client/src/workspace/store/prevent_sync_pattern.rs
FirelightFlagboy: done

libparsec/crates/client/src/workspace/store/prevent_sync_pattern.rs
FirelightFlagboy: done

libparsec/crates/client/src/workspace/store/prevent_sync_pattern.rs
FirelightFlagboy: done

libparsec/crates/client/src/workspace/store/prevent_sync_pattern.rs
FirelightFlagboy: done

libparsec/crates/client/src/workspace/store/prevent_sync_pattern.rs
```diff
@@ -0,0 +1,87 @@
...
+    Ok(())
+}
+
+fn decode_folder(
+    encoded: &[u8],
+    symkey: &SecretKey,
+) -> Result<Option<LocalFolderManifest>, ApplyPreventSyncPatternError> {
+    match LocalFolderManifest::decrypt_and_load(encoded, symkey) {
+        Ok(manifest) => Ok(Some(manifest)),
+        Err(DataError::Decryption) => Err(ApplyPreventSyncPatternError::DecryptionError),
```
touilleMan, line 82: I don't think we need a dedicated error for this

see for instance what we do here:
https://github.com/Scille/parsec-cloud/blob/b0bbab24069b6fadd38699b8eeb8ebf166282051/libparsec/crates/client/src/workspace/store/cache.rs#L219-L222

touilleMan (changes requested):

libparsec/crates/client/src/workspace/store/prevent_sync_pattern.rs
FirelightFlagboy: done

touilleMan: would be better to have an `anyhow::anyhow!("Local database contains invalid data: {}", err),` to help in the error message instead of just `err.into()`

touilleMan (approved):

PR #7849, opened 2024-07-31
Add types `RawEncrypted{Block,Chunk,Bytes}`

FirelightFlagboy:
Closes #7846

PR #7851, opened 2024-07-31
Proposal: Use `Bytes` everywhere in platform workspace storage to store encrypted data

FirelightFlagboy:


FirelightFlagboy:
Depends on #7849

touilleMan (changes requested):
See https://github.com/Scille/parsec-cloud/pull/7854#pullrequestreview-2266184485 using `Vec` is simple and fit the bill very well given the lifecycle of encrypted data in workspace storage

PR #7854, opened 2024-07-31
Proposal: Use `Cow` for raw encrypted data

FirelightFlagboy:
The goal of using `Cow` is to limit allocation from and to the data


FirelightFlagboy:
Depends on #7849

AureliaDolo (approved):
LGTM :+1: 

I'm curious, have you been able to measure the impact ?

touilleMan (changes requested):
As we discussed earlier, I'm against this change: encrypted data flow is very clear `<object>` <-> `serialized bytes` <-> `<encrypted bytes>`, hence relying on `Vec` is simpler and works very well here given the bytes have a single purpose and are never shared or anything that would require copy-on-write optimizations

PR #7865, opened 2024-08-01
Force zstd for the client build during packaging

FirelightFlagboy:


client/electron/snap/snapcraft.yaml
```diff
@@ -55,6 +55,8 @@ parts:
       - -libparsec.node
     build-attributes:
       - enable-patchelf
+    build-environment:
+      - LIBPARSEC_FORCE_ZSTD: real
```
touilleMan, line 59: This is not needed since on linux the real zstd is always used

.github/workflows/package-client.yml
```diff
@@ -292,6 +292,8 @@ jobs:
       - name: Build Electron bindings
         run: npm run build:release
         working-directory: bindings/electron
+        env:
+          LIBPARSEC_FORCE_ZSTD: real
```
touilleMan, line 296: Add a comment `# Be explicit about ZSTD implementation to avoid warning on Windows&macOS`

touilleMan (changes requested):

client/electron/snap/snapcraft.yaml
FirelightFlagboy: Better safe than sorry

.github/workflows/package-client.yml
FirelightFlagboy: done

touilleMan (approved):

PR #7870, opened 2024-08-01
Split installation of npm dependencies for electron

FirelightFlagboy:


.github/workflows/package-client.yml
```diff
@@ -280,15 +280,22 @@ jobs:
...
         timeout-minutes: 2
 
-      - name: Install dependencies
-        shell: bash
-        run: |
-          npm clean-install
-          (cd electron && npm clean-install)
-          (cd ../bindings/electron && npm clean-install)
+      - name: Install client dependencies
+        # Use `--ignore-scripts` to prevent the postinstall script trying to be smarter than us and install electron dependencies.
```
touilleMan, line 284:
```suggestion
        # Use `--ignore-scripts` to prevent the postinstall script trying to be smarter than us and
        # install electron dependencies (as it would use `npm install` instead of `npm clean-install`).
```

client/electron/snap/snapcraft.yaml
```diff
@@ -121,7 +121,8 @@ parts:
       # Patch client package.json to use local pre-packaged megashark-lib
       sed -i "s;megashark-lib\": \".*\";megashark-lib\": \"file:${CRAFT_STAGE}/megashark-lib.tar.gz\";" package.json
 
-      npm install
+      # Use `--ignore-scripts` to prevent the postinstall script trying to be smarter than us and install electron dependencies.
```
touilleMan, line 124: same as above

touilleMan (approved):

client/electron/snap/snapcraft.yaml
FirelightFlagboy:
```suggestion
      # Use `--ignore-scripts` to prevent the postinstall script trying to be smarter than us and 
      # install electron dependencies (as it would use `npm install` instead of `npm clean-install`).
```

Issue #7640, opened 2024-06-26
[🐛 | Bug]: Windows error when copying data from V2 to V3 via the File Explorer


mmmarcos:
### Parsec version tested on:

3.0.0-b.11

### Platforms tested on:

Windows

### Bug description:

Context: 
- `Scille-SIRH` organization migration from V2 to V3.
- Parsec V3 installed together with Parsec V2 on Windows (Kelly's PC) 

Steps to reproduce:
1. Login into `Scille-SIRH` for V2 and V3 for some organization.
2. Open a workspace from V2 and another one from V3 (both on Windows File Explorer)
3. Select all directories from V2 workspace and drag-and-drop to V3 workspace (both on Windows File Explorer)
    - Windows displays an error (screenshot needed)
4. Select all directories from V2 workspace and copy to V3 workspace via Ctrl+C, Ctrl+V (both on Windows File Explorer)
    - Windows displays an error (screenshot needed)
5. Open Parsec V2 GUI, enable offline availability for the workspace and wait until 100%.
6. Re-attempt 3 or 4
    - Windows displays an error (screenshot needed)

Eventually, files could be copied, so I'm not sure if it was a synchronization issue. 

We should try to reproduce to have more information.

Max-7:
![Screenshot 2024-06-26 161025](https://github.com/Scille/parsec-cloud/assets/18075498/62be3b95-f4ed-40f9-9377-3d95a636c2d2)

Error is `Invalid MS-DOS function`

FirelightFlagboy:
To enable debug log for the mountpoint, you need to run the executable with the environment `RUST_LOG=libparsec_platform_mountpoint=debug` set

mmmarcos:
I could not reproduce this issue on Linux. Tested with `v3.0.0-b.12.dev.19932+0ef95e2` installed from snap channel `v3/beta`. 

I copied 500+ files ~400MB from a V2 workspace to a V3 workspace with no error or lag. Tested on WS with available offline enabled and on another WS with available offline disabled.

PR #7873, opened 2024-08-01
WinFSP improvements

touilleMan:


libparsec/crates/client/src/workspace/store/file_updater.rs
```diff
@@ -26,7 +26,7 @@ pub(crate) enum ForUpdateFileError {
     #[error("Path doesn't exist")]
     EntryNotFound,
     #[error("Path points to entry that is not a file")]
-    EntryNotAFile,
+    EntryNotAFile { entry_id: VlobID },
```
FirelightFlagboy, line 29: The value `entry_id` is not used in the error message

libparsec/crates/client/src/workspace/transactions/open_file.rs
```diff
@@ -59,7 +59,7 @@ pub enum WorkspaceOpenFileError {
     #[error("Path doesn't exist")]
     EntryNotFound,
     #[error("Path points to entry that is not a file")]
-    EntryNotAFile,
+    EntryNotAFile { entry_id: VlobID },
```
FirelightFlagboy, line 62: The value `entry_id` is not used in the error message

libparsec/crates/platform_mountpoint/tests/unit/operations/windows_winfsp_tests.rs
```diff
@@ -63,7 +63,6 @@ async fn ok(tmp_path: TmpPath, env: &TestbedEnv) {
                         "-create_related_test",
                         // TODO: investigate misc tests
                         "-create_notraverse_test",
-                        "-delete_mmap_test",
```
FirelightFlagboy, line 66: Why we remove the `mmap` test ?

libparsec/crates/client/src/workspace/store/file_updater.rs
```diff
@@ -134,9 +134,11 @@ pub(super) async fn for_update_file(
 
     let manifest = match manifest {
         ArcLocalChildManifest::File(manifest) => manifest,
-        ArcLocalChildManifest::Folder(_) => {
+        ArcLocalChildManifest::Folder(manifest) => {
             release_guard_on_error!(update_guard);
-            return Err(ForUpdateFileError::EntryNotAFile);
+            return Err(ForUpdateFileError::EntryNotAFile {
```
FirelightFlagboy, line 139: **nitpick**: If find that we loose information about the type of the related entry id

I think it could be nice to have an additional field like `got: ManifestType`.

That would improve the error message:

`Entry id <id> is not a file but a <type>`

libparsec/crates/platform_mountpoint/src/windows/filesystem.rs
```diff
@@ -617,20 +678,36 @@ impl FileSystemContext for ParsecFileSystemContext {
...
+            last_write_time,
+            change_time,
+        );
+
+        // Note this method actually does nothing !
+        // But if we don't define it, `explorer.exe` displays "Invalid MS-DOS function" errors
+        // when copying some directory (see https://github.com/Scille/parsec-cloud/issues/7640).
+
+        self.tokio_handle
+            .block_on(async move { self.get_file_info_async(&fc).await })
```
FirelightFlagboy, line 707: Given that, this PR close #7640 ?

libparsec/crates/platform_mountpoint/src/windows/filesystem.rs
```diff
@@ -49,20 +49,51 @@ fn is_entry_info_path(path: &FsPath) -> bool {
...
+        // FILE_ATTRIBUTE_NORMAL which means "no other attributes is set"
+        // Also, this is what the winfsp memfs does.
+        .set_file_attributes(FileAttributes::ARCHIVE | FileAttributes::NOT_CONTENT_INDEXED)
+        .set_creation_time(created)
+        .set_last_access_time(updated)
+        .set_last_write_time(updated)
+        .set_change_time(updated)
+        // TODO: We truncate the EntryID to 64bits as index number, given EntryID is a UUIDv4 this
+        // increases the risk of collision... We should investigate to see if this is really needed.
+        .set_index_number(stat.id.as_u128() as u64)
```
FirelightFlagboy, line 92: Should you be more explicit on the truncate between little/big-endian ?

libparsec/crates/platform_mountpoint/tests/unit/operations/windows_winfsp_tests.rs
touilleMan: Because this test is now passing \o/

(the `-delete_mmap_test` should be read here "disable test `delete_mmap_test`")

FirelightFlagboy: T_T

libparsec/crates/platform_mountpoint/src/windows/filesystem.rs
touilleMan: YEEEEEEEEES \o/

FirelightFlagboy: I've linked the issue then

libparsec/crates/client/src/workspace/store/file_updater.rs
touilleMan: done. (I've also apply this to other errors with similar `entry_id` :+1: )

libparsec/crates/client/src/workspace/store/file_updater.rs
touilleMan: Currently there is only two types of entry: file and folder, so this doesn't bring much information ^^

I think the important thing here is to keep in mind the primary goal of this error (i.e. "the operation failed because the entry wasn't a file").
This is for this reason I avoided naming this error `EntryISAFolder`, which would convey the same information (since there is only file or folder), but is less clear given here we actually care about file, not folder.
It is always tempting to add more information to an error (especially if there are easily available in the function creating the error !), but I think it makes things more complicated for no reason.

However the current situation is interesting:
- winfsp needs a way to open a file or a folder
- but we can only open files in libparsec API, hence "opening" a folder only consist of retrieving its entry ID
- opening the file or getting the entry ID if it's a folder should be done in an atomic way, otherwise an ugly loop is needed to handle the case the entry is first a folder when we try to open it as a file, then becomes a file consider it as folder and try to get it entry ID :/

So in this situation, going against the rule I just stated is very appealing since it makes thing a bit less coherent in theory, but a lot better in practice !
This is the kind of trade-of that can only be judged on the spot, hence why I don't think we should add more information than needed.

libparsec/crates/platform_mountpoint/src/windows/filesystem.rs
touilleMan: I have added a comment :+1:

FirelightFlagboy (approved):

PR #7896, opened 2024-08-02
 Add `prevent_sync_pattern` to the client configuration

### LINKED ISSUE #7707: Make the pattern list configurable
Add the `prevent_sync_pattern` path to `config.rs`:

https://github.com/Scille/parsec-cloud/blob/fc76ec52b7ac92370d3ed24cce3cbd14c50b5e9b/libparsec/crates/client/src/config.rs#L45

### LINKED ISSUE #7828: Use default prevent sync pattern when no configuration is set up.
The simplest way would be to include the `libparsec/crates/client/src/workspace/default_pattern.ignore` using `include_str` and use that value as fallback in case no file is set to read the pattern from

FirelightFlagboy:
Closes #7828, closes #7707


libparsec/crates/client/src/config.rs
```diff
@@ -33,6 +33,9 @@ pub enum MountpointMountStrategy {
     Disabled,
 }
 
+/// This is the internal configuration of the client.
+/// Currently, this configuration is "complete" meaning all of his value are set.
```
AureliaDolo, line 37:
```suggestion
/// Currently, this configuration is "complete" meaning all of its value are set.
```

libparsec/src/config.rs
```diff
@@ -11,13 +11,18 @@ pub use libparsec_platform_device_loader::{
 };
 pub use libparsec_types::prelude::*;
 
+/// Partial client configuration.
+/// This configuration should be configured by the user before converting it to
+/// the proper configuration [`libparsec_client::ClientConfig`] by using default values as fallback.
 #[derive(Debug, Clone)]
 pub struct ClientConfig {
```
AureliaDolo, line 18: A note for later: could we rename this to something like `PartialClientConfig` or `RawClientConfig` to make things more explicit ?

FirelightFlagboy: I've created the issue #7897 for that, that would be nice to rename it :+1:

AureliaDolo (approved):

libparsec/src/config.rs
```diff
@@ -38,6 +44,8 @@ impl From<ClientConfig> for Arc<libparsec_client::ClientConfig> {
     }
 }
 
+const DEFAULT_PREVENT_SYNC_PATTERN: &str = std::include_str!("default_pattern.ignore");
```
touilleMan, line 47: I'd rather avoid having the default prevent sync pattern in the `libparsec_client` crate.

The issue is `libparsec_client` is high level stuff without tests (it is expected that GUI + CLI relying on it are enough to detect the issues)

On the other hand, the default prevent sync pattern is most likely something we want to test !

So I think what we should do:
- move the default prevent sync pattern into the `libparsec_types` crate
- Add a `Regex::default()` method that build the regex from the default prevent sync pattern

This way it is simple to add tests in `libparsec_types` checking the behavior of the default prevent sync pattern (and compare it to the empty and custom patterns ^^), and building `ClientConfig` is trivial:
```rust
            prevent_sync_pattern: match config.prevent_sync_pattern_file.unwrap_or_default()
```

(this works this way given the bindings code should introduce a new `Regex` type which is `StrBasedType` and hence does the string validation in the javascript-to-rust conversion step)

touilleMan (changes requested):

libparsec/src/config.rs
```diff
@@ -47,6 +54,14 @@ impl From<ClientConfig> for libparsec_client::ClientConfig {
             workspace_storage_cache_size: config.workspace_storage_cache_size,
             proxy: ProxyConfig::default(),
             with_monitors: config.with_monitors,
+            prevent_sync_pattern: match config.prevent_sync_pattern {
+                Some(pattern) => Regex::from_glob_reader(
+                    "client_prevent_sync_pattern",
+                    std::io::Cursor::new(pattern),
+                )
+                .expect("Cannot process provided prevent sync pattern file"),
```
touilleMan, line 62: `Regex` type should be exposed at bindings level, so providing an invalid `prevent_sync_pattern` cause a conversion error that javascript can deal with instead of having a panic)

See for instance `ParsecAddr` on how to implement the conversion code:
https://github.com/Scille/parsec-cloud/blob/9926f07483e6a7777a5aa87528d3a1957a282b80/bindings/generator/api/addr.py#L19-L23

FirelightFlagboy: Will be done in another PR, an issue was created: https://github.com/Scille/parsec-cloud/issues/7923

touilleMan (approved):

PR #7909, opened 2024-08-05
Change prevent sync pattern type to be mandatory

FirelightFlagboy:
That allow to detected placed where the pattern is not passed when it should be. I've placed TODO comments where the pattern should be passed as it will be fixed in another PR.


touilleMan (approved):
:+1:

PR #7910, opened 2024-08-05
Pass prevent sync pattern used by the workspace store

FirelightFlagboy:


libparsec/crates/client/src/workspace/store/mod.rs
```diff
@@ -117,6 +117,7 @@ pub(super) struct WorkspaceStore {
     /// Given accessing `storage` requires exclusive access, it is better to have it
     /// under its own lock so that all cache hit operations can occur concurrently.
     storage: AsyncMutex<Option<WorkspaceStorage>>,
+    prevent_sync_pattern: Regex,
```
touilleMan, line 120: The prevent sync pattern should be part of `ClientConfig` instead

(this is convenient given the config is already part of `WorkspaceOps`)

touilleMan (changes requested):

libparsec/crates/client/src/workspace/store/mod.rs
FirelightFlagboy: > The prevent sync pattern should be part of `ClientConfig` instead

It's already the case 🤨

> this is convenient given the config is already part of WorkspaceOps

Yes

> there is no reason to have the prevent sync pattern in the workspace store since it is not related to what is stored in the database

I mean ... we store the pattern in the database

The other reason is the pattern is required at multiple places inside the store module where it don't have access the the `WorkspaceOps`.

If you want to not have the pattern stored in the store, that would require to pass the pattern as argument in a lot of places.

touilleMan: > I mean ... we store the pattern in the database

Yes, but we never use the database for the need of retrieving the pattern. So from an outside view, `WorkspaceStore` consumes the pattern but never produces it.

> The other reason is the pattern is required at multiple places inside the store module where it don't have access the the WorkspaceOps.

In this case we can either:
1. Give the whole config object to `WorkspaceOps` 
2. Keep having a `prevent_sync_pattern` field in `WorkspaceOps`, but keeping it private since it is only for internal use (so for instance outbount/inbound sync would use `ops.config.prevent_sync_pattern` instead of `obs.store.get_prevent_sync_pattern()`)

Regarding which solution to use, I would say we should go with solution 2 given:
- we only use a single field in the config
- It is almost already implemented (should just turn `WorkspaceOps::prevent_sync_pattern` private)
- It avoids having to change code in the tests to pass a config (which on top of that is a more complicated structure to build than a simple `Regex`)

libparsec/crates/client/src/workspace/store/mod.rs
```diff
@@ -191,9 +192,14 @@ impl WorkspaceStore {
             certificates_ops,
             current_view_cache: Mutex::new(CurrentViewCache::new(Arc::new(root_manifest.into()))),
             storage: AsyncMutex::new(Some(storage)),
+            prevent_sync_pattern: pattern.clone(),
         })
     }
 
+    fn prevent_sync_pattern(&self) -> &Regex {
```
touilleMan, line 199: You can remove the getter and go directly with `self.prevent_sync_pattern` since this is private stuff

touilleMan (approved):

libparsec/crates/client/src/workspace/store/mod.rs
FirelightFlagboy: done

PR #7926, opened 2024-08-06
Use macro to define share option over multiple struct

FirelightFlagboy:
Previously we used the structs `*SharedOpts` to share the same configuration for multiple commands. But I don't find that flexible enough, so I created a macro to define the shared options over multiple structs.

The advantage of the struct is already flattened

### DIFF
--- a/cli/src/macro_opts.rs
+++ b/cli/src/macro_opts.rs
@@ -1,0 +1,155 @@
+/// This macros builds a Clap argument parser by combining the provided structure
+/// with common CLI options.
+///
+/// Available common CLI options:
+/// - `config_dir` (add `config_dir: std::path::PathBuf`)
+/// - `device` (add `device: Option<String>`)
+/// - `addr` (add `addr: libparsec::ParsecAddr`)
+/// - `token` (add `token: String`)
+///
+/// Example:
+///
+/// ```rust
+/// crate::clap_parser_with_shared_opts_builder!(
+///     #[with = config_dir]
+///     pub struct ListUsers {
+///         #[arg(short, long, default_value_t)]
+///         skip_revoked: bool,
+///     }
+/// );
+/// ```
+///
+/// Becomes:
+///
+/// ```rust no_run
+/// #[derive(clap::Parser)]
+/// pub struct ListUsers {
+///     #[arg(short, long, default_value_t)]
+///     skip_revoked: bool,
+///     #[doc = "Parsec config directory"]
+///     #[arg(short, long, default_value_os_t = libparsec::get_default_config_dir(), env = $crate::utils::PARSEC_CONFIG_DIR)]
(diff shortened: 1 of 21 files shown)


cli/src/macro_opts.rs
```diff
@@ -0,0 +1,122 @@
+#[macro_export]
+macro_rules! shared_opts_wrapper {
+    // Config dir option
+    (
+        #[with = config_dir $(,$modifier:ident)*]
```
touilleMan, line 5: I find the `#[with = a, b]` a bit unclear given it reads like `with = a` + `b` instead of `with = (a, b)`

would it be possible to replace the coma by a plus sign: `#[with = a + b]` ?

But don't bother too much if it's not a trivial change (or if you don't find it more readable), as this is a minor readability improvement ^^

cli/src/macro_opts.rs
```diff
@@ -0,0 +1,122 @@
+#[macro_export]
+macro_rules! shared_opts_wrapper {
```
touilleMan, line 2: Given this struct applies the `#[derive(clap::Parsec)]` to the struct provided as input, I would call this macro `clap_parser_builder` (or `clap_parser_with_shared_opts_builder` if we want to be extra explicit)

On top of that, adding a small example as comment would help a lot understanding this macro
```rust
/// This macros builds a Clap argument parser by combining the provided structure
/// with common CLI options.
///
/// Available common CLI options:
/// - `config_dir` (add `config_dir: std::path::PathBuf`)
/// - `device` (add `device: Option<String>`)
/// - `addr` (add `addr: libparsec::ParsecAddr`)
/// - `token` (add `token: String`)
///
/// Example:
///
/// ```rust
/// crate::shared_opts_wrapper!(
///     #[with = config_dir]
///     pub struct ListUsers {
///         #[arg(short, long, default_value_t)]
///         skip_revoked: bool,
///     }
/// );
/// ```
///
/// Becomes:
///
/// ```rust
/// #[derive(clap::Parser)]
/// pub struct ListUsers {
///     #[arg(short, long, default_value_t)]
///     skip_revoked: bool,
///     #[doc = "Parsec config directory"]
///     #[arg(short, long, default_value_os_t = libparsec::get_default_config_dir(), env = $crate::utils::PARSEC_CONFIG_DIR)]
///     pub(crate) config_dir: std::path::PathBuf,
/// }
/// ```
```

touilleMan (approved):
This macros is really cool :smile:

cli/src/macro_opts.rs
FirelightFlagboy: done

cli/src/macro_opts.rs
FirelightFlagboy: Currently, that not possible rust-lang/rust#18700

PR #7929, opened 2024-08-06
Remove callback in `load_device_file`

FirelightFlagboy:
Also improve error message when multiple devices are found for a given short ID by listing the matched devices over the whole list of devices.

### DIFF
--- a/cli/src/utils.rs
+++ b/cli/src/utils.rs
@@ -1,10 +1,6 @@
 // Parsec Cloud (https://parsec.cloud) Copyright (c) BUSL-1.1 2016-present Scille SAS
 
-use std::{
-    future::Future,
-    path::{Path, PathBuf},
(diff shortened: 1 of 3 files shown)


cli/src/utils.rs
```diff
@@ -54,8 +55,11 @@ pub const RED: &str = "\x1B[91m";
 pub const RESET: &str = "\x1B[39m";
 pub const YELLOW: &str = "\x1B[33m";
 
-pub fn format_devices(devices: &[AvailableDevice]) {
-    let n = devices.len();
+pub fn format_devices(
```
touilleMan, line 58: Good thing to have the `format_` function only doing formatting :+1:

touilleMan (approved):
Wow this change makes the code much more readable ! :1st_place_medal:

PR #7930, opened 2024-08-06
[MS] Client area: Improve user feedback on personal data update

NicoTuxx:
- inject informationManager in ClientAreaPage and pass it as prop to sub views
 - fix: add translationPrefixes to reduce mistakes
 - feat: add toast message on modals success

depends on #7947

### DIFF
--- a/client/src/views/client-area/personal-data/PersonalInfoModal.vue
+++ b/client/src/views/client-area/personal-data/PersonalInfoModal.vue
@@ -3,10 +3,10 @@
 <template>
   <ion-page class="modal">
     <ms-modal
-      title="clientArea.personalDataPage.modals.personalInfo.title"
+      :title="`${translationPrefix}.title`"
       :close-button="{ visible: true }"
       :confirm-button="{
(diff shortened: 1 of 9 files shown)


client/src/views/client-area/personal-data/PersonalInfoModal.vue
```diff
@@ -94,7 +95,7 @@ async function submit(): Promise<boolean> {
     errors.value = response.errors ?? [];
     return false;
   }
-  return await modalController.dismiss({}, MsModalResult.Confirm);
+  return await modalController.dismiss(`${translationPrefix}.success`, MsModalResult.Confirm);
```
Max-7, line 98: The modal should return a status, and this status should be used by the calling component to show a message, it should not return the message directly.

Max-7 (changes requested):

client/src/views/client-area/personal-data/PersonalInfoModal.vue
NicoTuxx: done

Max-7 (approved):

PR #7937, opened 2024-08-07
Add command to copy a local file to a remote workspace

### LINKED ISSUE #7936: Allow to deposit a file using the CLI

FirelightFlagboy:
Closes #7936

### DIFF
--- a/cli/src/commands/mod.rs
+++ b/cli/src/commands/mod.rs
@@ -21,3 +21,4 @@ pub mod share_workspace;
 pub mod stats_organization;
 pub mod stats_server;
 pub mod status_organization;
+pub mod workspace_import;
--- a/cli/src/commands/workspace_import.rs
+++ b/cli/src/commands/workspace_import.rs
@@ -1,0 +1,95 @@
+use std::{path::PathBuf, vec};
+
+use libparsec::{anyhow::Context, FsPath, OpenOptions, VlobID};
+use tokio::io::AsyncReadExt;
+
+use crate::utils::load_client;
+
+crate::clap_parser_with_shared_opts_builder!(
+    #[with = config_dir, device, password_stdin]
+    pub struct WorkspaceImport {
+        /// Workspace id
+        #[arg(short, long, value_parser = VlobID::from_hex)]
+        workspace_id: VlobID,
+        /// Local file to copy
+        src: PathBuf,
+        /// Workspace destination path
+        dest: FsPath,
+    }
+);
+
+pub async fn workspace_import(args: WorkspaceImport) -> anyhow::Result<()> {
+    let WorkspaceImport {
+        src,
+        dest,
+        workspace_id,
+        password_stdin,
+        device,
+        config_dir,
+    } = args;
+
+    log::trace!(
+        "workspace_import: {src} -> {workspace_id}:{dst}",
+        src = src.display(),
+        dst = dest
+    );
+
+    let client = load_client(&config_dir, device, password_stdin).await?;
+    let workspace = client.start_workspace(workspace_id).await?;
+    let fd = workspace
+        .open_file(
+            dest,
+            OpenOptions {
+                read: false,
+                write: true,
+                truncate: true,
+                create: true,
+                create_new: false,
+            },
+        )
+        .await?;
+
+    let file = tokio::fs::File::open(&src)
+        .await
+        .context("Cannot open local file")?;
+    let mut buf_file = tokio::io::BufReader::new(file);
+    let mut buffer = vec![0_u8; 4096];
+    let mut dst_offset = 0_usize;
+
+    log::debug!("Copying file to workspace");
+    loop {
+        let bytes_read = buf_file
+            .read(&mut buffer)
+            .await
+            .context("Cannot read local file")?;
+
+        if bytes_read == 0 {
+            break;
+        }
+
+        let mut bytes_written = 0_usize;
+        while bytes_written < bytes_read {
+            bytes_written = workspace
+                .fd_write(fd, dst_offset as u64, &buffer[bytes_written..bytes_read])
+                .await
+                .context("Cannot write to workspace")? as usize;
+            dst_offset += bytes_written;
+        }
+    }
+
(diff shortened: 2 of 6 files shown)


cli/src/copy_to.rs
```diff
@@ -0,0 +1,78 @@
...
+        #[arg(short, long, value_parser = VlobID::from_hex)]
+        workspace_id: VlobID,
+        /// Local file to copy
+        src: PathBuf,
+        /// Workspace destination path
+        dest: FsPath,
+    }
+);
+
+pub async fn copy_to(args: CopyTo) -> anyhow::Result<()> {
```
touilleMan, line 21: I find the naming too misleading here, `copy-to` let me feel I can use this command:
-  to copy file from within the workspace (`cp wksp1/foo.txt wksp1/bar.txt`)
- if this command can import data, I would expect to also be able to export data 

so I propose the following naming:
- `parsec workspace import --workspace 42 --source <path on the OS> <path in the workspace>` (copy file or folder in the workspace)
- `parsec workspace import --workspace 42 --source-stdin <path in the workspace>` (copy data from stdin in the workspace)
- `parsec workspace export --workspace 42 <path in the workspace> <path on the OS>`
- `parsec workspace cp --workspace 42 <path in the workspace> <path in the workspace>`
- `parsec workspace rm --workspace 42 <path in the workspace> <path in the workspace>`
- `parsec workspace touch --workspace 42 <path in the workspace>`
- `parsec workspace mkdir --workspace 42 <path in the workspace>`
- `parsec workspace ls --workspace 42 <path in the workspace>`

(or `workspace-xxx` instead of `workspace xxx` if it's too complicated to create sub-groups with our current CLI design)

touilleMan (changes requested):

cli/src/tests.rs
```diff
@@ -973,6 +973,73 @@ async fn setup_shamir(tmp_path: TmpPath) {
...
+#[tokio::test]
+async fn workspace_import_file(tmp_path: TmpPath) {
+    let _ = env_logger::builder().is_test(true).try_init();
+    let tmp_path_str = tmp_path.to_str().unwrap();
+    let config = get_testenv_config();
+    let (url, [alice, ..], _) = run_local_organization(&tmp_path, None, config)
+        .await
+        .unwrap();
+
+    set_env(&tmp_path_str, &url);
```
touilleMan, line 986: This function internally uses `std::env::set_var` so it sets environ variable for the whole process, this will cause random issue when multiple tests are run concurrently in the same process (so if we use `cargo test` instead of `cargo nextest`)

Instead it would be better to pass the environ variable to the `Command::cargo_bin` command (so that only this subprocess has those variable configured)

This is outside the scope of this PR (since `set_env` is used in each test), so can you create an issue about this ?

cli/src/tests.rs
```diff
@@ -973,6 +973,73 @@ async fn setup_shamir(tmp_path: TmpPath) {
...
+    let workspace = client.start_workspace(wid).await.unwrap();
+    let entries = workspace
+        .stat_folder_children(&"/".parse().unwrap())
+        .await
+        .unwrap();
+
+    assert_eq!(entries.len(), 1);
+    let (name, stat) = &entries[0];
+    assert_eq!(name.as_ref(), "test.txt");
+    assert!(matches!(stat, libparsec::EntryStat::File { size, .. } if size == &13));
```
touilleMan, line 1040: you should also read the content of the file to ensure it is correct

cli/src/tests.rs
```diff
@@ -973,6 +973,73 @@ async fn setup_shamir(tmp_path: TmpPath) {
...
+    // Create the workspace used to copy the file to
+    let wid = client
+        .create_workspace("new-workspace".parse().unwrap())
+        .await
+        .unwrap();
+    client.ensure_workspaces_bootstrapped().await.unwrap();
+
+    // Create a file to copy
+    let file = tmp_path.join("test.txt");
+    std::fs::write(&file, "Hello, World!").unwrap();
```
touilleMan, line 1011: It would be better to have a file that is bigger than the buffer size internally used (so > 4096), and not aligned on it, so typically a 10ko file:
```
let buff = "1234567890".repeat(1000);
assert_eq!(buff.len(), 10000);
```

cli/src/workspace_import.rs
```diff
@@ -0,0 +1,78 @@
...
+        while bytes_written < bytes_read {
+            bytes_written = workspace
+                .fd_write(fd, dst_offset as u64, &buffer[bytes_written..bytes_read])
+                .await? as usize;
+            dst_offset += bytes_written;
+        }
+    }
+
+    log::debug!("Flushing and closing file");
+    workspace.fd_flush(fd).await?;
```
touilleMan, line 75: flush is automatically done on close, so this line can be removed

cli/src/workspace_import.rs
```diff
@@ -0,0 +1,78 @@
...
+        let bytes_read = buf_file.read(&mut buffer).await?;
+
+        if bytes_read == 0 {
+            break;
+        }
+
+        let mut bytes_written = 0_usize;
+        while bytes_written < bytes_read {
+            bytes_written = workspace
+                .fd_write(fd, dst_offset as u64, &buffer[bytes_written..bytes_read])
```
touilleMan, line 68: `fd_write` always write the whole buffer given to it

I realize returning the size of data written is confusing here, so I think we should just return an empty tuple instead (I've create an issue about this)

touilleMan (approved):

cli/src/tests.rs
FirelightFlagboy: https://github.com/Scille/parsec-cloud/issues/8079

cli/src/workspace_import.rs
FirelightFlagboy: If that the case, I propose to rename `fd_write` to `fd_write_all` to be similar to [`Write::write_all`](https://doc.rust-lang.org/std/io/trait.Write.html#method.write_all)

touilleMan (approved):

PR #7951, opened 2024-08-08
Re-organization parsec-cli

FirelightFlagboy:
- Move sub-command to cli.
- Create `testenv_utils` module to handle test environment setup (shared with `run_testenv` sub-command).


cli/src/testenv_utils.rs
```diff
@@ -1,5 +1,3 @@
-// Parsec Cloud (https://parsec.cloud) Copyright (c) BUSL-1.1 2016-present Scille SAS
```
AureliaDolo, line 1: Shouldn't have the CI caught that ?

AureliaDolo (approved):

cli/src/testenv_utils.rs
FirelightFlagboy: The CI as run the license check:

```
 license_headers..........................................................Passed
- hook id: license_headers
- duration: 0.08s
```

But something seems wrong with the script :/

FirelightFlagboy: Waiting for #8060 to run check

PR #7954, opened 2024-08-08
Split cli test file

FirelightFlagboy:


cli/Cargo.toml
```diff
@@ -7,8 +7,7 @@ name = "parsec_cli"
 publish.workspace = true
 repository.workspace = true
 version.workspace = true
-
-# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
+autotests = false
```
AureliaDolo, line 10: I'm not well versed into that, but should we specify the test target somewhere in this file for tests to run ?

FirelightFlagboy: We don't need to specify a test file in `Cargo.toml` because we already do that in `main.rs`:

```rust
#[cfg(test)]
#[path = "../tests/integration/mod.rs"]
mod tests;
```

> To be pedantic, the above define a unit test, but we need the internal of `parsec_cli` to perform the integration test :sweat_smile:

AureliaDolo (approved):

cli/Cargo.toml
AureliaDolo: Thanks !

PR #7960, opened 2024-08-09
Add macro to reduce the boilerplate to run CLI commands in tests

FirelightFlagboy:


cli/tests/integration/mod.rs
```diff
@@ -97,3 +97,23 @@ async fn bootstrap_cli_test(
...
+
+#[macro_export]
+macro_rules! assert_cmd_success {
+    (with_password=$pass:expr, $($cmd:expr),+) => {
+        $crate::assert_cmd_success!(__internal__ $($cmd),+ , "--password-stdin")
+            .write_stdin(format!("{password}\n", password = $pass))
+            .assert()
+            .success()
+    };
+    (__internal__ $($cmd:expr),+) => {
```
AureliaDolo, line 109: What does the `__internal__` mean ?

AureliaDolo (approved):

cli/tests/integration/mod.rs
FirelightFlagboy: I use it to define the internal part of the macro, look how it's shared between the `with_password` and the other case

AureliaDolo: Thanks !

PR #7971, opened 2024-08-09
Re-enable test `share_workspace`

FirelightFlagboy:
The issue was we where using `other_alice` device instead of `bob` device.


AureliaDolo (approved):
:sparkles:

PR #7992, opened 2024-08-13
Add test where multiple devices are found for the same short hex id

FirelightFlagboy:


cli/tests/integration/device_option.rs
```diff
@@ -93,3 +93,148 @@ async fn missing_option(tmp_path: TmpPath) {
...
+
+    let mut devices = [first_device, second_device];
+    devices.sort_by_cached_key(|d| d.device_id.hex());
+
+    let mut available_devices_string_list = String::new();
+    format_device::<3>(&devices, &mut available_devices_string_list).unwrap();
+
+    crate::assert_cmd_failure!("list-users", "--device", first_dev_id_prefix).stderr(
+        predicates::str::contains(format!(
+            "Error: Multiple devices found for `{first_dev_id_prefix}`:"
```
touilleMan, line 170: Note having two devices begining with the same characters is possible and shouldn't be an issue.

In Parsec v2 we would start with a 3 characters short hex id, then make it bigger whenever the number of characters become too small to uniquely identify each device.

So I think we should open an issue about this

touilleMan (approved):

cli/tests/integration/device_option.rs
FirelightFlagboy: I've openned an issue: #8034

PR #7997, opened 2024-08-13
Return enum error instead of str for HumanHandle parsing

### LINKED ISSUE #7993: Make `HumanHandle::new` return a proper error enum instead of raw string

FirelightFlagboy:
Close #7993

### DIFF
--- a/libparsec/crates/types/src/id.rs
+++ b/libparsec/crates/types/src/id.rs
@@ -520,19 +520,35 @@ impl AsRef<str> for HumanHandle {
     }
 }
 
+#[derive(Debug, thiserror::Error)]
+pub enum HumanHandleParseError {
+    #[error("Email is missing")]
+    MissingEmail,
+    #[error("Invalid email address")]
+    InvalidEmail,
+    #[error("Invalid label")]
+    InvalidLabel,
+}
+
 impl TryFrom<&str> for HumanHandle {
-    type Error = &'static str;
+    type Error = HumanHandleParseError;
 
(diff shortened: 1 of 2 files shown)


libparsec/crates/types/src/id.rs
```diff
@@ -625,10 +641,10 @@ impl HumanHandle {
 }
 
 impl TryFrom<(&str, &str)> for HumanHandle {
-    type Error = &'static str;
+    type Error = HumanHandleParseError;
 
-    fn try_from(id: (&str, &str)) -> Result<Self, Self::Error> {
-        Self::new(id.0, id.1)
+    fn try_from((email, label): (&str, &str)) -> Result<Self, Self::Error> {
```
touilleMan, line 646: :+1:

libparsec/crates/types/src/id.rs
```diff
@@ -520,19 +520,35 @@ impl AsRef<str> for HumanHandle {
     }
 }
 
+#[derive(Debug, thiserror::Error)]
+pub enum HumanHandleParseError {
+    #[error("Email is missing")]
+    MissingEmail,
+    #[error("Invalid email address '{}'", .0)]
+    InvalidEmail(String),
```
touilleMan, line 528: I'd rather avoid storing the bad email/label here as this prevent us from leaking sensitive data in logs & sentry reports.

touilleMan (changes requested):

libparsec/crates/types/src/id.rs

PR #8019, opened 2024-08-13
Fix ignored tests in LocalFolderManifest on prevent sync pattern

touilleMan:


vxgmichel:
Already addressed in https://github.com/Scille/parsec-cloud/pull/8682

PR #8020, opened 2024-08-13
Add error type for parsing `UserProfile`

### LINKED ISSUE #8000: Add error type for `UserProfile` parsing
Currently `types::UserProfile::from_str` return a raw string

https://github.com/Scille/parsec-cloud/blob/40d92a3e49fd5c7e95678a979ad5fd77c9f0c9cb/libparsec/crates/types/src/id.rs#L665-L677

Create a proper error type

FirelightFlagboy:
Closes #8000

### DIFF
--- a/libparsec/crates/types/src/id.rs
+++ b/libparsec/crates/types/src/id.rs
@@ -675,15 +675,26 @@ pub enum UserProfile {
     Outsider,
 }
 
+#[derive(Debug, Clone, Copy)]
+pub struct InvalidUserProfile;
+
+impl std::error::Error for InvalidUserProfile {}
+
+impl std::fmt::Display for InvalidUserProfile {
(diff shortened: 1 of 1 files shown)


libparsec/crates/types/src/id.rs
```diff
@@ -675,15 +675,26 @@ pub enum UserProfile {
     Outsider,
 }
 
+#[derive(Debug, Clone)]
+pub struct InvalidUserProfile(String);
```
touilleMan, line 679:
```suggestion
pub struct InvalidUserProfile;
```

No need to display the invalid data

touilleMan (changes requested):

libparsec/crates/types/src/id.rs

touilleMan (approved):

PR #8021, opened 2024-08-13
Add error type for parsing `InvitationType`

### LINKED ISSUE #8001: Add error type for `InvitationType` parsing
Currently `types::InvitationType::from_str` return a raw string

https://github.com/Scille/parsec-cloud/blob/40d92a3e49fd5c7e95678a979ad5fd77c9f0c9cb/libparsec/crates/types/src/invite.rs#L29-L40

Create a proper error type

FirelightFlagboy:
Closes #8001

### DIFF
--- a/libparsec/crates/types/src/invite.rs
+++ b/libparsec/crates/types/src/invite.rs
@@ -26,14 +26,25 @@ pub enum InvitationType {
     Device,
 }
 
+#[derive(Debug, Clone)]
+pub struct InvitationTypeParseError;
+
+impl std::error::Error for InvitationTypeParseError {}
+
+impl std::fmt::Display for InvitationTypeParseError {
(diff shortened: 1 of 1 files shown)


libparsec/crates/types/src/invite.rs
```diff
@@ -26,14 +26,25 @@ pub enum InvitationType {
     Device,
 }
 
+#[derive(Debug, Clone)]
+pub struct InvitationTypeParseError(String);
```
touilleMan, line 30:
```suggestion
pub struct InvitationTypeParseError;
```

No need to display the invalid data

touilleMan (changes requested):

libparsec/crates/types/src/invite.rs
FirelightFlagboy: done

touilleMan (approved):

PR #8028, opened 2024-08-14
Document server architecture and causility/concurrency strategy

touilleMan:


server/README.md
```diff
@@ -0,0 +1,236 @@
...
+- `common`
+- `sequester`, which depends on `common`
+- `realm`, which depends on `common` and uses the realm ID as discriminant
+- `shamir_recovery`, which depends on `common` and use the ID of the user to recover as discriminant
+
+So the actual rule used in the server is:
+
+- Have lock in the server for each couple topic + discriminant
+- Take the lock for the topic + discriminant couple the certificate we want to add
+  belongs to AND to other topic our topic may depend on (handles concurrency).
```
FirelightFlagboy, line 124:
```suggestion
  belongs to AND to other topics our topic may depend on (handles concurrency).
```

FirelightFlagboy, line 124: If I understand correctly, currently all topics depends on `common`, that mean each topic need to lock `common`.

That feels like a _single big lock_, I'm missing something ?

touilleMan: The difference is the `common` lock is taken for read only, and another lock is taken for write.

So for instance we can have concurrent add of certificates in two different realms (both locking the `common` lock for read).

On the other hand, in a big lock architecture the lock would be always taken for write.

FirelightFlagboy: When the `common` lock is taken in write mode /

server/README.md
```diff
@@ -0,0 +1,236 @@
...
+Blocks are encrypted piece of data that belong to a given realm and cannot
+be modified.
+
+This makes concurrency & causality pretty simple:
+
+- Each block is considered independent, hence there is no need for lock dedicated to blocks.
+- A block is identified by a client-provided UUID, then there is no risk for double-creation
+  under concurrency (a trivial unicity check on insertion prevents this).
+- Block creation date must be more recent than the last certificate in `common` topic and
+  in the `realm` topic corresponding to the block's realm.
```
vxgmichel, line 142: This causality check is only treated one way (checking the topic when creating a block). The opposite is not mentioned in the above section.

vxgmichel: Also this section does not mention locking: is the `common` topic lock taken why the block is uploaded?

server/README.md
```diff
@@ -0,0 +1,236 @@
...
+
+> Notes:
+>
+> - It could be tempting to only lock the `realm` topic (given it's this topic that contains
+>   the certificate describing the role of a user in the realm). However a revoked user is also
+>   forbidden from adding data, hence making it mandatory to also lock the `common` topic here.
+> - Uploading blocks (and vlobs) are the most common task on the server, so we may want to
+>   uptomise it in the future to avoid having to lock two topics.
```
vxgmichel, line 150: That sounds quite bad. The proposed strategy should already address this problem IMO.

server/README.md
```diff
@@ -0,0 +1,236 @@
...
+>   A possible solution would be to have an optimistic locking approach: first read the
+>   last timestamp in the topics, then do the checks, then on the final data insertion check
+>   again the timestamp haven't changed (and if they have we re-do everything from the start).
+
+### Vlobs
+
+Vlobs are encrypted metadata that belong to a given realm. Unlike block they can be
+modified and hence have a version field that must increase linearly.
+
+Concurrency & causality is handle is a similar fashion than for blocks.
```
vxgmichel, line 160: Really? Vlobs require a database write, but blocks require bother a database write and upload to the object storage. I would find very surprising if both used the same approach in the end.

server/README.md
```diff
@@ -0,0 +1,236 @@
...
+  up to the server to check for causility to know if it should accept it.
+- Here the client merely ask the server do to something, then the server check it if it
+  possible *now* then do it *now*. No causility can be broken given we only work in the present.
+
+Regarding concurrency, a lock is needed to protect the creation of invitation
+(see [example in Data concurrency paragraph](#data-concurrency)).
+
+### Implementation: PostgreSQL
+
+In PostgreSQL we have multple tables (`common_topic`, `realm_topic` etc.) dediticated to
```
vxgmichel, line 186:
```suggestion
In PostgreSQL we have multiple tables (`common_topic`, `realm_topic` etc.) dedicated to
```
Many typos haven't been caught by cspell

server/README.md
```diff
@@ -0,0 +1,236 @@
...
+- `realm`, which depends on `common` and uses the realm ID as discriminant
+- `shamir_recovery`, which depends on `common` and use the ID of the user to recover as discriminant
+
+So the actual rule used in the server is:
+
+- Have lock in the server for each couple topic + discriminant
+- Take the lock for the topic + discriminant couple the certificate we want to add
+  belongs to AND to other topics our topic may depend on (handles concurrency).
+- Ensure the new certificate's timestamp is newer than any existing certificate
+  in the topics we have locked (handles causality).
```
vxgmichel, line 126: Also, the section does not mention the causality dependencies that exist between certificates and other types of data such as blocks or vlobs.

vxgmichel (commented):
This document does not mention the role of `CertificateBasedActionIdempotentOutcome` and what the returned timestamp corresponds to.

Also it does not list the commands to indicate which topics are read/write locked.

vxgmichel:
Also this document doesn't mention the role of `RequireGreaterTimestamp`.

Also the version 2 used to explicitly list the causality dependencies that exists between the different certificates and vlobs, for instance:

 https://github.com/Scille/parsec-cloud/blob/d397c035d1369e498a48dea26837d93b7291dfe4/parsec/backend/realm.py#L351-L376

This document does not mention it and goes straight to the "reject if the timestamp is lower than any of the related topic and vlob timestamps". This is an approach that has its strength and weaknesses that should be presented as such.

Ideally the document should list the strict causality rules that should be enforced and compare both approaches. Here are some interesting elements about both:
- The "shortcut" approach allow to check the causality at the same time as the locking process (e.g with a method like `lock_for_timestamp(x)` that can file if the timestamp is too low).
- The strict approach greatly diminishes the number of false negative, thus increasing server performance and responsiveness of 

Note that we're only talking about causality here, per-topic locking is still compatible with strict causality checking.

server/README.md
touilleMan: `common` lock is taken when adding common certificates, so for instance when adding/revoking a new

This shouldn't have an impact on vlob/block performances given adding common certificates should be rare.

server/README.md
```diff
@@ -0,0 +1,236 @@
...
+
+- RCP API (`parsec/asgi/rpc.py`): this implement the HTTP API used by the Parsec client
+  and defined in the crate `libparsec_protocol`.
+- Administration API (`parsec/asgi/administration.py`): this is a JSON REST API designed
+  for the administration of the server (e.g. create a new organization).
+- Redirection (`parsec/asgi/redirect.py`): allow conversion between `https://` -> `parsec://`
+  URLs (used in email given `parsec://` URLs are often not displayed correctly by email clients)
+- Misc user-facing URLs (`parsec/asgi/__init__.py`), e.g. `index.html`, page not found etc.
+
+I should be obvious now that the main part of the server is those components that both
```
AureliaDolo, line 35:
```suggestion
It should be obvious now that the main part of the server is those components that both
```

server/README.md
```diff
@@ -0,0 +1,236 @@
...
+
+On top of that those components have two implementations:
+
+- PostgreSQL `parsec/components/postgresql`
+- In memory `parsec/components/memory`
+
+While the PostgerSQL implementation is the only one to use in production, the memory
+implementation is very convenient:
+
+- It simplicity allows for fast hack when testing new ideas.
```
AureliaDolo, line 46:
```suggestion
- Its simplicity allows for fast hack when testing new ideas.
```

server/README.md
```diff
@@ -0,0 +1,236 @@
...
+> Notes:
+>
+> - It could be tempting to only lock the `realm` topic (given it's this topic that contains
+>   the certificate describing the role of a user in the realm). However a revoked user is also
+>   forbidden from adding data, hence making it mandatory to also lock the `common` topic here.
+> - Uploading blocks (and vlobs) are the most common task on the server, so we may want to
+>   uptomise it in the future to avoid having to lock two topics.
+>   A possible solution would be to have an optimistic locking approach: first read the
+>   last timestamp in the topics, then do the checks, then on the final data insertion check
+>   again the timestamp haven't changed (and if they have we re-do everything from the start).
```
AureliaDolo, line 153: What would prevent a DoS if new certificates are continuously produced ?

server/README.md
AureliaDolo, line 160:
```suggestion
Concurrency & causality is handle is a similar fashion as for blocks.
```

server/README.md
```diff
@@ -0,0 +1,236 @@
...
+can run).
+
+However this is not perfect: adding an `await` to an async function seems harmless, but in
+our case it would break the assumption !
+
+On top of that event dispatch is asynchronous event in memory (this is needed to follow
+more closely what happen in PostgreSQL when a event may be recieved long after the
+request that sent it has finished).
+This is not an issue nevetheless given sending event is always the very last operation
+done, at which point the modification in database (and there correspondin checks) have
```
AureliaDolo, line 221:
```suggestion
done, at which point the modification in database (and their corresponding checks) have
```

server/README.md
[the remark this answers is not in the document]
AureliaDolo: Should we always take the lock on common, as everything depends on it ?

AureliaDolo:
> This is an approach that has its strength and weaknesses that should be presented as such. 

Do we have a target max query per second that we expect to manage per server ?

For example if a user wants to automatically upload logs from a talkative server should it be rate limited ? 

Or do we only expect real humans to produce certificates, hence the qps should correlate with the number of users.

Having these targets in mind may be useful in comparing strength and weaknesses of both approaches

server/README.md
touilleMan: You can think of both vlob and block upload as:

1) fetch info from database
2) do stuff in Python
3) store the vlob/block in the database

In the case of vlob, step 2 is only about checking access rights and vlob existence. On the other hand block also includes the write to the object storage.

However this is irrelevant from the database point of view: in both case step 3 is only allowed if data in step 1 hasn't changed.

Obviously this approach means we can end up with data in the  object storage while the SQL transaction got cancelled.
However this is acceptable:
- Given the block ID is determined client-side, re-trying to upload the block will lead to another write in the object storage at the same location.
- The object storage follows an eventual consistency strategy, hence the new block upload is guaranteed to overwrite the previous one. The only issue would be if the client upload a different block with the same ID between each try, in this case eventual consistency means it is possible for a short period of time to get a previous value when querying the block (but this is purely theoretical since there is no reason for the client to do that !).

server/README.md
touilleMan: :+1:

server/README.md
touilleMan: I've run dictionary check and corrected plenty !

server/README.md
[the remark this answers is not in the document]
touilleMan: Do you think I should add something to stress the importance of this here ?

server/README.md
touilleMan: :+1:

server/README.md
touilleMan: :+1:

server/README.md
touilleMan: All API where lock is involved are authenticated, hence it is trivial for server administrator to freeze the users doing the DoS using the administration REST API.

server/README.md
touilleMan: :+1:

touilleMan:
> Do we have a target max query per second that we expect to manage per server ?
>
> For example if a user wants to automatically upload logs from a talkative server should it be rate limited ?
>
> Or do we only expect real humans to produce certificates, hence the qps should correlate with the number of users.
>
> Having these targets in mind may be useful in comparing strength and weaknesses of both approaches

The assumption we have is:
- Certificates & invitation changes are very low volume
- Vlob & block changes are high volume

For instance, your "user wants to automatically upload logs from a talkative server" example only produce new vlobs & blocks.

If we compare a Parsec organization to a group of people doing some work, certificates & invitation changes should be seen as modifying the *structure* of the group while vlob & block changes are the actual work being done. Hence as long as your group is not [the Bolivian government](https://en.wikipedia.org/wiki/Coups_d%27%C3%A9tat_in_Bolivia) you should be fine :trollface:

server/README.md
touilleMan: That's a subtle one !

touilleMan:
> This document does not mention the role of CertificateBasedActionIdempotentOutcome and what the returned timestamp corresponds to.

I've added a small paragraph about this. I'm planning on improving it tomorrow though ;-)

> Also it does not list the commands to indicate which topics are read/write locked.

I have added a paragraph for this :+1:

server/README.md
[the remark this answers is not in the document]
AureliaDolo: I think so, it' quite important to remember that common is not just another topic

AureliaDolo:
> If we compare a Parsec organization to a group of people doing some work, certificates & invitation changes should be seen as modifying the structure of the group while vlob & block changes are the actual work being done. 

So if at the beginning of a big organization lifetime, if an admin spams admin creation, for these admins to onboard users automatically, there could be an issue ?

server/README.md
```diff
@@ -0,0 +1,370 @@
...
+authenticated | ping                               |        |        |           |        |
+authenticated | events_listen                      |        |        |           |        |
+authenticated | certificate_get                    |        |        |           |        |
+anonymous     | organization_bootstrap             |        |        |           |        |
+authenticated | invite_list                        |        |        |           |        |
+invited       | invite_info                        |        |        |           |        |
+authenticated | invite_new_device                  | read   |        |           |        |
+authenticated | invite_new_user                    | read   |        |           |        |
+authenticated | invite_cancel                      | read   |        |           |        |
+authenticated | invite_1_greeter_wait_peer         |        |        |           |        |
```
AureliaDolo, line 257: There will be the new invite command to add here after #7761 is merged

FirelightFlagboy: I would propose to write a script to update the table as it is highly likely that we will forget about it

server/README.md
FirelightFlagboy: I understand that's trivial for server admin, but how that's possible on the SaaS ?

server/README.md
```diff
@@ -0,0 +1,370 @@
...
+>   A possible solution would be to have an optimistic locking approach: first read the
+>   last timestamp in the topics, then do the checks, then on the final data insertion check
+>   again the timestamp haven't changed (and if they have we re-do everything from the start).
+
+### Vlobs
+
+Vlobs are encrypted metadata that belong to a given realm. Unlike block they can be
+modified and hence have a version field that must increase linearly.
+
+Concurrency & causality is handle is a similar fashion as for blocks.
```
FirelightFlagboy, line 184:
```suggestion
Concurrency & causality are handled in a similar fashion as for blocks.
```

server/README.md
```diff
@@ -0,0 +1,370 @@
...
+>   `common` and `realm` topics are involved).
+
+#### `RequireGreaterTimestamp` vs `CertificateBasedActionIdempotentOutcome`
+
+On top of command-specific checks, a server may reject a valid certificate by returning
+`RequireGreaterTimestamp` or `CertificateBasedActionIdempotentOutcome`.
+
+`RequireGreaterTimestamp` is simple to understand: a certificate has a timestamp more
+recent than the one provided in the command, hence the client should use a timestamp
+strictly superior than the one returned in the error to have a chance to succeed.
```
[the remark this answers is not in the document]
FirelightFlagboy: Should be but I'm in doubt, just want a confirmation :shrug:

server/README.md
touilleMan: done.

server/README.md
vxgmichel: I didn't mean the syntax, I mean that maybe it's not something to optimize in the future but deal with now in this design document.

server/README.md
touilleMan: I guess the whole point of SAAS is to not having to deal with things such as DoS (as it's the SAAS's administrators job) ;-)

server/README.md
touilleMan: > I would propose to write a script to update the table as it is highly likely that we will forget about it

There is a script now ! (and it checks more than just that ^^)

> There will be the new invite command

I've added them (because the script forced me :laughing: ), but I have the feeling `invite_greeter_start_greeting_attempt` & `invite_claimer_start_greeting_attempt` also requires some form of manual locking in the style of what is needed when creating an invitation: given the attempt is created by using the user ID as discriminant, but multiple attempts with the same discriminant are allowed in the database (only a single one of them is not removed, but this is not something the database can handle on its own with a unique constraint)

@AureliaDolo  what do you think ?

server/README.md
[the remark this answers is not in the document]
touilleMan: yes, the server checks the timestamp ballpark to prevent this !

server/README.md
```diff
@@ -0,0 +1,370 @@
...
+- Have lock in the server for each couple topic + discriminant
+- Take the lock for the topic + discriminant couple the certificate we want to add
+  belongs to AND to other topics our topic may depend on (handles concurrency).
+- Ensure the new certificate's timestamp is newer than any existing certificate
+  in the topics we have locked (handles causality).
+
+> Notes:
+>
+> - Also organizations are totally isolated from each other on the server, so
+>   in a way the organization ID is an additional discriminant on each certificate topic.
```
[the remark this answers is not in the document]
touilleMan: We have a `organization` field for (almost) all tables, e.g.

https://github.com/Scille/parsec-cloud/blob/8aa93ec2b2e926d3420ebdd3b2364700f3ee0fed/server/parsec/components/postgresql/migrations/datamodel.sql#L42

server/README.md
```diff
@@ -0,0 +1,370 @@
...
+- `common`
+- `sequester`, which depends on `common`
+- `realm`, which depends on `common` and uses the realm ID as discriminant
+- `shamir_recovery`, which depends on `common` and use the ID of the user to recover as discriminant
+
+So the actual rule used in the server is:
+
+- Have lock in the server for each couple topic + discriminant
+- Take the lock for the topic + discriminant couple the certificate we want to add
+  belongs to AND to other topics our topic may depend on (handles concurrency).
```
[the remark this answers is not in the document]
touilleMan: That's an interesting point, I've added this information to the sentence.

touilleMan:
> So if at the beginning of a big organization lifetime, if an admin spams admin creation, for these admins to onboard users automatically, there could be an issue ?

I don't think so, even in the case of a big organization (say 1000s of users) with plenty of realms (again 1000s with 1000s of user in each), we still end up with a fairly low number of items to insert in the database and with no huge famine (the users insertions fight for the `common` lock, then all realm role insertion are per-realm isolated).
So in the end: a couple of millions insertions, and not all of them need the same locks, so we would end up with a couple of hours to do the job at a very conservative 300 insert/s which is perfectly acceptable for a one-time operation for such a huge organization.

mmmarcos (commented):
LGTM overall. Some suggestions + re-arrange of sections.

I found it difficult to understand what strategy is actually being applied for each case (Certificates, blocks, vlobs, and invitations) mostly because there are lots of explanations and side-notes. The subject is complex, that's sure, but I think it would be easier to understand if there is a somehow "rigid" structure for each of this section. E.g.: 
```
Certificates

<Brief description of what certificates are>

<Brief description of *how* causality & concurrency are *actually* handled (general case)>

<More detailed explanation on *why* is handled like that (general case)>

<Extended description of *how* and *why* causality & concurrency are *actually* handled (specific cases)>
```
IHMO this would simplify understanding as problems and specific cases are presented incrementally.

On the other hand, this may require a considerable re-writing of the PR. So maybe it is OK to merge it as-is, and then improve with time.

server/README.md
```diff
@@ -0,0 +1,478 @@
...
+authenticated | invite_4_greeter_communicate           |        |       |           |        |                     |                  |                     |
+invited       | invite_1_claimer_wait_peer             |        |       |           |        |                     |                  |                     |
+invited       | invite_2a_claimer_send_hashed_nonce    |        |       |           |        |                     |                  |                     |
+invited       | invite_2b_claimer_send_nonce           |        |       |           |        |                     |                  |                     |
+invited       | invite_3a_claimer_signify_trust        |        |       |           |        |                     |                  |                     |
+invited       | invite_3b_claimer_wait_peer_trust      |        |       |           |        |                     |                  |                     |
+invited       | invite_4_claimer_communicate           |        |       |           |        |                     |                  |                     |
```
AureliaDolo, line 356: Should be removed

server/README.md
```diff
@@ -0,0 +1,472 @@
...
+  (see [example in Data concurrency paragraph](#data-concurrency)). This is handled
+  by a dedicated `invitation_creation` advisory lock that must be taken before the
+  invitation creation procedure starts any checks involving the invitations.
+
+### What is locked by who ?
+
+The following array recapitulate which topics (and other related locks) each command
+needs to lock (and in which mode).
+
+Topic locks: `common`, `realm`, `sequester`, `shamir`
```
vxgmichel, line 322: syntax is wrong, should add `\n`

server/README.md
```diff
@@ -0,0 +1,472 @@
...
+Manual locks: `invitation_creation`
+Special checks:
+
+- `last_vlob_timestamp`: Compare the to-be-inserted data's timestamp with the timestamp
+  of the most recent vlob.
+- `user_revoke_safe`: See [User revocation and concurrent operation involving users](#user-revocation-and-concurrent-operation-involving-users)
+
+<!-- See `misc/check_server_readme.py` script that checks this table is up to date ! -->
+<!-- start-table-what-is-lock-by-who -->
+API Family    |  command                               | common | realm | sequester | shamir | invitation_creation | user_revoke_safe | last_vlob_timestamp |
```
vxgmichel, line 332: Should also add the invitation row lock used by invitation cancel/complete/step etc.)

server/README.md
```diff
@@ -0,0 +1,472 @@
...
+>   to another write in the object storage at the same location.
+> - The object storage follows an eventual consistency strategy, hence the new block upload
+>   is guaranteed to overwrite the previous one.
+> - The only issue would be if multiple attempts of the same block ID would be done with
+>   different data on each try. In this case eventual consistency means it is possible
+>   for a short period of time to get a previous value when querying the block.
+>   However this is purely theoretical since block ID is random (hence there should be
+>   no concurrency between clients) and a client considers a block as immutable.
+
+### Invitations
```
vxgmichel, line 289: This section should be updated once https://github.com/Scille/parsec-cloud/issues/9302 is treated.

PR #8036, opened 2024-08-14
Add suffix `.unsigned` to windows & macos artifacts

FirelightFlagboy:
Add this suffix to indicate that the artifacts are not signed.


vxgmichel (approved):
LGTM :+1:

PR #8041, opened 2024-08-16
[CLI] Fix method calculating the short device id size 

### LINKED ISSUE #8034: [CLI] device list may provide short device id that aren't distiguishable
When the user provide a short device id with the option `--device` that conflict with multiple device, 
We display a list of conflicting devices with a short device id.
But on certain condition that short id may be not enough to identify which device to use, see error below.

```
Error: Multiple devices found for `08877b61`:
088 - TestOrg-fc47f825-1f0d-4304-9395-: First <first@dev1.com> @ dev1
088 - TestOrg-fc47f825-1f0d-4304-9395-: Second <second@dev2.com> @ second
```

The fix will need to update the function `format_devices` to display a short device id that is unique for the given set of devices will remaining small.

FirelightFlagboy:
Previously, we determine the short device id (**sid**) using the number of devices to display.
But this is not optimal if some device ids share the same prefix as we may risk display multiple similar **sid** that reference different devices.

```
088 - TestOrg-fc47f825-1f0d-4304-9395-: First <first@dev1.com> @ dev1
088 - TestOrg-fc47f825-1f0d-4304-9395-: Second <second@dev2.com> @ second
```

To fix that, we need to calculate the smallest **sid** that ensure it's unique for the whole set of devices.

---

To make things more interesting, I've added a benchmark using `criterion` to bench the different versions of the function `get_short_device_id_size` where I got from `40ms` down to `10µs` on my PC.

Closes #8034

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -53,13 +53,15 @@ resolver = "2"
 #
 # https://doc.rust-lang.org/cargo/reference/workspaces.html#the-package-table
 [workspace.package]
-publish = false
 authors = ["@Scille/parsec-dev", "@Scille/rust-code-owners"]
 edition = "2021"
 homepage = "https://parsec.cloud"
 license = "BUSL-1.1"
-version = "3.0.1-a.0"
+publish = false
 repository = "https://github.com/Scille/parsec-cloud"
+# We use the same version for all crates.
+# This comment is used to prevent `taplo` from adding multiple spaces between the version and the comment.
+version = "3.0.1-a.0" # __PARSEC_VERSION__
 
 [workspace.lints.clippy]
 undocumented_unsafe_blocks = "deny"
@@ -82,22 +84,22 @@ unexpected_cfgs = { level = "warn", check-cfg = [
 # check the default feature and making sure they are actually needed ;-)
 [workspace.dependencies]
 libparsec = { path = "libparsec", default-features = false }
-libparsec_protocol = { path = "libparsec/crates/protocol", default-features = false }
+libparsec_client = { path = "libparsec/crates/client", default-features = false }
+libparsec_client_connection = { path = "libparsec/crates/client_connection", default-features = false }
 libparsec_crypto = { path = "libparsec/crates/crypto", default-features = false }
-libparsec_types = { path = "libparsec/crates/types", default-features = false }
+libparsec_platform_async = { path = "libparsec/crates/platform_async", default-features = false }
 libparsec_platform_device_loader = { path = "libparsec/crates/platform_device_loader", default-features = false }
+libparsec_platform_http_proxy = { path = "libparsec/crates/platform_http_proxy", default-features = false }
 libparsec_platform_ipc = { path = "libparsec/crates/platform_ipc", default-features = false }
 libparsec_platform_mountpoint = { path = "libparsec/crates/platform_mountpoint", default-features = false }
 libparsec_platform_storage = { path = "libparsec/crates/platform_storage", default-features = false }
+libparsec_protocol = { path = "libparsec/crates/protocol", default-features = false }
 libparsec_serialization_format = { path = "libparsec/crates/serialization_format", default-features = false }
 libparsec_testbed = { path = "libparsec/crates/testbed", default-features = false }
 libparsec_tests_fixtures = { path = "libparsec/crates/tests_fixtures", default-features = false }
-libparsec_client_connection = { path = "libparsec/crates/client_connection", default-features = false }
-libparsec_platform_async = { path = "libparsec/crates/platform_async", default-features = false }
-libparsec_platform_http_proxy = { path = "libparsec/crates/platform_http_proxy", default-features = false }
-libparsec_client = { path = "libparsec/crates/client", default-features = false }
-libparsec_tests_macros = { path = "libparsec/crates/tests_macros", default-features = false }
 libparsec_tests_lite = { path = "libparsec/crates/tests_lite", default-features = false }
+libparsec_tests_macros = { path = "libparsec/crates/tests_macros", default-features = false }
+libparsec_types = { path = "libparsec/crates/types", default-features = false }
 libparsec_zstd = { path = "libparsec/crates/zstd", default-features = false }
 
 # Third parties
@@ -105,8 +107,8 @@ android_logger = { version = "0.13", default-features = false }
 anyhow = { version = "1.0.89", default-features = false }
 argon2 = { version = "0.5.3", default-features = false }
 assert_cmd = { version = "2.0.16", default-features = false }
-async-lock = { version = "3.4.0", default-features = false }
 async-broadcast = { version = "0.5.1", default-features = false }
+async-lock = { version = "3.4.0", default-features = false }
 base32 = { version = "0.4.0", default-features = false }
 base64 = { version = "0.22.1", default-features = false }
 blake2 = { version = "0.10.6", default-features = false }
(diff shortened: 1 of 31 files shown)


misc/version_updater.py
```diff
@@ -414,7 +414,12 @@ def refresh_npm_package_lock(update_files: set[Path]) -> set[Path]:
     },
     ROOT_DIR / "Cargo.toml": {
         Tool.License: [TOML_LICENSE_FIELD],
-        Tool.Parsec: [ReplaceRegex(r'^version = ".*"$', 'version = "{version}"')],
+        Tool.Parsec: [
+            ReplaceRegex(
+                r'^version = ".*" (\s*)# __PARSEC_VERSION__$',
```
touilleMan, line 419:
```suggestion
                r'^version = ".*"\W*# __PARSEC_VERSION__$',
```

- remove parenthesis given we don't need the group in the output
- remove the whitespace between the quote and the parenthesis to reduce the assumptions we rely on (if for whatever reason we end up with `version = "1.2.3"#__PARSEC_VERSION__`, then the code is still valid Rust but our assumption would break this script)
- `\W` to match any whitechar (so also `\t`), again to reduce the amount of assumptions we rely on

misc/version_updater.py
```diff
@@ -414,7 +414,12 @@ def refresh_npm_package_lock(update_files: set[Path]) -> set[Path]:
...
+                'version = "{version}"' + " " * 35 + "# __PARSEC_VERSION__",
```
touilleMan, line 420: 35 blanks ??? I don't find this more readable than with a single blank
```toml
version = "3.0.0-rc.1+dev"                                   # __PARSEC_VERSION__
version = "3.0.0-rc.1+dev" # __PARSEC_VERSION__
```

and it makes the code here more complex:
```python
'version = "{version}"' + " " * 35 + "# __PARSEC_VERSION__",
```
vs
```python
'version = "{version}" # __PARSEC_VERSION__',
```

touilleMan (changes requested):

cli/src/utils.rs
```diff
@@ -41,6 +39,31 @@ pub fn format_devices(
...
+        // We use the fact that we will display the id in hex format to compare both ids using the bitwise XOR and count the leading zeros.
+        // Since it takes 4 bits to represent a hex digit, we divide the leading zeros by 4 to get the position of the first different hex digit.
+        .map(|(a, b)| {
+            let cmp = a.as_u128() ^ b.as_u128();
+            let position = cmp.leading_zeros() / 4;
+            position as usize + 1
+        })
+        // Find the max value or fallback to 1 (meaning we have only one device)
+        .max()
+        .unwrap_or(1)
```
touilleMan, line 64: I would keep a lower bound on three characters (that's what we did on Parsec v2), given having a single character as ID makes it harder to understand it is an ID, and harder to copy&paste (double click on a word to select it is much simpler than selecting a single character)

cli/src/utils.rs
```diff
@@ -41,6 +39,31 @@ pub fn format_devices(
...
 
+pub fn get_minimal_short_id_size<'a, I>(devices: I) -> usize
+where
+    I: Iterator<Item = &'a libparsec::DeviceID>,
+{
+    use itertools::Itertools;
+
+    devices
+        // Ensure we do not have duplicates in the set.
+        .sorted_by(|a, b| a.cmp(b))
```
touilleMan, line 50: I don't think dedup is needed here:

The list provided by the caller is expected to be printed with the IDs shortened according to what we return.
So in case of duplication we do want to display the full IDs an not a subset, otherwise we will end up with multiple devices with the same short IDs, making us believe we do have multiple devices available instead of realizing there is a bug in the code !

misc/version_updater.py
FirelightFlagboy: - `\W` matches any non-word character (equivalent to `[^a-zA-Z0-9_]`)
- `\s` matches any whitespace character (equivalent to `[\r\n\t\f\v ]`)

That almost the same

cli/src/utils.rs
FirelightFlagboy: > So in case of duplication we do want to display the full IDs an not a subset, otherwise we will end up with multiple devices with the same short IDs

The algorithm already handle that case, see the test case where only the last hex char is different.

misc/version_updater.py
FirelightFlagboy: It's `taplo` that align the comment to column 62 because of the authors line that end at col 60

FirelightFlagboy:
```toml
[workspace.package]
authors = [
    "@Scille/parsec-dev",       # Every parsec dev
    "@Scille/rust-code-owners", # And the code owners
]
edition = "2021"
homepage = "https://parsec.cloud"
license = "BUSL-1.1"
publish = false
repository = "https://github.com/Scille/parsec-cloud"
version = "3.0.0-rc.1+dev" # __PARSEC_VERSION__
```

I can propose this format, we keep 1 space between the version and comment but we need to have the authors entries on multiple line

mmmarcos:
I am just curious: what's the goal of this "short device ID", what's is used for?

I am wondering if it is really worth it to have all this code instead of just displaying the whole ID without any processing. The select/copy from the terminal shouldn't be any more complicated, regardless of its length. 

Or maybe simply adding an index:
```
Index    Device ID    ....   
1        TestOrg-fc47f825-1f0d-4304-9395-: First <first@dev1.com> @ dev1
2        TestOrg-fc47f825-1f0d-4304-9395-: Second <second@dev2.com> @ second
```

misc/version_updater.py
FirelightFlagboy: Adding a comment before `version` field solve this issue

misc/version_updater.py
FirelightFlagboy: Fixed the multiple space issue in another way

cli/src/utils.rs
FirelightFlagboy: done

Cargo.toml
AureliaDolo, line 1: Is there something in place to ensure sorting ?

AureliaDolo (approved):

Cargo.toml
FirelightFlagboy: Yes, `taplo`: https://github.com/tamasfe/taplo

AureliaDolo:
I'm wondering if it would not be worth introducing an upper limit to the short id before defaulting to the real id: there is not much difference between typing 15 digits and 32 from a human point of view.

And if we would want to really simplify, I'd rather have a fixed length short id (say 4 or 5 digits) for when a real human is expected to interact with the CLI and a mode where only full ids are used.

AureliaDolo (approved):
It would be nice to have a blogpost/entry in the doc to explain a bit that the displayed id may change when new devices are created, and maybe dive into the benchmarks and so on ^^

cli/src/utils.rs
```diff
@@ -41,6 +39,38 @@ pub fn format_devices(
...
+/// we want to keep a lower bound for the size of the short id to keep it simple to select the value via double click.
+pub(crate) const MINIMAL_SHORT_ID_SIZE: usize = 3;
+
+pub fn get_minimal_short_id_size<'a, I>(devices: I) -> usize
+where
+    I: Iterator<Item = &'a libparsec::DeviceID> + Clone,
+{
+    use itertools::Itertools;
+
+    // If we have duplicate, we short circuit to the max value to represent the id in hex format, which is 32 char.
```
touilleMan, line 52: I don't get the point of doing this:
- the default case should take care of this fine (i.e. removing this part requires no other changes in the function), so better keep things simple
- In practice this shouldn't occur so there is no need for performance optimization here (and even if it occurs, the list of devices is most likely very small so again performances are no big deal)

touilleMan (approved):

PR #8063, opened 2024-08-20
[MS] Custom order Stats

### LINKED ISSUE #7968: [Client - Sellsy] Statistics page
https://www.figma.com/design/8WXUivEfT1x4FTzCr3WNxd/Parsec-Client?node-id=216-6732&t=M6Eot8DLK14rZkZo-11

fabienSvtr:
## Describe your changes

image

## Checklist

- [ ] Describe your changes
   
closes #7968

### DIFF
--- a/client/src/views/client-area/statistics/CustomOrderStatisticsPage.vue
+++ b/client/src/views/client-area/statistics/CustomOrderStatisticsPage.vue
@@ -1,0 +1,667 @@
+<!-- Parsec Cloud (https://parsec.cloud) Copyright (c) BUSL-1.1 2016-present Scille SAS -->
+
+<template>
+  <div class="client-page-statistics">
+    <template v-if="organizationStats && contractDetails">
+      <!-- active users -->
+      <div class="users active">
+        <div class="users-text">
+          <ion-title class="users-active-text__title title-h2">
+            {{ $msTranslate('clientArea.statistics.titles.users') }}
+            <span class="subtitles-normal">
+              {{ $msTranslate('clientArea.statistics.titles.active') }}
+            </span>
+          </ion-title>
+        </div>
+        <div class="users-cards-list">
+          <ion-card class="users-cards-list-item">
+            <div class="users-cards-list-item-icons">
+              <ion-icon
+                class="users-cards-list-item-icons__main"
+                :icon="person"
+              />
+              <ion-icon
+                class="users-cards-list-item-icons__secondary"
+                :icon="star"
+              />
+            </div>
+            <div class="users-cards-list-item-text">
+              <ion-text class="users-cards-list-item-text__number title-h1">{{ organizationStats.adminUsersDetail.active }}</ion-text>
+              <ion-text class="users-cards-list-item-text__text subtitles-sm">
+                {{ $msTranslate({ key: 'clientArea.statistics.admin', count: organizationStats.adminUsersDetail.active }) }}
+              </ion-text>
+            </div>
+          </ion-card>
+          <ion-card class="users-cards-list-item">
+            <div class="users-cards-list-item-icons">
+              <ion-icon
+                class="users-cards-list-item-icons__main"
+                :icon="person"
+              />
+              <ion-icon
+                class="users-cards-list-item-icons__secondary"
+                :icon="pencil"
+              />
+            </div>
+            <div class="users-cards-list-item-text">
+              <ion-text class="users-cards-list-item-text__number title-h1">{{ organizationStats.standardUsersDetail.active }}</ion-text>
+              <ion-text class="users-cards-list-item-text__text subtitles-sm">
+                {{ $msTranslate({ key: 'clientArea.statistics.standard', count: organizationStats.standardUsersDetail.active }) }}
+              </ion-text>
+            </div>
+          </ion-card>
+          <ion-card class="users-cards-list-item">
+            <div class="users-cards-list-item-icons">
+              <ion-icon
+                class="users-cards-list-item-icons__main"
+                :icon="person"
+              />
+              <ion-icon
+                class="users-cards-list-item-icons__secondary"
+                :icon="arrowForward"
+              />
+            </div>
+            <div class="users-cards-list-item-text">
+              <ion-text class="users-cards-list-item-text__number title-h1">{{ organizationStats.outsiderUsersDetail.active }}</ion-text>
+              <ion-text class="users-cards-list-item-text__text subtitles-sm">
+                {{ $msTranslate({ key: 'clientArea.statistics.outsider', count: organizationStats.outsiderUsersDetail.active }) }}
+              </ion-text>
+            </div>
+          </ion-card>
+        </div>
+      </div>
+      <!-- revoked users -->
+      <div class="users revoked">
+        <div class="users-text">
+          <ion-title class="users-active-text__title title-h2">
+            {{ $msTranslate('clientArea.statistics.titles.users') }}
+            <span class="subtitles-normal">
+              {{ $msTranslate('clientArea.statistics.titles.revoked') }}
+            </span>
+          </ion-title>
+        </div>
+        <div class="users-cards-list">
+          <ion-card class="users-cards-list-item">
+            <div class="users-cards-list-item-icons">
+              <ion-icon
+                class="users-cards-list-item-icons__main"
+                :icon="person"
+              />
+              <ion-icon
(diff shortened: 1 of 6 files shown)


client/src/views/client-area/statistics/CustomOrderStatisticsPage.vue
```diff
@@ -0,0 +1,719 @@
...
+                class="users-cards-list-item-icons__secondary"
+                :icon="star"
+              />
+            </div>
+            <div class="users-cards-list-item-text">
+              <ion-text class="users-cards-list-item-text__number title-h1">{{ organizationStats.adminUsersDetail }}</ion-text>
+              <ion-text class="users-cards-list-item-text__text subtitles-sm">
+                {{
+                  $msTranslate(
+                    organizationStats.adminUsersDetail.active > 1 ? 'clientArea.statistics.admins' : 'clientArea.statistics.admin',
```
Max-7, line 33: Use the plural translation. You can define in the translation file:
`"admin": "Administrator | Administrators"`
and then use
`$msTranslate({ key: 'clientArea.statistics.admin', count: organizationStats.adminUsersDetail.active })`
and it will automatically choose whether to display `Administrator` or `Administrators` based on `count`.

client/src/views/client-area/statistics/CustomOrderStatisticsPage.vue
```diff
@@ -0,0 +1,719 @@
...
+                class="users-cards-list-item-icons__secondary"
+                :icon="pencil"
+              />
+            </div>
+            <div class="users-cards-list-item-text">
+              <ion-text class="users-cards-list-item-text__number title-h1">{{ organizationStats.standardUsersDetail }}</ion-text>
+              <ion-text class="users-cards-list-item-text__text subtitles-sm">
+                {{
+                  $msTranslate(
+                    organizationStats.standardUsersDetail.active > 1 ? 'clientArea.statistics.standards' : 'clientArea.statistics.standard',
```
Max-7, line 55: Same here, use the plural.

client/src/views/client-area/statistics/CustomOrderStatisticsPage.vue
```diff
@@ -0,0 +1,719 @@
...
+                class="users-cards-list-item-icons__secondary"
+                :icon="arrowForward"
+              />
+            </div>
+            <div class="users-cards-list-item-text">
+              <ion-text class="users-cards-list-item-text__number title-h1">{{ organizationStats.outsiderUsersDetail }}</ion-text>
+              <ion-text class="users-cards-list-item-text__text subtitles-sm">
+                {{
+                  $msTranslate(
+                    organizationStats.outsiderUsersDetail.active > 1 ? 'clientArea.statistics.outsiders' : 'clientArea.statistics.outsider',
```
Max-7, line 77: Same.

client/src/views/client-area/statistics/CustomOrderStatisticsPage.vue
```diff
@@ -0,0 +1,719 @@
...
+                class="users-cards-list-item-icons__main"
+                :icon="person"
+              />
+              <ion-icon
+                class="users-cards-list-item-icons__secondary"
+                :icon="star"
+              />
+            </div>
+            <div class="users-cards-list-item-text">
+              <ion-text class="users-cards-list-item-text__number title-h1">{{ organizationStats.adminUsersDetail.revoked || 0 }}</ion-text>
```
Max-7, line 108: Why `|| 0`?

client/src/views/client-area/statistics/CustomOrderStatisticsPage.vue
```diff
@@ -0,0 +1,719 @@
...
+        </div>
+      </div>
+      <!-- storage -->
+      <div class="storage">
+        <ion-text class="storage__title title-h2">{{ $msTranslate('clientArea.statistics.titles.storage') }}</ion-text>
+        <div class="storage-data">
+          <div class="storage-data-global">
+            <ion-text
+              class="storage-data-global__total title-h1"
+              :class="useStoragePercentage >= 100 ? 'overused' : ''"
```
Max-7, line 179: It can be written as `:class="{ 'overused': useStoragePercentage >= 100}"`, which I find a little cleaner.

client/src/views/client-area/statistics/CustomOrderStatisticsPage.vue
```diff
@@ -0,0 +1,719 @@
...
+          </div>
+          <div class="storage-data-consumption">
+            <ion-text class="storage-data-consumption__title title-h5">
+              {{ $msTranslate('clientArea.statistics.consumptionDetail') }}
+            </ion-text>
+
+            <div class="consumption">
+              <ms-report-text
+                :theme="MsReportTheme.Warning"
+                v-if="useStoragePercentage === 100"
```
Max-7, line 203: Shouldn't we warn when it's above 90% or something? 100% is already too late.

client/src/views/client-area/statistics/CustomOrderStatisticsPage.vue
```diff
@@ -0,0 +1,719 @@
...
+import { isDefaultOrganization } from '@/views/client-area/types';
+import { MsReportTheme, MsReportText } from 'megashark-lib';
+
+const props = defineProps<{
+  organization: BmsOrganization;
+}>();
+
+const organizationStats = ref<OrganizationStatsResultData | undefined>(undefined);
+const contractDetails = ref<CustomOrderDetailsResultData | undefined>(undefined);
+const useStoragePercentage = ref<number>(0);
```
Max-7, line 391: I think all those variables and functions could be replaced by computed. This way we avoid defining a variable and a function just to set that variable and we avoid having `get` functions that set values/

```
const storageSize = computed(...);
const storageWidth = computed(...);
const remainingStorageSize = computed(...);
const storagePercentage = computed(...);
...
```

client/src/views/client-area/statistics/CustomOrderStatisticsPage.vue
```diff
@@ -0,0 +1,719 @@
...
+
+function getUseStoragePercentage(): number {
+  if (!organizationStats.value) {
+    return 0;
+  }
+  const ordered = getStorageSize();
+  const current = totalData.value;
+
+  if (current / ordered >= 1) {
+    return (useStoragePercentage.value = 100);
```
Max-7, line 420: Never set a variable in a function named `get`. That's called a side-effect (https://en.wikipedia.org/wiki/Side_effect_(computer_science)) and it is huge source of bugs (some languages like C++ even have a `const` keyword on functions to make sure that a function does not change a value).

client/src/views/client-area/statistics/CustomOrderStatisticsPage.vue
```diff
@@ -0,0 +1,719 @@
...
+}
+
+function getUseStoragePercentage(): number {
+  if (!organizationStats.value) {
+    return 0;
+  }
+  const ordered = getStorageSize();
+  const current = totalData.value;
+
+  if (current / ordered >= 1) {
```
Max-7, line 419: You can replace this if/else with
`return Math.min(Math.round((current / ordered) * 100), 100)`.

`round` makes sure that if your percentage is `13.333333333` it gets rounded to `13`.
`min()` takes the lesser values of the two: if your percentage goes above 100, it will pick the second parameter (100) because it's the lowest of the two.

Max-7 (changes requested):
I avoided repeating some comments but they apply everywhere.

I also strongly encourage you to try with some mock values (you can change the values in `bms/api.ts`) to try with different user count, storage space and all that.

client/src/views/client-area/statistics/CustomOrderStatisticsPage.vue
```diff
@@ -0,0 +1,666 @@
...
+            <div class="consumption">
+              <ms-report-text
+                :theme="MsReportTheme.Warning"
+                v-if="storagePercentage === 100"
+              >
+                {{ $msTranslate('clientArea.statistics.fullUsed') }}
+              </ms-report-text>
+              <ms-report-text
+                :theme="MsReportTheme.Error"
+                v-if="storagePercentage > 90 && storagePercentage < 100"
```
Max-7, line 182: You could just write `v-else-if="storagePercentage > 90"`. The case where it's 100 would be covered by the first `v-if`.

Max-7 (approved):

PR #8080, opened 2024-08-21
PostgreSQL server side implementation of the non-blocking invite transport

AureliaDolo:


server/parsec/components/postgresql/invite.py
```diff
@@ -1063,6 +1063,20 @@ async def test_dump_all_invitations(
...
+        invitation_created_by: UserID,
+        greeter_id: UserID,
+        greeter_profile: UserProfile,
+    ) -> bool:
+        if invitation_type == InvitationType.DEVICE:
+            return invitation_created_by == greeter_id
+        elif invitation_type == InvitationType.USER:
+            return greeter_profile == UserProfile.ADMIN
+        else:
+            raise NotImplementedError
```
mmmarcos, line 1078: This logic is also implemented in `components/memory/invite.py`:
https://github.com/Scille/parsec-cloud/blob/9ed86452345a54058ac6f17f5b5d266d7bac01f6/server/parsec/components/memory/invite.py#L658-L664

Can it be moved up to `components/invite.py`? to share it between `memory` and `postgresql` implementations?

server/parsec/components/postgresql/migrations/0002_non_blocking_invitation.sql
mmmarcos, line 1: In case there are existing pending invitations, should they be migrated?

vxgmichel: Invitations are left as they are: the new transport only affect the greeting sessions

server/tests/common/data.py
```diff
@@ -343,8 +343,10 @@ async def wksp1_bob_becomes_owner_and_changes_alice(
         assert isinstance(outcome, RealmRoleCertificate)
 
 
-HttpCommonErrorsTesterDoCallback = Callable[[], Coroutine[None, None, None]]
-HttpCommonErrorsTester = Callable[[HttpCommonErrorsTesterDoCallback], Coroutine[None, None, None]]
+type HttpCommonErrorsTesterDoCallback = Callable[[], Coroutine[None, None, None]]
```
touilleMan, line 346: :open_mouth:

server/parsec/components/postgresql/invite.py
```diff
@@ -1097,55 +1374,563 @@ async def test_dump_all_invitations(
...
+                assert greeter == greeter_user_id
+                pass
+            case CheckDeviceBadOutcome.DEVICE_NOT_FOUND:
+                return InviteGreeterStartGreetingAttemptBadOutcome.AUTHOR_NOT_FOUND
+            case CheckDeviceBadOutcome.USER_NOT_FOUND:
+                return InviteGreeterStartGreetingAttemptBadOutcome.AUTHOR_NOT_FOUND
+            case CheckDeviceBadOutcome.USER_REVOKED:
+                return InviteGreeterStartGreetingAttemptBadOutcome.AUTHOR_REVOKED
+
+        invitation_info = await self.get_invitation_info(conn, organization_id, token)
```
touilleMan, line 1757: There is no lock here, how is handled the concurrency ? (i.e. multiple calls to `greeter_start_greeting_attempt` leading to more than one active attempt)

server/parsec/components/postgresql/invite.py
```diff
@@ -1097,55 +1374,563 @@ async def test_dump_all_invitations(
...
+
+    async def new_attempt_for_claimer(
+        self,
+        conn: AsyncpgConnection,
+        organization_id: OrganizationID,
+        greeter: UserID,
+        invitation_internal_id: int,
+        now: DateTime,
+    ) -> GreetingAttemptID:
+        (
```
touilleMan, line 1475: First condition can be added to the while loop

server/parsec/components/postgresql/invite.py
```diff
@@ -1097,55 +1374,563 @@ async def test_dump_all_invitations(
...
+            greeting_attempt_internal_id,
+            greeting_attempt_id,
+        ) = await self.get_active_attempt(
+            conn,
+            organization_id=organization_id,
+            greeter=greeter,
+            invitation_internal_id=invitation_internal_id,
+        )
+        is_active = await self.claimer_join_or_cancel(conn, greeting_attempt_internal_id, now)
+        while not is_active:
```
touilleMan, line 1485: I seems weird to have a loop here: given we are in a transaction we should be able to lock the session row to handle concurrency

touilleMan (commented):
Regarding concurrency, I would say we should lock the invitation row for all write operations (so everything involving a given invitation is sequential \o/), and have an advisory lock for invitation creation (given at this point there is no row  to lock)

server/parsec/components/postgresql/invite.py
```diff
@@ -1097,55 +1374,563 @@ async def test_dump_all_invitations(
...
+        await conn.execute(
+            *_q_cancel_greeting_attempt(
+                greeting_attempt_internal_id=greeting_attempt_internal_id,
+                cancelled_by=origin.str,
+                cancelled_reason=reason.str,
+                cancelled_on=now,
+            )
+        )
+
+    class StepOutcome(Enum):
```
AureliaDolo, line 1639: Why is StepOutcome defined inside PGInviteComponent ?

server/parsec/components/postgresql/test_queries.py
```diff
@@ -114,11 +114,27 @@
...
+deleted_greeting_steps AS (
+    DELETE FROM greeting_step
+    WHERE greeting_attempt in (select * from deleted_greeting_attempts)
+    RETURNING _id
 )
 SELECT 1
 """
 )
 
+
```
AureliaDolo, line 137:
```suggestion
```

server/poetry.lock
```diff
@@ -1,4 +1,4 @@
-# This file is automatically @generated by Poetry 1.8.3 and should not be changed by hand.
+# This file is automatically @generated by Poetry 1.8.2 and should not be changed by hand.
```
AureliaDolo, line 1: Is it weird that poetry version downgraded ?

server/tests/api_v4/authenticated/test_invite_greeter_start_greeting_attempt.py
```diff
@@ -1,29 +1,38 @@
...
-# TODO: Remove once PostgreSQL is supported
-@pytest.fixture(autouse=True)
-def _skip_if_postgresql(skip_if_postgresql: None) -> None:  # type: ignore
-    pass
-
-
 async def test_authenticated_invite_greeter_start_greeting_attempt_ok(
     coolorg: CoolorgRpcClients,
 ) -> None:
+    # This is a scenario where the greeter starts first.
```
AureliaDolo, line 17: Should we add a scenario where the claimer starts first ?

(the test that assert that all tests are here doesn't checks if there are more tests rigth?)

server/tests/common/data.py
AureliaDolo: what does that type mean ?

AureliaDolo (commented):
:+1:

server/parsec/components/postgresql/invite.py
```diff
@@ -1154,26 +1939,179 @@ async def greeter_step(
...
+        # Only the greeter or the claimer can complete the invitation
+        if not self.is_greeter_allowed(invitation_info, author_user_id, current_profile):
+            if not invitation_info.claimer_email == author_info.human_handle.email:
+                return InviteCompleteBadOutcome.AUTHOR_NOT_ALLOWED
+
+        await conn.execute(
+            *_q_delete_invitation(
+                invitation_internal_id=invitation_info.internal_id,
+                on=now,
+                reason="FINISHED",  # TODO: use an enum
```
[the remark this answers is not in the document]
AureliaDolo: see  #8224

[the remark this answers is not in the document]
FirelightFlagboy: Just need to update the comment then

server/parsec/components/postgresql/invite.py
vxgmichel: To be closer to its usage

server/poetry.lock
vxgmichel: eeeerh sorry ^^

server/tests/api_v4/authenticated/test_invite_greeter_start_greeting_attempt.py
vxgmichel: Sorry it's an artefact from a test I moved in `test_greeting_attempt.py` it should be removed

server/tests/common/data.py
vxgmichel: It defines a type alias: https://docs.python.org/3/library/typing.html#type-aliases

server/parsec/components/postgresql/invite.py
vxgmichel: This cannot happen due to :
```sql
-- Makes sure that there is only one active greeting attempt per session
CREATE UNIQUE INDEX unique_active_attempt ON greeting_attempt (greeting_session)
WHERE cancelled_on IS NULL;
```

vxgmichel:
@touilleMan 
> Regarding concurrency, I would say we should lock the invitation row for all write operations (so everything involving a given invitation is sequential \o/), and have an advisory lock for invitation creation (given at this point there is no row to lock)

That sounds good (no performance issue obviously). However, we must be careful about the order of things being locked in order to avoid deadlocks. In particular, remember that the `invitation_token` is not provided for 4 commands.


Here's how the current resources are managed by the 7 commands that have been added:


## greeter/claimer_start_greeting_attempt

Resources (in order)
- organization 
	- use `organization._get`
- author / greeter
	- use `user._check_device` for greeter 
	- used `user._check_user` for claimer, but removed in #8210
- invitation
	- no lock, but fine for invitations?
- greeting session
	- atomic "create if not exist"
	- has no internal state to protect
- greeting attempt
	- atomic "create if not exist"
	- atomic "join or cancel"
	- loop until it works

## greeter/claimer_cancel_greeting_attempt

Resources (in order)
- organization 
	- use `organization._get`
-  author (for greeter)
	- use `user._check_device`  
- greeting attempt
	- lock for update
- invitation
	- no lock, but fine for invitations?
- greeter 
	- used `user._check_user` but ordering is wrong!
- proceed to update the greeting attempt

## greeter/claimer_step

Resources (in order)
- organization 
	- use `organization._get`
-  author (for greeter)
	- use `user._check_device`  
- greeting attempt
	- lock for update
- invitation
	- no lock, but fine for invitations?
- greeter 
	- used `user._check_user` but ordering is wrong!
- proceed to update the greeting attempt steps

## Invitation complete

Resources (in order)
- organization 
	- use `organization._get`
- author
	- use `user._check_device`
- invitation
	- no lock, but fine for invitations?
- proceed to update the invitation

## Alternative

> we should lock the invitation row for all write operations

The 7 commands are all write operations. 

In order to have a consistent locking scheme, we could do the following:

- organization 
   - use `organization._get`
- invitation
   - get it either from the invitation token, or the attempt id
   - lock it for update
- author/greeter
   - get it from the attempt id for `claimer_cancel_greeting_attempt` and `claimer_step`
   - use either `_check_device` or `check_user` to get the profile
- greeting attempt
   - no locking required, it's already part of the locked invitation
- perform the change in the `greeting_attempt` or `greeting_step` table

Also note that it's not crucial to lock the author/greeter during the transaction, as it would not produce a race condition on any data if the author/greeter profile was to be changed during the transaction. But I guess it's good practice nevertheless.

@touilleMan What do you think?

touilleMan:
I think the alternative approach is more in-line with the topic based system of locking (i.e. having big locks easy to use instead of tailoring the lock needs for each command)

> In particular, remember that the invitation_token is not provided for 4 commands

:trollface: 

> However, we must be careful about the order of things being locked in order to avoid deadlocks

Totally !

Deadlock can only occur if multiple things get locked (so A does 1 then 2, while B does 2 then 1), hence:
1 - we lock only the invitation row
2 - or, if we want to lock multiple rows, we only lock once using a join

Approach 1 looks like:
```SQL
WITH my_invitation AS (
  SELECT
    _id,
    -- select what you need
  FROM invitation
  WHERE
    -- We use a nested select here instead of a join to avoid locking the
    -- `greeting_session`&`greeting_attempt` rows
    _id = (SELECT greeting_session.invitation
              FROM greeting_attempt
              JOIN greeting_session ON greeting_attempt.session = greeting_session._id
              WHERE
                   greeting_attempt.organization = $organization_internal_id
                   AND greeting_attempt.greeting_attempt_id = $greeting_attempt_id
    )
  FOR UPDATE
)

...
```

Approach 2 looks like:
```SQL
WITH my_invitation AS (
  SELECT
    invitation._id,
    -- select what you need
  FROM invitation
  INNER JOIN greeting_attempt ON invitation._id = greeting_attempt.invitation
  INNER JOIN greeting_session ON greeting_attempt.session = greeting_session._id
  WHERE
       greeting_attempt.organization = $organization_internal_id
       AND greeting_attempt.greeting_attempt_id = $greeting_attempt_id
  -- Note we are locking invitation + greeting_session + greeting_attempt rows here !
  FOR UPDATE
)

...
```

However there is trick here !
`greeting_attempt` row will get locked any time we modify it (I suppose `greeting_session` is never modified after insert), so we will end up with two independant locks (for our explicit `FOR UPDATE`, then the implicit lock done by the `UPDATE` query) in the case of approach 1. This is fine though since the locks are always done in the same order (hence the need to have a helper function that does the lock to ensure it is always taken the same way !).

So in the end I'd rather go with solution 1:
- It makes a simpler rule: "any invitation operation must first lock for write the invitation row"
- It makes the query code simpler since you can have a template code containing `SELECT ... FROM invitation WHERE _id = { invitation_id }` (with `invitation_id` being replaced by the subquery when getting the invitation from the attempt id)

(I'm not strongly again solution 2, so I you'd rather take it just let me know so that I update the documentation I'm writting ^^)

>     use organization._get

So you know, I plan on totally remove `organization._get` (it is currently still there only for the needs of invitation commands that I haven't modified)

The reason for this is that all the components class are stateless (since all data are in the... database ^^), hence this method can be a helper function instead.

This turns components classes into [merely an aggregate](https://github.com/Scille/parsec-cloud/blob/master/server/parsec/components/postgresql/realm.py) of commands [implementation](https://github.com/Scille/parsec-cloud/blob/master/server/parsec/components/postgresql/realm_rename.py), which I find much more readable that the huge file containing helper methods used in others huge files ;-)

Currently [there is few helpers](https://github.com/Scille/parsec-cloud/tree/master/server/parsec/components/postgresql/queries) dedicated to handling the common errors (e.g. organization not found) and lock the topics.

I think ideally we should have a `q_auth_lock_invitation` that contains `auth_and_lock_invitation(token: InvitationToken)` and `auth_and_lock_invitation_from_attempt(token: GreetingAttemptID)`. Both doing a write lock on the invitation row as discussed above.

> greeter/claimer_start_greeting_attempt
>
> use user._check_device for greeter
> used user._check_user for claimer
> [...]
> Also note that it's not crucial to lock the author/greeter during the transaction, as it would not produce a race condition on any data if the author/greeter profile was to be changed during the transaction. But I guess it's good practice nevertheless.

I considered in my design document that all the invitation API never take common topic lock, except for the invitation create (given we must ensure no user exist for this invitation !).

The rational being that the invitation system is an isolated system that doesn't use certificates, hence a recently revoked user modifying an invitation due to concurrency won't break consistency (everybody will just consider the user wasn't revoked when it made the change, this is especially true given most change do not register when they occurred !)

Thinking more about it, I think this approach is too fragile. Instead we should stick to the "lock the common topic whenever a user is doing something" rule.

So that would mean:
- for all greeting commands do a read lock on common (using the fancy [auth_and_lock_common_read](https://github.com/Scille/parsec-cloud/blob/598bcc34d933087a1a3a75dff962dad77f725582/server/parsec/components/postgresql/queries/q_auth_lock_common_only.py#L114-L118) ^^) *then* write lock the invitation.
- for all claiming commands just do a write lock on the invitation (since the invitation *is* what allow the claimer to authenticate and do things)

## What about creating a user without invitation ?

This makes me think of a corner case we haven't addressed: what happen if a user gets created outside of the invitation system ? (e.g. using the pki)
In this case, any existing invitation for to-be-created-user should get cancelled (and do we need to also cancel any pending greeting attempt ?).

This also means the user_create command will lock the invitation row for updating it (and possibly the greeting attempts rows too), it shouldn't be a big deal though.

Should I do a PR on this ?

server/tests/common/data.py
AureliaDolo: My question was more about the callable and coroutine part

vxgmichel: Ah yes, it's corresponds to an async function with the following prototype
```python
async def my_func() -> None
```
It's equivalent to `Callable[[], Awaitable[None]]` which I find more readable.

server/parsec/components/postgresql/invite.py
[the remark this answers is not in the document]
vxgmichel: fixed [2381875](https://github.com/Scille/parsec-cloud/pull/8080/commits/2381875c23a51d6429bc44df2d29d1559b50f9cf)

server/parsec/components/postgresql/invite.py
```diff
@@ -1376,13 +1374,13 @@ async def test_dump_all_invitations(
...
 
-    async def get_active_attempt(
+    async def get_greeting_session(
         self,
         conn: AsyncpgConnection,
         organization_id: OrganizationID,
         greeter: UserID,
         invitation_internal_id: int,
-    ) -> tuple[int, GreetingAttemptID]:
+    ) -> int:
```
AureliaDolo, line 1383: Here it's an int because a greeting session only has an id in the DB ?

It's a matter of personal preference, but I'd add a comment or a type alias to make that explicit

server/parsec/components/postgresql/invite.py
```diff
@@ -1405,95 +1410,72 @@ async def get_active_attempt(
...
-                greeter=greeter,
-                invitation_internal_id=invitation_internal_id,
+                greeting_session_id=greeting_session_id,
             )
-            is_active = await self.claimer_join_or_cancel(conn, greeting_attempt_internal_id, now)
-        return greeting_attempt_id
+            # Join the attempt
+            is_active = await self.join_or_cancel(peer, conn, greeting_attempt_internal_id, now)
+            # It has to be active since we just created it
+            assert is_active
```
AureliaDolo, line 1477: From what I understand, asserts are not checked if code is optimized, and as attempt creation and joining are not atomic, should we add a warning or a real error ?

server/parsec/components/postgresql/user.py
```diff
@@ -107,17 +106,23 @@ def __init__(self, pool: AsyncpgPool, event_bus: EventBus) -> None:
...
         self.event_bus = event_bus
 
+    async def _check_common_topic(
+        self, conn: AsyncpgConnection, organization_id: OrganizationID
+    ) -> DateTime:
+        common_timestamp = await conn.fetchval(
+            *_q_check_common_topic(organization_id=organization_id.str)
+        )
+        if common_timestamp is None:
+            common_timestamp = DateTime.from_timestamp_micros(0)
```
AureliaDolo, line 116: I don't get why the default timestamp here is 0

vxgmichel:
@touilleMan Concurrency is now handled correctly, with the following lock order being used for all requests:
- get organization
- common topic for share
- invitation for update

> This makes me think of a corner case we haven't addressed: what happen if a user gets created outside of the invitation system ? (e.g. using the pki)
In this case, any existing invitation for to-be-created-user should get cancelled (and do we need to also cancel any pending greeting attempt ?).

Good point

> Should I do a PR on this ?

No I think the current implementation is OK, what you suggest is an enhancement. I'll create an issue about it so we can plan it after the release.

server/parsec/components/postgresql/invite.py
vxgmichel: We never disable the asserts when running the server, so asserts are used for errors that would indicate a bug. They are logged by sentry and properly managed by the backend (hopefully)

server/parsec/components/postgresql/user.py
vxgmichel: That's the lower date possible, most likely epoch. That makes it a good default when no certificates are present.

AureliaDolo: Ok ^^

But is a situation when there is no certificate at all expected ?

vxgmichel (approved):
Enough reviews, let's merge this

server/parsec/components/postgresql/user.py
vxgmichel: yes, when bootstrapping the organization if I remember correctly.

server/parsec/components/postgresql/invite.py
vxgmichel: I wanted to add a type alias but I forgot your comment and merged the PR. I'll add it later if I have to come back to fix some stuff

PR #8086, opened 2024-08-21
Server memory topic locks

touilleMan:


server/parsec/components/memory/datamodel.py
```diff
@@ -79,6 +93,64 @@ class MemoryOrganization:
...
+
+            # 4) Release our locks and leave
+
+            self._topic_write_locked.difference_update(write)
+            for topic_and_discriminant in read:
+                self._topic_read_locked[topic_and_discriminant] -= 1
+
+            if self._notify_me_on_topic_lock_release is not None:
+                self._notify_me_on_topic_lock_release.set()
+                self._notify_me_on_topic_lock_release = Event()
```
vxgmichel, line 150: Use a `try-finally`, otherwise the lock won't be released when an exception is raised

vxgmichel (changes requested):

server/parsec/components/memory/datamodel.py
touilleMan: Good catch ! :1st_place_medal: 

fixed.

PR #8094, opened 2024-08-21
PostgreSQL implementation of `user_revoke`  handles newer vlob/block

### LINKED ISSUE #8093: PostgreSQL implementation of `user_revoke`/`realm_share`/`realm_unshare`  don't handle newer vlob/realm_certificate
https://github.com/Scille/parsec-cloud/blob/1f2739d6db3e33b23eba6b12934e74966268d740/server/parsec/components/postgresql/user.py#L1158-L1169

This code should be checking the revoked user hasn't uploaded data that are newer than the last vlob/blocks he has uploaded.

(also note the comment in the code that doesn't correspond to the code, as it has been directly taken from the memory implementation)

touilleMan:
fix #8093


touilleMan:
replaced by #8210

PR #8101, opened 2024-08-21
Use vite mode during packaging to customize the client static configuration

FirelightFlagboy:


FirelightFlagboy:
It's not a solution that I like, but as a temporary fix it's enough.

What's missing here is multiple `.env.[mode]` files for each possible `version.type`

NicoTuxx (approved):

PR #8177, opened 2024-08-26
Truncate block payloads in server logs

vxgmichel:
`block_create` requests and `block_read` replies have by far the largest payloads (up to 512K).

In debug mode, each call to those commands will add up to 1.5MB of content to the logs (due to ASCII representation).

Truncating those payloads allows to reduce the logs to a reasonable size.

### DIFF
--- a/server/parsec/asgi/rpc.py
+++ b/server/parsec/asgi/rpc.py
@@ -54,6 +54,46 @@ from parsec.logging import get_logger
 logger = get_logger()
 
 
+def block_repr(block: bytes, MAX_LOGGED_BLOCK_SIZE=64) -> str:
+    if len(block) <= MAX_LOGGED_BLOCK_SIZE:
+        return repr(block)
+    return f"{block[:MAX_LOGGED_BLOCK_SIZE]}... ({len(block)} bytes)"
+
+
+@dataclass
+class LoggedReq:
+    req: object
+
+    def __repr__(self) -> str:
+        # The `block_create` requests have by far the largest payloads (up to 512K).
+        # In debug mode, each call to those commands will add up to 1.5MB of content to the logs (due to ASCII representation).
+        # Truncating those payloads allows to reduce the logs to a reasonable size.
+        if isinstance(self.req, authenticated_cmds.latest.block_create.Req):
+            return f"BlockCreateReq {{ block: {block_repr(self.req.block)} }}"
+        else:
+            return repr(self.req)
+
+
+@dataclass
+class LoggedRep:
+    rep: object
+
+    def __repr__(self) -> str:
+        # The `block_read` replies have by far the largest payloads (up to 512K).
+        # In debug mode, each call to those commands will add up to 1.5MB of content to the logs (due to ASCII representation).
+        # Truncating those payloads allows to reduce the logs to a reasonable size.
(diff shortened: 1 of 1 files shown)


server/parsec/asgi/rpc.py
```diff
@@ -54,6 +54,32 @@
 logger = get_logger()
 
 
+def block_repr(block: bytes) -> str:
+    if len(block) <= 64:
+        return repr(block)
+    return f"{block[:64]}... ({len(block)} bytes)"
```
mmmarcos, line 60: Maybe make it more explicit?
```suggestion
MAX_LOGGED_BLOCK_SIZE = 64

def block_repr(block: bytes) -> str:
    """Truncate block repr to save space in logs"""
    if len(block) <= MAX_LOGGED_BLOCK_SIZE:
        return repr(block)
    return f"{block[:MAX_LOGGED_BLOCK_SIZE]}... ({len(block)} bytes)"
```

server/parsec/asgi/rpc.py
```diff
@@ -54,6 +54,32 @@
...
+
+
+@dataclass
+class LoggedReq:
+    req: object
+
+    def __repr__(self) -> str:
+        if not isinstance(self.req, authenticated_cmds.latest.block_create.Req):
+            return repr(self.req)
+        return f"BlockCreateReq {{ block: {block_repr(self.req.block)} }}"
```
mmmarcos, line 70: Maybe redundant but how about adding the comments you put in the PR description? 
Also I would reverse condition (matter of taste ^^)
```suggestion
        # block_create has a large payload (up to 512K)
        # truncate it to save space in logs 
        if isinstance(self.req, authenticated_cmds.latest.block_create.Req):
            return f"BlockCreateReq {{ block: {block_repr(self.req.block)} }}"
        return repr(self.req)
```

server/parsec/asgi/rpc.py
```diff
@@ -54,6 +54,32 @@
...
+
+
+@dataclass
+class LoggedRep:
+    rep: object
+
+    def __repr__(self) -> str:
+        if not isinstance(self.rep, authenticated_cmds.latest.block_read.RepOk):
+            return repr(self.rep)
+        return f"Ok {{ block: {block_repr(self.rep.block)}, key_index: {self.rep.key_index!r}, needed_realm_certificate_timestamp: {self.rep.needed_realm_certificate_timestamp!r} }}"
```
mmmarcos, line 80:
```suggestion
        # block_read has a large payload (up to 512K)
        # truncate it to save space in logs 
        if isinstance(self.rep, authenticated_cmds.latest.block_read.RepOk):
            return f"Ok {{ block: {block_repr(self.rep.block)}, key_index: {self.rep.key_index!r}, needed_realm_certificate_timestamp: {self.rep.needed_realm_certificate_timestamp!r} }}"
        return repr(self.rep)
```

mmmarcos (approved):
LGTM. Some suggestions to take or leave.

server/parsec/asgi/rpc.py
vxgmichel: 4556e098267f7fb6060d38048074439c1100a41f

server/parsec/asgi/rpc.py
vxgmichel: 4556e098267f7fb6060d38048074439c1100a41f

PR #8184, opened 2024-08-26
Add tests for organization_expired/author_revoked/author_frozen/invitation_already_used for each API command on the server

touilleMan:


server/tests/test_api.py
```diff
@@ -27,6 +27,8 @@ def test_each_cmd_req_rep_has_dedicated_test() -> None:
             for cmd_spec in cmd_specs:
                 cmd_name = cmd_spec["req"]["cmd"]
                 cmd_statuses = [rep_spec["status"] for rep_spec in cmd_spec["reps"]]
+                # Add a fake status corresponding to the tests of HTTP codes 460/461/462
```
mmmarcos, line 30:
```suggestion
                # Add fake status corresponding to common Parsec HTTP errors such as 404/460/461/462
                # The corresponding test function should test that all common errors are returned 
                # by the RPC API command
```

server/tests/test_api.py
```diff
@@ -27,6 +27,8 @@ def test_each_cmd_req_rep_has_dedicated_test() -> None:
...
+                cmd_statuses.append("http_46x")
```
mmmarcos, line 31: IMO `http_46x` is not very meaningful: it requires extra knowledge about API and does not say much about what is being tested. Besides, it is also testing `invitation_already_used` (410) and there are TODOs suggesting that `organization_not_found` (404) should also be tested. 

What about something like:
```suggestion
                cmd_statuses.append("http_common_errors")
```
```suggestion
                cmd_statuses.append("common_errors")
```
```suggestion
                cmd_statuses.append("common_parsec_errors")
```

server/tests/common/data.py
```diff
@@ -211,3 +215,167 @@ async def _alice_generated_realm_data(now: DateTime) -> None:
...
     return _alice_generated_realm_data
+
+
+Http46xTesterDoCallback = Callable[[], Coroutine[None, None, None]]
+Http46xTester = Callable[[Http46xTesterDoCallback], Coroutine[None, None, None]]
+
+
+# TODO: not sure how to test "organization_not_found"
+@pytest.fixture(params=("organization_expired", "author_revoked", "author_frozen"))
+async def http_46x_authenticated_tester(
```
mmmarcos, line 226: I'd rename this functions (See comment below in `test_api.py`). Also I would put family first, I think is clearer when using it as fixture.

Something like:
```suggestion
async def authenticated_common_errors_tester(
```

server/tests/common/data.py
```diff
@@ -211,3 +215,167 @@ async def _alice_generated_realm_data(now: DateTime) -> None:
...
+            assert err.rep.status_code == expected_http_status
+
+    yield _http_46x_authenticated_tester
+
+    assert tester_called
+
+
+# TODO: not sure how to test "organization_not_found" & "invitation_not_found"
+@pytest.fixture(params=("organization_expired", "invitation_already_used"))
+async def http_46x_invited_tester(
```
mmmarcos, line 307: I'd rename this functions (See comment below in `test_api.py`). Also I would put family first, I think is clearer when using it as fixture.

Something like:
```suggestion
async def invited_http_common_errors_tester(
```

server/tests/common/data.py
```diff
@@ -211,3 +215,167 @@ async def _alice_generated_realm_data(now: DateTime) -> None:
...
+            assert err.rep.status_code == expected_http_status, err
+
+    yield _http_46x_invited_tester
+
+    assert tester_called
+
+
+# TODO: not sure how to test "organization_not_found"
+@pytest.fixture(params=("organization_expired",))
+async def http_46x_anonymous_tester(
```
mmmarcos, line 352: I'd rename this functions (See comment below in `test_api.py`). Also I would put family first, I think is clearer when using it as fixture.

Something like:
```suggestion
async def anonymous_http_common_errors_tester(
```

server/tests/test_api.py
touilleMan: I would go with `http_common_errors` then

I think it's useful to mention http and error in the name, so that it hints the related code is in `server/asgi/rpc.py` (i.e. where the HTTP endpoints for the API are defined)

PR #7427, opened 2024-05-29
Improve debug/info log for RPC request in the server

touilleMan:
new format example:
image

and with log level debug:
image

Issue #5478, opened 2023-10-11
Add server log on handlers notify and listen


Landeers:
Add this logs on debug level on Parsec V2:

- all request payload
- notify event on handles

Add this logs on info level:

- listen event on handles

vxgmichel:
After investigation, debug logs already exist for request payloads:
https://github.com/Scille/parsec-cloud/blob/3c65dd012b2a645c3d2afc5f85bec26db4961f34/server/parsec/asgi/rpc.py#L601-L614

I'll check the server configuration.

vxgmichel:
Done here: https://github.com/Scille/parsec-cloud/pull/7427

PR #8191, opened 2024-08-27
Add logging to administration routes

vxgmichel:
Fix #https://github.com/Scille/parsec-cloud/issues/5478
Follow up after #7427


mmmarcos (approved):
LGTM

PR #8192, opened 2024-08-27
Fix missing prevent sync pattern handling in create file and create folder transaction helpers

### LINKED ISSUE #8068: [🐛 | Bug]: Creating file that match the prevent sync pattern make them being removed
### Parsec version tested on:

3.0.0-rc.1

### Platforms tested on:

Linux

### Bug description:

Creating (or renaming) a file/folder to make it match the prevent sync pattern make the said file to be removed (or at least it disappear from the list)

Reproduction steps:

1. Use the default prevent sync pattern.
2. Create a file named `foo.tmp`.
3. Observe the file still present until "synced".
4. After "sync" the file disappear. 

### Relevant output:

_No response_

vxgmichel:
Fix #8068

Many more tests are missing tho.

### DIFF
--- a/libparsec/crates/client/tests/unit/workspace/folder_transactions.rs
+++ b/libparsec/crates/client/tests/unit/workspace/folder_transactions.rs
@@ -159,7 +159,7 @@ async fn add_temporary_entry(
             builder.workspace_data_storage_fetch_workspace_vlob(
                 "alice@dev1",
                 wksp1_id,
-                Regex::from_regex_str("\\.tmp").unwrap(),
+                Regex::from_regex_str(r"\.tmp$").unwrap(),
             );
         } else {
             builder
@@ -172,7 +172,7 @@ async fn add_temporary_entry(
                 "alice@dev1",
                 wksp1_id,
                 wksp1_foo_id,
-                Regex::from_regex_str("\\.tmp").unwrap(),
+                Regex::from_regex_str(r"\.tmp$").unwrap(),
             );
         }
     })
--- a/libparsec/crates/client/tests/unit/workspace/folder_transactions.rs
+++ b/libparsec/crates/client/tests/unit/workspace/folder_transactions.rs
@@ -6,7 +6,10 @@ use libparsec_tests_fixtures::prelude::*;
 use libparsec_types::prelude::*;
 
 use super::utils::{ls, workspace_ops_factory};
-use crate::{workspace::MoveEntryMode, EventWorkspaceOpsOutboundSyncNeeded};
+use crate::{
+    workspace::{EntryStat, MoveEntryMode},
+    EventWorkspaceOpsOutboundSyncNeeded,
+};
 
 #[parsec_test(testbed = "minimal_client_ready")]
 async fn good(#[values(true, false)] root_level: bool, env: &TestbedEnv) {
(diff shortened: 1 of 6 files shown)


libparsec/crates/client/tests/unit/workspace/utils.rs
```diff
@@ -23,7 +23,7 @@ pub(crate) async fn workspace_ops_factory(
         workspace_storage_cache_size: WorkspaceStorageCacheSize::Default,
         proxy: ProxyConfig::default(),
         with_monitors: false,
-        prevent_sync_pattern: Regex::empty(),
+        prevent_sync_pattern: Regex::from_regex_str("\\.tmp").unwrap(),
```
touilleMan, line 26: Doesn't this matches `foo.tmp.bar` ?

(as a regex, I would expect something like `\\.tmp$` or `^.*\\.tmp$`)

libparsec/crates/client/tests/unit/workspace/folder_transactions.rs
```diff
@@ -134,3 +137,65 @@ async fn good(#[values(true, false)] root_level: bool, env: &TestbedEnv) {
...
+            p_assert_eq!(need_sync, false);
+        }
+        _ => panic!("Expected a folder"),
+    }
+
+    // Create a temporary directory
+    let target = "/test.tmp".parse().unwrap();
+    let child_id = ops.create_folder(target).await.unwrap();
+
+    // Check that the child is marked as needing sync
```
touilleMan, line 211: If the child is marked as need sync and a `EventWorkspaceOpsOutboundSyncNeeded` event has been dispatched, then the outbound sync monitor will upload this manifest on the server...

I didn't see any protection in `WorkspaceOps::outbound_sync` against this so I guess this is currently the behavior.

The "good" news is it should be invisible to the user (since the parent folder doesn't sync the entry name).


Regarding how to fix this, I see two solutions:
1. change `WorkspaceOps::outbound_sync`  to check for any confined entry in any parent. 
2. don't flag the confined entries as `need_sync`

At first glace I would say solution 2 seems better since:
- it's weird to have `need_sync` flags set for something that shouldn't be synced
- it should be simple to add (in the code responsible for scanning the database on prevent sync pattern chance) a comparison step between the local and the base to retrieve if a manifest has been locally modified
- on the contrary, retrieving all parents for a given entry in `WorkspaceOps::outbound_sync` means having to deal with concurrency issues and more locking (see the code of in `resolve_path.rs`, I'd rather avoid adding yet another kind of path resolver)

touilleMan (approved):

libparsec/crates/client/tests/unit/workspace/utils.rs
vxgmichel: 695cbccddb9bab1fa66450f1f87c282a3d2e7098

libparsec/crates/client/tests/unit/workspace/folder_transactions.rs
vxgmichel: Parsec v2 uses solution 1:
https://github.com/Scille/parsec-cloud/blob/d397c035d1369e498a48dea26837d93b7291dfe4/parsec/core/fs/workspacefs/sync_transactions.py#L344-L369

It's a better solution, because the new file/folder might get rename over time, and rename operations don't update the child manifest (there is no reason to). 

More broadly, the `need_sync` semantics means that the item is not up to date with the server. Whether a file is temporary or not has nothing to do with the file being up to date or not. For instance, you could have a file that has been synchronized (so it is up to date, hence not `need_sync`), but gets renamed into a temporary file.

This is why it's up to the synchronization engine to check whether the file is confined or not in order to allow or postpone its synchronization.

> on the contrary, retrieving all parents for a given entry in WorkspaceOps::outbound_sync means having to deal with concurrency issues

Most likely no. Like you noted, synchronizing a temporary file is not a big deal. So a simple `get_confinement_point` check before performing the synchronization is enough, as long as the synchronization engine can reschedule the entry when it gets renamed.

touilleMan: > It's a better solution, because the new file/folder might get renamed over time, and rename operations don't update the child manifest (there is no reason to).

That's a good point :+1: 

So solution 2 is just not possible (snif !) and we have no choice but to use solution 1

However I would another weakness of solution 1: at every start of the client, all confined entries are detected as need sync and hence the outbound sync monitor will try to sync them. It's not a big issue for now though (as it's not a bug, and the number of confined items is most likely low in a workspace), however we should open an issue about it just to track it, what do you think ?

> More broadly, the need_sync semantics means that the item is not up to date with the server

Then we need a better name than `need_sync` here (or at least some good docstring) !

Obviously the confinement system introduced three states (no local change | local change that don't need to be synced | local change that need to be synced) for a local manifest where there was only two before (i.e. no local change | local change that need to be synced)

On top of that, as you said above, those three states  cannot be encoded in the manifest itself since the fact the sync is not needed depends on the parent.

> So a simple get_confinement_point check before performing the synchronization is enough

`get_confinement_point` should in theory be atomic, and done in the same operation that the one fetching the manifest to synchronize. Hence my fear of having to introduce yet another resolver in `resolve_path.rs` (this one going from the entry ID, and going up to the root to find the top most confinement point).

I guess we could also use a non-atomic approach, but it may create weird behavior (think multiple rename/move while we are walking up the path, in the end we may endup with a path than never existed in the first place...). Then we have to identify if those behavior are acceptable (think for example that we detect the path doesn't lead to the root, we could consider the entry has been removed and hence we can remove from the database the corresponding vlob... is this acceptable if the path resolution is not atomic ?)

>  as long as the synchronization engine can reschedule the entry when it gets renamed.

That means the synchronization engine must detect a rename occurred and unconfined entries, I don't see how that's possible since the locally confined entry are not part of the base (so there is nothing to compare to).

I guess another solution is to send a dedicated signal instead.

The the scan would need to go through the entire tree of children of a given folder that was renamed out of confinement...


I  wonder if it would be better to instead add a dedicated `confined_by` field in the vlob table in the sqlite database:
- This would be set by the outbound sync (whenever it detects the current manifest that was in its queue for sync is in fact confined).
- On rename, a list of unconfined entry ID is transmitted along the manifest to store in sqlite. Hence sqlite can clean the `confined_by` field for any entry that used to be confined (at this point we would also need an event  to inform the sync monitor it should look in the local storage for new stuff to sync)

vxgmichel: I created the corresponding issue: https://github.com/Scille/parsec-cloud/issues/8198

PR #8222, opened 2024-09-02
Fix AppImage packaging for nightly build

FirelightFlagboy:
An extra `--` was added when generating the AppImage for nightly build, causing the flag `--nightly` being parsed as a target alongside `appimage`.

To fix that we factorize adding the `--` part if either _We make a nightly build_ or _Package an AppImage_.

### DIFF
--- a/.github/workflows/package-client.yml
+++ b/.github/workflows/package-client.yml
@@ -331,8 +331,9 @@ jobs:
         # we need to escape the flag another time by adding another `--`.
         run: >-
           npm run electron:release
-          ${{ matrix.platform == 'linux' && '-- appimage' || '' }}
-          ${{ inputs.nightly_build && '-- --nightly' || '' }}
+          ${{ (matrix.platform == 'linux' || inputs.nightly_build) && '--' }}
+          ${{ matrix.platform == 'linux' && 'appimage' || '' }}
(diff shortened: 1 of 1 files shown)


.github/workflows/package-client.yml
```diff
@@ -331,8 +331,9 @@ jobs:
         # we need to escape the flag another time by adding another `--`.
         run: >-
           npm run electron:release
-          ${{ matrix.platform == 'linux' && '-- appimage' || '' }}
-          ${{ inputs.nightly_build && '-- --nightly' || '' }}
+          ${{ (matrix.platform == 'linux' || inputs.nightly_build) && '--' }}
```
mmmarcos, line 334: Are spaces added automatically by GitHub?
```suggestion
          ${{ (matrix.platform == 'linux' || inputs.nightly_build) && '-- ' }}
```

FirelightFlagboy: Not from GitHub but from the yaml spec for [_folded multiline_ (`>`)]

- https://yaml-multiline.info/

[_folded multiline_ (`>`)]: https://yaml.org/spec/1.2.2/#folded-style

mmmarcos (approved):

PR #8227, opened 2024-09-03
Correct Outsider&realm interactions in libparsec&server

### LINKED ISSUE #8226: Iron out outsider and user update profile interactions
In Parsec a user can have 3 roles:
- STANDARD: can read/write/create/share workspace
- ADMIN: like STANDARD but can invite people and change other users' profile
- OUTSIDER: restricted profile that can only read/write workspace

## How it worked in Parsec v2

However there is a trick !

"Workspace" is a client side concept, on the server side we talk about "realm" which is a bunch of opaque data which common access control.
In practice that means the realm is used for storing workspace's data, but also for storing the user manifest.

Given each user have user manifest, that means an OUTSIDER must be able to create realms.
However user manifest are strictly personal, hence a realm created by an OUTSIDER cannot be shared with anyone (or put another way: an OUTSIDER is not allowed to share a realm even if he is OWNER)

## In Parsec v3 we allow user profile update

Changing a user's profile means anybody can become an OUTSIDER at any time.

### Client approach

Currently in the client's certificate integrity check, it is [invalid to have an OUTSIDER being MANAGER of any realm, or being OWNER of a realm with other users](https://github.com/Scille/parsec-cloud/blob/598bcc34d933087a1a3a75dff962dad77f725582/libparsec/crates/client/src/certif/add.rs#L1237-L1270)

This is not great since it means it is not always possible for an ADMIN user to change the profile of another user. And if we are in this case the error is pretty cryptic for the ADMIN without a clear way to solve it (e.g. "the user Bob cannot be turned into OUTSIDER since it has access MANAGER to realm 4f95c814239342c18eb2805d6490b73b, realm which you don't have access to so there is nothing you can do except try to find an OWNER and beg him... Ho ! and Bob is also part of 38 other realm with similar issues")

So obviously we should consider it is always possible to change a user's profile.

### Server approach

The server currently always allow the profile to be changed (this is probably the right approach as discussed above, but it means currently server allows things that will be considered inconsistent on client side...).

On top of that, the server currently doesn't check user profile in realm sharing/unsharing, this means:
- An OUTSIDER can share the realm he has created, this is obviously a bug
- An OUTSIDER can share a realm he has created or has been shared with him *before he become an OUTSIDER*. In this case things are more subtle since preventing the only OWNER-turned-OUTSIDER from sharing a workspace is kind of stuck

## A possible solution

The complexity here lies in the fact the rules of interaction between realm and OUTSIDER contains exceptions (e.g. "OUTSIDER cannot create realm... except for itself")

The good news is those rules come from the need to sync user manifest, however in Parsec v3 we no longer have user manifest (we have *local* user manifest, but they never get synchronized).

So what we can do:
- An OUTSIDER cannot use `realm_create` (so we should add a dedicated `author_not_allowed` error status)
- An OUTSIDER cannot be given role `MANAGER` or `OWNER`. This is only checked when giving the role, i.e. in `realm_share` (this is already implemented)
- All other realm_* commands only care about the author's realm role (this is already implemented)
- If we need to sync user manifest in the future (e.g. to store some configuration shared between the user's devices), we will introduce a new set of ad-hoc API commands instead of relying on the realms.

This way, an OUTSIDER is still able to operate normally on all the realm he used to be part of.

touilleMan:
close #8226 

New rules are:
- An OUTSIDER cannot create a realm.
- When sharing with an OUTSIDER, the roles MANAGER and OWNER cannot be given.
- A user becoming OUTSIDER retains his existing role in each realm he is part of.

### DIFF
--- a/libparsec/crates/client/src/client/mod.rs
+++ b/libparsec/crates/client/src/client/mod.rs
@@ -46,8 +46,8 @@ use crate::{
     event_bus::EventBus,
     monitors::{
         start_certif_poll_monitor, start_connection_monitor, start_server_config_monitor,
-        start_user_sync_monitor, start_workspaces_bootstrap_monitor,
-        start_workspaces_process_needs_monitor, start_workspaces_refresh_list_monitor, Monitor,
+        start_workspaces_bootstrap_monitor, start_workspaces_process_needs_monitor,
+        start_workspaces_refresh_list_monitor, Monitor,
     },
     user::UserOps,
 };
@@ -177,8 +177,20 @@ impl Client {
                 start_workspaces_refresh_list_monitor(client.event_bus.clone(), client.clone())
                     .await;
 
-            let user_sync_monitor =
-                start_user_sync_monitor(client.user_ops.clone(), client.event_bus.clone()).await;
+            // TODO: rework user manifest sync once needed
+            //
+            // User manifest currently doesn't need to be synced (i.e. we only
+            // use local user manifest, and rebuild it from scratch on every new
+            // client).
+            //
+            // The current user manifest sync system relies on realm to store the
+            // data, which is no longer allowed for users with OUTSIDER profile.
+            //
+            // Hence enabling the user sync monitor should have no effect, but it's
+            // safer to disable it entirely.
+
+            // let user_sync_monitor =
+            //     start_user_sync_monitor(client.user_ops.clone(), client.event_bus.clone()).await;
 
             let certif_poll_monitor = start_certif_poll_monitor(
                 client.certificates_ops.clone(),
@@ -198,7 +210,7 @@ impl Client {
             monitors.push(workspaces_bootstrap_monitor);
             monitors.push(workspaces_process_needs_monitor);
             monitors.push(workspaces_refresh_list_monitor);
-            monitors.push(user_sync_monitor);
+            // monitors.push(user_sync_monitor);
             monitors.push(certif_poll_monitor);
             monitors.push(server_config_monitor);
             monitors.push(connection_monitor);
--- a/libparsec/crates/client/src/monitors/user_sync.rs
+++ b/libparsec/crates/client/src/monitors/user_sync.rs
@@ -15,6 +15,10 @@ use crate::{
 
 const USER_SYNC_MONITOR_NAME: &str = "user_sync";
 
+// Local user manifest currently doesn't need to be synchronized, hence the
+// user sync monitor is never used (especially since it current implementation
+// is now incompatible with OUTSIDER users now that they can created realm).
+#[allow(unused)]
 pub(crate) async fn start_user_sync_monitor(
     user_ops: Arc<UserOps>,
     event_bus: EventBus,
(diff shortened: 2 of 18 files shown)


libparsec/crates/testbed/src/template/events.rs
```diff
@@ -1367,6 +1376,21 @@ impl TestbedEventShareRealm {
...
+
+            match role {
+                None | Some(RealmRole::Reader | RealmRole::Contributor) => (),
+                Some(RealmRole::Owner | RealmRole::Manager) => {
+                    match get_user_current_profile(&builder.events, user) {
+                        UserProfile::Admin | UserProfile::Standard => (),
+                        UserProfile::Outsider => {
+                            panic!(
+                                "User {} is Outsider, realm cannot be shared with him as {:?}",
+                                user, role
```
FirelightFlagboy, line 1388:
```suggestion
                Some(role @ RealmRole::Owner | RealmRole::Manager) => {
                    match get_user_current_profile(&builder.events, user) {
                        UserProfile::Admin | UserProfile::Standard => (),
                        UserProfile::Outsider => {
                            panic!(
                                "User {user} is Outsider, realm cannot be shared with him as {role}"
```

Use a binding to remove the debug print

vxgmichel (approved):
LGTM :+1:

libparsec/crates/client/src/client/mod.rs
```diff
@@ -151,8 +151,20 @@ impl Client {
                 start_workspaces_refresh_list_monitor(client.event_bus.clone(), client.clone())
                     .await;
 
-            let user_sync_monitor =
-                start_user_sync_monitor(client.user_ops.clone(), client.event_bus.clone()).await;
+            // TODO: rework user manifest sync once needed
```
AureliaDolo, line 154: Is there a related issue ?

libparsec/crates/client/src/monitors/user_sync.rs
```diff
@@ -88,6 +92,13 @@ fn task_future_factory(user_ops: Arc<UserOps>, event_bus: EventBus) -> impl Futu
                             event_bus.wait_server_online().await;
                             continue;
                         }
+                        UserSyncError::AuthorNotAllowed => {
+                            // OUTSIDER user are not allowed to create realm in Parsec v3.
+                            // In practice this is not an issue given we don't need to sync
+                            // user manifest for the moment.
+                            log::warn!("Rejected upload, we must have OUTSIDER profile",);
```
AureliaDolo, line 99:
```suggestion
                            log::warn!("Rejected upload: OUTSIDER is not allowed to create realm");
```

libparsec/crates/client/src/user/sync.rs
```diff
@@ -21,6 +21,8 @@ pub enum UserSyncError {
     Offline,
     #[error("Component has stopped")]
     Stopped,
+    #[error("Author")]
```
AureliaDolo, line 24:
```suggestion
    #[error("Author not allowed")]
```

libparsec/crates/client/src/monitors/user_sync.rs
touilleMan: :+1: done.

libparsec/crates/client/src/client/mod.rs
touilleMan: No, because we just don't need user manifest sync for now

In theory we should just remove entirely the user sync stuff, however it's a lot of code that would be needed again as soon as we want to share configuration between a user's devices (while slightly modified given we no longer can use the realm for this use).

So in the end I think it's just the pragmatical approach to let this code "as is" for the moment (and hence why there is no issue, there is nothing to do for the moment, not even removal ^^)

AureliaDolo: IMHO we could have an issue to 1/ link in all the places that are impacted 2/ centralize and document that discussion. But at the end of the day, it's project management boilerplate ^^

libparsec/crates/testbed/src/template/events.rs
touilleMan: :+1: done.

PR #8229, opened 2024-09-03
Remove fallback to the same env var for setting testbed URL

FirelightFlagboy:
The PR #8225 only made a basic search&replace, but we have some logique to remove since the `TESTBED_SERVER` env var was merged with `TESTBED_SERVER_URL`.

That the case when we start the testbed to rust tests where we would fallback to `TESTBED_SERVER_URL` if `TESTBED_SERVER` was not set.


AureliaDolo (approved):
:+1:

PR #8230, opened 2024-09-03
Perform inbound sync when listing files in CLI

FirelightFlagboy:


cli/src/commands/ls.rs
```diff
@@ -29,6 +29,19 @@ pub async fn ls(args: Ls) -> anyhow::Result<()> {
 
     let client = crate::utils::load_client(&config_dir, device, password_stdin).await?;
     let workspace = client.start_workspace(workspace_id).await?;
+    workspace.refresh_realm_checkpoint().await?;
+
+    loop {
```
touilleMan, line 34: You should add this behind a flag (typically `--fresh`), otherwise this command becomes unusable if we are offline

touilleMan (changes requested):

cli/src/commands/ls.rs
FirelightFlagboy: I've introduced the `--offline` flag instead

FirelightFlagboy:
Superseded by #8259

PR #8236, opened 2024-09-03
Fix testenv default device password

mmmarcos:


touilleMan (approved):

cli/src/testenv_utils.rs
```diff
@@ -21,7 +21,7 @@ use crate::{
 };
 
 pub const DEFAULT_ADMINISTRATION_TOKEN: &str = "s3cr3t";
-pub(crate) const DEFAULT_DEVICE_PASSWORD: &str = "test";
+pub(crate) const DEFAULT_DEVICE_PASSWORD: &str = "P@ssw0rd.";
```
FirelightFlagboy, line 24: I was happy with that password, why we change that ?

Issue #8179, opened 2024-08-26
Large amount of requests when synchronizing a file


vxgmichel:
While uploading a small file using the testbed server and the test environments (`run_testenv`), I noticed that many requests were performed by the client, about 4000: [logs.txt](https://github.com/user-attachments/files/16748272/logs.txt)

More specifically, 2 requests are being spammed:
- `vlob_update`, returning a `BadVlobVersion`
- `vlob_read_batch` returning `OK`

Here's a sample:
```
[...]

PR #8240, opened 2024-09-04
Fix server incorrectly dispatching Vlob event to it own author

touilleMan:


FirelightFlagboy (approved):

server/parsec/events.py
```diff
@@ -206,7 +206,12 @@ class EventVlob(BaseModel, ClientBroadcastableEvent):
 
     @override
     def is_event_for_client(self, client: RegisteredClient) -> bool:
-        return self.organization_id == client.organization_id and self.realm_id in client.realms
+        return (
+            self.organization_id == client.organization_id
+            # Skip the author of the event, given he obviously already knows about this vlob !
+            and self.author != client.device_id
+            and self.realm_id in client.realms
```
mmmarcos, line 213: I would move the longer explanation here, because this is the event that "breaks the rule".

BTW, for other events below, you commented "certificates are always integrated in the client by first polling them from the server to ensure no concurrently created certificate...". But for `VlobEvent` you say "**obviously** already knows about it !".

IMO it is not so "obvious" why the first problem does not apply to VlobEvents? Couldn't there be other VlobEvents that the client is not yet aware of? 
If VlobEvents are treated in a different way than certificates, maybe it is worth mentioning (feel free to update my suggestion below).

```suggestion
        # In general, the author of the event is notified about it because certificates 
        # are always integrated into the client by first polling them from the server 
        # (in order to ensure no concurrently-created certificate, of which the client 
        # may not be aware yet, is older than the certificate it has just created). 
        # EventVlob is a special case since the author of the event already knows
        # about the vlob and does not need to poll new events from the server.
        return (
            self.organization_id == client.organization_id
            and self.realm_id in client.realms
            # Skip the author of the event
            and self.author != client.device_id
```

server/parsec/events.py
```diff
@@ -234,6 +239,12 @@ class EventCommonCertificate(BaseModel, ClientBroadcastableEvent):
 
     @override
     def is_event_for_client(self, client: RegisteredClient) -> bool:
+        # Note, unlike for vlob, the certificate author still get a event about it.
+        #
+        # This is because certificates are always integrated in the client by first
+        # polling them from the server (in order to ensure no concurrently created
+        # certificate we are not yet aware of is older than the certificate we've
+        # just created).
```
mmmarcos, line 247:
```suggestion
        # Note, unlike for vlob, the certificate author still gets notified about the event.
        #
	    # This is because certificates are always integrated into the client by first 
	    # polling them from the server (in order to ensure no concurrently-created 
	    # certificate, of which the client may not be aware yet, is older than the 
	    # certificate it has just created).
```

server/parsec/events.py
```diff
@@ -255,6 +266,12 @@ class EventSequesterCertificate(BaseModel, ClientBroadcastableEvent):
...
```
mmmarcos, line 274:
```suggestion
        # Note, unlike for vlob, the certificate author still gets notified about the event.
        #
	    # This is because certificates are always integrated into the client by first 
	    # polling them from the server (in order to ensure no concurrently-created 
	    # certificate, of which the client may not be aware yet, is older than the 
	    # certificate it has just created).
```

server/parsec/events.py
```diff
@@ -277,6 +294,12 @@ class EventShamirRecoveryCertificate(BaseModel, ClientBroadcastableEvent):
...
```
mmmarcos, line 302:
```suggestion
        # Note, unlike for vlob, the certificate author still gets notified about the event.
        #
	    # This is because certificates are always integrated into the client by first 
	    # polling them from the server (in order to ensure no concurrently-created 
	    # certificate, of which the client may not be aware yet, is older than the 
	    # certificate it has just created).
```

server/parsec/events.py
```diff
@@ -303,6 +326,12 @@ class EventRealmCertificate(BaseModel, ClientBroadcastableEvent):
...
```
mmmarcos, line 334:
```suggestion
        # Note, unlike for vlob, the certificate author still gets notified about the event.
        #
	    # This is because certificates are always integrated into the client by first 
	    # polling them from the server (in order to ensure no concurrently-created 
	    # certificate, of which the client may not be aware yet, is older than the 
	    # certificate it has just created).
```

mmmarcos (approved):
This seems to fix #8179 (at least I am no longer able to reproduce it).

server/parsec/events.py
touilleMan: > IMO it is not so "obvious" why the first problem does not apply to VlobEvents? Couldn't there be other VlobEvents that the client is not yet aware of?

The difference comes from the fact certificates are only valid if received in a strict order (the order the server received  them) given their timestamp must be strictly growing.

On the other hand, multiple vlob can be upload concurrently without issues.

So a client that have a vlob needing sync can upload it in the server and mark it as sync=false in its sqlite database.

If it does the same approach for certificates (i.e. upload a certificates on the server, then store de certificate in its sqlite database), then it may receive from the server another certificates concurrently upload by another client that have a timestamp older (and hence an inconsistency is going to be detected).
So once the certificate uploaded, the client must poll the server to get all new certificates (among them being the certificate the client has just uploaded).

mmmarcos: Thanks for the explanation! So what about something like this:

```suggestion
        # In general, the author of the event is notified about it because certificates 
        # are always integrated into the client by first polling them from the server 
        # (in order to ensure no concurrently-created certificate, of which the client 
        # may not be aware yet, is older than the certificate it has just created). 
        #
        # For EventVlob, the client does not need to receive the vlob events in a strict
        # order (no need to poll new events from the server). So the client that uploads 
        # a vlob already knows about it and can mark it sync=false in its local sqlite db.
        # Therefore, the author of the event already knows about the vlob and does not
        # need to be notified about it.
        return (
            self.organization_id == client.organization_id
            and self.realm_id in client.realms
            # Skip the author of the event
            and self.author != client.device_id
```

FirelightFlagboy (approved):

PR #8241, opened 2024-09-04
Fix testbed init in GUI to use the in-memory database

touilleMan:


client/src/main.ts
```diff
@@ -105,8 +105,9 @@ async function setupApp(): Promise<void> {
     window.isDev = (): boolean => false;
     window.isLinux = (): boolean => isLinux;
 
-    if (configPath) {
-      window.getConfigDir = (): string => configPath;
+    if (testbedDiscriminantPath) {
+      window.getConfigDir = (): string => testbedDiscriminantPath;
+      window.getDataBaseDir = (): string => testbedDiscriminantPath;
```
FirelightFlagboy, line 110: @max, why those attribute are not constant ? because you want to be absolutely sure that are not overwritten ?

client/src/main.ts
```diff
@@ -145,6 +146,14 @@ async function setupApp(): Promise<void> {
       return [libparsec, nextStage];
     };
   } else if (import.meta.env.VITE_TESTBED_SERVER) {
```
FirelightFlagboy, line 148: For me it's strange that the creation of testbed org are fixed on compilation:

If you set `TESTBED_SERVER` during build of the client, this `if` will always be true (`import.meta.env.FOO` is replace by it's value during build).
Meaning you need to recompile if you do not want this behavior

FirelightFlagboy (approved):

PR #8253, opened 2024-09-04
Fix workspace sync monitor event queue refreshing due time

touilleMan:


FirelightFlagboy (approved):

libparsec/crates/client/src/workspace/store/file_updater.rs
```diff
@@ -168,6 +172,12 @@ impl FileUpdater {
         new_chunks: impl Iterator<Item = (ChunkID, &[u8])>,
         removed_chunks: impl Iterator<Item = ChunkID>,
     ) -> Result<(), UpdateFileManifestAndContinueError> {
+        // Sanity check to ensure the caller is not buggy
+        #[cfg(debug_assertions)]
+        {
+            assert_eq!(manifest.base.id, self.entry_id);
```
AureliaDolo, line 178: It's a nit, but why not use [`debug_assert_eq`](https://doc.rust-lang.org/std/macro.debug_assert_eq.html) directly ?

AureliaDolo (approved):

libparsec/crates/client/src/workspace/store/file_updater.rs
FirelightFlagboy: > debug_assert_eq! statements are only enabled in non optimized builds by default. An optimized build will not execute debug_assert_eq! statements unless -C debug-assertions is passed to the compiler.

Yes, we should use `debug_assert`

touilleMan: done.

PR #8254, opened 2024-09-05
[MS] Replace outsider to external

### LINKED ISSUE #8250: Update locale `outsider` to `external`

fabienSvtr:
## Describe your changes

Updated the profile `Outsider` to `External`

## Checklist

- [ ] Describe your changes
   
related to #8250

### DIFF
--- a/client/src/plugins/libparsec/definitions.ts
+++ b/client/src/plugins/libparsec/definitions.ts
@@ -60,7 +60,7 @@ export enum RealmRole {
 
 export enum UserProfile {
     Admin = 'UserProfileAdmin',
-    Outsider = 'UserProfileOutsider',
+    External = 'UserProfileExternal',
     Standard = 'UserProfileStandard',
 }
 export type DeviceID = string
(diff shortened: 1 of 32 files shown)


Max-7 (changes requested):
Not needed, only translations should be affected.

client/src/plugins/libparsec/definitions.ts
```diff
@@ -60,7 +60,7 @@ export enum RealmRole {
 
 export enum UserProfile {
     Admin = 'UserProfileAdmin',
-    Outsider = 'UserProfileOutsider',
+    External = 'UserProfileExternal',
     Standard = 'UserProfileStandard',
 }
```
FirelightFlagboy, line 65: You are searching for a bad time :sweat_smile: 

> Don't take the phrase above, too seriously. 

This file is auto generated meaning this change will be overwriteen sooner or later:

https://github.com/Scille/parsec-cloud/blob/94e90845923b9a265e15a8286e8353183fec6e7e/client/src/plugins/libparsec/definitions.ts#L3-L5

A proper way to change `UserProfile` enum on your side only would be to isolate `UserProfile` from libparsec and convert it to your `UserProfile`.

Two things about the refactoring:

- We cannot change this variant without requiring change on the Rust side.
- Updating the script to handle that case is not something trivial

fabienSvtr:
If the change is not defined in Parsec Rust, the RP will not be performed.

PR #8259, opened 2024-09-05
Add command to sync a workspace with the server

FirelightFlagboy:


AureliaDolo (approved):
I may be wrong, but in my understanding, there was at least some amount of sync included in some(all?) command ?

What is the use case of this command ?

FirelightFlagboy:
We discuss about it in https://scille.slack.com/archives/C07E7C8UKQT/p1725371267564449

@touilleMan prefer a sync command over a `--offline` flag to disable server communication during a specific command

PR #8269, opened 2024-09-06
Improve logging in the parsec server

vxgmichel:
More specifically:
- consistently log the different outcomes of RPC requests
- reduce the request id size to 4 characters
- log the different steps and outcomes in the SSE context
- bind the command name the the client context logger
- log base exceptions in debug mode, so cancellation or other base exception can appear

~It also adds debug logs to postgre transactions, although these could be dropped when the server crash investigation is over, up to debate.~ removed.

Also add a `--with-postgresql` option to the testbed server

### DIFF
--- a/server/parsec/asgi/administration.py
+++ b/server/parsec/asgi/administration.py
@@ -121,6 +121,9 @@ def log_request[**P, T: BaseModel | Response](
         except Exception as e:
             logger.error(f"{request.method} {request.url.path} exception", exc_info=e)
             raise
+        except BaseException as e:
+            logger.debug(f"{request.method} {request.url.path} base exception", exc_info=e)
+            raise
         if isinstance(result, Response):
             debug_extra = {
                 "status_code": result.status_code,
--- a/server/parsec/backend.py
+++ b/server/parsec/backend.py
@@ -2,6 +2,7 @@
 
 from __future__ import annotations
 
+import dataclasses
 from contextlib import asynccontextmanager
 from dataclasses import dataclass, field
 from typing import TYPE_CHECKING, Any, AsyncGenerator, Type, TypeAlias
@@ -38,6 +39,7 @@ from parsec.components.shamir import BaseShamirComponent
 from parsec.components.user import BaseUserComponent
 from parsec.components.vlob import BaseVlobComponent
 from parsec.config import BackendConfig
+from parsec.logging import get_logger
 from parsec.webhooks import WebhooksComponent
 
 if TYPE_CHECKING:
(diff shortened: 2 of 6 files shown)


server/parsec/cli/testbed.py
```diff
@@ -212,19 +218,29 @@ async def test_drop(raw_organization_id: str, request: Request) -> Response:
...
+async def testbed_backend_factory(
+    server_addr: ParsecAddr, with_postgresql: str | None
+) -> AsyncIterator[TestbedBackend]:
+    db_url = "MOCKED" if with_postgresql is None else with_postgresql
+    blockstore_config = (
+        MockedBlockStoreConfig() if with_postgresql is None else PostgreSQLBlockStoreConfig()
+    )
+    # Same as the defaults in `db_server_options`
+    db_min_connections = 1 if with_postgresql is None else 5
+    db_max_connections = 1 if with_postgresql is None else 7
```
mmmarcos, line 230: Would it be useful to log this config when running the testbed server?

Update: maybe actually all the BackendConfig should be logged at server startup (if it's not the case already)

server/parsec/client_context.py
```diff
@@ -119,7 +119,7 @@ class AuthenticatedClientContext:
 
     def __post_init__(self):
         self.logger = logger.bind(
-            request=uuid4().hex,
+            request=uuid4().hex[:4],
```
mmmarcos, line 122: Magic number? Maybe add a comment  
```suggestion
            # No need to have the full request uuid in logs
            request=uuid4().hex[:4],
```

mmmarcos (approved):
LGTM

server/parsec/cli/testbed.py
vxgmichel: 3141c68cb650bd27bac43690fbfc09a9cb84d1ee

server/parsec/client_context.py
vxgmichel: 9a697851672a9249cc9fa6db55d35fa2aa7f3558

server/parsec/asgi/administration.py
```diff
@@ -121,6 +121,9 @@ async def wrapped(*args: P.args, **kwargs: P.kwargs) -> T:
         except Exception as e:
             logger.error(f"{request.method} {request.url.path} exception", exc_info=e)
             raise
+        except BaseException as e:
```
touilleMan, line 124: so we are the pyo3 guys now ? :trollface:

server/parsec/client_context.py
```diff
@@ -118,8 +118,11 @@ class AuthenticatedClientContext:
     logger: ParsecBoundLogger = field(init=False)
 
     def __post_init__(self):
+        # Generate a request ID just for the logs
+        # It doesn't have to be as long as a UUID, 4 hex characters should be enough
+        request_id = (uuid4().hex[:4],)
```
touilleMan, line 123: Alternative solution: `random.randbytes(2).hex()`

server/parsec/backend.py
```diff
@@ -52,9 +54,12 @@
         TestbedEvent: TypeAlias = Any  # pyright: ignore[reportRedeclaration]
         TestbedTemplateContent: TypeAlias = Any  # pyright: ignore[reportRedeclaration]
 
+logger = get_logger()
+
 
 @asynccontextmanager
 async def backend_factory(config: BackendConfig) -> AsyncGenerator[Backend, None]:
+    logger.info("Backend configuration", **dataclasses.asdict(config))
```
touilleMan, line 62: I think it would be better to add a `__str__` method to `BackendConfig`, this would prevent risking leaking sensible values every time we add one to the config !

PR #8278, opened 2024-09-07
Improve `block_create`/`block_read` to release PostgreSQL connection during blockstore access

### LINKED ISSUE #8274: Possible deadlock in block create and block read
Block create and block read postgresql implementation keep an open transaction while accessing the blockstore.

This is bad in general since we shouldn't hold a transaction that itself locks rows while accessing an external service: this service might take time to respond, and this would block any other command trying to access these locked rows.

This is especially bad in postgersql blockstore mode, where a new connection is acquired while already holding an existing connection. If the max db connection is configured to be `N`, then `N` concurrent `block_read` or `block_create` requests might be able to start N connections. In this case, they will all bock trying to create a new connection for the blockstore.

This does happen in practice since blocks are typically uploaded in parallel.

touilleMan:
close #8274

### DIFF
--- a/server/parsec/components/postgresql/block_create.py
+++ b/server/parsec/components/postgresql/block_create.py
@@ -38,10 +38,9 @@ VALUES (
 """
 )
 
-
-# `block_create` being performance critical, we rely on a single big query to
-# fetches everything needed for access checks
-_q_create_fetch_data = Q(
+# `block_create` being performance critical, we rely on a single big query to both
+# lock `common`/`realm` topics and fetches everything needed for access checks.
+_q_create_fetch_data_and_lock_topics = Q(
     """
 WITH my_organization AS (
     SELECT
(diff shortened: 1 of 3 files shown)


server/parsec/components/postgresql/block_create.py
```diff
@@ -39,9 +39,9 @@
 )
 
 
-# `block_create` being performance critical, we rely on a single big query to both
-# lock `common`/`realm` topics and fetches everything needed for access checks
-_q_create_fetch_data_and_lock_topics = Q(
+# `block_create` being performance critical, we rely on a single big query to
+# fetches everything needed for access checks
+_q_create_fetch_data = Q(
```
vxgmichel, line 44: Why the renaming? the query still locks the topics right?

vxgmichel (approved):
LGTM :+1:

PR #8297, opened 2024-09-09
Improve wait peer methods during greeting attempt

vxgmichel:
Two improvements:
- Increase the number of retries when waiting for the other peer from 1 to 8
- Improve reactiveness by increasing the polling rate during the first attempts


libparsec/crates/client/src/invite/common.rs
```diff
@@ -2,6 +2,16 @@
...
 
+// The maximum number of attempts to wait for a peer to be ready.
+// Note that we only retry if the other peer has cancelled the ongoing
+// greeting attempt by starting a new one (which is done automatically)
+// This number should be high enough to cover the cases where the peer
+// cancels the greeting attempt while the nonces are being exchanged.
+// However, it shouldn't be too high to prevent the other peer from
+// being able re-roll the nonces too many times, for security reasons.
+// Any value between 4 and 32 should match those criteria.
+pub const WAIT_PEER_MAX_ATTEMPTS: usize = 8;
```
FirelightFlagboy, line 13: How did you determine that a value between 4 and 32 is OK for security reasons ?

libparsec/crates/client/src/invite/greeter.rs
```diff
@@ -422,32 +422,37 @@ struct BaseGreetInitialCtx {
...
+            // without requiring the front-end to deal with it.
+            if let Err(GreetInProgressError::GreetingAttemptCancelled {
+                origin: GreeterOrClaimer::Claimer,
+                reason: CancelledGreetingAttemptReason::AutomaticallyCancelled,
+                ..
+            }) = result
+            {
+                if attempt < WAIT_PEER_MAX_ATTEMPTS {
+                    continue;
+                }
```
FirelightFlagboy, line 440: The logic seems odd, if `attempt` is `>=` to `WAIT_PEER_MAX_ATTEMPTS` we advance to the next step

vxgmichel: Yes, so we can return the error at the next `?`.

libparsec/crates/client/src/invite/claimer.rs
```diff
@@ -304,31 +304,36 @@ struct BaseClaimInitialCtx {
...
-        {
-            result = self._do_wait_peer().await
+        // Loop over wait peer attempts
+        for attempt in 0.. {
+            let result = self._do_wait_peer().await;
+            // If the attempt was automatically cancelled by the other peer, try again (at most 8 times).
+            // Previous attempts are automatically cancelled when a new start greeting attempt is made.
+            // This way, the peers can synchronize themselves more easily during the wait-peer phase,
+            // without requiring the front-end to deal with it.
+            if let Err(ClaimInProgressError::GreetingAttemptCancelled {
```
touilleMan, line 314: The `if let Err(..)` followed by a `result?;` is hard to read (see @FirelightFlagboy comment)

You should instead a match:
```rust
match result {
  Ok(()) => (),
  // If the attempt was automatically cancelled by the other peer, try again (at most 8 times).
  // Previous attempts are automatically cancelled when a new start greeting attempt is made.
  // This way, the peers can synchronize themselves more easily during the wait-peer phase,
  // without requiring the front-end to deal with it.
  Err(ClaimInProgressError::GreetingAttemptCancelled { origin, reason, ..})  if attempt < WAIT_PEER_MAX_ATTEMPTS => continue,
  Err(err) => return err,
}
```

libparsec/crates/client/src/invite/common.rs
vxgmichel: Because that range is also achievable by human interaction (by manually restarting the process).

libparsec/crates/client/src/invite/claimer.rs
vxgmichel: 0e4e548241110f76284bc5f7fb82038f05eb1d43

libparsec/crates/client/src/invite/greeter.rs
vxgmichel: 0e4e548241110f76284bc5f7fb82038f05eb1d43

PR #8298, opened 2024-09-09
[MS] Added documentation link and e2e tests

Max-7:
Adds documentation link, adds e2e tests, replace some old cypress tests.

- [X] Describe your changes


client/src/locales/en-US.json
```diff
@@ -265,7 +266,8 @@
     "MenuPage": {
         "about": "About",
         "helpLink": "{signUrl}/contact",
-        "contactLink": "{signUrl}/contact"
+        "contactLink": "{signUrl}/contact",
+        "documentationLink": "https://docs.parsec.cloud/en/{version}"
```
FirelightFlagboy, line 270: If the version is passed as is, that will not work for nightly version since RTD only generate documentation for release tag

Max-7: Yeah but I don't think it matters. I pondered whether I should try and clean the version and decided against it. It adds complexity for something not really important. As long as it works when we release I think we're good.

PR #8304, opened 2024-09-09
Server fix migration 0003

touilleMan:


server/tests/test_migrations.py
```diff
@@ -49,9 +49,399 @@ def pg_psql(pg_cluster: Cluster) -> str:
...
+
+    # zack_u = UserCertificate(
+    #     author=coolorg.alice.device_id,
+    #     timestamp=DateTime.now(),
+    #     user_id=UserID.new(),
+    #     human_handle=HumanHandle("zack@invalid.com", "Zack"),
+    #     public_key=PrivateKey.generate().public_key,
+    #     algorithm=PrivateKeyAlgorithm.X25519_XSALSA20_POLY1305,
+    #     profile=UserProfile.STANDARD,
+    # )
```
vxgmichel, line 102: Debug artefacts?

vxgmichel (approved):
LGTM :+1:

server/tests/test_migrations.py
touilleMan: yup, removed (well moved to another file given I think it's still useful to have this at hand whenever we want to generate data or do copy/pasting)

PR #8316, opened 2024-09-10
Track caller in `Testbed::get_stuff` to provide that info in the panic message

FirelightFlagboy:


AureliaDolo (approved):
Is there a chance you still have the stack traces before and after that changes ?

FirelightFlagboy:
I don't have the stack trace, but the panic message stay the same (you just have `(caller ..)` added at the end)

PR #8317, opened 2024-09-10
Add Sentry support to Parsec server

### LINKED ISSUE #8299: [Server] Add Sentry support
Sentry should be configured server-side (it has already been added client-side in https://github.com/Scille/parsec-cloud/issues/6458)

In principle, set up is quite straightforward:
https://docs.sentry.io/platforms/python/integrations/fastapi/

However, legacy code should be investigated because it does not probably work (or maybe it was never used?). 
For example:
https://github.com/Scille/parsec-cloud/blob/49abce83813825e30255348e7e36db77bebfa9f4/server/parsec/logging.py#L147-L301

mmmarcos:
Closes #8299


server/parsec/logging.py
```diff
@@ -181,32 +171,52 @@ def _structlog_to_sentry_processor(
         if exc_info:
             sentry_event, _ = event_from_exception(
                 exc_info=exc_info,
-                client_options=sentry_client.options,
+                client_options=sentry_sdk.get_client().options,
```
vxgmichel, line 174: Are you sure it's ok to call `get_client()` for each log? just checking :)

vxgmichel (approved):
LGTM :+1: 

Have you been able to test it? I saw no errors in the `parsec3-server` project.

mmmarcos:
> LGTM 👍
> 
> Have you been able to test it? I saw no errors in the `parsec3-server` project.

Yes. I tested lots of combination (20+ issues/errors), but I removed them to start from a clean base... I should have left them for us to discuss 🤦‍♂️. I will generate a new one so you can see what it looks like.
In any case, I guess we can always tune it to our needs.

server/parsec/logging.py
mmmarcos: Not sure what you mean... are you suggesting that it might expensive to call it?

The client is retrieved from the current scope, and I do not see anything in their code that can be expensive. https://github.com/getsentry/sentry-python/blob/e6ca5a28dd139097ad7c8cb468e0b9232185b728/sentry_sdk/scope.py#L372

Also, keep in mind that this `get_client()` call is only executed when an Exception occurs AND the message contains the `exc_info`. That shouldn't happen too often, right?

PR #8322, opened 2024-09-11
Fix no field `entry_id` on type `&FileUpdater` when compiling in release

FirelightFlagboy:


touilleMan (approved):
use ```rust
#[cfg(debug_assertions)]
{
    assert_eq!(manifest.base.id, self.entry_id);
}```
format instead (consistent with the rest, and it seems using `debug_assert_eq` is risky given it uses `if cfg!` that only bite us when compiling in release (i.e. during new version release...)

PR #8341, opened 2024-09-11
Fix `remove_entry` not using `evolve_children_and_mark_updated` and add test.

### LINKED ISSUE #8306: Test `libparsec_client::workspace::transactions::remove_entry` with confined entries
Currently not using `evolve_children_and_mark_updated`

We should test if removing a confined entry update the confined entry list of the parent where the removed entry is removed from the list

The test should include using `WorkspaceOps::remove_entry(path)` and then check using the `WorkspaceStore` if the folder manifest as is local confined list udpated to not include the removed entry (also check before for the entry to be present in the list)

Test plan:

1. Create a confined file `foo.tmp`
2. Check the id of the confined file is present in parent confined list
3. Remove the confined file
4. Check the id of the confined file is not present in parent confined list

AureliaDolo:
Fix #8306


libparsec/crates/client/tests/unit/workspace/remove_entry.rs
```diff
@@ -288,3 +293,171 @@ async fn remove_folder_with_local_changes(
         EntryStat::Folder { need_sync, .. } if !need_sync
     );
 }
+
+#[parsec_test(testbed = "minimal_client_ready", with_server)]
+async fn remove_confined_entry(
```
FirelightFlagboy, line 298: We are testing the transaction, I would say to move the test to `libparsec/crates/client/tests/unit/workspace/folder_transactions.rs`.

That way to could use the helper defined here like `bootstrap_env`

libparsec/crates/client/tests/unit/workspace/remove_entry.rs
```diff
@@ -288,3 +293,171 @@ async fn remove_folder_with_local_changes(
...
+    // Parent must be updated
+    spy.assert_next(|e: &EventWorkspaceOpsOutboundSyncNeeded| {
+        p_assert_eq!(e.realm_id, wksp1_id);
+        p_assert_eq!(e.entry_id, parent_id);
+    });
+
+    // Check that the child is NOT in the parent's children
+    let children = ops.stat_folder_children_by_id(parent_id).await.unwrap();
+    assert!(children.is_empty());
+}
```
FirelightFlagboy, line 463: Also verify that the `LocalFolderManifest` is correctly updated by using `WorkspaceStore` directly.

What need to be added is the check for local_confinment_points is empty (since you've only added a single file that was confined).

Since you have access to the local manifest, you could also replace the stat_folder by directly checking the `children` field in the manifest.

FirelightFlagboy (changes requested):
I would say the test still need change since it didn't fail without a fix

libparsec/crates/client/tests/unit/workspace/folder_transactions.rs
```diff
@@ -305,6 +305,120 @@ async fn add_confined_entry(
...
+        ArcLocalChildManifest::Folder(folder_manifest) => {
+            assert_eq!(
+                folder_manifest.local_confinement_points,
+                HashSet::from([confined_child_id])
+            )
+        }
+        ArcLocalChildManifest::File(_) => {
+            panic!("parent expected to be a folder")
+        }
+    }
```
FirelightFlagboy, line 391:
```rust
    let ArcLocalChildManifest::Folder(folder) = ops.store.get_manifest(parent_id).await.unwrap()
    else {
        panic!("Expected a folder");
    };
```

Remainder that this syntax exist

vxgmichel (approved):
LGTM :+1:

Issue #8129, opened 2024-08-23
[🐛 | Bug]: Allow to go back during organization creation


mmmarcos:
During organization creation there are a couple of screens where options need to be selected.

Currently it is not possible to "go back" one page to check or change one of the entered values. The user needs to either:
- cancel the whole process and start again. 
- go through until the last screen where a confirmation is displayed 

We should add a "Previous" button next to "Next" button. This applies to _all_ the screens in this process. 

![image](https://github.com/user-attachments/assets/4cb4848d-f862-44c5-81bf-4ab3822dda79)

PR #8342, opened 2024-09-11
[MS] Added back buttons after server choice on organization creation

Ironicbay:
Closes https://github.com/Scille/parsec-cloud/issues/8129

### DIFF
--- a/client/src/views/client-area/BmsLogin.vue
+++ b/client/src/views/client-area/BmsLogin.vue
@@ -107,8 +107,18 @@
           </ion-text>
         </div>
 
-        <!-- login button -->
+        <!-- back and login buttons -->
         <div class="saas-login-button">
+          <ion-text
+            v-show="showBackButton"
+            :disabled="querying"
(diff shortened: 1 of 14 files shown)


client/src/views/client-area/BmsLogin.vue
```diff
@@ -191,11 +201,13 @@ import { Env } from '@/services/environment';
 const props = defineProps<{
   email?: string;
   hideHeader?: boolean;
+  hideBackButton?: boolean;
```
Max-7, line 204: As a general rule (no need to change anything here, just a note for future PRs), when adding optional parameters/props, if they are not set, the behavior of the function/component should not change. Here, if the new prop is not set, the behavior changes (breaking the API), so if you forgot to change a use of `BmsLogin` somewhere, now we've got a bug.

We don't have a lot of usage of BmsLogin so it's very probably fine here, just a general rule.

Max-7 (changes requested):

Max-7 (approved):

Issue #8276, opened 2024-09-06
Confinement points are not properly set in `stat_folder` and `stat_entry_by_id`


vxgmichel:
https://github.com/Scille/parsec-cloud/blob/597a0d9e0e9db234a88225f4527ba57d2bd24c76/libparsec/crates/client/src/workspace/transactions/stat_entry.rs#L108-L109

https://github.com/Scille/parsec-cloud/blob/597a0d9e0e9db234a88225f4527ba57d2bd24c76/libparsec/crates/client/src/workspace/transactions/stat_entry.rs#L118-L119

https://github.com/Scille/parsec-cloud/blob/597a0d9e0e9db234a88225f4527ba57d2bd24c76/libparsec/crates/client/src/workspace/transactions/read_folder.rs#L98-L99

vxgmichel:
Fixed by #8343

PR #8343, opened 2024-09-11
Add `retrieve_path_from_id` and fix missing confinement points

vxgmichel:
Fix issue #8276

TODO: 
- [x] Improve naming
- [x] Add tests

### DIFF
--- a/libparsec/crates/client/src/workspace/store/resolve_path.rs
+++ b/libparsec/crates/client/src/workspace/store/resolve_path.rs
@@ -36,15 +36,6 @@ impl From<PathConfinementPoint> for Option<VlobID> {
     }
 }
 
-impl From<Option<VlobID>> for PathConfinementPoint {
-    fn from(val: Option<VlobID>) -> Self {
-        match val {
-            None => PathConfinementPoint::NotConfined,
-            Some(id) => PathConfinementPoint::Confined(id),
-        }
-    }
-}
-
 #[derive(Debug, thiserror::Error)]
 pub(crate) enum ResolvePathError {
     #[error("Cannot reach the server")]
--- a/libparsec/crates/client/src/workspace/store/resolve_path.rs
+++ b/libparsec/crates/client/src/workspace/store/resolve_path.rs
@@ -23,14 +23,14 @@ use super::{
 /// that takes precedence.
 #[derive(Debug, Clone, Copy, PartialEq, Eq)]
 pub(crate) enum PathConfinementPoint {
-    None,
+    NotConfined,
     Confined(VlobID),
 }
 
 impl From<PathConfinementPoint> for Option<VlobID> {
     fn from(val: PathConfinementPoint) -> Self {
         match val {
-            PathConfinementPoint::None => None,
+            PathConfinementPoint::NotConfined => None,
             PathConfinementPoint::Confined(id) => Some(id),
         }
     }
@@ -39,7 +39,7 @@ impl From<PathConfinementPoint> for Option<VlobID> {
 impl From<Option<VlobID>> for PathConfinementPoint {
     fn from(val: Option<VlobID>) -> Self {
         match val {
-            None => PathConfinementPoint::None,
+            None => PathConfinementPoint::NotConfined,
             Some(id) => PathConfinementPoint::Confined(id),
         }
     }
@@ -251,7 +251,7 @@ fn cache_only_path_resolution(
                     StepKind::Root => {
                         let manifest =
                             ArcLocalChildManifest::Folder(cache.manifests.root_manifest().clone());
-                        let confinement = PathConfinementPoint::None;
+                        let confinement = PathConfinementPoint::NotConfined;
                         return CacheOnlyPathResolutionOutcome::Done {
                             manifest,
                             confinement,
@@ -263,7 +263,10 @@ fn cache_only_path_resolution(
         };
 
         let (parent_manifest, parent_confinement) = match &last_step {
-            StepKind::Root => (cache.manifests.root_manifest(), PathConfinementPoint::None),
+            StepKind::Root => (
+                cache.manifests.root_manifest(),
+                PathConfinementPoint::NotConfined,
+            ),
             StepKind::Child {
                 manifest,
                 confinement,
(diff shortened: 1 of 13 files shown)


libparsec/crates/client/src/workspace/store/resolve_path.rs
```diff
@@ -646,3 +664,108 @@ pub(crate) async fn resolve_path_for_reparenting(
...
+/// they could get outdated by the time you use them. This is fine however,
+/// since this is typically used for providing information to the user
+/// (e.g. for file and folder stats)
+pub(crate) async fn retrieve_path_from_id(
+    store: &super::WorkspaceStore,
+    entry_id: VlobID,
+) -> Result<(FsPath, PathConfinementPoint), ResolvePathError> {
+    // Initialize the results
+    let mut parts = Vec::new();
+    let mut confinement: Option<VlobID> = None;
```
FirelightFlagboy, line 679:
```suggestion
    let mut confinement = PathConfinementPoint::None;
```

Directly use the specific type

libparsec/crates/client/src/workspace/store/resolve_path.rs
```diff
@@ -27,6 +27,24 @@ pub(crate) enum PathConfinementPoint {
     Confined(VlobID),
 }
 
+impl From<PathConfinementPoint> for Option<VlobID> {
+    fn from(val: PathConfinementPoint) -> Self {
+        match val {
+            PathConfinementPoint::None => None,
```
FirelightFlagboy, line 33: Could it be possible to rename the variant `None` to `NotConfined` ?

libparsec/crates/client/src/workspace/store/resolve_path.rs
vxgmichel: done

libparsec/crates/client/src/workspace/store/resolve_path.rs
vxgmichel: done

libparsec/crates/client/src/workspace/store/resolve_path.rs
```diff
@@ -646,3 +658,106 @@ pub(crate) async fn resolve_path_for_reparenting(
...
+            super::GetManifestError::EntryNotFound => ResolvePathError::EntryNotFound,
+            super::GetManifestError::NoRealmAccess => ResolvePathError::NoRealmAccess,
+            super::GetManifestError::InvalidKeysBundle(err) => {
+                ResolvePathError::InvalidKeysBundle(err)
+            }
+            super::GetManifestError::InvalidCertificate(err) => {
+                ResolvePathError::InvalidCertificate(err)
+            }
+            super::GetManifestError::InvalidManifest(err) => ResolvePathError::InvalidManifest(err),
+            super::GetManifestError::Internal(err) => err.context("cannot retrieve path").into(),
```
AureliaDolo, line 692: nit: You could maybe implement `From<GetManifestError> for ResolvePathError` and use it below too. I'm not sure about how to do with the internal variant though.

libparsec/crates/client/tests/unit/workspace/stat_entry.rs
```diff
@@ -121,6 +121,161 @@ async fn stat_entry(#[values(true, false)] local_cache: bool, env: &TestbedEnv)
...
 }
 
+async fn stat_entry_by_id_helper(
+    ops: &WorkspaceOps,
+    entry_id: VlobID,
+    kind: &'_ str,
+    expected_confinement_point: Option<VlobID>,
+) -> (EntryStat, Option<VlobID>) {
+    match kind {
+        "vanilla" => (
```
AureliaDolo, line 131: In this context, vanilla means we care about the confinement point but we don't know it ?

libparsec/crates/client/src/workspace/store/resolve_path.rs
```diff
@@ -23,10 +23,19 @@ use super::{
 /// that takes precedence.
 #[derive(Debug, Clone, Copy, PartialEq, Eq)]
 pub(crate) enum PathConfinementPoint {
-    None,
+    NotConfined,
```
AureliaDolo, line 26: Following our discussion about using that instead of Option<VlobID>, I'd rather use PathConfinementPoint anywhere it's possible (without to much hassle). 

But that will wait :hourglass:

AureliaDolo (approved):

libparsec/crates/client/src/workspace/store/resolve_path.rs
```diff
@@ -646,3 +658,106 @@ pub(crate) async fn resolve_path_for_reparenting(
...
+        // Update the confinement point result
+        // Give priority to the top most confinement point
+        confinement = if parent_manifest
+            .local_confinement_points
+            .contains(&current_entry_id)
+        {
+            PathConfinementPoint::Confined(current_parent_id)
+        } else {
+            confinement
+        };
```
FirelightFlagboy, line 753: `Top most` is the furthest file from root (`/`)?

e.g.: `/foo/test.tmp` (furthest) vs `/foo.tmp` (nearest)

FirelightFlagboy:
```suggestion
        // Update the confinement point result
        // Give priority to the root most confinement point
        confinement = if parent_manifest
            .local_confinement_points
            .contains(&current_entry_id)
        {
            PathConfinementPoint::Confined(current_parent_id)
        } else {
            confinement
        };
```

Currently, it's prioritizing setting confinement to the entry nearest to the root (If that not wanted, adding an `or_else` should do the trick).

I propose replace `top` to `root` in the comment

AureliaDolo:
```mermaid
flowchart TD
    root --> folder1tmp--> folder2tmp  --> file
```

In that case the path confinement point would be folder1tmp ?

vxgmichel: > Currently, it's prioritizing setting confinement to the entry nearest to the root (If that not wanted, adding an or_else should do the trick).

Yes that's the behavior we want (first version of this PR got it wrong)

vxgmichel: > In that case the path confinement point would be folder1tmp ?

It would be root even, because it's the folder that tagged `folder1tmp` as confined

vxgmichel: > I propose replace top to root in the comment

Yes good point

libparsec/crates/client/tests/unit/workspace/stat_entry.rs
vxgmichel: 571c530cf9aacf14e2379652fc3b6187774559f0

libparsec/crates/client/src/workspace/store/resolve_path.rs
vxgmichel: 571c530cf9aacf14e2379652fc3b6187774559f0

PR #8347, opened 2024-09-12
Add handle for client event callback

FirelightFlagboy:


libparsec/src/invite.rs
```diff
@@ -111,7 +111,7 @@ pub async fn bootstrap_organization(
     sequester_authority_verify_key: Option<SequesterVerifyKeyDer>,
 ) -> Result<AvailableDevice, BootstrapOrganizationError> {
     let config: Arc<libparsec_client::ClientConfig> = config.into();
-    let events_plugged = OnEventCallbackPlugged::new(on_event_callback);
+    let events_plugged = OnEventCallbackPlugged::new(handle, on_event_callback);
```
touilleMan, line 114: Always pass `0` for handle here (i.e. "always invalid handle"), and add a comment to explain it is the pragmatical approach

libparsec/src/client.rs
```diff
@@ -120,7 +122,7 @@ pub async fn client_start(
 
     // 3) Actually start the client
 
-    let on_event = OnEventCallbackPlugged::new(on_event_callback);
+    let on_event = OnEventCallbackPlugged::new(handle, on_event_callback);
```
touilleMan, line 125:
```suggestion
    let on_event = OnEventCallbackPlugged::new(initializing.handle(), on_event_callback);
```

`initializing.handle` filed is private (and owner shouldn't be able to modify it), so you should add a `InitializingGuard::handle(&self) ->`Handle ` getter

touilleMan (approved):

libparsec/src/client.rs
FirelightFlagboy: done

libparsec/src/invite.rs
FirelightFlagboy: done

bindings/electron/src/meths.rs
```diff
@@ -7451,22 +7454,25 @@ fn claimer_retrieve_info(mut cx: FunctionContext) -> JsResult<JsPromise> {
...
-                        .apply::<JsValue, _>(&mut cx)?;
-                }
-                Ok(())
-            });
-        }) as std::sync::Arc<dyn Fn(libparsec::ClientEvent) + Send + Sync>
+        std::sync::Arc::new(
+            move |handle: libparsec::Handle, event: libparsec::ClientEvent| {
+                let callback2 = callback.clone();
+                callback.channel.send(move |mut cx| {
+                    // TODO: log an error instead of panic ? (it is a bit harsh to crash
```
AureliaDolo, line 7461: Is there an issue or something out there to maybe remember to do that some day ?

AureliaDolo (approved):
I'm not well versed into typescript but otherwise LGTM

PR #8348, opened 2024-09-12
[CLI] Group server commands under the same group

FirelightFlagboy:


cli/src/commands/admin/create_organization.rs
```diff
@@ -8,7 +8,7 @@ use crate::utils::*;
 
 crate::clap_parser_with_shared_opts_builder!(
     #[with = addr, token]
-    pub struct CreateOrganization {
+    pub struct Args {
```
AureliaDolo, line 11: So that's why the organization group contains only the bootstrap command ?

AureliaDolo (approved):

cli/src/main.rs
```diff
@@ -22,6 +22,10 @@ struct Arg {
 
 #[derive(Subcommand)]
 enum Command {
+    /// Contains subcommands related to admin.
```
AureliaDolo, line 25:
```suggestion
    /// Contains subcommands related to server and organization admin.
```

To make clear that some organization commands are here, not in the organization group

PR #8350, opened 2024-09-12
[CLI] Group organization commands under the same group 

FirelightFlagboy:


AureliaDolo (approved):
That's the only command related to an organization ? :open_mouth:

FirelightFlagboy:
Not really, CF its parent #8348, some organization related are grouped in `admin` group

AureliaDolo:
Thanks, so I'm not sure about that group: are they some other organization commands that we could add that don't require the admin token ?

PR #8353, opened 2024-09-12
Fix prevent sync pattern in tests

### LINKED ISSUE #8016: Incorrect (but with no impact on the tests !) pattern in LocalFolderManifest tests
see e.g.
https://github.com/Scille/parsec-cloud/blob/40d92a3e49fd5c7e95678a979ad5fd77c9f0c9cb/libparsec/crates/types/tests/unit/local_folder_manifest.rs#L693

The issue is the regex `.tmp` matches stuff like `xtmp`, which is obviously not what we want here (we only want to match names ending with `.tmp`)

The solution is simply to correct the pattern into `\.tmp` or to use `Regex::from_glob_pattern` instead of `Regex::from_regex_str`

FirelightFlagboy:
Closes #8016

### DIFF
--- a/libparsec/crates/types/tests/unit/local_folder_manifest.rs
+++ b/libparsec/crates/types/tests/unit/local_folder_manifest.rs
@@ -12,6 +12,8 @@ use libparsec_tests_lite::prelude::*;
 
 type AliceLocalFolderManifest = Box<dyn FnOnce(&Device) -> (&'static [u8], LocalFolderManifest)>;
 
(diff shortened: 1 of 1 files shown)


AureliaDolo (approved):
Nit: we could have a `const PREVENT_SYNC_PATTERN_TMP: &str = r"\.tmp$".

Kinda like `PREVENT_SYNC_PATTERN_EMPTY_PATTERN` maybe ?

touilleMan (approved):

AureliaDolo (approved):

PR #8385, opened 2024-09-16
[MS] Updated how external links are opened + fixed changelog link

### LINKED ISSUE #8378: [🐛 | Bug]: Missing `v` prefix to documentation link
### Parsec version tested on:

3.0.0-rc.7

### Platforms tested on:

Linux

### Bug description:

On the device list page (home), the documentation link use an invalid link <https://docs.parsec.cloud/en/3.0.0-rc.7/> where is should be <https://docs.parsec.cloud/en/v3.0.0-rc.7/>

> Note the added `v` before `3.0.0`

### Relevant output:

_No response_

Max-7:
Closes #8378 

- [X] Describe your changes

### DIFF
--- a/client/src/services/environment.ts
+++ b/client/src/services/environment.ts
@@ -1,6 +1,6 @@
 // Parsec Cloud (https://parsec.cloud) Copyright (c) BUSL-1.1 2016-present Scille SAS
 
-import { EnvironmentType } from 'megashark-lib';
+import { EnvironmentType, I18n } from 'megashark-lib';
 
 export const APP_VERSION = __APP_VERSION__;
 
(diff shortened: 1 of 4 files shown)


client/src/services/environment.ts
```diff
@@ -60,10 +60,45 @@ function getTrialServers(): Array<string> {
   return DEFAULT_TRIAL_SERVERS;
 }
 
+const CLEAN_APP_VERSION = `${APP_VERSION.replace('+dev', '')}`;
```
FirelightFlagboy, line 63: Note: For nightly build, `+{local}` is set to a short commit id (it replace `dev`).

I'm think If it would be simple to handle that part with `misc/version_updater.py` :thinking:, we already have a function (`hide_patch_version`) that can remove the patch version of python, we could have something similar for `CLEAN_APP_VERSION`

Ironicbay (approved):

client/src/services/environment.ts
Max-7: I update the cleaning part slightly to remove anything after the `+` (if present). Probably good enough.

PR #8390, opened 2024-09-16
Log the parsec version in the backend factory

vxgmichel:
This is useful to make sure we deployed the correct version. 

This PR also moves the project versions to a dedicated toml file (this commit is optional, although useful IMO)


FirelightFlagboy (approved):

AureliaDolo (approved):

.github/workflows/_parse_version.yml
```diff
@@ -84,6 +84,12 @@ jobs:
           ref: ${{ inputs.commit_sha }}
         timeout-minutes: 5
 
+      - name: Install python
+        uses: actions/setup-python@29a37be0a3d3e8bf5bc1eb19cd0502922f5b312a  # pin v5.2.0
+        id: setup-python
+        with:
+          python-version: 3.12
+
```
FirelightFlagboy, line 92: Why do you need to install python in `Parse version` job ? the version provided by the system is not good ?

vxgmichel: We want the same version of python to be used everywhere to avoid unexpected issues.

Typically, `tomllib` has been introduced in python 3.11

touilleMan (approved):

PR #8415, opened 2024-09-17
[MS] Reduce 'done' items to 10 in import menu and rename it to 'last …

### LINKED ISSUE #8393: [🐛 | Bug]: GUI breaks when `FileOperationMenu` tab with many items is open
### Parsec version tested on:

3.0.0-rc7

### Platforms tested on:

MacOS

### Bug description:

GUI becomes very laggy and navigating through the folder tree bugs out (fails to find paths, incorrect breadcrumbs)

To replicate:
- Import many files, in my case a folder with a few hundred files and some subfolders, which I import twice or thrice to build a layered folder tree
- Leave the `FileOperationMenu` open on the `Done` tab (which displays the few hundred elements imported)
- Enjoy

### Relevant output:

_No response_

Ironicbay:
Closes #8393

### DIFF
--- a/client/src/views/files/FileOperationMenu.vue
+++ b/client/src/views/files/FileOperationMenu.vue
@@ -40,8 +40,8 @@
         @click="currentTab = Tabs.Done"
       >
         <div class="item-container">
-          <ion-text>{{ $msTranslate('FoldersPage.ImportFile.tabs.done') }}</ion-text>
-          <span class="text-counter">{{ doneItems.length > 99 ? '+99' : doneItems.length }}</span>
+          <ion-text>{{ $msTranslate('FoldersPage.ImportFile.tabs.lastUploaded') }}</ion-text>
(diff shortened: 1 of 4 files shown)


client/src/views/files/FileOperationMenu.vue
```diff
@@ -232,24 +197,34 @@ function getFileOperationComponent(item: OperationItem): Component | undefined {
 }
 
 function updateImportState(id: string, state: FileOperationState, stateData?: StateData): void {
-  const operation = fileOperations.value.find((op) => op.data.id === id);
+  const operation = inProgressItems.value.find((op) => op.data.id === id);
```
Max-7, line 200: You could use `findIndex` right here and use it to get the matching operation, just to avoid going through the list againt when removing the item.

Max-7 (changes requested):

Max-7 (approved):

Max-7 (approved):

PR #8418, opened 2024-09-17
Lower `EVENT_VLOB_MAX_BLOB_SIZE` 

### LINKED ISSUE #8410: InvalidParameterValueError: payload string too long
Sentry Issue: [PARSEC3-SERVER-K](https://scille.sentry.io/issues/5862061221/?referrer=github_integration)

```
InvalidParameterValueError: payload string too long
(13 additional frame(s) were not displayed)
...
  File "asyncio/base_events.py", line 685, in run_until_complete
    return future.result()
  File "contextlib.py", line 231, in __aexit__
    await self.gen.athrow(value)
  File "contextlib.py", line 231, in __aexit__
    await self.gen.athrow(value)
  File "contextlib.py", line 231, in __aexit__
    await self.gen.athrow(value)
  File "asyncpg/protocol/protocol.pyx", line 207, in bind_execute

InterfaceError: pool is closed
(4 additional frame(s) were not displayed)
...
  File "contextlib.py", line 210, in __aenter__
    return await anext(self.gen)

SSE session exception
```

vxgmichel:
... so the payload doesn't exceed 8000 bytes

Also remove the useless b64 conversion in `send/parse_signal`

Fix #8410

### DIFF
--- a/server/parsec/events.py
+++ b/server/parsec/events.py
@@ -31,8 +31,8 @@ if TYPE_CHECKING:
 # can be passed with the event (on top of that we keep the events in cache for SSE
 # last-event-id, so better avoid too much pressure of the RAM)
 # We keep a large margin with PostgreSQL limit given the event also contains additional
-# fields and must be encoded in base64 (which alone adds ~33% in size).
-EVENT_VLOB_MAX_BLOB_SIZE = 4096
+# fields and the blob is hex encoded.
+EVENT_VLOB_MAX_BLOB_SIZE = 1024
 
 
 OrganizationIDField = Annotated[
--- a/.cspell/custom-words.txt
+++ b/.cspell/custom-words.txt
@@ -358,6 +358,7 @@ makensis
 manylinux
 mapbox
 mapboxgl
+maxsplit
 mediaservice
 megashark
 memfs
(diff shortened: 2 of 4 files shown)


server/parsec/events.py
```diff
@@ -31,8 +31,8 @@
 # can be passed with the event (on top of that we keep the events in cache for SSE
 # last-event-id, so better avoid too much pressure of the RAM)
 # We keep a large margin with PostgreSQL limit given the event also contains additional
-# fields and must be encoded in base64 (which alone adds ~33% in size).
-EVENT_VLOB_MAX_BLOB_SIZE = 4096
+# fields and the blob is hex encoded.
+EVENT_VLOB_MAX_BLOB_SIZE = 1024
```
FirelightFlagboy, line 35: Why 1024 and not 2048 ?

mmmarcos (approved):
LGTM

AureliaDolo (approved):

server/parsec/events.py
vxgmichel: Honestly, I took an extra margin so we don't run again into this kind of issue.

Adding a test would have been better, but as @mmmarcos mentioned it, we should simply get rid of the vlob data in the vlob event: https://github.com/Scille/parsec-cloud/issues/8420

PR #8449, opened 2024-09-19
Bump version `3.0.1+dev` to `3.0.1-a.0`

FirelightFlagboy:
Following a discussion with @vxgmichel, we choose to set the current dev version as alpha:

The semver speck indicate that build metadata in version (everything after `+`, so `+dev`) is not used when comparing versions.
We don't want that `3.0.1 == 3.0.1+dev` so we choose to indicate that the version is a dev version using the prerelease part of the version.
We don't not use `-dev` as pre-release since the literal part of the pre-release is ascii sorted, that would result in `3.0.1-beta.5 < 3.0.1-dev.0`.

To make this change perennial, the release script was updated:

When building a release, the script will increment the pre-release part of the version (or set it to alpha 0 by default) corresponding to the new in-dev version.
This require additional changes to allow the release script to build a release using the current version.


mmmarcos (approved):
LGTM

PR #8452, opened 2024-09-19
Bump Rust toolchain

FirelightFlagboy:


Cargo.toml
```diff
@@ -63,6 +63,12 @@ repository = "https://github.com/Scille/parsec-cloud"
 undocumented_unsafe_blocks = "deny"
 unwrap_used = "deny"
 
+[workspace.lints.rust]
+unexpected_cfgs = { level = "warn", check-cfg = [
+    'cfg(use_pure_rust_but_dirty_zstd)',
+    'cfg(skip_fuse_atime_option)',
+] }
```
AureliaDolo, line 70: I don't quite get what this setting does ? It warns if any of these feature is used ?

AureliaDolo (approved):
I didn't know about `inspect` it's nice :sparkles:

Cargo.toml
FirelightFlagboy: It's to "register" our custom cfg

PR #8458, opened 2024-09-19
Manage RequireGreaterTimestamp errors

### LINKED ISSUE #8270: `RequireGreaterTimestamp` replies are not handled correctly
Here's how all `RequireGreaterTimestamp` replies are handled:

https://github.com/Scille/parsec-cloud/blob/11c6d575d63e897ac0b0fad3320343a9cd811a9c/libparsec/crates/client/src/certif/user_revoke.rs#L71-L81

That means that if a client clock is behind, and there's a newer certificate, that client will be asked to try again. With the code above, the client will try again using the `strictly_greater_than` provided timestamp. However, and as the name implies, client is expected to try again with a timestamp **strictly greater** than the one being provided (i.e, the timestamp of the most recent certificate).

For that reason, the next client request will fail as well, starting a full speed polling loop until the client clock finally passes the `strictly_greater_than` timestamp. However, another certificate might be uploaded by a non-drifting client in the meantime, causing the polling loop to keep going, until it reaches the timestamp of that new certificate.

For reference, this is how those timestamps were managed in parsec v2:
```python
        timestamp = self.device.timestamp()
        if timestamp_greater_than is not None:
            timestamp = max(
                timestamp, timestamp_greater_than.add(microseconds=ROLE_CERTIFICATE_STAMP_AHEAD_US)
            )
```

Note that an arbitrary value is added to the `timestamp_greater_than` timestamp.

Those offset were defined as such:
```python
# This value is used to increment the timestamp provided by the backend
# when a manifest restamping is required. This value should be kept small
# compared to the certificate stamp ahead value, so the certificate updates have
# priority over manifest updates.
MANIFEST_STAMP_AHEAD_US = 100_000  # microseconds, or 0.1 seconds

# This value is used to increment the timestamp provided by the backend
# when a certificate restamping is required. This value should be kept big
# compared to the manifest stamp ahead value, so the certificate updates have
# priority over manifest updates.
ROLE_CERTIFICATE_STAMP_AHEAD_US = 500_000  # microseconds, or 0.5 seconds

# Archiving doesn't have to use a value as large 0.5 seconds
ARCHIVING_CERTIFICATE_STAMP_AHEAD_US = 1_000  # microseconds, or 1 milliseconds
```

AureliaDolo:
Fix #8270 and adds a regression test.

### DIFF
--- a/libparsec/crates/client/src/certif/mod.rs
+++ b/libparsec/crates/client/src/certif/mod.rs
@@ -113,9 +113,11 @@ impl From<GreaterTimestampOffset> for i64 {
     }
 }
 
-/// ops to have the time provider
-/// offset depending on the type of certificate
-/// strictly_greater_than, the time expected by the server
+/// Return the timestamp to be used when the server requires a greater timestamp
+///
+/// The timestamp to be used will be whichever is greater between:
+/// - now (obtained from `time_provider`)
+/// - `strictly_greater_than` (expected by the server) + `offset` (depending on certificate type)
 pub(crate) fn manage_require_greater_timestamp(
     time_provider: &TimeProvider,
     offset: GreaterTimestampOffset,
--- a/libparsec/crates/client/src/certif/mod.rs
+++ b/libparsec/crates/client/src/certif/mod.rs
@@ -51,6 +51,74 @@ use libparsec_types::prelude::*;
 
 use crate::{event_bus::EventBus, ClientConfig};
 
+// The following values define an offset to be used when processing
+// a require greater timestamp error. They are different to give priority
+// to some kind of certificates (ex: a certificate that revokes a user
+// should have more chances to be accepted than a vlob update one)
+// The base offset is the average network round trip time to compensate
+// the first round trip.
+
+// An optimization could be to measure the real round trip time for each request.
+// The issue with that is that we tackling two distinct issues
+// 1. the client clock is lagging behind
+// 2. the client connection is degraded
+// The current approach is only compensating for case 1.
+// Case 2 would benefit from a offset depending on the real round trip time,
+// but if we base our estimate of the round trip time on the expected timestamp
+// provided by the server, case 2 would benefit from an unjustified priority, as
+// it would pass for a client with a very degraded connection (within the limits of the ballpark though).
+// That issue would not arise if we monitor the round trip time only from the client point of view.
+
+/// This value is used to increment the timestamp provided by the backend
+/// when a manifest restamping is required. This value should be kept small
+/// compared to the certificate stamp ahead value, so the certificate updates have
+/// priority over manifest updates.
+/// # in microseconds, or 0.1 seconds, 100 ms (average network round trip time)
+pub const MANIFEST_STAMP_AHEAD_US: i64 = 100_000;
+
+/// This value is used to increment the timestamp provided by the backend
+/// when a certificate restamping is required. This value should be kept big
+/// compared to the manifest stamp ahead value, so the certificate updates have
+/// priority over manifest updates.
+///  # microseconds, or 0.5 seconds
+pub const ROLE_CERTIFICATE_STAMP_AHEAD_US: i64 = 500_000;
+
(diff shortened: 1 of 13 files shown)


vxgmichel (approved):
LGTM :+1:

libparsec/crates/client/src/certif/mod.rs
```diff
@@ -51,6 +51,82 @@ use libparsec_types::prelude::*;
...
+            GreaterTimestampOffset::User => USER_CERTIFICATE_STAMP_AHEAD_US,
+            GreaterTimestampOffset::Archive => ARCHIVING_CERTIFICATE_STAMP_AHEAD_US,
+            GreaterTimestampOffset::Realm => REALM_CERTIFICATE_STAMP_AHEAD_US,
+        }
+    }
+}
+
+/// ops to have the time provider
+/// offset depending on the type of certificate
+/// strictly_greater_than, the time expected by the server
```
mmmarcos, line 118: Minor suggestion to use structure from https://doc.rust-lang.org/rustdoc/how-to-write-documentation.html

```suggestion
/// Return the timestamp to be used when the server requires a greater timestamp
///
/// The timestamp to be used will be whichever is greater between:
/// - now (obtained from `time_provider`)
/// - `strictly_greater_than` (expected by the server) + `offset` (depending on certificate type)
```

libparsec/crates/client/src/certif/mod.rs
```diff
@@ -51,6 +51,82 @@ use libparsec_types::prelude::*;
...
+pub(crate) fn manage_require_greater_timestamp(
```
mmmarcos, line 119: Manage seems too generic, but I cannot think of a much better name, maybe one of: 
```suggestion
pub(crate) fn compute_greater_timestamp(
```
```suggestion
pub(crate) fn obtain_greater_timestamp(
```
```suggestion
pub(crate) fn determine_greater_timestamp(
```
```suggestion
pub(crate) fn greater_timestamp(
```

mmmarcos (commented):
Kudos for this PR and the helpful docs!

mmmarcos (changes requested):
Forgot to mention: this PR fixes bug #8270, a newsfragment should be added to briefly describe the fixed bug.

libparsec/crates/client/src/certif/mod.rs
```diff
@@ -51,6 +51,82 @@ use libparsec_types::prelude::*;
...
+// Case 2 would benefit from a offset depending on the real round trip time,
+// but if we base our estimate of the round trip time on the expected timestamp
+// provided by the server, case 2 would benefit from an unjustified priority, as
+// it would pass for a client with a very degraded connection (within the limits of the ballpark though).
+// That issue would not arise if we monitor the round trip time only from the client point of view.
+
+/// This value is used to increment the timestamp provided by the backend
+/// when a manifest restamping is required. This value should be kept small
+/// compared to the certificate stamp ahead value, so the certificate updates have
+/// priority over manifest updates.
```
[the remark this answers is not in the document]
AureliaDolo: No because if the offset is small there is a greater chance that something else was acknowledged by the server.
For example, if a server answers with requireGreaterTimestamp(T0) to client A, then client B sends a new certificate at T0+x, and the server accepts it. When client A will answer if the offset is smaller or equal to x, the corresponding certificate will be refused, whereas if it's greater it will be accepted.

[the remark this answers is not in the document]
AureliaDolo: I'll add that to the doc, it's a bit counter intuitive.

FirelightFlagboy (approved):

newsfragments/8270.bugfix.rst
```diff
@@ -0,0 +1,2 @@
+Make sure than when the server asks the client for a greater timestamp,
+the client offsets the received timestamp.
```
mmmarcos, line 2:
```suggestion
Make sure than when the server asks for a strictly greater timestamp,
the client offsets the timestamp sent by the server.
```

mmmarcos (approved):

PR #8467, opened 2024-09-20
Add command to display history for current release

mmmarcos:
Also simplify news fragment check in the CI (https://github.com/Scille/parsec-cloud/pull/8467/commits/48898831914c81f6fb65606f44c266614acbae43): with the previous checks it was impossible to update the README file or update an existing news fragment.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -180,9 +180,7 @@ jobs:
             )
           && steps.newsfragment-have-changed.outputs.newsfragments == 'true'
         run: |
-          whereis git
-          git fetch origin ${{ github.base_ref }}
-          python .github/scripts/check_newsfragments.py --base=${{ github.base_ref }}
+          python .github/scripts/check_newsfragments.py
         timeout-minutes: 5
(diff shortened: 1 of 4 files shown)


.github/workflows/ci.yml
```diff
@@ -182,7 +182,7 @@ jobs:
         run: |
           whereis git
           git fetch origin ${{ github.base_ref }}
```
FirelightFlagboy, line 184: Remove `git fetch` since it's not used by the script

FirelightFlagboy (commented):
This simplify the script, but I see that we will be checking for each newsfragment not only the one added/modified by a PR.

I would say it's not too costly to check for each fragments every time. but if it is we could perform what was done before in a single command:

```
git diff --name-only `git merge-base origin/master HEAD` -- newsfragment
```

FirelightFlagboy (approved):

.github/workflows/ci.yml
mmmarcos: done

PR #8470, opened 2024-09-20
Use base64 for event encoding in server's PostgreSQL NOTIFY

touilleMan:


server/parsec/events.py
```diff
@@ -97,12 +98,12 @@ def hex_bytes_validator(val: object) -> bytes:
     elif isinstance(val, bytearray):
         return bytes(val)
     elif isinstance(val, str):
-        return bytes.fromhex(val)
+        return b64decode(val)
     raise ValidationError()
 
 
 def hex_bytes_serializer(val: bytes) -> str:
```
FirelightFlagboy, line 105: This function no longer encode bytes into hexa, it should be rename. It would suggest:

- `bytes_serializer`: If the method of serialization is not important, we would remove it from the name.
- `b64_bytes_serializer`: Like the current name, to indicate the method of the serialization

touilleMan: Good catch ! I've renamed to `base64_bytes_serializer` & `Base64Bytes`

FirelightFlagboy (approved):

PR #8483, opened 2024-09-23
[MS] Adds accept tos modal

### LINKED ISSUE #7633: Implement T&Cs acceptance client-side
- On the enrollment screen, a checkbox should be added, indicating "I accept the CGU". => UI 0.5h.j
- The enrollment process is blocked as long as the checkbox is unchecked (i.e. "Next" button is disabled).
- On this same screen, a link must be added to the CGU (e.g. a static file added to the Parsec server). => UI 0.5h.j
  - CGU are server configured, not organization
- Parsec client retrieves link configuration from backend. => 2h.j
- Displays the checkbox only if a link to the CGU is present (backend link configuration). => UI 1h.j
- Information of date of acceptance should be sent to the server
- Add link on Menu for CGU to be able to read it at anytime. => UI 1h.j

Max-7:
Closes #7633 
Depends on #8471 

To test:
1. Don't forget to recompile, `python make.py i && python make.py ei`
2. Start the testbed
3. Open electron. On the device page, note the org name
4. Use the little script `python update_tos.py <OrgName>`
5. Log in to the org

Don't forget to change the locale to check if the link is updated, you can add additional locales to update_tos.py, change their name, use `en_US` or `en-us` or `en` or some kind of variations to check if the GUI manages to find the most appropriate version.

```python
# upload_tos.py. It needs requests.
import requests
import sys

response = requests.patch(
    f"http://localhost:6770/administration/organizations/{sys.argv[1]}",
    headers={"Authorization": f"Bearer s3cr3t"},
    json={"tos": {
        "en_US": "https://www.youtube.com/watch?v=dQw4w9WgXcQ",
        "fr_BE": "https://www.youtube.com/watch?v=hEmODTcKJmE",
    }},
)

print(response.status_code, response.content)
```

- [X] Describe your changes


FirelightFlagboy:
Stacked PR should be done by setting the appropriate base (master vs eula), that was not the case

client/src/locales/en-US.json
```diff
@@ -203,7 +203,18 @@
             "tosLink": "https://my.parsec.cloud/assets/pdf/Parsec-CGVU-en.pdf",
             "and": "and ",
             "privacyPolicy": "Privacy Policy",
-            "privacyPolicyLink": "https://parsec.cloud/en/politique-de-confidentialite/"
+            "privacyPolicyLink": "https://parsec.cloud/en/politique-de-confidentialite/",
+            "update": {
+                "title": "New terms of service",
+                "subtitle": "We have updated our terms of service. Please read and accept them to continue.",
+                "tosButton": "Terms Of Service",
+                "url": "https://my.parsec.cloud/assets/pdf/Parsec-CGVU-fr.pdf",
```
FirelightFlagboy, line 211: Placeholder ?

mmmarcos (changes requested):
The "normal" use cases seem to work fine: 
1. ToS defined in server, user logins, ToS dialog is displayed, user accepts ToS, login (OK)
2. User logout, then login, no ToS dialog is displayed (OK)
3. User logout, ToS updates in server, user login --> same as 1 (OK)

## Issues

### Wrong login status

- User is logged out
- TOS are updated in the server
- User tries to login, accepts the new TOS
- Login status is not updated (**KO**)
image
- Go to home, shows user is logged-in
image
- Return to org, login status is still offline (**KO**)
- Logout, then login, -- login status is online (OK)

### App state if logged-in during ToS update

If the user is logged-in and the TOS are updated on the server. The application behaves unexpectedly: 
- At first, nothing seems to change from the user perspective (menus are accesible, workspaces can be created, folders can be opened, files can be imported, etc.).
- However, some actions like attempting to sharing a workspace, result in an unrelated error: 
image
This is very disturbing for the user, who is most certainly not aware of TOS update, and the apps seems to fail wirhout any reason.

What should be done in this case it should be carefully discussed: 
- should the user be forced to log out? not great, since it could be importing files or editing a document...
- should we ask for TOS acceptance, but let them continue using until next login? asking for acceptance could potentially also block a current user operation
- should we ignore TOS update until next login? this seems the more natural and user-friendly alternative. Although it may need changes on the current TOS logic.

client/src/locales/en-US.json
```diff
@@ -203,7 +203,18 @@
...
+                "checkbox": "I have read and accept the new terms.",
```
mmmarcos, line 212:
```suggestion
                "checkbox": "I have read and accept the terms of service.",
```

client/src/locales/fr-FR.json
```diff
@@ -203,7 +203,18 @@
             "tosLink": "https://my.parsec.cloud/assets/pdf/Parsec-CGVU-fr.pdf",
             "and": "et la ",
             "privacyPolicy": "Politique de Confidentialité",
-            "privacyPolicyLink": "https://parsec.cloud/fr/politique-de-confidentialite/"
+            "privacyPolicyLink": "https://parsec.cloud/fr/politique-de-confidentialite/",
+            "update": {
+                "title": "Nouvelles conditions générales",
```
mmmarcos, line 208: To be consistent with button?
```suggestion
                "title": "Nouvelles conditions générales de vente et d'utilisation",
```

client/src/locales/fr-FR.json
```diff
@@ -203,7 +203,18 @@
...
+                "subtitle": "Nous avons mis à jour les conditions générales d'utilisation. Veuillez les consulter avant de les accepter.",
```
mmmarcos, line 209:
```suggestion
                "subtitle": "Nous avons mis à jour les conditions générales d'utilisation. Veuillez les consulter et les accepter pour continuer.",
```

mmmarcos (commented):
Minor suggestions

client/src/locales/fr-FR.json
mmmarcos: We'll stick with "condition générales ~de vente~ d'utilisation"

newsfragments/7633.feature.rst
```diff
@@ -0,0 +1 @@
+Added TOS acceptance on login if the server provides them
```
mmmarcos, line 1: Terms of Services (ToS) can now be defined in the server. Users are requested to accept them on first connection and whenever they are updated.

server/parsec/components/events.py
```diff
@@ -256,60 +257,72 @@ def __init__(self, config: BackendConfig, event_bus: EventBus):
...
+                for registered in self._registered_clients.values():
+                    if (
+                        registered.organization_id == event.organization_id
+                        and registered.user_id == event.user_id
+                    ):
+                        registered.cancel_scope.cancel()
+
+            case EventOrganizationTosUpdated():
+                # All user in the organization must re-accept the TOS before being
+                # able to use the server again.
```
mmmarcos, line 317:
```suggestion
                # All users in the organization must re-accept the TOS before being
                # able to communicate with the server again.
```

client/src/locales/en-US.json
mmmarcos: Not sure what this URL text is about, but it should probably not point to Parsec CGVU.

client/src/locales/fr-FR.json
```diff
@@ -203,7 +203,18 @@
             "tosLink": "https://my.parsec.cloud/assets/pdf/Parsec-CGVU-fr.pdf",
             "and": "et la ",
             "privacyPolicy": "Politique de Confidentialité",
-            "privacyPolicyLink": "https://parsec.cloud/fr/politique-de-confidentialite/"
+            "privacyPolicyLink": "https://parsec.cloud/fr/politique-de-confidentialite/",
+            "update": {
+                "title": "Nouvelles conditions générales d'utilisation",
+                "subtitle": "Nous avons mis à jour les conditions générales d'utilisation. Veuillez les consulter et les accepter pour continuer. Vous serez automatiquement déconnecté si vous les refusez.",
+                "tosButton": "Conditions Générales d'Utilisation",
+                "url": "https://my.parsec.cloud/assets/pdf/Parsec-CGVU-fr.pdf",
```
mmmarcos, line 211: Remove?

client/src/parsec/terms_of_service.ts
```diff
@@ -0,0 +1,40 @@
...
+    return await libparsec.clientAcceptTos(handle, updatedOn.toSeconds() as any as DateTime);
+  } else {
+    return { ok: true, value: null };
+  }
+}
+
+export async function getTOS(): Promise<Result<Tos, ClientGetTosError>> {
+  const handle = getParsecHandle();
+
+  if (handle !== null && !needsMocks()) {
```
mmmarcos, line 22: I'm not familiar with the client code: No handle means there is an error in the application? I suppose it should not happen, but could there be an error that makes `handle == null`? In that case, what is presented to the user?

Why not handling it like:
```ts
if (handle !== null) {
    // error case
}

if (!needsMocks()){
    // normal case
} else {
    // mock case
}
```

client/src/parsec/terms_of_service.ts
```diff
@@ -0,0 +1,40 @@
...
+      result.value.updatedOn = DateTime.fromSeconds(result.value.updatedOn as any as number);
+    }
+    return result;
+  } else {
+    return {
+      ok: true,
+      value: {
+        perLocaleUrls: new Map<string, string>([
+          ['en-US', 'https://www.youtube.com/watch?v=dQw4w9WgXcQ'],
+          ['fr-FR', 'https://www.youtube.com/watch?v=hEmODTcKJmE'],
```
mmmarcos, line 34: If there is even the slightest chance that this ends-up being shown to the user (see above comment), I would not put these URLs in the source code. We can just put some invalid placeholder URLs.

mmmarcos (commented):
Some comments otherwise LGTM (I've mostly reviewed changes to the `server/` directory).

client/src/parsec/terms_of_service.ts
Max-7: I like the thought of rickrolling the user in case an error occurs. But yeah, you're right, I'll replace those.

client/src/parsec/terms_of_service.ts
Max-7: No handle shouldn't be possible unless the user messes with the URL, which is not possible with Electron (only in web mode).
I'll have to do a big pass on everything related to how we mock data and how to handle these kind of errors once we got the web bindings.

libparsec/crates/client/src/event_bus.rs
```diff
@@ -218,6 +218,15 @@ impl_events!(
...
     Offline,
+    /// This event is fired by `Client::accept_tos` when the TOS acceptance has been
+    /// successful (hence it's most likely the server will now accept our next connection
+    /// attempt).
+    ///
+    /// This event is used to signify to the connection monitor that it should try
+    /// to connect to the server without delay. This is needed since, when offline,
+    /// the connection monitor has a backoff mechanism to avoid hammering the server
+    /// with connection attempts.
+    ShouldRetryConnectionNow,
```
AureliaDolo, line 229: It's an idea for later. but could that event be useful in some other places too ?

AureliaDolo (approved):
I've looked mostly at the libparsec part and it's looking good.

client/src/parsec/terms_of_service.ts
mmmarcos: Me too, and this is probably not a big issue with SaaS users. But I was thinking specially in the context of Parsec as a security solution for government agencies or in the context of product certifications.

fabienSvtr (approved):

libparsec/crates/client/src/client/tos.rs
```diff
@@ -87,7 +88,14 @@ pub async fn accept_tos(
     let request = tos_cmds::latest::tos_accept::Req { tos_updated_on };
     let rep = client_ops.cmds.send(request).await?;
     match rep {
-        tos_cmds::v4::tos_accept::Rep::Ok => Ok(()),
+        tos_cmds::v4::tos_accept::Rep::Ok => {
+            // If the TOS acception was required, it most likely means the connection
```
mmmarcos, line 92:
```suggestion
            // If TOS acceptance was required, it most likely means the connection
```

libparsec/crates/client/src/event_bus.rs
```diff
@@ -218,6 +218,15 @@ impl_events!(
     ///
     /// The SSE connection with the server has been lost.
     Offline,
+    /// This event is fired by `Client::accept_tos` when the TOS acceptance has been
+    /// successful (hence it's most likely the server will now accept our next connection
+    /// attempt).
```
mmmarcos, line 223:
```suggestion
    /// This event is fired by `Client::accept_tos` when the TOS have been accepted.
    /// It's most likely the server will now accept the next connection attempt.
```

mmmarcos (approved):
LGTM (server part)

client/electron/src/setup.ts
```diff
@@ -200,6 +200,10 @@ export class ElectronCapacitorApp {
     await this.updater?.checkForUpdates();
   }
 
+  addAuthorizedURL(url: string): void {
+    ALLOWED_URL_LIST.push(url);
```
FirelightFlagboy, line 204: Should check if url is already present in allowed list

PR #8495, opened 2024-09-24
Make parsec use $HOME/.config as the base configuration directory for the snap package

### LINKED ISSUE #8464: [🐛 | Bug]: If snap user remove the snap they lost their devices
### Parsec version tested on:

3.0.0

### Platforms tested on:

Linux

### Bug description:

Repro: 

> [!Caution]
> Be mindfull when trying to reproduce the bug, you will lose your devices!

1. Install parsec using snap:
   ```shell
   snap install parsec --channel=v3
   ```

2. Create a trial organization
3. Remove parsec
   ```shell
   snap remove parsec
   ```
4. Re-install parsec using snap (see step 1)
5. Observe your trial device no longer exist

## Explanation

On snap, the devices are stored under `~/snap/parsec/{version}/.config/parsec3/libparsec/devices`.

Upon removal of the snap, the folder `~/snap/parsec/{version}` will be removed alongside it.

### Relevant output:

_No response_

FirelightFlagboy:
Also migrate previous devices located in snap user data

Closes #8464

### DIFF
--- a/client/electron/scripts/snap_migrate_devices_from_snap_user_data.sh
+++ b/client/electron/scripts/snap_migrate_devices_from_snap_user_data.sh
@@ -1,0 +1,24 @@
+#!/bin/bash
+# This scripts copies device files (.keys) from the legacy device directory (based on $HOME/snap)
+# to the new device directory (based on $HOME/.config)
+
+config_dir=${PARSEC_BASE_CONFIG_DIR:-$HOME/.config}
+device_dir=$config_dir/parsec3/libparsec/devices
+mkdir -pv "$device_dir"
+# Iterate over device directory from all installed parsec snaps
+for old_device_dir in $(find "$HOME"/snap -type f -name '*.keys' 2>/dev/null | grep parsec | xargs dirname | sort | uniq); do
+    sentinel_file=$old_device_dir/.parsec-migrated-devices
+    # Skip directory if it contains the same devices that are stored in the sentinel file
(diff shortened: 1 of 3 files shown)


client/electron/scripts/snap_migrate_devices_from_snap_user_data.sh
```diff
@@ -0,0 +1,13 @@
+#!/bin/bash
+
```
mmmarcos, line 2:
```suggestion
# This scripts copies device files (.keys) from the legacy device directory (based on $HOME/snap)
# to the new device directory (based on $HOME/.config)

```

newsfragments/8464.bugfix.rst
```diff
@@ -0,0 +1,2 @@
+Fix for Linux snap's users where devices are removed when the snap is removed (``snap remove parsec``).
+The fix include set ``PARSEC_BASE_CONFIG_DIR`` to ``$HOME/.config`` and have a script that copy devices present in snap user data to the new location.
```
mmmarcos, line 2: The migration script would be executed on `install` or `refresh`, but not on `remove` is that so?
I would rephrase it a little bit:
```suggestion
Fix for lost devices when parsec snap is removed: devices are now stored in ``$HOME/.config`` and are independent from snap. Existing devices are migrated.
```

mmmarcos (changes requested):

mmmarcos (approved):
LGTM

Issue #8437, opened 2024-09-18
[🚀 | Feature request]: Go-to-home button is not always visible in customer area


vxgmichel:
### What you would like to accomplish

When a single organization is present in the customer area, only the "go to organization" button is visible.

### How the feature should work

"Go to home" should always be there, maybe as another button close to the "log out" button.

PR #8501, opened 2024-09-24
[MS] Made 'Go to home' button in client area accessible when an org is chosen

Ironicbay:
Closes https://github.com/Scille/parsec-cloud/issues/8437

### DIFF
--- a/client/src/views/client-area/ClientAreaSidebar.vue
+++ b/client/src/views/client-area/ClientAreaSidebar.vue
@@ -79,14 +79,18 @@
       <!-- button: go to organization -->
       <div
         class="organization-card-button custom-button custom-button-fill"
-        v-show="showMenu"
-        @click="goToHome"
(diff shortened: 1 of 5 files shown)


client/src/views/client-area/ClientAreaSidebar.vue
```diff
@@ -346,12 +368,12 @@ async function openOrganizationChoice(event: Event): Promise<void> {
   }
 }
 
+async function goToOrganization(): Promise<void> {
+  await navigateTo(Routes.Home, { skipHandle: true, query: { bmsOrganizationId: props.organization.parsecId } });
```
Max-7, line 372: Also add `replace: true` in both options.

Max-7 (approved):

NicoTuxx (changes requested):
Not what was expected

Max-7 (approved):

NicoTuxx (approved):

Issue #8445, opened 2024-09-19
Aggregate operations (imports, moves, copies) when there are too many in the process (1st improvement)


fabienSvtr:
## How it should be improved
On the left "quick version", on the right a little better one if you have more time.
![image](https://github.com/user-attachments/assets/dc8f7618-6372-44e8-ba2c-66234150e3ff)

PR #8506, opened 2024-09-24
[MS] Aggregate file operations

Ironicbay:
Closes https://github.com/Scille/parsec-cloud/issues/8445

### DIFF
--- a/client/src/services/fileOperationManager.ts
+++ b/client/src/services/fileOperationManager.ts
@@ -26,7 +26,7 @@ import {
 import { wait } from '@/parsec/internals';
 import { v4 as uuid4 } from 'uuid';
 
-const MAX_SIMULTANEOUS_IMPORT_JOBS = 5;
+const MAX_SIMULTANEOUS_OPERATIONS = 3;
 
 export const FileOperationManagerKey = 'fileOperationManager';
 
@@ -422,6 +422,7 @@ class FileOperationManager {
 
   private async doMove(data: MoveData): Promise<void> {
     await this.sendState(FileOperationState.MoveStarted, data);
+    await this.sendState(FileOperationState.OperationProgress, data, { progress: 0 });
 
     let moveResult = await moveEntry(data.workspaceHandle, data.srcPath, data.dstPath, data.forceReplace);
     if (!moveResult.ok) {
@@ -441,15 +442,15 @@ class FileOperationManager {
         moveResult = await moveEntry(data.workspaceHandle, data.srcPath, data.dstPath, false);
         if (moveResult.ok) {
           data.dstPath = dstPath;
-          this.sendState(FileOperationState.EntryMoved, data);
+          await this.sendState(FileOperationState.EntryMoved, data);
         }
         i++;
       }
       if (!moveResult.ok) {
-        this.sendState(FileOperationState.MoveFailed, data, { error: moveResult.error.tag });
+        await this.sendState(FileOperationState.MoveFailed, data, { error: moveResult.error.tag });
       }
     } else {
-      this.sendState(FileOperationState.EntryMoved, data);
+      await this.sendState(FileOperationState.EntryMoved, data);
     }
   }
 
@@ -558,7 +559,7 @@ class FileOperationManager {
         break;
       }
 
-      if (this.operationJobs.length >= MAX_SIMULTANEOUS_IMPORT_JOBS) {
+      if (this.operationJobs.length >= MAX_SIMULTANEOUS_OPERATIONS) {
         // Remove the files that have been cancelled but have not yet
         // started their import
         for (const cancelId of this.cancelList.slice()) {
--- a/client/src/views/files/FileOperationMenu.vue
+++ b/client/src/views/files/FileOperationMenu.vue
@@ -30,7 +30,7 @@
       >
         <div class="item-container">
           <ion-text>{{ $msTranslate('FoldersPage.ImportFile.tabs.inProgress') }}</ion-text>
-          <span class="text-counter">{{ inProgressItems.length > 99 ? '+99' : inProgressItems.length }}</span>
+          <span class="text-counter">{{ inProgressAndQueuedLength > 99 ? '+99' : inProgressAndQueuedLength }}</span>
         </div>
       </ion-item>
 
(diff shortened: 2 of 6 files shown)


client/src/views/files/FileOperationMenu.vue
```diff
@@ -60,14 +60,19 @@
     <ion-list class="upload-menu-list">
       <template v-if="currentTab === Tabs.InProgress">
         <component
-          v-for="item in inProgressItems"
+          v-for="item in inProgressItems.slice(0, MAX_SIMULTANEOUS_OPERATIONS)"
```
Max-7, line 63: I'm not sure this list is always ordered. Instead, use another list for items that have the state `OperationProgress` (and maybe rename the `inProgressItems` list.

client/src/services/fileOperationManager.ts
```diff
@@ -26,7 +26,7 @@ import {
 import { wait } from '@/parsec/internals';
 import { v4 as uuid4 } from 'uuid';
 
-const MAX_SIMULTANEOUS_IMPORT_JOBS = 5;
+export const MAX_SIMULTANEOUS_OPERATIONS = 3;
```
Max-7, line 29: This shouldn't be exported. We may change things a bit later on and allow some control over how many operations should be done simultaneously, but the components using the manager shouldn't rely on this variable, they should only rely on the events.

client/src/views/files/FileOperationMenu.vue
```diff
@@ -60,14 +60,19 @@
...
           :is="getFileOperationComponent(item)"
           :key="item.data.id"
           :state="item.state"
           :state-data="item.stateData"
           :operation-data="item.data"
           @cancel="cancelOperation"
         />
+        <file-queued-item
+          v-if="inProgressItems.length > MAX_SIMULTANEOUS_OPERATIONS"
```
Max-7, line 72: Don't rely on this variable, only use the state of the item.

Max-7 (changes requested):

client/src/views/files/FileOperationMenu.vue
```diff
@@ -236,6 +253,17 @@ async function cancelOperation(importId: string): Promise<void> {
   await fileOperationManager.cancelOperation(importId);
 }
 
+async function cancelQueuedOperations(): Promise<void> {
+  errorHappenedInUpload.value = true;
+
+  for (const operation of queuedItems.value) {
```
Max-7, line 259: Only call `cancelOperation`. The fileOperationManager should then raise an event FileOperationState.Cancelled to warn you that the operation has been cancelled, at which time you can remove it from one of your queues.

client/src/views/files/FileOperationMenu.vue
```diff
@@ -197,8 +203,19 @@ function getFileOperationComponent(item: OperationItem): Component | undefined {
 }
 
 function updateImportState(id: string, state: FileOperationState, stateData?: StateData): void {
-  const index = inProgressItems.value.findIndex((op) => op.data.id === id);
+  let index: number;
+
+  if (state === FileOperationState.OperationProgress) {
+    // Check if item has to be moved from queued
+    index = queuedItems.value.findIndex((op) => op.data.id === id);
```
Max-7, line 210: I would do the opposite: check if the item is already in `inProgressItems`, if not push it and then try to remove it from queuedItems.

The reason being that `inProgressItems` will probably be quite short (it shouldn't contain more than MAX_SIMULTANEOUS_OPERATIONS) which means a find() would be very quick, whereas `queuedItems` could become very big.

Max-7 (changes requested):

client/src/views/files/FileOperationMenu.vue
```diff
@@ -219,6 +274,33 @@ function updateImportState(id: string, state: FileOperationState, stateData?: St
       inProgressItems.value.splice(index, 1);
       errorHappenedInUpload.value = true;
     }
+  } else if (state === FileOperationState.Cancelled) {
+    let arr: Array<OperationItem>;
+    switch (data.getDataType()) {
```
Max-7, line 279: We can also cancel an item that's in progress.

client/src/views/files/FileOperationMenu.vue
```diff
@@ -258,9 +346,27 @@ async function onFileOperationEvent(
       }
       break;
     case FileOperationState.FileAdded:
+      queuedUploadItems.value.push({
+        data: fileOperationData as FileOperationData,
+        state: state,
+        stateData: stateData,
+      });
+      if (uploadMenuList.value?.$el) {
```
Max-7, line 354: Could you extract this in a function?

Max-7 (changes requested):

client/src/services/fileOperationManager.ts
```diff
@@ -421,6 +421,7 @@ class FileOperationManager {
   }
 
   private async doMove(data: MoveData): Promise<void> {
+    await this.sendState(FileOperationState.OperationProgress, data, { progress: 0 });
```
Max-7, line 424: After the `MoveStarted` event. Cannot have Progress if it isn't started yet.

client/src/views/files/FileOperationMenu.vue
```diff
@@ -196,16 +221,67 @@ function getFileOperationComponent(item: OperationItem): Component | undefined {
...
+  let index: number;
+
+  if (state === FileOperationState.OperationProgress) {
+    // Check if item has to be moved from queued to inProgress
+    if (!inProgressItems.value.find((op) => op.data.id === data.id)) {
+      switch (data.getDataType()) {
+        case FileOperationDataType.Import:
+          index = queuedUploadItems.value.findIndex((op) => op.data.id === data.id);
+          if (queuedUploadItems.value[index]) {
+            inProgressItems.value.push(queuedUploadItems.value[index]);
```
Max-7, line 234: You can avoid one lookup by doing what you did, except just on two lines:
```
const item = queuedUploadItems.value.splice(index, 1);
inProgressItems.value.push(items);
```

client/src/views/files/FileOperationMenu.vue
```diff
@@ -196,16 +221,67 @@ function getFileOperationComponent(item: OperationItem): Component | undefined {
...
+          if (queuedMoveItems.value[index]) {
+            inProgressItems.value.push(queuedMoveItems.value[index]);
+            queuedMoveItems.value.splice(index, 1);
+          }
+          break;
+      }
+    }
+  }
+
+  let currentOperationArray: Array<OperationItem> = [];
```
Max-7, line 256: You could init `currentOperationArray` to `inProgressItems.value` directly and store the index. To me, it's a bit easier to follow.

```typescript
let currentOperationArray = inProgressItems.value;
index = inProgressItems.value.findIndex((op) => op.data.id === data.id);

if (index === -1) {
  switch (data.getDataType()) {
    ...
  }
  index = currentOperationArray.findIndex((op) => op.data.id === data.id);
}

const operator = currentOperationArray[index];
```

Max-7 (changes requested):

Max-7 (approved):

PR #8507, opened 2024-09-24
Fix nightly release failure because of invalid current branch

FirelightFlagboy:
The current branch check was failing when releasing a nightly build when the patch != 0. This check is here to enforce the use of the release branch for the next patch version. But that doesn't apply to nightly builds.

### DIFF
--- a/misc/releaser.py
+++ b/misc/releaser.py
@@ -471,7 +471,7 @@ def ensure_working_in_a_clean_git_repo() -> None:
 
 
 def ensure_working_on_the_correct_branch(
-    release_branch: str, base_ref: str | None, version: Version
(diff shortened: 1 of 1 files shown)


misc/releaser.py
```diff
@@ -480,7 +480,7 @@ def ensure_working_on_the_correct_branch(
         print("Already in the release branch, nothing to do")
         return
 
-    if version.patch == 0:
+    if version.patch == 0 or nightly:
```
mmmarcos, line 483:
```suggestion
    # Force use of the release branch for the next patch version or nightly build
    if version.patch == 0 or nightly:
```

mmmarcos (approved):

PR #8511, opened 2024-09-25
Add Sentry's AsyncioIntegration

mmmarcos:
This PR 
- adds `AsyncioIntegration` because it is _not_ automatically added by Sentry
- removes `LoggingIntegration` because it is automatically added by Sentry
  - also, the configuration we used (`level = logging.INFO`, `event_level=logging.ERROR`) corresponds to the default


vxgmichel (approved):
LGTM :+1:

PR #8522, opened 2024-09-26
Acknowledge release 3.0.1

FirelightFlagboy:
- Bump version 3.0.1-a.0 -> 3.0.2-a.0+dev
- Rework the logic to generate the next development version
- When using current version for release, remove the local part of the version
- Fix missing strip of the output of `git rev-parse`


LICENSE
```diff
@@ -19,7 +19,7 @@ Additional Use Grant: You may make production use of the Licensed Work, provided
                       2) use the Licensed Work for use in a governmental or any
                       for-profit entity with more than fifty (50) users.
 
-Change Date:  Sep 17, 2028
+Change Date:  Sep 25, 2028
```
AureliaDolo, line 22: How is that date chosen ?

docs/adminguide/installation.rst
```diff
@@ -9,9 +9,11 @@ Install Linux Parsec CLI
 
 Parsec is also available as a command line interface (CLI) for Linux. It is a standalone binary that is provided in our GitHub releases.
 
+.. _Parsec CLI v3.0.2-a.0+dev: https://github.com/Scille/parsec-cloud/releases/download/v3.0.2-a.0+dev/parsec-cli_3.0.2-a.0+dev_linux_x86_64
```
AureliaDolo, line 12: I'm not sure exactly who is the target of that doc, but if it's users and not just developers, I'd rather suggest installing the latest released version, not the currently developed one.

AureliaDolo (approved):

LICENSE
FirelightFlagboy: The script `releaser.py` update the license files:

https://github.com/Scille/parsec-cloud/blob/9cae172727d64e29e93bb77703eb48386bb2b7fc/misc/releaser.py#L588

docs/adminguide/installation.rst
FirelightFlagboy: You're right, but that already handle, when we will publish a new version that include this part of the documentation, the version will match the one released (that version is handled by `version_updater.py`).

LICENSE
AureliaDolo: Thanks !

docs/adminguide/installation.rst
AureliaDolo: So for now this doc is not published ?

AureliaDolo: Thanks !

PR #8566, opened 2024-10-01
Reenable manifest update lock guard release check

touilleMan:
close https://github.com/Scille/parsec-cloud/issues/8579

### DIFF
--- a/libparsec/crates/client/src/workspace/store/per_manifest_update_lock.rs
+++ b/libparsec/crates/client/src/workspace/store/per_manifest_update_lock.rs
@@ -48,28 +48,25 @@ pub(super) enum ManifestUpdateLockTakeOutcome {
 #[derive(Debug)] // Note this should not be made `Clone` !
 pub(super) struct ManifestUpdateLockGuard {
     entry_id: VlobID,
-    #[allow(unused)]
     is_released: bool,
 }
 
-// TODO: Lock is useful to detect misuse of the guard, but it is also very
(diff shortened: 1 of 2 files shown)


FirelightFlagboy (commented):
Does this PR fix an issue ? (that is not linked)

For me, this PR lack a test that would have failed without the patch

touilleMan:
> Does this PR fix an issue ? (that is not linked)

Nop, I've created #8579
(I don't think a newsfragment is needed for this PR though, given it fixes something that is most likely invisible for the end user)

> For me, this PR lack a test that would have failed without the patch

That's the trick: without this patch we can get guard not released but right at the moment we stop the WorkspaceOps, so no deadlock should occur since everything is expected to go down.

However the key point here is we introduce deadlocking code without realizing it which is very bad !

Hence the re-enabling of the lock-released-before-drop check to protect us against this kind of errors (that can be much more serious that what we got here !).

FirelightFlagboy:
It's not possible to show that `ManifestUpdateLockGuard` was not released prior to being dropped before the patch ?

:thinking: That would be possible to `log::error` was replaced by a panic, but ... that may not be something we want

touilleMan:
> That would be possible to log::error was replaced by a panic, but ... that may not be something we want

That was precisely what the commented code was doing, see the comment about why this was commented in the first place

FirelightFlagboy (approved):

PR #8583, opened 2024-10-02
Improve FUSE config and document it

touilleMan:


libparsec/crates/platform_mountpoint/src/unix/filesystem.rs
```diff
@@ -1086,6 +1169,17 @@ impl fuser::Filesystem for Filesystem {
...
+        let log = std::path::PathBuf::from(format!(
+            "/home/emmanuel/projects/parsec-cloud/fs_{ino:x}_{offset:x}.log"
+        ));
+        std::fs::File::options()
+            .create(true)
+            .append(true)
+            .write(true)
+            .open(log)
+            .unwrap()
+            .write_all(data)
```
vxgmichel, line 1181: Debug artefacts?

libparsec/crates/platform_mountpoint/src/unix/filesystem.rs
```diff
@@ -216,8 +217,90 @@ impl fuser::Filesystem for Filesystem {
...
+            .expect("Capability available");
+
+        // Adaptative readdirplus optimization support: still use readdir when lookup is
+        // has already been done by the previous readdirplus.
+        // See https://github.com/libfuse/libfuse/blob/2aeef499b84b596608181f9b48d589c4f8ffe24a/include/fuse_common.h#L311
+        config
+            .add_capabilities(fuser::consts::FUSE_READDIRPLUS_AUTO)
+            .expect("Capability available");
+
+        // TODO: `FUSE_SPLICE_READ` seems to require `write_buf()` to be implemented, which is
```
FirelightFlagboy, line 278: Why read require `write_buf`, did you mean `read_buf` ?

Can you create an issue on `fuser` side ?

libparsec/crates/platform_mountpoint/src/unix/filesystem.rs
```diff
@@ -1086,6 +1169,17 @@ impl fuser::Filesystem for Filesystem {
...
+            "/home/emmanuel/projects/parsec-cloud/fs_{ino:x}_{offset:x}.log"
+        ));
+        std::fs::File::options()
+            .create(true)
+            .append(true)
+            .write(true)
+            .open(log)
+            .unwrap()
+            .write_all(data)
+            .unwrap();
```
FirelightFlagboy, line 1182: To remove

libparsec/crates/platform_mountpoint/src/unix/filesystem.rs
touilleMan: No, it's a low tech backdoor :ninja: :trollface:

libparsec/crates/platform_mountpoint/src/unix/filesystem.rs
touilleMan: done.

libparsec/crates/platform_mountpoint/src/unix/filesystem.rs
touilleMan: Nop it's `write_buf` 

> Can you create an issue on fuser side ?

This is not related to fuser, it's [documented as such on libfuse](https://github.com/libfuse/libfuse/blob/2aeef499b84b596608181f9b48d589c4f8ffe24a/include/fuse_common.h#L227-L233)

I've created an issue on fuser side to try to better understand what is the support of `write_buf` on fuser: https://github.com/cberner/fuser/issues/298

libparsec/crates/platform_mountpoint/Cargo.toml
```diff
@@ -17,7 +17,12 @@ regex = { workspace = true }
 winfsp_wrs = { workspace = true }
 
 [target.'cfg(target_family = "unix")'.dependencies]
-fuser = { workspace = true, features = ["libfuse", "abi-7-21"] }
+# ABI 7.31 has been introduced in Linux 5.2 (2019-06-08), should be old enough
```
mmmarcos, line 20: Shouldn't this be carefully tested?

A quick search shows that Ubuntu `22.10` (and older) started with initial kernel version _older_ than `5.2`. On Debian, only `12` (stable) started with a `6.x` version. 
Most up-to-date systems will meet the requirement, but I wouldn't be surprised if the administration still uses "old" RHEL/CentOS laptops. Maybe we should check that?

touilleMan: I wouldn't be concerned about this: if your kernel is still from 2019 that means your have never updated your system... [then you have much bigger issues than not being able to run Parsec :trollface: ](https://www.cvedetails.com/product/47/Linux-Linux-Kernel.html?vendor_id=33)

FirelightFlagboy (approved):

PR #8585, opened 2024-10-02
Add bindings for recovery devices.

### LINKED ISSUE #7044: Functions to create and import recovery devices in the bindings
Should not be a priority since:
1. we're not compatible with v2 recovery devices anyway so we're not expecting any user to import recovery devices created with v2 into v3
2. the keyring should be the most used way to authenticate, and in case a user prefers to use a password, there's less chance of them forgetting it

Also check if this feature could be completly replaced by the shamir.

AureliaDolo:
Fix #7044

- [x] agree on the interface, see below (thanks @Max-7)
- [x] Adapt/Implement low-level functions in libparsec crate `platform_device_loader` [^1]
  - Adapt existing functions to use bytes instead of paths
  - Since this is OS-agnostic, it could go in the common part (not native/web specific implementation)
- [ ] Adapt CLI existing commands to use the new low-level functions see #8805 
- [x] Implement high-level functions in libparsec 
- [x] Add tests to https://github.com/Scille/parsec-cloud/blob/aa583a0c3565e1e2b69c1e342789eafebe4f8fa0/libparsec/crates/platform_device_loader/tests/recovery.rs
  - Create a recovery device, import it, check if workspace is accessible
  - Bad data tests: wrong passphrase, etc.
  - Dump recovery device and use it in tests, to check for regression in future versions (can be done later, create an issue)
- [x] Expose high-level functions in bindings
  - Translate interface to Python for bindings
  - Create specific enum for error cases
  - Generate bindings

```rust
pub async fn import_recovery_device(
    recovery_device: Vec<u8>,
    passphrase: String,
    device_label: DeviceLabel,
    save_strategy: DeviceSaveStrategy,
) -> Result<AvailableDevice, ImportRecoveryDeviceError> {
    todo!()
}

pub async fn export_recovery_device(
    clientHandle: Handle
) -> Result<(String, Vec<u8>), ExportRecoveryDeviceError> {
    todo!()
}
```

[^1]: in the existing we are using paths because it's only used in the CLI see https://github.com/Scille/parsec-cloud/blob/1e819d6f2b2edb96a6ee7c51042cfd06d8f97962/libparsec/crates/platform_device_loader/src/lib.rs#L226-L241
, but that won't work with the GUI \o/

### DIFF
--- a/libparsec/crates/client/src/certif/shamir_recovery_setup.rs
+++ b/libparsec/crates/client/src/certif/shamir_recovery_setup.rs
@@ -176,17 +176,10 @@ async fn create_shamir_recovery_device(
 ) -> Result<CreateShamirRecoveryDeviceOutcome, CertifShamirError> {
     let author = &certificate_ops.device;
 
-    let recovery_device = LocalDevice::generate_new_device(
-        certificate_ops.device.organization_addr.clone(),
-        certificate_ops.device.initial_profile,
-        certificate_ops.device.human_handle.clone(),
+    let recovery_device = LocalDevice::from_existing_device_for_user(
+        &author.clone(),
         DeviceLabel::try_from(format!("shamir-recovery-{timestamp}").as_str())
             .expect("Invalid device label"),
-        Some(certificate_ops.device.user_id),
-        None,
-        None,
-        None,
-        None,
     );
 
     let device_cert = DeviceCertificate {
--- a/libparsec/crates/client/src/client/mod.rs
+++ b/libparsec/crates/client/src/client/mod.rs
@@ -2,6 +2,7 @@
 
 #![allow(dead_code)]
 
+mod recovery_device;
 mod shamir_setup_create;
 mod tos;
 mod user_revoke;
@@ -44,7 +45,11 @@ use crate::{
 };
 use libparsec_client_connection::AuthenticatedCmds;
 use libparsec_platform_async::lock::Mutex as AsyncMutex;
+
 use libparsec_types::prelude::*;
+pub use recovery_device::{
+    import_recovery_device, ClientExportRecoveryDeviceError, ImportRecoveryDeviceError,
+};
 
 // Re-exposed for public API
 pub use crate::certif::{
@@ -532,6 +537,13 @@ impl Client {
 
     pub async fn accept_tos(&self, tos_updated_on: DateTime) -> Result<(), ClientAcceptTosError> {
         tos::accept_tos(self, tos_updated_on).await
+    }
+
+    pub async fn client_export_recovery_device(
+        &self,
+        device_label: DeviceLabel,
+    ) -> Result<(SecretKeyPassphrase, Vec<u8>), ClientExportRecoveryDeviceError> {
+        recovery_device::export_recovery_device(self, device_label).await
     }
 }
 
--- a/libparsec/crates/client/src/client/recovery_device.rs
+++ b/libparsec/crates/client/src/client/recovery_device.rs
@@ -1,0 +1,357 @@
+// Parsec Cloud (https://parsec.cloud) Copyright (c) BUSL-1.1 2016-present Scille SAS
+
+use std::{path::Path, sync::Arc};
+
+use libparsec_client_connection::{AuthenticatedCmds, ConnectionError, ProxyConfig};
+use libparsec_platform_device_loader::{
+    get_default_key_file, save_device, PlatformImportRecoveryDeviceError, SaveDeviceError,
+};
+use libparsec_platform_storage::certificates::PerTopicLastTimestamps;
+use libparsec_protocol::authenticated_cmds::v4::device_create;
+use libparsec_types::{
+    AvailableDevice, DeviceLabel, DeviceSaveStrategy, LocalDevice, SecretKeyPassphrase,
+};
+
+use crate::{
+    greater_timestamp, CertifPollServerError, GreaterTimestampOffset, InvalidCertificateError,
+};
+use libparsec_types::prelude::*;
+
+use super::Client;
+
+pub async fn export_recovery_device(
+    client: &Client,
+    device_label: DeviceLabel,
+) -> Result<(SecretKeyPassphrase, Vec<u8>), ClientExportRecoveryDeviceError> {
+    let (passphrase, data, recovery_device) =
+        libparsec_platform_device_loader::export_recovery_device(&client.device, device_label)
+            .await;
+
+    // save recovery device
+    let latest_known_timestamp =
+        register_new_device(&client.cmds, &recovery_device, &client.device).await?;
+
+    let latest_known_timestamp = PerTopicLastTimestamps::new_for_common(latest_known_timestamp);
+
+    client
+        .certificates_ops
+        .poll_server_for_new_certificates(Some(&latest_known_timestamp))
+        .await
+        .map_err(|e| match e {
+            CertifPollServerError::Stopped => ClientExportRecoveryDeviceError::Stopped,
+            CertifPollServerError::Offline => ClientExportRecoveryDeviceError::Offline,
+            CertifPollServerError::InvalidCertificate(err) => {
+                ClientExportRecoveryDeviceError::InvalidCertificate(err)
+            }
+            CertifPollServerError::Internal(err) => err
+                .context("Cannot poll server for new certificates")
+                .into(),
+        })?;
+    Ok((passphrase, data))
+}
+
+#[derive(Debug, thiserror::Error)]
+pub enum ClientExportRecoveryDeviceError {
+    #[error("Component has stopped")]
+    Stopped,
+    #[error("Cannot reach the server")]
+    Offline,
+    #[error(transparent)]
+    Internal(#[from] anyhow::Error),
+
+    #[error("Our clock ({client_timestamp}) and the server's one ({server_timestamp}) are too far apart")]
+    TimestampOutOfBallpark {
+        server_timestamp: DateTime,
+        client_timestamp: DateTime,
+        ballpark_client_early_offset: f64,
+        ballpark_client_late_offset: f64,
+    },
+    #[error(transparent)]
+    InvalidCertificate(#[from] Box<InvalidCertificateError>),
+}
+
+impl From<RegisterNewDeviceError> for ClientExportRecoveryDeviceError {
+    fn from(value: RegisterNewDeviceError) -> Self {
+        match value {
+            RegisterNewDeviceError::Stopped => ClientExportRecoveryDeviceError::Stopped,
+            RegisterNewDeviceError::Offline => ClientExportRecoveryDeviceError::Offline,
+            RegisterNewDeviceError::Internal(error) => {
+                ClientExportRecoveryDeviceError::Internal(error)
+            }
+            RegisterNewDeviceError::TimestampOutOfBallpark {
+                server_timestamp,
+                client_timestamp,
+                ballpark_client_early_offset,
+                ballpark_client_late_offset,
+            } => ClientExportRecoveryDeviceError::TimestampOutOfBallpark {
+                server_timestamp,
+                client_timestamp,
+                ballpark_client_early_offset,
+                ballpark_client_late_offset,
+            },
+            RegisterNewDeviceError::InvalidCertificate(invalid_certificate_error) => {
+                ClientExportRecoveryDeviceError::InvalidCertificate(invalid_certificate_error)
+            }
+        }
+    }
+}
+
+#[derive(Debug, thiserror::Error)]
+pub enum ImportRecoveryDeviceError {
+    #[error("Component has stopped")]
+    Stopped,
+    #[error("Cannot reach the server")]
+    Offline,
+    #[error(transparent)]
+    Internal(#[from] anyhow::Error),
+
+    #[error("Our clock ({client_timestamp}) and the server's one ({server_timestamp}) are too far apart")]
+    TimestampOutOfBallpark {
+        server_timestamp: DateTime,
+        client_timestamp: DateTime,
+        ballpark_client_early_offset: f64,
+        ballpark_client_late_offset: f64,
+    },
+    #[error(transparent)]
+    InvalidCertificate(#[from] Box<InvalidCertificateError>),
+    #[error("Cannot deserialize file content")]
+    InvalidData,
+    #[error("Passphrase format is invalid")]
+    InvalidPassphrase,
+    #[error("Failed to decrypt file content")]
+    DecryptionFailed,
+    #[error(transparent)]
+    InvalidPath(anyhow::Error),
+}
+
+impl From<PlatformImportRecoveryDeviceError> for ImportRecoveryDeviceError {
+    fn from(value: PlatformImportRecoveryDeviceError) -> Self {
+        match value {
+            PlatformImportRecoveryDeviceError::InvalidData => {
+                ImportRecoveryDeviceError::InvalidData
+            }
+            PlatformImportRecoveryDeviceError::InvalidPassphrase => {
+                ImportRecoveryDeviceError::InvalidPassphrase
+            }
+            PlatformImportRecoveryDeviceError::DecryptionFailed => {
+                ImportRecoveryDeviceError::DecryptionFailed
+            }
+        }
+    }
+}
+
+impl From<RegisterNewDeviceError> for ImportRecoveryDeviceError {
+    fn from(value: RegisterNewDeviceError) -> Self {
+        match value {
+            RegisterNewDeviceError::Stopped => ImportRecoveryDeviceError::Stopped,
+            RegisterNewDeviceError::Offline => ImportRecoveryDeviceError::Offline,
+            RegisterNewDeviceError::Internal(error) => ImportRecoveryDeviceError::Internal(error),
+            RegisterNewDeviceError::TimestampOutOfBallpark {
+                server_timestamp,
+                client_timestamp,
+                ballpark_client_early_offset,
+                ballpark_client_late_offset,
+            } => ImportRecoveryDeviceError::TimestampOutOfBallpark {
+                server_timestamp,
+                client_timestamp,
+                ballpark_client_early_offset,
+                ballpark_client_late_offset,
+            },
+            RegisterNewDeviceError::InvalidCertificate(invalid_certificate_error) => {
+                ImportRecoveryDeviceError::InvalidCertificate(invalid_certificate_error)
+            }
+        }
+    }
+}
+
+impl From<SaveDeviceError> for ImportRecoveryDeviceError {
+    fn from(value: SaveDeviceError) -> Self {
+        match value {
+            SaveDeviceError::InvalidPath(error) => ImportRecoveryDeviceError::InvalidPath(error),
+            SaveDeviceError::Internal(error) => ImportRecoveryDeviceError::Internal(error),
+        }
+    }
+}
+
+async fn register_new_device(
+    cmds: &AuthenticatedCmds,
+    new_device: &LocalDevice,
+    author: &LocalDevice,
+) -> Result<DateTime, RegisterNewDeviceError> {
+    // Loop is needed to deal with server requiring greater timestamp
+    let mut timestamp = author.now();
+
+    loop {
+        let outcome = internal_register_new_device(cmds, new_device, author, timestamp).await?;
+
+        match outcome {
+            DeviceInternalsOutcome::Done(timestamp) => return Ok(timestamp),
+            DeviceInternalsOutcome::RequireGreaterTimestamp(strictly_greater_than) => {
+                // TODO: handle `strictly_greater_than` out of the client ballpark by
+                // returning an error
+                timestamp = greater_timestamp(
+                    &author.time_provider,
+                    GreaterTimestampOffset::User,
+                    strictly_greater_than,
+                );
+            }
+        }
+    }
+}
+
+pub(crate) struct DeviceCertificatesBytes {
+    pub full: Bytes,
+    pub redacted: Bytes,
+}
+
+/// generates certificates for new device signed by author at now
+pub(crate) fn generate_new_device_certificates(
+    new_device: &LocalDevice,
+    author: &LocalDevice,
+    now: DateTime,
+) -> DeviceCertificatesBytes {
+    let device_cert = DeviceCertificate {
+        author: CertificateSignerOwned::User(author.device_id),
+        timestamp: now,
+        user_id: new_device.user_id,
+        device_id: new_device.device_id,
+        device_label: MaybeRedacted::Real(new_device.device_label.clone()),
+        verify_key: new_device.verify_key(),
+        algorithm: SigningKeyAlgorithm::Ed25519,
+    };
+
+    let device_certificate = device_cert.dump_and_sign(&author.signing_key).into();
+
+    let redacted_device_cert = device_cert.into_redacted();
+
+    let redacted_device_certificate = redacted_device_cert
+        .dump_and_sign(&author.signing_key)
+        .into();
+
+    DeviceCertificatesBytes {
+        full: device_certificate,
+        redacted: redacted_device_certificate,
+    }
+}
+
+#[derive(Debug, thiserror::Error)]
+enum RegisterNewDeviceError {
+    #[error("Component has stopped")]
+    Stopped,
+    #[error("Cannot reach the server")]
+    Offline,
+    #[error(transparent)]
+    Internal(#[from] anyhow::Error),
+    #[error("Our clock ({client_timestamp}) and the server's one ({server_timestamp}) are too far apart")]
+    TimestampOutOfBallpark {
+        server_timestamp: DateTime,
+        client_timestamp: DateTime,
+        ballpark_client_early_offset: f64,
+        ballpark_client_late_offset: f64,
+    },
+    #[error(transparent)]
+    InvalidCertificate(#[from] Box<InvalidCertificateError>),
+}
+
+impl From<ConnectionError> for RegisterNewDeviceError {
+    fn from(value: ConnectionError) -> Self {
+        match value {
+            ConnectionError::NoResponse(_) => Self::Offline,
+            err => Self::Internal(err.into()),
+        }
+    }
+}
+
+#[derive(Debug)]
+enum DeviceInternalsOutcome {
+    Done(DateTime),
+    RequireGreaterTimestamp(DateTime),
+}
+
+async fn internal_register_new_device(
+    cmds: &AuthenticatedCmds,
+    new_device: &LocalDevice,
+    author: &LocalDevice,
+    now: DateTime,
+) -> Result<DeviceInternalsOutcome, RegisterNewDeviceError> {
+    let DeviceCertificatesBytes {
+        full: device_certificate,
+        redacted: redacted_device_certificate,
+    } = generate_new_device_certificates(new_device, author, now);
+    match cmds
+        .send(device_create::Req {
+            device_certificate,
+            redacted_device_certificate,
+        })
+        .await?
+    {
+        device_create::Rep::Ok => Ok(DeviceInternalsOutcome::Done(now)),
+        device_create::Rep::RequireGreaterTimestamp {
+            strictly_greater_than,
+        } =>
+        // The retry is handled by the caller
+        {
+            Ok(DeviceInternalsOutcome::RequireGreaterTimestamp(
+                strictly_greater_than,
+            ))
+        }
+        device_create::Rep::TimestampOutOfBallpark {
+            server_timestamp,
+            client_timestamp,
+            ballpark_client_early_offset,
+            ballpark_client_late_offset,
+            ..
+        } => Err(RegisterNewDeviceError::TimestampOutOfBallpark {
+            server_timestamp,
+            client_timestamp,
+            ballpark_client_early_offset,
+            ballpark_client_late_offset,
+        }),
+        bad_rep @ (device_create::Rep::UnknownStatus { .. }
+        | device_create::Rep::InvalidCertificate
+        | device_create::Rep::DeviceAlreadyExists) => {
+            Err(anyhow::anyhow!("Unexpected server response: {:?}", bad_rep).into())
+        }
+    }
+}
+
+pub async fn import_recovery_device(
+    config_dir: &Path,
+    recovery_device: &[u8],
+    passphrase: String,
+    device_label: DeviceLabel,
+    save_strategy: DeviceSaveStrategy,
+) -> Result<AvailableDevice, ImportRecoveryDeviceError> {
+    // 0) Load the recovery device
+
+    let recovery_device: Arc<_> = libparsec_platform_device_loader::import_recovery_device(
+        recovery_device,
+        passphrase.into(),
+    )
+    .await?
+    .into();
+    let cmds = AuthenticatedCmds::new(config_dir, recovery_device.clone(), ProxyConfig::default())?;
+
+    // We first register the new device to the server, and only then save it on disk.
+    // This is to ensure the device found on disk are always valid, however the drawback
+    // is it can lead to losing the device if something goes wrong between the two steps.
+    //
+    // This is considered acceptable given 1) the error window is small and
+    // 2) if this occurs, the user can always re-import the recovery device.
+
+    let new_device = LocalDevice::from_existing_device_for_user(&recovery_device, device_label);
+
+    // 1) Upload the device on the server
+
+    register_new_device(&cmds, &new_device, &recovery_device).await?;
+
+    // 2) Save the device on disk
+
+    let access = {
+        let key_file = get_default_key_file(config_dir, &new_device.device_id);
+        save_strategy.into_access(key_file)
+    };
+    let new_available_device = save_device(config_dir, &access, &new_device).await?;
+
+    Ok(new_available_device)
+}
--- a/libparsec/crates/client/tests/unit/client/recovery.rs
+++ b/libparsec/crates/client/tests/unit/client/recovery.rs
@@ -1,0 +1,117 @@
+// Parsec Cloud (https://parsec.cloud) Copyright (c) BUSL-1.1 2016-present Scille SAS
+
+use crate::DeviceInfo;
+
+use libparsec_platform_device_loader::{get_default_key_file, load_device};
+use libparsec_tests_fixtures::prelude::*;
+use libparsec_types::prelude::*;
+
+use super::utils::client_factory;
+
+fn assert_device(info: &DeviceInfo, expected_device: impl TryInto<DeviceID>, env: &TestbedEnv) {
+    let (certif, _) = env.get_device_certificate(expected_device);
+
+    let DeviceInfo {
+        id,
+        device_label,
+        created_on,
+        created_by,
+    } = info;
+    p_assert_eq!(id, &certif.device_id);
+    let expected_device_label = match &certif.device_label {
+        MaybeRedacted::Real(device_label) => device_label,
+        MaybeRedacted::Redacted(_) => unreachable!(),
+    };
+    p_assert_eq!(device_label, expected_device_label);
+    p_assert_eq!(created_on, &certif.timestamp);
+    let expected_created_by = match &certif.author {
+        CertificateSignerOwned::User(author) => Some(author),
+        CertificateSignerOwned::Root => None,
+    };
+    p_assert_eq!(created_by.as_ref(), expected_created_by);
+}
+
+fn devices_belongs_to_same_user(dev1: &LocalDevice, dev2: &LocalDevice) -> bool {
+    dev1.organization_addr == dev2.organization_addr
+        && dev1.user_id == dev2.user_id
+        && dev1.human_handle == dev2.human_handle
+        && dev1.private_key == dev2.private_key
+        && dev1.initial_profile == dev2.initial_profile
+        && dev1.user_realm_id == dev2.user_realm_id
+        && dev1.user_realm_key == dev2.user_realm_key
+}
+
+#[parsec_test(testbed = "coolorg", with_server)]
+async fn ok(env: &TestbedEnv) {
+    // Part 1: create recovery device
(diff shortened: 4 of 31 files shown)


bindings/generator/api/common.py
FirelightFlagboy, line 1: Revert change made to this file

libparsec/crates/client/src/certif/device.rs
```diff
@@ -0,0 +1,178 @@
...
+                    strictly_greater_than,
+                );
+            }
+        }
+    }
+}
+
+// returns (device_certificate, redacted_device_certificate)
+pub(crate) fn generate_new_device(
+    new_device: LocalDevice,
```
FirelightFlagboy, line 54:
```suggestion
    new_device: &LocalDevice,
```

Could be a ref, that would limit the requirement of multiple clone

libparsec/crates/platform_device_loader/src/lib.rs
```diff
@@ -264,3 +264,41 @@ pub enum RemoveDeviceError {
...
+    platform::import_recovery_device(recovery_device, passphrase, device_label).await
+}
+
+/// returns (secret key passphrase, recovery device encrypted data, recovery device)
+pub async fn inner_export_recovery_device(
+    device: &LocalDevice,
+    device_label: DeviceLabel,
+) -> Result<(SecretKeyPassphrase, Vec<u8>, LocalDevice), PlatformExportRecoveryDeviceError> {
+    platform::export_recovery_device(device, device_label).await
+}
```
FirelightFlagboy, line 304: They don't need specific platform impl (or don't seems to), move them directly here

libparsec/crates/client/src/client/mod.rs
```diff
@@ -533,6 +539,165 @@ impl Client {
...
+    #[error(transparent)]
+    DataError(#[from] libparsec_types::DataError),
+    // #[error(transparent)]
+    // EncryptionError(#[from] CertifEncryptForUserError),
+    #[error(transparent)]
+    GetCertificateError(#[from] GetCertificateError),
+    #[error(transparent)]
+    CertifDeviceError(#[from] CertifDeviceError),
+    #[error(transparent)]
+    PlatformImportRecoveryDeviceError(#[from] PlatformImportRecoveryDeviceError),
```
FirelightFlagboy, line 696: I have the feeling that should be Internal and not a specific variant :shrug:

libparsec/crates/client/src/client/mod.rs
```diff
@@ -533,6 +539,165 @@ impl Client {
...
+    #[error(transparent)]
+    CertifDeviceError(#[from] CertifDeviceError),
+    #[error(transparent)]
+    PlatformExportRecoveryDeviceError(#[from] PlatformExportRecoveryDeviceError),
+    #[error(transparent)]
+    ConnectionError(#[from] ConnectionError),
+}
+
+#[derive(Debug, thiserror::Error)]
+
```
FirelightFlagboy, line 667:
```suggestion
#[derive(Debug, thiserror::Error)]
```

Remove extra newline

libparsec/crates/types/src/lib.rs
```diff
@@ -123,6 +123,9 @@ pub type IndexInt = u64;
 // pub type IndexInt = std::num::NonZeroU64;
 // pub type IndexOffsetInt = u64;
 
+/// Used to identify client handle
+pub type Handle = u32;
```
FirelightFlagboy, line 127: This type is specific to the implementation provided for the JS Client, I suggest to keep it in `libparsec` crate and not add it to `libparsec_types`

FirelightFlagboy (changes requested):

libparsec/crates/client/src/client/mod.rs
AureliaDolo: I split this variant, to be able to distinguish invalid path from invalid passphrase.

Max-7 (approved):
Had no problem with the API while implementing #8742, so it's good for me from an outsider perspective. I'll let better minds judge the code quality.

libparsec/crates/platform_device_loader/tests/change_auth.rs
```diff
@@ -27,7 +27,7 @@ async fn same_key_file(tmp_path: TmpPath) {
         None,
         None,
         None,
-        None,
+        None,None,None,
```
FirelightFlagboy, line 30: format ?

libparsec/crates/platform_device_loader/tests/change_auth.rs
```diff
@@ -94,7 +94,7 @@ async fn different_key_file(tmp_path: TmpPath) {
...
```
FirelightFlagboy, line 97: format ?

libparsec/crates/platform_device_loader/tests/save_load.rs
```diff
@@ -26,7 +26,7 @@ async fn save_load(#[values("keyring", "password")] kind: &str, tmp_path: TmpPat
         None,
         None,
         None,
-        None,
+        None,None,None,
```
FirelightFlagboy, line 29: format ?

libparsec/crates/platform_device_loader/tests/recovery.rs
```diff
@@ -1,3 +1,40 @@
...
+            // cspell:disable-next-line
+            "parsec3://127.0.0.1:6770/Org?no_ssl=true&p=xCD7SjlysFv3d4mTkRu-ZddRjIZPGraSjUnoOHT9s8rmLA"
+        ).unwrap(),
+        UserProfile::Admin,
+        HumanHandle::new("alice@dev1", "alice").unwrap(),
+        "alice label".parse().unwrap(),
+        None,
+        None,
+        None,
+        None,None,None,None,
```
FirelightFlagboy, line 27: format ?

libparsec/crates/client/tests/unit/workspace/file_transactions_stateful.rs
```diff
@@ -160,7 +160,7 @@ impl AsyncStateMachineTest for FileTransactionStateMachine {
             None,
             None,
             None,
-            None,
+            None,None,None,
```
FirelightFlagboy, line 163: format ?

libparsec/crates/platform_device_loader/src/web/mod.rs
FirelightFlagboy, line 1: To revert

libparsec/crates/platform_device_loader/src/lib.rs
```diff
@@ -264,3 +264,113 @@ pub enum RemoveDeviceError {
...
+#[derive(Debug, thiserror::Error)]
+pub enum PlatformExportRecoveryDeviceError {
+    #[error(transparent)]
+    Internal(#[from] anyhow::Error),
+    #[error(transparent)]
+    LoadDeviceError(#[from] LoadDeviceError),
+}
+
+/// Returns recovery device
+pub async fn inner_import_recovery_device(
```
FirelightFlagboy, line 292: Remove `inner_` prefix

libparsec/crates/platform_device_loader/src/lib.rs
```diff
@@ -264,3 +264,113 @@ pub enum RemoveDeviceError {
...
+    };
+
+    Ok(recovery_device)
+}
+
+/// create a new recovery device from provided local device
+/// returns the passphrase associated with the key that
+/// encrypted device data, the dumped recovery device data
+/// and the recovery device itself
+pub async fn inner_export_recovery_device(
```
FirelightFlagboy, line 324: Remove `inner_` prefix

libparsec/crates/client/tests/unit/workspace/file_transactions_stateful.rs
AureliaDolo, line 163: It's weird that the pre commit didn't catch that :thinking:

libparsec/crates/client/src/certif/shamir_recovery_setup.rs
```diff
@@ -187,6 +187,8 @@ async fn create_shamir_recovery_device(
         None,
         None,
         None,
+        Some(certificate_ops.device.user_realm_id),
+        Some(certificate_ops.device.user_realm_key.clone()),
```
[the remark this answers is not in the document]
vxgmichel: Yes, good catch

libparsec/crates/client/src/client/mod.rs
```diff
@@ -533,8 +541,192 @@ impl Client {
...
+    }
+}
+pub async fn create_device_from_recovery(
+    cmds: Arc<AuthenticatedCmds>,
+    recovery_device: &LocalDevice,
+    device_label: &DeviceLabel,
+    save_strategy: DeviceSaveStrategy,
+    config_dir: PathBuf,
+) -> Result<AvailableDevice, ImportRecoveryDeviceError> {
+    let new_device = LocalDevice::generate_new_device(
```
vxgmichel, line 595: Let's add a new method `LocalDevice::new_device_from_existing_device` so we don't run into more problems in the future

FirelightFlagboy: :+1: 

I would like if we go with `from_existing_device`, for me the prefix `new_device` is already implied with the `from`.

libparsec/crates/client/src/certif/device.rs
```diff
@@ -0,0 +1,161 @@
...
+use super::{
+    encrypt::CertifEncryptForUserError, greater_timestamp, CertificateBasedActionOutcome,
+    GreaterTimestampOffset, InvalidCertificateError,
+};
+
+pub async fn create_new_device(
+    cmds: Arc<AuthenticatedCmds>,
+    new_device: LocalDevice,
+    author: Arc<LocalDevice>,
+) -> Result<CertificateBasedActionOutcome, CertifDeviceError> {
```
vxgmichel, line 22: Maybe rename this function `register_new_device`, since it receives a `new_device` as an argument (I understand that it's going to be sending a `create_device` request to the server, but in this context it's unclear whether `create` refers to the local device creation or its registration on the server)

libparsec/crates/client/tests/unit/workspace/file_transactions_stateful.rs
AureliaDolo: It was a too long string in the parameters that messed up rustfmt :/

AureliaDolo:
@touilleMan 767e2ce65f2b91db1d0cdb1635c8d9ff15b28802 could be cherry picked if needed

libparsec/crates/types/src/local_device.rs
```diff
@@ -60,13 +62,29 @@ impl LocalDevice {
...
+    ) -> LocalDevice {
+        LocalDevice {
+            // device specif fields
+            device_label,
+            device_id: DeviceID::default(),
+            signing_key: SigningKey::generate(),
+            time_provider: TimeProvider::default(),
+            // the other fields are user related
+            ..device.clone()
+        }
```
FirelightFlagboy, line 85: I would have use the method `LocalDevice::generate_new_device` instead, so we only have a single place where the field of `LocalDevice` are initialized

libparsec/crates/types/src/local_device.rs
```diff
@@ -60,13 +62,29 @@ impl LocalDevice {
...
-            user_realm_key: SecretKey::generate(),
+            user_realm_id: user_realm_id.unwrap_or_default(),
+            user_realm_key: user_real_key.unwrap_or_else(SecretKey::generate),
             local_symkey: SecretKey::generate(),
             time_provider: time_provider.unwrap_or_default(),
         }
     }
 
+    /// generates a new device for a given user
+    pub fn from_existing_device_for_user(
```
FirelightFlagboy, line 73: Is the suffix `_for_user` useful ? for me a device is always linked to an user/machine

libparsec/crates/platform_device_loader/src/lib.rs
```diff
@@ -264,3 +264,101 @@ pub enum RemoveDeviceError {
...
+        .to_string()
+    };
+
+    let (passphrase, key) = SecretKey::generate_recovery_passphrase();
+
+    let recovery_device = LocalDevice::from_existing_device_for_user(device, device_label);
+    let ciphertext = {
+        let mut cleartext = recovery_device.dump();
+        let ciphertext = key.encrypt(&cleartext);
+        cleartext.zeroize(); // Scrub the buffer given it contains keys in clear
```
FirelightFlagboy, line 345: Not using `zeroizing` ?

libparsec/crates/types/src/local_device.rs
AureliaDolo: It's to convey that it's for the *same* user

libparsec/crates/types/src/local_device.rs
AureliaDolo: I'd rather keep it that way, as I find it highlights better which fields are generated and which are the same.

libparsec/crates/types/src/local_device.rs
FirelightFlagboy: But that's already implied by `existing_device`, no ? :thinking:

libparsec/crates/client/src/client/mod.rs
```diff
@@ -533,8 +541,179 @@ impl Client {
...
+    cmds: Arc<AuthenticatedCmds>,
+    recovery_device: &LocalDevice,
+    device_label: &DeviceLabel,
+    save_strategy: DeviceSaveStrategy,
+    config_dir: PathBuf,
+) -> Result<AvailableDevice, ImportRecoveryDeviceError> {
+    let new_device =
+        LocalDevice::from_existing_device_for_user(recovery_device, device_label.clone());
+    // save recovery device
+    // TODO save after upload
```
FirelightFlagboy, line 596:
```suggestion
    // TODO: save after upload
```

libparsec/crates/client/src/client/mod.rs
```diff
@@ -533,8 +541,179 @@ impl Client {
...
+    let access = {
+        let key_file = dbg!(get_default_key_file(&config_dir, &new_device.device_id));
```
FirelightFlagboy, line 598:
```suggestion
        let key_file = get_default_key_file(&config_dir, &new_device.device_id);
```

libparsec/crates/platform_device_loader/tests/recovery.rs
```diff
@@ -1,3 +1,43 @@
...
+use libparsec_types::HumanHandle;
+use libparsec_types::ParsecOrganizationAddr;
+use libparsec_types::UserProfile;
+use libparsec_types::{DeviceLabel, LocalDevice};
+
+#[parsec_test]
+async fn test_ok_recovery() {
+    let url = ParsecOrganizationAddr::from_any(
+        // cspell:disable-next-line
+        "parsec3://127.0.0.1:6770/Org?no_ssl=true&p=xCD7SjlysFv3d4mTkRu-ZddRjIZPGraSjUnoOHT9s8rmLA",
```
FirelightFlagboy, line 16: Do we use the testbed here ? if thinking if the value of the URL is important and if it's value could change

libparsec/crates/types/src/local_device.rs
AureliaDolo: I don't know, it feels more explicit that way IMHO

libparsec/crates/platform_device_loader/tests/recovery.rs
AureliaDolo: It only has to be a valid organization address

vxgmichel (approved):
LGTM :+1:

libparsec/src/recovery.rs
```diff
@@ -0,0 +1,58 @@
...
+use crate::Handle;
+
+pub async fn import_recovery_device(
+    config: ClientConfig,
+    recovery_device: Vec<u8>,
+    passphrase: String,
+    device_label: DeviceLabel,
+    save_strategy: DeviceSaveStrategy,
+) -> Result<AvailableDevice, ImportRecoveryDeviceError> {
+    let recovery_device: Arc<_> = libparsec_platform_device_loader::import_recovery_device(
```
touilleMan, line 21: `libparsec` main crate is not convenient to test, so it should only contain the public API that is used for the javascript bindings (and also a bit of wrappers code to deal with the handle system)

So here you should put this `import_recovery_device` function in the `libparsec_client` crate (where it can easily be tested), then re-export it in `libparsec` (possibly creating a thin wrapper in `libparsec` if you need to deal with handles)

libparsec/crates/types/src/local_device.rs
```diff
@@ -60,13 +62,29 @@ impl LocalDevice {
...
+        device_label: DeviceLabel,
+    ) -> LocalDevice {
+        LocalDevice {
+            // device specif fields
+            device_label,
+            device_id: DeviceID::default(),
+            signing_key: SigningKey::generate(),
+            time_provider: TimeProvider::default(),
+            // the other fields are user related
+            ..device.clone()
```
touilleMan, line 84: I'd rather avoid using this implicit copy (and instead go with explicitly listing every field in the structure), since it means it will silently introduce a bug if we add a new field to `LocalDevice` that is specific to each device (and hence shouldn't be copied from the existing device)

libparsec/crates/types/src/local_device.rs
```diff
@@ -50,6 +50,8 @@ impl LocalDevice {
         signing_key: Option<SigningKey>,
         private_key: Option<PrivateKey>,
         time_provider: Option<TimeProvider>,
+        user_realm_id: Option<VlobID>,
```
touilleMan, line 53: I think this `generate_new_device` is too broad and we should instead replace it with multiple methods, each dedicated to a narrow usecase (especially since it naming is error prone, as we don't know if it means a new device for an existing or for a new user ^^).

`from_existing_device_for_user` is a good step in that direction (though I would rename this `new_device_for_user`)

So we can also add `new_user` that should be use when creating the first device for a user.

What do you think ?

libparsec/crates/types/src/local_device.rs
```diff
@@ -60,13 +62,29 @@ impl LocalDevice {
             signing_key: signing_key.unwrap_or_else(SigningKey::generate),
             private_key: private_key.unwrap_or_else(PrivateKey::generate),
             initial_profile,
-            user_realm_id: VlobID::default(),
-            user_realm_key: SecretKey::generate(),
+            user_realm_id: user_realm_id.unwrap_or_default(),
+            user_realm_key: user_real_key.unwrap_or_else(SecretKey::generate),
             local_symkey: SecretKey::generate(),
```
touilleMan, line 67: shouldn't`local_symkey` be corrected too ?

libparsec/src/handle.rs
```diff
@@ -9,6 +9,7 @@ use libparsec_client::{
 use libparsec_platform_async::event::Event;
 use libparsec_types::prelude::*;
 
+/// Used to identify client handle
```
touilleMan, line 12: Handle are also used to identify running workspace, mountpoint, and invitation progress (see the `HandleItem` enum

libparsec/crates/platform_device_loader/tests/recovery.rs
touilleMan, line 16: Don't use a real address if we don't plan on actually reaching it (otherwise this can lead to unexpected things if our code is buggy and the url is actually accessed)

Instead you should use the `.invalid` top level domain [that is a reserved TLD for this kind of needs](https://en.wikipedia.org/wiki/.invalid), so typically `parsec3://parsec.invalid/Org?...` (also note we don't need `no_ssl=true`)

libparsec/crates/platform_device_loader/tests/change_auth.rs
```diff
@@ -15,11 +15,13 @@ async fn same_key_file(tmp_path: TmpPath) {
         key_file: key_file.clone(),
         password: "P@ssw0rd.".to_owned().into(),
     };
+    let url = ParsecOrganizationAddr::from_any(
+        // cspell:disable-next-line
+        "parsec3://127.0.0.1:6770/Org?no_ssl=true&p=xCD7SjlysFv3d4mTkRu-ZddRjIZPGraSjUnoOHT9s8rmLA",
```
touilleMan, line 20: See my other comment about `invalid` TLD

libparsec/crates/client/tests/unit/workspace/file_transactions_stateful.rs
```diff
@@ -148,11 +148,12 @@ impl AsyncStateMachineTest for FileTransactionStateMachine {
             .await
             .unwrap();
 
+        let url = ParsecOrganizationAddr::from_any(
+            // cspell:disable-next-line
+            "parsec3://127.0.0.1:6770/Org?no_ssl=true&p=xCD7SjlysFv3d4mTkRu-ZddRjIZPGraSjUnoOHT9s8rmLA"
```
touilleMan, line 153: See my comment about `invalid` TLD

libparsec/crates/client/src/certif/device.rs
```diff
@@ -0,0 +1,175 @@
...
+        .into();
+
+    DeviceCertificatesBytes {
+        full: device_certificate,
+        redacted: redacted_device_certificate,
+    }
+}
+
+#[derive(Debug, thiserror::Error)]
+pub enum CertifDeviceError {
```
touilleMan, line 86: The function using this error is called `register_new_device`, so it should be renamed `CertifRegisterNewDeviceError`

libparsec/crates/client/src/certif/device.rs
```diff
@@ -0,0 +1,175 @@
...
+    cmds: Arc<AuthenticatedCmds>,
+    new_device: LocalDevice,
+    author: Arc<LocalDevice>,
+) -> Result<CertificateBasedActionOutcome, CertifDeviceError> {
+    // Loop is needed to deal with server requiring greater timestamp
+    let mut timestamp = author.now();
+
+    loop {
+        let outcome = internal_register_new_device(
+            cmds.clone(),
```
touilleMan, line 28: I think you should be able to pass those parameters by reference fine instead of using clone here

libparsec/crates/client/src/client/mod.rs
```diff
@@ -533,8 +541,175 @@ impl Client {
     pub async fn accept_tos(&self, tos_updated_on: DateTime) -> Result<(), ClientAcceptTosError> {
         tos::accept_tos(self, tos_updated_on).await
     }
+
+    pub async fn export_recovery_device(
```
touilleMan, line 545: You should put your code in a dedicated `client/recovery_device.rs` module in order to keep this file readable considering how many method `Client` has

libparsec/crates/client/src/certif/device.rs
```diff
@@ -0,0 +1,175 @@
...
+    anyhow, thiserror, Bytes, CertificateSignerOwned, DateTime, DeviceCertificate, LocalDevice,
+    MaybeRedacted, SigningKeyAlgorithm, UserID,
+};
+
+use super::{
+    encrypt::CertifEncryptForUserError, greater_timestamp, CertificateBasedActionOutcome,
+    GreaterTimestampOffset, InvalidCertificateError,
+};
+
+pub async fn register_new_device(
```
touilleMan, line 18: Given this code is only used by `Client::export_recovery_device` you may want to put it in the client/recovery_device.rs` module I was talking about in another comment.

I guess we could argue this code is fine here (given we are producing a certificate, so it's kind of okay to do that in the `certif` module), however currently we generate the user&device certificates directly in the code that need it (i.e. bootstrap organization and user/device greeter)  so we would stay consistent.

The `certif` module has currently a single goal: implement `CertificateOps`, however `registered_new_device` is not related to `CertificateOps` (given we don't need knowledge on the previous certificates to upload a new user/device certificate, which is not the case for e.g. `revoke_user` that hence is part of `CertificateOps`)

On top of that `register_new_device` is kind of broad (in a similar way to `LocalDevice::generate_new_device`), so I think it's better to keep in private in a dedicated module that only has a single goal, instead of having it public and risking to have it re-used later on for something it is not designed for...

libparsec/crates/client/tests/unit/client/recovery.rs
```diff
@@ -0,0 +1,119 @@
...
+    let config = client.config.clone();
+    let config_dir = config.config_dir.clone();
+    let recovery_device_label = DeviceLabel::try_from("recovery").unwrap();
+    let (passphrase, data) = client
+        .export_recovery_device(recovery_device_label.clone())
+        .await
+        .unwrap();
+    client.refresh_workspaces_list().await.unwrap();
+
+    // but why do you test workspace count ?
```
touilleMan, line 50: Shouldn't we compare `alice` with `recovery_device` and `new_device` on the fields that should have been preserved instead ? (again with an explicit destructuring to be explicit about the fields that are not compared in order to be resilient against newly added fields)

touilleMan (changes requested):

libparsec/crates/types/src/local_device.rs
AureliaDolo: That makes sense.

libparsec/crates/types/src/local_device.rs
AureliaDolo: And if we avoid using generate_new_device after #8790 it won't be needed.

AureliaDolo: Actually that fix is not needed anymore

libparsec/crates/client/tests/unit/client/recovery.rs
AureliaDolo: I wasn't able to compare recovery device. The associated local device is not exposed anymore.

But I compare the new device below ^^

libparsec/crates/client/src/client/mod.rs
```diff
@@ -44,7 +45,12 @@ use crate::{
 };
 use libparsec_client_connection::AuthenticatedCmds;
 use libparsec_platform_async::lock::Mutex as AsyncMutex;
+
 use libparsec_types::prelude::*;
+pub use recovery_device::{
+    create_device_from_recovery, import_recovery_device, ExportRecoveryDeviceError,
```
touilleMan, line 51: `create_device_from_recovery` is only used in `libparsec/crates/client/src/client/recovery_device.rs` so it should stay private

libparsec/crates/types/src/local_device.rs
touilleMan: As you say, we should deal with this question in #8790

libparsec/crates/platform_device_loader/src/lib.rs
```diff
@@ -264,3 +264,100 @@ pub enum RemoveDeviceError {
 }
 
 pub use platform::remove_device;
+use zeroize::Zeroizing;
+
+#[derive(Debug, thiserror::Error)]
+pub enum PlatformImportRecoveryDeviceError {
+    #[error(transparent)]
+    InvalidPath(anyhow::Error),
```
touilleMan, line 272: This error type is never used, should remove it

libparsec/crates/platform_device_loader/src/lib.rs
```diff
@@ -264,3 +264,100 @@ pub enum RemoveDeviceError {
...
+    #[error("Cannot deserialize file content")]
+    InvalidData,
+    #[error("Passphrase format is invalid")]
+    InvalidPassphrase,
+    #[error("Failed to decrypt file content")]
+    DecryptionFailed,
+    #[error(transparent)]
+    SaveDeviceError(#[from] SaveDeviceError),
```
touilleMan, line 280: This is also never used

libparsec/crates/platform_device_loader/tests/recovery.rs
```diff
@@ -1,3 +1,43 @@
...
+        export_recovery_device(&local_device, recovery_device_label.clone())
+            .await
+            .unwrap();
+    let imported_recovery_device = import_recovery_device(&data, secret_passphrase)
+        .await
+        .unwrap();
+
+    assert_eq!(exported_recovery_device, imported_recovery_device);
+    assert_eq!(imported_recovery_device.device_label, recovery_device_label);
+}
```
touilleMan, line 43: Given you have `secret_passphrase` and `data` available here, it's convenient to add test the error case for import at the end of this test function:
```rust

    // It's convenient to also test the import bad cases here

    // Invalid data
    p_assert_matches!(
        import_recovery_device(
            &b"dummy", secret_passphrase.clone()
        )
            .await,
        Err(ImportRecoveryDeviceError::InvalidData)
    );

    // Invalid passphrase

    p_assert_matches!(
        import_recovery_device(
            &data, b"dummy".to_string().into()
        )
            .await,
        Err(ImportRecoveryDeviceError::InvalidPassphrase)
    );

    // Decryption failed

    let different_secret_passphrase = {
        let mut different_secret_passphrase = secret_passphrase.clone();
        different_secret_passphrase[0] += 1; // Change the first byte
        different_secret_passphrase
    };
    p_assert_matches!(
        import_recovery_device(
            &data, different_secret_passphrase
        )
            .await,
        Err(ImportRecoveryDeviceError::DecryptionFailed)
    );
```

libparsec/crates/client/src/client/recovery_device.rs
```diff
@@ -0,0 +1,373 @@
...
+    #[error("Component has stopped")]
+    Stopped,
+    #[error("Cannot reach the server")]
+    Offline,
+    #[error(transparent)]
+    Internal(#[from] anyhow::Error),
+    #[error("User {0} is revoked")]
+    UserRevoked(UserID),
+    #[error("Threshold is 0")]
+    ThresholdIsZero,
```
touilleMan, line 268: Never used, to remove

libparsec/crates/client/src/client/recovery_device.rs
```diff
@@ -0,0 +1,373 @@
...
+        .into();
+
+    DeviceCertificatesBytes {
+        full: device_certificate,
+        redacted: redacted_device_certificate,
+    }
+}
+
+#[derive(Debug, thiserror::Error)]
+pub enum CertifRegisterNewDeviceError {
```
touilleMan, line 258: should be called `Certif...` since it's not something from the `client/src/certf`

libparsec/crates/client/src/client/recovery_device.rs
```diff
@@ -0,0 +1,373 @@
...
+    #[error("Component has stopped")]
+    Stopped,
+    #[error("Cannot reach the server")]
+    Offline,
+    #[error(transparent)]
+    Internal(#[from] anyhow::Error),
+    #[error("User {0} is revoked")]
+    UserRevoked(UserID),
```
touilleMan, line 266: Never used, to remove

libparsec/crates/client/src/client/recovery_device.rs
```diff
@@ -0,0 +1,373 @@
...
+    let (passphrase, data, recovery_device) =
+        libparsec_platform_device_loader::export_recovery_device(&client.device, device_label)
+            .await?;
+
+    // save recovery device
+    let outcome =
+        register_new_device(client.cmds.clone(), recovery_device, client.device.clone()).await?;
+
+    let latest_known_timestamps = match outcome {
+        CertificateBasedActionOutcome::LocalIdempotent => return Ok((passphrase, data)), // not sure how it could happen though
```
touilleMan, line 40: Your comment is right: `CertificateBasedActionOutcome::LocalIdempotent` is never occurs here !

(this variant is used when the action you want to do has already been done, so if you download the latest certificates you end up in the same state than if your action had succeeded. This works typically when revoking a user or giving a new role to a user in a workspace)

Here things are different: if somebody else create a device with the same `DeviceID`[^1], then you cannot carry on since it means the private keys for this device you own are invalid.

[^1]: Of course `DeviceID` is a UUID so in practice this should never occur under normal circumstances !

I think it's better to define your own outcome type (that could be private) instead of relying on `CertificateBasedActionOutcome`, this way you don't end up with this additional variant that is of no use here.

libparsec/crates/client/src/client/recovery_device.rs
```diff
@@ -0,0 +1,373 @@
...
+    cmds: Arc<AuthenticatedCmds>,
+    recovery_device: &LocalDevice,
+    device_label: &DeviceLabel,
+    save_strategy: DeviceSaveStrategy,
+    config_dir: PathBuf,
+) -> Result<AvailableDevice, ImportRecoveryDeviceError> {
+    let new_device =
+        LocalDevice::from_existing_device_for_user(recovery_device, device_label.clone());
+    // save recovery device
+    // TODO: save after upload
```
touilleMan, line 78: Is there an issue to create about this ?

What is the tradeof here ? I guess it's something similar to what we have when we create a device during enrollment (i.e. if our machine goes down at the wrong time we have the choice between losing the new device, or having in our client a device the server doesn't know about [^1])

In the end I think we should follow the same strategy in both places  (so saving the device *after* we have uploaded it to the server)

[^1]: Of course we can have the best of both world with a more complex approach based on saving a temporary type that represent the fact the device might or might have been uploaded on the server... but it's too much complexity for a corner case, at least for the moment ;-)

libparsec/crates/client/src/client/recovery_device.rs
```diff
@@ -0,0 +1,373 @@
...
+            certificate_timestamp,
+            certificate_timestamp,
+        ),
+    };
+
+    Ok(saved_device)
+}
+
+#[derive(Debug, thiserror::Error)]
+pub enum ExportRecoveryDeviceError {
```
touilleMan, line 103: Call it `ClientExportRecoveryDeviceError` since it will be used when calling `Client::export_recovery_device`

libparsec/crates/client/src/client/recovery_device.rs
```diff
@@ -0,0 +1,373 @@
...
+    #[error(transparent)]
+    CertifDeviceError(#[from] CertifRegisterNewDeviceError),
+    #[error(transparent)]
+    PlatformExportRecoveryDeviceError(#[from] PlatformExportRecoveryDeviceError),
+    #[error(transparent)]
+    ConnectionError(#[from] ConnectionError),
+}
+
+#[derive(Debug, thiserror::Error)]
+pub enum ImportRecoveryDeviceError {
```
touilleMan, line 135: Call it `ClientImportRecoveryDeviceError` since it will be used when calling `Client::import_recovery_device`

libparsec/crates/client/src/client/mod.rs
```diff
@@ -44,7 +45,11 @@ use crate::{
 };
 use libparsec_client_connection::AuthenticatedCmds;
 use libparsec_platform_async::lock::Mutex as AsyncMutex;
+
 use libparsec_types::prelude::*;
+pub use recovery_device::{
+    import_recovery_device, ExportRecoveryDeviceError, ImportRecoveryDeviceError,
```
touilleMan, line 51: I think it is a good idea to put this `import_recovery_device` in client (since the export part is there and the two are closely related) even if it is really related to `Client`

So I wonder if they would be a gain to expose `import_recovery_device` as a `Client` method anyway:
```rust
impl Client {
    [...]
    pub async fn import_recovery_device(
        &self,
       [...]
    ) -> Result<AvailableDevice, ClientImportRecoveryDeviceError> {
        recovery_device::import_recovery_device(self, [...]).await
    }
    [...]
}
```

The main gain I see is we are more consistent when naming the errors:
- Since we have `Client::export_recovery_device`, we should have`ClientExportRecoveryDeviceError`
- But if `import_recovery_device` is not a method of `Client`, it may be a bit weird to have in error named `ClientImportRecoveryDeviceError`
- And if its error is called `ImportRecoveryDeviceError`, then it is also weird since it this error is defined in `src/client` where all other public errors have a `Client` prefix (so it feels like an error in the naming...)

Why do you think ?

touilleMan (changes requested):
We are basically good here :+1: 

However you should have a look at the errors variants to make sure you only define types that are actually used (the good thing is it has a cascading effect: removing a variant often also remove a variant in another error type that do re-publishing \o/)

libparsec/crates/client/src/client/mod.rs
AureliaDolo: We could do that, it justifies the client prefix, but it's kinda artificial :shrug: 
```diff

File: libparsec/crates/client/src/client/recovery_device.rs (hunk 42 above)
Comment (AureliaDolo) [reply libparsec/crates/client/src/client/recovery_device.rs:78]: see #8810

File: libparsec/crates/client/src/client/recovery_device.rs [hunk 46]
```diff
@@ -0,0 +1,327 @@
...
+                ClientImportRecoveryDeviceError::InvalidPassphrase
+            }
+            PlatformImportRecoveryDeviceError::DecryptionFailed => {
+                ClientImportRecoveryDeviceError::DecryptionFailed
+            }
+        }
+    }
+}
+
+pub async fn register_new_device(
```
touilleMan, line 162: should be private

libparsec/crates/client/src/client/recovery_device.rs
```diff
@@ -0,0 +1,327 @@
...
+        .into();
+
+    DeviceCertificatesBytes {
+        full: device_certificate,
+        redacted: redacted_device_certificate,
+    }
+}
+
+#[derive(Debug, thiserror::Error)]
+pub enum RegisterNewDeviceError {
```
touilleMan, line 224: should be private

libparsec/crates/client/src/client/mod.rs
touilleMan: You're right, this is totally artificial :/
On top of that, I've just realize it doesn't play nice with the functions exposed in `libparsec` where anything starting with `client_` should provide a client Handle (same thing for anything starting with `workspace_` should provide a workspace Handle)

So in the end we should just name it `import_revovery_device` & `ImportRecoveryDeviceError` and keep it as a standalone function
(but of course we still have `Client::export_recovery_device`, so the error is `ClientExportRecoveryDeviceError` and the binding function in `libparsec` is `client_export_recovery_device`)

Aaaaand sorry for all thus fuzz on such a boring topic :pray:

touilleMan (changes requested):

touilleMan (approved):

PR #8598, opened 2024-10-03
Improve libparsec client connection tests

touilleMan:
depends on #8471


FirelightFlagboy (changes requested):
Extract the commit that add support for dict in the serialization crate, that would make the PR simpler to review

touilleMan:
> Extract the commit that add support for dict in the serialization crate, that would make the PR simpler to review

This commit is part of #8471 (on which this PR is based on), here you should only need to review the last 3 commits

FirelightFlagboy:
> This commit is part of https://github.com/Scille/parsec-cloud/pull/8471 (on which this PR is based on), here you should only need to review the last 3 commits

Stacked PR should be done by setting the appropriate base (`master` vs `eula`), that was not the case

FirelightFlagboy (approved):

AureliaDolo (approved):

PR #8605, opened 2024-10-03
[CLI] Remove device data alongside the device

FirelightFlagboy:


cli/src/commands/device/remove.rs
```diff
@@ -17,7 +18,14 @@ pub async fn main(args: Args) -> anyhow::Result<()> {
         config_dir.display(),
     );
 
-    let device = load_device_file(&config_dir, device).await?;
+    // FIXME: https://github.com/Scille/parsec-cloud/issues/8591
```
AureliaDolo, line 21: I don't get why the device deletion is related to the config

libparsec/crates/platform_device_loader/tests/remove.rs
```diff
@@ -0,0 +1,27 @@
...
+        key_file: key_file.clone(),
+        password: "FooBar".to_owned().into(),
+    };
+    save_device(&tmp_path, &access, &device).await.unwrap();
+
+    // 2. Remove the device.
+    remove_device(&key_file).await.unwrap();
+
+    // 3. Check that the device as been removed.
+    assert!(!key_file.exists(), "Device file should have been archived");
```
AureliaDolo, line 26:
```suggestion
    assert!(!key_file.exists(), "Device file should have been removed");
```

AureliaDolo (approved):

cli/src/commands/device/remove.rs
FirelightFlagboy: I use the wrong link :x

Should be:

```suggestion
    // FIXME: https://github.com/Scille/parsec-cloud/issues/8604
```

We need the configuration for deleting the data associated to the device (we need the value of `data_base_dir`).

Currently both the client and CLI use default value

PR #8611, opened 2024-10-04
Fix #8594: Improve error message when greeting attempt is cancelled

### LINKED ISSUE #8594: [CLI]: Improve error message in `invite claim` when greeter choose the wrong code
```txt
Greeting attempt cancelled by Greeter because InvalidSasCode on 2024-10-03T08:36:36.084506Z
```

FirelightFlagboy:


AureliaDolo (approved):
Do you know if there was a specific reason to use a screaming case previously ?

FirelightFlagboy:
> Do you know if there was a specific reason to use a screaming case previously ?

IDK, maybe @AureliaDolo knows :trollface: 

Maybe it's for a roudtrip parsing of raw string: `FromStr -> Display -> FromStr`
But we implement `serde::{Serialization,Deserialization}` for that :shrug:

Issue #8557, opened 2024-10-01
Aggregate `Done` and `Error` operations in `FileOperationMenu.vue`


Ironicbay:
Limit `Done` and `Error` items to a set small number and show the remaining items in a `+123 operations done`

PR #8636, opened 2024-10-04
[MS] Aggregate Done & Error items in FileOperationMenu.vue

Ironicbay:
Closes https://github.com/Scille/parsec-cloud/issues/8557

### DIFF
--- a/client/src/locales/en-US.json
+++ b/client/src/locales/en-US.json
@@ -573,10 +573,12 @@
             "browse": "Go to location",
             "failedDetails": "Please make sure you are online."
         },
-        "uploadsLeft": "Importing {amount} items",
-        "copiesLeft": "Copying {amount} items",
-        "movesLeft": "Moving {amount} items",
-        "operationsLeft": "{amount} operations left"
+        "uploadsLeft": "{amount} imports pending",
+        "copiesLeft": "{amount} copies pending",
+        "movesLeft": "{amount} moves pending",
+        "itemsDoneAmount": "One more operation done | {amount} more operations done",
+        "itemsErrorAmount": "One more operation cancelled or failed | {amount} more operations cancelled or failed",
+        "operationAmount": "{amount} operations"
     },
     "FileDetails": {
         "title": "Details on {name}",
--- a/client/package-lock.json
+++ b/client/package-lock.json
@@ -26,7 +26,7 @@
                 "@types/luxon": "^3.3.4",
                 "axios": "^1.7.4",
                 "luxon": "^3.4.4",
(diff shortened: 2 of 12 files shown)


client/src/services/fileOperationManager.ts
```diff
@@ -51,6 +51,12 @@ enum FileOperationState {
   EntryCopied,
 }
 
+enum FileOperationTabs {
```
Max-7, line 54: Nop. `tabs` are how we chose to display it, it has nothing to do with the operation manager. It doesn't care what we do with the information it provides.

client/src/components/files/FileAggregateItem.vue
```diff
@@ -34,20 +35,22 @@
...
-import { MsImage, Copy, FileImport, Move } from 'megashark-lib';
+import { MsImage, Copy, FileImport, Move, NoImportDone, NoImportError } from 'megashark-lib';
 import { IonIcon, IonItem, IonProgressBar, IonText, IonButton } from '@ionic/vue';
 import { closeCircle } from 'ionicons/icons';
-import { FileOperationDataType } from '@/services/fileOperationManager';
+import { FileOperationDataType, FileOperationTabs } from '@/services/fileOperationManager';
 
 const props = defineProps<{
   amount: number;
+  tab: FileOperationTabs;
```
Max-7, line 53: `Tabs` should only exist in the `FileOperationMenu`, it shouldn't be exported. The item has no need to know if it's displayed in a tab.

Beside the naming problem, I would just dupplicate the component. Have one for done items, one for errors, one for in progress. It would avoid those weird state switches, which are only going to get weirder once @fabienSvtr starts adding classes everywhere. It can be done using a `<slot />` to avoid too much copy/pasting.

Max-7 (changes requested):

Ironicbay:
Includes changes done in https://github.com/Scille/parsec-cloud/pull/8649

client/src/locales/en-US.json
```diff
@@ -573,10 +573,12 @@
...
             "failedDetails": "Please make sure you are online."
         },
-        "uploadsLeft": "Importing {amount} items",
-        "copiesLeft": "Copying {amount} items",
-        "movesLeft": "Moving {amount} items",
-        "operationsLeft": "{amount} operations left"
+        "uploadsLeft": "{amount} imports pending",
+        "copiesLeft": "{amount} copies pending",
+        "movesLeft": "{amount} moves pending",
+        "itemsDoneAmount": "{amount} more operations done",
```
Max-7, line 579: Should use the plural for this.
`One more operation done | {amount} more operations done`
and use it with
`{ key: 'xxx.itemsDoneAmount', data: { amount: amount }, count: amount }`

Max-7 (approved):

NicoTuxx (approved):
LGTM

PR #8648, opened 2024-10-07
Specify shell in all GitHub Actions workflows when needed

FirelightFlagboy:
Summary
-------

- `shell: bash` was added when a run command used pipes (including command that used `set -e -o pipefail`).

   For the documentation: <https://docs.github.com/en/actions/writing-workflows/workflow-syntax-for-github-actions#exit-codes-and-error-action-preference>

   Setting `shell` to that value will automatically add the `-o pipefail` (alongside `-e`)

- `shell: bash ... {0}` was used for run commands that used `set ...`

Fixes
-----

- Fix `yq` syntax (GA use `yq-go` not `yq` (the python wrapper of `jq`))


mmmarcos (approved):
LGTM

PR #8661, opened 2024-10-07
Workspace history

touilleMan:


Max-7 (approved):
Had no problem using the API to implement #8743.

libparsec/crates/client/src/workspace/fetch.rs
```diff
@@ -272,3 +274,131 @@ pub(super) async fn fetch_block(
...
+        manifests.push(manifest);
+    }
+
+    Ok(manifests)
+}
+
+struct VlobVersionsData {
+    needed_common_certificate_timestamp: DateTime,
+    needed_realm_certificate_timestamp: DateTime,
+    items: Vec<(VlobID, IndexInt, DeviceID, VersionInt, DateTime, Bytes)>,
```
AureliaDolo, line 366: Maybe put that into a struct ?

AureliaDolo (approved):

libparsec/crates/client/src/workspace/fetch.rs
touilleMan: I didn't use a struct here given `VlobVersionsData` is not a public structure but something only used to allow code splitting between `fetch_versions_remote_workspace_manifest` and `fetch_versions_vlob`.

The `Vec<(VlobID, IndexInt, DeviceID, VersionInt, DateTime, Bytes)>` type (let's call it "the complex type" from now on ^^) comes from the API protocol, so we have to deal with it no matter what.

From this we have two choices:

1) deal with the complex type in `fetch_versions_remote_workspace_manifest`  (this is the current solution)
2) deal with the complex type in `fetch_versions_vlob` so that `fetch_versions_remote_workspace_manifest` can work with a better type (this is your proposed solution)

The trick is no matter what we do, we will end up with a loop dealing with the complex type (in case 1 it is to do the actual work, in case 2 it is to convert it to a better type).

So in the end I don't think there is a gain here.

(but again, if `VlobVersionsData` were a public structure, the logic would be much different and we would most likely benefit from having a cleaner type)

AureliaDolo: At the end of the day, it's a matter of personal taste ^^

I like having named fields when possible. But a comment explain what are the field is enough.

touilleMan: I've added a comment to provide a better documentation on the meaning of each field (and also why adding why there is no dedicated type in use here)

> At the end of the day, it's a matter of personal taste ^^

To make you feel better, the performance should be better when reusing the tuple (since the memory layout between the tuple and a dedicated type is likely different[^1] and hence the compiler cannot re-use the vec as-is) :trollface: 

[^1]: my guess is the layout is the same as long as the parameters are declared in the same order, but this is implicit without strong guarantee, so I wouldn't rely on that ^^

PR #8675, opened 2024-10-08
Add option to enable proxy header

### LINKED ISSUE #8427: [🐛 | Bug]: Organization bootstrap fails
### Parsec version tested on:

v3.0.0-rc.8

### Platforms tested on:

Windows

### Bug description:

On custom server while bootstrap (join organization with bootstrap link), client received an `BootstrapOrganizationErrorInternal` error

### Relevant output:

Server capture on debug level
Capture d’écran 2024-09-18 094414

### LINKED ISSUE #8626: [Server] Add option to handle header set by a proxy
A Parsec server could be configured behind a reverse proxy (like apache or traefik).
But currently the server does not use the headers added by the proxy (`Forwarded-*` or `X-Forwarded-*`).

> [!IMPORTANT]
> The server should not use those headers by default, the sysadmin should explicitly enable the server to use those headers.

This could cause issue when the server generate link like for bootstrapping an organization where it only use the scheme of the request instead of looking for the forwarded header:

https://github.com/Scille/parsec-cloud/blob/6a56633e5483b611d4c0cf21132c73fbe67ad213/server/parsec/asgi/administration.py#L179

We could take inspiration of `keycloak` has the optional option `--proxy-headers {forwarded|xforwarded}`

FirelightFlagboy:
Add a new option that configure the list of trusted address to parse the proxy headers from.
The parsing is handled by `uvicorn` which currently only support `x-forwarded-{for,proto}` headers.

Closes #8626, closes #8427

### DIFF
--- a/docs/hosting/deployment/parsec.env
+++ b/docs/hosting/deployment/parsec.env
@@ -7,9 +7,6 @@ PARSEC_SSL_KEYFILE=/run/secrets/parsec-pem-key
 # The SSL certificate file.
 PARSEC_SSL_CERTFILE=/run/secrets/parsec-pem-crt
 
-# Enforce HTTPS by redirecting HTTP request.
-PARSEC_FORWARD_PROTO_ENFORCE_HTTPS=X-Forwarded-Proto:https
-
 # The granularity of Error log outputs.
 PARSEC_LOG_LEVEL=WARNING
 
--- a/newsfragments/8626.feature.rst
+++ b/newsfragments/8626.feature.rst
@@ -1,0 +1,1 @@
+Add server option ``--proxy-trusted-addresses`` to enable parsing of proxy headers from trusted addresses. By default, the server will trust the proxy headers from localhost.
--- a/server/parsec/asgi/__init__.py
+++ b/server/parsec/asgi/__init__.py
@@ -60,25 +60,11 @@ def app_factory() -> AsgiApp:
 app: AsgiApp = app_factory()
 
 
-# TODO: implement forward_proto_enforce_https
-#     # Do https redirection if incoming request doesn't follow forward proto rules
-#     if backend.config.forward_proto_enforce_https:
-#         header_key, header_expected_value = backend.config.forward_proto_enforce_https
-
-#         @app.before_request
-#         def redirect_unsecure() -> ResponseReturnValue | None:
-#             header_value = request.headers.get(header_key)
-#             # If redirection header match and protocol match, then no need for a redirection.
-#             if header_value is not None and header_value != header_expected_value:
-#                 if request.url.startswith("http://"):
-#                     return quart_redirect(request.url.replace("http://", "https://", 1), code=301)
-#             return None
-
-
 async def serve_parsec_asgi_app(
     app: AsgiApp,
     host: str,
     port: int,
+    proxy_trusted_addresses: list[str],
     ssl_certfile: Path | None = None,
     ssl_keyfile: Path | None = None,
     workers: int | None = None,
@@ -93,6 +79,7 @@ async def serve_parsec_asgi_app(
         v_major, _ = parsec_version.split(".", 1)
         # ex: parsec/3
         server_header = f"parsec/{v_major}"
+
     # Note: Uvicorn comes with default values for incoming data size to
     # avoid DoS abuse, so just trust them on that ;-)
     config = uvicorn.Config(
@@ -106,6 +93,13 @@ async def serve_parsec_asgi_app(
         ssl_keyfile=ssl_keyfile,  # type: ignore
         ssl_certfile=ssl_certfile,
         workers=workers,
+        # Enable/Disable X-Forwarded-Proto, X-Forwarded-For to populate remote address info.
+        # When enabled, is restricted to only trusting connecting IPs in forwarded-allow-ips.
+        # See: https://www.uvicorn.org/settings/#http
+        # Currently uvicorn only supports X-Forwarded-* headers (https://github.com/encode/uvicorn/issues/2237)
+        proxy_headers=(proxy_trusted_addresses != []),
+        # Comma separated list of IP Addresses, IP Networks, or literals (e.g. UNIX Socket path) to trust with proxy headers
+        forwarded_allow_ips=proxy_trusted_addresses,
         # TODO: configure access log format:
         # Timestamp is added by the log processor configured in `parsec.logging`,
         # here we configure peer address + req line + rep status + rep body size + time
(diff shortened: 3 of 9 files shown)


server/parsec/asgi/__init__.py
```diff
@@ -106,6 +93,12 @@ async def serve_parsec_asgi_app(
         ssl_keyfile=ssl_keyfile,  # type: ignore
         ssl_certfile=ssl_certfile,
         workers=workers,
+        # Enable/Disable X-Forwarded-Proto, X-Forwarded-For to populate remote address info.
+        # Currently uvicorn only supports X-Forwarded headers
+        # An issue exist to support Forwarded headers: https://github.com/encode/uvicorn/issues/2237
```
mmmarcos, line 98:
```suggestion
        # Enable/Disable X-Forwarded-Proto, X-Forwarded-For to populate remote address info.
        # When enabled, is restricted to only trusting connecting IPs in forwarded-allow-ips.
        # See: https://www.uvicorn.org/settings/#http
        # Currently uvicorn only supports X-Forwarded-* headers (https://github.com/encode/uvicorn/issues/2237)
```

docs/hosting/deployment/parsec.env
```diff
@@ -7,9 +7,6 @@ PARSEC_SSL_KEYFILE=/run/secrets/parsec-pem-key
 # The SSL certificate file.
 PARSEC_SSL_CERTFILE=/run/secrets/parsec-pem-crt
 
-# Enforce HTTPS by redirecting HTTP request.
-PARSEC_FORWARD_PROTO_ENFORCE_HTTPS=X-Forwarded-Proto:https
```
mmmarcos, line 11: I would document the added variable 
```env
# Comma-separated list of IP Addresses, IP Networks or literals to trust with proxy headers.
# Set this value to allow the server to use X-Forwarded-* headers from those clients.
# PARSEC_PROXY_TRUSTED_ADDRESSES=
```

server/parsec/cli/run.py
```diff
@@ -292,20 +279,15 @@ def handle_parse_result(
...
-    help=(
-        "Enforce HTTPS by redirecting incoming request that do not comply with the provided header."
-        " This is useful when running Parsec behind a forward proxy handing the SSL layer."
-        " You should *only* use this setting if you control your proxy or have some other"
-        " guarantee that it sets/strips this header appropriately."
-        " Typical value for this setting should be `X-Forwarded-Proto:https`."
-    ),
+    help="""\b
+        Comma-separated list of IP Addresses, IP Networks or literals to trust with proxy headers.
+        Settings this value will allow the server to use the forwarded headers from those clients.
```
mmmarcos, line 289:
```suggestion
        Set this value to allow the server to use X-Forwarded-* headers from those clients.
```

newsfragments/8626.feature.rst
```diff
@@ -0,0 +1,8 @@
+Add server option ``--proxy-trusted-addresses`` to enabled parsing of proxy headers from trusted addresses.
+
+.. important::
+  Breaking change: Previous the server used the default value provided by ``uvicorn`` which always enabled the proxy header for localhost proxy.
+  If you want to have the same behavior as before, you need to set the ``--proxy-trusted-addresses`` to ``localhost,127.0.0.1,::1``.
+
+.. tip::
+  The env variable ``PARSEC_PROXY_TRUSTED_ADDRESSES`` instead of the option.
```
mmmarcos, line 8: Better avoid fancy formatting and stick to oneliners for changelog (see below)
```suggestion
Add server option ``--proxy-trusted-addresses`` to enable parsing of proxy headers from trusted addresses.
```

server/parsec/asgi/__init__.py
```diff
@@ -106,6 +93,12 @@ async def serve_parsec_asgi_app(
...
+        proxy_headers=(proxy_trusted_addresses != []),
+        # Comma separated list of IP Addresses, IP Networks, or literals (e.g. UNIX Socket path) to trust with proxy headers
+        forwarded_allow_ips=proxy_trusted_addresses,
```
mmmarcos, line 101: As stated in uvicorn docs regarding `forwarded_allow_ips`: 
> Defaults to the $FORWARDED_ALLOW_IPS environment variable if available, **or '127.0.0.1'**. 

Why change the default behavior? 
Why not leaving `proxy_headers` enabled (the default value) and setting `forwarded_allow_ips` as:

```suggestion
        forwarded_allow_ips=proxy_trusted_addresses+["localhost", "127.0.0.1", "::1"],
```

FirelightFlagboy: The main reason is explicit vs implicit configuration.

Previously we cannot disable the parsing of forwarded header. Now we can and allow the user to provide a specific list.

I will not apply your suggestion as your are hard-coding values that cannot be overwritten by the user.

newsfragments/8626.feature.rst
FirelightFlagboy: I'd like to have the migration steps in the changelog

docs/hosting/deployment/parsec.env
FirelightFlagboy: Should be done in #8676

server/parsec/cli/run.py
FirelightFlagboy:
```suggestion
        Set this value to allow the server to use the forwarded headers from those clients.
```

Don't specify `X-Forwarded` in case `uvicorn` add support for it.

server/parsec/asgi/__init__.py
mmmarcos: OK, so you suggest to set those default values somewhere else?

FirelightFlagboy: done, in another way (default value for option)

newsfragments/8626.feature.rst
FirelightFlagboy: outdated

newsfragments/8626.feature.rst
```diff
@@ -0,0 +1,5 @@
+Add server option ``--proxy-trusted-addresses`` to enabled parsing of proxy headers from trusted addresses.
+By default, the server will trust the proxy headers from localhost
+
+.. tip::
+  You can also configure the trusted addresses using the environment variable ``PARSEC_PROXY_TRUSTED_ADDRESSES``.
```
mmmarcos, line 5: I reiterate my previous outdated comment:

We should still stick to one-liner without fancy formatting, as it is the case for [almost all changelog entries](https://github.com/Scille/parsec-cloud/blob/master/HISTORY.rst?rgh-link-date=2024-10-09T10%3A09%3A28Z). If more detail is needed, it should be added to the issue referenced by the news fragment (https://github.com/Scille/parsec-cloud/issues/8626) which will be displayed in the changelog.

```suggestion
Add server option ``--proxy-trusted-addresses`` to enable parsing of proxy headers from trusted addresses. By default, the server will trust the proxy headers from localhost
```

mmmarcos (approved):

PR #8682, opened 2024-10-09
Improve confinement logic by adding the UnconfinedLocalFolderManifest type

### LINKED ISSUE #7815: Improve existing confinement point design
This issue is a follow-up to the following discussions:
- https://github.com/Scille/parsec-cloud/pull/7756#pullrequestreview-2197086671
- https://github.com/Scille/parsec-cloud/pull/7756#issuecomment-2254466406

The last comment suggests to introduce a new type called `UnconfinedLocalFolderManifest` that has the same fields as `LocalFolderManifest`, except for the `local_confinement_points` and `remote_confinement_points` fields which are absent.

This type would implement the following type conversions:
```
RemoteFolderManifest -> UnconfinedLocalFolderManifest -> RemoteFolderManifest
                            v                 ^
                            LocalFolderManifest
```

Those 4 types conversions are implemented using the following methods in `UnconfinedLocalFolderManifest`:
- `to_remote(self) -> RemoteManifest`
- `from_remote(remote: RemoteManifest) -> Self`
- `apply_confinement(self, existing_local_manifest: Option<LocalManifest>, prevent_sync_pattern: Pattern) -> LocalManifest`
- `remove_confinement(local_manifest: LocalManifest) -> Self`

This way, the full confinement logic is maintained in `apply_confinement` and `remove_confinement`.  More precisely:
- `apply_confinement` is equivalent to the the old `filter_remote_entries + restore_local_confinement_points`
- `remove_confinement` is equivalent to the the old `filter_local_confinement_points + restore_remote_confinement_points`

With this design, the semantics of the fields `remote/local_confinement_points` of `LocalFolderManifest` don't change over time, and more robust invariants can be defined and enforced.

Note that `LocalFolderManifest.apply_prevent_sync_pattern` is now defined as:
```rust
local_folder_manifest
	.remove_confinement(local_manifest) // -> UnconfinedLocalFolderManifest
	.apply_confinement(local_manifest, new_prevent_sync_pattern) // -> LocalFolderManifest
```

The conversion of a remote manifest into a local manifest is:
```rust
UnconfinedLocalFolderManifest::from_remote(remote_manifest)
	.apply_confinement(existing_local_manifest, prevent_sync_pattern) // -> LocalFolderManifest
```

And the conversion of a local manifest into a remote manifest is:
```rust
local_folder_manifest
	.remove_confinement(local_manifest) // -> UnconfinedLocalFolderManifest
	.to_remote() // -> FolderManifest
```

The motivation for this change is to improve maintainability and readability, although the core logic is not expected to change (unless a specific bug is found during implementation).

Note: the merging of `filter_remote_entries` and `restore_local_confinement_points` together into `apply_confinement` would allow to deal more explicitly with the subtleties explained in this comment: https://github.com/Scille/parsec-cloud/pull/7756#discussion_r1694001074

Somewhat related: https://github.com/Scille/parsec-cloud/issues/7738

### LINKED ISSUE #7924: Investigate failing tests related to `apply_prevent_sync_pattern` & `merge_folder_manifests`

vxgmichel:
Fix #7815
Fix #7924

### DIFF
--- a/libparsec/crates/client/src/workspace/merge.rs
+++ b/libparsec/crates/client/src/workspace/merge.rs
@@ -249,10 +249,10 @@ pub(super) fn merge_local_folder_manifest(
     let LocalFolderManifest {
         base:
             FolderManifest {
-                id: local_base_id,
-                version: local_base_version,
-                children: local_base_children,
-                parent: local_base_parent,
+                id: _,
+                version: _,
+                children: _,
+                parent: _,
                 // Ignored, we don't merge data that change on each sync
                 author: _,
                 // `created` should never change, so in theory we should have
@@ -266,7 +266,7 @@ pub(super) fn merge_local_folder_manifest(
                 updated: _,
             },
         children: _,
-        parent: local_parent,
+        parent: _,
         need_sync: _,
         speculative: local_speculative,
         // Ignored, that field is merged in `from_remote_with_local_context`
@@ -278,10 +278,10 @@ pub(super) fn merge_local_folder_manifest(
     } = local;
 
     // 0) Sanity checks, caller is responsible to handle them properly !
-    debug_assert_eq!(local_base_id, &remote.id);
+    debug_assert_eq!(local.base.id, remote.id);
 
     // 1) Shortcut in case the remote is outdated
-    if remote.version <= *local_base_version {
+    if remote.version <= local.base.version {
         return MergeLocalFolderManifestOutcome::NoChange;
     }
 
@@ -373,7 +373,7 @@ pub(super) fn merge_local_folder_manifest(
         } = &mut merge_in_progress;
 
         *merged_speculative = false;
-        *merged_parent = *local_parent;
+        *merged_parent = local.parent;
         *merged_need_sync = local.need_sync;
         *merged_updated = local.updated;
         local.children.clone_into(merged_children);
@@ -385,7 +385,7 @@ pub(super) fn merge_local_folder_manifest(
 
     // Solve the folder conflict
     let merged_children = merge_children(
-        local_base_children,
+        &local.base.children,
         &local.children,
         // Why not using `remote.children` here ?
         // This is to handle the confinement points: given `merge_children` has no
@@ -398,7 +398,7 @@ pub(super) fn merge_local_folder_manifest(
         // entries are also present with same ID&name in `local_children`).
         &merge_in_progress.children,
     );
-    let merged_parent = merge_parent(*local_base_parent, *local_parent, remote.parent);
+    let merged_parent = merge_parent(local.base.parent, local.parent, remote.parent);
 
     // Children merge can end up with nothing to sync.
     //
--- a/libparsec/crates/client/src/workspace/merge.rs
+++ b/libparsec/crates/client/src/workspace/merge.rs
@@ -265,7 +265,7 @@ pub(super) fn merge_local_folder_manifest(
                 // Ignored, we don't merge data that change on each sync
                 updated: _,
             },
-        children: local_children,
+        children: _,
         parent: local_parent,
         need_sync: local_need_sync,
         speculative: local_speculative,
@@ -386,7 +386,7 @@ pub(super) fn merge_local_folder_manifest(
     // Solve the folder conflict
     let merged_children = merge_children(
         local_base_children,
-        local_children,
+        &local.children,
         // Why not using `remote.children` here ?
         // This is to handle the confinement points: given `merge_children` has no
         // concept of confinement points, it will see as conflict any confined entry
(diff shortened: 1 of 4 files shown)


libparsec/crates/client/src/workspace/merge.rs
FirelightFlagboy, line 1: Could you revert the change in this file ? they are not required, outside of the change made to `let merge_need_sync = ...`

libparsec/crates/types/src/local_manifest/folder.rs
```diff
@@ -203,7 +203,10 @@ impl LocalFolderManifest {
                     actually_updated = true;
                 }
                 // Add new entry
-                self.children.insert(name, entry_id);
+                if let Some(previous_id) = self.children.insert(name, entry_id) {
+                    // Remove previous id from local confinement points if it was there
+                    self.local_confinement_points.remove(&previous_id);
+                };
```
FirelightFlagboy, line 209: That step is already done at the lines 183-187 (removing children from the map & local confinement points), why it's needed again ?

libparsec/crates/types/src/local_manifest/folder.rs
```diff
@@ -215,238 +218,254 @@ impl LocalFolderManifest {
...
+            self,
+            prevent_sync_pattern,
+            timestamp,
+        )
+    }
 
-        self.local_confinement_points.clear();
-        self
+    pub fn from_remote(remote: FolderManifest, prevent_sync_pattern: &Regex) -> Self {
+        UnconfinedLocalFolderManifest::apply_confinement_from_remote(remote, prevent_sync_pattern)
```
FirelightFlagboy, line 234: For me, this method should not exist in `UnconfinedLocalFolderManifest` since it produce a `LocalFolderManifest`.

I would have expected that we go that route: `RemoteFolder -> UnconfinedLocalFolder -> LocalFolder`

Is it possible to make that explicit ?

FirelightFlagboy (changes requested):

libparsec/crates/types/src/local_manifest/folder.rs
vxgmichel: Very good catch, I read too fast and thought the the steps 183-187 were only done for explicitly removed entries. I'll just remove this bit.

libparsec/crates/client/src/workspace/merge.rs
vxgmichel: They are actually required. Some bugs were due to the fact the local was replaced by a new version of local with the prevent sync pattern applied:
```rust
let local = &local.apply_prevent_sync_pattern(prevent_sync_pattern, timestamp);
```
But the destruct was done before this process, so some of the extracted variables got outdated. This was quite confusing so I just removed them all but kepts the destruct because of this comment:
```
// Destruct local and remote manifests to ensure this code with fail to compile whenever a new field is introduced.
```
Let met know if you have a suggestion, but I find it less error prone as it is.

libparsec/crates/types/src/local_manifest/folder.rs
vxgmichel: Ahah I kind of hoped it would go unnoticed ^^

So yes, `UnconfinedLocalFolderManifest::apply_confinement_from_remote` is conceptually similar to 
```rust
UnconfinedLocalFolderManifest::from_remote(remote).apply_confinement(None, prevent_sync_pattern, None)
```
if `apply_confinement` were to take `existing_local_manifest` as an `Option<LocalFolderManifest>` (and same for the timestamp).

So I had to choose between:
- 1. Doing just that, using optional arguments, essentially removing `apply_confinement_from_remote` entirely. I didn't like it because `apply_confinement` is already pretty tricky as it is, and it's not obvious at first glance than it will behave correctly for the simple case where the existing manifest doesn't exist. More specifically, in the case of `apply_confinement_from_remote`, you know that the produced `UnconfinedLocalFolderManifest` has its children equal to its base children, which is not the case in general for `apply_confinement` (e.g when it's used to change the prevent sync pattern). I figured, better be explicit with this specific case, as it seems more readable and allows for optimizations.
- 2. I also could have put this specific logic in `LocalFolderManifest::from_remote`, which is where it belongs. But I wanted the confinement handling implementation to be gathered in a single place, for better readability, and so logic can be compared.
- 3. Which is why I ended up specializing `apply_confinement_from_remote` as a dedicated static method.

So I could be convinced that approach 2 is less confusing, but I'd rather stay away from approach 1.

FirelightFlagboy: So we could move `apply_confinement_from_remote` to `LocalFolderManifest` without putting the logic into `from_remote`.

vxgmichel: So you're suggesting approach 2 + renaming `LocalFolderManifest::from_remote` to `LocalFolderManifest::apply_confinement_from_remote`? This does not change the point I made in my comment, I would like to keep `apply_confinement_from_remote` close to `apply_confinement` since the latter is a specialized implementation of the former.

FirelightFlagboy (approved):

Issue #8198, opened 2024-08-28
Manage confined entries in synchronization monitor


vxgmichel:
As mentioned in [this conversation](https://github.com/Scille/parsec-cloud/pull/8192#discussion_r1733128470), the `WorkspaceOpsOutboundSyncEntryConfined` event has not been implemented.

In parsec v2, it was emitted when a confined entry tries to get synchronized: 
https://github.com/Scille/parsec-cloud/blob/d397c035d1369e498a48dea26837d93b7291dfe4/parsec/core/fs/workspacefs/sync_transactions.py#L362-L369

It was linked to a handler in the sync monitor:
https://github.com/Scille/parsec-cloud/blob/d397c035d1369e498a48dea26837d93b7291dfe4/parsec/core/sync_monitor.py#L445-L452

Which allowed the sync monitor to keep track of the confined entries and their respective cause.

This information was then used here:
https://github.com/Scille/parsec-cloud/blob/d397c035d1369e498a48dea26837d93b7291dfe4/parsec/core/sync_monitor.py#L176-L188

This way, all the confined entries corresponding to a given cause would get re-integrated into the synchronization system when the cause gets updated.

mmmarcos:
According to @vxgmichel this still needs to be implemented, but it is not a priority since #8068 was fixed by #8192.

PR #8710, opened 2024-10-12
Improve confinement

### LINKED ISSUE #7885: `merge_local_folder_manifest` should expect `local` to respect the `prevent_sync_pattern`
In parsec v2,  `merge_local_folder_manifest` accept `local` manifest with arbitrary prevent sync pattern.

The idea was it was a very unlikely  event (a new prevent sync pattern have been set and is not fully applied), and in this case a flag could be passed to tell `merge_local_folder_manifest` to manually apply the new prevent sync pattern to `local`

https://github.com/Scille/parsec-cloud/blob/cb5cce5972e8e8228069762bbf2c0e8becd23c29/parsec/core/fs/workspacefs/sync_transactions.py#L201-L204

In practice however this was not needed since the prevent sync pattern is only set when starting the `WorskpaceFS`, which in turn ensure the prevent sync pattern is fully applied as part of it initialization routine

https://github.com/Scille/parsec-cloud/blob/cb5cce5972e8e8228069762bbf2c0e8becd23c29/parsec/core/fs/workspacefs/workspacefs.py#L222-L223

On top of that, having `merge_local_folder_manifest` accepting outdated `local` makes the function significantly harder to test (see https://github.com/Scille/parsec-cloud/pull/7756/files#diff-4e496067608e0a1d8d2067e4121b7465f612c1007dc9f35128cf0054ab3adc99R605-R625)

So we could simplify this by just considering `merge_local_folder_manifest` only accept `local` manifest in sync with `prevent_sync_pattern` (and adding a docstring to this fonction to document this !)

vxgmichel:
More improvements over #8682

Fix #7885

Namely:

1. Swap the logic in `merge_local_folder_manifest`: the original implementation performed the merging of the remote and local children using confined version of the correspond manifests (i.e applying the local confinement context on the remote). But in order to that, both manifests need to use the same `prevent_sync_pattern` for consistency, so there had to be an extra step of applying the new prevent sync pattern on the local manifest, even though it hasn't changed in basically all cases. This made the implementation of `merge_local_folder_manifest` more confusing than it needs to be: instead, the merging of the children can be made on the **unconfined** version of the corresponding manifests. Now the manifests have no reasons to be outdated, which remove the need for the extra conversion entirely. 
   To sum it up, the new implementation does this:
   ```
        local manifest
            |
            [remove confinement]
            |
            v
        unconfined local manifest
            |
            [merge] <- using unconfined remote manifest 
            |
            v
        new unconfined local manifest
            |
            [apply confinement] <- using local manifest 
            |
            v
        new local manifest
   ```
   Note that this process is similar to applying a new pattern (or re-applying the existing one) if the remote manifest includes no changes.
   Note that this doesn't affect the overall behavior of `merge_local_folder_manifest` aside for some specific cases here and there.
   Note that this fixes #7885

2. A change to `apply_confinement` allows clients to keep their synchronized files locally when the happen to match a new prevent_sync_pattern. In the previous implementation, the file would have simply been removed. This made sense because in the general case, we want to avoid referencing an id locally (and possibly update its corresponding content) when it's also remotely confined. But in the special case where the entry is both confined locally and remotely, there is no problem keeping the id both locally and remotely, as long as the corresponding confined entry is not synchronized. Handling this specific case then allows for client to keep their temporary files when their names starts matching the prevent sync pattern, which is likely a desirable behavior.
   However, it depends on the following issue being addressed, which is not the case at the moment: #8198


libparsec/crates/client/src/workspace/merge.rs
```diff
@@ -308,30 +297,28 @@ pub(super) fn merge_local_folder_manifest(
...
 
-    // 5) The remote changes are ours (our current local changes occurs while
+    // The following case was present in the original v2 code.
+    // TODO: investigate why it's not present anymore and check whether
+    // it is still relevant.
+    // ```python
+    // # All the local changes have been successfully uploaded
+    // if local_manifest.match_remote(remote_manifest):
+    //     return local_from_remote
+    // ```
```
FirelightFlagboy, line 319: Is it something you will investigate during this PR or at a later time ?

If it's the later, could you create an issue and add its link in the TODO command ?

libparsec/crates/types/src/local_manifest/folder.rs
```diff
@@ -405,17 +415,28 @@ impl UnconfinedLocalFolderManifest {
...
+                // Case 1
                 if existing_local_manifest
                     .local_confinement_points
                     .contains(entry_id)
                 {
-                    // Insert the locally confined entry in the new manifest
-                    existing_local_confined_entries.insert(name.clone(), Some(*entry_id));
-
                     // Perform a rename if the entry id is already present in the new manifest
                     if let Some(&previous_name) = new_entry_ids.get(entry_id) {
```
FirelightFlagboy, line 424: Should we check if `previous != name` ?

libparsec/crates/client/src/workspace/merge.rs
vxgmichel: The issue has been created: https://github.com/Scille/parsec-cloud/issues/8750

> Is it something you will investigate during this PR or at a later time ?
I'm not entirely sure yet

libparsec/crates/types/src/local_manifest/folder.rs
vxgmichel: 68d9f8304ecbabeec2e568d953f194162c09d219

libparsec/crates/client/src/workspace/merge.rs
vxgmichel: I don't have time to fix this issue, but I added a bit more information

libparsec/crates/types/src/local_manifest/folder.rs
vxgmichel: Good point :+1:

PR #8720, opened 2024-10-14
[MS] Serialize objects when logging them

Max-7:
- [X] Describe your changes

### DIFF
--- a/client/src/views/organizations/TOSModal.vue
+++ b/client/src/views/organizations/TOSModal.vue
@@ -84,7 +84,9 @@ function getLinkForLocale(): string {
 
   window.electronAPI.log(
     'info',
-    `No terms of service link was found matching the locale '${currentLocale}', using '${links.values().next().value}'`,
+    `No terms of service link was found matching the locale \
+'${currentLocale}' in ${JSON.stringify(props.tosLinks.entries())}, \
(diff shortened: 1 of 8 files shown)


client/src/views/organizations/TOSModal.vue
```diff
@@ -84,7 +84,9 @@ function getLinkForLocale(): string {
 
   window.electronAPI.log(
     'info',
-    `No terms of service link was found matching the locale '${currentLocale}', using '${links.values().next().value}'`,
+    `No terms of service link was found matching the locale \
+'${currentLocale}' in ${JSON.stringify(props.tosLinks.entries())}, \
+using '${links.values().next().value}'`,
   );
   // Still not found, worst cast scenario, return the first value
   return links.values().next().value;
```
FirelightFlagboy, line 92:
```suggestion
  const first_link = links.values().next().value;
  window.electronAPI.log(
    'info',
    `No terms of service link was found matching the locale \
'${currentLocale}' in ${JSON.stringify(props.tosLinks.entries())}, \
using '${first_link}'`,
  );
  // Still not found, worst cast scenario, return the first value
  return first_link;
```

I recommend creating the var `first_link` to not repeat the logic to retrieve the default link

NicoTuxx (approved):

PR #8745, opened 2024-10-16
Rework sentry release creation

FirelightFlagboy:


server/parsec/logging.py
```diff
@@ -319,7 +320,8 @@ def enable_sentry_logging(dsn: str, environment: str) -> None:
     sentry_sdk.init(
         dsn=dsn,
         environment=environment,
-        release=__version__,
+        release="parsec@" + __version__,
+        dist=platform.system(),
```
mmmarcos, line 324: Not sure this is useful because this information is already included by Sentry as a tag. But I guess it doesn't do any harm to keep it.

mmmarcos (commented):
LGTM overall, except for:

- release name does not seem consistent between
  - `parsec@...` in vite config
  - `parsec-electron@...` in electron config
  - `parsec@...` in server config
- release name should also be configured for libparsec:
https://github.com/Scille/parsec-cloud/blob/046db762af85815d14d381f6e0421596dd5cf240/bindings/electron/src/lib.rs#L20-L30

PR #8752, opened 2024-10-18
Quick and dirty fix for macOS mountpoint rename issue

### LINKED ISSUE #6144: Implement mountpoint for macOS

FirelightFlagboy:
Closes #6144


FirelightFlagboy:
Superseded by #8766

PR #8766, opened 2024-10-21
Patch `fuser` with a fix for MacFUSE rename op

FirelightFlagboy:


Cargo.toml
```diff
@@ -211,6 +211,9 @@ x25519-dalek = { version = "2.0.1", default-features = false }
 zeroize = { version = "1.8.1", default-features = false }
 zstd = { version = "0.13.2", default-features = false }
 
+[patch.crates-io]
```
touilleMan, line 214: Should add a comment explaining why we rely on this fork, and also a `TODO` comment linking an issue to track when we can switch back to the main release of fuser

touilleMan (approved):

Cargo.toml
FirelightFlagboy: done

PR #8770, opened 2024-10-22
Split shamir setup into two retry loops

### LINKED ISSUE #8206: Infinite loop when using `Client::shamir_setup_create`
## The issue

```
$ cargo nextest r -p libparsec_client client::tests::shamir_setup_create::ok
        SLOW [> 10.000s] libparsec_client client::tests::shamir_setup_create::ok
        SLOW [> 20.000s] libparsec_client client::tests::shamir_setup_create::ok
 TERMINATING [> 30.000s] libparsec_client client::tests::shamir_setup_create::ok
     TIMEOUT [  30.006s] libparsec_client client::tests::shamir_setup_create::ok
   Canceling due to test failure
------------
     Summary [  30.007s] 1 test run: 0 passed, 1 timed out, 738 skipped
     TIMEOUT [  30.006s] libparsec_client client::tests::shamir_setup_create::ok
error: test run failed
```

When looking at the requests to the testbed server we see plenty of `device_create ok` then `shamir_recovery_setup require_greater_timestamp`

The issue comes from the fact the same timestamp is used to generate the payload of both `device_create` and `shamir_recovery_setup`, hence the former request detects the timestamp is equal to the one in the current common topic !

https://github.com/Scille/parsec-cloud/blob/e9a0189dc12a963a1a00918cc7ac502c80fab1aa/libparsec/crates/client/src/certif/shamir_recovery_setup.rs#L223-L295

## The fix

Obviously a different timestamp must be generated for the data of each command.

However I think things should be refactored a bit further.

From the client point of view, shamir recovery setup consist of two distinct operations:
1) obtain a device suitable for shamir recovery
2) actually submit the `shamir_recovery_setup` command to the server

In the current code we have bundled together the two operations, if 2 fails gets a require_greater_timestamp then also we re-do 1.

I think it would be better to split function `shamir_internals` into `create_shamir_recovery_device` and `do_shamir_recovery_setup`, each function being called within `shamir_internals` with it own dedicated retry loop (and hence timestamp handling).

vxgmichel:
Fix #8206

### DIFF
--- a/libparsec/crates/client/src/certif/mod.rs
+++ b/libparsec/crates/client/src/certif/mod.rs
@@ -605,19 +605,10 @@ impl CertificateOps {
 
     pub async fn shamir_setup_create(
         &self,
-        author_device: DeviceID,
-        author_id: UserID,
         share_recipients: HashMap<UserID, u8>,
         threshold: u8,
     ) -> Result<CertificateBasedActionOutcome, CertifShamirError> {
-        shamir_recovery_setup::shamir_setup_create(
-            self,
-            author_device,
-            author_id,
-            share_recipients,
-            threshold,
-        )
-        .await
+        shamir_recovery_setup::shamir_setup_create(self, share_recipients, threshold).await
     }
 }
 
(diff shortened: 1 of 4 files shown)


FirelightFlagboy (approved):

AureliaDolo (approved):
Do we have something specific in case the device certificate is uploaded but not the shamir one ?

vxgmichel:
> Do we have something specific in case the device certificate is uploaded but not the shamir one ?

In this case there's not much to do, we simply ignore the created device.

PR #8772, opened 2024-10-22
[MS] Add go back to login in org switch popover

### LINKED ISSUE #8724: Where is home?
As a user, I find it a quite upsetting not to have a clear notion of "home" page or how to come back to it. 

For me, the "home" page is the one listing the workspaces (because it is the first page shown on login), but:
- there is no clear way to come back to this page
- AND the app uses "home" for other pages too!

For example, take the following screen:
image

A user may ask itself, how do I return to the "home" page? 

- Maybe if I click `Go to home`? --> WRONG this button takes back to the (Welcome?) organization page :dizzy_face: 
- Maybe is `Go back`? --> WRONG this depends on where I was before opening My Profile
- Maybe is the `Workspaces` in the left panel --> NOPE, not clickable
- Maybe there is a home button in the main menu? --> NIET

Moreover, it's also disturbing to having different notions of "Home" in the same page. Here, the same icon is used for different pages:
image

----

The home page usually acts as an anchor for the user to feel it is back to a "known" place.

So my suggestions:
- Use "Welcome" or "Login" page for the page listing the organizations/devices. 
- Use "Home" for the page listing the workspaces
- Use different icons for these two pages.
- Ensure that at least the "Home" page is always available (and if possible in the same place) no matter where the user is.

Max-7:
Closes #8724 

Two things to note:
1. The button "Back to workspaces" is hidden is we are on the workspaces page
2. It's not possible to go to the customer area from here, we don't know if the current user has one.

- [X] Describe your changes


client/src/components/organizations/OrganizationSwitchPopover.vue
```diff
@@ -16,6 +16,14 @@
           {{ currentOrg?.id }}
         </ion-text>
       </div>
+      <ion-button
+        @click="goToLogin"
+        fill="outline"
+      >
+        <ion-icon :icon="repeat" />
```
Ironicbay, line 23: We could maybe use `swap-horizontal` ?

client/src/locales/en-US.json
```diff
@@ -1227,7 +1227,9 @@
         "loggedInOrgs": "Connected organizations",
         "myOrgs": "My organizations",
         "invite": "Invite a member",
-        "manage": "Manage"
+        "manage": "Manage",
+        "goToLogin": "Switch organization",
```
Ironicbay, line 1231: This is a bit misleading since we can switch to another logged in organization from this popover. But I'm not sure what to write instead, it would be nice to mention that it's to log in to a logged-off organization.

client/src/views/sidebar-menu/SidebarMenuPage.vue
```diff
@@ -387,9 +383,7 @@ const isTrialOrg = ref(false);
 const expirationDuration = ref<Duration | undefined>(undefined);
 const isExpired = ref(false);
 const loggedInDevices = ref<LoggedInDeviceInfo[]>([]);
-const canSwitchOrg = computed(() => {
-  return isDesktop() && loggedInDevices.value.length > 1;
-});
+const canSwitchOrg = ref(true);
```
Ironicbay, line 386: This variable seems obsolete

client/src/locales/en-US.json
```diff
@@ -1227,7 +1227,9 @@
...
+        "goToCustomerArea": "Go to customer area"
```
Ironicbay, line 1232: Is this translation added by mistake ?

Ironicbay (changes requested):

Ironicbay (approved):

client/src/components/organizations/OrganizationSwitchPopover.vue
```diff
@@ -16,6 +16,21 @@
           {{ currentOrg?.id }}
         </ion-text>
       </div>
+      <div
+        class="section-buttons custom-button custom-button-fill"
+        @click="goToLogin"
+      >
+        <ion-icon
```
Ironicbay, line 23: Missing `IonIcon` import

client/src/components/organizations/OrganizationSwitchPopover.vue
```diff
@@ -112,11 +128,13 @@ onMounted(async () => {
 
 async function onOrganizationClick(org: ConnectedOrganization): Promise<void> {
   if (org.handle !== getConnectionHandle()) {
-    await popoverController.dismiss({
-      handle: org.handle,
-    });
+    await popoverController.dismiss({ handle: org.handle }, MsModalResult.Confirm);
```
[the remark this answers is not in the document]
Max-7: Web is a special case: everything is mocked but the `listDevices`. So it displays "real" devices, but when you try to log with them, it's all fake. No real login, no real workspaces, files, users. It uses a fake handle (the same for all devices).

It doesn't work because you're trying to go from `/42/workspaces` to `/42/workspaces`, which is exactly the same URL, so no reason to navigate.

It could be possible to use different handles and all that but it's web, nothing is supposed to work. When it seems like it's doing something, it's only because we faked it to look that way.

Issue #8726, opened 2024-10-14
[🐛 | Bug]: Email field is not fully visible on Contact details


mmmarcos:
### Parsec version tested on:

3.1.0-rc1

### Platforms tested on:

Any

### Bug description:

The email field is not fully visible on the Contact details dialog.  

![image](https://github.com/user-attachments/assets/f90ffbba-5e2b-4ed3-8804-77b4343c25d7)

PR #8792, opened 2024-10-25
[MS] Fix disabled email field display in UserInformation.vue

Ironicbay:
Closes https://github.com/Scille/parsec-cloud/issues/8726

Screenshot 2024-10-28 at 12 05 00

### DIFF
--- a/client/src/views/users/GreetUserModal.vue
+++ b/client/src/views/users/GreetUserModal.vue
@@ -82,6 +82,7 @@
           class="step user-info-page"
         >
           <user-information
+            class="user-details"
             :default-email="''"
             :default-name="''"
             :email-enabled="false"
@@ -519,5 +520,9 @@ onUnmounted(async () => {
 
 .spinner {
   padding-bottom: 0.4rem;
+}
+
+.user-details {
+  width: 100%;
 }
 </style>


client/src/components/users/UserInformation.vue
```diff
@@ -23,6 +24,19 @@
       @on-enter-keyup="$emit('onEnterKeyup', email)"
       :validator="emailValidator"
     />
+    <div
+      v-else
+      class="disabled-email-address"
+    >
+      <ion-text class="form-label">
```
Max-7, line 31: Missing `IonText` import
```
[Vue warn]: Failed to resolve component: ion-text
```

client/src/components/users/UserInformation.vue
```diff
@@ -23,6 +24,19 @@
...
+        {{ $msTranslate('CreateOrganization.email') }}
+      </ion-text>
+      <div class="disabled-email-address-value">
```
Max-7, line 34: Is there a way to make this take the whole width instead of just half (when greeting the user)?

Max-7 (changes requested):

Max-7 (approved):

Issue #8789, opened 2024-10-25
[🐛 | Bug]: glibc 2.28 unsupported


Landeers:
### Parsec version tested on:

3.1.0

### Platforms tested on:

Linux

### Bug description:

Platform details : 
On RHEL 8.10, glibc last version available is 2.28 (https://distrowatch.com/table.php?distribution=redhat).

Is it possible to make parsec-cli compatible with this version or to embed the glibc directly into the binary?

### Relevant output:

```shell
$ ldd parsec-cli_3.1.0_linux_x86_64
./parsec-cli_3.1.0_linux_x86_64: /lib64/libm.so.6: version `GLIBC_2.29' not found (required by ./parsec-cli_3.1.0_linux_x86_64)
./parsec-cli_3.1.0_linux_x86_64: /lib64/libc.so.6: version `GLIBC_2.32' not found (required by ./parsec-cli_3.1.0_linux_x86_64)
./parsec-cli_3.1.0_linux_x86_64: /lib64/libc.so.6: version `GLIBC_2.33' not found (required by ./parsec-cli_3.1.0_linux_x86_64)
./parsec-cli_3.1.0_linux_x86_64: /lib64/libc.so.6: version `GLIBC_2.34' not found (required by ./parsec-cli_3.1.0_linux_x86_64)
        linux-vdso.so.1 (0x00007ffe21281000)
        libfuse3.so.3 => /lib64/libfuse3.so.3 (0x00007fd7276cd000)
        libssl.so.3 => /lib64/libssl.so.3 (0x00007fd7273df000)
        libcrypto.so.3 => /lib64/libcrypto.so.3 (0x00007fd726cca000)
        libgcc_s.so.1 => /lib64/libgcc_s.so.1 (0x00007fd726ab2000)
        libm.so.6 => /lib64/libm.so.6 (0x00007fd726730000)
        libc.so.6 => /lib64/libc.so.6 (0x00007fd72635a000)
        /lib64/ld-linux-x86-64.so.2 (0x00007fd728e08000)
        libdl.so.2 => /lib64/libdl.so.2 (0x00007fd726156000)
        libpthread.so.0 => /lib64/libpthread.so.0 (0x00007fd725f36000)
        libz.so.1 => /lib64/libz.so.1 (0x00007fd725d1e000)
```

FirelightFlagboy:
`ld` is able to find a compatible `libc` here (see that `libc.so.6` is defined to a path vs `not found`)

[...]

PR #8804, opened 2024-10-28
Build CLI with musl

FirelightFlagboy:
Use `cross` to build CLI for linux target GNU & Musl.

The Musl target provide a statically linked binary that do not use
shared library.

During the packaging of the CLI, we also vendord openssl so we don't
depends on system libssl.

Closes https://github.com/Scille/parsec-cloud/issues/8789


.github/workflows/package-cli.yml
```diff
@@ -55,6 +55,17 @@ jobs:
     if: ${{ inputs.version_patch_run_id != '' && always() || success() }}
     name: 📦 Packaging CLI for linux 🐧
     runs-on: ubuntu-24.04
+    env:
+      CARGO: cargo
+      CROSS_VERSION: v0.2.5
```
FirelightFlagboy, line 60: Need to handle that with `version_updater` script

libparsec/Cargo.toml
FirelightFlagboy, line 1: `mountpoint` feature need to be set when building `libparsec` for electron

.github/workflows/package-cli.yml
```diff
@@ -55,6 +55,17 @@ jobs:
     if: ${{ inputs.version_patch_run_id != '' && always() || success() }}
     name: 📦 Packaging CLI for linux 🐧
     runs-on: ubuntu-24.04
+    env:
+      CARGO: cargo
```
touilleMan, line 59: Why not just use `cargo` directly ?

.github/workflows/package-cli.yml
```diff
@@ -83,30 +94,43 @@ jobs:
...
       - name: Build CLI
-        run: cargo build -p parsec-cli --release
+        run: >-
+          ${{ env.CARGO }} build -p parsec-cli
+          --release
+          --target=${{ matrix.target }}
+          --features=vendored-openssl
         timeout-minutes: 10
 
+      - name: Debug target folder
```
touilleMan, line 119: I guess the target folder will contain a lot of stuff (e.g. all the .o files)

so is this really useful to keep, or should we consider this a debugging artifact that can be removed now that the build runs fine ?

libparsec/src/workspace.rs
```diff
@@ -404,7 +404,7 @@ pub async fn workspace_mount(
     )))
 }
 
-#[cfg(not(target_arch = "wasm32"))]
+#[cfg(all(not(target_arch = "wasm32"), feature = "mountpoint"))]
```
touilleMan, line 407: You must also change the `#[cfg]` on the other implementation of this function (see above), right now there is no implementation of this function if not in wasm32 and `mountpoint` feature is disabled

.github/workflows/package-cli.yml
touilleMan: Given `env.CARGO` is only used in the `Build CLI` step, I'd rather try to keep as much local as possible

I guess this environment variable has to do with the `echo "CARGO=cross" >> $GITHUB_ENV` in the `Use cross` step.

This issue here are:
- `CARGO` environment variable may seems like a standard environment variable that is used internally in some rust tools
- It makes the code more complex, since we have to guess what `${{ env.CARGO }}` stands for (and often got the info wrong since the variable is defined at the top of the file, so it's easy to miss the overwrite that occurs in `Use Cross`)

~So in the end I would just remove this `CARGO` variable and instead call `cross` directly~

PLOT TWIST ! `CARGO` *is* indeed [a special environ variable](https://doc.rust-lang.org/cargo/reference/environment-variables.html) (I've just discover that now ^^)

So I think we should declare this environment variable only in the step that need it (so in `Build CLI`, and we consider `Use Cross` only job is to make available in PATH the `cross` command [^1]), and with a comment explaining (something like `Tell Cargo to forward this value instead of its own auto-detected path when it builds crates and executes build scripts and external subcommands`)

[^1]: so `Use Cross` can be renamed `Install Cross`

libparsec/Cargo.toml
```diff
@@ -22,6 +22,7 @@ test-utils = [
     "libparsec_platform_ipc/test-with-testbed",
     "libparsec_types/test-mock-time",
 ]
+mountpoint = ["dep:libparsec_platform_mountpoint"]
```
touilleMan, line 25: What is the need here ?
Given this feature is used in all builds, and the build is broken when not enabling it (see my other comment) I think we should just get rid of it

touilleMan (changes requested):

libparsec/Cargo.toml
FirelightFlagboy: The PR description state:

> `libparsec` as a new feature `mountpoint` that provide mountpoint. That
simplify compilation of static cli since we do not have to include
`libfuse`.

touilleMan: There is no simplification here since we add code to support a new feature, feature which is always enabled.

On top of that we want to have fuse support in the CLI, and this PR was able to do that fine ;-)

So I'd rather not have this feature entirely to stay KISS

.github/workflows/package-cli.yml
FirelightFlagboy: outdated

libparsec/src/workspace.rs
FirelightFlagboy: reverted

libparsec/Cargo.toml
FirelightFlagboy: reverted

.github/workflows/package-cli.yml
FirelightFlagboy: Now the step is only run if the job enable the debug mode

.github/workflows/package-cli.yml
```diff
@@ -55,6 +55,18 @@ jobs:
...
     runs-on: ubuntu-24.04
+    env:
+      cross-version: v0.2.5
+    strategy:
+      fail-fast: false
+      matrix:
+        include:
+          - target: x86_64-unknown-linux-musl
+            suffix: linux-x86_64-musl
+            type: static
```
touilleMan: Remove this flag, or rename it to `vendored-openssl`

.github/workflows/package-cli.yml
```diff
@@ -55,6 +55,18 @@ jobs:
     if: ${{ inputs.version_patch_run_id != '' && always() || success() }}
     name: 📦 Packaging CLI for linux 🐧
     runs-on: ubuntu-24.04
+    env:
+      cross-version: v0.2.5
+    strategy:
+      fail-fast: false
+      matrix:
```
touilleMan, line 62: Add a comment to explain we have two build strategy:
```
Parsec CLI has three potential dynamic dependancies: `libc`, `libfuse3` and `libopenssl`, hence we have two build strategy here:
- fully static binary with musl
- fully dynamic binary with gnu
```

.github/workflows/package-cli.yml
```diff
@@ -83,30 +95,43 @@ jobs:
...
+          curl -LO "https://github.com/cross-rs/cross/releases/download/${{ env.cross-version }}/cross-x86_64-unknown-linux-musl.tar.gz"
+          tar xf cross-x86_64-unknown-linux-musl.tar.gz
         timeout-minutes: 5
 
       - name: Build CLI
-        run: cargo build -p parsec-cli --release
-        timeout-minutes: 10
+        run: >-
+          cross build -p parsec-cli
+          --release
```
touilleMan: pass flags with `$(make.py cli-release-libparsec-cargo-flags)`

The `--target` should be kept here, and vendored-openssl passed through the `MAYBE_FORCE_VENDORED_OPENSSL` environ variable

touilleMan (approved):

Cross.toml
```diff
@@ -0,0 +1,26 @@
+[target.x86_64-unknown-linux-gnu]
+# FIXME: Remove me once cross-rs provide an updated image for x86_64-unknown-linux-gnu that do not use ubuntu-xenial.
+# Use a more recent version of the container since the one provided by 0.2.5 is ubuntu-xenial (16.04) which is too old for our needs
+# https://github.com/cross-rs/cross/issues/517#issuecomment-1500585369
+image = "ghcr.io/cross-rs/x86_64-unknown-linux-gnu@sha256:cf091033b6c4b0416a03d181080ad45c418248a262867a623306748a2c02a3c2" # pin ghcr.io/cross-rs/x86_64-unknown-linux-gnu:main
+pre-build = [
+    "dpkg --add-architecture $CROSS_DEB_ARCH",
+    # We do not add the prefix $CROSS_DEB_ARCH since neither libfuse3-dev nor libssl-dev is available for musl-linux-amd64
```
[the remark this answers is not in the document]
FirelightFlagboy: The comment was remove, it refer to a previous commit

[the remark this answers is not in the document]
FirelightFlagboy: The comment was remove, it refer to a previous version

.github/workflows/package-cli.yml
FirelightFlagboy: The flag was removed

PR #8840, opened 2024-11-04
[MS] Create the mountpoint folder before adding a link to it in the sidebar

### LINKED ISSUE #8820: Make sure that mountpoint folder exists before creating a link to it in the sidebar

Max-7:
Closes #8820 

- [X] Describe your changes


client/electron/src/setup.ts
```diff
@@ -213,6 +214,7 @@ export class ElectronCapacitorApp {
   }
 
   updateMountpoint(path: string): void {
+    fs.mkdirSync(path, { recursive: true });
```
FirelightFlagboy, line 217: Something is off here:

When searching for a suitable mounpoint for Windows, we need that the mountpoint path to not exist: https://github.com/Scille/parsec-cloud/blob/4d8960905d968a361163dc763a0cc71bd1be152a/libparsec/crates/platform_mountpoint/src/windows/mount.rs#L162-L175

Here we require that it exist, I've the felling we're going to have concurrency issues

Max-7: But in this case you're mounting the workspace. I'm just creating the folder where the mountpoints are located (`~/Parsec` on Linux). That folder is always the same, it's not deleted when the app closes, and I just make sure it's created right at the start of the app, before we try to mount anything.

client/electron/src/setup.ts
```diff
@@ -213,6 +214,7 @@ export class ElectronCapacitorApp {
   }
 
   updateMountpoint(path: string): void {
```
FirelightFlagboy, line 216:
```suggestion
  createAndRegisterBaseMountpointFolder(path: string): void {
```

I suggest to rename the method to something that reflect what the method do and give more hint about the path value

Max-7: This method just informs the ElectronCapacitorApp that a new mountpoint path has been set. What the class does with that information is up to the class.
If I had named it `addMountpointToQuickAccess` before, now I would have to rename it `createFolderAndAddMountpointToQuickAccess`, and later on when we'd want to add a shortcut on the desktop for the mountpoint I would have to rename it `createFolderAndAddMountpointToQuickAccessAndAddDesktopShortcut`.

fabienSvtr (approved):

Issue #8694, opened 2024-10-11
[🚀 | Feature request]: Display user freeze state & filter  


mmmarcos:
When a user is frozen, this information is not displayed in the Users page.

The goal of this issue is **not** to change frozen status in the GUI. Currently, this is only available at server-level via API, we are not sure if we should open this to organization admins. 

Instead, the idea is to only display the frozen status in the Users page, probably in the status column:
![image](https://github.com/user-attachments/assets/b9708f21-8cb7-4262-8b7c-dc29a5b5b482).

Note that the status is not displayed in the Grid view (is it wanted?)   
![image](https://github.com/user-attachments/assets/e4ee7173-a0d6-4d1d-a769-6d896d9f0691)

## To keep in mind

- Frozen status should be added as filter
- Priority of the displayed status: Revoked > Frozen > Active
- Explain what each status means (tooltip)

## Task List

- [x] #8695
- [x] #8696
- [x] #8697

### Setup

TODO: Describe how to local setup for testing this feature (testbed ? local server ? ...)

### Tests

TODO: List some test cases to be performed when reviewing the feature.

AureliaDolo:
On the backend side:

- [ ] expose get_user_info protocol in authenticated cmds (there is already a method retrieving that info from the db/memory, and server's userinfo alreadu has the frozen field)
- [ ] (add a protocol cmd to get_user_info)
- [ ] add frozen field to libparsec's UserInfo
- [ ] in client_list_user 
	1. retrieve userinfo from certifops
	2. update it with current frozen status through a authenticated cmds
- [ ] update binding's generator's ClientInfo with frozen status
- [ ] generate bindings
- [ ] add tests	


Open questions: 
- Should the frozen info be somewhere in the certifops ?
	- No, there is no associated certificate, but we won't know which users are frozen without a specific request
	- Yes, it is data about a user. But when should it be updated ?
- Is there already a protocol command that could be adapted to this usage ?

AureliaDolo:
Possible implementations:
1. update current client_list_user
	a. add a get_freeze_info(vec userId) -> vec UserInfo authenticated cmds
	b. add a get_frozen_users() -> Vec UserId
	- use a or b it in client list_user after retrieving user info from local certificates to update the user info certifops provided 
2. Add a get_remote_users_info(vec UserId) -> vec (userID, bool) binding
	- the client side implementation would only propagate the query to the server and do nothing else
	- in the GUI it would be used after list users to complete data with server's data. 
[...]

PR #8850, opened 2024-11-05
Expose user freeze state

### LINKED ISSUE #8685: Add TOS acceptance when creating an organization
- Saas orgs: https://www.figma.com/design/XW3SNWHV8BhPsPJpsOCV2h/Parsec-App?node-id=24760-130079&node-type=frame&t=d5Pm99OFxHrbH2Cf-0
- Trial orgs: https://www.figma.com/design/XW3SNWHV8BhPsPJpsOCV2h/Parsec-App?node-id=30696-33982&node-type=frame&t=BlZwmEnelh6J5nmf-11

### LINKED ISSUE #8695: Expose user freeze status in bindings

### LINKED ISSUE #8697: Server: Add freeze status when requesting user list
See details on #8694

AureliaDolo:
Fix #8695, fix #8697
Related #8694

TODO
- [x] check if author is admin
- [x] add client test

Open questions
- in postgre do we really need to first check that the organization exists ?


libparsec/crates/client/tests/unit/client/list_frozen_users.rs
```diff
@@ -0,0 +1,24 @@
+// Parsec Cloud (https://parsec.cloud) Copyright (c) BUSL-1.1 2016-present Scille SAS
+
+use libparsec_tests_fixtures::prelude::*;
+
+use super::utils::client_factory;
+use crate::ClientListFrozenUsersError;
+
+#[parsec_test(testbed = "coolorg", with_server)]
+async fn ok(env: &TestbedEnv) {
```
FirelightFlagboy, line 9: I would slip the test into 2 part, one is the happy path with alice and the other is the unhappy path with the `AutorNotAllowed` error using mallory

libparsec/crates/protocol/schema/authenticated_cmds/list_frozen_users.json5
```diff
@@ -0,0 +1,27 @@
...
+                        "name": "frozen_users",
+                        "type": "List<UserID>"
+                    }
+                ]
+            },
+            {
+                "status": "author_not_allowed"
+            },
+            {
+                "status": "author_not_found"
```
touilleMan, line 23: This status is not required: if the author doesn't exist then the authentication will fail and the server will return a HTTP 403

server/parsec/components/user.py
```diff
@@ -698,3 +709,20 @@ async def api_tos_accept(
...
+    ) -> authenticated_cmds.latest.list_frozen_users.Rep:
+        outcome = await self.list_frozen_users(client_ctx.organization_id, client_ctx.user_id)
+        match outcome:
+            case UserListFrozenUsersBadOutcome.ORGANIZATION_NOT_FOUND:
+                client_ctx.organization_not_found_abort()
+            case UserListFrozenUsersBadOutcome.AUTHOR_NOT_ALLOWED:
+                return authenticated_cmds.latest.list_frozen_users.RepAuthorNotAllowed()
+            case UserListFrozenUsersBadOutcome.AUTHOR_NOT_FOUND:
+                return authenticated_cmds.latest.list_frozen_users.RepAuthorNotFound()
+            case frozen_users:
```
touilleMan, line 727: Move the happy case first (this is to stay consistent with the rest of the codebase, and also given internally Python turns the match in a chain of `if .. elif .. elif ...`, so we avoid comparing to all the error cases)

On top of that, avoid using untyped variable the case: given they  accept any type it may introduce bug if we add new error types.

```suggestion
            case list() as frozen_users:
```

server/parsec/components/user.py
```diff
@@ -698,3 +709,20 @@ async def api_tos_accept(
...
                 return tos_cmds.latest.tos_accept.RepTosMismatch()
+
+    @api
+    async def api_list_frozen_users(
+        self,
+        client_ctx: AuthenticatedClientContext,
+        req: authenticated_cmds.latest.list_frozen_users.Req,
+    ) -> authenticated_cmds.latest.list_frozen_users.Rep:
+        outcome = await self.list_frozen_users(client_ctx.organization_id, client_ctx.user_id)
+        match outcome:
```
touilleMan, line 720: You should also add a `case UserListFrozenUsersBadOutcome.AUTHOR_REVOKED:` 

Note the test `server/tests/api_v4/authenticated/test_list_frozen_users.py::test_authenticated_list_frozen_users_http_common_errors[author_revoked]` has passed even without this: this is because the authentication system is triggered first and handle the case.

So this is mostly to stay consistent with the rest of the APIs (we have plan in the long run to remove the organization/user exists/not revoked/not expired/not frozen check from all the APIs in order have this totally factorized, but now is not the time ^^)

server/parsec/components/postgresql/user_list_frozen_users.py
```diff
@@ -0,0 +1,67 @@
...
+)
+
+_q_get_organization_frozen_users = Q(
+    f"""
+SELECT
+    user_.user_id
+
+FROM user_
+WHERE
+    user_.organization = { q_organization_internal_id("$organization_id") }
```
touilleMan, line 30: Not `auth_no_lock` returns a `AuthNoLockData` that contains the internal id of the organization, so you don't need to use  `q_organization_internal_id` \o/

server/parsec/components/postgresql/user_list_frozen_users.py
```diff
@@ -0,0 +1,67 @@
...
+"""
+)
+
+
+async def user_list_frozen_users(
+    conn: AsyncpgConnection,
+    organization_id: OrganizationID,
+    user_id: UserID,
+) -> list[UserID] | UserListFrozenUsersBadOutcome:
+    u_row = await conn.fetchrow(
```
touilleMan, line 42: There is a helper function to do those common checks easily: `match await auth_no_lock(conn, organization_id, author):` (see what is done in `certificates_get` for instance)

touilleMan (changes requested):

server/parsec/components/postgresql/user_list_frozen_users.py
AureliaDolo: That's handy !

FirelightFlagboy (approved):

server/parsec/components/postgresql/user_list_frozen_users.py
```diff
@@ -0,0 +1,64 @@
...
+    rows = await conn.fetch(
+        *_q_get_organization_frozen_users(organization_id=internal_organization_id)
+    )
+
+    users = []
+    for row in rows:
+        match row["user_id"]:
+            case str() as raw_user_id:
+                user_id = UserID.from_hex(raw_user_id)
+            case _:
```
touilleMan, line 60: If we end up in the fall-through case, it means their is corupted data in PostgreSQL

So instead of ignoring it, we should report it by raising an exception containing the invalid value:
```python
        case unknown:
            assert False, repr(unknown)
```

server/parsec/components/postgresql/user_list_frozen_users.py
```diff
@@ -0,0 +1,64 @@
...
+)
+
+_q_get_organization_frozen_users = Q(
+    """
+SELECT
+    user_.user_id
+
+FROM user_
+WHERE
+    user_.organization = $organization_id
```
touilleMan, line 23:
```suggestion
    user_.organization = $organization_internal_id
```

touilleMan (approved):

PR #8855, opened 2024-11-05
Add `purpose` field to device certificate

### LINKED ISSUE #8748: Update device model to include a `type` field
We have multiple device type, and are working to introduce some more:

- Standard device
- Recovery device
- Shamir device
- Parsec Auth device

We would like to be able to differentiate the types of device.

We could add a field (like type) to identify the type of device.

Currently we rely on the untagged deserialization provided by serde to load each device type but we would like to provide that information upfront to the lib user.

> Adding a type field, would also make serde more efficient at deserializing the devices since it can directly identify the type of device.

touilleMan:
close #8748

Used to flag recovery devices from regular ones


libparsec/crates/types/schema/certif/device_certificate.json5
```diff
@@ -41,6 +47,23 @@
...
+                    "name": "Standard",
+                    "discriminant_value": "STANDARD"
+                },
+                {
+                    "name": "ShamirRecovery",
+                    "discriminant_value": "SHAMIR_RECOVERY"
+                },
+                {
+                    "name": "KeyFileRecovery",
+                    "discriminant_value": "KEY_FILE_RECOVERY"
```
FirelightFlagboy, line 64: I'm thinking if that variant should be called `PassphraseRecovery`, since we do not use a file as is but need a passphrase.

`Shamir` could also work that way (but we don't want that) where the secret that is reconstructed is the key to decrypt the file.

vxgmichel: I'm ok with both but passphrase sounds a bit better, so let's do that :+1:

libparsec/crates/types/schema/certif/device_certificate.json5
```diff
@@ -41,6 +47,27 @@
...
+                    "discriminant_value": "SHAMIR_RECOVERY"
+                },
+                {
+                    "name": "PassphraseRecovery",
+                    "discriminant_value": "PASSPHRASE_RECOVERY"
+                },
+                {
+                    "name": "WebAuth",
+                    "discriminant_value": "WEB_AUTH"
+                }
```
FirelightFlagboy, line 69: :face_with_peeking_eye:

FirelightFlagboy (approved):

AureliaDolo (approved):

PR #8873, opened 2024-11-07
Add pre-commit hook for powrap

mmmarcos:
Changes in this PR:
 
- Add pre-commit hook to automatically apply `powrap` on PO files
- Add Makefile target to locally wrap files via `powrap`
- Apply wrap to all existing `.po` files.


server/pyproject.toml
```diff
@@ -63,6 +63,7 @@ httpx-sse = ">=0.3.1,<0.5.0"
 maturin = "1.7.4"
 patchelf = { version = "^0.17.2.0", markers = "platform_system=='Linux'" }
 poetry-lock-package = ">=0.4.4,<0.6.0"
+poutils = "^0.15.4"
```
FirelightFlagboy, line 66: I would suggest to put that dependency in a dedicated group

mmmarcos: How would you name this group? `precommit`? `ci`? Would you also move other dependencies such as `deptry` or `sqlfluff` that are used in pre-commit?

After some thought, I am not sure if it is a good idea. If we create another group (say `precommit` or `ci`):
- If you are a developer: you would need to install both `dev` group AND this new group to be able to commit
- If you are the ci: you would also need to install both groups.

So, what would be the benefit of having a different group?

FirelightFlagboy: > How would you name this group?

> If you are a developer: you would need to install both dev group AND this new group to be able to commit

That's OK for me since that would mean to dev is trying to work on the documentation, which we could add a dedicated step in the documentation for the setup.

> If you are the ci: you would also need to install both groups.

Set once and forget

FirelightFlagboy (commented):
Can you call `make wrap` in `ci-docs` ? and check for modified files

FirelightFlagboy (commented):
LGTM, there is something missing as this PR is still a draft ?

mmmarcos:
> LGTM, there is something missing as this PR is still a draft ?

It seems that `msgcat` (which is used by `powrap` and is part of `gettext` package) is not installed in the GA runner, but I don't know how to check that: https://github.com/Scille/parsec-cloud/actions/runs/11740418118/job/32706939530?pr=8873#step:6:28

FirelightFlagboy:
I had the same problem, that's fixed by installing the missing package, cf the following exemple commit:

[98465aa8b](https://github.com/Scille/parsec-bms/pull/1048/commits/98465aa8b67565cab7e3a4094260eda1e3448b79)

To see which software are installed on a runner, go to the first set `Set up job` open-up the `Runner Image` details, that will reveal a line starting with `Included Software` with a like to a README listing all install software+version on the runner

FirelightFlagboy (approved):

PR #8879, opened 2024-11-08
Fix Python crypto bindings

### LINKED ISSUE #8864: Remove useless `VerifyKey.__bytes__` in Python bindings
`VerifyKey.encode` does the same thing and is more consistent with the other keys classes

see: 
https://github.com/Scille/parsec-cloud/blob/557fec107dfafa968f8d84158a30fd98c0594877/server/parsec/_parsec_pyi/crypto.pyi#L45

### LINKED ISSUE #8865: `PrivateKey.encode` returns a `list[int]` instead of `bytes` in Python bindings
@vxgmichel PyO3 has bitten us again ! :trollface: 

```python
>>> from parsec._parsec import PrivateKey
>>> PrivateKey.generate().encode()
[220, 39, 111, 82, 33, 93, 107, 31, 31, 217, 70, 228, 115, 161, 50, 79, 0, 110, 251, 62, 34, 98, 184, 102, 70, 122, 162, 4, 28, 247, 33, 125]
```

https://github.com/Scille/parsec-cloud/blob/557fec107dfafa968f8d84158a30fd98c0594877/server/parsec/_parsec_pyi/crypto.pyi#L59

It's possible some other key types have the same issue, so we should check also check them...

touilleMan:
close #8865 
close #8864


vxgmichel (approved):
LGTM :+1:

FirelightFlagboy (approved):

AureliaDolo (approved):

PR #8900, opened 2024-11-12
[MS] Batch workspace sharing

### LINKED ISSUE #8326: [🚀 | Feature request]: Batch workspace sharing

Allow an user to share a workspace to multiple users in 1 go

On the workspace sharing model, display a "selectable" list (like the one for files) where the user can select one or more users to assign them to a specific role

Figma link: https://www.figma.com/design/XW3SNWHV8BhPsPJpsOCV2h/Parsec-App?node-id=25763-142489&node-type=section&t=gb5llDfZsvTDGJTC-11

### Setup

TODO: If applicable, describe how to local setup for testing this feature (testbed ? local server ? mocked data...)

### Tests

TODO: List some test cases to be performed when reviewing the feature.

Ironicbay:
Closes #8326 

https://github.com/user-attachments/assets/5b26a93c-0a9b-44ab-8212-a5167fec3b8c

### DIFF
--- a/client/src/locales/en-US.json
+++ b/client/src/locales/en-US.json
@@ -320,7 +320,6 @@
         }
     },
     "WorkspaceSharing": {
-        "title": "Share this workspace",
         "searchPlaceholder": "e.g.: John Smith",
         "currentUserLabel": "Me",
         "unshareSuccess": "The workspace is no longer shared with {user}.",
@@ -331,8 +330,22 @@
             "alreadyNotShared": "This workspace is not shared with {user}.",
             "alreadyHasRole": "{user} is already {role} on this workspace."
         },
+        "batchSharing": {
+            "outsiderRoleWarning": "{external} profiles can only have {contributor} or {reader} roles.",
+            "external": "External",
+            "contributor": "Contributor",
+            "reader": "Reader",
+            "buttonSelect": "Multiple selection",
+            "buttonFinish": "Finish",
+            "chooseRole": "Choose a role",
+            "noChanges": "Selected members were already {role}.",
+            "someFailed": "Some role assignments failed.",
+            "allFailed": "And error has occurred, role assignment has failed. (Reason: {reason})",
+            "updateRoleSuccess": "Selected members' roles have been updated to {role}.",
+            "unshareSuccess": "The workspace is no longer shared with selected members."
+        },
         "onlyOwnerWarning": "To prevent data loss if you ever leave the organization, we recommend promoting someone else to {owner} on this workspace.",
-        "owner": "owner",
+        "owner": "Owner",
         "noMatch": "No user found that matches '{query}'."
     },
     "SettingsModal": {
--- a/client/src/views/workspaces/WorkspaceSharingModal.vue
+++ b/client/src/views/workspaces/WorkspaceSharingModal.vue
@@ -3,19 +3,9 @@
 <template>
   <ion-page class="modal">
     <ms-modal
-      :title="'WorkspaceSharing.title'"
+      :title="I18n.valueAsTranslatable(workspaceName)"
       :close-button="{ visible: true }"
     >
-      <div class="modal-head-content">
-        <ion-text class="modal-title title-h4">
-          {{ workspaceName }}
-        </ion-text>
-        <ms-search-input
-          class="modal-head-content__search"
-          v-model="search"
-          :placeholder="'WorkspaceSharing.searchPlaceholder'"
-        />
-      </div>
       <ms-report-text
         :theme="MsReportTheme.Info"
         v-if="isOnlyOwner()"
@@ -30,6 +20,55 @@
           </template>
         </i18n-t>
       </ms-report-text>
+      <div class="modal-head-content">
+        <ms-search-input
+          class="modal-head-content__search"
+          v-model="search"
+          :placeholder="'WorkspaceSharing.searchPlaceholder'"
+          @change="updateSelectedUsers()"
+        />
+        <div class="modal-head-content-right">
+          <ion-text
+            id="batch-activate-button"
+            @click="onBatchSharingActivate()"
+            fill="clear"
+            class="button-small"
+          >
+            {{ $msTranslate(showCheckboxes ? 'WorkspaceSharing.batchSharing.buttonFinish' : 'WorkspaceSharing.batchSharing.buttonSelect') }}
+          </ion-text>
+          <ms-dropdown
+            v-show="showCheckboxes"
+            ref="dropdownRef"
+            class="dropdown"
+            :options="options"
+            :disabled="!someUserSelected"
+            :label="'WorkspaceSharing.batchSharing.chooseRole'"
+            :appearance="MsAppearance.Outline"
+            @change="onBatchRoleChange($event.option)"
+          />
+        </div>
+      </div>
+      <ms-report-text
+        :theme="MsReportTheme.Info"
+        id="profile-assign-info"
+        v-if="showCheckboxes && orgHasExternalUsers"
+      >
+        <i18n-t
+          keypath="WorkspaceSharing.batchSharing.outsiderRoleWarning"
+          scope="global"
+        >
+          <template #external>
+            <strong> {{ $msTranslate('WorkspaceSharing.batchSharing.external') }} </strong>
+          </template>
+          <template #contributor>
+            <strong> {{ $msTranslate('WorkspaceSharing.batchSharing.contributor') }} </strong>
+          </template>
+          <template #reader>
+            <strong> {{ $msTranslate('WorkspaceSharing.batchSharing.reader') }} </strong>
+          </template>
+        </i18n-t>
+      </ms-report-text>
+
       <!-- content -->
       <div class="modal-container">
         <ion-list class="user-list">
@@ -38,8 +77,7 @@
             class="no-match-result body"
             v-show="
               userRoles.length > 0 &&
-                filteredSharedUserRoles.length === 0 &&
-                filteredNotSharedUserRoles.length === 0 &&
+                filteredUserRoles.length === 0 &&
                 !currentUserMatchSearch()
             "
           >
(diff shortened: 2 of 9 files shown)


client/src/views/workspaces/WorkspaceSharingModal.vue
```diff
@@ -132,37 +191,54 @@ const props = defineProps<{
   informationManager: InformationManager;
 }>();
 
+interface UserRole {
+  role: [UserTuple, WorkspaceRole | null];
```
Max-7, line 195: Might as well extract the role. It was done as `[UserTuple, Workspace | null]` before to avoid declaring an interface, but since you're now declaring an interface, you can write
```
interface XX {
  user: UserTuple;
  role: WorkspaceRole | null;
  isSelected: boolean;
}
```

client/src/views/workspaces/WorkspaceSharingModal.vue
```diff
@@ -191,11 +267,73 @@ function isSelectDisabled(role: WorkspaceRole | null): boolean {
...
+  return lowestProfile;
+}
+
+function getHighestSelectedRole(): WorkspaceRole {
+  let highestRole = WorkspaceRole.Reader;
+  for (const user of selectedUsers.value) {
+    if (!user.role[1]) {
+      continue;
+    }
+    if (user.role[1] in [WorkspaceRole.Manager, WorkspaceRole.Owner]) return user.role[1];
```
Max-7, line 299: I don't think you can use `in` here, it doesn't work as in Python.
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Expressions_and_operators#in
It's used to check if an object has a property and in this case, if an array as an index (not a value).

```
> a = [2,3,4,5];
> 4 in a;
false
> 2 in a;
true
```

client/src/views/workspaces/WorkspaceSharingModal.vue
```diff
@@ -310,6 +448,16 @@ async function updateUserRole(
   }
   await refreshSharingInfo();
 }
+
+async function onBatchRoleChange(newRoleOption: MsOption): Promise<void> {
+  for (const userRole of selectedUsers.value) {
+    if (!userRole.role[1] || newRoleOption.key !== userRole.role[1]) {
+      await updateUserRole(userRole.role[0], userRole.role[1], newRoleOption.key, () => {});
```
Max-7, line 455: Don't use `updateUserRole`, it does too much. It handles errors individually, which means that if you try to change the role of 5 users at the same time and the 5 fail, you'll get five toasts on top of one another.
It also calls `refreshSharingInfo` which updates `userRoles` which will probably update `selectedUsers` while you're still iterating on the list.

client/src/views/workspaces/WorkspaceSharingModal.vue
```diff
@@ -191,11 +267,73 @@ function isSelectDisabled(role: WorkspaceRole | null): boolean {
...
+    user.isSelected = !membersCheckbox.value;
+  }
+}
+
 async function refreshSharingInfo(): Promise<void> {
   const result = await getWorkspaceSharing(props.workspaceId, true);
 
   if (result.ok) {
-    userRoles.value = result.value;
+    userRoles.value = [];
```
Max-7, line 333: You can use `.map` to convert an array of object to an array of other objects (or to add a property)

client/src/locales/en-US.json
```diff
@@ -330,8 +330,17 @@
             "alreadyNotShared": "This workspace is not shared with {user}.",
             "alreadyHasRole": "{user} is already {role} on this workspace."
         },
+        "batchSharing": {
+            "outsiderRoleWarning": "{external} profiles only be set as {contributor} or {reader}.",
```
mmmarcos, line 334:
```suggestion
            "outsiderRoleWarning": "{external} profiles can only have {contributor} or {reader} role.",
```

client/src/views/workspaces/WorkspaceSharingModal.vue
```diff
@@ -3,18 +3,51 @@
 <template>
   <ion-page class="modal">
     <ms-modal
-      :title="'WorkspaceSharing.title'"
+      :title="workspaceName"
```
Max-7, line 6: This will create an error since title is expecting a `Translatable`. It will pass the workspaceName to `$msTranslate` but there won't be any key for it. You should use `I18n.valueAsTranslatable` in this case (https://github.com/Scille/megashark-lib/blob/cff992d182818775dc4084185c2854e8435060d5/lib/services/translation.ts#L173)

client/src/views/workspaces/WorkspaceSharingModal.vue
```diff
@@ -3,18 +3,51 @@
...
+          v-show="showCheckboxes"
+          ref="dropdownRef"
+          class="dropdown"
+          :options="options"
+          :disabled="!someUserSelected()"
+          :label="'WorkspaceSharing.batchSharing.chooseRole'"
+          :appearance="MsAppearance.Outline"
+          @change="onBatchRoleChange($event.option)"
+        />
+        <ion-button
```
Max-7, line 45: Mission import for `IonButton`

client/src/views/workspaces/WorkspaceSharingModal.vue
```diff
@@ -65,35 +105,54 @@
...
               v-for="entry in filteredSharedUserRoles"
-              :disabled="isSelectDisabled(entry[1])"
-              :key="entry[0].id"
-              :user="entry[0]"
-              :role="entry[1]"
-              :client-profile="ownProfile"
-              :client-role="ownRole"
-              @role-update="updateUserRole"
-            />
+              :key="`${entry.user.id}-${entry.role}`"
```
Max-7, line 110: Is there a reason why you updated the key to add the role? The userId should already be unique.

client/src/views/workspaces/WorkspaceSharingModal.vue
```diff
@@ -65,35 +105,54 @@
...
               v-for="entry in filteredNotSharedUserRoles"
-              :disabled="isSelectDisabled(entry[1])"
-              :key="entry[0].id"
-              :user="entry[0]"
-              :role="entry[1]"
-              :client-profile="ownProfile"
-              :client-role="ownRole"
-              @role-update="updateUserRole"
-            />
+              :key="`${entry.user.id}-${entry.role}`"
```
Max-7, line 139: Same here, why use the role?

client/src/views/workspaces/WorkspaceSharingModal.vue
```diff
@@ -132,37 +192,63 @@ const props = defineProps<{
...
+const showCheckboxes = ref<boolean>(false);
+const membersCheckbox = ref<boolean>(false);
+const dropdownRef = ref();
+
+const userRoles: Ref<Array<UserRole>> = ref([]);
 const filteredSharedUserRoles = computed(() => {
   const searchString = search.value.toLocaleLowerCase();
   return userRoles.value
-    .filter(([user, role]) => {
+    .filter(({ user: user, role: role }) => {
```
Max-7, line 211: `.filter((userRole: UserRole) => { userRole.user.humanHandle.xxx })`

client/src/views/workspaces/WorkspaceSharingModal.vue
```diff
@@ -132,37 +192,63 @@ const props = defineProps<{
...
     })
-    .sort((item1, item2) => item1[0].humanHandle.label.localeCompare(item2[0].humanHandle.label));
+    .sort((item1, item2) => item1.user.humanHandle.label.localeCompare(item2.user.humanHandle.label));
 });
 
 const filteredNotSharedUserRoles = computed(() => {
   const searchString = search.value.toLocaleLowerCase();
   return userRoles.value
-    .filter(([user, role]) => {
+    .filter(({ user: user, role: role }) => {
```
Max-7, line 224: Same here, `.filter((userRole: UserRole) => {})`

client/src/views/workspaces/WorkspaceSharingModal.vue
```diff
@@ -132,37 +192,63 @@ const props = defineProps<{
...
+    .sort((item1, item2) => item1.user.humanHandle.label.localeCompare(item2.user.humanHandle.label));
+});
+
+const selectedUsers = computed(() => {
+  return userRoles.value.filter((user) => user.isSelected === true);
 });
 
 const countSharedUsers = computed(() => filteredSharedUserRoles.value.length + (clientInfo.value ? 1 : 0));
 
+function someUserSelected(): boolean {
```
Max-7, line 240: Use `computed()` for these three functions.

client/src/views/workspaces/WorkspaceSharingModal.vue
```diff
@@ -310,6 +461,72 @@ async function updateUserRole(
...
+      const result = await shareWorkspace(props.workspaceId, userRole.user.id, newRoleOption.key);
+      if (result.ok) {
+        changesMade = true;
+      } else {
+        errorOccured = true;
+      }
+    }
+  }
+
+  if (!changesMade) {
```
Max-7, line 480: Error handling is insufficient here. 
You can test it in web mode by updating the mocks for `shareWorkspace` and add (as an example)
`return { ok: false, error: { tag: ClientShareWorkspaceErrorTag.Offline, error: 'Offline' } };`

The error I get is `Selected members were already Contributor.`, which is false. I'm not even sure that this error is possible, if you try to assign the same role to a user, I think it'll just work.

Handling multiple errors is always a pain. The best thing to do imo is to count the number of successes, the number of errors, and store the first error (if any). You can then compare it to the number of operations: 

- `successesCount === selectedUsers.value.length` means all good
- `errorsCount === selectedUsers.value.length` means all failed, probably a very general case like being offline, being revoked, having insufficient permission (you can check using the error you stored and assume it's the same for all)
- some errors is a complicated case, but a generic error like "some failed" is probably enough

Max-7 (changes requested):

client/src/views/workspaces/WorkspaceSharingModal.vue
```diff
@@ -310,6 +454,71 @@ async function updateUserRole(
...
+          key: 'WorkspaceSharing.batchSharing.noChanges',
+          data: {
+            role: I18n.translate(getWorkspaceRoleTranslationKey(newRoleOption.key).label),
+          },
+        },
+        level: InformationLevel.Info,
+      }),
+      PresentationMode.Toast,
+    );
+  } else if (errorOccured) {
```
Max-7, line 486: It's still not enough. If error is set, it could mean that all failed, or that only some failed. It's important to make the distinction, and in case they all failed, to have a bit more information (I mentioned storing the first error).

Replacing `changesMade` and `errorOccured` by counters would make this way more easy, and you'd also be able to compare to `selectedUsers.value.count` to know how many operation were planned, how many failed, how many succeeded.

Max-7 (changes requested):

client/src/views/workspaces/WorkspaceSharingModal.vue
```diff
@@ -310,6 +478,68 @@ async function updateUserRole(
...
+      if (!result.ok) {
+        if (errorCount === 0) {
+          error = result.error;
+        }
+        errorCount += 1;
+      }
+    }
+  }
+
+  if (errorCount > 0 && errorCount < selectedUsers.value.length) {
```
Max-7, line 498: Technically no, since you only call `shareWorkspace` on some of the selected users (condition on line 487), you could have a case where all operations failed with a number of errors inferior to the number of selected users. You do need to know how many attempts were made.

client/src/views/workspaces/WorkspaceSharingModal.vue
```diff
@@ -310,6 +478,68 @@ async function updateUserRole(
   }
   await refreshSharingInfo();
 }
+
+async function onBatchRoleChange(newRoleOption: MsOption): Promise<void> {
+  let errorCount: number = 0;
+  let error = undefined;
```
Max-7, line 484: Type it, it's a `ClientShareWorkspaceError | undefined`

Max-7 (approved):

PR #8819, opened 2024-10-30
Add support for shamir recovery invitation

vxgmichel:
Part of https://github.com/Scille/parsec-cloud/issues/6090

Close #7359

Fix #8429

Tested with:
```shell
$ . ./cli/run_testenv.sh 
[...]
Alice & Bob devices (password: P@ssw0rd.):
- 624 // Alice
- 196 // Alice 2nd device
- 1f2 // Bob

$ echo P@ssw0rd. | target/debug/parsec-cli shared-recovery create --password-stdin -d 1f2
Shamir setup has been created

$ echo P@ssw0rd. | target/debug/parsec-cli invite shared-recovery --password-stdin -d 624 --email bob@example.com
Invitation URL: parsec3://127.0.0.1:41997/Org?no_ssl=true&a=claim_shamir_recovery&p=xBCqEy9RIO8HPyycmlus9Qzg

$ echo P@ssw0rd. | target/debug/parsec-cli invite list --password-stdin -d 624
aa132f5120ef073f2c9c9a5bacf50ce0        idle    shamir recovery (Bob <bob@example.com>)

$ echo P@ssw0rd. | target/debug/parsec-cli invite cancel --password-stdin -d 624 --token aa132f5120ef073f2c9c9a5bacf50ce0
Invitation deleted

$ echo P@ssw0rd. | target/debug/parsec-cli invite list --password-stdin -d 624 
aa132f5120ef073f2c9c9a5bacf50ce0        cancelled       shamir recovery (Bob <bob@example.com>)
```

PR #8925, opened 2024-11-14
Add TestbedEventFreezeUser/TestbedEventUpdateOrganization/RemoveShamirRecovery  events and shamir template to the testbed

touilleMan:


libparsec/crates/testbed/src/template/events.rs
```diff
@@ -1875,14 +1894,25 @@ impl TestbedEventArchiveRealm {
  * TestbedEventNewShamirRecovery
  */
 
+pub struct TestbedEventNewShamirRecoveryCache {
+    pub certificates: Vec<TestbedTemplateEventCertificate>,
+    pub ciphered_data: Bytes,
+}
+
 #[derive(Serialize, Deserialize, Clone)]
```
AureliaDolo, line 1902: I get that the testbed is a mysterious place, but why does a testbed event needs to be serialized/ deserialized?

AureliaDolo: Not sure I want an answer though

libparsec/crates/testbed/src/template/events.rs
```diff
@@ -257,11 +264,12 @@ where
...
     ArchiveRealm(L),
     NewShamirRecovery(M),
+    RemoveShamirRecovery(N),
     Other,
 }
 
-impl<A, B, C, D, E, F, G, H, I, J, K, L, M> Iterator
-    for TestbedEventCertificatesIterator<A, B, C, D, E, F, G, H, I, J, K, L, M>
+impl<A, B, C, D, E, F, G, H, I, J, K, L, M, N> Iterator
+    for TestbedEventCertificatesIterator<A, B, C, D, E, F, G, H, I, J, K, L, M, N>
```
AureliaDolo, line 272: Wow

libparsec/crates/testbed/src/template/events.rs
```diff
@@ -1909,96 +1945,323 @@ impl TestbedEventNewShamirRecovery {
...
+        // Share certificates
 
-                let brief_certif = TestbedTemplateEventCertificate {
-                    certificate: AnyArcCertificate::ShamirRecoveryBrief(Arc::new(certif)),
-                    signed_redacted: signed.clone(),
-                    signed,
-                };
-                certifs.push(brief_certif);
+        for (recipient, _) in &self.per_recipient_shares {
+            // TODO: Put a real share here once shamir recovery is implemented
```
AureliaDolo, line 2026: Depends on #8819 or #6090 ?

libparsec/crates/types/schema/certif/removed_shamir_recovery_certificate.json5
AureliaDolo, line 1: So we introduce a real shamir deletion certificate here ?

Should there be an associated protocol command too ?

touilleMan: yes that's the idea, but it is something for another PR ;-)

libparsec/crates/testbed/src/template/events.rs
touilleMan: Serialization is need for two things:
- to compute le CRC, this allows the rust code running the test and the testbed server to ensure they have the same template (otherwise a single change in the template definition would cause subtle de-synchronization that would make our test fail in all kind of weird ways)
- to transmit customization of a template from the rust code running the test to the testbed server (in practice the rust code serialize the additional events it has created in during the customization, then ask the testbed server to create a regular organization with a given template, then ask the testbed server to load the additional events)

libparsec/crates/testbed/src/template/events.rs
touilleMan: Ah I missed that ! good catch !

touilleMan: done.

This snowballed a bit though: I notice some issues with the shamir schemas (incorrect serialization format, inconsistent load doing a signature verification while the dump doesn't sign, no tests) so I also fixed that ;-)

server/tests/api_v4/authenticated/test_shamir_recovery_setup.py
```diff
@@ -85,46 +86,49 @@ async def test_authenticated_shamir_recovery_setup_share_inconsistent_timestamp(
     assert rep == authenticated_cmds.v4.shamir_recovery_setup.RepShareInconsistentTimestamp()
 
 
+@pytest.mark.xfail(
+    reason="TODO: currently there is a unique shamir topic, we should switch to a per-user shamir topic instead"
+)
```
vxgmichel, line 91: Is that still necessary?

vxgmichel (approved):
Minor comments, LGTM otherwise :+1:

libparsec/crates/types/src/shamir.rs
```diff
@@ -102,6 +110,13 @@ impl ShamirShare {
     pub fn dump(&self) -> Vec<u8> {
         Vec::from(&self.0)
     }
+    pub fn load(raw: &[u8]) -> CryptoResult<Self> {
+        let share = sharks::Share::try_from(raw).map_err(|_| CryptoError::KeySize {
+            expected: 2,
```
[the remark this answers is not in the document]
touilleMan: I based this on the error condition in sharks:
https://github.com/c0dearm/sharks/blob/e7e23ba899c2a80d622f57f96c61e814325f8e20/src/share.rs#L59

server/tests/api_v4/authenticated/test_shamir_recovery_setup.py
touilleMan: Yes because currently the memory implementation of shamir in server has a single shamir topic for the whole organization (not a shamir topic per-user)

AureliaDolo (approved):
It would be nice to check if the shamir rfc is still upto date.

Issue #8333, opened 2024-09-11
[🚀 | Feature request]: Display last updater of a file


Landeers:
The **last updater** of a file/folder should be displayed in the **list view**.

![image](https://github.com/user-attachments/assets/2624c28d-407b-45c5-b572-37d6416183b0)

NicoTuxx:
Check if still relevant

Max-7:
Requires update to parsec, the last updater of a file is not listed among the `EntryStat` data https://github.com/Scille/parsec-cloud/blob/c014eb83ab8cdb9f2ea4877f927fe563c3487117/client/src/plugins/libparsec/definitions.ts#L1175

mmmarcos:
All tasks completed.

PR #8974, opened 2024-11-21
[libparsec] Expose last updater.

AureliaDolo:
Related #8333


Max-7 (commented):
When exposing the API in the bindings, just be mindful of the fact that while a `DeviceID` is enough information to retrieve the associated `HumanHandle` using the function `clientGetUserDevice`, having to do it for each file could have a severe impact on performances. The GUI could add a cache if needed but there may be better alternatives on the libparsec side.

mmmarcos:
Superseded by #10633

PR #8993, opened 2024-11-25
Replace UserOrDevice with InvitationType

vxgmichel:
Single PR to make sure this change doesn't break compatibility


FirelightFlagboy (approved):
:+1:

PR #8996, opened 2024-11-26
[MS] UsersPage search input

### LINKED ISSUE #8964: Add searchbar on UserPage

Ironicbay:
Closes #8964

### DIFF
--- a/client/src/components/users/types.ts
+++ b/client/src/components/users/types.ts
@@ -46,6 +46,7 @@ export interface UserFilterChangeEvent {
 export class UserCollection {
   users: Array<UserModel>;
   filters: UserFilterLabels;
+  _searchFilter: string;
 
   constructor() {
     this.users = [];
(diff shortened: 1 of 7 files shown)


client/src/components/users/types.ts
```diff
@@ -89,9 +91,23 @@ export class UserCollection {
     });
   }
 
+  getSearchFilterUsers(): Array<UserModel> {
```
Max-7, line 94: You don't need to add all of this. `getUsers()` already returns the visible users only. So just add the searchFilter to `userIsVisible`:
```
if (!user.humanHandle.label.toLocaleLowerCase().includes(this.searchFilter) &&
    !user.humanHandle.email.toLocaleLowerCase().includes(this.searchFilter) {
  return false;
}
```
and everything should take care of itself.

Since we want `searchFilter` to also be lower case, you could rename the attribute `_searchFilter` and add a setter that makes it lowerCase:

```
public set searchFilter(value: string) {
  this._searchFilter = value.toLocaleLowerCase();
}
```

Max-7 (changes requested):

Max-7 (approved):

PR #9086, opened 2024-12-02
Factorize the initialization of a client in the CLI

FirelightFlagboy:


cli/src/macros.rs
```diff
@@ -1,3 +1,26 @@
+/// This macro builds a main function that takes the `Args` struct defined in the same module.
+///
+/// Once called, it will initialize a client using the provided arguments and then call the sub-task and cleanup the client ressource before returning.
```
AureliaDolo, line 3:
```suggestion
/// Once called, it will initialize a client using the provided arguments and then call the sub-task and cleanup the client ressource before returning.
///
/// the callback takes (args: Args, client: &StartedClient) as arguments and returns anyhow::Result<()>
```

AureliaDolo (approved):

PR #9101, opened 2024-12-03
Fix `WorkspaceOps::watch_entry_oneshot` not being triggered when a missing child of the watched folder is finally synchronized

### LINKED ISSUE #9102: [🐛 | Bug]: GUI sometime not (or partially) refreshing the folder content when news files/folders are added
### Parsec version tested on:

3.12

### Platforms tested on:

Any

### Bug description:

see case 4.a in https://github.com/Scille/parsec-cloud/issues/9002#issuecomment-2511933641

>    1. on a shared workspace, user 1 creates a folder
>    2. user 2 goes into the new folder in the Parsec GUI
>    3. user 1 copies multiple files (~3, each with a size around 60ko to 2.5mo) from within Parsec GUI by using drap&drop (user 1 is on Windows)
>    4. a) user 2 sometime sees only a subset of the copied files (in this case a simple refresh is enough to see the full content of the folder)
>       b) user 2 sometime doesn't see any copied files (i.e. the folder appears still empty).
>
> Case 4.a) and 4.b) are most likely uncorrelated bug

### Relevant output:

_No response_

touilleMan:
close #9102

### DIFF
--- a/bindings/electron/src/index.d.ts
+++ b/bindings/electron/src/index.d.ts
@@ -2697,6 +2697,7 @@ export function clientNewDeviceInvitation(
 ): Promise<Result<NewInvitationInfo, ClientNewDeviceInvitationError>>
 export function clientNewShamirRecoveryInvitation(
     client: number,
+    claimer_user_id: string,
     send_email: boolean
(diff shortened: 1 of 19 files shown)


vxgmichel (approved):
LGTM :+1:

Issue #9079, opened 2024-12-02
[🐛 | Bug]: Wrong name when copying a file in place several times


Ironicbay:
### Parsec version tested on:

3.2.0

### Platforms tested on:

MacOS

### Bug description:

When copying `File.txt` in place several times, we obtain:
- File (2).txt
- File (2) (3).txt
- File (2) (3) (4).txt

etc.

PR #9114, opened 2024-12-04
[MS] Fix copy suffixes

Ironicbay:
Closes https://github.com/Scille/parsec-cloud/issues/9079


client/src/services/fileOperationManager.ts
```diff
@@ -462,7 +466,10 @@ class FileOperationManager {
           return;
         }
         if (!copied) {
-          await deleteFile(data.workspaceHandle, dstPath);
+          // We don't want to delete anything if count is 10 since no file has been created
+          if (filenameCount < 10) {
```
Max-7, line 470: Condition is a bit weird, I wonder if it would not be safer to do `if (!Path.areSame(srcPath, dstPath))` instead. This way we're covered.

The `Path.areSame` function is dumb, it would benefit from calling `Path.normalize()` on both paths before doing the comparaison, which means it should become async. It can be done either in this PR or in another one.

Max-7 (approved):

PR #9127, opened 2024-12-06
Cli shared recovery create - ask for threshold if not provided.

### LINKED ISSUE #9087: [CLI] shared-recovery create: make threshold optional if recipients is not specified
Threshold arg should not be required when recipients are not specified (it should have the same `requires = "recipients",` as weights)

Default recipients are all admins expect author.

Default for threshold could be 
- max(1, ⌊admin_count * k⌋)  with k between 0 and 1. 

What should be k default value, should it be configurable ? (no)

I'd rather go for 2/3

AureliaDolo:
Fix #9087


FirelightFlagboy (commented):
Can you add a test for the threshold input ?

cli/src/commands/shared_recovery/create.rs
```diff
@@ -84,6 +85,16 @@ pub async fn main(shamir_setup: Args) -> anyhow::Result<()> {
...
     };
 
+    let threshold = if let Some(t) = threshold {
+        t
+    } else {
+        Input::<NonZeroU8>::new()
+        .with_prompt(format!(
+            "Choose a threshold between 1 and {}\nThe threshold is the minimum number of recipients that one must gather to recover the account",
+            per_recipient_shares.len()
+        )) .interact_text()?
```
vxgmichel, line 95: This is a blocking sync call in an async function, we should at least acknowledge it with a comment

vxgmichel (approved):
A single comment, LGTM otherwise :+1:

Cargo.toml
```diff
@@ -172,6 +173,7 @@ regex = { version = "1.11.1", default-features = false }
 regex-syntax = { version = "0.8.4", default-features = false }
 reqwest = { version = "0.12.9", default-features = false }
 reqwest-eventsource = { version = "0.6.0", default-features = false }
+rexpect = "0.6.0"
```
FirelightFlagboy, line 176: Can you create an issue to update the cli tests for device/user invitation dance ?

AureliaDolo: #9145

FirelightFlagboy (approved):

PR #9129, opened 2024-12-06
[MS] Replace MsActionBar by new FileControls component

NicoTuxx:


client/src/components/viewers/FileControlsButton.vue
```diff
@@ -0,0 +1,64 @@
...
+
+<script setup lang="ts">
+import { MsImage, Translatable } from 'megashark-lib';
+import { IonButton, IonIcon } from '@ionic/vue';
+
+defineProps<{
+  buttonLabel?: Translatable;
+  icon?: string;
+  image?: string;
+  isDropdown?: boolean;
```
Max-7, line 31: I'm guessing the dropdown is here to add an arrow. Be careful with that, you'll probably want to invert the arrow when the dropdown is opened and you'll be entering a world of pain. Probably better to do a separated dropdown-button if that's the case.

Max-7 (approved):

PR #9132, opened 2024-12-06
Workspace store deadlock protection

touilleMan:
- Hide `WorkspaceStore`'s cache & storage (i.e. local db) internals
- Access must be done through callback API to limit the scope the lock is taken
- A runtime deadlock detector is present in debug build \o/

### DIFF
--- a/libparsec/crates/client/src/workspace/store/file_updater.rs
+++ b/libparsec/crates/client/src/workspace/store/file_updater.rs
@@ -2,6 +2,7 @@
 
 use std::sync::Arc;
 
+use libparsec_platform_async::event::EventListener;
 use libparsec_types::prelude::*;
 
 use crate::certif::{InvalidCertificateError, InvalidKeysBundleError, InvalidManifestError};
@@ -49,8 +50,9 @@ pub(super) async fn for_update_file(
     // Guard's drop will panic if the lock is not released
     macro_rules! release_guard_on_error {
         ($update_guard:expr) => {
-            let mut cache_guard = store.current_view_cache.lock().expect("Mutex is poisoned");
-            cache_guard.lock_update_manifests.release($update_guard);
+            store.data.with_current_view_cache(|cache| {
+                cache.lock_update_manifests.release($update_guard);
+            });
         };
     }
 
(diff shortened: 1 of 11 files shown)


libparsec/crates/client/src/workspace/store/file_updater.rs
```diff
@@ -62,36 +64,48 @@ pub(super) async fn for_update_file(
...
-
-            // 1) Lock for update
+        // 1) Lock for update
 
-            let outcome = cache_guard.lock_update_manifests.take(entry_id);
-            match outcome {
+        enum LockForUpdateOutcome {
+            WaitAndRetryStep1(EventListener),
+            GoToStep3(ManifestUpdateLockGuard),
+            GoToStep4((ManifestUpdateLockGuard, ArcLocalChildManifest)),
```
AureliaDolo, line 72:
```suggestion
            WaitAndRetryStep1(EventListener),
            // There is no GoToStep2 because it's the dispatching step
            GoToStep3(ManifestUpdateLockGuard),
            GoToStep4((ManifestUpdateLockGuard, ArcLocalChildManifest)),
```

Or something like that to clear up why there is no step 2 here

libparsec/crates/client/src/workspace/store/folder_updater.rs
```diff
@@ -60,33 +62,43 @@ pub(super) async fn for_update_folder(
...
         }
 
-        let update_guard = {
-            let mut cache_guard = store.current_view_cache.lock().expect("Mutex is poisoned");
-
-            // 1) Lock for update
-
-            let outcome = cache_guard.lock_update_manifests.take(entry_id);
-            match outcome {
+        enum LockForUpdateOutcome {
```
AureliaDolo, line 65: Isn't that the same struct as in `file_updater.rs` ?

libparsec/crates/client/src/workspace/store/mod.rs
```diff
@@ -107,18 +107,205 @@ pub(super) enum EnsureManifestExistsWithParentError {
...
+    //! the risk of deadlocks.
+    //!
+    //! Indeed, the current view cache and storage (i.e. local database) are both protected
+    //! by a mutex and used in conjunction, so deadlock may occur due to the wrong order of
+    //! acquisition or if a lock it kept too long (especially since locks are not reentrant).
+    //!
+    //! So to solve this issue we only give access to store and storage through a callback
+    //! based API:
+    //! - It better limits and makes explicit the scope of the lock.
+    //! - It is not possible to acquire the storage when the lock is acquired (since the
```
AureliaDolo, line 121:
```suggestion
    //! - It is not possible to acquire the storage lock when the view cache lock is acquired (since the
```

libparsec/crates/client/src/workspace/store/mod.rs
```diff
@@ -107,18 +107,205 @@ pub(super) enum EnsureManifestExistsWithParentError {
...
+        pub fn with_current_view_cache<T>(&self, cb: impl FnOnce(&mut CurrentViewCache) -> T) -> T {
+            #[cfg(debug_assertions)]
+            let _lock_tracker_guard =
+                DataLockTrackerGuard::register(&self.lock_tracking, DataLock::Cache);
+
+            let mut guard = self.current_view_cache.lock().expect("Mutex is poisoned !");
+            cb(&mut guard)
+        }
+
+        // Acquire the lock on the storage.
```
AureliaDolo, line 170:
```suggestion
        /// Acquire the lock on the storage.
```

libparsec/crates/client/src/workspace/store/file_updater.rs
touilleMan: :+1:

libparsec/crates/client/src/workspace/store/folder_updater.rs
touilleMan: Yes, but I don't think it is worth it to factorize it since this is purely implementation detail of `for_update_*` code (that's why it is defined within the function body).

Basically factorize just a type would create an artificial relationship between two unrelated pieces of code that use it, which typically cause issue when modifying one piece but not the other (on top of adding complexity when reading the code since the type definition end up in another file).

However you're right on the fact both `for_update_*` functions have very similar implementations, so we should be able to factorize the entire implementation. I will look into this in my next PR (see #9153) that will modify part of this code anyway.

libparsec/crates/client/src/workspace/store/mod.rs
touilleMan: :+1:

FirelightFlagboy (approved):

PR #9133, opened 2024-12-06
Backport fix inbound sync parent before child event

touilleMan:


newsfragments/9002.bugfix.rst
FirelightFlagboy, line 1: Update the fragment following @mmmarcos review in #9134

FirelightFlagboy (approved):

PR #9152, opened 2024-12-10
Introduce `WorkspaceStore::for_update_sync`

### LINKED ISSUE #8749: Resolve path and confinement point in `for_update_sync_local_only`
This is the first step needed for #8198

This should be done using: https://github.com/Scille/parsec-cloud/blob/96eb831587aa68f860ea7452b4b423876c1679c0/libparsec/crates/client/src/workspace/store/resolve_path.rs#L744

touilleMan:
close #8749

### DIFF
--- a/libparsec/crates/client/src/workspace/store/resolve_path.rs
+++ b/libparsec/crates/client/src/workspace/store/resolve_path.rs
@@ -12,7 +12,8 @@ use crate::{
 
 use super::{
     cache::{
-        populate_cache_from_local_storage_or_server, PopulateCacheFromLocalStorageOrServerError,
+        populate_cache_from_local_storage, populate_cache_from_local_storage_or_server,
+        PopulateCacheFromLocalStorageError, PopulateCacheFromLocalStorageOrServerError,
     },
     per_manifest_update_lock::ManifestUpdateLockGuard,
 };
@@ -89,7 +90,10 @@ pub(super) async fn resolve_path(
 ) -> Result<(ArcLocalChildManifest, PathConfinementPoint), ResolvePathError> {
     resolve_path_maybe_lock_for_update(store, path, false)
         .await
-        .map(|(manifest, confinement_point, _)| (manifest, confinement_point))
+        .map(|(manifest, confinement_point, lock_guard)| {
+            debug_assert!(lock_guard.is_none());
+            (manifest, confinement_point)
+        })
 }
 
 async fn resolve_path_maybe_lock_for_update(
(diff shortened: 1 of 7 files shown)


libparsec/crates/client/src/workspace/store/resolve_path.rs
```diff
@@ -89,7 +90,10 @@ pub(super) async fn resolve_path(
 ) -> Result<(ArcLocalChildManifest, PathConfinementPoint), ResolvePathError> {
     resolve_path_maybe_lock_for_update(store, path, false)
         .await
-        .map(|(manifest, confinement_point, _)| (manifest, confinement_point))
+        .map(|(manifest, confinement_point, lock_guard)| {
+            debug_assert!(lock_guard.is_none());
```
vxgmichel, line 94: Is that a debugging artifact?

libparsec/crates/client/src/workspace/store/resolve_path.rs
```diff
@@ -727,69 +795,273 @@ fn cache_only_retrieve_path_from_id(
...
-                        PopulateCacheFromLocalStorageOrServerError::Internal(err) => {
-                            err.context("cannot fetch manifest").into()
-                        }
-                    })?;
+            CacheOnlyPathRetrievalOutcome::NeedPopulateCache(cache_miss_entry_id) => {
+                match populate_cache_from_local_storage_or_server(store, cache_miss_entry_id).await
+                {
+                    // Happy case :)
+                    Ok(_) => (),
+                    // The entry seems to just not exist, we cannot just
```
vxgmichel, line 1033: we cannot just...?

libparsec/crates/client/src/workspace/store/resolve_path.rs
```diff
@@ -727,69 +795,273 @@ fn cache_only_retrieve_path_from_id(
...
+
+            // We got a cache miss (entry is missing)
+            CacheOnlyPathRetrievalOutcome::NeedPopulateCache(cache_miss_entry_id)
+                if cache_miss_entry_id == entry_id =>
+            {
+                match populate_cache_from_local_storage(store, cache_miss_entry_id).await {
+                    // Happy case :)
+                    Ok(_) => (),
+                    Err(PopulateCacheFromLocalStorageError::EntryNotFound) => {
+                        last_not_found_during_populate = Some(cache_miss_entry_id);
```
vxgmichel, line 928: Isn't it more readable to return a `RetrievePathFromIDEntry::Missing` here? 

Or maybe you need re-enter `cache_only_retrieve_path_from_id` for some lock handling reason?

vxgmichel (approved):

libparsec/crates/client/src/workspace/store/resolve_path.rs
touilleMan: yes that's the thing: `cache_only_retrieve_path_from_id` is responsible to resolve the path and provide the update lock in one atomic step

libparsec/crates/client/src/workspace/store/resolve_path.rs
touilleMan: don't try to argue, WE CANNOT JUST !!!! :trollface:

libparsec/crates/client/src/workspace/store/resolve_path.rs
touilleMan: nope, it's a proper guard

`debug_assert` means the check is only present in debug build, which is because this assert is very defensive (since there is a strong and direct relation between the `false` passed to `resolve_path_maybe_lock_for_update` and the `None` returned).

libparsec/crates/client/src/workspace/store/resolve_path.rs
touilleMan: done.

PR #9193, opened 2024-12-13
The CLI command `user revoke` now takes the email as an argument instead of an option

FirelightFlagboy:


AureliaDolo (approved):

AureliaDolo:
The newsfragment is missing

PR #9194, opened 2024-12-13
Remove env var that interferes with the tests

FirelightFlagboy:


cli/tests/integration/mod.rs
```diff
@@ -44,6 +44,13 @@ fn set_env(tmp_dir: &str, url: &ParsecAddr) {
     // Hidden environ variable only used for CLI testing to customize the throttling time
     // in invitation polling, without that the tests would be much slower for no reason.
     std::env::set_var("_PARSEC_INVITE_POLLING_THROTTLE_MS", "10");
+    std::env::remove_var("PARSEC_DEVICE_ID");
```
touilleMan, line 47: Should add a comment to explain we want to isolate the test from rogue environ variables here

touilleMan (approved):

PR #9225, opened 2024-12-18
[MS] Adds BMS API for calls to create custom orders

### LINKED ISSUE #9223: Add API for creating a new custom order
- The call in services/bms
- The mock in services/bms
- The mock in playwright

Max-7:
Closes #9223 

- [X] Describe your changes

### DIFF
--- a/client/src/services/bms/api.ts
+++ b/client/src/services/bms/api.ts
@@ -10,6 +10,7 @@ import {
   CONNECTION_ERROR_STATUS,
   ChangePasswordQueryData,
   ClientQueryData,
+  CreateCustomOrderRequestQueryData,
   CreateOrganizationQueryData,
   CustomOrderQueryData,
   DataType,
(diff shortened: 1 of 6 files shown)


client/src/services/bms/api.ts
NicoTuxx, line 1: I'm okay with putting dummy value like `IGNORED` until we get the API from the backend, but at the moment you will probably get troubles with the serializer because your dummy value does not match the password rules for example.

NicoTuxx (changes requested):
I would rename to `createCustomOrderRequest` or `requestCustomOrder`

client/src/services/bms/api.ts
Max-7: Ah yeah, thought about it for the email and the postal code but forgot about the password.

NicoTuxx (approved):
LGTM

Issue #9108, opened 2024-12-04
[CLI] Improve display when listing invitation


FirelightFlagboy:
Currently:

```
02ff2c07042dd5bde23fe108b048f76f	cancelled	user (email=[redacted]
801167da3548c3f4613235528f6399f3	finished	device
7b9bbc4c33fa062a8e297b73b8a472ed	idle	device
```

Things that could be improvede:

- The missing `)` for the user email
- The `status` column

Landeers:
+ display invitation link ?

PR #9232, opened 2024-12-19
Aurelia/cli/invit/shared recovery

### LINKED ISSUE #9180: [CLI] Implement shared recovery invitation process
Depends on #8995 
Part of #6090 

- [ ] cli/src/commands/invite/claim.rs
- [ ] cli/src/commands/invite/greet.rs

Search for "Shamir recovery invitation is not supported yet"

TODO: describe manual tests

AureliaDolo:
Fix #9180


cli/src/commands/invite/claim.rs
```diff
@@ -69,9 +78,58 @@ pub async fn main(args: Args) -> anyhow::Result<()> {
             let ctx = step4_device(ctx).await?;
             save_device(ctx, save_mode).await
         }
-        AnyClaimRetrievedInfoCtx::ShamirRecovery(_) => Err(anyhow::anyhow!(
-            "Shamir recovery invitation is not supported yet"
-        )),
+        AnyClaimRetrievedInfoCtx::ShamirRecovery(ctx) => {
+            let mut pick_ctx = ctx;
+
+            let mut device_ctx = loop {
```
FirelightFlagboy, line 84: What happen if the user kill the CLI mid-process to selecting the recipients?
Do we have a cleanup to do?

libparsec/crates/client/src/invite/claimer.rs
```diff
@@ -356,6 +356,13 @@ impl ShamirRecoveryClaimPickRecipientCtx {
         self.shamir_recovery_created_on
     }
 
+    pub fn yet_to_contact_recipients(&self) -> Vec<&ShamirRecoveryRecipient> {
```
FirelightFlagboy, line 359:
```suggestion
    pub fn recipients_without_a_share(&self) -> Vec<&ShamirRecoveryRecipient> {
```

cli/src/commands/invite/claim.rs
```diff
@@ -69,9 +78,58 @@ pub async fn main(args: Args) -> anyhow::Result<()> {
...
+                let ctx = shamir_pick_recipient(&pick_ctx)?;
+                let ctx = step1_shamir(ctx).await?;
+                let ctx = step2_shamir(ctx).await?;
+                let ctx = step3_shamir(ctx).await?;
+                let share_ctx = step4_shamir(ctx).await?;
+                let maybe = pick_ctx.add_share(share_ctx)?;
```
FirelightFlagboy, line 90: We don't allow the user to add more share?

cli/src/commands/invite/claim.rs
```diff
@@ -89,6 +147,19 @@ async fn step0(
     Ok(ctx)
 }
 
+/// Step 0.5: choose recipient
+fn shamir_pick_recipient(
+    ctx: &ShamirRecoveryClaimPickRecipientCtx,
+) -> anyhow::Result<ShamirRecoveryClaimInitialCtx> {
+    let recipients = ctx.yet_to_contact_recipients();
+    let human_recipients: Vec<_> = recipients.iter().map(|r| r.human_handle.clone()).collect();
```
FirelightFlagboy, line 155: If `ShamirRecoveryRecipient` implemented `fmt::Display` we would not need to create a new vec :thinking:

FirelightFlagboy, line 155: Is it required to clone the `human_handle`? could we use a ref instead?

cli/src/commands/invite/claim.rs
```diff
@@ -153,6 +245,23 @@ async fn step2_device(ctx: DeviceClaimInProgress1Ctx) -> anyhow::Result<DeviceCl
...
+/// Step 2: signify trust
+async fn step2_shamir(
+    ctx: ShamirRecoveryClaimInProgress1Ctx,
+) -> anyhow::Result<ShamirRecoveryClaimInProgress2Ctx> {
+    let sas_codes = ctx.generate_greeter_sas_choices(3);
+
+    let selected_sas = Select::new()
+        .items(&sas_codes)
+        .with_prompt("Select code provided by greeter")
+        .interact()?;
```
FirelightFlagboy, line 257: Update the previous selectors for consistency, I recommend creating an helper function

cli/src/commands/invite/greet.rs
```diff
@@ -164,6 +211,22 @@ async fn step3_device(ctx: DeviceGreetInProgress2Ctx) -> anyhow::Result<DeviceGr
...
 
+/// Step 3: signify trust
+async fn step3_shamir(
+    ctx: ShamirRecoveryGreetInProgress2Ctx,
+) -> anyhow::Result<ShamirRecoveryGreetInProgress3Ctx> {
+    let sas_codes = ctx.generate_claimer_sas_choices(3);
+    let selected_sas = Select::new()
+        .items(&sas_codes)
+        .with_prompt("Select code provided by claimer")
+        .interact()?;
```
FirelightFlagboy, line 222: Same reflection concerning updating the already existing selectors

FirelightFlagboy (changes requested):

cli/tests/integration/invitations/shared_recovery.rs
```diff
@@ -0,0 +1,204 @@
...
+
+        if sas_code == first[first.len() - 4..] {
+            locked.send_line("j").unwrap();
+        } else if sas_code == second[second.len() - 4..] {
+            locked.send_line("jj").unwrap();
+        } else if sas_code == third[third.len() - 4..] {
+            locked.send_line("jjj").unwrap();
+        } else {
+            panic!("no corresponding sas code available")
+        }
```
FirelightFlagboy, line 142: Possible to have a more generic approach to scan for the code?

FirelightFlagboy, line 142: Can you extract the logic to reuse it for the greeter part?

cli/tests/integration/invitations/shared_recovery.rs
```diff
@@ -0,0 +1,204 @@
...
+        .unwrap()
+        .exp_string("Code to provide to claimer:")
+        .unwrap();
+    let (_, matched) = p_greeter.lock().unwrap().exp_regex("[A-Z0-9]{4}").unwrap();
+    let sas_code = dbg!(matched[matched.len() - 4..].to_string()); // last 4 chars are the sas code
+
+    // code selection
+
+    claimer.await.unwrap();
+    let cloned_claimer = p_claimer.clone();
```
FirelightFlagboy, line 125: Why do you need to clone here?

cli/tests/integration/invitations/shared_recovery.rs
```diff
@@ -0,0 +1,204 @@
...
+
+    let claimer = tokio::task::spawn(async move {
+        let mut locked = claimer_cloned.lock().unwrap();
+        locked
+            .exp_string("Choose a person to contact now:")
+            .unwrap();
+        // down to choose bob
+        locked.send_line("j").unwrap();
+
+        //  p_claimer.exp_string(&format!("Waiting the greeter {} to start the invitation procedure", bob.human_handle)).unwrap();
```
FirelightFlagboy, line 108: dead code

cli/src/commands/invite/claim.rs
FirelightFlagboy: The number of share is setup during the creation of the device with the command `shared-recovery create`

cli/src/commands/invite/claim.rs
FirelightFlagboy: I understand

cli/tests/integration/invitations/shared_recovery.rs
AureliaDolo: Because I move that into the block just below, and I need to keep access for the last part.

And I'd rather lock the greeter and claimer sessions in their own scope to avoid confusion.

cli/src/commands/invite/claim.rs
AureliaDolo: see #9334

cli/src/commands/invite/list.rs
```diff
@@ -13,6 +13,11 @@ crate::build_main_with_client!(main, list_invite);
 
 pub async fn list_invite(_args: Args, client: &StartedClient) -> anyhow::Result<()> {
     log::trace!("Listing invitations");
+    {
+        let mut spinner = start_spinner("Poll server for new certificates".into());
+        client.poll_server_for_new_certificates().await?;
+        spinner.stop_with_symbol(GREEN_CHECKMARK);
+    }
```
FirelightFlagboy, line 20: **polish (non-blocking):** I've seen this block multiple time, could we create a utility function for that?

cli/src/commands/invite/list.rs
```diff
@@ -31,7 +36,7 @@ pub async fn list_invite(_args: Args, client: &StartedClient) -> anyhow::Result<
                     status,
                     token,
                     ..
-                } => (token, status, format!("user (email={claimer_email}")),
+                } => (token, status, format!("user (email={claimer_email})")),
```
FirelightFlagboy, line 39: **note:** Partially fix #9108 (concerning the closing paren.)

cli/src/commands/invite/claim.rs
```diff
@@ -69,9 +78,61 @@ pub async fn main(args: Args) -> anyhow::Result<()> {
...
+                    ShamirRecoveryClaimMaybeFinalizeCtx::Offline(ctx) => {
+                        let retry = Select::new()
+                            .default(0)
+                            .with_prompt("Unable to join server, do you want to retry ?")
+                            .items(&["yes", "no"])
+                            .interact()?;
+
+                        if retry == 0 {
+                            // yes
+                            device_ctx = ctx;
```
FirelightFlagboy, line 116: **question:** If the server is not available, should we wait a little before retrying? (maybe we should consider an exponential back-off :thinking: )

cli/src/commands/invite/claim.rs
```diff
@@ -69,9 +78,61 @@ pub async fn main(args: Args) -> anyhow::Result<()> {
...
+                        break shamir_recovery_claim_recover_device_ctx;
+                    }
+                    // need more shares
+                    ShamirRecoveryClaimMaybeRecoverDeviceCtx::PickRecipient(
+                        shamir_recovery_claim_pick_recipient_ctx,
+                    ) => pick_ctx = shamir_recovery_claim_pick_recipient_ctx,
+                }
+            };
+
+            let final_ctx = loop {
```
AureliaDolo, line 104: If the server is offline and the user keep retrying it could go forever :/

FirelightFlagboy (approved):

cli/src/commands/invite/claim.rs
AureliaDolo: In his case the little is the reaction time of the user.

But in an ideal world, yes, we should do that.

FirelightFlagboy:
I do not see a newsfragment, that expected?

newsfragments/9180.feature.rst
```diff
@@ -0,0 +1 @@
+[CLI] Implement invite claim and greet for shared recovery.
```
vxgmichel, line 1: Those newsfragments are meant to be user oriented. I think we should have only one of them for the whole #6090 meta-issue, introducing the shared recovery feature in general

cli/src/commands/invite/list.rs
AureliaDolo: see #9346

cli/src/commands/invite/claim.rs
```diff
@@ -214,6 +346,34 @@ async fn step4_device(ctx: DeviceClaimInProgress3Ctx) -> anyhow::Result<DeviceCl
     Ok(ctx)
 }
 
+/// Step 4: retrieve device
```
vxgmichel, line 349: Retrieve share

cli/src/commands/invite/claim.rs
```diff
@@ -89,6 +150,20 @@ async fn step0(
...
+fn shamir_pick_recipient(
+    ctx: &ShamirRecoveryClaimPickRecipientCtx,
+) -> anyhow::Result<ShamirRecoveryClaimInitialCtx> {
+    let recipients = ctx.recipients_without_a_share();
+    let human_recipients: Vec<&_> = recipients.iter().map(|r| &r.human_handle).collect();
+    let selection = Select::new()
+        .default(0)
+        .with_prompt("Choose a person to contact now")
+        .items(&human_recipients)
+        .interact()?;
```
vxgmichel, line 163: Here we ask the user to pick the next recipient but we do not display the current progress (like `2 out of 3 shares already retrieved`). Also, we don't show how the number of shares owned by each recipient.

PR #9235, opened 2024-12-20
[MS] Match available devices using the server addr

### LINKED ISSUE #8922: Use server to match a device against a file link
I think historically, AvailableDevice did not have the server address. But now that it has, we can use it in addition to the org name to match the device.

https://github.com/Scille/parsec-cloud/blob/a394e685551c0884eb876c6444a87e9d9918bac2/client/src/main.ts#L419

Max-7:
Closes #8922 

- [X] Describe your changes


NicoTuxx (approved):
LGTM

PR #9239, opened 2024-12-20
Set mountpoint RW/RO flags based on workspace role

### LINKED ISSUE #9197: Windows file explorer shows "Catastrophic Failure" when creating a file on read-only mountpoint
See #9174 

Probably a way to tell the Windows Explorer that it's alright, the volume is just not writable.

394868530-1b47697a-535e-44ea-9998-9ed4516860cf

### LINKED ISSUE #9229: Implement read-only mountpoint for user with READER role
When user has the READER role on a workspace, a mountpoint flag should be set to read-only. 

Currently this is not implemented.

Linux: https://github.com/Scille/parsec-cloud/blob/d7006acf204265b282949fe1cd8c279555e0cd3b/libparsec/crates/platform_mountpoint/src/unix/mount.rs#L66

Windows: https://github.com/Scille/parsec-cloud/blob/d7006acf204265b282949fe1cd8c279555e0cd3b/libparsec/crates/platform_mountpoint/src/windows/mount.rs#L80-L81

NOTE: The re-mount between RO <-> RW (on user role update) can probably be implemented later. But at least, we should set the right flags on first mount.

----

For reference, this is how it was implemented in V2:
https://github.com/Scille/parsec-cloud/blob/7bd32ebff01202c69793f58608aea3c0af1b5e42/parsec/core/fs/workspacefs/workspacefs.py#L292-L297
https://github.com/Scille/parsec-cloud/blob/7bd32ebff01202c69793f58608aea3c0af1b5e42/parsec/core/mountpoint/winfsp_runner.py#L245

mmmarcos:
Fixes #9229 
Fixes #9197 

NOTE: Re-mount is yet to be implemented in case the user role changes (e.g. from `READER` to `CONTRIBUTOR`) while logged-in. Currently, the user needs to logout and login again in order for workspaces to be re-mounted with the right flags.

### DIFF
--- a/libparsec/crates/platform_mountpoint/src/windows/mount.rs
+++ b/libparsec/crates/platform_mountpoint/src/windows/mount.rs
@@ -41,7 +41,7 @@ impl Mountpoint {
         winfsp_wrs::init()
             .map_err(|err| anyhow::anyhow!("Cannot load the WinFSP DLL: error {}", err))?;
 
-        let (workspace_name, _) = ops.get_current_name_and_self_role();
+        let (workspace_name, role) = ops.get_current_name_and_self_role();
 
         let mountpoint_path = match &ops.config().mountpoint_mount_strategy {
             MountpointMountStrategy::Directory { base_dir } => {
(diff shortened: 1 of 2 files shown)


FirelightFlagboy (approved):

libparsec/crates/platform_mountpoint/src/windows/mount.rs
```diff
@@ -78,7 +81,7 @@ impl Mountpoint {
                 .set_reparse_point_access_check(false)
                 .set_named_streams(false)
                 // TODO: Should detect and re-mount when the workspace switched between read-only and read-write
```
FirelightFlagboy, line 83: This `TODO` should also be present on unix side

FirelightFlagboy:
One notable blocker for this PR is currently we do not re-mount the mountpoint on workspace role change.
For me, that not an ideal solution to ask the user to log-out/in (But that a behavior similar to adding a group on linux)

mmmarcos:
Superseded by #9240

PR #9257, opened 2024-12-23
Improve server logging

touilleMan:
Improve logging configuration in the server

- Don't generate breadcrumb when logs trigger sentry report
- Add `CONSOLE_NO_TIMESTAMP` format (useful in PAAS when the log collecter already provide a timestamp)
- Support colored output from uvicorn logs
- Avoid useless breadcrumb in structlog sentry handler when the log already cause a report
- Remove no longer needed explicit call to exc_info formatter in structlog processors
- Make consistent the log format between stdlib and structlog

### DIFF
--- a/server/parsec/logging.py
+++ b/server/parsec/logging.py
@@ -226,7 +226,7 @@ def _structlog_to_sentry_processor(
         sentry_sdk.capture_event(sentry_event)
 
     # Info & Warning log levels
-    if level_number >= logging.INFO:
+    elif level_number >= logging.INFO:
         # Breadcrumbs do not generate an issue. Instead, they are recorded
         # by Sentry until an event/error occurs, then they are added to
         # the issue as contextual information (useful for debugging!).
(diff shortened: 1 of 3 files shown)


server/parsec/logging.py
```diff
@@ -103,7 +132,9 @@ def info_with_debug_extra(
 # - Configure stdlib with a structlog formatter to have consistent output format
 # - Configure Sentry as a processor in structlog before any formatting processor
 #   (so Sentry doesn't end up with a single string log blob)
-# - Configure Sentry stdlib integration (so 3rd party logs are also captured)
+# - Configure Sentry stdlib integration (so 3rd party logs are also captured).
+#   Note this integration uses it own handler, so it will be totally isolated
```
mmmarcos, line 136:
```suggestion
#   Note this integration uses its own handler, so it will be totally isolated
```

mmmarcos (approved):
LGTM

PR #9261, opened 2024-12-24
[MS] Adds API and mocks for custom order requests list

Max-7:
- [X] Describe your changes

### DIFF
--- a/client/src/services/bms/mockApi.ts
+++ b/client/src/services/bms/mockApi.ts
@@ -9,6 +9,7 @@ import {
   ClientQueryData,
   CreateCustomOrderRequestQueryData,
   CustomOrderQueryData,
+  CustomOrderRequestStatus,
   CustomOrderStatus,
   DataType,
   InvoiceStatus,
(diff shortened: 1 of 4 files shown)


client/src/services/bms/mockApi.ts
```diff
@@ -212,4 +213,59 @@ export const MockedBmsApi = {
       };
     },
   ),
+  getCustomOrderRequests: createMockFunction(
+    'getCustomOrderRequests',
+    async (_token: AuthenticationToken) => {
+      return {
+        status: 200,
+        isError: false,
+        requests: [
```
fabienSvtr, line 222: I think you missed the `type`:
```
data: {
    type: DataType.GetCustomOrderRequests,
    requests: [
          {
...`

File: client/src/services/bms/types.ts [hunk 2]
```diff
@@ -169,6 +178,20 @@ interface RefreshTokenResultData {
...
 }
 
+interface GetCustomOrderRequestsResultData {
+  type: DataType.GetCustomOrderRequests;
+  requests: Array<{
+    id: string;
+    organizationId?: string;
+    describedNeeds: string;
+    users: number;
+    storage: number;
```
fabienSvtr, line 188: The type coming from Parsec Sign is a `string`.
i.e.: `From 100 Go to 300 Go`

client/src/services/bms/types.ts
```diff
@@ -169,6 +178,20 @@ interface RefreshTokenResultData {
...
+    status: CustomOrderRequestStatus;
+    comment: string;
+    orderDate: DateTime;
+  }>;
```
fabienSvtr, line 192: It's an array in your mocks:
`}[];`

fabienSvtr (changes requested):

client/src/services/bms/types.ts
Max-7: No it's not, https://github.com/Scille/parsec-sign/blob/be56bec136eabc482693ce407562a529ea963121/src/services/bmsApi.ts#L116
And it shouldn't be. You're not displaying the string as it is (you're not doing `{{ data.storage }}`, you're also matching the value to the right translation. It becomes very hard to do and to maintain properly if you're trying to match `From 100 Go to 300 Go` to display it correctly in French for example.

You're defining the value that is sent yourself in https://github.com/Scille/parsec-cloud/blob/edc5a147228cba04fa5c175bf2b9d059953d3cb4/client/src/views/client-area/orders/NewOrderModal.vue#L100. The value you send is the `key`, a number, and what you display to the user is the label.

client/src/services/bms/types.ts
Max-7: It is defined as an Array. `requests: Array<{...}>`. I used the `Array<{...}>` syntax instead of `{...}[]`.

fabienSvtr (approved):

Issue #9213, opened 2024-12-16
[MS] FileControlsFullscreen component


NicoTuxx:
see [File viewer specifications](https://github.com/Scille/parsec-cloud/issues/9011)

PR #9272, opened 2025-01-02
[MS] FileControlsFullscreen component

Ironicbay:
Closes https://github.com/Scille/parsec-cloud/issues/9213

### DIFF
--- a/client/src/views/viewers/ImageViewer.vue
+++ b/client/src/views/viewers/ImageViewer.vue
@@ -5,6 +5,7 @@
     <template #viewer>
       <img
         v-if="src.length"
+        ref="imgElement"
         :src="src"
       />
     </template>
@@ -14,6 +15,10 @@
           @change="onChange"
           ref="zoomControl"
         />
+        <file-controls-button
+          @click="toggleFullScreen"
+          :icon="scan"
+        />
       </file-controls>
     </template>
   </file-viewer-wrapper>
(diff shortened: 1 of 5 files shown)


client/src/views/viewers/AudioViewer.vue
```diff
@@ -17,19 +17,20 @@
...
           :icon="getPlaybackIcon()"
           @click="togglePlayback"
-        />
-        <file-controls-button
+        /> -->
+        <!-- <file-controls-button
           :icon="getVolumeIcon()"
           @click="toggleVolume"
-        />
+        /> -->
```
fabienSvtr, line 30: Why did you add comments here?

Ironicbay: Everything is commented on master, I'm un-commenting one component at a time as they are implemented.

client/src/views/viewers/AudioViewer.vue
```diff
@@ -95,6 +97,10 @@ function updateMediaData(event: Event): void {
 //       return pause;
 //   }
 // }
+
+function toggleFullScreen(): void {
+  audioElement.value?.requestFullscreen();
```
Max-7, line 102: > The Element.requestFullscreen() method issues an asynchronous request (https://developer.mozilla.org/en-US/docs/Web/API/Element/requestFullscreen)

So all those functions should be async, and the call to `requestFullscreen` should be awaited. Valid for all calls in all viewers.

Max-7 (changes requested):

Max-7 (approved):

Issue #9031, opened 2024-11-29
[MS] (File viewer features) Viewer section - Accurate rendering (fonts, tables, images, graphics…)


NicoTuxx:
- [x] ImageViewer: #9053 
- [x] DocumentViewer: #9059 
- [x] PdfViewer: #9061 
- [x] SpreadsheetViewer: #9058 : #9561 
- [x] TextViewer: #9057 
- [x] VideoViewer: #9054 

see [File viewer specifications](https://github.com/Scille/parsec-cloud/issues/9011)

PR #9290, opened 2025-01-07
[MS] Improve docx files viewer UX

NicoTuxx:
- replace mammoth.js with docx-preview
- add zoom control to docx files viewer
- add pagination control to docx files viewer
- improve FileControlsInput component (expose isEditing function)
- improve FileControlsPagination component (add page prop to be able to set the current page in the parent component)
- update pagination on scroll in docx files viewer

related to #9031 

https://github.com/user-attachments/assets/16d47eea-bab1-4e8c-93f1-25aea230447e


NicoTuxx:
https://github.com/user-attachments/assets/16d47eea-bab1-4e8c-93f1-25aea230447e

client/src/views/viewers/DocumentViewer.vue
```diff
@@ -7,32 +7,107 @@
...
 import { MsSpinner } from 'megashark-lib';
 
 const props = defineProps<{
   contentInfo: FileContentInfo;
 }>();
 
 const loading = ref(true);
-const htmlContent = ref('');
+const documentContent = ref();
+const pages = ref<NodeList>();
```
Max-7, line 45: I think it would be better to type it as array directly and do the conversion right when we're getting the pages. This way we manipulate an `Array<HTMLElement>` everytime instead of a weird NodeList.
The type would be `ref<Array<HTMLElement>> = ref([])` and we can do `pages.value = Array.from(documentContent.value.querySelectorAll('section.docx-page'));` to get the pages.
It means no conversion later on, and you're not forced to use `forEach` to iter over the list.

client/src/views/viewers/DocumentViewer.vue
```diff
@@ -7,32 +7,107 @@
...
   loading.value = true;
-  const result = await mammoth.convertToHtml({ arrayBuffer: props.contentInfo.data.buffer });
-  htmlContent.value = result.value;
+  await renderAsync(props.contentInfo.data.buffer, documentContent.value, undefined, {
+    ignoreLastRenderedPageBreak: false,
+    className: 'docx-page',
+  });
+
+  // recommended way to get all pages by the library developer
+  pages.value = documentContent.value.querySelectorAll('section.docx-page');
```
Max-7, line 58: I've tried with two documents, both of them had only one `.docx-page`, meaning the pagination always displays `1 / 1`

client/src/views/viewers/DocumentViewer.vue
```diff
@@ -7,32 +7,107 @@
...
+    (targetPage as HTMLElement).scrollIntoView({ behavior: 'smooth' });
+  }
+  currentPage.value = pageIndex;
+}
+
+function onScroll(): void {
+  if (!pages.value) {
+    return;
+  }
+  // const target = event.target as HTMLElement;
```
Max-7, line 83: Clean this comment.

Max-7 (changes requested):

Max-7:
You could also handle errors right now. A simple try/catch around the `renderAsync` that displays an error message. It can be tested in web mode in FileViewer, just return the DocumentViewer for images and try to open an image.

client/src/views/viewers/DocumentViewer.vue
NicoTuxx: As explained on slack, only documents containing manual pagebreaks or Word automatic pagebreaks are supported.

client/src/views/viewers/DocumentViewer.vue
NicoTuxx: Done

client/src/services/storageManager.ts
```diff
@@ -171,7 +171,7 @@ export class StorageManager {
       confirmBeforeQuit: data.confirmBeforeQuit !== undefined ? data.confirmBeforeQuit : StorageManager.DEFAULT_CONFIG.confirmBeforeQuit,
       meteredConnection: data.meteredConnection !== undefined ? data.synchroWifiOnly : StorageManager.DEFAULT_CONFIG.meteredConnection,
       unsyncFiles: data.unsyncFiles !== undefined ? data.unsyncFiles : StorageManager.DEFAULT_CONFIG.unsyncFiles,
-      skipViewers: true, // data.skipViewers !== undefined ? data.skipViewers : StorageManager.DEFAULT_CONFIG.skipViewers,
+      skipViewers: false, // data.skipViewers !== undefined ? data.skipViewers : StorageManager.DEFAULT_CONFIG.skipViewers,
```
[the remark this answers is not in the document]
NicoTuxx: good catch

client/src/views/viewers/DocumentViewer.vue
NicoTuxx: done

Max-7 (approved):

PR #9308, opened 2025-01-09
Add workflow dispatch for nightly releaser workflow

vxgmichel:
Also bump softprops/action-gh-release to v2.2.1


FirelightFlagboy (approved):
LGTM :eyes:

PR #9318, opened 2025-01-10
[MS] Improve PdfViewer (multi-pages, zoom, scroll, etc.)

NicoTuxx:


client/src/views/viewers/PdfViewer.vue
```diff
@@ -59,24 +63,22 @@ const props = defineProps<{
 
 const loading = ref(true);
 const error = ref('');
-const canvas = ref();
-const pdfRef = ref();
 const currentPage = ref(1);
 const pdf: Ref<pdfjs.PDFDocumentProxy | null> = shallowRef(null);
-const actions: Ref<Array<{ icon?: string; text?: Translatable; handler: () => void }>> = ref([]);
+const pdfContainer = ref();
```
Max-7, line 68: Type the different ref when possible.

client/src/views/viewers/PdfViewer.vue
```diff
@@ -100,45 +108,162 @@ async function loadPage(pageIndex: number): Promise<void> {
 
   try {
     const page = await pdf.value.getPage(pageIndex);
-    const viewport = page.getViewport({ scale: scale.value });
+
```
Max-7, line 111: You could add
```typescript
if (needsMocks() && pageIndex === 3) {
  throw new Error('Failed to load page');
}
```
this would make a page not load and allow us to easily see how errors are rendered (and have e2e tests for this case).
I mention it because it currently bugs (maybe the deviceAspectRatio is not taken into account here or some other minor thing).
Screenshot 2025-01-16 072711

client/src/views/viewers/PdfViewer.vue
```diff
@@ -100,45 +108,162 @@ async function loadPage(pageIndex: number): Promise<void> {
...
+    const canvasElement = canvas.value.at(pageIndex - 1);
+    if (!canvasElement) {
+      loading.value = false;
```
Max-7, line 114: You don't need to set `loading` to false, it is handled by the `finally`

client/src/views/viewers/PdfViewer.vue
```diff
@@ -100,45 +108,162 @@ async function loadPage(pageIndex: number): Promise<void> {
...
-      transform: transform,
-      viewport: viewport,
-    };
-    page.render(renderContext);
-    currentPage.value = pageIndex;
+    canvasElement.removeAttribute('data-rendered');
   } catch (e: any) {
+    const canvasElement = canvas.value.at(pageIndex - 1);
+    if (!canvasElement) {
+      loading.value = false;
```
Max-7, line 129: Same here, the `finally` will take care of that.

Max-7 (approved):

client/src/views/viewers/PdfViewer.vue
NicoTuxx: Fix PDF page loading error on initialization

PR #9319, opened 2025-01-10
Add missing shamir recovery GUI bindings

### LINKED ISSUE #8998: [Shamir port] add GUI bindings

vxgmichel:
Close #8998

### DIFF
--- a/bindings/generator/templates/binding_electron_meths.rs.j2
+++ b/bindings/generator/templates/binding_electron_meths.rs.j2
@@ -14,6 +14,8 @@
 JsObject
 {%- elif type.kind == "list" -%}
 JsArray
+{%- elif type.kind == "set" -%}
+JsArray
 {%- elif type.kind == "tuple" -%}
 JsArray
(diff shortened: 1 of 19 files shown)


bindings/generator/templates/binding_electron_meths.rs.j2
```diff
@@ -14,6 +14,8 @@
 JsObject
 {%- elif type.kind == "list" -%}
 JsArray
+{%- elif type.kind == "set" -%}
+JsArray
```
FirelightFlagboy, line 18: **suggestion:** Ideally, we should be using `JsSet` but that type is not exposed in `neon`, that would require to do something similar to `Map`

https://github.com/Scille/parsec-cloud/pull/9319/files#diff-a2fa8ef212d1ee4f61387826d485369fbca8bd6e66bd656cb99a432bbb3957d1R354-R364

touilleMan (approved):

Max-7 (approved):

PR #9345, opened 2025-01-14
[MS] Replaced workspaces by recent workspaces in sidebar

### LINKED ISSUE #9321: Update sidebar with quick access
- Don't show all workspaces
- Display last 5 workspaces accessed

Max-7:
Closes #9321 

Remember which workspaces were added to the quick access list, and store the list so it stays when the app closes, when user logs out, or when switching org.
A workspace is added to the list when clicked on directly, or when accessed by a file link.

- [X] Describe your changes


Ironicbay (changes requested):
- The `Recent workspaces` list is not updated when accessing a favorite from the sidebar
- The workspace order follows a "most recent" order when accessing workspaces, but the order is shuffled after relogging, and is not updated when visiting a workspace in the `recent` list. If the list displays `Workspace 1` `Workspace 2` `Workspace 3` (1 being at the top and most recent), when visiting `Workspace 3` I expect it to be bumped to the top of the list.

Ironicbay (approved):

Issue #9173, opened 2024-12-11
[↩️ Backport][🐛 Bug]: [MacOS] Cannot drop file to finder folder of mountpoint


mmmarcos:
Backport fix #8907 to branch `releases/3.2`

FirelightFlagboy:
Fixed by #9351

Issue #8991, opened 2024-11-25
[🐛 | Bug]: MacOS: file saving not working from native MacOS software


Ironicbay:
### Parsec version tested on:

3.2.0

### Platforms tested on:

MacOS

### Bug description:

This had to be fixed in V2 as well, see https://github.com/Scille/parsec-cloud/pull/2211 and https://github.com/Scille/parsec-cloud/pull/4198

Saving works fine from non-Apple software, so it's likely the same issue we previously encountered.

To replicate:
- Open a `.txt` file inside a Parsec workspace, it should open with `TextEdit` by default
- `cmd+S` fails.

FirelightFlagboy:
Fixed by #9351

Issue #8976, opened 2024-11-22
[🐛 | Bug]: Copy/paste on macOS mountpoint not working


fabienSvtr:
### Parsec version tested on:

3.2.0 rc3

### Platforms tested on:

MacOS

### Bug description:

We can't copy/paste a file:
- from the finder to a mountpoint
- from the mountpoint himself inside another folder

The file appears on the app but with size of 0 byte.

### Relevant output:
https://github.com/user-attachments/assets/8a75c885-cd49-4934-8bbe-9366ee79b7e7

Ironicbay:
Fixed in https://github.com/Scille/parsec-cloud/pull/9114

FirelightFlagboy:
Fixed by #9351

PR #9351, opened 2025-01-16
Fix mountpoint issues on macOS

FirelightFlagboy:


libparsec/crates/platform_mountpoint/src/unix/filesystem.rs
```diff
@@ -1121,7 +1113,23 @@ impl fuser::Filesystem for Filesystem {
 
         // TODO: support atime/utime change ?
 
-        reply.manual().error(libc::ENOSYS);
+        // Nothing to set, just return the file attr
```
touilleMan, line 1116: The comment is a bit off since the previous one stated that there is things to change BUT we don't support them

touilleMan (commented):
Can you add a test, that would rely on a subprocess running your snippet of osascript 

(typically put in `libparsec/crates/platform_mountpoint/tests/unit/operations/unix_setattr.rs`)

And we also need some linux tests given the code change impact them too (and besides there is no tests for setattr right now !)

For instances:
- macos only test with osascript the copy of a file
- (linux, macos) test setattr used for truncate to an arbitrary size (same size as current one, bigger size, smaller size, zero size)
- (linux, macos) test setattr used to change the others fields allowed by it but not supported (atime, etc.)
- (linux, macos) test setattr on a non-existing file
- (linux, macos) test setattr to change size on a folder

Obviously we can create an issue to implement part of them in a separated PR

libparsec/crates/platform_mountpoint/src/unix/filesystem.rs
```diff
@@ -1033,6 +1024,7 @@ impl fuser::Filesystem for Filesystem {
 
                 return;
             } else {
+                // FIXME: Currently I'm only returning the file attributes without truncating the file
```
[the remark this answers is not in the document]
FirelightFlagboy: The current code is buggy: It should Resize according the `size` param and not truncate the file

libparsec/crates/platform_mountpoint/src/unix/filesystem.rs
```diff
@@ -1121,7 +1113,23 @@ impl fuser::Filesystem for Filesystem {
 
         // TODO: support atime/utime change ?
```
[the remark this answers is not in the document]
FirelightFlagboy: None, `setattr` was call with all Option set to None.
I'm 90% sure that caused by the fuser's fix for an invalid ctime

libparsec/crates/platform_mountpoint/src/unix/filesystem.rs
FirelightFlagboy: For me it's nothing to set on parsec side either because we ignore them or we don't handle them.
I'm open to suggestion to improve the comment

deny.toml
```diff
@@ -93,6 +93,14 @@ We are not affected since sqlx is used on the client side to store the files' da
 Before being stored in SQLite, that data is splitted into chunk and encoded, so we are below the overflow limit of 4GB.
 """
 
+[[advisories.ignore]]
```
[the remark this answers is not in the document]
FirelightFlagboy: We use `blahaj` on master

mmmarcos:
@FirelightFlagboy this PR fixes multiple issues (#8976, #8991, #9173) but it is not clearly specified which changes fix which issues. Please summarize it in the PR description.

libparsec/crates/platform_mountpoint/src/unix/filesystem.rs
```diff
@@ -1121,7 +1113,23 @@ impl fuser::Filesystem for Filesystem {
...
+        let ops = self.ops.clone();
+
+        self.tokio_handle.spawn(async move {
+            let res = getattr_from_path(&ops, path, ino, uid, gid).await;
+
+            match res {
+                Ok(attr) => reply.manual().attr(&TTL, &attr),
+                Err(errno) => reply.manual().error(errno),
+            }
+        });
```
mmmarcos, line 1132: Is this code what fixes both copy/paste and save issues? In any case, we should add a comment explaining briefly why the changes below are needed. 

We (you) had a hard time investigating and figuring out what was going on. Let's track that here as a comment.

FirelightFlagboy:
> @FirelightFlagboy this PR fixes multiple issues (#8976, #8991, #9173) but it is not clearly specified which changes fix which issues. Please summarize it in the PR description.

Those are specified in the body of the first 2 commits

libparsec/crates/platform_mountpoint/src/unix/filesystem.rs
FirelightFlagboy: There is a comment just above

FirelightFlagboy:
> Can you add a test, that would rely on a subprocess running your snippet of osascript [...]

I've created the issue #9353

libparsec/crates/platform_mountpoint/src/unix/filesystem.rs
mmmarcos: I'd also mention #8976 and #8991 here in the comment because it is not evident that this was needed to fix those problems.

libparsec/crates/platform_mountpoint/src/unix/filesystem.rs
mmmarcos: https://github.com/Scille/parsec-cloud/pull/9351#discussion_r1918227957

mmmarcos:
> > Can you add a test, that would rely on a subprocess running your snippet of osascript [...]
> 
> I've created the issue #9353 

I'd still add some tests in this PR because it fixes major bugs. IMO we should add at least test regarding `unix_setattr.rs`.

PR #9356, opened 2025-01-16
Bump to API v5

vxgmichel:
This is required to the changes made in https://github.com/Scille/parsec-cloud/pull/6644


libparsec/crates/protocol/src/version.rs
```diff
@@ -14,6 +14,7 @@ use libparsec_types::prelude::*;
 // - v4.1 (Parsec 3.2+):
 //   * Add `ShamirRecovery` variants to `invite_list` and `invite_info`
 //   * Add `invite_new_shamir_recovery` command
+// v5 (Parsec 3.3+): TODO
```
touilleMan, line 17: Should create an issue to track the fact we need to update this before 3.3 release

touilleMan (approved):

libparsec/crates/protocol/src/version.rs
vxgmichel: ad797d8d07527399bca8277bc2b393a61fd0aff5

touilleMan (approved):

PR #9384, opened 2025-01-20
Use Bigint to bind u64/i64

touilleMan:


FirelightFlagboy (commented):
@Max-7 The usage on Number on API that use BigInt seems brittle to me.

Is it just a temporary fix?

FirelightFlagboy (approved):

Ironicbay (approved):

PR #9421, opened 2025-01-24
fix(l/c/t/s/local_device_file.rs): Rm passwd from debug ouput of `Device{Save,Access}Stategy`

FirelightFlagboy:


libparsec/crates/types/src/local_device_file.rs
```diff
@@ -202,8 +202,19 @@ impl DeviceSaveStrategy {
     }
 }
 
+impl std::fmt::Debug for DeviceSaveStrategy {
```
touilleMan, line 205: I guess the main issue is the fact `Password` is implemented as `pub type Password = zeroize::Zeroizing<String>;`

Can we instead implement it as a propert struct (so that we can implement a Debug trait that hide the actual password), then implement a Deref to string.

This seems a more robust approach since it ensures the issue won't occur again whenever we add this `Password` type to a new structure.

touilleMan: Besides, I guess `pub type SecretKeyPassphrase = zeroize::Zeroizing<String>;` should be corrected the same way too

FirelightFlagboy:
Superseded by #9424

PR #9422, opened 2025-01-24
[MS] Update file size when sync starts and ends

### LINKED ISSUE #9388: [🐛 | Bug]: Invalid displayed file's size after copy of file from mountpoint
### Parsec version tested on:

3.2.4-rc.2

### Platforms tested on:

Linux

### Bug description:

1. Have parsec already open at a workspace folder
2. From the OS explorer navigate to the workspace folder in the mountpoint
3. Copy a file (I had the issue with a file that was ~24M big)
4. During the copy the size value is updated on the app but get stuck at one point (for me it was ~2.5MB) an never update after

> This is a minor display issue, if you refresh the view (change folder then going back) the size is displayed with the correct value

### Relevant output:

_No response_

Max-7:
Closes #9388 

File stat when sync starts and ends. Not ideal, the file size will still not be updated while the copy is progressing, but it will update eventually.

- [X] Describe your changes

### DIFF
--- a/newsfragments/9388.bugfix.rst
+++ b/newsfragments/9388.bugfix.rst
@@ -1,1 +1,1 @@
-Correctly refreshes the size of a file when copying from the file explorer
+Correctly refreshes the displayed size of a file when copying from the file explorer
--- a/newsfragments/9388.bugfix.rst
+++ b/newsfragments/9388.bugfix.rst
@@ -1,0 +1,1 @@
+Correctly refreshes the size of a file when copying from the file explorer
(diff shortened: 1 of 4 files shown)


newsfragments/9388.bugfix.rst
```diff
@@ -0,0 +1 @@
+Correctly refreshes the size of a file when copying from the file explorer
```
FirelightFlagboy, line 1: **nitpick:** Since it's just a display issue, I would specify that in the fragment

```suggestion
Correctly refreshes the displayed size of a file when copying from the file explorer
```

NicoTuxx (approved):
LGTM

PR #9441, opened 2025-01-27
Protect against concurrent realm creation

### LINKED ISSUE #9428: Realm create is not protected against concurrent creation
When running the test `libparsec_client client::tests::with_monitors::sharing` using the testbed server running with a postgresql cluster, an error is sometimes generated (about 1 in 2).

This is due to this test performing a two realm creation request concurrently:
```
=== realm_create ===
9b47efcd-9c62-4332-be2f-6fc3e6aa101d
=== realm_create ===
9b47efcd-9c62-4332-be2f-6fc3e6aa101d
=== done ===
=== done ===
```

This causes one of these two `_q_create_realm` queries to not return any row:
https://github.com/Scille/parsec-cloud/blob/abce6ef375da4ee46ad6e5d7fefa2da1835dd033/server/parsec/components/postgresql/realm_create.py#L34-L99

Which causes in turn this `assert` to fail:
https://github.com/Scille/parsec-cloud/blob/abce6ef375da4ee46ad6e5d7fefa2da1835dd033/server/parsec/components/postgresql/realm_create.py#L167

A couple of comments:

1. Is this normal that the client ends up performing concurrent realm creation in this test or is this the symptom of another bug?

2. Using `UNION` with `LIMIT 1` as a way to return a default value if the previous `WITH` statement did not return any row is very unreliable as far as I recall. In particular, this assumes that `UNION` lists the rows in order which Postgresql does not guarantee
    > UNION effectively appends the result of query2 to the result of query1 (although there is no guarantee that this is the order in which the rows are actually returned).  
    https://www.postgresql.org/docs/current/queries-union.html#QUERIES-UNION

3. The request returns no row at all because the realm creation happens in several steps:
   - insertion into `realm`
   - insertion into `realm_user_role`
   - insertion into `realm_topic`
   
   That means that the second request can notice that the realm id is part of the `realm` table but the corresponding realm does not appear in the `realm_topic` table

vxgmichel:
Also protect against concurrent key rotation

Fix #9428

### DIFF
--- a/server/parsec/components/postgresql/queries/q_auth_lock_realm.py
+++ b/server/parsec/components/postgresql/queries/q_auth_lock_realm.py
@@ -29,6 +29,22 @@ class LockRealmWriteRealmData:
     realm_user_current_role: RealmRole
 
 
+# Two important notes about preventing race conditions during realm-related operations:
+# - 1. Selecting the `realm_topic` table with a `FOR SHARE` or `FOR UPDATE` clause
+#   allows us to effectively lock the access to the realm, for read and write operations
+#   respectively. More specifically, concurrent read-locking queries on the same realm
+#   will block until the ongoing write transaction is done. Similarly, concurrent write-locking
+#   queries on the same realm will block until all ongoing transactions are done (read or write).
+# - 2. It is tempting to use a `WITH` clause to have a single query that both
+#   locks the realm topic and fetches information about the realm. However, this
+#   would expose use to race conditions, as the `WITH` clause does not provide
+#   any guarantee on the order of execution of its subqueries. More generally,
+#   the order of execution of subqueries in an SQL query is not guaranteed, whether
+#   it is using a `WITH` clause, the `(SELECT ...)` syntax, or the `FROM ... JOIN ...`.
+#   This is why we have to split the `lock_realm_*` functions in two queries, one to
+#   lock the realm topic and one to fetch the information about the realm.
+
+
 _Q_LOCK_REALM_TEMPLATE = """
 SELECT realm._id AS realm_internal_id
 FROM realm
@@ -110,9 +126,14 @@ async def _do_lock_realm(
     realm_internal_id = await conn.fetchval(
         *lock_query(organization_internal_id=organization_internal_id, realm_id=realm_id)
     )
-    if realm_internal_id is None:
-        return LockRealmWriteRealmBadOutcome.REALM_NOT_FOUND
-    assert isinstance(realm_internal_id, int)
+
+    match realm_internal_id:
+        case int() as realm_internal_id:
+            pass
+        case None:
+            return LockRealmWriteRealmBadOutcome.REALM_NOT_FOUND
+        case unknown:
+            assert False, repr(unknown)
 
     row = await conn.fetchrow(
         *_q_realm_info_after_lock(
--- a/server/parsec/components/postgresql/queries/q_auth_lock_realm.py
+++ b/server/parsec/components/postgresql/queries/q_auth_lock_realm.py
@@ -3,7 +3,6 @@ from __future__ import annotations
 
 from dataclasses import dataclass
 from enum import auto
-from typing import Any
 
 from parsec._parsec import (
     DateTime,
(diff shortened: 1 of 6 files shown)


server/tests/api_v5/authenticated/test_realm_rotate_key.py
```diff
@@ -676,3 +676,69 @@ async def test_authenticated_realm_rotate_key_sequester_service_unavailable(
...
+            per_sequester_service_keys_bundle_access=None,
+            keys_bundle=b"<keys bundle>",
+        )
+        for certif in certifs
+        for _ in range(2)
+    ]
+
+    # Run all the coroutines concurrently
+    reps = await asyncio.gather(*coroutines)
+    for rep in reps:
```
touilleMan, line 736:
```suggestion
    assert len(reps) == 10
    assert len(r for r in reps if rep == authenticated_cmds.latest.realm_rotate_key.RepOk()) == 1
    assert len(r for r in reps if isinstance(rep, (authenticated_cmds.latest.realm_rotate_key.RepBadKeyIndex, authenticated_cmds.latest.realm_rotate_key.RepRequireGreaterTimestamp)) == 9
```

server/tests/api_v5/authenticated/test_realm_rotate_key.py
```diff
@@ -676,3 +676,69 @@ async def test_authenticated_realm_rotate_key_sequester_service_unavailable(
...
+        assert isinstance(
+            rep,
+            (
+                authenticated_cmds.latest.realm_rotate_key.RepOk,
+                authenticated_cmds.latest.realm_rotate_key.RepBadKeyIndex,
+                authenticated_cmds.latest.realm_rotate_key.RepRequireGreaterTimestamp,
```
touilleMan, line 742: Why is there two different possible errors here ?

server/tests/api_v5/authenticated/test_realm_create.py
```diff
@@ -240,3 +242,42 @@ async def do():
...
+        coolorg.alice.realm_create(
+            realm_role_certificate=certif.dump_and_sign(coolorg.alice.signing_key)
+        )
+        for certif in certifs
+        for _ in range(2)
+    ]
+
+    # Run all requests concurrently
+    reps = await asyncio.gather(*coroutines, return_exceptions=True)
+    for rep in reps:
```
touilleMan, line 276:
```suggestion

    assert len(reps) == 10
    assert len(r for r in reps if rep == authenticated_cmds.latest.realm_create.RepOk()) == 1
    assert len(r for r in reps if isinstance(rep, authenticated_cmds.latest.realm_create.RepRealmAlreadyExists) == 9
```

server/parsec/components/postgresql/queries/q_auth_lock_realm.py
```diff
@@ -96,32 +89,42 @@ async def lock_realm_read(
...
 ) -> LockRealmWriteRealmData | LockRealmWriteRealmBadOutcome:
-    row = await conn.fetchrow(*args)
+    # 0) Lock realm
+
+    realm_internal_id = await conn.fetchval(
+        *lock_query(organization_internal_id=organization_internal_id, realm_id=realm_id)
+    )
+    if realm_internal_id is None:
+        return LockRealmWriteRealmBadOutcome.REALM_NOT_FOUND
+    assert isinstance(realm_internal_id, int)
```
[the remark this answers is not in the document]
touilleMan: Or use the good old `match ...: case int(): ... case None: ... case unknown: ...` format, as it tells right away we are validating data from the DB here

touilleMan (approved):

server/tests/api_v5/authenticated/test_realm_rotate_key.py
vxgmichel: Because it depends in which order the concurrent requests are executed (we want to run them concurrently but still one has to run slightly before the other one in order to display a potential concurrency issue on the insertion query)

server/tests/api_v5/authenticated/test_realm_rotate_key.py
vxgmichel: 0eddeee5f805ee1acfcee74c29f9f22a8b7bdc07

server/tests/api_v5/authenticated/test_realm_create.py
vxgmichel: 0eddeee5f805ee1acfcee74c29f9f22a8b7bdc07

server/parsec/components/postgresql/queries/q_auth_lock_realm.py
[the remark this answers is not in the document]
vxgmichel: 0eddeee5f805ee1acfcee74c29f9f22a8b7bdc07

server/parsec/components/postgresql/realm_create.py
```diff
@@ -31,70 +30,71 @@
 )
 from parsec.events import EventRealmCertificate
 
+_q_lock_realm_creation = Q(
```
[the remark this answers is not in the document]
vxgmichel: 0eddeee5f805ee1acfcee74c29f9f22a8b7bdc07

server/parsec/components/postgresql/queries/q_auth_lock_realm.py
```diff
@@ -31,47 +30,42 @@ class LockRealmWriteRealmData:
 
 
 _Q_LOCK_REALM_TEMPLATE = """
```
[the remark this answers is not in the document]
vxgmichel: 0eddeee5f805ee1acfcee74c29f9f22a8b7bdc07

server/parsec/components/postgresql/realm_create.py
```diff
@@ -152,37 +152,46 @@ async def realm_create(
...
-
-    match row["last_timestamp"]:
-        case DateTime() as last_realm_certificate_timestamp:
-            pass
-        case unknown:
-            assert False, unknown
+    # The realm already exists
 
-    if not inserted:
+    if realm_internal_id is None:
```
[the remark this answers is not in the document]
vxgmichel: 0eddeee5f805ee1acfcee74c29f9f22a8b7bdc07

PR #9463, opened 2025-01-30
Fix `/administration/stats` endpoint when `at` is provided and before an existing organization

### LINKED ISSUE #9427: [Bug] Server stats at a date when the organization doesn't exist yet
It's unclear from the code base whether those stats should be absent from the response or should be all set to zero.

The memory implementation sets all the stats to zero without considering the organization `created_on` date:
https://github.com/Scille/parsec-cloud/blob/abce6ef375da4ee46ad6e5d7fefa2da1835dd033/server/parsec/components/memory/organization.py#L259

The postgresql implementation returns an error if the organization did not exist at that time:
https://github.com/Scille/parsec-cloud/blob/abce6ef375da4ee46ad6e5d7fefa2da1835dd033/server/parsec/components/postgresql/organization_stats.py#L94-L96

However this exception is not properly handled in `organization_server_stats` which causes a `TypeError`: 
https://github.com/Scille/parsec-cloud/blob/abce6ef375da4ee46ad6e5d7fefa2da1835dd033/server/parsec/components/postgresql/organization_stats.py#L165

There is a `test_server_stats_at` test:
https://github.com/Scille/parsec-cloud/blob/abce6ef375da4ee46ad6e5d7fefa2da1835dd033/server/tests/administration/test_stats.py#L301

But it doesn't test this case since the minimal organization is created at linux epoch.

vxgmichel:
Fix #9427

### DIFF
--- a/server/parsec/components/memory/organization.py
+++ b/server/parsec/components/memory/organization.py
@@ -254,14 +254,22 @@ class MemoryOrganizationComponent(BaseOrganizationComponent):
         if user.current_profile != UserProfile.ADMIN:
             return OrganizationStatsAsUserBadOutcome.AUTHOR_NOT_ALLOWED
 
-        return self._stats(org, at)
+        match self._stats(org, at):
+            case OrganizationStats() as stats:
+                return stats
+            case OrganizationStatsBadOutcome.ORGANIZATION_NOT_FOUND:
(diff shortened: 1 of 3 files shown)


mmmarcos (approved):
LGTM :+1:

PR #9471, opened 2025-01-30
Ignore unknown events in the client, at deserialization.

### LINKED ISSUE #9410: Ignore unkown events in the client
In order to be able to add new events without breaking compatibility, we should ignore unrecognized events in the client.

At the moment, getting an unknown event would stop the monitor:
https://github.com/Scille/parsec-cloud/blob/02313cdb7eade3fe66fd191217e21d1ea6fe98fa/libparsec/crates/client/src/monitors/connection.rs#L218

AureliaDolo:
Fix #9410 

The added test case fails with BadContent without the patch.


libparsec/crates/client_connection/src/authenticated_cmds/sse.rs
```diff
@@ -94,10 +94,15 @@ where
             let raw = BASE64
                 .decode(event.data.as_bytes())
                 .map_err(|_| ConnectionError::BadContent)?;
-            let cooked =
-                T::api_load_response(raw.as_ref()).map_err(|_| ConnectionError::BadContent)?;
-            SSEResponseOrMissedEvents::Response(cooked)
+            match T::api_load_response(raw.as_ref()) {
+                Ok(cooked) => SSEResponseOrMissedEvents::Response(dbg!(cooked)),
+                Err(_e) => {
```
FirelightFlagboy, line 99: **suggestion:** If we had to debug why an event is not processed by the client, it would be simpler if we log the error

```suggestion
                Err(e) => {
                    log::trace!("Invalid event data: {e}")
```

FirelightFlagboy (approved):

Issue #9019, opened 2024-11-29
[MS] FileControlsDropdownItem component


NicoTuxx:
see [File viewer specifications](https://github.com/Scille/parsec-cloud/issues/9011)
- [ ] Toggle
- [ ] List checkable option

PR #9476, opened 2025-01-31
[MS] FileControlsDropdown & FileControlsDropdownItem

### LINKED ISSUE #9018: FIleControlsDropdown component
see [File viewer specifications](https://github.com/Scille/parsec-cloud/issues/9011)

Ironicbay:
Closes #9018 and #9019 

Capture d’écran 2025-02-06 à 11 55 11
Capture d’écran 2025-02-06 à 11 55 20

### DIFF
--- a/client/src/components/viewers/controls/FileControlsDropdownPopover.vue
+++ b/client/src/components/viewers/controls/FileControlsDropdownPopover.vue
@@ -1,0 +1,96 @@
+<!-- Parsec Cloud (https://parsec.cloud) Copyright (c) BUSL-1.1 2016-present Scille SAS -->
+
+<template>
+  <ion-list class="file-controls-dropdown-popover">
+    <div
+      v-if="parentHistory.length > 0"
+      @click="onBackClicked"
+      class="back-button"
+    >
+      <ion-icon :icon="chevronBack" />
+      <ion-text class="title">
+        {{ parentHistory.slice(-1)[0].name }}
+      </ion-text>
+    </div>
+    <div
+      class="divider"
+      v-if="parentHistory.length > 0"
+    />
+    <file-controls-dropdown-item
+      v-for="(item, index) in displayedItems"
+      :key="index"
+      :is-active="item.isActive"
+      :item="item"
+      @click="onOptionClick"
+    />
+  </ion-list>
+</template>
+
+<script setup lang="ts">
+import { FileControlsDropdownItem, FileControlsDropdownItemContent } from '@/components/viewers';
+import { IonIcon, IonList, IonText, popoverController } from '@ionic/vue';
+import { ref } from 'vue';
+import { chevronBack } from 'ionicons/icons';
+import { Translatable } from 'megashark-lib';
+
+const props = defineProps<{
+  items: FileControlsDropdownItemContent[];
+}>();
+
+const displayedItems = ref<FileControlsDropdownItemContent[]>(props.items);
+const parentHistory = ref<Array<{ items: FileControlsDropdownItemContent[]; name: Translatable }>>([]);
+
+async function onOptionClick(option: FileControlsDropdownItemContent): Promise<void> {
+  if (option.children) {
+    parentHistory.value.push({ items: displayedItems.value, name: option.label });
+    displayedItems.value = option.children;
+  }
+  if (option.callback) {
+    await option.callback();
+  }
+  if (option.dismissToParent) {
+    await onBackClicked();
+  } else if (option.dismissPopover) {
+    await popoverController.dismiss({
+      option: option,
+    });
+  }
+}
+
+async function onBackClicked(): Promise<void> {
+  const parentItem = parentHistory.value.pop();
+  if (parentItem) {
+    displayedItems.value = parentItem.items;
+  }
+}
+</script>
+
+<style lang="scss" scoped>
+.file-controls-dropdown-popover {
+  overflow-y: auto;
+  max-height: 20em;
+
+  .back-button {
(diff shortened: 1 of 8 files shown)


client/src/components/viewers/types.ts
```diff
@@ -0,0 +1,18 @@
+// Parsec Cloud (https://parsec.cloud) Copyright (c) BUSL-1.1 2016-present Scille SAS
+
+import { Translatable } from 'megashark-lib';
+
+export interface FileControlsDropdownItemContent {
+  id: number;
```
Max-7, line 6: The `id` makes defining the items a bit more annoying, is there really a use for it?

client/src/components/viewers/types.ts
```diff
@@ -0,0 +1,18 @@
...
+  label: Translatable;
+  icon?: string;
+  children?: FileControlsDropdownGroup;
+  dismissPopover?: boolean;
+  dismissToParent?: boolean;
+  callback?: () => any;
```
Max-7, line 12: Shouldn't callback be mandatory? And since it's awaited, define it as `Promise<any>` instead.

client/src/components/viewers/types.ts
```diff
@@ -0,0 +1,18 @@
...
+}
+
+export interface FileControlsDropdownGroup {
+  items: FileControlsDropdownItemContent[];
+  activeItem: number | undefined;
```
Max-7, line 17: This tells which item is the active one? Why not add an optional `isActive?: boolean` to the item? It would mean that you could just have `Array<FileControlsDropdownItemContent>` instead of creating a whole `FileControlsDropdownGroup` interface.
It would make the API more intuitive to use, and it would probably makes the code simpler (not necessarily shorter because you'd have to go through the list to toggle the selection, but simpler).

client/src/views/viewers/SpreadsheetViewer.vue
```diff
@@ -68,18 +80,23 @@ const currentPage = ref('');
...
 const zoomControl = ref();
 const zoomLevel = ref(1);
+const dropdownItems = ref<FileControlsDropdownGroup>({ activeItem: 0, items: [] });
 
 onMounted(async () => {
   try {
     workbook = XLSX.read(props.contentInfo.data, { type: 'array' });
     pages.value = workbook.SheetNames;
     await switchToPage(workbook.SheetNames[0]);
+    let i = 0;
```
Max-7, line 93: You should be able to iter with `for (const [index, page] of pages.value.entries())` to get the index. Just a little tip.

client/src/components/viewers/controls/FileControlsDropdownPopover.vue
```diff
@@ -0,0 +1,101 @@
...
+import { ref } from 'vue';
+import { chevronBack } from 'ionicons/icons';
+import { Translatable } from 'megashark-lib';
+
+const props = defineProps<{
+  items: FileControlsDropdownGroup;
+}>();
+
+const displayedItems = ref<FileControlsDropdownGroup>(props.items);
+const parentHistory = ref<Array<FileControlsDropdownGroup>>([]);
```
Max-7, line 41: Since those variables seem to be synchronous (you always do the same operation on both at once, unless I missed something) I would group them together:
`const blabla: Array<{history: FileControlsDropdownGroup, names: Translatable}> = [];`
This ensures that they stay synced.

client/src/components/viewers/types.ts
```diff
@@ -0,0 +1,13 @@
+// Parsec Cloud (https://parsec.cloud) Copyright (c) BUSL-1.1 2016-present Scille SAS
+
+import { Translatable } from 'megashark-lib';
+
+export interface FileControlsDropdownItemContent {
+  label: Translatable;
+  callback?: (arg?: any) => Promise<any>;
```
Max-7, line 7: Why would you need arguments? You're calling the callback yourself and you don't pass any.

client/src/views/viewers/SpreadsheetViewer.vue
```diff
@@ -26,11 +26,16 @@
     <template #controls>
       <file-controls v-show="!loading && workbook">
         <file-controls-group>
-          <file-controls-button
+          <!-- <file-controls-button
```
Max-7, line 29: Remove the commented part.

Max-7 (approved):

client/src/components/viewers/controls/FileControlsButton.vue
NicoTuxx, line 1: Are this modifications still relevant, since it does not seems to be used by the new components.

client/src/views/viewers/SpreadsheetViewer.vue
```diff
@@ -26,11 +26,10 @@
...
         <file-controls-group>
-          <file-controls-button
-            v-for="(action, key) in actions"
-            :key="key"
-            @click="action.handler"
-            :label="action.text"
+          <file-controls-dropdown
+            :items="dropdownItems"
+            :title="I18n.valueAsTranslatable(currentPage)"
+            :icon="chevronUp"
```
NicoTuxx, line 32: Icon may be chevronDown? It seems weird at the moment even if it's opening to the top?
@fabienSvtr your opinion on this?

image

client/src/views/viewers/VideoViewer.vue
NicoTuxx, line 1: The settings menu was great to test your component, but do we want to keep this?
Maybe comment, mock or remove it?

NicoTuxx (approved):

PR #9462, opened 2025-01-29
Restrict admin invitations

Max-7:
- [X] Keep changes in the pull request as small as possible
- [X] Ensure the commit history is sanitized
- [X] Give a meaningful title to your PR
- [X] Describe your changes
- [X] Link any related issue in the description

PR #9486, opened 2025-01-31
Add comments and renaming about inital ctx cancellation

vxgmichel:
A commit addressing @touilleMan's comments I forgot to push on #9462.


touilleMan (approved):
lgtm

PR #9503, opened 2025-02-03
Add, expose and use claimer_user_wait_all_peers

### LINKED ISSUE #9413: Share user invite between all administrators
Follow up after #9078 

Now that the server and data model supports it, this feature can be fully implemented.

Two different approaches have been considered:

1. Add an extra step in the API so that the user can explicitly choose which administrator to contact. This requires to update `libparsec`, the bindings and the frontend.

2. Update `UserClaimInitialCtx::wait_peer` to simultaneously wait for all the administrators until one of them joins. This still requires to update `libparsec` but less changes to bindings and the frontend.

The simpler approach should first be done as a PoC to validate the feature.

vxgmichel:
Close #9413

TODO:
- [x] Add newsfragment


vxgmichel:
Screen shot of the updated text when a preferred greeter is available:
image

And when a preferred user is not available:
image

Max-7 (approved):

touilleMan (approved):

PR #9525, opened 2025-02-03
Update and fix outdated invitation event

### LINKED ISSUE #9411: Remove or update outdated invitation event
This `Invitation` event is unused and outdated. It should be removed or updated in API v5:

https://github.com/Scille/parsec-cloud/blob/02313cdb7eade3fe66fd191217e21d1ea6fe98fa/libparsec/crates/protocol/schema/authenticated_cmds/events_listen.json5#L53-L65

This also includes removing the `claimers_ready` attribute from the `BaseInviteComponent` which is also unused
https://github.com/Scille/parsec-cloud/blob/02313cdb7eade3fe66fd191217e21d1ea6fe98fa/server/parsec/components/invite.py#L574

The invitation status structure might also need to be updated since the `READY` variant is not used anymore:
https://github.com/Scille/parsec-cloud/blob/02313cdb7eade3fe66fd191217e21d1ea6fe98fa/libparsec/crates/types/src/invite.rs#L69-L74

vxgmichel:
Close #9411


server/parsec/components/postgresql/invite.py
```diff
@@ -928,9 +927,35 @@ async def q_take_invitation_create_write_lock(
...
 _q_list_administrators = Q(
     """
+SELECT
+    user_id
+FROM user_
+INNER JOIN organization ON user_.organization = organization._id
+WHERE
+    organization.organization_id = $organization_id
+    AND user_.current_profile = 'ADMIN'
+    AND user_.revoked_on IS NULL
```
FirelightFlagboy, line 939: **question:** Why do we use a `inner join` here, can't we directly filter the associated organization of the user using the param `$organization_id`?

vxgmichel: `user._organization` is the internal id to the organization (an integer), while `$organization_id` is the `OrganizationID` (like `CoolOrg`). We have the same kind of indirection for user/user_id, and device/device_id.

Some requests sometimes re-use an internal id that has been returned by a previous query in order the avoid the extra join, but it's not the case here. It could be done though, it's just an optimization for later.

touilleMan (approved):
:pizza: :pinched_fingers:

PR #9554, opened 2025-02-04
Web compact to load devices

FirelightFlagboy:


libparsec/crates/platform_device_loader/src/web/internal.rs
```diff
@@ -0,0 +1,306 @@
...
+    }
+
+    pub(crate) fn save_device_file(
+        &self,
+        key: &str,
+        file_data: &DeviceFile,
+    ) -> Result<(), SaveDeviceFileError> {
+        let data = file_data.dump();
+        let b64_data = BASE64_STANDARD.encode(&data);
+        self.storage
```
touilleMan, line 186: I'd rather avoid having to rely on a non-atomic operation here (i.e. storing two things in the localStorage, since there is no transaction in localStorage)

Non-atomicity is always more paint since we have to take into account weird corner cases. like an item is present in the list, but its corresponding device is not, or vice-versa. Regarding the last point, it theory it shouldn't occur since we store the device first, then modify the list, however I didn't see any guarantee in localStorage specification about that so who knows....

Also it saves more error handling when adding/removing devices since we no longer have to deal with the fact we might not be able to deserialize the list (and this case should we overwrite the list and risk losing access ? or should we do nothing and potentially never be able to add new device ever again...)

On top of that if `add_item_to_list`/`remove_item_from_list` is buggy at some point, we will lose access to the previous device :/

So ideally we should remove the list entirely and rely on querying the localStorage to discover the available devices. Given the number of devices is expected to be very low, I think we can simply do something like this:
```js
> sessionStorage.setItem("device/123", "<device 123 payload>")
undefined
> sessionStorage.setItem("device/abc", "<device abc payload>")
undefined
> for (i = 0; i < localStorage.length; i++)  { var key = sessionStorage.key(i); if (key.startsWith("device/")) { console.log(sessionStorage.getItem(key)); } }
<device abc payload>
<device 123 payload>
```

libparsec/crates/platform_device_loader/src/web/error.rs
```diff
@@ -0,0 +1,205 @@
...
+        #[display("Failed to remove item {key} from storage ({error:?})")]
+        RemoveItemStorage {
+            key: String,
+            error: JsValue
+        }
+    };
+    JsonDeserializationError = {
+        JsonDecode(serde_json::Error)
+    };
+    JsonSerError = {
```
touilleMan, line 62: In the codebase we consider the encoding is always possible (since we encode only valid data), so there is no need for this error (we should just panic if it occurs)

see what is done in the `libparsec_types::format_v0_dump` function

libparsec/crates/platform_device_loader/src/web/error.rs
```diff
@@ -0,0 +1,205 @@
...
+        || InvalidPathError;
+}
+
+impl From<LoadDeviceError> for crate::LoadDeviceError {
+    fn from(value: LoadDeviceError) -> Self {
+        match value {
+            LoadDeviceError::InvalidFileType => Self::InvalidData,
+            LoadDeviceError::GetSecretKey(_) => Self::DecryptionFailed,
+            LoadDeviceError::DecryptAndLoad(e) => e.into(),
+            LoadDeviceError::InvalidPath { .. } => Self::InvalidPath(anyhow::anyhow!("{value}")),
```
touilleMan, line 126: In web, the path is totally under our control:
- in prod it is a dummy value set at compilation time
- in test, we are generating this value to whatever suits us

So I think it is safe to assume the path is always valid utf8, and just panic if that's not the case (we should just have a comment next to panic to explain why it should never occur in practice)

libparsec/crates/platform_device_loader/src/web/error.rs
```diff
@@ -0,0 +1,205 @@
...
+
+impl_from_new_storage_error!(
+    crate::LoadDeviceError,
+    crate::SaveDeviceError,
+    crate::ChangeAuthentificationError,
+    crate::ArchiveDeviceError,
+    crate::RemoveDeviceError
+);
+
+error_set::error_set! {
```
touilleMan, line 37: This approach is interesting, we should definitely investigate to see how well suited it is to simplify our current cumbersome error handling system ^^

Next time, it would be better to inform us prior to introduce change that add new dependencies or challenge the standard way we do certain things :pray:

Cargo.lock
```diff
@@ -2153,7 +2181,9 @@ dependencies = [
 name = "libparsec_platform_device_loader"
 version = "3.3.1-a.0+dev"
 dependencies = [
+ "base64",
```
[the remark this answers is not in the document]
FirelightFlagboy: done

libparsec/crates/platform_device_loader/Cargo.toml
```diff
@@ -7,6 +7,7 @@ homepage.workspace = true
 license.workspace = true
 version.workspace = true
 repository.workspace = true
+autotests = false
```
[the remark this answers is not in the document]
FirelightFlagboy: done

libparsec/crates/platform_device_loader/src/web/internal.rs
```diff
@@ -0,0 +1,306 @@
...
+impl Storage {
+    /// Entry in the local storage where the devices keys are stored.
+    const LIST_DEV_KEY: &'static str = "parsec_devices";
+    /// Entry in the local storage where archived devices keys are stored.
+    const ARCHIVED_DEV_KEY_LIST: &'static str = "archived_parsec_devices";
+
+    pub(crate) fn new() -> Result<Self, NewStorageError> {
+        let window = web_sys::window().ok_or(NewStorageError::NoWindow)?;
+        let storage = if cfg!(test) {
+            // Use session storage for tests, So the browser does not try to persist the data.
```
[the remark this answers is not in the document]
FirelightFlagboy: We will hit this code path if the testbed abstraction return nothing in the function defined in `lib.rs`.

Do you think about something?

libparsec/crates/platform_device_loader/src/web/error.rs
FirelightFlagboy: I wanted to do an introduction on Slack but I was side tracked by more urgent issues :sweat_smile:

libparsec/crates/platform_device_loader/src/web/error.rs
FirelightFlagboy: done

libparsec/crates/platform_device_loader/tests/units/list.rs
```diff
@@ -26,28 +30,21 @@ async fn list_no_devices(tmp_path: TmpPath, #[case] path_exists: bool) {
 #[parsec_test]
 async fn ignore_invalid_items(tmp_path: TmpPath) {
     let devices_dir = tmp_path.join("devices");
+    #[cfg(not(target_arch = "wasm32"))]
```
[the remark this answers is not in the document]
FirelightFlagboy: done

libparsec/crates/platform_device_loader/src/web/error.rs
FirelightFlagboy: done

libparsec/crates/platform_device_loader/src/web/internal.rs
FirelightFlagboy: done

libparsec/crates/platform_device_loader/tests/units/list.rs
```diff
@@ -26,311 +30,160 @@ async fn list_no_devices(tmp_path: TmpPath, #[case] path_exists: bool) {
...
-        .unwrap();
-    }
+    crate::tests::utils::create_device_file(&devices_dir.join("dir/a_file.keys"), b"dummy");
 
     let devices = list_available_devices(&tmp_path).await;
     p_assert_eq!(devices, []);
 }
 
 #[parsec_test]
 async fn list_devices(tmp_path: TmpPath) {
```
FirelightFlagboy, line 48: **chore (blocking):** Restore the test master

libparsec/crates/platform_device_loader/src/web/internal.rs
```diff
@@ -0,0 +1,230 @@
...
+        Ok(Self { storage })
+    }
+
+    pub(crate) fn list_available_devices(
+        &self,
+        config_dir: &Path,
+    ) -> Result<Vec<AvailableDevice>, ListAvailableDevicesError> {
+        let device_prefix = config_dir
+            .to_str()
+            .expect("The internal path is not a valid UTF-8 string");
```
touilleMan, line 36: It is not obvious why `config_dir` is guaranteed to be a valid UTF8 path

`config_dir` is a config variable, so it is suggested that it can be set to whatever value by the user the application.

Of course here we are on web so `config_dir` is a fake config value that doesn't get configured (hence why this conversion is safe).
However this feels rather flimsy: if later on we start using `config_dir` for it original purpose, then we would have introduced a bug here that will silently sail to production :/

So given the performance gain is negligible here, I think we'd rather instead use the safe conversion of [to_string_lossy](https://doc.rust-lang.org/std/path/struct.Path.html#method.to_string_lossy)

```suggestion
         let device_prefix = config_dir.to_string_lossy();
```

libparsec/crates/platform_device_loader/src/web/internal.rs
```diff
@@ -0,0 +1,230 @@
...
+            .map_err(Into::into)
+    }
+
+    pub(crate) fn load_device(
+        &self,
+        access: &DeviceAccessStrategy,
+    ) -> Result<(Arc<LocalDevice>, DateTime), LoadDeviceError> {
+        let key_path = access.key_file();
+        let key = key_path
+            .to_str()
```
touilleMan, line 97: same as above

touilleMan (approved):

libparsec/crates/platform_device_loader/src/web/internal.rs
FirelightFlagboy: done

libparsec/crates/platform_device_loader/src/web/internal.rs
FirelightFlagboy: done

libparsec/crates/platform_device_loader/src/web/internal.rs
```diff
@@ -0,0 +1,230 @@
...
+        &self,
+        access: &DeviceAccessStrategy,
+        device: &LocalDevice,
+        created_on: DateTime,
+    ) -> Result<AvailableDevice, SaveDeviceError> {
+        let protected_on = device.now();
+        let server_url = crate::server_url_from_device(device);
+        let key_file_path = access.key_file();
+        let key_file = key_file_path
+            .to_str()
```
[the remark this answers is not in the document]
FirelightFlagboy: done

libparsec/crates/platform_device_loader/src/web/internal.rs
```diff
@@ -0,0 +1,230 @@
...
+            .expect("The internal path is not a valid UTF-8 string");
+
+        match access {
+            DeviceAccessStrategy::Keyring { .. } => todo!("Save keyring device"),
+            DeviceAccessStrategy::Password { password, .. } => {
+                let key_algo = crate::generate_default_password_algorithm_parameters();
+                let key = crate::secret_key_from_password(password, &key_algo)
+                    .expect("Salt has the correct length");
```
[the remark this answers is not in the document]
FirelightFlagboy: done

libparsec/crates/platform_device_loader/src/web/internal.rs
```diff
@@ -0,0 +1,230 @@
...
+            .set_item(key, &b64_data)
+            .map_err(|e| SaveDeviceFileError::SetItemStorage {
+                key: key.to_owned(),
+                error: e,
+            })
+    }
+
+    pub(crate) fn archive_device(&self, path: &Path) -> Result<(), ArchiveDeviceError> {
+        let key = path
+            .to_str()
```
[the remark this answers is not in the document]
FirelightFlagboy: done

libparsec/crates/platform_device_loader/src/web/internal.rs
```diff
@@ -0,0 +1,230 @@
...
+            .remove_item(key)
+            .map_err(|error| ArchiveDeviceError::RemoveItemStorage {
+                key: key.to_owned(),
+                error,
+            })
+    }
+
+    pub(crate) fn remove_device(&self, path: &Path) -> Result<(), RemoveDeviceError> {
+        let key = path
+            .to_str()
```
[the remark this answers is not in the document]
FirelightFlagboy: done

PR #9555, opened 2025-02-04
Add GreetingAttemptReady and GreetingAttemptCancelled event

### LINKED ISSUE #9412: Add new events dedicated to invitation handling
Those new events should be aimed at helping with the synchronization of claimers and greeters. 

This is especially useful for the greeting of a claimer during a user invitation. Since the user can be greeted by any administrator, an event could be generated in order to notify that a given claimer is waiting for an administrator willing to invite them. If a claimer chooses to wait for a specific greeter, then another event could be sent to notify the user that a claimer is specifically waiting for them. Another useful event would to signify that a claimer has been joined by a given greeter, for instance to notify other administrators that the invitation is already being taking care of.

To sum it up, there could be:
- a `ClaimerConnected` event
- a `ClaimerWaitingForPeer` event
- a `ClaimerJoinedGreeter` event

vxgmichel:
Close #9412


touilleMan (approved):
LGTM

PR #9583, opened 2025-02-08
Set file permissions to 0o500 when the workspace is read-only on linux

### LINKED ISSUE #9389: [🐛 | Bug]: Mountpoint show write perm when user as the Reader profile
### Parsec version tested on:

3.2.4-rc.2

### Platforms tested on:

Linux

### Bug description:

1. Have parsec with a workspace where you are `reader`
2. In the explorer, go to the mountpoint of the workspace 
3. Observe that the mountpoint folder and its children have the permission `0o700` (where it should be `0o500`)

### Relevant output:

_No response_

vxgmichel:
Fix #9389

### DIFF
--- a/libparsec/crates/platform_mountpoint/src/unix/filesystem.rs
+++ b/libparsec/crates/platform_mountpoint/src/unix/filesystem.rs
@@ -33,9 +33,12 @@ const TTL: std::time::Duration = std::time::Duration::ZERO;
 const GENERATION: u64 = 0;
 /// The default block size, set to 512 KB.
 const BLOCK_SIZE: u64 = 512 * 1024;
-/// Default permissions for files and folders.
+/// Read-write permissions for files and folders.
 /// Equivalent to `chmod` flags `all=,u=rwx`.
-const PERMISSIONS: u16 = 0o700;
+const READ_WRITE_PERMISSIONS: u16 = 0o700;
+/// Read-only permissions for files and folders.
+/// Equivalent to `chmod` flags `all=,u=rx`.
+const READ_ONLY_PERMISSIONS: u16 = 0o500;
 
 fn os_name_to_entry_name(name: &OsStr) -> EntryNameResult<EntryName> {
     name.to_str()
@@ -43,9 +46,20 @@ fn os_name_to_entry_name(name: &OsStr) -> EntryNameResult<EntryName> {
         .ok_or(EntryNameError::InvalidName)?
 }
 
-fn file_stat_to_file_attr(stat: FileStat, inode: Inode, uid: u32, gid: u32) -> fuser::FileAttr {
+fn file_stat_to_file_attr(
+    stat: FileStat,
+    inode: Inode,
+    uid: u32,
+    gid: u32,
+    is_read_only: bool,
+) -> fuser::FileAttr {
     let created: std::time::SystemTime = stat.created.into();
     let updated: std::time::SystemTime = stat.updated.into();
+    let perm = if is_read_only {
+        READ_ONLY_PERMISSIONS
+    } else {
+        READ_WRITE_PERMISSIONS
+    };
     fuser::FileAttr {
         ino: inode,
         size: stat.size,
@@ -55,7 +69,7 @@ fn file_stat_to_file_attr(stat: FileStat, inode: Inode, uid: u32, gid: u32) -> f
         ctime: updated,
         crtime: created,
         kind: fuser::FileType::RegularFile,
-        perm: PERMISSIONS,
+        perm,
         nlink: 1,
         uid,
         gid,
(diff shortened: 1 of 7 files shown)


libparsec/crates/platform_mountpoint/src/unix/filesystem.rs
```diff
@@ -120,25 +143,57 @@ fn entry_stat_to_file_attr(stat: EntryStat, inode: Inode, uid: u32, gid: u32) ->
     }
 }
 
+enum MaybeReadOnly {
+    ReadOnlyMountpoint,
+    ReadWriteMountpoint,
+}
+
+impl MaybeReadOnly {
+    fn get(&self, ops: &WorkspaceOps) -> bool {
```
touilleMan, line 152: It is not needed to check the current role since access control is done when calling the workspace ops methods, instead we can just rely on the `is_read_only` computed once at mountpoint start, and return the correct error when workspace ops return a not allowed error code

libparsec/crates/platform_mountpoint/src/unix/filesystem.rs
```diff
@@ -35,17 +35,29 @@ const GENERATION: u64 = 0;
...
+const READ_ONLY_PERMISSIONS: u16 = 0o500;
 
 fn os_name_to_entry_name(name: &OsStr) -> EntryNameResult<EntryName> {
     name.to_str()
         .map(|s| s.parse())
         .ok_or(EntryNameError::InvalidName)?
 }
 
-fn file_stat_to_file_attr(stat: FileStat, inode: Inode, uid: u32, gid: u32) -> fuser::FileAttr {
+fn file_stat_to_file_attr(
```
touilleMan, line 47: you can move this function (and `entry_stat_to_file_attr` & `reply_with_lookup`) as `Filesystem` methods if you think it is simpler/cleaner to have `self.is_read_only` (hence not having to modify the call sites) instead of passing it as parameter

touilleMan (approved):

libparsec/crates/platform_mountpoint/src/unix/filesystem.rs
```diff
@@ -35,17 +35,29 @@ const GENERATION: u64 = 0;
 const BLOCK_SIZE: u64 = 512 * 1024;
 /// Default permissions for files and folders.
```
FirelightFlagboy, line 36: **nitpick:** Comment not up-to-date

libparsec/crates/platform_mountpoint/src/unix/filesystem.rs
vxgmichel: 28a6bdf858bce7419dfadb38c812d9f486d02aa7

libparsec/crates/platform_mountpoint/src/unix/filesystem.rs
vxgmichel: 28a6bdf858bce7419dfadb38c812d9f486d02aa7

libparsec/crates/platform_mountpoint/src/unix/filesystem.rs
vxgmichel: I don't think we can due to the use of `self.tokio_handle.spawn(async move { ... })`

libparsec/crates/platform_mountpoint/src/unix/filesystem.rs
```diff
@@ -33,19 +33,33 @@ const TTL: std::time::Duration = std::time::Duration::ZERO;
 const GENERATION: u64 = 0;
 /// The default block size, set to 512 KB.
 const BLOCK_SIZE: u64 = 512 * 1024;
-/// Default permissions for files and folders.
+/// Read-write permissions for files and folders.
 /// Equivalent to `chmod` flags `all=,u=rwx`.
-const PERMISSIONS: u16 = 0o700;
+const READ_WRITE_PERMISSIONS: u16 = 0o700;
+/// Read-only permissions for files and folders.
+/// Equivalent to `chmod` flags `all=,u=r.x`.
```
FirelightFlagboy, line 40: **nitpick:** 
```suggestion
/// Equivalent to `chmod` flags `all=,u=rx`.
```

newsfragments/9389.bugfix.rst
```diff
@@ -0,0 +1 @@
+Do not report write permissions on linux when a workspace is read-only.
```
FirelightFlagboy, line 1: **nitpick:** macOS is also affected

```suggestion
Do not report write permissions on unix when a workspace is read-only.
```

```suggestion
Do not report write permissions on linux and macOS when a workspace is read-only.
```

vxgmichel: 0ae9e99bf9934842c99122f10fd2199d9e174ae1

libparsec/crates/platform_mountpoint/src/unix/filesystem.rs
vxgmichel: 3d9870f5c3b89fb553ce0aacf705ca9adb5f690d

PR #9584, opened 2025-02-09
Check `new_chunks` in `fd_flush` to properly reshape when flushing a file 

### LINKED ISSUE #7478: Reshape in fd_flush does not use manifest.new_chunks to build the new data
Copied from https://github.com/Scille/parsec-cloud/pull/7431#issuecomment-2143554327

Reshape in `fd_flush` does not use `manifest.new_chunks` to build the new data to turn into a block:
https://github.com/Scille/parsec-cloud/blob/5da5a48b635e1c02e7486a0411a729540e1f83c8/libparsec/crates/client/src/workspace/transactions/fd_flush.rs#L128-L155

Compare with what's done in `fd_read`:
https://github.com/Scille/parsec-cloud/blob/5da5a48b635e1c02e7486a0411a729540e1f83c8/libparsec/crates/client/src/workspace/transactions/fd_read.rs#L82-L103

In practice, this causes the file manifests to not be fully reshaped when `force_reshape_and_flush` is called.

The corresponding test should be added.

vxgmichel:
Fix #7478

### DIFF
--- a/libparsec/crates/client/src/workspace/transactions/fd_flush.rs
+++ b/libparsec/crates/client/src/workspace/transactions/fd_flush.rs
@@ -136,24 +136,48 @@ async fn reshape(
         let mut local_miss = false;
         let start = reshape.destination().start;
         for chunk_view in reshape.source().iter() {
-            let outcome = ops.store.get_chunk_or_block_local_only(chunk_view).await;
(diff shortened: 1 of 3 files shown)


touilleMan (approved):
:+1:

PR #9594, opened 2025-02-10
fix(rust-dev-loader): Re-enable list device test

FirelightFlagboy:
- Update the generated blob
- Add logging when a device is ignored because of an error

### DIFF
--- a/libparsec/crates/platform_device_loader/Cargo.toml
+++ b/libparsec/crates/platform_device_loader/Cargo.toml
@@ -24,6 +24,7 @@ libparsec_types = { workspace = true }
 # `alloc` feature is required to properly erase struct containing vec/string
 zeroize = { workspace = true, features = ["alloc"] }
 log = { workspace = true }
+itertools = { workspace = true }
 
 [target.'cfg(not(target_arch = "wasm32"))'.dependencies]
 dirs = { workspace = true }
--- a/libparsec/crates/platform_device_loader/tests/list.rs
+++ b/libparsec/crates/platform_device_loader/tests/list.rs
@@ -55,143 +55,282 @@ async fn ignore_invalid_items(tmp_path: TmpPath) {
 }
 
 #[parsec_test]
-#[ignore = "TODO: scheme has changed, must regenerate the dump"]
 async fn list_devices(tmp_path: TmpPath) {
-    let alice_file_path = tmp_path
-        .join("devices/52e905dc5a505f068cbec94298768f877054016080c0a0d09992730385966db6.keys");
+    // 1. Generate raw data
+
+    // Keyring
+
+    let keyring_expected = DeviceFileKeyring {
+        created_on: "2000-01-01T00:00:00Z".parse().unwrap(),
+        protected_on: "2000-01-01T00:00:01Z".parse().unwrap(),
+        server_url: "https://parsec.invalid".to_string(),
+        organization_id: "CoolOrg".parse().unwrap(),
+        user_id: "alice".parse().unwrap(),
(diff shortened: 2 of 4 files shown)


libparsec/crates/platform_device_loader/tests/list.rs
```diff
@@ -55,97 +55,104 @@ async fn ignore_invalid_items(tmp_path: TmpPath) {
...
-    //   device_id: "alice@dev1"
-    //   organization_id: "CoolOrg"
-    //   slug: "f78292422e#CoolOrg#alice@dev1"
-    //   algorithm:
-    //     type: "ARGON2ID"
-    //     salt: b"salt"
-    //     opslimit: 1
-    //     memlimit_kb: 8
-    //     parallelism: 1
+    // Generated from Rust implementation (Parsec v3.2.5-a.0+dev)
```
touilleMan, line 65: Should add another one item with `DeviceFile::Recovery` flavor

libparsec/crates/platform_device_loader/Cargo.toml
```diff
@@ -45,3 +46,4 @@ libparsec_tests_lite = { workspace = true }
 libparsec_tests_fixtures = { workspace = true, features = [
     "test-with-platform-device-loader-testbed",
 ] }
+hex = { workspace = true }
```
touilleMan, line 49: This is not needed as `hex!` is already provided as part of `libparsec_tests_fixtures::prelude::*`

FirelightFlagboy: The `hex` does not provide the macro `hex!` but the function `encode` that is used to generate the blob for the device.
I wanted to keep the dependency listed to simplify updating the blob, but maybe it's not needed

libparsec/crates/platform_device_loader/tests/list.rs
FirelightFlagboy: done

libparsec/crates/platform_device_loader/tests/list.rs
```diff
@@ -55,101 +55,135 @@ async fn ignore_invalid_items(tmp_path: TmpPath) {
...
+    // let alice_device = DeviceFile::Password(DeviceFilePassword {
```
touilleMan, line 67: Use `misc/test_expected_payload_cooker.py` to generate the payload description, as this provides an actual view on the actual content of the serialized data, and not how the serialized data has been generated (hence ensuring the serialization code is not doing unexpected things)

libparsec/crates/platform_device_loader/Cargo.toml
touilleMan: I don't thing this is needed since we already have `misc/test_expected_payload_cooker.py` that deals with this in a standardized way.

touilleMan (changes requested):

libparsec/crates/platform_device_loader/tests/list.rs
touilleMan: btw a fix way to correct this is to simply replace both comment and raw content by the stuff from commit 1881bfd6bbe21dfa88b350acf20256993a6f533f ^^

FirelightFlagboy: done

libparsec/crates/platform_device_loader/Cargo.toml
FirelightFlagboy: done

touilleMan (approved):

PR #9641, opened 2025-02-11
Use send_signal to send events in postgresql server implementation

### LINKED ISSUE #9611: [🐛 | Bug]: Not informed of new invitations
### Parsec version tested on:

3.3.0-rc.6

### Platforms tested on:

Windows

### Bug description:

When inviting a user, the invitation pop-up is not displayed. Need to log out and log in again to refresh it.

### Relevant output:

_No response_

vxgmichel:
Fix #9611

### DIFF
--- a/server/parsec/components/postgresql/events.py
+++ b/server/parsec/components/postgresql/events.py
@@ -27,22 +27,24 @@ logger = get_logger()
 
 
 class PGEventBus(EventBus):
+    """The `EventBus.send` method is not implemented for the PostgreSQL event bus.
+
+    Events are typically sent after a change in the database, meaning that there is
+    always an open transaction to commit. For better concurrency handling, it makes
+    more sense to send the event as part of the transaction, i.e using:
+
+        await send_signal(conn, event)
+
+    instead of using another connection such as the notification connection used
+    in `test_send` for testing purposes.
+    """
+
     def __init__(self, conn: AsyncpgConnection):
         super().__init__()
         self._conn = conn
 
     @override
-    async def send(self, event: Event) -> None:
-        """This method is only here for testing purposes.
-
-        Events are typically sent after a change in the database, meaning that there is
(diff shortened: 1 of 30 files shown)


server/parsec/components/postgresql/events.py
```diff
@@ -29,44 +27,32 @@
...
 
 class PGEventBus(EventBus):
-    def __init__(self, send_events_channel: MemoryObjectSendStream[Event]):
+    def __init__(self, conn: AsyncpgConnection):
         super().__init__()
-        self._send_events_channel = send_events_channel
+        self._conn = conn
 
     @override
     async def send(self, event: Event) -> None:
```
touilleMan, line 35: since the method is only for testing, better to name it with a `test_` prefix so that it pops out during code review whenever it is wrongly used.

```suggestion
    async def test_send(self, event: Event) -> None:
```

(I guess you couldn't do it because it was used in the `server/tests` code where `EventBus` can be either memory or postgresql, if so it's probably not worth it to bother with this change ^^)

server/parsec/components/postgresql/events.py
```diff
@@ -77,22 +63,29 @@ def _on_notification(conn: object, pid: int, channel: str, payload: object) -> N
                 "Invalid notif received", pid=pid, channel=channel, payload=payload, exc_info=exc
             )
             return
+        logger.info_with_debug_extra(
+            "Received internal event",
```
touilleMan, line 67:
```suggestion
            "Received event",
```

The event is no longer internal since it comes from postgresql ^^

touilleMan (approved):

server/parsec/components/postgresql/events.py
vxgmichel: 55c12c1d0e384293ef21a8d0a192af44d5650931

PR #9681, opened 2025-02-12
[PoC] Add biometrics authentication method 

vxgmichel:
Port of #6557 

The new authentication method is called `biometrics` because this is how [bitwarden calls it](https://github.com/bitwarden/clients/blob/f5e6fc8ed5a4756d7866ae05118ff795ec68a374/apps/desktop/desktop_native/core/src/biometric/windows.rs).

In practice it's an interface to Windows Hello, and more specifically the [Windows.Security.Credentials.KeyCredential](https://learn.microsoft.com/en-us/uwp/api/windows.security.credentials.keycredentialmanager.issupportedasync?view=winrt-22621#windows-security-credentials-keycredentialmanager-issupportedasync) API.

To quote the docs:
> An application cannot provision a key credential until the user has successfully done the following:
> - Connected their user account to their Microsoft account.
> - Provided an unlock gesture (PIN or biometric) to protect the container that stores their key credential.

That means that the parsec device key file becomes protected by the machine TPM which gets unlocked by either facial recognition, fingerprint recognition, or PIN code.

The `is_biometrics_available` function is exposed to specifically check those requirements.

This PR also adds the `parsec device change-authentication` method for testing purposes. 

Logging using the biometrics access strategy is also supported in the electron app, but it's not yet available when saving a new device. In practice it would make sense to use `biometrics` when available and fall back to `keyring` if necessary, without exposing both to the user.


vxgmichel:
Reviewers should pay attention and address those TODOs:

1. Focusing the windows security prompt:
```rust
        // TODO: This Alt key hack has been recently removed from bitwarden:
        //   https://github.com/bitwarden/clients/pull/12255
        // We should investigate if it's still necessary, or if the approach can be improved.
```

2. Identifying the credential using its public key:
```rust
    // TODO: The approach here is to create a new key credential if it doesn't exist.
    // This allows us to expose only a single `derive_key_from_biometrics` API for
    // both loading and saving.
    // However, this means we cannot detect when an existing key credential has been
    // removed from the system (e.g when the pin code has been removed and recreated).
    // Another approach would be to retrieve the corresponding public key when saving
    // a new device using `KeyCredential::RetrievePublicKey()`. This way, this information
    // could be stored in the device key file and use it to detect when the key credential
    // has been removed or changed.
```

3. Use an async API to not block during the prompt:
```rust
// TODO: this function is synchronous but will block the thread while waiting for the
// user to enter this pin code. Either this function should be async or we make sure
// the caller uses `tokio::task::spawn_blocking`
```

4. Choose a challenge format:
```rust
    // TODO: Is this the formatting we want for the challenge?
    // Is the device ID the right thing to use here?
    let challenge = format!("BIOMETRICS_CHALLENGE:{}", device_id.hex());
```

PR #9746, opened 2025-02-18
feat(device-loader): Extract more agnostic stuff from native impl

FirelightFlagboy:


libparsec/crates/platform_device_loader/src/lib.rs
```diff
@@ -416,3 +472,14 @@ fn secret_key_from_password(
         ),
     }
 }
+
+#[cfg_attr(target_arch = "wasm32", expect(dead_code))]
+fn server_url_from_device(device: &LocalDevice) -> String {
```
mmmarcos, line 477: Now that this is has been extracted we should add some unit tests for the function.

libparsec/crates/platform_device_loader/src/lib.rs
```diff
@@ -396,6 +394,64 @@ fn load_available_device_from_blob(
     })
 }
 
+#[cfg_attr(target_arch = "wasm32", expect(dead_code))]
+fn encrypt_device(device: &LocalDevice, key: &SecretKey) -> Bytes {
+    let mut cleartext = zeroize::Zeroizing::new(device.dump());
+    let ciphertext = key.encrypt(&cleartext);
+    cleartext.zeroize();
```
mmmarcos, line 401: Isn't the point of `Zeroizing` to automatically `zeroize`?

libparsec/crates/platform_device_loader/src/lib.rs
```diff
@@ -396,6 +394,64 @@ fn load_available_device_from_blob(
...
+fn decrypt_device_file(
+    device_file: &DeviceFile,
+    key: &SecretKey,
+) -> Result<LocalDevice, DecryptDeviceFileError> {
+    let mut cleartext = key
+        .decrypt(device_file.ciphertext())
+        .map_err(DecryptDeviceFileError::Decrypt)
+        .map(zeroize::Zeroizing::new)?;
+    let res = LocalDevice::load(&cleartext).map_err(DecryptDeviceFileError::Load);
+    cleartext.zeroize();
```
mmmarcos, line 441: Isn't the point of `Zeroizing` to automatically `zeroize`?

libparsec/crates/platform_device_loader/src/native/mod.rs
```diff
@@ -280,25 +259,11 @@ pub async fn save_device(
...
-            .expect("Salt has the correct length");
-
-            let ciphertext = {
-                let cleartext = Zeroizing::new(device.dump());
-                let ciphertext = key.encrypt(&cleartext);
-                ciphertext.into()
-            };
+            let key_algo = super::new_default_pbkdf_algo();
+            let key = super::secret_key_from_password(password, &key_algo)
+                .expect("Salt has the correct length");
```
mmmarcos, line 264: Message in expect should be for the error case.
```suggestion
                .expect("Salt does not have the correct length");
```
or more generic since errors could come from things other than the salt:
```suggestion
                .expect("Failed to derive key from password");
```

libparsec/crates/platform_device_loader/src/lib.rs
```diff
@@ -396,6 +394,64 @@ fn load_available_device_from_blob(
...
+    res
+}
+
+#[cfg_attr(target_arch = "wasm32", expect(dead_code))]
+fn new_default_pbkdf_algo() -> DeviceFilePasswordAlgorithm {
```
touilleMan, line 446: From what I understand, [PBKDF is the name of a specific algorithm for key derivation](https://en.wikipedia.org/wiki/PBKDF2), so we should choose another name here
```suggestion
fn default_device_file_password_algorithm() -> DeviceFilePasswordAlgorithm {
```

libparsec/crates/platform_device_loader/src/lib.rs
FirelightFlagboy: The line was removed

libparsec/crates/platform_device_loader/src/lib.rs
FirelightFlagboy: The line was removed

libparsec/crates/platform_device_loader/src/lib.rs
FirelightFlagboy: done

libparsec/crates/platform_device_loader/Cargo.toml
```diff
@@ -7,6 +7,7 @@ homepage.workspace = true
 license.workspace = true
 version.workspace = true
 repository.workspace = true
+autotests = false
```
touilleMan, line 10: I'd rather stay consistent since we don't do this with the other crates

Can you instead create a `tests/unit/` folder containing all the tests files (and the `mod.rs` you added)

touilleMan (approved):

Issue #9042, opened 2024-11-29
[MS] (File viewer features) Controls section - Playback speed


NicoTuxx:
see [File viewer specifications](https://github.com/Scille/parsec-cloud/issues/9011)

depends on #9018 #9019 #9020 

Ex: (0.5x, 1.5x, 2x, etc.)

Issue #9045, opened 2024-11-29
[MS] (File viewer features) Controls section - Picture-in-Picture mode


NicoTuxx:
see [File viewer specifications](https://github.com/Scille/parsec-cloud/issues/9011)

depends on #9018 #9019

PR #9750, opened 2025-02-18
[MS] Add loop + picture in picture + playback speed in audio & video viewers

Ironicbay:
Closes https://github.com/Scille/parsec-cloud/issues/9042 and https://github.com/Scille/parsec-cloud/issues/9045 (no issue on `loop`)

## Checklist

- [x] Describe your changes
   
If this PR is related to GUI:

If this PR affects the user:

### DIFF
--- a/client/src/locales/en-US.json
+++ b/client/src/locales/en-US.json
@@ -1855,6 +1855,20 @@
         "text": {
             "loadDocumentError": "Could not load this document.",
             "loadingDocument": "Loading document..."
+        },
+        "players": {
+            "playback": {
+                "title": "Playback speed",
+                "speed": {
+                    "0_25": "x0.25",
+                    "0_5": "x0.5",
+                    "1": "x1",
+                    "1_5": "x1.5",
+                    "2": "x2"
+                }
+            },
+            "pip": "Picture in Picture",
+            "loop": "Loop"
         }
     }
 }
--- a/client/src/views/viewers/AudioViewer.vue
+++ b/client/src/views/viewers/AudioViewer.vue
@@ -37,6 +37,10 @@
           v-model="volume"
           :muted="muted"
         />
+        <file-controls-dropdown
+          :items="dropdownItems"
+          :icon="cog"
+        />
       </file-controls>
     </template>
   </file-viewer-wrapper>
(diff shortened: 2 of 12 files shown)


client/src/locales/en-US.json
```diff
@@ -1811,6 +1811,20 @@
         "text": {
             "loadDocumentError": "Could not load this document.",
             "loadingDocument": "Loading document..."
+        },
+        "players": {
+            "playback": {
+                "title": "Playback speed",
+                "speed": {
+                    "0": "x0.25",
```
Max-7, line 1819: May as well name the key `0_25` or `2`, so we don't have to ask "What's 'playback.speed.0'?" when reading the code.

client/src/views/viewers/AudioViewer.vue
```diff
@@ -114,6 +147,24 @@ function togglePlayback(): void {
   }
 }
 
+async function changePlaybackSpeed(value: number): Promise<void> {
+  const playbackSpeedValues = [0.25, 0.5, 1, 1.5, 2];
```
Max-7, line 151: We don't need an array, you could pass the value directly. And if we want to restrict the possible values, an enum would be more appropriate. Something like `await changePlaybackSpeed(PlaybackSpeed.Speed0_25)`.
I'm guessing you used an array because the zoom does, but for the zoom, a user can enter a value that has to be rounded, it's easier to do with an array. Here the values are fixed.

client/src/views/viewers/VideoViewer.vue
```diff
@@ -102,103 +104,45 @@ const cancelVolumeWatch = watch(
...
+    label: 'fileViewers.players.loop',
+    icon: infinite,
+    callback: toggleLoop,
+    isActive: false,
+  },
+]);
 
 onMounted(async () => {
   src.value = URL.createObjectURL(new Blob([props.contentInfo.data], { type: props.contentInfo.mimeType }));
+  document.addEventListener('enterpictureinpicture', () => onTogglePictureInPicture(true), false);
```
Max-7, line 137: Is there a reason why you add it to the document instead of the video element?

If not, add those directly on the video element,
```
<video
  @enterpictureinpicture="onTogglePictureInPicture(true)"
/>
This way you don't have to remove the event listener, Vue will do it for you.
```

Max-7 (approved):

NicoTuxx (approved):
LGTM

NicoTuxx (approved):

PR #9756, opened 2025-02-19
Fix missing manifest lock guard release in `SyncUpdater::into_sync_conflict_updater` when populate errors out

touilleMan:


libparsec/crates/client/src/workspace/store/per_manifest_update_lock.rs
```diff
@@ -161,8 +154,9 @@ impl PerManifestUpdateLock {
         // We remove the entry altogether, this is the way to signify it is free to take now
         self.per_manifest_lock.swap_remove(index);
 
-        // Finally mark the guard as released to pass the sanity check in it drop
-        guard.is_released = true;
+        // Finally forget about the guard to avoid running its drop (since it
+        // is expected to be called only when the guard is improperly dropped).
+        std::mem::forget(guard);
```
FirelightFlagboy, line 159: **chore (blocking, non-blocking):** This result in a memory leak of `entry_id`

FirelightFlagboy (changes requested):

libparsec/crates/client/src/workspace/store/per_manifest_update_lock.rs
touilleMan: `entry_id` is a `VlobID`, internally only containing a `uuid::Uuid` which is itself just a `[u8;16]` (cf. https://docs.rs/uuid/latest/src/uuid/lib.rs.html#291)

Hence both `VlobID` and `uuid::Uuid` don't implement a drop so we can safely use `std::mem::forget` without having any memory leak

FirelightFlagboy (commented):
Is it possible to have a test that show the issue being resolved?

touilleMan:
> Is it possible to have a test that show the issue being resolved?

There is now a test \o/

FirelightFlagboy (approved):

PR #9757, opened 2025-02-19
feat(rfc): Add RFC Parsec auth

### LINKED ISSUE #8482: RFC Parsec Auth
We should describe what it needs to be done to implement Parsec Auth and its impact on Parsec Server and libparsec:

- Parsec Auth API
  - Account creation (email+password)
  - Password update
  - User authentication + 2FA support
  - Get _device registration key_ (allows to create _device securisation key_, the "classic" Parsec device key)
  - Store _device registration key_ after identity validation (i.e. after invite)
  - List organizations for current authenticated user
  - List invitations for current authenticated user
  - Account recovery (TBD)
    - List Shamir configurations for current authenticated user
    - List devices already registered (per organization)
  - For the (not too distant) future
    - SSO support
    - Passkey/WebAuthn support
    - PKI support

- Parsec Server API (all token-protected routes)
  - List organizations of a user
  - List invitations of a user. Note these are _received_ (not _sent_) invitations 
  - List shamir of a user
 
- Libparsec
  - Support new APIs from Parsec Auth/Server
  - Generation of _device registration key_ to be stored in Parsec Auth 
  - Password derivation (from _device registration key_?)
  - Disable mountpoint
  - Decrypt documents on RAM (for visualization of supported file-types)
  - Download a document (for non-supported file types) 
  - Device storage

Multi-tab browser support (for a different RFC?)

FirelightFlagboy:
Parsec auth would be a new service that store the device of the users securely and remotly to allow them to access those devices anywhere on any machine.

Close #8482

### DIFF
--- a/docs/rfcs/1013-parsec-account.md
+++ b/docs/rfcs/1013-parsec-account.md
@@ -1,0 +1,510 @@
+<!-- Parsec Cloud (https://parsec.cloud) Copyright (c) BUSL-1.1 2016-present Scille SAS -->
+
+# Store Parsec device on a remote server
+
+## Overview
+
+This RFC will discuss the implementation of a new service to store the user's devices remotely while still ensuring that only the user can use them.
+
+> [!NOTE]
+> The creation of Parsec devices is not discussed in this RFC
+> as this is still performed as usual by the client application.
+
+> [!IMPORTANT]
+> Mentions to FIDO2 as an alternative authentication method should not be taken as decision made in stone.
+> We are currently lacking actual experience to be able to correctly specify what is needed.
+> For instance a FIDO2 extension allows generating a symmetric key that could be used as password
+> (so FIDO2 implementation would only be a small layer on top of the password).
+
+## Background & Motivation
+
+During our discussions about providing a client application that could run in a web-browser, came up the following questions:
+
+- How should we store the device in the browser?.
+
+  Some concerns were raised about the persistence of a Parsec device in the browser.
+
+- What would the user experience be like?
+
+  - What should happen when the user uses a different browser?
+  - What is the user expecting when using another computer?
+
+  We assume that the user should be able to connect from any browser/computer.
+
+That is how we came to the idea of storing the device on a remote server.
+
+- That will solve the issue of the browser cleaning its data.
+- The devices would be accessible from anywhere (given the user is online).
+
+## Goals
+
+1. Describe how to store/access the user devices on a remote service.
+2. Propose an API to implement this feature.
+
+## Non-Goals
+
+- Discuss security concerns regarding the web client application.
+
+## Design
+
+### General principles
+
+An authentication service (Parsec Auth) would be used to store user devices. This service will be requested before connecting to the metadata server (Parsec server) in order to retrieve a specific user device. The device will then be used to authenticate to the metadata server.
+
+<!-- Device stored on a third-party service -->
+```mermaid
+sequenceDiagram
+  autonumber
+  participant Client
+  participant S_Auth as Authentification server
+  participant S_Data as Metadata server
+
+  Client ->> S_Auth: Retrieve the device of the user
+  S_Auth ->> Client: Device of the user
+
+  Client ->> S_Data: Exchange with the server
+```
+
+### Account creation
+
+Creating the account for a user will require some information for the system to work:
+
+- Some information about the user:
+
+  - Its username (would be used to pre-fill the label part of the human handle when creating a device)
+  - An identifier for the user (email)
+
+    The service needs to verify that the user has access to the email (using a code or sending a link to continue the process).
+
+- An ID for the authentication method
+
+  That value is generated from the chosen method since it needs to be deterministic from the values available to the user and specific to the method itself.
+  For password, it means the value will likely be derived from itself using `PBKDF_A`
+
+- A HMAC shared key to authenticate the user (`AUTH_METHOD_HMAC_KEY`)
+
+  When using a password, the HMAC key is derived from it using the function `PBKDF_B` (this operation is performed on the client side to avoid leaking the password).
+
+  When using FIDO2, the included challenge is used as is for the HMAC key since the private key is not shared with the server.
+
+- An asymmetric key pair `AUTH_METHOD_PUB_KEY`/`AUTH_METHOD_PRIV_KEY` used to encrypt `ACCOUNT_MANIFEST_SYM_KEY` described below
+
+  When using a password, the asymmetric key pair is generated using a cryptographic secure pseudo-random number generator (CSPRNG).
+
+  When using FIDO2, it already provides the asymmetric keys securely so they are used as is.
+
+- A symmetric key (`AUTH_MEDIUM_SYM_KEY`) used to encrypt `AUTH_MEDIUM_PRIV_KEY` alongside the parameters used to generate it `AUTH_MEDIUM_SYM_KEY_PARAMS`.
+
+  When using a password, it's generated from the function `PBKDF_C`
+
+  For FIDO2, the symmetric key is not needed because `AUTH_METHOD_PRIV_KEY` is not shared since it's stored on the device.
+
+- A randomly generated symmetric key `ACCOUNT_MANIFEST_SYM_KEY` used to encrypt the account manifest.
+
+- Generate the account manifest `ACCOUNT_MANIFEST` to store:
+
+  - The list of authentication public keys (`AUTH_METHOD_PUB_KEY`)
+  - The list of keys used to encrypt the devices (the list is empty at the beginning)
+
+  > The manifest will be encrypted to prevent the server to temper with it.
+
+> [!CAUTION]
+> The output of `PBKDF_{A,B,C}` should be different to avoid leaking the private key.
+
+
+```mermaid
+sequenceDiagram
+  autonumber
+  participant Alice
+  participant S as Authentification server
+
+  Alice ->> S: Provide its email for the account creation
+  S ->> Alice: Send an email with a unique code needed to continue the process
+  alt Choose between password or fido2
+    Alice ->> Alice: Choose a password
+    Alice ->> Alice: Generate `auth_method_id` using the password and `PBKDF_A`
+    Alice ->> Alice: Derive `AUTH_METHOD_HMAC_KEY` using the password and `PBKDF_B`
+    Alice ->> Alice: Create `AUTH_METHOD_SYM_KEY` using the password and `PBKDF_C`
+    Alice ->> Alice: Generate `AUTH_METHOD_PRIV_KEY`, `AUTH_METHOD_PUB_KEY`
+    Alice ->> Alice: Encrypt `AUTH_METHOD_PRIV_KEY` with `AUTH_METHOD_SYM_KEY`
+  else
+    Alice ->> Alice: Generate `auth_method_id` using fido2 device.
+    Alice ->> Alice: Generate `AUTH_METHOD_HMAC_KEY` using the fido2 device.
+    Alice ->> Alice: Use the public key of the FIDO2 device as `AUTH_METHOD_PUB_KEY`
+  end
+
+  Alice ->> Alice: Create a bundle containing `AUTH_METHOD_PUB_KEY` (`ACCOUNT_MANIFEST`)
+  Alice ->> Alice: Create a symmetric key `ACCOUNT_MANIFEST_SYM_KEY`
+  Alice ->> Alice: Encrypt the bundle `ACCOUNT_MANIFEST` with `ACCOUNT_MANIFEST_SYM_KEY` (`enc(ACCOUNT_MANIFEST)`)
+  Alice ->> Alice: Encrypt `ACCOUNT_MANIFEST_SYM_KEY` with `AUTH_METHOD_PUB_KEY` (`enc(ACCOUNT_MANIFEST_SYM_KEY)`)
+  Alice ->> S: Send the require data to finish the account creation
+```
+
+At the end of the process, the server should save the following information:
+
+- An entry for the user account:
+
+  - A UUID (`account_id`)
+  - Username
+
+  - The email of the user
+
+- An entry for the account manifest
+
+  - Identified with the provided `account_id`
+  - A UUID (`account_manifest_id`)
+  - Denoted as the first version of the manifest
+  - The encrypted account manifest (`enc(ACCOUNT_MANIFEST)`)
+
+- An entry for the authentication method linked to the user:
+
+  - An ID that is provided by the user (`auth_method_id`)
+
+    > That require the ID to be generated by the user with information available at the moment (likely derived from its secret).
+
+  - The ID of the related account manifest (`account_manifest_id`)
+
+  - The HMAC secret (`AUTH_METHOD_HMAC_KEY`)
+
+    The secret should be securely stored by either encrypting its value or using a vault.
+
+  - The encrypted private key (`enc(AUTH_METHOD_PRIV_KEY)`)
+
+    > Could be empty when the user uses a FIDO2 device.
+
+  - The encrypted key used to decrypt the account manifest (`enc(ACCOUNT_MANIFEST_SYM_KEY)`)
+  - The parameters used to generate the symmetric key (`AUTH_METHOD_SYM_KEY`) that encrypted `enc(AUTH_METHOD_PRIV_KEY)` (`AUTH_METHOD_SYM_KEY_PARAMS`)
+
+> [!WARNING]
+> A user should be able to register multiple authentication methods.
+
+### The authentication process
+
+We use a HMAC to authenticate each request to the server with the secret `AUTH_METHOD_HMAC_KEY` that was previously shared.
+
+The HMAC code is generated using:
+
+- `auth_method_id` (It will be provided in the header of the request)
+- A timestamp (used as a nonce value to prevent replay attack)
+- The checksum of the body
+
+### Obtain the account manifest from the server
+
+The server stores the encrypted account manifest that contain the important information about the user.
+For the user to obtain the account manifest from the server they will do the following:
+
+1. Authenticate with the server.
+2. Retrieve the following encrypted data from the sever:
+
+   1. Account manifest (`enc(ACCOUNT_MANIFEST)`)
+   2. Account manifest symmetric key (`enc(ACCOUNT_MANIFEST_SYM_KEY)`)
+   3. Authentication method private key (`enc(AUTH_METHOD_PRIV_KEY)`)
+   4. The parameters used to generate the symmetric key that encrypt the private key (`AUTH_METHOD_SYM_KEY_PARAMS`)
+
+   > The retrieve data is related to the used authentication method
+
+3. Re-generate `AUTH_METHOD_SYM_KEY` from the parameters and the password
+4. Decrypt `enc(AUTH_METHOD_PRIV_KEY)` with `AUTH_METHOD_SYM_KEY` to obtain the private key (`AUTH_METHOD_PRIV_KEY`)
+5. Decrypt `enc(ACCOUNT_MANIFEST_SYM_KEY)` with `AUTH_METHOD_PRIV_KEY` (`ACCOUNT_MANIFEST_SYM_KEY`)
+6. Decrypt `enc(ACCOUNT_MANIFEST)` with `ACCOUNT_MANIFEST_SYM_KEY`
+7. Decrypt the account manifest
+
+```mermaid
+sequenceDiagram
+  autonumber
+  participant Alice
+  participant S as  Account server
+
+  Note over Alice,S: Alice is already authenticated to the server
+
+  S ->> Alice: Retrieve `enc(ACCOUNT_MANIFEST)`, `enc(ACCOUNT_MANIFEST_SYM_KEY)`,<br>`enc(AUTH_METHOD_PRIV_KEY)` & `AUTH_METHOD_SYM_KEY_PARAMS`
+  Alice ->> Alice: Generate `AUTH_METHOD_SYM_KEY` using the password and `AUTH_METHOD_SYM_KEY_PARAMS`
+  Alice ->> Alice: Decrypt `enc(AUTH_METHOD_PRIV_KEY)` with `AUTH_METHOD_SYM_KEY`
+  Alice ->> Alice: Decrypt `enc(ACCOUNT_MANIFEST_SYM_KEY)` with `AUTH_METHOD_PRIV_KEY`
+  Alice ->> Alice: Decrypt `enc(ACCOUNT_MANIFEST)` with `ACCOUNT_MANIFEST_SYM_KEY`
+```
+
+### Uploading a new device
+
+To upload a new device, the user would need to:
+
+1. [Obtain the account manifest from the server](#obtain-the-account-manifest-from-the-server).
+2. Create a new symmetric key for encrypting the device (`DEVICE_SYM_KEY`).
+3. Encrypt the new device with `DEVICE_SYM_KEY` (`enc(DEVICE)`).
+4. Add `DEVICE_SYM_KEY` to the account manifest.
+5. Encrypt the account manifest with the symmetric key `ACCOUNT_MANIFEST_SYM_KEY`.
+6. Upload the organization and device ID, the updated encrypted account manifest and the encrypted device.
+7. The server verify the device's email (by asking the metadata server) if it corresponds to the account email
+
+   > Later we could add a list of verified email to allow uploading devices using different emails.
+
+```mermaid
+sequenceDiagram
+  autonumber
+  participant Alice
+  participant S as Authentification server
+
+  Note over Alice,S: Alice is already authenticated to the server
+
+  S ->> Alice: Obtain the account manifest from the server following the procedure
+  Alice ->> Alice: Generate `DEVICE_SYM_KEY`
+  Alice ->> Alice: Encrypt the device `DEVICE` with `DEVICE_SYM_KEY` (`enc(DEVICE)`)
+  Alice ->> Alice: Add `DEVICE_SYM_KEY` to the account manifest
+  Alice ->> S: Upload `org_id` & `dev_id`, `enc(DEVICE)`, `enc(ACCOUNT_MANIFEST)`
+  alt The device email corresponds with the account's email
+    S ->> Alice: The server accept the request, the manifest is stored as a new version
+  else
+    S ->> Alice: The server reject the request
+  end
+```
+
+> [!NOTE]
+> The user can upload multiple devices for the same organization.
+
+### List available devices
+
+Once authenticated, the user can list the devices associated to their account.
+
+The list contains the information about each device like:
+
+- `organization_id`
+- `device_id`
+- `user_id` (information retrieved from the metadata server)
+- `human_handle` (information retrieved from the metadata server)
+- `created_on`
+- `revoked_on` (information retrieved from the metadata server)
+
+The list will not contain the `enc(DEVICE)` because most of the time the user will only need one device (a specific API is proposed for that in the next section).
+
+### Retrieve a specific device
+
+The server provides an API to retrieve the blob `ENC_DEV` of a device for a user and a given organization.
+The client application can then decrypt the blob on its side to obtain the device using the appropriate key store in its account manifest.
+
+### Adding a new authentication method
+
+To add a new authentication method, it requires the client application to:
+
+1. Be already authenticated to the server.
+2. [Obtain the account manifest from the server](#obtain-the-account-manifest-from-the-server).
+3. Generate `AUTH_METHOD_HMAC_KEY'`, `AUTH_METHOD_SYM_KEY'`, `AUTH_METHOD_SYM_KEY_PARAMS'`, `AUTH_METHOD_PRIV_KEY'`, `AUTH_METHOD_PUB_KEY'` for the added authentication method.
+4. Add `AUTH_METHOD_PUB_KEY'` to the account manifest.
+5. Encrypt `AUTH_METHOD_PRIV_KEY'` with `AUTH_METHOD_SYM_KEY'` (`enc(AUTH_METHOD_PRIV_KEY')`).
+6. Encrypt `ACCOUNT_MANIFEST_SYM_KEY` with `AUTH_METHOD_PUB_KEY'` (`enc(ACCOUNT_MANIFEST_SYM_KEY)`).
+7. Upload the updated manifest, `enc(AUTH_METHOD_PRIV_KEY')`, `enc(ACCOUNT_MANIFEST_SYM_KEY)` and `AUTH_METHOD_SYM_KEY_PARAMS'`.
+
+   > The server will store the updated manifest as a new version
+
+```mermaid
+sequenceDiagram
+  autonumber
+  participant Alice
+  participant S as Authentification server
+
+  Note over Alice,S: Alice is already authenticated to the server
+
+  S ->> Alice: Retrieve the account manifest from the server
+  Alice ->> Alice: Generate the required keys for the new authentication method
+  Alice ->> Alice: Add `AUTH_METHOD_PUB_KEY'` to the account manifest
+  Alice ->> Alice: Encrypt `AUTH_METHOD_PRIV_KEY'` with `AUTH_METHOD_SYM_KEY'` (`enc(AUTH_METHOD_PRIV_KEY')`)
+  Alice ->> Alice: Encrypt `ACCOUNT_MANIFEST_SYM_KEY` with `AUTH_METHOD_PUB_KEY'` (`enc(ACCOUNT_MANIFEST_SYM_KEY)`)
+  Alice ->> S: Upload the new/updated data to the server
+```
+
+### Removing an authentication method
+
+To remove an authentication method from a user, it requires the client application to:
+
+1. Be already authenticated to the server.
+2. [Obtain the account manifest from the server](#obtain-the-account-manifest-from-the-server).
+3. Remove the public of the authentication method to remove from the account manifest.
+4. Create a new `ACCOUNT_MANIFEST_SYM_KEY'`.
+5. For each public key `AUTH_METHOD_PUB_KEY` in the account manifest, encrypt `ACCOUNT_MANIFEST_SYM_KEY` with it (`enc(ACCOUNT_MANIFEST_SYM_KEY)`).
+6. Upload the encrypted & updated account manifest and the list of `enc(AUTH_METHOD_SYM_KEY')`.
+
+   > The server will store the updated manifest as a new version
+
+```mermaid
+sequenceDiagram
+  autonumber
+  participant Alice
+  participant S as Authentification server
+
+  Note over Alice,S: Alice is already authenticated to the server
+
+  S ->> Alice: Get the account manifest from the server
+  Alice ->> Alice: Remove the public key of the authentication method from the account manifest
+  Alice ->> Alice: Create a new `ACCOUNT_MANIFEST_SYM_KEY'`
+  loop For each public key in the manifest
+    Alice ->> Alice: Encrypt `ACCOUNT_MANIFEST_SYM_KEY'` with the public key (`enc(ACCOUNT_MANIFEST_SYM_KEY')`)
+  end
+  Alice ->> S: Upload the encrypted & updated account manifest, the list of `enc(ACCOUNT_MANIFEST_SYM_KEY')`<br>and the ID of the authentication method to remove
+```
+
+### Removing the account
+
+If the user wants to remove its account, it requires to just send a request to the server.
+
+```mermaid
+sequenceDiagram
+  autonumber
+  participant Alice
+  participant S as Authentification server
+
+  Note over Alice,S: Alice is already authenticated to the server
+
+  Alice ->> S: Remove the account from the server
+  S ->> Alice: Send email with confirmation link to confirm suppression
+  Alice ->> S: Click on the link confirming the deletion of the account
+```
+
+> [!IMPORTANT]
+> Only account data stored in the authentication server is deleted (see [Account creation](<#Account creation>)), no data is deleted in the metadata server.
+
+### Recovering access to the account
+
+In case the user forget is authentication method, they still could recover access to their account.
+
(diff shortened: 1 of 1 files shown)


FirelightFlagboy:
The file is rendered at https://github.com/Scille/parsec-cloud/blob/8482-rfc-parsec-auth/docs/rfcs/1012-parsec-auth.md

docs/rfcs/1012-parsec-auth.md
```diff
@@ -0,0 +1,285 @@
...
+How would a user would store/access its devices from a remote service.
+What are the API that would be needed to implement this feature.
+
+## Non-Goals
+
+How to ensure that the client is not compromised.
+
+## Design
+
+The authentification service would be a used before connecting to the parsec-cloud server to retrieve the device of the user that will then be used to authenticate to parsec-cloud.
```
AureliaDolo, line 40:
```suggestion
The authentification service would be used before connecting to the parsec-cloud server to retrieve the device of the user that will then be used to authenticate to parsec-cloud.
```

docs/rfcs/1012-parsec-auth.md
```diff
@@ -0,0 +1,285 @@
...
+> That list does not contain the blob `ENC_DEV`.
+
+### Retrieve a device from the service
+
+The server provide an API to retrieve the blob `ENC_DEV` of a device.
+The user can then decrypt the blob on its side.
+
+### Adding a new authentification method
+
+To add a new authentification method, it require the use to:
```
AureliaDolo, line 192:
```suggestion
To add a new authentification method, it require the user to:
```

docs/rfcs/1012-parsec-auth.md
```diff
@@ -0,0 +1,285 @@
...
+  Alice ->> Alice: Create a new asymmetric key pair (`PRIV_AUTH_KEY_B`/`PUB_AUTH_KEY_B`)
+  S ->> Alice: Retrieve the key `enc(SYM_KEY)` related to `PUB_AUTH_KEY`
+  Alice ->> Alice: Decrypt the key `enc(SYM_KEY)` with `PRIV_AUTH_KEY`
+  Alice ->> Alice: Encrypt the key `SYM_KEY` with `PUB_AUTH_KEY_B` (`enc'(SYM_KEY)`)
+  Alice ->> S: Upload `enc'(SYM_KEY)`, `PUB_AUTH_KEY_B` and the secret of the authentification method
+```
+
+### Removing an authentification method
+
+To remove an authentification method from a user, it require the user to:
```
AureliaDolo, line 217:
```suggestion
To remove an authentification method from a user, it requires the user to:
```

docs/rfcs/1012-parsec-auth.md
```diff
@@ -0,0 +1,285 @@
...
+N/A
+
+## Operations
+
+That require managing a new service that will store the devices of the user.
+
+## Security/Privacy/Compliance
+
+Consideration should be taken about the method used to save the device on the remote server.
+The service would need to identify the user so we will need to have some information about the user (email mostly).
```
AureliaDolo, line 277: Should we generate a parsec auth user id  ? That way a user may have the option to change their email.

FirelightFlagboy: Indeed that would be better

docs/rfcs/1012-parsec-auth.md
FirelightFlagboy: done

docs/rfcs/1012-parsec-auth.md
FirelightFlagboy: done

docs/rfcs/1012-parsec-auth.md
FirelightFlagboy: done

docs/rfcs/1012-parsec-auth.md
FirelightFlagboy: done

docs/rfcs/1012-parsec-auth.md
```diff
@@ -0,0 +1,287 @@
...
+  Alice ->> S: Upload the list of encrypted device `ENC_DEV_B`
+  Alice ->> S: Upload the list of key `enc(SYM_KEY_B)`
+  Alice ->> S: Remove the authentication method from the server
+```
+
+### Using the service alongside the parsec-cloud server
+
+The parsec client would use the authentication service to store a special device, who will only be used to create a new local device.
+
+That new device would be stored on the local storage and if the storage happens to be cleaned, the user could still re-create a new device from the special one stored on the service.
```
mmmarcos, line 264: This section needs further details. We should explain a little bit why this "special device" and local devices are needed, why not directly using the special device to login, and _how_ exactly the special device is used to "create a new local device".

Also, we should probably give them more meaningful names. I think we've already called them "device registration key" (the one that is uploaded and allows to register new devices) and "device security key" (the "classic" Parsec device key).

docs/rfcs/1012-parsec-auth.md
```diff
@@ -0,0 +1,303 @@
...
+  end
+
+  Alice ->> Alice: Create the symmetric key `SYM_KEY`
+  Alice ->> Alice: Encrypt the key `SYM_KEY` with `PUB_AUTH_KEY`
+  Alice ->> S: Send the require data to finish the account creation (`enc(SYM_KEY)`, `PUB_AUTH_KEY`, `AUTH_SECRET`, email)
+```
+
+At the end of the process, the server should save the following information in a new entry:
+
+- A UUID (`user_id`)
```
AureliaDolo, line 120: Maybe make explicit that this user_id is not related to any organization's user_id.

docs/rfcs/1012-parsec-auth.md
```diff
@@ -0,0 +1,303 @@
...
+
+> [!WARNING]
+> A user should be able to register multiple authentication methods,
+> so the user entry should allow for multiple `AUTH_SECRET`, `PUB_AUTH_KEY` and `enc(SYM_KEY)`.
+
+### The authentication process
+
+To be able to authenticate the user, it will need to provide the following information:
+
+- Its email
```
AureliaDolo, line 138:
```suggestion
- Their email
```

docs/rfcs/1012-parsec-auth.md
```diff
@@ -0,0 +1,303 @@
...
+
+To upload a new device, the user would need to:
+
+0. Authenticate on the server to obtain the `AUTH_TOKEN` (see previous section).
+1. Retrieve the symmetric key `enc(SYM_KEY)` from the server.
+2. Decrypt the key with `PRIV_AUTH_KEY`.
+3. Encrypt the new device with `SYM_KEY` (`ENC_DEV`).
+4. Upload the encrypted device (`ENC_DEV`) and the ID of the organization (`org_id`) to the server.
+
+The server creates an entry whose primary key is the tuple `user_id` and `org_id`.
```
AureliaDolo, line 169: Just to be sure, it's the auth server user_id ? Not the organization user_id ?

docs/rfcs/1012-parsec-auth.md
FirelightFlagboy: I've added a note to be more specific

docs/rfcs/1012-parsec-auth.md
FirelightFlagboy: I've added a note to clarify that

docs/rfcs/1012-parsec-auth.md
```diff
@@ -0,0 +1,287 @@
...
+That require managing a new service that will store the devices of the user.
+
+## Security/Privacy/Compliance
+
+Consideration should be taken about the method used to save the device on the remote server.
+The service would need to identify the user so we will need to have some information about the user (email mostly).
+
+## Risks
+
+The devices are not stored securely to only allow the user to access them.
```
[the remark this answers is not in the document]
FirelightFlagboy: I've added an explanation

docs/rfcs/1012-parsec-auth.md
```diff
@@ -0,0 +1,287 @@
...
+  For Fido2, the private already exists in the fido2 device so we use it directly.
+
+  > [!NOTE]
+  > Should verify if it's possible to use fido2 to encrypt random blob of data.
+
+- A symmetric key (`SYM_KEY`) that is generated during the account creation at the end by the client.
+  It's encrypted with `PUB_AUTH_KEY` and then uploaded to the service.
+
+> [!CAUTION]
+> The output of `PBKDF_A` and `PBKDF_B` should be different to avoid leaking the private key.
```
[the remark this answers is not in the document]
FirelightFlagboy: That done in the paragraph `### Consideration on PBKDF algorithms`

docs/rfcs/1012-parsec-auth.md
FirelightFlagboy: I've added more details why the device is not used as is

docs/rfcs/1012-parsec-auth.md
```diff
@@ -0,0 +1,351 @@
...
+> [!IMPORTANT]
+> We only delete the account in parsec-auth, no data are deleted in the metadata server.
+
+### Integration with the Parsec server
+
+The parsec client would use the authentication service to store a special device (let's call it remote device), who will only be used to create a new local device.
+That new device would be stored on the local storage and if the storage happens to be cleaned, the user could still re-create a new device from the remote device.
+
+We could directly use the remote device to communicate with the metadata server, but we choose the same approach used by the recovery device where it's only used to create new devices.
+One advantage is it provide a list of connected browsers/computers.
```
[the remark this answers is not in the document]
FirelightFlagboy: > BTW the list of linked browsers

The ability to list the devices for a user (in a orga) is already implemented in the desktop

docs/rfcs/1012-parsec-auth.md
```diff
@@ -0,0 +1,430 @@
...
+  When using a password, it's not used as is, but it's derived to create a HMAC key using the function `PBKDF_A` (this operation is done on the client side to not leak the master password).
+  The new secret will be provided as is to the server (like a password).
+
+  For the FIDO2 case the included challenge is used as is since the private key is not shared with the server.
+
+- An asymmetric key pair (`AUTH_MEDIUM_PUB_KEY`, `AUTH_MEDIUM_PRIV_KEY`)
+
+  They're generated using a cryptographic secure pseudo-random number generator (CSPRNG).
+
+  > This step is not needed when using a fido2 device since, it provides the asymmetric keys securely.
```
AureliaDolo, line 81:
```suggestion
  > This step is not needed when using a fido2 device since it provides the asymmetric keys securely.
```

docs/rfcs/1012-parsec-auth.md
```diff
@@ -0,0 +1,430 @@
...
+  participant Alice
+  participant S as Authentification server
+
+  Alice ->> S: Provide its email for the account creation
+  S ->> Alice: Send an email with a unique code need to continue the process
+  alt Choose between password or fido2
+    Alice ->> Alice: Choose a password
+    Alice ->> Alice: Derive `AUTH_MEDIUM_HMAC_KEY` using the password and `PBKDF_A`
+    Alice ->> Alice: Create `AUTH_MEDIUM_SYM_KEY` using the password and `PBKDF_B`
+    Alice ->> Alice: Generate `AUTH_MEDIUM_PRIV_KEY`, `AUTH_MEDIUM_PUB_KEY`
```
AureliaDolo, line 116: is there a real difference between create and generate ? 

Is it because the symetric key uses an input , and the asymetric pair doesn't ?

docs/rfcs/1012-parsec-auth.md
```diff
@@ -0,0 +1,430 @@
...
+  Alice ->> Alice: Encrypt the bundle `ACCOUNT_MANIFEST` with `ACCOUNT_MANIFEST_SYM_KEY` (`enc(ACCOUNT_MANIFEST)`)
+  Alice ->> Alice: Encrypt `ACCOUNT_MANIFEST_SYM_KEY` with `AUTH_MEDIUM_PUB_KEY` (`enc(ACCOUNT_MANIFEST_SYM_KEY)`)
+  Alice ->> S: Send the require data to finish the account creation (`enc(AUTH_MEDIUM_PRIV_KEY)`, `AUTH_MEDIUM_SYM_KEY_PARAMS`, `enc(ACCOUNT_MANIFEST)`, `enc(ACCOUNT_MANIFEST_SYM_KEY)`, `AUTH_MEDIUM_HMAC_KEY`, email validation token)
+```
+
+At the end of the process, the server should save the following information:
+
+- An entry for the user account:
+
+  - A UUID (`user_id`)
```
AureliaDolo, line 134: nit: maybe it could be renamed to account_id to avoid confusion.

docs/rfcs/1012-parsec-auth.md
```diff
@@ -0,0 +1,430 @@
...
+> A user should be able to register multiple authentication methods,
+
+### The authentication process
+
+We use a HMAC to authenticate each request to the server with the secret `AUTH_MEDIUM_HMAC_KEY` that was previously shared.
+
+### Obtain the account manifest from the server
+
+The server store the encrypted account manifest that contain the important information about the user.
+For the user to obtain the account manifest from the server it will do:
```
AureliaDolo, line 162:
```suggestion
For the user to obtain the account manifest from the server they will do the following:
```

docs/rfcs/1012-parsec-auth.md
```diff
@@ -0,0 +1,430 @@
...
+  Alice ->> Alice: Generate `DEVICE_OWNERSHIP_PROOF`
+  Alice ->> S: Upload `org_id`, `enc(DEVICE)`, `DEVICE_OWNERSHIP_PROOF`, `enc(ACCOUNT_MANIFEST)`
+```
+
+> [!NOTE]
+> The user can upload multiple devices for the same organization.
+
+### List available devices
+
+Once authenticated, the user can list the devices associated to its account.
```
AureliaDolo, line 231:
```suggestion
Once authenticated, the user can list the devices associated to their account.
```

docs/rfcs/1012-parsec-auth.md
```diff
@@ -0,0 +1,430 @@
...
+
+The list contain the information about the device like:
```
AureliaDolo, line 233:
```suggestion
The list contains the information about the device like:
```

docs/rfcs/1012-parsec-auth.md
```diff
@@ -0,0 +1,430 @@
...
+The new local device would be stored on the local storage. If the storage happens to be cleaned, a new device could still be created from the remote device.
+
+We could have directly used the remote device to communicate with the metadata server,
+but this approach (similar to the one used to create devices from a recovery device)
+has the advantage of allowing to provide a list of linked browsers/computers.
+
+For existing devices (created before/without the authentication service) the client would need to create and upload a remote device to the service (cf [Uploading a new device](#uploading-a-new-device)) in order to access Parsec from a web environment.
+
+
+### Global flow diagram
```
AureliaDolo, line 335: :heart:

docs/rfcs/1012-parsec-auth.md
FirelightFlagboy: done

mmmarcos (commented):
There is confusing usage of the terms "authentication method" (well explained in this RFC) and "authentication medium" (mentioned only once but used in all names, e.g. `AUTH_MEDIUM_HMAC_KEY`). Are those different concepts? Otherwise we should stick to "authentication method" and replace all the names accordingly (e.g. `AUTH_METHOD_HMAC_KEY`).

docs/rfcs/1012-parsec-auth.md
```diff
@@ -0,0 +1,432 @@
...
+  - The email of the user
+
+  - The encrypted account manifest (`enc(ACCOUNT_MANIFEST)`)
+
+- An entry for the authentication method linked to the user:
+
+  - A UUID (`auth_medium_id`)
+  - The HMAC secret (`AUTH_MEDIUM_HMAC_KEY`)
+
+    The secret should be securely stored by either encrypting its value or using a vault.
```
touilleMan, line 145: This is not needed since this secret has a similar role to what would have a regular authentication token.

docs/rfcs/1012-parsec-auth.md
```diff
@@ -0,0 +1,432 @@
...
+
+  They're generated using a cryptographic secure pseudo-random number generator (CSPRNG).
+
+  > This step is not needed when using a fido2 device since it provides the asymmetric keys securely.
+
+- A symmetric key used to encrypt `AUTH_MEDIUM_PRIV_KEY` (`AUTH_MEDIUM_SYM_KEY`) alongside its parameters used to generate it `AUTH_MEDIUM_SYM_KEY_PARAMS`.
+
+  When using a password, it's generated from the function `PBKDF_B`
+
+  For FIDO2, we can bypass that step since we do not share `AUTH_MEDIUM_PRIV_KEY` since it's stored on the device.
```
touilleMan, line 87: Regarding FIDO2, I think we are currently lacking actual experience to be able to correctly specify what is needed

For instance we know that we are going to need the FIDO2 extension that allows to generate a symmetric key.
It would be interesting to use this symmetric key as a password for our system, this way the FIDO2 implementation would only be a small layer on top of the password one.

So I think we should either
- focus on building a small POC showcasing how to use FIDO2 to create a symmetric key, doing auth, reseting auth etc.
- or dropping FIDO2 entirely from this RFC, and instead just consider we should support multiple authentication methods all based on a user-provided secret

docs/rfcs/1012-parsec-auth.md
```diff
@@ -0,0 +1,432 @@
...
+
+  - A UUID (`account_id`)
+
+  - The email of the user
+
+  - The encrypted account manifest (`enc(ACCOUNT_MANIFEST)`)
+
+- An entry for the authentication method linked to the user:
+
+  - A UUID (`auth_medium_id`)
```
touilleMan, line 142: I think the auth medium id should be derived from the secret, otherwise there is no way for the client to retrieve it !

docs/rfcs/1012-parsec-auth.md
```diff
@@ -0,0 +1,432 @@
...
+
+We use a HMAC to authenticate each request to the server with the secret `AUTH_MEDIUM_HMAC_KEY` that was previously shared.
+
+### Obtain the account manifest from the server
+
+The server store the encrypted account manifest that contain the important information about the user.
+For the user to obtain the account manifest from the server it will do:
+
+1. Be authenticated with the server.
+2. Retrieve the following encrypted data from the sever:
```
touilleMan, line 167: Also note that since we authenticate the request with a given auth medium id, we should only be able to get back the data related to this specific auth medium.

docs/rfcs/1012-parsec-auth.md
```diff
@@ -0,0 +1,432 @@
...
+To upload a new device, the user would need to:
+
+1. [Obtain the account manifest from the server](#obtain-the-account-manifest-from-the-server).
+2. Create a new symmetric key for encrypting the device (`DEVICE_SYM_KEY`).
+3. Encrypt the new device with `DEVICE_SYM_KEY` (`enc(DEVICE)`).
+4. Add `DEVICE_SYM_KEY` to the account manifest.
+5. Encrypt the account manifest with the symmetric key `ACCOUNT_MANIFEST_SYM_KEY`.
+5. Generate a device ownership proof (`DEVICE_OWNERSHIP_PROOF`):
+
+   The proof is generated by using the signature key of the device to sign the blob `device_id` + `auth_medium_id` currently used to authenticate the request.
```
touilleMan, line 206:
```suggestion
   The proof is generated by using the signature key of the device to sign the blob `device_id` + `account_id` currently used to authenticate the request.
```

It is more logical to use account id since any other auth medium id sharing the same account id will be able to have access to this device

docs/rfcs/1012-parsec-auth.md
```diff
@@ -0,0 +1,432 @@
...
+
+6. Upload the organization, the encrypted device, the device ownership proof, and the updated encrypted account manifest.
```
touilleMan, line 208: what do you mean by "Upload the organization" ?

docs/rfcs/1012-parsec-auth.md
```diff
@@ -0,0 +1,432 @@
...
+## Security/Privacy/Compliance
+
+Consideration should be taken about the method used to save the device on the remote server.
+The service would need to identify the user so we will need to have some information about the user (email mostly).
+
+### Consideration on PBKDF algorithms
+
+The `PBKDF_A` (used to derive a secret from the password) and `PBKDF_B` (used to derive the private key from the password) output should be different in order to avoid leaking the private key.
+
+To make the output different, we could:
```
touilleMan, line 408: Keep things simple: we already use argon2, so that's what we are going to use, with the salt composed of the email + a constant unique for each derivation.

docs/rfcs/1012-parsec-auth.md
```diff
@@ -0,0 +1,432 @@
...
+In [Account creation](<#Account creation>), we use the password to generate 2 things:
+
+- An intermediate secret using `PBKDF_A` provided as is to the server.
+- The key pair using `PBKDF_B`.
+
+We have the section [Consideration of PBKDF algorithms](<#Consideration on PBKDF algorithms>) that state some of the requirement.
+
+## Remarks & open questions
+
+- The client application will need to communicate with the authentication service: should this be integrated into `libparsec`? Or it's the JS side that handle that?
```
touilleMan, line 429: Since we already do client/server communication in `libparsec`, I think it's straightforward to continue doing that ^^

docs/rfcs/1012-parsec-auth.md
```diff
@@ -0,0 +1,432 @@
...
+- What to do when a user is revoked from an organization?
+  Should we remove the device, or should we just provide the information that the device for the given organization is revoked?
```
touilleMan, line 431: I think it is better to just inform  the user is revoked

This way, the question of removing a device stays orthogonal: even if the user is not revoked, we may want to remove the device from parsec auth (for instance if we have by mistake uploaded multiple devices for the same user ?)

docs/rfcs/1012-parsec-auth.md
```diff
@@ -0,0 +1,432 @@
...
+  - The parameters used to generate the symmetric key (`AUTH_MEDIUM_SYM_KEY`) that encrypted `enc(AUTH_MEDIUM_PRIV_KEY)` (`AUTH_MEDIUM_SYM_KEY_PARAMS`)
+
+> [!WARNING]
+> A user should be able to register multiple authentication methods,
+
+### The authentication process
+
+We use a HMAC to authenticate each request to the server with the secret `AUTH_MEDIUM_HMAC_KEY` that was previously shared.
+
+### Obtain the account manifest from the server
```
touilleMan, line 161: Regarding the account manifest, I think we need two things:

1. The account manifest should be versioned, as this would prevent concurrent overwritting
2. We should be able to create multiple independent account manifests
     - any given auth medium would be related to a single account manifest ID
     - any uploaded device could be related to multiple account manifest ID

Regarding 2, this would allow a clean way to recover once the password has been lost:
- the Parsec auth admin generate a password reset link and send it the user
- password reset link is used to create a brand new account manifest and a auth medium using a new password
- once the user is connected, he sees he currently has access to no device (since his account manifest is empty), but that he used to have access to device X/Y/Z that are protected by some old auth medium (typically showing something like `john@example.com on Coolorg, protected by a password configured on 2000-01-01T00:00:00Z`).
- The user can then try to use an old auth medium to get access the the previous auth medium, if he succeed then the content of the old account manifest is merged with the new one.
- Since we store by which account manifest a device is accessible by, we can hide from the recovery screen the devices present in older account manifests once they have been recovered.

docs/rfcs/1012-parsec-auth.md
```diff
@@ -0,0 +1,432 @@
...
+
+- An entry for the user account:
+
+  - A UUID (`account_id`)
+
+  - The email of the user
+
+  - The encrypted account manifest (`enc(ACCOUNT_MANIFEST)`)
+
+- An entry for the authentication method linked to the user:
```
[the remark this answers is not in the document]
touilleMan: > Does it makes sense to add an authentication method (e.g. "password") multiple times? If not, how will the server prevent it?

I think we still want this (typically FIDO2 relies on the fact you end up with a different authentication source on each of the machine you want to connect from)

> If there are multiple authentication methods for an account, and the users wants to authenticate with "password", how will the server know which of the stored authentication methods to use?

The auth medium should have an ID derived from its secret that is send to the server as part of the request Authentication header ([similar to what is done for the authenticated API](https://github.com/Scille/parsec-cloud/blob/7e136748b4051ae47aeb092758e3feb83818cb1b/libparsec/crates/client_connection/src/authenticated_cmds/mod.rs#L357))

docs/rfcs/1012-parsec-auth.md
```diff
@@ -0,0 +1,432 @@
...
+### Integration with the Parsec server
+
+The parsec client would use the authentication service to store a special device (let's call it remote device), which will only be used to create a new local device.
+The new local device would be stored on the local storage. If the storage happens to be cleaned, a new device could still be created from the remote device.
+
+We could have directly used the remote device to communicate with the metadata server,
+but this approach (similar to the one used to create devices from a recovery device)
+has the advantage of allowing to provide a list of linked browsers/computers.
+
+For existing devices (created before/without the authentication service) the client would need to create and upload a remote device to the service (cf [Uploading a new device](#uploading-a-new-device)) in order to access Parsec from a web environment.
```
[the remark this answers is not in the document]
touilleMan: > So the user never uploads the existing local devices (created before auth service)? It's always a new remote device that needs to be created and uploaded?

Yes we want to have dedicated device stored in parsec auth (and maybe later even mark them in their device certificate as only allowed to create other devices).

> BTW, the client will need to create and upload 1 remote device per existing local device? Or is it 1 remote device per organization?

Good point ! I think the client should simply check if the account manifest already contains a valid device.

Since we said that the device in parsec auth is a specific one that is only used for creating new device, then the process of connecting an existing device to a parsec auth account result in the same outcome no matter if the parsec auth account already contained or not a device: in both case we end up with a device in parsec auth that has the same organization ID / user ID than the device on the client, but has a different device ID.

> Currently, it was proposed that the creation (and upload) of a remote device was transparent for the user and happen on login. Should it be instead a manual operation triggered by the user?

I'm not sure what you mean by transparent... when the client does a login on an existing device store on the local machine, there is no parsec auth involved so nothing possible.

On the other hand, if the end-user choose to do parsec auth login on his client,  he will end up on a page listing:
- the devices stored on parsec auth and that no device on the machine has this user ID ==> a "connect" button is displayed for each one of them (and clicking it means we create a new device and store it on the machine)
- the devices stored on parsec auth and some device on the machine has this user ID ==> they are hidden
- the devices stored on the machine and whose user ID is on parsec auth ==> a "connect" button is displayed
- the devices stored on the machine and whose user ID is not on parsec auth (e.g. device created before parsec auth existed) ==> a "store in parsec auth" button[^1] is displayed (and clicking it upload a new device in parsec auth)

[^1]: there is no "connect" button since the device is not parsec of parsec auth, so if the user doesn't want to add it to parsec auth he should connect to it at the step he connected to parsec auth

docs/rfcs/1012-parsec-auth.md
```diff
@@ -0,0 +1,432 @@
...
+  S_Auth -->> Client: Device of the user
+
+  Client  ->> S_Data: Exchange with the server
+```
+
+### Account creation
+
+Creating the account for a user will require some information for the system to work:
+
+- An identifier for the user (email)
```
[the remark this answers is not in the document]
FirelightFlagboy: done

docs/rfcs/1012-parsec-auth.md
```diff
@@ -0,0 +1,432 @@
...
+> The output of `PBKDF_A` and `PBKDF_B` should be different to avoid leaking the private key.
+
+
+```mermaid
+sequenceDiagram
+  autonumber
+  participant Alice
+  participant S as Authentification server
+
+  Alice ->> S: Provide its email for the account creation
```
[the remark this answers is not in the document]
FirelightFlagboy: Could be send at the last step like the other values

docs/rfcs/1012-parsec-auth.md
```diff
@@ -0,0 +1,432 @@
...
+  The service needs to verify that the user has access to the email (using a code or sending a link to continue the process).
+
+- A HMAC shared key to authenticate the user (`AUTH_MEDIUM_HMAC_KEY`)
+
+  When using a password, it's not used as is, but it's derived to create a HMAC key using the function `PBKDF_A` (this operation is done on the client side to not leak the master password).
+  The new secret will be provided as is to the server (like a password).
+
+  For the FIDO2 case the included challenge is used as is since the private key is not shared with the server.
+
+- An asymmetric key pair (`AUTH_MEDIUM_PUB_KEY`, `AUTH_MEDIUM_PRIV_KEY`)
```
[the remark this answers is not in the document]
FirelightFlagboy: done

docs/rfcs/1012-parsec-auth.md
```diff
@@ -0,0 +1,432 @@
...
+  For FIDO2, we can bypass that step since we do not share `AUTH_MEDIUM_PRIV_KEY` since it's stored on the device.
+
+- A randomly generated symmetric key `ACCOUNT_MANIFEST_SYM_KEY`, it will be used to encrypt the account manifest.
+
+- Generate the account manifest `ACCOUNT_MANIFEST`,
+
+  the manifest will store:
+
+  - The list of authentication public keys
+  - The list of keys used to encrypt the devices (the list is empty at the beginning)
```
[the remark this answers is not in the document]
FirelightFlagboy: done

docs/rfcs/1012-parsec-auth.md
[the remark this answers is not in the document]
FirelightFlagboy: included at the last step

docs/rfcs/1012-parsec-auth.md
FirelightFlagboy: The secret need to be store securely.
If we store it in the database and we have a leak someone could impersonate the user

docs/rfcs/1012-parsec-auth.md
FirelightFlagboy: I've added precision to the organization id

docs/rfcs/1012-parsec-auth.md
```diff
@@ -0,0 +1,432 @@
...
+3. Encrypt the new device with `DEVICE_SYM_KEY` (`enc(DEVICE)`).
+4. Add `DEVICE_SYM_KEY` to the account manifest.
+5. Encrypt the account manifest with the symmetric key `ACCOUNT_MANIFEST_SYM_KEY`.
+5. Generate a device ownership proof (`DEVICE_OWNERSHIP_PROOF`):
+
+   The proof is generated by using the signature key of the device to sign the blob `device_id` + `auth_medium_id` currently used to authenticate the request.
+
+6. Upload the organization, the encrypted device, the device ownership proof, and the updated encrypted account manifest.
+
+   > The server will ask the metadata server to verify the device ownership proof since it has access to the verify key of the device.
```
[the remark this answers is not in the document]
FirelightFlagboy: done

docs/rfcs/1012-parsec-auth.md
FirelightFlagboy: The goal was to note the different possibilities.
I've added our choice for using the salt method at the end of the list

docs/rfcs/1012-parsec-auth.md
FirelightFlagboy: I've noted you reply in the RFC

docs/rfcs/1012-parsec-auth.md
FirelightFlagboy: I've noted your reply in the RFC

docs/rfcs/1012-parsec-auth.md
FirelightFlagboy: The proof is only use to validate that the use can use the uploaded device that server not other purpose.
As such it's not save on the server side

docs/rfcs/1012-parsec-auth.md
FirelightFlagboy: I understand that we lake experience concerning the usage of FIDO2 devices for authentication and such.

If I need to choose, it would prefer to drop reference to FIDO2 from the RFC for it to be describe in a future RFC

docs/rfcs/1012-parsec-auth.md
FirelightFlagboy: I've changed the UUID by an ID provided by the client with a note

docs/rfcs/1012-parsec-auth.md
mmmarcos: I agree we lack actual experience but at the same time FIDO2 was already discussed (analyzed?) during the study to support ProConnect for Resana.

So I'm fine with keeping FIDO2 in stand-by for the moment and just consider multiple auth methods based on a user-provided secret. Just make sure that we are not closing the door to support FIDO2.

docs/rfcs/1012-parsec-auth.md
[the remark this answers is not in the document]
mmmarcos: What you call the "connect" button is what we were thinking to happen automatically when the user clicks on the organization (as if a login to org was done). So there are no distinction if the device is on Parsec auth or not: when a user clicks on the organization, Parsec will either "login to org" OR it will "connect + login to org".

This of course could be re-discussed, as front-end development is just about starting.

docs/rfcs/1012-parsec-auth.md
FirelightFlagboy: For confirmation: We do not remove the FIDO2 references from this RFC ? @mmmarcos

docs/rfcs/1012-parsec-auth.md
FirelightFlagboy: I explicitly indicated that the encrypted blob correspond to the current authentication method

docs/rfcs/1012-parsec-auth.md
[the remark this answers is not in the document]
FirelightFlagboy: Following an internal discussion with the front-end team, the behavior is not decided yet but the provided API will not corner us one way or another.

Since there is no actionable action from this discussion I will close it

docs/rfcs/1012-parsec-auth.md
mmmarcos: @FirelightFlagboy I am not sure. I think it will be a pity to lose the information on what's already been analyzed. I guess it depends on how relevant the information is and how much do you think it would be useful in the future.

So choose one of the following: 

1. Keep references to FIDO2 --> Add a note in the Overview section (or non goals section) based on @touilleMan comment:
```
Mentions to FIDO2 as an alternative authentication method should not be 
taken as decision made in stone. We are currently lacking actual experience 
to be able to  correctly specify what is needed. For instance a FIDO2 
extension allows to generate a symmetric key that could be used as password 
(so FIDO2 implementation would only be a small layer on top of the password).
```

2. Remove references to FIDO2 to have a "cleaner" RFC --> Move them somewhere (issue? draft rfc?).

FirelightFlagboy: I've added the note in the overview section

mmmarcos (approved):
LGTM

FirelightFlagboy (commented):
Rename `parsec-auth` to `parsec-account`

docs/rfcs/1012-parsec-auth.md
```diff
@@ -0,0 +1,511 @@
...
+  Alice ->> S: Send the require data to finish the account creation
+```
+
+At the end of the process, the server should save the following information:
+
+- An entry for the user account:
+
+  - A UUID (`account_id`)
+  - User firstname
+  - User lastname
```
FirelightFlagboy, line 149: **todo:** 
```suggestion
  - User username
```

docs/rfcs/1012-parsec-auth.md
```diff
@@ -0,0 +1,511 @@
...
+
+  - The encrypted private key (`enc(AUTH_METHOD_PRIV_KEY)`)
+
+    > Could be empty when the user uses a FIDO2 device.
+
+  - The encrypted key used to decrypt the account manifest (`enc(ACCOUNT_MANIFEST_SYM_KEY)`)
+  - The parameters used to generate the symmetric key (`AUTH_METHOD_SYM_KEY`) that encrypted `enc(AUTH_METHOD_PRIV_KEY)` (`AUTH_METHOD_SYM_KEY_PARAMS`)
+
+> [!WARNING]
+> A user should be able to register multiple authentication methods,
```
FirelightFlagboy, line 180: **todo:** 
```suggestion
> A user should be able to register multiple authentication methods
```

docs/rfcs/1012-parsec-auth.md
```diff
@@ -0,0 +1,511 @@
...
+- The authentication service storing users devices will need to be managed and probably be deployed separately from the Parsec metadata service
+- GDPR will need to be considered for the data stored by the service. T&Cs will need to be adapted in order to mention the service and how to obtain and delete the data stored.
+
+## Security/Privacy/Compliance
+
+Consideration should be taken about the method used to save the device on the remote server.
+The service would need to identify the user so we will need to have some information about the user (email mostly).
+
+### Consideration on PBKDF algorithms
+
```
FirelightFlagboy, line 467: **todo:** Specific sub-derivation principle using `black2`

docs/rfcs/1012-parsec-auth.md
```diff
@@ -0,0 +1,511 @@
...
+```mermaid
+sequenceDiagram
+  autonumber
+  participant Alice
+  participant S as Authentification server
+
+  Note over Alice,S: Alice is already authenticated to the server
+
+  Alice ->> S: Remove the account from the server
```
FirelightFlagboy, line 362: **todo:** 
```suggestion
  Alice ->> S: Remove the account from the server
  S ->> Alice: Send confirmation email with a URL to use to confirm the suppression
  
  Alice ->> S: Use the confirmation link to delete the account
```

FirelightFlagboy:
Superseded by #10090

PR #9765, opened 2025-02-20
[CLI] Support sequester in organization bootstrap.

### LINKED ISSUE #5930: Handle `sequestered organization` in `CLI` during `bootstrap-organization`

AureliaDolo:
Fix #5930

### DIFF
--- a/cli/src/commands/organization/bootstrap.rs
+++ b/cli/src/commands/organization/bootstrap.rs
@@ -1,10 +1,10 @@
 // Parsec Cloud (https://parsec.cloud) Copyright (c) BUSL-1.1 2016-present Scille SAS
 
-use std::sync::Arc;
+use std::{path::PathBuf, sync::Arc};
 
 use libparsec::{
     AvailableDevice, ClientConfig, DeviceLabel, DeviceSaveStrategy, HumanHandle,
-    ParsecOrganizationBootstrapAddr, Password,
+    ParsecOrganizationBootstrapAddr, Password, SequesterVerifyKeyDer,
 };
 
 use crate::utils::*;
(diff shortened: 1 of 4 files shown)


cli/src/commands/organization/bootstrap.rs
```diff
@@ -24,6 +24,9 @@ pub struct Args {
     /// User email
     #[arg(short, long)]
     email: String,
+    /// Sequester authority verify key path
+    #[arg(short, long)]
```
FirelightFlagboy, line 28: **nitpick (non-blocking):** 
I would remove the short option as `-s` don't mean much here (I think this is applicable to a lot of other place)

```suggestion
    /// Sequester authority verify key path
    #[arg(long)]
```

cli/src/commands/organization/bootstrap.rs
```diff
@@ -35,6 +38,7 @@ pub async fn bootstrap_organization_req(
     device_label: DeviceLabel,
     human_handle: HumanHandle,
     password: Password,
+    sequester_key: Option<PathBuf>,
```
FirelightFlagboy, line 41: **chore:** Since the other args are already processed value, I would do the same here

```suggestion
    sequester_key: Option<SequesterVerifyKeyDer>,
```

touilleMan: Yes the key should be passed as a path (obviously it would be a mess to have to provide the key as cleartext on the command line ^^)

and I agree with @FirelightFlagboy , it would be better to deal with the key loading in the `main` function :+1:

FirelightFlagboy (approved):

PR #9776, opened 2025-02-21
Sign CLI & GUI release script + doc

touilleMan:


docs/development/guide-release-parsec.md
FirelightFlagboy, line 1: **chore (blocking):** You need to update the TOC at the top of the file

touilleMan:
not needed, will instead get the changes from PR #9778 when version 3.3 is acknowledge into `master`

Issue #9712, opened 2025-02-14
List all invitations for a given user


mmmarcos:
We want to list the invitations that a specific user as received in parsec-auth.

For that parsec-auth would need to request the server and it would need to perform a lookup in the `invitations` table searching for invitation with `user_invitation_claimer_email=$email && type=USER`:

https://github.com/Scille/parsec-cloud/blob/7739f08858c8c4fd3d8f4a2446c2783eb19c24bd/server/parsec/components/postgresql/migrations/datamodel.sql#L228-L273

mmmarcos:
Discussed again with @touilleMan today: we still think that we should wait until implementing Parsec ~Auth~ Account. For the moment there is no consensus if this should be part of the Administration API or part of another router.

mmmarcos:
Done in #10657

PR #9806, opened 2025-02-24
feat(server): Add function to check if service request is authenticated

FirelightFlagboy:


AureliaDolo (approved):

server/parsec/config.py
```diff
@@ -197,6 +197,9 @@ class BackendConfig:
 
     # Bearer token used to authenticate the administration API
     administration_token: str
+    # Bearer token used to authenticate the service API
+    # If not defined, the API is not accessible
+    service_token: str | None = None
```
touilleMan, line 202: What is this service API ?

If we end up with two unrelated APIs both using a token, we must further document what each of them are for.

FirelightFlagboy:
Reason for Closing: https://github.com/Scille/parsec-cloud/issues/9712#issuecomment-2789578496

PR #9807, opened 2025-02-24
Add route to list organization per users.

### LINKED ISSUE #9710: List all organizations for a given user

AureliaDolo:
Fix #9710 

TODO 
- [ ] update after #9806 
- [ ] tests

Some other fields could be added to know more about how a user is in an organization:
- human name is trivial
- user_id, if they are revoked, or frozen, their initial and current profile are a join with _user table away

Also note that in memory we must iterate on every user of every organization because of the way the datamodel is build. One may want to keep that index aside to improve performance, but the memory implementation is mostly used to test.

### DIFF
--- a/server/parsec/asgi/administration.py
+++ b/server/parsec/asgi/administration.py
@@ -54,7 +54,12 @@ from parsec.components.sequester import (
     SequesterUpdateConfigForServiceStoreBadOutcome,
     WebhookSequesterService,
 )
-from parsec.components.user import UserFreezeUserBadOutcome, UserInfo, UserListUsersBadOutcome
+from parsec.components.user import (
+    UserFreezeUserBadOutcome,
+    UserInfo,
+    UserListOrganizationsBadOutcome,
+    UserListUsersBadOutcome,
+)
 from parsec.events import ActiveUsersLimitField, DateTimeField, OrganizationIDField, UserIDField
 from parsec.logging import get_logger
 from parsec.types import Base64Bytes, SequesterServiceIDField, Unset, UnsetType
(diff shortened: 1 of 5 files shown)


server/parsec/asgi/administration.py
```diff
@@ -755,3 +760,36 @@ async def administration_organization_sequester_service_update_config(
         status_code=200,
         content={},
     )
+
+
+class AdministrationUsersOrganizations(BaseModel):
+    user_email: str
+
+
+@administration_router.get("/administration/users/organizations")
```
touilleMan, line 769: Since this API is about getting back organization ID, it should be

```suggestion
@administration_router.get("/administration/organizations")
```
Used as `GET /administration/organization?human_email=john@example.com`

server/parsec/asgi/administration.py
```diff
@@ -755,3 +760,36 @@ async def administration_organization_sequester_service_update_config(
         status_code=200,
         content={},
     )
+
+
+class AdministrationUsersOrganizations(BaseModel):
+    user_email: str
```
touilleMan, line 766: In Parsec, user refers to a user certificate with a UserID

Here we want to know all the organizations that contains a given email, so we should uses human (as in Parsec human correspond to the email+label couple)



so basically you should replace every occurence of user in this PR by human ^^

server/parsec/components/user.py
```diff
@@ -489,6 +499,11 @@ async def list_frozen_users(
     ) -> list[UserID] | UserListFrozenUsersBadOutcome:
         raise NotImplementedError
 
+    async def list_organizations(
```
touilleMan, line 502: Move this in the organization component since it is about retrieving organizations

server/parsec/components/user.py
```diff
@@ -366,6 +372,10 @@ class UserListFrozenUsersBadOutcome(BadOutcomeEnum):
     AUTHOR_REVOKED = auto()
 
 
+class UserListOrganizationsBadOutcome(BadOutcomeEnum):
```
touilleMan, line 375: I don't think this is needed: all organization not containing the email are simply omitted from the result, so returning an empty list as a result is enough

server/parsec/asgi/administration.py
touilleMan: However if we want to also provide information about the users with the given human handle in each organization, then we probably provide this route with a more specialized name, for instance:

```json5
$ curl https://parsec.cloud/administration/organizations_with_human?email=john@example.com
{
    organizations: [
      {
         organization_id: CoolOrg,
         users: [ // Ordered by created_on
            {
               user_id: b63f033a5a0042faa81be96c9e6f3ee7,
               created_on: "2000-01-02T00:00:00Z",
               current_profile: ADMINISTRATOR,
               frozen: false,
               revoked_on: None,
             },
            {
               user_id: b63f033a5a0042faa81be96c9e6f3ee7,
               created_on: "2000-01-01T00:00:00Z",
               current_profile: STANDARD,
               frozen: false,
               revoked_on: "2000-01-02T00:00:00Z",
             }
         ]
       }
    ]
}
```

touilleMan (changes requested):

Issue #9626, opened 2025-02-11
Improve spacing in Organization configuration column


FirelightFlagboy:
My issue here is the row `User limit ...` should have more spacing before the tag `unlimited`

![Image](https://github.com/user-attachments/assets/19160a8f-cadc-459b-b434-0bf661841088)

PR #9809, opened 2025-02-25
[MS] Add margin for chips in organization user info

Ironicbay:
Closes https://github.com/Scille/parsec-cloud/issues/9626

Capture d’écran 2025-02-25 à 10 41 10

### DIFF
--- a/client/src/views/organizations/OrganizationInformationPage.vue
+++ b/client/src/views/organizations/OrganizationInformationPage.vue
@@ -326,6 +326,7 @@ ion-title {
       justify-content: space-between;
       display: flex;
       align-items: center;
+      gap: 1rem;
     }
   }
 


client/src/views/organizations/OrganizationInformationPage.vue
```diff
@@ -326,6 +326,10 @@ ion-title {
       justify-content: space-between;
       display: flex;
       align-items: center;
```
fabienSvtr, line 328: It is more usual use to add a `gap` instead of `margin`.

fabienSvtr (changes requested):

client/src/views/organizations/OrganizationInformationPage.vue
fabienSvtr: `gap: 1rem;` but on the parent

fabienSvtr (approved):

PR #9832, opened 2025-02-27
feat(rfcs): Add RFC to describe the protocol for parsec-auth

FirelightFlagboy:


AureliaDolo (commented):
It's cumbersome to read but it would be nice to have the same format as in others RFCs (like [here](https://github.com/Scille/parsec-cloud/blob/master/docs/rfcs/1000-shamir-based-recovery.md#1---create-a-shamir-recovery-setup)) aka the json5 format

docs/rfcs/1013-parsec-auth-scheme.md
```diff
@@ -0,0 +1,223 @@
...
+    ErrUnauthenticated,
+    ErrUnknownPublicKey,
+    ErrInternal
+  }
+}
+```
+
+```rust
+mod UploadDevice {
+  #[use_auth_token]
```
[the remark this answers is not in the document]
FirelightFlagboy: This macro is to denote that this request should be authenticated with an auth token

docs/rfcs/1013-parsec-auth-scheme.md
```diff
@@ -0,0 +1,223 @@
...
+  struct UploadDeviceReq {
+    organization_id: OrganizationId,
+    encrypted_device: EncryptedDevice,
```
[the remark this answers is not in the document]
FirelightFlagboy: Nope, it's mostly a specialized `Bytes` type

docs/rfcs/1013-parsec-auth-scheme.md
```diff
@@ -0,0 +1,204 @@
...
+    - status: invalid_credentials
+    - status: internal_error
+```
+
+### Uploading a new device
+
+To upload a new device, the client first need to retrieve the current symmetric key:
+
+```yml
+get_enc_sym_key:
```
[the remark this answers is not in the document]
touilleMan: With this route we plan to have only a single master key ?

The more I think about it the more I'm convinced we should have a master key per encrypted device.
This way each encrypted device is independent, meaning in case a user lose his password he can start all over with a new password and last regain access to the old device if he recall his old password (typically by having a page in the app listing all available encrypted device with a status showing if the device is available with the current key, and a button to try another decryption method).

Another issue with a single master key is we need a way to rotate this key (since getting access to this key means getting access to all the futures encrypted device), which add more complexity.

The drawback of having one master key per encrypted device is that all devices (at least the ones we have access to) must be re-encrypted whenever the password is changed. I guess this is no big deal since a device key file is ~1ko and we are only going to have a couple of them (but if it's an issue we encrypt with the password an intermediary key used to encrypt the device key file while being different for each device).

docs/rfcs/1013-parsec-auth-scheme.md
```diff
@@ -0,0 +1,204 @@
...
+    - status: unknown_public_key
+```
+
+Now that the client can encrypt the device with the symmetric key, it can upload it:
+
+```yml
+upload_device:
+  use_auth_token: true
+  req-fields:
+    organization_id: OrganizationId
```
touilleMan, line 89: Do we want to only being able to store a single device by organization ?

On one hand it makes things simpler, however it also means information can get overwritten by mistake (typically if you join an organization you are already part of)...

On top of that I think it would be good to keep a way to access previous device (typically useful to regain access to data by using the realm export feature).

So I think we should provide here:
- human handle
- user id
- device id
- created on

This way we can list for each organization all the human handles available, and try to login with the most recent device ID of the most recent user ID (since the created on is supposed to be the information from the certificate, we know a more recent user ID with a given human handle means older user ID with the same human handle have been revoked)

docs/rfcs/1013-parsec-auth-scheme.md
```diff
@@ -0,0 +1,204 @@
...
+## Design
+
+### Account creation
+
+To start the account creation, the client start by sending its email.
+
+```yml
+account_creation_step1_email_validation:
+  req-fields:
+   email: Email
```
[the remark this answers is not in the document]
FirelightFlagboy: A specialized string yes

docs/rfcs/1013-parsec-auth-scheme.md
```diff
@@ -0,0 +1,204 @@
...
+On `ok`, the server would have sent a mail with a unique token used for next the request used to register the authentication method:
+
+```yml
+account_creation_step2_password:
+  req-fields:
+    email: Email
+    email_validation_token: EmailValidationToken
+    enc_sym_key: EncSymKey
+    secret: DerivedPassword
+    public_key: PublicKey
```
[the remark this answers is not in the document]
FirelightFlagboy: Creating the account for a user will require some information for the system to work:

[...]

An asymmetric key pair (PUB_AUTH_KEY, PRIV_AUTH_KEY)

When using a password, the private key is generated by using the function PBKDF_B with the password

> A key property here is we only need the password to generate the required keys. 
> Another approach would be to use the password as a symmetric key to encrypt a randomly generated private key that is sent alongside the public key. 
> But we would lose that property.

For FIDO2, the private key already exists in the FIDO2 device so we can use it directly.

> [!NOTE] 
> Should verify if it's possible to use fido2 to encrypt random blob of data.

[Account Creation](https://github.com/Scille/parsec-cloud/blob/8482-rfc-parsec-auth/docs/rfcs/1012-parsec-auth.md#account-creation)

docs/rfcs/1013-parsec-auth-scheme.md
```diff
@@ -0,0 +1,204 @@
...
+To start the account creation, the client start by sending its email.
+
+```yml
+account_creation_step1_email_validation:
+  req-fields:
+   email: Email
+  reps:
+    - status: ok
+    - status: invalid_email
+    - status: email_already_exists
```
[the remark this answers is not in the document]
FirelightFlagboy: Should we return `invalid_email` if the email is already used?

docs/rfcs/1013-parsec-auth-scheme.md
```diff
@@ -0,0 +1,204 @@
...
+
+On `ok`, the server would have sent a mail with a unique token used for next the request used to register the authentication method:
+
+```yml
+account_creation_step2_password:
+  req-fields:
+    email: Email
+    email_validation_token: EmailValidationToken
+    enc_sym_key: EncSymKey
+    secret: DerivedPassword
```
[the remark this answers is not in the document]
FirelightFlagboy: Yes we will change to use the signature mekanism.

This will replace the secret with a verify-key

docs/rfcs/1013-parsec-auth-scheme.md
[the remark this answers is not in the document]
FirelightFlagboy: We return `ok` and ignore the request

docs/rfcs/1013-parsec-auth-scheme.md
```diff
@@ -0,0 +1,204 @@
...
+    - status: invalid_email
+    - status: email_already_exists
+```
+
+On `ok`, the server would have sent a mail with a unique token used for next the request used to register the authentication method:
+
+```yml
+account_creation_step2_password:
+  req-fields:
+    email: Email
```
touilleMan, line 35: `email` was already provided on step 1, we shouldn't re-provide it since it would allow cunning user to bypass email validation by providing a different email at this step...

docs/rfcs/1013-parsec-auth-scheme.md
```diff
@@ -0,0 +1,267 @@
...
+  "cmd": "account_creation_step2_registration",
+  "req": {
+    "email_validation_token": "EmailValidationToken",
+    "account_manifest": "EncryptedAccountManifest",
+    "auth_medium": "AuthMedium"
+  },
+  "reps": [
+    {
+      "status": "ok",
+      "auth_medium_id": "AuthMediumID"
```
touilleMan, line 50: tl;dr: remove this field

Long version:

auth medium ID should either be defined client-side or be returned by the server

Since we store the auth medium ID in the encrypted `AccountManifest`, we have no choice but to have have auth medium ID be defined client-side. And so returning auth medium ID to the client is useless.

docs/rfcs/1013-parsec-auth-scheme.md
```diff
@@ -0,0 +1,267 @@
...
+    },
+    {
+      "status": "invalid_email_validation_token"
+    }
+  ],
+  "custom_types": {
+    "AuthMedium": {
+      "id": "AuthMediumID",
+      "account_manifest_encrypted_sym_key": "EncryptedSymKey",
```
touilleMan, line 59: We need a comment for each key field here explaining what they are needed for
(typically that `hmac_key` is provided in cleartext given it's a shared secret between the client and the server in order to do authenticate the client requests)

docs/rfcs/1013-parsec-auth-scheme.md
```diff
@@ -0,0 +1,267 @@
...
+```
+
+`EncryptedAccountManifest`, `EncryptedSymKey`, `HMACKey`, `EncryptedPrivKey` are specialized `Bytes` types.
+
+`EncryptedAccountManifest` is generated by encrypting the following rust structure:
+
+```rust
+struct AccountManifest {
+  /// List of registered auth medium for the account.
+  auth_mediums: HashMap<AuthMediumID, PublicKey>,
```
touilleMan, line 75:
```suggestion
  // The account manifest is itself encrypted for each auth medium,
  // hence we need those keys to be able to re-encrypt it every time
  // it is modified (i.e. when adding/removing a device/auth medium).
  per_medium_pub_key: HashMap<AuthMediumID, PublicKey>,
```

docs/rfcs/1013-parsec-auth-scheme.md
```diff
@@ -0,0 +1,267 @@
...
+      "status": "ok",
+      "account_manifest": "EncryptedAccountManifest",
+      "encrypted_priv_key_key_algorithm": "KeyAlgorithm",
+      "encrypted_priv_key": "EncryptedPrivKey"
+    }
+  ]
+}
+```
+
+And now request the account manifest:
```
touilleMan, line 102: `auth_medium_get` already contains a `account_manifest` field in its response

Besides, I'd rather avoid avoid having another route if possible to lower the overall complexity

docs/rfcs/1013-parsec-auth-scheme.md
```diff
@@ -0,0 +1,267 @@
...
+}
+```
+
+### Deleting the account
+
+To delete the account, the client needs to be authenticated and provide the derived password:
+
+```yml
+{
+  "cmd": "delete_account",
```
touilleMan, line 259:
```suggestion
  "cmd": "account_delete",
```

docs/rfcs/1013-parsec-auth-scheme.md
```diff
@@ -0,0 +1,267 @@
...
+
+### Obtain the account manifest
+
+Before obtaining the encrypted account manifest, the user retrieves the information related to its authentication medium:
+
+```yml
+{
+  "cmd": "auth_medium_get",
+  "req": {
+    "id": "AuthMediumID"
```
touilleMan, line 89: How the client knows the auth medium ID that should be provided here ?

docs/rfcs/1013-parsec-auth-scheme.md
```diff
@@ -0,0 +1,267 @@
...
+  }
+}
+```
+
+`EncryptedAccountManifest`, `EncryptedSymKey`, `HMACKey`, `EncryptedPrivKey` are specialized `Bytes` types.
+
+`EncryptedAccountManifest` is generated by encrypting the following rust structure:
+
+```rust
+struct AccountManifest {
```
touilleMan, line 73: This structure is also a schema, so should be described with the json5 format

docs/rfcs/1013-parsec-auth-scheme.md
```diff
@@ -0,0 +1,267 @@
...
+  /// List of registered auth medium for the account.
+  auth_mediums: HashMap<AuthMediumID, PublicKey>,
+  /// List of devices registered for the account.
+  device_keys: HashMap<(OrganizationID, DeviceID), SymmetricKey>,
```
touilleMan, line 77:
```suggestion
  // Each device is stored on the server encrypted with it own symmetric key.
  // Note we call this key "decryption" even if it's a symmetric one since the device
  // never gets re-encrypted and hence the key is only used to decrypt it.
  per_device_decryption_keys: HashMap<(OrganizationID, DeviceID), SymmetricKey>,
```

docs/rfcs/1013-parsec-auth-scheme.md
```diff
@@ -0,0 +1,267 @@
...
+  ]
+}
+```
+
+> [!NOTE]
+> The user can upload multiple devices for the same organization.
+
+### List available devices
+
+To list the devices registered in the service, the client only needs to be authenticated:
```
touilleMan, line 149: This route is no longer useful since those data should be store in the account manifest

docs/rfcs/1013-parsec-auth-scheme.md
```diff
@@ -0,0 +1,267 @@
...
+
+### Uploading a new device
+
+Now that the client can encrypt the device with the symmetric key, it can upload it:
+
+```yml
+{
+  "cmd": "device_upload",
+  "req": {
+    "account_manifest": "EncryptedAccountManifest",
```
touilleMan, line 125: we probably need a versioning on the account manifest, otherwise we open ourselves to concurrent overwritting here

docs/rfcs/1013-parsec-auth-scheme.md
```diff
@@ -0,0 +1,267 @@
...
+
+## Design
+
+### Account creation
+
+To start the account creation, the client start by sending its email.
+
+```yml
+{
+  "cmd": "account_creation_step1_email_validation",
```
mmmarcos, line 22: Most of existing commands use the `<object>_<verb>` syntax  (`realm_create`, `vlob_create`, `user_revoke`, etc.) so I would suggest using `account_create`. Also, including a "step number" was used in the invite protocol and later removed.

So what about something like:
```suggestion
  "cmd": "account_create_get_token",
```
```suggestion
  "cmd": "account_create_get_validation_token",
```
```suggestion
  "cmd": "account_create_get_email_validation_token",
```

docs/rfcs/1013-parsec-auth-scheme.md
```diff
@@ -0,0 +1,267 @@
...
+    "account_manifest": "EncryptedAccountManifest",
+    "auth_medium": "AuthMedium"
+  },
+  "reps": [
+    {
+      "status": "ok",
+      "auth_medium_id": "AuthMediumID"
+    },
+    {
+      "status": "invalid_email_validation_token"
```
mmmarcos, line 53: Besides email validation, are there other checks performed server-side? (like the integrity of the `account_manifest` or the `auth_medium`?). If so, we will probably have specific response status.

docs/rfcs/1013-parsec-auth-scheme.md
mmmarcos: I guess a new route is needed to "list" the registered auth ~mediums~ methods. That would seem consistent with the proposed routes to "add" and "remove" an auth method.

BTW, based on [this comment](https://github.com/Scille/parsec-cloud/pull/9832/files#r1997741237), I would rename this route to be about the account manifest (and not the auth medium), so i suggest:

- `account_manifest_get` (this route, to get the account manifest)
- `auth_method_add` (already described below)
- `auth_method_remove` (already described below)
- `auth_method_list` (route to be added below, allows to get the auth ~medium~ method ID )

docs/rfcs/1013-parsec-auth-scheme.md
```diff
@@ -0,0 +1,267 @@
...
+{
+  "cmd": "auth_medium_get",
+  "req": {
+    "id": "AuthMediumID"
+  },
+  "reps": [
+    {
+      "status": "ok",
+      "account_manifest": "EncryptedAccountManifest",
+      "encrypted_priv_key_key_algorithm": "KeyAlgorithm",
```
mmmarcos, line 95: I know is a key algorithm but maybe we can remove the duplicate word:
```suggestion
      "encrypted_priv_key_algorithm": "KeyAlgorithm",
```

docs/rfcs/1013-parsec-auth-scheme.md
mmmarcos: I agree. See my comment above.

docs/rfcs/1013-parsec-auth-scheme.md
```diff
@@ -0,0 +1,267 @@
+<!-- Parsec Cloud (https://parsec.cloud) Copyright (c) BUSL-1.1 2016-present Scille SAS -->
+
+# Schema for parsec-auth
+
+## Overview
+
+This RFC defines the schema used for communicating with the parsec-auth service.
```
mmmarcos, line 7:
```suggestion
This RFC defines the schema used for communicating with the parsec-auth service 
described in [RFC 1012](https://github.com/Scille/parsec-cloud/blob/rfc-parsec-auth-schema/docs/rfcs/1012-parsec-auth.md).
```

docs/rfcs/1013-parsec-auth-scheme.md
```diff
@@ -0,0 +1,267 @@
...
+
+## Goals and Non-Goals
+
+The goal is to define the required schema used when communicating with the parsec-auth service.
+For simplicity, we will skip the fido2 part that will be defined in a later RFC.
+
+## Design
+
```
mmmarcos, line 15: It would be nice to add a summary of the commands proposed in this RFC.

```suggestion

### Summary

This RFC proposes to add the commands below to manage _accounts_,
_authentication methods_ and _devices_. All these commands would 
be added to a new command family named `account_cmds`.

For the account:
- `account_create_get_token`: get a verification token for a given email
- `account_create`: create an account for a given email 
- `account_manifest_get`: get the account manifest for a given authentication method.
- `account_delete`: remove the account and all the data associated to it

For the authentication methods:
- `auth_method_add`: add an authentication method
- `auth_method_remove`: remove an existing authentication method
- `auth_method_list`: list registered authentication methods

For the devices managed by the service:
- `device_upload`: upload a new device for an organization
- `device_list`: list all registered devices
- `device_get`: get the encrypted device for a given organization and device ID

```

docs/rfcs/1013-parsec-auth-scheme.md
mmmarcos: All the encrypted devices would be stored in the account manifest? So is the client that directly stores that in the manifest?

If that's the case there is no need to "list" or "get" devices, but there is also no need to "upload" them either??

docs/rfcs/1013-parsec-auth-scheme.md
```diff
@@ -0,0 +1,267 @@
...
+}
+```
+
+### Adding a new authentication method
+
+To add a new authentication method, the client just needs to provide the required information:
+
+```yml
+{
+  "cmd": "auth_medium_new",
```
mmmarcos, line 203:
```suggestion
  "cmd": "auth_method_add",
```
or to be consistent with existing commands (`realm_create`, `user_create`, etc.)
```suggestion
  "cmd": "auth_method_create",
```

docs/rfcs/1013-parsec-auth-scheme.md
```diff
@@ -0,0 +1,267 @@
...
+
+> `delete_auth_medium_id` is optional and can be used to delete the authentication method identified by the ID.
+
+### Removing an authentication method
+
+To remove an authentication method
+
+```yml
+{
+  "cmd": "auth_medium_delete",
```
mmmarcos, line 235: I'd use remove to be consistent with the section title (or change the title)
```suggestion
  "cmd": "auth_method_remove",
```

docs/rfcs/1013-parsec-auth-scheme.md
FirelightFlagboy: done

docs/rfcs/1013-parsec-auth-scheme.md
FirelightFlagboy: done

docs/rfcs/1013-parsec-auth-scheme.md
FirelightFlagboy: done

docs/rfcs/1013-parsec-auth-scheme.md
```diff
@@ -0,0 +1,267 @@
...
+  "cmd": "account_creation_step1_email_validation",
+  "req": {
+    "email": "Email"
+  },
+  "reps": [
+    {
+      "status": "ok"
+    },
+    {
+      "status": "invalid_email"
```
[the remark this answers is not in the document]
FirelightFlagboy: We will send the OK response to not create an oracle about whose email is registered in the service

docs/rfcs/1013-parsec-auth-scheme.md
```diff
@@ -0,0 +1,267 @@
...
+    }
+  ]
+}
+```
+
+On `ok`, the server would have sent a mail with a unique token used for next the request used to register the authentication method:
```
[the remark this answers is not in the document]
FirelightFlagboy: I imagine yes, the token will expire after some time

docs/rfcs/1013-parsec-auth-scheme.md
FirelightFlagboy: No other checks: data being encrypted mean the server cannot read them

docs/rfcs/1013-parsec-auth-scheme.md
FirelightFlagboy: `auth_method_list` could not be implemented since it will be an authenticated route whose require `AuthMediumID` to generate the HMAC signature

FirelightFlagboy: It was stated in the prior RFC that `AuthMediumID` is generated from the secret

docs/rfcs/1013-parsec-auth-scheme.md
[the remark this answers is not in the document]
mmmarcos: So maybe we should clarify that in the paragraph below (``On `ok` the server would have sent a mail...``) by adding something like:

```
If the email is already registered in the service, an `ok` response will still 
be sent (without token creation) to avoid creating an oracle about emails 
registered in the service.
```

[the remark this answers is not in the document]
FirelightFlagboy: done

docs/rfcs/1013-parsec-auth-scheme.md
FirelightFlagboy: We still need the list for the `revoked_on` field, current the manifest only store symmetric key to decrypt the devices.

Should we change that?

docs/rfcs/1013-parsec-auth-scheme.md
mmmarcos: Or
```suggestion
  "cmd": "account_send_email_validation_token",
```

docs/rfcs/1013-parsec-auth-scheme.md
```diff
@@ -0,0 +1,350 @@
...
+}
+```
+
+### Adding a new authentication method
+
+To add a new authentication method, the client just needs to provide the required information:
+
+```yml
+{
+  "cmd": "auth_method_new",
```
FirelightFlagboy, line 287: **polish:** 
```suggestion
  "cmd": "auth_method_create",
```

docs/rfcs/1013-parsec-auth-scheme.md
mmmarcos: Actually, we've already used `delete`, so let's stick with that.

docs/rfcs/1013-parsec-auth-scheme.md
FirelightFlagboy: This field no longer exist

FirelightFlagboy:
Superseded by #10090

Issue #9607, opened 2025-02-11
Billing details page in customer area has multiple problems


Max-7:
- [x] The update button can be clicked even if all fields are empty (maybe what we want?)
- [x] The `Update` button can be spammed
- [ ] No auto-complete for adress

![Image](https://github.com/user-attachments/assets/2d9b9c8e-ae4c-427e-956d-e868a7e3eabb)

PR #9852, opened 2025-03-03
[MS] Fix billing page issues

Ironicbay:
Closes https://github.com/Scille/parsec-cloud/issues/9607


client/src/views/client-area/billing-details/BillingDetailsPage.vue
```diff
@@ -135,6 +164,13 @@ async function submit(): Promise<void> {
       PresentationMode.Toast,
     );
   }
+  updateCurrentAddress();
+  isEditing.value = false;
```
Max-7, line 168: Should disable right at the start to prevent multiple click.

Max-7 (approved):

PR #9863, opened 2025-03-05
Remove explicit dependency on base64 & base32 crates from `Cargo.toml`

touilleMan:
We use `data-encoding` crate instead for this


FirelightFlagboy (commented):
Before accepting, can you confirm that `Cargo.lock` does not need to be updated?

touilleMan:
> Before accepting, can you confirm that Cargo.lock does not need to be updated?

I confirm: `base64` crate is still used by some 3rd party dependencies (e.g. `reqwest`) and `base32` was not used at all (it was used for the Parsec v2 file sharing URL, which was modified in Parsec v3)

FirelightFlagboy (approved):

PR #9867, opened 2025-03-05
Fix libparsec client tests web compat

touilleMan:


libparsec/crates/client/tests/unit/workspace_history/fd_close.rs
```diff
@@ -18,9 +18,13 @@ async fn ok(
...
     let wksp1_bar_txt_id: VlobID = *env.template.get_stuff("wksp1_bar_txt_id");
     let wksp1_v2_timestamp: DateTime = *env.template.get_stuff("wksp1_v2_timestamp");
-    let ops = strategy
+    let ops = match strategy
         .start_workspace_history_ops_at(env, wksp1_v2_timestamp)
-        .await;
+        .await
+    {
+        Ok(ops) => ops,
+        Err(StartWorkspaceHistoryOpsError::UnsupportedDataAccessStrategy) => return,
```
FirelightFlagboy, line 26: **question:** Why do we stop the test here?

libparsec/crates/client/tests/unit/workspace_history/utils.rs
```diff
@@ -176,13 +180,19 @@ impl std::ops::DerefMut for WorkspaceHistoryOpsWithMaybeTmpPath {
     }
 }
 
+#[derive(Debug)]
+pub enum StartWorkspaceHistoryOpsError {
+    #[cfg_attr(not(target_arch = "wasm32"), allow(dead_code))]
```
FirelightFlagboy, line 185: **nitpick:** 
```suggestion
    #[cfg_attr(not(target_arch = "wasm32"), expect(dead_code))]
```

touilleMan: Great point ! I didn't know that :+1:

libparsec/crates/client/tests/unit/workspace_history/fd_close.rs
touilleMan: This is because the realm export flavor of workspace history is not compatible with web

In theory it would have been cleaner to express this by sprinkling `#[cfg(target_arch = "wasm32")]` to the parametrize of the test:
```rust
    #[cfg_attr(not(target_arch = "wasm32"), values(DataAccessStrategy::Server, DataAccessStrategy::RealmExport))]
    #[cfg_attr(target_arch = "wasm32", values(DataAccessStrategy::Server))]
    strategy: DataAccessStrategy,
```

However this doesn't work (I suspect the rstest macro is not able to detect the `values`  param when it is enclosed in `cfg_attr`)

I'm going to rename the error `StartWorkspaceHistoryOpsError::RealmExportNotSupportedOnWeb` to make the reason more explicit

FirelightFlagboy (approved):

PR #9887, opened 2025-03-07
feat(serialization_format): Move `cmd` field into protocol object

FirelightFlagboy:
The `cmd` field was present in `req` but it's used to define both the request and response.
So I've decided to put the field at the same level of `req` & `reps`


touilleMan (commented):
I don't see changes for  `misc/gen_protocol_typings.py`,  have you checked it still works ?

(also you should ensure we didn't miss any other script, typically by doing a `grep -R json5 .`)

touilleMan (approved):

PR #9899, opened 2025-03-10
[MS] Adds a splashscreen to electron

Max-7:
- [X] Describe your changes


client/electron/assets/splash.html
```diff
@@ -0,0 +1,27 @@
...
+
+      img {
+        display: block;
+        overflow-y: none;
+        overflow-x: none;
+      }
+    </style>
+  </head>
+  <body>
+    <img
```
FirelightFlagboy, line 23: **question (non-blocking):** Can we provide the image as a proper file over including it as base64?

client/electron/src/splashscreen.ts
```diff
@@ -0,0 +1,65 @@
...
+  splash: BrowserWindow;
+  loaded: boolean = false;
+
+  constructor(from: BrowserWindow) {
+    const x = from.getPosition()[0] + (from.getSize()[0] / 2 - 624 / 2);
+    const y = from.getPosition()[1] + (from.getSize()[1] / 2 - 424 / 2);
+
+    this.splash = new BrowserWindow({
+      width: 624,
+      height: 424,
```
FirelightFlagboy, line 16: **question (non-blocking):** Should the sizes inherited from `from`?

Max-7:
Can be done with capacitor-electron, no need for this PR anymore.

PR #9910, opened 2025-03-11
feat(mountpoint): Check unix metadata

FirelightFlagboy:


Cargo.lock
```diff
@@ -2236,6 +2236,7 @@ dependencies = [
  "libparsec_tests_lite",
  "libparsec_types",
  "log",
+ "nix 0.29.0",
```
touilleMan, line 2239: I'd rather use a unsafe with [stdlib's `getuid`/`getgid`](https://docs.rs/libc/latest/libc/fn.getuid.html) instead of adding yet another dependency

touilleMan (changes requested):

Cargo.lock
FirelightFlagboy: done

mmmarcos (approved):
LGTM

touilleMan (approved):

PR #9916, opened 2025-03-11
fix(rust-async): Fix long sleep on web

FirelightFlagboy:


libparsec/crates/platform_async/src/web/mod.rs
```diff
@@ -9,12 +9,24 @@ pub use pretend_future::{pretend_future_is_send_on_web, pretend_stream_is_send_o
...
+    // Prevent panic when duration is > u32::MAX milliseconds
+    // In `gloo_timers::future::sleep`
+    const MAX_DURATION: std::time::Duration = std::time::Duration::from_millis(u32::MAX as u64);
+    while duration > MAX_DURATION {
+        duration -= MAX_DURATION;
+        let not_send_future = gloo_timers::future::sleep(MAX_DURATION);
+        pretend_future_is_send_on_web(not_send_future).await
+    }
+
+    if duration > std::time::Duration::ZERO {
```
FirelightFlagboy, line 26: **note:** I've choose that `sleep(0)` does not sleep, is that OK

FirelightFlagboy: We should sleep when `duration == 0` to be able to yield the future.
This behavior is expected from the `yield` implementation just below the `sleep` function.

touilleMan (changes requested):
why not using commit e4c793aba6bdf87112a120f66815028aa99d46c4 ?

(besides using u32::MAX is doesn't work, see explanation in the commit)

FirelightFlagboy:
Because I didn't see that commit

touilleMan (approved):

PR #9921, opened 2025-03-12
[MS] Update capacitor community

Max-7:
- [X] Describe your changes


client/package.json
```diff
@@ -33,7 +33,7 @@
         "megashark:local:stop": "npm unlink megashark-lib --no-save && npm install"
     },
     "dependencies": {
-        "@capacitor-community/electron": "^5.0.1",
+        "@capacitor-community/electron": "github:Max-7/capacitor-electron",
```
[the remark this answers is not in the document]
Max-7: I think it's because of the scripts used to launch electron, but it can probably be removed. I'll check later.

client/.gitignore
```diff
@@ -35,6 +35,7 @@ npm-debug.log*
 /electron/build/
 /electron/dist/
 /electron/out/
+/electron/electron/
```
FirelightFlagboy, line 38: Curious about why this line was added, can you explain it to me?

FirelightFlagboy (approved):

client/.gitignore
Max-7: It's a result of removing capacitor-electron from client/.  The commands were previously ran from `client/`, now we do `cd electron` before. When doing an electron:open, it places some stuff inside `electron/electron`. I tried playing with CAPACITOR_ROOT_DIR (https://github.com/Scille/capacitor-electron/blob/19e2c79f354c385c98a1dd1507ea248f6c91e9f2/src/cli-scripts/open.ts#L8) with no success. And since it doesn't seem to change anything...

Ironicbay (approved):

PR #9925, opened 2025-03-12
Sanitize workspace name before using it to determine mountpoint path on Windows

touilleMan:


mmmarcos (approved):
LGTM

Landeers (commented):
Live test on Windows OK

PR #9926, opened 2025-03-12
Drop IndexedDb before testbed init on web

touilleMan:


libparsec/crates/platform_storage/src/web/model.rs
```diff
@@ -3,13 +3,73 @@
...
+use std::path::Path;
 
 use libparsec_types::prelude::*;
 
 use crate::certificates::{FilterKind, GetCertificateQuery};
 use crate::workspace::{RawEncryptedChunk, RawEncryptedManifest};
 use crate::PREVENT_SYNC_PATTERN_EMPTY_PATTERN;
 
+pub(super) fn get_certificates_storage_db_name(
+    #[cfg_attr(not(feature = "test-with-testbed"), allow(unused_variables))] data_base_dir: &Path,
```
FirelightFlagboy, line 15: **question:** Why don't we use the `data_base_dir` value anyway? 

We already does that when loading the devices and that would simplify the functions even if the value is an empty path

libparsec/crates/platform_storage/src/testbed.rs
```diff
@@ -53,134 +64,145 @@ pub(crate) async fn maybe_populate_certificate_storage(data_base_dir: &Path, dev
...
-                        AnyArcCertificate::RevokedUser(certif) => update
-                            .add_certificate(certif.deref(), encrypted)
-                            .await
-                            .unwrap(),
+        #[cfg(target_arch = "wasm32")]
+        drop_existing_web_indexed_db(&format!(
+            "{}-{}-certificates",
+            data_base_dir.display(),
+            device.device_id.hex()
+        ))
```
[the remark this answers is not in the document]
touilleMan: That was my plan initially, however it is combersome to do since it requires exposing `drop_existing_web_indexed_db` (so we must modify `drop_existing_web_indexed_db` visibility and have `web/mod.rs` exposing it)

So I considered it wasn't worth the added complexity since the worst we risk here is to have `drop_existing_web_indexed_db` called on the wrong database name, in such case we will be informed something is wrong pretty fast since running the tests simply fail (given the tested database won't be cleaned)

libparsec/crates/platform_storage/src/web/model.rs
touilleMan: I just copied existing code here, but I totally agree with your point :+1: 

I'm going to change this then

libparsec/crates/platform_storage/src/testbed.rs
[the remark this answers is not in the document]
FirelightFlagboy: Why not exposing `get_certificates_storage_db_name` instead?

libparsec/crates/platform_storage/src/web/model.rs
```diff
@@ -3,13 +3,46 @@
...
+        "{}-{}-certificates",
+        data_base_dir.display(),
+        device_id.hex()
+    )
+}
+
+pub(super) fn get_user_storage_db_name(data_base_dir: &Path, device_id: DeviceID) -> String {
+    format!(
+        "{}-{}-user",
+        data_base_dir.to_str().unwrap(),
```
FirelightFlagboy, line 28: **chore (blocking):** The other function use `Path::display()`

```suggestion
        data_base_dir.display(),
```

libparsec/crates/platform_storage/src/testbed.rs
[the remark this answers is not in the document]
touilleMan: sorry I mixed `drop_existing_web_indexed_db` with `get_certificates_storage_db_name` in my previous message

(so I'm not exposing `get_certificates_storage_db_name` since it is cumbersome to change its visibility, and the benefit is low given if the wrong value is used the tests are going to fail anyway)

libparsec/crates/platform_storage/src/web/model.rs
touilleMan: :+1:

FirelightFlagboy (approved):

FirelightFlagboy (dismissed):

libparsec/crates/platform_storage/tests/unit/native/cleanup.rs
```diff
@@ -26,9 +26,11 @@ async fn cleanup_device_data(tmp_path: TmpPath, env: &TestbedEnv) {
...
     // 3) Cleanup the device data.
-    remove_device_data(&tmp_path, &alice.device_id)
+    remove_device_data(&tmp_path, alice.device_id)
         .await
         .unwrap();
     assert!(!wksp_file.exists(), "wksp file still exists");
     assert!(!device_dir.exists(), "device dir still exists");
 }
+
+// TODO: Convert this test be compatible with web
```
FirelightFlagboy, line 36: **todo (blocking):** Create the related issue and link it in the comment

libparsec/crates/client/src/certif/store.rs
```diff
@@ -211,43 +211,43 @@ impl CertificatesStore {
...
-        };
+        let outcome = storage
+            .for_update(async |updater| {
+                let mut write_guard = CertificatesStoreWriteGuard {
+                    store: self,
+                    storage: updater,
+                };
 
-        let outcome = cb(&mut write_guard).await;
+                cb(&mut write_guard).await
```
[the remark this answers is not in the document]
touilleMan: `CertificatesStoreWriteGuard` provides the business logic methods (e.g. `CertificatesStoreWriteGuard::add_next_common_certificate`)

libparsec/crates/platform_storage/src/certificates.rs
```diff
@@ -687,11 +687,22 @@ impl CertificatesStorage {
...
+        cb: impl AsyncFnOnce(CertificatesStorageUpdater) -> Result<R, E>,
+    ) -> anyhow::Result<Result<R, E>> {
         self.platform
-            .for_update()
+            .for_update(async |platform_updater| {
+                let updater = CertificatesStorageUpdater {
+                    platform: platform_updater,
+                };
+                cb(updater).await
+            })
```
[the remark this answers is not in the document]
touilleMan: not sure what you mean by "restriction" :thinking:

libparsec/crates/platform_storage/src/native/certificates.rs
```diff
@@ -669,17 +680,58 @@ impl PlatformCertificatesStorage {
...
+            }
+            Err(err) => {
+                transaction.rollback().await?;
+                Ok(Err(err))
+            }
+        }
     }
 
     pub async fn get_last_timestamps(&mut self) -> anyhow::Result<PerTopicLastTimestamps> {
         // TODO: transaction shouldn't be needed here (but it's currently easier to implement this way)
```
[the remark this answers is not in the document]
touilleMan: Using `self.for_update` is not very elegant since it may cause cause all read operations to be sequential.

Currently we don't use [`BEGIN IMMEDIATE`](https://www.sqlite.org/lang_transaction.html) when opening a transaction, so transactions are in read mode by default (and hence multiple can be run concurrently). However for whatever reason we change this in `for_update`, then the read operations will all use write transactions that can only be run in sequential order.

libparsec/crates/platform_storage/src/native/certificates.rs
```diff
@@ -700,18 +752,20 @@ impl PlatformCertificatesStorage {
         limit: Option<u32>,
     ) -> anyhow::Result<Vec<(DateTime, Vec<u8>)>> {
         // TODO: transaction shouldn't be needed here (but it's currently easier to implement this way)
```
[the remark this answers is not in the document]
touilleMan: https://github.com/Scille/parsec-cloud/pull/9926#discussion_r2014312923

libparsec/crates/platform_storage/tests/unit/native/cleanup.rs
touilleMan: done (see https://github.com/Scille/parsec-cloud/issues/10007)

libparsec/crates/platform_storage/src/web/certificates.rs
```diff
@@ -127,223 +149,699 @@ impl<'a> PlatformCertificatesStorageForUpdateGuard<'a> {
...
 pub(crate) struct PlatformCertificatesStorage {
-    conn: Arc<IdbDatabase>,
+    conn: Database,
 }
 
-// Safety: PlatformCertificatesStorage is read only
+// SAFETY: see `pretend_future_is_send_on_web`'s documentation for the full explanation.
+// `PlatformCertificatesStorage` contains `indexed_db::Database` which is `!Send` since
+// it interfaces with the browser's IndexedDB API that is not thread-safe.
+// Since are always mono-threaded in web, we can safely pretend to be `Send`, which is
```
FirelightFlagboy, line 754: **typo:** 
```suggestion
// Since we are always mono-threaded in web, we can safely pretend to be `Send`, which is
```

FirelightFlagboy (approved):

Max-7 (approved):

libparsec/crates/platform_storage/src/web/certificates.rs
touilleMan: done :+1:

PR #9953, opened 2025-03-17
[MS] Parsec Auth first draft

Max-7:
We add a page right at the start of the app to log in to Parsec auth. Once logged in, we progress to the home page normally, which will list the devices.

- With Playwright, we straight up skip the login page. Maybe we can find a way to test it later on, but it will just slow down most tests.
- We provide a setting option to skip the login page for people who don't care about Parsec Auth
- Once logged in to Parsec Auth, or if we skip it, we forward the query and params to make sure invitation and file links get processed
- We provide a way to mock or change the server for Parsec Auth
- We redirect `/` to the auth page if Parsec Auth is not skipped

This should be easier than adding another state into the HomePage.

- [X] Describe your changes


client/src/views/home/HomePageHeader.vue
```diff
@@ -113,14 +119,26 @@ import { onMounted, onUnmounted, ref, inject, Ref } from 'vue';
...
 const eventDistributor = injectionProvider.getDefault().eventDistributor;
 let eventCbId: string | null = null;
 const updateAvailability: Ref<UpdateAvailabilityData | null> = ref(null);
+const hotkeyManager: HotkeyManager = inject(HotkeyManagerKey)!;
+
+let hotkeys: HotkeyGroup | null = null;
 
 onMounted(async () => {
+  hotkeys = hotkeyManager.newHotkeys();
+  hotkeys.add({ key: ',', modifiers: Modifiers.Ctrl, platforms: Platforms.Desktop, disableIfModal: true }, openSettingsModal);
```
Ironicbay, line 139: You should remove these 2 keybinds from `HomePage.vue` if you're adding them here

Ironicbay (approved):

PR #9962, opened 2025-03-19
Bump sqlx and libsqlite3-sys

### LINKED ISSUE #8052: Migrate to `sqlx@8`
Changelog: https://github.com/launchbadge/sqlx/blob/v0.8.0/CHANGELOG.md
Example migration: https://github.com/cloudwalk/stratus/pull/1673/files

mmmarcos:
Bump `sqlx` from `0.7.2` to `0.8.3`
Bump `libsqlite3-sys` from `0.26.0` to `0.30.1`

The `fetch_many` function is deprecated. It was intended for executing a single query string containing multiple SQL statements while still using bind parameters. See: https://github.com/launchbadge/sqlx/issues/3108

Since we were not using it for that, the `fetch_all` function seems to be enough.

Closes #8052

### DIFF
--- a/libparsec/crates/platform_storage/src/native/workspace.rs
+++ b/libparsec/crates/platform_storage/src/native/workspace.rs
@@ -4,7 +4,7 @@
 // validation work and takes care of handling concurrency issues.
 // Hence no unique violation should occur under normal circumstances here.
 
-use libparsec_platform_async::stream::{StreamExt, TryStreamExt};
+use libparsec_platform_async::future::TryFutureExt;
 use sqlx::{
     sqlite::{SqliteConnectOptions, SqliteJournalMode, SqliteSynchronous},
     ConnectOptions, Connection, Row, SqliteConnection,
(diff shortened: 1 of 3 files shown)


libparsec/crates/platform_storage/src/native/workspace.rs
```diff
@@ -665,21 +664,15 @@ async fn db_list_manifests(
...
-        .try_collect()
-        .await;
+        .fetch_all(executor);
+
+    let res: anyhow::Result<Vec<RawEncryptedManifest>> = match rows.await {
+        Ok(v) => v
+            .iter()
+            .map(|row| row.try_get(0).map_err(anyhow::Error::from))
+            .collect(),
+        Err(e) => Err(e.into()),
```
FirelightFlagboy, line 674: **nitpick (if-minor):** I would have used the `map_ok` and `err_into` from [`TryFutureExt`](https://docs.rs/futures/latest/futures/future/trait.TryFutureExt.html)

mmmarcos: Not sure how to use those, can you add a code suggestion?

FirelightFlagboy:
```suggestion
    rows
      .await
      .map_ok(|v| v
            .iter()
            .map(|row| row.try_get(0).map_err(anyhow::Error::from))
            .collect::<Vec<_>>()
      ) 
      .err_into::<anyhow::Error>()
```

I would also merge that block with the `fetch_all()` above

FirelightFlagboy (approved):

PR #9983, opened 2025-03-24
Fix web bindings tuple type

touilleMan:


FirelightFlagboy (approved):

touilleMan:
> I would squash the 2 commits together ;)

That's a matter of style, but so far we have always keep the commits of generated code apart ;-)

(the rational for keeping it apart is: 1. simpler handling of git conflict since the commit can just be discarded and re-generated, 2. less clutter when consulting the git history since the actual important change is not drawn in a lots of related-but-not-pertinent autogenerated code)

Max-7 (approved):

PR #9986, opened 2025-03-24
feat(webapp): Define `base_href` to `./`

FirelightFlagboy:
`vite` support relative base
(<https://vite.dev/guide/build#relative-base>).

This is fill our gap where we do not know in advance the how the app will be hosted.

### DIFF
--- a/.github/workflows/package-client.yml
+++ b/.github/workflows/package-client.yml
@@ -126,6 +126,9 @@ jobs:
         run: npm run web:release
         env:
           PARSEC_APP_SENTRY_AUTH_TOKEN: ${{ secrets.SENTRY_AUTH_TOKEN }}
+          # Configure the base href of the WebApp,
+          # we use a relative path since we want to provide a bundle that does not enforce a way to distribute the app.
+          PARSEC_APP_BASEHREF: ./
         working-directory: client
 
       - name: Zip webapp content
--- a/client/vite.config.ts
+++ b/client/vite.config.ts
@@ -13,12 +13,12 @@ import wasmPack from './scripts/vite_plugin_wasm_pack';
 const plugins: PluginOption[] = [vue(), topLevelAwait()];
 let platform: string;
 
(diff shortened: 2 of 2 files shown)


client/vite.config.ts
```diff
@@ -13,12 +13,12 @@ import wasmPack from './scripts/vite_plugin_wasm_pack';
 const plugins: PluginOption[] = [vue(), topLevelAwait()];
 let platform: string;
 
-// Vite only expose in `import.meta.env` the environ variables with a `PARSEC_APP_` prefix,
+// Vite only expose in `import.meta.env` the environment variables with a `PARSEC_APP_` prefix,
 // but to make it a bit easier, we're also letting the user provide TESTBED_SERVER
 if (process.env.PARSEC_APP_TESTBED_SERVER || process.env.TESTBED_SERVER) {
-  // Why this if guard ? Guess what kiddo !
+  // Why this if guard? Guess what kiddo !
```
touilleMan, line 19:
```suggestion
  // Why this if guard? Guess what kiddo!
```

.github/workflows/package-client.yml
```diff
@@ -126,6 +126,9 @@ jobs:
         run: npm run web:release
         env:
           PARSEC_APP_SENTRY_AUTH_TOKEN: ${{ secrets.SENTRY_AUTH_TOKEN }}
+          # Configure the base href of the WebApp,
+          # we use a relative path since we want to provide a bundle that does not enforce a way to distribute the app.
+          PARSEC_APP_BASEHREF: ./
```
touilleMan, line 131: Do we have a reason to make this configurable ?

My concern here is we are going to to end up with different default depending on if we generate the release from the CI or with a `npm run web:release` run from the dev machine (which will lead to some headache on why the release fails when hosted from the server...)

FirelightFlagboy: Do you prefer that we change the default used in `vite.config.ts` ?

FirelightFlagboy:
Superseded by #9992

PR #9991, opened 2025-03-24
Add param `--with-client-web-app` to server run CLI

touilleMan:


server/parsec/asgi/__init__.py
```diff
@@ -49,16 +50,24 @@ def app_factory(backend: Backend) -> AsgiApp:
...
+            return templates.TemplateResponse(request, "index.html")
+
+    app.get("/")(root)
+
+    app.mount("/static", StaticFiles(packages=[("parsec", "static")]))
 
     async def page_not_found(scope: Scope, receive: Receive, send: Send) -> None:
         request = Request(scope)
-        response = templates.TemplateResponse("404.html", {"request": request}, status_code=404)
+        response = templates.TemplateResponse(request, "404.html", status_code=404)
```
FirelightFlagboy, line 70: **nitpick (non-blocking):** I would use full named parameters

```suggestion
        response = templates.TemplateResponse(request=request, name="404.html", status_code=404)
```

server/parsec/cli/run.py
```diff
@@ -305,6 +305,26 @@ def handle_parse_result(
...
+    show_default=True,
+    type=int,
+    callback=lambda ctx, param, value: None if value is None or value <= 0 else value,
+    envvar="PARSEC_SSE_KEEPALIVE",
+    show_envvar=True,
+    help="Keep SSE connection open by sending keepalive messages to client, in seconds (pass <= 0 to disable)",
+)
+@click.option(
+    "--with-client-web-app",
+    type=click.Path(exists=True, file_okay=False),
```
FirelightFlagboy, line 320: **nitpick (non-blocking):** I would be explicit about we accept dir and not file

```suggestion
    type=click.Path(exists=True, file_okay=False, dir_okay=True),
```

FirelightFlagboy, line 320: **polish (non-blocking):** By reading the documentation about `click.Path`, we could specify the type of the option to `pathlib.Path`

```suggestion
    type=click.Path(exists=True, file_okay=False, path_type=pathlib.Path),
```

FirelightFlagboy (approved):

server/parsec/asgi/__init__.py
touilleMan: done.

server/parsec/cli/run.py
touilleMan: done.

PR #10009, opened 2025-03-26
add fastapi middleware to process SPA router

Landeers:
## Describe your changes

Base url of SPA into Parsec server is `/client`. So when user go to `/`, server redirect to `/client`. Then SPA homepage is `/home`. And final url is `/client/home`.
But if user refresh page, `/client/home` do not exist (404). So we must deal with this case by adding a fastapi middleware to response `index.html` file on 404 response.

## Checklist

- [x] Describe your changes

### DIFF
--- a/server/parsec/asgi/__init__.py
+++ b/server/parsec/asgi/__init__.py
@@ -3,7 +3,7 @@ from __future__ import annotations
 
 import os
 from pathlib import Path
-from typing import Any, Awaitable, Callable, cast
+from typing import cast
 
 import anyio
 import uvicorn
@@ -12,8 +12,6 @@ from fastapi.middleware.cors import CORSMiddleware
 from fastapi.responses import RedirectResponse
 from fastapi.staticfiles import StaticFiles
 from fastapi.templating import Jinja2Templates
-from starlette.middleware.base import BaseHTTPMiddleware
-from starlette.responses import FileResponse
 from starlette.types import Receive, Scope, Send
 
 from parsec._version import __version__ as parsec_version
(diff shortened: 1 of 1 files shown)


FirelightFlagboy (commented):
Why not using an embedded app ?

```python
client_app = FastAPI()
client_app.mount("/", StaticFiles(directory=with_client_web_app, html=True))

async def default_to_index(scope: Scope, receive: Receive, send: Send) -> None:
    await FileResponse(with_client_web_app / "index.html")

client_app.router.default = default_to_index

app.mount(client_base_url, client_app)
```

server/parsec/asgi/__init__.py
```diff
@@ -41,6 +44,23 @@
     },
 ]
 
+client_base_url = "/client"
```
touilleMan, line 47:
```suggestion
WEB_APP_BASE_URL = "/client"
```

server/parsec/asgi/__init__.py
```diff
@@ -41,6 +44,23 @@
...
+
+
+class SPARedirectMiddleware(BaseHTTPMiddleware):
```
touilleMan, line 50: Instead of providing a middleware that gets called for every route, we can instead just overload the `StaticFiles` class:

```python
class StaticFilesWithSPARedirect(StaticFiles):
    def lookup_path(self, path: str) -> tuple[str, os.stat_result | None]:
        match super().lookup_path(path):
            case (_, None):
                return super().lookup_path("index.html")
            case found:
                return found
```
(I've quickly tested this code, it seems to work as intended)

FirelightFlagboy: Sees better than my solution for embedded app

Landeers: Test ok on my server !
done.

FirelightFlagboy (approved):
LGTM, a little squash would be nice ;)

touilleMan (approved):

PR #10017, opened 2025-03-27
Add forget all certificates & poll to client CLI

touilleMan:


cli/src/commands/certificate/poll.rs
```diff
@@ -0,0 +1,21 @@
...
+crate::clap_parser_with_shared_opts_builder!(
+    #[with = config_dir, device, password_stdin]
+    pub struct Args {
+    }
+);
+
+crate::build_main_with_client!(main, poll);
+
+pub async fn poll(_args: Args, client: &StartedClient) -> anyhow::Result<()> {
+    let mut spinner = start_spinner("Poll server for new certificates".into());
```
AureliaDolo, line 14: [Here](https://github.com/Scille/parsec-cloud/blob/84d9c33331ff0e6d40cc61707e28fe3db89b51de/cli/src/utils.rs#L432) is an helper

AureliaDolo (approved):
A few nits but otherwise LGTM

cli/src/commands/certificate/poll.rs
touilleMan: I used it initially, however this helper doesn't display the number of new certificates

I think not showing the number of new certificates is fine when the poll is a side task, but for this CLI command it is the main thing so reporting is a good thing

libparsec/crates/client/src/certif/mod.rs
```diff
@@ -303,7 +317,7 @@ impl CertificateOps {
     pub async fn poll_server_for_new_certificates(
         &self,
         latest_known_timestamps: Option<&PerTopicLastTimestamps>,
-    ) -> Result<(), CertifPollServerError> {
+    ) -> Result<usize, CertifPollServerError> {
```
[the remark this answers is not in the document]
touilleMan: returns the number of new certificates

I've added a comment (there was already one for `Client::poll_server_for_new_certificates` but not for `CertficateOps::poll_server_for_new_certificates`

libparsec/crates/client/tests/unit/certif/forget_all_certificates.rs
```diff
@@ -0,0 +1,52 @@
...
+#[parsec_test(testbed = "minimal")]
+async fn ok(env: &TestbedEnv) {
+    env.customize(|builder| {
+        builder.certificates_storage_fetch_certificates("alice@dev1");
+    })
+    .await;
+    let alice = env.local_device("alice@dev1");
+    let ops = certificates_ops_factory(env, &alice).await;
+
+    let add_first_certifcate = async || {
```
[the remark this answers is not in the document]
touilleMan: :+1: done.

cli/src/commands/certificate/forget_all_certificates.rs
```diff
@@ -0,0 +1,35 @@
...
+
+pub async fn forget_all_certificates(_args: Args, client: &StartedClient) -> anyhow::Result<()> {
+    let short_id = &client.device_id().hex()[..3];
+    let organization_id = client.organization_id();
+    let human_handle = client.human_handle();
+    let device_label = client.device_label();
+
+    println!("You are about to clear the local certificates database for device:");
+    println!("{YELLOW}{short_id}{RESET} - {organization_id}: {human_handle} @ {device_label}");
+    println!("Are you sure? (y/n)");
```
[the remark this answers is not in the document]
touilleMan: I'm all okay for additional shininess, though I'd rather not do this change in this PR

This is because I've copied this code from another command, so it should also be modified

So I've opened issue #10020

client/src/parsec/login.ts
```diff
@@ -396,22 +396,27 @@ export async function getCurrentAvailableDevice(): Promise<Result<AvailableDevic
...
       accessStrategy.tag === DeviceAccessStrategyTag.Password &&
       (accessStrategy as DeviceAccessStrategyPassword).password !== 'P@ssw0rd.'
     ) {
-      return { ok: false, error: { tag: ClientChangeAuthenticationErrorTag.DecryptionFailed, error: 'Invalid password' } };
+      return { ok: false, error: { tag: UpdateDeviceErrorTag.DecryptionFailed, error: 'Invalid password' } };
     }
-    return { ok: true, value: null };
+    return {
+      ok: true,
+      // @ts-ignore: Too cumbersome to create a full `AvailableDevice` object,
```
Max-7, line 416: You can just c/c from https://github.com/Scille/parsec-cloud/blob/050fceb7664512f9f1f1ca863a165f9a999711a6/client/src/parsec/claim_device.ts#L253 It may be easier than trying to trick the linter.
This code will be gone very soon anyway.

Max-7 (approved):

FirelightFlagboy (changes requested):
Missing newsfragment

cli/src/commands/certificate/poll.rs
FirelightFlagboy: The function could easily updated to return the number of certificate

touilleMan:
@FirelightFlagboy I've added the newsfragment :pray:

client/src/parsec/login.ts
touilleMan: done, thanks :pray:

cli/src/commands/device/change_authentication.rs
```diff
@@ -0,0 +1,95 @@
...
+
+use crate::utils::*;
+
+crate::clap_parser_with_shared_opts_builder!(
+    #[with = config_dir, device, password_stdin]
+    pub struct Args {
+   #[clap(long, short, action)]
+        password: bool,
+   #[clap(long, short, action)]
+        keyring: bool,
```
FirelightFlagboy, line 13: **suggestion (blocking):** Use a enum instead that simply the pattern matching

FirelightFlagboy (changes requested):
- Missing test for cli `change-authentication`
- Missing test for cli `overwrite-server-url`

cli/src/commands/device/change_authentication.rs
touilleMan: I've created issue #10028 to track this, since we want to also standardize the other CLI commands

touilleMan:
@FirelightFlagboy I've added a test for each new CLI command (and I found an error in the naming of a command thank this them ^^)

libparsec/crates/platform_device_loader/src/web/mod.rs
```diff
@@ -56,34 +59,52 @@ pub async fn save_device(
...
 
     if key_file != new_key_file {
-        storage.remove_device(key_file)?;
+        if let Err(err) = storage.remove_device(key_file) {
+            log::warn!("Cannot remove old key file {key_file:?}: {err}");
+        }
     }
 
-    Ok(available_device)
+    Ok((available_device, old_server_addr))
```
mmmarcos, line 97: The old_server_address doesn't seem to be used anywhere, is it needed?

libparsec/crates/platform_device_loader/src/lib.rs
```diff
@@ -198,19 +196,40 @@ impl From<SaveDeviceError> for ChangeAuthentificationError {
...
+    }
+
+    platform::update_device(current_access, new_access, None)
+        .await
+        .map(|(available_device, _)| available_device)
+}
+
+/// Note `config_dir` is only used as discriminant for the testbed here
+///
+/// Returns the old server address
```
mmmarcos, line 217: Not sure if this is relevant but there seem to be an asymmetry between
- `update_device_change_authentication` returning the updated device, and
- `update_device_overwrite_server_address` returning the old server address

Why not returning the device in both cases? or the changed element in both cases?
Also, why "overwrite" and not simply "change"?

There is beauty in symmetry ^^

cli/src/commands/device/overwrite_server_url.rs
mmmarcos, line 1: I'll rename it to `change_server_url` to be consistent with `change_authentication_method` (see my comment in libparsec)

libparsec/crates/platform_device_loader/src/web/mod.rs
touilleMan: This is used in the CLI to display the old address before overwritting it

cli/src/commands/device/overwrite_server_url.rs
touilleMan: I don't mind changing the name, however the important thing to consider is that `change_authentication_method` is a legit operation while `overwrite_server_url` is more of a hack for advanced usages (hence why the `overwrite` has a way to indicate some information is going to be lost).

mmmarcos: OK, so let's stick with overwrite then, specially if it's not intended for normal uses.

libparsec/crates/platform_device_loader/src/lib.rs
mmmarcos: https://github.com/Scille/parsec-cloud/pull/10017#discussion_r2018511704

cli/tests/integration/device/change_authentication.rs
```diff
@@ -0,0 +1,48 @@
...
+use libparsec_tests_fixtures::prelude::*;
+
+use crate::{
+    integration_tests::{bootstrap_cli_test, wait_for},
+    testenv_utils::{TestOrganization, DEFAULT_DEVICE_PASSWORD},
+};
+
+#[rstest::rstest]
+#[tokio::test]
+async fn ok(tmp_path: TmpPath) {
```
FirelightFlagboy, line 14: **issue (blocking):** The test is not complete, we need to check if the device can be unlocked with the new authentication

cli/tests/integration/device/overwrite_server_url.rs
```diff
@@ -0,0 +1,58 @@
...
+use libparsec_tests_fixtures::prelude::*;
+
+use crate::{
+    integration_tests::{bootstrap_cli_test, wait_for},
+    testenv_utils::{TestOrganization, DEFAULT_DEVICE_PASSWORD},
+};
+
+#[rstest::rstest]
+#[tokio::test]
+async fn ok(tmp_path: TmpPath) {
```
FirelightFlagboy, line 14: **todo:** Add a todo comment about inspecting the info of the device via `device info` (command not exiting for now)

libparsec/crates/platform_device_loader/src/lib.rs
touilleMan: Returning `AvailableDevice` is useful to display the new device (e.g. showing its short ID in the CLI, displaying the organization & human handle when importing a recovery device)

However in case of `update_device_overwrite_server_address` we don't care about this (since those information don't change at all), but want something totally unrelated (i.e. the old server url)

cli/tests/integration/device/change_authentication.rs
touilleMan: done :+1:

cli/tests/integration/device/overwrite_server_url.rs
touilleMan: done.

PR #10024, opened 2025-03-28
[MS] Update GUI to use bindings in web mode

### LINKED ISSUE #9368: Test web app GUI without mock
Test the web GUI without mocking

- list device
- login
- invite user /  greet
- workspace create / share
- file download

How does the mountpoint behaves? are there any errors? should it be disabled?

Max-7:
Closes #9368

Parsec is fully functionnal on web (as far as I can tell). We're still missing a few tests that have been marked as skipped for now (see #10029)

Quite a lot has changed.
The old mocks bypassed libparsec entirely, and just returned fake data. Now libparsec is always called and returns what data is available. This does mean that if you want to see the file page with files, you have to import them first. Same with the users page, you have to invite new users first.
While using the testbed, your data are not remanent. You're switching organization everytime you refresh the page. The testbed does provide three users (admin/member/external) and a default workspace named `wksp1`  created by the admin that is already shared with the member user.
To ease the pain, a special page named `DevLayout` has been added. It loads first when you visit a page, and checks if you're on the testbed and what env variables are defined.   Some env variables (already added in .env.development and documented in the README) will import a few files into the default workspace. It can also log in automatically, this way if you refresh the page while editing the Users page for example, it will keep you on this page with a default device instead of forcing you to log back in again.

It also means that to work on device/user greet/join you need two tabs, you can't just click `next` anymore.

- [X] Describe your changes


client/playwright.config.ts
```diff
@@ -37,6 +37,16 @@ export default defineConfig({
...
+  testIgnore: [
+    '**/client_*',
+    '**/workspace_history.spec.ts',
+    '**/my_profile_page.spec.ts',
+    '**/organization_switch_popover.spec.ts',
+    '**/file_details.spec.ts',
+    '**/export_recovery_device.spec.ts',
+    '**/import_recovery_device.spec.ts',
+    '**/device_greet.spec.ts',
+  ],
```
mmmarcos, line 49: Please create an issue to track this.

Max-7: #10029

client/src/views/layouts/ConnectedLayout.vue
```diff
@@ -88,7 +71,7 @@ onMounted(async () => {
 
   recentDocumentManager.resetHistory();
   await recentDocumentManager.loadFromStorage(storageManager, handle);
-  await mockRecentDocuments();
+  // await mockRecentDocuments();
```
NicoTuxx, line 74: why keep this comment when you removed the function?

NicoTuxx (approved):
LGTM

PR #10027, opened 2025-03-28
[MS] Improve prettier & lint commands

### LINKED ISSUE #10026: GUI linter checks differ between pre-commit and CI
Pre-commit has to manually handle the dependencies:
https://github.com/Scille/parsec-cloud/blob/050fceb7664512f9f1f1ca863a165f9a999711a6/.pre-commit-config.yaml#L262-L301

While the CI doesn't use pre-commit to run the commands:
https://github.com/Scille/parsec-cloud/blob/050fceb7664512f9f1f1ca863a165f9a999711a6/.github/workflows/ci-web.yml#L75-L88

This is not great since it means there is no single source of truth for those rules, and the CI may fail even after pre-commit accepted a given commit :/

On top of that those rules are disabled in pre-commit (they are in manual mode), this is also not great since it means errors are only found later on (it is better to have it enabled and let the dev use `SKIP=eslint` if he want to commit faster).

touilleMan:
close #10026


Max-7 (approved):

.pre-commit-config.yaml
```diff
@@ -257,47 +257,63 @@ repos:
...
+  # We use the same trick that for Python here: we run the tools with `npx`
+  # from the GUI project directory to reuse the installed dependencies.
   - repo: local
     hooks:
       - id: eslint
-        name: eslint
-        entry: eslint
-        language: node
+        name: eslint (using `npx`)
+        # Note we don't use `npx --prefix=./client eslint` because it causes
```
FirelightFlagboy, line 266: **suggestion (non-blocking):** 
```suggestion
        # Note we don't use `npx eslint` with `--prefix=./client` because it causes
```

.pre-commit-config.yaml
```diff
@@ -257,47 +257,63 @@ repos:
...
-          - eslint-plugin-import@2.27.5
-          - eslint-plugin-n@15.6.1
-          - eslint-plugin-promise@6.1.1
-          - eslint-plugin-vue@8.7.1
 
-  - repo: local
-    hooks:
+      - id: vue-tsc
+        name: vue-tsc (using `npx`)
+        # Note we don't use `npx --prefix=./client vue-tsc` because it causes
```
FirelightFlagboy, line 284: **suggestion (non-blocking):** 
```suggestion
        # Note we don't use `npx vue-tsc` with `--prefix=./client` because it causes
```

.pre-commit-config.yaml
```diff
@@ -257,47 +257,63 @@ repos:
...
       - id: prettier
-        name: prettier
-        entry: >-
-          npx prettier
-          --config=client/.prettierrc.json5
-          --ignore-path=client/.prettierignore
-          --write
-        language: node
+        name: prettier (using `npx`)
+        # Note we don't use `npx --prefix=./client prettier` because it causes
```
FirelightFlagboy, line 302: **suggestion (non-blocking):** 
```suggestion
        # Note we don't use `npx prettier` with `--prefix=./client` because it causes
```

.github/workflows/ci.yml
```diff
@@ -213,7 +213,7 @@ jobs:
         with:
           config-file: ${{ steps.patched-pre-commit-config.outputs.path }}
         env:
-          SKIP: clippy,fmt,cspell,ruff,ruff-format,pyright,eslint,deptry,sqlfluff,powrap
+          SKIP: clippy,fmt,cspell,ruff,ruff-format,pyright,eslint,prettier,deptry,sqlfluff,powrap
```
FirelightFlagboy, line 216: **todo (blocking):** Missing `vue-tsc`

FirelightFlagboy (changes requested):

FirelightFlagboy (approved):

PR #10030, opened 2025-03-28
fix(client): Allow using relative base url by using baseURI

FirelightFlagboy:
`vue-router` does not support relative `BASE_URL` and need an absolute value.
We configured the project to use a relative `BASE_URL` to simplify hosting and to not have specific builds.

The solution was to use `document.baseURI`


FirelightFlagboy:
Superseded by #10034

PR #10071, opened 2025-04-01
Fix protocol generation

AureliaDolo:
So it can account for two words families

### DIFF
--- a/libparsec/crates/serialization_format/src/protocol.rs
+++ b/libparsec/crates/serialization_format/src/protocol.rs
@@ -512,7 +512,8 @@ fn quote_cmd(family: &str, cmd_version: u32, cmd: &GenCmd) -> (TokenStream, Toke
     let command_name = snake_case_name;
 
     // `authenticated` -> `Authenticated`
-    let family_enum_variant = format_ident!("{}{}", family[..1].to_uppercase(), family[1..]);
+    // 'authenticated_account' -> 'AuthenticatedAccount'
+    let family_enum_variant = format_ident!("{}", snake_to_pascal_case(family));
 
     let module = match &cmd.spec {
         GenCmdSpec::ReusedFromVersion { version } => {
--- a/misc/gen_protocol_typings.py
+++ b/misc/gen_protocol_typings.py
@@ -65,7 +65,7 @@ SpecsByMajor: TypeAlias = dict[MajorVersion, CmdSpecs]
 SpecsByFamily: TypeAlias = dict[str, SpecsByMajor]
 
 
-def snake_case_to_camel_case(name: str) -> str:
+def snake_case_to_upper_camel_case(name: str) -> str:
     words = name.split("_")
 
     if len(words) == 1:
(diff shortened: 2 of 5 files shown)


libparsec/crates/serialization_format/src/protocol.rs
```diff
@@ -503,6 +503,12 @@ fn quote_versioned_cmds(family: &str, version: u32, cmds: &[GenCmd]) -> (Ident,
     (versioned_cmds_mod, code)
 }
 
+fn format_family_name_to_ident(s: &str) -> String {
+    s.split("_")
+        .map(|w| format!("{}{}", &w[..1].to_uppercase(), &w[1..]))
+        .fold(String::new(), |acc, w| acc + &w)
+}
```
FirelightFlagboy, line 510: **chore:** We already have `snake_to_pascal_case` for that

AureliaDolo: Thanks !

misc/gen_protocol_typings.py
```diff
@@ -549,9 +549,10 @@ def load(cls, raw: bytes) -> {"|".join(cmd + ".Req" for cmd in cmds_names)}: ...
     test_rpc_code_need_import_types: set[str] = set()
     for family_name, specs in sorted(collect_specs().items()):
         family_mod_name = f"{family_name}_cmds"
+        camel_case_family_name = snake_case_to_camel_case(family_name)
```
FirelightFlagboy, line 552: **nitpick (non-blocking):** Should not it be `PascalCase` instead of `camelCase`?

```suggestion
        pascal_case_family_name = snake_case_to_pascal_case(family_name)
```

AureliaDolo: So that shows I don't know how to read ^

AureliaDolo: So just to be sure, but [snake_case_to_camel_case](https://github.com/Scille/parsec-cloud/blob/6f8a42a756a77955c6acd9c31d1a9ec9fdad9cfc/misc/gen_protocol_typings.py#L68) is actually pascal case right ? The first letter is capitalized

FirelightFlagboy: Indeed, I you could rename the function instead :pray:

AureliaDolo: Sooo wikipedia says that pascal case is camel case https://en.wikipedia.org/?title=Pascal_Case&redirect=no :shrug:

FirelightFlagboy: :pitchfork:

FirelightFlagboy: > The more specific terms `Pascal case` and `upper camel case` refer to a joined phrase where the first letter of each word is capitalized

AureliaDolo: > Some people and organizations, notably [Microsoft](https://en.wikipedia.org/wiki/Microsoft), use the term camel case only for lower camel case, designating Pascal case for the upper camel case

FirelightFlagboy (approved):

PR #10073, opened 2025-04-01
Misc fixes in testbed

touilleMan:


mmmarcos (approved):

libparsec/crates/testbed/src/template/events.rs
```diff
@@ -561,15 +561,17 @@ impl TestbedEventBootstrapOrganization {
                 (human_handle, device_id)
             }
             None => {
-                let human_handle = HumanHandle::new_redacted(first_user_id);
+                let label = first_user_id.hex();
+                let email = format!("{}@example.com", &label);
```
FirelightFlagboy, line 565: **nitpick:** 
```suggestion
                let email = format!("{label}@example.com");
```

libparsec/crates/testbed/src/template/events.rs
```diff
@@ -561,15 +561,17 @@ impl TestbedEventBootstrapOrganization {
...
+                let human_handle = HumanHandle::new(&email, &label).unwrap();
                 let device_id = builder.counters.next_device_id();
                 (human_handle, device_id)
             }
         };
 
         let device_label = match device_id.test_nickname() {
             Some(nickname) => format!("My {} machine", nickname).parse().unwrap(),
```
FirelightFlagboy, line 573: **nitpick:** 
```suggestion
            Some(nickname) => format!("My {nickname} machine").parse().unwrap(),
```

libparsec/crates/testbed/src/template/events.rs
```diff
@@ -943,7 +945,9 @@ impl TestbedEventNewUser {
                 (human_handle, device_id)
             }
             None => {
-                let human_handle = HumanHandle::new_redacted(user_id);
+                let label = user_id.hex();
+                let email = format!("{}@example.com", &label);
```
FirelightFlagboy, line 949: **nitpick:** 
```suggestion
                let email = format!("{label}@example.com");
```

server/parsec/components/postgresql/user_update_user.py
```diff
@@ -184,7 +184,6 @@ async def user_update_user(
 
     # 4) All checks are good, now we do the actual insertion
 
-    # TODO: validate it's okay not to check this
     # Note an OUTSIDER is not supposed to be OWNER/MANAGER of a shared realm. However this
```
FirelightFlagboy, line 187: **polish:** I use an highlighter for TODO, NOTE, etc if they are formatted like so `<type>:`

```suggestion
    # NOTE: an OUTSIDER is not supposed to be OWNER/MANAGER of a shared realm. However this
```

libparsec/crates/testbed/src/template/events.rs
touilleMan: done.

libparsec/crates/testbed/src/template/events.rs
touilleMan: done.

libparsec/crates/testbed/src/template/events.rs
touilleMan: done.

server/parsec/components/postgresql/user_update_user.py
touilleMan: I don't think this is needed:
- currently we have a lot more `Note` than `NOTE` in the codebase (534 vs 21 resuts)
- unlike `TODO` which is pretty standard and serves a purpose (i.e. being able to grep the codebase) , using `NOTE` in a standardized way doesn't seem to bring any benefit (typically in the current code, the comment could have been phrased as "An OUTSIDER is ..." without changing it purpose while circumventing any rule about the use of "note").

PR #10090, opened 2025-04-02
docs(rfc): Add RFC about parsec-account

FirelightFlagboy:
Parsec Account is a new service that would allow user to store devices on a remote server.
This would be used to simplify enrolling/recovering new end devices (computer, phones, web browsers)

### DIFF
--- a/docs/rfcs/1013-parsec-account.md
+++ b/docs/rfcs/1013-parsec-account.md
@@ -1,0 +1,525 @@
+<!-- Parsec Cloud (https://parsec.cloud) Copyright (c) BUSL-1.1 2016-present Scille SAS -->
+
+# Store Parsec device on a remote server
+
+## Overview
+
+This RFC will discuss the implementation of a new service to store the user's devices remotely while still ensuring that only the user can use them.
+
+## Background & Motivation
+
+During our discussions about providing a client application that could run in a web-browser, came up the following questions:
+
+- How should we store the device in the browser?.
+
+  Some concerns were raised about the persistence of a Parsec device in the browser.
+
+- What would the user experience be like?
+
+  - What should happen when the user uses a different browser?
+  - What is the user expecting when using another computer?
+
+  We assume that the user should be able to connect from any browser/computer.
+
+That is how we came to the idea of storing the device on a remote server.
+
+- That will solve the issue of the browser cleaning its data.
+- The devices would be accessible from anywhere (given the user is online).
+
+## Goals
+
+- Describe how to store/access the user devices on a remote service.
+- Propose an API to implement this feature.
+- Focus on allowing users to connect using a password
+
+## Non-Goals
+
+- Discuss security concerns regarding the web client application.
+- How the devices are created (nothing new, the creation is performed as usual by the client)
+- Support authentication methods other than password (will be addressed in a new RFC)
+
+## Design
+
+### General principles
+
+An authentication service (Parsec Account) would be used to store user devices.
+This service will be requested before connecting to the metadata server (Parsec server) in order to retrieve a specific user device.
+The device will then be used to authenticate to the metadata server.
+
+<!-- Device stored on a third-party service -->
+```mermaid
+sequenceDiagram
+  autonumber
+  participant Client
+  participant S_Acc as Account server
+  participant S_Data as Metadata server
+
+  Client ->> S_Acc: Retrieve the device of the user
+  S_Acc ->> Client: Device of the user
+
+  Client ->> S_Data: Exchange with the server
+```
+
+### Glossary
+
+- **Parsec Account:** The name of the new service that will store the devices for a user.
+- **Account:** The user's account inside Parsec account.
+- **Vault:**: A pool of encrypted data accessible to a group of authentication methods.
+- **Protected device**: The devices that the user want to protect.
+
+- `auth_method_master_key`: A secret obtained from the authentication method.
+- `auth_method_hmac_key`: A secret used for HMAC based authentication.
+- `auth_method_secret_key`: A symmetric key to encrypt the vault key
+- `vault_key`: A symmetric key used to encrypt the vault.
+
+### Key generation
+
+- `auth_method_master_key` is generated from a password using a PBKDF algorithm (likely `argon2`).
+- `auth_method_hmac_key` and `auth_method_secret_key` are generated from `auth_method_master_key` using a KDF algorithm (the algorithm will be configured differently to not use the same key).
+- `vault_key` is generated from a CSPRNG.
+
+
+### How we authenticate the request
+
+Some indices are already present above, we will use an HMAC based authentication.
+
+> The reason for this choice is that we do not need additional request to get an OAuth token.
+
+The HMAC code will be put in the `Authorization` HTTP header and will have the following format:
+
+```bash
+PARSEC-HMAC-BLAKE2B.${id}.${account_email}.${timestamp}.${signature}
+```
+
+The signature is generated like so:
+
+```math
+\begin{gather}
+body\_sha256 = sha256(body) \\
+content = \text{"PARSEC-HMAC-BLAKE2B"} \Vert email \Vert timestamp \Vert body\_sha256 \\
+code = hmac_{black2b}(shared\_secret, content) \\
+signature = base64(code)
+\end{gather}
+```
+
+> [!TIP]
+> $\Vert$ is the concatenation operator.
+
+> We pass the checksum of the body and not the body itself to simplify generating the signature on low-power device (smartcard)
+> since they would have less data to process.
+
+### The data model
+
+#### Account
+
+[`Account`]: #account
+
+```yaml
+{
+  "label": "Account",
+  "type": "account",
+  "other_fields": {
+    "email": "EmailStr",
+    "label": "EmailLabel"
+  }
+}
+```
+
+The `label` is not used by Parsec Account server but will be used by the client to pre-fill the device label information (and the email field).
+
+#### Vault
+
+[`Vault`]: #vault
+
+```yaml
+{
+  "label": "Vault",
+  "type": "vault",
+  "other_fields": {
+    "created_on": "Timestamp",
+    "created_by": "Creator"
+  }
+}
+```
+
+> `Creator` is a string-like value containing IP address and User-Agent to
+> identify the creator of the vault.
+
+- There may be multiple vaults for an account.
+- For any given account, only a single vault is active.
+- The older vaults are kept for recovery purposes.
+
+#### Auth Method
+
+[`AuthMethod`]: #auth-method
+
+```yaml
+{
+  "label": "AuthMethod",
+  "type": "auth_method",
+  "other_fields": {
+    "algorithm": "KeyAlgorithm",
+    "hmac_key": "HMACKey",
+    "account_id": "AccountID",
+    "vault_id": "VaultID",
+    "vault_encrypted_key": "EncryptedSymKey",
+    "created_on": "Timestamp",
+    "created_by": "Creator"
+  }
+}
+```
+- The algorithm field contains the full configuration needed to do the derivations operations. Typically, for the password it should be `{"name": "Argon2iBlake2", "argon2i_opslimit": 65536, "argon2i_memlimit_kb": 3, "argon2i_parallelism": 1}`
+- There can be only a single password authentication method per vault.
+- Other authentication methods (such as FIDO2) may be added in the future.
+
+#### Protected Device
+
+[`ProtectedDevice`]: #protected-device
+
+```yaml
+{
+  "label": "ProtectedDevice",
+  "type": "protected_device",
+  "other_fields": {
+    "organization_id": "OrganizationID",
+    "device_id": "DeviceID",
+    "vault_id": "ID",
+    "encrypted_device": "EncryptedDevice"
+  }
+}
+```
+- Device are protected within a vault
+- Any given device can be protected in multiple vaults (this is typically the case when the devices have been recovered from an older vault).
+
+### API
+
+> [!IMPORTANT]
+> All requests are authenticated with a [HMAC signature](#how-we-authenticate-the-request) except stated otherwise.
+
+#### Account creation
+
+To create an account, the client starts by requesting an email validation token.
+
+```yaml
+{
+  "cmd": "account_send_email_validation_token",
+  "req": {
+    "email": "EmailStr"
+  },
+  "reps": [
+    {
+      "status": "ok"
+    },
+    {
+      "status": "invalid_email"
+    }
+  ]
+}
+```
+
+> This request is not authenticated with the server.
+
+On `ok`, the server would have sent a mail with a unique token used for next the request used to register the authentication method.
+
+> The token will be valid for a duration that will be defined at the implementation (likely will be configurable)
+
+If the email is already registered in the service, a `ok` response will still be sent (without sending an email)
+to avoid creating an oracle about emails registered in the service.
+
+```yml
+{
+  "cmd": "account_password_create",
+  "req": {
+    "email_validation_token": "EmailValidationToken",
+    "account_label": "String",
+    "auth_method_key_params": "KeyAlgorithm",
+    "auth_method_hmac_key": "HMACKey",
+    "encrypted_vault_key": "EncryptedSymKey"
+  },
+  "reps": [
+    {
+      "status": "ok",
+    },
+    {
+      "status": "invalid_email_validation_token"
+    }
+  ]
+}
+```
+
+> This request is authenticated with the newly generated key.
+> But it's possible that we make it unauthenticated in the implementation to not have a dedicated verification process for the request
+> (since it would need to use the data provided in the request vs save in the server).
+
+Upon receiving the request, the server will create the [`Account`], [`AuthMethod`], and [`Vault`].
+
+The IP address and User-agent of the creator are recorded (see Data Model).
+
+#### Uploading a new device
+
+To upload a new device the user need to retrieve the current vault information
+
+```yaml
+{
+  "cmd": "vault_get",
+  "req": {},
+  "reps": {
+    "ok": {
+      "vault_id": "VaultID",
+      "encrypted_key": "EncryptedSymKey",
+      "created_on": "Timestamp",
+      "created_by": "Creator",
+      # True if using the active vault (latest vault).
+      "active": "bool"
+    }
+  }
+}
+```
+
+It returns the vault information associated with the authentication method
+
+```yaml
+{
+  "cmd": "protected_device_upload",
+  "req": {
+    # We add the vault_id to limit concurrency issue in case another client perform a vault rotation.
+    "vault_id": "VaultID",
+    "device": "UploadedDevice"
+  },
+  "reps": {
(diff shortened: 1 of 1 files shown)


docs/rfcs/1013-parsec-account.md
```diff
@@ -0,0 +1,376 @@
...
+
+Once a device is recovered (i.e.: Was using an old vault and re-encrypted using the current `vault_key`),
+The server should clean up older devices with matching `organization_id` and `device_id` when processing the `protected_device_upload`.
+
+## Risks
+
+- The devices are not stored securely to only allow the user to access them.
+- During [perform clean up when recovering devices](#perform-clean-up-when-recovering-devices), the user can lose access to his devices if an attacker takes control of his account and replace the protected devices with dummy data).
+
+  > To remediate this, it is up to the database administrator to restore a backup.
```
FirelightFlagboy, line 360: **issue:** This is really not ideal if we have to rollback the database every-time someone account is compromised

docs/rfcs/1013-parsec-account.md
```diff
@@ -0,0 +1,376 @@
...
+#### Account deletion
+
+```yaml
+cmd: account_delete
+req: {}
+reps:
+  ok: {}
+```
+
+The server will send an email with a link to confirm the suppression of the account.
```
AureliaDolo, line 326: Will the link use a route like the following ?
```yaml
cmd: account_delete_validate
req:
    delete_token: DeleteToken
reps:
    ok: {}
    invalid_token: {}
```

docs/rfcs/1013-parsec-account.md
```diff
@@ -0,0 +1,376 @@
...
+## Risks
+
+- The devices are not stored securely to only allow the user to access them.
+- During [perform clean up when recovering devices](#perform-clean-up-when-recovering-devices), the user can lose access to his devices if an attacker takes control of his account and replace the protected devices with dummy data).
+
+  > To remediate this, it is up to the database administrator to restore a backup.
+
+## Remarks & open questions
+
+- The client will interact with parsec-account via `libparsec`, that way we reuse our RFC communication stack.
```
AureliaDolo, line 364:
```suggestion
- The client will interact with parsec-account via `libparsec`, that way we reuse our RPC communication stack.
```

docs/rfcs/1013-parsec-account.md
```diff
@@ -0,0 +1,376 @@
...
+- What to do when a user is revoked from an organization?
+  Should we remove the device, or should we just provide the information that the device for the given organization is revoked?
+
+  > Just inform the user about the revocation
+  >
+  > @touilleMan
+
+- A similar question applies for devices that are no longer used (e.g. a linked browser that is no longer used)
```
AureliaDolo, line 372: This could also apply to a old laptop's device.  But I think it's more a parsec metadata issue of having the option to remove unused devices (and then propagating this data to parsec account)

docs/rfcs/1013-parsec-account.md
```diff
@@ -0,0 +1,376 @@
...
+```yaml
+label: Vault
+type: vault
+other_fields:
+  created_on: Timestamp
+  created_by: Creator
+```
+
+> `Creator` is a string like value that represent by whom created the element,
+> it will include the user IP and user-agent.
```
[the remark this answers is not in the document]
FirelightFlagboy: It's not really useful to do that since a vaults are not made to shared with other users.
But if need we are still able to retrieve the account linked to a vault `Account <-ref- Auth Method -ref-> Vault`

docs/rfcs/1013-parsec-account.md
FirelightFlagboy: I've imagined that we would use a standalone link, would it be better to reuse the same system for the account creation? (where the user would need to enter the token in a input field that `libparsec` does the final request)

docs/rfcs/1013-parsec-account.md
```diff
@@ -0,0 +1,376 @@
...
+```
+
+> [!CAUTION]
+> The returned list does not include device from the current vault (since we already have access to them and do not need to recover them)
+
+#### Updating the password
+
+```yaml
+cmd: auth_method_password_update
+
```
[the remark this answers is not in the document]
FirelightFlagboy: done

docs/rfcs/1013-parsec-account.md
```diff
@@ -0,0 +1,376 @@
...
+> Meaning they would not access the devices (encrypted by the new `vault_key`).
+
+#### Account recovery
+
+```yaml
+cmd: account_recovery_start
+req:
+  email: EmailStr
+reps:
+  ok: {}
```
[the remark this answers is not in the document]
FirelightFlagboy: done

docs/rfcs/1013-parsec-account.md
```diff
@@ -0,0 +1,386 @@
...
+```yaml
+label: Vault
+type: vault
+other_fields:
+  created_on: Timestamp
+  created_by: Creator
+```
+
+> `Creator` is a string like value that represent by whom created the element,
+> it will include the user IP and user-agent.
```
mmmarcos, line 138: I'd reword it a little bit

```suggestion
> `Creator` is a string-like value containing IP address and User-Agent to 
> identify the creator of the vault.
```

docs/rfcs/1013-parsec-account.md
```diff
@@ -0,0 +1,436 @@
...
+  "other_fields": {
+    "created_on": "Timestamp",
+    "created_by": "Creator"
+  }
+}
+```
+
+> `Creator` is a string like value that represent by whom created the element,
+> it will include the user IP and user-agent.
+
```
mmmarcos, line 139: I would add notes discussed:
 
```suggestion
- There may be multiple vaults for an account.
- For any given account, only a single vault is active.
- The older vaults are kept for recovery purposes.

```

docs/rfcs/1013-parsec-account.md
```diff
@@ -0,0 +1,436 @@
...
+    "hmac_key": "HMACKey",
+    "account_id": "AccountID",
+    "vault_id": "VaultID",
+    "vault_encrypted_key": "EncryptedSymKey",
+    "created_on": "Timestamp",
+    "created_by": "Creator"
+  }
+}
+```
+
```
mmmarcos, line 159:
```suggestion
- The algorithm field contains the full configuration needed to do the derivations operations. Typically for the password it should be `{"name": "Argon2iBlake2", "argon2i_opslimit": 65536, "argon2i_memlimit_kb": 3, "argon2i_parallelism": 1}`
- There can be only a single password authentication method per vault.
- Other authentication methods (such as FIDO2) may be added in the future.

```

docs/rfcs/1013-parsec-account.md
```diff
@@ -0,0 +1,436 @@
...
+  "type": "protected_device",
+  "other_fields": {
+    "organization_id": "OrganizationID",
+    "device_id": "DeviceID",
+    "vault_id": "ID",
+    "encrypted_device": "EncryptedDevice"
+  }
+}
+```
+
```
mmmarcos, line 176:
```suggestion
- Device are protected within a vault
- Any given device can be protected in multiple vaults (this is typically the case when the devices have been recovered from an older vault).
    
```

docs/rfcs/1013-parsec-account.md
```diff
@@ -0,0 +1,436 @@
...
+### API
+
+#### Account creation
+
+To start the account creation, the client start by sending its email.
```
mmmarcos, line 181:
```suggestion
To create an account, the client starts by requesting an email validation token.
```

docs/rfcs/1013-parsec-account.md
```diff
@@ -0,0 +1,436 @@
...
+      "status": "ok",
+    },
+    {
+      "status": "invalid_email_validation_token"
+    }
+  ]
+}
+```
+
+Upon receiving the request, the server will create the [`Account`], [`AuthMethod`], and [`Vault`]
```
mmmarcos, line 225:
```suggestion
Upon receiving the request, the server will create the [`Account`], [`AuthMethod`], and [`Vault`].

The IP address and User-agent of the creator are recorded (see Data Model).
```

docs/rfcs/1013-parsec-account.md
mmmarcos: Good point. Taking into account the importance of the information stored AND that this action cannot be undone, it seems reasonable to protect the user against mistakes. 

So yes, we should probably use the send-token-to-confirm-deletion approach.

docs/rfcs/1013-parsec-account.md
```diff
@@ -0,0 +1,436 @@
...
+  },
+  "reps": {
+    "ok": {}
+  }
+}
+```
+
+The server will update the unique [`AuthMethod`] linked to the account with the updated field.
+At that point, the client will need to use the update key for the next requests.
+
```
mmmarcos, line 340:
```suggestion

The current password will be overwritten in the current authentication method.

The IP address and User-agent of the creator are recorded (see Data Model).
        
```

docs/rfcs/1013-parsec-account.md
```diff
@@ -0,0 +1,436 @@
...
+    "ok": {}
+  }
+}
+```
+
+Upon receiving this request, the server creates a new [`Vault`] and new [`ProtectedDevice`].
+
+> [!IMPORTANT]
+> Doing a vault rotation mean the other authentication methods would not have access to the new `vault_key`.
+> Meaning they would not have access to the devices (encrypted by the new `vault_key`).
```
mmmarcos, line 360:
```suggestion
> Doing a vault key rotation means the other authentication methods would not have access to the new `vault_key`,
> and therefore would not have access to the devices (which will be encrypted by the new `vault_key`).
```

docs/rfcs/1013-parsec-account.md
```diff
@@ -0,0 +1,456 @@
...
+    "organization_id": "OrganizationID",
+    "device_id": "DeviceID",
+    "encrypted_device": "EncryptedDevice"
+  }
+}
+```
+
+#### List devices
+
+To list the device linked to the authentication method's vault:
```
mmmarcos, line 257:
```suggestion
To list the devices linked to the authentication method's vault:
```

docs/rfcs/1013-parsec-account.md
```diff
@@ -0,0 +1,456 @@
...
+    "device_id": "DeviceID",
+    "user_id": "UserID",
+    "HumanHandle": "HumanHandle",
+    "created_on": "Timestamp",
+    "revoked_on": "Option<Timestamp>",
+    "encrypted_device": "EncryptedDevice"
+  }
+}
+```
+
```
mmmarcos, line 288:
```suggestion
The encrypted device file is typically ~1KB in size. Considering the number of devices is generally small, it is included here to simplify the API. In the future, a dedicated route may be added specifically to obtain the encrypted device.
          
```

docs/rfcs/1013-parsec-account.md
```diff
@@ -0,0 +1,456 @@
...
+      "algorithm": "KeyAlgorithm",
+      "created_on": "Timestamp",
+      "created_by": "Creator"
+    }
+  }
+}
+```
+
+> [!CAUTION]
+> The returned list does not include device from the current vault (since we already have access to them and do not need to recover them)
```
mmmarcos, line 320: I wouldn't put this in a CAUTION, but as part of the description of the route

```suggestion
The returned list does not include devices from the current vault (since user already has access to them and do not need to recover them). Only the vaults that still contain to-be-recovered protected devices are listed.

Recovery can be done in two ways:

- By remembering the old authentication method's password (since vaults are not removed)
- By recovering the Vault key. For this purpose, we should consider providing a way to export the vault key.
```

docs/rfcs/1013-parsec-account.md
```diff
@@ -0,0 +1,456 @@
...
+
+  > Just inform the user about the revocation
+  >
+  > @touilleMan
+
+- A similar question applies for devices that are no longer used (e.g. a linked browser that is no longer used)
+
+  > The uploaded device in parsec-account are not linked to a specific browser, so this does not apply
+  >
+  > @FirelightFlagboy
```
mmmarcos, line 459: Is this intended to be part of the RFC?

mmmarcos (commented):
LGTM overall :+1: 

I only miss the summary section listing proposed routes. It might be worth adding it back.

docs/rfcs/1013-parsec-account.md
FirelightFlagboy: Yes

docs/rfcs/1013-parsec-account.md
FirelightFlagboy: So not a standalone link send in the email?

docs/rfcs/1013-parsec-account.md
```diff
@@ -0,0 +1,478 @@
...
+
+## Security/Privacy/Compliance
+
+Consideration should be taken about the method used to save the device on the remote server.
+The service would need to identify the user so we will need to have some information about the user (email mostly).
+
+### Perform clean up when recovering devices
+
+Once a device is recovered (i.e.: Was using an old vault and re-encrypted using the current `vault_key`),
+The server should clean up older devices with matching `organization_id` and `device_id` when processing the `protected_device_upload`.
```
FirelightFlagboy, line 455: **issue:** This is a notable change on how we work with database, traditionally we never remove data from the database.

This is causing some challenge about how the user can recover it's account (https://github.com/Scille/parsec-cloud/pull/10090/files#r2025053545)

docs/rfcs/1013-parsec-account.md
```diff
@@ -0,0 +1,478 @@
...
+    "ok": {}
+  }
+}
+```
+
+Upon receiving this request, the server creates a new [`Vault`] and new [`ProtectedDevice`].
+
+> [!IMPORTANT]
+> Doing a vault key rotation means the other authentication methods would not have access to the new `vault_key`,
+> and therefore would not have access to the devices (which will be encrypted by the new `vault_key`).
```
FirelightFlagboy, line 379: **issue:** If we do not remove the previous protected devices, the other auth method could still have access to them using the previous vault.

docs/rfcs/1013-parsec-account.md
```diff
@@ -0,0 +1,386 @@
...
+#### Account
+
+[`Account`]: #account
+
+```yaml
+label: Account
+type: account
+other_fields:
+  email: EmailStr
+  label: EmailLabel
```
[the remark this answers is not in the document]
FirelightFlagboy: I've added a small paragraph below

docs/rfcs/1013-parsec-account.md
FirelightFlagboy: Conclusion, we will use a token with an authenticated

FirelightFlagboy: done

docs/rfcs/1013-parsec-account.md
```diff
@@ -0,0 +1,507 @@
...
+```bash
+PARSEC-HMAC-BLAKE2B.${id}.${account_email}.${timestamp}.${signature}
+```
+
+The signature is generated like so:
+
+```math
+\begin{gather}
+body\_sha256 = sha256(body) \\
+content = \text{"PARSEC-HMAC-BLAKE2B"} | email | timestamp | body\_sha256 \\
```
AureliaDolo, line 99: What does | mean in this context ?

docs/rfcs/1013-parsec-account.md
```diff
@@ -0,0 +1,507 @@
...
+    }
+  ]
+}
+```
+
+On `ok`, the server would have sent a mail with a unique token used for next the request used to register the authentication method.
+
+> The token will be valid for a duration that will be defined at the implementation (likely will be configurable)
+
+If the email is already registered in the service, a `ok` response will still be sent (without sending an email)
```
AureliaDolo, line 214: It's a remark for later, but should we send an email to warn the user that another account creation attempt has occurred ?

docs/rfcs/1013-parsec-account.md
```diff
@@ -0,0 +1,507 @@
...
+    {
+      "status": "invalid_email_validation_token"
+    }
+  ]
+}
+```
+
+Upon receiving the request, the server will create the [`Account`], [`AuthMethod`], and [`Vault`].
+
+The IP address and User-agent of the creator are recorded (see Data Model).
```
AureliaDolo, line 240: Missing link to the section ?

docs/rfcs/1013-parsec-account.md
```diff
@@ -0,0 +1,507 @@
...
+{
+  "cmd": "vault_get",
+  "req": {},
+  "reps": {
+    "ok": {
+      "vault_id": "VaultID",
+      "encrypted_key": "EncryptedSymKey",
+      "created_on": "Timestamp",
+      "created_by": "Creator",
+      # True if using the active vault (latest vault).
```
AureliaDolo, line 256: I'm not sure how we could get not the latest vault here

docs/rfcs/1013-parsec-account.md
```diff
@@ -0,0 +1,507 @@
...
+    # We add the vault_id to limit concurrency issue in case another client perform a vault rotation.
+    "vault_id": "VaultID",
+    "device": "UploadedDevice"
+  },
+  "reps": {
+    "ok": {},
+    # The server compare the human handle of the device to the email of parsec-account.
+    "invalid_human_handle": {},
+    # Also check if `vault_id` correspond to the latest vault in use.
+    "not_using_active_vault": {}
```
AureliaDolo, line 278: nit:
```suggestion
    "inactive_vault": {}
```

AureliaDolo (approved):
The snippets are still marked are yaml, but they look like json now ^^

Also I don't really know where, but I'd make explicit somewhere that all commands are authenticated, expect the account_send_email_validation_token one.

So except a few nits and clarifying question LGTM :tada:

mmmarcos (approved):

docs/rfcs/1013-parsec-account.md
FirelightFlagboy: It's the concatenation operator but wikipedia says it should be `||` instead.
So I will fix this

docs/rfcs/1013-parsec-account.md
FirelightFlagboy: When writing the RFC I though of the same thing but I do not see how actionable the email would be to the user.

docs/rfcs/1013-parsec-account.md
FirelightFlagboy: The link**s** are in the above phrase

docs/rfcs/1013-parsec-account.md
FirelightFlagboy: Currently it will always be true since it will always return the latest known vault. 
I'm anticipating that we add a `id: Option<VaultID>` in the request field later on

docs/rfcs/1013-parsec-account.md
FirelightFlagboy: done

PR #10096, opened 2025-04-03
test(playwright): Retain trace on failure

FirelightFlagboy:
Keep the trace instead of enabling them on first retry to allow investigating the test in case the retry work (meaning flaky) while still keeping trace of retries if they also fail.


Max-7 (approved):

FirelightFlagboy:
`retain-trace-on-failure` take more time since that trace is generated for all tests.
Will require #10097 to take less time (by having less tests to execute by jobs)

PR #10128, opened 2025-04-07
ci(npm): Install dependencies without running audit

FirelightFlagboy:
The audit phase only list know vulnerabilities, we do not use that capabilities during CI run/build.


FirelightFlagboy:
No notable gain here

PR #10130, opened 2025-04-08
build(electron): Use `neon dist` instead of `cargo-cp-artifact`

FirelightFlagboy:


FirelightFlagboy:
Waiting reply from https://github.com/neon-bindings/neon/issues/1098

FirelightFlagboy:
> Waiting reply from [neon-bindings/neon#1098](https://github.com/neon-bindings/neon/issues/1098)

TL,DR: `neon dist` is the new method, but the previous one `cargo-cp-artifact` should still be working

PR #10132, opened 2025-04-08
[MS] Added log API for web

Max-7:
```typescript
// Example
async function sendBugReport(): Promise<void> {
  const mockFiles = await import('@/parsec/mock_files');

  // Query to BMS /feedback (fails currently because invalid data)
  await BmsApi.reportBug(
    { name: 'Max', email: 'maxime.grandcolas@scille.fr', description: 'THIS IS A TEST' },
    { includeLogs: true, files: [{ name: 'hell_yeah.png', mimeType: 'image/png', data: await mockFiles.getMockFileContent(mockFiles.MockFileType.PNG) }]},
  );

  // Open logs in a new tab
  await Env.Links.openLogs();
}
```

- [X] Describe your changes

### DIFF
--- a/client/src/services/bms/api.ts
+++ b/client/src/services/bms/api.ts
@@ -1,5 +1,6 @@
 // Parsec Cloud (https://parsec.cloud) Copyright (c) BUSL-1.1 2016-present Scille SAS
 
+import { getDefaultDeviceName } from '@/common/device';
 import {
   AddPaymentMethodQueryData,
   AuthenticationToken,
@@ -8,6 +9,7 @@ import {
   BmsInvoice,
   BmsOrganization,
   BmsResponse,
+  BugReportQueryData,
   CONNECTION_ERROR_STATUS,
   ChangePasswordQueryData,
   ClientQueryData,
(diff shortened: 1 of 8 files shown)


client/src/services/bms/api.ts
```diff
@@ -719,6 +721,60 @@ async function getCustomOrderInvoices(
...
+      type: 'bug_report',
+      // See if this can be made more precise (OS Version, browser, ...)
+      platform: getDefaultDeviceName(),
+      name: query.name ?? query.email,
+      email: query.email,
+      title: 'Bug Report',
+      description: query.description,
+      version: APP_VERSION,
+      timestamp: DateTime.now().toISO(),
+      logs: opts?.includeLogs ? await window.electronAPI.getLogs() : [],
```
FirelightFlagboy, line 750: **nitpick (non-blocking):** I would have put the logs in an additional file, that would allow for more flexibility for how we handle the form data in the backend.

Max-7: How is it more flexible to add a txt rather than all the lines in an array? The backend is able to create a txt from that if needed.
I could add the logs as a file in addition to the array but I don't see why that would be necessary.

FirelightFlagboy: That would allow to directly stream the file to storage and make the data file smaller

Issue #10076, opened 2025-04-02
[Auth] User authentication via login+password


mmmarcos:

FirelightFlagboy:
Closing, all tasks complete

PR #10148, opened 2025-04-10
feat(client_connection): Generate HMAC signature for authentication

FirelightFlagboy:
Currently the signature is generated with a random key each time until we have access to a stable secret.

It also use an hardcoded email since we do not have an account struct to store the user information.

Related to #10076

### DIFF
--- a/libparsec/crates/client_connection/src/authenticated_account_cmds.rs
+++ b/libparsec/crates/client_connection/src/authenticated_account_cmds.rs
@@ -1,12 +1,13 @@
 // Parsec Cloud (https://parsec.cloud) Copyright (c) BUSL-1.1 2016-present Scille SAS
 
 use bytes::Bytes;
+use data_encoding::BASE64URL;
 use reqwest::{
     header::{HeaderMap, HeaderValue, AUTHORIZATION, CONTENT_LENGTH, CONTENT_TYPE},
     Client, RequestBuilder, Url,
 };
-use std::fmt::Debug;
 use std::path::Path;
+use std::{fmt::Debug, sync::Arc};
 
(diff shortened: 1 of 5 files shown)


libparsec/crates/client_connection/src/authenticated_account_cmds.rs
```diff
@@ -157,12 +174,50 @@ fn prepare_request(
...
+        v.extend_from_slice(auth_id.as_bytes());
+        v.extend_from_slice(email.as_bytes());
+        v.extend_from_slice(timestamp.as_bytes());
+        v.extend_from_slice(body_sha256.as_ref());
+        v
+    };
+    let signature = secret.hmac_full(&content);
+    let b64_signature = BASE64URL.encode(&signature);
+
+    format!("Bearer {PARSEC_AUTH_METHOD}.{auth_id}.{email}.{timestamp}.{b64_signature}")
```
FirelightFlagboy, line 218: **question (non-blocking):** Since the email contain `.` char, should keep it as a raw value or should we encode it (thinking of b64url)

touilleMan: that's a very good point !

I think we should base64 it.

But I'm also wondering if we need to add it: if we provide the  `auth_id` (which I guess corresponds to the authentication method ID), then the server can retrieve the email on its side so need to include it :thinking:

FirelightFlagboy: Currently, `auth_id` is a constant value set to `password`. since we decided to simply parsec-account for now by remove the problem of how the user obtain it's `auth_id` before authenticating

mmmarcos (approved):
LGTM

PR #10173, opened 2025-04-15
feat(server): Parse authentication header for authenticated account request

### LINKED ISSUE #10170: [Account] Server parse authorization header
Need some similar to [`AuthenticationToken`], parse algo type, auth type, b64 email, timestamp, b64 hmac signature.

[`AuthenticationToken`]: https://github.com/Scille/parsec-cloud/blob/5b390f1f59cb21cb57f247498942de348cd6aa83/server/parsec/components/auth.py#L97

FirelightFlagboy:
Closes #10170


server/tests/test_token.py
```diff
@@ -0,0 +1,54 @@
...
+import base64
+import hashlib
+
+from parsec._parsec import DateTime, SecretKey
+from parsec.components.auth import AccountAuthenticationToken, AccountAuthenticationType
+
+
+def gen_account_auth_token() -> tuple[AccountAuthenticationToken, SecretKey, bytes]:
+    email = "foo@example.com"
+    timestamp = DateTime.from_timestamp_seconds(1744721639)
```
AureliaDolo, line 12: It's a specific value ?

server/parsec/components/auth.py
```diff
@@ -155,6 +162,77 @@ def verify_signature(self, verify_key: VerifyKey) -> bool:
...
+            email = urlsafe_b64decode(raw_email).decode()
+            validate_email(email)
+            timestamp = DateTime.from_timestamp_seconds(int(raw_timestamp_us))
+            signature = urlsafe_b64decode(raw_signature)
+        except (
+            ValueError,
+            # Raised when validating email
+            PydanticCustomError,
+        ) as e:
+            raise ValueError("Invalid token format") from e
```
AureliaDolo, line 217: Maybe this would be confusing if the error is raised when the source is the email format.

AureliaDolo (approved):
A few nits, but otherwise LGTM

server/tests/test_token.py
FirelightFlagboy: It was the timestamp at when I written the test, but it does not hold a particular value

server/parsec/components/auth.py
FirelightFlagboy: You prefer that the implementation return a dedicated error message for the email error?

server/parsec/components/auth.py
```diff
@@ -155,6 +162,77 @@ def verify_signature(self, verify_key: VerifyKey) -> bool:
...
+class AccountAuthenticationToken:
+    auth_type: AccountAuthenticationType
+    timestamp: DateTime
+    email: EmailStr
+    header_and_payload: tuple[bytes, ...]
+    signature: bytes
+
+    HEADER = b"PARSEC-HMAC-BLAKE2B"
+
+    @staticmethod
```
touilleMan, line 179: `AuthenticatedToken` has this comment for `generate_raw`: `# Only used for tests, but coherent to have it here`

Should keep the comment if it's still the case

server/parsec/components/auth.py
```diff
@@ -155,6 +162,77 @@ def verify_signature(self, verify_key: VerifyKey) -> bool:
...
+                raise ValueError
+            auth_type = AccountAuthenticationType(raw_auth_type)
+            email = urlsafe_b64decode(raw_email).decode()
+            validate_email(email)
+            timestamp = DateTime.from_timestamp_seconds(int(raw_timestamp_us))
+            signature = urlsafe_b64decode(raw_signature)
+        except (
+            ValueError,
+            # Raised when validating email
+            PydanticCustomError,
```
touilleMan, line 215: [`PydanticCustomError` inherits `ValueError`](https://docs.pydantic.dev/latest/api/pydantic_core/#pydantic_core.PydanticCustomError) so this can be removed

server/parsec/components/auth.py
```diff
@@ -155,6 +162,77 @@ def verify_signature(self, verify_key: VerifyKey) -> bool:
             return False
 
 
+class AccountAuthenticationType(enum.Enum):
+    Password = b"password"
+
+
+@dataclass
+class AccountAuthenticationToken:
```
touilleMan, line 170: I think we shouldn't make this structure generic, and instead only focus on the email-based authentication.

Any non email-based authentication method will need a different cooking anyway (since the email field is replaced by something else, the `generate_raw` will take different argument, and the `from_raw` will cook differently)

So in the end it's just better to:
- call this class `AccountPasswordAuthenticationToken`
- change the `HEADER` for something like `PARSEC-PASSWORD-HMAC-BLAKE2B`

This way it will be trivial to add different authentication type and just know which one to use according to the `HEADER` value

touilleMan (changes requested):

server/tests/test_token.py
mmmarcos: It would be nice to add a comment, it does not cost much and may save time for our future selves.
```suggestion
    // Timestamp from the moment the test was written, does not hold a particular value 
    timestamp = DateTime.from_timestamp_seconds(1744721639)
```
Or just use a plain readable date as with the other dummy values :)
```suggestion
    timestamp = DateTime.from_ymd_hms_us(2025, ...)
```

FirelightFlagboy: Going with the second suggestion

server/parsec/components/auth.py
```diff
@@ -155,6 +162,77 @@ def verify_signature(self, verify_key: VerifyKey) -> bool:
...
+    @staticmethod
+    def generate_raw(
+        auth_type: AccountAuthenticationType,
+        timestamp: DateTime,
+        email: EmailStr,
+        body: bytes,
+        key: SecretKey,
+    ) -> bytes:
+        raw_timestamp_us = str(timestamp.as_timestamp_seconds()).encode("ascii")
+        raw_email = urlsafe_b64encode(email.encode())
```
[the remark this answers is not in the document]
FirelightFlagboy: done

server/parsec/components/auth.py
```diff
@@ -155,6 +162,77 @@ def verify_signature(self, verify_key: VerifyKey) -> bool:
...
+    @classmethod
+    def from_raw(cls, raw: bytes) -> "AccountAuthenticationToken":
+        try:
+            raw_header_and_payload, raw_signature = raw.rsplit(b".", 1)
+            header_and_payload = raw_header_and_payload.split(b".")
+            header, raw_auth_type, raw_email, raw_timestamp_us = header_and_payload
+            if header != cls.HEADER:
+                raise ValueError
+            auth_type = AccountAuthenticationType(raw_auth_type)
+            email = urlsafe_b64decode(raw_email).decode()
```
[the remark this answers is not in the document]
FirelightFlagboy: done

server/parsec/components/auth.py
FirelightFlagboy: done

server/parsec/components/auth.py
```diff
@@ -155,6 +162,77 @@ def verify_signature(self, verify_key: VerifyKey) -> bool:
...
+        raw_timestamp_us = str(timestamp.as_timestamp_seconds()).encode("ascii")
+        raw_email = urlsafe_b64encode(email.encode())
+        header_and_payload = (
+            AccountAuthenticationToken.HEADER,
+            auth_type.value,
+            raw_email,
+            raw_timestamp_us,
+        )
+        body_sha256 = hashlib.sha256(body).digest()
+        raw_signature = key.hmac_full(b"".join(header_and_payload) + body_sha256)
```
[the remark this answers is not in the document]
FirelightFlagboy: done

server/parsec/components/auth.py
FirelightFlagboy: done

server/parsec/components/auth.py
```diff
@@ -155,6 +162,77 @@ def verify_signature(self, verify_key: VerifyKey) -> bool:
             return False
 
 
+class AccountAuthenticationType(enum.Enum):
+    Password = b"password"
```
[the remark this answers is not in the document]
FirelightFlagboy: The enum no longer exist

server/parsec/components/auth.py
```diff
@@ -155,6 +162,77 @@ def verify_signature(self, verify_key: VerifyKey) -> bool:
...
+
+    @classmethod
+    def from_raw(cls, raw: bytes) -> "AccountAuthenticationToken":
+        try:
+            raw_header_and_payload, raw_signature = raw.rsplit(b".", 1)
+            header_and_payload = raw_header_and_payload.split(b".")
+            header, raw_auth_type, raw_email, raw_timestamp_us = header_and_payload
+            if header != cls.HEADER:
+                raise ValueError
+            auth_type = AccountAuthenticationType(raw_auth_type)
```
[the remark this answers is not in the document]
FirelightFlagboy: The enum no longer exist

server/parsec/components/auth.py
```diff
@@ -155,6 +162,77 @@ def verify_signature(self, verify_key: VerifyKey) -> bool:
...
+class AccountAuthenticationType(enum.Enum):
+    Password = b"password"
+
+
+@dataclass
+class AccountAuthenticationToken:
+    auth_type: AccountAuthenticationType
+    timestamp: DateTime
+    email: EmailStr
+    header_and_payload: tuple[bytes, ...]
```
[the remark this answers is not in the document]
FirelightFlagboy: done

touilleMan (approved):

server/parsec/components/auth.py
```diff
@@ -155,6 +160,66 @@ def verify_signature(self, verify_key: VerifyKey) -> bool:
...
 
 
+@dataclass
+class AccountPasswordAuthenticationToken:
+    timestamp: DateTime
+    email: EmailStr
+    header_and_payload: bytes
+    signature: bytes
+
+    HEADER = b"PARSEC-PASSWORD-HMAC-BLAKE2B"
```
touilleMan, line 170: Can you also update the RFC since the header value has changed

PR #10174, opened 2025-04-15
[account] Implement `account_send_email_validation_token` route.

### LINKED ISSUE #9956: [Auth] Account creation (email + password)

AureliaDolo:
Part of #9956


libparsec/crates/protocol/schema/anonymous_account_cmds/account_send_email_vaidation_token.json5
AureliaDolo, line 1: I'll copy paste that in the rfc

server/parsec/api.py
```diff
@@ -74,7 +81,7 @@ def api(fn: ApiFnWithSelf) -> ApiFnWithSelf:  # pyright: ignore[reportMissingTyp
         case _:
             raise ValueError(f"unexpected family {m_family}")
     assert types["client_ctx"] == ForwardRef(expected_type_name), (
-        f"Expect `client_ctx` to be an `{expected_type_name}`"
+        f"Expect `client_ctx`({types['client_ctx']}) to be an `{expected_type_name}`"
```
AureliaDolo, line 84: I was a bit stuck on that 
```
E   AssertionError: Expect `client_ctx`(<class 'parsec.client_context.AnonymousAccountClientContext'>) to be an `AnonymousAccountClientContext
````

because `from __future__ import annotations` was missing from account.py 

What do y'all think about adding that, as I'm quite sure it's not the first time I stumble on this
```suggestion
        f"Expect `client_ctx`({types['client_ctx']}) to be an `{expected_type_name}.` Did you add `from __future__ import annotations` ?"
```

server/parsec/components/memory/account.py
```diff
@@ -1,9 +1,15 @@
 # Parsec Cloud (https://parsec.cloud) Copyright (c) BUSL-1.1 2016-present Scille SAS
+
+
```
AureliaDolo, line 3: Remove line

server/parsec/components/memory/account.py
```diff
@@ -17,7 +23,11 @@ async def create(self, email: str) -> None | CreateAccountBadOutcome:
         if email in self._data.accounts:
             return CreateAccountBadOutcome.ACCOUNT_ALREADY_EXISTS
         else:
-            self._data.accounts[email] = MemoryAccount(email)
+            try:
+                user_email = UserAccountDataValidationModel(user_email=email)
+                self._data.accounts[email] = MemoryAccount(user_email=user_email.user_email)
```
AureliaDolo, line 28: @FirelightFlagboy I'm not exactly sure that's the canonical way to use pydantic for data validation

server/parsec/components/memory/datamodel.py
```diff
@@ -855,4 +855,8 @@ class MemoryShamirShare:
 
 @dataclass(slots=True)
 class MemoryAccount:
-    user_email: str | None
+    user_email: EmailStr
+
+
+class UserAccountDataValidationModel(BaseModel):
```
AureliaDolo, line 861: Used for data validation, I'm not sure that's the right way to do that

server/parsec/components/postgresql/account.py
```diff
@@ -1,4 +1,6 @@
 # Parsec Cloud (https://parsec.cloud) Copyright (c) BUSL-1.1 2016-present Scille SAS
+
+
```
AureliaDolo, line 3: Remove line

server/tests/api_v5/anonymous_account/test_account_send_email_validation.py
```diff
@@ -0,0 +1,45 @@
...
+from parsec._parsec import anonymous_account_cmds
+from tests.common import Backend, HttpCommonErrorsTester
+from tests.common.client import AccountRpcClient
+
+
+async def test_anonymous_account_account_send_email_validation_token_ok(
+    account: AccountRpcClient, backend: Backend, with_postgresql: bool
+) -> None:
+    if with_postgresql:
+        pytest.xfail("TODO: postgre not implemented yet")
```
mmmarcos, line 14: You can avoid these `if`s with some handy fixtures from `conftest.py`:

https://github.com/Scille/parsec-cloud/blob/0327720b970b9217bc1e9e1ea8c366022776236b/server/tests/conftest.py#L251-L260

server/parsec/components/account.py
```diff
@@ -27,3 +32,20 @@ async def test_new_account(self) -> str:
...
+        self,
+        client_ctx: AnonymousAccountClientContext,
+        req: anonymous_account_cmds.latest.account_send_email_validation_token.Req,
+    ) -> anonymous_account_cmds.latest.account_send_email_validation_token.Rep:
+        outcome = await self.create(req.email)
+        match outcome:
+            case CreateAccountBadOutcome.INVALID_EMAIL:
+                return anonymous_account_cmds.latest.account_send_email_validation_token.RepInvalidEmail()
+            case CreateAccountBadOutcome.ACCOUNT_ALREADY_EXISTS:
+                # Do nothing
```
mmmarcos, line 47:
```suggestion
                # Respond OK without sending token to prevent creating 
                # an oracle of email addresses registered in the service.
```

server/parsec/components/memory/datamodel.py
FirelightFlagboy: You can add a test with a invalid email as `user_email`, I except `pydantic` to raise an exception.
But since they also provide `validate_email` we may have some surprise.

Anyway the best email validation is when we can send email to that address :shrug:

server/parsec/components/memory/account.py
FirelightFlagboy, line 28: **todo:** Use the same key

```suggestion
                self._data.accounts[user_email.user_email] = MemoryAccount(user_email=user_email.user_email)
```

server/parsec/components/memory/datamodel.py
FirelightFlagboy: Concerning the usage of that class, you want to use [`TypeAdapter`](https://docs.pydantic.dev/latest/concepts/type_adapter/) instead.

server/parsec/components/memory/account.py
```diff
@@ -17,7 +23,12 @@ async def create(self, email: str) -> None | CreateAccountBadOutcome:
         if email in self._data.accounts:
             return CreateAccountBadOutcome.ACCOUNT_ALREADY_EXISTS
         else:
-            self._data.accounts[email] = MemoryAccount(email)
+            try:
+                email_adapter = TypeAdapter(EmailStr)
```
AureliaDolo, line 27: @FirelightFlagboy what do you think ?

server/tests/api_v5/anonymous_account/test_account_send_email_validation.py
AureliaDolo: There are a few other places where this is the case, see #10200

libparsec/crates/protocol/schema/anonymous_account_cmds/account_send_email_vaidation_token.json5
FirelightFlagboy, line 1: **todo:** format with `jq .`

docs/rfcs/1013-parsec-account.md
```diff
@@ -203,21 +203,32 @@ The `label` is not used by Parsec Account server but will be used by the client
...
-    {
-      "status": "ok"
+```json
+[
+  {
+    "major_versions": [
+            5
+        ],
+    "cmd": "account_send_email_validation_token",
+    "req": { "fields":
```
FirelightFlagboy, line 213: **todo:** format with `jq .`

server/parsec/components/memory/account.py
```diff
@@ -17,7 +23,12 @@ async def create(self, email: str) -> None | CreateAccountBadOutcome:
...
+                email_parsed = email_adapter.validate_python(email)
+                self._data.accounts[email_parsed] = MemoryAccount(email_parsed)
```
FirelightFlagboy, line 29: **nitpick:** Only the `validate_python` is expected to fail so a would just wrap that in the `try/except`

server/parsec/components/memory/account.py
FirelightFlagboy: That what a was expecting, I would move it at the top of the file and name it like a type (`EmailAdapter`)

server/parsec/components/memory/account.py
FirelightFlagboy, line 29: **issue (blocking):** The created account should be marked a not verified.

Since the creation is only complete with the email token & the first authentication method.
We surely want to be able to perform some cleanup of not-verified account in the future.

We may want to use a dedicated `NotVerifiedAccount` and promote it to `Account` :thinking: (The reason to use a dedicated type is to not use an account that is not verified where it should have been)

mmmarcos: This may need further discussion as the RFC 1013 does not specifically address how to manage unverified emails. 

Some things to keep in mind:

1. The user verifies an email address, _not_ an account. This subtle difference will probably guide our data model design, as we could store separately: 
  - unverified_emails (temporarily?)
  - accounts (user has verified they has access to the email address)

2. An account shouldn't probably be created before email verification.

3. RFC says that token should not be sent if account already exists, but what if the email has not yet been verified? In this case, the server should re-send email otherwise, if the user has lost/deleted the original mail, it won't be able to create an account.

server/tests/api_v5/anonymous_account/test_account_send_email_validation.py
```diff
@@ -0,0 +1,37 @@
+# Parsec Cloud (https://parsec.cloud) Copyright (c) BUSL-1.1 2016-present Scille SAS
+
+
+from parsec._parsec import anonymous_account_cmds
+from tests.common import Backend, HttpCommonErrorsTester
+from tests.common.client import AccountRpcClient
+
+
+async def test_anonymous_account_account_send_email_validation_token_ok(
+    account: AccountRpcClient, backend: Backend, skip_if_postgresql: None
```
mmmarcos, line 10: I'd use `xfail_if_postgresql` which indicates this is temporary (see message in the fixture). We use `skip` for permanent skips.

```suggestion
    account: AccountRpcClient, backend: Backend, xfail_if_postgresql: None
```

server/parsec/components/memory/account.py
touilleMan: In RFC 1013, we are clear about having a `account_send_email_validation_token` route that must first be used to obtain a token proving the email address is owned.
The token is then used in the `account_password_create` to actually create the account.

So there is no concept of not verified account, all account are verified since creating them requires to provide the proof the email has been verified.

>  The user verifies an email address, not an account. This subtle difference will probably guide our data model design, as we could store separately:
>  - unverified_emails (temporarily?)
>  - accounts (user has verified they has access to the email address)

Yes this is totally the idea ! This should simplify the implementation by avoiding states for the account, and instead having a simple separated table keeping track of email / validation token correspondence.

> RFC says that token should not be sent if account already exists, but what if the email has not yet been verified?

If the email hasn't been verified, then the token should be re-send (otherwise a user deleting the email containing the verification token before using it would no longer be able to register to Parsec account !)

server/parsec/api.py
```diff
@@ -74,7 +81,7 @@ def api(fn: ApiFnWithSelf) -> ApiFnWithSelf:  # pyright: ignore[reportMissingTyp
         case _:
             raise ValueError(f"unexpected family {m_family}")
     assert types["client_ctx"] == ForwardRef(expected_type_name), (
-        f"Expect `client_ctx` to be an `{expected_type_name}`"
+        f"Expect `client_ctx`({types['client_ctx']}) to be an `{expected_type_name}`. Did you add `from __future__ import annotations` ?"
```
touilleMan, line 84: I don't understand the ``` Did you add `from __future__ import annotations` ? ``` part, can you explain ?

server/parsec/components/account.py
```diff
@@ -1,17 +1,22 @@
...
 
 
 class CreateAccountBadOutcome(BadOutcomeEnum):
     ACCOUNT_ALREADY_EXISTS = auto()
+    INVALID_EMAIL = auto()
 
 
 class BaseAccountComponent:
-    async def create(self, email: str) -> None | CreateAccountBadOutcome:
+    async def create_token(self, email: str) -> EmailValidationToken | CreateAccountBadOutcome:
```
touilleMan, line 19:
```suggestion
    async def create_email_valitation_token(self, email: str) -> EmailValidationToken | CreateAccountBadOutcome:
```

Better be more explicit about what this does since there is plenty of different token in this codebase ^^

server/tests/api_v5/anonymous_account/test_account_send_email_validation.py
```diff
@@ -0,0 +1,37 @@
+# Parsec Cloud (https://parsec.cloud) Copyright (c) BUSL-1.1 2016-present Scille SAS
+
+
+from parsec._parsec import anonymous_account_cmds
+from tests.common import Backend, HttpCommonErrorsTester
+from tests.common.client import AccountRpcClient
+
+
```
touilleMan, line 8: You should add a `test_anonymous_account_account_send_email_validation_token_send_email_bad_outcome` working similarly to the existing test `test_authenticated_invite_new_device_send_email_bad_outcome` to test what happen if the email cannot be send

libparsec/crates/protocol/schema/anonymous_account_cmds/account_send_email_vaidation_token.json5
```diff
@@ -0,0 +1,24 @@
...
+          "type": "String"
+        }
+      ]
+    },
+    "reps": [
+      {
+        "status": "ok"
+      },
+      {
+        "status": "invalid_email"
```
touilleMan, line 20: I think we missed the error status related to the email not being able to be sent.

In the `invite_new_user`/`invite_new_device` commands we have a `email_sent` field in the result with the following possibilities:
- `SUCCESS`
- `SERVER_UNAVAILABLE`
- `RECIPIENT_REFUSED`


The need for a dedicated field in the success result is needed in those command since an invitation can be successfully created even if the email fails to be sent. However this is not something irrelevant to `account_send_email_validation_token` since the whole point of this command is to send an email !

So I think we should have:
- add a `email_server_unavailable`
- rename `invalid_email` to `email_recipient_refused`

server/parsec/components/account.py
```diff
@@ -24,6 +29,23 @@ async def test_new_account(self) -> str:
...
+        req: anonymous_account_cmds.latest.account_send_email_validation_token.Req,
+    ) -> anonymous_account_cmds.latest.account_send_email_validation_token.Rep:
+        outcome = await self.create_token(req.email)
+        match outcome:
+            case CreateAccountBadOutcome.INVALID_EMAIL:
+                return anonymous_account_cmds.latest.account_send_email_validation_token.RepInvalidEmail()
+            case CreateAccountBadOutcome.ACCOUNT_ALREADY_EXISTS:
+                # Respond OK without sending token to prevent creating
+                return anonymous_account_cmds.latest.account_send_email_validation_token.RepOk()
+            case EmailValidationToken():  # as token:
```
touilleMan, line 49: For consistency with the rest of the codebase, can you move the happy case (i.e. `EmailValidationToken()`) first ?

server/parsec/components/memory/account.py
```diff
@@ -13,11 +21,17 @@ def __init__(self, data: MemoryDatamodel, event_bus: EventBus) -> None:
...
 
     @override
-    async def create(self, email: str) -> None | CreateAccountBadOutcome:
+    async def create_token(self, email: str) -> EmailValidationToken | CreateAccountBadOutcome:
         if email in self._data.accounts:
             return CreateAccountBadOutcome.ACCOUNT_ALREADY_EXISTS
         else:
-            self._data.accounts[email] = MemoryAccount(email)
+            try:
+                email_parsed = EmailAdapter.validate_python(email)
```
touilleMan, line 29: Note this is not how we handle the email address for `invite_new_user`/`invite_new_device` (i.e. we don't do any validation and pass it directly to Python `smtplib`, which in turn [only complain](https://github.com/python/cpython/blob/293f2895d577dbabcd7202950ddb6089ea70b959/Lib/smtplib.py#L153) if it [contains UTF8](https://github.com/python/cpython/blob/293f2895d577dbabcd7202950ddb6089ea70b959/Lib/smtplib.py#L959-L964))

It is important to be consistent across the codebase on what validation strategy we use. So we should either:
- Don't do validation here
- Add validation to  `invite_new_user`/`invite_new_device`

I'd think it's better to go with option 2. Typically we could introduce a dedicated email type in the serialization schema, this way we simply wouldn't have to bother with doing manual checks (and risking forgetting them ^^)

One additional benefit is to avoid to rely on two different way to validate the same things (since it could lead to different results for the same input, which attackers typically like to abuse...):
- In libparsec we have `HumanHandle::email_is_valid` that relies on the `email-address-parser` crate
- In Python we use the `pydantic` module

Can you open an issue about this ?

touilleMan (changes requested):

server/parsec/api.py
AureliaDolo: before I added that to the account component it was failing because client ctx was a forward ref, not the actual client context. I spend more time that I wanted to on that, and I'm pretty sure I already encountered this error while setting up the shamir component. Maybe I can just put a comment describing that  before the assert and not include it in the error message.

server/parsec/components/account.py
AureliaDolo: Indeed ^^

server/parsec/api.py
FirelightFlagboy: The original discussion: https://github.com/Scille/parsec-cloud/pull/10174#discussion_r2044984114

libparsec/crates/protocol/schema/anonymous_account_cmds/account_send_email_vaidation_token.json5
FirelightFlagboy: off-topic: In case the email cannot be send, should the server perform retries in background ?

libparsec/crates/types/tests/unit/token.rs
```diff
@@ -98,3 +98,4 @@ macro_rules! base_token_tests {
 
 base_token_tests!(invitation_token, InvitationToken);
 base_token_tests!(bootstrap_token, BootstrapToken);
+base_token_tests!(account_token, EmailValidationToken);
```
FirelightFlagboy, line 101: **nitpick:** It's `EmailVAlidationToken` in snake_case

```suggestion
base_token_tests!(email_validation_token, EmailValidationToken);
```

server/parsec/components/memory/account.py
FirelightFlagboy, line 29: **nitpick:** Move this validation step in the base implementation

`create_token` should only create the token (and provided with an already validated token) and let the base implementation do the rest

server/parsec/components/account.py
mmmarcos: I would also rename the `BadOutcome` for consistency:

 ```suggestion
    async def create_email_validation_token(self, email: str) -> EmailValidationToken | CreateEmailValidationTokenBadOutcome:
```

**Comment by touilleMan** [comment]
> off-topic: In case the email cannot be send, should the server perform retries in background ?

We should follow what is already done for `invite_new_{user|device}`

File: server/parsec/components/account.py [hunk 19]
```diff
@@ -24,6 +29,23 @@ async def test_new_account(self) -> str:
...
+                # TODO send email with token
```
mmmarcos, line 50: This TODO is tracked in #9956 as part of the added route. Will you be adding it to this PR?

mmmarcos (commented):
LGTM overall, although the actual email send is missing? :thinking:   

Note that there is a typo in the commit message (_va**l**idation_): [Add account_send_email_vaidation_token route.](https://github.com/Scille/parsec-cloud/pull/10174/commits/e1c99b2b9afe9abf3642136dd2467d08fec7f801)

server/parsec/components/account.py
AureliaDolo: As it's its own can of worms, and we want to keep PRs small I'd rather do it separately

libparsec/crates/protocol/schema/anonymous_account_cmds/account_send_email_vaidation_token.json5
AureliaDolo: And should we have an option to not actually send the email ? For example for tests purpose ?

AureliaDolo: to be continued in #10216

mmmarcos (approved):
LGTM

server/parsec/api.py
touilleMan: > Maybe I can just put a comment describing that before the assert and not include it in the error message.

Yes, just adding as comment with the error message from https://github.com/Scille/parsec-cloud/pull/10174#discussion_r2044984114 is a good idea

libparsec/crates/protocol/schema/anonymous_account_cmds/account_send_email_vaidation_token.json5
touilleMan: > off-topic: In case the email cannot be send, should the server perform retries in background ?

We don't do that in user/device invite, so no (on top of that we haven't had any reported issue regarding email so far, so I'd say there is no need opening an issue about that either)

touilleMan: > And should we have an option to not actually send the email ? For example for tests purpose ?

This is already covered by the fact we can [configure a mock email sender](https://github.com/Scille/parsec-cloud/blob/20cf7387df909aaee2e9bb36b5c78ee7b0a99f66/server/parsec/config.py#L167)

touilleMan (approved):

PR #10178, opened 2025-04-16
Add fallback code to support web when shared worker is not available

touilleMan:


client/src/plugins/libparsec/trampoline-web.ts
```diff
@@ -108,3 +126,117 @@ export async function LoadWebLibParsecPlugin(): Promise<any> {
...
+        };
+      }
+
+      if (name === 'clientStop') {
+        return async (client: Handle): Promise<Result<null, ClientStopError>> => {
+          const ret = (await target.clientStop(client)) as Result<null, ClientStopError>;
+
+          if (ret.ok) {
+            releaseTabLock(undefined, client);
+          }
```
FirelightFlagboy, line 229: **nitpick:** 
```suggestion
          // We only release the lock if `libparsec` was able to stop the client
          // else we consider that it's still in use and the device is not ready to start a new client
          if (ret.ok) {
            releaseTabLock(undefined, client);
          }
```

client/src/plugins/libparsec/trampoline-web.ts
```diff
@@ -108,3 +126,117 @@ export async function LoadWebLibParsecPlugin(): Promise<any> {
...
+
+async function withoutSharedWorker(): Promise<any> {
+  await init_module();
+  module.initLogger();
+
+  // Map of key path + callback to release the lock
+  // Note only the locks that have been acquired in the current tab are present,
+  // so you shouldn't look into this map to check if a lock is already taken !
+  const tabLocksWithReleaser: Map<string, (val: any) => void> = new Map();
+  const tabLocks: Array<[Handle, string, (val: any) => void]> = [];
```
FirelightFlagboy, line 138: **question:** Where this array modified?

I see that we remove value from it in `releaseTabLock` but I didn't see where we add value

client/src/plugins/libparsec/trampoline-web.ts
FirelightFlagboy, line 1: **nitpick:** 

image

client/src/plugins/libparsec/trampoline-web.ts
touilleMan: Good catch ! I switched from a `Map` to an array half way into the code, and forget to fix some parts :/

fixed.

client/src/plugins/libparsec/trampoline-web.ts
```diff
@@ -108,3 +134,150 @@ export async function LoadWebLibParsecPlugin(): Promise<any> {
...
+          x[0] = handle;
+        }
+        return true;
+      }
+      return false;
+    });
+  }
+
+  function isLockFromThisTab(keyFile: string): boolean {
+    for (let i = 0; i < tabLocks.length; i += 1) {
```
FirelightFlagboy, line 181: **todo:** use `tabLockes.find`

client/src/plugins/libparsec/trampoline-web.ts
```diff
@@ -108,3 +134,150 @@ export async function LoadWebLibParsecPlugin(): Promise<any> {
...
+  function registerTabLockClientHandle(keyFile: string, handle: Handle): void {
+    tabLocks.find((x) => {
+      if (x[1] === keyFile) {
+        if (x[0] === undefined) {
+          x[0] = handle;
+        }
+        return true;
+      }
+      return false;
+    });
```
FirelightFlagboy, line 177: **nitpick:** I would use use `Object.assign` instead: https://stackoverflow.com/a/66017339

It's strange to mutate a value in `find()` (at least in rust we have `filter_map`)

client/src/plugins/libparsec/trampoline-web.ts
```diff
@@ -108,3 +134,150 @@ export async function LoadWebLibParsecPlugin(): Promise<any> {
...
+  await init_module();
+  module.initLogger();
+
+  // Array of [<started client handle>, <keyFile>, <callback to release the lock>]
+  // Note the client handle can be undefined since we first acquire the lock, then
+  // start the client, and only then update the lock info to register the handle.
+  //
+  // Note only the locks that have been acquired in the current tab are present,
+  // so you shouldn't look into this array to check if a lock is already taken !
+  const tabLocks: Array<[Handle | undefined, string, (val: any) => void]> = [];
```
FirelightFlagboy, line 148: **nitpick:** I would put object in it so we have named field instead

```suggestion
  interface TabLockInfo {
    handle?: Handle,
    keyFile: string,
    releaseLock(val: any) => void
  }
  const tabLocks: Array<TabLockInfo> = [];
```

client/src/plugins/libparsec/trampoline-web.ts
touilleMan: I've replaced the `find` by a good old for loop, this way both `releaseTabLock` & `registerTabLockClientHandle` has similar code.

(on top of being pretty exotic, using `Object.assign` seems rather cumbersome since it would require to test the output of `find` to bail out in case of `undefined`)

client/src/plugins/libparsec/trampoline-web.ts
touilleMan: That was my idea at fist, however there is no quick way to unwrap the result

So the code would look like this:

```javascript
const found = tabLocks.find((x) => x[1] === keyFile);
if (found === undefined) {
  return false;
} else {
  return true;
}
```

Which I don't find better, so I stick with the good old C++ style ;-p

client/src/plugins/libparsec/trampoline-web.ts
touilleMan: I've added a test \o/

Max-7 (changes requested):
Does not work on Chrome on Android (tested with 136 & 135. I managed to get the logs with remote debugging:

```
GET https://saas-demo-v3-mightyfairy.parsec.cloud/libparsec_bindings_web_bg-51ba43bb.wasm 404 (Not Found)
`WebAssembly.instantiateStreaming` failed because your server does not serve Wasm with `application/wasm` MIME type. Falling back to `WebAssembly.instantiate` which is slower. Original error:
 TypeError: Failed to execute 'compile' on 'WebAssembly': HTTP status code is not ok
Uncaught (in promise) CompileError: WebAssembly.instantiate(): expected magic word 00 61 73 6d, found 3c 21 64 6f @+0
```
It tries to get the `libparsec.wasm` at `/` when it should be `/client/assets` in this case.

Max-7 (approved):

FirelightFlagboy (approved):

Issue #10204, opened 2025-04-23
Missing sequester service tests


mmmarcos:
The following tests are marked `xfail` even though sequester support was added in v3.3.0:

- rejected_by_sequester_service
  - `test_authenticated_vlob_create_rejected_by_sequester_service`
  - `test_authenticated_vlob_update_rejected_by_sequester_service`
  - `test_authenticated_realm_rotate_key_rejected_by_sequester_service`

- sequester_service_unavailable 
  - `test_authenticated_vlob_create_sequester_service_unavailable` 
  - `test_authenticated_vlob_update_sequester_service_unavailable`
  - `test_authenticated_realm_rotate_key_sequester_service_unavailable`

touilleMan:
This is intended.

The sequester service can be used for two things:
1. The basic use case where the sequester service at a specific time to consult existing workspace data.
2. The "anti-virus" style use case, where the sequester service is used continuously to scan the data right before the server store them, and potentially tell the server to reject them.

For the 2 use case to work, the sequester service must be configured with a webhook that is used by the server to notify the 3rd party service (which has access to the sequester service private key) that some new data are to be scanned.
However this feature has been left on the side for the moment since it is not strictly needed and add some complexity (mostly the fact that we shouldn't use the webhook from within a SQL transaction to avoid slowing down the whole system)

mmmarcos:
OK, so I've just found #6259 which I guess refer to these tests. 

Since there is now more info here, I would close #6259 and keep this issue open.

PR #10214, opened 2025-04-25
Use test fixture xfail_if_postgresql

### LINKED ISSUE #10200: [Clean up] use skip if postgre when neeed
You can avoid these `if`s with some handy fixtures from `conftest.py`:

https://github.com/Scille/parsec-cloud/blob/0327720b970b9217bc1e9e1ea8c366022776236b/server/tests/conftest.py#L251-L260

_Originally posted by @mmmarcos in https://github.com/Scille/parsec-cloud/pull/10174#discussion_r2047022347_
            

Here are the incriminated files
- [ ] anonymous account test_ping
- [ ] authenticated account test ping
- [x] test realm rotate key
- [x] test vlob create
- [x] test vlob update

mmmarcos:
Remaining cases are fixed by #10174.

Fix #10200

### DIFF
--- a/server/tests/api_v5/authenticated/test_realm_rotate_key.py
+++ b/server/tests/api_v5/authenticated/test_realm_rotate_key.py
@@ -560,11 +560,8 @@ async def test_authenticated_realm_rotate_key_rejected_by_sequester_service(
     sequestered_org: SequesteredOrgRpcClients,
     backend: Backend,
     monkeypatch: pytest.MonkeyPatch,
-    with_postgresql: bool,
+    xfail_if_postgresql: None,
 ) -> None:
-    if with_postgresql:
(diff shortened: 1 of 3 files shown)


FirelightFlagboy (approved):

FirelightFlagboy:
We loose the comment about why it's marked `xfail` should we keep that somewhere?

mmmarcos:
> We loose the comment about why it's marked `xfail` should we keep that somewhere?

I don't think is that necessary since there is now #10204 that tracks what it needs to be done. Do you prefer to add a link to the issue as a comment?

FirelightFlagboy:
> > We loose the comment about why it's marked `xfail` should we keep that somewhere?
> 
> I don't think is that necessary since there is now #10204 that tracks what it needs to be done. Do you prefer to add a link to the issue as a comment?

Not necessarily

PR #10226, opened 2025-04-28
[MS] Custom order requests synchro bms

### LINKED ISSUE #10201: OrderPage, update the Custom Request from API
Waiting for https://github.com/Scille/parsec-bms/issues/1111.

Update OrderPage depending the method used in getCustomRequest from the API and update the mockAPI to test all cases.

fabienSvtr:
## Checklist

- [ ] Describe your changes
   
closes #10201


client/src/services/bms/mockApi.ts
```diff
@@ -97,7 +97,7 @@ function createCustomOrderInvoices(
       type: DataType.CustomOrderDetails,
       id: i,
       link: `https://unknown/link${i}.pdf`,
-      number: `Invoice_${i}`,
+      number: `156740546545645${i}`,
```
Max-7, line 100: Why this change?

client/src/components/client-area/OrderInProgress.vue
```diff
@@ -6,10 +6,11 @@
     class="order-progress"
     :class="{ open: open }"
   >
+    {{ request.orderDate }}
```
Max-7, line 9: Just displayed here without any formatting?

client/src/components/client-area/OrderInProgress.vue
```diff
@@ -212,8 +260,9 @@
         <!-- invoice details -->
         <div
           v-if="
-            getStatus(customOrderBmsStatus, customOrderSellsyStatus)?.stepName === (StatusStep.Confirmed || StatusStep.InvoiceToBePaid) &&
-            orderDetails
+            orderDetails &&
+            (getStatus(customOrderBmsStatus, customOrderSellsyStatus)?.stepName === StatusStep.Confirmed ||
```
Max-7, line 264: You can use `[StatusStep.Confirmed, StatusStep.InvoiceToBePaid].includes(getStatus(customOrderBmsStatus, customOrderSellsyStatus)?.stepName)`. Makes it a bit easier to add or remove valide status.

client/src/components/client-area/OrderInProgress.vue
```diff
@@ -68,8 +111,10 @@
         <div
           class="details-list"
           v-if="
-            getStatus(customOrderBmsStatus, customOrderSellsyStatus)?.stepName === (StatusStep.Confirmed || StatusStep.Available) &&
-            orderDetails
+            orderDetails &&
+            (getStatus(customOrderBmsStatus, customOrderSellsyStatus)?.stepName === StatusStep.Confirmed ||
```
Max-7, line 115: You can use the same technique here.

client/src/components/client-area/OrderInProgress.vue
```diff
@@ -36,7 +37,7 @@
 
     <!-- order details -->
     <div
-      v-if="open && orderDetails?.type === DataType.CustomOrderDetails"
+      v-if="open && request"
```
Max-7, line 40: Why would request not be defined? It's a non-optional prop.

client/src/components/client-area/OrderInProgress.vue
```diff
@@ -6,10 +6,11 @@
     class="order-progress"
     :class="{ open: open }"
   >
+    {{ request.orderDate }}
     <!-- order header -->
     <div
       class="order-header"
-      v-if="customOrderBmsStatus"
+      v-if="customOrderBmsStatus && request"
```
Max-7, line 13: Same here, request is a non-optional prop.

Max-7 (changes requested):

Max-7 (commented):
I'm also surprised that you didn't have to update some e2e tests for this. We're probably missing some, they need to be added.

client/src/services/bms/mockApi.ts
fabienSvtr: Just to test with a number what it looks like but I can undo the modification.

client/src/components/client-area/OrderInProgress.vue
fabienSvtr: I removed it.

client/src/components/client-area/OrderInProgress.vue
```diff
@@ -68,8 +109,9 @@
         <div
           class="details-list"
           v-if="
-            getStatus(customOrderBmsStatus, customOrderSellsyStatus)?.stepName === (StatusStep.Confirmed || StatusStep.Available) &&
-            orderDetails
+            orderDetails &&
+            [StatusStep.Confirmed, StatusStep.InvoiceToBePaid, StatusStep.Available].includes(
+              getStatus(customOrderBmsStatus, customOrderSellsyStatus)?.stepName ?? StatusStep.Unknown)
```
Max-7, line 114: You don't need the `?? StatusStep.Unknown`. `?.stepName` can be unknown, which will not be in the array.

client/src/components/client-area/OrderInProgress.vue
```diff
@@ -212,8 +257,9 @@
         <!-- invoice details -->
         <div
           v-if="
-            getStatus(customOrderBmsStatus, customOrderSellsyStatus)?.stepName === (StatusStep.Confirmed || StatusStep.InvoiceToBePaid) &&
-            orderDetails
+            orderDetails &&
+            [StatusStep.Confirmed, StatusStep.InvoiceToBePaid].includes(
+              getStatus(customOrderBmsStatus, customOrderSellsyStatus)?.stepName ?? StatusStep.Unknown)
```
Max-7, line 262: Same here, you don't neek the `?? StatusStep.Unknown`

Max-7 (changes requested):

client/src/views/client-area/orders/OrdersPage.vue
```diff
@@ -35,7 +33,7 @@
     </template>
 
     <!-- no orders -->
-    <template v-if="!querying && !error && !orderRequests">
+    <template v-if="!querying && !error && orderRequests && orderRequests.requests.length === 0">
```
Ironicbay, line 36: I get that an error should be displayed if `orderRequests` is `undefined` as it should not have been assigned in the `onMounted`. For clarity, I would change the next condition (line 43) from
```
<template v-if="error">
```
to
```
<template v-if="error || !orderRequests">
```

client/src/views/client-area/ClientAreaPage.vue
```diff
@@ -250,20 +245,21 @@ onMounted(async () => {
...
-      // If there are no organization has been paid, we set orders, otherwise we default to contracts
+      // If there are no paid organizations, we show orders, otherwise we default to contracts
       currentPage.value = ClientAreaPages.Orders;
       // case where the user selected one specific organization
       if (!isDefaultOrganization(currentOrganization.value)) {
         if (await _isCustomOrderPaid(currentOrganization.value)) {
           currentPage.value = ClientAreaPages.Contracts;
         }
       } else {
+        let allPaid = organizations.value.length > 0;
```
Ironicbay, line 256: I'm not so sure about the necessity of this part.
From what I'm reading [here](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Logical_AND_assignment), `allPaid` would be updated only if it's set to `true` in the first place. If the outcome of `_isCustomOrderPaid(org)` is `false` at some point, continuing the loop would be useless as `allPaid` would be locked to `false`, and the outcome of this line 262 seems to contradict the previous implementation.

The previous behaviour was `if SOME org is _isCustomOrderPaid then go to Contracts`, the current behaviour is `if EVERY org is _isCustomOrderPaid then go to Contracts`

Since `organizations.value` is an array you can simplify all this by using array operators:
```
const allPaid = organizations.value.some(async (org) => await _isCustomOrderPaid(org));
```
or, depending on the intended logic
```
const allPaid = organizations.value.every(async (org) => await _isCustomOrderPaid(org));
```
and ditch the `for` loop below.

client/src/components/client-area/OrderInProgress.vue
```diff
@@ -145,7 +186,7 @@
               {{ $msTranslate('clientArea.orders.new.comment') }}
             </ion-text>
             <ion-text class="details-list-item__data subtitles-normal">
-              {{ $msTranslate(request.comment) }}
+              {{ $msTranslate(request.formula) }}
```
Ironicbay, line 189: Might put a `v-if` since `formula` is optional

client/src/components/client-area/OrderInProgress.vue
```diff
@@ -261,29 +301,24 @@ import { formatFileSize } from '@/common/file';
...
   CustomOrderRequest,
   DataType,
 } from '@/services/bms';
 import { ref, onMounted } from 'vue';
-import { isDefaultOrganization } from '@/views/client-area/types';
 import OrderTrackingProgress from '@/components/client-area/OrderTrackingProgress.vue';
 import { Translatable, I18n } from 'megashark-lib';
 
 const error = ref<string>('');
 const querying = ref(true);
```
Ironicbay, line 315: `querying` is unused in this view

client/src/components/client-area/OrderInProgress.vue
```diff
@@ -261,29 +301,24 @@ import { formatFileSize } from '@/common/file';
...
   BmsAccessInstance,
   CustomOrderRequest,
   DataType,
 } from '@/services/bms';
 import { ref, onMounted } from 'vue';
-import { isDefaultOrganization } from '@/views/client-area/types';
 import OrderTrackingProgress from '@/components/client-area/OrderTrackingProgress.vue';
 import { Translatable, I18n } from 'megashark-lib';
 
 const error = ref<string>('');
```
Ironicbay, line 314: `error` is unused in this view

Ironicbay (changes requested):

client/src/views/client-area/ClientAreaPage.vue
Max-7: Seen with @fabienSvtr, we do want to only show the contract page if all organization have been paid (or whatever the function checks for).

An early break in the loop is not a bad idea, especially since we're talking about network queries.

However, I don't think `some` or `every` handle promises: the function will return a Promise, it will be evaluted to `true`, and thus `allPaids` would always be true, and the queries will only resolve after the loop is executed.

client/src/locales/en-US.json
```diff
@@ -1862,8 +1862,9 @@
         },
         "orders": {
             "loading": "Your orders are loading...",
+            "noOrders": "You don't have any pending orders.",
```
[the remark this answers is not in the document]
fabienSvtr: I have replaced each time we have this spelling

client/src/views/client-area/orders/OrdersPage.vue
fabienSvtr: I can update the condition, but I'm not sure about the boolean. If we need to add more error messages, we should have similar code on each page.

Ironicbay (approved):

Max-7 (approved):

PR #10242, opened 2025-05-05
chore(cargo-deny): Ignore RUSTSEC-2025-0026

FirelightFlagboy:
A new advisory was created for the now `unmaitained` crate `registry` that
manipulate Windows registry.

We depend on that crate from `winfsp-wrs-sys`, so until
Scille/winfsp_wrs#36 is fixed we have to
ignore it.

### DIFF
--- a/deny.toml
+++ b/deny.toml
@@ -105,6 +105,13 @@ reason = """
 `PyString::from_object` is not used in the codebase.
 """
 
+[[advisories.ignore]]
+id = "RUSTSEC-2025-0026"
+reason = """
+registry is no longer maintained but in a ok state.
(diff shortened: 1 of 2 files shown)


deny.toml
```diff
@@ -105,6 +96,13 @@ reason = """
 `PyString::from_object` is not used in the codebase.
 """
 
+[[advisories.ignore]]
+id = "RUSTSEC-2025-0363"
```
touilleMan, line 100: Is it by pure chance that the ID of the removed item is so similar to the added one ?
`RUSTSEC-2024-0363` vs `RUSTSEC-2025-0363`

FirelightFlagboy: Oups https://rustsec.org/advisories/RUSTSEC-2025-0026

mmmarcos (approved):

PR #10259, opened 2025-05-12
refactor(device-loader): Use `StorageManager` over `LocalStorage` on web

FirelightFlagboy:
Since `libparsec` is run from a shared worker, we no longer have access to `localStorage` via the `window` property.

This commit refactor the web impl to use `StorageManager` over `LocalStorage` that is available in both browser and shared worker.

Another fix would have been to store the device in IndexedDB.

### DIFF
--- a/.github/workflows/ci-rust.yml
+++ b/.github/workflows/ci-rust.yml
@@ -162,6 +162,10 @@ jobs:
       - name: Test platform-device-loader on wasm
         if: inputs.run-wasm-tests
         run: wasm-pack test --headless --firefox libparsec/crates/platform_device_loader
+        env:
+          # RUST_LOG: wasm_bindgen_test_runner
+          # Default timeout is 20, increase it for slow runner
+          WASM_BINDGEN_TEST_TIMEOUT: 40
         timeout-minutes: 10
 
       - name: Build CLI binary
--- a/libparsec/crates/platform_device_loader/src/web/internal.rs
+++ b/libparsec/crates/platform_device_loader/src/web/internal.rs
@@ -1,107 +1,110 @@
 // Parsec Cloud (https://parsec.cloud) Copyright (c) BUSL-1.1 2016-present Scille SAS
 
-use std::{path::Path, sync::Arc};
+use std::{path::Path, rc::Rc, sync::Arc};
 
-use itertools::Itertools;
+use libparsec_platform_async::{lock::Mutex, stream::StreamExt};
 use libparsec_types::{
     AvailableDevice, DateTime, DeviceAccessStrategy, DeviceFile, DeviceFilePassword, LocalDevice,
 };
 
-use super::error::*;
+use crate::web::wrapper::{DirEntry, DirOrFileHandle, OpenOptions};
+
+use super::{
+    error::*,
+    wrapper::{Directory, File},
+};
 
 pub struct Storage {
-    storage: web_sys::Storage,
+    root_dir: Directory,
 }
 
 impl Storage {
-    pub(crate) fn new() -> Result<Self, NewStorageError> {
-        let window = web_sys::window().ok_or(NewStorageError::NoWindow)?;
-        let storage = if cfg!(test) {
-            // Use session storage for tests, So the browser does not try to persist the data.
-            window.session_storage()
-        } else {
-            window.local_storage()
-        }
-        .map_err(NewStorageError::WindowError)
-        .and_then(|v| v.ok_or(NewStorageError::NoLocalStorage))?;
-        Ok(Self { storage })
+    pub(crate) async fn new() -> Result<Self, NewStorageError> {
+        let root_dir = Directory::get_root().await?;
+        Ok(Self { root_dir })
     }
 
-    pub(crate) fn list_available_devices(
+    pub(crate) async fn list_available_devices(
         &self,
         config_dir: &Path,
     ) -> Result<Vec<AvailableDevice>, ListAvailableDevicesError> {
         let devices_dir = crate::get_devices_dir(config_dir);
-        let device_prefix = format!("{}/", devices_dir.display());
-        log::debug!("Will list devices starting with `{device_prefix}`");
-        let items_count =
-            self.storage
-                .length()
-                .map_err(|e| ListAvailableDevicesError::GetItemStorage {
-                    key: "length".to_owned(),
-                    error: e,
-                })?;
-        Ok((0..items_count)
-            .filter_map(|i| {
-                let key = self.storage.key(i).ok().flatten();
-                match key {
-                    Some(key)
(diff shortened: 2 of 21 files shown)


libparsec/crates/platform_device_loader/tests/units/utils.rs
```diff
@@ -6,59 +6,76 @@ pub use platform::*;
 mod platform {
     use std::path::Path;
 
-    pub fn create_device_file(path: &Path, content: &[u8]) {
-        std::fs::create_dir_all(path.parent().unwrap()).unwrap();
-        std::fs::write(path, content).unwrap();
+    pub async fn create_device_file(path: &Path, content: &[u8]) {
+        tokio::fs::create_dir_all(path.parent().unwrap())
```
touilleMan, line 10: Since this is a test helper, it is okay to use the sync api to access the filesystem here l. I guess you changed to the Tokio one given the function was required to be asynchronous anyway due to constraints on the web implémentation ?
(If that's the case I'm okay with it ^^)

.github/workflows/ci-rust.yml
```diff
@@ -161,7 +161,37 @@ jobs:
 
       - name: Test platform-device-loader on wasm
         if: inputs.run-wasm-tests
-        run: wasm-pack test --headless --firefox libparsec/crates/platform_device_loader
+        shell: bash -o pipefail -ex {0}
```
touilleMan, line 164: Should be reverted to launching all tests (or there should be a comment explaining what's going on here ^^)

libparsec/crates/platform_device_loader/tests/units/utils.rs
FirelightFlagboy: Yes it's now async given the constrain of the web implementation.

libparsec/crates/tests_fixtures/src/tmp_path.rs
```diff
@@ -16,7 +16,7 @@ pub struct TmpPath(PathBuf);
 impl TmpPath {
     pub fn create() -> Self {
         let mut path = if cfg!(target_arch = "wasm32") {
-            PathBuf::default()
+            "/".into()
```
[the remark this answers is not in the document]
FirelightFlagboy: The comment is still correct:

- On web, it's `/`
- On native, it's `temp_dir()` (likely return `/tmp`)

Then we append `parsec-tests/${uuidv4()}` to it, so it's a random path either case.

libparsec/crates/platform_device_loader/tests/units/load.rs
```diff
@@ -5,7 +5,9 @@
 #![allow(clippy::unwrap_used)]
 
 use crate::{load_device, LoadDeviceError};
-use libparsec_tests_fixtures::prelude::*;
+use libparsec_tests_fixtures::prelude::{
```
touilleMan, line 8: stick to `use libparsec_tests_fixtures::prelude::*;` plz

libparsec/crates/platform_device_loader/src/web/internal.rs
```diff
@@ -1,103 +1,111 @@
...
+        {
+            Ok(dir) => dir,
+            Err(GetDirectoryHandleError::NotFound { .. }) => {
+                log::debug!("Could not found devices dir");
+                return Ok(Vec::new());
+            }
+            Err(e) => return Err(e.into()),
+        };
+        let mut devices = Vec::<AvailableDevice>::new();
+        let dirs_to_explore = Rc::new(Mutex::new(Vec::from([dir])));
```
touilleMan, line 53: Nitpick: import `libparsec_async::lock::Mutex` as `AsyncMutex` (this is what is done in the rest of the codebase and makes it obvious we are not dealing with the sync version)

touilleMan: BTW, the `RC<Mutex<>>` is not needed here since the variable only lives in this function (hence it is never shared).

libparsec/crates/platform_device_loader/src/web/wrapper.rs
```diff
@@ -0,0 +1,432 @@
...
+    }
+
+    fn new(handle: web_sys::FileSystemDirectoryHandle, parent: &Path) -> Self {
+        Self {
+            path: parent.join(handle.name()),
+            handle,
+        }
+    }
+
+    pub async fn get_dir_handle_from_path(
```
touilleMan, line 78: Nitpicks: For consistency in the naming we should choose:

- `dir` vs `directory` (the former being currently used in methods names, the later in the name of the struct)
- `file/folder` vs `file_handle/folder_handle` (the former being currently used in the name of the struct, the former in the methods names)

I'd rather avoid shortening the names whenever possible, so `directory` has my preference over `dir`

Regarding the `_handle` suffix, I'd rather removing it:
- As a concept, a handle is supposed to be an opaque object (e.g. UNIX file descriptor) that represent a resource owned by another system. However here the `File/Directory` structs are not just an opaque identifier, but an actual class that exposes methods (it is a wrapper that abstract away the need to deal with the actual handler).
- It is shorter so we get back the space lost by not shortering `directory` ;-p

libparsec/crates/platform_device_loader/src/web/wrapper.rs
```diff
@@ -0,0 +1,432 @@
...
+use wasm_bindgen_futures::{stream::JsStream, JsFuture};
+use web_sys::{js_sys, wasm_bindgen::JsCast, DomException};
+
+use super::error::{
+    CastError, GetDirectoryHandleError, GetFileHandleError, GetRootDirectoryError, ReadToEndError,
+    RemoveEntryError, WriteAllError,
+};
+
+#[derive(Clone)]
+pub struct Directory {
```
touilleMan, line 22: I guess all the wrapper stuff are internal helpers only needed by the web implementation, so there visibility should be `pub(super)` (or at least `pub(crate)`)

FirelightFlagboy: The module itself (wrapper) is not exposed so it's good ?

libparsec/crates/platform_device_loader/src/web/wrapper.rs
FirelightFlagboy: done

libparsec/crates/platform_device_loader/tests/units/load.rs
FirelightFlagboy: done

libparsec/crates/platform_device_loader/src/web/internal.rs
FirelightFlagboy: done

Issue #8422, opened 2024-09-17
Display app icon on windows selector


mmmarcos:
The Parsec icon is correctly displayed on Gnome app selector:
![image](https://github.com/user-attachments/assets/b6509a7c-1ad1-46d0-80f3-2ac9eb25a4cc)

But not on Gnome windows selector:
![image](https://github.com/user-attachments/assets/fbb691fa-f422-45a3-a13e-2e0bf24ec1f8)

Tested with latest snap on Debian.

PR #10265, opened 2025-05-13
Fix minor issues with app desktop entry

mmmarcos:
Related to #8422

Changes in this PR:

- Added extra info about app in `snapcraft.yaml` and `package.json`.
- Added instructions on how to build snap locally
- Moved desktop entry to a specific file in the `assets` directory. The following keys were changed:
  - Added `SingleMainWindow=true` so that the desktop manager does not include a "New Window" action
  - Added `StartupNotify=false` to be explicit that this is not managed
  - Added `StartupWMClass=parsec` to be explicit about the wmclass
  - Fixed `Categories` to use `;` separator
  - Added `Keywords` to make it easier to find the application
  - Added `GenericName` and improve `Comment`

See: https://specifications.freedesktop.org/desktop-entry-spec/latest/recognized-keys.html

### DIFF
--- a/client/electron/README.md
+++ b/client/electron/README.md
@@ -1,8 +1,10 @@
 # Electron app for Parsec
 
-## Signing Windows releases
+## Signing a Windows release
 
-The `package-client` workflow produces an artefact called `Windows-X64-electron-app-exe-pre-built`. This artefact contains all the assets, build artefacts and scripts necessary to build a windows release (i.e an NSIS installer). More precisely, the artefact contains the following files and directories:
+The `package-client` workflow produces an artefact called `Windows-X64-electron-app-exe-pre-built`.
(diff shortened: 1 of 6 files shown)


client/electron/README.md
```diff
@@ -40,3 +42,63 @@ There are two requirements:
...
+6. Create a symbolic link to the snap directory from the repo root directory
+
+   ```shell
+   ln -sv client/electron/snap
+   ```
+
+7. (optional) Clean your repo so the default lxd storage (30Go) does not run out of space
+
+   ```shell
+   cargo clean
```
FirelightFlagboy, line 91: **nitpick (non-blocking):** Also cleanup `node_modules`

```suggestion
   fd node_modules --type d --prune --no-ignore --exec-batch rm -rf
   cargo clean
```

PR #10329, opened 2025-05-20
[Account] Add account_create_with_password_proceed cmd.

### LINKED ISSUE #10223: account_password_create route (authenticated)
- [x] implementation
- [ ] server tests
- [ ] libparsec tests

AureliaDolo:
Fix #10223


docs/rfcs/1013-parsec-account.md
```diff
@@ -422,78 +422,84 @@ To avoid DOS (Deny Of Service) attack, the server will limit the number of reque
 Anonymous account API:
 
 ```json5
```
FirelightFlagboy, line 424: **todo (blocking):** Format the code block using `jq .`

libparsec/crates/protocol/schema/anonymous_account_cmds/account_create_with_password_proceed.json5
```diff
@@ -0,0 +1,78 @@
+
```
FirelightFlagboy, line 1: **todo (blocking):** Format the file with `jq .`

server/parsec/components/memory/account.py
```diff
@@ -39,6 +45,42 @@ async def create_email_validation_token(
...
+        self,
+        token: EmailValidationToken,
+        now: DateTime,
+        hmac_key: SecretKey,
+        vault_key_access: bytes,
+        human_label: str,
+    ) -> None | CreateAccountWithPasswordBadOutcome:
+        # check token
+        # TODO check time
+        unverified_email = [k for (k, v) in self._data.unverified_emails.items() if v[0] == token]
```
FirelightFlagboy, line 58: **polish (blocking):** use `next()` instead, it will return `T | None` so you wont need the `[0]` below

server/parsec/components/memory/account.py
```diff
@@ -39,6 +45,42 @@ async def create_email_validation_token(
...
+        if not unverified_email:  # the provided token is not in unverified emails
+            return CreateAccountWithPasswordBadOutcome.INVALID_TOKEN
+        email = unverified_email[0]
+
+        # create authentication method
+        auth_method = MemoryAuthenticationMethod(
+            created_on=now,
+            created_by_ip="127.0.0.0",  # TODO
+            created_by_user_agent="agent",  # TODO
```
FirelightFlagboy, line 67: **todo:** Need to be moved outside (could be hard-coded in the caller), the method should take 2 params (or a struct) the `client_ip` and `client_user_agent`, both would be found in the HTTP headers.

server/parsec/components/memory/account.py
```diff
@@ -39,6 +45,42 @@ async def create_email_validation_token(
...
+            hmac_key=hmac_key,
+            vault_key_access=vault_key_access,
+            master_secret_algorithm='{"name": "ARGON2ID", "opslimit": 65536, "memlimit_kb": 3, "parallelism": 1}',  # TODO
```
FirelightFlagboy, line 70: **todo:** Like for ip and `user_agent`, the value should be provided by the caller

server/parsec/components/memory/datamodel.py
```diff
@@ -859,4 +860,31 @@ class MemoryShamirShare:
...
+    vaults: list[bytes]
+    authentication_methods: list[MemoryAuthenticationMethod]
+
+
+@dataclass(slots=True)
+class MemoryAuthenticationMethod:
+    created_on: DateTime
+    # IP address of the HTTP request that created the authentication method
+    # (either by account creation, vault key rotation or account recovery)
+    created_by_ip: str
```
FirelightFlagboy, line 877: **nitpick:** use [`IPvAnyAddress`](https://docs.pydantic.dev/latest/api/networks/#pydantic.networks.IPvAnyAddress)

server/parsec/components/memory/datamodel.py
```diff
@@ -859,4 +860,31 @@ class MemoryShamirShare:
...
+    created_by_user_agent: str
+    # Secret key used for HMAC based authentication with the server
+    hmac_key: SecretKey
+    # Vault key encrypted with the `auth_method_secret_key` see rfc 1014
+    vault_key_access: bytes
+    # The algorithm to obtain the `auth_method_master_secret`, it depends on
+    # the authentication method used, this attribute should contains the full configuration
+    # needed to do this operation (Typically, for the password it should be something like
+    # `{"name": "ARGON2ID", "opslimit": 65536, "memlimit_kb": 3, "parallelism": 1}`).
+    master_secret_algorithm: str  # TODO KeyAlgorithm
```
FirelightFlagboy, line 888: **thought:** I'm tempted to say we store a `any` here 

1. Given it's provided as an object from the client and only useful it.
2. In postgres we can store JSON data.
3. We would not need to perform a deserialize op when the list want to list the registred auth method

server/parsec/components/memory/datamodel.py
```diff
@@ -859,4 +860,31 @@ class MemoryShamirShare:
...
+    # (server side-only): Indicate if this method is enabled or not.
```
FirelightFlagboy, line 889: **nitpick:** We are the the sever side
```suggestion
    # Indicate if this method is enabled or not.
```

server/parsec/_parsec_pyi/protocol/anonymous_account_cmds/v5/account_create_with_password_proceed.pyi
FirelightFlagboy, line 1: **question:** The `pyi` was updated but not `parsec` crate (under `server/src`), is this an oversight?

server/parsec/components/account.py
```diff
@@ -30,6 +31,10 @@ class CreateEmailValidationTokenBadOutcome(BadOutcomeEnum):
     TOO_SOON_AFTER_PREVIOUS_DEMAND = auto()
 
 
+class CreateAccountWithPasswordBadOutcome(BadOutcomeEnum):
+    INVALID_TOKEN = auto()
+
+
 class BaseAccountComponent:
```
FirelightFlagboy, line 38: **todo:** The Base class should provided a way for other component to access the authentication method of an account

server/parsec/components/memory/datamodel.py
touilleMan: # Overview

I see three ways of storing the algorithm data:

Considering two possible algorithms:
- `{"name": "AlgoA", "param_a": <int>`
- `{"name": "AlgoB", "param_b": <str>`

## 1) flatten all possible fields in the main table

```sql
CREATE TYPE algorithm_type AS ENUM ('ALGO_A', 'ALGO_B'');

CREATE TABLE account (
  algorithm_type: algorithm_type,
  algorithm_algo_a_param_a: INT,
  algorithm_algo_a_param_b: TEXT
  ...
);
```

## 2) Store config as a JSON blob

```sql
CREATE TYPE algorithm_type AS ENUM ('ALGO_A', 'ALGO_B'');

CREATE TABLE account (
  algorithm_type: algorithm_type,
  algorithm_config: JSONB,
  ...
);
```

## 3) Store config in dedicated tables


```sql
CREATE TABLE password_algorithm_algo_a (
  _id SERIAL PRIMARY KEY,
  algorithm_algo_a_param_a: INT,
  algorithm_algo_a_param_b: TEXT
);
CREATE TABLE password_algorithm_algo_b (
  _id SERIAL PRIMARY KEY,
  algorithm_algo_a_param_b: TEXT
);

CREATE TABLE account (
  -- One of those (and only one !) of those field must be set
  algorithm_algo_a: INTEGER REFERENCE password_algorithm_algo_a(_id),
  algorithm_algo_b: INTEGER REFERENCE password_algorithm_algo_b(_id),
  ...
);
```

# Which one should we use ?

Solution 1 is the simplest but grows in complexity for every new algorithm format we need to support.

Right at the opposite of the spectrum, solution 2 shines if we have plenty of algorithm available, on the other hand it makes the data more opaque (since json can have any arbitrary fields).

Solution 3 is kind of a middle ground.

Since we are not planning to add any new algorithm for the time being (we would basically only add a new one if Argon2ID becomes deprecated), I would say solution 1 is the one to go with (hell, we don't even need to have an `algorithm_type` enum type for now since there is only a single choice !)

On top of that, we will be able to easily switch to solution 2 or 3 in the future by doing a straightforward migration.

server/parsec/components/memory/datamodel.py
FirelightFlagboy: The IP is not provided by the client's message but by the network stack

server/parsec/components/account.py
```diff
@@ -23,12 +24,25 @@
...
+)
+
 
 class CreateEmailValidationTokenBadOutcome(BadOutcomeEnum):
     ACCOUNT_ALREADY_EXISTS = auto()
     TOO_SOON_AFTER_PREVIOUS_DEMAND = auto()
 
 
+class CreateAccountWithPasswordBadOutcome(BadOutcomeEnum):
+    INVALID_TOKEN = auto()
```
touilleMan, line 38: I guess we should also have an `ACCOUNT_ALREADY_CREATED` error (similar to what we do during organization bootstrap with the `ORGANIZATION_ALREADY_BOOTSTRAPPED`  error)

server/parsec/client_context.py
```diff
@@ -173,6 +173,8 @@ class AnonymousAccountClientContext:
     client_api_version: ApiVersion
     settled_api_version: ApiVersion
     logger: ParsecBoundLogger = field(init=False)
+    user_agent: str
+    user_ip: str
```
touilleMan, line 177: Since ASGI doesn't guarantee that the ip can be provided, we should type it as `str | None` and add a comment (typically linking to https://asgi.readthedocs.io/en/latest/specs/www.html#http-connection-scope)

server/parsec/client_context.py
```diff
@@ -175,6 +175,8 @@ class AnonymousAccountClientContext:
...
```
FirelightFlagboy, line 179: **todo (blocking):** Add a todo comment with a link to #10384

libparsec/crates/types/src/addr.rs
```diff
@@ -40,6 +40,7 @@ const PARSEC_ACTION_CLAIM_USER: &str = "claim_user";
 const PARSEC_ACTION_CLAIM_DEVICE: &str = "claim_device";
 const PARSEC_ACTION_CLAIM_SHAMIR_RECOVERY: &str = "claim_shamir_recovery";
 const PARSEC_ACTION_PKI_ENROLLMENT: &str = "pki_enrollment";
+const PARSEC_ACTION_ACCOUNT_EMAIL_VALIDATION: &str = "account_email_validation";
```
touilleMan, line 43: nitpick:
```suggestion
const PARSEC_ACCOUNT_ACTION_EMAIL_VALIDATION: &str = "account_email_validation";
```

I would use "account" as a prefix to indicate this action is unrelated to the other ones (since we now have two types of action URLs)

In a subsequent PR, I think we should also rename `ParsceActionAddr` -> `ParsecOrganizationActionAddr` (and also `PARSEC_ACTION_XXX` -> `PARSEC_ORGANIZATION_ACTION_XXX`)

server/parsec/components/account.py
touilleMan: > if the implementation is not done correctly we could end with an oracle that list registered user.

If that's the case, we would have bigger problem since it means an invalid token would pass validation fine !

The point is here is to decide what to do if a *valid* token has been already used.

> That also mean to do that correctly we need to keep the token in the state for longer than required

"longer than required" is the interesting part: it is possible than for user experience reasons we are required to keep the token virtually forever, typically:

- If the user has made a mistake while pasting by hand the URL he should get "token invalid"
- If token has a lifetime and the user uses it too late, he should get "token too old" (and the GUI tells the user to re-validate the email)
- If the token has been used to create the account, the user re-using it should get "token already used" (and the GUI tells the user to authenticate)

@Max-7 Do you think we have those (or maybe others) needs here ?

Also so far in the database we strongly refrain from deleting data right away. So here I guess we should also refrain from deleting the row containing the token (but instead link it to the created account). In the future if we want to do some cleanup in the database, I guess we would rely on providing a script that should be run in a scheduled fashion.

FirelightFlagboy (approved):

PR #10370, opened 2025-05-22
refactor(libparsec-web): Add log on trampoline-web

FirelightFlagboy:
- Add log to indicate the mode `libparsec` is run
- Add event watcher for worker error
- Add event watcher for worker port error


client/src/plugins/libparsec/trampoline-web.ts
```diff
@@ -48,8 +48,10 @@ export async function LoadWebLibParsecPlugin(): Promise<any> {
   // So in this case we fallback to have libparsec running in the tab, with some
   // web locks to avoid having multiple tabs using the same device.
   if (window.SharedWorker !== undefined) {
+    console.log('Starting libparsec in shared worker mode')
     return await withSharedWorker();
   } else {
+    console.log('Starting libparsec in non worker mode')
```
Max-7, line 54: Isn't it still a worker, just not a shared one?

Max-7 (approved):

PR #10373, opened 2025-05-22
Update recovery files texts

mmmarcos:
I propose using "Recovery File" instead of "Organization Recovery" since everything in the profile is already to the organization and updated texts explain that this is done per-organization.

Also some minor changes.


client/src/locales/fr-FR.json
```diff
@@ -979,7 +979,7 @@
             },
             "finish": {
                 "first": "Vous pouvez maintenant vous connecter à Parsec.",
-                "second": "Nous vous conseillons de créer un fichier de récupération au cas où vous oublieriez votre mot de passe."
+                "second": "Nous vous conseillons vivement de créer un fichier de récupération au cas où vous oublieriez votre mot de passe ou perdriez d'accès à votre appareil."
```
Max-7, line 982: `perdriez l'accès`

Max-7 (approved):

PR #10412, opened 2025-05-27
feat(types): Add `HostAddr`

### LINKED ISSUE #10384: Add custom IP validation

FirelightFlagboy:
Need to rename `*_ip` to `*_host` since the value may be a hostname instead of a ip.

Fix #10384


touilleMan:
Also since we have the type `EmailAddress`, I'd rather stay with `Address` suffix instead of `Addr`

As a rule of thumb avoid shortening names in types.

However I plead guilty to have broken this rule a fair number of time (see `ParsecAddr` for instance :disappointed: ), but the codebase seems to have settle toward this (I would say short name are only really used sparingly when naming local variable when it makes things more readable)

FirelightFlagboy:
The `request.client.host` can be a hostname, depending on how it's configured with the proxy_header:

https://github.com/encode/uvicorn/blob/fa3d9d27dbe7ecbc68e0992309d0ba49ead5640b/uvicorn/middleware/proxy_headers.py#L27

On some condition, it replace the client ip by the value contained in `x-forwarded-for` which on top of IPs, could also be a hostname

touilleMan:
> The request.client.host can be a hostname, depending on how it's configured with the proxy_header:

But since we end up with either a hostname or an ip or nothing, what is the purpose of adding this type instead of simply relying on a String ?

I think we should choose here to either:
- stick to a string that is only aimed at providing info for human
- stick with a `IPAddress` type and consider store None whenever the input data doesn't correspond

(I think we'd better choose the first choice unless we have an actual need to handle ip adress programatically)

PR #10431, opened 2025-05-28
[Account] js bindings for account creation

### LINKED ISSUE #10388: [Account] Add bindings for account create
Bindings API for account create

- [ ] `accountCreateSendValidationEmail` send email with validation code
  - Req: email
  - Res: OK or `AccountCreateSendValidationEmailError`
- [ ] `accountCreateProceed` (name TBC) check if code entered by user is valid
  - Req: 
      -  step check code: email, validation token/code
      - step create: human handle, email,  password
  - Res: OK or `AccountCreateProceedError`

AureliaDolo:
Fix #10388 

- [x] account client setup
- account_send_email_validation_token
    - [x] libparsec_account
    - [x] libparsec_std
    - [ ] tests
    - [x] generate bindings

- account_create_with_password_proceed
    - [x] libparsec_account
    - [x] libparsec_std
    - [ ] tests
    - [x] generate bindings

### DIFF
--- a/bindings/generator/api/account.py
+++ b/bindings/generator/api/account.py
@@ -1,0 +1,82 @@
+# Parsec Cloud (https://parsec.cloud) Copyright (c) BUSL-1.1 2016-present Scille SAS
+
+
+from .common import (
+    HumanHandle,
+    Password,
+    Ref,
+    Path,
+    Result,
+    EmailAddress,
+    ErrorVariant,
+    Variant,
+    ValidationCode,
+)
+from .addr import ParsecAddr
+
+
+class AccountSendEmailValidationTokenError(ErrorVariant):
+    class Stopped:
+        pass
+
+    class Offline:
+        pass
+
+    class Internal:
+        pass
+
+    class EmailRecipientRefused:
+        pass
+
+    class EmailServerUnavailable:
+        pass
+
+    class EmailParseError:
+        pass
+
+
+class AccountCreateProceedError(ErrorVariant):
+    class Stopped:
+        pass
+
+    class Offline:
+        pass
+
+    class Internal:
+        pass
+
+    class InvalidValidationCode:
+        pass
+
+    class AuthMethodIdAlreadyExists:
+        pass
+
+    class CryptoError:
+        pass
+
+
+async def account_create_send_validation_email(
+    email: Ref[str],
+    config_dir: Ref[Path],
+    addr: ParsecAddr,
+) -> Result[None, AccountSendEmailValidationTokenError]:
+    raise NotImplementedError
+
+
+class AccountCreateStep(Variant):
+    class CheckCode:
+        validation_code: ValidationCode
+        email: EmailAddress
+
+    class Create:
+        human_handle: HumanHandle
+        password: Password
+        validation_code: ValidationCode
+
+
+async def account_create_proceed(
+    step: AccountCreateStep,
+    config_dir: Ref[Path],
+    addr: ParsecAddr,
+) -> Result[None, AccountCreateProceedError]:
+    raise NotImplementedError
--- a/bindings/generator/api/common.py
+++ b/bindings/generator/api/common.py
@@ -268,6 +268,13 @@ class InvitationToken(StrBasedType):
     )
 
 
+class ValidationCode(StrBasedType):
+    custom_from_rs_string = "|s: String| -> Result<libparsec::ValidationCode, _> { libparsec::ValidationCode::from_str(&s).map_err(|e| e.to_string()) }"
+    custom_to_rs_string = (
+        "|x: libparsec::ValidationCode| -> Result<String, &'static str> { Ok(x.into()) }"
+    )
+
+
 class GreetingAttemptID(StrBasedType):
     custom_from_rs_string = "|s: String| -> Result<libparsec::GreetingAttemptID, _> { libparsec::GreetingAttemptID::from_hex(s.as_str()).map_err(|e| e.to_string()) }"
     custom_to_rs_string = (
--- a/client/src/plugins/libparsec/definitions.ts
+++ b/client/src/plugins/libparsec/definitions.ts
@@ -103,6 +103,7 @@ export type Path = string
 export type SASCode = string
 export type SequesterServiceID = string
 export type UserID = string
+export type ValidationCode = string
 export type VlobID = string
 export type SequesterVerifyKeyDer = Uint8Array
 export type NonZeroU8 = number
@@ -416,6 +417,111 @@ export interface WorkspaceUserAccessInfo {
     currentProfile: UserProfile
     currentRole: RealmRole
 }
+
+// AccountCreateProceedError
+export enum AccountCreateProceedErrorTag {
+    AuthMethodIdAlreadyExists = 'AccountCreateProceedErrorAuthMethodIdAlreadyExists',
+    CryptoError = 'AccountCreateProceedErrorCryptoError',
+    Internal = 'AccountCreateProceedErrorInternal',
+    InvalidValidationCode = 'AccountCreateProceedErrorInvalidValidationCode',
+    Offline = 'AccountCreateProceedErrorOffline',
+    Stopped = 'AccountCreateProceedErrorStopped',
+}
+
+export interface AccountCreateProceedErrorAuthMethodIdAlreadyExists {
+    tag: AccountCreateProceedErrorTag.AuthMethodIdAlreadyExists
+    error: string
+}
+export interface AccountCreateProceedErrorCryptoError {
+    tag: AccountCreateProceedErrorTag.CryptoError
+    error: string
+}
+export interface AccountCreateProceedErrorInternal {
+    tag: AccountCreateProceedErrorTag.Internal
+    error: string
+}
+export interface AccountCreateProceedErrorInvalidValidationCode {
+    tag: AccountCreateProceedErrorTag.InvalidValidationCode
+    error: string
+}
+export interface AccountCreateProceedErrorOffline {
+    tag: AccountCreateProceedErrorTag.Offline
+    error: string
+}
+export interface AccountCreateProceedErrorStopped {
+    tag: AccountCreateProceedErrorTag.Stopped
+    error: string
+}
+export type AccountCreateProceedError =
+  | AccountCreateProceedErrorAuthMethodIdAlreadyExists
+  | AccountCreateProceedErrorCryptoError
+  | AccountCreateProceedErrorInternal
+  | AccountCreateProceedErrorInvalidValidationCode
+  | AccountCreateProceedErrorOffline
+  | AccountCreateProceedErrorStopped
+
+// AccountCreateStep
+export enum AccountCreateStepTag {
+    CheckCode = 'AccountCreateStepCheckCode',
+    Create = 'AccountCreateStepCreate',
+}
+
+export interface AccountCreateStepCheckCode {
+    tag: AccountCreateStepTag.CheckCode
+    validationCode: ValidationCode
+    email: EmailAddress
+}
+export interface AccountCreateStepCreate {
+    tag: AccountCreateStepTag.Create
+    humanHandle: HumanHandle
+    password: Password
+    validationCode: ValidationCode
+}
+export type AccountCreateStep =
+  | AccountCreateStepCheckCode
+  | AccountCreateStepCreate
+
+// AccountSendEmailValidationTokenError
+export enum AccountSendEmailValidationTokenErrorTag {
+    EmailParseError = 'AccountSendEmailValidationTokenErrorEmailParseError',
+    EmailRecipientRefused = 'AccountSendEmailValidationTokenErrorEmailRecipientRefused',
+    EmailServerUnavailable = 'AccountSendEmailValidationTokenErrorEmailServerUnavailable',
+    Internal = 'AccountSendEmailValidationTokenErrorInternal',
+    Offline = 'AccountSendEmailValidationTokenErrorOffline',
+    Stopped = 'AccountSendEmailValidationTokenErrorStopped',
+}
+
+export interface AccountSendEmailValidationTokenErrorEmailParseError {
+    tag: AccountSendEmailValidationTokenErrorTag.EmailParseError
+    error: string
+}
+export interface AccountSendEmailValidationTokenErrorEmailRecipientRefused {
+    tag: AccountSendEmailValidationTokenErrorTag.EmailRecipientRefused
+    error: string
+}
+export interface AccountSendEmailValidationTokenErrorEmailServerUnavailable {
+    tag: AccountSendEmailValidationTokenErrorTag.EmailServerUnavailable
+    error: string
+}
+export interface AccountSendEmailValidationTokenErrorInternal {
+    tag: AccountSendEmailValidationTokenErrorTag.Internal
+    error: string
+}
+export interface AccountSendEmailValidationTokenErrorOffline {
+    tag: AccountSendEmailValidationTokenErrorTag.Offline
+    error: string
+}
+export interface AccountSendEmailValidationTokenErrorStopped {
+    tag: AccountSendEmailValidationTokenErrorTag.Stopped
+    error: string
+}
+export type AccountSendEmailValidationTokenError =
+  | AccountSendEmailValidationTokenErrorEmailParseError
+  | AccountSendEmailValidationTokenErrorEmailRecipientRefused
+  | AccountSendEmailValidationTokenErrorEmailServerUnavailable
+  | AccountSendEmailValidationTokenErrorInternal
+  | AccountSendEmailValidationTokenErrorOffline
+  | AccountSendEmailValidationTokenErrorStopped
 
 // ActiveUsersLimit
 export enum ActiveUsersLimitTag {
@@ -4446,6 +4552,16 @@ export type WorkspaceWatchEntryOneShotError =
   | WorkspaceWatchEntryOneShotErrorStopped
 
 export interface LibParsecPlugin {
+    accountCreateProceed(
+        step: AccountCreateStep,
+        config_dir: Path,
+        addr: ParsecAddr
+    ): Promise<Result<null, AccountCreateProceedError>>
+    accountCreateSendValidationEmail(
+        email: string,
+        config_dir: Path,
+        addr: ParsecAddr
+    ): Promise<Result<null, AccountSendEmailValidationTokenError>>
     archiveDevice(
         device_path: Path
     ): Promise<Result<null, ArchiveDeviceError>>
--- a/libparsec/crates/account/src/lib.rs
+++ b/libparsec/crates/account/src/lib.rs
@@ -1,9 +1,12 @@
 // Parsec Cloud (https://parsec.cloud) Copyright (c) BUSL-1.1 2016-present Scille SAS
 
-use std::path::Path;
+use std::{path::Path, str::FromStr};
 
 use libparsec_client_connection::{
     AccountAuthMethod, AnonymousAccountCmds, AuthenticatedAccountCmds, ConnectionError, ProxyConfig,
+};
+use libparsec_protocol::anonymous_account_cmds::v5::{
+    account_create_proceed, account_create_send_validation_email,
 };
 use libparsec_types::prelude::*;
 
@@ -29,6 +32,24 @@ pub struct Account {
 pub enum AccountLoginWithPasswordError {
     #[error("Server provided an invalid password algorithm config: {0}")]
     BadPasswordAlgorithm(CryptoError),
+    #[error("Component has stopped")]
+    Stopped,
+    #[error("Cannot communicate with the server: {0}")]
+    Offline(#[from] ConnectionError),
+    #[error(transparent)]
+    Internal(#[from] anyhow::Error),
+}
+
+#[derive(Debug, thiserror::Error)]
+pub enum AccountSendEmailValidationTokenError {
+    #[error("Email recipient refused")]
+    EmailRecipientRefused,
+    #[error("Email server unavailable")]
+    EmailServerUnavailable,
+    #[error("Unable to parse email")]
+    EmailParseError(#[from] EmailAddressParseError),
+    #[error("Component has stopped")]
+    Stopped,
     #[error("Cannot communicate with the server: {0}")]
     Offline(#[from] ConnectionError),
     #[error(transparent)]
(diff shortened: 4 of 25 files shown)


bindings/generator/api/account.py
```diff
@@ -0,0 +1,86 @@
...
+async def account_create_proceed(
+    config_dir: Ref[Path],
+    addr: ParsecAnonymousAccountAddr,
+    auth_method_hmac_key: SecretKey,
+    human_label: str,
+    password_algorithm: Optional[PasswordAlgorithm],
+    validation_token: EmailValidationToken,
+    vault_key_access: Bytes,
+    auth_method_id: AccountAuthMethodID,
+) -> Result[None, AccountCreateProceedError]:
```
FirelightFlagboy: **question:** Is it possible to have a more simple exposed function? it take a lot of params where most IMO should be generated in the lib

libparsec/crates/account/src/lib.rs
FirelightFlagboy, line 1: **issue:** The implementation is different with the one proposed by @touilleMan in #10471

libparsec/crates/account/src/lib.rs
```diff
@@ -1 +1,153 @@
...
+        bad_rep @ account_create_proceed::Rep::UnknownStatus { .. } => {
+            Err(anyhow::anyhow!("Unexpected server response: {:?}", bad_rep).into())
+        }
+    }
+}
+
+#[allow(unused)]
+struct Account {
+    cmds: AuthenticatedAccountCmds,
+    auth_method: AccountAuthMethod,
```
FirelightFlagboy: **question:** Why do we need the `cmds` & `auth_method`?

libparsec/crates/client_connection/src/authenticated_account_cmds.rs
```diff
@@ -24,7 +24,7 @@ const API_LATEST_MAJOR_VERSION: u32 = API_LATEST_VERSION.version;
 /// Method name that will be used for the `Authorization` header
 pub const PARSEC_AUTH_METHOD: &str = "PARSEC-MAC-BLAKE2B";
 
-#[derive(Debug)]
+#[derive(Debug, Clone)]
 pub struct AccountAuthMethod {
```
FirelightFlagboy: **nitpick:** `AuthenticatedCmds` use an `Arc<LocalDevice>`, for me `AccountAuthMethod` is in a similar case and should be wrapped in an `Arc` instead of deriving `Clone`

libparsec/crates/client_connection/src/lib.rs
```diff
@@ -14,7 +14,7 @@ mod testbed_send_hook_helpers;
 
 pub use anonymous_account_cmds::AnonymousAccountCmds;
 pub use anonymous_cmds::AnonymousCmds;
-pub use authenticated_account_cmds::AuthenticatedAccountCmds;
+pub use authenticated_account_cmds::{AccountAuthMethod, AuthenticatedAccountCmds};
```
FirelightFlagboy: **polish:** It feels strange that `client_conn` expose a new type, maybe we should move it to `types`?

libparsec/crates/account/src/lib.rs
AureliaDolo: It should not be an issue, both implemented methods are anonymous, thus they don't rely on the account struct. Here the account struct is just a placeholder.

libparsec/crates/client_connection/src/lib.rs
AureliaDolo: I don't know, authenticated_cmds also exposes quite a few things too.

bindings/generator/api/common.py
```diff
@@ -243,6 +243,21 @@ class SequesterVerifyKeyDer(BytesBasedType):
...
+    pass
+
+
+class Bytes(BytesBasedType):
+    """
+    libparsec's bytes
+    """
+
+    custom_from_rs_bytes = (
+        "|b: &[u8]| -> Result<libparsec::Bytes,  String> { Ok(libparsec::Bytes::from(b.to_vec())) }"
```
touilleMan: You may be able to use [Bytes::copy_from_slice(b)](https://docs.rs/bytes/latest/bytes/struct.Bytes.html#method.copy_from_slice) instead

bindings/generator/api/common.py
```diff
@@ -243,6 +243,21 @@ class SequesterVerifyKeyDer(BytesBasedType):
     pass
 
 
+class SecretKey(BytesBasedType):
+    pass
+
+
+class Bytes(BytesBasedType):
```
touilleMan: I don't see this type in `bindings/generator/api/account.py` is it used ?

bindings/generator/api/account.py
```diff
@@ -0,0 +1,73 @@
...
+        pass
+
+    class AuthMethodIdAlreadyExists:
+        pass
+
+    class CryptoError:
+        pass
+
+
+async def account_send_email_validation_token(
```
touilleMan: This should be named `account_create_send_validation_email` since it's the name of the API (and especially since it makes very clear this should method is related to `account_create_proceed`)

bindings/generator/api/account.py
```diff
@@ -0,0 +1,73 @@
...
+    email: Ref[str],
+    config_dir: Ref[Path],
+    addr: ParsecAnonymousAccountAddr,
```
touilleMan:
```suggestion
    addr: ParsecAddr,
```

`ParsecAnonymousAccountAddr` is being removed since `ParsecAddr` can be used instead without any drawback

libparsec/crates/client_connection/src/lib.rs
touilleMan: `AccountAuthMethod` is a simple structure and is only public to be used in `AuthenticatedAccountCmds::new/AuthenticatedAccountCmds::from_client` constructors

So I think we'd better entirely remove `AccountAuthMethod`, then pass separate arguments to the constructors (i.e. `time_provider/auth_method_id/auth_method_mac_key`) and store them directly in the `AuthenticatedAccountCmds` structure.

libparsec/crates/types/src/account.rs
```diff
@@ -62,6 +63,14 @@ pub enum PasswordAlgorithm {
     },
 }
 
+/*
+*  VaultKeyAccess
+*/
+#[derive(Debug, Clone, Serialize, Deserialize, PartialEq, Eq)]
+pub struct VaultKeyAccess {
```
touilleMan: `VaultKeyAccess` is going to be serialized, so it should be defined as a json5 in `types/schema/` following the RFC1014 specs

If your goal was to have a stub instead of an actual implementation, then you should add a big TODO since it is very easy to miss this ^^

libparsec/src/account.rs
```diff
@@ -0,0 +1,33 @@
...
+
+pub async fn account_send_email_validation_token(
+    email: &str,
+    config_dir: &Path,
+    addr: ParsecAnonymousAccountAddr,
+) -> Result<(), AccountSendEmailValidationTokenError> {
+    libparsec_account::account_send_email_validation_token(email, config_dir, addr).await
+}
+
+#[allow(clippy::too_many_arguments)]
```
touilleMan:
```suggestion
#[expect(clippy::too_many_arguments)]
```
?

client/src/plugins/libparsec/definitions.ts
```diff
@@ -4442,6 +4537,18 @@ export type WorkspaceWatchEntryOneShotError =
   | WorkspaceWatchEntryOneShotErrorStopped
 
 export interface LibParsecPlugin {
+    accountCreateProceed(
```
Max-7, line 4555: Would it be possible to separate this into three functions instead:
1. send the email
2. validate the code
3. create the account

It makes it easier to follow what was planned in the mock-ups.

client/src/plugins/libparsec/definitions.ts
```diff
@@ -4442,6 +4537,18 @@ export type WorkspaceWatchEntryOneShotError =
...
+        human_label: string,
```
Max-7: It would be better to separate `firstName` and `lastName`. For UI purposes, we don't always want to display the full name. We may want to have `Hi {firstName}!` in an email for example.

client/src/plugins/libparsec/definitions.ts
```diff
@@ -4442,6 +4537,18 @@ export type WorkspaceWatchEntryOneShotError =
...
+        validation_token: EmailValidationToken,
+        config_dir: Path,
+        addr: ParsecAnonymousAccountAddr,
+        password: Password
+    ): Promise<Result<null, AccountCreateProceedError>>
+    accountSendEmailValidationToken(
+        email: string,
+        config_dir: Path,
```
Max-7, line 4562: Why do you need `config_dir` in this case? Are you saving any data locally? If that the case, could we avoid having the `server` parameter for both functions?

AureliaDolo: Good catch, it's needed to create AnonymousAccountCmds but it's not actually used to create that.

AureliaDolo: Sooo it's used when testing with the testbed to get the send hook

bindings/generator/api/common.py
```diff
@@ -243,6 +243,10 @@ class SequesterVerifyKeyDer(BytesBasedType):
     pass
 
 
+class SecretKey(BytesBasedType):
```
touilleMan: `SecretKey` is some low level crypto stuff, we should not have to expose this to the GUI (as this would typically mean the GUI would have to generate its own `SecretKey`...)

Instead it should be the `libparsec`'s responsibility to handle this

bindings/generator/api/common.py
```diff
@@ -268,6 +272,20 @@ class InvitationToken(StrBasedType):
     )
 
 
+class EmailValidationToken(StrBasedType):
```
touilleMan: Note this is going to be replaced by `ValidationCode` (see https://github.com/Scille/parsec-cloud/pull/10576)

bindings/generator/api/account.py
```diff
@@ -0,0 +1,82 @@
...
+) -> Result[None, AccountSendEmailValidationTokenError]:
+    raise NotImplementedError
+
+
+class AccountCreateStep(Variant):
+    class CheckCode:
+        code: str
+
+    class Create:
+        human_label: str
```
touilleMan: `code` is also needed for the creation

libparsec/crates/account/src/lib.rs
```diff
@@ -93,6 +118,140 @@ impl Account {
...
+
+            let auth_method = AccountAuthMethod {
+                time_provider: TimeProvider::default(),
+                id: AccountAuthMethodID::from(
+                    auth_method_master_secret.derive_uuid_from_uuid(AUTH_METHOD_ID_DERIVATION_UUID),
+                ),
+                mac_key: auth_method_master_secret
+                    .derive_secret_key_from_uuid(AUTH_METHOD_MAC_KEY_DERIVATION_UUID),
+            };
+            let vault_key = auth_method_master_secret
```
touilleMan: It seems you forgot about `auth_method_secret_key` here ^^

`vault_key` should be a regular `SecretKey` randomly generated and serialized into the `vault_key_access` struct
`auth_method_secret_key` is `SecretKey` that is derived from `auth_method_master_secret` and is used to encrypt `vault_key_access`

libparsec/crates/account/src/lib.rs
```diff
@@ -17,6 +20,8 @@ pub const AUTH_METHOD_MAC_KEY_DERIVATION_UUID: uuid::Uuid =
     uuid::uuid!("11111111-1111-1111-1111-111111111111");
 pub const AUTH_METHOD_SECRET_KEY_DERIVATION_UUID: uuid::Uuid =
     uuid::uuid!("22222222-2222-2222-2222-222222222222");
+pub const AUTH_METHOD_VAULT_KEY_DERIVATION_UUID: uuid::Uuid =
```
touilleMan: `vault_key` is not derived, you should use `AUTH_METHOD_SECRET_KEY_DERIVATION_UUID` to derive `auth_method_secret_key` instead

libparsec/crates/account/src/lib.rs
```diff
@@ -28,6 +33,26 @@ pub struct Account {
...
+    Stopped,
+    #[error("Cannot communicate with the server: {0}")]
+    Offline(#[from] ConnectionError),
+    #[error(transparent)]
+    Internal(#[from] anyhow::Error),
+}
+
+#[derive(Debug, thiserror::Error)]
+pub enum AccountSendEmailValidationTokenError {
+    #[error("Email recipient refused")]
```
touilleMan, line 45: When sending an email to invite a user/device, the status for the email are:
```rust
pub enum InvitationEmailSentStatus {
    Success,
    ServerUnavailable,
    RecipientRefused,
}
```

So why here do we also have `InvalidEmail` and `EmailParseError` ?
And, beside, `InvalidEmail` and `EmailParseError` seem very similar, shouldn't we merge them together ?

bindings/generator/api/addr.py
```diff
@@ -48,6 +48,11 @@ class ParsecPkiEnrollmentAddr(StrBasedType):
     custom_to_rs_string = "|addr: libparsec::ParsecPkiEnrollmentAddr| -> Result<String, &'static str> { Ok(addr.to_url().into()) }"
 
 
+class ParsecAnonymousAccountAddr(StrBasedType):
```
touilleMan: see https://github.com/Scille/parsec-cloud/pull/10431#discussion_r2135240550

bindings/generator/api/common.py
```diff
@@ -268,6 +272,20 @@ class InvitationToken(StrBasedType):
...
+    custom_from_rs_string = "|s: String| -> Result<libparsec::EmailValidationToken, _> { libparsec::EmailValidationToken::from_hex(s.as_str()).map_err(|e| e.to_string()) }"
+    custom_to_rs_string = (
+        "|x: libparsec::EmailValidationToken| -> Result<String, &'static str> { Ok(x.hex()) }"
+    )
+
+
+class AccountAuthMethodID(StrBasedType):
```
touilleMan: You can remove this since we don't need expose the ID to the bindings

The only part where it could be useful to expose this ID would be in `vault_item_recovery_list`, but currently we just don't return it ;-)

libparsec/crates/protocol/schema/anonymous_account_cmds/account_create_proceed.json5
```diff
@@ -59,6 +28,65 @@
...
+                "name": "code",
+                "type": "String"
+              }
+            ]
+          },
+          {
+            "name": "Number1Create",
+            "discriminant_value": "NUMBER_1_CREATE",
+            "fields": [
+              {
```
touilleMan, line 53: validation_code is missing

libparsec/crates/protocol/schema/anonymous_account_cmds/account_create_proceed.json5
```diff
@@ -59,6 +28,65 @@
...
+        "name": "AccountCreateStep",
+        "discriminant_field": "step",
+        "variants": [
+          {
+            "name": "Number0CheckCode",
+            "discriminant_value": "NUMBER_0_CHECK_CODE",
+            "fields": [
+              {
+                /// Code received by email following use of `account_create_send_validation_email`
+                "name": "code",
```
touilleMan:
```suggestion
                "name": "validation_code",
```

touilleMan (dismissed):

libparsec/crates/protocol/schema/anonymous_account_cmds/account_create_proceed.json5
AureliaDolo: It has been validated at step zero, is it still needed here ?

FirelightFlagboy: IMO, yes it's needed else we would need to have a statefull backend with cookies and stuff

bindings/generator/api/account.py
```diff
@@ -0,0 +1,82 @@
...
+    email: Ref[str],
+    config_dir: Ref[Path],
+    addr: ParsecAddr,
+) -> Result[None, AccountSendEmailValidationTokenError]:
+    raise NotImplementedError
+
+
+class AccountCreateStep(Variant):
+    class CheckCode:
+        code: str
```
FirelightFlagboy: **todo:** 
```suggestion
        code: ValidationCode
```

libparsec/crates/protocol/schema/anonymous_account_cmds/account_create_proceed.json5
```diff
@@ -7,43 +7,12 @@
     "req": {
       "fields": [
         {
-          // Token received by email following use of `account_create_send_validation_email`
-          "name": "validation_token",
-          "type": "EmailValidationToken"
+          "name": "email",
```
touilleMan: `email` field should also be moved in the step:
- as `email: EmailAddress` for `Number0CheckCode`
- as `human_handle: HumanHandle` for `Number1Create`

This is because in `Number1Create` we currently also provide the `human_label` field, but as a string so without any guarantee regarding its content.
But `HumanHandle` constructor does some validation on the label (e.g. a label cannot contain a `<` character), and we don't have a dedicated type that represent a validated human label.

libparsec/crates/account/src/lib.rs
AureliaDolo: The email parse error comes from the the emailAdress parser, it provides more detail on which part has failed

And actually the server never responds invalid email (because we check it first) so we can get rid of invalid email altogether

libparsec/crates/protocol/schema/anonymous_account_cmds/account_create_proceed.json5
AureliaDolo: I don't get why we would need cookies, the only data we need at this point is the email

FirelightFlagboy: The protocol do not require that we do the check step, we can directly confirm the account creation

AureliaDolo: From what I understand we want to tweak the protocol to first check the code and then in another step proceed with the account creation to allow the front to ask separately for the code.

If we allow to do it all in one step, it kinda defeat that purpose ?

FirelightFlagboy: > If we allow to do it all in one step, it kinda defeat that purpose ?

I mean the step is called `CHECK_CODE`, not `VALIDATE_CODE`.

by not providing the code in the second step, that require to ensure that `CHECK_CODE` was called before with a valid code.

For me, it seems simpler to just pass the code again

AureliaDolo: I think it could introduce some issues : what happens if step0 is ok then step1 is called but with a different code ? 

And maybe it was poor choice of words on my part, but what does step0 do if it does not validate / check that the provided code is the expected one?

if we validate the code only in step1, there is no point in having step0 IMHO

FirelightFlagboy: > what happens if step0 is ok then step1 is called but with a different code ?

The API is stateless, so server would check the code and find that it's invalid.

> but what does step0 do if it does not validate / check that the provided code is the expected one?

It's expected that the check code step, check the code :shrug:  

> if we validate the code only in step1, there is no point in having step0 IMHO

It still has value as the client could send a request to validate that the user does enter the correct code before continuing further (i.e.: requesting the other elements).

AureliaDolo: IMHO we should keep the code only in step 0 and that should have the effect to change the status in unverified email to code_checked, and then we can call step 1 without the code to actually create the account, only for unverified emails where the code has been checked.

That way we won't have an optional step during account creation. Kinda like during the invitation process, every step is needed.

If we want to repeat the code in step 1, there could be issues with the number of tries: 
- if step0 succeeds on the last try, it should fail on step1 with a valid code because it's out of tries
- if we don't check for the number of tries in either step, it's bad 

We could provide a one time use token on a successful step0 to provide at step1 to ensure there in no possibility that the call to step1 is made by someone different that step0

libparsec/crates/protocol/schema/anonymous_account_cmds/account_create_proceed.json5
```diff
@@ -59,6 +24,75 @@
...
+            ]
+          },
+          {
+            "name": "Number1Create",
+            "discriminant_value": "NUMBER_1_CREATE",
+            "fields": [
+              {
+                /// Code received by email following use of `account_create_send_validation_email`
+                "name": "validation_code",
+                "type": "String"
```
FirelightFlagboy: **todo:** We have a dedicated type for the code

```suggestion
                "type": "ValidationCode"
```

libparsec/crates/protocol/schema/anonymous_account_cmds/account_create_proceed.json5
```diff
@@ -59,6 +24,75 @@
...
+        "variants": [
+          {
+            /// This step is optional
+            "name": "Number0CheckCode",
+            "discriminant_value": "NUMBER_0_CHECK_CODE",
+            "fields": [
+              {
+                /// Code received by email following use of `account_create_send_validation_email`
+                "name": "validation_code",
+                "type": "String"
```
FirelightFlagboy: **todo:** 
```suggestion
                "type": "ValidationCode"
```

libparsec/crates/account/src/lib.rs
```diff
@@ -249,6 +270,144 @@ impl Account {
...
+        .send(account_create_proceed::Req {
+            account_create_step: req_step,
+        })
+        .await?
+    {
+        account_create_proceed::Rep::Ok => Ok(()),
+        account_create_proceed::Rep::AuthMethodIdAlreadyExists => {
+            Err(AccountCreateProceedError::AuthMethodIdAlreadyExists)
+        }
+        account_create_proceed::Rep::InvalidValidationToken => {
```
FirelightFlagboy: **todo (non-blocking):** Should be rename to `InvalidCode` but can be done later

libparsec/crates/serialization_format/src/protocol_python_bindings.rs
FirelightFlagboy, line 1: **note:** I've fixed the same thing in #10533

AureliaDolo:
> With the change needed to the protocol + server, I would tempted to directly update that, then do the bindings 🤔

@FirelightFlagboy I'd rather merge as is, to avoid having to rebase again, and so that the front guys can move on

FirelightFlagboy (approved):

PR #10440, opened 2025-06-02
Add account `PasswordAlgorithm` builtin type

touilleMan:


docs/rfcs/1013-parsec-account.md
```diff
@@ -153,6 +153,18 @@ on the server.
...
 (for the storing part) and [RFC 1015](1015-registration-device.md) (for how the registration device works).
 
+### 2.6 - Password vs password-less authentication methods
+
+From the client point of view, account authentication always starts by retrieving `auth_method_master_secret`,
+from which all `auth_method_id`/`auth_method_mac_key`/`auth_method_secret_key` are derived (see [`Keys`]).
+
+The detail of how to retrieve `auth_method_master_secret` depends of the authentication method, which can be of two types:
+
+- Standalone, for which the client is able to store `auth_method_master_secret` all by itself.
```
FirelightFlagboy: **nitpick:** Standalone does not mean much about how the password is retrieved, I would greatly prefer a more descriptive name

```suggestion
- Keyring, for which the client is able to store `auth_method_master_secret` all by itself.
```

FirelightFlagboy: **issue:** A Standalone auth method prevent connecting from another device.

We should anticipate that by either requiring that a password exist or display a message to the user to add a password auth method to its account.

docs/rfcs/1014-account-vault-for-device-stored-on-server.md
```diff
@@ -448,68 +422,39 @@ To list all protected device from all vault except current/active one:
...
+                        "name": "created_on",
+                        "type": "DateTime"
+                    },
+                    {
+                        "name": "disabled_on",
+                        "type": "RequiredOption<DateTime>"
+                    },
+                    {
+                        // TODO: Use `IPAddress` type once implemented
+                        // see https://github.com/Scille/parsec-cloud/issues/10384
```
FirelightFlagboy: **todo:** The issue is closed as not planned ;)

libparsec/crates/protocol/schema/authenticated_account_cmds/vault_item_recovery_list.json5
```diff
@@ -37,68 +37,39 @@
...
+                        "name": "created_on",
+                        "type": "DateTime"
+                    },
+                    {
+                        "name": "disabled_on",
+                        "type": "RequiredOption<DateTime>"
+                    },
+                    {
+                        // TODO: Use `IPAddress` type once implemented
+                        // see https://github.com/Scille/parsec-cloud/issues/10384
```
FirelightFlagboy: **todo:** The issue is closed as not planned ;)

docs/rfcs/1013-parsec-account.md
```diff
@@ -443,11 +457,15 @@ Anonymous account API:
...
-          "type": "PasswordAlgorithm"
-        },
+            // Auth method can be of two types:
+            // - Standalone, for which the client is able to store
+            //   `auth_method_master_secret` all by itself.
+            // - Password, for which the client must obtain some configuration
+            //   (i.e. this field !) from the server in order to know how
+            //   to turn the password into `auth_method_master_secret`.
+            "name": "auth_method_password_algorithm",
+            "type": "Option<PasswordAlgorithm>"
```
FirelightFlagboy: **nitpick:** I would prefer that we do not wrap in an Option, the value should indicate how to retrieve the key

docs/rfcs/1014-account-vault-for-device-stored-on-server.md
touilleMan: done.

libparsec/crates/protocol/schema/authenticated_account_cmds/vault_item_recovery_list.json5
touilleMan: done.

docs/rfcs/1013-parsec-account.md
touilleMan: > Standalone does not mean much about how the password is retrieved

That's because in standalone mode there is *no* password ^^

> I would greatly prefer a more descriptive name

I agree, standalone is probably too generic to be a good name !
(also I don't think `Keyring` is the right name either, since the authentication method can originate from something else than the OS keyring, e.g. a smartcard)

@mmmarcos @AureliaDolo any idea for a better name ?

To summarize what we are naming:

The account authentication system relies on `auth_method_master_secret` which can be considered as the root secret.
Here we are naming the strategies to obtain this `auth_method_master_secret`:
- Password obviously means we start from a password and derive it
- Standalone means we already know the `auth_method_master_secret`

The key point here is that a standalone strategy means the client doesn't need to obtain configuration from the server before being able to authenticate (as in password mode the algorithm used for the password derivation must be retrieved first).

I guess `ClientStored` or `ClientSideStored` then ?

touilleMan: > issue: A Standalone auth method prevent connecting from another device.

That's an interesting point !
I guess it depends on how the auth method is stored client-side, for instance storing it on a smartcard avoid this issue.

> We should anticipate that by either requiring that a password exist or display a message to the user to add a password auth method to its account.

However choosing to protect his account explicitly without password can also be a valid need: consider a corporate user that is only allowed to authenticate with smartcard to avoid having to deal with password and its associated risks.

So I would say the statue-quo is good enough here:
- Having a single authentication method and losing it falls into the "password lost" use-case, so it's already covered
- Right now we only have the password auth method really functional. I guess we will have a better view on best practices regarding this kind of issues when implementing other auth methods

docs/rfcs/1013-parsec-account.md
touilleMan: I agree, enums are better than options ;-)

So that was my plan originally, however I couldn't only have a `AuthMethodConfig` type since `auth_method_password_update` and `account_get_password_algorithm` both deal only with the password, hence a `PasswordAlgorithm` is also required.

So my 2nd attempt was to define `AuthMethodConfig` as an enum, basically:
```rust
enum AuthMethodConfig {
  Password(PasswordAlgorithm),
  Standalone
}
```
However going this way add yet another layer of temptation: since we have an enum, why not also list the different types of authentication possible:
```rust
enum AuthMethodConfig {
  Password(PasswordAlgorithm),
  KeyringStored,
  Fido2,
}
```

But the issue with this approach is that we need to update this list whenever we add a new authentication method type, while from the server point of view there is only a single things that matters: are we password (where config must be returned and only a single instance is allowed) or not.

So in the end I though  "are we password or not" translates well with a `Option<PasswordAlgorithm>`:
- It saved me from yet another types with related Python bindings containing inheritance :/
- Having an option means the field can be named `auth_method_password_algorithm` which is obvious. On the other hand having an enum means the field should be named something like `auth_method_config` making it much less obvious this is only a field about password algorithm maybe present...

docs/rfcs/1013-parsec-account.md
mmmarcos: It seems to me that it's more about _providing_ than _storing_ it. If it's the client that will provide it, how about:
- `Supplied` or `SuppliedByClient` or `ClientSupplied`
- `Provided` or `ProvidedByClient` or `ClientProvided`

In the sense that `Password` actually means "derived from password".

AureliaDolo: > the client doesn't need to obtain configuration from the server before being able to authenticate

related to that idea we could go with independent or autonomous, but maybe that's too close to standalone.

Otherwise CLientSupplied or ClientProvided sounds good.  (I have a slight preference for provided)

docs/rfcs/1013-parsec-account.md
AureliaDolo: :+1: for the option for now, maybe we could keep this reflection at the end of the rfc/ in a draft rfc to keep discussions open for the time we'll implement the other auth methods

server/parsec/client_context.py
```diff
@@ -176,9 +176,9 @@ class AnonymousAccountClientContext:
     settled_api_version: ApiVersion
     logger: ParsecBoundLogger = field(init=False)
     client_user_agent: str
-    # there is no guarantee that asgi is able to provide the ip address
+    # In ASGI the client IP address is optional
     # see https://asgi.readthedocs.io/en/latest/specs/www.html#http-connection-scope
-    client_ip: str | None  # TODO replace by ipaddr #10384
+    client_ip_address: str | None
```
AureliaDolo: nit: Is there a strong reason to keep the None variant ? In VaultItemRecoveryAuthMethod the string can be empty

edit: and we are not checking the ip format

server/parsec/components/account.py
```diff
@@ -124,9 +109,9 @@ async def create_account_with_password(
         human_label: str,
         created_by_user_agent: str,
         created_by_ip: str | None,
```
AureliaDolo: Same as above, we may allow the str to be empty

server/parsec/components/account.py
```diff
@@ -353,27 +336,28 @@ async def api_account_vault_item_recovery_list(
...
+                            password_algorithm = None
+                        case PasswordAlgorithmArgon2id() as a:
+                            password_algorithm = PasswordAlgorithmArgon2id(
+                                salt=a.salt,
+                                opslimit=a.opslimit,
+                                memlimit_kb=a.memlimit_kb,
+                                parallelism=a.parallelism,
                             )
+                        # `PasswordAlgorithm` is an abstract type
+                        case PasswordAlgorithm() as unknown:
```
AureliaDolo, line 351: In the same vein as the discussion on enum vs option for auth method, how far in the future do we expect to have other password algorithm ? And depending on that, is it worth to get rid of the abstract password algorithm ?

touilleMan: > how far in the future do we expect to have other password algorithm ?

There is only two reasons for having a new algorithm here:
- strong requirement from a client
- argon2 becoming less secure

So we are not planning on any new algorithm, but cannot rule out we won't have one to implement.

> is it worth to get rid of the abstract password algorithm ?

I'd rather keep it this way since:
- We have already [future-proofed our other use of crypto algorithms](https://github.com/Scille/parsec-cloud/blob/09bf901b552a135f7863c6aa835a57ec3d63f67e/libparsec/crates/types/schema/certif/realm_key_rotation_certificate.json5#L45-L64)
- Having `PasswordAlgorithm` defined as an enum means it is serialized with its type. if we had it as a struct instead, we would have a much tougher time handling deserialization once other algorithms are added (basically `PasswordAlgorithm` would eventually be turned into an enum, but during deserialization must account for the legacy case where it is a struct :/)
- Python bindings with inheritance is annoying to get right, so I'd rather have it already set (so that when we need to add the new algorithm type we can solely focus on what matter)

server/parsec/client_context.py
touilleMan: The conversion `None` -> `""` can take place at multiple levels:
1. when client_context is created
2. when the data is stored (e.g. in `MemoryAccountComponent.create_account`)
3. when the data is displayed (e.g. in `MemoryAccountComponent.vault_item_recovery_list`)

I was reluctant to do that early on in step 1 since it means "losing information" (i.e. the database contains cooked data, which is generally not a good idea since the need for cooking those data may change with time, so better store raw data and only do the cooking when fetching them)

However I think you are right here: if the cooking had an actual impact then we wouldn't consider `None` is equivalent to `""` in the first place !

So I'm changing this code to do the cooking in step 1 :+1: 
(I'm also using `client_ip_address: str | Literal[""]` as typing in order to make it extra-clear that the empty string is an allowed value)

docs/rfcs/1013-parsec-account.md
touilleMan: I've changed for `ClientProvided` then \o/

AureliaDolo (approved):

PR #10447, opened 2025-06-03
Server implementation of Account `auth_method_password_get_algorithm` API

touilleMan:


libparsec/crates/protocol/schema/anonymous_account_cmds/auth_method_password_get_algorithm.json5
FirelightFlagboy, line 1: **todo:** format file

server/tests/api_v5/anonymous_account/test_auth_method_password_get_algorithm.py
```diff
@@ -0,0 +1,69 @@
...
+            parallelism=1,
+        )
+    )
+
+    # Also ensure config is stable for a dummy address
+
+    rep2 = await anonymous_account.auth_method_password_get_algorithm(
+        email=EmailAddress("zack@example.com")
+    )
+    assert rep2 == rep
```
FirelightFlagboy: **nitpick:** Use a dedicated tests for that, it's different than getting an existing auth method

libparsec/crates/protocol/schema/anonymous_account_cmds/auth_method_password_get_algorithm.json5
touilleMan: :+1: done.

server/tests/api_v5/anonymous_account/test_auth_method_password_get_algorithm.py
touilleMan: :+1: done.

libparsec/crates/platform_device_loader/src/lib.rs
```diff
@@ -23,6 +23,9 @@ use native as platform;
 #[cfg(target_arch = "wasm32")]
 use web as platform;
 
+// ⚠️ Changing Argon2id default config also require updating the fake config generator
+// on server side (see `server/parsec/components/account.py`), otherwise it would
+// become trivial for a attacker to determine which config is fake !
```
mmmarcos:
```suggestion
// ⚠️ Changing Argon2id default config also requires updating the fake config generator
// on server side (see `server/parsec/components/account.py`), otherwise it would
// become trivial for an attacker to determine which config is fake!
```

FirelightFlagboy (approved):
LGTM minor the spelling issue found by the CI

server/parsec/components/account.py
```diff
@@ -96,6 +96,23 @@ class BaseAccountComponent:
     def __init__(self, config: BackendConfig):
         self._config = config
 
+    async def get_password_algorithm_or_fake_it(self, email: EmailAddress) -> PasswordAlgorithm:
```
mmmarcos, line 99: I would remove the `or_fake_it` from the name and add more detail in docstring (see below)
```suggestion
    async def get_password_algorithm(self, email: EmailAddress) -> PasswordAlgorithm:
```

server/parsec/components/account.py
```diff
@@ -96,6 +96,23 @@ class BaseAccountComponent:
...
+        """
+        Note that if email doesn't correspond to a valid account, a made-up
+        algorithm config is returned.
```
mmmarcos:
```suggestion
        Return the password algorithm configuration corresponding to the account
        of the given email address.

        If the email address doesn't correspond to a valid account, a fake (stable) 
        configuration is returned in order to prevent an attacker guessing whether 
        the email address is present in the database or not.
```

server/parsec/components/account.py
```diff
@@ -96,6 +96,23 @@ class BaseAccountComponent:
...
+        """
+        raise NotImplementedError
+
+    def _generate_fake_password_algorithm(self, email: EmailAddress) -> PasswordAlgorithm:
+        # TODO: the generated config must be realistic, unpredictable and stable
+        #       (introduce a `SECRET_KEY` environ variable and derive from there ?)
```
mmmarcos, line 112: :+1:

server/parsec/components/memory/account.py
```diff
@@ -45,6 +45,22 @@ def __init__(
...
+        except KeyError:
+            active_authentication_methods = ()
+
+        for auth_method in active_authentication_methods:
+            if auth_method.password_algorithm is not None:
+                return auth_method.password_algorithm
+
+        else:
+            # No password found
+            return self._generate_fake_password_algorithm(email)
```
mmmarcos: The `else` seems unnecessary since the `for` has no `break` (only a `return`). 
```suggestion
        # Account does not exists or it does not have an active password algorithm.
        # In either case, a fake configuration is returned. 
        return self._generate_fake_password_algorithm(email)
```

libparsec/crates/platform_device_loader/src/lib.rs
touilleMan: :+1: done.

server/parsec/components/account.py
touilleMan: I'd rather keep the naming as-is since it is very exotic to return dummy data instead of returrning `None` or raising an exception

server/parsec/components/account.py
touilleMan: :+1: done.

server/parsec/components/memory/account.py
touilleMan: :+1: done.

PR #10449, opened 2025-06-03
feat(server): Bump pyo3 to v22

FirelightFlagboy:
- Import trait `PyModuleMethods`
- Enable `gil-refs` feature

  Allow to use `as_byes` and `FromPyObject` on some types

- fix incompatible types

  Replace `to_owned` to `into_py(py)`

- Add required signature

  <https://pyo3.rs/v0.25.0/migration.html#deprecation-of-implicit-default-for-trailing-optional-arguments>

- Ignore warning about useless conversion

  <https://github.com/PyO3/pyo3/issues/4828>

- fix unexpected feature when using create_exception

  <https://github.com/PyO3/pyo3/issues/4743>


mmmarcos:
Is this bump needed for a specific issue (if so, which one)? or is it regular dependency maintenance?

Is there any reason why not bumping to higher versions such as v23, v24 or v25?

FirelightFlagboy:
> Is this bump needed for a specific issue (if so, which one)? or is it regular dependency maintenance?

The later: dependency maintenance.
Dependabot never proposed to bump pyo3.
So it's only recently I've discover that we are multiple version behind with breaking change that prevent us to bump the edition of the Rust code to 2024

PR #10452, opened 2025-06-03
feat(server): Add deleted_on on Account model

FirelightFlagboy:
Closes https://github.com/Scille/parsec-cloud/issues/10451

### DIFF
--- a/server/parsec/components/memory/account.py
+++ b/server/parsec/components/memory/account.py
@@ -117,18 +117,14 @@ class MemoryAccountComponent(BaseAccountComponent):
         """Use only in tests, nothing is checked."""
         return self._data.unverified_emails[email][0]
 
-    def auth_method_id_already_exists(self, id: AccountAuthMethodID) -> bool:
-        for account in self._data.accounts.values():
-            for vault in (account.current_vault, *account.previous_vaults):
-                if id in vault.authentication_methods:
-                    return True
-        return False
+    def auth_method_id_already_exists(self, auth_method_id: AccountAuthMethodID) -> bool:
+        return self._data.get_account_from_any_auth_method(auth_method_id) is not None
 
     @override
     async def vault_item_upload(
         self, auth_method_id: AccountAuthMethodID, item_fingerprint: HashDigest, item: bytes
     ) -> None | AccountVaultItemUploadBadOutcome:
-        match self._data.get_account_from_auth_method(auth_method_id=auth_method_id):
+        match self._data.get_account_from_active_auth_method(auth_method_id=auth_method_id):
             case (account, _):
                 pass
             case None:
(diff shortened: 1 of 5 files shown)


FirelightFlagboy:
Any suggestion how to test that? we have multiple API that is affected

server/parsec/components/memory/account.py
```diff
@@ -178,10 +178,8 @@ async def vault_key_rotation(
             case None:
                 return AccountVaultKeyRotation.ACCOUNT_NOT_FOUND
 
-        for account in self._data.accounts.values():
-            for vault in (account.current_vault, *account.previous_vaults):
-                if new_auth_method_id in vault.authentication_methods:
-                    return AccountVaultKeyRotation.NEW_AUTH_METHOD_ID_ALREADY_EXISTS
+        if self.auth_method_id_already_exists(new_auth_method_id):
```
touilleMan, line 177: BTW can you change `auth_method_id_already_exists` implementation for something similar to the old code here (since the current implementation relies on iterating over each element instead of taking advantage of the fact those elements are keys in a dict)

server/parsec/components/memory/account.py
```diff
@@ -119,13 +119,9 @@ def test_get_token_by_email(self, email: EmailAddress) -> EmailValidationToken |
 
     def auth_method_id_already_exists(self, id: AccountAuthMethodID) -> bool:
         for account in self._data.accounts.values():
-            for auth_id in account.current_vault.authentication_methods.keys():
-                if auth_id == id:
+            for vault in (account.current_vault, *account.previous_vaults):
```
touilleMan: :+1:

server/parsec/components/memory/datamodel.py
```diff
@@ -111,6 +111,9 @@ def get_account_from_auth_method(
         self, auth_method_id: AccountAuthMethodID
     ) -> tuple[MemoryAccount, MemoryAuthenticationMethod] | None:
         for account in self.accounts.values():
+            # Skip deleted account
```
touilleMan: I would expect that all auth methods for a given account gets deactivated when said account is deleted.

- Question: Is this the case ?
- We should document this, typically adding a comment next to `MemoryAccount.deleted_on` field in the datamodel (e.g. `Note any active auth methods gets disabled when the account is deleted`)
- If that's the case then we don't need to check for account deleted here (we can have a sanity check assert to ensure the account found is not deleted though)

EDIT: done.

touilleMan:
> Any suggestion how to test that? we have multiple API that is affected

I think there is no tests to add right now since the `deleted_on` is never used (since even `get_account_from_auth_method` only uses it as a sanity check)
We will add tests for it when we add the delete account API.

server/parsec/components/memory/datamodel.py
FirelightFlagboy: In this PR only the field `deleted_on` was added on `MemoryAccount`.
For the rest is to be define.

I'll put additional information to the parent issue, because disabling the auth methods add another safeguard in case we forget to check the account's `deleted_on`

PR #10458, opened 2025-06-03
Remove section Advanced concepts from the user docs

mmmarcos:


mmmarcos:
> Do we have a copy somewhere?

I don't think so. I'll add it to a Parsec workspace within the Scille organization.

mmmarcos:
> > Do we have a copy somewhere?
> 
> I don't think so. I'll add it to a Parsec workspace within the Scille organization.

Done. (Uploaded to Parsec (Technique)

PR #10460, opened 2025-06-03
test(account): Refactor fixture to use component method

FirelightFlagboy:
Create the alice & bob account use the `MemoryAccountComponent` methods

### DIFF
--- a/server/tests/common/account.py
+++ b/server/tests/common/account.py
@@ -12,15 +12,16 @@ import base64
 import pytest
 from httpx import AsyncClient
 
-from parsec._parsec import AccountAuthMethodID, DateTime, EmailAddress, SecretKey
-from parsec.backend import Backend
-from parsec.components.account import PasswordAlgorithmArgon2ID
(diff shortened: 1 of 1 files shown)


server/tests/common/account.py
```diff
@@ -40,35 +41,56 @@
...
+    created_by_user_agent: str,
+    created_on: DateTime,
+    human_label: str,
+    mac_key: SecretKey,
+    password_secret_algorithm: PasswordAlgorithmArgon2ID,
+    vault_key_access: bytes,
+):
+    assert account_email not in account_component._data.accounts
+    email_token = EmailValidationToken.new()
+    account_component._data.unverified_emails[account_email] = (email_token, created_on)
```
touilleMan: There is still a TODO to put here since this code won't work with the PostgreSQL implementation

touilleMan (approved):

server/tests/common/account.py
FirelightFlagboy: Instead of adding a todo, I've directly use the method to get an email token

PR #10470, opened 2025-06-04
Improve server pydantic types

touilleMan:


server/parsec/config.py
```diff
@@ -253,6 +253,7 @@ class BackendConfig:
         default_factory=lambda: ActiveUsersLimit.NO_LIMIT
     )
     organization_initial_user_profile_outsider_allowed: bool = True
+    # Note minimum archiving period must be a positive value !
     organization_initial_minimum_archiving_period: int = 2592000  # seconds (i.e 30 days)
```
FirelightFlagboy, line 257: **question:** Why we do not use `NonNegativeInt` here? likewise for `sse_keepalive`

FirelightFlagboy: Annotation are not used by pyright for type checking

server/parsec/types.py
```diff
@@ -37,39 +50,130 @@ class UnsetType(Enum):
...
+
+# -- Field types annotations --------------------------------------------------
+
+
+# Base64BytesField
+#
+# Why a custom type if Pydantic already has `Base64Bytes` type?
+#
+# Long story short, Pydantic's `Base64Bytes` wants the input to deserialize to
+# be bytes, which doesn't work for JSON since a str is provided.
```
FirelightFlagboy: **typo:** 

> deser[..] to be bytes

mmmarcos: Not following... what is the typo?

FirelightFlagboy: The phrasing is feel strange to me at the "to be bytes": "the input to deserialize to be bytes"

mmmarcos: Yep, I agree. I've updated the phrasing.

FirelightFlagboy (approved):

PR #10471, opened 2025-06-04
Introduce `libparsec_account::Account`

touilleMan:


libparsec/crates/account/src/lib.rs
```diff
@@ -1 +1,100 @@
...
+pub const AUTH_METHOD_ID_DERIVATION_UUID: uuid::Uuid =
+    uuid::uuid!("00000000-0000-0000-0000-000000000000");
+pub const AUTH_METHOD_MAC_KEY_DERIVATION_UUID: uuid::Uuid =
+    uuid::uuid!("11111111-1111-1111-1111-111111111111");
+pub const AUTH_METHOD_SECRET_KEY_DERIVATION_UUID: uuid::Uuid =
+    uuid::uuid!("22222222-2222-2222-2222-222222222222");
+
+#[derive(Debug)]
+pub struct Account {
+    #[allow(unused)]
```
FirelightFlagboy: **todo:** 
```suggestion
    #[expect(unused)]
```

libparsec/crates/crypto/src/sodiumoxide/key_derivation.rs
```diff
@@ -28,17 +32,42 @@ impl KeyDerivation {
...
     pub fn derive_secret_key_from_uuid(&self, id: uuid::Uuid) -> SecretKey {
+        let raw: [u8; SecretKey::SIZE] = self.derive_raw_form_uuid(id).into();
+        SecretKey::from(raw)
+    }
+
+    pub fn derive_uuid_from_uuid(&self, id: uuid::Uuid) -> uuid::Uuid {
+        uuid::Uuid::from_bytes(self.derive_raw_form_uuid(id).into())
+    }
+
+    fn derive_raw_form_uuid<Size>(&self, id: uuid::Uuid) -> GenericArray<u8, Size>
```
FirelightFlagboy: **typo:** 
```suggestion
    fn derive_raw_from_uuid<Size>(&self, id: uuid::Uuid) -> GenericArray<u8, Size>
```

libparsec/crates/crypto/src/rustcrypto/key_derivation.rs
```diff
@@ -25,6 +30,18 @@ impl KeyDerivation {
...
 
     pub fn derive_secret_key_from_uuid(&self, id: uuid::Uuid) -> SecretKey {
+        SecretKey(self.derive_raw_form_uuid(id))
+    }
+
+    pub fn derive_uuid_from_uuid(&self, id: uuid::Uuid) -> uuid::Uuid {
+        uuid::Uuid::from_bytes(self.derive_raw_form_uuid(id).into())
+    }
+
+    fn derive_raw_form_uuid<Size>(&self, id: uuid::Uuid) -> GenericArray<u8, Size>
```
FirelightFlagboy: **typo:** 
```suggestion
    fn derive_raw_from_uuid<Size>(&self, id: uuid::Uuid) -> GenericArray<u8, Size>
```

server/parsec/components/account.py
```diff
@@ -96,6 +96,27 @@ class BaseAccountComponent:
...
+        raise NotImplementedError
+
+    def _generate_fake_password_algorithm(self, email: EmailAddress) -> PasswordAlgorithm:
+        # TODO: the generated config must be realistic, unpredictable and stable
+        #       (introduce a `SECRET_KEY` environ variable and derive from there ?)
+        return PasswordAlgorithmArgon2id(
+            memlimit_kb=128 * 1024,  # 128 Mo
+            opslimit=3,
+            parallelism=1,
+            salt=b"0" * 16,
```
FirelightFlagboy: **issue:** Use another value instead of a salt with full 0

Until we stabilize the fake algo gen, I would prefer that we use a different salt that is not composed of a single value.

I propose:

- Email
- `sha256(email)[:16]`
- A RNG

FirelightFlagboy (changes requested):

libparsec/crates/crypto/src/sodiumoxide/utils.rs
```diff
@@ -26,3 +35,33 @@ pub(super) use impl_try_from;
...
+) -> Result<GenericArray<u8, Size>, CryptoError>
+where
+    Size: ArrayLength<u8> + IsLessOrEqual<U64>,
+    LeEq<Size, U64>: NonZero,
+{
+    let mut key = GenericArray::<u8, Size>::default();
+
+    // Libsodium only support parallelism of 1
+    if parallelism != 1 {
+        return Err(CryptoError::DataSize);
```
AureliaDolo: Could it be useful to add the info on which datasize is invalid ? (example: `Datasize("parallelism")`) Albeit it would be done in another issue.

server/parsec/components/account.py
AureliaDolo: :+1: for the rng for the unpredictability

libparsec/crates/account/src/lib.rs
touilleMan: :+1: done.

libparsec/crates/crypto/src/sodiumoxide/key_derivation.rs
touilleMan: good catch ! done.

libparsec/crates/crypto/src/sodiumoxide/utils.rs
touilleMan: I think the wrong error variant is used here, it would better to use `CryptoError::Algorithm("Argon2id with paralellism > 1 is not supported".to_string())` (whose error message corresponds `Unsupported algorithm: {0}` well to our usecase)

Can you create an issue for this ?

server/parsec/components/account.py
touilleMan: > I would prefer that we use a different salt that is not composed of a single value.

I chose to use a salt full of 0 to make it more obvious this is temporary placeholder code, but I guess it's a matter of taste ^^

> 👍 for the rng for the unpredictability

https://xkcd.com/221/ :trollface: 

More seriously, RNG is not a good solution since we require a stable output (otherwise the attacker can call the route twice and check if the outputs changes)
(Also note neither is `sha56(email)` since, while stable, the outpout is predictable since the attacker can compute it on its side and compare it to what the server returns...) 



Anyway, I've changed the code to have a proper implementation of `_generate_fake_password_algorithm` \o/

libparsec/crates/crypto/src/sodiumoxide/utils.rs
mmmarcos: https://github.com/Scille/parsec-cloud/issues/10514

server/parsec/config.py
```diff
@@ -238,6 +238,11 @@ class BackendConfig:
     # Bearer token used to authenticate the administration API
     administration_token: str
 
+    # Random value used to make unpredictable (but still stable & realistic) the password
+    # algorithm configuration returned for non-existing accounts, hence preventing an
+    # attacker from using it as an oracle do determine if a given email has an account.
```
mmmarcos:
```suggestion
    # attacker from using it as an oracle to determine if a given email has an account.
```

libparsec/crates/account/src/lib.rs
```diff
@@ -1 +1,88 @@
 // Parsec Cloud (https://parsec.cloud) Copyright (c) BUSL-1.1 2016-present Scille SAS
+
```
mmmarcos, line 2: I'd add a REAMDE to briefly document this new crate.
```suggestion
#![doc = include_str!("../README.md")]

```

OR at least describe the crate's goal/scope in the `Internal crates` section of https://github.com/Scille/parsec-cloud/blob/master/libparsec/README.md?plain=1#L17

libparsec/crates/account/src/lib.rs
```diff
@@ -1 +1,88 @@
...
+}
+
+impl Account {
+    pub fn new(
+        config_dir: &Path,
+        proxy: ProxyConfig,
+        addr: ParsecAddr,
+        auth_method_master_secret: KeyDerivation,
+    ) -> anyhow::Result<Self> {
+        let auth_method = AccountAuthMethod {
```
mmmarcos, line 51: This is already specified elsewhere, but it might be useful here too
```suggestion
        // The `auth_method_master_secret` is derived locally, 
        // no information is required from the server.
        let auth_method = AccountAuthMethod {
```

libparsec/crates/account/src/lib.rs
```diff
@@ -1 +1,88 @@
...
+    }
+
+    pub async fn from_password(
+        config_dir: &Path,
+        proxy: ProxyConfig,
+        addr: ParsecAddr,
+        email: EmailAddress,
+        password: Password,
+    ) -> Result<Self, AccountFromPasswordError> {
+        let password_algorithm = {
```
mmmarcos, line 71: This is already specified elsewhere, but it might be useful here too
```suggestion
        // The password algorithm configuration is obtained from the server
        // to know how to turn the password into `auth_method_master_secret`.
        let password_algorithm = {
```

mmmarcos (approved):
LGTM specially `./server`. 

For libparsec I would describe the goal and scope of the new `account` crate (either in `./libparsec/crates/account/README.md` or in `./libparsec/README.md`)

.github/workflows/package-cli.yml
FirelightFlagboy, line 1: **todo (blocking):** Remove the cherry-picked commit that add that and rebase.

It was fixed in #10513

libparsec/crates/crypto/src/common/password.rs
```diff
@@ -0,0 +1,83 @@
...
+
+use crate::{compute_from_password, CryptoError, KeyDerivation, Password, SecretKey};
+
+// ⚠️ Changing Argon2id default config also requires updating the fake config generator
+// on server side (see `server/parsec/components/account.py`), otherwise it would
+// become trivial for a attacker to determine which config is fake !
+pub const ARGON2ID_DEFAULT_MEMLIMIT_KB: u32 = 128 * 1024; // 128 Mo
+pub const ARGON2ID_DEFAULT_OPSLIMIT: u32 = 3;
+// Be careful when changing parallelism: libsodium only supports 1 thread !
+pub const ARGON2ID_DEFAULT_PARALLELISM: u32 = 1;
```
FirelightFlagboy, line 25: **note:** Those are valid values following `OWASP`

https://cheatsheetseries.owasp.org/cheatsheets/Password_Storage_Cheat_Sheet.html#argon2id

libparsec/crates/crypto/src/rustcrypto/key_derivation.rs
```diff
@@ -34,11 +51,11 @@ impl KeyDerivation {
         let mut personal: [u8; 16] = [0u8; 16];
```
FirelightFlagboy: **nitpick:** Remove the duplicate typing

```suggestion
        let mut personal = [0u8; 16];
```

libparsec/crates/crypto/src/rustcrypto/key_derivation.rs
```diff
@@ -34,11 +51,11 @@ impl KeyDerivation {
...
         personal[..8].copy_from_slice(&id.as_bytes()[8..]);
 
-        let subkey = Blake2bMac256::new_with_salt_and_personal(&self.0, &salt, &personal)
+        let subkey = Blake2bMac::<Size>::new_with_salt_and_personal(&self.0, &salt, &personal)
```
FirelightFlagboy: **question:** Half of `personal` remain empty (full of 0) is it expected ?

Can we use a smaller array or it's the implementation that require an array of 16 bytes

libparsec/crates/crypto/src/rustcrypto/password.rs
```diff
@@ -0,0 +1,47 @@
...
+            let argon = Argon2::new(
+                Algorithm::Argon2id,
+                Version::V0x13,
+                Params::new(
+                    *memlimit_kb,
+                    *opslimit,
+                    *parallelism,
+                    Some(Size::to_usize()),
+                )
+                .map_err(|_| CryptoError::DataSize)?,
```
FirelightFlagboy, line 37: **nitpick:** Returning `DataSize` would be confusing if we need to debug an issue with that

```suggestion
                .map_err(|_| CryptoError::InvalidParam)?,
```

libparsec/crates/crypto/src/sodiumoxide/key_derivation.rs
```diff
@@ -28,17 +32,30 @@ impl KeyDerivation {
...
     pub fn derive_secret_key_from_uuid(&self, id: uuid::Uuid) -> SecretKey {
+        let raw: [u8; SecretKey::SIZE] = self.derive_raw_from_uuid(id).into();
+        SecretKey::from(raw)
+    }
+
+    pub fn derive_uuid_from_uuid(&self, id: uuid::Uuid) -> uuid::Uuid {
+        uuid::Uuid::from_bytes(self.derive_raw_from_uuid(id).into())
+    }
+
+    fn derive_raw_from_uuid<Size>(&self, id: uuid::Uuid) -> GenericArray<u8, Size>
```
FirelightFlagboy, line 43: **issue (blocking):** The `id` is not used in the same way as `rustcrypto` impl

Here we use the full value, but the `rustcrypto` impl only use the first-half

server/parsec/cli/run.py
```diff
@@ -247,6 +250,31 @@ def handle_parse_result(
     show_envvar=True,
     help="Delay before resending an account creation confirmation email (in seconds)",
 )
+@click.option(
+    "--fake-account-password-algorithm-seed",
+    show_default=True,
+    envvar="PARSEC_FAKE_ACCOUNT_PASSWORD_ALGORITHM_SEED",
```
FirelightFlagboy, line 257: **todo:** Update the hosting documentation

server/parsec/cli/run.py
```diff
@@ -247,6 +250,31 @@ def handle_parse_result(
...
+    ),
+    help="""Random value used to make unpredictable the password algorithm configuration
+    returned for non-existing accounts.
+
+    By default a random value will be generated at each startup, however setting a value
+    once and for all offers better protection against attackers trying to determine
+    if a given email has an account.
+
+    A typical way to generate a good seed is to use:
+        $ python -c 'import secrets; print(secrets.token_hex())'
```
FirelightFlagboy: **nitpick:** In the hosting documentation, we already use `openssl` to generate the admin token, I would reuse the same method:

```shell
openssl rand 63 | base64 --wrap=86
```

FirelightFlagboy (changes requested):

server/parsec/config.py
touilleMan: :+1: done.

libparsec/crates/account/src/lib.rs
touilleMan: Good point, I've added a section in `libparsec/README.md`

libparsec/crates/account/src/lib.rs
touilleMan: Done. I've also added a comment next to the definition of the derivation constants (e.g. `AUTH_METHOD_ID_DERIVATION_UUID`)

libparsec/crates/account/src/lib.rs
touilleMan: :+1: done.

.github/workflows/package-cli.yml
touilleMan: done.

libparsec/crates/crypto/src/common/password.rs
touilleMan: This configuration was taken from libsodium's MODERATE profile

I've added a comment to explain this, and also precise this is in range of what recommends OWASP :+1:

libparsec/crates/crypto/src/rustcrypto/key_derivation.rs
touilleMan: :+1: done.
(and also for the `salt` param)

libparsec/crates/crypto/src/rustcrypto/key_derivation.rs
touilleMan: > Half of personal remain empty (full of 0) is it expected ?

Yes, this code is strictly following what libsodium does (and we have a test to ensure the output is consistent between both implementations)

See the original code where the last 8 bytes of the 16 bytes array being zeroed: https://github.com/jedisct1/libsodium/blob/4a15ab7cd0a4b78a7356e5f488d5345b8d314549/src/libsodium/crypto_kdf/blake2b/kdf_blake2b.c#L40

> Can we use a smaller array or it's the implementation that require an array of 16 bytes

Again, we strictly follow what some people smarter than us have done in libsodium ^^

libparsec/crates/crypto/src/rustcrypto/password.rs
touilleMan: ~done.~

This is not possible after all: [libsodium's `derive_key` function](https://docs.rs/sodiumoxide/latest/sodiumoxide/crypto/pwhash/argon2id13/fn.derive_key.html) doesn't provide a dedicated error if the parameter have the wrong value, so we must stick to a single error type that correspond to invalid config / salt too small / etc.

libparsec/crates/crypto/src/sodiumoxide/key_derivation.rs
touilleMan: In both implementation we split `id` into two 8bytes parts, then provide them as the two parameters blake2b takes.
However those two parameters are named differently:
- `subkey_id` & `context` in libsodium
- `salt` & `personal` in RustCrypto

This makes things a bit confusing, but under the hood they both strictly correspond to the same things

Also this code has not been introduced from this PR (only the API has been slightly modified), and there is a test ensuring both implementation are equivalent ;-)

FirelightFlagboy: I didn't see how the salt was generated on `rustcrypto`.

Could you normalize the name of `id_low` & `id_high` for `salt` & `personal`?

server/parsec/cli/run.py
touilleMan: I'd rather keep the python trick since having a base64 is less convenient:

```
$ openssl rand 63 | base64 --wrap=86
KGT1sMbJ3l5XHEHQqA/WLtMyT9WnlR5u6i27++14remGwGwuRRrWLPDiry/KN33viSeDs+X8H8rUYPcjpF+L
$ python -c 'import secrets; print(secrets.token_hex())'
bb3f4f92ec3d6b6b72b1ff33726ca9f451da2f7b275c3dba4c3c184aacc844ff
```

The Python token uses 32 bytes of entropy which is strong enough, and the punctuation characters used in base64 doesn't play well with double click selection.

However using `openssl` makes for a simpler command, so I've updated both my `python ...` and the existing `openssl rand 63 | base64 --wrap=86` to `openssl rand -hex 32` instead

server/parsec/cli/run.py
touilleMan: :+1:  done.

docs/HOSTING.md
FirelightFlagboy, line 1: **thought:** Should we remove that file? We already have `docs/hosting/deployment`

docs/hosting/deployment/index.rst
```diff
@@ -110,13 +110,11 @@ Below you will find a simple script to generate a token:
...
   :linenos:
 
-The script will generate a random token (:bash:`openssl rand 63 | base64 --wrap=86`) and create the env file ``parsec-admin-token.env``
+The script will generate a random token (:bash:`openssl rand -hex 32`) and create the env file ``parsec-admin-token.env``
 
 .. note::
 
-  The step :bash:`TOKEN=$(openssl rand 63 | base64 --wrap=86)` could also be replaced by a value generated by a password-generator for example.
-
-  It doesn't need to be encoded in ``base64`` (we encode it in the script just to have printable characters).
```
FirelightFlagboy, line 119: **nitpick:** I would conserve this phrase, as we do not require the TOKEN to only be hex value

docs/hosting/deployment/index.rst
FirelightFlagboy, line 1: **todo:** You did not mention the generation of the secret `PARSEC_FAKE_ACCOUNT_PASSWORD_ALGORITHM_SEED`.

You would put it in `gen-admin-token.sh` (maybe also renaming that script) since it also generate another token.

libparsec/crates/account/src/lib.rs
```diff
@@ -1 +1,98 @@
...
+                Rep::Ok { password_algorithm } => password_algorithm,
+                bad_rep @ Rep::UnknownStatus { .. } => {
+                    return Err(anyhow::anyhow!("Unexpected server response: {:?}", bad_rep).into());
+                }
+            }
+        };
+
+        let auth_method_master_secret = password_algorithm
+            .compute_key_derivation(&password)
+            .map_err(AccountFromPasswordError::BadPasswordAlgorithm)?;
```
FirelightFlagboy, line 90: **issue:** The algorithm values are not validated

The server could send a weak configuration to try to brute force the master key

libparsec/crates/crypto/src/common/password.rs
```diff
@@ -0,0 +1,90 @@
...
+    pub fn generate_fake_from_seed(email: &str, seed: &SecretKey) -> Self {
+        // Currently there is only a single algorithm type in use and
+        // no customization over its parameters, so we only need to generated
+        // a fake salt.
+        // Note that we have a test (`test_libparsec_actual_password_algorithm`)
+        // doing some dark magic in order to detect whenever this changes.
+
+        let raw = seed.mac_512(email.as_bytes());
+
+        let salt = raw[..16].to_vec();
```
FirelightFlagboy: **nitpick:** From where the `16` come from? Can you use a constant?

libparsec/crates/crypto/src/rustcrypto/key_derivation.rs
```diff
@@ -25,20 +30,32 @@ impl KeyDerivation {
...
+    {
         // We follow what libsodium does here
         // (see https://github.com/jedisct1/libsodium/blob/4a15ab7cd0a4b78a7356e5f488d5345b8d314549/src/libsodium/crypto_kdf/blake2b/kdf_blake2b.c#L31-L52)
 
-        let mut salt: [u8; 16] = [0u8; 16];
+        let mut salt = [0u8; 16];
         salt[..8].copy_from_slice(&id.as_bytes()[..8]);
 
-        let mut personal: [u8; 16] = [0u8; 16];
+        let mut personal = [0u8; 16];
```
FirelightFlagboy: **nitpick (if-minor):** Can we use constants for those `16` values?

libparsec/crates/crypto/tests/password.rs
```diff
@@ -0,0 +1,178 @@
...
+                    opslimit: 1,
+                    memlimit_kb: 8,
+                    parallelism: 0,
+                };
+
+                p_assert_matches!(
+                    algo.$name(&"Passw0rd".to_owned().into()),
+                    Err(CryptoError::DataSize)
+                );
+            }
```
FirelightFlagboy: **nitpick:** Could be factorized with `rstest`

libparsec/crates/crypto/src/sodiumoxide/key_derivation.rs
touilleMan: I'd rather keep the naming as it is:
- There is a comment in the RustCrypto implementation explaining the reasoning for the code with a link on the libsodium implementation
- The parameters `salt/personal` vs `subkey_id/context` correspond to the name of the param in the function that are called ( `Blake2bMac::new_with_salt_and_personal` for rustcrypto vs `derive_from_key` for sodiumoxide)
- The parameters are cooked in a not obvious way while their types are the same, thus is it easy to mess up the order those parameters should be passed. Using the same name for the variable and the function parameter makes this much more obvious.

server/parsec/_parsec_pyi/crypto.pyi
```diff
@@ -106,3 +102,30 @@ class SequesterVerifyKeyDer:
 def generate_nonce() -> bytes: ...
 
 class CryptoError(Exception): ...
+
+class PasswordAlgorithm:
+    @staticmethod
+    def generate_argon2id() -> PasswordAlgorithm: ...
+    @staticmethod
+    def generate_fake_from_seed(email: str, seed: SecretKey) -> PasswordAlgorithm: ...
```
FirelightFlagboy, line 110: **todo (blocking):** Every other places use `EmailAddress`, can you add a comment why the exception ?

server/parsec/cli/run.py
```diff
@@ -247,6 +250,31 @@ def handle_parse_result(
...
+    envvar="PARSEC_FAKE_ACCOUNT_PASSWORD_ALGORITHM_SEED",
+    show_envvar=True,
+    callback=(
+        lambda ctx, param, value: SecretKey(secrets.token_bytes(32))
+        if not value
+        else SecretKey(
+            blake2b(
+                b"fake_account_password_algorithm_seed" + value.encode("utf8"), digest_size=32
+            ).digest()
+        )
```
FirelightFlagboy: **nitpick:** IMO I would not provide a default value as we do not want it to be used in production.

server/parsec/config.py
```diff
@@ -238,6 +238,11 @@ class BackendConfig:
     # Bearer token used to authenticate the administration API
     administration_token: str
 
+    # Random value used to make unpredictable (but still stable & realistic) the password
+    # algorithm configuration returned for non-existing accounts, hence preventing an
+    # attacker from using it as an oracle to determine if a given email has an account.
+    fake_account_password_algorithm_seed: SecretKey
```
FirelightFlagboy, line 240: **todo:** Should be put in `AccountConfig`

docs/hosting/deployment/index.rst
touilleMan: :+1: done.

libparsec/crates/crypto/src/common/password.rs
touilleMan: Good point ! I've reworked this part (and factorized random generation code)

libparsec/crates/crypto/src/rustcrypto/key_derivation.rs
touilleMan: done. (constant with a link on libsodium's code)

libparsec/crates/crypto/tests/password.rs
touilleMan: :+1: done.

server/parsec/config.py
touilleMan: I've removed the `AccountConfig` entirely instead, there is no reason to separated the config related to account this way (especially since there is config such as `organization_initial_account_vault_strategy` that make blurry the separation)

Note the other sub-config structures  are required because they correspond to different possible configurations (e.g. `BaseBlockStoreConfig` that can be S3 or Swift each with different fields).

server/parsec/_parsec_pyi/crypto.pyi
touilleMan: `EmailAddress` is defined in `libparsec_types` which depends on `libparsec_crypto`, hence we cannot use `EmailAddress` type here.

But that is not an actual issue: `email` parameter is only used here as a raw value, we could have called it `discriminant` instead of `email` (but calling it `email` makes it more obvious of course !)

I've added a comment about this :+1:

server/parsec/cli/run.py
touilleMan: Good call, I've removed the default behavior :+1:

docs/hosting/deployment/index.rst
touilleMan: :+1: done.

I didn't renamed that script considering `PARSEC_FAKE_ACCOUNT_PASSWORD_ALGORITHM_SEED` is not very important.

BTW, I wonder if we should simplify the documentation on the `.env` files: the documentation they contain is less good than the one obtains from `parsec run --help` (or from `HOSTING.md`)
Given it's complicated to keep the documentation in sync everywhere, what we could do:
- add a small precommit script to check that each environ variable in `run.py` is also present in `HOSTING.md`. This would prevent forgetting to update `HOSTING.md` every-time we add a new config var (it won't detect more advanced case like when the description of a param is updated, but it is a good low-hanging fruit anyway ^^)
- remove the `.env` files and replace them by a single `.env` file that contains only the mandatory params, with each param commented (so that using this config without doing the necessary customization won't do anything)
- remove `gen-admin-token.sh` , we can add a note explaining the `openssl` command to generate token

WDYT ?

docs/HOSTING.md
```diff
@@ -277,38 +300,39 @@ Example:
...
 
-- ``--dev``
+- `--dev`
 
 Equivalent to:
 
 ```txt
 --debug --db=MOCKED --blockstore=MOCKED --administration-token=s3cr3t
 --email-sender=no-reply@parsec.com --email-host=MOCKED
+ --fake-account-password-algorithm-seed=F4k3
```
FirelightFlagboy: **nitpick:** 
```suggestion
--fake-account-password-algorithm-seed=F4k3
```

docs/hosting/deployment/gen-admin-token.sh
```diff
@@ -3,8 +3,12 @@ set -euo pipefail
 
 ENV_FILE=parsec-admin-token.env
 if [ ! -f $ENV_FILE ]; then
-    TOKEN=$(openssl rand 63 | base64 --wrap=86)
-    echo "PARSEC_ADMINISTRATION_TOKEN=$TOKEN" > $ENV_FILE
+    PARSEC_ADMINISTRATION_TOKEN=$(openssl rand -hex 32)
+    echo "PARSEC_ADMINISTRATION_TOKEN=$PARSEC_ADMINISTRATION_TOKEN" > $ENV_FILE
+
+    PARSEC_FAKE_ACCOUNT_PASSWORD_ALGORITHM_SEED=$(openssl rand -hex 32)
+    echo "PARSEC_FAKE_ACCOUNT_PASSWORD_ALGORITHM_SEED=$PARSEC_FAKE_ACCOUNT_PASSWORD_ALGORITHM_SEED"
```
FirelightFlagboy: **todo:** Must be added to env file

```suggestion
    echo "PARSEC_FAKE_ACCOUNT_PASSWORD_ALGORITHM_SEED=$PARSEC_FAKE_ACCOUNT_PASSWORD_ALGORITHM_SEED" >> $ENV_FILE
```

libparsec/crates/crypto/src/common/password.rs
FirelightFlagboy: The raw value is still here

server/parsec/_parsec_pyi/crypto.pyi
FirelightFlagboy: I did not see the comment

docs/HOSTING.md
touilleMan: :+1: done.

docs/hosting/deployment/gen-admin-token.sh
touilleMan: :sweat_smile: that's some dumb error ! thx !

server/parsec/_parsec_pyi/crypto.pyi
touilleMan: I added the comment in the Rust implementation (this is only a comment and not a docstring since we describe implementation detail and not how to use the function)

https://github.com/Scille/parsec-cloud/pull/10471/files#diff-739f4c41abd8f2f1d7ea80d499d3036f82ea11285a6ba230ad1437243688315eR69-R72

libparsec/crates/crypto/src/common/password.rs
touilleMan: :+1: done.

docs/HOSTING.md
touilleMan: This is a good question, but I'd rather keep this out of this PR ;-)

libparsec/crates/account/src/lib.rs
touilleMan: This is a very good point, I will address it in a subsequent PR :+1:

PR #10517, opened 2025-06-11
[MS] Added support for custom branding

### LINKED ISSUE #10500: Dynamic management of resources
The app should be able to load resources dynamically (used for custom branding).

Using the logo as example:

- the app should first try to load the logo from a server (defined in config)
- if it fails, it should try to load the image for a folder (defined in config, with default)
- if it fails, it should display the default logo

Max-7:
Closes #10500 

This adds the support for custom branding.
Files can be put inside `public/custom` for web, or `electron/custom` for electron, and they will be used if available. For now, it only uses a few files defined in `resourcesManager.ts`. Texts can be fully replaced with `en_US.json` and `fr_FR.json`.

- [X] Describe your changes

### DIFF
--- a/client/src/main.ts
+++ b/client/src/main.ts
@@ -47,12 +47,13 @@ import { Events } from '@/services/eventDistributor';
 import { HotkeyManager, HotkeyManagerKey } from '@/services/hotkeyManager';
 import { Information, InformationDataType, InformationLevel, InformationManager, PresentationMode } from '@/services/informationManager';
 import { InjectionProvider, InjectionProviderKey } from '@/services/injectionProvider';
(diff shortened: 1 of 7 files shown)


client/src/main.ts
```diff
@@ -674,6 +692,7 @@ declare global {
       seeInExplorer: (path: string) => void;
       getLogs: () => Promise<Array<string>>;
       initError: (error?: string) => void;
+      readCustomFile: (ffile: string) => Promise<ArrayBuffer | undefined>;
```
Ironicbay: typo on `file` argument

NicoTuxx (approved):

PR #10519, opened 2025-06-11
Use constant for cmd families in RPC test fixture

### LINKED ISSUE #10483: Python test RPC fixture should use Constant values
Unrelated to this PR: we should use constants (or enum) for command families here 
```python
AUTHENTICATED_ACCOUNT="authenticated_account"
...
raw_rep = await self._do_request(req.dump(), AUTHENTICATED_ACCOUNT)
```

_Originally posted by @mmmarcos in https://github.com/Scille/parsec-cloud/pull/10480#discussion_r2131853956_

mmmarcos:
Closes #10483

### DIFF
--- a/server/tests/common/rpc.py
+++ b/server/tests/common/rpc.py
@@ -30,6 +30,15 @@ from parsec._parsec import (
     tos_cmds,
 )
 
+# Command families
+
+TOS_FAMILY = "tos"
+INVITED_FAMILY = "invited"
+ANONYMOUS_FAMILY = "anonymous"
+AUTHENTICATED_FAMILY = "authenticated"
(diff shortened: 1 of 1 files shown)


mmmarcos:
> https://github.com/Scille/parsec-cloud/blob/6346a36834ec3161776d033fe08836fe21f40829/server/tests/common/rpc.py#L3
> 
> Will not work on the next generation

Ouch, I completely missed that. If this code is auto-generated, the need for constants is much less important! If you agree we can close the PR and the issue.

FirelightFlagboy:
Do as you see fit, seems it's auto generated I'm not bothered about the lake of constant here

PR #10539, opened 2025-06-13
docs(RFC): Add RFC about simplifying user-facing token

FirelightFlagboy:
Render: [Simple User Facing Token](https://github.com/Scille/parsec-cloud/blob/rfc-simple-user-facing-token/docs/rfcs/0000-simple-user-facing-token.md)

### DIFF
--- a/docs/rfcs/0000-simple-user-facing-token.md
+++ b/docs/rfcs/0000-simple-user-facing-token.md
@@ -1,0 +1,66 @@
+<!-- Parsec Cloud (https://parsec.cloud) Copyright (c) BUSL-1.1 2016-present Scille SAS -->
+<!-- cspell: words 4OIJ-VLMK-EJOM-H77B-WT4N-XFVS-IM -->
+
+# Simple User Facing Token
+
+## Overview
+
+How to ease the user input for entering a token
+
+## Background & Motivation
+
+When designing the API for Parsec Account more specifically for the account creation and deletion confirmation steps, we require the user to provide a token.
+Currently, those tokens consist of 16 bytes randomly generated
+
+If it where provided to the user in [base32] encoding, that represent 32 character:
+
+```shell
+$ openssl rand 16 | base32 | tr -d = | sed 's/\(.\{4,4\}\)/\1 /g' | xargs echo | tr ' ' '-'
+4OIJ-VLMK-EJOM-H77B-WT4N-XFVS-IM
+```
+
+> We choose [base32] because it contains only upper letter and digit with $1$ and $0$ removed
+> to prevent confusion with `I` and `O`.
+> It's easy on the client side to convert lower to upper letter and $1$ to `I` and $0$ to `O`.
+
+[base32]: https://en.wikipedia.org/wiki/Base32
+
+## Goals and Non-Goals
+
+The goal is to simply the user input of the tokens to confirm the account creation/deletion.
+
+## Design
+
+Instead of using plain token for those operations, we instead use a text code:
+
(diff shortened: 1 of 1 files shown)


docs/rfcs/0000-simple-user-facing-token.md
```diff
@@ -0,0 +1,65 @@
...
+
+- Initially, we discussed generating a token by deriving that code (the server sent that code, the client derivate it to get a token).
+
+  That approach does not provide any benefit as the generated token would have been constraint by code (we consider that code as less entropy than the token).
+  But does allow reusing the current protocol definition (i.e. the client provide a token)
+
+- If we do not need the length of the code to be configurable, we could have something similar to the SAS code (but longer than 4 char)
+
+  This change sightly the approach as it would be `libparsec` that generate that code, that simplify the implementation as we would not need to add another command for the code length for UI configuration.
+  We could update to the configurable length approach later on.
```
FirelightFlagboy, line 54: **thought:** That approach seems better than the configurable approach as it require less upfront work

docs/rfcs/0000-simple-user-facing-token.md
```diff
@@ -0,0 +1,65 @@
...
+The goal is to simply the user input of the tokens to confirm the account creation/deletion.
+
+## Design
+
+Instead of using plain token for those operations, we instead use a text code:
+
+The server would generate those codes using the [base32] alphabet, their length would be configurable via a settings (cli option, env var).
+Those codes are then sent via email (the current communication channel).
+
+That would require minor modification on the protocol for account creation/deletion as it would take a `AccountCreationCode`/`AccountDeletionCode` of arbitrary length (remember the length is server configurable).
```
AureliaDolo, line 39: Should we set a not configurable minimum length for the code ? For example not accept a length under 8 chars.

docs/rfcs/0000-simple-user-facing-token.md
```diff
@@ -0,0 +1,65 @@
...
+## Goals and Non-Goals
+
+The goal is to simply the user input of the tokens to confirm the account creation/deletion.
+
+## Design
+
+Instead of using plain token for those operations, we instead use a text code:
+
+The server would generate those codes using the [base32] alphabet, their length would be configurable via a settings (cli option, env var).
+Those codes are then sent via email (the current communication channel).
```
AureliaDolo, line 37: For when the mobile usage is more integrated, but we could show the code as a qr code to be scanned.

FirelightFlagboy: That's given the user does not use their phone to read emails, that seems unlikely today

FirelightFlagboy:
> It would be nice to add a default value (maybe even included the estimated brute force time)

What do you mean by "default value"?

For the estimated brute force time: https://github.com/Scille/parsec-cloud/blob/1a8fb273bd258bb98309959a27f3e66d1937f20c/docs/rfcs/0000-simple-user-facing-token.md?plain=1#L61

docs/rfcs/0000-simple-user-facing-token.md
FirelightFlagboy: I've added that the client should check that the code is not too small in the RFC

touilleMan:
I don't think we need a new RFC for this since the impacted RFC (i.e. RFC1013) is not yet in production

Instead we should directly modify RFC1013 :+1:

FirelightFlagboy:
Deprecated in favor of #10569

PR #10546, opened 2025-06-16
fix(account-creation): Check token validity

### LINKED ISSUE #10532: Server does not check if the creation token is still valid
In parsec-account, during the account creation, the server does not check if the token is still valid

FirelightFlagboy:
Also refactor the account creation test

Closes #10532


touilleMan:
We've discussed this in last meeting, this doesn't need to be implemented yet since the token will be only valid for a couple of trials instead

PR #10547, opened 2025-06-16
refactor(releaser): Rework push force setting in nightly mode

FirelightFlagboy:
- Rename `--force-push` to `git-force` as its value is used on different `git` commands.
- Add `GitForceMode` to represent to different force mode.

> You can test by setting the env var `DRY_GIT=1` and observing the printed `git` commands

### DIFF
--- a/misc/releaser.py
+++ b/misc/releaser.py
@@ -30,6 +30,7 @@ A typical release process looks as follow:
 from __future__ import annotations
 
 import argparse
+import enum
 import math
 import os
 import re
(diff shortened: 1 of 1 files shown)


misc/releaser.py
```diff
@@ -992,6 +999,30 @@ def history_main(args: argparse.Namespace) -> None:
     print(gen_rst_release_entry(version, release_date, newsfragment_rst))
 
 
+class GitForceMode(enum.Enum):
+    NoForce = "none"
+    WithLease = "with-lease"
+    Force = "force"
```
mmmarcos: I'd sort
```suggestion
    Force = "force"
    NoForce = "none"
    WithLease = "with-lease"
```

Also, values are not really meaningful here, we can use auto:
```suggestion
    Force = auto()
    NoForce = auto()
    WithLease = auto()
```

Unless we give meaning to them:
```suggestion
    Force = "--force"
    ForceWithLease = "--force-with-lease"
    NoForce = ""
```
So you can replace the match in `push_args` with `return [self.value]`.

FirelightFlagboy: The values are important, they're used for `argparse` option value

mmmarcos (approved):

PR #10567, opened 2025-06-18
Introduce `TrustedPasswordAlgorithm`&`UntrustedPasswordAlgorithm`

touilleMan:


libparsec/crates/crypto/src/common/password.rs
```diff
@@ -55,6 +52,137 @@ impl PasswordAlgorithm {
...
+                opslimit,
+                parallelism,
+                salt,
+            },
+        }
+    }
+}
+
+/// Password algorithm configuration provided by a 3rd party, with no for us to
+/// validate if the derived key is what we expected.
```
FirelightFlagboy: **typo:** 
```suggestion
/// Password algorithm configuration provided by a 3rd party, with no way for us to
/// validate if the derived key is what we expected.
```

libparsec/crates/crypto/src/common/password.rs
```diff
@@ -55,6 +52,137 @@ impl PasswordAlgorithm {
...
+///
+/// Typical use-case fort this is the account auth method which can be of two types:
```
FirelightFlagboy: **typo:** 🏰
```suggestion
/// Typical use-case for this is the account auth method which can be of two types:
```

libparsec/crates/crypto/src/common/password.rs
```diff
@@ -55,6 +52,137 @@ impl PasswordAlgorithm {
...
+/// - ClientProvided, for which the client is able to store
+///   `auth_method_master_secret` all by itself.
+/// - Password, for which the client must obtain some configuration
+///   (i.e. this type !) from the server in order to know how
```
FirelightFlagboy: **typo:** 
```suggestion
///   (i.e. this type!) from the server in order to know how
```

libparsec/crates/crypto/src/common/password.rs
```diff
@@ -70,36 +198,18 @@ impl PasswordAlgorithm {
         // - `libparsec_types` itself depends on `libparsec_crypto` (i.e. this crate !)
         // - `email` here is only used as-is as a discriminant. Providing an invalid
         //   email has not impact whatsoever on the quality of the output.
-        email: &str,
-        seed: &SecretKey,
+        _email: &str,
+        _seed: &SecretKey,
```
FirelightFlagboy, line 204: **nitpick:** If we no longer need those params, remove them?

Or did you intend to further customize the parameters?

FirelightFlagboy (commented):
Some typo before ready to go

docs/rfcs/1013-parsec-account.md
```diff
@@ -264,6 +264,10 @@ Attributes:
 >   still be used to recover the vault in case of account recovery (see below).
 > - Currently, only password authentication is supported, other authentication
 >   methods (such as FIDO2) may be added in the future.
+> - Given the server control the password algorithm, the client must ensure the
+>   provided configuration is not too weak (which would allow the server to brute
+>   force the password by using `auth_method_id` as oracle).
```
AureliaDolo, line 269: What should be the expected client behavior if the configuration is too weak ? Just saying that the configuration is too weak is not really helpful, and there is nothing a user could really do to solve that. Maybe something like "contact an administrator " ?

FirelightFlagboy: The client would refuse to connect to the server that provide a weak configuration as it may be a MiM attack.

I think a message of some sort:

> During connection attempt to parsec-account, the server sent a configuration that result in weaker security than the default one.
> As so we refuse to connect to it.
> You must contact your system administrator.

libparsec/crates/crypto/src/common/password.rs
touilleMan: :+1: done.

libparsec/crates/crypto/src/common/password.rs
touilleMan: :+1: done.

libparsec/crates/crypto/src/common/password.rs
touilleMan: :+1: done.

libparsec/crates/crypto/src/common/password.rs
touilleMan: Yes we are going to need them whenever we add a new algorithm, or if we start supporting custom parameters for argon2i.

docs/rfcs/1013-parsec-account.md
touilleMan: :+1:

libparsec/crates/crypto/src/common/password.rs
```diff
@@ -55,6 +52,137 @@ impl PasswordAlgorithm {
...
+            } => {
+                if memlimit_kb < ARGON2ID_DEFAULT_MEMLIMIT_KB
+                    || opslimit < ARGON2ID_DEFAULT_OPSLIMIT
+                    || parallelism < ARGON2ID_DEFAULT_PARALLELISM
+                {
+                    return Err(CryptoError::Algorithm("Config too weak".to_string()));
+                }
+
+                let salt: [u8; ARGON2ID_SALTBYTES] = blake2b_hash(
+                    [b"PARSEC_PASSWORD_ARGON2ID.".as_ref(), email.as_bytes()].into_iter(),
```
FirelightFlagboy: **nitpick:** Use a constant value for `b"PAR[...]2ID"`

touilleMan: :+1: done.

PR #10569, opened 2025-06-18
Update RFC1013 to use a validation code instead of token and ParsecAccountActionAddr

touilleMan:


docs/rfcs/1013-parsec-account.md
```diff
@@ -320,31 +320,19 @@ The Parse Account relies on email address validation for multiple operations:
...
-`AccountRecovery` | `account_recovery` | `base64(msgpack(<validation_token>))`
-`AccountDeletion` | `account_delete`   | `base64(msgpack(<validation_token>))`
-
-> [!NOTE]
-> We should probably rename `ParsecActionAddr` -> `ParsecOrganizationActionAddr`
-> to avoid confusion with this new `ParsecAccountActionAddr` (this is a purely internal
-> code change, so no risk of backward compatibility issue).
+Since this validation code must be copy&pasted by a end user, it is of much lower entropy
+than the tokens are used in `ParsecInvitationAddr`/`ParsecOrganizationBootstrapAddr`. For
+this reason, a validation code can only be attempted three times before becoming invalid.
```
FirelightFlagboy: **polish:** I would add a note about how we are able to detect that a code is invalid.

```suggestion
than the tokens are used in `ParsecInvitationAddr`/`ParsecOrganizationBootstrapAddr`. For
this reason, a validation code can only be attempted three times before becoming invalid.

> [!NOTE]
> We know about the failed attempts because the user also provide the email related to the account to create
```

docs/rfcs/1013-parsec-account.md
```diff
@@ -447,9 +434,10 @@ Anonymous account API:
     "req": {
       "fields": [
         {
-          // Token received by email following use of `account_create_send_validation_email`
-          "name": "validation_token",
-          "type": "EmailValidationToken"
+          // Code received by email following use of `account_create_send_validation_email`
+          // Should be 6 base32 characters.
+          "name": "validation_code",
+          "type": "String"
```
FirelightFlagboy: **nitpick:** I would use a dedicated type for the code

```suggestion
          "type": "AccountCreationCode"
```

docs/rfcs/1013-parsec-account.md
```diff
@@ -492,7 +480,12 @@ Anonymous account API:
...
       },
       {
-        "status": "invalid_validation_token"
+        "status": "invalid_validation_code"
+      },
+      {
+        // After 3 bad attempts, the server no longer check the validation
+        // code and directly return this error.
+        "status": "too_many_attempts_on_validation_code"
       },
```
FirelightFlagboy: **nitpick (non-blocking):** I would have just returned `invalid_code`

```suggestion
```

docs/rfcs/1013-parsec-account.md
```diff
@@ -652,8 +646,10 @@ Anonymous account API:
   "req": {
     "fields": [
       {
-        "name": "validation_token",
-        "type": "Token"
+        // Code received by email following use of `account_create_send_validation_email`
+        // Should be 6 base32 characters.
+        "name": "validation_code",
+        "type": "String"
```
FirelightFlagboy: **nitpick:** I would use a dedicated type

```suggestion
        "type": "AccountRecoveryCode"
```

docs/rfcs/1013-parsec-account.md
```diff
@@ -726,16 +727,23 @@ Anonymous account API:
   "cmd": "account_delete_proceed",
   "req": [
     {
-      "name": "validation_token",
-      "type": "Token"
+      // Code received by email following use of `account_create_send_validation_email`
+      // Should be 6 base32 characters.
+      "name": "validation_code",
+      "type": "String"
```
FirelightFlagboy: **nitpick:** I would use a dedicated type

```suggestion
      "type": "AccountDeletionCode"
```

docs/rfcs/1013-parsec-account.md
touilleMan: Good point ! Since the code cannot be used as a ID, we need to add the email in the commands

docs/rfcs/1013-parsec-account.md
touilleMan: The three validation code are the same, so I'm against having one different type for every one of them

But we can introduce a single `ValidationCode` type for all of them, however I'm not sure it is worth the trouble since this field is only used for simple comparison.

docs/rfcs/1013-parsec-account.md
```diff
@@ -320,31 +320,19 @@ The Parse Account relies on email address validation for multiple operations:
...
 
-1. The Parsec client sends an initial API request to the server.
-2. The server sends a confirmation email to the user containing a link with a validation token.
-3. The user clicks on the link in the email, which open the Parsec client.
-4. The Parsec client sends a second API request with the validation token to the server.
-5. The server proceed with the actual operation.
+1. The user select the operation from his Parsec client.
+2. The Parsec client sends an initial API request to the server.
+3. The server sends a confirmation email to the user containing a validation code.
+4. The user checks his email to get the validation code, and paste it in his Parsec client.
```
AureliaDolo:
```suggestion
4. The user checks his email to get the validation code, and pastes it in his Parsec client.
```

docs/rfcs/1013-parsec-account.md
AureliaDolo: For clarification: If a user has legitimately noodle fingers and fails at the validation code step, what's the expected process ?
- they restart the process from scratch after some time after the last attempt ?
- they contact an administrator to be manually unlocked ?
- should their mail be kept in unverified_emails after their 3rd attempt, or moved to a suspicious emails list ?

docs/rfcs/1013-parsec-account.md
AureliaDolo: I'm not sure, if the code is invalid the user may want to try another code, whereas after too many attempt that won't work anymore.

docs/rfcs/1013-parsec-account.md
```diff
@@ -505,12 +498,13 @@ Anonymous account API:
 ```
 
 > [!NOTE]
-> Since the request is authenticated with the token field it contains, it could
+> Since the request is authenticated with the validation code field it contains, it could
```
AureliaDolo:
```suggestion
> Since the request is authenticated with the validation code and the email field it contains, it could
```

I'm not sure this part is still relevant though

docs/rfcs/1013-parsec-account.md
touilleMan: Serendipity caught me, now we have a PR properly implementing `ValidationCode` type  https://github.com/Scille/parsec-cloud/pull/10576 ^^

docs/rfcs/1013-parsec-account.md
touilleMan: done with `ValidationCode`

docs/rfcs/1013-parsec-account.md
touilleMan: done with `ValidationCode`

docs/rfcs/1013-parsec-account.md
touilleMan: :+1: done.

docs/rfcs/1013-parsec-account.md
touilleMan: The user only has to restar the process from scratch (i.e. resending a new validation email containing a new code)

I've added a comment to make this more explicit

docs/rfcs/1013-parsec-account.md
touilleMan: I've removed this part :+1:

docs/rfcs/1013-parsec-account.md
```diff
@@ -719,23 +732,34 @@ Authenticated account API:
...
 {
   "cmd": "account_delete_proceed",
   "req": [
     {
-      "name": "validation_token",
-      "type": "Token"
+      "name": "email",
+      "type": "EmailAddress"
+    },
+    {
```
FirelightFlagboy: **todo (blocking):** It's the authenticated API, we already have the email of the user

```suggestion
```

docs/rfcs/1013-parsec-account.md
FirelightFlagboy: I understand, I'm fine with the proposition it just that I had a small doubt

docs/rfcs/1013-parsec-account.md
```diff
@@ -417,14 +418,13 @@ Anonymous account API:
...
+On `ok`, the server would have sent a mail with a unique validation code to use in the next part of account creation.
 
 If an account already exists with this email, a `ok` response will still be sent (without sending an email)
 to avoid creating an oracle about emails registered in the service.
-If an account creation has already been requested for an email and the email is not validated yet, a new email with a new token may or may not been sent depending on how long ago the previous one was sent.
 
-The token is valid for a default duration of 24h, which can be changed in the server configuration (
-`PARSEC_ACCOUNT_EMAIL_VALIDATION_TOKEN_VALIDITY` specifying the duration in seconds).
+If an account creation has already been requested for an email and the email is not validated yet, a new email with a
+new validation code may or may not been sent depending on how long ago the previous one was sent.
```
AureliaDolo, line 427: It should not be an issue, but if both the time before sending a new token and the validity duration are configurable, it could make sense to check that the latter is bigger.

If the validity duration is too short the server may refuse to send a new email, whereas it  should have created a new code.

docs/rfcs/1013-parsec-account.md
```diff
@@ -745,14 +769,18 @@ At that point the server can remove the client, and its related data.
...
 > creation & recovery, we just want to remove data from the server).
 >
-> For this reason, we could in the validation email a link to a simple HTTP GET
-> route in the server that would do the removal (so no Parsec client involved).
+> For this reason, we could have in the validation email a link to a simple
+> HTTP GET route in the server that would do the removal (so no Parsec client
+> involved).
 >
 > Instead, we still use a regular validation link that redirect to the Parsec client
 > to be consistent with other operations.
```
AureliaDolo: Is that still true if we remove the actionAccountParsecAddr ?

docs/rfcs/1013-parsec-account.md
AureliaDolo: This issue may or may not arise depending on the order in which we check token validity vs ability to send an email

docs/rfcs/1013-parsec-account.md
touilleMan: Good point ! done. :+1:

docs/rfcs/1013-parsec-account.md
touilleMan: I think we don't want to make token validity configurable: since it is considered to be a synchronous operation (i.e. the user has to keep his Parsec client open on the page asking him for the code), a standard 1h value is basically good for any usecase.

docs/rfcs/1013-parsec-account.md
touilleMan: :+1: I've corrected this part

docs/rfcs/1013-parsec-account.md
AureliaDolo: :+1: And someone setting the time before resending the new email to more than 1h seems a lot .

PR #10576, opened 2025-06-19
Introduce `ValidationCode` type

touilleMan:
Following changes on RFC1013


libparsec/crates/types/src/account.rs
```diff
@@ -42,6 +42,60 @@ macro_rules! impl_decrypt_and_load {
...
+/// The code format is a 6 characters string using base32 (i.e. RFC 4648).
+/// Example: `AD3FXJ`
+#[serde_as]
+#[derive(Clone, Serialize, Deserialize, PartialEq, Eq)]
+#[serde(try_from = "&str", into = "String")]
+#[non_exhaustive] // Prevent initialization without going through the factory
+pub struct ValidationCode([u8; VALIDATION_CODE_SIZE]);
+
+impl TryFrom<&str> for ValidationCode {
+    type Error = &'static str;
```
FirelightFlagboy: **todo:** Use proper enum

FirelightFlagboy (changes requested):

libparsec/crates/types/src/account.rs
touilleMan: done.

libparsec/crates/types/src/account.rs
```diff
@@ -42,6 +43,70 @@ macro_rules! impl_decrypt_and_load {
...
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        f.debug_tuple("ValidationCode")
+            .field(&self.as_ref())
+            .finish()
+    }
+}
+
+impl AsRef<str> for ValidationCode {
+    fn as_ref(&self) -> &str {
+        std::str::from_utf8(&self.0).expect("valid")
```
AureliaDolo:
```suggestion
        std::str::from_utf8(&self.0).expect("validationCode uses a subset of UTF8")
```
or even if you want to be really explicit
```
match std::str::from_utf8(&self.0) {
    Ok(s) => s,
    Err(_) => unreachable!("validationCode uses a subset of UTF8")
}
```

AureliaDolo (approved):

libparsec/crates/types/src/account.rs
```diff
@@ -42,6 +43,70 @@ macro_rules! impl_decrypt_and_load {
...
+
+#[derive(Error, Debug)]
+pub enum ValidationCodeError {
+    #[error("Bad size")]
+    BadSize,
+    #[error("Not base32")]
+    NotBase32,
+}
+
+const BASE32_ALPHABET: &[u8; 32] = b"ABCDEFGHIJKLMNOPQRSTUVWXYZ234567";
```
FirelightFlagboy: **nitpick (non-blocking):** We define a similar alphabet for SAS code, can this alphabet be also used for SAS code?

FirelightFlagboy (approved):

libparsec/crates/types/src/account.rs
FirelightFlagboy: Clippy would say to use `.unwrap_or_else(|_| unrea...`

touilleMan: done.

libparsec/crates/types/src/account.rs
touilleMan: Good idea, I've factorized the code.

libparsec/crates/types/src/account.rs
```diff
@@ -42,6 +44,81 @@ macro_rules! impl_decrypt_and_load {
...
+
+#[derive(Error, Debug)]
+pub enum ValidationCodeParseError {
+    #[error("Bad size")]
+    BadSize,
+    #[error("Not base32")]
+    NotBase32,
+}
+
+const VALIDATION_CODE_CHARS: &[u8; 32] = crate::BASE32_ALPHABET;
```
FirelightFlagboy, line 59: **nitpick:** I would use an alias over redefining the constant

```suggestion
use crate::BASE32_ALPHABET as VALIDATION_CODE_CHARS;
```

libparsec/crates/types/src/lib.rs
```diff
@@ -42,6 +42,10 @@ pub const REALM_RENAME_KEY_DERIVATION_UUID: uuid::Uuid =
 pub const PATH_URL_KEY_DERIVATION_UUID: uuid::Uuid =
     uuid::uuid!("00000000-0000-0000-0000-000000000003");
 
+/// Base32 alphabet used for human facing codes (see `SASCode` & `ValidationCode`)
+/// (Note I/1 and 0/O are skipped to avoid visual confusion)
+const BASE32_ALPHABET: &[u8; 32] = b"ABCDEFGHJKLMNPQRSTUVWXYZ23456789";
```
FirelightFlagboy, line 47: **nitpick:** I think we should just stick to base32 alphabet to not have a confusing constant name.

```suggestion
const BASE32_ALPHABET: &[u8; 32] = b"ABCDEFGHIJKLMNOPQRSTUVWXYZ234567";
```

I think it's OK to have `I` and `O` since we can easily fix user typo if it use `1` or `0`

libparsec/crates/types/src/account.rs
```diff
@@ -42,6 +44,81 @@ macro_rules! impl_decrypt_and_load {
...
+pub struct ValidationCode([u8; VALIDATION_CODE_SIZE]);
+
+impl Default for ValidationCode {
+    fn default() -> Self {
+        let mut bytes = [0u8; VALIDATION_CODE_SIZE];
+        for c in bytes.iter_mut() {
+            let r = rand::thread_rng().gen_range(0..32);
+            *c = VALIDATION_CODE_CHARS[r];
+        }
+        Self(bytes)
```
FirelightFlagboy, line 80: **suggestion:** Using the standard base32 alphabet would allow to do that:

```suggestion
        use data_encoding::BASE32;
        let num: u32 = rand::thread_rng().gen_range(0..0xff_ff_ff_ff);
        let code = BASE32.encode(&num.to_be_bytes());
        Self(
            code.as_bytes()[..VALIDATION_CODE_SIZE]
                .try_into()
                .unwrap_or_else(|_| unreachable!("The slice is of correct size")),
        )
```

touilleMan: Since this alphabet is already used for the SAS code, we shouldn't modify it to avoid introducing weird behavior when doing enrollment between between different versions of Parsec clients :/

PR #10619, opened 2025-06-23
refactor(server): Rework how we store information about deletion code

FirelightFlagboy:


server/parsec/components/account.py
```diff
@@ -97,6 +97,33 @@ class VaultItems:
...
+VALIDATION_CODE_VALIDITY_DURATION_SECONDS: int = 3600
+"""
+How much seconds a validation code remain usable (valid)
+"""
+
+VALIDATION_CODE_MAX_FAILED_ATTEMPS: int = 3
+"""How much an user can "guess" the code"""
+
+
+@dataclass
```
touilleMan:
```suggestion
@dataclass(slots=True)
```

server/parsec/components/account.py
```diff
@@ -97,6 +97,33 @@ class VaultItems:
...
+class ValidationCodeInfo:
+    code: ValidationCode
+    created_at: DateTime
+    failed_attemps: int = 0
+
+    def is_expired(self, now: DateTime) -> bool:
+        """Check that the validation code is not expired"""
+        return self.created_at.add(seconds=VALIDATION_CODE_VALIDITY_DURATION_SECONDS) >= now
```
touilleMan: The result is the same, but I would find `(now - self.created_at) > VALIDATION_CODE_VALIDITY_DURATION_SECONDS` more readable

touilleMan (approved):

PR #10621, opened 2025-06-23
Rework `liparsec_account::Account` login methods

touilleMan:


libparsec/crates/account/src/lib.rs
```diff
@@ -35,51 +35,134 @@ pub enum AccountFromPasswordError {
...
+
+    async fn _login_with_master_secret(
+        config_dir: &Path,
+        proxy: ProxyConfig,
+        addr: ParsecAddr,
+        auth_method_master_secret: KeyDerivation,
+        // We must recycle anonymous_cmds if it exists, otherwise its drop will complain
+        // during tests if some request registered with `test_register_sequence_of_send_hooks`
+        // has yet to occur.
+        anonymous_cmds: Option<AnonymousAccountCmds>,
```
FirelightFlagboy, line 102: **note:** That also allow to recycle the http client

So we potentially skip the proxy config and tls handcheck

libparsec/crates/account/src/lib.rs
```diff
@@ -35,51 +35,134 @@ pub enum AccountFromPasswordError {
...
             mac_key: auth_method_mac_key,
         };
-        let cmds = AuthenticatedAccountCmds::new(config_dir, addr, proxy, auth_method)
-            .context("Cannot configure server connection")?;
 
-        Ok(Self { cmds })
+        let cmds = match anonymous_cmds {
+            None => AuthenticatedAccountCmds::new(config_dir, addr, proxy, auth_method)
+                .context("Cannot configure server connection")?,
+            Some(anonymous_cmds) => anonymous_cmds.into_authenticated(auth_method),
```
FirelightFlagboy: **nitpick:** I would prefer that we expose a `AuthenticatedAccountCmds::from_anonymous`

because it's not anonymous job to know how to construct an authenticated client

libparsec/crates/account/src/lib.rs
touilleMan: I've added a comment in `AuthenticatedAccountCmds::from_anonymous` about this :+1:

libparsec/crates/account/src/lib.rs
touilleMan: done :+1:

libparsec/crates/account/src/lib.rs
```diff
@@ -35,51 +35,136 @@ pub enum AccountFromPasswordError {
...
+}
+
+pub struct AccountListRegistrationDeviceItem {
+    pub organization_id: OrganizationID,
+    pub user_id: UserID,
+}
+
+pub struct AccountListRegistrationDeviceItems {
+    pub items: Vec<AccountListRegistrationDeviceItem>,
+    pub bad_entries: Vec<()>,
```
FirelightFlagboy, line 66: **question:** :raised_eyebrow:

libparsec/crates/account/src/lib.rs
```diff
@@ -91,13 +176,76 @@ impl Account {
...
+            match item {
+                AccountVaultItem::RegistrationDevice(registration_device) => {
+                    res.items.push(AccountListRegistrationDeviceItem {
+                        organization_id: registration_device.organization_id,
+                        user_id: registration_device.user_id,
+                    });
+                }
+                AccountVaultItem::WebLocalDeviceKey(_) => (),
+            }
+        }
```
FirelightFlagboy, line 246: **nitpick:** If you are only settings `items`, use a `collect` with `filter_map`

libparsec/crates/account/src/lib.rs
touilleMan: Yes, this is some WIP code ^^

However I'd rather merge it now since it is in a "good enough" state and the login changes are rather important

(I pondered not including `list_registration_device` at all from the PR, but it is not ideal since it means multiple fields from `Account` are never used and must be marked as such...)

libparsec/crates/account/src/lib.rs
touilleMan: This code is WIP anyway (hence the `bad_entries` field that I was thinking to use instead of the `log::warn` in the loop...)

The key point here is I wonder if I should decrypt the data right in the `list_registration_device` or should I wait for the method that actually ask to use the registration device ( and also so I keep some cache between the two calls ?)

So too many questions, that should be easier to answer once I'll start implementing the method using the registration device (i.e. in the next PR, stay tuned ^^)

FirelightFlagboy (approved):

PR #10623, opened 2025-06-23
Update `list_workspace_users` to include revoked & unshared

### LINKED ISSUE #10615: Modify `list_workspace_user` to retrieve all users of a workspace
Modify `libparsec_client::Client::list_workspace_users` (so `libparsec_client::CertificateOps::list_workspace_users` in practice ^^) to always return the users that used to be part of the workspace.

> It's better to always return users who no longer belong, since otherwise we'll have to add a flag (include vs don't include), but this won't change the return type, so you'll have to check the fields marked None anyway...

So the returned type would be a `vec` of:
```
#[derive(Debug, PartialEq, Eq)]
pub struct WorkspaceUserAccessInfo {
    pub user_id: UserID,
    pub human_handle: HumanHandle,
    pub current_profile: Option<UserProfile>, // None if no longer part of the workspace
    pub current_role: RealmRole,  // None if user hase been revoked
```

Also, we'll probably need to to change `CertificatesStoreReadGuard::get_realm_current_users_roles` (since it doesn't return users who have been revoked). And be careful, because `get_realm_current_users_role` is also used in `libparsec/crates/client/src/certif/realm_key_rotation.rs::generate_realm_rotate_key_req`

mmmarcos:
`WorkspaceUserAccessInfo` now uses an `Option` for `current_role` (`None` if unshared) & `current_profile` (`None` if revoked).

This required to update `get_realm_current_users_roles` to include unshared certificates: the flag `include_unshared` was added so other calls to `get_realm_current_users_roles` are not impacted.

Closes #10615

### DIFF
--- a/libparsec/crates/client/src/certif/list.rs
+++ b/libparsec/crates/client/src/certif/list.rs
@@ -314,12 +314,18 @@ pub(super) async fn get_user_device(
 pub struct WorkspaceUserAccessInfo {
     pub user_id: UserID,
     pub human_handle: HumanHandle,
-    pub current_profile: UserProfile,
-    pub current_role: RealmRole,
+    /// The user's current profile or `None` if the user has been revoked
+    pub current_profile: Option<UserProfile>,
+    /// The user's current role or `None` if the user no longer has access to the workspace
+    pub current_role: Option<RealmRole>,
 }
 
-/// List users currently part of the given workspace (i.e. user not revoked
-/// and with a valid role)
+/// List all users who have (or had at some point) access to the workspace.
+///
+/// This includes:
+/// - Users who have access to the workspace, their current role & profile are set accordingly.
+/// - Users who no longer have access to the workspace (i.e. unshared), their current role is set to `None`.
+/// - Users who are no longer part of the organization (i.e. revoked), their current profile is set to `None`.
 pub(super) async fn list_workspace_users(
     ops: &CertificateOps,
     realm_id: VlobID,
@@ -329,26 +335,13 @@ pub(super) async fn list_workspace_users(
     ops.store
         .for_read(async |store| {
             let per_user_certifs = store
-                .get_realm_current_users_roles(UpTo::Current, realm_id)
+                .get_realm_current_users_roles(UpTo::Current, realm_id, true)
                 .await?;
 
             for (user_id, role_certif) in per_user_certifs {
-                let current_role = role_certif
-                    .role
-                    .expect("unshared user should not be listed");
-
-                // Ignore revoked users
-                let maybe_revoked = store
-                    .get_revoked_user_certificate(UpTo::Current, user_id)
-                    .await?;
-                if maybe_revoked.is_some() {
-                    continue;
-                }
-
                 let user_certif = match store.get_user_certificate(UpTo::Current, user_id).await {
                     Ok(user_certif) => user_certif,
-                    // We got the user ID from the certificate store, it is guaranteed to
-                    // be present !
+                    // We got the user ID from the certificate store, it is guaranteed to be present!
                     Err(
                         GetCertificateError::NonExisting
                         | GetCertificateError::ExistButTooRecent { .. },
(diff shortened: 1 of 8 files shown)


libparsec/crates/client/src/certif/store.rs
```diff
@@ -974,7 +977,13 @@ macro_rules! impl_read_methods {
...
                     }
                     None => {
-                        per_user_role.remove(&certif.user_id);
+                        if !include_unshared
+                        {
+                            per_user_role.remove(&certif.user_id);
+                        }
+                        else{
+                            per_user_role.insert(certif.user_id, certif);
+                        }
```
FirelightFlagboy, line 986: **todo:** format

libparsec/crates/client/tests/unit/client/share_workspace.rs
```diff
@@ -280,22 +284,42 @@ async fn can_unshare_with_revoked_recipient(
...
         .share_workspace(wksp1_id, "bob".parse().unwrap(), None)
         .await
         .unwrap();
 
-    let wksp1_access_info = client.list_workspace_users(wksp1_id).await.unwrap();
+    let wksp1_user_access_info = client.list_workspace_users(wksp1_id).await.unwrap();
+    let (alice_access_info, bob_access_info) = {
+        p_assert_eq!(wksp1_user_access_info.len(), 2);
+        let a = &wksp1_user_access_info[0];
+        let b: &WorkspaceUserAccessInfo = &wksp1_user_access_info[1];
```
FirelightFlagboy, line 299: **question:** Why the annotation needed for `b` and not for `a`?

mmmarcos: I think is residual: code was copied from another test where  `b` had the annotation because it was an `Option`. Now is no longer needed.

libparsec/crates/client/src/certif/store.rs
mmmarcos: I wonder why it didn't format automatically :thinking:

FirelightFlagboy: Because we are inside a macro

libparsec/crates/client/src/certif/list.rs
```diff
@@ -314,12 +314,18 @@ pub(super) async fn get_user_device(
...
 }
 
-/// List users currently part of the given workspace (i.e. user not revoked
-/// and with a valid role)
+/// List all users who have (or had at some point) access to the workspace.
+///
+/// This includes:
+/// - Users who have access to the workspace, their current role & profile are set accordingly.
+/// - Users who no longer have access to the workspace (i.e. unshared), their current role is set to `None`.
+/// - Users who are no longer part of the organization (i.e. revoked), their current profile is set to `None`.
```
touilleMan, line 328: Also note a revoked user might appear with a role for the workspace: this is because the removal of the role is automatically done, but only one a workspace owners's Parsec client has  detected the user revocation.

libparsec/crates/client/tests/unit/certif/list_workspace_users.rs
```diff
@@ -75,13 +75,13 @@ async fn multiple(env: &TestbedEnv) {
         [
             (
                 "bob".parse().unwrap(),
-                UserProfile::Standard,
-                RealmRole::Contributor
+                Some(UserProfile::Standard),
```
touilleMan, line 78: Would be good to also test the case a user is revoked, no longer part of the workspace, or both

This can be easily done by adding  to the `customize` call

For revoked:
```rust
builder.new_user("mallory");
[...]
builder.share_realm(realm_id, "mallory", RealmRole::Contributor);
builder.revoke_user("mallory");
[...]
```

For no longer part of the workspace:
```rust
builder.new_user("mallory");
[...]
builder.share_realm(realm_id, "mallory", RealmRole::Contributor);
builder.share_realm(realm_id, "mallory", None);
[...]
```

(I guess the easiest way would be to create 3 additional users in the `customize`, each representing a different case. This way we avoid having to parametrize the test.

libparsec/crates/client/tests/unit/client/list_workspace_users.rs
```diff
@@ -89,36 +84,43 @@ async fn ok(
                 MaybeRedacted::Redacted(_) => unreachable!(),
             }
         );
-        p_assert_eq!(*current_profile, UserProfile::Admin);
-        p_assert_eq!(*current_role, RealmRole::Owner);
+        p_assert_eq!((*current_profile).unwrap(), UserProfile::Admin);
```
touilleMan, line 87:
```suggestion
        p_assert_eq!(*current_profile, Some(UserProfile::Admin));
```

libparsec/crates/client/tests/unit/certif/list_workspace_users.rs
touilleMan: I've seen those tests were already present in the `client` upper layer, so we probably don't need to bother to also put them in `certif` ;-)

mmmarcos:
See https://github.com/Scille/parsec-cloud/issues/10615#issuecomment-2999517206

PR #10627, opened 2025-06-24
refactor(crypto): Use `OsRng` over `thread_rng`

### LINKED ISSUE #10501: Use `OsRng` over `thread_rng`

FirelightFlagboy:
To not have to think if `thread_rng` can be used as a CSRNG

Closes #10501

### DIFF
--- a/libparsec/crates/crypto/src/rustcrypto/sequester.rs
+++ b/libparsec/crates/crypto/src/rustcrypto/sequester.rs
@@ -1,5 +1,6 @@
 // Parsec Cloud (https://parsec.cloud) Copyright (c) BUSL-1.1 2016-present Scille SAS
 
+use rand::rngs::OsRng;
 use rsa::{
     oaep::Oaep,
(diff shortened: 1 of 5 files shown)


libparsec/crates/crypto/src/rustcrypto/sequester.rs
```diff
@@ -148,7 +149,7 @@ impl SequesterPublicKeyDer {
     // Encryption format:
     //   <algorithm name>:<encrypted secret key with RSA key><encrypted data with secret key>
     pub fn encrypt(&self, data: &[u8]) -> Vec<u8> {
-        let mut rng = rand::thread_rng();
+        let mut rng = OsRng;
```
touilleMan: maybe remove this `rng` variable since it is only used onced and we inline it in the others calls

FirelightFlagboy: done

PR #10639, opened 2025-06-25
[MS] Fix client area personal info change with empty phone number

### LINKED ISSUE #10344: [🐛 | Bug]: Phone number required when updating personal info when it's not required in monthly subscription
### Parsec version tested on:

3.4.0-rc2

### Platforms tested on:

MacOS

### Bug description:

In the client area, when trying to update personal info (first and last name) on monthly subscription, a phone number is required even though it's only required when subscribing to a custom order.

https://github.com/user-attachments/assets/c4965c48-65e3-4755-84fe-cbead2a909f3

### Relevant output:

_No response_

Ironicbay:
Closes #10344

### DIFF
--- a/client/src/services/bms/access.ts
+++ b/client/src/services/bms/access.ts
@@ -114,7 +114,7 @@ class BmsAccess {
       client: {
         firstname: data.firstname,
         lastname: data.lastname,
-        phone: data.phone,
+        phone: data.phone ?? null,
         country: data.country,
         company: data.company,
         job: data.job,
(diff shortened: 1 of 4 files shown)


client/src/services/bms/access.ts
```diff
@@ -100,7 +100,7 @@ class BmsAccess {
   async updatePersonalInformation(data: {
     firstname?: string;
     lastname?: string;
-    phone?: string;
+    phone?: string | null;
```
Max-7: I don't think this change is necessary since we're always sending all the parameters. I would do this instead:

```
    const response = await this.api.updatePersonalInformation(this.tokens.access, {
      userId: this.customerInformation.id,
      client: {
        ...
        phone: data.phone ?? null,
      },
```
to keep the API as is and not introduce another falsish value. Only the change in `UpdatePersonalInformationQueryData` is needed, but everything else can stay the same.

PR #10633, opened 2025-06-24
Add last updater to EntryStat & WorkspaceHistoryEntryStat 

mmmarcos:
- `EntryStat`
  - [ ] Add `last_updater` to `EntryStat::File` & `EntryStat::Folder`
  - [ ] Add tests to check `last_updater` is correctly updated
  - [ ] Update bindings using `EntryStat`
- `WorkspaceHistoryEntryStat`
  - [ ] Add `last_updater` to `WorkspaceHistoryEntryStat::File` & `WorkspaceHistoryEntryStat::Folder`
  - [ ] Add tests to check `last_updater` is correctly updated
  - [ ] Update bindings using `WorkspaceHistoryEntryStat`

Closes #10614

PR #10640, opened 2025-06-25
[MS] Display last updater

### LINKED ISSUE #10616: Display last updater in list view in GUI

Max-7:
Depends on #10633 

Closes #10616

### DIFF
--- a/client/src/locales/en-US.json
+++ b/client/src/locales/en-US.json
@@ -742,10 +742,13 @@
     "FileDetails": {
         "title": "Details on {name}",
         "stats": {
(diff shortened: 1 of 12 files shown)


client/src/locales/en-US.json
```diff
@@ -742,10 +742,13 @@
     "FileDetails": {
         "title": "Details on {name}",
         "stats": {
+            "generalInfo": "General information",
+            "lastUpdate": "Last update",
```
mmmarcos, line 746: What's the difference between `lastUpdate` et `updated`?

PR #10644, opened 2025-06-26
[MS] Tests warning appearing on Safari

Max-7:
My initial thought was to test
- Warning on Safari
- Warning on small display
- Warning if indexedDB is not available

Sadly, I cannot add a test for the small display because we disable it when using the testbed to not be too annoying when developping, and I don't want to have specific code in the app just for one specific test.
For indexedDB, I haven't found a way to disable it in Playwright: no permissions, and it's used in the web worker so I can't mock `window.indexedDB` or something like that.

We end up we a simple small test to check that the warning is properly displayed on Safari.

### DIFF
--- a/client/tests/e2e/specs/home_page.spec.ts
+++ b/client/tests/e2e/specs/home_page.spec.ts
@@ -216,6 +216,6 @@ msTest('Warn on Safari', async ({ context }) => {
   await setupNewPage(page, { mockBrowser: 'Safari' });
   const modal = page.locator('.incompatible-environment');
   await expect(modal).toBeVisible();
-  await expect(modal.locator('.incompatible-content__title')).toHaveText('Your browser is not compatibly with Parsec yet.');
+  await expect(modal.locator('.incompatible-content__title')).toHaveText('Your browser is not compatible with Parsec yet.');
   await context.close();
 });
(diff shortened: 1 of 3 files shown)


client/tests/e2e/specs/home_page.spec.ts
```diff
@@ -208,3 +208,14 @@ msTest('Open feedback', async ({ home }) => {
...
 });
+
+msTest('Warn on Safari', async ({ context }) => {
+  const page = (await context.newPage()) as MsPage;
+  // Tried changing the user agent but the detection is too good,
+  // so we need a little extra step.
+  await setupNewPage(page, { mockBrowser: 'Safari' });
+  const modal = page.locator('.incompatible-environment');
+  await expect(modal).toBeVisible();
+  await expect(modal.locator('.incompatible-content__title')).toHaveText('Your browser is not compatibly with Parsec yet.');
```
Ironicbay:
```suggestion
  await expect(modal.locator('.incompatible-content__title')).toHaveText('Your browser is not compatible with Parsec yet.');
```
```suggestion
  await expect(modal.locator('.incompatible-content__title')).toHaveText('Your browser is not compatibly with Parsec yet.');
```

PR #10645, opened 2025-06-26
Account list invitations protocol schema & server implementation

touilleMan:


server/parsec/components/invite.py
```diff
@@ -47,29 +47,16 @@
...
-                )
-            case InvitationCreatedByExternalService(service_label):
-                return invited_cmds.latest.invite_info.InvitationCreatedByExternalService(
-                    service_label
-                )
-            case unknown:
-                assert False, unknown
-
-    def for_invite_list(self) -> InviteListInvitationCreatedBy:
+    def cook_for_api[T](
```
AureliaDolo, line 50: It's more of a question: but how does this work ? you pass the two callable everytime ?

AureliaDolo (approved):

server/parsec/components/invite.py
touilleMan: Yes pass the two callbacks and let the function choose what it needs

And as a matter of fact, the two callbacks are in fact just the classes that should be instantiated, but the typing is easier to express them as callbacks ^^

Issue #10312, opened 2025-05-16
Check for server addr once


AureliaDolo:
This check should be moved to `api_account_send_email_validation_token`, before the call to `self.create_email_validation_token`, otherwise server will be creating and storing tokens unnecessarily.

_Originally posted by @mmmarcos in https://github.com/Scille/parsec-cloud/pull/10249#discussion_r2089223464_

mmmarcos:
Suggested by @FirelightFlagboy: check once when loading the config and, if not available, replace smtp with a mock implementation.

PR #10649, opened 2025-06-26
[Account] Use validation code in account creation.

### LINKED ISSUE #10554: [Account] Allow to validate email address before account create
Depends on #10553

The token/code sent to the user's email address for account creation is stored in Parsec server.

~Add a route `account_create_check_code` (name TBC) to check if the code is valid before actually performing the request to create an account.~

Instead of adding a route, @AureliaDolo suggests re-using the existing `account_create_proceed` with an enum with 2 variants `Validate` & `Create` where the create variant contains all params except the token.

The route will check the token regardless of the variant, but the `Create` one will create the account.

### LINKED ISSUE #10599: Update `account_create_proceed` to use `ValidationCode` over a token

### LINKED ISSUE #10600: `account_create_send_validation_email` should send a code over a token

### LINKED ISSUE #10667: Finish test account creation after email validation in `libparsec_account`
See test `ok_with_server_and_create`:
https://github.com/Scille/parsec-cloud/pull/10665/files#diff-b3aa0d7c6c9f7f1c7f21b06bf0df30e1e48b34b0fa496a2099d2d2a0fbd5850eR44

tl;dr:  Get validation code from the email and use to to actually create the account once migration out of invitation addr is done

### LINKED ISSUE #10668: Handle attempt counter during account creation code check

### LINKED ISSUE #10669: Handle attempt counter during account creation code proceed

### LINKED ISSUE #10670: Handle code validity on account create check

### LINKED ISSUE #10671: Handle code validity on account create proceed

AureliaDolo:
Fix #10600, fix #10599, fix #10668, fix #10669, fix #10670, fix #10671, fix #10554
fix #10667


server/tests/common/account.py
```diff
@@ -48,19 +49,19 @@ async def create_account(
     created_by_ip: str,
     created_by_user_agent: str,
     created_on: DateTime,
-    human_label: str,
+    human_handle: HumanHandle,
```
touilleMan, line 51: both `account_email` and `human_handle` contain the email, so you should remove `account_email` field

server/parsec/components/memory/datamodel.py
```diff
@@ -907,7 +908,7 @@ class MemoryAccount:
     account_email: EmailAddress
```
touilleMan, line 910: We want a single source of truth, so convert this field into a property:

```python
@property
def account_email(self) -> EmailAddress:
    return self.human_label.email
```

The alternative is to keep email and label separated. So keep `human_label: str` as-is, but to add a comment explaining it is guaranteed to be a valid human handle label since it has been validated beforehand (note this is basically what we always do when storing data into PostgreSQL ^^).

FirelightFlagboy:
Need rebase

server/parsec/components/memory/account.py
```diff
@@ -84,54 +85,59 @@ async def account_info(
...
+    ) -> ValidationCodeComparison:
+        info = self._data.unverified_emails.get(email)
+        if info is None:
+            return ValidationCodeComparison.NOT_FOUND
+
+        (info, last_sent_email) = info
+        if info.is_expired(now):
+            return ValidationCodeComparison.EXPIRED
+        if not info.has_remaining_attempts():
+            return ValidationCodeComparison.NO_REMAINING_ATTEMPTS
```
FirelightFlagboy: **polish (non-blocking):** Should this check remain that precise or should we rely on `ValidationCodeInfo::can_be_used`?

server/parsec/components/memory/datamodel.py
```diff
@@ -105,7 +105,8 @@ class MemoryDatamodel:
     # accounts are not associated to one organization
     # (email, account)
     accounts: dict[EmailAddress, MemoryAccount] = field(default_factory=dict)
-    unverified_emails: dict[EmailAddress, tuple[EmailValidationToken, DateTime]] = field(
+    # (email, ((code, code_created_at, remaining attempts), last_email_sent_at))
+    unverified_emails: dict[EmailAddress, tuple[ValidationCodeInfo, DateTime]] = field(
```
FirelightFlagboy: **todo:** `ValidationCodeInfo` already contain a date (`created_at`) that also correspond when the code was sent.

server/tests/api_v5/anonymous_account/test_account_create_proceed.py
```diff
@@ -58,53 +76,64 @@ async def test_anonymous_account_account_create_proceed_invalid_validation_code(
     assert rep == anonymous_account_cmds.latest.account_create_send_validation_email.RepOk()
 
     # retrieve alice's token that was sent by email
-    token = backend.account.test_get_token_by_email(email)
+    token = backend.account.test_get_code_by_email(email)
     assert token is not None
 
-    # other_token = EmailValidationToken.new()
+    other_code = ValidationCode.generate()
```
FirelightFlagboy: **polish:** unlikely, but need an assert that check that `other_code != code`

server/tests/api_v5/anonymous_account/test_account_create_proceed.py
```diff
@@ -58,53 +76,64 @@ async def test_anonymous_account_account_create_proceed_invalid_validation_code(
     assert rep == anonymous_account_cmds.latest.account_create_send_validation_email.RepOk()
 
     # retrieve alice's token that was sent by email
-    token = backend.account.test_get_token_by_email(email)
+    token = backend.account.test_get_code_by_email(email)
```
FirelightFlagboy: **polish (blocking):** The variable should be renamed to `code`

```suggestion
    code = backend.account.test_get_code_by_email(email)
```

server/tests/api_v5/anonymous_account/test_account_create_proceed.py
```diff
@@ -58,53 +76,64 @@ async def test_anonymous_account_account_create_proceed_invalid_validation_code(
...
+            auth_method_hmac_key=SecretKey.generate(),
+            vault_key_access=b"vault_key_access",
+            auth_method_id=AccountAuthMethodID.from_hex("9aae259f748045cc9fe7146eab0b132e"),
+        )
+    )
+    assert rep == anonymous_account_cmds.latest.account_create_proceed.RepInvalidValidationCode()
 
     # Alice's mail kept in unverified emails
-    token = backend.account.test_get_token_by_email(email)
+    token = backend.account.test_get_code_by_email(email)
```
FirelightFlagboy: **polish:** The variable should be renamed to `code`
```suggestion
    code = backend.account.test_get_code_by_email(email)
```

server/tests/api_v5/anonymous_account/test_account_create_proceed.py
```diff
@@ -118,43 +147,237 @@ async def test_anonymous_account_account_create_proceed_auth_method_id_already_e
...
+    assert rep == anonymous_account_cmds.latest.account_create_send_validation_email.RepOk()
+
+    # retrieve alice's token that was sent by email
+
+    code = backend.account.test_get_code_by_email(email)
+    assert code is not None
+
+    # Set config to not wait an hour
+    monkeypatch.setattr("parsec.components.account.VALIDATION_CODE_VALIDITY_DURATION_SECONDS", 1)
+    await sleep(1)
```
FirelightFlagboy: **issue (blocking):** Do not sleep in tests

libparsec/crates/protocol/schema/anonymous_account_cmds/account_create_proceed.json5
```diff
@@ -19,6 +19,14 @@
       {
         "status": "invalid_validation_code"
       },
+      {
```
touilleMan, line 22: Note this schema differs from the one in RFC1013 (`AccountCreateStep` doesn't exist)

server/parsec/components/memory/datamodel.py
AureliaDolo: The difference is that if the send email api is called before the code expires, but after the interval before sending a new email, last_email_sent_at will be updated when the new email is sent but not the created at, because the code is still the same.

FirelightFlagboy: I thought that when we send an email, we always generate a new code

AureliaDolo: You're right, my bad !

server/parsec/components/memory/account.py
```diff
@@ -84,54 +85,58 @@ async def account_info(
...
         elif email in self._data.unverified_emails:
-            (_, last_email_datetime) = self._data.unverified_emails[email]
-            if not self.can_send_new_code_email(last_email_datetime, now):
-                return AccountCreateEmailValidationTokenBadOutcome.TOO_SOON_AFTER_PREVIOUS_DEMAND
+            info = self._data.unverified_emails[email]
+            if not self.can_send_new_code_email(info.created_at, now):
+                return AccountCreateEmailValidationCodeBadOutcome.TOO_SOON_AFTER_PREVIOUS_DEMAND
+
+        code = ValidationCode.generate()
+        info = ValidationCodeInfo(code, now, 0)
```
FirelightFlagboy: **nitpick (non-blocking):** The `failed_attempts` field use a default value, so we're not required to provide it when initializing a new instance.

```suggestion
        info = ValidationCodeInfo(code, now)
```

If you prefer explicitly setting it to 0, I would use a named parameters:

```suggestion
        info = ValidationCodeInfo(code, now, failed_attempts=0)
```

server/parsec/components/memory/account.py
```diff
@@ -151,20 +156,20 @@ async def create_account(
...
-        self._data.unverified_emails.pop(email)
+        # if nothing is removed, that means the code has not been checked
+        assert self._data.unverified_emails.pop(human_handle.email) is not None
 
     @override
-    def test_get_token_by_email(self, email: EmailAddress) -> EmailValidationToken | None:
+    def test_get_code_by_email(self, email: EmailAddress) -> ValidationCode | None:
         """Use only in tests, nothing is checked."""
-        return self._data.unverified_emails[email][0]
+        return self._data.unverified_emails[email].code
```
FirelightFlagboy: **todo:** The method should return None if the email is not found and not raise an exception

```suggestion
        code_info = self._data.unverified_emails.get(email)
        if code_info:
        	return code_info.code
```

libparsec/crates/protocol/schema/anonymous_account_cmds/account_create_proceed.json5
touilleMan: done.

server/parsec/components/memory/account.py
touilleMan: done :+1:

server/parsec/templates/email/account_validation.html.j2
```diff
@@ -18,12 +18,12 @@
       ]
     ),
     namespace(
-      text="Once installed, click on the link below to confirm your email address.",
+      title="Use validation code",
+      text="To confirm the validation of your account, use the following code:",
```
mmmarcos:
```suggestion
      text="To confirm your email address, use the following code:",
```

server/parsec/components/account.py
```diff
@@ -132,26 +131,71 @@ class VaultItems:
...
+        )
+
+    # Used by `delete_proceed` implementations
+    async def _send_account_delete_validation_email(
+        self,
+        email: EmailAddress,
+        validation_code: ValidationCode,
+    ) -> None | SendEmailBadOutcome:
+        if not self._config.server_addr:
+            return SendEmailBadOutcome.BAD_SMTP_CONFIG
```
mmmarcos, line 186: Probably for another issue/PR: this check is used in many places, it could be done by a decorator provided by the `email` component.

AureliaDolo (commented):
@touilleMan :+1: for the changes you made

server/parsec/components/account.py
AureliaDolo: There is already an issue for that ^^ https://github.com/Scille/parsec-cloud/issues/10312

server/parsec/templates/email/account_validation.html.j2
touilleMan: done :+1:

PR #10650, opened 2025-06-26
Rework `libparsec_platform_device_loader` so that it is not responsible for generating the new device when importing a recovery device

touilleMan:


touilleMan:
> Would be nice to document a it what is each method's responsibility

done :+1:

PR #10651, opened 2025-06-27
Add newsfragment for last updater

mmmarcos:
Related to #8333

### DIFF
--- a/newsfragments/10524.feature.rst
+++ b/newsfragments/10524.feature.rst
@@ -1,1 +1,1 @@
-Display login and logout button in the homepage
+Display login and logout buttons in the homepage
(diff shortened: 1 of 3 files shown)


newsfragments/10524.feature.rst
```diff
@@ -1 +1 @@
-Display login and logout button in the homepage
+Display login and logout buttons in the homepage
```
Max-7: This newsfragment should be removed. It's linked to Parsec Account, which is not enabled yet.

Max-7 (changes requested):

Max-7 (approved):

PR #10657, opened 2025-06-27
Add fetch/list/register registration devices methods to `liparsec_account::Account`

touilleMan:


libparsec/crates/account/src/lib.rs
```diff
@@ -23,391 +30,148 @@ const AUTH_METHOD_SECRET_KEY_DERIVATION_UUID: uuid::Uuid =
...
-    pub organization_id: OrganizationID,
-    pub user_id: UserID,
-}
+    /// Since the validation code is supposed to be copied by the user from his
+    /// email client to the Parsec client by hand, this method can be used to
+    /// ensure the validation code + email couple is valid and can lead.
+    ///
+    /// Also note that calling too many time this method with an invalid
+    /// validation code leads to a full invalidation of the attempt and requires
+    /// a new validation email to be sent.
```
AureliaDolo, line 62: Maybe add that despite the naming this step is actually optional.

libparsec/crates/account/src/login.rs
```diff
@@ -0,0 +1,163 @@
...
+    addr: ParsecAddr,
+    auth_method_master_secret: KeyDerivation,
+    // We must recycle anonymous_cmds if it exists, otherwise its drop will complain
+    // during tests if some request registered with `test_register_sequence_of_send_hooks`
+    // has yet to occur.
+    anonymous_cmds: Option<AnonymousAccountCmds>,
+    // Since human handle is supposed to be fetched from the server, this parameter is
+    // always set to `None`... Unless for `Account::test_new` where we precisely want
+    // to avoid this initial server request!
+    human_handle: Option<HumanHandle>,
```
AureliaDolo, line 58: Maybe something like that would work,if we wanna be nit-picky
```suggestion
    #[cfg(test)]
    human_handle: HumanHandle,
```
and latter 
```rust
let human_handle = if cfg!(test) {
    human_handle
} else {
    // do the request
};
```

libparsec/crates/account/src/register_new_device.rs
```diff
@@ -0,0 +1,114 @@
...
+}
+
+pub(super) async fn account_register_new_device(
+    account: &Account,
+    organization_id: OrganizationID,
+    user_id: UserID,
+    new_device_label: DeviceLabel,
+    save_strategy: DeviceSaveStrategy,
+) -> Result<AvailableDevice, AccountRegisterNewDeviceError> {
+    // 1. Retreive the registration device from the cache
```
AureliaDolo:
```suggestion
    // 1. Retrieve the registration device from the cache
```

libparsec/crates/account/src/lib.rs
touilleMan: done.

libparsec/crates/account/src/login.rs
touilleMan: I might work, bu it's rather cumbersome since we need to add `#[cfg(test)]` both in the function signature + the function body + the call sites...

So I'd rather stick with the current approach (also note that in release the compiler is supposed to be smart enough to determine we always call the function with a `None` constant and inline this away)

libparsec/crates/account/src/register_new_device.rs
touilleMan: :'( done

touilleMan:
> Should we rename (in another issue) the steps in account create proceed schema to step 1 and step 2 to be consistent with the js bindings ?

I would normally agree on this, however in the current case I don't think this is needed:
1. The number of steps is very low (step 1: send the validation mail, step 2: proceed to do the actual operation)
2. The number of steps differs between the client/server API and the bindings (since the step 2 in the API is used for both step 2 and 3 in the bindings)

The last point is especially important: having different amount of steps for what is seemingly the same component on different layers is very puzzling :/

So the good solution I see would be to rename steps 2 and 3 into step 2a and step 2b (this way it's obvious those steps both rely on the client/server API's step 2).

However even with this I feel like this is more Yak-shaving that something actually important (see point 1), while we still have plenty of work to do on this feature...

PR #10665, opened 2025-06-30
Add testbed mailbox check

touilleMan:


libparsec/crates/testbed/src/env.rs
```diff
@@ -1161,3 +1161,61 @@ pub async fn test_drop_testbed(discriminant_dir: &Path) -> anyhow::Result<()> {
 
     Ok(())
 }
+
+pub async fn test_check_mailbox(
```
AureliaDolo, line 1167:
```suggestion
/// Returns (sender, timestamp, body)
pub async fn test_check_mailbox(
```

libparsec/src/testbed.rs
```diff
@@ -85,3 +85,21 @@ pub async fn test_drop_testbed(discriminant_dir: &Path) -> Result<(), TestbedErr
 pub async fn test_drop_testbed(_discriminant_dir: &Path) -> Result<(), TestbedError> {
     Err(TestbedError::Disabled)
 }
+
+#[cfg(feature = "test-utils")]
```
AureliaDolo, line 89:
```suggestion
#[cfg(feature = "test-utils")]
/// Returns (sender, timestamp, body)
```

libparsec/crates/testbed/src/env.rs
touilleMan: :+1: done.

libparsec/src/testbed.rs
touilleMan: :+1: done.

libparsec/crates/testbed/src/env.rs
```diff
@@ -1161,3 +1161,114 @@ pub async fn test_drop_testbed(discriminant_dir: &Path) -> anyhow::Result<()> {
 
     Ok(())
 }
+
+/// Returns the list of received emails (sender, timestamp, body), ordered
+/// from oldest to newest.
+pub async fn test_check_mailbox(
+    server_addr: &ParsecAddr,
+    email: &EmailAddress,
+) -> anyhow::Result<Vec<(EmailAddress, DateTime, String)>> {
```
FirelightFlagboy, line 1170: **nitpick:** I would prefer a struct with named field over relying on a doc-comment

libparsec/crates/testbed/src/env.rs
```diff
@@ -1161,3 +1161,114 @@ pub async fn test_drop_testbed(discriminant_dir: &Path) -> anyhow::Result<()> {
...
+            let as_bytes = data_encoding::BASE64
+                .decode(b64.as_bytes())
+                .map_err(|_| ())?;
+            String::from_utf8(as_bytes).map_err(|_| ())
+        };
+
+        let (sender, timestamp, body) = match (
+            s.next().map(EmailAddress::from_str),
+            s.next().map(DateTime::from_str),
+            s.next().map(base64_to_str),
```
FirelightFlagboy, line 1208: **issue (blocking):** Not all emails are sent in base64, you need to check the charset

libparsec/crates/testbed/src/env.rs
```diff
@@ -1161,3 +1161,114 @@ pub async fn test_drop_testbed(discriminant_dir: &Path) -> anyhow::Result<()> {
...
+            }
+        };
+
+        mails.push((sender, timestamp, body));
+    }
+
+    Ok(mails)
+}
+
+pub async fn test_new_account(
```
FirelightFlagboy, line 1228: **polish:** This function is not related to this PR, could you make it simpler to review be proposing that in another PR?

FirelightFlagboy (changes requested):

libparsec/crates/testbed/src/env.rs
touilleMan: I agree they are not strictly related

However this is not much code, in both case they are related to the testbed system and splitting the PR is much more work (since the testbed server has to be regenerate for both routes).

So I'd rather keep that in the PR and merge it asap ;-)

libparsec/crates/testbed/src/env.rs
touilleMan: On server side we use Python's `email::Message::as_string()` to return the mail body as a `str`, then in `GET /testbed/mailbox/{raw_email}` we encode the mail body with utf8 then base64 the result.

So the base64 encoding here is not related how email might use base64.

libparsec/crates/testbed/src/env.rs
touilleMan: ~:+1: done.~

EDIT: this is more complex to implement that expected: the introducted type must be exposed in the bindings, however `libparsec_testbed` is only compiled when the `test-utils` feature is enabled.
Hence:
- Since the type is declared in `libparsec_testbed`, it must be redefined in `libparsec::testbed` when `test-utils` is not enabled (otherwise release compilation fails)
- The type re-declared in `libparsec::testbed` must be the same type than in `libparsec_testbed` (so not a stub like `pub type TestCheckMailboxMail = ();` since the bindings code actually uses it
- Going out-of-sync between the actual type and the type re-declared is a low risk (since this type is not supposed to change often if at all), however it creates a bug that is only going to be detected when the release compilation is done (so very last, typically when we want to release a new version)
- Another approach is to declare the type in `libparsec_types`, but this is cumbersome since we end up with a type used in a single very narrow usecase but defined very far away from it...

So I feel like there is no clear winner here, and the least worst approach is to stick we what we currently have :/

libparsec/crates/testbed/src/env.rs
FirelightFlagboy: `Message::as_string` can return you a string that is encoded in base64, that's the issue else `export-email` would have been simpler to implement

server/parsec/config.py
```diff
@@ -168,7 +169,11 @@ class MockedEmailConfig:
     type = "MOCKED"
 
     sender: EmailAddress
-    tmpdir: str
+    # Emails sent with recipient as key and  list of <sender, timestamp, body> as value.
+    # Keeping the emails is needed for testbed server's `GET /testbed/mailbox/{email}` route.
+    sent_emails: dict[EmailAddress, list[tuple[EmailAddress, DateTime, str]]] = field(
```
FirelightFlagboy: **polish:** Also use a named tuple here (or dataclass)

touilleMan: :+1: done.

libparsec/crates/testbed/src/env.rs
touilleMan: > Message::as_string can return you a string that is encoded in base64

This is irrelevant here: `GET /testbed/mailbox/{raw_email}` only focus on returning the email body, whatever its contains

Of course this totally relevant on the test layer that will actually dig into the email to find the validation code (or whatever it is looking for)
So if the whole mail is encoded as base64 due to some weird mime-type, the test layer is going to have a hard time finding what it is looking for.

But this is not an issue for this PR (and it is most likely not an issue at all since I've already checked that the email we send have their HTML content dumped as-is in the stringified body)

server/parsec/cli/testbed.py
```diff
@@ -233,6 +244,83 @@ async def test_drop(raw_organization_id: str, request: Request) -> Response:
...
+
+    # Body should be something like `<sender_email>\t<timestamp>\t<base64(body)`,
+    # also multiple mails can be send one after the other separated by `\n`
+
+    def _encode_mail(mail: MockedSentEmail) -> bytes:
+        return f"{mail.sender.str}\t{mail.timestamp.to_rfc3339()}\t".encode() + b64encode(
+            mail.body.encode("utf8")
+        )
+
+    rep_body = b"\n".join(_encode_mail(mail) for mail in sent_emails)
```
FirelightFlagboy, line 271: **question:** Why not just json ? seems simpler than custom encoding of multiple values

libparsec/crates/testbed/src/env.rs
FirelightFlagboy: The issue still remain that it's not clear whether the body is encoded in base64 or not

server/parsec/cli/testbed.py
touilleMan: JSON is simpler on the Python side, but is still cumbersome on the Rust side since we need to define a struct with serde support and add a json decoder lib :/

That why so far we have managed to use really crude encoding for each testbed route. This current route is the one having the most complexity but is still manageable to encode by hand so I'd rather stay consistent with the rest.

libparsec/crates/testbed/src/env.rs
touilleMan: > The issue still remain that it's not clear whether the body is encoded in base64 or not

This will be dealt with in PR https://github.com/Scille/parsec-cloud/pull/10649 that is planned to implement a test to retrieve the validation code from the email ;-p

PR #10666, opened 2025-06-30
Add libparsec client `get_user_info` and bindings

### LINKED ISSUE #6485: Add a `getUserInfo` functions to bindings
Something that takes a `UserID` and returns the associated `UserInfo` (or a not found error or whatever).

This can be done in typescript only using a find on the list returned by `clientListUsers()` (it is in fact how I've implemented it for now) but I'm guessing that Parsec can do this a bit better.

mmmarcos:
Closes #6485

### DIFF
--- a/bindings/generator/api/client.py
+++ b/bindings/generator/api/client.py
@@ -228,6 +228,24 @@ async def client_revoke_user(
     raise NotImplementedError
 
 
+class ClientGetUserInfoError(ErrorVariant):
+    class Stopped:
+        pass
+
+    class NonExisting:
+        pass
+
+    class Internal:
+        pass
+
+
+async def client_get_user_info(
+    client: Handle,
+    user_id: UserID,
+) -> Result[UserInfo, ClientGetUserInfoError]:
+    raise NotImplementedError
+
+
 class ClientListUsersError(ErrorVariant):
     class Stopped:
         pass
--- a/bindings/electron/src/index.d.ts
+++ b/bindings/electron/src/index.d.ts
@@ -1095,6 +1095,25 @@ export type ClientGetUserDeviceError =
   | ClientGetUserDeviceErrorStopped
 
 
+// ClientGetUserInfoError
+export interface ClientGetUserInfoErrorInternal {
+    tag: "Internal"
+    error: string
+}
+export interface ClientGetUserInfoErrorNonExisting {
+    tag: "NonExisting"
+    error: string
+}
+export interface ClientGetUserInfoErrorStopped {
+    tag: "Stopped"
+    error: string
+}
+export type ClientGetUserInfoError =
+  | ClientGetUserInfoErrorInternal
+  | ClientGetUserInfoErrorNonExisting
+  | ClientGetUserInfoErrorStopped
+
+
 // ClientInfoError
 export interface ClientInfoErrorInternal {
     tag: "Internal"
@@ -3958,6 +3977,10 @@ export function clientGetUserDevice(
     client: number,
     device: string
 ): Promise<Result<[UserInfo, DeviceInfo], ClientGetUserDeviceError>>
+export function clientGetUserInfo(
+    client: number,
+    user_id: string
+): Promise<Result<UserInfo, ClientGetUserInfoError>>
 export function clientInfo(
     client: number
 ): Promise<Result<ClientInfo, ClientInfoError>>
--- a/bindings/electron/src/meths.rs
+++ b/bindings/electron/src/meths.rs
@@ -6222,6 +6222,33 @@ fn variant_client_get_user_device_error_rs_to_js<'a>(
     Ok(js_obj)
 }
 
+// ClientGetUserInfoError
+
+#[allow(dead_code)]
+fn variant_client_get_user_info_error_rs_to_js<'a>(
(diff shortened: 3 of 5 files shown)


bindings/generator/api/client.py
```diff
@@ -228,6 +228,24 @@ async def client_revoke_user(
...
+    class NonExisting:
+        pass
+
+    class Internal:
+        pass
+
+
+async def client_get_user_info(
+    client: Handle,
+    user_id: bool,
```
FirelightFlagboy: **issue (blocking):** Seems like a typo, why `user_id` is a boolean and not `UserID` ?

mmmarcos: A typo indeed, now fixed.

libparsec/crates/client/tests/unit/certif/get_user_info.rs
```diff
@@ -0,0 +1,82 @@
...
+    let ops = certificates_ops_factory(env, &alice).await;
+
+    let res = ops.get_user_info(alice.user_id).await.unwrap();
+
+    p_assert_eq!(res.id, alice.user_id);
+    p_assert_eq!(res.human_handle, alice.human_handle);
+    p_assert_eq!(res.current_profile, UserProfile::Admin);
+    p_assert_eq!(
+        res.created_on,
+        DateTime::from_ymd_hms_us(2000, 1, 2, 0, 0, 0, 0).unwrap()
```
touilleMan: nitpick:

```suggestion
        "2000-01-02T00:00:00Z".parse().unwrap()
```

libparsec/crates/client/src/certif/list.rs
```diff
@@ -151,6 +131,90 @@ pub(super) async fn list_users(
...
+impl From<CertifStoreError> for CertifGetUserInfoError {
+    fn from(value: CertifStoreError) -> Self {
+        match value {
+            CertifStoreError::Stopped => Self::Stopped,
+            CertifStoreError::Internal(err) => err.into(),
+        }
+    }
+}
+
+pub(super) async fn get_user_info_for_user_certificates(
```
touilleMan:
```suggestion
async fn get_user_info_for_user_certificates(
```

FirelightFlagboy (approved):

libparsec/crates/client/src/certif/list.rs
touilleMan: And I would rename the method since in the other functions "get" means the certificate are going to be obtains from the storage, while here we provide it

something like `cook_user_info_from_certificates()`

libparsec/crates/client/src/certif/list.rs
```diff
@@ -151,6 +131,90 @@ pub(super) async fn list_users(
...
+    user_certificate: std::sync::Arc<UserCertificate>,
```
touilleMan: In theory it cleaner to use references here:
```suggestion
    user_certificate: &UserCertificate,
    revoked_certificate: Option<&RevokedUserCertificate>,
```

And if I recall correctly you can turn a `Option<Arc<X>>` into a `Option<&X>` in the caller just by doing `x.as_ref()

However it is an internal helper, so if the caller becomes too cumbersome we should just stick to the simpler approach :+1:

libparsec/crates/client/src/certif/list.rs
```diff
@@ -151,6 +131,90 @@ pub(super) async fn list_users(
...
+    revoked_certificate: Option<std::sync::Arc<RevokedUserCertificate>>,
+    update_certificate: Option<std::sync::Arc<UserUpdateCertificate>>,
+) -> UserInfo {
+    let (revoked_on, revoked_by) = match revoked_certificate {
+        None => (None, None),
+        Some(certif) => (Some(certif.timestamp), Some(certif.author.to_owned())),
+    };
+    let current_profile = match update_certificate {
+        None => user_certificate.profile,
+        Some(certif) => certif.new_profile,
```
touilleMan:
```suggestion
        Some(update_certificate) => update_certificate.new_profile,
```
(I'd rather make it extra clear the split between `user_certificate` vs `update_certificate` on where the profile is fetched)

libparsec/crates/client/src/certif/list.rs
mmmarcos: `x.as_ref()` returns `Option<&Arc<X>>` (not `Option<&X>`). 

The only way I found is to take a ref to the Option and map `Option<Arc<X>>` to `Option<&X>` with `x.as_ref().map(|x| x.as_ref())`

So caller changes from:
```rust
cook_user_info_from_certificates(user_certificate, revoked_certificate, update_certificate)
```
to
```rust
cook_user_info_from_certificates(
    user_certificate.as_ref(),
    revoked_certificate.as_ref().map(|x| x.as_ref()),
    update_certificate.as_ref().map(|x| x.as_ref()),
)
```
Any preference?

FirelightFlagboy: @mmmarcos What you are looking is for `as_deref`

mmmarcos: Done, thx @FirelightFlagboy!

FirelightFlagboy (approved):

PR #10678, opened 2025-07-02
Libparsec account create registration device

touilleMan:


libparsec/crates/account/src/create_registration_device.rs
```diff
@@ -0,0 +1,150 @@
...
+        let req = Req {
+            item_fingerprint: item.fingerprint(),
+            item: item.dump().into(),
+        };
+        match account.cmds.send(req).await? {
+            Rep::Ok => (),
+            bad_rep @ (
+                // Unexpected since the fingerprint is a hash containing, among
+                // other things, the randomly chosen device ID.
+                Rep::FingerprintAlreadyExists
```
AureliaDolo, line 127: Maybe it could be useful to keep these errors separated, as for the fingerprint error, the front could retry but not for the unknown status

touilleMan: As the comment says, this error is never supposed to occur under normal circumstances since the fingerprint is a SHA256 of UUIDs (device & user IDs)

PR #10683, opened 2025-07-02
feat(server): Add hardened default TLS cipher suite

FirelightFlagboy:
Followed the recommendation made by [ANSSI - Recommendation TLS](https://cyber.gouv.fr/sites/default/files/2017/07/anssi-guide-recommandations_de_securite_relatives_a_tls-v1.2.pdf)


server/parsec/cli/run.py
```diff
@@ -367,6 +367,33 @@ def handle_parse_result(
...
+    envvar="PARSEC_SSL_CIPHER",
+    show_envvar=True,
+    # List of recommended cipher suite by ANSSI
+    # https://cyber.gouv.fr/sites/default/files/2017/07/anssi-guide-recommandations_de_securite_relatives_a_tls-v1.2.pdf
+    default=":".join(
+        [
+            # For TLSv1.3
+            "TLS_AES_256_GCM_SHA384",
+            "TLS_AES_128_GCM_SHA256",
+            "TLS_AES_128_CCM_SHA256",
```
mmmarcos: What about `TLS_CHACHA20_POLY1305_SHA256`?

server/parsec/cli/run.py
```diff
@@ -367,6 +367,33 @@ def handle_parse_result(
     show_envvar=True,
     help="SSL certificate file. This setting enables serving Parsec over SSL.",
 )
+@click.option(
+    "--ssl-ciphers",
+    type=str,
+    envvar="PARSEC_SSL_CIPHER",
```
mmmarcos:
```suggestion
    envvar="PARSEC_SSL_CIPHERS",
```

server/parsec/cli/run.py
```diff
@@ -367,6 +367,33 @@ def handle_parse_result(
...
+    show_envvar=True,
+    # List of recommended cipher suite by ANSSI
```
mmmarcos:
```suggestion
    # List of recommended cipher suites by ANSSI
```

server/parsec/cli/run.py
FirelightFlagboy: Just forgotten :shrug:

FirelightFlagboy: done

server/parsec/cli/run.py
```diff
@@ -367,6 +367,34 @@ def handle_parse_result(
...
+            "TLS_ECDHE_ECDSA_WITH_AES_256_CCM",
+            "TLS_ECDHE_ECDSA_WITH_AES_128_CCM",
+            "TLS_ECDHE_ECDSA_WITH_CHACHA20_POLY1305_SHA256",
+            # For TLSv1.2 using RSA
+            "TLS_ECDHE_RSA_WITH_AES_256_GCM_SHA384",
+            "TLS_ECDHE_RSA_WITH_AES_128_GCM_SHA256",
+            "TLS_ECDHE_RSA_WITH_CHACHA20_POLY1305_SHA256",
+        ]
+    ),
+    help="SSL Ciphers list separated by colon (:)",
```
touilleMan:
```suggestion
    help="Comma-separated list of allowed SSL Ciphers",
```

Use `,` as separator that's what we already use in `--organization-initial-tos`

server/parsec/cli/run.py
```diff
@@ -367,6 +367,34 @@ def handle_parse_result(
...
+)
```
touilleMan, line 402: Is it possible to check the validity of the provided value and return list of allowed values if an invalid value is provided ?

[SSLContext.get_ciphers](https://docs.python.org/3/library/ssl.html#ssl.SSLContext.get_ciphers) might be what we are looking for:
```python
>>> import ssl
>>> >>> [x['name'] for x in ssl.SSLContext(ssl.PROTOCOL_TLS_SERVER).get_ciphers()]
['TLS_AES_256_GCM_SHA384', 'TLS_CHACHA20_POLY1305_SHA256', 'TLS_AES_128_GCM_SHA256', 'ECDHE-ECDSA-AES256-GCM-SHA384', 'ECDHE-RSA-AES256-GCM-SHA384', 'ECDHE-ECDSA-AES128-GCM-SHA256', 'ECDHE-RSA-AES128-GCM-SHA256', 'ECDHE-ECDSA-CHACHA20-POLY1305', 'ECDHE-RSA-CHACHA20-POLY1305', 'ECDHE-ECDSA-AES256-SHA384', 'ECDHE-RSA-AES256-SHA384', 'ECDHE-ECDSA-AES128-SHA256', 'ECDHE-RSA-AES128-SHA256', 'DHE-RSA-AES256-GCM-SHA384', 'DHE-RSA-AES128-GCM-SHA256', 'DHE-RSA-AES256-SHA256', 'DHE-RSA-AES128-SHA256']
```

Note however that the name of some ciphers doesn't match the ones listed in the default config, e.g. `TLS_ECDHE_ECDSA_WITH_AES_256_GCM_SHA384` vs `ECDHE-ECDSA-AES256-GCM-SHA384`

So I might be due to:
- An error on the naming of the default config (taken from https://cyber.gouv.fr/sites/default/files/2017/07/anssi-guide-recommandations_de_securite_relatives_a_tls-v1.2.pdf ?), but in this case it would mean uvicorn accept unknown name of ciphers without error which is weird :/
- The config provided to uvicorn requires a different name than the one we obtain. This would most likely mean uvicorn relies on some other component to obtain the available ciphers, so we shouldn't use `get_ciphers()` (since it doesn't actually validate anything in this case !)

In any case, if it is too cumbersome to add this, we can also just provide a link in the param documentation to some website that provides the list of allowed ciphers :+1:

server/parsec/cli/run.py
FirelightFlagboy: I do not make the rules, it's what `openssl` is expecting

> The cipher list consists of one or more cipher strings separated by colons
>
> https://docs.openssl.org/1.1.1/man1/ciphers/#cipher-list-format

server/parsec/cli/run.py
FirelightFlagboy: > but in this case it would mean uvicorn accept unknown name of ciphers without error which is weird :/

Not weird, you provide a list and the SSL layout try to work with it.
Depending on your openssl version, not all ciphers may be available

> The config provided to uvicorn requires a different name than the one we obtain

Univcorn is simple, it use the `ssl` module

server/parsec/cli/run.py
FirelightFlagboy: I've reworked the option to be comma seperated, and build the list for openssl when creating the uvicorn config

docs/hosting/deployment/index.rst
```diff
@@ -431,3 +435,43 @@ The important takeaways are:
     The ``Forwarded`` header (`RFC-7239 <https://datatracker.ietf.org/doc/html/rfc7239>`_) is not used by Parsec, but it may be in the future.
 
 - Set the header ``host`` to the accessible address. Here we force the value to be ``example.com``, but you can set it to ``$host`` like for ``X-Forwarded-Host``.
+
+TLS Recommendation
+******************
+
+We recommend that connection to the service are using a TLS layer.
```
touilleMan:
```suggestion
We recommend that connections to the service are made using a TLS layer.
```

docs/hosting/deployment/index.rst
```diff
@@ -431,3 +435,43 @@ The important takeaways are:
...
+- `Traefik TLS configuration <https://doc.traefik.io/traefik/https/tls>`__
+
+Or if you do not use a reverse proxy, see how to configure :ref:`TLS on the server <server-tls-config>`
+
+.. _server-tls-config:
+
+TLS Server configuration
+************************
+
+We recommend that when user directly connect to the server (without using a :ref:`reverse proxy<behind_a_proxy>`)
```
touilleMan:
```suggestion
We recommend that when user directly connects to the server (i.e. without using a :ref:`reverse proxy<behind_a_proxy>`)
```

server/parsec/asgi/__init__.py
```diff
@@ -177,6 +178,8 @@ async def serve_parsec_asgi_app(
         log_config=None,
         ssl_keyfile=ssl_keyfile,
         ssl_certfile=ssl_certfile,
+        # openssl expect the list to be colon separated.
```
touilleMan:
```suggestion
        # openssl expects the list to be colon separated.
```

server/tests/cli/test_run.py
```diff
@@ -40,7 +40,7 @@ def ssl_conf(request: pytest.FixtureRequest) -> Generator[SSLConf, None, None]:
             server_cert.private_key_pem.tempfile() as server_keyfile,
         ):
             yield SSLConf(
-                backend_opts=f" --ssl-keyfile={server_keyfile} --ssl-certfile={server_certfile} ",
+                backend_opts=f" --ssl-keyfile={server_keyfile} --ssl-certfile={server_certfile} --ssl-ciphers=DEFAULT",
```
touilleMan: `--ssl-ciphers=DEFAULT` is already the default value when nothing is specified, so no need to add it explicitly

server/parsec/cli/run.py
```diff
@@ -367,6 +367,42 @@ def handle_parse_result(
...
+    envvar="PARSEC_SSL_CIPHERS",
+    show_envvar=True,
+    callback=lambda ctx, param, value: value.split(","),
+    # List of recommended cipher suites by ANSSI
+    # https://cyber.gouv.fr/sites/default/files/2017/07/anssi-guide-recommandations_de_securite_relatives_a_tls-v1.2.pdf
+    default=",".join(
+        [
+            # We need to translate the cipher names to be compatible with openssl naming,
+            # We use https://docs.openssl.org/3.5/man1/openssl-ciphers/#cipher-suite-names for that.
+            # For TLSv1.3, openssl use the same naming so no translation.
```
touilleMan:
```suggestion
            # For TLSv1.3, openssl uses the same naming so no translation.
```

server/parsec/cli/run.py
```diff
@@ -367,6 +367,42 @@ def handle_parse_result(
     show_envvar=True,
     help="SSL certificate file. This setting enables serving Parsec over SSL.",
 )
+@click.option(
+    "--ssl-ciphers",
+    type=str,
+    envvar="PARSEC_SSL_CIPHERS",
+    show_envvar=True,
+    callback=lambda ctx, param, value: value.split(","),
```
touilleMan, line 394: Since gunicorn uses ssl, we can use this code to improve the error message:

```python
def _parse_ssl_ciphers(value: str) -> list[str]:
    # Use a list instead of a set here since this is not performance critical and
    # it preserves protocol order in the error message
    available_protocols = [x['name'] for x in ssl.SSLContext(ssl.PROTOCOL_TLS_SERVER).get_ciphers()]
    protocols = []
    for protocol in value.split(","):
        if protocol not in available_protocols:
            raise ValueError("Unknown protocol `{protocol}`, allowed values are:\n" + "\n".join(f"- {p}" for p in available_protocols)
        protocols.append(protocol)
    return protocols
```

server/parsec/cli/run.py
touilleMan: > Not weird, you provide a list and the SSL layout try to work with it.
> Depending on your openssl version, not all ciphers may be available

This seems very error prone (as a typo in the cipher name won't be detected and instead lead to the cipher not being used at all :scream: )

So doing our own validation of SSL cipher name first seems like a good idea (see my other comment suggesting how to do that)

server/parsec/cli/run.py
touilleMan: we dig further, it is more complicated than expected (to handle the default value, and also the generic values such as `TLS13` that select multiple ciphers)

server/parsec/cli/run.py
touilleMan: see https://github.com/Scille/parsec-cloud/pull/10683#discussion_r2221848477

server/tests/cli/test_run.py
touilleMan: Add a comment to explain that `DEFAULT` is openssl's default and not our default config

server/parsec/cli/run.py
```diff
@@ -386,6 +386,42 @@ def handle_parse_result(
...
+            # https://docs.openssl.org/3.5/man1/openssl-ciphers/#chacha20-poly1305-cipher-suites-extending-tls-v12
+            "ECDHE-ECDSA-CHACHA20-POLY1305",  # TLS_ECDHE_ECDSA_WITH_CHACHA20_POLY1305_SHA256
+            # using RSA
+            "ECDHE-RSA-AES256-GCM-SHA384",  # TLS_ECDHE_RSA_WITH_AES_256_GCM_SHA384
+            "ECDHE-RSA-AES128-GCM-SHA256",  # TLS_ECDHE_RSA_WITH_AES_128_GCM_SHA256
+            # https://docs.openssl.org/3.5/man1/openssl-ciphers/#chacha20-poly1305-cipher-suites-extending-tls-v12
+            "ECDHE-RSA-CHACHA20-POLY1305",  # TLS_ECDHE_RSA_WITH_CHACHA20_POLY1305_SHA256
+        ]
+    ),
+    help="Comma separated of TLS cipher suite to use",
```
touilleMan: Should improve the help with something similar than https://docs.gunicorn.org/en/stable/settings.html#ciphers

typically we need to provide a link to the list of valid cipher names

touilleMan (approved):

FirelightFlagboy (commented):
- Remove default ciphers values
- Set ciphers values in parsec.env + explain
- Remove ssl-ciphers option in ssl test

server/tests/cli/test_run.py
FirelightFlagboy: The change was revert

server/parsec/cli/run.py
FirelightFlagboy: I've taken inspiration from guvicorn documentation

PR #10686, opened 2025-07-03
[MS] More custom branding

Ironicbay:


client/src/views/home/IncompatibleEnvironmentModal.vue
```diff
@@ -7,9 +7,15 @@
       :close-button="{ visible: false }"
     >
       <div class="incompatible-image-container">
+        <ms-image
```
Max-7, line 11: I don't htink you need the ``, you should be able to use only `ms-image`:

```
<ms-image :image="ResourcesManager.instance().get(Resources.LogoIncompatible, IncompatibleIcon) as string">

import IncompatibleIcon from '@/assets/images/logo-incompatible.svg?raw';
```

client/src/views/organizations/creation/ServerTypeChoice.vue
```diff
@@ -29,7 +29,13 @@
             src="@/assets/images/background/background-shapes-small.svg"
             id="center"
           />
+          <ms-image
```
Max-7, line 33: Same here, `ms-image` should be enough.

client/src/views/organizations/creation/OrganizationCreationPage.vue
```diff
@@ -22,6 +22,13 @@
 <script setup lang="ts">
 import { IonPage, IonText } from '@ionic/vue';
 import { MsImage, MsSpinner, LogoIconGradient, LogoIconWhite, useWindowSize } from 'megashark-lib';
+import { Resources, ResourcesManager } from '@/services/resourcesManager';
+
+function getResourceImage(custom: Resources, defaultImage: string): string {
+  console.log(custom, defaultImage, ResourcesManager.instance().get(custom, defaultImage) as string);
```
Max-7: Remove this log.

Max-7 (approved):

PR #10688, opened 2025-07-03
Account create/delete/recover email sending rate limit

touilleMan:


server/parsec/cli/options.py
```diff
@@ -474,8 +474,6 @@ def blockstore_server_options[**P, T](fn: Callable[P, T]) -> Callable[P, T]:
 Each configuration must be provided with the form
 `<raid_type>:<node>:<config>` with `<raid_type>` RAID0/RAID1/RAID5, `<node>` a
 integer and `<config>` the MOCKED/POSTGRESQL/S3/SWIFT config.
-
-\b
```
FirelightFlagboy, line 478: **issue (if-minor):** That `\b` is important for the padding if I remember correctly

server/parsec/cli/run.py
```diff
@@ -243,18 +261,29 @@ def handle_parse_result(
     default=AccountVaultStrategy.ALLOWED,
 )
 @click.option(
-    "--validation-email-cooldown-delay",
-    default=60,
+    "--validation-email-rate-limit",
+    default="60:3",
```
FirelightFlagboy: **nitpick:** I would have used `nargs=2` with `type=(int, int)`

server/parsec/email_rate_limit.py
```diff
@@ -0,0 +1,96 @@
...
+
+class EmailRateLimit:
+    def __init__(
+        self,
+        cooldown_delay,
+        max_per_hour,
+    ):
+        self.cooldown_delay = cooldown_delay
+        self.max_per_hour = max_per_hour
+        self._last_sent_mails: deque[SentEmail] = deque()
```
AureliaDolo, line 37:
```suggestion
        # should be sorted by timestamp from older to more recent
        self._last_sent_mails: deque[SentEmail] = deque()
```

server/parsec/email_rate_limit.py
```diff
@@ -0,0 +1,96 @@
...
+                if m.recipient == recipient or m.client_ip_address == client_ip_address
+            ]
+
+        print("===", now)
+        print(
+            "=======>",
+            client_ip_address,
+            recipient.str,
+            [(x.client_ip_address, x.recipient.str, now - x.timestamp) for x in relevant_emails],
+        )
```
FirelightFlagboy: **todo:** Debug print to remove or convert to log

AureliaDolo: Forgotten print ?

server/parsec/cli/options.py
touilleMan: Yes, but not this one: in the rest of the documentations we only use a `\b` at the beginning of the comment but not at the end, so I thing it's better to stay consistent

server/parsec/cli/run.py
```diff
@@ -440,7 +470,7 @@ def run_cmd(
     organization_initial_tos: dict[TosLocale, TosUrl] | None,
     organization_initial_allowed_client_agent: AllowedClientAgent,
     organization_initial_account_vault_strategy: AccountVaultStrategy,
-    validation_email_cooldown_delay: int,
+    validation_email_rate_limit: tuple[int, int],
```
AureliaDolo, line 456:
```suggestion
    # (cooldown in seconds, max number of email per hour)
    validation_email_rate_limit: tuple[int, int],
```

AureliaDolo (approved):

server/parsec/cli/run.py
touilleMan: Good point :+1: done.

server/parsec/email_rate_limit.py
touilleMan: :+1: done.

touilleMan: :+1: done.

server/parsec/email_rate_limit.py
touilleMan: :+1: done.

server/parsec/cli/run.py
touilleMan: :+1: done.

PR #10692, opened 2025-07-04
Expose `get_bootstrap_date`.

### LINKED ISSUE #8414: Expose org bootstrap date to the bindings
Add a function `getOrganizationBootstrapDate` to the bindings.

### Context

This is needed to fix #8405.

### Analysis

Note that there are two dates related to organization creation:

- Org _creation_ date (before bootstrap): this will need a server request as the information is not present locally. This could be added as a field to `OrganizationInfo` returned by the already existing function `clientOrganizationInfo()`, but **this is out of scope for this issue**.
- Org _bootstrap_ date: easy to obtain locally since the info can be obtained from local certificates.

So this issue is about adding a new function to the bindings to obtain the **organization bootstrap date**.

AureliaDolo:
Fix #8414

### DIFF
--- a/libparsec/crates/client/src/client/organization_info.rs
+++ b/libparsec/crates/client/src/client/organization_info.rs
@@ -2,7 +2,9 @@
 
 use libparsec_client_connection::ConnectionError;
 use libparsec_protocol::authenticated_cmds::latest::organization_info::{Rep, Req};
-use libparsec_types::anyhow;
+use libparsec_types::{anyhow, DateTime};
+
+use crate::{CertifPollServerError, CertifStoreError, InvalidCertificateError};
 
 #[derive(Debug, PartialEq, Eq)]
 pub struct OrganizationInfo {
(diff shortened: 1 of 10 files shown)


libparsec/crates/client/src/client/organization_info.rs
```diff
@@ -34,3 +36,35 @@ pub(super) async fn organization_info(
...
+    client: &super::Client,
+) -> Result<DateTime, ClientGetOrganizationBootstrapDateError> {
+    Ok(client
+        .certificates_ops
+        .list_users(false, None, None)
+        .await?
+        .into_iter()
+        .find(|info| info.created_by.is_none()) // User was created by root verify key
+        .ok_or(ClientGetOrganizationBootstrapDateError::BootstrapDateNotFound)?
+        .created_on)
```
FirelightFlagboy: **nitpick:** use `.map` so you wont need the `?` and wrapping around `Ok()`

FirelightFlagboy (approved):

libparsec/crates/client/src/client/organization_info.rs
```diff
@@ -34,3 +36,35 @@ pub(super) async fn organization_info(
...
+            CertifStoreError::Stopped => Self::Stopped,
+            CertifStoreError::Internal(err) => err.into(),
+        }
+    }
+}
+
+pub(super) async fn get_organization_bootstrap_date(
+    client: &super::Client,
+) -> Result<DateTime, ClientGetOrganizationBootstrapDateError> {
+    client
```
mmmarcos: It may seem obvious now, but I would describe the strategy to get bootstrap date
```suggestion
    // The organization bootstrap date coincides with the date the first user was created. 
    // Since the first user is the only one not invited by another user, the strategy here is to 
    // look for this user and get its creation date.
    client
```

libparsec/crates/client/src/client/organization_info.rs
```diff
@@ -34,3 +36,35 @@ pub(super) async fn organization_info(
...
     }
 }
+
+#[derive(Debug, thiserror::Error)]
+pub enum ClientGetOrganizationBootstrapDateError {
+    #[error(transparent)]
+    Internal(#[from] anyhow::Error),
+    #[error("Component has stopped")]
+    Stopped,
+    #[error("Bootstrap date not found")]
```
mmmarcos, line 46: Can this ever happened? If so, shouldn't we add a test covering this case?

AureliaDolo: It can happen if we don't have the certificates yet

libparsec/crates/client/src/client/organization_info.rs
```diff
@@ -34,3 +36,75 @@ pub(super) async fn organization_info(
...
+            .list_users(false, None, None)
+            .await?
+    } else {
+        users
+    };
+
+    users
+        .into_iter()
+        .find(|info| info.created_by.is_none()) // User was created by root verify key)
+        .ok_or(ClientGetOrganizationBootstrapDateError::BootstrapDateNotFound)
```
touilleMan, line 108: We should have a loop to poll forward r certificates until the user certif is available

This would allow to remove this error

FirelightFlagboy: Done above: https://github.com/Scille/parsec-cloud/pull/10692/files#diff-bffbb6da9df29c83b4f017fa002b35bf0b7b02a9c342d441f0610de4d6d5eaf5R91-R103

But we does not loop, if we loop and the user is offline the function will never return

PR #10706, opened 2025-07-07
[MS] Use the bindings to get the org creation date

### LINKED ISSUE #8405: [🐛 | Bug]: Trial days remaining is incorrectly computed from the device creation date
### Parsec version tested on:

3.0.0-rc.7

### Platforms tested on:

Windows

### Bug description:

That causes incorrect values to be displayed for a given user if that user is invited after the first day.

### Relevant output:

_No response_

Max-7:
Closes #8405 

The detection of a trial server is done via an environment variable, so if you have devices, it's possible to edit `.env.development.local` to change the trial server (for example, `PARSEC_APP_TRIAL_SERVERS="localhost:6770?no_ssl=true;localhost:6770"` will detect the local server as a trial server).

We can also modify the function that returns the creation date. This bypasses the bindings, so it's not a complete test, but it lets you see how the GUI reacts.
```
diff --git a/client/src/parsec/organization.ts b/client/src/parsec/organization.ts
index ac79f5cf1..0de288128 100644
--- a/client/src/parsec/organization.ts
+++ b/client/src/parsec/organization.ts
@@ -169,7 +169,7 @@ export async function getOrganizationCreationDate(
   }
   const result = await libparsec.clientGetOrganizationBootstrapDate(handle);
   if (result.ok) {
-    result.value = DateTime.fromSeconds(result.value as any as number);
+    result.value =  DateTime.now().minus({ days: 4 }); //DateTime.fromSeconds(result.value as any as number);
   }
   return result;
 }
```
for example, to set the creation date to 4 days in the past.

### DIFF
--- a/client/src/common/organization.ts
+++ b/client/src/common/organization.ts
@@ -42,6 +42,5 @@ export function formatExpirationTime(duration: Duration): Translatable {
 export function isTrialOrganizationDevice(device: AvailableDevice): boolean {
   const url = new URL(device.serverUrl);
   const serverType = getServerTypeFromHost(url.hostname, url.port.length > 0 ? parseInt(url.port) : undefined);
-
   return serverType === ServerType.Trial;
 }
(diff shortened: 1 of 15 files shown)


NicoTuxx (approved):
LGTM

PR #10720, opened 2025-07-09
[MS] Use env.BASE_URL in resourcesManager.ts

Ironicbay:


client/src/services/resourcesManager.ts
```diff
@@ -71,7 +71,7 @@ class WebResourcesProvider implements IResourcesProvider {
 
   async load(res: Resources): Promise<NonNullable<unknown>> {
     const axiosInstance = axios.create({
-      baseURL: this.path,
+      baseURL: import.meta.env.BASE_URL ?? this.path,
```
Max-7: Nop, you still need to use `this.path`. Should look something like
```
baseURL: `${import.meta.env.BASE_URL ?? ''}${this.path}`
```
Maybe also make sure that `this.path` starts with a `/` in the constructor.

Max-7 (approved):

PR #10750, opened 2025-07-17
Add is_file_content_local api.

### LINKED ISSUE #10326: [🚀 | Feature request]: Function in libparsec to know if a file is available

Seen with @touilleMan 

When opening a file while offline, we only know if it's available when trying to read.
- in the case of a file viewer, it's a bit late, we would prefer getting that info on stats() or on open()
- when using an external app, we have no control over what is going to be displayed (the app will fail reading the file and display whatever error it wants)

We should instead have a function to get the availibility of a file.

The function could retrieve the file manifest and check that all blocks id are available locally.

We considered adding this to the `entryStat` function, but it would be too costly. We don't need the information all the time, only when trying to open a file.

AureliaDolo:
Fix #10326

### DIFF
--- a/libparsec/crates/client/src/workspace/transactions/open_file.rs
+++ b/libparsec/crates/client/src/workspace/transactions/open_file.rs
@@ -10,7 +10,7 @@ use super::WorkspaceCreateFileError;
 use crate::{
     certif::{InvalidCertificateError, InvalidKeysBundleError, InvalidManifestError},
     workspace::{
-        store::{ForUpdateFileError, ResolvePathError},
+        store::{ForUpdateFileError, ReadChunkOrBlockLocalOnlyError, ResolvePathError},
         FileUpdater, OpenedFile, OpenedFileCursor, ReadMode, WorkspaceOps, WriteMode,
     },
 };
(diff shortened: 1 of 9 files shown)


libparsec/crates/client/src/workspace/transactions/open_file.rs
```diff
@@ -50,6 +50,88 @@ impl OpenOptions {
...
+    InvalidManifest(#[from] Box<InvalidManifestError>),
+
+    // specific errors
+    #[error("Path resolves to a folder, not a file")]
+    NotAFile,
+}
+
+/// To check if all blocks are locally present
+/// to avoid unexpected errors when trying to
+/// open offline a file with missing blocks
```
mmmarcos: A couple of suggestion based on rustdoc book on [Documenting components](https://doc.rust-lang.org/rustdoc/how-to-write-documentation.html#documenting-components):

```suggestion
/// Checks if all blocks are locally present for the given file `path`
///
/// This check allows to avoid unexpected errors when trying to open a file with 
/// missing blocks while being offline.
///
/// No check is performed if `path` is a folder, instead an error is returned.
```

libparsec/crates/client/tests/unit/workspace/open_file.rs
mmmarcos, line 1: If I am not mistaken, current tests are:
- existing file available locally --> `Ok(true)`
- new (empty?) file available locally --> `Ok(true)`
- non existing file --> `EntryNotFound`

We should add a test for the following cases:
- folder --> `NotAFile`
- existing (or new) file, not present locally --> `Ok(false)`

The first one should be rather straightforward, but I am not sure about the second one.

mmmarcos (approved):

PR #10783, opened 2025-07-21
Improve mocked email display on server

touilleMan:
When email sending is mocked, the server only print the plain text part of the email (currently the whole mail is printed which is pretty big since it also includes all the CSS and alternative mime stuff)


server/parsec/components/email.py
```diff
@@ -66,7 +67,18 @@ def _smtp_send_email(
 def _mocked_send_email(
     email_config: MockedEmailConfig, to_addr: EmailAddress, message: Message
 ) -> None | SendEmailBadOutcome:
-    message_body = message.as_string()
+    # Try to only display the plain text part for brevity...
+    for part in message.walk():
+        if (
+            part.get("Content-Type") == 'text/plain; charset="utf-8"'
```
FirelightFlagboy: **todo:** You are missing part where it's encoded in `latin-ascii` are we have changed how jinja does it?

touilleMan: This is totally empirical code: I've check what is the format of mailed returned when inviting a new user and consider the other emails should work similarly.

Since it's only test code, I'd rather keep thing simple and wait until this code is not enough to improve it (or maybe to fix encoding the incompatible mailing code instead)

touilleMan: You were right, `us-ascii` is needed since it seems automatically used by Python whenever the text string can be encoded this way ;-)

PR #10786, opened 2025-07-21
ci(q&a): Chech newsfragment now take GH token

FirelightFlagboy:
During certain run of `check_newsfragments` in the CI, we had some `rate limited exceeded`. We get those errors because we make anonymous request, adding `Authorization` header to each HTTP request with a valid token make them authenticated with a much greater rate limit.

I've also added `X-GitHub-Api-Version` header to make the script more robust in case of API version bump on GitHub side.

### DIFF
--- a/.github/scripts/check_newsfragments.py
+++ b/.github/scripts/check_newsfragments.py
@@ -10,6 +10,7 @@ directory are valid:
 
 import argparse
 import json
+import os
 from concurrent.futures import ThreadPoolExecutor
 from pathlib import Path
 from typing import Optional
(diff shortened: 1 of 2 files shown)


FirelightFlagboy:
CF failure on the following run: https://github.com/Scille/parsec-cloud/actions/runs/16416228506/job/46382717622?pr=10718

FirelightFlagboy:
You can test the script with

```shell
GITHUB_TOKEN=$(gh auth token) python .github/scripts/check_newsfragments.py
```

mmmarcos (approved):

PR #10787, opened 2025-07-21
refactor(libparsec): Simplify default value for log level

FirelightFlagboy:
`log::Level` provide the method `as_str` that we can use instead of having to use `String`


libparsec/src/lib.rs
```diff
@@ -114,35 +114,29 @@ pub fn libparsec_init_set_on_event_callback(
...
 #[cfg(not(target_arch = "wasm32"))]
 fn init_logger(config: &ClientConfig) {
     let log_env = env_logger::Env::default()
         .filter_or(
             "PARSEC_RUST_LOG",
-            config
-                .log_level
-                .map(|lvl| lvl.to_string())
-                .unwrap_or_else(|| "info".to_string()),
+            config.log_level.map(|lvl| lvl.as_str()).unwrap_or("info"),
```
touilleMan: :+1:

libparsec/crates/crypto/src/sodiumoxide/mod.rs
```diff
@@ -17,3 +17,7 @@ pub use secret::*;
 pub use sequester::*;
 pub use sign::*;
 pub(crate) use utils::*;
+
+pub fn init() -> anyhow::Result<()> {
+    sodiumoxide::init().map_err(|_| anyhow::anyhow!("Failed to init sodiumoxide"))
```
touilleMan: nitpick: I would rather use `map_err(|()| anyhow::...`, the `()` instead of `_` makes it obvious that `sodiumoxide::init()` doesn't return a precise error (since in that case it that would be convenient to add it to the returned error)

touilleMan: Also, since we just panic in `libparsec_init()` instead of returning the error through the bindings, we should just move the panic here (given it simplifies the code, and sodiumoxide is never supposed to not be able to initialize)

FirelightFlagboy: done

AureliaDolo (approved):

FirelightFlagboy:
Superseded by #10809

Issue #10758, opened 2025-07-18
Adapt open strategy for file edition


mmmarcos:
Once Parsec is able to edit files (#10757), we should consider the following:

- There should be two actions (View, Edit) accessible via buttons / contextual menu
- Double-click should choose a default action (View ? Edit ?), possibly with fallback (CryptPad server/editor may not be available)
- Should this depend on the file type? 
- Should this depend on the user role? (i.e `Reader` --> always `View`?)
- If no viewer/editor is available for file type, should propose download?
- Web vs Electron?
- Option to skip or configure in Settings?
- Check if CryptPad have a read-only mode to open files

## Analysis

After reviewing a few existing solutions, here are some propositions:

1. 👑
- Single click to select file, eventually display **preview** when file is selected
- **Double click** defaults to open **Viewer**
- Option visible to open **Editor** from **Viewer**
- **Editor** available in right-click contextual menu on file
- If neither is available, suggest downloading the file
- On electron, suggest to open file with another software / on device

2.
- Single click to select, eventually display **preview** when file is selected
- Double click defaults to opens **Editor** if available and has write access
- Double click opens **Viewer** if available and has read-only access or **Editor** is not compatible
- Option to open **Viewer** from right click
- Suggest download if neither is available

3.
- Single click to select, eventually display **preview** when file is selected
- Double click opens a **dialog** asking if user wants open in **Viewer** or **Editor** (if available)
- Both options available on right click
- Default option could be set in settings

4.
- Double click defaults to whatever option the user sets (via settings or dialog at first open)
- Visible `View` and `Edit` icons in `FileListItem` and `FileGridItem` that opens either with a single click on the icon

mmmarcos:
Let's stick with option 1 (dbl click = view) for the moment. 

Option 2 (dbl click = edit) may bring some issues: opening a file for editing may take some time and does not depend on Parsec. A slow cryptpad server will have an impact on Parsec user experience. Moreover, if "edit" is the default action, users will never select "view" thus wasting resources (session creation, key management, etc.) unnecessarily every time it.

Options 3 (dialog) may unnecessary interrupt user workflow, so I would say no.

Option 4 (settings) IMO will be nice to have some user feedback before making that configurable.

NicoTuxx:
A basic approach of the desired strategy is done in #10795.
What's missing will be done separetely in other issues: #10820 #10891 #10890

PR #10790, opened 2025-07-21
Add RFC 1020 for collaborative edition

### LINKED ISSUE #11010: RFC CryptPad integration
RFC to describe integration and identify to do list.

mmmarcos:
Close #11010


docs/rfcs/1020-collaborative-edition.md
```diff
@@ -0,0 +1,140 @@
...
+
+# Collaborative edition via CryptPad
+
+## Overview
+
+This RFC describes how to enable collaborative edition in Parsec via [CryptPad](https://cryptpad.org/).
+
+## Background & Motivation
+
+CryptPad is fully end-to-end encrypted and doesn't know anything about Parsec users, so it cannot manage
```
FirelightFlagboy: **issue:** I though that the encryption was a setting to that can be disabled

docs/rfcs/1020-collaborative-edition.md
```diff
@@ -0,0 +1,140 @@
...
+access-rights with user sessions. Editing sessions are based exclusively on *session keys*: either a user knows
+the session key and can access the document, or don't.
+
+To manage the session keys, CryptPad API provides 2 alternatives:
+
+1. Keys generated by CryptPad: Parsec is informed via `onNewKey` handler to store and distribute the new key.
+2. Keys generated by Parsec: CryptPad won't perform any check and will trust Parsec to rotate keys correctly.
+
+Only the latter is acceptable for Parsec, so this RFC focuses on option 2.
```
FirelightFlagboy: **polish:** I prefer to be explicit about the choice of Parsec over Cryptpad to generate the key:

It's Parsec that will handle key rotation and most important the RBAC

docs/rfcs/1020-collaborative-edition.md
```diff
@@ -0,0 +1,140 @@
...
+When session key changes, CryptPad won't destroy the session instantly but the connected users won't be able to save further changes.
+
+CryptPad only closes the session when all users quit. There is a configurable timeout, before actually closing the
+session so it's possible to reconnect to the session for a user having the session key.
+
+## Goals and Non-Goals
+
+The goal is to describe:
+
+- how to correctly **store**, **share** and **rotate the editing session keys** in Parsec.
```
FirelightFlagboy: **thought:** We want to store the key in an ephemeral way since the session come and go

docs/rfcs/1020-collaborative-edition.md
```diff
@@ -0,0 +1,140 @@
...
+However, note that Parsec will continue to perform key rotations on the realm key as usual (e.g. if a user is revoked).
+This should not be an issue.
+
+##### Key rotation - Changes related to a user in the session
+
+In this simple scenario, Alice and Bob collaborate to edit a document. The session key needs to be rotated if:
+
+1. Alice (or Bob) is revoked from the organization.
+2. Alice (or Bob) is "unshared" from the workspace (its role is *removed*)
+3. Alice (or Bob) role changes from `Owner` or `Manager` or `Contributor` to `Reader` (or vice-versa)
```
FirelightFlagboy: **question:** Does CryptPad provide a way to join a session in read-only mode?

docs/rfcs/1020-collaborative-edition.md
```diff
@@ -0,0 +1,140 @@
...
+
+#### Key rotation - Share new key
+
+CryptPad does not provides a mechanism to force session closing. This means that in order to share the new key:
+
+- Document saves triggered from users within the existing session **must be disabled**
+- All users are informed that a "reload" is needed
+- All users are disconnected from the session
+  - This will need to be managed by Parsec GUI since CryptPad does not provides a mechanism for this
+- CryptPad closes the session after a timeout
```
FirelightFlagboy: **question:** Is it require to exit the editor? can it be possible to keep it open and just use the new key?

docs/rfcs/1020-collaborative-edition.md
```diff
@@ -0,0 +1,140 @@
...
+- Users are informed of the new key
+- A new session is started with the new session key, and users re-join the new session.
+
+We need to ensure that this process is smooth for the Parsec user.
+
+#### Key storage
+
+> TODO: it seems that Parsec does not need to store the session keys as any user can derive it from the realm key.
```
FirelightFlagboy: **todo (blocking):** This document does not explain how the key is generated, that's the most important piece.

Depending how it's done we would see if we need to store it or not (I we need it, I think that would be an ephemeral storage)

docs/rfcs/1020-collaborative-edition.md
```diff
@@ -0,0 +1,140 @@
...
+
+## Security/Privacy/Compliance
+
+> What security/privacy/compliance aspects should be considered?
+> If you're not certain, never assume there aren’t any. Always talk to the security team.
+
+## Risks
+
+- CryptPad instances do not forbid users to connect directly to them. This needs further analysis
+  since security measures applied by Parsec could be easily by-passed.
```
FirelightFlagboy: **thought:** I think of a potential user brute forcing the key, we may need to use service like `fail2ban` to jail an user

docs/rfcs/1020-collaborative-edition.md
```diff
@@ -0,0 +1,140 @@
...
+
+We need to ensure that this process is smooth for the Parsec user.
+
+#### Key storage
+
+> TODO: it seems that Parsec does not need to store the session keys as any user can derive it from the realm key.
+
+### Document Save
+
+TODO
```
FirelightFlagboy: **note:** Some point to take care

- Limit the explosion version numbers of a document (to limit consumption of storage)
- Who save the document? What happen when that user exit the session?

FirelightFlagboy (changes requested):

docs/rfcs/1020-collaborative-edition.md
mmmarcos: I aimed to be explicit. If you think is not explicit enough, can you provide a suggestion on how you would rephrase it?

docs/rfcs/1020-collaborative-edition.md
mmmarcos: We don't know. That should be investigated as part of #10758

docs/rfcs/1020-collaborative-edition.md
mmmarcos: If you mean the CryptPad editor I don't think this is possible because CryptPad session needs to be fully closed.

That said, Parsec editor (around CryptPad iframe) would certainly remain open, so there maybe a way to "hide" the switch to new session from the user.

docs/rfcs/1020-collaborative-edition.md
mmmarcos: Yep, I wonder if we can always derive the same session key from the same realm key, so no storage will be needed.

Anyway, I am leaving this part for the experts available: @touilleMan and you ;)

docs/rfcs/1020-collaborative-edition.md
mmmarcos: From what I understand, a save can be triggered:
- by any user in the session, manually
- by auto-save at configurable interval
  - This is managed by CryptPad, it detects the last user who have made a change and performs auto-save only from that user, unless the save fail and it authorizes another user to save (see below)
- when the last user exit the session, if unsaved changes, CryptPad asks to save before exit.

According to CryptPad developers:

> The user who triggers the save would normally be the last user to have made a change to the document when the auto-save timer allows saving. There is an internal system that prevents several users from triggering save at the same time, and this system allows another user to be authorized to save if the action has failed with the first.

docs/rfcs/1020-collaborative-edition.md
```diff
@@ -0,0 +1,140 @@
...
+
+CryptPad only provides a collaborative session that multiple users can join to edit the same document.
+
+The data is end-to-end encrypted on the CryptPad side but doesn't persist. The integrating service (Parsec) is expected
+to manage storage and access rights / sharing keys.
+
+Editing sessions are identified via a *session key*. Users who want to edit the same document would use the same
+session key and they will end up in the same session (the first user will create the session, the others will join).
+
+Session keys must be rotated when when someone's access to a document has been revoked. This will guarantee they won't
```
FirelightFlagboy: **typo:** 
```suggestion
Session keys must be rotated when someone's access to a document has been revoked. This will guarantee they won't
```

docs/rfcs/1020-collaborative-edition.md
```diff
@@ -0,0 +1,140 @@
...
+be able to use the session key to extract the document content from CryptPad directly.
```
FirelightFlagboy: **thought:** When a user is revoked, we consider it as access to every document up to the revocation moment.
Here we need to rotate the key because we do not want the user to access up to date / new documents

docs/rfcs/1020-collaborative-edition.md
```diff
@@ -0,0 +1,168 @@
...
+
+- Randomly generated when a user is the first to open a CryptPad session.
+
+  This system provides random session key that are unique to each session,
+  it cannot be determined by other users so we need to add a system to share the keys among them.
+
+  Given that we need to shared something only to a subgroup of the workspace, we need to do it securely.
+  So we need to encrypt it: If encryption secret is the realm key, it would require the server to perform some access check on the data.
+  Or we use a dedicated secret, but we fall back to the first approach.
+
```
AureliaDolo: Is one approach favored over the other ?

docs/rfcs/1020-collaborative-edition.md
```diff
@@ -0,0 +1,168 @@
...
+##### Key rotation - Changes not related to the users in a session
+
+Following the previous scenario, if the changes are not related to Alice or Bob (e.g. Mallory is revoked or unshared),
+a key rotation may still be needed. This is for the same reasons that the realm key is rotated: Mallory **had** access
+to the session key, which mean that it would continue having access if no rotation is performed.
+
+Parsec GUI should carefully present this information as it may seem strange for Alice and Bob to be forced to reload
+the session due to a key rotation.
+
```
AureliaDolo: example : "A user has no longer access to this file. Please reload the session to ensure there is no unauthorized access."
"Un utilisateur n'a plus accès à ce fichier. Veuillez rafraichir la session pour éviter un accès non autorisé"

And maybe include a link to read more about why exactly the reload is needed

docs/rfcs/1020-collaborative-edition.md
```diff
@@ -0,0 +1,168 @@
...
+#### Key rotation - Share new key
+
+CryptPad does not provides a mechanism to force session closing. This means that in order to share the new key:
+
+- Document saves triggered from users within the existing session **must be disabled**
+- All users are informed that a "reload" is needed
```
AureliaDolo: It would be nice if the user would be able to for example finish their sentence before reloading the session. But if we don't want to risk the user postponing the reload we could have a countdown (e.g. 1 min) before an automatic reload.

docs/rfcs/1020-collaborative-edition.md
AureliaDolo: Maybe it would be nice to differentiate an autosave (and not increment the parsec file version) from a save triggered by a user (manually or by being the last to exit a session).

docs/rfcs/1020-collaborative-edition.md
```diff
@@ -0,0 +1,168 @@
...
+## Security/Privacy/Compliance
+
+> What security/privacy/compliance aspects should be considered?
+> If you're not certain, never assume there aren’t any. Always talk to the security team.
+
+## Risks
+
+- CryptPad instances do not forbid users to connect directly to them. This needs further analysis
+  since security measures applied by Parsec could be easily by-passed.
+- A user saving a document may not be the one who made the modification.
```
AureliaDolo: This is already the case inside crypt pad ? Can this be mitigated by parsec ?

docs/rfcs/1020-collaborative-edition.md
```diff
@@ -0,0 +1,189 @@
...
+##### Shared secret Pros&Cons
+
+Pros:
+
+- Known by other writers ahead of time
+
+Cons:
+
+- Cannot be extended to include a reader session key later on
+- Need to add a now key rotation system
```
AureliaDolo:
```suggestion
- Need to add a new key rotation system
```

FirelightFlagboy: done

docs/rfcs/1020-collaborative-edition.md
```diff
@@ -0,0 +1,211 @@
...
+1. Keys generated by CryptPad: Parsec is informed via `onNewKey` handler to store and distribute the new key.
+2. Keys generated by Parsec: CryptPad won't perform any check and will trust Parsec to rotate keys correctly.
+
+Only the latter is acceptable for Parsec (as CryptPad is not connected to the RBAC system of Parsec), so this RFC focuses on option 2.
+
+### About CryptPad sessions
+
+CryptPad only provides a collaborative session that multiple users can join to edit the same document.
+
+The data is end-to-end encrypted on the CryptPad side but doesn't persist. The integrating service (Parsec) is expected
```
AureliaDolo:
```suggestion
The data is end-to-end encrypted on CryptPad's side but doesn't persist. The integrating service (Parsec) is expected
```

docs/rfcs/1020-collaborative-edition.md
```diff
@@ -0,0 +1,211 @@
...
+
+## Goals and Non-Goals
+
+The goal is to describe:
+
+- how to correctly **store**, **share** and **rotate the editing session keys** in Parsec.
+- how to manage document saves
+
+> [!IMPORTANT]
+> We only cater to users who have write access as do lack experience on how CryptPad work with read-only access and how we can generate a session key with that restriction. (See open questions)
```
AureliaDolo:
```suggestion
> We only cater to users who have write access as we do lack experience on how CryptPad work with read-only access and how we can generate a session key with that restriction. (See open questions)
```

docs/rfcs/1020-collaborative-edition.md
```diff
@@ -0,0 +1,211 @@
...
+  type: HashMap<UserID, Blob>
+```
+
+- **editing_session_key**: Correspond to the current editing session_key to join the CryptPad session, but it's encrypted with the public key of all writers identified by `UserID`.
+
+##### `EditingManifest` & Key rotation
+
+When the writers list change (added or removed) or the creation of the manifest is `< now() - CRYPTPAD_SESSION_TIMEOUT * 0.9`, we need to rotate the session key.
+
+> `CRYPTPAD_SESSION_TIMEOUT` is configured on the CryptPad server side, it defines how long a session is keep alive once the last user quit the session.
```
AureliaDolo:
```suggestion
> `CRYPTPAD_SESSION_TIMEOUT` is configured on the CryptPad server side, it defines how long a session is kept alive once the last user quit the session.
```

docs/rfcs/1020-collaborative-edition.md
```diff
@@ -0,0 +1,211 @@
...
+> That value should be know by the client (the value should be provided by the server as it's can change depending on the CryptPad server)
```
AureliaDolo:
```suggestion
> That value should be know by the client (the value should be provided by the server as it can change depending on the CryptPad server)
```

docs/rfcs/1020-collaborative-edition.md
```diff
@@ -0,0 +1,211 @@
...
+
+
+### Document Save
+
+CryptPad dispatch `onSave` event, it's apparently the last user who made a change that will receive the event.
+
+> We consider that only a single user will receive such event, and we will not have to handle concurrent save.
+
+During the `onSave` event, CryptPad provide the full document up-to-date to save.
+For that reason would be best to introduce a differential saving (e.i: compare old vs new to only save changed block, or not save at all if identical)
```
AureliaDolo:
```suggestion
For that reason would be best to introduce a differential saving (i.e.: compare old vs new to only save changed block, or not save at all if identical)
```

docs/rfcs/1020-collaborative-edition.md
```diff
@@ -0,0 +1,211 @@
...
+
+> The comparison could be on the whole file or block per block if we want more granularity on the uploaded blocks.
+
+In the bindings, we likely want to expose a function to do that:
+
+```rust
+/// Compare `new_content` with current content of file at `path` and only register the modified block if any.
+/// It always compare against the latest known version of the file.
```
AureliaDolo:
```suggestion
/// It always compares against the latest known version of the file.
```

docs/rfcs/1020-collaborative-edition.md
```diff
@@ -0,0 +1,211 @@
...
+```
+
+## Security/Privacy/Compliance
+
+> What security/privacy/compliance aspects should be considered?
+> If you're not certain, never assume there aren’t any. Always talk to the security team.
+
+## Risks
+
+- CryptPad provide a system to share a document and the deployed server MUST have that feature disabled
```
AureliaDolo:
```suggestion
- CryptPad provides a system to share a document and the deployed server MUST have that feature disabled
```

mmmarcos:
Collaborative edition would be addressed by an upcoming RFC (1031?), see `collaborative-editics.md` (WIP) in the [editics-crytpad-oo](https://github.com/Scille/parsec-cloud/tree/editics-crytpad-oo/docs/rfcs) branch.

PR #10791, opened 2025-07-21
Server postgresql org test no drop

touilleMan:
This saves ~10s  (~15%)  on my machine:
```shell
$ git checkout master
$ py.test --postgresql tests/
1096 passed, 62 skipped, 90 xfailed, 1 xpassed in 58.98s
$ git checkout server-postgresql-org-test-no-drop
$ py.test --postgresql tests/
1096 passed, 62 skipped, 90 xfailed, 1 xpassed in 69.80s
```


server/tests/api_v5/anonymous_account/test_account_create_proceed.py
```diff
@@ -203,7 +203,7 @@ async def test_anonymous_account_account_create_proceed_send_validation_email_re
             ),
             auth_method_mac_key=SecretKey.generate(),
             vault_key_access=b"vault_key_access",
-            auth_method_id=bob_account.auth_method_id,
+            auth_method_id=AccountAuthMethodID.new(),
```
FirelightFlagboy: **question:** This is different from the rest of the code, why we need a new ID here?

server/tests/realm_export/test_export.py
FirelightFlagboy, line 1: **polish:** If we no longer use the constant `ORG_ID`, remove them

FirelightFlagboy:
The CI also have a -10sec on the Postgre test but that may be a fluke

server/tests/api_v5/anonymous_account/test_account_create_proceed.py
touilleMan: Good catch ! I've corrected to use a pre-generated dummy ID :+1:

server/tests/realm_export/test_export.py
touilleMan: done. :+1:

PR #10792, opened 2025-07-21
[MS] Implements Cryptpad API

NicoTuxx:
## Checklist

- [ ] Describe your changes
   
If this PR is related to GUI:

If this PR affects the user:


client/README.md
```diff
@@ -137,3 +137,5 @@ ionic cap add ios
 | `PARSEC_APP_ENABLE_ACCOUNT`             | `boolean`                 | Enable Parsec Account                                                                   |                                                                                                                         |
 | `PARSEC_APP_ENABLE_ACCOUNT_AUTO_LOGIN`  | `boolean`                 | Enable Account auto login                                                               | Only for development purposes! Only works if `PARSEC_APP_ENABLE_ACCOUNT` and `PARSEC_APP_MOCK_ACCOUNT` are set to true. |
 | `PARSEC_APP_ENABLE_CUSTOM_BRANDING`     | `boolean`                 | Enable the custom branding                                                              |                                                                                                                         |
+| `PARSEC_APP_ENABLE_CRYPTPAD_EDITION`    | `boolean`                 | Enable Cryptpad edition poc                                                                    |                                                                                                 |
```
Max-7: `editing`, not `edition`. But I would just use `PARSEC_APP_ENABLE_CRYPTAD`, I think it's clear enough and it keeps the variable name short.

client/README.md
```diff
@@ -137,3 +137,5 @@ ionic cap add ios
...
+| `PARSEC_APP_CRYPTPAD_EDITION_SERVER`    | `url`                 | Cryptpad edition server to use                                                                    |                                                                                                 |
```
Max-7: Would call that `PARSEC_APP_CRYPT_DEFAULT_SERVER` instead, since ultimately we'll want the server to change depending on the parsec server or the org.

client/src/services/cryptpad.ts
```diff
@@ -0,0 +1,160 @@
...
+    default:
+      return CryptpadDocumentType.Unsupported;
+  }
+}
+
+export function isEnabledCryptpadDocumentType(extension: string): boolean {
+  return ENABLED_DOCUMENT_TYPES.includes(getDocumentTypeFromExtension(extension));
+}
+
+function isFirefox(): boolean {
```
Max-7: Already exists: https://github.com/Scille/parsec-cloud/blob/a0857a3c42b0522f43b524e6ee4bfd6974f18eae/client/src/parsec/environment.ts#L51

client/src/services/cryptpad.ts
```diff
@@ -0,0 +1,160 @@
...
+  CryptpadDocumentType.Sheet,
+  CryptpadDocumentType.Doc,
+  CryptpadDocumentType.Presentation,
+  CryptpadDocumentType.Code,
+];
+
+export interface CryptpadConfig {
+  document: {
+    url: string;
+    fileType: string;
```
Max-7, line 25: What's that `fileType`?

client/src/services/cryptpad.ts
```diff
@@ -0,0 +1,160 @@
...
+  };
+}
+
+export class Cryptpad {
+  private scriptId = 'cryptpad-api-js';
+  private containerElement: HTMLElement;
+  private script: HTMLScriptElement | null = null;
+  private scriptLoaded = false;
+
+  constructor(containerElement: HTMLElement) {
```
Max-7: Should take the server as a parameter.

client/src/services/cryptpad.ts
```diff
@@ -0,0 +1,160 @@
...
+    if (!this.script) {
+      this.script = document.createElement('script');
+      this.script.id = this.scriptId;
+      this.script.src = `${this.getServer()}/cryptpad-api.js`;
+      this.script.async = true;
+      document.head.appendChild(this.script);
+    }
+  }
+
+  isEnabled(): boolean {
```
Max-7: I would remove `isEnabled` and `getServer` from the class. Controlling whether cryptpad is enabled and what server to use should be done outside, this class just does its job and instanciates cryptpad.

client/src/services/cryptpad.ts
```diff
@@ -0,0 +1,160 @@
...
+  private scriptLoaded = false;
+
+  constructor(containerElement: HTMLElement) {
+    this.containerElement = containerElement;
+    if (!this.isEnabled()) {
+      return;
+    }
+    this.script = document.getElementById(this.scriptId) as HTMLScriptElement | null;
+    if (!this.script) {
+      this.script = document.createElement('script');
```
Max-7: Type `this.script` as `HTMLScriptElement` and just throw an error if it can't be found/created. It'll make using the class easier (methods won't have to check if script is null) and we don't really want to have an instance of this class that's not properly initialized, it's better to abort right at the start.

client/src/services/cryptpad.ts
```diff
@@ -0,0 +1,160 @@
...
+          console.log('CryptPad API script loaded successfully.');
+        }
+        resolve();
+      };
+
+      // Add error handling for HTTPS/security issues
+      this.script!.onerror = (error): void => {
+        console.error('Failed to load CryptPad script:', error);
+        console.error('This might be due to HTTPS requirements. Check if CryptPad server requires secure context.');
+        resolve();
```
Max-7: Should use `reject` instead in case of an error.

client/src/services/cryptpad.ts
```diff
@@ -0,0 +1,160 @@
...
+  }
+
+  async init(): Promise<void> {
+    if (!this.script) return;
+    if (this.scriptLoaded) return;
+    await new Promise<void>((resolve) => {
+      this.script!.onload = (): void => {
+        if (typeof (window as any).CryptPadAPI === 'function') {
+          this.scriptLoaded = true;
+          console.log('CryptPad API script loaded successfully.');
```
Max-7: Use `window.electronAPI.log()` when possible, so it gets logged.

client/src/services/cryptpad.ts
```diff
@@ -0,0 +1,160 @@
...
+    if (!ENABLED_DOCUMENT_TYPES.includes(config.documentType)) {
+      console.warn(`CryptPad edition for document type ${config.documentType} is not enabled.`);
+      return;
+    }
+
+    if (!this.containerElement) {
+      console.warn('Container element is not initialized. Please call init() before open().');
+      return;
+    }
+    if (!this.scriptLoaded) {
```
Max-7, line 103: Just have a `isInit` variable that we can check, instead of having to check two.

client/src/services/cryptpad.ts
```diff
@@ -0,0 +1,160 @@
...
+
+  isEnabled(): boolean {
+    return Env.isCryptpadEnabled();
+  }
+
+  getServer(): string {
+    return Env.getCryptpadServer();
+  }
+
+  async init(): Promise<void> {
```
Max-7, line 67: This should return or throw some king of error if the initialization failed.

client/src/services/cryptpad.ts
```diff
@@ -0,0 +1,160 @@
...
+      console.warn('Container element is not initialized. Please call init() before open().');
+      return;
+    }
+    if (!this.scriptLoaded) {
+      console.warn('CryptPad API script is not loaded yet. Please wait for it to load before calling open().');
+      return;
+    }
+
+    // Firefox-specific key prefix for POC
+    if (config.document && config.document.key && isFirefox()) {
```
Max-7: Remove this.

client/src/services/cryptpad.ts
```diff
@@ -0,0 +1,160 @@
...
+
+export const ENABLED_DOCUMENT_TYPES = [
+  CryptpadDocumentType.Pad,
+  CryptpadDocumentType.Sheet,
+  CryptpadDocumentType.Doc,
+  CryptpadDocumentType.Presentation,
+  CryptpadDocumentType.Code,
+];
+
+export interface CryptpadConfig {
```
Max-7: This config is Cryptpad internal config, I'm not sure we should expose it directly, we can probably simplify the API.

client/src/services/cryptpad.ts
```diff
@@ -0,0 +1,160 @@
...
+      this.script!.onerror = (error): void => {
+        console.error('Failed to load CryptPad script:', error);
+        console.error('This might be due to HTTPS requirements. Check if CryptPad server requires secure context.');
+        resolve();
+      };
+    });
+  }
+
+  async open(config: CryptpadConfig): Promise<void> {
+    await this.init();
```
Max-7, line 92: Initialization should be made before, not from open.

client/README.md
NicoTuxx: done

client/README.md
NicoTuxx: was going to rename to EDITICS anyway.

client/src/services/cryptpad.ts
NicoTuxx: The document extension, used to convert the initial document provided and export to the correct format.

NicoTuxx: Since the JS script does not give us proper typing, this interface is here to cover that.

client/src/services/cryptpad.ts
NicoTuxx: deleted.

client/src/services/cryptpad.ts
NicoTuxx: deleted.

client/README.md
NicoTuxx: update: for now we decided in last meeting to use env variables to change the server so I'm renaming it.

client/src/services/cryptpad.ts
NicoTuxx: done

client/src/services/cryptpad.ts
NicoTuxx: reject does not seem to exist, are you sure about that?

client/src/services/cryptpad.ts
NicoTuxx: done

client/src/services/cryptpad.ts
```diff
@@ -0,0 +1,150 @@
...
+        if (typeof (window as any).CryptPadAPI === 'function') {
+          this.scriptLoaded = true;
+          console.log('CryptPad API script loaded successfully.');
+        }
+        resolve();
+      };
+
+      // Add error handling for HTTPS/security issues
+      this.script!.onerror = (error): void => {
+        console.error('Failed to load CryptPad script:', error);
```
NicoTuxx: @Max-7 how to use window.electronAPI.log on this kind of cases?

client/README.md
Max-7: We decided that the app should be able to handle multiple servers (one per org or one per server), so this would just be used as the default server if we have nothing else. Env variables are set at build time.

client/src/services/cryptpad.ts
Max-7:
```
new Promise((resolve, reject) => {
  // everything good
  resolve();
  // something bad happened
  reject(error);
})
```

So you can use the promise with
```
promise.then((result) => {
  // good
}).catch((error) => {
  // bad
})
```
or
```
try {
  await new Promise(...);
} catch (e: any) {

}
```

client/src/services/cryptpad.ts
Max-7: For normal prints, for debugging purposes for example, `console.log` would do the job.

In the case we want to log something, since it's serialized to the main process that will write to a file, it's better to unpack complex objects to basic types ourselves (it also adds a bit of security)

Since we don't know the error instance type:
``window.electronAPI.log('error', `Failed to load CryptPad script: ${error.toString()}`)``;

client/README.md
NicoTuxx: yup lgtm

client/src/services/cryptpad.ts
NicoTuxx: done

client/src/services/cryptpad.ts
NicoTuxx: lgtm

client/src/services/cryptpad.ts
NicoTuxx: done

client/src/services/cryptpad.ts
NicoTuxx: done

client/src/services/cryptpad.ts
NicoTuxx: done

client/src/services/cryptpad.ts
NicoTuxx: done

client/src/services/cryptpad.ts
```diff
@@ -0,0 +1,134 @@
...
+      document.head.appendChild(script);
+    }
+    this.script = script;
+  }
+
+  async init(): Promise<void> {
+    if (this.scriptLoaded) return;
+    await new Promise<void>((resolve, reject) => {
+      this.script.onload = (): void => {
+        if (typeof (window as any).CryptPadAPI === 'function') {
```
Max-7, line 71: What happens if that's not the case? Maybe just add a little log+call to reject() instead of resolve(), I don't think it'll ever happen but it will be painful to debug.

client/src/services/environment.ts
```diff
@@ -201,6 +201,23 @@ async function openTOS(tosLink: string): Promise<void> {
...
+
+const ENABLE_EDITICS_VARIABLE = 'PARSEC_APP_ENABLE_EDITICS';
+const CRYPTPAD_SERVER_VARIABLE = 'PARSEC_APP_DEFAULT_CRYPTPAD_SERVER';
+const DEFAULT_CRYPTPAD_SERVER = 'https://centakina.ddns.net';
+
+function isEditicsEnabled(): boolean {
+  return import.meta.env[ENABLE_EDITICS_VARIABLE] === 'true';
+}
+
+function getCryptpadServer(): string {
```
Max-7: `getCryptpadDefaultServer`, technically.

Max-7 (approved):

PR #10793, opened 2025-07-21
[MS] Electron security config for Cryptpad usage

NicoTuxx:


client/electron/src/setup.ts
```diff
@@ -551,15 +552,14 @@ export class ElectronCapacitorApp {
...
-            ? `default-src ${customScheme}://* 'unsafe-inline' devtools://* 'unsafe-eval' https://* 'unsafe-eval' data: blob:`
-            : `default-src ${customScheme}://* 'unsafe-inline' 'unsafe-eval' https://* 'unsafe-eval' data: blob:`,
-        ],
-      },
-    });
+    const responseHeaders = { ...details.responseHeaders };
+
+    responseHeaders['Content-Security-Policy'] = [[
+      `default-src ${customScheme}://* 'unsafe-inline' 'unsafe-eval' https://* data: blob:${electronIsDev ? ' devtools://*' : ''}`,
+      `frame-src ${customScheme}://* https://* data: blob:`,
```
Max-7: If I understand this correctly, this allows a script to connect to any server and use an iframe from any server, when we only want to allow those from a cryptpad server. We should probably investigate this a bit more, maybe make the list dynamic (since we can have multiple servers), maybe log when a server is added to the list, ...

NicoTuxx: done

Max-7 (approved):

Issue #10761, opened 2025-07-18
Properly manage CryptPad unavailability


mmmarcos:
Some cases to be managed:

- When feature flag is disabled we want to redirect the user somewhere else. The page &  route must not load.
=> done in #10795 

- CryptPad server not available: the CryptPad GUI (iframe) would not load, should be handled properly by Parsec GUI.
  - We could simply display an error, but if there will be no server for a particular deployment, it could be frustrating seeing the "Edit" action. Should we hide editing actions?
  => will be treated in #10894 #10895 #10896 
 
- CryptPad server temporarily offline: the CryptPad GUI (iframe) is loaded but the connection was lost, how to handle this case?
=> should be separated in a new issue (not top-priority)

Note that Parsec server maybe offline: this shouldn't be an issue since local saves can be made via libparsec. However if this is related to a connection issues, this would probably mean CryptPad server would also temporarily offline.

Issue #10760, opened 2025-07-18
Error management when using CryptPad API


mmmarcos:
CryptPad API errors should be properly managed by Parsec.

This is not the case in the POC https://github.com/Scille/parsec-cloud/pull/10518

NicoTuxx:
All known errors are treated, this issue is kept to investigate what Cryptpad could throw when loading / editiing a document.

PR #10794, opened 2025-07-21
[MS] Implements FileEditor page & routing

NicoTuxx:


client/src/components/files/handler/viewer/index.ts
Max-7, line 1: Why `handler/viewer` instead of just `viewer`? Everything related to the editor will be put into `handler` too? The name `handler` kinda sucks. I'm guessing you tried to find something neutral and `editor` and `viewer` where taken.
It's probably a mistake to separate both though, I think we should see the feature as `file editing` as a whole, it's just that some files are only read-only. Even if technically we currently have different pages, the more it's going to evolve, the more they will probably blend together.

client/src/router/types.ts
```diff
@@ -180,6 +181,11 @@ const routes: Array<RouteRecordRaw> = [
                     name: Routes.MyProfile,
                     component: () => import('@/views/users/MyProfilePage.vue'),
                   },
+                  {
```
Max-7: How are we treating the feature being deactivated? Do we allow the URL to be loaded but the page displays an error message? Or are we treating like that it doesn't exist and hiding it completely?
If it's the second one, this needs a bit more work to add checks and a redirect.

client/src/views/files/handler/FileHandler.vue
```diff
@@ -109,35 +108,37 @@ import {
...
+const { isHeaderVisible, toggleHeader: toggleMainHeader, showHeader, hideHeader } = useHeaderControl();
+const handlerReady = ref(false);
+
+const emits = defineEmits<{
+  (e: 'fileLoaded', contentInfo: FileContentInfo, fileInfo: DetectedFileType, fileInfoId: string): void;
+}>();
 
 const cancelRouteWatch = watchRoute(async () => {
-  if (!currentRouteIs(Routes.Viewer)) {
+  console.log('FileHandler: route changed, reloading file');
```
Max-7: Don't forget to remove those logs.

client/src/views/files/handler/FileHandler.vue
```diff
@@ -248,6 +252,7 @@ async function _getFileInfo(
       }
       offset += readResult.value.byteLength;
     }
+    fileInfoId.value = statsResult.value.id;
```
Max-7: What's the point of this variable?

client/src/views/files/handler/FileHandler.vue
```diff
@@ -452,46 +429,29 @@ async function onClick(event: MouseEvent): Promise<void> {
...
 
   if (!contentInfo.value) {
     window.electronAPI.log('error', 'No content info when trying to download a file');
     return;
   }
 
-  const config = await storageManager.retrieveConfig();
-  if (!config.disableDownloadWarning) {
-    const { result, noReminder } = await askDownloadConfirmation();
+  try {
```
Max-7: Incorrect conflict resolution, what you're deleting here is the most recent version.

client/src/views/files/handler/FileHandler.vue
```diff
@@ -61,28 +61,26 @@
...
             </ion-buttons>
           </div>
 
-          <!-- file-viewer component -->
-          <component
-            :is="viewerComponent"
-            :content-info="contentInfo"
-          />
+          <!-- file-handler sub-component -->
+          <slot @handlerReady="handlerReady = true" />
```
Max-7: This seems very weird.

If I understand correctly, you have a component that emits a signal to the parent that emits a signal to its child. I don't know if it's technically wrong but on a design level, it probably is. A parent should pass props to a child, a child should emit signals to the parent.

client/src/views/files/handler/editor/FileEditor.vue
```diff
@@ -0,0 +1,146 @@
...
+  emits('handlerReady');
+}
+
+async function getUrlFromFile(workspaceHandle: WorkspaceHandle, path: FsPath): Promise<string | undefined> {
+  const statsResult = await entryStat(workspaceHandle, path);
+  if (!statsResult.ok || !statsResult.value.isFile()) {
+    return;
+  }
+
+  const openResult = await openFile(workspaceHandle, path, { read: true });
```
Max-7: You're reading the whole file twice.

client/src/views/files/handler/editor/FileEditor.vue
```diff
@@ -0,0 +1,146 @@
...
+  const fileUrl = await getUrlFromFile(getWorkspaceHandle()!, getDocumentPath());
+  console.log('Cryptpad file URL:', fileUrl);
+  if (!fileUrl) {
+    window.electronAPI.log('error', 'Failed to retrieve file URL for Cryptpad edition');
+    return;
+  }
+  const cryptpadConfig = {
+    document: {
+      url: fileUrl,
+      fileType: contentInfoRef.value?.extension || '',
```
Max-7: What happens if we give an empty fileType or title?
`?`, `|` and `|| ''` do not handle errors, they shouldn't be used as quick shortcut to shut the linter up.

client/src/views/files/handler/editor/FileEditor.vue
```diff
@@ -0,0 +1,146 @@
...
+      title: contentInfoRef.value?.fileName || '',
+      key: fileInfoIdRef.value,
+    },
+    documentType: documentType.value!,
+    editorConfig: {
+      lang: longLocaleCodeToShort(I18n.getLocale()),
+    },
+    autosave: 10,
+    events: {
+      // eslint-disable-next-line @typescript-eslint/explicit-function-return-type
```
Max-7: Why not add an explicit return type instead?

client/src/views/files/handler/editor/FileEditor.vue
```diff
@@ -0,0 +1,146 @@
...
+      onSave: async (file: Blob, callback: () => void) => {
+        console.log('Cryptpad onSave event triggered');
+        // Handle save logic here
+        const workspaceHandle = getWorkspaceHandle();
+        const openResult = await openFile(workspaceHandle!, getDocumentPath(), { write: true, truncate: true });
```
Max-7: What happens if the workspaceHandle is not set?
`getWorkspaceHandle` and `getDocumentPath` use the URL to find those info. This callback should be a closure instead, having the information directly in its scope.

client/src/views/files/handler/editor/FileEditor.vue
```diff
@@ -0,0 +1,146 @@
...
+
+        if (!openResult.ok) {
```
Max-7, line 125: Probably should handle errors here.

client/src/views/files/handler/editor/FileEditor.vue
```diff
@@ -0,0 +1,146 @@
...
+        await writeFile(workspaceHandle!, fd, 0, new Uint8Array(arrayBuffer));
+        await closeFile(workspaceHandle!, fd);
+        callback();
+      },
+      onHasUnsavedChanges: (unsaved: boolean): void => {
+        console.log('Cryptpad unsaved changes:', unsaved);
+      },
+    },
+  };
+  if (Env.isEditicsEnabled()) {
```
Max-7: What happens otherwise?

client/src/views/files/handler/viewer/FileViewer.vue
```diff
@@ -0,0 +1,77 @@
...
+
+const emits = defineEmits<{
+  (e: 'handlerReady'): void;
+}>();
+
+async function onFileLoaded(contentInfo: FileContentInfo, fileInfo: DetectedFileType): Promise<void> {
+  contentInfoRef.value = contentInfo;
+
+  const component = await getComponent(fileInfo);
+  if (!component) {
```
Max-7, line 52: What does the page display in this case? We show a toast, and what else? Since `handlerReady` will never be called, I'm guessing the state is a bit uncertain.

Max-7 (changes requested):

client/src/components/files/handler/viewer/index.ts
NicoTuxx: Nope.
Handler is the common view above editor & viewer (that are 2 ways of handling a file).
Here viewer components are really specific to File viewers & have nothing to do with cryptpad / editing components.
We are not sure at the moment that this components will disappear (and I would not bet on it), so for now, it's easier to keep them separated.
If we managed to replace everything by only Cryptpad read-only mode, then the "editor" part will naturally be the only one remaining.

client/src/router/types.ts
NicoTuxx: Good question. I'm not sure how we want to treat this, I'll ask about it on call-meeting & adapt here in consequence.

NicoTuxx: Nevertheless it seems like a subject to address in:
[#10761](https://github.com/Scille/parsec-cloud/issues/10761)

client/src/views/files/handler/FileHandler.vue
NicoTuxx: done

client/src/views/files/handler/FileHandler.vue
NicoTuxx: use as file id in FileEditor

client/src/views/files/handler/FileHandler.vue
NicoTuxx: How to get what was here before?

client/src/views/files/handler/editor/FileEditor.vue
NicoTuxx: Where?

client/src/views/files/handler/editor/FileEditor.vue
NicoTuxx: [#10760](https://github.com/Scille/parsec-cloud/issues/10760)

client/src/views/files/handler/editor/FileEditor.vue

client/src/views/files/handler/editor/FileEditor.vue
NicoTuxx: I'm not sure to get what's the point here, you want the callback to take the needed info as params?

client/src/views/files/handler/editor/FileEditor.vue
NicoTuxx: [#10760](https://github.com/Scille/parsec-cloud/issues/10760)

client/src/views/files/handler/editor/FileEditor.vue
NicoTuxx: [#10761](https://github.com/Scille/parsec-cloud/issues/10761)

client/src/views/files/handler/viewer/FileViewer.vue
NicoTuxx: Maybe should be improved with changing how the handlerReady is working now.

client/src/views/files/handler/FileHandler.vue
NicoTuxx: I've found examples working like this & I've got no problems when testing this the way it is now.
But I agree that it looks weird, how would you replace this?

Max-7: Using a prop to pass the data from the parent to the child, and an event from the child to the parent if needed.

client/src/views/files/handler/editor/FileEditor.vue
Max-7: `onFileLoaded` gets a `ContentInfo` containing the file, and you're reading it again here.

client/src/views/files/handler/FileHandler.vue
Max-7: Can't this be put inside the content info with the other data instead of using a separate variable?

client/src/views/files/handler/editor/FileEditor.vue
Max-7: Not as its parameters since you have no control over those, but in its scope, instead of retrieving in the callback itself.

```typescript
function a() {
  const workspaceHandle = XX;
   
  return {
    events: {
      onSave: (data) => {
        console.log(workspaceHandle);
      }
    }
  }
}
```

client/src/views/files/handler/FileHandler.vue
Max-7: I think you'll have to manually copy.

client/src/views/files/handler/editor/FileEditor.vue
Max-7: I think it's a mistake to avoid dealing with errors right now. By that I don't mean displaying a beautiful error message, I just mean that how we deal with errors will have an impact on the design (interrupt the loading, perform a redirection, ...) Silencing the linter and think that's it's simply a matter of adding a toast later may cost us in the future.

client/src/views/files/handler/FileHandler.vue
NicoTuxx: Yh I did not understand what was the point at first, the rebase was failed, I'm working on this. 👌

client/src/views/files/handler/editor/FileEditor.vue
NicoTuxx: okay so basically it means that #10760 should be included in this PR.

client/src/views/files/handler/FileHandler.vue
NicoTuxx: fixed

client/src/views/files/handler/editor/FileEditor.vue
NicoTuxx: We need to decide what we want here & apply it in the given issue.
I propose that if the feature is disabled, the page should not load & redirect to somewhere else.

client/src/router/types.ts
NicoTuxx: We need to decide what we want here & apply it in the given issue.
I propose that if the feature is disabled, the page should not load & redirect to somewhere else.

client/src/views/files/handler/FileHandler.vue
NicoTuxx: yh totally. done.

client/src/views/files/handler/editor/FileEditor.vue
NicoTuxx: We discussed it with @Max-7 on slack we'll do as previous comment says.

client/src/router/types.ts
NicoTuxx: We discussed it with @Max-7 on slack we'll do as previous comment says.

client/src/views/files/handler/editor/FileEditor.vue

client/src/views/files/handler/editor/FileEditor.vue

client/src/views/files/handler/editor/FileEditor.vue
NicoTuxx: done

client/src/views/files/handler/editor/FileEditor.vue

PR #10795, opened 2025-07-21
[MS] Open FileViewer page on file double click opening & add open editor button in viewer

NicoTuxx:


client/src/components/files/FileOpenFallbackChoice.vue
```diff
@@ -38,17 +40,34 @@
...
 }>();
 
+const storageManager: StorageManager = inject(StorageManagerKey)!;
+
 async function onViewerClick(): Promise<void> {
   await modalController.dismiss(OpenFallbackChoice.View);
 }
 
 async function onDownloadClick(): Promise<void> {
+  const config = await storageManager.retrieveConfig();
```
Max-7: I think you can factorize the whole config thing. It's going to be used multiple times.

client/src/components/files/FileOpenFallbackChoice.vue
```diff
@@ -0,0 +1,68 @@
+<template>
+  <ms-modal
+    :title="'openFallback.title'"
```
Max-7: `title="openFallback.title"`, it's a little cleaner.

client/src/services/fileOpener.ts
Max-7, line 1: I mentionned this to @NicoTuxx, I think this whole file (and the whole `contentType` thing) needs to be factorized or cleaned or re-thought. That `openPath` function is one of the ugliest piece of code I've ever written, the logic is impossible to follow.

client/src/services/fileOpener.ts
```diff
@@ -40,6 +49,7 @@ const ENABLED_FILE_VIEWERS = [
 ];
 
 const OPEN_FILE_SIZE_LIMIT = 15_000_000;
+const fileOperationManager: FileOperationManager = inject(FileOperationManagerKey)!;
```
Max-7: You can't inject in something that's not a component, this has to be passed as an argument to functions that need it.

client/src/services/fileOpener.ts
```diff
@@ -86,19 +96,20 @@ async function showInExplorer(workspaceHandle: WorkspaceHandle, path: FsPath, in
...
 }
 
-async function openPath(
+async function getEntryStat(
   workspaceHandle: WorkspaceHandle,
   path: FsPath,
   informationManager: InformationManager,
   options: OpenPathOptions,
-): Promise<void> {
+  workspaceId?: string,
```
Max-7: You can get the workspaceId from the workspaceHandle, using `getWorkspaceInfo`.

client/src/services/fileOpener.ts
```diff
@@ -115,8 +126,155 @@ async function openPath(
...
+  path: FsPath,
+  workspaceHandle: WorkspaceHandle,
+  options: OpenPathOptions,
+  informationManager: InformationManager,
+  workspaceId?: string,
+  contentType?: DetectedFileType,
+): Promise<void> {
+  try {
+    if (!Env.isEditicsEnabled()) {
+      return;
```
Max-7: Probably won't call this function if Editics is not enabled, but we should still display a log here.

client/src/services/fileOpener.ts
```diff
@@ -115,8 +126,155 @@ async function openPath(
...
+            documentPath: entry.path,
+            timestamp: options.atTime?.toMillis().toString(),
+            fileTypeInfo: Base64.fromObject(contentType),
+          },
+          params: {
+            mode: FileHandlerMode.Edit,
+          },
+        });
+      }
+    }
```
Max-7, line 237: Same here, at minimum we should have a log if we have the wrong file type, instead of nothing happening.

client/src/router/checks.ts
```diff
@@ -79,3 +80,13 @@ export function currentRouteIsLoggedRoute(): boolean {
     Routes.FileHandler,
   ]);
 }
+
+export function getFileHandlerMode(): FileHandlerMode | null {
+  const router = getRouter();
+  const currentRoute = router.currentRoute.value;
+
+  if (currentRoute.name === Routes.FileHandler) {
+    return currentRoute.params.mode as FileHandlerMode;
```
Max-7: The `mode` param may not be present, which means you're casting `undefined` as a `FileHandlerMode`, silencing every error the linter could have found. Or it could be a weird value. I think you can replace the return type by `FileHandlerMode | undefined`, check that it's present and is a FileHandlerMode before casting, and otherwise just return undefined.

client/src/views/files/handler/FileHandler.vue
```diff
@@ -366,8 +380,8 @@ function loadComponent(): void {
 }
 
 onMounted(async () => {
+  handlerMode.value = getFileHandlerMode();
```
Max-7, line 383: What happens if we don't have a mode?
I think it's easier to just default to viewer all the time and not let it be undefined.

Max-7 (changes requested):

client/src/router/checks.ts
NicoTuxx: done
I've added an auto-redirect on the router if no mode is set in params

client/src/views/files/handler/FileHandler.vue
NicoTuxx: I've added a condition in the router to avoid no handler mode, the route is not usable without one now.

client/src/services/fileOpener.ts
NicoTuxx: done

client/src/services/fileOpener.ts
NicoTuxx: was done in openPath, but I've replaced in each call.

client/src/services/fileOpener.ts
NicoTuxx: Added logs & throw error to fall into the catch block with fallback options.

client/src/services/fileOpener.ts
NicoTuxx: Added logs & throw error to fall into the catch block with fallback options.

PR #10799, opened 2025-07-22
feat(client-conn): Configure min TLS version allowed by client

FirelightFlagboy:
We cannot set it to TLSv1.3 as it's only supported by `reqwest` when using `rustls-tls` feature.


FirelightFlagboy (commented):
- [x] Test TLS config in a test env

PR #10805, opened 2025-07-22
Only do user frozen check during authentication  in PostgreSQL

touilleMan:
Only do user frozen check during authentication
    
- Simplify code by not having this check in each command logic code
- Simplify testing since the check in each command logic code is shadowed by the one in the authentication layer
- Returning a "revoked" error for frozen in the command logic code is confusing
- Allow to use `backend.xxx` methods to do something even if the user is frozen (useful in tests)


server/parsec/components/postgresql/migrations/datamodel.sql
```diff
@@ -108,6 +108,12 @@ CREATE TABLE user_ (
     -- This field is altered in an `ALTER TABLE` statement below
     -- in order to avoid cross-reference issues
     shamir_recovery INTEGER,
+    -- `frozen`  is just a service-side access control flag to indicate the
```
FirelightFlagboy: **nitpick:** 
```suggestion
    -- `frozen` is just a service-side access control flag to indicate the
```

FirelightFlagboy (commented):
Can you send me a link where the authentication step check for the frozen state of the user?

server/parsec/components/postgresql/migrations/datamodel.sql
touilleMan: done. :+1:

touilleMan:
> Can you send me a link where the authentication step check for the frozen state of the user?

https://github.com/Scille/parsec-cloud/blob/873837bfaa79319fefe192fbd819d0ce63979164/server/parsec/components/postgresql/auth.py#L341-L347

FirelightFlagboy:
And in memory mode: https://github.com/Scille/parsec-cloud/blob/873837bfaa79319fefe192fbd819d0ce63979164/server/parsec/components/memory/auth.py#L114-L115

PR #10808, opened 2025-07-22
[MS] Account recovery page

### LINKED ISSUE #9823: [🚀 | Feature request]: Forgotten password

Figma link: https://www.figma.com/design/XW3SNWHV8BhPsPJpsOCV2h/Parsec-App?node-id=21196-36534&p=f&t=MCB1l49mUST1Uzxd-11

## Preview
Image

If the user has forgiven his Parsec password.

_No response_

Max-7:
Closes #9823

Removed `PARSEC_APP_MOCK_ACCOUNT` since it's no longer needed.

You can probably test using autologin and copying the generated email, but the easiest way is probably to create an account, log out, do the whole recovery process and check that you can log in with the new password.


client/src/views/account/RecoverAccountPage.vue
```diff
@@ -0,0 +1,481 @@
...
+
+async function codeEntered(): Promise<void> {
+  if (code.value.length !== 6) {
+    return;
+  }
+  error.value = '';
+  step.value = Steps.NewAuthentication;
+}
+
+async function onPasswordChosen(): Promise<void> {
```
Ironicbay, line 293: Inputing an already used password does not seem to work and raises a generic error
If an user had password A then changed it to password B, and went to recover their account, they are not able to set either password A nor B and the error message is misleading.

client/src/views/account/RecoverAccountPage.vue
```diff
@@ -0,0 +1,481 @@
...
+    resendDisabled.value = false;
+  }, 5000);
+}
+
+const validAuth = asyncComputed(async () => {
+  return step.value === Steps.NewAuthentication && passwordInputRef.value && (await passwordInputRef.value.areFieldsCorrect());
+});
+
+const watchCancel = watchRoute(async (_newRoute, oldRoute) => {
+  if (oldRoute.name !== Routes.RecoverAccount) {
```
Ironicbay: Why use this condition? It would be nice to clear all fields once the recovery is done / page is left.

client/src/locales/en-US.json
```diff
@@ -1663,6 +1663,38 @@
             "password": "Password",
             "confirmPassword": "Confirm password"
         },
+        "recoverAccount": {
+            "step1": {
+                "title": "Recover your account",
+                "description": "No worries, we will help you recover your account.",
```
Ironicbay: I would change this line to something more basic
```
Please enter the e-mail address used for your account and check that the server address is correct.
```

client/src/locales/fr-FR.json
```diff
@@ -1662,6 +1662,38 @@
             "password": "Mot de passe",
             "confirmPassword": "Confirmer le mot de passe"
         },
+        "recoverAccount": {
+            "step1": {
+                "title": "Récupérer votre compte",
+                "description": "Pas d'inquiétude, nous allons vous aider à récupérer votre compte.",
```
Ironicbay: Same here
```
Veuillez renseigner l'adresse e-mail liée à votre compte, et vérifier que l'adresse du serveur est correcte.
```

client/src/locales/fr-FR.json
```diff
@@ -1662,6 +1662,38 @@
...
+                "title": "Vérification de votre email",
+                "description": "Un code de vérification a été envoyé à votre adresse électronique. Veuillez le saisir pour mettre à jour votre mot de passe.",
+                "resendCode": "Renvoyer un code",
+                "nextButton": "Valider",
+                "update": "Modifier",
+                "info": "Si vous ne recevez pas le code, veuillez vérifier votre dossier de spam ou de courrier indésirable."
+            },
+            "step3": {
+                "title": "Choisissez votre nouveau mot de passe",
+                "description": "Vérifiez que votre nouveau mot de passe comprend les critères minimaux pour être valide.",
```
```suggestion
                "description": "Votre nouveau mot de passe doit respecter les critères minimaux de robustesse.",
```

client/src/locales/en-US.json
```diff
@@ -1663,6 +1663,38 @@
...
+                "title": "Email Verification",
+                "description": "A verification code has been sent to your email address. Please enter it to update your password.",
+                "resendCode": "Resend code",
+                "nextButton": "Validate",
+                "update": "Update",
+                "info": "If you do not receive the code, please check your spam or junk folder."
+            },
+            "step3": {
+                "title": "Choose your new password",
+                "description": "Make sure your new password meets the minimum criteria to be valid.",
```
```suggestion
                "description": "Make sure your new password meets the minimum strength criteria.",
```

client/src/views/account/RecoverAccountPage.vue
Max-7: It's here because for whatever reason, when navigating to a new route, the router keeps the component alive. It makes no sense to me, it makes everything painful, but that's what it does. So if you're at the code stage and you're clicking to `back to login` and then coming back to the account recovery, you want the whole process to restart.

client/src/views/account/RecoverAccountPage.vue
Max-7: Yep it is a "limitation" right now, we're only ever adding authentication instead of replacing. I'm not really sure what to do about it. A message like "well, good news, you remembered your old password" seems a bit weird.

client/src/locales/en-US.json
Max-7: I don't really want to mention the server address. Ideally we just set a default one and the user never has to change it, it's just here so it can be changed if needed. The user has no idea whether it's correct or not.

Ironicbay (approved):

PR #10809, opened 2025-07-22
refactor(crypto): Migrate to `libsodium-rs`

FirelightFlagboy:


libparsec/crates/crypto/src/sodiumoxide/hash.rs
```diff
@@ -1,18 +1,18 @@
 // Parsec Cloud (https://parsec.cloud) Copyright (c) BUSL-1.1 2016-present Scille SAS
```
touilleMan, line 1: Rename `libparsec/crates/crypto/src/sodiumoxide` -> `libparsec/crates/crypto/src/libsodium` ?

libparsec/crates/crypto/src/sodiumoxide/mod.rs
```diff
@@ -17,3 +17,9 @@ pub use secret::*;
 pub use sequester::*;
 pub use sign::*;
 pub(crate) use utils::*;
+
+pub fn init() {
```
touilleMan: remove since libsodium-rs already self-initialize on load

.github/workflows/ci-rust.yml
```diff
@@ -97,7 +97,7 @@ jobs:
       - name: Install build dependencies
         shell: bash
         run: |
-          until sudo apt-get -y install fuse3 libfuse3-dev libdbus-1-dev; do
+          until sudo apt-get -y install fuse3 libfuse3-dev libdbus-1-dev libsodium-dev; do
```
touilleMan: Why is it needed ? How `libsodium-rs` choose to use the libsodium code provided from the OS or to download its own version ?

My guess is we do not need this, since `libsodium-rs` always takes care of downloading `libsodium` source code since it must get a precise version that depends of the version of `libsodium-rs` (while doing `apt install libsodium-dev` provide an arbitrary version of libsodium)

FirelightFlagboy: It's faster to use the system provided libsodium than for Rust to compile it from source

libparsec/crates/crypto/src/sodiumoxide/mod.rs
FirelightFlagboy: done

libparsec/crates/crypto/src/sodiumoxide/hash.rs
FirelightFlagboy: done

.github/workflows/ci-rust.yml
touilleMan: What I understand [from libsodium-sys's `build.rs`](https://github.com/jedisct1/libsodium-sys-stable/blob/4f2a39d84f34cff2422ec724bc17df2dfce558c1/build.rs#L509-L532) is that installing libsodium on the system is only used if the feature `use-pkg-config` is set.

I didn't see this feature enable in this PR, and I could run `cargo build -p libparsec_crypto --features use-libsodium` on my machine without `libsodium-dev` installed.

Besides, I think it is not worth it to depend on a system dependency here:
- the build cache should prevent the compilation most of the time
- we don't want to have slightly different versions between CI / dev machines / actual release

Cargo.toml
```diff
@@ -153,7 +153,9 @@ js-sys = { version = "0.3.70", default-features = false }
 keyring = { version = "3.6.2", default-features = false }
 lazy_static = { version = "1.5.0", default-features = false }
 libc = { version = "0.2.174", default-features = false }
-libsodium-sys = { version = "0.2.7", default-features = false }
+
+# We use the version which depends `libsodium-rs`
```
touilleMan:
```suggestion
# We use the version on which depends `libsodium-rs`
```

libparsec/crates/crypto/src/sodium/key_derivation.rs
```diff
@@ -50,18 +50,20 @@ impl KeyDerivation {
...
         let context = id_high;
+        debug_assert_eq!(context.len(), blake2b::CONTEXTBYTES);
 
-        let mut subkey = GenericArray::default();
-        derive_from_key(&mut subkey, subkey_id, *context, &self.0)
+        let subkey = blake2b::derive_from_key(Size::USIZE, subkey_id, context, &self.0)
             .expect("subkey has always a valid size");
 
-        subkey
+        // generic-array@1.2.0 provide better way to convert from an array but we are stuck by
```
touilleMan:
```suggestion
        // TODO: replace `from_exact_iter` by `from_array` once generic-array is updated to v1.0+
```

Cheap recycle since this comment is already in the codebase ^^
The important this though is to have a `TODO` marker so we can easily go back to it later on

.github/workflows/ci-rust.yml
FirelightFlagboy: done

PR #10810, opened 2025-07-23
[MS] Updated download from file viewers

Max-7:
- Update the download feature from the file viewers to use the fileOperationManager, so that downloads get added to the file operations menu
- Added the file operation menu to the viewers, so that the download is showing
- Fixed playwright with SSL
- Little bit of factorization for download
- Added a test for download from the file viewers

### DIFF
--- a/client/src/router/types.ts
+++ b/client/src/router/types.ts
@@ -124,6 +124,11 @@ const routes: Array<RouteRecordRaw> = [
                         name: Routes.History,
                         component: () => import('@/views/workspaces/WorkspaceHistoryPage.vue'),
                       },
+                      {
+                        path: `/:handle(\\d+)/${Routes.Viewer}`,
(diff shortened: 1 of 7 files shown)


client/playwright.config.ts
```diff
@@ -28,13 +28,15 @@ export default defineConfig({
   reporter: IN_CI ? 'blob' : 'list',
   webServer: {
     command: 'npm run dev -- --port 8080',
-    url: 'http://localhost:8080',
+    ignoreHTTPSErrors: true,
```
FirelightFlagboy: **question:** What kind of HTTPS error did you get?

Ironicbay (approved):

client/playwright.config.ts
Max-7: Self-signed certificate errors.

PR #10817, opened 2025-07-24
[MS] Option to download multiple files/folders as an archive

### LINKED ISSUE #10748: Download multiple files (or folders)
Currently, it's not possible to download a folder, and downloading multiple files at once only works properly on Chrome because other browsers do not support the FileSystemAPI fully (https://caniuse.com/?search=showDirectoryPicker).

We could instead archive the files. This means only using `showSaveFilePicker` (or the shim that we use currently) which seems to work properly to download a single file: an archive.

Neither compression nor performances matter much, the priority is being able to stream the archive creation to avoid saturating the memory with one big archive at any given moment. Being able to display the progress and cancel the archive creation would also be nice.

I've done some testing using [zipjs](https://github.com/gildas-lormeau/zip.js/). I chose it over [fflate](https://github.com/101arrowz/fflate) because we don't really care about performances and it seems way easier to use, as the following code shows:

```typescript
// Get the handle for the file on disk
const saveHandle = await showSaveFilePicker({
  _preferPolyfill: false,
  suggestedName: 'archive.zip',
});

// Get the stream to write to
const wStream = await saveHandle.createWritable();
// Create a zip write, giving it the stream to write to
const zipWriter = new zip.ZipWriter(wStream);

// Iterating over all the selected files
for (const entry of selectedEntries) {
  // Create a readable stream for this file
  const rStream = await createReadStream(workspaceInfo.value.handle, entry.path);

  try {
    // Add the file to the zip with its name and the readable stream
    await zipWriter.add(entry.name, rStream);
  } catch (e: any) {
    window.electronAPI.log('warn', `Error while downloading file: ${e?.toString()}`);
  }
}
await zipWriter.close();
```
It does create a zip file containing every file in its entirety, does so asynchronously and streams the result (I checked, the file on disk gets increasingly larger, it does not appear suddenly).

In theory it could also be possible to add a password to the archive.

Required features:
- Inform the user about the size of the archive before creating it
- Cancel the creation
- See progress (at least file by file)
- Handle multiple files
- Handle folders recursively
- Option to set a password to the archive?

Max-7:
Closes #10748


client/src/components/files/FileDownloadArchive.vue
```diff
@@ -0,0 +1,327 @@
...
+import { checkmarkCircle, closeCircle } from 'ionicons/icons';
+import { computed, onMounted, ref } from 'vue';
+
+const props = defineProps<{
+  operationData: DownloadArchiveData;
+  state: FileOperationState;
+  stateData?: StateData;
+}>();
+
+const workspaceName = ref('');
```
Ironicbay: Unused variable

client/src/components/files/FileDownloadArchive.vue
```diff
@@ -0,0 +1,327 @@
...
+const icon = File.Zip;
+const progress = computed(() => {
+  if (props.state === FileOperationState.Cancelled || props.state === FileOperationState.EntryDownloaded) {
+    return { name: undefined, size: undefined, totalProgress: 100 };
+  } else if (props.state === FileOperationState.OperationProgress && props.stateData) {
+    const data = props.stateData as DownloadOperationProgressStateData;
+    return { name: data.currentFile, size: data.currentFileSize, totalProgress: data.progress };
+  }
+  return { name: undefined, size: undefined, totalProgress: 0 };
+});
```
Ironicbay, line 146: `size` attribute seems unused

client/src/locales/en-US.json
```diff
@@ -799,7 +799,13 @@
             "openFailed": "Could not open the file. You can try using your browser's downloads list.",
             "cancel": "Cancel",
             "download": "Download file",
-            "notRemindMe": "Do not remind me"
+            "notRemindMe": "Do not remind me",
+            "maximumRecursionReached": "Maximum recursion reached, cannot download",
```
Ironicbay: Not sure if `recursion` is clear enough for a standard user.
```
Maximum subfolder depth reached, cannot download.
```

client/src/locales/fr-FR.json
```diff
@@ -798,7 +798,13 @@
             "openFailed": "Impossible d'ouvrir le fichier. Vous pouvez essayer en utilisant la liste des téléchargement de votre navigateur.",
             "cancel": "Annuler",
             "download": "Télécharger le fichier",
-            "notRemindMe": "Ne plus me le rappeler"
+            "notRemindMe": "Ne plus me le rappeler",
+            "maximumRecursionReached": "Récursion maximale atteinte, impossible de télécharger",
```
```suggestion
            "maximumRecursionReached": "Profondeur maximale de sous-dossiers atteinte, impossible de télécharger",
```

client/src/locales/fr-FR.json
```diff
@@ -798,7 +798,13 @@
...
+            "maximumFilesReached": "Trop de fichiers, impossible de télécharger",
```
```suggestion
            "maximumFilesReached": "Trop de fichiers sélectionnés, impossible de télécharger",
```

client/src/locales/en-US.json
```diff
@@ -799,7 +799,13 @@
...
+            "maximumFilesReached": "Too many files, cannot download",
```
```suggestion
            "maximumFilesReached": "Too many selected files, cannot download",
```

client/src/parsec/file.ts
```diff
@@ -171,7 +171,11 @@ export async function parseFileLink(link: string): Promise<Result<ParsedParsecAd
   return result as Result<ParsedParsecAddrWorkspacePath, ParseParsecAddrError>;
 }
 
-export async function createReadStream(workspaceHandle: WorkspaceHandle, path: FsPath): Promise<ReadableStream> {
+export async function createReadStream(
+  workspaceHandle: WorkspaceHandle,
+  path: FsPath,
+  cb?: (readSize: number) => Promise<void>,
```
Ironicbay: Maybe rename it explicitly to `callback`

client/src/services/fileOperationManager.ts
```diff
@@ -808,6 +868,80 @@ class FileOperationManager {
     }
   }
 
+  private async doDownloadArchive(data: DownloadArchiveData): Promise<void> {
```
Ironicbay, line 871: The generated zip file is incompatible with MacOS 
Screenshot 2025-07-29 at 10 29 06

client/src/components/files/FileDownloadArchive.vue
Max-7: Yep, I've kept it just in case @fabienSvtr wanted to display the size of the current file somewhere.

client/src/services/fileOperationManager.ts
```diff
@@ -808,6 +868,86 @@ class FileOperationManager {
...
+    let wStream: FileSystemWritableFileStream | undefined;
+    let cancelled = false;
+    try {
+      wStream = await data.saveHandle.createWritable();
+      zipWriter = new zipjs.ZipWriter(wStream, {
+        level: 3,
+        compressionMethod: 0x08,
+        useUnicodeFileNames: true,
+        supportZip64SplitFile: false,
+        zip64: false,
```
FirelightFlagboy, line 887: **question:** Why not using 64bits format?

So we are not limited to 4GB

Max-7: I haven't been able to generate a single zip with this option set to true that is compatible with MacOS :/

FirelightFlagboy: MacOS still living like it's 2006 :confused:

FirelightFlagboy: Do you use/check for the 32bits limit? (number of file, uncompressed size, compressed size, ...)

Max-7: Right now we don't, we'll have to run some tests but it probably takes 30min to generate an archive of that size, and I don't know what criteria to use to display an error. 
I've opened #10843 as a reminder to run all those tests.

PR #10824, opened 2025-07-25
Account `auth_method_create`

touilleMan:

PR #10825, opened 2025-07-25
Rework `Account.(login|create_3_proceed|recover_2_proceed)` to allow both password and master secret auth

touilleMan:


libparsec/crates/account/src/account_create.rs
```diff
@@ -92,40 +87,43 @@ pub(super) async fn account_create(
...
+                        .compute_key_derivation(password)
+                        .expect("algorithm config is valid");
+
+                    let auth_method_keys = derive_auth_method_keys(&auth_method_master_secret);
+                    (
+                        Some(auth_method_password_algorithm.into()),
+                        auth_method_keys,
+                    )
+                }
             };
```
FirelightFlagboy: **nitpick:** This block is repeated multiple time, can you factorize it?

touilleMan: I only see this block in `account_create` and `account_recover_proceed`, so I don't think it is worth it to factorize (also note the most error-prone stuff has already been factorize in `derive_auth_method_keys()`)

FirelightFlagboy: The block is duplicated again in the next PR #10824

libparsec/crates/account/src/account_create.rs
```diff
@@ -92,40 +87,43 @@ pub(super) async fn account_create(
         },
         AccountCreateStep::Proceed {
             human_handle,
-            password,
+            auth_method_strategy,
```
mmmarcos, line 91: Would use same order as enum definition:

```rust
        AccountCreateStep::Proceed {
            validation_code,
            human_handle,
            auth_method_strategy,
        }
```

Max-7 (approved):

libparsec/crates/account/src/account_create.rs
touilleMan: Ok you convinced me :smiley: , done :+1:

mmmarcos (approved):
LGTM

libparsec/crates/account/src/account_create.rs
touilleMan: done :+1:

PR #10837, opened 2025-07-29
Fix sqlfluff linter errors in auth component

mmmarcos:


server/parsec/components/postgresql/auth.py
```diff
@@ -55,18 +55,21 @@
...
-        SELECT type FROM invitation WHERE organization = organization._id AND token = $token LIMIT 1
-    ) as invitation_type,
+        SELECT invitation.type FROM invitation
+        WHERE invitation.organization = organization._id AND invitation.token = $token LIMIT 1
+    ) AS invitation_type,
     (
-        SELECT deleted_on FROM invitation WHERE organization = organization._id AND token = $token LIMIT
-    1) as invitation_deleted_on
+        SELECT invitation.deleted_on FROM invitation
+        WHERE invitation.organization = organization._id AND invitation.token = $token LIMIT 1
```
FirelightFlagboy, line 71: **question:** Can these sub-queries be factorized?

below we have some example using `WITH name AS (...)`

mmmarcos: I'm not sure. Probably in this case both forms are equivalent, but I prefer to limit changes to fixing linter errors, just in case.

PR #10844, opened 2025-07-30
Fix `libparsec_protocol` not detecting a rebuild is needed when a .json5 schema is added

touilleMan:


FirelightFlagboy:
1. I do not see how that fix the issue
2. That increase the size of the generated data which slow down compilation

touilleMan:
> I do not see how that fix the issue

consider the following:
1. `cargo build -p libparsec_protocol`
2. `git checkout otherbranch`
3. `cargo build -p libparsec_protocol`

If `otherbranch` introduces a new schema, that means a new .json5 file is present (let's say in `libparsec/crates/protocol/schema/authenticated_cmds/new_command.json5`).
However build done at step 1 has produced `cargo:rerun-if-changed=` commands according to a `glob("schema/**/*.json5")` that occured *before* this file was present !
So in the end the build done is step 3 doesn't detect a re-build is needed and does nothing, which in turn produce rather strange errors such as (real life example ^^):
```
[INFO]: 🌀  Compiling to Wasm...
   Compiling libparsec_account v3.4.1-a.0+dev (/home/max/parsec-cloud/libparsec/crates/account)
error[E0432]: unresolved import `libparsec_protocol::authenticated_account_cmds::latest::auth_method_create`
  --> libparsec/crates/account/src/create_auth_method.rs:27:65
   |
27 |     use libparsec_protocol::authenticated_account_cmds::latest::auth_method_create::{Rep, Req};
   |                                                                 ^^^^^^^^^^^^^^^^^^ could not find `auth_method_create` in `latest`

For more information about this error, try `rustc --explain E0432`.
error: could not compile `libparsec_account` (lib) due to 1 previous error```
```

> That increase the size of the generated data which slow down compilation

Regarding runtime, the compiler is supposed to be smart enough to totally strip those data since they are private to the module and never used.
Regarding the slowdown in compilation speed, I would be really surprised this has any significant impact since it just boils down to reading a bunch of files (I'm an old C/C++ dev, reading files are all across the compilation with the `#include` ^^).

touilleMan:
I've just tested the compilation time:
With this patch
```shell
$ cargo clean -p libparsec_protocol
     Removed 277 files, 277.2MiB total

$ time cargo build -p libparsec_protocol
   Compiling libparsec_protocol v3.4.1-a.0+dev (/home/emmanuel/projects/parsec-cloud/libparsec/crates/protocol)
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 5.35s

________________________________________________________
Executed in    5,23 secs    fish           external
   usr time    8,63 secs    0,00 millis    8,63 secs
   sys time    1,22 secs    1,04 millis    1,22 secs
```

Without this patch:
```shell
$ cargo clean -p libparsec_protocol
     Removed 277 files, 277.2MiB total

$ time cargo build -p libparsec_protocol
   Compiling libparsec_protocol v3.4.1-a.0+dev (/home/emmanuel/projects/parsec-cloud/libparsec/crates/protocol)
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 6.51s

________________________________________________________
Executed in    5,17 secs    fish           external
   usr time    8,49 secs  143,00 micros    8,49 secs
   sys time    1,26 secs  970,00 micros    1,25 secs
```

(I've run both build multiple times and got time in the range 5.1s to 6.4s, so no significant difference seems to be introduced by this patch)

touilleMan:
I've also compile the Python bindings with and without this patch and obtain the same size

FirelightFlagboy:
I understand the scenario, but I don't see how this PR solve it

touilleMan:
> I understand the scenario, but I don't see how to this PR solve it

Good catch ! My assumption were wrong and this "fix" doesn't fix anything :smile: 

However there is still hope: I've just discovered that `cargo:rerun-if-changed` also support directories \o/
So I've updated this PR to use this instead (and it seems to work fine now ^^)

FirelightFlagboy (approved):

FirelightFlagboy:
Can you do the same for types ?

touilleMan:
> Can you do the same for types ?

I've added this check for the python bindings :+1: 

And I've totally removed this for the `libparsec_types` crate, the reason being that this crate only use the `libparsec_serialization_format::parsec_data` macro that takes the path to a .json5 file (and not to a folder like what is done in `libparsec_protocol`).
Hence if a new .json5 is added, it must be referenced by a new call to `libparsec_serialization_format::parsec_data` in the Rust code, hence automatically triggering a rebuild without the need for any `cargo:rerun-if-changed` \o/

touilleMan:
> I think we can remove the glob dependencies now 🤔

:+1: done !

PR #10880, opened 2025-08-04
ci(python): Use `poetry` to run `make.py`

FirelightFlagboy:
Poetry will then use the python version setup with `poetry env use` during the `setup-python-poetry` step


FirelightFlagboy:
Superseded by #10883

PR #10932, opened 2025-08-14
[MS] Show full path in file details modal

### LINKED ISSUE #10918: [MS] Details on a files doesn't show the full path
Image

also note that it doesn't go fully to the right part or the greyed area. 

Possible solutions:
- on hoover or click show the full path
- allow resizing the modal / the path box

fabienSvtr:
## Preview
image

closes #10918

### DIFF
--- a/client/src/views/files/FileDetailsModal.vue
+++ b/client/src/views/files/FileDetailsModal.vue
@@ -94,7 +94,9 @@
           </ion-label>
           <div class="file-info-path-value">
             <ion-text class="file-info-path-value__text body">
-              {{ shortenFileName(entry.path, { maxLength: 60, prefixLength: 20, suffixLength: 30 }) }}
+              <span>
+                {{ showFullPath ? entry.path : shortenFileName(entry.path, { maxLength: 60, prefixLength: 20, suffixLength: 30 }) }}
+              </span>
             </ion-text>
             <template v-if="isDesktop()">
               <ion-button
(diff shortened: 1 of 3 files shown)


client/src/views/files/FileDetailsModal.vue
```diff
@@ -123,6 +126,12 @@
               </ion-text>
             </template>
           </div>
+          <ion-text
+            class="file-info-path__full button-small"
+            @click="toggleShowFullPath = !toggleShowFullPath"
+          >
+            {{ $msTranslate('FileDetails.stats.fullPath') }}
+          </ion-text>
```
Ironicbay: I would either hide this text if full path is shown (user has willingly clicked to expand, it isn't very invasive and should reset anyway with the next modal opening), or have two texts `See full path` `See cropped path`/`Hide full path` depending on the boolean state.

client/src/views/files/FileDetailsModal.vue
```diff
@@ -94,7 +94,10 @@
           </ion-label>
           <div class="file-info-path-value">
             <ion-text class="file-info-path-value__text body">
-              {{ shortenFileName(entry.path, { maxLength: 60, prefixLength: 20, suffixLength: 30 }) }}
+              <span v-if="toggleShowFullPath">
```
Ironicbay: Logic is reversed, cropped path should be shown by default and full path should be shown after clicking

Ironicbay (changes requested):

client/src/views/files/FileDetailsModal.vue
fabienSvtr: I updated it to show the button only if the path length is > to 60 characters. And I had a toggle version for the button text (show/hide).

Ironicbay (approved):

PR #10950, opened 2025-08-19
[MS] Checklist recommandations responsive

### LINKED ISSUE #10705: [MS] Responsive - Recommandation checklist
Following the issue #10646 
## What to do
- Mockups : https://www.figma.com/design/XW3SNWHV8BhPsPJpsOCV2h/Parsec-App?node-id=35251-99978&t=RtNiqDQf6DztMAtg-11
- Dev

fabienSvtr:
closes #10705

### DIFF
--- a/client/src/views/header/HeaderPage.vue
+++ b/client/src/views/header/HeaderPage.vue
@@ -94,6 +94,27 @@
                 :icon="search"
               />
             </ion-button>
+            <div
+              v-if="!isMobile() && securityWarningsCount > 0 && securityWarnings && isSmallDisplay"
+              id="trigger-checklist-button"
+              class="topbar-right-button__item unread"
+              @click="openSecurityWarningsModal()"
+              ref="checklistSecurityButton"
+            >
+              <ion-icon :icon="checkmarkCircle" />
+              <div class="checklist-security-levels">
+                <span
+                  v-if="securityWarnings?.isWorkspaceOwner"
+                  class="security-level"
+                  :class="{ 'security-level--done': securityWarningsCount < 3 }"
+                />
+                <span
+                  class="security-level"
+                  :class="{ 'security-level--done': securityWarningsCount < 2 }"
+                />
+                <span class="security-level" />
+              </div>
+            </div>
             <ion-button
               v-if="!isMobile()"
               slot="icon-only"
@@ -133,8 +154,9 @@ import { pxToRem } from '@/common/utils';
 import HeaderBackButton from '@/components/header/HeaderBackButton.vue';
 import HeaderBreadcrumbs, { RouterPathNode } from '@/components/header/HeaderBreadcrumbs.vue';
 import InvitationsButton from '@/components/header/InvitationsButton.vue';
-import { ClientInfo, Path, UserProfile, getClientInfo, isMobile, getWorkspaceName } from '@/parsec';
+import { ClientInfo, Path, UserProfile, getClientInfo, isMobile, getWorkspaceName, WorkspaceRole } from '@/parsec';
 import {
+  ProfilePages,
   Routes,
   currentRouteIs,
   currentRouteIsFileRoute,
@@ -152,10 +174,11 @@ import { HotkeyGroup, HotkeyManager, HotkeyManagerKey, Modifiers, Platforms } fr
 import { InformationManager, InformationManagerKey } from '@/services/informationManager';
 import useSidebarMenu from '@/services/sidebarMenu';
 import useHeaderControl from '@/services/headerControl';
-import { Translatable, MsImage, SidebarToggle, useWindowSize } from 'megashark-lib';
+import { Translatable, MsImage, SidebarToggle, useWindowSize, MsModalResult } from 'megashark-lib';
 import NotificationCenterPopover from '@/views/header/NotificationCenterPopover.vue';
 import NotificationCenterModal from '@/views/header/NotificationCenterModal.vue';
 import ProfileHeaderOrganization from '@/views/header/ProfileHeaderOrganization.vue';
+import RecommendationChecklistPopoverModal from '@/components/misc/RecommendationChecklistPopoverModal.vue';
 import { openSettingsModal } from '@/views/settings';
 import {
   IonButton,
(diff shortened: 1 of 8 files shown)


client/src/views/header/HeaderPage.vue
```diff
@@ -187,6 +202,17 @@ const notificationPopoverIsVisible: Ref<boolean> = ref(false);
 const informationManager: InformationManager = inject(InformationManagerKey)!;
 const eventDistributor: EventDistributor = inject(EventDistributorKey)!;
 const notificationCenterButtonRef = useTemplateRef('notificationCenterButton');
+const securityWarnings = ref<SecurityWarnings | undefined>();
+const securityWarningsCount = computed(() => {
```
Ironicbay, line 216: Since you're only checking if there is at least 1 warning and not how many there are, you can change this const to a boolean to simplify it. It's only done this way in large display to show as text (`X remaining tasks`) which is not the case here.

```
const activeSecurityWarnings = computed(() => {
  if (!securityWarnings.value) {
    return false;
  }
  return (
    securityWarnings.value.hasRecoveryDevice ||
    securityWarnings.value.hasMultipleDevices ||
    securityWarnings.value.soloOwnerWorkspaces.length === 0
  );
});
```

Ironicbay (commented):
I would use a blue or grey notification circle instead of red, since this is an advice/info (and presented as such in large display)

client/src/views/header/HeaderPage.vue
```diff
@@ -187,6 +215,17 @@ const notificationPopoverIsVisible: Ref<boolean> = ref(false);
...
+  if (!securityWarnings.value) {
+    return false;
+  }
+  return (
+    securityWarnings.value.hasRecoveryDevice ||
+    securityWarnings.value.hasMultipleDevices ||
+    securityWarnings.value.soloOwnerWorkspaces.length === 0
+  );
+});
```
Ironicbay: Now that you're using the warning count and not just a boolean, you need to revert this to a `number`
```
const securityWarningsCount = computed(() => {
  if (!securityWarnings.value) {
    return 0;
  }
  return (
    (securityWarnings.value.hasRecoveryDevice ? 0 : 1) +
    (securityWarnings.value.hasMultipleDevices ? 0 : 1) +
    (securityWarnings.value.soloOwnerWorkspaces.length === 0 ? 0 : 1)
  );
});
```

client/src/views/header/HeaderPage.vue
```diff
@@ -94,6 +94,30 @@
...
               />
             </ion-button>
+            <div
+              v-if="!isMobile() && securityWarningsCount && isSmallDisplay"
+              id="trigger-checklist-button"
+              class="topbar-right-button__item"
+              :class="{ unread: securityWarningsCount > 0 }"
+              @click="openSecurityWarningsModal()"
+              ref="checklistSecurityButton"
+            >
```
Ironicbay: - Explicit the condition since the count is back to being a `number`
- This div should be shown only if there is at least 1 warning, so it will always be `unread`
```suggestion
            <div
              v-if="!isMobile() && securityWarningsCount > 0 && isSmallDisplay"
              id="trigger-checklist-button"
              class="topbar-right-button__item unread"
              @click="openSecurityWarningsModal()"
              ref="checklistSecurityButton"
            >
```

client/src/views/header/HeaderPage.vue
```diff
@@ -94,6 +94,30 @@
...
+              <ion-icon :icon="checkmarkCircle" />
+              <div class="checklist-security-levels">
+                <span
+                  class="security-level"
+                  :class="{ 'security-level--done': securityWarningsCount }"
+                />
```
Ironicbay:
```suggestion
                <span
                  class="security-level"
                  :class="{ 'security-level--done': securityWarningsCount < 3 }"
                />
```

client/src/views/header/HeaderPage.vue
```diff
@@ -94,6 +94,30 @@
...
+                <span
+                  class="security-level"
+                  :class="{ 'security-level--done': securityWarningsCount.length > 1 }"
+                />
```
Ironicbay:
```suggestion
                <span
                  class="security-level"
                  :class="{ 'security-level--done': securityWarningsCount.length === 1 }"
                />
```

(or `< 2`, whichever you prefer)

client/src/views/header/HeaderPage.vue
```diff
@@ -94,6 +94,30 @@
...
+                <span
+                  class="security-level"
+                  :class="{ 'security-level--done': securityWarningsCount.length > 2 }"
+                />
```
Ironicbay: Third one will never be done, we hide the button if it is
```suggestion
                <span class="security-level" />
```

client/src/views/header/HeaderPage.vue
```diff
@@ -94,6 +94,29 @@
...
+                  class="security-level"
+                  :class="{ 'security-level--done': securityWarningsCount < 3 }"
+                />
+                <span
+                  class="security-level"
+                  :class="{ 'security-level--done': securityWarningsCount <= 1 }"
+                />
+                <span
+                  class="security-level"
+                  :class="{ 'security-level--done': securityWarningsCount === 0 }"
```
Ironicbay: You can remove this line, this check will never be true

client/src/views/header/HeaderPage.vue
```diff
@@ -94,6 +94,29 @@
...
+            >
+              <ion-icon :icon="checkmarkCircle" />
+              <div class="checklist-security-levels">
+                <span
+                  class="security-level"
+                  :class="{ 'security-level--done': securityWarningsCount < 3 }"
+                />
+                <span
+                  class="security-level"
+                  :class="{ 'security-level--done': securityWarningsCount <= 1 }"
```
Ironicbay: You can use `< 2` or `=== 1` since the only possible cases here can be 1, 2 or 3

client/src/views/header/HeaderPage.vue
```diff
@@ -94,6 +94,29 @@
...
+              class="topbar-right-button__item unread"
+              @click="openSecurityWarningsModal()"
+              ref="checklistSecurityButton"
+            >
+              <ion-icon :icon="checkmarkCircle" />
+              <div class="checklist-security-levels">
+                <span
+                  class="security-level"
+                  :class="{ 'security-level--done': securityWarningsCount < 3 }"
+                />
```
Ironicbay: For non-workspace owners, you can add this condition on the first `span`
```suggestion
                <span
                  v-if="securityWarnings.isWorkspaceOwner"
                  class="security-level"
                  :class="{ 'security-level--done': securityWarningsCount < 3 }"
                />
```

The style will need tweaking

client/src/views/header/HeaderPage.vue
fabienSvtr: I think `< 2` is better because if we want to add a task, at least the condition will be ok.

client/src/views/header/HeaderPage.vue
```diff
@@ -94,6 +94,33 @@
...
+                <span
+                  class="security-level"
+                  :class="{
+                    'security-level--done': securityWarnings.isWorkspaceOwner ? securityWarningsCount <= 1 : securityWarningsCount < 1,
+                  }"
+                />
+                <span
+                  v-if="securityWarnings.isWorkspaceOwner"
+                  class="security-level"
+                />
```
Ironicbay: This is more complicated than it needs to be. The first `<span>` is only wanted if there are 3 potential warnings. So you can add a `v-if="securityWarnings?.isWorkspaceOwner"` to it.
The second and third can then be the same whatever the situation: you'll have one span active if there is at most one security warning, and the last one is just forever not `done`.

See my suggestion: 
```suggestion
                <span
                  v-if="securityWarnings?.isWorkspaceOwner"
                  class="security-level"
                  :class="{ 'security-level--done': securityWarningsCount < 3 }"
                />
                <span
                  class="security-level"
                  :class="{ 'security-level--done': securityWarningsCount < 2 }"
                />
                <span class="security-level" />
```

PR #10951, opened 2025-08-19
feat(misc): Add script to extract important dependencies versions

FirelightFlagboy:


misc/list-important-dependencies.py
```diff
@@ -0,0 +1,374 @@
+#!/usr/bin/env python3
+# Parsec Cloud (https://parsec.cloud) Copyright (c) BUSL-1.1 2016-present Scille SAS
+
+"""
+List what we consider important dependencies of our client & server application.
```
AureliaDolo:
```suggestion
List versions of what we consider important dependencies of our client & server application.
```

AureliaDolo (approved):

misc/list-important-dependencies.py
```diff
@@ -0,0 +1,374 @@
+#!/usr/bin/env python3
```
touilleMan: Please, use snake_case instead of kebab-case for naming python scripts :pray:

misc/list-important-dependencies.py
```diff
@@ -0,0 +1,374 @@
...
+        ]
+    )
+
+    deps_from_lock = list_dependencies_from_toml_file(
+        PYTHON_LOCK_FILE, PoetryLock, IMPORTANT_PYTHON_DEPS
+    )
+    deps.update(deps_from_lock.dependencies)
+
+    if deps_from_lock.not_found:
+        log.warning(
```
touilleMan: I think we should raise an error here instead: if the dependency is missing it means this script is broken and must be fixed.
```suggestion
        assert not deps_from_lock.not_found, f"Some expected dependencies where not found in lock file {', '.join(deps_from_lock.not_found)}"
```

misc/list-important-dependencies.py
```diff
@@ -0,0 +1,374 @@
...
+    ROOT_DIR / "client/package-lock.json",
+    ROOT_DIR / "client/electron/package-lock.json",
+)
+CARGO_LOCK_FILE = ROOT_DIR / "Cargo.lock"
+# We do not include the lock file for the documentations as it's not important for the application.
+PYTHON_LOCK_FILE = ROOT_DIR / "server/poetry.lock"
+RUST_TOOLCHAIN_FILE = ROOT_DIR / "rust-toolchain.toml"
+VERSION_FILE = ROOT_DIR / "misc/versions.toml"
+
+LOG_LEVEL = logging.DEBUG
```
touilleMan: nitpick: logging module is cumbersome and doesn't provide much compared to just using a print.

Typically here:
- `LOG_LEVEL` is hardcoded, so it cannot changed without modifying the script
- sub-logger can in theory be used to obtain only a subset of the logs, but here again the config is hardcoded so this has no use.
- The amount of logs is going to be very low anyway

So I would argue for a different philosophy:
- only add print for thing that have actual value under normal use (here I guess it would be nothing since the output the user is interested in is a csv written to a file whose path as been specified by the user in the first place)
- when the script has to be debugged, the dev just add additional temporary prints wherever he needs
- (not needed here) if a more verbose output might be of value under certain cases, add a `--verbose` option to the script to enable/disable certain prints

This means simpler code and simpler output for the end user ;-)

misc/list_important_dependencies.py
```diff
@@ -0,0 +1,364 @@
...
+        "electron-builder",
+        "electron-builder-squirrel-windows",
+        "electron-is-dev",
+        "electron-updater",
+        "electron-window-state",
+        "regedit",
+        "typescript",
+    )
+
+    deps = set([Dependency("nodejs", versions["node"], DependencyType.Javascript)])
```
touilleMan: nitpic: `set([1, 2])` can also be written as `{1, 1}`

touilleMan (approved):

misc/list_important_dependencies.py
```diff
@@ -0,0 +1,364 @@
...
+def write_deps(csv: csv.DictWriter[Dependency], deps: set[Dependency]):
+    deps_list = list(deps)
+    deps_list.sort(key=lambda dep: dep.name)
+    csv.writerows(DependencyModel(dep).model_dump() for dep in deps_list)
+
+
+def parse_args() -> argparse.Namespace:
+    parser = argparse.ArgumentParser()
+
+    parser.add_argument("--output", default=Path("versions.csv"), type=Path)
```
touilleMan, line 91: nitpick: default output file name is not displayed when using `--help`, nor is printed when running the command:
```shell
$ python misc/list_important_dependencies.py
Loading client/package-lock.json content
Loading client/electron/package-lock.json content
Loading rust-toolchain.toml content
Loading Cargo.lock content
Loading server/poetry.lock content
```

Since you have already put some `Loading .... content` prints, I think the consistent thing to do is to add a final `Writting versions CSV ./output.csv` print

misc/list_important_dependencies.py
FirelightFlagboy: don

misc/list_important_dependencies.py
FirelightFlagboy: done

PR #10960, opened 2025-08-20
chore(rust): Bump to rust-1.87

FirelightFlagboy:
- Fix some enum with to large variant where easily possible
- Clippy auto-fix


libparsec/crates/testbed/src/template/events.rs
```diff
@@ -141,6 +141,7 @@ macro_rules! single_certificate_event {
  */
 
 #[derive(Serialize, Deserialize, Debug, Clone)]
+#[expect(clippy::large_enum_variant)]
```
touilleMan, line 144: :+1:

libparsec/src/handle.rs
```diff
@@ -10,7 +10,7 @@ pub type Handle = u32;
 const INVALID_HANDLE_ERROR_MSG: &str = "Invalid Handle";
 
 enum RegisteredHandleItem {
-    Open(HandleItem),
+    Open(Box<HandleItem>),
```
touilleMan, line 13: RAM manufacturers hate him: A young software engineer has found a way to reduce memory leak from 500 to only 8 octets !

libparsec/crates/types/src/certif.rs
```diff
@@ -108,6 +108,13 @@ macro_rules! impl_unsecure_load {
                 pub fn hint(&self) -> String {
                     format!("{:?}", self.unsecure)
                 }
+
+                /// Convert to it's verified variant by check its signature
```
touilleMan:
```suggestion
                /// Convert to its verified variant by check its signature
```

To be honest I'm not convinced by what added information this docstring brings (since the signature makes it already clear that a `UnsecureFoo` is turned into a `Foo` by doing a `verify_signature` ...), however I don't mind keeping it if you think otherwise :+1:

touilleMan (approved):

PR #10961, opened 2025-08-20
chore(rust): Bump rust to 1.88

FirelightFlagboy:
- Clippy included an auto-fix for format string parameter


touilleMan (approved):
:tada:

PR #10977, opened 2025-08-26
[MS] Improve UI and add back button (Create account)

### LINKED ISSUE #10983: [MS] Parsec Account - UI UX issues
### Parsec version

3.5.0

- [ ] Any
- [x] macOS
- [ ] Linux
- [ ] Windows
- [x] Web

### Bug description

- Subtitle too small
- Several buttons are not with the right height
- Missing back button when creating the account

Image

Image

Image

### Reproduction steps

```bash
1.
2.
3.
...
```

### Relevant output:

_No response_

fabienSvtr:
https://github.com/user-attachments/assets/0136713a-8873-4121-8dda-156bcf339d9d

closes #10983

### DIFF
--- a/client/src/locales/en-US.json
+++ b/client/src/locales/en-US.json
@@ -1674,6 +1674,7 @@
             "alreadyHaveAnAccount": "Already have an account?",
             "loginButton": "Log in",
             "nextButton": "Next",
+            "sendCode": "Send a validation code",
             "resendCode": "Resend code",
             "welcome": "Welcome to Parsec",
             "welcomeDescription": "Your Parsec Share account has been created successfully. You now have access to Parsec.",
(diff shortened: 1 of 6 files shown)


client/src/locales/en-US.json
```diff
@@ -1674,6 +1674,7 @@
             "alreadyHaveAnAccount": "Already have an account?",
             "loginButton": "Log in",
             "nextButton": "Next",
+            "sendCode": "Send a validation code",
```
Ironicbay, line 1677: Why not keep it at `Next`? Wouldn't it be confusing for some users to have such text when they expect to just go forward with the process?

fabienSvtr: In order to be more precise about what is accepted afterwards.

PR #10987, opened 2025-08-27
[MS] Edit server address in account login

### LINKED ISSUE #10986: Disabled server address by default
Image

fabienSvtr:
image
image

closes #10986


client/tests/e2e/specs/account_create.spec.ts
```diff
@@ -39,6 +39,12 @@ msTest('Parsec account create account', async ({ parsecAccount }) => {
   const userInfoNext = userInfoContainer.locator('.account-login-content-button__item');
   await expect(userInfoNext).toHaveText('Next');
   await expect(userInfoNext).toBeTrulyDisabled();
+  await expect(container.locator('.input-edit-button')).toBeVisible();
+  await expect(container.locator('.input-edit-button')).toHaveText('Edit');
+  await expect(inputContainers.nth(0).locator('ion-input')).toHaveClass(/(^|\s)login-server-input-disabled(\s|$)/);
```
Max-7: You shuld be able to use `toHaveTheClass('login-server-input-disabled')`

client/tests/e2e/specs/account_create.spec.ts
```diff
@@ -39,6 +39,12 @@ msTest('Parsec account create account', async ({ parsecAccount }) => {
...
+  await expect(inputContainers.nth(0).locator('ion-input')).toHaveAttribute('disabled');
+  await parsecAccount.locator('.create-account-page').locator('.input-edit-button').click();
+  await expect(inputContainers.nth(0).locator('ion-input')).not.toBeDisabled();
```
Max-7: Nop, should be the opposite of the previous condition : not have the class and not have the attribute.

Max-7 (dismissed):

mmmarcos:
@fabienSvtr if it's not too complicated, it would be nice to have something like https://vikunja.canine.tools/login

client/src/views/account/RecoverAccountPage.vue
```diff
@@ -436,6 +470,11 @@ async function backToLogin(): Promise<void> {
   box-shadow: var(--parsec-shadow-soft);
   padding: 2rem;
   gap: 1rem;
+  position: relative;
+
+  .input-edit-button {
+    right: 2.25rem;
+  }
```
Ironicbay, line 483: I guess this css class interacts with the one in `_inputs.scss`, and the button does nothing as a result

Ironicbay (commented):
When `isEditingServer` is set as `true`, the button text remains as `Edit`, is greyed out and hovering doesn't show a pointer, however we can still interact with it to confirm the input.

Two options:s
- Change the text to `Confirm` and add a `cursor: pointer` + underline on hover
- Change the text to `Cancel`, restore the default server address on click, and add `cursor: pointer` + underline on hover

client/src/views/account/RecoverAccountPage.vue
fabienSvtr: No, we need it on the RecoveryAccountPage because the structure styling is not done in the same way.

Ironicbay: What I meant was that the `Edit` button does nothing on this page

fabienSvtr: I edit the page to fix it.

PR #10994, opened 2025-08-28
[MS] Fix sidebar menu visibility management for file viewers

fabienSvtr:
## Describe the issue
The existing problem:
- A file is uploaded (visible sidebar ✅ )
- It is opened with the viewer (invisible sidebar ✅ ) <-- should not be visible if the user has not performed any actions
- The page is reloaded
- The user goes to the workspaces page (invisible sidebar :x:)

https://github.com/user-attachments/assets/fff1d980-9f3e-42fc-9d00-b19e1d1a749e

### DIFF
--- a/client/src/services/sidebarMenu.ts
+++ b/client/src/services/sidebarMenu.ts
@@ -1,38 +1,123 @@
 // Parsec Cloud (https://parsec.cloud) Copyright (c) BUSL-1.1 2016-present Scille SAS
 
-import { ref } from 'vue';
+import { StorageManager, StorageManagerKey } from '@/services/storageManager';
+import { SIDEBAR_MENU_DATA_KEY, SidebarDefaultData, SidebarSavedData } from '@/views/menu';
+import { computed, ComputedRef, inject, Ref, ref } from 'vue';
 
-const defaultWidth = 300;
-const hiddenWidth = 0;
-const computedWidth = ref<number>(defaultWidth);
-const storedWidth = ref<number>(defaultWidth);
+const DEFAULT_WIDTH = 300;
+const HIDDEN_WIDTH = 0;
+const computedWidth = ref<number>(DEFAULT_WIDTH);
+const storedWidth = ref<number>(DEFAULT_WIDTH);
+const isInitialized = ref<boolean>(false);
+const isVisible = computed(() => {
+  return !isInitialized.value ? false : computedWidth.value > HIDDEN_WIDTH;
+});
 
-export default function useSidebarMenu(): any {
-  function isVisible(): boolean {
-    return computedWidth.value > 4;
-  }
+interface SidebarMenu {
+  width: Ref<number>;
+  isVisible: ComputedRef<boolean>;
+  reset: (persist?: boolean) => void;
+  hide: (persist?: boolean) => void;
+  show: (persist?: boolean) => void;
+  setWidth: (width: number, persist?: boolean) => void;
+}
 
-  function reset(): void {
-    computedWidth.value = storedWidth.value;
-  }
+export default function useSidebarMenu(): SidebarMenu {
+  const storageManager: StorageManager | null = inject(StorageManagerKey, null);
 
-  function hide(): void {
-    if (isVisible()) {
-      storedWidth.value = computedWidth.value;
-      computedWidth.value = hiddenWidth;
+  // Start async initialization
+  init()
+    .then()
+    .catch((error) => {
+      console.warn('Failed to initialize sidebar menu:', error);
+    })
+    .finally(() => {
+      isInitialized.value = true;
+    });
+
+  async function init(): Promise<void> {
+    if (!storageManager) {
+      return;
+    }
+
+    try {
+      const savedData = await storageManager.retrieveComponentData<SidebarSavedData>(SIDEBAR_MENU_DATA_KEY, SidebarDefaultData);
+
+      const savedWidth = savedData.width || DEFAULT_WIDTH;
+      const isVisible = savedData.visible !== undefined ? savedData.visible : true; // Default to visible
+
+      if (isVisible) {
+        computedWidth.value = savedWidth;
+        storedWidth.value = savedWidth;
+      } else {
+        computedWidth.value = HIDDEN_WIDTH;
+        storedWidth.value = savedWidth;
+      }
+    } catch (error) {
+      console.warn('Failed to load sidebar state from storage:', error);
+      // Default to visible state with default width on storage errors
+      computedWidth.value = DEFAULT_WIDTH;
+      storedWidth.value = DEFAULT_WIDTH;
     }
   }
 
-  function show(): void {
+  async function _saveToStorage(): Promise<void> {
+    if (!storageManager) {
+      return;
+    }
+
+    try {
+      await storageManager.updateComponentData<SidebarSavedData>(
+        SIDEBAR_MENU_DATA_KEY,
+        {
+          width: computedWidth.value <= HIDDEN_WIDTH ? storedWidth.value : computedWidth.value,
+          visible: isVisible.value,
+        },
+        SidebarDefaultData,
+      );
+    } catch (error) {
+      console.warn('Failed to save sidebar state to storage:', error);
+    }
+  }
+
+  function reset(persist = true): void {
+    computedWidth.value = DEFAULT_WIDTH;
+    storedWidth.value = DEFAULT_WIDTH;
+    if (persist) {
+      _saveToStorage();
+    }
+  }
+
+  function hide(persist = true): void {
+    if (isVisible.value) {
+      storedWidth.value = computedWidth.value;
+      computedWidth.value = HIDDEN_WIDTH;
+      if (persist) {
+        _saveToStorage();
+      }
+    }
+  }
+
+  function show(persist = true): void {
     computedWidth.value = storedWidth.value;
+    if (persist) {
+      _saveToStorage();
+    }
+  }
+
+  function setWidth(width: number, persist = true): void {
+    computedWidth.value = width;
+    if (persist) {
+      _saveToStorage();
+    }
   }
 
   return {
-    computedWidth,
-    storedWidth,
+    width: computedWidth,
     isVisible,
     reset,
     hide,
     show,
+    setWidth,
   };
 }
--- a/client/tests/e2e/specs/file_viewers.spec.ts
+++ b/client/tests/e2e/specs/file_viewers.spec.ts
@@ -458,3 +458,183 @@ msTest('Text viewer', async ({ documents }) => {
   // Didn't manage to make a better regex, I have no idea why but nothing matches
   await expect(container.locator('.editor-scrollable')).toHaveText(new RegExp('^.*Parsec.*$'));
 });
+
+// File viewer sidebar management tests
+msTest('File viewer auto-hides sidebar on entry', async ({ documents }) => {
+  // Start with visible sidebar
+  await expect(documents.locator('.sidebar')).toBeVisible();
+
+  // Open file viewer
+  const entries = documents.locator('.folder-container').locator('.file-list-item');
+  await entries.nth(2).dblclick();
+  await expect(documents.locator('.ms-spinner-modal')).toBeHidden();
+  await expect(documents).toBeViewerPage();
+
+  // Sidebar should be auto-hidden in file viewer
+  await expect(documents.locator('.sidebar')).toBeHidden();
+  await expect(documents.locator('#connected-header .topbar')).toBeHidden();
+});
+
+msTest('File viewer sidebar can be toggled in large display', async ({ documents }) => {
+  const entries = documents.locator('.folder-container').locator('.file-list-item');
+  await entries.nth(2).dblclick();
+  await expect(documents).toBeViewerPage();
+
+  // Sidebar initially hidden
+  await expect(documents.locator('.sidebar')).toBeHidden();
+
+  const toggleButton = documents.locator('.file-handler-topbar-buttons__item.toggle-menu');
+  const toggleSidebarButton = documents.locator('#trigger-toggle-menu-button');
+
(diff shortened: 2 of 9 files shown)


client/tests/e2e/specs/file_viewers.spec.ts
```diff
@@ -458,3 +458,121 @@ msTest('Text viewer', async ({ documents }) => {
...
+    await recentDocs.nth(1).click();
+    await expect(documents.locator('.ms-spinner-modal')).toBeHidden();
+
+    // Sidebar should remain visible
+    await expect(documents.locator('.sidebar')).toBeVisible();
+    await expect(documents.locator('#connected-header .topbar')).toBeVisible();
+  }
+});
+
+msTest('File viewer restores previous sidebar state on exit', async ({ documents }) => {
```
Max-7: Is the other state tested? If the sidebar is visible, is it correctly restored when exiting the viewer?

client/tests/e2e/specs/file_viewers.spec.ts
```diff
@@ -458,3 +458,121 @@ msTest('Text viewer', async ({ documents }) => {
...
+  const isToggleVisible = await toggleButton.isVisible();
+
+  if (!isToggleVisible) {
+    // Small display might handle sidebar differently
+    await expect(documents.locator('.sidebar')).toBeHidden();
+  } else {
+    // If toggle is available, test it works in small display
+    await toggleButton.click();
+    // Behavior might be different in small display (modal overlay vs sidebar)
+  }
```
Max-7: Missing at least one expect(). Nothing is really tested.

client/tests/e2e/specs/sidebar.spec.ts
```diff
@@ -321,3 +321,79 @@ msTest('Recent and pinned workspaces are updated when workspace is renamed', asy
...
+
+  // Hide sidebar in large display
+  await toggleButton.click();
+  await expect(sidebar).toBeHidden();
+
+  // Switch to small display
+  await documents.setDisplaySize(DisplaySize.Small);
+
+  // In small display, sidebar behavior might be different
+  // The toggle button might be hidden or behave differently
```
Max-7: Why? It should be consistent, else it's a bug.

client/tests/e2e/specs/file_viewers.spec.ts
```diff
@@ -458,3 +458,121 @@ msTest('Text viewer', async ({ documents }) => {
...
+  await expect(documents.locator('#connected-header .topbar')).toBeVisible();
+
+  await toggleSidebarButton.click();
+  await expect(documents.locator('.sidebar')).toBeVisible();
+
+  // Navigate to another file via sidebar
+  const sidebar = documents.locator('.sidebar');
+  const recentDocs = sidebar.locator('#sidebar-files').locator('.list-sidebar-content').getByRole('listitem');
+
+  if ((await recentDocs.count()) > 1) {
```
Max-7: I don't understand this check. The test should be consistent, it shouldn't have multiple files at one time, 0 at another, or else there's a bug in the app or the test is not written properly.

client/tests/e2e/specs/file_viewers.spec.ts
```diff
@@ -458,3 +458,121 @@ msTest('Text viewer', async ({ documents }) => {
...
+    await recentDocs.nth(1).click();
+    await expect(documents.locator('.ms-spinner-modal')).toBeHidden();
+
+    // Sidebar should remain visible
+    await expect(documents.locator('.sidebar')).toBeVisible();
+    await expect(documents.locator('#connected-header .topbar')).toBeVisible();
+  }
```
Max-7: Nothing gets tested if the initial condition doesn't match.

client/tests/e2e/specs/sidebar_persistence.spec.ts
Max-7, line 1: This can be moved to sidebar.spec.ts, we don't need two test files for the sidebar.

client/tests/e2e/specs/sidebar_persistence.spec.ts
```diff
@@ -0,0 +1,65 @@
...
+async function toggleSidebar(page: MsPage): Promise<void> {
+  // Look for toggle button in standard locations - simplified approach
+  const toggleButton = page.locator('#trigger-toggle-menu-button').first();
+
+  // Wait for button to be available and click it
+  await toggleButton.waitFor({ state: 'visible', timeout: 10000 });
+  await toggleButton.click();
+}
+
+msTest('Sidebar state persists across page navigation', async ({ documents, workspaces }) => {
```
Max-7: I would advice against using two fixtures. I don't know if playwright will consider them only as aliases or if can cause some bugs in the plumping.

client/tests/e2e/specs/sidebar_persistence.spec.ts
```diff
@@ -0,0 +1,65 @@
...
+  const sidebar = page.locator('.sidebar');
+  return await sidebar.isVisible();
+}
+
+async function toggleSidebar(page: MsPage): Promise<void> {
+  // Look for toggle button in standard locations - simplified approach
+  const toggleButton = page.locator('#trigger-toggle-menu-button').first();
+
+  // Wait for button to be available and click it
+  await toggleButton.waitFor({ state: 'visible', timeout: 10000 });
```
Max-7: Why not just `await expect(toggleButton).toBeVisible()`? Also 10s timeout seems way overkill. As a general rule: if you have to rely on timeouts, the test is going to be flaky.

client/tests/e2e/specs/sidebar_persistence.spec.ts
```diff
@@ -0,0 +1,65 @@
...
+
+  // Navigate back to documents
+  await workspaces.locator('.workspace-card-item').nth(0).click();
+  await expect(documents.locator('.folder-container')).toBeVisible();
+
+  // Verify sidebar still hidden
+  await expect(documents.locator('.sidebar')).toBeHidden();
+});
+
+msTest('Sidebar loading state prevents flash on page load', async ({ documents }) => {
```
Max-7: Two things on this test:
- you can use the fixture `workspaces` instead of document, so we don't lose time importing documents that are not used (probably true for some other tests too)
- it doesn't do what you think: reloading the page while logged in will (at least it should) return you to the homepage since the handle is no longer valid. The test "works" because you're expecting the sidebar to be hidden, which it technically is since it's not in the DOM. In dev it doesn't happen because we set the variable `PARSEC_APP_TESTBED_AUTO_LOGIN` that will create a new handle when using the testbed. If it does the same thing with playwright, it's a bug and should be fixed (somewhere here https://github.com/Scille/parsec-cloud/blob/0531c5ee1b0544dc1912324cea1fc5ccfc82668a/client/src/views/layouts/DevLayout.vue#L45, auto-login should not be activated while testing).

client/src/services/sidebarMenu.ts
```diff
@@ -1,38 +1,125 @@
...
-const storedWidth = ref<number>(defaultWidth);
+const DEFAULT_WIDTH = 300;
+const HIDDEN_WIDTH = 0;
+const computedWidth = ref<number>(DEFAULT_WIDTH);
+const storedWidth = ref<number>(DEFAULT_WIDTH);
+const isLoaded = ref<boolean>(false);
+const isVisible = computed(() => {
+  return !isLoaded.value ? false : computedWidth.value > HIDDEN_WIDTH;
+});
+let isInitialized = false;
```
Max-7: I'm a bit unsure with this. Do we want to initialize those variables everytime `useSidebarMenu` is called, or only the first time? It seems weird to have functions living in the scope of another function `useSidebarMenu` but working on global variables.

Max-7 (changes requested):
First test I've done fails: login on the workspace page, hide the sidebar, refresh the page, the sidebar is visible.

client/src/services/sidebarMenu.ts
NicoTuxx: Did you get problems / bugs due to this? Because it was working for me like this is now, I don't really now how to change it without risking the service to cause problems again, suggestions?

NicoTuxx: As discussed on slack, could not do an other way, we agreed to stay like this for now.

client/tests/e2e/specs/file_viewers.spec.ts
NicoTuxx: Changed it.

client/tests/e2e/specs/file_viewers.spec.ts
NicoTuxx: Done

client/tests/e2e/specs/sidebar.spec.ts
NicoTuxx: Changed it.

client/tests/e2e/specs/file_viewers.spec.ts
NicoTuxx: Done

client/tests/e2e/specs/sidebar_persistence.spec.ts
NicoTuxx: removed it since what I was trying to test can not happen.

client/tests/e2e/specs/sidebar_persistence.spec.ts
NicoTuxx: Done

client/src/services/sidebarMenu.ts
```diff
@@ -1,38 +1,125 @@
...
 
-const defaultWidth = 300;
-const hiddenWidth = 0;
-const computedWidth = ref<number>(defaultWidth);
-const storedWidth = ref<number>(defaultWidth);
+const DEFAULT_WIDTH = 300;
+const HIDDEN_WIDTH = 0;
+const computedWidth = ref<number>(DEFAULT_WIDTH);
+const storedWidth = ref<number>(DEFAULT_WIDTH);
+const isLoaded = ref<boolean>(false);
```
Max-7: Can't you use isInitialized for this? You have two variables with a similar meaning.

client/src/services/sidebarMenu.ts
```diff
@@ -1,38 +1,125 @@
...
+    .finally(() => {
+      isLoaded.value = true;
+    });
+
+  async function init(): Promise<void> {
+    if (isInitialized || !storageManager) {
+      return;
+    }
+
+    isInitialized = true;
```
Max-7: Technically it should only be initialized at the end, not right at the start.

client/tests/e2e/helpers/utils.ts
```diff
@@ -76,7 +76,7 @@ export async function fillInputModal(root: Locator | Page, text: string, clear?:
   const okButton = modal.locator('.ms-modal-footer-buttons').locator('#next-button');
   await fillIonInput(modal.locator('ion-input'), text);
   await okButton.click();
-  await expect(modal).toBeHidden();
+  await expect(modal).toBeHidden({ timeout: 10000 });
```
Max-7: Timeouts should not be changed. If something doesn't show in the default 5s, the test is flaky.

client/tests/e2e/specs/file_viewers.spec.ts
```diff
@@ -458,3 +458,186 @@ msTest('Text viewer', async ({ documents }) => {
...
+  // Show header and sidebar
+  await toggleButton.click();
+  await expect(documents.locator('#connected-header .topbar')).toBeVisible();
+
+  await toggleSidebarButton.click();
+  await expect(documents.locator('.sidebar')).toBeVisible();
+
+  // Go back to documents page and open a second file to ensure we have 2 recent files
+  // The back button might be in different locations depending on menu state
+  let backButton = documents.locator('.topbar-left-content').locator('.back-button');
```
Max-7: Why? I don't get why the button is moving around. If there's randomness, there's a bug.

client/tests/e2e/specs/file_viewers.spec.ts
```diff
@@ -458,3 +458,186 @@ msTest('Text viewer', async ({ documents }) => {
...
+  // If not found, try the file handler topbar location
+  if (!(await backButton.isVisible())) {
+    backButton = documents.locator('.file-handler-topbar').locator('.back-button');
+  }
+
+  await backButton.click();
+  await expect(documents.locator('.folder-container')).toBeVisible();
+
+  // Wait a moment for the first file to be registered in recent documents
+  await documents.waitForTimeout(500);
```
Max-7: Use expect() to check if recent documents contain what you want. The expect() will retry (for 5s) until it finds the expected state.

client/tests/e2e/specs/file_viewers.spec.ts
```diff
@@ -458,3 +458,186 @@ msTest('Text viewer', async ({ documents }) => {
...
+
+  // Open second file
+  await entries.nth(3).dblclick();
+  await expect(documents).toBeViewerPage();
+  await expect(documents.locator('.ms-spinner-modal')).toBeHidden();
+
+  // Wait for file to be fully loaded and registered
+  await documents.waitForTimeout(500);
```
Max-7: Use expect to check for the state.

client/tests/e2e/specs/sidebar.spec.ts
Max-7, line 1: Didn't you have a test to check the sidebar after a reload?

client/tests/e2e/specs/file_viewers.spec.ts
NicoTuxx: It's just a wrong comment coming from a code suggestion that I forgot to remove.
We have 2 back buttons (one in header, one in file handler topbar when the header is hidden), the confusion comes from that.

client/tests/e2e/specs/sidebar.spec.ts
NicoTuxx: added some new

Max-7 (approved):

PR #10995, opened 2025-08-28
[MS] Hide update button for external users

### LINKED ISSUE #10743: User name are not displayed for file last updater or file creator for external users
Since external users do not have access to user names, a raw id is displayed instead, which is not very elegant.

fabienSvtr:
image
image

closes #10743


client/src/components/files/explorer/FileListDisplay.vue
```diff
@@ -126,14 +131,15 @@ import FileListItemProcessing from '@/components/files/explorer/FileListItemProc
...
-import { FsPath, WorkspaceRole } from '@/parsec';
+import { FsPath, WorkspaceRole, UserProfile } from '@/parsec';
 import { IonText, IonList, IonLabel, IonListHeader, IonIcon } from '@ionic/vue';
 import { arrowUp, arrowDown } from 'ionicons/icons';
 import { computed, useTemplateRef } from 'vue';
 import { MsCheckbox, useWindowSize, MsSorterChangeEvent } from 'megashark-lib';
 
 const { isLargeDisplay, isSmallDisplay } = useWindowSize();
 const props = defineProps<{
+  profile: UserProfile;
```
Ironicbay: I'd suggest renaming every new `profile` variable/prop from this PR to `ownProfile` to make it clearer, as it's done in a few places in the app already.

PR #11018, opened 2025-09-01
Prevent client using API < 5.2 from using `invite_(claimer|greeter)_step` commands

### LINKED ISSUE #11017: Prevent client using API 5.1 to enroll with API 5.2+
Once https://github.com/Scille/parsec-cloud/issues/11015 is merged, patched and unpatched clients become incompatible in the enrollment process.

The simple fix is to prevent any client using API < 5.2 from using `invite_(claimer|greeter)_step` commands.
Typically the server should reject early on the old client's attempt with a `UnsupportedApiVersion` error code (that should hopefuly lead to a dialogue in the GUI inviting the user to upgrade Parsec).

touilleMan:
closes #11017

### DIFF
--- a/server/parsec/components/invite.py
+++ b/server/parsec/components/invite.py
@@ -12,6 +12,7 @@ from enum import auto
 from jinja2 import Environment
 
 from parsec._parsec import (
+    ApiVersion,
     CancelledGreetingAttemptReason,
     DateTime,
     DeviceID,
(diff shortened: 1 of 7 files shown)


server/parsec/components/invite.py
```diff
@@ -46,6 +47,9 @@
 InviteListInvitationCreatedBy = authenticated_cmds.latest.invite_list.InvitationCreatedBy
 
 
+NEW_SAS_CODE_ALGORITHM_INTRODUCED_IN_VERSION = ApiVersion(version=5, revision=2)
```
FirelightFlagboy, line 50: **nitpick:** Add a short comment about what differ from the previous version (i.e.: how we generate the code that make client using an older version incompatible)

FirelightFlagboy (commented):
issue: The issue of using incompatible clients during invitation still persiste if they connect to an older server that does not have such verification step, But it's nothing we could do (else introduce a client version in invitation field :thinking:)

PR #11024, opened 2025-09-03
[MS] Fix file handler refresh

Max-7:


client/src/views/files/handler/FileHandler.vue
```diff
@@ -239,15 +239,18 @@ const cancelRouteWatch = watchRoute(async () => {
...
 
   // Same file, no need to reload
   if (contentInfo.value && contentInfo.value.path === getDocumentPath() && atDateTime.value?.toMillis() === timestamp) {
-    const fileHandlerMode = getFileHandlerMode();
     if (fileHandlerMode === handlerMode.value) {
       return;
     }
     handlerMode.value = getFileHandlerMode();
+  } else if (fileHandlerMode !== handlerMode.value) {
+    handlerMode.value = getFileHandlerMode();
```
Ironicbay: Why not assign the value of `fileHandlerMode` instead of calling `getFileHandlerMode()` again? (same for line 251)

Ironicbay (approved):

client/src/views/files/handler/FileHandler.vue
Max-7: Yep, you're right. I refactored it a bit in the process.

PR #11036, opened 2025-09-04
[MS] Tried to preload cryptpad to maybe gain a few seconds when opening a file

Max-7:
I thought that by loading preloading a file, some things would stay in cache and the opening would later be faster, but no. Keeping it just in case.


Max-7:
No point in keeping this open.

PR #11046, opened 2025-09-04
Move `libparsec_crypto`'s tests to `tests/unit` and make them compatible with wasm32

touilleMan:


libparsec/crates/crypto/tests/unit/mod.rs
```diff
@@ -0,0 +1,22 @@
...
+mod private;
+mod sas;
+mod secret;
+mod sequester;
+mod sign;
+mod utils;
+
+#[cfg(target_arch = "wasm32")]
+mod platform {
+    wasm_bindgen_test::wasm_bindgen_test_configure!(run_in_browser);
```
FirelightFlagboy: **todo (blocking):** Should also test in `shared_worker`

FirelightFlagboy (approved):

libparsec/crates/crypto/tests/unit/mod.rs
touilleMan: Not sure what you mean... is there an option to enable in the wasm_bindgen configuration ?

FirelightFlagboy: Yes, else we only run the test in a browser context, we also want to run them in a shared worker

touilleMan: done :+1:

PR #11050, opened 2025-09-04
Fix RustCrypto's incorrect implementation of `blake2b_hash`

touilleMan:


FirelightFlagboy:
Need rebase and it does not seems that we have a test for `blake2b_hash`

touilleMan:
rebased, and I've added one additional test directly testing `blake2b_hash` for extra sanity ^^

FirelightFlagboy (approved):

PR #11072, opened 2025-09-08
Fix `Account::create_auth_method()` vault key handling

touilleMan:
`Account::create_auth_method()` was incorrectly uploading a new vault key instead of using the existing one


libparsec/crates/account/tests/unit/create_auth_method.rs
```diff
@@ -60,20 +60,27 @@ async fn ok(#[values("master_secret", "password")] kind: &str, env: &TestbedEnv)
...
+    let account2 = Account::login(
         env.discriminant_dir.clone(),
         ProxyConfig::default(),
         env.server_addr.clone(),
         login_strategy,
     )
     .await
     .unwrap();
+
+    // Also make sure both auth methods agree on the vault key
```
FirelightFlagboy, line 72: **note:** I wouldn't say "agreed", more like "under new management"

FirelightFlagboy (approved):

PR #11089, opened 2025-09-10
Optimize vlob_poll_changes's SQL query to correctly make use of `realm_vlob_update`'s unique index

touilleMan:


touilleMan:
> why a left join can be replaced by an inner join here.

tl;dr: this is a small style adjustment, the real fix is the condition inversion in the `WHERE`

long version:

`a LEFT JOIN b where a.x = b.y` means all rows from `a` will end up in the final result, even if there is no `b` that fulfill `a.x = b.y`
`a UNION JOIN b where a.x = b.y` means only rows fulfilling `a.x = b.y` will end up in the result

However the key point here is that the condition `realm_vlob_update.vlob_atom = vlob_atom._id` is always fulfilled for one and only one couple of rows (i.e. each `realm_vlob_update` refers to a single `vlob_atom`)
So in this case `LEFT JOIN` and `UNION JOIN` are equivalent, however I find `UNION JOIN` to be more explicit on the desired outcome (i.e. `UNION JOIN` means "we want each returned row to be based on a`vlob_atom` and a `realm_vlob_update` row", while `LEFT JOIN` would mean "we accept that some returned row be based on a`realm_vlob_update` row and no `vlob_atom` row")

Issue #11107, opened 2025-09-11
[MS] Cancelling a replacing file import deletes the original file


Ironicbay:
### Parsec version

3.5.0-a9

### Which platform(s) you have observed the problem

- [ ] Any
- [x] macOS
- [ ] Linux
- [ ] Windows
- [ ] Web

### Bug description

When replacing `fileName.ext` with another `fileName.ext`, the user is prompted to replace or cancel the import.

When choosing to replace, if the user then cancels the import, the original "full" file is then deleted, and what remains is only what has been imported before the cancellation instead of the usable original file.

### Reproduction steps

```bash
1. Import a duplicate within a workspace (same file name as one already present)
2. Click `Replace`
3. Cancel the import
4. The remaining file has a lower size than the original and is unusable
```

Issue #10054, opened 2025-03-31
[🐛 | Bug]: Drag & drop a file from Finder to Parsec GUI in same directory loops endlessly


Ironicbay:
### Parsec version tested on:

3.3.2

### Platforms tested on:

MacOS

### Bug description:

To reproduce:
- Navigate to a workspace containing files
- Open current directory in file explorer
- Drag & drop a file from file explorer to Parsec GUI (in the same directory)
- Choose to replace the file when prompted
- Copy / replace never occurs nor finishes and prevents the app from closing

### Relevant output:

The `await reader.read()` [here](https://github.com/Scille/parsec-cloud/blob/master/client/src/services/fileOperationManager.ts#L1015) raises:
```
TypeError: network error
```

mmmarcos:
To be tested on Linux/Windows as it does not seem platform-specific.

PR #11129, opened 2025-09-15
[MS] Fix import cancel & import replace issues

Ironicbay:
Closes https://github.com/Scille/parsec-cloud/issues/11107
Closes https://github.com/Scille/parsec-cloud/issues/10054

### DIFF
--- a/client/src/parsec/file.ts
+++ b/client/src/parsec/file.ts
@@ -55,9 +55,12 @@ export async function rename(
   workspaceHandle: WorkspaceHandle,
   path: FsPath,
   newName: EntryName,
+  canReplace = false,
 ): Promise<Result<FsPath, WorkspaceMoveEntryError>> {
   const newPath = await Path.join(await Path.parent(path), newName);
-  const result = await libparsec.workspaceMoveEntry(workspaceHandle, path, newPath, { tag: MoveEntryModeTag.NoReplace });
+  const result = await libparsec.workspaceMoveEntry(workspaceHandle, path, newPath, {
+    tag: canReplace ? MoveEntryModeTag.CanReplace : MoveEntryModeTag.NoReplace,
+  });
   if (result.ok) {
     return { ok: true, value: newPath };
   }
--- a/client/src/services/fileOperationManager.ts
+++ b/client/src/services/fileOperationManager.ts
@@ -27,6 +27,7 @@ import {
   moveEntry,
   openFile,
   readFile,
+  rename,
   resizeFile,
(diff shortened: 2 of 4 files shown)


client/src/services/fileOperationManager.ts
```diff
@@ -1007,12 +1012,35 @@ class FileOperationManager {
...
+        // Delete the file
+        await deleteFile(data.workspaceHandle, filePath);
+        // Restore file from tmp
+        if (data.needsTmp) {
+          const tmpFilePath = await Path.join(data.path, getTmpFileName(data.file.name));
+          await rename(data.workspaceHandle, tmpFilePath, data.file.name);
+        } else {
+        }
+        // Inform about the cancellation
+        await this.sendState(FileOperationState.FileImported, data);
```
FirelightFlagboy: **question:** Why we use the state `FileImported` here when the read on the file as failed?

client/src/views/files/utils.ts
```diff
@@ -13,6 +13,10 @@ import { modalController, popoverController } from '@ionic/vue';
 import { Answer, askQuestion, I18n, MsModalResult } from 'megashark-lib';
 import { showSaveFilePicker } from 'native-file-system-adapter';
 
+export function getTmpFileName(fileName: string): string {
+  return `._${fileName}`;
```
Max-7: What if a file with this name already exists? It's easier to just use something like `crypto.randomUUID()` to avoid potential collisions. This way we never have to worry about them.

Max-7 (commented):
This seems way more complex than it needs to be, but maybe I'm missing something. I would just go inside `doImport` in the `fileOperationManager.ts` and do

- open a hidden file (with a pattern matching the exclusion, in the right folder) using `crypto.randomUUID()`
- do the whole import with this file, if the import gets cancelled or some error happens, it doesn't matter, the file gets deleted
- at the end, just rename the temporary file into the real file name

Nothing else needs to be updated, it should work for the copy too. The temporary file is used systematically for all imports.

client/src/parsec/file.ts
```diff
@@ -55,9 +55,12 @@ export async function rename(
   workspaceHandle: WorkspaceHandle,
   path: FsPath,
   newName: EntryName,
+  canReplace?: boolean,
```
Max-7: Better to use `canReplace = false` in this case. It doesn't change much but it's always better to restrict the type as much as possible.

Max-7 (approved):

PR #11148, opened 2025-09-17
feat(pki): Sign message using cert store

FirelightFlagboy:
(Windows only for now)

- Add an example that sign a provided message by using a referenced certificate.

### DIFF
--- a/libparsec/crates/platform_pki/src/lib.rs
+++ b/libparsec/crates/platform_pki/src/lib.rs
@@ -13,11 +13,17 @@ pub enum CertificateHash {
     SHA256(Box<[u8; 32]>),
 }
 
-impl AsRef<[u8]> for CertificateHash {
-    fn as_ref(&self) -> &[u8] {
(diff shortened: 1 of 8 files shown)


libparsec/crates/platform_pki/src/lib.rs
```diff
@@ -69,3 +82,24 @@ pub struct CertificateDer {
     pub cert_ref: CertificateReferenceIdOrHash,
     pub der_content: Bytes,
 }
+
+#[derive(Debug, thiserror::Error)]
+pub enum SignMessageError {
+    #[error("Cannot open certificate store: {}", .0)]
+    CannotOpenStore(std::io::Error),
+    #[error("Cannot found certificate")]
```
AureliaDolo:
```suggestion
    #[error("Cannot find certificate")]
```

FirelightFlagboy: done

PR #11162, opened 2025-09-18
feat(pki): Allow to decrypt message using cert store

### LINKED ISSUE #11079: Migrate `smartcard::platform::decrypt_secret_key` windows only
See https://github.com/Scille/parsec-extensions/blob/5f0ae5ac44d23d775d92e241b546fc0cac327a0a/parsec_ext/smartcard/win32.py#L518-L529

## How it works

1. Open cert store
2. Find related certificate from ids
3. Get `NCrypt` handle
4. Decrypt the message using the handle

FirelightFlagboy:
(Windows only)

- Add example to decrypt message

Closes #11079


libparsec/crates/platform_pki/src/windows/mod.rs
```diff
@@ -381,3 +381,98 @@ fn ncrypt_encrypt_message(message: &[u8], handle: &NcryptKey) -> std::io::Result
...
+        // We do not support a CryptoAPI provider as its API is marked for depreciation by windows.
+        PrivateKey::CryptProv(..) => {
+            todo!("Use CryptGetUserKey to get the keypair")
+        }
+        // Handle to a CryptoGraphy Next Generation (CNG) API
+        PrivateKey::NcryptKey(handle) => {
+            ncrypt_decrypt_message(encrypted_message, &handle).map(Into::into)
+        }
+    }
+    .map_err(DecryptMessageError::CannotDecrypt)?;
```
AureliaDolo: It seems like this part is repeated a few times (expect for the specific error) would it be worth putting that in a dedicated function ?

FirelightFlagboy: Could be, If so it could be interesting to generate the errors using `error_set` to simplify conversion of errors from that function

AureliaDolo: That, or implement the conversion manually. Not fun but it's done in just one place.

FirelightFlagboy: done

PR #11187, opened 2025-09-23
ci(publish): Improve determining `RELEASE_TAG`

FirelightFlagboy:
At least allow for better debugging experience if something goes wrong.

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -30,7 +30,7 @@ on:
         required: true
   release:
     types:
-      # triggered when a draft release is published 
+      # triggered when a draft release is published
       # and also if the release is edited later on
       - edited
   pull_request:
@@ -43,10 +43,6 @@ on:
 concurrency:
   group: publish-package-${{ github.workflow }}-${{ github.ref }}
   cancel-in-progress: true
-
-
-env:
-  RELEASE_TAG: ${{ (github.event_name == 'release' && github.ref_name) || (github.event_name == 'pull_request' && 'nightly') || inputs.release_tag }}
 
 jobs:
   publish:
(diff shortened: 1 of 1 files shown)


FirelightFlagboy:
Fail because `nightly` is already on snap, but the important step is in success state

.github/workflows/publish.yml
```diff
@@ -63,6 +59,34 @@ jobs:
             misc
         timeout-minutes: 5
 
+      - name: Determine release tag to use
+        shell: bash -eux -o pipefail {0}
+        id: tag
```
mmmarcos: nit:
```suggestion
        id: release-tag
```

.github/workflows/publish.yml
```diff
@@ -63,6 +59,34 @@ jobs:
...
+        run: |
+          case '${{ github.event_name }}' in
+            # In the conext of a release event, `ref_name` is set to the tag of the release.
+            release)
+              export tag='${{ github.ref_name }}'
```
mmmarcos: Why not using `RELEASE_TAG` instead of `tag`?

Also, it is good practice to assign and export in different lines (see [SC2155](https://github.com/koalaman/shellcheck/wiki/SC2155)). Even if does not apply for literal values, I would only assign here and export at the end (e.g. before tee'ing to `GITHUB_ENV`).

```suggestion
              RELEASE_TAG='${{ github.ref_name }}'
```

.github/workflows/publish.yml
```diff
@@ -63,6 +59,34 @@ jobs:
...
+              export tag='${{ inputs.release_tag }}'
+              ;;
+          esac
+
+          if [ -z "$tag" ]; then
+            echo 'Failed to determine tag from event' >&2
+            exit 1
+          fi
+
+          echo "RELEASE_TAG=$tag" | tee -a "$GITHUB_ENV"
```
mmmarcos, line 87: With previous comment, this would become
```suggestion
          export RELEASE_TAG
          echo "RELEASE_TAG=$RELEASE_TAG" | tee -a "$GITHUB_ENV"
```

.github/workflows/publish.yml
FirelightFlagboy: I've removed the `id` since we do not need it

.github/workflows/publish.yml
FirelightFlagboy: Shorter name that don't require using caps lock, I've made the change about removing the `export` but note that the impacted line would still always by in success since they'll always be a string (even empty)

.github/workflows/publish.yml
FirelightFlagboy: We do not need the `export`, only the `tee` to `GITHUB_ENV`

Issue #11128, opened 2025-09-15
[MS] Editics - Select which editors to use with CryptPad


mmmarcos:
- Disable txt? -> Update to Monaco Editor if possible
- Disable csv? -> Yes it's not working so disabled it
- ...

It depends on what was testing here: https://github.com/Scille/parsec-cloud/issues/10764

Max-7:
There's already a whole list of text files defined here https://github.com/Scille/parsec-cloud/blob/2d2783c14ce36bcc5ba057e7959cd56930fee0d5/client/src/common/fileTypes.ts#L30.

Also make sure text uses Code, not Pad, and that rtf do work with Pad.

PR #11194, opened 2025-09-24
[DRAFT] [MS] Select editors to use with cryptpad

Ironicbay:
Closes https://github.com/Scille/parsec-cloud/issues/11128


mmmarcos:
To outline it here, according to @Ironicbay:

> So far, I've mainly been testing what is compatible and what isn't, and I've come to the somewhat annoying conclusion that there are file types that are supported by the editor but not by the viewer, and vice versa. 
> So I wanted to discuss the approach we wanted to take: 
> - do we want to offer the editor only if the viewer is available, 
> - or do we offer both on as many files as possible regardless of whether the other one works?

IMO, supporting editing a file without a viewer may be confusing for the end-user. I suggest sticking to files where a viewer _is_ supported (we could always add editor for other filetypes later). 

So, for the moment, available actions should be
- ✅ View
- ✅ View + Edit
- ❌ Edit

client/src/services/cryptpad.ts
```diff
@@ -158,23 +159,51 @@ export class Cryptpad {
 }
 
 export function getDocumentTypeFromExtension(extension: string): CryptpadDocumentType {
+  // CryptpadDocumentType.Pad adds html content to text files, avoiding using it for now
+  // CryptpadDocumentType.Presentation works with .pptx, but is disabled as long as there's no viewer for it
   switch (extension.toLowerCase()) {
+    case 'xml':
```
Max-7: Why not use what's in `fileTypes.ts`? We have two different files for the same thing. We have the whole `detectFileType` that's used for the viewers, why not use this to choose what editor to use?

client/src/services/cryptpad.ts
```diff
@@ -157,29 +158,25 @@ export class Cryptpad {
...
-    case 'md':
-      return CryptpadDocumentType.Code;
     default:
       return CryptpadDocumentType.Unsupported;
   }
 }
 
-export function isEnabledCryptpadDocumentType(extension: string): boolean {
-  return ENABLED_DOCUMENT_TYPES.includes(getDocumentTypeFromExtension(extension));
+export function isEnabledCryptpadDocumentType(contentType: FileContentType | undefined): boolean {
```
Max-7: Doesn't make much sense for this function to take an arg that's undefined. I know why you added it, but it's not a good reason: this function shouldn't have to care whether others functions may or may not return undefined, they are separate.

PR #11196, opened 2025-09-24
feat(pki): Add algorithm ID

FirelightFlagboy:
Used to identify which algorithm was used and which to use even tho we only support RSA for now

### DIFF
--- a/libparsec/crates/platform_pki/examples/decrypt_message.rs
+++ b/libparsec/crates/platform_pki/examples/decrypt_message.rs
@@ -5,7 +5,9 @@ use std::path::PathBuf;
 
 use anyhow::Context;
 use clap::Parser;
-use libparsec_platform_pki::{decrypt_message, CertificateHash, CertificateReference};
+use libparsec_platform_pki::{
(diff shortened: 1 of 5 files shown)


libparsec/crates/platform_pki/examples/decrypt_message.rs
```diff
@@ -39,11 +44,13 @@ fn main() -> anyhow::Result<()> {
...
         .context("Failed to decode hex encoded data")?;
 
-    let res = decrypt_message(&data, &cert_ref).context("Failed to decrypt message")?;
+    let res =
+        decrypt_message(args.algorithm, &data, &cert_ref).context("Failed to decrypt message")?;
 
     println!(
-        "Decrypted by cert with id: {}",
-        data_encoding::BASE64.encode_display(&res.cert_ref.id)
+        "Decrypted by cert with id {{{}}} with algo {}",
```
touilleMan, line 51: `{{{}}}` ? o_O'

FirelightFlagboy: `{{` and `}}` is the syntax to espace `{` and `}` respectivelly

PR #11205, opened 2025-09-25
fix(mountpoint): Use `umount` as fallback of `fusermount`

FirelightFlagboy:
During the cleanup of previous mountpoints we now use `umount` as a fallback of `fusermount` in case the later fail. We also ensure the commands exit with a success status


libparsec/crates/platform_mountpoint/src/unix/mount.rs
```diff
@@ -240,15 +240,7 @@ pub async fn clean_base_mountpoint_dir(
...
-                    if let Err(err) = std::process::Command::new("fusermount")
-                        .arg("-u")
-                        .arg(&entry_path)
-                        .status()
-                    {
-                        log::warn!(
-                            "Base home dir cleanup, failed to unmount: {} ({err})",
-                            entry_path.display()
-                        );
+                    if try_unmount_path(&entry_path).is_break() {
```
touilleMan: I don't see the need for using `ControlFlow` over just returning a boolean:
```rust
if !try_unmount_path(&entry_path) {
  // Cannot unmount this path, better leave it alone...
  continue;
}
```

libparsec/crates/platform_mountpoint/src/unix/mount.rs
```diff
@@ -368,3 +360,34 @@ fn create_suitable_mountpoint_dir(
 
     unreachable!()
 }
+
+fn try_unmount_path(path: &Path) -> ControlFlow<()> {
+    try_unmount_path_using_command("fusermount", &["-u"], path)
```
touilleMan, line 366: same as above: I find it much simpler to just have `try_unmount_path_using_command` returns a boolean and do:
```rust
fn try_unmount_path(path: &Path) -> bool {
  try_unmount_path_using_command("fusermount", ...) || try_unmount_path_using_command("mount", ...)
}
```

libparsec/crates/platform_mountpoint/src/unix/mount.rs
```diff
@@ -368,3 +360,34 @@ fn create_suitable_mountpoint_dir(
...
+        // for the given path.
+        .map(ControlFlow::Continue)
+        // Otherwise, the caller should stop the cleanup process for the given path.
+        .unwrap_or(ControlFlow::Break(()))
+}
+
+fn try_unmount_path_using_command(command: &str, prefix_args: &[&str], path: &Path) -> Option<()> {
+    let display_path = path.display();
+    log::debug!("Try unmount {display_path} using `{command}`");
+    match Command::new(command).args(prefix_args).arg(path).status() {
```
touilleMan, line 373: We should handle the case that the command is simply not available on the OS, in such case we should not log a warning (but just a debug) since it is an expected behavior

libparsec/crates/platform_mountpoint/src/unix/mount.rs
```diff
@@ -240,15 +240,7 @@ pub async fn clean_base_mountpoint_dir(
                     // (e.g. Parsec crashed without properly unmounting).
                     // The only way to remove it is to manually force unmount.
                     log::debug!("Base home dir cleanup, unmounting: {}", entry_path.display());
-                    if let Err(err) = std::process::Command::new("fusermount")
-                        .arg("-u")
-                        .arg(&entry_path)
-                        .status()
-                    {
-                        log::warn!(
```
touilleMan, line 248: We should keep this log since it indicates a mountpoint couldn't be unmounted, which is not something that can be logged from `try_unmount_path_using_command` (since this function represent only one attempt at unmounting, but another one might succeed)

PR #11234, opened 2025-10-01
[MS] Sort users by roles in workspace sharing modal

### LINKED ISSUE #8570: Sort by role in workspace sharing dialog
In the workspace sharing dialog users are sorted by name, but it is more common (at least in this dialog) to search users by their role first.

image

A minima: it should be possible to _always_ sort by role-first:
- sort by role first (Owner > Manager > Contributor > Reader) 
- and then by name (alphabetically)
 
Bonus: sort by role-first by default, but allow _also_ to sort by name

Bonus x2: Consider "Ascending" and "Descending" sorts

fabienSvtr:
https://github.com/user-attachments/assets/d496479b-bec8-402e-9359-4e08e7e45409

closes #8570

### DIFF
--- a/client/src/views/workspaces/WorkspaceSharingModal.vue
+++ b/client/src/views/workspaces/WorkspaceSharingModal.vue
@@ -274,9 +274,18 @@ const filteredUserRoles = computed(() => {
 });
 
 const filteredSharedUserRoles = computed(() => {
+  const roleOrder = [WorkspaceRole.Owner, WorkspaceRole.Manager, WorkspaceRole.Contributor, WorkspaceRole.Reader];
   return filteredUserRoles.value
     .filter((userRole: UserRole) => userRole.role !== null)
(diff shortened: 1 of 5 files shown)


client/src/views/workspaces/WorkspaceSharingModal.vue
```diff
@@ -276,7 +276,11 @@ const filteredUserRoles = computed(() => {
 const filteredSharedUserRoles = computed(() => {
   return filteredUserRoles.value
     .filter((userRole: UserRole) => userRole.role !== null)
-    .sort((item1, item2) => item1.user.humanHandle.label.localeCompare(item2.user.humanHandle.label));
+    .sort((item1, item2) => item1.user.humanHandle.label.localeCompare(item2.user.humanHandle.label))
```
Max-7: You're sorting the whole list twice, this is not the same as sorting by role first and by name second.

Max-7 (changes requested):

Max-7 (approved):

PR #11271, opened 2025-10-07
Fix server/pyproject.toml syntax

mmmarcos:
It seems that optional groups are not yet fully supported for `[dependency-groups]`. They are supported by Poetry but dependabot parser fails (see https://github.com/Scille/parsec-cloud/actions/runs/18305430818)


FirelightFlagboy (approved):
If dependabot is too limiting, we could use [renovate](https://docs.renovatebot.com/)

FirelightFlagboy:
Note: that we still want to migrate to `uv` as it's simpler/faster

Issue #10379, opened 2025-05-22
[🚀 | Feature request]: Handle parsec links in web mode


Max-7:
### What you would like to accomplish

Parsec uses links for referencing to a file/workspace or for invitations. They use the protocol `parsec3://`.

The custom protocol can pose a bit of problem with modern mail client (like gmail), and so the link in emails is provided as an https link that directs to the server, and the server does a redirection to the parsec3 protocol. The browser then tries to open the link using the app registered for this protocol.

When a user copies a file link, it is directly using the parsec3 protocol.

In web mode, we need to find a way to consume those links.

### How the feature should work

The GUI already has the ability to process links using the query part of the URL. `/home?fileLink=<URLEncodedLink>` and `/home?claimLink=<URLEncodedLink>` and `/home?bootstrapLink=<URLEncodedLink>. It does need a bit of tuning but it can be made functional with minimum effort.

If the server is able to redirect a HTTPS link to the web app by filling up the right queries, we can handle links very easily.
In case no matching device is found on the browser, the app could just propose to open the link with the `parsec3://` protocol instead, making it fairly transparent for someone who doesn't use the web app.

FirelightFlagboy:
Need to ask @touilleMan about that

Max-7:
The whole process would look like:

- every link provided by the app are https links. When the user clicks on it, they get redirected to the server
- if the server doesn't have the web version of Parsec running, it can either have a link to a web server, or it redirects the link to parsec3:// (as done currently)
- if it does have the web service running, it redirects the link (with parsec3:// protocol) to the web version and adds the link as a query parameter (https://app.parsec.coud/client?link=parsec3://<the url-encoded link>)
- the GUI looks at the link, if it has the organization it tries to open the link with it. If it doesn't or if it fails, it redirects/gives the link to the user to open the desktop version

One caveat it that by providing links only with `https://`, a user has to be online for a link to work. A user shouldn't have trouble understanding it though since the web browser is going to open and display a connection error. It also makes the links easier to share by mail.

PR #11281, opened 2025-10-08
[MS] Handle Parsec URLs in web mode

Ironicbay:
Closes https://github.com/Scille/parsec-cloud/issues/10379

How to test:
1.  Build a webapp 
```
BASE_URL=/client npm run web:release
```
2. Run a local server using this webapp
```
poetry -P server run python -m parsec run --dev --with-client-web-app client/dist/ --spontaneous-organization-bootstrap
```
3. Create an organization on this server through the custom server process using `parsec3://localhost:6777?no_ssl=true`
4. Generate any parsec link
5. You can test the redirection page by replacing `parsec3://localhost:6777/` with `http://localhost:6777/redirect/`

### DIFF
--- a/client/src/locales/en-US.json
+++ b/client/src/locales/en-US.json
@@ -343,6 +343,14 @@
             }
         }
     },
+    "WebRedirectPage": {
+        "title": "Launching Parsec App",
+        "subtitle": "You can choose to open this link in Parsec app or in the browser.",
+        "desktop": "Open in Parsec app",
+        "web": "Continue in browser",
+        "downloadText": "Don't have Parsec app yet?",
+        "downloadLink": "Download Parsec"
+    },
     "CreateOrganization": {
         "fullname": "Full name",
         "fullnamePlaceholder": "John Smith",
--- a/client/src/locales/fr-FR.json
+++ b/client/src/locales/fr-FR.json
@@ -342,6 +342,14 @@
             }
         }
     },
+    "WebRedirectPage": {
+        "title": "Lancement de l'application Parsec",
+        "subtitle": "Vous pouvez choisir d'ouvrir ce lien dans l'application Parsec ou dans votre navigateur web.",
+        "desktop": "Ouvrir dans l'application Parsec",
+        "web": "Continuer dans votre navigateur web",
+        "downloadText": "Vous n'avez pas encore l'application Parsec ?",
+        "downloadLink": "Télécharger Parsec"
+    },
     "CreateOrganization": {
         "fullname": "Nom complet",
         "fullnamePlaceholder": "Jean Martin",
--- a/client/src/services/linkHandler.ts
+++ b/client/src/services/linkHandler.ts
@@ -1,0 +1,129 @@
+// Parsec Cloud (https://parsec.cloud) Copyright (c) BUSL-1.1 2016-present Scille SAS
+
+import { availableDeviceMatchesServer } from '@/common/device';
+import { bootstrapLinkValidator, claimLinkValidator, fileLinkValidator } from '@/common/validators';
+import { AvailableDevice, getOrganizationHandle, listAvailableDevices, parseFileLink } from '@/parsec';
+import {
+  backupCurrentOrganization,
+  currentRouteIs,
+  getConnectionHandle,
+  navigateTo,
+  RouteBackup,
+  Routes,
+  switchOrganization,
+} from '@/router';
+import { Information, InformationLevel, InformationManager, PresentationMode } from '@/services/informationManager';
+import { modalController, popoverController } from '@ionic/vue';
+import { Base64, Validity } from 'megashark-lib';
+
+async function handleJoinLink(link: string): Promise<void> {
+  if (!currentRouteIs(Routes.Home)) {
+    await switchOrganization(null, true);
+  }
+  await navigateTo(Routes.Home, { query: { claimLink: link } });
+}
+
+async function handleBootstrapLink(link: string): Promise<void> {
+  if (!currentRouteIs(Routes.Home)) {
+    await switchOrganization(null, true);
+  }
+  await navigateTo(Routes.Home, { query: { bootstrapLink: link } });
+}
+
+async function handleFileLink(link: string, informationManager: InformationManager): Promise<void> {
+  const result = await parseFileLink(link);
+  if (!result.ok) {
+    informationManager.present(
+      new Information({
+        message: 'link.invalidFileLink',
+        level: InformationLevel.Error,
+      }),
+      PresentationMode.Toast,
+    );
+    return;
+  }
+  const linkData = result.value;
+  // Check if the org we want is already logged in
+  const handle = await getOrganizationHandle({ id: linkData.organizationId, server: { hostname: linkData.hostname, port: linkData.port } });
+
+  // We have a matching organization already opened
+  if (handle) {
+    if (getConnectionHandle() && handle !== getConnectionHandle()) {
+      await switchOrganization(handle, true);
+    }
+
+    const routeData: RouteBackup = {
+      handle: handle,
+      data: {
+        route: Routes.Workspaces,
+        params: { handle: handle },
+        query: { fileLink: link },
+      },
+    };
+    await navigateTo(Routes.Loading, { skipHandle: true, replace: true, query: { loginInfo: Base64.fromObject(routeData) } });
+  } else {
+    // Check if we have a device with the org
+    const devices = await listAvailableDevices();
+    let matchingDevice: AvailableDevice | undefined;
+    for (const device of devices) {
+      // Pre-matching on org id only, in case we don't find a server match. This could happen
+      // with different DNS, IP addresses instead of hostname, ...
+      if (device.organizationId === linkData.organizationId && !matchingDevice) {
+        matchingDevice = device;
+      }
+      if (
+        (await availableDeviceMatchesServer(device, { hostname: linkData.hostname, port: linkData.port })) &&
+        device.organizationId === linkData.organizationId
+      ) {
+        matchingDevice = device;
+        // Full match takes priority, we can break out the loop
+        break;
+      }
+    }
+    if (!matchingDevice) {
+      await informationManager.present(
+        new Information({
+          message: { key: 'link.orgNotFound', data: { organization: linkData.organizationId } },
+          level: InformationLevel.Error,
+        }),
+        PresentationMode.Modal,
+      );
+      return;
+    }
+    if (!currentRouteIs(Routes.Home)) {
+      await backupCurrentOrganization();
+    }
+    await navigateTo(Routes.Home, { replace: true, skipHandle: true, query: { deviceId: matchingDevice.deviceId, fileLink: link } });
+  }
+}
+
+export async function handleParsecLink(link: string, informationManager: InformationManager): Promise<void> {
+  if (await modalController.getTop()) {
+    informationManager.present(
+      new Information({
+        message: 'link.appIsBusy',
+        level: InformationLevel.Error,
+      }),
+      PresentationMode.Toast,
+    );
+    return;
+  }
+  if (await popoverController.getTop()) {
+    await popoverController.dismiss();
+  }
+  if ((await claimLinkValidator(link)).validity === Validity.Valid) {
+    await handleJoinLink(link);
+  } else if ((await fileLinkValidator(link)).validity === Validity.Valid) {
+    await handleFileLink(link, informationManager);
+  } else if ((await bootstrapLinkValidator(link)).validity === Validity.Valid) {
+    await handleBootstrapLink(link);
+  } else {
+    await informationManager.present(
+      new Information({
+        message: 'link.invalid',
+        level: InformationLevel.Error,
+      }),
+      PresentationMode.Modal,
+    );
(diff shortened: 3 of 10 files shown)


server/parsec/asgi/redirect.py
```diff
@@ -55,4 +56,34 @@ def redirect_parsec_url(
...
+    with_client_web_app: bool = request.app.state.with_client_web_app
+    query = quote_plus(
+        location_url_query,
+        safe="/",
+        encoding="utf8",
+        errors="strict",
+    )
+    if with_client_web_app:
+        web_location_url = urlunsplit((
+            'http',
```
FirelightFlagboy: **todo:** Be dynamic on the used schema

FirelightFlagboy (changes requested):

server/parsec/asgi/redirect.py
```diff
@@ -55,4 +56,16 @@ def redirect_parsec_url(
...
     )
 
+    with_client_web_app: bool = request.app.state.with_client_web_app
+
+    if with_client_web_app:
+        query = quote_plus(location_url_query, safe="/", encoding="utf8", errors="strict")
+        location_url = urlunsplit((
+            'http',
+            server_addr_split.netloc,
+            f'client/webRedirect?webRedirectUrl={server_addr_split.scheme}://{server_addr_split.netloc}/{path}',
```
FirelightFlagboy: **issue:** Should be quoted alongside `query` argument to be correctly escaped

FirelightFlagboy: **nitpick:** Since the path already contain `webRedirect`, the query param could be shorter

```suggestion
            f'client/webRedirect?url={server_addr_split.scheme}://{server_addr_split.netloc}/{path}',
```

Ironicbay: I'd keep it that way for clarity in the `Query` [interface](https://github.com/Scille/parsec-cloud/pull/11281/files#diff-1a99729bde2ee3e065d5f2951042141649e096169844d8f2e5d9208ecf4434bcR308). But it is redundant with the route name indeed.

client/src/locales/en-US.json
```diff
@@ -343,6 +343,14 @@
             }
         }
     },
+    "WebRedirectPage": {
+        "title": "Launching Parsec App",
+        "subtitle": "You can choose whether to open this link in the Parsec desktop app or in this browser.",
+        "desktop": "Open in desktop app",
```
```suggestion
        "desktop": "Open in Parsec app",
```

client/src/locales/en-US.json
```diff
@@ -343,6 +343,14 @@
...
+        "web": "Continue in browser",
+        "downloadText": "Don't have the app yet?",
```
```suggestion
        "downloadText": "Don't have Parsec app yet?",
```

client/src/locales/en-US.json
```diff
@@ -343,6 +343,14 @@
             }
         }
     },
+    "WebRedirectPage": {
+        "title": "Launching Parsec App",
+        "subtitle": "You can choose whether to open this link in the Parsec desktop app or in this browser.",
```
```suggestion
        "subtitle": "You can choose to open this link in Parsec app or in the browser.",
```

client/src/locales/fr-FR.json
```diff
@@ -342,6 +342,14 @@
             }
         }
     },
+    "WebRedirectPage": {
+        "title": "Lancement de l'application Parsec",
+        "subtitle": "Vous pouvez choisir d'ouvrir ce lien dans l'application Parsec ou simplement dans votre navigateur.",
```
```suggestion
        "subtitle": "Vous pouvez choisir d'ouvrir ce lien dans l'application Parsec ou dans votre navigateur web.",
```

client/src/locales/fr-FR.json
```diff
@@ -342,6 +342,14 @@
...
+        "desktop": "Ouvrir dans l'application Parsec",
+        "web": "Continuer dans votre navigateur",
```
```suggestion
        "web": "Continuer dans votre navigateur web",
```

client/src/locales/fr-FR.json
```diff
@@ -342,6 +342,14 @@
...
+        "downloadText": "Vous n'avez pas l'application ?",
```
```suggestion
        "downloadText": "Vous n'avez pas encore l'application Parsec ?",
```

server/parsec/asgi/redirect.py
```diff
@@ -55,4 +55,27 @@ def redirect_parsec_url(
         (server_addr_split.scheme, server_addr_split.netloc, path, location_url_query, None)
     )
 
+    try:
+        with_client_web_app: bool = request.app.state.with_client_web_app
+    except AttributeError:
+        with_client_web_app = bool(0)
+
+    if with_client_web_app:
+        query = quote_plus(location_url_query, safe="/", encoding="utf8", errors="strict")
```
FirelightFlagboy: **question:** Why do we need to re-encode `location_url_query`?

server/parsec/asgi/redirect.py
```diff
@@ -55,4 +55,27 @@ def redirect_parsec_url(
...
+            f"{server_addr_split.scheme}://{server_addr_split.netloc}/{path}",
+            safe="/",
+            encoding="utf8",
+            errors="strict",
+        )
+        location_url = urlunsplit(
+            (
+                request.url.scheme,
+                server_addr_split.netloc,
+                f"client/webRedirect?webRedirectUrl={redirection_url}",
```
FirelightFlagboy: **todo:** put the query param `webRedirectUrl` into `query` to keep that string as a path

FirelightFlagboy: **todo:** Use `WEB_APP_BASE_URL` constant

server/parsec/asgi/redirect.py
```diff
@@ -55,4 +55,27 @@ def redirect_parsec_url(
...
+        redirection_url = quote_plus(
+            f"{server_addr_split.scheme}://{server_addr_split.netloc}/{path}",
+            safe="/",
+            encoding="utf8",
+            errors="strict",
+        )
+        location_url = urlunsplit(
+            (
+                request.url.scheme,
+                server_addr_split.netloc,
```
FirelightFlagboy: **issue:** We want to use `request.root_path` to handle reverse proxy configuration

server/tests/test_redirect.py
```diff
@@ -63,3 +63,19 @@ async def test_get_redirect_invitation(use_ssl: bool, client: httpx.AsyncClient,
     assert rep.status_code == 302
     location_addr = ParsecInvitationAddr.from_url(rep.headers["location"])
     assert location_addr == invitation_addr
+
+
+@pytest.mark.parametrize("use_ssl", (False, True))
+async def test_get_redirect_when_client_web_app_available(
```
FirelightFlagboy, line 71: **todo:** Also test that `webRedirectUrl` is correctly encoded

client/src/services/linkHandler.ts
mmmarcos, line 1: I would rename this file in order to be consistent with the existing services (instead of the gerund/noun form (_ing_), they are named`...Manager`, `...Logger`, `...Handler`).

So `linkHandler.ts` should be better.

server/parsec/asgi/redirect.py
Ironicbay: `location_url_query` is not encoded. It is used as is in the standard redirection, but here we need to encode it the same way as we do for the `redirection_url`.

newsfragments/10379.feature.rst
```diff
@@ -0,0 +1 @@
+Added a redirection page to handle Parsec links in both web and desktop mode, if the server has a webapp available.
```
mmmarcos: The second part seems redundant: if there is no web app, links won't be handled in web mode
```suggestion
Added a redirection page to handle Parsec links in both web and desktop mode
```

server/parsec/asgi/redirect.py
```diff
@@ -55,4 +55,27 @@ def redirect_parsec_url(
         (server_addr_split.scheme, server_addr_split.netloc, path, location_url_query, None)
     )
 
+    try:
+        with_client_web_app: bool = request.app.state.with_client_web_app
+    except AttributeError:
+        with_client_web_app = bool(0)
```
mmmarcos: why `bool(0)`?
```suggestion
        with_client_web_app = False
```

FirelightFlagboy: **nitpick:** 
```suggestion
        with_client_web_app = False
```

server/parsec/asgi/redirect.py
FirelightFlagboy: `location_url_query` come from `urlencode` as the function name implies, it encode a value

client/src/services/linkHandler.ts
```diff
@@ -0,0 +1,129 @@
...
+      return;
+    }
+    if (!currentRouteIs(Routes.Home)) {
+      await backupCurrentOrganization();
+    }
+    await navigateTo(Routes.Home, { replace: true, skipHandle: true, query: { deviceId: matchingDevice.deviceId, fileLink: link } });
+  }
+}
+
+export async function handleParsecLink(link: string, informationManager: InformationManager): Promise<void> {
```
mmmarcos, line 100: This function should be documented

```suggestion
/** Handle a "parsec3://" link
 * 
 * ...details about what this function does: parse link + redirect?...
 * ...errors...
 */
export async function handleParsecLink(link: string, informationManager: InformationManager): Promise<void> {
```

server/parsec/asgi/redirect.py
```diff
@@ -55,4 +55,27 @@ def redirect_parsec_url(
...
+
+    if with_client_web_app:
```
mmmarcos, line 73: Please add some code comments for clarity: what's the redirect strategy when dealing with client web app.
```suggestion
    # ...summarize the redirect strategy when dealing with client web app...
    if with_client_web_app:
```

server/parsec/asgi/redirect.py
Ironicbay: `urlencode` encodes the elements given to build the url, here we want the URL operators to be encoded as well
Without `quote_plus`: 
```
parsec3%3A//parsec.invalid/foo/bar?a=1&b=2
```
With `quote_plus`: 
```
parsec3%3A//parsec.invalid/foo/bar?a%3D1%26b%3D2
```

server/parsec/asgi/redirect.py
Ironicbay: Habit of writing `false` not capitalized and python not knowing this keyword 😅

server/tests/test_redirect.py
Ironicbay: Isn't this already done in the last `assert`? What would you do on top of it?

server/parsec/asgi/redirect.py
Ironicbay: Fixed, I simplified the function.

server/parsec/asgi/redirect.py
Ironicbay: There's no `root_path` on `request`, there is `request.app.root_path` but it seems empty in my testings. Could you elaborate?

server/parsec/asgi/redirect.py
```diff
@@ -55,4 +57,33 @@ def redirect_parsec_url(
...
+    except AttributeError:
+        with_client_web_app = False
+
+    # If there's no webapp available, we proceed to a standard redirection using
+    # the location_url, which will be opened in the desktop app if available.
+    # If a webapp is available, we instead redirect to a redirection page hosted
+    # on this web client using standard http(s) on the path WEB_APP_BASE_URL.
+    # The client then handles the parsec URL in this redirection page communicated
+    # through the webRedirectUrl argument.
+
```
mmmarcos: A couple of suggestions. Also, this comment is just above the `if with_client_web_app:` so I would start explaining that case. 
```suggestion
    # If web app is available, redirect to a page hosted in the client that will 
    # ask the user if they want to open the link in desktop app or in the browser.
    # The original `location_url` is therefore passed as a `redirection_url` to 
    # the client that will handle it in the redirection page.
    #
    # If web app is not available, the standard redirection is performed to 
    # `location_url`, which will be opened in the desktop app if available.
```

server/tests/test_redirect.py
FirelightFlagboy: The assert only check for an expected value, not if the value is correct (can be decoded into an url)

server/parsec/asgi/redirect.py
FirelightFlagboy: I mean `request.base_url` :pray: https://fastapi.tiangolo.com/reference/request/#fastapi.Request.base_url

server/parsec/asgi/redirect.py
```diff
@@ -55,4 +57,34 @@ def redirect_parsec_url(
...
+            safe="/",
+            encoding="utf8",
+            errors="strict",
+        )
+        location_url = urlunsplit(
+            (
+                request.base_url.scheme,
+                request.base_url.netloc,
+                WEB_APP_BASE_URL,
+                f"webRedirectUrl={redirection_url}{request.app.root_path}",
```
FirelightFlagboy: **question:** I'm not understanding why we need to suffix the query with `request.app.root_path`

Ironicbay: Left by mistake ^^

Max-7 (approved):

Issue #11298, opened 2025-10-09
[MS] CryptPad editor doesn't update language when locale changes


NicoTuxx:
# CryptPad editor doesn't update language when locale changes

## Current Behavior

When a user changes the application language in settings while a file is open in the CryptPad editor, the editor's UI language doesn't update automatically. The language only changes when:
- The page is fully reloaded, OR
- The user navigates away and opens a file again

## Expected Behavior

The CryptPad editor should automatically update its UI language when the user changes the language in the application settings, without requiring a page reload or file re-opening.

## Technical Context

The CryptPad integration API accepts the language via `editorConfig.lang` when initializing an editor:

```typescript
const cryptpadConfig = {
  editorConfig: {
    lang: longLocaleCodeToShort(I18n.getLocale()),
    // ... other config
  },
  // ... rest of config
};
```

However, CryptPad stores this language in `localStorage` (as `CRYPTPAD_LANG`) during initialization and doesn't automatically detect changes to this value once the editor is loaded.

## Edge Cases to Handle

1. User changes language → opens the same file again (may use cached editor instance)
2. User changes language → opens a different file (new editor instance but may reuse the script)
3. Multiple files open in different tabs/windows
4. User changes language while editing (unsaved changes must be preserved)

## Proposed Solution

Implement a **global locale change listener** that:

1. Watches for `I18n.getLocale()` changes at the application level (not component level)
2. When detected, either:
   - **Option A**: Completely reload the CryptPad script (`cryptpad-api.js`) and re-initialize all editor instances
   - **Option B**: Update localStorage and trigger a reload of the editor iframe for active file editors
3. Preserve any unsaved changes during the reload process

## Files to Review

- `client/src/services/cryptpad.ts` - CryptPad service and configuration
- `client/src/views/files/handler/editor/FileEditor.vue` - File editor component
- `client/src/views/files/handler/FileHandler.vue` - File handler wrapper
- `client/src/components/settings/SettingsList.vue` - Where `I18n.changeLocale()` is called

## Related Code

The language is currently passed during editor initialization:

```typescript
// In FileEditor.vue, openFileWithCryptpad()
editorConfig: {
  lang: longLocaleCodeToShort(I18n.getLocale()),
  user: {
    name: userInfo.humanHandle.label,
  },
},
```

And CryptPad stores it in localStorage (in `cryptpad/www/integration/main.js`):

```javascript
if (data.editorConfig.lang) {
[...]

Issue #11297, opened 2025-10-09
[MS] Manage Cryptpad theme properly when dark mode is effective


NicoTuxx:
At the moment, we force Cryptpad theme on light mode because Parsec does not have an official dark mode theme.

If a dark mode is properly implemented, we need to do:
- remove the custom code that is forcing light theme in cryptpad-deploy repository
- decide if we want cryptpad to work as default behavior (it matches the user preferences to be in a light or dark theme)
- if we want the theme to follow the one selected in Parsec, we need to implement a way to manage this (in both repos)

N.B. OnlyOffice provide a param for that: https://api.onlyoffice.com/docs/docs-api/usage-api/config/editor/customization/customization-standard-branding/#uitheme
BUT Cryptpad is used not only with OO and other pads would not be compatible with this. Secondly Cryptpad has its own theme management system that overrides the OO one, we can take a look at this param, but it's probably not a good idea to base a solution on this.

PR #11300, opened 2025-10-09
feat(cryptpad): Provide Cryptpad with user preferences & info

NicoTuxx:
# CryptPad User Info and Preferences Implementation

## Overview
This PR implements passing user information and preferences to the CryptPad integration, including username, language, and read-only mode functionality.

### ✅ Username
- **Status**: Fully implemented and working
- **Implementation**: Added `user` field to `CryptpadConfig.editorConfig` with `name` property
- **How it works**: The username is passed from `userInfo.humanHandle.label` to CryptPad's configuration
- **Code location**: `client/src/services/cryptpad.ts` (interface), `client/src/views/files/handler/editor/FileEditor.vue` (implementation)

### ✅ Language (Initial)
- **Status**: Fully implemented for initial file opening
- **Implementation**: Already existed in the codebase, using `I18n.getLocale()` converted to short locale code
- **How it works**: Language is set when the CryptPad iframe is initialized
- **Code location**: `client/src/views/files/handler/editor/FileEditor.vue`
- **Limitation**: Language updates require CryptPad iframe reload - see issue #11298

### ✅ Read-Only Mode
- **Status**: Fully implemented with automatic detection
- **Implementation**: 
  - Added `CryptpadAppMode` enum with `View` and `Edit` values
  - Added `mode` field to `CryptpadConfig`
  - Implemented automatic read-only mode detection for users with Reader role on workspace
- **How it works**: 
  - The `readOnly` prop is passed to FileEditor component and converted to `CryptpadAppMode.View` or `CryptpadAppMode.Edit`
  - When FileHandler mounts, it checks if the current user has the `WorkspaceRole.Reader` role on the workspace using `getWorkspaceSharing()`
  - If the user is a Reader, `readOnly` is automatically set to `true`
  - This ensures Reader users always open files in read-only mode regardless of how they access the file
- **Code location**: 
  - `client/src/services/cryptpad.ts` (enum and interface)
  - `client/src/views/files/handler/editor/FileEditor.vue` (mode conversion)
  - `client/src/views/files/handler/FileHandler.vue` (workspace role check)

### ❌ Theme
- **Status**: Not implemented - not supported by CryptPad integration API
- **Reason**: CryptPad's integration API (version 7.3.0) does not expose theme customization parameters. The theme is determined by CryptPad's own configuration and user's system preferences.
- **Related Issue**: Created issue #11297 to track potential future implementation at the CryptPad deployment level

### ❌ Dynamic Language Updates
- **Status**: Not implemented in this PR
- **Reason**: Updating the language requires reloading the CryptPad iframe, which has edge cases when users have multiple files open or reopen the same file
- **Related Issue**: Created issue #11298 to track this as a separate feature request

## Files Modified

1. **`client/src/services/cryptpad.ts`**
   - Added `CryptpadAppMode` enum (`View` and `Edit`)
   - Extended `CryptpadConfig` interface with `user` field in `editorConfig`
   - Extended `CryptpadConfig` interface with `mode` field

2. **`client/src/views/files/handler/editor/FileEditor.vue`**
   - Added `userInfo` prop to receive user information
   - Pass `user.name` to CryptPad config from `userInfo.humanHandle.label`
   - Convert `readOnly` prop to `CryptpadAppMode.View` or `CryptpadAppMode.Edit`

3. **`client/src/views/files/handler/FileHandler.vue`**
   - Added imports for `WorkspaceRole` and `getWorkspaceSharing`
   - Implemented workspace role check in `onMounted` to automatically detect Reader users
   - Set `readOnly` to `true` when user has `WorkspaceRole.Reader` on the workspace

## What Works Out of the Box

- ✅ Username is displayed in CryptPad's collaborative features
- ✅ Language is set correctly when opening a file
- ✅ Read-only mode prevents editing when enabled
- ✅ **Users with Reader role on workspace automatically open files in read-only mode**
- ✅ Manual read-only mode can still be set via query parameter

## What Requires Additional Work

- ⚠️ **Theme customization**: Would require changes in the CryptPad deployment configuration (tracked in issue #11297)
- ⚠️ **Dynamic language updates**: Would require iframe reload logic with proper handling of edge cases (tracked in issue #11298)

## Testing Notes

To test this implementation:
1. Open a CryptPad-supported file (e.g., `.odt`, `.xlsx`)
2. Verify your username appears in CryptPad's user list
3. Change application language, refresh the page and open a new file - verify CryptPad uses the correct language
4. Open a file with `?readOnly=true` query parameter - verify editing is disabled
5. As a user with Reader role on a workspace, open any file - verify it automatically opens in read-only mode

## Related Issues

- #11297: Track theme customization feature request
- #11298: Track dynamic language update feature request

### DIFF
--- a/client/src/services/cryptpad.ts
+++ b/client/src/services/cryptpad.ts
@@ -12,6 +12,11 @@ export enum CryptpadDocumentType {
   Unsupported = 'unsupported',
 }
 
+export enum CryptpadAppMode {
+  View = 'view',
+  Edit = 'edit',
+}
+
 export const ENABLED_DOCUMENT_TYPES = [
   CryptpadDocumentType.Pad,
   CryptpadDocumentType.Sheet,
@@ -30,8 +35,13 @@ interface CryptpadConfig {
   documentType: CryptpadDocumentType;
   editorConfig: {
     lang: string;
+    user?: {
+      name: string;
+      id: string;
+    };
   };
   autosave: number;
+  mode?: CryptpadAppMode;
   events: {
     onSave: (file: Blob, callback: () => void) => void;
     onNewKey?: (data: { new: string; old: string }, callback: (key: string) => void) => void;
--- a/client/src/services/fileOpener.ts
+++ b/client/src/services/fileOpener.ts
@@ -34,6 +34,7 @@ interface OpenPathOptions {
   onlyViewers?: boolean;
   atTime?: DateTime;
   useEditor?: boolean;
+  readOnly?: boolean;
 }
 
 // Uncomment here to enable file viewers on desktop; should be removed when all file viewers are implemented
@@ -206,12 +207,14 @@ async function openInEditor(
             contentType: contentType,
           });
         }
+
         await navigateTo(Routes.FileHandler, {
           query: {
             workspaceHandle: workspaceHandle,
             documentPath: entry.path,
             timestamp: options.atTime?.toMillis().toString(),
             fileTypeInfo: Base64.fromObject(contentType),
+            readOnly: options.readOnly,
           },
           params: {
             mode: FileHandlerMode.Edit,
--- a/client/src/views/files/FoldersPage.vue
+++ b/client/src/views/files/FoldersPage.vue
@@ -1441,7 +1441,11 @@ async function openEntryContextMenu(event: Event, entry: EntryModel, onFinished?
   const actions = new Map<FileAction, (file: EntryModel[]) => Promise<void>>([
     [FileAction.Preview, async (entries: EntryModel[]): Promise<void> => await openEntries(entries, { skipViewers: false })],
     [FileAction.Rename, renameEntries],
-    [FileAction.Edit, async (entries: EntryModel[]): Promise<void> => await openEntries(entries, { useEditor: true })],
+    [
+      FileAction.Edit,
+      async (entries: EntryModel[]): Promise<void> =>
+        await openEntries(entries, { useEditor: true, readOnly: ownRole.value === parsec.WorkspaceRole.Reader }),
+    ],
     [FileAction.MoveTo, moveEntriesTo],
     [FileAction.MakeACopy, copyEntries],
     [FileAction.Open, async (entries: EntryModel[]): Promise<void> => await openEntries(entries, { skipViewers: true })],
--- a/client/src/views/files/handler/FileHandler.vue
+++ b/client/src/views/files/handler/FileHandler.vue
@@ -190,9 +190,9 @@
             :is="handlerComponent"
             :content-info="contentInfo"
             :file-info="detectedFileType"
-            :read-only="readOnly"
             @file-loaded="handlerReadyRef = true"
             v-on="isComponentEditor() ? { onSaveStateChange: onSaveStateChange } : {}"
+            v-bind="isComponentEditor() ? { userInfo: userInfo, readOnly: readOnly } : {}"
           />
         </div>
       </div>
@@ -285,7 +285,7 @@ const handlerReadyRef = ref(false);
 const handlerComponent: Ref<typeof FileEditor | typeof FileViewer | null> = shallowRef(null);
 const handlerMode = ref<FileHandlerMode | undefined>(undefined);
 const sidebarMenuVisibleOnMounted = ref(false);
-const userInfo: Ref<ClientInfo | null> = ref(null);
+const userInfo: Ref<ClientInfo | undefined> = ref(undefined);
 const saveState = ref<SaveState>(SaveState.None);
 const savedIconRef = useTemplateRef<InstanceType<typeof IonIcon>>('savedIcon');
 const unsavedIconRef = useTemplateRef<InstanceType<typeof MsImage>>('unsavedIcon');
@@ -536,6 +536,14 @@ onMounted(async () => {
   handlerMode.value = getFileHandlerMode();
   const query = getCurrentRouteQuery();
   readOnly.value = Boolean(query.readOnly);
+
+  const clientInfoResult = await getClientInfo();
+  if (clientInfoResult.ok) {
+    userInfo.value = clientInfoResult.value;
+  } else {
+    window.electronAPI.log('error', `Failed to retrieve user info ${JSON.stringify(clientInfoResult.error)}`);
+  }
+
   await loadFile();
 
   hideHeader();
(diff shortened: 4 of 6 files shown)


client/src/services/fileOpener.ts
```diff
@@ -206,12 +207,14 @@ async function openInEditor(
...
           });
         }
+
         await navigateTo(Routes.FileHandler, {
           query: {
             workspaceHandle: workspaceHandle,
             documentPath: entry.path,
             timestamp: options.atTime?.toMillis().toString(),
             fileTypeInfo: Base64.fromObject(contentType),
+            readOnly: options.readOnly ? true : undefined,
```
Max-7: Can just use `options.readOnly` since it's an optional boolean.

client/src/views/files/FoldersPage.vue
```diff
@@ -322,7 +322,8 @@ const tabBarActions = computed(() => {
     if (isFileEditable(selectedEntries[0].name) && !isReadOnly) {
       actions.push({
         label: 'FoldersPage.tabbar.edit',
-        action: async () => await openEntries(getSelectedEntries(), { useEditor: true }),
+        action: async () =>
+          await openEntries(getSelectedEntries(), { useEditor: true, readOnly: ownRole.value === parsec.WorkspaceRole.Reader }),
```
Max-7: A bit redundant here since we only push this action `if (!readOnly)`. I don't think readers can see the edit button at the moment so it's not necessary to change this file I think, appart from testing.

client/src/views/files/FoldersPage.vue
```diff
@@ -1441,7 +1442,11 @@ async function openEntryContextMenu(event: Event, entry: EntryModel, onFinished?
   const actions = new Map<FileAction, (file: EntryModel[]) => Promise<void>>([
     [FileAction.Preview, async (entries: EntryModel[]): Promise<void> => await openEntries(entries, { skipViewers: false })],
     [FileAction.Rename, renameEntries],
-    [FileAction.Edit, async (entries: EntryModel[]): Promise<void> => await openEntries(entries, { useEditor: true })],
+    [
+      FileAction.Edit,
+      async (entries: EntryModel[]): Promise<void> =>
```
Max-7, line 1446: Doesn't seem to be a filter in this case, perfect for testing, but otherwise we probably want to avoid showing an `Edit` button to a read only user.

client/src/views/files/FoldersPage.vue
```diff
@@ -1558,7 +1563,7 @@ const actionBarOptionsFoldersPage = computed(() => {
         label: 'FoldersPage.fileContextMenu.actionEdit',
         icon: create,
         onClick: async () => {
-          await openEntries(getSelectedEntries(), { useEditor: true });
+          await openEntries(getSelectedEntries(), { useEditor: true, readOnly: ownRole.value === parsec.WorkspaceRole.Reader });
```
Max-7: Same here, we only display this button `if (ownRole.value !== parsec.WorkspaceRole.Reader)`

client/src/views/files/handler/FileHandler.vue
```diff
@@ -536,6 +537,14 @@ onMounted(async () => {
   handlerMode.value = getFileHandlerMode();
   const query = getCurrentRouteQuery();
   readOnly.value = Boolean(query.readOnly);
+
+  const clientInfoResult = await getClientInfo();
```
Max-7, line 540: Why add that here instead of the `FileEditor` directly? You only need it in the `openFileWithCryptpad` function in FileEditor.

client/src/views/files/handler/editor/FileEditor.vue
```diff
@@ -190,8 +198,12 @@ async function openFileWithCryptpad(): Promise<boolean> {
     documentType: documentType.value as CryptpadDocumentType, // Safe since we checked for null earlier
     editorConfig: {
       lang: longLocaleCodeToShort(I18n.getLocale()),
+      user: {
+        name: userInfo.humanHandle.label,
```
Max-7: You also get the userId, may as well set it.

client/src/services/cryptpad.ts
```diff
@@ -30,8 +35,13 @@ interface CryptpadConfig {
   documentType: CryptpadDocumentType;
   editorConfig: {
     lang: string;
+    user?: {
```
Max-7, line 38: Only the `user` object should be optional, name and id should be mandatory.

client/src/views/files/handler/FileHandler.vue
```diff
@@ -193,6 +193,7 @@
             :read-only="readOnly"
             @file-loaded="handlerReadyRef = true"
             v-on="isComponentEditor() ? { onSaveStateChange: onSaveStateChange } : {}"
+            v-bind="isComponentEditor() ? { userInfo: userInfo } : {}"
```
Max-7: Don't need to pass the prop here, but I did notice that the `read-only` prop causes error when opening a viewer (since viewers don't have the prop) so this v-bind should be used for read-only.

client/src/views/files/handler/FileHandler.vue
NicoTuxx: I did not add it, it was already in this file, I just move it a little bit earlier in the onMounted.
We wanted to avoid trying showing a handler Component if we don't have the info we need. (It's used for userDetails on this page too, it would be repetitive to call it a second time in Editor, since we already need it here.)

client/src/views/files/FoldersPage.vue
NicoTuxx: For now we should still pass it, we can open a new issue to think (with mock-ups) how we want and what we want to show/hide to the user.

client/src/views/files/FoldersPage.vue
NicoTuxx: same as next comment

client/src/views/files/handler/editor/FileEditor.vue
```diff
@@ -77,10 +84,11 @@ const cryptpadInstance = ref<Cryptpad | null>(null);
 const fileUrl = ref<string | null>(null);
 const error = ref('');
 
-const { contentInfo, fileInfo, readOnly } = defineProps<{
+const { contentInfo, fileInfo, readOnly, userInfo } = defineProps<{
   contentInfo: FileContentInfo;
   fileInfo: DetectedFileType;
   readOnly?: boolean;
+  userInfo: ClientInfo;
```
Max-7: userInfo is masked as `ClientInfo | null` in FileHandler, and you pass this prop directly, which means this prop should also be of type `ClientInfo | null`. I would change the clientInfo in FileHandler to `ClientInfo | undefined` instead, and mark this one as `ClientInfo?`.

client/src/views/files/FoldersPage.vue
Max-7: I kinda agree with the other one, since it's there, might as well not change it for now, but in this case it's completely useless: you enter the if only if ownRole.value !== parsec.WorkspaceRole.Reader, so it's the same as passing `readOnly: true`, which is the same as not passing any variable at all.

PR #11307, opened 2025-10-10
Ms 3.5.1 backports newsfragments

fabienSvtr:
Add missing newsfragments.

### DIFF
--- a/newsfragments/11107.bugfix.rst
+++ b/newsfragments/11107.bugfix.rst
@@ -1,0 +1,1 @@
+Fixed a bug causing files to be incorrectly deleted when cancelling an import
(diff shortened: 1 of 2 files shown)


newsfragments/11107.bugfix.rst
```diff
@@ -0,0 +1 @@
+Fix import while cancelling or importing files
```
Ironicbay:
```suggestion
Fixed a bug causing files to be incorrectly deleted when cancelling an import
```

newsfragments/11155.bugfix.rst
```diff
@@ -0,0 +1 @@
+Fix sorting files in Firefox
```
Ironicbay:
```suggestion
Fixed a bug on Firefox where files could only be sorted by name
```

PR #11327, opened 2025-10-13
refactor(types): Promote `CertificateReference` to standalone type

FirelightFlagboy:
`CertificateReference` will be used for other element notably for `LocalPendingEnrollment` & `DeviceFileSmartcard`


libparsec/crates/types/schema/pki/certificate_reference.json5
```diff
@@ -0,0 +1,14 @@
+{
+    "label": "CertificateReference",
+    "type": "certificate_reference",
+    "other_fields": [
```
touilleMan, line 4: Why not a variant type ?
Is none for both ID and Hash acceptable ?

libparsec/crates/types/schema/pki/certificate_reference.json5
```diff
@@ -0,0 +1,14 @@
...
+        {
+            "name": "certificate_id",
+            "type": "RequiredOption<Bytes>"
+        },
+        {
+            "name": "certificate_hash",
```
touilleMan, line 10: no mention of the algorithm ?
I feel like we want to specify this given both sha1 and sha256 are often used to identify x509 certificates
```suggestion
            "name": "certificate_sha256_hash",
```

libparsec/crates/types/src/pki.rs
```diff
@@ -367,22 +366,17 @@ impl TryFrom<CertificateReferenceData> for CertificateReference {
...
-                certificate_id: None,
-            },
-            CertificateReference::IdOrHash(id_or_hash) => {
-                let CertificateReferenceIdOrHash { id, hash } = id_or_hash;
-                CertificateReferenceData {
-                    certificate_hash: Some(hash.to_string()),
-                    certificate_id: Some(id),
-                }
+        let (hash, id) = match value {
+            CertificateReference::Id(id) => (None, Some(id)),
```
touilleMan, line 370: ID is capitalized everywher (e.g. `DeviceID`, `AccountAuthMethodID`, `DataError::UnexpectedRealmID` etc.), so we should stay consistent with this

```suggestion
            CertificateReference::ID(id) => (None, Some(id)),
```

libparsec/crates/types/src/pki.rs
```diff
@@ -367,22 +366,17 @@ impl TryFrom<CertificateReferenceData> for CertificateReference {
...
+            CertificateReference::Hash(hash) => (Some(hash), None),
+            CertificateReference::IdOrHash(CertificateReferenceIdOrHash { id, hash }) => {
```
touilleMan, line 372:
```suggestion
            CertificateReference::IDAndHash(CertificateReferenceIdOrHash { id, hash }) => {
```

`IdOrHash` seems to refer to what you can do with this object, not what this object contains (i.e. the object contains both an ID and a Hash, hence its ID or its hash can be used to )

I'd rather name it after what it contains

libparsec/crates/types/schema/pki/certificate_reference.json5
```diff
@@ -0,0 +1,14 @@
+{
```
touilleMan, line 1: Is there a reason for both ID and hash being optional ?

Note the `libparsec/crates/types/schema/pki/x509_certificate.json5` (that is not currently used, but has been migrated from Parsec v2) contains a non optional `certificate_sha1` field.

Regarding the ID, it is very unusual to have an optional ID (and a missing ID most likely represents to absence of something, not the presence of something that has no ID but could have one...), so we should add a comment here to explain what is the logic  here.

From what I recall, the ID is something that depends of the smartcard API (Windows CNG vs PKCS#11) used to access the private key.
So we can always obtain an ID from the smartcard API for a given key, but it may or may not work later on when using it to access the key (i.e. if a different smartcard API is used to obtained and to use the key).
In such case we should have the this id field mandatory but rename it so that it is specificity are obvious, typically:
- change the `id` suffix to `path`
- turn this type into a dict so that each API can have its own path (and we can skip access by path if the available path is not compatible with the smartcard API we are currently using)
- Add a comment to explain why this is so complicated ^^

```json5
{
    "label": "X509CertificateReference",
    "other_fields": [
         ...
        {
            // There is no standard ID for X509 certificate.
            // Instead each cryptographic API (e.g. PKCS#11, Windows CNG) has its own way of
            // identifying the path leading to the secret keys corresponding to the X509 certificate.
            "name": "certificate_path",
            "type": "Dict<X509CertificateCryptoAPI, Bytes>"
        }
        ...
    ],
    "nested_types": [
        {
            "name": "X509CertificateCryptoAPI",
            "variants": [
                {
                    "name": "PKCS11",
                    "discriminant_value": "PKCS11"
                },
                {
                    "name": "WindowsCNG",
                    "discriminant_value": "WINDOWS_CNG"
                }
            ]
        }
    ]
}
```

libparsec/crates/types/schema/pki/certificate_reference.json5
```diff
@@ -0,0 +1,14 @@
...
+    "label": "CertificateReference",
```
touilleMan, line 2: "Certificate" has already a meaning in the codebase that is different from x509, so we should choose a more precise name here, typically `X509CertificateReference`
```suggestion
    "label": "X509CertificateReference",
```

libparsec/crates/types/schema/pki/certificate_reference.json5
```diff
@@ -0,0 +1,14 @@
...
+    "type": "certificate_reference",
+    "other_fields": [
+        {
+            "name": "certificate_id",
+            "type": "RequiredOption<Bytes>"
+        },
+        {
+            "name": "certificate_hash",
+            "type": "RequiredOption<String>"
```
touilleMan, line 11: Hash is made of bytes
```suggestion
            "type": "RequiredOption<Bytes>"
```

libparsec/crates/types/schema/pki/certificate_reference.json5
FirelightFlagboy: I've created the issue #11330 for that

libparsec/crates/types/schema/pki/certificate_reference.json5
FirelightFlagboy: `certificate_hash` is an SRIHash so the used algorithm is encoded in the string

libparsec/crates/types/schema/pki/certificate_reference.json5
FirelightFlagboy: Yes, but here it's an SRI hash so it's represented as a string

libparsec/crates/types/schema/pki/certificate_reference.json5
FirelightFlagboy: Because we don't know how to write an enum using the schema.

For a cert ref to be valid, at least one value should be provided so either ID, hash or both

libparsec/crates/types/schema/pki/certificate_reference.json5
touilleMan: we should name `certificate_path` -> `certificate_uri` since that is the name used in PKCS11 and it is less misleading than path that may seem to refer to the OS filesystem

libparsec/crates/types/schema/pki/certificate_reference.json5
touilleMan: what is SRI ?

libparsec/crates/types/schema/pki/certificate_reference.json5
touilleMan: Why is that so ? Is this something standard ? Is it something that can be used as-is from both PKCS11 and Windows CNG APIs ?

I don't know about SRI, but https://www.srihash.org/ seems to suggest it is related to checking integrity of a webpage. So I don't understand the relationship with a x509 certificate...

Besides we only need sha256 and possibly sha1, but SRI seems to only define sha256/sha284/sha512 (cf. https://www.w3.org/TR/sri/#terms). So in the end I don't see what we gain over the providing the sha256 as bytes in a KISS way.

In anyway, if we use that we should add a comment about this, and change the name to something like `certificate_sri_hash`

libparsec/crates/types/schema/pki/certificate_reference.json5
FirelightFlagboy: SRI Hash is a format for checksum values, it a string composed of hash id + hash value in base64

https://developer.mozilla.org/en-US/docs/Web/Security/Subresource_Integrity

FirelightFlagboy:
Superseded by #11379

PR #11340, opened 2025-10-15
[MS] Warn before leaving the file editor if there are unsaved changes

### LINKED ISSUE #11329: [MS] Warning message when quitting a file edition with unsaved changes
When a user is editing a file and decide to leave the editor, we need atleast to warn him with a message (toast? modal? @fabienSvtr) about the unsaved changes since this could lead to a data loss.

Max-7:
Closes #11329

### DIFF
--- a/client/src/locales/en-US.json
+++ b/client/src/locales/en-US.json
@@ -2546,9 +2546,9 @@
             "tooltipError": "Failed to save the changes.",
             "readOnly": "Read only document",
             "notSavedTitle": "Discard changes?",
-            "notSavedQuestion": "Some changes have not be saved yet. Do you want to discard them?",
-            "notSavedDiscard": "Discard the changes",
-            "notSavedStay": "Stay on the page"
+            "notSavedQuestion": "Some changes have not been saved yet. Do you want to discard them?",
+            "notSavedDiscard": "Discard changes",
+            "notSavedStay": "Stay on page"
         }
     },
     "openFallback": {


client/src/locales/en-US.json
```diff
@@ -2544,7 +2544,11 @@
             "tooltipSaved": "Changes have been saved.",
             "tooltipUnsaved": "Changes haven't been saved yet.",
             "tooltipError": "Failed to save the changes.",
-            "readOnly": "Read only document"
+            "readOnly": "Read only document",
+            "notSavedTitle": "Discard changes?",
+            "notSavedQuestion": "Some changes have not be saved yet. Do you want to discard them?",
```
```suggestion
            "notSavedQuestion": "Some changes have not been saved yet. Do you want to discard them?",
```

client/src/locales/en-US.json
```diff
@@ -2544,7 +2544,11 @@
...
+            "notSavedDiscard": "Discard the changes",
```
```suggestion
            "notSavedDiscard": "Discard changes",
```

client/src/locales/en-US.json
```diff
@@ -2544,7 +2544,11 @@
...
+            "notSavedStay": "Stay on the page"
```
```suggestion
            "notSavedStay": "Stay on page"
```

client/src/views/files/handler/FileHandler.vue
```diff
@@ -532,6 +533,27 @@ function loadComponent(): void {
...
+    }
+  }
+  return true;
+}
+
+onBeforeRouteLeave(async () => {
+  return await checkSaved();
+});
+
+onBeforeRouteUpdate(async () => {
```
Ironicbay, line 553: What's the difference between this and `onBeforeRouteLeave`? Any specific cases where `onBeforeRouteUpdate` is necessary?

Max-7: I don't know if it can realistically happen in our app specifically but it would be if the page stays the same but the file changes, so if you're on the fileHandler editing one document and you're clicking on another document in the sidebar: the path stays the same but the parameters change, which should lead to an update. I don't think we can really trigger that right now since you can't edit a file while you're editing a file (the sidebar only use the viewers). This case specifically probably needs to be tested anyway, I just wanted to be thorough.

Ironicbay (approved):

PR #11342, opened 2025-10-15
refactor(pki): Rename `CertificateReference` to `X509CertificateReference`

### LINKED ISSUE #11341: Rename `CertificateReference` to `X509CertificateReference`

FirelightFlagboy:
Closes #11341

### DIFF
--- a/bindings/generator/api/pki.py
+++ b/bindings/generator/api/pki.py
@@ -18,15 +18,15 @@ class CertificateHash(Variant):
         data: Sha256BoxData
 
 
-class CertificateReferenceIdOrHash(Structure):
+class X509CertificateReferenceIdOrHash(Structure):
     id: Bytes
     hash: CertificateHash
 
 
-class CertificateReference(Variant):
+class X509CertificateReference(Variant):
     Id = VariantItemTuple(Bytes)
     Hash = VariantItemTuple(CertificateHash)
-    IdOrHash = VariantItemTuple(CertificateReferenceIdOrHash)
+    IdOrHash = VariantItemTuple(X509CertificateReferenceIdOrHash)
 
 
 class ShowCertificateSelectionDialogError(ErrorVariant):
@@ -36,6 +36,6 @@ class ShowCertificateSelectionDialogError(ErrorVariant):
 
 
 async def show_certificate_selection_dialog_windows_only() -> Result[
-    Optional[CertificateReference], ShowCertificateSelectionDialogError
+    Optional[X509CertificateReference], ShowCertificateSelectionDialogError
 ]:
     raise NotImplementedError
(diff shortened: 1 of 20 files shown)


libparsec/crates/types/tests/unit/local_device_file.rs
```diff
@@ -316,7 +316,7 @@ fn smartcard_device_file(alice: &Device) {
     .as_ref();
     let expected = DeviceFile::Smartcard(DeviceFileSmartcard {
         encrypted_key: b"foo".as_ref().into(),
-        certificate_ref: CertificateReference::Id(Bytes::copy_from_slice(b"foo")),
+        certificate_ref: X509CertificateReference::Id(Bytes::copy_from_slice(b"foo")),
```
touilleMan, line 319:
```suggestion
        certificate_ref: X509CertificateReference::Id(Bytes::from_static(b"foo")),
```

bindings/generator/api/pki.py
```diff
@@ -18,15 +18,15 @@ class SHA256:
...
 
 
-class CertificateReferenceIdOrHash(Structure):
+class X509CertificateReferenceIdOrHash(Structure):
     id: Bytes
     hash: CertificateHash
 
 
-class CertificateReference(Variant):
+class X509CertificateReference(Variant):
```
touilleMan, line 26: Since hash is expected to be always present, `X509CertificateReferenceIdOrHash` can be removed and `X509CertificateReference` becomes a simple struct with a mandatory `id` field and an optional `sha256_hash` field.

bindings/generator/api/pki.py
```diff
@@ -18,15 +18,15 @@ class SHA256:
         data: Sha256BoxData
 
 
-class CertificateReferenceIdOrHash(Structure):
+class X509CertificateReferenceIdOrHash(Structure):
     id: Bytes
```
touilleMan, line 22: Those types are over-engineered, the GUI doesn't care about `Bytes` and `CertificateHash` since it is only going to manipulate them as opaque objects (or at most the GUI may want to display the hash as hexa, but that should be a trivial conversion the GUI can do on its side)

So we should just provide them as regular `bytes`

Also note this over-engineering also impact libparsec internals: `libparsec_types::CertificateReference` should also be simplified by only using `Vec<u8>` for `id` and `[u8; 32]` for `sha256_hash` (this removes the need for the `CertificateHash` type)

libparsec/crates/platform_device_loader/tests/units/list.rs
```diff
@@ -165,7 +165,9 @@ async fn list_devices(tmp_path: TmpPath) {
             .parse()
             .unwrap(),
         device_label: "PC1".parse().unwrap(),
-        certificate_ref: CertificateReference::Id(Bytes::copy_from_slice(b"Mallory's certificate")),
+        certificate_ref: X509CertificateReference::Id(Bytes::copy_from_slice(
```
touilleMan, line 168:
```suggestion
        certificate_ref: X509CertificateReference::Id(Bytes::from_static(
```

PR #11343, opened 2025-10-15
refactor(bindings): Simplify `CertificateHash` to be string derivated

FirelightFlagboy:

### DIFF
--- a/bindings/generator/api/pki.py
+++ b/bindings/generator/api/pki.py
@@ -6,26 +6,28 @@ from .common import (
     Bytes,
     ErrorVariant,
     Result,
-    Sha256BoxData,
+    StrBasedType,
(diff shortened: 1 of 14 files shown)


bindings/generator/api/pki.py
```diff
@@ -6,16 +6,18 @@
...
     Structure,
     Variant,
     VariantItemTuple,
 )
 
 
-class CertificateHash(Variant):
-    class SHA256:
-        data: Sha256BoxData
+class CertificateHash(StrBasedType):
```
touilleMan:
```suggestion
class X509CertificateHash(StrBasedType):
```

FirelightFlagboy: done

PR #11348, opened 2025-10-15
[MS] Filtered menu and header actions depending on profile and role

### LINKED ISSUE #11232: [GUI] Externals can create workspaces
### Parsec version

3.5.0

- [ ] Any
- [ ] macOS
- [ ] Linux
- [x] Windows
- [x] Web

### Bug description

On Parsec client (GUI), outsider can create workspaces but the content is not synchronized.
Condition `v-show="clientProfile != UserProfile.Outsider"` seems not working.

### Reproduction steps

```bash
1. Create an outsider user
2. Connect with outsider profile
3. Create a workspace
```

### Relevant output:

https://github.com/user-attachments/assets/f46e4fb2-d219-4cda-8191-d27ccc3e9c27

Ironicbay:
Closes #11232 
- Removed infos about invitations in the organisation page for `Standard` users
- Removed access to organisation page for `External` users in small display
- Filtered workspace contextual menu options for `Reader` role
- Removed the option to create a workspace for `External` users
- Added tests for these changes

### DIFF
--- a/client/src/views/workspaces/WorkspacesPage.vue
+++ b/client/src/views/workspaces/WorkspacesPage.vue
@@ -91,7 +91,7 @@
               }}
             </ion-text>
             <ion-button
-              v-show="clientProfile != UserProfile.Outsider"
+              v-show="clientProfile !== UserProfile.Outsider"
               id="new-workspace"
               fill="outline"
               @click="openCreateWorkspaceModal()"
@@ -510,6 +510,10 @@ function onMsSorterChange(event: MsSorterChangeEvent): void {
 }
 
 async function createWorkspace(name: WorkspaceName): Promise<void> {
+  // Externals shouldn't be able to access here, but putting this as safety
+  if (clientProfile.value === UserProfile.Outsider) {
+    return;
+  }
   const result = await parsecCreateWorkspace(name);
   if (result.ok) {
     informationManager.present(
@@ -595,14 +599,16 @@ async function onOpenWorkspaceContextMenu(workspace: WorkspaceInfo, event: Event
 const actionBarOptionsWorkspacesPage = computed(() => {
   const actionsArray = [];
 
-  actionsArray.push({
-    id: 'button-new-workspace',
-    label: 'WorkspacesPage.createWorkspace',
-    icon: addCircle,
-    onClick: async (): Promise<void> => {
-      await openCreateWorkspaceModal();
-    },
-  });
+  if (clientProfile.value !== UserProfile.Outsider) {
+    actionsArray.push({
+      id: 'button-new-workspace',
+      label: 'WorkspacesPage.createWorkspace',
+      icon: addCircle,
+      onClick: async (): Promise<void> => {
+        await openCreateWorkspaceModal();
+      },
+    });
+  }
   return actionsArray;
 });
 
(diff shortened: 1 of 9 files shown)


client/src/views/workspaces/WorkspacesPage.vue
```diff
@@ -510,6 +510,10 @@ function onMsSorterChange(event: MsSorterChangeEvent): void {
 }
 
 async function createWorkspace(name: WorkspaceName): Promise<void> {
+  // Externals shouldn't be able to access here, but putting this as safety
```
Max-7, line 513: What's the behavior of `parsecCreateWorkspace` when the user is an outsider? Does it works or does it return an error?
- If it returns an error, then that check is counter-productive: we should never get here, but if we do, it's better to display an error rather than nothing happening when the user presses the button.
- If it works it's all good.

Ironicbay: Yes there's no restriction for externals to create a workspace. On master & latest version, they can create and browse local workspaces that won't be synced with the server.

client/tests/e2e/specs/organization_information.spec.ts
```diff
@@ -49,6 +51,50 @@ for (const displaySize of [DisplaySize.Large, DisplaySize.Small]) {
...
+    await expect(configContainer.locator('.info-list-item').locator('.info-list-item__title')).toHaveText([
+      'External profile',
+      'User limit (excluding users with External profile)',
+      'Server address',
+    ]);
+    await expect(configContainer.locator('.info-list-item').nth(0).locator('.info-list-item__value')).toHaveText(['Enabled']);
+    await expect(configContainer.locator('.info-list-item').nth(1).locator('.info-list-item__value')).toHaveText(['Unlimited']);
+    await expect(configContainer.locator('.info-list-item').nth(2).locator('.server-address-value__text')).toHaveText(/^parsec3:\/\/.+$/);
+
+    await expect(usersContainer.locator('.users-list-item').nth(0).locator('.users-list-item__title')).toHaveText('3');
```
Max-7: Creating intermediate variables can make this a little clearer.

client/tests/e2e/specs/organization_information.spec.ts
```diff
@@ -49,6 +51,50 @@ for (const displaySize of [DisplaySize.Large, DisplaySize.Small]) {
...
+    const storageContainer = container.locator('.organization-storage');
+    const emailInvitationCard = usersContainer.locator('.invitation-card-list-item').nth(0);
+    const PkiRequestCard = usersContainer.locator('.invitation-card-list-item').nth(1);
+    await usersContainer.locator('#invitations-button').isVisible();
+
+    if (displaySize === DisplaySize.Small) {
+      await organizationPageStandard.setDisplaySize(DisplaySize.Small);
+    }
+
+    await expect(emailInvitationCard).not.toBeVisible();
```
Max-7: I think using `toBeHidden` is better. I dont really know how this works, but I would guess that it executes the `toBeVisible`, it tries to check for the full timeout (5s by default), and then applies the not. Which means the expect always take the full 5s.

client/tests/e2e/specs/organization_information.spec.ts
```diff
@@ -49,6 +51,50 @@ for (const displaySize of [DisplaySize.Large, DisplaySize.Small]) {
     await expect(storageContainer.locator('.storage-list-item__value').nth(1)).toHaveText(/^\d+(\.\d{2})? (B|KB|MB)$/);
   });
 
+  msTest(`Org info default state on ${displaySize} display on standard user`, async ({ organizationPageStandard }) => {
```
Max-7, line 55: Do we check somewhere that the org page is not available as an external?

PR #11337, opened 2025-10-14
Remove unicode support for email address

touilleMan:
close #11336

PR #11350, opened 2025-10-15
Add `server/parsec/components/postgresql/migrations/0015_drop_email_unicode_support.sql` migration

touilleMan:


touilleMan:
integrated to https://github.com/Scille/parsec-cloud/pull/11337

PR #11352, opened 2025-10-15
DeviceFileOpenBao schema

### LINKED ISSUE #11290: Create new device type for Openbao
- openbao_url (e.g. https://openbao-dev.parsec.cloud/)
- secret path (e.g. `/v1/parsec-keys/data/{user.id}/{secret-id}`)
- auth_path (e.g. `/v1/auth/hexagone/oidc/auth_url`)
- auth_type (e.g. `HEXAGONE`)

touilleMan:
closes #11290

### DIFF
--- a/client/src/components/profile/AuthenticationCard.vue
+++ b/client/src/components/profile/AuthenticationCard.vue
@@ -114,6 +114,13 @@ const methodConfig: Record<
     methodName: '',
     description: '',
   },
+  // TODO
+  [DeviceSaveStrategyTag.OpenBao]: {
+    imageSrc: '',
+    imageAlt: '',
+    methodName: '',
+    description: '',
+  },
   [CustomDeviceSaveStrategyTag.SSO]: {
     imageSrc: personCircleGradient,
     imageAlt: 'Smartcard',


client/src/components/profile/AuthenticationCard.vue
```diff
@@ -114,6 +114,13 @@ const methodConfig: Record<
     methodName: '',
     description: '',
   },
+  // TODO
+  [DeviceSaveStrategyTag.OpenBao]: {
```
Max-7, line 118: It should replace the `CustomDeviceSaveStrategyTag.SSO` just below.

libparsec/crates/types/schema/local_device/device_file_openbao.json5
```diff
@@ -0,0 +1,90 @@
...
+            "type": "String"
+        },
+        {
+            // e.g. /v1/auth/hexagone/oidc/auth_url
+            "name": "openbao_auth_path",
+            "type": "String"
+        },
+        {
+            "name": "openbao_auth_type",
+            "type": "OpenBaoAuthType"
```
FirelightFlagboy, line 66: **issue:** I'm not sure of that field being constrained, we are artificially limiting to hexagone and pro-connect which doesn't make sense in OIDC where we want instead to allow any system to implement the protocol.

For me, the field should just be a string

FirelightFlagboy (changes requested):

client/src/components/profile/AuthenticationCard.vue
touilleMan: I guess the `imageAlt: 'Smartcard',` should be changed, and also `sso` should be renamed to `openbao` for consistency.
So what about doing this is another PR ? ^^

touilleMan:
replaced by #11530

PR #11364, opened 2025-10-16
Implement pki_enrollment_submit cmd following rfc 1020.

### LINKED ISSUE #11304: Migrate `pki_enrollment_submit` ano cmd (memory only)
Depends on:

- [x] #11303 
- [x] #11407

---

- [ ] Copy schema in RFC: https://github.com/Scille/parsec-cloud/blob/master/docs/rfcs/1020-pki-support.md#submit-the-request-to-join-an-organization
- [ ] Add minimum test required on libparsec side for the new protocol
- [ ] Add minimum test required on server side for the new protocol
- [ ] Add method in `client` crate to allow anonymous user to submit the request

AureliaDolo:
Fix #11304


docs/rfcs/1020-pki-support.md
```diff
@@ -155,6 +155,12 @@ Part of the anonymous API [`pki_enrollment_submit`] allow a user to request join
         "name": "der_x509_certificate",
         "type": "Bytes"
       },
+      {
+        // This email should be the same as the one in the certificate
+        // It is provided here for convenience
+        "name": "der_x509_certificate_email",
+        "type": "EmailAddress"
+      },
```
FirelightFlagboy: **issue (blocking):** Already in payload

server/parsec/components/memory/pki.py
```diff
@@ -78,7 +76,8 @@ async def submit(
                 return PkiEnrollmentSubmitBadOutcome.ENROLLMENT_ID_ALREADY_USED
 
             try:
-                PkiEnrollmentSubmitPayload.load(submit_payload)
+                payload = PkiEnrollmentSubmitPayload.load(submit_payload)
+                submitter_email = payload.human_handle.email
```
FirelightFlagboy: **nitpick:** The `try except` is only for the deserialization of the payload, after that operation, we are sure that the human handle contain a "valid" email so we can move that line out of the `try` block

Issue #11102, opened 2025-09-11
Switch authentication method from/to smart card


mmmarcos:
It's easy enough as the SaveStrategy/AccessStrategy mechanism is simple enough, but from smartcards is not necessarily something we want to authorized.

Should we forbid "switching out" from smart card auth? (e.g. hide this possibility from the GUI). In theory it would not be possible to forbid an advanced user to switch out (e.g. using CLI), so there is no real security. Instead, we could simply inform the user that "switching out" from smartcard auth would be "less secure".

---

@fabienSvtr: I suggest to add a modal while the user is willing to switch authentication method.

<img width="1542" height="852" alt="Image" src="https://github.com/user-attachments/assets/9e70189c-ac5e-4c1a-9cd9-6c034ce2300b" />

PR #11368, opened 2025-10-17
[MS] Added Smartcard option to change authentication modal and fixed style

Ironicbay:
Closes https://github.com/Scille/parsec-cloud/issues/11102
Active card without checkmark (auth page)
Screenshot 2025-10-17 at 10 41 14

Change auth modal with keyring as active option
Screenshot 2025-10-17 at 10 45 22'

Change auth modal on web, system auth disabled, `Password` is active but is shown in `Default` state since the user can choose a new password
Screenshot 2025-10-17 at 10 55 02

Screenshot 2025-10-17 at 10 50 00

### DIFF
--- a/client/tests/e2e/helpers/greet.ts
+++ b/client/tests/e2e/helpers/greet.ts
@@ -152,7 +152,7 @@ export async function addUser(
   await expect(usersPage.locator('#users-page-user-list').getByRole('listitem')).toHaveCount(currentUserCount + 1);
 
   // Joiner sets password
-  const authRadio = joinData.content.locator('.choose-auth-page').locator('.radio-list-item');
+  const authRadio = joinData.content.locator('.choose-auth-page').locator('.radio-list-item:visible');
   await expect(authRadio).toHaveCount(2);
   await expect(authRadio.nth(0)).toHaveTheClass('radio-disabled');
   await expect(authRadio.nth(0).locator('.authentication-card-text__title')).toHaveText('System authentication');
--- a/client/tests/e2e/specs/auth_with_sso.spec.ts
+++ b/client/tests/e2e/specs/auth_with_sso.spec.ts
@@ -37,7 +37,7 @@ msTest('Go through trial org creation process, auth SSO', async ({ context }) =>
   await expect(authContainer).toBeVisible();
   await expect(authContainer.locator('.modal-header-title__text')).toHaveText('Authentication');
 
-  const authRadio = authContainer.locator('.choose-auth-page').locator('.radio-list-item');
+  const authRadio = authContainer.locator('.choose-auth-page').locator('.radio-list-item:visible');
   await expect(authRadio).toHaveCount(3);
   await expect(authRadio.nth(0)).toHaveTheClass('radio-disabled');
   await expect(authRadio.nth(0).locator('.authentication-card-text__title')).toHaveText('System authentication');
--- a/client/tests/e2e/specs/create_organization_custom.spec.ts
+++ b/client/tests/e2e/specs/create_organization_custom.spec.ts
@@ -136,7 +136,7 @@ msTest('Go through custom org creation process', async ({ home }) => {
   await expect(authNext).toBeVisible();
   await expect(authNext).toHaveDisabledAttribute();
 
-  const authRadio = authContainer.locator('.choose-auth-page').locator('.radio-list-item');
+  const authRadio = authContainer.locator('.choose-auth-page').locator('.radio-list-item:visible');
   await expect(authRadio).toHaveCount(2);
   await expect(authRadio.nth(0)).toHaveTheClass('radio-disabled');
   await expect(authRadio.nth(0).locator('.authentication-card-text__title')).toHaveText('System authentication');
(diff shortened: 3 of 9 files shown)


client/src/components/devices/ChooseAuthentication.vue
```diff
@@ -43,22 +43,25 @@
...
           />
         </ion-radio>
 
         <!-- TO DO: uncomment for Smartcard -->
         <!-- <ion-radio
-          v-if="false"
           class="item-radio radio-list-item"
           label-placement="end"
           justify="start"
+          :value="DeviceSaveStrategyTag.Smartcard"
```
Max-7, line 55: Instead of commenting, since we'll get the info of wether the smartcard is available or not, add a function somewhere in parsec (like device.ts or something), `async function isSmartcardAvailable(): Promise<bool> { return false; }` and add a v-if to hide those elements.
Since that's how we'll know if the smartcard is available later on, we'll just have to change the content of this function with a real call to libparsec.
Removing the comment also means that the editor and the linter can parse the code properly.

client/src/components/devices/ChooseAuthentication.vue
```diff
@@ -185,6 +194,25 @@ async function onChange(_value: any): Promise<void> {
...
+  switch (auth) {
+    case AvailableDeviceTypeTag.Password:
+      return AuthenticationCardState.Default;
+    case AvailableDeviceTypeTag.Keyring:
+      if (!keyringAvailable.value) {
+        return AuthenticationCardState.Unavailable;
+      }
+      return auth === props.activeAuth ? AuthenticationCardState.Active : AuthenticationCardState.Default;
+    case AvailableDeviceTypeTag.Smartcard:
+      if (isWeb()) {
```
Max-7: You can remove this condition. The goal is to have smartcards on web too, there's no need to add a condition right now, and if we ever have one it will be a call to libparsec like `isSmartcardAvailable()` or something like that, not `isWeb()`

client/src/components/profile/AuthenticationCard.vue
```diff
@@ -62,6 +62,7 @@ const props = defineProps<{
   state: AuthenticationCardState;
   authMethod: DeviceSaveStrategyTag | CustomDeviceSaveStrategyTag;
   disabled?: boolean;
+  hideCheckmark?: boolean;
```
Max-7: I don't like props like this. The parent shouldn't have to worry about the child displaying a checkbox or not, it makes the parent too aware of its children structure.

The prop should be a some kind of general thing and the component determines that if it's in this state, it should display or not a checkmark. For example, on Github for each commit you have a verified checkmark: the prop should be something like `isVerified: boolean`, not `displayCheckmark`.

client/src/components/profile/AuthenticationCard.vue
```diff
@@ -99,6 +100,7 @@ const methodConfig: Record<
     imageAlt: 'Smartcard',
     methodName: 'Authentication.method.smartcard.title',
     description: 'Authentication.method.smartcard.description',
+    unavailableExplanation: 'Authentication.keyringUnavailableOnWeb',
```
Max-7: Just unavailable in general, not specific to web.

client/src/views/profile/AuthenticationPage.vue
```diff
@@ -7,13 +7,22 @@
...
+          :hide-checkmark="true"
           v-show="currentDevice && currentDevice.ty.tag === AvailableDeviceTypeTag.Keyring"
         />
         <authentication-card
           :auth-method="DeviceSaveStrategyTag.Password"
           :state="AuthenticationCardState.Active"
+          :hide-checkmark="true"
           v-show="currentDevice && currentDevice.ty.tag === AvailableDeviceTypeTag.Password"
         />
+        <!-- TO DO: uncomment for Smartcard -->
```
Max-7: Same here, have a function in parsec.

Max-7: This one doesn't need to be commented, no device has `SmartCard` as a tag anyway.

client/src/parsec/device.ts
```diff
@@ -110,3 +110,7 @@ export async function createDeviceInvitation(sendEmail: boolean): Promise<Result
   }
   return generateNoHandleError<ClientNewDeviceInvitationError>();
 }
+
+export function isSmartcardAvailable(): boolean {
```
Max-7: Make it async. The call to libparsec later on will be async, it'll be easier to update later. It's a bit annoying, you'll probably have to create variables to store that result.

client/src/components/profile/AuthenticationCard.vue
```diff
@@ -62,6 +62,7 @@ const props = defineProps<{
   state: AuthenticationCardState;
   authMethod: DeviceSaveStrategyTag | CustomDeviceSaveStrategyTag;
   disabled?: boolean;
+  isVerified?: boolean;
```
Max-7: I meant `isVerified` in my example xD In this case I don't really know what triggers that checkmark. Is it to show the current auth method? In this case, `isCurrent` or something like that.

client/src/components/devices/ChooseAuthentication.vue
```diff
@@ -43,24 +44,27 @@
...
         </ion-radio>
 
-        <!-- TO DO: uncomment for Smartcard -->
-        <!-- <ion-radio
-          v-if="false"
+        <ion-radio
           class="item-radio radio-list-item"
           label-placement="end"
           justify="start"
+          :value="DeviceSaveStrategyTag.Smartcard"
```
Max-7, line 55: While we don't have any implementation at all, it's still better to `v-if="smartCardAvailable"` I think. We'll update later.

PR #11378, opened 2025-10-20
refactor(dependabot): Configure extra groups for rust deps

FirelightFlagboy:
- Add group for crypto, test, js, windows, android type of dependencies
- Also ignore bump of `generic-array` to v1 until crypto deps are update to use it.

  (to prevent compat issues where we need to convert a generic array from v1 to v0)

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -48,12 +48,58 @@ updates:
       - dependency-name: rand
         versions:
           - ">=0.9"
+      # The crypto dependencies need to be updated to be compatible with generic-array@v1
+      - dependency-name: generic-array
+        versions:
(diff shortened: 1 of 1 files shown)


.github/dependabot.yml
```diff
@@ -48,12 +48,58 @@ updates:
       - dependency-name: rand
         versions:
           - ">=0.9"
+      # Need the crypto dependencies to be update to be compatible with generic-array@v1
```
mmmarcos:
```suggestion
      # The crypto dependencies need to be updated to be compatible with generic-array@v1
```

mmmarcos (approved):

PR #11389, opened 2025-10-20
[MS] Displays a warning if creating a workspace with a similar name to one existing

### LINKED ISSUE #8391: [🚀 | Feature request]: Warn user when a workspace with the same name already exists

Avoid seeing two workspaces with the same name.

When creating a new workspace, check if there is already one workspace with this name.

The easy way would be to check that at organization level, but this would leak existing workspace names.

One possibility would be to check only workspaces the user has access to. But that raises the question of what happens when sharing a workspace to a user that already has access to a workspace with that name. 

We could either 
- accept that in this case some duplicates could appear
- force a renaming when sharing the workspace if necessary 
- have a nickname for the workspace for a given user (either a custom one or in the form of "from username's space")

_No response_

Max-7:
Closes #8391

### DIFF
--- a/client/src/locales/fr-FR.json
+++ b/client/src/locales/fr-FR.json
@@ -651,7 +651,11 @@
             "pageTitle": "Créer un nouvel espace de travail",
             "label": "Nom de l'espace de travail",
             "placeholder": "Mon espace",
-            "create": "Créer"
+            "create": "Créer",
+            "sameNameExistsTitle": "Nom d'espace dupliqué",
+            "sameNameExistsQuestion": "Un autre espace de travail existe avec ce nom ou un nom similaire. Souhaitez-vous créer cet espace?",
+            "createAnyway": "Créer",
+            "cancel": "Annuler"
         },
         "RenameWorkspaceModal": {
             "pageTitle": "Renommer l'espace de travail",
--- a/client/src/views/workspaces/WorkspacesPage.vue
+++ b/client/src/views/workspaces/WorkspacesPage.vue
@@ -183,6 +183,8 @@ import {
   MsSearchInput,
   MsSpinner,
   useWindowSize,
+  askQuestion,
+  Answer,
 } from 'megashark-lib';
 import {
   WORKSPACES_PAGE_DATA_KEY,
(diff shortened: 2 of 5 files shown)


client/src/views/workspaces/WorkspacesPage.vue
```diff
@@ -540,21 +542,57 @@ async function createWorkspace(name: WorkspaceName): Promise<void> {
...
+      const current = wi.currentName.toLocaleLowerCase();
+
+      // If we find a case-insensitive match or one name contains the other and the name is a bit longer that a few letters,
+      // both names are not too far off each other in length (3 characters difference at most)
+      // For example, we already have a 'Workspace', and we're trying to create 'Workspace 1'
+      if (newName === current) {
+        return true;
+      }
+      return (
+        newName.length > 6 && Math.abs(newName.length - current.length) <= 3 && (current.includes(newName) || newName.includes(current))
```
Ironicbay, line 576: It's unfortunate that small workspaces name are exempt from this check. But good enough for the issue addressed

client/src/locales/fr-FR.json
```diff
@@ -651,7 +651,11 @@
             "pageTitle": "Créer un nouvel espace de travail",
             "label": "Nom de l'espace de travail",
             "placeholder": "Mon espace",
-            "create": "Créer"
+            "create": "Créer",
+            "sameNameExistsTitle": "Nom d'espace dupliqué",
+            "sameNameExistsQuestion": "Un autre espace existe avec ce nom ou un nom similaire. Souhaitez-vous créer cet espace?",
```
```suggestion
            "sameNameExistsQuestion": "Un autre espace de travail existe avec ce nom ou un nom similaire. Souhaitez-vous créer cet espace?",
```

client/src/views/workspaces/WorkspacesPage.vue
Max-7: Yeah, I wanted to avoid cases where you have a workspace named `A` and it just matches with everything. I think the condition is decent enough. It's just meant as a helper, not some rule with really want to enforce.

PR #11390, opened 2025-10-20
feat(types): Hash in cert ref is always present

### LINKED ISSUE #11382: The hash value should always be present

FirelightFlagboy:
Closes #11382


libparsec/crates/types/src/pki/cert_ref.rs
```diff
@@ -62,21 +62,30 @@ impl FromStr for X509CertificateHash {
...
     }
 }
 
-#[derive(Debug, Clone, Eq, PartialEq, Serialize, Deserialize)]
-pub struct X509CertificateReferenceIdOrHash {
-    pub id: Bytes,
-    pub hash: X509CertificateHash,
+impl X509CertificateReference {
+    pub fn add_uri(mut self, uri: Bytes) -> Self {
+        let _ = self.uri.replace(uri);
```
AureliaDolo, line 83: I'm not exactly sure if/how that could be a problem, but here if there was another value we are just dropping it, without any warning.  Is there a good reason for which we would want to change the uri ?

libparsec/crates/types/src/pki/cert_ref.rs
```diff
@@ -62,21 +62,30 @@ impl FromStr for X509CertificateHash {
...
+        self
+    }
+
+    pub fn get_uri(&self) -> Option<&Bytes> {
+        self.uri.as_ref()
+    }
+
+    pub fn uris(&self) -> impl Iterator<Item = &Bytes> {
```
AureliaDolo, line 91: I find it a bit misleading: there is at most on uri in an option. And it seems to be a duplicate for get_uri then.

FirelightFlagboy: For now it's a simple Option but on the next PR, it will be a "special HashMap" at could yield multiple values

libparsec/crates/types/src/pki/cert_ref.rs
FirelightFlagboy: It come from the next PR, in the future a cert ref could have multiple URIs (one for each system (pkcs11, windows CNG)). Maybe `insert_or_replace` would be better :thinking: WDYT?

FirelightFlagboy: Will use `add_or_replace`

PR #11410, opened 2025-10-22
feat(pki): Use cert ref in `LocalPendingEnrollment`

### LINKED ISSUE #11409: Update `LocalPendingEnrollment` schema to use a `CertificateReference` instead of `x509Certificate`

FirelightFlagboy:
Closes #11409

### DIFF
--- a/libparsec/crates/types/schema/pki/local_pending_enrollment.json5
+++ b/libparsec/crates/types/schema/pki/local_pending_enrollment.json5
@@ -14,8 +14,8 @@
             "type": "OrganizationID"
         },
         {
-            "name": "x509_certificate",
-            "type": "X509Certificate"
+            "name": "x509_certificate_ref",
+            "type": "X509CertificateReference"
         },
         {
             "name": "submitted_on",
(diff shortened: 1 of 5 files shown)


libparsec/crates/types/schema/pki/local_pending_enrollment.json5
```diff
@@ -26,16 +26,36 @@
...
+        },
         {
             "name": "ciphertext",
+            // Correspond to `PrivateParts` that is serialized in msgpack format and encrypted with `encrypted_key`.
             "type": "Bytes"
         }
+    ],
+    "nested_types": [
+        {
+            "name": "PrivateParts",
```
AureliaDolo, line 48: Where is this type used ?

FirelightFlagboy: When an user make a pki enrollment request, it generate 2 keypair for encryption & signature.

The private parts of those keypairs are stored in the locally on the device (but encrypted by the pki)

PR #11419, opened 2025-10-23
fix(bindings): Fix inconsistency on tag value generation

FirelightFlagboy:


Max-7 (approved):
Don't know about the code quality but it does fix my problem.

PR #11425, opened 2025-10-24
fix(binding): Converting bytes to js buffer using typed array

FirelightFlagboy:
The previous method was using `ArrayBuffer` which is not typed: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Typed_arrays neon allow to generate a typed array from a slice so that also simplify the binding code


Max-7 (approved):
Don't know about the code quality but it does fix my problem.

PR #11428, opened 2025-10-24
[MS] Properly catch exception from the bindings

Max-7:


client/src/plugins/libparsec/index.ts
```diff
@@ -25,22 +25,46 @@ const _libparsec = registerPlugin<LibParsecPlugin>('LibParsec', {
...
+        const result = await target[name as keyof LibParsecPlugin](...args);
+        if (
+          result &&
+          (result as any).ok === false &&
+          !FUNCTIONS_TO_SKIP.find((sd) => sd.function === name && (result as any).error && (result as any).error.tag === sd.tag)
+        ) {
+          const resultError = result as { ok: boolean; error: { tag: string; error: string } };
+          logError('warn', `Error when calling ${name}: ${JSON.stringify(resultError.error)}`);
         }
+        return result;
```
FirelightFlagboy, line 63: **nitpick:** Could that block be moved out of the `try catch`? 

so the try is only for the call of the binding function or is it not possible because `result` is scoped?

Max-7: `result` is indeed scoped, moving it out would be a bit ugly, but doable. I just don't see any good reason why. If there's some other problems (like `JSON.stringify` failing) we probably want to catch it too so the GUI can keep operating.

PR #11446, opened 2025-10-27
[MS] Adds smartcard auth method

Max-7:
To be able to test, just return `true` in `parsec/pki.ts:isSmartcardAvailable`. It's only possible to select a "real" certificate on Windows, but you can cheat by replacing the function `parsec/pki.ts:selectCertificate` with something like
```typescript
export async function selectCertificate(): Promise<Result<X509CertificateReference | null, ShowCertificateSelectionDialogError>> {
  return { ok: true, value: { hash: 'HASH', uris: [] } };
}
```
It won't get you very far (you won't be able to create a device), but at least you'll be able to test the choice of the auth method.

### DIFF
--- a/client/src/locales/en-US.json
+++ b/client/src/locales/en-US.json
@@ -1160,7 +1160,7 @@
                 },
                 "selectButton": "Choose a certificate",
                 "selectOther": "Update",
-                "selectInstructions": "A pop-up will open, prompting you to chose the certificate.",
+                "selectInstructions": "A pop-up will open, prompting you to choose the certificate.",
                 "certificateSelected": "Certificate selected"
             },
             "sso": {


client/src/components/devices/ChooseAuthentication.vue
```diff
@@ -53,7 +53,7 @@
           label-placement="end"
           justify="start"
           :value="DeviceSaveStrategyTag.Smartcard"
-          :disabled="activeAuth === AvailableDeviceTypeTag.Smartcard"
+          :disabled="!smartcardAvailable || activeAuth === AvailableDeviceTypeTag.Smartcard"
```
Ironicbay, line 56: There is already `v-show="smartcardAvailable` on this element. Is this supposed to be shown if smartcard is not available?

client/src/locales/en-US.json
```diff
@@ -1156,7 +1157,11 @@
                     "subtitle": "Smartcard authentication is considered the most secure method to log in. By changing your authentication method, be mindful that your device security might be compromised.",
                     "yes": "Continue",
                     "no": "Keep current method"
-                }
+                },
+                "selectButton": "Choose a certificate",
+                "selectOther": "Update",
+                "selectInstructions": "A pop-up will open, prompting you to chose the certificate.",
```
```suggestion
                "selectInstructions": "A pop-up will open, prompting you to choose the certificate.",
```

client/src/components/devices/ChooseCertificate.vue
```diff
@@ -0,0 +1,131 @@
...
+
+  &-selected {
+    display: flex;
+    align-items: center;
+    gap: 0.25rem;
+    color: var(--parsec-color-light-secondary-text);
+
+    &__icon {
+      font-size: 1rem;
+    }
```
Ironicbay, line 128: Maybe change the icon color? 
Screenshot 2025-10-30 at 10 15 26

client/src/components/devices/ChooseCertificate.vue
```diff
@@ -0,0 +1,131 @@
...
+onMounted(async () => {
+  smartcardAvailable.value = await isSmartcardAvailable();
+});
+
+function getCertificate(): X509CertificateReference | undefined {
+  return certificate.value;
+}
+
+async function selectCertificate(): Promise<void> {
+  const result = await parsecSelectCertificate();
```
Ironicbay, line 73: Does this check the certificate validity?

client/src/components/devices/ChooseAuthentication.vue
Max-7: Yeah we'll remove the v-show once #11492 is available.

client/src/components/devices/ChooseCertificate.vue
Max-7: Depends on what you mean by validity. It opens a windows pop-up so you're able to select what windows considers to be a valid certificate on the machine. Right now there's no restrictions on what type of certificate can be selected.

PR #11456, opened 2025-10-27
Simplify `DeviceAccessStrategy`

touilleMan:


libparsec/crates/types/src/local_device_file.rs
```diff
@@ -340,34 +337,38 @@ impl DeviceAccessStrategy {
...
-    pub fn into_save_strategy(self) -> DeviceSaveStrategy {
-        match self {
-            DeviceAccessStrategy::Keyring { .. } => DeviceSaveStrategy::Keyring,
+            DeviceAccessStrategy::Keyring { .. } => {
+                if matches!(extra_info, AvailableDeviceType::Keyring) {
+                    Some(DeviceSaveStrategy::Keyring)
+                } else {
+                    None
+                }
+            }
```
FirelightFlagboy, line 353: **nitpick:** I would prefer this pattern and we return None at the end
```suggestion
            DeviceAccessStrategy::Keyring { .. } if matches!(extra_info, AvailableDeviceType::Keyring) => return Some(DeviceSaveStrategy::Keyring),
```

libparsec/crates/platform_device_loader/src/native/mod.rs
```diff
@@ -121,56 +121,73 @@ pub(super) async fn load_ciphertext_key(
...
+                    })?
+                    .into();
+
+                let key = SecretKey::from_recovery_passphrase(passphrase)
+                    .map_err(|_| LoadCiphertextKeyError::DecryptionFailed)?;
+
+                Ok(key)
+            } else {
+                Err(LoadCiphertextKeyError::InvalidData)
+            }
```
FirelightFlagboy, line 151: **nitpick:** Like the other places return `InvalidData` by default at the end
```suggestion
        DeviceAccessStrategy::Keyring { .. } if let DeviceFile::Keyring(device) = device_file => {
                let entry = KeyringEntry::new(&device.keyring_service, &device.keyring_user)
                    .map_err(|e| {
                        LoadCiphertextKeyError::Internal(anyhow::anyhow!("OS Keyring error: {e}"))
                    })?;

                let passphrase = entry
                    .get_password()
                    .map_err(|e| {
                        LoadCiphertextKeyError::Internal(anyhow::anyhow!(
                            "Cannot retrieve password from OS Keyring: {e}"
                        ))
                    })?
                    .into();

                let key = SecretKey::from_recovery_passphrase(passphrase)
                    .map_err(|_| LoadCiphertextKeyError::DecryptionFailed)?;

                return Ok(key)
            
```

libparsec/crates/platform_device_loader/src/web/mod.rs
```diff
@@ -58,25 +58,34 @@ pub(super) async fn load_ciphertext_key(
...
+            if let DeviceFile::Password(device) = device_file {
+                let key = device
+                    .algorithm
+                    .compute_secret_key(password)
+                    .map_err(|_| LoadCiphertextKeyError::InvalidData)?;
+
+                Ok(key)
+            } else {
+                Err(LoadCiphertextKeyError::InvalidData)
+            }
```
FirelightFlagboy, line 76: **nitpick:** 
```suggestion
        DeviceAccessStrategy::Password { password, .. } if let DeviceFile::Password(device) = device_file => {
                let key = device
                    .algorithm
                    .compute_secret_key(password)
                    .map_err(|_| LoadCiphertextKeyError::InvalidData)?;

               return Ok(key)
          
```

libparsec/crates/types/src/local_device_file.rs
touilleMan: See the comment it the beginning of the match, using `<pattern> if <condition>` means we need a default accept-all case for when the condition is false, but then this default case would also swallow any new variant (hence shadowing this code has to be updated).

The solution would be to have this default case be expressed as:
```rust
match self {
    // Simplified example: let's pretend we only have Keyring and Password variants
    DeviceAccessStrategy::Keyring { .. } if matches!(extra_info, AvailableDeviceType::Keyring) => Some(...),
    DeviceAccessStrategy::Password { password, .. } if matches!(extra_info, AvailableDeviceType::Password) => Some(...),
    // Default case explicitly listing all the variants we know about
    DeviceAccessStrategy::Keyring { .. } | DeviceAccessStrategy::Password { .. } => None,

}
```
However this looks rather cumbersome, so I don't think it add anything compared to the current `if/else`

libparsec/crates/platform_device_loader/src/native/mod.rs
touilleMan: same as above, I want to avoid using a default case here

libparsec/crates/platform_device_loader/src/web/mod.rs
touilleMan: same as above, I want to avoid using a default case here

PR #11464, opened 2025-10-28
Implement pki_enrollment_accept client part.

### LINKED ISSUE #11360: Expose a method in `AuthenticatedClient` to accept pki request

AureliaDolo:
Fix #11360

- what about timestamp out of ball park ? -> move to certif folder ?
- [ ] test
- [x] retrieve submit payload
- poll server for new certificate

### DIFF
--- a/libparsec/crates/client/src/client/mod.rs
+++ b/libparsec/crates/client/src/client/mod.rs
@@ -4,6 +4,7 @@
 
 mod list_frozen_users;
 mod organization_info;
+mod pki_enrollment_accept;
 mod pki_enrollment_list;
 mod pki_enrollment_reject;
 mod pki_enrollment_submit;
@@ -33,6 +34,7 @@ use std::{
 
 pub use self::{
     list_frozen_users::ClientListFrozenUsersError,
+    pki_enrollment_accept::PkiEnrollmentAcceptError,
     pki_enrollment_list::{PkiEnrollmentListError, PkiEnrollmentListItem},
     pki_enrollment_reject::PkiEnrollmentRejectError,
     pki_enrollment_submit::{pki_enrollment_submit, PkiEnrollmentSubmitError},
@@ -687,11 +689,31 @@ impl Client {
     ) -> Result<Vec<PkiEnrollmentListItem>, PkiEnrollmentListError> {
         pki_enrollment_list::list_enrollments(&self.cmds).await
     }
+
     pub async fn pki_enrollment_reject(
         &self,
         enrollment_id: EnrollmentID,
     ) -> Result<(), PkiEnrollmentRejectError> {
         pki_enrollment_reject::reject(&self.cmds, enrollment_id).await
+    }
+
+    pub async fn pki_enrollment_accept(
+        &self,
+        profile: UserProfile,
+        enrollment_id: EnrollmentID,
+        human_handle: &HumanHandle,
+        cert_ref: &X509CertificateReference,
+        submit_payload: &PkiEnrollmentSubmitPayload,
+    ) -> Result<(), PkiEnrollmentAcceptError> {
+        pki_enrollment_accept::accept(
+            self,
+            profile,
+            enrollment_id,
+            human_handle,
+            cert_ref,
+            submit_payload,
+        )
+        .await
     }
 }
 
--- a/libparsec/crates/client/src/client/pki_enrollment_accept.rs
+++ b/libparsec/crates/client/src/client/pki_enrollment_accept.rs
@@ -1,0 +1,212 @@
+// Parsec Cloud (https://parsec.cloud) Copyright (c) BUSL-1.1 2016-present Scille SAS
+
+use libparsec_client_connection::ConnectionError;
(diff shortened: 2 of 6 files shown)


libparsec/crates/client/src/client/mod.rs
```diff
@@ -674,6 +676,15 @@ impl Client {
     ) -> Result<DateTime, ClientGetOrganizationBootstrapDateError> {
         organization_info::get_organization_bootstrap_date(self).await
     }
+
+    pub async fn pki_enrollment_accept(
+        &self,
+        profile: UserProfile,
+        enrollment_id: EnrollmentID,
+        submit_payload: PkiEnrollmentSubmitPayload,
+    ) -> Result<(), ClientPkiEnrollmentAcceptError> {
```
FirelightFlagboy: **issue:** The admin need to be able to overwrite the provided human handle.

Since we are not relying on the user information provided by the x509 certificate but by the user.

libparsec/crates/client/src/client/pki_enrollment_accept.rs
```diff
@@ -0,0 +1,256 @@
...
+    // sign payload with accepter x509certificate
+    let (payload_signature, accepter_der_x509_certificate) =
+        if let Some(cert) = show_certificate_selection_dialog_windows_only()? {
+            (
+                sign_message(&payload, &cert)?.signature,
+                get_der_encoded_certificate(&cert)?.der_content,
+            )
+        } else {
+            return Err(ClientPkiEnrollmentAcceptError::NoCertificateChosen);
+        };
```
FirelightFlagboy: **issue (blocking):** The GUI should already have asked the user about which certificate to use.

The function must be provided with a certificate reference.

It also causing issue because you're using something that is only available on windows.

libparsec/crates/client/src/client/pki_enrollment_accept.rs
```diff
@@ -0,0 +1,256 @@
...
+}
+
+/// Helper to prepare the creation of a new user.
+/// Returns:
+/// - new user certificate
+/// - redacted ser certificate
+/// - device certificate
+/// - redacted certificate
+/// - accept answer payload
+fn create_new_signed_user_certificates(
```
FirelightFlagboy, line 170: **chore:** That function is already defined in `client/src/invite/greeter.rs`

A refactor is needed to extract the similar bits (returning the serialized device&user certificates) and keep the specific one (generating `InvitetuserConfirmation` & `PkiEnrollmentAnswerPayload`)

libparsec/crates/client/src/client/pki_enrollment_accept.rs
```diff
@@ -0,0 +1,227 @@
...
+) -> Result<PkiAcceptOutcome, ClientPkiEnrollmentAcceptError> {
+    // retrieve payload from  enrollment_id
+    let accepter_der_x509_certificate = get_der_encoded_certificate(cert_ref)?.der_content;
+
+    let local_file = libparsec_platform_device_loader::get_default_local_pending_file(
+        &client.config().config_dir,
+        enrollment_id,
+    );
+
+    let local_pending = load_pki_local_pending(local_file).await?;
```
FirelightFlagboy: **issue (blocking):** The accepter won't have access to the local pending save on the submitter filesystem.

The payload is obtain from `pki_enroll_list` and the value should likely be provided to the function

FirelightFlagboy (changes requested):

libparsec/crates/platform_device_loader/src/lib.rs
```diff
@@ -770,3 +770,31 @@ pub async fn save_pki_local_pending(
...
+impl From<ReadFileError> for LoadPkiLocalPendingError {
+    fn from(value: ReadFileError) -> Self {
+        match value {
+            ReadFileError::StorageNotAvailable => Self::StorageNotAvailable,
+            ReadFileError::Internal(error) => Self::Internal(error),
+        }
+    }
+}
+
+pub async fn load_pki_local_pending(
```
FirelightFlagboy: **issue (blocking):** We want need to load a local pending.

Instead we would list all available pending request and at that point we won't need to re-read a request on the filesystem has we would have everything we need extracted from the first step (#11524)

libparsec/crates/client/src/client/pki_enrollment_accept.rs
```diff
@@ -0,0 +1,215 @@
...
+    #[error(transparent)]
+    CertificateSelectionError(#[from] ShowCertificateSelectionDialogError),
+    #[error(transparent)]
+    SignatureError(#[from] SignMessageError),
+    #[error("active users limit reached")]
+    ActiveUsersLimitReached,
+    #[error("human handle already taken")]
+    HumanHandleAlreadyTaken,
+    #[error(transparent)]
+    GetDerEncodedCertificateError(#[from] GetDerEncodedCertificateError),
```
FirelightFlagboy: **nitpick:** During yesterday discussion we decided to merge every pki operation error into a single variant for pki submit.

I think that also apply here

FirelightFlagboy (approved):

PR #11474, opened 2025-10-28
Better `Account(Save|Access)Strategy` for the GUI

touilleMan:
With this change, the GUI no longer has to first deal with the opaque key upload/fetch in the account and then the actual operation (org bootstrap, enrollment finalize etc.) \o/

### DIFF
--- a/libparsec/crates/platform_device_loader/src/web/error.rs
+++ b/libparsec/crates/platform_device_loader/src/web/error.rs
@@ -90,12 +90,6 @@ error_set::error_set! {
         ReadToEnd(ReadToEndError),
         RmpDecode(libparsec_types::RmpDecodeError)
     }
-    InvalidPathError := {
-        #[display("Invalid path {}", path.display())]
-        InvalidPath {
-            path: PathBuf
-        }
-    }
     ReadFile := {
         GetFile(GetFileHandleError),
         ReadFile(ReadToEndError),


Max-7 (approved):
GUI side looks good

libparsec/crates/platform_device_loader/Cargo.toml
```diff
@@ -68,3 +69,5 @@ libparsec_tests_lite = { workspace = true }
 libparsec_tests_fixtures = { workspace = true, features = [
     "test-with-platform-device-loader-testbed",
 ] }
+
+crc32fast = { workspace = true }
```
FirelightFlagboy, line 73: **question:** Only used by testbed ?

libparsec/src/device.rs
```diff
@@ -3,10 +3,189 @@
...
+        Smartcard {
+            certificate_reference: X509CertificateReference,
+        },
+        AccountVault {
+            account_handle: Handle,
+        },
+    }
+
+    impl DeviceSaveStrategy {
+        pub fn convert(
```
touilleMan: Add comment to explain why not `TryFrom` (i.e. side effect are in use)

libparsec/src/device.rs
```diff
@@ -3,10 +3,189 @@
...
+            key_file: PathBuf,
+        },
+        AccountVault {
+            key_file: PathBuf,
+            account_handle: Handle,
+        },
+    }
+
+    impl DeviceAccessStrategy {
+        pub fn convert(
```
touilleMan: Add comment to explain why not TryFrom (i.e. side effect are in use)

libparsec/src/device.rs
touilleMan: done.

libparsec/src/device.rs
touilleMan: done.

PR #11496, opened 2025-10-30
refactor(binding/gen): Use union type format

FirelightFlagboy:
`ruff` keep rewriting `Optional[T]` to `T | None`, this make `generate.py` support the `UnionType`.

Additionally:

- List `ruff` sort import order & remove unused imports
- Use `list`, `dict`, ... over their `type.[..]` equivalent

### DIFF
--- a/bindings/generator/generate.py
+++ b/bindings/generator/generate.py
@@ -5,16 +5,15 @@ from __future__ import annotations
 import argparse
 import subprocess
 from collections import OrderedDict, namedtuple
+from collections.abc import Callable, Iterable
 from dataclasses import dataclass, field
 from importlib import import_module
 from inspect import isclass, iscoroutinefunction, isfunction, signature
 from pathlib import Path
-from types import ModuleType
-from typing import (
+from types import ModuleType, UnionType
+from typing import (  # noqa: UP035
     Any,
-    Callable,
     Dict,
-    Iterable,
     List,
     Set,
     Tuple,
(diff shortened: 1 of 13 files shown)


bindings/generator/generate.py
```diff
@@ -193,33 +188,35 @@ class BaseTypeInUse:
...
-        if origin is Union:  # Python resolves `Optional[Foo]` as `Union[Foo, None]`
+        if origin is Union or isinstance(
+            param, UnionType
+        ):  # Python resolves `Optional[Foo]` as `Union[Foo, None]`
             assert len(args) == 2 and type(None) in args
             elem_param = next(x for x in args if x is not type(None))
             return OptionalTypeInUse(BaseTypeInUse.parse(elem_param))
 
-        elif origin in (list, List):
+        elif origin is list:
```
touilleMan, line 208: This change means the binding declaration can no longer use `List/Dict/Set/Tuple` from `typing`
I'm in favor of this, however it may be hard to detect the issue for somebody using by mistake those types.

So I would add an additional assert at the beginning of the final `else` to better explain the issue:
```python
    else:
        assert origin in (List, Dict, Set, Tuple), "Don't use List/Dict/Set/Tuple from `typing` module, use instead `list/dict/set/tuple`!"
        [...]
```

FirelightFlagboy: done

PR #11544, opened 2025-11-04
Add js bindings for pki enrollment accept.

### LINKED ISSUE #11459: Bindings: Accept pki request (accepter)
Depends on:

- [x] #11360

AureliaDolo:
Fix #11459


libparsec/src/client.rs
```diff
@@ -709,16 +709,18 @@ pub async fn client_pki_enrollment_accept(
     enrollment_id: EnrollmentID,
     human_handle: &HumanHandle,
     cert_ref: &X509CertificateReference,
-    submit_payload: &PkiEnrollmentSubmitPayload,
+    submit_payload: Bytes,
```
FirelightFlagboy, line 712: **issue:** I'm gonna close my eyes on that 🫣

The client should give you back the payload in a struct format but I've discovered an issue with pki_list listing unstrusted value with raw payload (not deserialized)

AureliaDolo: see #11568

PR #11545, opened 2025-11-05
[MS] Added default values for window width and height

Ironicbay:
Following https://github.com/Scille/parsec-cloud/pull/11531


client/electron/src/setup.ts
```diff
@@ -402,6 +402,8 @@ export class ElectronCapacitorApp {
     this.MainWindow = new BrowserWindow({
       icon: appIcon,
       show: false,
+      width: 1500,
```
mmmarcos: Have you considered `1600` to use the same ratio used for minimum values (16:9)?
https://en.wikipedia.org/wiki/16:9_aspect_ratio#Common_resolutions
```suggestion
      width: 1600,
```

PR #11573, opened 2025-11-06
fix(protocol): Add missing signature algo

### LINKED ISSUE #11536: Missing used signature algorithm during enrollment list
https://github.com/Scille/parsec-cloud/blob/0040a244bd4f4d5e2a75e101a079c5ff0b5b1e20/libparsec/crates/protocol/schema/authenticated_cmds/pki_enrollment_list.json5#L44-L48

FirelightFlagboy:
Closes #11536


bindings/generator/api/pki.py
```diff
@@ -84,11 +84,21 @@ async def show_certificate_selection_dialog_windows_only() -> Result[
     raise NotImplementedError
 
 
+CUSTOM_TO_RS_STRING_DISPLAY = (
+    "|v| -> Result<_, std::convert::Infallible> { Ok(std::string::ToString::to_string(&v)) }"
+)
```
AureliaDolo, line 89: If it's here and not inside the PkiSignatureAlgorithm class because it's generic, it could be moved to common.py

AureliaDolo (approved):

bindings/generator/api/pki.py
```diff
@@ -84,11 +84,21 @@ async def show_certificate_selection_dialog_windows_only() -> Result[
...
+
+
+class PkiSignatureAlgorithm(StrBasedType):
+    custom_to_rs_string = CUSTOM_TO_RS_STRING_DISPLAY
+
+
 class PkiEnrollmentListItem(Structure):
     enrollment_id: EnrollmentID
     submitted_on: DateTime
     der_x509_certificate: Bytes
```
touilleMan, line 99: Why are we exposing certificate&payload info to the GUI ? Those are opaque buffers that the GUI won't be able to do anything with it...

bindings/generator/api/pki.py
touilleMan: YAGNI, if it is currently only used in a single place, better not try to factorize the code given it lowers the readability

server/parsec/components/memory/datamodel.py
```diff
@@ -52,6 +52,7 @@
     VerifyKey,
     VlobID,
 )
+from parsec._parsec_pyi.pki import PkiSignatureAlgorithm
```
touilleMan, line 55: Be careful when using auto-complete for imports: it doesn't know a bout `parsec._parsec` and hence do the import from the `.pyi` typing stub files which is going to fail at runtime
```suggestion
from parsec._parsec import PkiSignatureAlgorithm
```

server/parsec/components/memory/pki.py
```diff
@@ -16,6 +16,7 @@
     UserProfile,
     VerifyKey,
 )
+from parsec._parsec_pyi.pki import PkiSignatureAlgorithm
```
touilleMan, line 19:
```suggestion
from parsec._parsec import PkiSignatureAlgorithm
```
(should be merged in the previous import)

server/parsec/components/pki.py
```diff
@@ -15,6 +15,7 @@
     anonymous_cmds,
     authenticated_cmds,
 )
+from parsec._parsec_pyi.pki import PkiSignatureAlgorithm
```
touilleMan, line 18: same as above

touilleMan (changes requested):

FirelightFlagboy:
Deprecated by #11583

PR #11575, opened 2025-11-06
PostgreSQL implementation of PKI enrollment APIs

touilleMan:


server/parsec/components/postgresql/pki_reject.py
```diff
@@ -0,0 +1,107 @@
...
+    AuthAndLockCommonOnlyBadOutcome,
+    auth_and_lock_common_read,
+)
+from parsec.components.postgresql.utils import Q
+from parsec.events import EventPkiEnrollment
+
+_q_get_enrollment = Q("""
+SELECT
+    _id AS enrollment_internal_id,
+    enrollment_state
```
mmmarcos: This differs from `pki_accept.py` where the request already checks for the submitted state:
```python
enrollment_state = 'SUBMITTED' AS in_submitted_state
```

For the reject, you check the state below which, in the end, it should have the same result, but better be consistent and use either method in both cases.

IMO it seems better to put the `'SUBMITTED'` in the SQL query because it's a value stored in the database.

touilleMan: good catch ! done :+1:

Issue #11567, opened 2025-11-05
Client pki list does not validate submitted request


FirelightFlagboy:
The code at

https://github.com/Scille/parsec-cloud/blob/ffb33420487d666c2fcebb26d6a091501bb57d37/libparsec/crates/client/src/client/pki_enrollment_list.rs#L25-L39

Only list the value from the server without doing any validation (do we trust the request)

PR #11578, opened 2025-11-07
feat(bindgen): Add `PkiSignatureAlgorithm`

FirelightFlagboy:


AureliaDolo (approved):

FirelightFlagboy:
Maybe it's not require once #11567 is fixed and the returned type simplified

FirelightFlagboy:
Needed because we need #11536 first

PR #11584, opened 2025-11-07
Update macos from 13 to 15 in package CI

Ironicbay:


.github/workflows/package-desktop.yml
```diff
@@ -191,7 +191,7 @@ jobs:
             artifact_tag: windows-exe
           - name: 🍎 macOS x86
             platform: macos
-            os: macos-13
+            os: macos-15
```
FirelightFlagboy: **issue:** `macos-15` by default use arm cpu, we need to specify intel

```suggestion
            os: macos-15-intel
```

PR #11611, opened 2025-11-12
fix(pki): Verify pending requests

### LINKED ISSUE #11567: Client pki list does not validate submitted request

FirelightFlagboy:
`client::pki_enrollment_list` did not check pending request for trust and valid payload. Now invalid request is skipped by `pki_enrollment_list`.

Closes #11567


libparsec/crates/client/src/client/pki_enrollment_list.rs
FirelightFlagboy, line 1: **todo (blocking):** Use `load_submit_payload`

PR #11616, opened 2025-11-13
Added missing newsfragments for release 3.6.1

Ironicbay:


FirelightFlagboy (approved):

newsfragments/11529.bugfix.rst
```diff
@@ -0,0 +1 @@
+Fixed an issue that would prevent the application window from displaying when installed with snap
```
FirelightFlagboy: **nitpick:** 
```suggestion
Linux: Fixed an issue that would prevent the application window from displaying when installed with snap
```

PR #11618, opened 2025-11-13
Openboa support: part 1 `device_file_openbao.json` schema & `libparsec_openbao`

touilleMan:


cli/src/commands/device/change_authentication.rs
```diff
@@ -77,6 +77,15 @@ pub async fn main(args: Args) -> anyhow::Result<()> {
         AvailableDeviceType::Recovery => {
             return Err(LoadAndUnlockDeviceError::UnsupportedAuthentication(device.ty).into());
         }
+
+        AvailableDeviceType::OpenBao { .. } => {
+            // In theory we should support this authentication method here,
+            // however:
+            // - It is cumbersome since it requires opening a browser window for login
+            //   and redirect it result to a server listening on localhost...
```
FirelightFlagboy: **typo:** 
```suggestion
            //   and redirect its result to a server listening on localhost...
```

cli/src/commands/device/overwrite_server_url.rs
```diff
@@ -56,6 +56,15 @@ pub async fn main(args: Args) -> anyhow::Result<()> {
         AvailableDeviceType::Recovery => {
             return Err(LoadAndUnlockDeviceError::UnsupportedAuthentication(device.ty).into());
         }
+
+        AvailableDeviceType::OpenBao { .. } => {
+            // In theory we should support this authentication method here,
+            // however:
+            // - It is cumbersome since it requires opening a browser window for login
+            //   and redirect it result to a server listening on localhost...
```
FirelightFlagboy: **typo:** 
```suggestion
            //   and redirect its result to a server listening on localhost...
```

cli/src/utils.rs
```diff
@@ -271,6 +271,16 @@ pub async fn load_and_unlock_device(
                 device.ty,
             ));
         }
+        AvailableDeviceType::OpenBao { .. } => {
+            // In theory we should support this authentication method here,
+            // however:
+            // - It is cumbersome since it requires opening a browser window for login
+            //   and redirect it result to a server listening on localhost...
```
FirelightFlagboy: **typo:** 
```suggestion
            //   and redirect its result to a server listening on localhost...
```

libparsec/crates/openbao/src/lib.rs
```diff
@@ -0,0 +1,176 @@
...
+        if !matches!(rep.status(), StatusCode::OK) {
+            return Err(OpenBaoFetchOpaqueKeyError::BadServerResponse(
+                anyhow::anyhow!("Bad status code: {}", rep.status()),
+            ));
+        }
+
+        // See https://openbao.org/api-docs/secret/kv/kv-v2/#read-secret-version
+        let rep_json = rep.json::<serde_json::Value>().await.map_err(|err| {
+            OpenBaoFetchOpaqueKeyError::BadServerResponse(anyhow::anyhow!("Not JSON: {:?}", err))
+        })?;
```
FirelightFlagboy, line 103: **issue:** Why not using a serde schema ? here we trade a serde error for a panic at runtime, I largely prefer the first

libparsec/crates/openbao/src/lib.rs
```diff
@@ -0,0 +1,176 @@
...
+use libparsec_types::{prelude::*, uuid::Uuid};
+
+#[derive(Debug)]
+pub struct OpenBaoCmds {
+    client: reqwest::Client,
+    // Don't store the server URL as `url::Url` since we use as base URL for building
+    // new ones, we also need to ensure that the URL is not a cannot-be-a-base.
+    // Hence it is simpler to parse and do the cannot-be-a-base check together
+    // where the new URL is needed (as `OpenBaoCmds::new` cannot fail).
+    openbao_server_url: String,
```
FirelightFlagboy, line 15: **question:** I fail to see how it's an interesting property to be able to create `OpenBaoCmds` with an invalid URL that will be checked later on instead of during the creation.

libparsec/crates/platform_device_loader/src/native/mod.rs
```diff
@@ -192,8 +193,14 @@ pub(super) async fn load_ciphertext_key(
...
-                        AccountVaultOperationsFetchOpaqueKeyError::Offline(err) => {
-                            LoadCiphertextKeyError::RemoteOpaqueKeyFetchOffline(err)
+                        // Note it's important to serialize the whole error (and not the sub-error
+                        // it wraps): by doing so we keep the information of which server was
+                        // involved (since different save strategy communicate with different
+                        // server, e.g. Parsec account vs OpenBao).
+                        err @ AccountVaultOperationsFetchOpaqueKeyError::Offline(_) => {
+                            LoadCiphertextKeyError::RemoteOpaqueKeyFetchOffline(anyhow::anyhow!(
+                                err
+                            ))
```
FirelightFlagboy: **issue:** The comment indicate something requiring a special treatment.
For me it's indicating that `anyhow` is not adapted here, and we should use an enum or a struct to indicated the provenance of the error

cli/src/commands/device/change_authentication.rs
touilleMan: :+1:

cli/src/commands/device/overwrite_server_url.rs
touilleMan: :+1:

cli/src/utils.rs
touilleMan: :+1:

libparsec/crates/openbao/src/lib.rs
touilleMan: there is no risk of panic at runtime here:
- if the payload is not a valid JSON, we get a `OpenBaoFetchOpaqueKeyError::BadServerResponse("NOT JSON: ...")`
- if the payload is a valid JSON, but its `data/data/opaque_key` is missing/invalid, we get a `OpenBaoFetchOpaqueKeyError::BadServerResponse("Missing or invalid `data/data/opaque_key` field")`

On top of that I'd rather avoid using a serde schema here:
- The schema to validate is extremely simple, so doing by hand is not cumbersome
- The `data/data/opaque_key` field is encoded as base64. It would require some advanced tricks to include the base64 deserialization in the schema declaration, and the alternative is to have the base64 deserialization done after the schema deserialization (hence ending up with something pretty similar to the current code)
- The schema is not that well known (see https://openbao.org/api-docs/secret/kv/kv-v2/#read-secret-version, the types of the field is not specified), so I'd rather only parse the minimum actually needed.

libparsec/crates/openbao/src/lib.rs
touilleMan: The gain is to have `OpenBaoCmds::new` not needing to return a `Result`

Best practice is to validate early, and pass object representing some data already validated.

So in theory we would have `OpenBaoCmds::new(openbao_server_url: Url)`.

However this `Url` type is not strict enough for our needs: it can represent a cannot-be-a-base URL which we don't want.
So we end up with a choice:
- do the `cannot-be-a-base` validation in `OpenBaoCmds::new` (but we need to return a `Result` then)
- take the url as-is and only complain about `cannot-be-a-base` later on

Now one additional point: we are going to use this URL in `OpenBaoCmds::fetch_opaque_key(openbao_ciphertext_key_path: &str)` to do construct a new URL
Since this URL construction can fail if `openbao_ciphertext_key_path` is invalid, we have no choice but to have ``OpenBaoCmds::fetch_opaque_key` returning a `Result` with an error type related to invalid URL.

So we can move the `cannot-be-a-base` validation in the `OpenBaoCmds::fetch_opaque_key` and re-use this invalid URL error.
And while we are at it, we can also do the Url parsing in `OpenBaoCmds::fetch_opaque_key` to further simplify the error handling from the caller (since the caller no longer have to manually do the raw url to `Url` parsing by himself).

So in the end it's just simpler for the caller: he can create a `OpenBaoCmds` without any error to deal with, then have a single type of error related to unvalid URL to deal with when calling `OpenBaoCmds::fetch_opaque_key`

libparsec/crates/platform_device_loader/src/native/mod.rs
touilleMan: Note the commend is about the fact we often do in the code:
```rust
  match err {
    A::Foo(err) => B::Foo(err),
  }
```
This unwrapping+rewrapping only has sens if the wrapper doesn't provide some information compared to the new wrapper.
But this is not the case here:
- the `AccountVaultOperationsFetchOpaqueKeyError::Offline` wrapper type gives the information that the server that couldn't be reached was the Parsec account server
- for the openbao strategy, there is a `OpenBaoOperationsFetchOpaqueKeyError::NoServerResponse` wrapper that gives the information that the openbao server couldn't be reached
Both errors end up being wrapped in a `LoadCiphertextKeyError::RemoteOpaqueKeyFetchOffline`, but it is important to keep track of what server couldn't be reached.

On top of that regarding the use of `anyhow` within `LoadCiphertextKeyError::RemoteOpaqueKeyFetchOffline`, there is two things to consider:
1. the need for the caller to differentiate between errors types
2. the need to further provide information about the error

Here 1 is done by having `RemoteOpaqueKeyFetchOffline` and `RemoteOpaqueKeyFetchFailed` errors variant (so the caller can deal differently depending on if the opaque key fetch failed because the server couldn't be reached or because some unexpected error occured)

Regarding 2, there is no need to introduce error variant for this given the caller is not going to deal with those different error in different ways, it is just going to be displayed as further explanation (typically useful for Sentry bug report and for end-user when the error is about some unexpected error to provide a bit more context).

In a nutshell we have limited the error variants to what the caller needs (which have the added benefit of not requiring change on the caller side if we create new ways of failing, as long as those new new still correspond to the broad `RemoteOpaqueKeyFetchOffline/RemoteOpaqueKeyFetchFailed` errors)

libparsec/crates/account/src/upload_opaque_key_in_vault.rs
```diff
@@ -11,8 +11,8 @@ pub enum AccountUploadOpaqueKeyInVaultError {
     BadVaultKeyAccess(DataError),
     #[error("Cannot communicate with the server: {0}")]
     Offline(#[from] ConnectionError),
-    #[error(transparent)]
-    Internal(#[from] anyhow::Error),
+    #[error("The Parsec account server returned an unexpected response: {0}")]
+    BadServerResponse(anyhow::Error),
```
AureliaDolo, line 15: nit : 
```suggestion
    BadAccountServerResponse(anyhow::Error),
```

libparsec/crates/openbao/src/lib.rs
AureliaDolo: What about having a CanBeABaseUrl custom type that could be passed to OpenBaoCmds::new so the caller has to pass something already checked ?  The caller will still have to deal with an error, though not during OpenBaoCmds creation.

libparsec/crates/openbao/src/lib.rs
AureliaDolo: > The schema is not that well known

serde ignores fields when they are not specified, so you can only define what we need 
https://play.rust-lang.org/?version=stable&mode=debug&edition=2024&gist=621cbd579f7890bcfdf7c14a0076c43c

libparsec/crates/openbao/src/lib.rs
touilleMan: > What about having a CanBeABaseUrl custom type that could be passed to OpenBaoCmds::new so the caller has to pass something already checked ?

I think this is too cumbersome:
- it introduces more code
- the caller has to deal with errors when converting a string into a `CanBeABaseUrl`
- the caller still has to deal with errors when calling `fetch_opaque_key`

On the other hand with the current solution, the caller only has to deal with error when calling `fetch_opaque_key`

libparsec/crates/account/src/upload_opaque_key_in_vault.rs
touilleMan: I'd rather stay with `BadServerResponse`:
- In most of the codebase there is only one server (i.e. the Parsec server) so it's always obvious what we are talking about (the only exception being the OpenBao server)
- If we specify in the error name "Account", we should also do that for the `Offline` error since it is also about communicating with this server...
- ...and we open the pandora's box for all the errors involving the server that are in the `libparsec_account` crate !

However I've update the error messages to mention the failure is related to the Parsec account server (and done the same for the errors in `fetch_opaque_key_in_vault`), this is because those methods are used the `Device(Save|Access)Strategy` where they are hidden behind a generic trait.
So those errors reach the end-user heavily wrapped and the message is the way to have an idea of which server (so openbao server or the parsec account server depending on the strategy has failed)

libparsec/crates/openbao/src/lib.rs
```diff
@@ -0,0 +1,176 @@
...
+                anyhow::anyhow!("Bad status code: {}", rep.status()),
+            ));
+        }
+
+        // See https://openbao.org/api-docs/secret/kv/kv-v2/#read-secret-version
+        let rep_json = rep.json::<serde_json::Value>().await.map_err(|err| {
+            OpenBaoFetchOpaqueKeyError::BadServerResponse(anyhow::anyhow!("Not JSON: {:?}", err))
+        })?;
+
+        let key = rep_json["data"]["data"]["opaque_key"]
```
touilleMan, line 108: Add a comment to explain this doesn't panic if the field is missing

libparsec/crates/platform_device_loader/src/testbed.rs
```diff
@@ -332,6 +332,21 @@ pub(crate) fn maybe_load_device(
...
                             return None;
                         }
+                        // This assert does nothing at runtime, however at compilation time it
+                        // ensures `DeviceAccessStrategy` (and incidentally `DeviceSaveStrategy`
+                        // since they are always in sync) contains the variants we expect.
+                        // This is important since `match (access, c_save_strategy)` is not
+                        // an exhaustive match, and hence any new variant added to
+                        // `DeviceAccessStrategy` will silently fall through the default
+                        // case and create a bug :/
+                        debug_assert!(matches!(
```
touilleMan: This doesn't work at compilation, replace it by:
```
 #[pragma ignore clippy error or something]
  match access {
    DeviceAccessStrategy::Password { .. }
      | DeviceAccessStrategy::Smartcard { .. }
      | DeviceAccessStrategy::Keyring { .. }
      | DeviceAccessStrategy::AccountVault { .. }
      | DeviceAccessStrategy::OpenBao { .. }
}
```

libparsec/crates/platform_device_loader/src/native/mod.rs
```diff
@@ -204,6 +211,32 @@ pub(super) async fn load_ciphertext_key(
...
+                    .fetch_opaque_key(device.openbao_ciphertext_key_path.clone())
+                    .await
+                    .map_err(|err| match err {
+                        // Note it's important to serialize the whole error (and not the sub-error
+                        // it wraps): by doing so we keep the information of which server was
+                        // involved (since different save strategy communicate with different
+                        // server, e.g. Parsec account vs OpenBao).
+                        err @ (OpenBaoOperationsFetchOpaqueKeyError::BadURL(_)
+                        | OpenBaoOperationsFetchOpaqueKeyError::BadServerResponse(_)) => {
+                            LoadCiphertextKeyError::RemoteOpaqueKeyFetchFailed(anyhow::anyhow!(err))
```
touilleMan:
```suggestion
                            LoadCiphertextKeyError::RemoteOpaqueKeyFetchFailed{
                               provenance: RemoteOperation::OpenBao,
                               error: anyhow::anyhow!(err))  // No need to wrap the whole initial error
                             }
```

libparsec/crates/platform_device_loader/src/testbed.rs
touilleMan: done.

libparsec/crates/openbao/src/lib.rs
touilleMan: done.

libparsec/crates/platform_device_loader/src/native/mod.rs
touilleMan: done.

touilleMan:
merged with #11666 \o/

PR #11620, opened 2025-11-13
Openboa support: part 3 `get_server_config` API

touilleMan:


server/parsec/cli/run.py
```diff
@@ -165,6 +195,90 @@ def handle_parse_result(
...
+    envvar="PARSEC_ALLOWED_CLIENT_AGENT",
+    show_envvar=True,
+    type=click.Choice(AllowedClientAgent),
+    help="""Limit the type of Parsec client allowed to connect
+(default: NATIVE_OR_WEB if `--with-client-web-app` is used, NATIVE_ONLY else)
+\b
+- NATIVE_ONLY: Only desktop client is allowed
+- NATIVE_OR_WEB: Desktop and web clients are allowed
+""",
+    default=None,
```
FirelightFlagboy, line 182: **question:** No default to `native_only` or `native_or_web` here ?

FirelightFlagboy: After reading the implementation below, I would like a small comment that the default value depends on other parameters (`webapp_path`)

server/parsec/cli/run.py
```diff
@@ -539,6 +628,44 @@ def run_cmd(
 
         jinja_env = get_environment(template_dir)
 
+        match (allowed_client_agent, with_client_web_app):
+            case (AllowedClientAgent() as cooked_allowed_client_agent, _):
+                pass
```
FirelightFlagboy, line 606: **nitpick:** I would have added a warning log if `with_client_web_app` is configured but the configuration does not allow to web clients

libparsec/src/config.rs
```diff
@@ -74,3 +76,75 @@ impl From<ClientConfig> for libparsec_client::ClientConfig {
...
+                libparsec_protocol::anonymous_server_cmds::v5::server_config::OpenBaoConfig::Disabled => None,
+                libparsec_protocol::anonymous_server_cmds::v5::server_config::OpenBaoConfig::Enabled { auths, secret, server_url } => Some(
+                    OpenBaoConfig {
+                        server_url,
+                        secret,
+                        auths: auths.into_iter().filter_map(|auth| match auth.id.as_ref() {
+                            "OIDC_HEXAGONE" => Some(OpenBaoAuthConfig::OIDCHexagone { mount_path: auth.mount_path }),
+                            "OIDC_PRO_CONNECT" => Some(OpenBaoAuthConfig::OIDCProConnect { mount_path: auth.mount_path }),
+                            // Unknown kind of auth, just ignore it
+                            _ => None,
```
AureliaDolo: Maybe add a warning here. IMHO in the best case we support a new auth and the client is late to the update, but worst case the server might be up to something weird.

server/parsec/cli/run.py
```diff
@@ -165,6 +195,90 @@ def handle_parse_result(
...
+with a just the account password.
+
+A key feature of the account is its vault allowing the user to store sensitive
+data (e.g. device keys) server-side, however it comes at the  cost of slightly
+less security (since the server can try to brute-force the user account's
+password to gain access to his vault).
+\b
+- DISABLED: Users cannot create account
+- ENABLED_WITH_VAULT: User can create account and store sensitive data in it.
+- ENABLED_WITHOUT_VAULT: User can create account, but should not store sensitive data
```
AureliaDolo, line 200:
```suggestion
- ENABLED_WITHOUT_VAULT: User can create account, but should not store sensitive data (advisory only)
```

server/parsec/cli/run.py
```diff
@@ -539,6 +628,44 @@ def run_cmd(
...
+                    OpenBaoAuthConfig(
+                        id=OpenBaoAuthType.PRO_CONNECT,
+                        mount_path=openbao_auth_pro_connect,
+                    )
+                )
+
+            openbao_config = OpenBaoConfig(
+                server_url=openbao_server_url,
+                secret_mount_path=openbao_secret_mount_path,
+                auths=auths,
```
AureliaDolo, line 644: What happens if neither hexagone nor pro connect are set up ?

server/parsec/cli/run.py
touilleMan: > I would like a small comment that the default value depends on other parameters (webapp_path)

I agree with this, though I'd rather not do it since this code is removed in the next PR (see https://github.com/Scille/parsec-cloud/pull/11628)

server/parsec/cli/run.py
touilleMan: same as above: a warning would be a nice improvement, but I'd rather not do it since this code is doesn't known it... but it is already dead ! NANI !???!

libparsec/src/config.rs
touilleMan: Good point ! I've added a warn log there :+1:

server/parsec/cli/run.py
touilleMan: :+1:

server/parsec/cli/run.py
touilleMan: Then the GUI won't display any SSO connection options, which is not great for the user, but not an issue from the Parsec server point of view ^^

This is a good point though: I've added a simple check there to return an error if no ` --openbao-auth-*` has been provided :+1:

server/parsec/cli/run.py
```diff
@@ -95,6 +98,33 @@ def handle_parse_result(
         return value, args
 
 
+class OpenBaoOption(click.Option):
```
touilleMan: Dead code to remove

server/parsec/components/events.py
```diff
@@ -477,3 +486,62 @@ async def sse_api_events_listen(
                 # It's vital to unregister the client here given the memory location of the
                 # client (and hence the id resulting of it) will most likely be re-used !
                 self._registered_clients.pop(id(client_ctx))
+
+    @api
```
touilleMan, line 495: Add a sorry comment

server/parsec/cli/run.py
touilleMan: done.

server/parsec/components/events.py
touilleMan: done.

touilleMan:
merged with #11666 \o/

PR #11628, opened 2025-11-14
Openboa support: part 5 remove allowed client agent config

touilleMan:


touilleMan:
> I think a newsfragment would be nice for that.


Done :+1:

touilleMan:
merged with #11666 \o/

PR #11639, opened 2025-11-17
[MS] Openbao with bindings

Max-7:
Don't forget to recompile everything.

Launch the testbed as usual.

Since the Parsec server is now responsible for giving all the information needed to connect with SSO, this configuration can be changed in `server/parsec/cli/testbed.py`.

If nothing is changed, is should work by mocking everything.
Here's what's needed to test with a real openbao server and real SSO.
```
diff --git a/server/parsec/cli/testbed.py b/server/parsec/cli/testbed.py
index 7bbb149ff..11a3e5182 100644
--- a/server/parsec/cli/testbed.py
+++ b/server/parsec/cli/testbed.py
@@ -713,8 +713,8 @@ async def testbed_backend_factory(
         allowed_client_agent=AllowedClientAgent.NATIVE_OR_WEB,
         organization_spontaneous_bootstrap=True,
         openbao_config=OpenBaoConfig(
-            server_url=server_addr.to_http_url("/testbed/mock/openbao"),
-            secret_mount_path="secret/parsec-keys",
+            server_url="https://openbao-dev.parsec.cloud",
+            secret_mount_path="parsec-keys",
             auths=[
                 OpenBaoAuthConfig(
                     id=OpenBaoAuthType.HEXAGONE,
@@ -722,7 +722,7 @@ async def testbed_backend_factory(
                 ),
                 OpenBaoAuthConfig(
                     id=OpenBaoAuthType.PRO_CONNECT,
-                    mount_path="auth/pro_connect",
+                    mount_path="auth/oidc",
                 ),
             ],
         ),
```

For ProConnect, the email `test@fia1.fr` can be used.

Both the real openbao server and the testbed should be tested to make sure they work properly.

What should be tested:
- if the SSO auth is available when using the local server (creating an org on a custom server for example)
- if the SSO auth is not available when using a server without the configuration (creating an org on the trial server for example)
- should be a valid auth mode for new users and devices too
- once a device is created with SSO, it can be accessed (even after a refresh or closing the app)
- check that errors are handled properly (using localhost for the server but shutting down internet access for example, and whatever else your deviant mind can think of)

### DIFF
--- a/client/src/views/devices/DevicesPage.vue
+++ b/client/src/views/devices/DevicesPage.vue
@@ -102,6 +102,7 @@ async function onAddDeviceClick(): Promise<void> {
   const modal = await modalController.create({
     component: GreetDeviceModal,
     canDismiss: true,
+    backdropDismiss: false,
     cssClass: 'greet-organization-modal',
     componentProps: {
       informationManager: informationManager,


.github/workflows/ci-web.yml
```diff
@@ -37,6 +37,8 @@ jobs:
         image: ghcr.io/scille/parsec-cloud/parsec-testbed-server:3.6.2-a.0.dev.20412.600fb06
         ports:
           - 6777:6777
+        env:
+          PARSEC_SERVER_ADDR: parsec3://parsec-testbed-server:6777?no_ssl=true
```
FirelightFlagboy, line 42: **todo:** Need a comment about why we set `PARSEC_SERVER_ADDR` to itself

.github/workflows/ci-web.yml
```diff
@@ -170,6 +172,8 @@ jobs:
...
```
FirelightFlagboy, line 178: **todo:** Need a comment about why we set `PARSEC_SERVER_ADDR` to itself

FirelightFlagboy (dismissed):

client/src/components/devices/ChooseAuthentication.vue
```diff
@@ -63,7 +63,7 @@
         </ion-radio>
 
         <ion-radio
-          v-show="isOpenBaoAvailable()"
+          v-show="serverConfig?.openbao && serverConfig?.openbao.auths.some((auth) => isSSOProviderHandled(auth.tag))"
```
Ironicbay: Might be nice to move this condition to a `computed` for clarity, esp. since it's used twice

client/src/components/devices/ChooseAuthentication.vue
```diff
@@ -132,10 +140,19 @@
           />
         </div>
         <sso-provider-card
-          :provider="SSOProvider.ProConnect"
-          @click="onSSOLoginClicked(SSOProvider.ProConnect)"
+          v-for="auth in serverConfig?.openbao.auths.filter((auth) => isSSOProviderHandled(auth.tag))"
```
Ironicbay, line 135: Same here, a `computed` might make this clearer.

client/src/components/devices/ChooseAuthentication.vue
```diff
@@ -247,20 +268,37 @@ async function areFieldsCorrect(): Promise<boolean> {
...
 
-async function onSSOLoginClicked(_provider: SSOProvider): Promise<void> {
-  const result = await openBaoConnect();
+async function onSSOLoginClicked(provider: OpenBaoAuthConfigTag): Promise<void> {
+  if (!props.serverConfig?.openbao) {
+    error.value = 'Authentication.openBaoUnavailable';
+    window.electronAPI.log('error', 'OpenBao not enabled on this server');
+    return;
+  }
+  const auth = props.serverConfig.openbao.auths.find((v) => v.tag === provider);
```
Ironicbay: What's `v`? Maybe use a more explicit name

client/src/services/openBao.ts
```diff
@@ -121,89 +167,95 @@ async function getToken(code: string, state: string): Promise<OpenBaoResult<{ to
...
+          reject();
+        }
       });
-    }
+    });
+  } else {
+    const bc = new BroadcastChannel('openbao-oidc');
+    resultPromise = new Promise<{ code: string; state: string }>((resolve, reject) => {
+      bc.onmessage = (event): void => {
+        // Set a 5min timeout
```
Ironicbay, line 201: Why did you choose 5 minutes? Isn't it a bit long for a connection timeout?

client/src/views/organizations/creation/CreateOrganizationSaas.vue
```diff
@@ -138,12 +142,25 @@ const currentError = ref<Translatable | undefined>(undefined);
 const saveStrategy = ref<DeviceSaveStrategy | undefined>(undefined);
 const availableDevice = ref<AvailableDevice | undefined>(undefined);
 const sequesterKey = ref<string | undefined>(undefined);
+const serverConfig = ref<ServerConfig | undefined>(undefined);
 
 onMounted(async () => {
+  let server!: string;
```
Ironicbay: You could declare `let server: string = Env.getSaasServers()[0] ?? '';` and remove the `else` repeated twice below

client/src/views/organizations/creation/CreateOrganizationTrial.vue
```diff
@@ -88,14 +89,25 @@ const saveStrategy = ref<DeviceSaveStrategy | undefined>(undefined);
 const availableDevice = ref<AvailableDevice | undefined>(undefined);
 const currentError = ref<Translatable | undefined>(undefined);
 const organizationName = ref<OrganizationID | undefined>(undefined);
-const { getOpenBaoClient, isOpenBaoAvailable } = useOpenBao();
+const serverConfig = ref<ServerConfig | undefined>(undefined);
 
 onMounted(async () => {
+  let server!: string;
```
Ironicbay: Same here, you could assign `server` to `getTrialServerAddress()` by default

client/src/components/devices/ChooseAuthentication.vue
```diff
@@ -122,7 +122,15 @@
...
       </div>
 
-      <div v-if="authentication === DeviceSaveStrategyTag.OpenBao">
+      <!-- eslint-disable vue/html-indent -->
+      <div
+        v-if="
+          authentication === DeviceSaveStrategyTag.OpenBao &&
+          serverConfig?.openbao &&
+          serverConfig?.openbao.auths.some((auth) => isSSOProviderHandled(auth.tag))
+        "
```
NicoTuxx: We should not disable eslint in html just to put long conditions on multi-line.
Anyway it is not really practical & readable, should be move to a variable / function.

client/src/components/devices/ChooseAuthentication.vue
```diff
@@ -247,20 +268,37 @@ async function areFieldsCorrect(): Promise<boolean> {
...
-  const result = await openBaoConnect();
+async function onSSOLoginClicked(provider: OpenBaoAuthConfigTag): Promise<void> {
+  if (!props.serverConfig?.openbao) {
+    error.value = 'Authentication.openBaoUnavailable';
+    window.electronAPI.log('error', 'OpenBao not enabled on this server');
+    return;
+  }
+  const auth = props.serverConfig.openbao.auths.find((v) => v.tag === provider);
+  if (!auth) {
+    error.value = 'Authentication.invalidOpenBaoData';
```
NicoTuxx, line 285: Could use an enum to list this error values.

client/src/views/home/LoginPage.vue
```diff
@@ -47,9 +47,14 @@
         </div>
       </ion-card-content>
       <ion-footer class="login-card-footer">
+        <sso-provider-card
+          v-if="device.ty.tag === AvailableDeviceTypeTag.OpenBao"
+          :provider="(device.ty as AvailableDeviceTypeOpenBao).openbaoPreferredAuthId as OpenBaoAuthConfigTag"
```
NicoTuxx, line 52: same as before, not that readable in the template, should be moved to TS

client/src/views/home/LoginPage.vue
```diff
@@ -96,25 +107,41 @@ const password = ref('');
...
+async function onLoginSSOClick(provider: OpenBaoAuthConfigTag): Promise<void> {
+  if (props.loginInProgress) {
+    return;
+  }
+  const serverConfigResult = await getServerConfig(props.device.serverAddr);
 
-  if (!connResult.ok) {
-    console.error('Failed to log in to openbao');
+  if (!serverConfigResult.ok || !serverConfigResult.value.openbao) {
+    errorMessage.value = 'Authentication.openBaoUnavailable';
```
NicoTuxx, line 123: same, could be listed in enum

client/src/views/organizations/creation/CreateOrganizationTrial.vue
```diff
@@ -88,14 +89,25 @@ const saveStrategy = ref<DeviceSaveStrategy | undefined>(undefined);
...
+    } else {
+      server = getTrialServerAddress();
     }
+  } else {
+    server = getTrialServerAddress();
+  }
+  const configResult = await getServerConfig(server);
+  if (configResult.ok) {
+    serverConfig.value = configResult.value;
+    console.log(serverConfig.value);
```
NicoTuxx: remove console log

NicoTuxx (changes requested):

client/src/services/openBao.ts
Max-7: It's not really a connection timeout in the network sense, it's the time the user has to enter their credentials.. It's arbitrary. I don't even know if it's really needed.

client/src/components/devices/ChooseAuthentication.vue
Max-7: You can't always do that. I can squeeze it in this case with a computed, but `v-if` is not only for show, it also tells the linter about the state of a variable in the scope, which it won't be able to deduce from a function or a variable. It's the reason why I didn't move it to a variable in the first place, I want the linter to know that everything in the scope of this div can access `serverConfig.openBao`, the variable will not be undefined.

And the eslint-disable in this case is because of a slight incompatibility between eslint and prettier on how the lines should be indented.

client/src/components/devices/ChooseAuthentication.vue
Max-7: Not really any point. Those are fixed strings passed to a function, we don't need a specific type for them.

client/src/views/home/LoginPage.vue
Max-7: Moving it to a function would make it more complex. Same as before, the `v-if` is also used to scope the variable: after this v-if, we can be sure that the device if of type OpenBao. Plus it's just a cast, there;s no logic.

client/src/components/devices/ChooseAuthentication.vue
NicoTuxx: to avoid to use pure string, prone to error.

client/src/views/home/LoginPage.vue
NicoTuxx: same reason as before.

client/src/components/devices/ChooseAuthentication.vue
```diff
@@ -178,6 +179,8 @@ const choosePasswordRef = useTemplateRef<InstanceType<typeof MsChoosePasswordInp
 const chooseCertificateRef = useTemplateRef<InstanceType<typeof ChooseCertificate>>('chooseCertificate');
 const openBaoClient = ref<undefined | OpenBaoClient>(undefined);
 const smartcardAvailable = ref(false);
+const isConnectWithSSO = ref(false);
```
Max-7: You don't need another variable, you already have `openBaoClient`

client/src/components/devices/SsoProviderCard.vue
```diff
@@ -27,13 +42,16 @@ import { OpenBaoAuthConfigTag } from '@/parsec';
...
+import { checkmarkCircle } from 'ionicons/icons';
+import { IonIcon, IonText } from '@ionic/vue';
 
 defineEmits<{
   (e: 'ssoSelected', provider: OpenBaoAuthConfigTag): void;
 }>();
 
 defineProps<{
   provider: OpenBaoAuthConfigTag;
+  isConnected: boolean;
```
Max-7: Make it optional so it doesn't break anything.

client/src/components/devices/SsoProviderCard.vue
```diff
@@ -19,6 +22,18 @@
...
     </div>
+    <div
+      class="connected"
+      v-else
+    >
+      <ion-icon
+        :icon="checkmarkCircle"
+        class="connected-icon"
+      />
+      <ion-text class="button-large connected-text">
```
Max-7: We could have 8 different SSOs here, which will make this a nightmare. Any way to make the "connected" part generic? Maybe display the SSO but disabled, and showing "You are connected" under?

client/src/views/home/LoginPage.vue
```diff
@@ -49,6 +49,7 @@
       <ion-footer class="login-card-footer">
         <sso-provider-card
           v-if="device.ty.tag === AvailableDeviceTypeTag.OpenBao"
+          :is-connected="isConnectWithSSO"
```
Max-7: It's not needed here: when the login is successful, we automatically call clientStart()

client/src/components/devices/ChooseAuthentication.vue
```diff
@@ -122,7 +122,7 @@
         <choose-certificate ref="chooseCertificate" />
       </div>
 
-      <div v-if="authentication === DeviceSaveStrategyTag.OpenBao">
+      <div v-if="authentication === DeviceSaveStrategyTag.OpenBao && serverConfig?.openbao && showOpenBaoAuth">
```
Ironicbay, line 125:
```suggestion
      <div v-if="authentication === DeviceSaveStrategyTag.OpenBao && showOpenBaoAuth">
```

Max-7: Nah, you do need that `v-if="serverConfig?.openBao"` so that the linter knows when I'm using `serverConfig.openBao` two lines later that it is actually set. It can't figure out that `showOpenBaoAuth` already tests for that.

PR #11646, opened 2025-11-17
[PKI] Fix pki_enroll_accept to accept the payload as struct instead of bytes

### LINKED ISSUE #11568: [pki] pass the submit payload as a struct format in pki accept
Current the expose `pki_enroll_accept` to the client expect submit_payload to be provided as bytes. instead we should provide it as a struct form since `pki_enroll_list` will provide struct

_Originally posted by @FirelightFlagboy in https://github.com/Scille/parsec-cloud/pull/11544#discussion_r2495205302_

AureliaDolo:
fix #11568

### DIFF
--- a/libparsec/src/client.rs
+++ b/libparsec/src/client.rs
@@ -709,11 +709,9 @@ pub async fn client_pki_enrollment_accept(
     enrollment_id: PKIEnrollmentID,
     human_handle: &HumanHandle,
     cert_ref: &X509CertificateReference,
-    submit_payload: Bytes,
+    submit_payload: PkiEnrollmentSubmitPayload,
 ) -> Result<(), PkiEnrollmentAcceptError> {
     let client = borrow_client(client)?;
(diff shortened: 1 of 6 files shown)


libparsec/src/client.rs
```diff
@@ -709,11 +709,9 @@ pub async fn client_pki_enrollment_accept(
     enrollment_id: PKIEnrollmentID,
     human_handle: &HumanHandle,
```
AureliaDolo, line 710: Already in the payload : remove

libparsec/crates/client/src/client/mod.rs
FirelightFlagboy, line 1: **issue:** I assumed that the correct human_handle is provided by the payload that is modified.

We should not modify the payload but keep the additional human_handle

AureliaDolo: Right ! I forgot that the greeter can modify the human handle

FirelightFlagboy (approved):

PR #11652, opened 2025-11-17
[MS] Implements async enrollments in the GUI (PKI+OpenBAO)

### LINKED ISSUE #12060: [MS] Asynchronous enrolment front implementation

• Combine all asynchronous enrolment request in the `InvitationsPage`

When the admin got a new invitation:

**For SSO**:
- He should Log in with the SSO provider (OpenBao for now)

**For PKI**:
- He must provide a certificate

### Mockups : 
https://github.com/Scille/parsec-cloud/issues/12015

_No response_

Max-7:
Closes #12060

# Setup

Testing this properly on a dev environment can be a bit tricky.

- Start with `python make.py i && python make.py ei && python make.py wi`
- Use the testbed server

### How to test

To purely check the design part (how it looks), it's possible to set `(window as any).TESTING_PKI = true` somewhere at the top of the main.ts, which will switch the PKI implementation to the mocked one. Usefull for e2e tests (since PKI is not available on web) or quickly checking the design, but of course, not enough to see if it actually works. It does help to check how some errors are displayed as they can be difficult to generate manually.

**It can only be properly tested on Electron Windows right now.**

- Since certificates are needed, a nice little script if provided. Launch a Powershell terminal, go to `parsec-cloud\libparsec\crates\platform_pki`, and run `.\examples\import_testpki_windows.ps1`. Just accept to add the certificate, and when you're killing the script, it will automatically remove the certificates for you (if you accept). You may need to provide specific permissions to be able to exec the script (something about `Exec-Policy`).
- Launch Electron, use `parsec3://127.0.0.1:6770/TestbedOrg1?no_ssl=true&a=async_enrollment` (change the number) to make an enrollment request. Select PKI, select Bob's certificate, and it should be good.
- Log in with Alice, in the requests page you should see the request and be able to accept it selecting Alice's certificate.

### What to test (non-exhaustive list)

- Enrollment from start to finish (making a request, validation, joining the org, share a file)
- Admin rejects the request
- Invalid certificate (old_bob is an expired certificate)
- Cancel the request
- Change authentication from/to smartcard
- Not displayed on web or platforms != Windows

## For SSO

Two things to check:

### With the mocks from the testbed

- OpenBao is mocked by the testbed, useful for testing
- Don't change anything, should work in both Electron and web
- Use `parsec3://127.0.0.1:6770/TestbedOrg1?no_ssl=true&a=async_enrollment` (change the number) to make a request. Depending on the platform, SSO should be the only available option. On Electron Windows, you get the choice between SSO and PKI. Click on ProConnect, everything should be automatic.
- Log in with Alice, in the request list you can accept the request.

### With a real OpenBao server

- Use the real OpenBao server but our testbed Parsec server, need a bit of configuration
- Apply this patch to remove HTTPS from our web server (openbao doesn't like it) andto setup Parsec's server properly
```
diff --git a/client/vite.config.ts b/client/vite.config.ts
index 4932f7777..9d89ee9cc 100644
--- a/client/vite.config.ts
+++ b/client/vite.config.ts
@@ -88,9 +88,9 @@ if (process.env.PARSEC_APP_SENTRY_AUTH_TOKEN) {
 }

 // 3) Add dev specific plugins
-if (process.env.NODE_ENV === 'development' || process.env.CI) {
-  plugins.push(basicSsl());
-}
+// if (process.env.NODE_ENV === 'development' || process.env.CI) {
+//   plugins.push(basicSsl());
+// }

 plugins.push(
   viteStaticCopy({
diff --git a/server/parsec/cli/testbed.py b/server/parsec/cli/testbed.py
index 94892ca1a..7bf95f503 100644
--- a/server/parsec/cli/testbed.py
+++ b/server/parsec/cli/testbed.py
@@ -984,8 +984,8 @@ async def testbed_backend_factory(
         fake_account_password_algorithm_seed=SecretKey(b"F" * 32),
         organization_spontaneous_bootstrap=True,
         openbao_config=OpenBaoConfig(
-            server_url=server_addr.to_http_url("/testbed/mock/openbao"),
-            secret_mount_path="secret",
+            server_url="https://openbao-dev.parsec.cloud",  # server_addr.to_http_url("/testbed/mock/openbao"),
+            secret_mount_path="parsec-keys",  #"secret",
             transit_mount_path="transit",
             auths=[
                 OpenBaoAuthConfig(
@@ -994,7 +994,7 @@ async def testbed_backend_factory(
                 ),
                 OpenBaoAuthConfig(
                     id=OpenBaoAuthType.PRO_CONNECT,
-                    mount_path="auth/pro_connect",
+                    mount_path="auth/oidc",
                 ),
             ],
         ),
```
- Same thing, use `test@fia1.fr` as the email

### What to test (non-exhaustive list)

- Enrollment from start to finish (making a request, validation, joining the org, share a file)
- Admin rejects the request
- Invalid certificate (old_bob is an expired certificate)
- Cancel the request
- Change authentication from/to openbao
- Available on all platforms

### DIFF
--- a/client/src/locales/en-US.json
+++ b/client/src/locales/en-US.json
@@ -1290,7 +1290,8 @@
             "sso": {
                 "title": "Single Sign-On",
                 "description": "Login with an external account",
-                "connected": "Connected"
+                "connected": "Connected",
+                "unavailable": "This method is not allowed by this server."
             }
         },
         "methodChosen": "Method chosen:",
--- a/client/src/locales/fr-FR.json
+++ b/client/src/locales/fr-FR.json
@@ -1290,7 +1290,8 @@
             "sso": {
                 "title": "Authentification unique (SSO)",
                 "description": "S'authentifier avec un compte externe",
-                "connected": "Connecté"
+                "connected": "Connecté",
+                "unavailable": "Cette méthode n'est pas autorisée par ce serveur."
             }
         },
         "methodChosen": "Méthode choisie :",
--- a/client/src/components/devices/ChooseAuthentication.vue
+++ b/client/src/components/devices/ChooseAuthentication.vue
@@ -25,6 +25,18 @@
           class="item-radio radio-list-item"
           label-placement="end"
           justify="start"
+          :value="DeviceSaveStrategyTag.Password"
+        >
+          <authentication-card
+            @click="onMethodSelected(DeviceSaveStrategyTag.Password)"
+            :state="getAuthCardState(AvailableDeviceTypeTag.Password)"
+            :auth-method="DeviceSaveStrategyTag.Password"
+          />
+        </ion-radio>
+        <ion-radio
+          class="item-radio radio-list-item"
+          label-placement="end"
+          justify="start"
           :value="DeviceSaveStrategyTag.Keyring"
           :disabled="!keyringAvailable || activeAuth === AvailableDeviceTypeTag.Keyring"
         >
@@ -32,18 +44,6 @@
             :auth-method="DeviceSaveStrategyTag.Keyring"
             :state="getAuthCardState(AvailableDeviceTypeTag.Keyring)"
             :disabled="!keyringAvailable"
-          />
-        </ion-radio>
-        <ion-radio
-          class="item-radio radio-list-item"
-          label-placement="end"
-          justify="start"
-          :value="DeviceSaveStrategyTag.Password"
-        >
-          <authentication-card
-            @click="onMethodSelected(DeviceSaveStrategyTag.Password)"
-            :state="getAuthCardState(AvailableDeviceTypeTag.Password)"
-            :auth-method="DeviceSaveStrategyTag.Password"
           />
         </ion-radio>
 
@@ -58,21 +58,22 @@
             @click="onMethodSelected(DeviceSaveStrategyTag.PKI)"
             :state="getAuthCardState(AvailableDeviceTypeTag.PKI)"
             :auth-method="DeviceSaveStrategyTag.PKI"
+            :disabled="!smartcardAvailable"
           />
         </ion-radio>
 
         <ion-radio
-          v-show="showOpenBaoAuth"
           class="item-radio radio-list-item"
           label-placement="end"
           :value="DeviceSaveStrategyTag.OpenBao"
           justify="start"
-          :disabled="activeAuth === AvailableDeviceTypeTag.OpenBao"
+          :disabled="!openBaoAuthAvailable || activeAuth === AvailableDeviceTypeTag.OpenBao"
         >
           <authentication-card
             :state="getAuthCardState(AvailableDeviceTypeTag.OpenBao)"
             @click="onMethodSelected(DeviceSaveStrategyTag.OpenBao)"
             :auth-method="DeviceSaveStrategyTag.OpenBao"
+            :disabled="!openBaoAuthAvailable"
           />
         </ion-radio>
       </ion-radio-group>
@@ -122,7 +123,7 @@
         <choose-certificate ref="chooseCertificate" />
       </div>
 
-      <div v-if="authentication === DeviceSaveStrategyTag.OpenBao && serverConfig?.openbao && showOpenBaoAuth">
+      <div v-if="authentication === DeviceSaveStrategyTag.OpenBao && serverConfig?.openbao && openBaoAuthAvailable">
         <div class="method-chosen">
           <ion-text class="method-chosen__title subtitles-sm">{{ $msTranslate('Authentication.methodChosen') }}</ion-text>
           <authentication-card
@@ -190,7 +191,7 @@ const querying = ref(false);
 
 const error = ref('');
 
-const showOpenBaoAuth = computed(() => {
+const openBaoAuthAvailable = computed(() => {
   return props.serverConfig?.openbao && props.serverConfig?.openbao.auths.some((auth) => isSSOProviderHandled(auth.tag));
 });
 
@@ -236,6 +237,9 @@ function getAuthCardState(auth: AvailableDeviceTypeTag): AuthenticationCardState
       }
       return auth === props.activeAuth ? AuthenticationCardState.Active : AuthenticationCardState.Default;
     case AvailableDeviceTypeTag.OpenBao:
+      if (!openBaoAuthAvailable.value) {
+        return AuthenticationCardState.Unavailable;
+      }
       return auth === props.activeAuth ? AuthenticationCardState.Active : AuthenticationCardState.Default;
     default:
       return AuthenticationCardState.Default;
--- a/client/src/components/profile/AuthenticationCard.vue
+++ b/client/src/components/profile/AuthenticationCard.vue
@@ -122,6 +122,7 @@ const methodConfig: Record<
     imageAlt: 'OpenBao',
     methodName: 'Authentication.method.sso.title',
     description: 'Authentication.method.sso.description',
+    unavailableExplanation: 'Authentication.method.sso.unavailable',
   },
 };
 
--- a/client/tests/e2e/helpers/assertions.ts
+++ b/client/tests/e2e/helpers/assertions.ts
@@ -484,4 +484,49 @@ export const expect = baseExpect.extend({
       pass: true,
     };
   },
+
+  async toHaveAuthentication(
+    authRadio: Locator,
+    state?: { passwordDisabled?: boolean; ssoDisabled?: boolean; pkiDisabled?: boolean; keyringDisabled?: boolean },
+  ): Promise<AssertReturnType> {
+    await baseExpect(authRadio).toHaveCount(4);
+    await baseExpect(authRadio.nth(0).locator('.authentication-card-text__title')).toHaveText('Password');
+    await baseExpect(authRadio.nth(1).locator('.authentication-card-text__title')).toHaveText('System authentication');
+    await baseExpect(authRadio.nth(2).locator('.authentication-card-text__title')).toHaveText('Smartcard');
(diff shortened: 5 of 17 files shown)


NicoTuxx:
When testing the PKI on Electron:
1. If we use the same link with the same certificate 2 times, the first request is cancelled and a new one pop, but on the user sight it's strange to see both:
image

2. If I select a certificate and click on Previous, then click on Continue, the certificate is not selected anymore:
image
image

3. When using an inappropriate certificate, the error message is really unclear for the end user:
image

4. The terms used in the admin interface needs improvements 'requests' hide in an 'invitation' menu is not obvious.
image

5. When managing a PKI request, if the certificate is inappropriate, the admin got the error message in a toast, only after setting up the user role, info... Can be really frustrating:
image

6. After creating 2 requests (Alice & Bob), I logged in as admin user, I accepted Alice request & rejected Bob's one. Then navigating to the home page, I can see what seems to be 2 problems:
- Bob request is shown as rejected, seems logic, but then the user has a modal when trying to remove the request, asking him to keep or clean this request => What's the goal here? Why a user would keep something that was rejected & he can do nothing about it?
- Alice request is shown with a "Join" button, it is unclear what is expected from the end user here, I already filled a request to join an organization, why do I have to do something more here?
image

7. After accepting the previous Alice request from step 6, I can see the organization on the home page in the list to log in. It's shown as clickable when hovering on the block, but when I click nothing happens at all:
image
In console:
image

8. Okay so big problem on this one I think, after steps 6 & 7, I created a new request with Bob certificate, then logged as admin again, I accepted the request BUT USING BOB CERTIFICATE INSTEAD OF ALICE ONE and it WORKS :|
image
Then I clicked on "Join" for Bob & used Alice certificate instead of Bob & it worked too, it should absolutely not.
Then the organization is showed in the home page list, but when clicked nothing happens & same error message than Alice on step 7:
image

9. When using an expired certificate (old_boby), the message is unclear for the end user:
image

10. When joining an organization, the UI is talking about PKI & certificates, but when switching auth mode from any to Smartcard, the terms are "Smartcard" & no PKI at all, it's confusing for the end user:
image

11. When switching auth mode from Smartcard, I can not choose to keep Smartcard but change my certificate, I'm forced to switch to something else like password & then comeback to Smartcard which is not user friendly at all:
image

12. Changing auth mode from Smartcard to something else does not work at all (tried password & system auth):
image

13. When changing auth mode from password to Smartcard, everything seems fine, but after logout, trying to login on home page, the organization is in the same state than step 7, with error in console:
image
Worth, after some seconds it seems to have impact the org list completely:
image

NicoTuxx:
When testing OpenBao mocked on Electron windows:

1. Could not do any testing, error at first step:
image
image

Max-7:
> 1. If we use the same link with the same certificate 2 times, the first request is cancelled and a new one pop, but on the user sight it's strange to see both:

We now automatically delete requests that have been cancelled. There's no point in every showing them since cancelling is the result of a user interaction.

> 2. If I select a certificate and click on Previous, then click on Continue, the certificate is not selected anymore:

The good old battle between v-if and v-show. I think it's a minor inconvenience and not worth making the code more complex.

> 5. When managing a PKI request, if the certificate is inappropriate, the admin got the error message in a toast, only after setting up the user role, info... Can be really frustrating:

Not much we can do, we can only validate at the end of the process. I think it's acceptable.
 
> 6. After creating 2 requests (Alice & Bob), I logged in as admin user, I accepted Alice request & rejected Bob's one. Then navigating to the home page, I can see what seems to be 2 problems:
> 
> * Bob request is shown as rejected, seems logic, but then the user has a modal when trying to remove the request, asking him to keep or clean this request => What's the goal here? Why a user would keep something that was rejected & he can do nothing about it?

I removed the modal, no point in asking.

> * Alice request is shown with a "Join" button, it is unclear what is expected from the end user here, I already filled a request to join an organization, why do I have to do something more here?

We do need the user to interact in order to transform the request into the real device. Is it a problem of how it's presented? Would something like "Finalize" be better?
I did remove the need to re-select a certificate, which already makes it better.

> 8. Okay so big problem on this one I think, after steps 6 & 7, I created a new request with Bob certificate, then logged as admin again, I accepted the request BUT USING BOB CERTIFICATE INSTEAD OF ALICE ONE and it WORKS :|

I don't really see that as a problem, it's more of a weird quirk when testing, you're not supposed to have everyone's certificates.

> Then I clicked on "Join" for Bob & used Alice certificate instead of Bob & it worked too, it should absolutely not.

Agreed, this is a problem.

> 11. When switching auth mode from Smartcard, I can not choose to keep Smartcard but change my certificate, I'm forced to switch to something else like password & then comeback to Smartcard which is not user friendly at all:

I've removed the condition, we can now change from PKI to PKI. It can look a bit weird as the `PKI` will have a checkmark next to it as it is the current auth method, but I don't really want to change it as it is a complex bit of code with enough weird code as is.

> <Different problems when saving a device>

It would seem those are specific to the testbed. It would have to be fixed, but it should work fine without using the testbed (by creating an org manually with `parsec3://localhost:6770?no_ssl=true`).

> When testing OpenBao mocked on Electron windows:
>
> Could not do any testing, error at first step:

Problem with the CSP on Electron that did not allow a connection to localhost:6770, it has been fixed.

NicoTuxx:
OpenBao with TestBed:

1. I create a request with SSO, then when the request is in "Waiting" status, I cancelled it. When login as Alice (admin), I can still see AND accept the cancelled request...
image
I can accept it & Bobby is  even enrolled properly...
image
But since I cancelled the request on Bobby's side, he will never be able to login to this organization:
image

NicoTuxx:
> We do need the user to interact in order to transform the request into the real device. Is it a problem of how it's presented? Would something like "Finalize" be better?
> I did remove the need to re-select a certificate, which already makes it better.

Yes I think it's only a presentation problem, maybe "Finalize" would make more sense. @fabienSvtr some thoughts on this matter?

fabienSvtr:
> > We do need the user to interact in order to transform the request into the real device. Is it a problem of how it's presented? Would something like "Finalize" be better?
> > I did remove the need to re-select a certificate, which already makes it better.
> 
> Yes I think it's only a presentation problem, maybe "Finalize" would make more sense. @fabienSvtr some thoughts on this matter?

I updated the text with `Add to my organizations` (find it more affording)
image

client/src/components/invitations/AsyncEnrollmentRequestItem.vue
```diff
@@ -0,0 +1,339 @@
...
+            class="request-type__label"
+            v-if="request.identitySystem.tag === AsyncEnrollmentIdentitySystemTag.PKI"
+          >
+            {{ $msTranslate('InvitationsPage.asyncEnrollmentRequest.type.pki') }}
+          </ion-text>
+          <ion-text
+            class="request-type__label"
+            v-if="request.identitySystem.tag === AsyncEnrollmentIdentitySystemTag.PKICorrupted"
+          >
+            {{ $msTranslate('InvitationsPage.asyncEnrollmentRequest.errors.pkiCorrupted') }}
```
Ironicbay:
```suggestion
            {{ $msTranslate('InvitationsPage.asyncEnrollmentRequest.type.pkiCorrupted') }}
```

client/src/components/organizations/OrganizationUserInformation.vue
```diff
@@ -117,15 +117,15 @@
           v-if="false"
```
Ironicbay, line 117: You can remove the `v-if="false"` \o/

client/src/components/invitations/AsyncEnrollmentRequestItem.vue
```diff
@@ -0,0 +1,339 @@
...
+  } else if (props.request.identitySystem.tag === AsyncEnrollmentIdentitySystemTag.PKICorrupted) {
+    canAccept.value = false;
+    reason = 'InvitationsPage.asyncEnrollmentRequest.errors.problemWithRequestCertificate';
+  } else {
+    canAccept.value = false;
+    reason = 'InvitationsPage.asyncEnrollmentRequest.errors.unknownIdentitySystem';
+  }
+
+  if (!canAccept.value && reason) {
+    attachMouseOverTooltip(rejectButtonRef.value?.$el, reason);
```
Ironicbay: Wouldn't it make more sense to attach the tooltip to the `request type` text instead of the button? In this case, `PKI (corrupted)`

Image

client/src/locales/en-US.json
```diff
@@ -1404,7 +1453,23 @@
...
+                "selected": "Certificate selected",
+                "description": "Certificate verified your identity and allows you to securely join the organization."
+            },
+            "notAvailable": "PKI is not available on this system.",
+            "buttons": {
+                "join": "Request to join organization",
+                "cancel": "Cancel",
+                "close": "Close"
+            }
+        }
```
Ironicbay: These locales aren't used

client/src/locales/fr-FR.json
```diff
@@ -1404,7 +1453,23 @@
...
+                "selected": "Certificat sélectionné",
+                "description": "Le certificat vérifie votre identité et vous permet de rejoindre l'organisation en toute sécurité."
+            },
+            "notAvailable": "La PKI n'est pas disponible sur ce système.",
+            "buttons": {
+                "join": "Demander à rejoindre l'organisation",
+                "cancel": "Annuler",
+                "close": "Fermer"
+            }
+        }
```
Ironicbay: Same, these are unused

NicoTuxx:
LGTM, jusrt when switching from SmartCard to SmartCard it's not clear that we can click on SmartCard choice, should be improved @fabienSvtr 
image
Can be done in another issue/PR

PR #11661, opened 2025-11-17
[MS] Fix new version changelog link

Ironicbay:


client/src/services/environment.ts
```diff
@@ -207,11 +216,12 @@ async function openContactLink(): Promise<void> {
 }
 
 async function openLicenseLink(): Promise<void> {
-  await openUrl(I18n.translate({ key: 'app.licenseLink', data: { version: APP_VERSION_PREFIX } }));
+  await openUrl(I18n.translate({ key: 'app.licenseLink', data: { version: CLEAN_APP_VERSION } }));
 }
 
 async function openChangelogLink(version?: string): Promise<void> {
-  await openUrl(I18n.translate({ key: 'app.history', data: { version: version ?? APP_VERSION_PREFIX } }));
+  const normalizedVersion = version ? await normalizeAppVersion(version) : undefined;
```
Max-7: Could just put `CLEAN_APP_VERSION` instead of `undefined`, which removes the need for another ternary.

client/src/services/environment.ts
```diff
@@ -163,8 +163,17 @@ function getOpenBaoServer(): string | undefined {
  Links
 */
 
-const CLEAN_APP_VERSION = `${APP_VERSION.slice(0, APP_VERSION.indexOf('+') === -1 ? undefined : APP_VERSION.indexOf('+'))}`;
-const APP_VERSION_PREFIX = `v${CLEAN_APP_VERSION}`;
+async function normalizeAppVersion(version: string): Promise<string> {
```
Max-7: Doesn't need to be async, it just does a little string formatting.

Max-7 (approved):

Issue #11648, opened 2025-11-17
Add server configuration for trust cert to use for pki certificate check


FirelightFlagboy:
The configuration should allow for a list of root certificates.
Th certs will then be used during `pki_submit` & `pki_accept` to check the trustchain of the provided certificates (leaf + intermediates)

PR #11665, opened 2025-11-17
[PKI] Add bindings to get pki enrollement addr.

### LINKED ISSUE #11651: [PKI] Bindings and client impl  to generate a pkiAddr
- [ ] bindings
- [ ] client implementation: must be admin to generate ParsecPkiEnrollmentAddr

AureliaDolo:
Fix #11651

### DIFF
--- a/libparsec/crates/client/src/client/pki_get_addr.rs
+++ b/libparsec/crates/client/src/client/pki_get_addr.rs
@@ -1,0 +1,9 @@
+// Parsec Cloud (https://parsec.cloud) Copyright (c) BUSL-1.1 2016-present Scille SAS
+
+use libparsec_types::ParsecPkiEnrollmentAddr;
+
+use crate::Client;
+
+pub async fn get_addr(client: &Client) -> ParsecPkiEnrollmentAddr {
+    ParsecPkiEnrollmentAddr::new(client.organization_addr(), client.organization_id().clone())
+}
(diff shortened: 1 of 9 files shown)


libparsec/crates/client/src/client/pki_get_addr.rs
```diff
@@ -0,0 +1,25 @@
...
+    Internal(#[from] anyhow::Error),
+}
+
+pub async fn get_addr(client: &Client) -> Result<ParsecPkiEnrollmentAddr, PkiGetAddrError> {
+    match client.get_current_self_profile().await? {
+        UserProfile::Admin => Ok(ParsecPkiEnrollmentAddr::new(
+            client.organization_addr(),
+            client.organization_id().clone(),
+        )),
+        UserProfile::Standard | UserProfile::Outsider => Err(PkiGetAddrError::AuthorNotAllowed),
```
FirelightFlagboy: **nitpick:** I don't think that's need since anyway the enrollment is validated by admins

AureliaDolo: Maybe if the link is shared too much, the admins may be flooded with illegitimate/invalid requests ?

FirelightFlagboy: The link being generated by an admin won't solve that as the link can be shared after

We have #11648 to prevent that

Max-7 (approved):
API looks good

PR #11666, opened 2025-11-17
Openboa support: part 6 correct `AvailableDevice::server_url` type

touilleMan:


client/src/views/home/HomePage.vue
```diff
@@ -360,10 +361,17 @@ async function handleQuery(): Promise<void> {
...
     const availableDevices = await listAvailableDevices();
-    const device = availableDevices.find((d) => {
-      const serverType = getServerTypeFromAddress(d.serverUrl);
-      return serverType === ServerType.Saas && d.organizationId === query.bmsOrganizationId;
-    });
+    let device = undefined;
+    for (const d of availableDevices) {
+      const result = await parseParsecAddr(d.serverAddr);
+      if (!result.ok) {
+        throw Error(`Invalid \`serverAddr\` field for AvailableDevice: ${d}`);
```
Max-7, line 368: Use `continue` instead of throwing an error. Any call to libparsec that fails is automatically logged so we're covered, might as well not interrupt the execution.

client/src/parsec/pki.ts
Max-7, line 1: Not necessary to update this file since it's all mocks that are going to be removed in the coming days. Let's avoid future git conflicts.

client/src/components/organizations/OrganizationSwitchPopover.vue
```diff
@@ -110,15 +110,15 @@ const filteredOrgs: Ref<Array<ConnectedOrganization>> = ref([]);
 
 onMounted(async () => {
   const result = await getLoggedInDevices();
-  connectedOrgs.value = result.map((info) => {
+  connectedOrgs.value = await result.map(async (info) => {
```
Max-7: I don't think that does what you expect it to do. I think it will return an array of Promises. Probably have to do `Promise.all()` on the result.

client/src/common/organization.ts
```diff
@@ -39,8 +40,11 @@ export function formatExpirationTime(duration: Duration): Translatable {
   return 'HomePage.organizationList.expiration.expired';
 }
 
-export function isTrialOrganizationDevice(device: AvailableDevice): boolean {
-  const url = new URL(device.serverUrl);
-  const serverType = getServerTypeFromHost(url.hostname, url.port.length > 0 ? parseInt(url.port) : undefined);
+export async function isTrialOrganizationDevice(device: AvailableDevice): Promise<boolean> {
+  const result = await libparsec.parseParsecAddr(device.serverAddr);
+  if (!result.ok) {
```
Max-7, line 45: Avoid throwing, just return false.

client/src/views/home/HomePage.vue
touilleMan: The idea here is `parseParsecAddr(d.serverAddr);` is always supposed to return an ok result since `d.serverAddr` has been obtained from a call to `libparsec` bindings.

So here what I want to do is an assert, but I don't know how to do that in Typescript... do you see a better way ?

> Any call to libparsec that fails is automatically logged

I guess when you say "that fails" you mean an unexpected error (typically passing the wrong type of argument). But here we focus on the the case the call succeed but returned an error result, which shouldn't be logged since it is a totally normal thing.

client/src/common/organization.ts
touilleMan: same as above: this code is an assert... but I don't know how to do assert in Typescript :/

client/src/components/organizations/OrganizationSwitchPopover.vue
touilleMan: done :+1:

client/src/parsec/pki.ts
touilleMan: This file has to be modified at least for its `confirmLocalJoinRequest()` function given it returns an `AvailableDevice`

Regarding the rest of the changes, I'd rather keep them since the mock are probably going to be the base for the actual code.
While the `server` -> `serverAddr` is trivial an just a minor cleanup, the conversion from a `ParsecOrganizationAddr` into a `ParsecAddr` made in `requestJoinOrganization` is more cunning (without the changes, the `ParsecOrganizationAddr` was directly used as a `ParsecAddr` which is a bug)

client/src/views/home/HomePage.vue
Max-7: Both exceptions and a function from libparsec returning an error (`.ok` is false) are logged. This can indeed lead to false positives but it's fine. You can take a look at https://github.com/Scille/parsec-cloud/blob/dee9e1c6b021df4fde4d2ab2148c6b315b457cd9/client/src/plugins/libparsec/index.ts#L51

For this specific error, I think continuing "silently" (it's still logged) is our best bet. We don't really have anything to show the user, we don't want to display an error, we don't want to interrupt the function.

bindings/generator/api/addr.py
```diff
@@ -56,24 +57,28 @@ class ParsedParsecAddr(Variant):
     class Server:
         hostname: str
         port: U32
```
FirelightFlagboy: **todo:** use u16 for port

libparsec/crates/types/src/data_macros.rs
```diff
@@ -27,3 +27,53 @@ macro_rules! impl_transparent_data_format_conversion {
 }
 
 pub use impl_transparent_data_format_conversion;
+
+#[macro_export]
+macro_rules! impl_data_format_conversion {
```
touilleMan: Add documentation with example

libparsec/crates/types/src/local_device_file.rs
```diff
@@ -7,16 +7,16 @@ use libparsec_serialization_format::parsec_data;
...
 };
 
 #[derive(Debug, Clone, Serialize, Deserialize, PartialEq, Eq)]
-#[serde(into = "DeviceFileKeyringData", from = "DeviceFileKeyringData")]
+#[serde(into = "DeviceFileKeyringData", try_from = "DeviceFileKeyringData")]
 pub struct DeviceFileKeyring {
     pub created_on: DateTime,
     pub protected_on: DateTime,
-    pub server_url: String,
+    pub server_url: ParsecAddr,
```
touilleMan, line 19: Add serialize/deserialize impl for `ParsecAddr` type instead of custom hooks

libparsec/crates/types/src/data_macros.rs
touilleMan: Remove from this PR as it is no longer needed

Create a dedicated PR and merge this macro with `impl_transparent_data_format_conversion`

libparsec/src/addr.rs
```diff
@@ -12,24 +12,28 @@ pub enum ParsedParsecAddr {
     Server {
         hostname: String,
         port: u32,
```
touilleMan:
```suggestion
        port: u16,
```
?

client/src/parsec/pki.ts
```diff
@@ -257,6 +257,11 @@ const REQUESTS: Array<_Request> = [];
 export async function requestJoinOrganization(
   link: ParsecOrganizationAddr,
 ): Promise<Result<LocalJoinRequest, RequestJoinOrganizationError>> {
+  const result = await parseParsecAddr(link);
+  if (!result.ok) {
+    throw new Error(`Invalid ParsecAddr \`${link}\``);
```
touilleMan:
```suggestion
    throw new Error(`Invalid ParsecOrganizationAddr \`${link}\``);
```

server/parsec/cli/testbed.py
```diff
@@ -537,7 +537,7 @@ async def test_openbao_hexagone_auth_url(request: Request):
         return Response("`redirect_uri` parameter required", status_code=400)
     redirect_uri = (
         # cspell: disable-next-line
-        f"{redirect_uri}?code={code}&state=st_VXANtcosv3cCZThfaPLz&iss={quote_plus(sso_base_url)}"  # 
+        f"{redirect_uri}?code={code}&state=st_VXANtcosv3cCZThfaPLz&iss={quote_plus(sso_base_url)}"  #
```
FirelightFlagboy: **todo:** 
```suggestion
        f"{redirect_uri}?code={code}&state=st_VXANtcosv3cCZThfaPLz&iss={quote_plus(sso_base_url)}"
```

libparsec/crates/platform_device_loader/tests/units/list.rs
```diff
@@ -533,6 +534,10 @@ async fn testbed(env: &TestbedEnv) {
     )
     .await
     .unwrap();
+    println!(
```
touilleMan: debug print to remove

libparsec/crates/platform_device_loader/tests/units/list.rs
```diff
@@ -480,6 +480,7 @@ async fn testbed(env: &TestbedEnv) {
     })
     .await;
     let organization_addr = env.local_device("alice@dev1").organization_addr.clone();
+    println!("!!!! {}", organization_addr);
```
touilleMan: debug print to remove

libparsec/crates/types/src/local_device_file.rs
touilleMan, line 19: Maybe a serde instruction here can allow for a rename of this field into `server_addr`...

libparsec/crates/platform_device_loader/tests/units/list.rs
touilleMan: done.

server/parsec/cli/testbed.py
touilleMan: done.

libparsec/crates/platform_device_loader/tests/units/list.rs
touilleMan: done.

client/src/parsec/pki.ts
touilleMan: done.

libparsec/src/addr.rs
touilleMan: done.

bindings/generator/api/addr.py
touilleMan: done.

libparsec/crates/types/src/local_device_file.rs
touilleMan: done.

libparsec/crates/types/src/data_macros.rs
touilleMan: done.

libparsec/crates/types/src/local_device_file.rs
touilleMan: This won't work: the serde rename cannot be used for this type since its (de)serialization is only drived by the `impl_transparent_data_format_conversion!()` macro (that itself choose which field in `XXXData` match with field in `XXX`)

PR #11678, opened 2025-11-18
[MS] Mount workspace after listing

fabienSvtr:
This PR will be necessary to mount/unmount workspace properly.

### DIFF
--- a/client/src/parsec/workspace.ts
+++ b/client/src/parsec/workspace.ts
@@ -39,7 +39,7 @@ import { DateTime } from 'luxon';
 export async function initializeWorkspace(
   workspaceId: WorkspaceID,
   connectionHandle: ConnectionHandle | null = null,
-): Promise<Result<StartedWorkspaceInfo, WorkspaceInfoError | ClientStartWorkspaceError | WorkspaceMountError>> {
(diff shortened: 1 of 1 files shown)


client/src/parsec/workspace.ts
```diff
@@ -261,6 +249,10 @@ export async function mountWorkspace(
   return await libparsec.workspaceMount(workspaceHandle);
 }
 
+export async function unmountWorkspace(workspace: WorkspaceHandle): Promise<Result<null, MountpointUnmountError>> {
```
Max-7: Remove this from now, we don't need it in this PR.

client/src/views/workspaces/WorkspacesPage.vue
```diff
@@ -458,7 +458,6 @@ async function refreshWorkspacesList(): Promise<void> {
     }
     workspaceList.value = result.value;
     const recentWorkspaces = recentDocumentManager.getWorkspaces();
-    // If a workspace is listed in recent workspaces but not present in the list, remove it
```
Max-7, line 461: Why is this comment removed? It seems valid.

fabienSvtr: I put it back

PR #11680, opened 2025-11-18
feat(pki): Add new errors for `pki_accept`

### LINKED ISSUE #11676: Add additional PKI error to `pki_accept` proto
Currently not used, but will be in the future once #11648 is implemented
 
- invalid_der_x509_certificate
- invalid_payload_signature
- invalid_x509_trustchain (Cannot trust provided leaf cert and/or intermediate certs)

FirelightFlagboy:
Currently not used but will be in the future

Closes #11676

### DIFF
--- a/libparsec/crates/protocol/schema/authenticated_cmds/pki_enrollment_accept.json5
+++ b/libparsec/crates/protocol/schema/authenticated_cmds/pki_enrollment_accept.json5
@@ -66,11 +66,7 @@
             },
             {
                 // The payload is not correctly formatted
-                "status": "invalid_payload_data"
-            },
-            {
-                // The provided certificate is not valid
-                "status": "invalid_certificate"
+                "status": "invalid_payload"
             },
             {
                 // The server did not found a request for the provided ID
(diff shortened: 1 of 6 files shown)


libparsec/crates/protocol/schema/authenticated_cmds/pki_enrollment_accept.json5
```diff
@@ -123,6 +119,17 @@
                         "type": "DateTime"
                     }
                 ]
+            },
+            {
+                // A provided certificates is invalid
```
AureliaDolo, line 124: should we expect data on which certificate in invalid ?

FirelightFlagboy: WDYM by data on which cert in invalid?

By data, you mean something to identify the invalid certificate?

AureliaDolo: Yes

FirelightFlagboy: Reply from @touilleMan:

> x509 certs are something stable and consistant one against another, So if one cert in invalid, the others are likely to be invalid.
> Anyway, we can add it later but we never needed a fingerprint until now.
> In case of such error, we would need the user to provide the list of used certificates

PR #11697, opened 2025-11-19
feat(editor): Enhance error management

NicoTuxx:
- add structured error titles and messages
- adapt modal level to info / warning depending on use case
- show info modal for unsupported document types
- show warning modal for timeout loading error cases
- redirect the user to previous location in modals
- show error page instead of toast / modal for init, load, opening error cases
- delete script node when script loading fails
- add onError callback on cryptpad api

### DIFF
--- a/client/src/locales/en-US.json
+++ b/client/src/locales/en-US.json
@@ -2503,10 +2503,13 @@
                 "impossibleToOpen": "Cannot edit this file",
                 "genericError": "Failed to open the file",
                 "unsupportedFileType": "File not supported by Parsec",
-                "editionNotAvailable": "Parsec Edition unavailable"
+                "editionNotAvailable": "Document editing is unavailable",
+                "tooLongToOpen": "The file is taking too long to open"
             },
             "informationEditDownload": "If you want to edit this file, you can download it and open it locally on your device.",
             "informationPreviewDownload": "If you want to preview this file, you can download it and open it locally on your device.",
+            "tooLongToOpenOnWeb": "You can try downloading it and opening with default app (right click on file > 'Download').",
+            "tooLongToOpenOnDesktop": "You can try opening it with default app (right click on file > 'Open with default app').",
             "unknownFileExtension": "This file type extension could not be identified. Please check that it's correctly written (.txt, .pdf, .doc, etc.).",
             "noContentFileType": "Could not open this file in the Parsec editor as its type could not be identified.",
             "noFolderPreview": "Cannot preview folders.",
--- a/client/src/locales/fr-FR.json
+++ b/client/src/locales/fr-FR.json
@@ -2504,10 +2504,13 @@
                 "impossibleToOpen": "Impossible d'éditer ce fichier",
                 "genericError": "Impossible d'ouvrir ce fichier",
                 "unsupportedFileType": "Type de fichier non pris en charge par Parsec",
-                "editionNotAvailable": "Parsec Edition indisponible"
+                "editionNotAvailable": "L'édition de documents n'est pas disponible",
+                "tooLongToOpen": "Le fichier met trop de temps à s'ouvrir"
             },
             "informationEditDownload": "Si vous souhaitez modifier ce fichier, vous pouvez le télécharger et l'ouvrir localement sur votre appareil.",
             "informationPreviewDownload": "Si vous souhaitez prévisualiser ce fichier, vous pouvez le faire dans Parsec ou le télécharger pour l'ouvrir localement sur votre appareil.",
+            "tooLongToOpenOnWeb": "Vous pouvez essayer de le télécharger et de l'ouvrir avec l'application par défaut (clic droit sur le fichier > 'Télécharger').",
+            "tooLongToOpenOnDesktop": "Vous pouvez essayer de l'ouvrir avec l'application par défaut (clic droit sur le fichier > 'Ouvrir avec l'app par défaut').",
             "unknownFileExtension": "L'extension de ce fichier n'a pas pu être reconnue. Veuillez vérifier qu'elle est correctement écrite (.txt, .pdf, .doc, etc.).",
             "noContentFileType": "Impossible d'ouvrir ce fichier, son type n'ayant pas pu être reconnu dans l'éditeur de Parsec.",
             "noFolderPreview": "Impossible d'afficher l'aperçu d'un dossier.",
@@ -2600,11 +2603,13 @@
     },
     "openFallback": {
         "title": "Parsec Edition n'est pas disponible",
-        "subtitleDownload": "Si vous souhaitez modifier ce fichier, vous pouvez le télécharger et l'ouvrir localement sur votre appareil.",
-        "subtitlePreviewDownload": "Si vous souhaitez prévisualiser ce fichier, vous pouvez le faire dans Parsec ou le télécharger pour l'ouvrir localement sur votre appareil.",
+        "subtitle": {
+            "download": "Si vous souhaitez modifier ce fichier, vous pouvez le télécharger et l'ouvrir localement sur votre appareil.",
+            "previewDownload": "Si vous souhaitez prévisualiser ce fichier, vous pouvez le faire dans Parsec ou le télécharger pour l'ouvrir localement sur votre appareil."
+        },
         "actions": {
-            "viewer": "Ouvrir dans le visualiseur",
-            "download": "Télécharger le fichier",
+            "viewer": "Prévisualiser (dans Parsec)",
+            "download": "Télécharger",
             "open": "Afficher dans l'explorateur",
             "cancel": "Annuler"
         }
--- a/client/src/services/cryptpad.ts
+++ b/client/src/services/cryptpad.ts
@@ -46,6 +46,8 @@ interface CryptpadConfig {
     onSave: (file: Blob, callback: () => void) => void;
     onNewKey?: (data: { new: string; old: string }, callback: (key: string) => void) => void;
     onHasUnsavedChanges?: (unsaved: boolean) => void;
+    onError?: (error: { message: string; errorType: string; originalError: string }) => void;
+    onReady?: () => void;
   };
 }
 
@@ -55,6 +57,7 @@ export enum CryptpadErrorCode {
   InitFailed = 'init-failed',
   ScriptElementCreationFailed = 'script-element-creation-failed',
   DocumentTypeNotEnabled = 'document-type-not-enabled',
+  LoadingTimeout = 'loading-timeout',
 }
 
 export class CryptpadError extends Error {
@@ -80,7 +83,11 @@ export class CryptpadInitError extends CryptpadError {
 export class CryptpadOpenError extends CryptpadError {
   public documentType: string;
 
-  constructor(code: CryptpadErrorCode.NotInitialized | CryptpadErrorCode.DocumentTypeNotEnabled, documentType: string, message?: string) {
+  constructor(
+    code: CryptpadErrorCode.NotInitialized | CryptpadErrorCode.DocumentTypeNotEnabled | CryptpadErrorCode.LoadingTimeout,
+    documentType: string,
+    message?: string,
+  ) {
     super(code, message ?? `Failed to open document type '${documentType}' with Cryptpad: ${code}`);
     this.documentType = documentType;
   }
@@ -132,11 +139,13 @@ export class Cryptpad {
           }
         };
 
-        // Add error handling for HTTPS/security issues
+        // Add error handling for HTTPS/security, network issues
         (script as HTMLScriptElement).onerror = (error): void => {
           window.electronAPI.log('error', `Failed to load CryptPad script: ${error.toString()}`);
           window.electronAPI.log('error', 'This might be due to HTTPS requirements. Check if CryptPad server requires secure context.');
           reject(new CryptpadInitError(CryptpadErrorCode.InitFailed, error.toString()));
+          // We remove the faulty script to allow retrying
+          script?.remove();
         };
       });
     } else {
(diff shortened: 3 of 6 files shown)


client/src/services/cryptpad.ts
```diff
@@ -164,7 +180,49 @@ export class Cryptpad {
       throw new CryptpadOpenError(CryptpadErrorCode.DocumentTypeNotEnabled, config.documentType);
     }
 
-    (window as any).CryptPadAPI(this.containerElement.id, { ...config });
+    // Set up a loading timeout (30 seconds) to detect stuck/corrupted files
+    const LOADING_TIMEOUT_MS = 30000;
+    this.clearLoadingTimeout();
+
+    // Create a promise that will be resolved when CryptPad successfully loads
+    let resolveLoading: (() => void) | null = null;
```
Max-7: Minor detail but since it should be always set, you can do `let resolveLoading!: (() => void)`, to let the linter know that it is going to have a value and will never be null/undefined. Just to avoid having to check later if the variable is set or not.

client/src/services/cryptpad.ts
```diff
@@ -164,7 +180,49 @@ export class Cryptpad {
...
+
+    try {
+      // Start CryptPad API (doesn't wait for loading to complete)
+      void (window as any).CryptPadAPI(this.containerElement.id, { ...config });
+
+      // Race between successful loading (onReady callback) and timeout
+      await Promise.race([loadingComplete, timeoutPromise]);
+
+      // If we reach here, onReady was called before timeout
+      this.clearLoadingTimeout();
```
Max-7: You can move `clearLoadingTimeout` to the finally, just to avoid calling it both here and in the catch.

client/src/services/cryptpad.ts
```diff
@@ -164,7 +180,49 @@ export class Cryptpad {
...
+    } catch (error) {
+      this.clearLoadingTimeout();
+      throw error;
```
Max-7: This should be converted to a  `CryptpadOpenError`. Basically doing what you did to handle the timeout error in the file editor, but here.

client/src/views/files/handler/editor/FileEditor.vue
```diff
@@ -158,50 +168,44 @@ async function loadCryptpad(): Promise<boolean> {
...
     if (e instanceof CryptpadError) {
       switch (e.code) {
         case CryptpadErrorCode.NotEnabled:
-          title = 'fileViewers.errors.titles.editionNotAvailable';
+          await openRedirectionModal(ErrorTitle.EditionNotAvailable, ErrorMessage.EditableOnlyOnSystem, InformationLevel.Warning);
           break;
         case CryptpadErrorCode.ScriptElementCreationFailed:
         case CryptpadErrorCode.InitFailed:
-          title = 'fileViewers.errors.titles.genericError';
+          error.value = ErrorMessage.EditableOnlyOnSystem;
```
Max-7, line 184: Missing a break here (even if it's the last one, it could create a bug in the future if we add more values).

client/src/views/files/handler/editor/FileEditor.vue
```diff
@@ -279,6 +283,11 @@ async function openFileWithCryptpad(): Promise<boolean> {
           emits('onSaveStateChange', SaveState.Unsaved);
         }
       },
+      onError: (errorData: { message: string; errorType: string; originalError: string }): void => {
```
Max-7, line 299: Do we know more about those errors? Could we maintain an enum on both side to be able to display a little more details to the user? Strings are a bit hard to match, it's fine for debugging but more complicated if we want to display better error messages.

client/src/views/files/handler/editor/FileEditor.vue
```diff
@@ -287,28 +296,22 @@ async function openFileWithCryptpad(): Promise<boolean> {
...
+      await openRedirectionModal(ErrorTitle.CorruptedFile, ErrorMessage.CorruptedFile, InformationLevel.Warning);
+    } else if (e instanceof CryptpadError) {
       switch (e.code) {
         case CryptpadErrorCode.NotInitialized:
-          title = 'fileViewers.errors.titles.genericError';
+          error.value = ErrorMessage.EditableOnlyOnSystem;
           break;
         case CryptpadErrorCode.DocumentTypeNotEnabled:
-          title = 'fileViewers.errors.titles.editionNotAvailable';
+          await openRedirectionModal(ErrorTitle.EditionNotAvailable, ErrorMessage.EditableOnlyOnSystem, InformationLevel.Info);
```
Max-7, line 319: Missing a break

client/src/views/files/handler/editor/FileEditor.vue
```diff
@@ -287,28 +296,22 @@ async function openFileWithCryptpad(): Promise<boolean> {
...
-    await informationManager.present(
-      new Information({
-        title,
-        message,
-        level,
-      }),
-      PresentationMode.Modal,
-    );
+    // Redirect user back to files list after modal closes
+    await routerGoBack();
```
Max-7: `openRedirectionModal` already does a `routerGoBack`.

Max-7 (changes requested):

client/src/locales/en-US.json
```diff
@@ -2503,10 +2503,12 @@
                 "impossibleToOpen": "Cannot edit this file",
                 "genericError": "Failed to open the file",
                 "unsupportedFileType": "File not supported by Parsec",
-                "editionNotAvailable": "Parsec Edition unavailable"
+                "editionNotAvailable": "Parsec Edition unavailable",
```
```suggestion
                "editionNotAvailable": "Document editing is unavailable",
```

client/src/locales/fr-FR.json
```diff
@@ -2504,10 +2504,12 @@
                 "impossibleToOpen": "Impossible d'éditer ce fichier",
                 "genericError": "Impossible d'ouvrir ce fichier",
                 "unsupportedFileType": "Type de fichier non pris en charge par Parsec",
-                "editionNotAvailable": "Parsec Edition indisponible"
+                "editionNotAvailable": "Parsec Edition indisponible",
```
```suggestion
                "editionNotAvailable": "L'édition de documents n'est pas disponible",
```

client/src/locales/en-US.json
```diff
@@ -2503,10 +2503,12 @@
...
+                "corruptedFile": "File is corrupted or too large"
             },
             "informationEditDownload": "If you want to edit this file, you can download it and open it locally on your device.",
             "informationPreviewDownload": "If you want to preview this file, you can download it and open it locally on your device.",
+            "corruptedFileMessage": "This file appears to be corrupted or is too large to load. Please try downloading it or opening a different file.",
```
mmmarcos: Suggestion to download the file should only be displayed on web. On desktop, we should suggest to "Open with default app". Should we have two message here (for web, for desktop)?

Also, if the user wants to open this file, suggesting to open another won't help.

On web:
```suggestion
            "corruptedFileMessage": "This file appears to be corrupted or is too large to open. You can try downloading it and opening with default app (right click on file > 'Download').",
```
On desktop:
```suggestion
            "corruptedFileMessage": "This file appears to be corrupted or is too large to open. You can try opening it with default app (right click on file > 'Open with default app').",
```

client/src/locales/fr-FR.json
```diff
@@ -2504,10 +2504,12 @@
...
+                "corruptedFile": "Fichier corrompu ou trop volumineux"
             },
             "informationEditDownload": "Si vous souhaitez modifier ce fichier, vous pouvez le télécharger et l'ouvrir localement sur votre appareil.",
             "informationPreviewDownload": "Si vous souhaitez prévisualiser ce fichier, vous pouvez le faire dans Parsec ou le télécharger pour l'ouvrir localement sur votre appareil.",
+            "corruptedFileMessage": "Ce fichier semble être corrompu ou est trop volumineux pour être chargé. Veuillez essayer de le télécharger ou d'ouvrir un autre fichier.",
```
mmmarcos: See comment in `en-US.json`

client/src/services/cryptpad.ts
```diff
@@ -164,7 +180,49 @@ export class Cryptpad {
...
+    // Create a promise that will be resolved when CryptPad successfully loads
+    let resolveLoading: (() => void) | null = null;
+    const loadingComplete = new Promise<void>((resolve) => {
+      resolveLoading = resolve;
+    });
+
+    const timeoutPromise = new Promise<void>((_, reject) => {
+      this.loadingTimeoutId = window.setTimeout(() => {
+        window.electronAPI.log('error', 'CryptPad loading timeout - file may be corrupted or too large');
+        reject(new Error('CryptPad loading timeout - the file may be corrupted, too large, or the server is not responding'));
```
mmmarcos: IMO a loading timeout shouldn't be an error.

Also, log and error messages are not consistent, shouldn't it be the same?
```suggestion
        const message = 'CryptPad loading timeout. The file may be corrupted or too large';
        window.electronAPI.log('error', message);
        reject(new Error(message));
```
```suggestion
        const message = 'CryptPad loading timeout. The file may be corrupted, too large, or the server is not responding';
        window.electronAPI.log('error', message);
        reject(new Error(message));
```

client/src/views/files/handler/editor/FileEditor.vue
```diff
@@ -287,28 +296,22 @@ async function openFileWithCryptpad(): Promise<boolean> {
...
     return true;
   } catch (e: unknown) {
-    let title: Translatable = 'fileViewers.errors.titles.genericError';
-    const message: Translatable = 'fileViewers.errors.informationEditDownload';
-    const level: InformationLevel = InformationLevel.Info;
-
-    if (e instanceof CryptpadError) {
+    // Check if this is a timeout error (corrupted file)
+    if (e instanceof Error && e.message.includes('timeout')) {
+      window.electronAPI.log('error', 'CryptPad loading timeout - file appears to be corrupted or too large');
```
mmmarcos: Timeout should not be an error. 

Also, the log message is already displayed on timeout, isn't it? (see my previous comment)

client/src/views/files/handler/editor/FileEditor.vue
NicoTuxx: At the moment I'm not totally certain about all errors that can fall in here. It will be addressed when contacting xwiki, I keep it in mind for future improvement.

client/src/services/cryptpad.ts
```diff
@@ -164,7 +185,45 @@ export class Cryptpad {
...
+        const message = 'CryptPad loading timeout - the file may be corrupted, too large, or the server is not responding.';
+        window.electronAPI.log('warn', message);
+        reject(new CryptpadOpenError(CryptpadErrorCode.LoadingTimeout, config.documentType, message));
+      }, LOADING_TIMEOUT_MS);
+    });
+
+    // Wrap the original onReady callback to signal successful loading
+    const originalOnReady = config.events.onReady;
+    config.events.onReady = () => {
+      if (resolveLoading) resolveLoading();
```
Max-7: Is that check still needed? `resolveLoading` should never be undefined.

client/src/views/files/handler/editor/FileEditor.vue
```diff
@@ -287,28 +298,25 @@ async function openFileWithCryptpad(): Promise<boolean> {
     window.electronAPI.log('info', 'CryptPad editor initialized successfully');
     return true;
   } catch (e: unknown) {
-    let title: Translatable = 'fileViewers.errors.titles.genericError';
-    const message: Translatable = 'fileViewers.errors.informationEditDownload';
-    const level: InformationLevel = InformationLevel.Info;
-
-    if (e instanceof CryptpadError) {
+    // Check if this is a timeout error (corrupted file)
+    if (e instanceof Error && e.message.includes('timeout')) {
```
Max-7: Timeout is converted to a CryptpadError

client/src/views/files/handler/editor/FileEditor.vue
```diff
@@ -287,28 +298,25 @@ async function openFileWithCryptpad(): Promise<boolean> {
     window.electronAPI.log('info', 'CryptPad editor initialized successfully');
     return true;
   } catch (e: unknown) {
-    let title: Translatable = 'fileViewers.errors.titles.genericError';
-    const message: Translatable = 'fileViewers.errors.informationEditDownload';
-    const level: InformationLevel = InformationLevel.Info;
-
-    if (e instanceof CryptpadError) {
+    // Check if this is a timeout error (corrupted file)
+    if (e instanceof CryptpadError && e.code === CryptpadErrorCode.LoadingTimeout) {
```
Max-7: Why not put that in the `else if` just below? It checks if the error is a CryptpadError, and it does a switch on the code, so it's just adding a case to the switch.

NicoTuxx: fixed

client/src/views/files/handler/editor/FileEditor.vue
```diff
@@ -77,6 +77,7 @@ import { IonButton, IonIcon, IonItem, IonList, IonText } from '@ionic/vue';
 import { checkmarkCircle } from 'ionicons/icons';
 import { I18n, Translatable } from 'megashark-lib';
 import { inject, onMounted, onUnmounted, ref, useTemplateRef } from 'vue';
+import { FileHandlerMode } from '../types';
```
mmmarcos: absolute path?
```suggestion
import { FileHandlerMode } from '@/views/files/handler/types';
```

client/src/services/cryptpad.ts
```diff
@@ -164,7 +173,45 @@ export class Cryptpad {
       throw new CryptpadOpenError(CryptpadErrorCode.DocumentTypeNotEnabled, config.documentType);
     }
 
-    (window as any).CryptPadAPI(this.containerElement.id, { ...config });
+    // Set up a loading timeout (30 seconds) to detect stuck/corrupted files
+    const LOADING_TIMEOUT_MS = 30000;
```
mmmarcos, line 177: I know it's arbitrary but 30 seconds seems too long without a spinner or animated progress bar. Maybe 15s?

client/src/views/files/handler/editor/FileEditor.vue
```diff
@@ -158,50 +169,51 @@ async function loadCryptpad(): Promise<boolean> {
...
+  await informationManager.present(
+    new Information({
+      title,
+      message,
+      level,
+    }),
+    PresentationMode.Modal,
+  );
+  await routerGoBack();
+  // TODO: find why we have router problems causing routerGoBack to stay on same page
```
mmmarcos, line 202: If there is an existing issue, I would add it here
```suggestion
  // TODO: investigate why we have router problems causing routerGoBack to stay on same page
```

client/src/services/cryptpad.ts
NicoTuxx: You want a modal to pop each second with a "wait" button? I was already not really convinced by letting the user with a solution that can continue forever, but each 15 sec seems a lot like a spam to me :/

NicoTuxx: we keep 30 for now

PR #11705, opened 2025-11-20
[MS] Show/hide workspace feature

### LINKED ISSUE #6969: "Hide" workspace for unmounting
Image

## Context
The user will need to have 2 functions for displaying the workspace which should be clear for him.
- 1️⃣: workspaces can be **activated or not** (via the mountpoint). To do this, the idea will be to show the user that it is possible to lock a workspace. 
- 2️⃣: workspaces can be **hidden** to reduce the number of visible workspaces (still active on the mountpoint).

## Userflow
- The user can hides a workspace by clicking on the contextual menu and `Hide this workspace`
- Then a modal will be shown to warn the user the workspace will be unmounted and hidden on the app (only the "Show hidden workspaces" is not checked)
- The user can check "Don’t show this again"
- The app shows a toast the confirm the action

fabienSvtr:
## Describe what's changed
- The user can open workspace context menu to "Show" or "Hide" a workspace.
  - **Parsec Web**: Hide a workspace has no consequences except hiding the workspace from the workspaces page
  - **Parsec Desktop**: Hide option will unmount the workspace

- If the user want to preview a file (from a mounted workspace) or an mounted workspace in his explorer, a modal appeared to ask if he wants to mount the workspace in order to view the file/workspace. This modal appeared in two cases:
  - Open file with default app
  - Show file/workspace in explorer

If the user doesn't want to mount the workspace, an error toast is shown.
- The user can toggle a checkbox to show the hidden workspaces. 

closes #6969


client/src/views/workspaces/WorkspacesPage.vue
Ironicbay, line 1: Line 137: if all of the user's workspaces are hidden and the toggle is off, the text shown is misleading: `No workspace match this search filter`.

I would make a special case if the user has access to workspaces that are all hidden.

client/src/views/workspaces/WorkspacesPage.vue
```diff
@@ -49,6 +49,21 @@
...
+          class="hidden-workspaces-checkbox"
+          id="show-hidden-workspaces-large"
+        >
+          <ion-text class="button-medium">
+            <span v-if="windowWidth > WindowSizeBreakpoints.LG">
+              {{ $msTranslate('WorkspacesPage.categoriesMenu.showHiddenWorkspaces') }}
+            </span>
+            <span v-else>{{ $msTranslate('WorkspacesPage.categoriesMenu.showHiddenWorkspacesShort') }}</span>
+          </ion-text>
+        </ms-checkbox>
```
Ironicbay: I would change the checkbox and `hiddenWorkspaces` variable behaviours:
- Show the checkbox only if there is at least one hidden workspace, otherwise it takes up space for nothing
- The `hiddenWorkspaces` state should be remembered somehow, maybe through the storage manager, currently the state resets to `false` as soon as we change page, it's jarring especially if we visit a hidden workspace and go back to the workspaces page
- Use `WorkspacesPage.categoriesMenu.showHiddenWorkspaces` text at all times in large display, and keep `WorkspacesPage.categoriesMenu.showHiddenWorkspacesShort` checkbox text only in the smallest size in small display

client/tests/e2e/specs/workspaces_list.spec.ts
```diff
@@ -255,6 +255,97 @@ for (const displaySize of [DisplaySize.Small, DisplaySize.Large]) {
     await favoriteWorkspacesButton.click({ force: true });
     await verifyActiveCategory(workspaces, 2);
   });
+
+  msTest(`Show/hide workspace ${displaySize} display`, async ({ workspaces }) => {
```
Ironicbay, line 284: You can add another test in which you create a new workspace (with the existing `createWorkspace` test function) to test the display with two workspaces, hiding one after the other and checking that both are displayed or not accordingly.

newsfragments/6969.feature.rst
```diff
@@ -0,0 +1 @@
+User can show or hide a workspace from the list of workspaces. A hidden workspace will be not visible from the explorer (on desktop app).
```
Ironicbay:
```suggestion
User can show or hide a workspace from the list of workspaces. A hidden workspace will be not visible from the explorer (on desktop app) and can be toggled to be visible or not in Parsec.
```

newsfragments/11865.feature.rst
Ironicbay, line 1: You can remove this newsfragment since you already have another

client/src/components/workspaces/utils.ts
```diff
@@ -184,22 +194,150 @@ export async function openWorkspaceContextMenu(
...
+          key: 'WorkspacesPage.showHideWorkspace.failedShown',
+          data: { workspace: workspace.currentName },
+        },
+        level: InformationLevel.Error,
+      }),
+      PresentationMode.Toast,
+    );
+  }
+
+  return ok;
```
Ironicbay: It's preferable to return early if something goes wrong and to avoid to have huge blocks of code with 1 or 2 extra indentations because of an early condition. Here the `ok` variable is a bit useless since we only care what the `mountWorkspace()` function returned if we're on desktop.

Also, the return value is never processed in your code so you can ditch the `Promise<boolean>` typing for `Promise<void>`

Your function could be structured like so (pseudo-code):

```typescript
export async function(...) {
  if (isDesktop()) {
    // Try to mount
    const result = await mountWorkspace();
    if (!result.ok) {
      // Handle error there
      informationManager.present( ... );
      // Log the error
      window.electronAPI.log('error', `Error while mounting workspace: ${result.error}`);
      return;
    }
  }
  
  // continue as usual, the function will begin here in web and will proceed as expected in desktop since we know the mount was successful
  workspaceAttributes.removeHidden(workspace.id);

  ...
}
```

client/src/components/workspaces/utils.ts
```diff
@@ -184,22 +194,150 @@ export async function openWorkspaceContextMenu(
...
+      new Information({
+        message: {
+          key: 'WorkspacesPage.showHideWorkspace.failedHidden',
+          data: { workspace: workspace.currentName },
+        },
+        level: InformationLevel.Error,
+      }),
+      PresentationMode.Toast,
+    );
+  }
```
Ironicbay: Same remark for this function. Try to unmount if desktop, if it failed, show toast and return early. Be mindful about logging the error if the unmount fails as I suggested for the mount function.

client/src/locales/en-US.json
```diff
@@ -703,10 +705,51 @@
...
+            "successWebHidden": "The workspace is now hidden in Parsec.",
+            "failedHidden": "Failed to hide the workspace. Please try again.",
+            "successDesktopShown": "The workspace is now visible in Parsec and your explorer.",
+            "successWebShown": "The workspace is now visible in Parsec.",
+            "failedShown": "Failed to show the workspace. Please try again."
+        },
+        "workspaceHiddenModal": {
+            "title": "Hide a workspace",
+            "affectedWorkspaces": "You are about to hide the workspace {workspace} and it will no longer be visible from the file explorer.",
+            "info": "You can check “Show hidden workspaces” to show it on Parsec.",
```
```suggestion
            "info": "You can check “Show hidden workspaces” to see it in Parsec.",
```

client/src/locales/fr-FR.json
```diff
@@ -702,10 +704,51 @@
...
+            "successWebHidden": "L'espace de travail est maintenant masqué dans Parsec.",
+            "failedHidden": "Impossible de masquer l'espace de travail. Veuillez réessayer.",
+            "successDesktopShown": "L'espace de travail est maintenant visible dans Parsec et votre explorateur.",
+            "successWebShown": "L'espace de travail est maintenant visible dans Parsec.",
+            "failedShown": "Impossible d'afficher l'espace de travail. Veuillez réessayer."
+        },
+        "workspaceHiddenModal": {
+            "title": "Masquer un espace de travail",
+            "affectedWorkspaces": "Vous êtes sur le point de masquer l'espace de travail {workspace}. Cet espace de travail ne sera plus visible depuis l'explorateur de fichiers.",
+            "info": "Vous pouvez cocher “Afficher les espaces masqués” pour le voir sur Parsec.",
```
```suggestion
            "info": "Vous pouvez cocher “Afficher les espaces masqués” pour le voir dans Parsec.",
```

client/src/services/storageManager.ts
```diff
@@ -197,6 +200,7 @@ export class StorageManager {
       skipLongPathsSupportWarning: data.skipLongPathsSupportWarning ?? StorageManager.DEFAULT_CONFIG.skipLongPathsSupportWarning,
       disableDownloadWarning: data.disableDownloadWarning ?? StorageManager.DEFAULT_CONFIG.disableDownloadWarning,
       hideParsecDownload: data.hideParsecDownload ?? StorageManager.DEFAULT_CONFIG.hideParsecDownload,
+      skipWorkspaceHiddenWarning: data.skipWorkspaceHiddenWarning ?? StorageManager.DEFAULT_CONFIG.skipWorkspaceHiddenWarning,
```
Ironicbay, line 203: It still feels weird to me that this is a one-time warning. I feel like there needs to be somewhere else we should remind the user that hiding a workspace makes the files inaccessible from the explorer. But I guess that's kinda explained when trying to click `Show in explorer` on a hidden workspace.

client/src/views/files/FoldersPage.vue
```diff
@@ -1403,11 +1406,31 @@ async function openEntry(entryToOpen: EntryModel, options: OpenPathOptions): Pro
...
+      return;
+    }
+
+    result = await showWorkspace(workspaceInfo.value, workspaceAttributes, informationManager, eventDistributor);
+  }
+
+  if (result) {
+    await pathOpener.openPath(workspaceHandle, entry.path, informationManager, options);
+    selectionEnabled.value = false;
+  }
```
Ironicbay, line 1337: You're adding a condition to the function, what was there before should not change.

Simply adapt your new code to return if the conditions aren't met:

```typescript
  // Your added code
  if (workspaceAttributes.isHidden(workspaceInfo.value.id)) {
    const answer = await askQuestion(
      'WorkspacesPage.openInExplorerModal.file.title',
      'WorkspacesPage.openInExplorerModal.file.description',
      {
        yesText: 'WorkspacesPage.openInExplorerModal.actionConfirm',
        noText: 'WorkspacesPage.openInExplorerModal.actionCancel',
      },
    );
    if (answer === Answer.No) {
      return;
    }
    const result = await showWorkspace(workspaceInfo.value, workspaceAttributes, informationManager, eventDistributor);
    if (!result) {
      return;
    }
  }

  // Previous code left as it was
  await pathOpener.openPath(workspaceHandle, entry.path, informationManager, options);
  selectionEnabled.value = false;
```

client/src/views/files/FoldersPage.vue
```diff
@@ -1528,13 +1551,28 @@ async function shareEntries(): Promise<void> {
...
+      },
+    );
+
+    if (answer === Answer.No) {
+      return;
+    }
+
+    result = await showWorkspace(workspaceInfo.value, workspaceAttributes, informationManager, eventDistributor);
+  }
+  if (result) {
```
Ironicbay: Same here.
```typescript
    ...
    const result = await showWorkspace(workspaceInfo.value, workspaceAttributes, informationManager, eventDistributor);
    if (!result) {
      return;
    }
  }
  await pathOpener.showInExplorer(workspaceInfo.value.handle, entries[0].path, informationManager);
```

client/src/services/storageManager.ts
fabienSvtr: It's a choice, if the user does the action multiple times, it's quite disturbing but since we don't have any feedback yet, we can decide which option to choose (add the checkbox or not).

client/src/locales/en-US.json
```diff
@@ -635,6 +635,7 @@
         "noWorkspacesExternal": "You do not have access to any workspaces. Workspaces that have been shared with you will be listed here.",
         "loading": "Loading your workspaces...",
         "noMatchingWorkspaces": "No workspace match this search filter.",
+        "allWorkspacesHidden": "All your workspaces are hidden. Deselect the \"Hidden workspaces\" filter to show them.",
```
```suggestion
        "allWorkspacesHidden": "All your workspaces are hidden. Toggle the \"Hidden workspaces\" filter to show them.",
```

client/src/locales/fr-FR.json
```diff
@@ -634,6 +634,7 @@
         "noWorkspacesExternal": "Vous n'avez pas encore accès à un espace de travail. Les espaces de travail qui vous seront partagés s'afficheront ici.",
         "loading": "Chargement de vos espaces de travail...",
         "noMatchingWorkspaces": "Aucun espace de travail ne correspond à cette recherche.",
+        "allWorkspacesHidden": "Tous vos espaces de travail sont masqués. Désélectionnez le filtre \"Espaces masqués\" pour les afficher.",
```
```suggestion
        "allWorkspacesHidden": "Tous vos espaces de travail sont masqués. Désactivez le filtre \"Espaces masqués\" pour les afficher.",
```

client/src/components/workspaces/utils.ts
```diff
@@ -184,22 +195,151 @@ export async function openWorkspaceContextMenu(
...
+
+  return ok;
+}
```
Ironicbay, line 270: Adapt this function as you did for `hideWorkspace()`: first check for the conditions specific to desktop, return early if necessary, continue as expected otherwise

```typescript
export async function showWorkspace(
  workspace: WorkspaceInfo | StartedWorkspaceInfo,
  workspaceAttributes: WorkspaceAttributes,
  eventDistributor: EventDistributor,
): Promise<boolean> {
  if (isDesktop()) {
    const result = await mountWorkspace(workspace.handle);
    if (!result.ok) {
      informationManager.present(
        new Information({
          message: {
            key: 'WorkspacesPage.showHideWorkspace.failedShown',
            data: { workspace: workspace.currentName },
          },
          level: InformationLevel.Error,
        }),
        PresentationMode.Toast,
      );
      window.electronAPI.log('error', `Error while mounting workspace: ${result.error}`);
      return false;
    }
  }

  workspaceAttributes.removeHidden(workspace.id);
  await workspaceAttributes.save();

  informationManager.present(
    new Information({
      message: {
        key: isDesktop() ? 'WorkspacesPage.showHideWorkspace.successDesktopShown' : 'WorkspacesPage.showHideWorkspace.successWebShown',
        data: { workspace: workspace.currentName },
      },
      level: InformationLevel.Success,
    }),
    PresentationMode.Toast,
  );
  await eventDistributor.dispatchEvent(Events.WorkspaceMountpointsSync, {
    workspaceId: workspace.id,
    isMounted: true,
  });

  return true;
}
```

Ironicbay (commented):
I like the approach to use the workspace filter to show or hide hidden workspaces.

What I'd change:
- Remove entirely the `Show hidden workspaces` checkmark on large display to lighten the page
- Have the checkbox OFF by default (which is the case at the moment), but with the filter button being active: this is where I'm not sure of this myself, but if a user clicks on `Hide workspace`, they should expect it to disappear, so it should be filtered by default. However, it would be jarring to have the filter button shown as active even if the user has no hidden workspaces. So what I would do:
  - Have the checkbox OFF by default (current behaviour)
  - Revert the check in `WorkspaceFilter.vue::missingFilters` to show the filter active when the hidden workspace checkbox is OFF
  - Add a `noHiddenWorkspaces: boolean` (or similar) prop in `WorkspaceFilter.vue` to tell from the `WorkspacesPage.vue` if the user has one or more workspaces hidden. If there is no hidden workspace, ignore this filter in the `missingFilters` const. If there are hidden workspaces, process the hidden filter as usual.

```typescript
// WorkspaceFilter.vue

const props = defineProps<{
  filters: WorkspacesPageFilters;
  noHiddenWorkspaces: boolean;
}>();

...

const missingFilters = computed(() => {
  if (props.noHiddenWorkspaces {
    return !props.filters.owner || !props.filters.manager || !props.filters.contributor || !props.filters.reader;
  }
  return !props.filters.owner || !props.filters.manager || !props.filters.contributor || !props.filters.reader || !props.filters.hidden;
});

File: client/src/components/users/UserFilterPopover.vue [hunk 14]
```diff
@@ -8,12 +8,23 @@
...
-          class="body-sm list-group-title"
-          id="filter-title-status"
-        >
-          {{ $msTranslate('UsersPage.filter.status') }}
-        </ion-text>
+        <div class="list-group-header">
+          <ion-text class="body-sm list-group-title">
+            {{ $msTranslate('UsersPage.filter.status') }}
+          </ion-text>
+          <ion-button
```
Ironicbay, line 15: Missing `IonButton` import

client/src/locales/en-US.json
```diff
@@ -699,14 +702,58 @@
         "linkNotCopiedToClipboard": "Failed to copy the link. Your browser or device does not seem to support copy/paste.",
         "filter": {
             "title": "Filter",
-            "roles": "Role"
+            "roles": "Role",
+            "hidden": "Workspace visibility",
+            "visibleOnly": "Visibles only",
```
```suggestion
            "visibleOnly": "Visible only",
```

client/src/locales/en-US.json
```diff
@@ -699,14 +702,60 @@
...
```
Ironicbay: You can remove this locale

client/src/locales/en-US.json
```diff
@@ -699,14 +702,60 @@
...
             "all": "All workspaces",
+            "allShort": "All",
             "recents": "Recently viewed",
-            "favorites": "Pinned",
+            "recentsShort": "Recent",
+            "favorites": "Favorites",
+            "hidden": "Hidden",
             "noRecentWorkspaces": "You have not consulted any workspaces yet. Recently consulted workspaces will be listed here.",
-            "noFavoriteWorkspaces": "You have not pinned any workspaces yet. Pin a workspace to have it be listed here."
+            "noFavoriteWorkspaces": "You have not favorited any workspaces yet. Favorite a workspace to have it be listed here.",
```
```suggestion
            "noFavoriteWorkspaces": "You have not set any workspaces as favorite. Favorite a workspace to have it be listed here.",
```

client/src/views/workspaces/WorkspacesPage.vue
```diff
@@ -24,6 +24,7 @@
           />
           <workspace-filter
             :filters="workspaceFilters"
+            :no-hidden-workspaces="workspaceAttributes.getHidden().value.length === 0"
```
Ironicbay: You can remove this line

client/src/views/workspaces/WorkspacesPage.vue
```diff
@@ -60,6 +61,7 @@
           <div class="mobile-filters-buttons">
             <workspace-filter
               :filters="workspaceFilters"
+              :no-hidden-workspaces="workspaceAttributes.getHidden().value.length === 0"
```
Ironicbay: Same

client/src/views/workspaces/WorkspacesPage.vue
```diff
@@ -499,22 +531,34 @@ async function refreshWorkspacesList(): Promise<void> {
...
 
+const showHiddenWorkspacesButton = computed(() => {
+  return (
+    workspaceAttributes.getHidden().value.length === workspaceList.value.length &&
+    workspaceList.value.length > 0 &&
+    filteredWorkspaces.value.length === 0
+  );
+});
+
 const filteredWorkspaces = computed(() => {
```
Ironicbay, line 540: Add a condition to filter hidden workspaces when not on the `Hidden` tab. Either add
```typescript
|| (workspaceAttributes.isHidden(workspace.id) && workspaceMenuState.value !== WorkspaceMenu.Hidden)
```
to the first `if` to filter it out right away, or add `&& !isHidden(workspace.id)` to every other case in the `switch/case` below.

If you choose the first, you can then remove the case for `WorkspaceMenu.All` since the hidden workspaces would be filtered before.

client/src/views/workspaces/WorkspacesPage.vue
```diff
@@ -468,7 +498,7 @@ async function refreshWorkspacesList(): Promise<void> {
       } else {
         window.electronAPI.log('warn', `Failed to get sharing for ${wk.currentName}`);
       }
-      if (isDesktop() && wk.mountpoints.length === 0) {
+      if (isDesktop() && wk.mountpoints.length === 0 && !hidden.value.includes(wk.id)) {
```
Ironicbay:
```suggestion
      if (isDesktop() && wk.mountpoints.length === 0 && !workspaceAttributes.isHidden(wk.id)) {
```

Ironicbay: And you can delete the `hidden` const line 491

client/src/components/users/UserFilterPopover.vue
Ironicbay, line 15: Still missing the `IonButton` import

PR #11707, opened 2025-11-20
[MS] Add authentication change to/from openbao

Ironicbay:


client/src/views/profile/AuthenticationPage.vue
```diff
@@ -57,8 +72,47 @@ import { inject, onMounted, Ref, ref } from 'vue';
 const currentDevice: Ref<AvailableDevice | null> = ref(null);
 const informationManager: InformationManager = inject(InformationManagerKey)!;
 const error = ref('');
+const serverConfig = ref<ServerConfig | undefined>(undefined);
+
+function multipleAuthAvailable(): boolean {
```
Max-7: Honestly, do we really care about displaying `Change password` instead of `Change authentication`? I understand the goal but its adds a lot of complexity (in the code, in the e2e tests, in the manual tests) for something insignificant.

client/src/views/profile/AuthenticationPage.vue
```diff
@@ -57,8 +72,47 @@ import { inject, onMounted, Ref, ref } from 'vue';
...
+  if (!isWeb()) {
+    return true;
+  }
+
+  return (serverConfig.value?.openbao && serverConfig.value?.openbao.auths.some((auth) => isSSOProviderHandled(auth.tag))) === true;
+}
+
+async function getOpenBaoClient(): Promise<OpenBaoClient | void> {
```
Max-7: Not a valid return type.

client/src/views/profile/AuthenticationPage.vue
```diff
@@ -68,6 +122,12 @@ async function openChangeAuthentication(): Promise<void> {
     if (answer === Answer.No) {
       return;
     }
+  } else if (currentDevice.value && currentDevice.value.ty.tag === AvailableDeviceTypeTag.OpenBao) {
+    const connResult = await getOpenBaoClient();
+    if (!connResult) {
```
Max-7: Missing some error handling here. What does the user see if we can't get the open bao client?

client/src/views/users/UpdateAuthenticationModal.vue
```diff
@@ -186,6 +194,26 @@ async function changeAuthentication(): Promise<void> {
       tag: DeviceAccessStrategyTag.Smartcard,
       keyFile: props.currentDevice.keyFilePath,
     };
+  } else if (props.currentDevice.ty.tag === AvailableDeviceTypeTag.OpenBao) {
+    if (!props.openBaoConnInfo) {
```
Max-7: This should be done earlier, before opening the modal. Still have to check here, but just a log here is fine.

client/src/views/users/UpdateAuthenticationModal.vue
```diff
@@ -127,28 +143,51 @@ const errorMessage: Ref<Translatable> = ref('');
...
 
 async function nextStep(): Promise<void> {
-  if (pageStep.value === ChangeAuthenticationStep.InputCurrentPassword) {
-    const access: DeviceAccessStrategyPassword = {
-      tag: DeviceAccessStrategyTag.Password,
-      password: currentPassword.value,
-      keyFile: props.currentDevice.keyFilePath,
-    };
+  if (pageStep.value === ChangeAuthenticationStep.CurrentAuthentication) {
+    let access!: DeviceAccessStrategyPassword | DeviceAccessStrategyOpenBao;
```
NicoTuxx, line 177: Maybe a switch case would be interesting here, if we plan to have multiple choices here.

PR #11711, opened 2025-11-20
[MS] Show modal hidden workspace

### LINKED ISSUE #11870: Show a modal when user hides a workspace
When a user wants to hide a workspace, a modal appears to inform the workspace will not be show in his files explorer 
Image

fabienSvtr:
### Describe the feature
If the user want to preview a file (from a mounted workspace) or an mounted workspace in his explorer, a modal appeared to ask if he wants to mount the workspace in order to view the file/worksapce. This modal appeared in two cases:
- Open file with default app
- Show file/workspace in explorer

If the user doesn't want to mount the workspace, an error toast is shown.

https://github.com/user-attachments/assets/64d45615-cc58-4eea-a6a2-c04625142312

closes #11870

### DIFF
--- a/client/src/components/workspaces/utils.ts
+++ b/client/src/components/workspaces/utils.ts
@@ -3,6 +3,7 @@
 import { workspaceNameValidator } from '@/common/validators';
 import {
   ClientRenameWorkspaceErrorTag,
+  StartedWorkspaceInfo,
   UserProfile,
   WorkspaceID,
   WorkspaceInfo,
@@ -10,6 +11,7 @@ import {
   WorkspaceRole,
   getClientProfile,
   getSystemPath,
+  getWorkspaceInfo,
   isDesktop,
   mountWorkspace,
   getPathLink as parsecGetPathLink,
@@ -17,16 +19,18 @@ import {
   unmountWorkspace,
 } from '@/parsec';
 import { Routes, navigateTo } from '@/router';
-import { EventDistributor } from '@/services/eventDistributor';
+import { EventDistributor, Events } from '@/services/eventDistributor';
 import { Information, InformationLevel, InformationManager, PresentationMode } from '@/services/informationManager';
 import { recentDocumentManager } from '@/services/recentDocuments';
+import { StorageManager } from '@/services/storageManager';
 import { WorkspaceAttributes } from '@/services/workspaceAttributes';
 import SmallDisplayWorkspaceContextMenu from '@/views/workspaces/SmallDisplayWorkspaceContextMenu.vue';
 import { WorkspaceAction } from '@/views/workspaces/types';
 import WorkspaceContextMenu from '@/views/workspaces/WorkspaceContextMenu.vue';
+import WorkspaceHiddenModal from '@/views/workspaces/WorkspaceHiddenModal.vue';
 import WorkspaceSharingModal from '@/views/workspaces/WorkspaceSharingModal.vue';
 import { modalController, popoverController } from '@ionic/vue';
-import { Clipboard, DisplayState, Translatable, getTextFromUser } from 'megashark-lib';
+import { Answer, Clipboard, DisplayState, MsModalResult, Translatable, askQuestion, getTextFromUser } from 'megashark-lib';
 
 export const WORKSPACES_PAGE_DATA_KEY = 'WorkspacesPage';
 
@@ -121,6 +125,7 @@ export async function openWorkspaceContextMenu(
   workspaceAttributes: WorkspaceAttributes,
   eventDistributor: EventDistributor,
   informationManager: InformationManager,
+  storageManager: StorageManager,
   fromSidebar = false,
   isLargeDisplay = true,
 ): Promise<void> {
@@ -177,7 +182,7 @@ export async function openWorkspaceContextMenu(
         await copyLinkToClipboard(workspace, informationManager);
         break;
       case WorkspaceAction.OpenInExplorer:
-        await openWorkspace(workspace, informationManager);
+        await seeInExplorer(workspace, informationManager, workspaceAttributes, eventDistributor);
         break;
       case WorkspaceAction.Rename:
         await openRenameWorkspaceModal(workspace, informationManager, isLargeDisplay);
@@ -190,10 +195,23 @@ export async function openWorkspaceContextMenu(
         await navigateTo(Routes.History, { query: { documentPath: '/', workspaceHandle: workspace.handle } });
         break;
       case WorkspaceAction.Mount:
-        await showWorkspace(workspace, workspaceAttributes, informationManager);
+        await showWorkspace(workspace, workspaceAttributes, informationManager, eventDistributor);
         break;
       case WorkspaceAction.UnMount:
-        await hideWorkspace(workspace, workspaceAttributes, informationManager);
+        if (isDesktop()) {
+          const refreshWorkspaces = await getWorkspaceInfo(workspace.handle);
+          if (refreshWorkspaces.ok) {
+            await unmountWorkspaceConfirmation(
+              workspaceAttributes,
+              refreshWorkspaces.value,
+              informationManager,
+              eventDistributor,
+              storageManager,
+            );
+          }
+        } else {
+          await hideWorkspace(workspace, workspaceAttributes, informationManager, eventDistributor);
+        }
         break;
       default:
         console.warn('No WorkspaceAction match found');
@@ -202,10 +220,11 @@ export async function openWorkspaceContextMenu(
 }
 
 export async function showWorkspace(
-  workspace: WorkspaceInfo,
+  workspace: WorkspaceInfo | StartedWorkspaceInfo,
   workspaceAttributes: WorkspaceAttributes,
   informationManager: InformationManager,
-): Promise<void> {
+  eventDistributor: EventDistributor,
+): Promise<boolean> {
   let ok = true;
   if (isDesktop()) {
     const result = await mountWorkspace(workspace.handle);
@@ -214,6 +233,8 @@ export async function showWorkspace(
 
   if (ok) {
     workspaceAttributes.removeHidden(workspace.id);
+    await workspaceAttributes.save();
+
     informationManager.present(
       new Information({
         message: {
@@ -224,6 +245,10 @@ export async function showWorkspace(
       }),
       PresentationMode.Toast,
     );
+    await eventDistributor.dispatchEvent(Events.WorkspaceMountpointsSync, {
+      workspaceId: workspace.id,
+      isMounted: true,
+    });
   } else {
     informationManager.present(
       new Information({
@@ -236,12 +261,15 @@ export async function showWorkspace(
       PresentationMode.Toast,
     );
   }
+
+  return ok;
 }
 
 export async function hideWorkspace(
-  workspace: WorkspaceInfo,
+  workspace: WorkspaceInfo | StartedWorkspaceInfo,
   workspaceAttributes: WorkspaceAttributes,
   informationManager: InformationManager,
+  eventDistributor: EventDistributor,
 ): Promise<void> {
   let ok = true;
   if (isDesktop()) {
@@ -251,6 +279,8 @@ export async function hideWorkspace(
 
   if (ok) {
     workspaceAttributes.addHidden(workspace.id);
+    await workspaceAttributes.save();
+
     informationManager.present(
       new Information({
         message: {
@@ -261,6 +291,10 @@ export async function hideWorkspace(
       }),
       PresentationMode.Toast,
     );
+    await eventDistributor.dispatchEvent(Events.WorkspaceMountpointsSync, {
+      workspaceId: workspace.id,
+      isMounted: false,
+    });
   } else {
     informationManager.present(
       new Information({
@@ -275,16 +309,35 @@ export async function hideWorkspace(
   }
 }
 
-async function openWorkspace(workspace: WorkspaceInfo, informationManager: InformationManager): Promise<void> {
-  const result = await getSystemPath(workspace.handle, '/');
+async function seeInExplorer(
+  workspace: WorkspaceInfo,
+  informationManager: InformationManager,
+  workspaceAttributes: WorkspaceAttributes,
+  eventDistributor: EventDistributor,
+): Promise<void> {
+  if (workspaceAttributes.isHidden(workspace.id)) {
+    const answer = await askQuestion(
+      'WorkspacesPage.openInExplorerModal.workspace.title',
+      'WorkspacesPage.openInExplorerModal.workspace.description',
+      {
+        yesText: 'WorkspacesPage.openInExplorerModal.actionConfirm',
+        noText: 'WorkspacesPage.openInExplorerModal.actionCancel',
+      },
+    );
 
+    if (answer === Answer.Yes) {
+      await showWorkspace(workspace, workspaceAttributes, informationManager, eventDistributor);
+    }
+  }
+
+  const result = await getSystemPath(workspace.handle, '/');
   if (!result.ok) {
     await informationManager.present(
       new Information({
         message: { key: 'FoldersPage.open.folderFailed', data: { name: workspace.currentName } },
         level: InformationLevel.Error,
       }),
-      PresentationMode.Modal,
+      PresentationMode.Toast,
     );
   } else {
     window.electronAPI.openFile(result.value);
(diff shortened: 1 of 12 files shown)


client/src/views/files/FoldersPage.vue
```diff
@@ -1509,6 +1529,22 @@ async function seeInExplorer(entries: EntryModel[]): Promise<void> {
...
+
+      if (answer === Answer.Yes) {
+        await showWorkspace(workspaceInfo.value, workspaceAttributes, informationManager);
+        await showInExplorer(workspaceInfo.value.handle, entries[0].path, informationManager);
+      }
+      return;
+    }
     await showInExplorer(workspaceInfo.value.handle, entries[0].path, informationManager);
   } else {
     await openPath(workspaceInfo.value.handle, entries[0].path, informationManager, fileOperationManager, { skipViewers: true });
```
Ironicbay: This can be made better: if the user tries it on a folder, only an error shows instead of the modal.
I would first ask the question to show workspace or not, and proceed as usual if yes:
```
async function seeInExplorer(entries: EntryModel[]): Promise<void> {
  if (entries.length > 1 || !workspaceInfo.value || !isDesktop()) {
    return;
  }
  if (workspaceAttributes.isHidden(workspaceInfo.value.id)) {
    const answer = await askQuestion(
      'WorkspacesPage.openInExplorerModal.file.title',
      'WorkspacesPage.openInExplorerModal.file.description',
      {
        yesText: 'WorkspacesPage.openInExplorerModal.actionConfirm',
        noText: 'WorkspacesPage.openInExplorerModal.actionCancel',
      },
    );

    if (answer === Answer.No) {
      return;
    }
    await showWorkspace(workspaceInfo.value, workspaceAttributes, informationManager);
  }
  if (entries[0].isFile()) {
    await showInExplorer(workspaceInfo.value.handle, entries[0].path, informationManager);
  } else {
    await openPath(workspaceInfo.value.handle, entries[0].path, informationManager, fileOperationManager, { skipViewers: true });
  }
}
```

client/src/views/files/FoldersPage.vue
```diff
@@ -1387,6 +1390,23 @@ async function openEntries(entries: EntryModel[], options: OpenPathOptions): Pro
...
+        noText: 'WorkspacesPage.openInExplorerModal.actionCancel',
+      },
+    );
+
+    if (answer === Answer.Yes) {
+      await showWorkspace(workspaceInfo.value, workspaceAttributes, informationManager);
+      await openPath(workspaceHandle, entry.path, informationManager, fileOperationManager, options);
+    }
+    return;
+  }
```
Ironicbay, line 1428: Same remark here, use the `askQuestion()` result to know if you return early or not, and if not, exit the `if` to proceed as usual. It's easier to debug and to change code later on since we'd be using the same code regardless of the context.
```
if (isHidden) {
  askQuestion();
  if (answer === Answer.No) {
    return;
  }
  await showWorkspace();
}
// Proceed as usual
```

client/src/components/workspaces/utils.ts
```diff
@@ -193,7 +197,15 @@ export async function openWorkspaceContextMenu(
         await showWorkspace(workspace, workspaceAttributes, informationManager);
         break;
       case WorkspaceAction.UnMount:
-        await hideWorkspace(workspace, workspaceAttributes, informationManager);
+        if (storageManager === undefined) {
+          console.error('StorageManager is required to unmount a workspace with confirmation');
+          break;
+        }
```
Ironicbay: This is raised every time `Hide this workspace` is selected from the sidebar.
Either:
- Remove the `error` + `break` and add the condition `storageManager !== undefined` as a condition to call `unmountWorkspaceConfirmation`

or

- Add the storage manager to `openWorkspaceContextMenu` call from the sidebar

client/src/components/workspaces/utils.ts
```diff
@@ -329,6 +359,41 @@ async function renameWorkspace(workspace: WorkspaceInfo, newName: WorkspaceName,
...
+  if (role === MsModalResult.Confirm) {
+    if (data?.skipWorkspaceHiddenWarning === true) {
+      config.skipWorkspaceHiddenWarning = true;
+      await storageManager.storeConfig(config);
+    }
+
+    await hideWorkspace(workspace, workspaceAttributes, informationManager);
+  }
+}
+
```
Ironicbay, line 420: This being a one-time warning if the user checks the checkbox can be a bit confusing, I guess I'm not the only person to dismiss such warnings without paying much attention only to be a bit lost on what actually happened after. 
But this is going to get clearer as the implementation progresses.

client/src/views/files/FoldersPage.vue
fabienSvtr: I update with the modification you suggest except the :
```
if (entries[0].isFile()) {
    await showInExplorer(workspaceInfo.value.handle, entries[0].path, informationManager);
  } else {
    await openPath(workspaceInfo.value.handle, entries[0].path, informationManager, fileOperationManager, { skipViewers: true });
  }
```

in this case, the folder can't be open in explorer.

client/src/components/workspaces/utils.ts
fabienSvtr: Yes, but I think for know, it's necessary to implement this checkbox, if the user decided to check it, he should understand the consequences. And whether he don't find the workspace, he can still see it in Parsec

client/src/components/workspaces/utils.ts
```diff
@@ -123,6 +126,7 @@ export async function openWorkspaceContextMenu(
   informationManager: InformationManager,
   fromSidebar = false,
   isLargeDisplay = true,
+  storageManager?: StorageManager,
```
Max-7: Why make it optional? If it's not present, unmount does nothing.

client/src/views/files/FoldersPage.vue
```diff
@@ -1406,6 +1409,22 @@ async function openEntry(entryToOpen: EntryModel, options: OpenPathOptions): Pro
...
+      },
+    );
+    if (answer === Answer.No) {
+      return;
+    }
+
+    await showWorkspace(workspaceInfo.value, workspaceAttributes, informationManager);
+  }
+
   await pathOpener.openPath(workspaceHandle, entry.path, informationManager, options);
```
Max-7: You don't know if the mount was successful.

client/src/views/files/FoldersPage.vue
```diff
@@ -1531,11 +1550,23 @@ async function seeInExplorer(entries: EntryModel[]): Promise<void> {
...
+      },
+    );
+
+    if (answer === Answer.No) {
+      return;
+    }
+
+    await showWorkspace(workspaceInfo.value, workspaceAttributes, informationManager);
   }
+  await pathOpener.showInExplorer(workspaceInfo.value.handle, entries[0].path, informationManager);
```
Max-7: Same here, you don't know if the mount was successful.

client/src/views/files/FoldersPage.vue
```diff
@@ -1531,11 +1550,23 @@ async function seeInExplorer(entries: EntryModel[]): Promise<void> {
   if (entries.length > 1 || !workspaceInfo.value || !isDesktop()) {
     return;
   }
-  if (entries[0].isFile()) {
```
Max-7, line 1534: We used two different functions here, you replaced them with only `showInExplorer`. Sure about that?

client/src/views/workspaces/WorkspaceHiddenModal.vue
```diff
@@ -0,0 +1,102 @@
...
+const skipWorkspaceHiddenWarning = ref(false);
+
+async function confirmHidden(): Promise<boolean> {
+  if (!props.workspaceName) {
+    return false;
+  }
+
+  return await modalController.dismiss(
+    {
+      workspaceName: props.workspaceName,
```
Max-7: Don't need the name here.

client/src/views/workspaces/WorkspaceHiddenModal.vue
```diff
@@ -0,0 +1,102 @@
...
+import { ref } from 'vue';
+
+const props = defineProps<{
+  workspaceName: WorkspaceName;
+}>();
+
+const skipWorkspaceHiddenWarning = ref(false);
+
+async function confirmHidden(): Promise<boolean> {
+  if (!props.workspaceName) {
```
Max-7: Cannot happen, workspaceName is not declared as optional. Well technically it can happen if the name is empty but that's the caller's fault.

client/src/views/workspaces/WorkspaceHiddenModal.vue
```diff
@@ -0,0 +1,102 @@
...
+      title="WorkspacesPage.workspaceHiddenModal.title"
+      :close-button="{ visible: true }"
+      :confirm-button="{
+        disabled: workspaceName === undefined,
+        label: 'WorkspacesPage.workspaceHiddenModal.actionConfirm',
+        onClick: confirmHidden,
+      }"
+      :cancel-button="{
+        label: 'WorkspacesPage.workspaceHiddenModal.actionCancel',
+        disabled: false,
```
Max-7, line 15: I think you can just omit `disabled` and `onClick`, it will cancel by default.

client/src/views/workspaces/WorkspaceHiddenModal.vue
```diff
@@ -0,0 +1,102 @@
+<!-- Parsec Cloud (https://parsec.cloud) Copyright (c) BUSL-1.1 2016-present Scille SAS -->
+
+<template>
+  <ion-page class="modal">
+    <ms-modal
+      title="WorkspacesPage.workspaceHiddenModal.title"
+      :close-button="{ visible: true }"
+      :confirm-button="{
+        disabled: workspaceName === undefined,
```
Max-7: Cannot happen.

client/src/views/workspaces/WorkspaceHiddenModal.vue
```diff
@@ -0,0 +1,102 @@
...
+      }"
+    >
+      <div class="workspace-hidden">
+        <ion-text class="workspace-hidden__description body-lg">
+          <i18n-t
+            keypath="WorkspacesPage.workspaceHiddenModal.affectedWorkspaces"
+            scope="global"
+          >
+            <template #workspace>
+              <strong>{{ workspaceName ? workspaceName : '' }}</strong>
```
Max-7: workspaceName is always defined.

Max-7 (changes requested):

client/src/views/workspaces/WorkspaceHiddenModal.vue
fabienSvtr: `disabled` is required but yes for the `onClick`

client/src/views/files/FoldersPage.vue
fabienSvtr: Yes, the function is only about opening the explorer so we don't want to `openPath`. In both cases (folder and file) we want to call the `showInExplorer` function

client/src/components/workspaces/utils.ts
```diff
@@ -193,7 +197,13 @@ export async function openWorkspaceContextMenu(
         await showWorkspace(workspace, workspaceAttributes, informationManager);
         break;
       case WorkspaceAction.UnMount:
-        await hideWorkspace(workspace, workspaceAttributes, informationManager);
+        if (storageManager !== undefined) {
```
Max-7: Remove this check, can never happen.

client/src/components/workspaces/utils.ts
```diff
@@ -202,10 +222,11 @@ export async function openWorkspaceContextMenu(
 }
 
 export async function showWorkspace(
-  workspace: WorkspaceInfo,
+  workspace: WorkspaceInfo | StartedWorkspaceInfo,
   workspaceAttributes: WorkspaceAttributes,
   informationManager: InformationManager,
-): Promise<void> {
+  eventDistributor?: EventDistributor,
```
Max-7: Why is that optional?

client/src/components/workspaces/utils.ts
```diff
@@ -275,16 +312,35 @@ export async function hideWorkspace(
   }
 }
 
-async function openWorkspace(workspace: WorkspaceInfo, informationManager: InformationManager): Promise<void> {
-  const result = await getSystemPath(workspace.handle, '/');
+async function seeInExplorer(
+  workspace: WorkspaceInfo,
+  informationManager: InformationManager,
+  workspaceAttributes: WorkspaceAttributes,
+  eventDistributor?: EventDistributor,
```
Max-7: Why is it optional?

client/src/services/eventDistributor.ts
```diff
@@ -124,6 +125,11 @@ interface EntryDeletedData {
   path: FsPath;
 }
 
+interface WorkspaceMountpointsSyncData {
```
Max-7: Workspace synchronization is something else. This is just about whether the workspace is mounted or not.

client/src/views/menu/MainMenu.vue
```diff
@@ -767,6 +782,14 @@ async function removeRecentFile(file: RecentFile): Promise<void> {
   recentDocumentManager.removeFile(file);
 }
 
+async function onOpenWorkspaceContextMenu(workspace: WorkspaceInfo, event: Event, onFinished?: () => void): Promise<void> {
```
Max-7: What's the `onFinished` param?

client/src/components/workspaces/utils.ts
fabienSvtr: I updated it and removed the option.

client/src/components/workspaces/utils.ts
fabienSvtr: I updated it and removed the option.

client/src/views/menu/MainMenu.vue
fabienSvtr: It was to call the callback of the child component but this may differ from what happens in `MainMenu`.

client/src/services/eventDistributor.ts
fabienSvtr: Changed it to `WorkspaceMountpointInfo`

client/src/components/workspaces/utils.ts
```diff
@@ -224,6 +245,11 @@ export async function showWorkspace(
       }),
       PresentationMode.Toast,
     );
+    if (eventDistributor !== undefined)
```
Max-7: Not declared as optional, there's no need to check this.

client/src/components/workspaces/utils.ts
```diff
@@ -236,21 +262,26 @@ export async function showWorkspace(
...
-  workspace: WorkspaceInfo,
+  workspace: WorkspaceInfo | StartedWorkspaceInfo,
   workspaceAttributes: WorkspaceAttributes,
   informationManager: InformationManager,
+  eventDistributor: EventDistributor,
 ): Promise<void> {
   let ok = true;
   if (isDesktop()) {
-    const result = await unmountWorkspace(workspace);
+    const result = await unmountWorkspace(workspace as WorkspaceInfo);
```
Max-7: No, `WorkspaceInfo` is different from `StartedWorkspaceInfo`. They may share attributes, but it's not okay to cast from one to the other. You can make `unmountWorkspace` take a `WorkspaceInfo | StartedWorkspaceInfo` though.

client/src/views/menu/MainMenu.vue
Max-7: Is it used somewhere? Or is it just a leftovers from a copy/paste?

client/src/views/workspaces/WorkspacesPage.vue
```diff
@@ -591,13 +592,18 @@ async function performWorkspaceAction(action: WorkspaceAction): Promise<void> {
...
+    workspaceAttributes,
+    eventDistributor,
+    informationManager,
+    storageManager,
+    false,
+    isLargeDisplay.value,
+  );
   await refreshWorkspacesList();
-
-  if (onFinished) {
```
Max-7, line 598: Why did you remove the `onFinished` here? It's legitimate.

Max-7 (changes requested):

client/src/views/workspaces/WorkspacesPage.vue
fabienSvtr: I just removed the wrong `onFinished` (the one in `WorkspacePage` instead of the one in `MainMenu`), I updated and set it back here.

client/src/views/menu/MainMenu.vue
fabienSvtr: See my comment below

client/src/components/workspaces/utils.ts
fabienSvtr: Updated ✅

Max-7 (approved):

PR #11739, opened 2025-11-24
refactor(pki): Rename x509::Certificate to CertificateInformation

FirelightFlagboy:


libparsec/crates/platform_pki/src/x509/mod.rs
```diff
@@ -20,21 +20,21 @@ error_set::error_set! {
 }
 
 #[derive(Debug)]
-pub struct Certificate {
+pub struct CertificateInformation {
```
touilleMan: `X509CertificateInformation`, I'd rather always call it `X509Certificate` since "certificate" in our codebase refers to the Parsec certificate (e.g. device certificate, user revoked certificate etc.)

libparsec/crates/platform_pki/src/lib.rs
```diff
@@ -122,7 +122,7 @@ pub struct SignedMessageFromPki {
 pub use errors::SignMessageError;
 pub use platform::sign_message;
 
-pub use shared::{verify_message, Certificate, SignedMessage};
+pub use shared::{verify_message, Certificate, EndCertificate, SignedMessage};
```
touilleMan: see my other comment: I'd rather have all PKI certificate have the x509 prefix (this can be done in another PR though ^^)

server/src/data/pki.rs
```diff
@@ -152,3 +152,47 @@ crate::binding_utils::gen_py_wrapper_class_for_enum!(
...
+        self.to_end_certificate()
+            .map(|v| PyBytes::new(py, v.issuer()))
+    }
+
+    fn subject<'py>(&self, py: Python<'py>) -> PyResult<Bound<'py, PyBytes>> {
+        self.to_end_certificate()
+            .map(|v| PyBytes::new(py, v.subject()))
+    }
+
+    fn sha256_fingerprint<'py>(&self, py: Python<'py>) -> Bound<'py, PyBytes> {
```
touilleMan: This code is supposed to be a wrapper, so it's weird to add logic code in it
- if `sha256_fingerprint` is something important, then it should be present in the main `libparsec_types::X509Certificate`, and only re-exported here
- Otherwise, Python already comes with sha256 support, so it is trivial to compute the fingerprint there from the der (typically by expose `self.0.as_ref()` to Python though a `get_der(self) -> bytes` method).

libparsec/crates/platform_pki/src/x509/mod.rs
FirelightFlagboy: done

libparsec/crates/platform_pki/src/lib.rs
FirelightFlagboy: done

server/src/data/pki.rs
FirelightFlagboy: Done by replacing that method with the getter `der`

PR #11744, opened 2025-11-25
[MS] Initial loading of the page

### LINKED ISSUE #11673: Slow First Contentful Paint when loading wepapp
On some devices with a slow connection the FCP take a long time (5s-10s).

Normally it's solve by caching the assets but some organization clear the assets caches 
often causing user to encounter against the slow loading of the webapp.

Max-7:
Closes #11673

### DIFF
--- a/client/index.html
+++ b/client/index.html
@@ -36,16 +36,369 @@
     />
     <meta
       name="apple-mobile-web-app-title"
-      content="Parsec"
+      content="Parsec app loading"
     />
     <meta
       name="apple-mobile-web-app-status-bar-style"
-      content="black"
+      content="#FCFCFC"
     />
+
+    <script>
+      (function () {
+        var lang = navigator.language || navigator.userLanguage;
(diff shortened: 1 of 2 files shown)


client/index.html
```diff
@@ -36,16 +36,369 @@
...
+                stop-color="#2394FF"
+              />
+            </linearGradient>
+          </defs>
+        </svg>
+      </div>
+
+      <div class="loading-container">
+        <div id="loading-title">Loading Parsec app</div>
+        <div id="loading-subtitle">Initializing modules...</div>
```
Ironicbay, line 399: Is there any way to have localization for these strings?

Ironicbay (approved):

client/index.html
FirelightFlagboy: Look above, there's a simple localization

Max-7: The localization module is not loaded yet, we can't use the usual system, so we use a a workaround: set fixed texts to English, on load try to detect the language, if it's French, get the elements by id and set a French text.

PR #11756, opened 2025-11-26
feat(pki): Use smartcard cert to determine which root cert to use for validation

### LINKED ISSUE #11758: Update `load_answer_payload` to use a trusted root

FirelightFlagboy:
Closes #11758


bindings/generator/api/config.py
AureliaDolo, line 1: It would be nice if we all had the same formatting config (and maybe for that to be the one on the CI), what do you use ?

FirelightFlagboy: I use `ruff` :shrug:

PR #11762, opened 2025-11-26
Add RFC 1023 & 1024 for asynchronous enrollment

touilleMan:
Question:
- we currently use `pki_` prefix for everything related to PKI, except for the local device protection that is named `smartcard`. I think we should settle on either `pki` or `smartcard` for all code related to PKI (so most likely just rename `DeviceFileSmartcard` -> `DeviceFilePKI`).


docs/rfcs/1023-async-enrollment.md
```diff
@@ -0,0 +1,1093 @@
...
+In regular synchronous enrollment, the two peers are designed as claimer
+(the one that wants to be enrolled) and greeter (the one that does the enrollment).
+
+However this term is changed to submitter & accepter in the asynchronous enrollment.
+This is notably to reflect two fundamental differences:
+
+- All administrator in the organization can deal with any given asynchronous enrollment (while
+  there is only a single greeter per synchronous enrollment).
+- An asynchronous enrollment is created by the submitter (while a synchronous one is created
+  by the greeter).
```
AureliaDolo:
```suggestion
- When the submitter creates an enrollment it is asynchronous, whereas when an accepter creates the enrollment, it is synchronous.
```
It kinda sounded like the people involved could choose what kind of enrollment they wanted, but maybe it's just me. (Also greeter -> accepter)

docs/rfcs/1023-async-enrollment.md
```diff
@@ -0,0 +1,1093 @@
...
+
+### 2.2 - Workflow
+
+ 0. Alice wants to join CoolOrg organization.
+ 1. Alice uses the organization submission link to submit a request. This submission link
+    is specific to the organization and is common to all submit request (typically this
+    link can be pinned to a newcomer guide documentation).
```
AureliaDolo:
```suggestion
    link can be referenced in a newcomer guide documentation).
```

docs/rfcs/1023-async-enrollment.md
```diff
@@ -0,0 +1,1093 @@
...
+ 2. Alice authenticates with an external identity system (either connect trough OIDC or uses his smartcard containing a PKI identity).
```
AureliaDolo:
```suggestion
 2. Alice authenticates with an external identity system (either connect trough OIDC or uses their smartcard containing a PKI identity).
```
```suggestion
 2. Alice authenticates with an external identity system (either connect trough OIDC or uses her smartcard containing a PKI identity).
```

docs/rfcs/1023-async-enrollment.md
```diff
@@ -0,0 +1,1093 @@
...
+ 3. Alice creates a user encryption key pair and a device signing key pair.
+ 4. Alice saves the private parts of the user & device keys pairs on his machine's filesystem.
```
AureliaDolo:
```suggestion
 4. Alice saves the private parts of the user & device keys pairs on their machine's filesystem.
```
```suggestion
 4. Alice saves the private parts of the user & device keys pairs on her machine's filesystem.
```

docs/rfcs/1023-async-enrollment.md
```diff
@@ -0,0 +1,1093 @@
...
+            // what is expected (e.g. when using PKI, it corresponds to the `subjectAltName`
+            // field in the submitter's x509 certificate).
+            "name": "requested_human_handle",
+            "type": "HumanHandle"
+        }
+    ]
+}
+```
+
+### 3.2 - List the pending request
```
AureliaDolo, line 262: Does this lists only valid, still waiting for approval enrollments ?
If yes, should we add a list_untrusted like here #11841?

(it's for the admin to be able to see all enrollments, even if they are invalid)

docs/rfcs/1023-async-enrollment.md
```diff
@@ -0,0 +1,1093 @@
...
+            {
+                // Only user with ADMIN profile can use this command
+                "status": "author_not_allowed"
+            },
+            {
+                // The server did not found a request for the provided ID
+                "status": "enrollment_not_found"
+            },
+            {
+                // The request is no longer in pending state (either accepted or rejected)
```
AureliaDolo:
```suggestion
                // The request is no longer in pending state (either accepted, cancelled or rejected)
```

docs/rfcs/1023-async-enrollment.md
```diff
@@ -0,0 +1,1093 @@
...
+            {
+                // `payload_signature` cannot be verified by the server
+                "status": "invalid_payload_signature"
+            },
+            {
+                // The server did not found a request for the provided ID
+                "status": "enrollment_not_found"
+            },
+            {
+                // The request is no longer in pending state (either accepted or rejected)
```
AureliaDolo:
```suggestion
                // The request is no longer in pending state (either accepted, cancelled or rejected)
```

docs/rfcs/1023-async-enrollment.md
```diff
@@ -0,0 +1,1093 @@
...
+                "status": "enrollment_no_longer_available"
+            },
+            {
+                // The organization has reached the maximum number of active users
+                "status": "active_users_limit_reached"
+            },
+            {
+                // The user already exist in the organization
```
AureliaDolo:
```suggestion
                // The user already exist in the organization (aka email already in use) 
```

I'm not sure about that though, this could be about the user_id too.

docs/rfcs/1023-async-enrollment.md
```diff
@@ -0,0 +1,1093 @@
...
+                                "name": "algorithm",
+                                "type": "PkiSignatureAlgorithm"
+                            },
+                            {
+                                // Certificate used by the accepter to sign the payload
+                                "name": "accepter_der_x509_certificate",
+                                "type": "Bytes"
+                            },
+                            {
+                                // Optional list of intermediate certificates needed to
```
AureliaDolo, line 595: What happens if it's not provided ? The trustchain is not verified ?

docs/rfcs/1023-async-enrollment.md
```diff
@@ -0,0 +1,1093 @@
...
+                    "fields": [
+                        {
+                            // `SecretKey` encrypted by asymmetric key from the smartcard.
+                            "name": "encrypted_key",
+                            "type": "Bytes"
+                        },
+                        {
+                            // Certificate here refers to the X509 certificate that describes what is
+                            // in the smartcard.
+                            "name": "certificate_ref",
```
AureliaDolo, line 978:
```suggestion
                            "name": "submitter_certificate_ref",
```

docs/rfcs/1023-async-enrollment.md
```diff
@@ -0,0 +1,1093 @@
...
+Such validation is done at two levels:
+
+- In the accepter client: here the x509 root certificate is the same as the one for the
+  accepter's x509 certificate.
+- Server : here the allowed x509 root certificate are configured at startup.
+
+> [!NOTE]
+> The server validation is not strictly needed from a security standpoint, but it
+> prevents ending up with bad request in the list of pending enrollments that then
+> have to be manually rejected.
```
AureliaDolo: The invalid requests still need to be visible (they can be hidden) but the admin must be able to kind them to be able to debug the case where a legitimate user submits unknowingly an invalid certificate. So a manual action would still be needed to clean up those certificates

docs/rfcs/1023-async-enrollment.md
touilleMan: > It kinda sounded like the people involved could choose what kind of enrollment they wanted

That's the point here: this paragraph show the different between the synchronous enrollment (i.e. the one we currently have with SASCode) and an asynchronous one.

> Also greeter -> accepter

Here we are precisely talking about the peer that creates the new user in a synchronous enrollment, so it is the greeter ^^


Anyway, I've rephrased this part to make it more clear :+1:

docs/rfcs/1023-async-enrollment.md
touilleMan: :+1: much better !

docs/rfcs/1023-async-enrollment.md
touilleMan: :+1:

docs/rfcs/1023-async-enrollment.md
touilleMan: :+1:

docs/rfcs/1023-async-enrollment.md
touilleMan: The fact that an enrollment request is valid or not is something determined with the client (while here we are only describing the API protocol between the client and the server).

> If yes, should we add a list_untrusted like here https://github.com/Scille/parsec-cloud/pull/11841?

Yes, I think in the client we want to have a method that returns the list of enrollment without checking them. Which is fine as long as the enrollement accept method do the check (note enrollment reject should *not* do the check, otherwise it would be impossible to remove invalid enrollment from the list of submitted enrollments !).

docs/rfcs/1023-async-enrollment.md
touilleMan: :+1:

docs/rfcs/1023-async-enrollment.md
touilleMan: :+1:

docs/rfcs/1023-async-enrollment.md
touilleMan: Yes, the email already in use correspond to the error `human_handle_already_taken`
this error is about user ID already taken

I've completed the documentation to make this more clear :+1:

docs/rfcs/1023-async-enrollment.md
touilleMan: The trustchain is composed of:
- a leaf certifitate (i.e. the user's one)
- a root cerificate (i.e. the one trusted no matter what)
- intermediate certificates (i.e. ones that are signed by a trusted certificate)

Typically if we consider https://touilleman.xyz x509 certificate, we have the following chain:
`www.touilleman.xyz` <-- signed by -- `R12` <-- signed by-- `ISRG Root X1`

So here `R12` is the intermediate certificate and `ISRG Root X1` the root certificate (i.e. the only certificate the OS knows about).
So without providing `R12` in the list of intermediate certificates, then `www.touilleman.xyz` cannot be verified at all.

docs/rfcs/1023-async-enrollment.md
touilleMan: I'd rather not add a `submitter_` suffix here, otherwise for consistency this prefix should be also used in:
- the fields in OpenBao (i.e. `submitter_openbao_entity_id`, and `submitter_openbao_preferred_auth_id`...)
- potentially the fields referring the informations the submitter has provided (i.e. `requested_device_label`, `requested_human_handle`, `ciphertext_signing_key`, `ciphertext_private_key`)

However this pose an interesting point:
The x509 certificate here has been used for two things:
1) Sign the submit payload
2) Encrypt the soon-to-become user&devie's signing and private keys

And will be used for one additional thing:
3) Validate the accept payload

Currently we only acknowledge step 2 in this schema, so I'm going to rename the `encrypted_key` field as `identity_system` (I already use this "identity system" name in the documentation) and document that it is used for multiple things.

docs/rfcs/1023-async-enrollment.md
touilleMan: Note that if the server validates the request payload, the invalid enrollment request will never show up in the admin list (and hence never needs to be rejected).

Obviously the client that has submitted the invalid enrollment request should get a correct dialog from the GUI to understand 

However this is for long-term goal, in the short/middle term we don't do any server validation: which means it's the administrator that will see the invalid enrollment request (and hence it's *his* GUI that should correctly explain why this request is invalid, so that the admin can then help the user select the correct smartcard)

In any way, an invalid enrollment request should be a very rare event since it would only occurs when the wrong smartcard is selected, but it's extremely rare to have end user with multiple smartcards anyway !

docs/rfcs/1022-pki-device-storage.md
```diff
@@ -0,0 +1,62 @@
...
+1. Enable PKI-based authentication for Parsec devices (i.e. device keys file) using external identity providers.
+2. Provide a seamless user experience where device authentication is tied to PKI credentials (e.g. smartcard).
+
+## 2 - Overview
+
+PKI-based device creation is rather straightforward:
+
+1. User selects a PKI credential to use (typically selecting smartcard containing a secret key)
+2. Parsec client generates a new opaque key
+3. Parsec client encrypts the opaque key using the secret key from the smartcard
```
FirelightFlagboy: **issue:** 
```suggestion
3. Parsec client encrypts the opaque key using the public key from the smartcard
```

Reminder that in async crypto:

- public => encrypt
- private => decrypt

docs/rfcs/1022-pki-device-storage.md
```diff
@@ -0,0 +1,62 @@
...
+PKI-based device access works in a similar fashion
+
+1. Client reads `DeviceFileSmartcard` from local storage
+2. User is prompted to unlock the PKI credential he used to secure the given `DeviceFileSmartcard`
+3. Parsec client decrypts the opaque key using the PKI credential's private key
+4. Parsec client decrypts the `LocalDevice` using the opaque key
+
+## 3 - Data model
+
+### 3.1 - New `DeviceFileSmartcard` schema
```
FirelightFlagboy: **todo:** Rename to `DeviceFilePKI`

docs/rfcs/1023-async-enrollment.md
```diff
@@ -0,0 +1,1137 @@
...
+    > [!NOTE]
+    > The verification implies both Alice and Bob uses the same external identity system.
+    > Typically, it is not possible to validate a submit signed with a PKI with an admin
+    > that can only authenticate with OIDC.
+    > Similarly, if Alice and Bob both use PKI, they also must share the same root
+    > certificate for the validation to be successful.
+
+ 8. If Bob accepts the request (after being validated), it creates the Alice's user and
+    device certificates with the provided user & device public keys.
+ 9. Bob sends to the Parsec the accept payload along with its signature. This is similar
```
FirelightFlagboy: **typo:** 
```suggestion
 9. Bob sends to the Parsec server the accept payload along with its signature. This is similar
```

docs/rfcs/1023-async-enrollment.md
```diff
@@ -0,0 +1,1137 @@
...
+    to step 5 (but in reverse order).
+10. Alice checks the Parsec server periodically (polling) until she obtains the accept payload.
```
FirelightFlagboy: **nitpick:** 
```suggestion
10. Alice checks the Parsec server periodically (polling) until she obtains a response for initial request (either accepted with an accept payload or refused).
```

docs/rfcs/1023-async-enrollment.md
```diff
@@ -0,0 +1,1137 @@
...
+                    "name": "submit_payload",
+                    "type": "Bytes"
+                },
+                {
+                    "name": "submit_payload_signature",
+                    "type": "SubmitPayloadSignature"
+                }
+            ]
+        },
+        "reps": [
```
FirelightFlagboy, line 114: **issue:** Laking PKI specific errors

FirelightFlagboy, line 114: **issue:** Missing `already_enrolled`

In PKI we have that error, it's considered OK because it would have been return after the PKI trustchain check

docs/rfcs/1023-async-enrollment.md
```diff
@@ -0,0 +1,1137 @@
...
+            "type": "DeviceLabel"
+        },
+        {
+            // Note the email in the HumanHandle should not be blindly trusted here !
+            // Instead the identity validation step should ensure it correspond with
+            // what is expected (e.g. when using PKI, it corresponds to the `subjectAltName`
+            // field in the submitter's x509 certificate).
+            "name": "requested_human_handle",
+            "type": "HumanHandle"
+        }
```
FirelightFlagboy, line 257: **issue:** If we to always do that check, this field is not needed as we could obtain that information from somewhere else

FirelightFlagboy (changes requested):

docs/rfcs/1023-async-enrollment.md
```diff
@@ -0,0 +1,1137 @@
...
+                    "name": "accept_payload",
+                    "type": "Bytes"
+                },
+                {
+                    "name": "accept_payload_signature",
+                    "type": "AcceptPayloadSignature"
+                }
+            ]
+        },
+        "reps": [
```
FirelightFlagboy, line 475: **issue:** Missing errors for PKI check

docs/rfcs/1023-async-enrollment.md
```diff
@@ -0,0 +1,1137 @@
...
+        },
+        {
+            // Submitter asks for device_label/human_handle, but accepter has final word on this
+            "name": "requested_device_label",
+            "type": "DeviceLabel"
+        },
+        {
+            "name": "requested_human_handle",
+            "type": "HumanHandle"
+        },
```
FirelightFlagboy, line 950: **issue:** The handle MUST be obtained by the IdP as we cannot trust that value

docs/rfcs/1023-async-enrollment.md
```diff
@@ -0,0 +1,1137 @@
...
+
+### 5.1 - PKI based
+
+In PKI mode, the submitter signs his request payload using his personal PKI's signing key, then
+provides it along with his personal x509 certificate (and any intermediary x509 certificates needed).
+
+This way submit payload validation is done by:
+
+1 - Validating the x509 certificate chain
+2 - Ensuring the x509 root certificate in the chain is expected (see below).
```
FirelightFlagboy: **issue:** The client will not provide the Root, it's the server that should be configured with a list of trusted root (#11648)

docs/rfcs/1023-async-enrollment.md
```diff
@@ -0,0 +1,1137 @@
...
+3 - Validating the submit payload signature according to the verify key in the submitter's x509 certificate.
+4 - Checking the email in the `requested_human_handle` field of the submit payload correspond
+    to the `subjectAltName` field in the submitter's x509 certificate.
```
FirelightFlagboy: **todo:** Use correct markdown ordered list format

FirelightFlagboy: **issue:** That operation is always done to validate that field meaning we could shortcut that field and directly use the value from the certificate

docs/rfcs/1022-pki-device-storage.md
touilleMan: :+1:

docs/rfcs/1022-pki-device-storage.md
touilleMan: :+1:

docs/rfcs/1023-async-enrollment.md
touilleMan: :+1:

docs/rfcs/1023-async-enrollment.md
touilleMan: :+1:

docs/rfcs/1023-async-enrollment.md
touilleMan: That's not necessarily true: typically a X509 certificate can contain multiple emails, so we wouldn't know which one to use in this case.

Similarly, we could also want to validate the human handle label, but it's not something that is always possible (typically in OpenBao we only have access to the email).

I think it's better to provide the human handle here since it means we can obtain this information without having to deal with the identity system, which is convenient when listing the enrollment.

docs/rfcs/1023-async-enrollment.md
touilleMan: :+1:

docs/rfcs/1023-async-enrollment.md
touilleMan: I guess the `in the chain` part is misleading: the root certificate is never supposed to be provided by the submitter (or by anyone)

I've reworded this part :+1:

docs/rfcs/1023-async-enrollment.md
touilleMan: see https://github.com/Scille/parsec-cloud/pull/11762#discussion_r2603356664

touilleMan: Beside, can we also check the human handle name against a field in the X509 certificate ?

docs/rfcs/1023-async-enrollment.md
touilleMan: `AsyncEnrollmentLocalPending` is only stored in local so the risk of it being modified is low.

Beside, if a modification occurs on the `requested_*` fields, this has no real impact since those fields are not used after the submit payload is send (they are stored here only as an easy way to know what this enrollment is about)

However opening the door to data modification is never a good idea, so we could modify the schema so that an actor without the ciphertext key cannot do a meaningful modification, this can be done in two ways:

1. Duplicate the fields: we still need the fields (organization_id, human_handle etc.) to be readable without decryption, but we can also duplicate them encrypted. This way both version are compared upon decryption.
2. Generate a hash of the unencrypted fields and store it encrypted. This way the hash can be recomputed and compared to the expected value upon decryption.

Solution 2 seems better at first, but it works poorly in case a field is latter added (since an older version of Parsec will simply ignore this field when decrypting the structure, hence leading to an invalid hash !).

(there is some other solutions like providing an embedded structure containing all the fields, or provided for each field a separated encrypted hash; but they are just more cumbersome anyway)

So I guess using solution 1 is good enough, in practice it means adding the following fields to the structure:
- `ciphertext_server_url`
- `ciphertext_organization_id`
- `ciphertext_submitted_on`
- `ciphertext_enrollment_id`
- `ciphertext_requested_device_label`
- `ciphertext_requested_human_handle`

(so only `identity_system` doesn't have it's encrypted twin since it is the field used to retrieve the decryption key)

docs/rfcs/1023-async-enrollment.md
FirelightFlagboy: We are already able to generate a human handle from a certificate, my point stand:

If we need to check against a source of truth, just use that source

docs/rfcs/1023-async-enrollment.md
touilleMan: :+1: (note the error has been renamed `email_already_enrolled`)

touilleMan: I've added `invalid_der_x509_certificate` and `invalid_x509_trustchain` :+1:

docs/rfcs/1023-async-enrollment.md
touilleMan: I've added `invalid_der_x509_certificate` and `invalid_x509_trustchain` :+1:

docs/rfcs/1023-async-enrollment.md
touilleMan: > If we need to check against a source of truth, just use that source

Again, there is a difference between being able to validate a value and generating this value:

- An X509 certificate can contain multiple email, which one should be kept ?
- Some identity system might not be able to provide the human handle label (that's the case with the current approach we have with OpenBao where only the email is available).
- It is convenient to skip entirely the identity system when validation is not required (i.e. when the admin wants to list the pending enrollments)

> We are already able to generate a human handle from a certificate

That what I though :smiley: , however my question was more "what is the field in the X509 certificate we use to obtain the human handle label" so that I can update the RFC with this information.

FirelightFlagboy: > what is the field in the X509 certificate we use to obtain the human handle label

We use a combination of multiple fields: commonName + either SAN.email or emailAddress

docs/rfcs/1023-async-enrollment.md
```diff
@@ -0,0 +1,1195 @@
...
+        "reps": [
+            {
+                "status": "ok"
+            },
+            {
+                // The user does not have the permission required to perform this action
+                "status": "author_not_allowed"
+            },
+            {
+                // One of the provided certificate is not valid
```
FirelightFlagboy: **nitpick:** 
```suggestion
                // One of the provided user or device certificate is not valid
```

docs/rfcs/1023-async-enrollment.md
```diff
@@ -0,0 +1,1195 @@
...
+> [!NOTE]
+>
+> - Obviously the X509 root certificate is never transmitted between clients,
+>   it is instead obtained from the clients's OS certificates store.
+> - The server validation is not strictly needed from a security standpoint, but it
+>   prevents ending up with bad request in the list of pending enrollments that then
+>   have to be manually rejected.
+
+Finally the accepter also signs his accept payload that has to be then validated
+by the submitter.
```
FirelightFlagboy: **nitpick:** The server also perform a verification for the request, it's a best effort so an admin does not provide a request signed by a PKI that is not trusted by the submitter

docs/rfcs/1023-async-enrollment.md
```diff
@@ -0,0 +1,1195 @@
...
+
+# Step 2: Create key
+POST https://openbao-dev.parsec.cloud/v1/transit/keys/user-d42286c0-a41d-bf0f-7dab-c9c27a0f0a58
+Headers: {"X-Vault-Token": "s.XXX"}
+<<< status: 200 (also 200 if key already exists)
+<<< {'request_id': '1cfb7461-5b77-333f-b952-21f34149267f', 'lease_id': '', 'renewable': False, 'lease_duration': 0, 'data': {'allow_plaintext_backup': False, 'auto_rotate_period': 0, 'deletion_allowed': False, 'derived': False, 'exportable': False, 'imported_key': False, 'keys': {'1': {'certificate_chain': '', 'creation_time': '2025-12-01T15:50:33.606080602Z', 'name': 'ed25519', 'public_key': 'ig9jt4uLZj4AxYSqdS3XkyymPamEijV4JaNHAfa1yUY='}}, 'latest_version': 1, 'min_available_version': 0, 'min_decryption_version': 1, 'min_encryption_version': 0, 'name': 'user-d42286c0-a41d-bf0f-7dab-c9c27a0f0a58', 'soft_deleted': False, 'supports_decryption': False, 'supports_derivation': True, 'supports_encryption': False, 'supports_signing': True, 'type': 'ed25519'}, 'wrap_info': None, 'warnings': None, 'auth': None}
+
+# Step 3: Sign payload
+POST https://openbao-dev.parsec.cloud/v1/transit/sign/user-d42286c0-a41d-bf0f-7dab-c9c27a0f0a58
+Headers: {"X-Vault-Token": "s.XXX"}
```
FirelightFlagboy: **nitpick:** Missing placeholder payload in example

docs/rfcs/1023-async-enrollment.md
```diff
@@ -0,0 +1,1195 @@
...
+<<< status: 200
+<<< {'request_id': '1288a880-525e-5e8d-a630-33e2c63847c6', 'lease_id': '', 'renewable': False, 'lease_duration': 0, 'data': {'key_version': 1, 'signature': 'vault:v1:C4jRZx+ymLou26tN8Q2KDy46dA47W7x/MH6nuEZVqdz+H0RvoaFbQUAHceBKhB+Qow2qXAXiRFAFWKGPZU93CQ=='}, 'wrap_info': None, 'warnings': None, 'auth': None}
+
+# Step 4&5 omitted
+
+# Step 6: Bob verify payload
+POST https://openbao-dev.parsec.cloud/v1/transit/verify/user-d42286c0-a41d-bf0f-7dab-c9c27a0f0a58
+Headers: {"X-Vault-Token": "s.YYY"}
```
FirelightFlagboy: **nitpick:** Missing placeholder payload in example

docs/rfcs/1023-async-enrollment.md
touilleMan: :+1:

docs/rfcs/1023-async-enrollment.md
touilleMan: :+1: I've added a word about this

docs/rfcs/1023-async-enrollment.md
touilleMan: :+1: good catch !

docs/rfcs/1023-async-enrollment.md
touilleMan: :+1:

docs/rfcs/1023-async-enrollment.md
```diff
@@ -0,0 +1,1197 @@
...
+            {
+                // `submit_payload` field doesn't correspond to a valid `AsyncEnrollmentSubmitPayload`
+                "status": "invalid_submit_payload"
+            },
+            {
+                // `submit_payload_signature` doesn't match with `submit_payload`
+                "status": "invalid_submit_payload_signature"
+            },
+            {
+                // A provided X509 certificate (either the submitter's or an intermediate) is invalid
```
AureliaDolo, line 151: Should we provide a way to show which cert is invalid ?

docs/rfcs/1023-async-enrollment.md
```diff
@@ -0,0 +1,1197 @@
...
+                // Only user with ADMIN profile can use this command
+                "status": "author_not_allowed"
+            },
+            {
+                // The server did not found a request for the provided ID
+                "status": "enrollment_not_found"
+            },
+            {
+                // The request is no longer in pending state (either accepted, cancelled or rejected)
+                "status": "enrollment_no_longer_available"
```
AureliaDolo, line 418: Should we distinguish between enrollment accepted vs the other, as in one case the situation is already as the admin wanted but in the accepted case it's quite the opposite.

docs/rfcs/1023-async-enrollment.md
AureliaDolo: So how optional is it ? From what I understand, it's optional iif the leaf cert is a root cert.

docs/rfcs/1023-async-enrollment.md
```diff
@@ -0,0 +1,1210 @@
...
+            },
+            {
+                // A provided X509 certificate (either the submitter's or an intermediate) is invalid
+                "status": "invalid_der_x509_certificate"
+            },
+            {
+                // Cannot trust provided trustchain.
+                // This usually means the trustchain is technically valid, but
+                // it relies on a X509 root certificate that is not among the
+                // ones trusted by the server.
```
FirelightFlagboy: **note:** During PKI that error is also returned in case the chain is too complex (more that 10 elements for now)

docs/rfcs/1023-async-enrollment.md
```diff
@@ -0,0 +1,1210 @@
...
+            // useful to list the pending enrollments without requiring authentication).
+            "name": "cleartext_content",
+            "type": "Bytes"
+        },
+        {
+            // `HashDigest` (i.e. sha256 hash) of the `cleartext_content` field,
+            // encrypted with the ciphertext key (i.e. a `SecretKey`).
+            // This is used to ensure the cleartext data hasn't been tempered.
+            "name": "ciphertext_cleartext_content_digest",
+            "type": "Bytes"
```
FirelightFlagboy, line 900: **issue:** We should standardize `X509CertificateHash` as it use the SRI-hash format that include information about the used hash algo and use it here

docs/rfcs/1023-async-enrollment.md
AureliaDolo: It's 100.

docs/rfcs/1023-async-enrollment.md
touilleMan: I'm going to promote `HashDigest` as a type usable in the schemas instead :+1:

docs/rfcs/1023-async-enrollment.md
touilleMan: I'm going to document that :+1:

docs/rfcs/1023-async-enrollment.md
touilleMan: we discussed about this: it is most likely the certificates are either all good or all bad, so it's not worth it to indicate what is the wrong certificate

docs/rfcs/1023-async-enrollment.md
touilleMan: Not needed: the admin just wants to reject the enrollment here, if he cannot then the GUI refresh his page and then the enrollment is no longer visible

FirelightFlagboy (approved):

PR #11765, opened 2025-11-27
refactor(ruff): Create global ruff config

FirelightFlagboy:

### DIFF
--- a/.github/workflows/ci-python.yml
+++ b/.github/workflows/ci-python.yml
@@ -190,6 +190,7 @@ jobs:
       - name: Check python code style
         shell: bash -eux {0}
         run: |
+          poetry -P server run ruff version
           for step in ruff-format ruff-check pyright ${{ !inputs.style-only && 'sqlfluff sqlfluff-in-python' || '' }}; do
             python \
               ${{ steps.pre-commit.outputs.install-path }} \
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -129,20 +129,7 @@ repos:
     hooks:
       - id: ruff-check
         name: ruff check (using `poetry`)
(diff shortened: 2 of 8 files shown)


.pre-commit-config.yaml
```diff
@@ -154,7 +154,6 @@ repos:
         require_serial: true
```
touilleMan, line 141: should also correct the ruff-check rule (and you can remove the "wow this looks ugly ..." comment since it is no longer the case

.github/workflows/ci-python.yml
```diff
@@ -198,7 +199,8 @@ jobs:
               --show-diff-on-failure \
               --verbose \
               --color=always \
-              ${{ steps.pre-commit.outputs.suggested-args }}
+              --all-files
+              # ${{ steps.pre-commit.outputs.suggested-args }}
```
touilleMan: why comment out the suggested-args ?
(note there is already the `pass_filenames: false` that can be put in a check config, see what does `ruff-check` for instance)

FirelightFlagboy: Hint: [`1a3b9a3` (#11765)](https://github.com/Scille/parsec-cloud/pull/11765/commits/1a3b9a377dcf4d469e34b84a1c33ff5181956cfa) ;)

.pre-commit-config.yaml
FirelightFlagboy: It's still the case as if you pass a file that should be ignored (e.g. `misc/bench.py`) `ruff check` will check it and generate errors

FirelightFlagboy: NVM, indeed we can remove the comment and simplify the `ruff-check` command

PR #11807, opened 2025-12-01
feat(pki): Return emails contained in cert

FirelightFlagboy:


libparsec/crates/platform_pki/src/x509/mod.rs
```diff
@@ -42,6 +42,33 @@ impl X509CertificateInformation {
             }
         })
     }
+
+    /// List emails contained in the certificate
+    ///
+    /// Will favor `subjectAltName` `emailAddress` over subject `emailAddress` as the later one is
```
touilleMan:
```suggestion
    /// Will favor `subjectAltName` over subject `emailAddress` as the later one is
```

PR #11815, opened 2025-12-02
feat(pki): Use human handle from cert for pki submit

### LINKED ISSUE #11811: Server: pki-submit should rely on provided cert to generate human handle

FirelightFlagboy:
Closes #11811


server/tests/api_v5/anonymous/test_pki_enrollment_submit.py
```diff
@@ -292,6 +321,8 @@ async def test_anonymous_pki_enrollment_submit_email_already_used(
             human_handle=coolorg.alice.human_handle,
         ).dump(),
     )
+
+    # 3. ...
```
AureliaDolo, line 325: something is missing ?

FirelightFlagboy: Not for me, the text would have been `check for email already submitted` but I found that redundant with the assert below

PR #11824, opened 2025-12-03
feat(editor): add timeout modal for file loading issues and enhance error handling

NicoTuxx:
## Testing the Timeout Modal

To test the scenario where the timeout modal appears but the file eventually loads:

**Temporarily make these changes** in `FileEditor.vue`:

1. **Reduce the timeout threshold** (around line 445):
```typescript
const LOADING_TIMEOUT_MS = 3000; // Reduced from 30000 for testing
```

2. **Add a simulated delay** to the `onReady` event (around line 369):
```typescript
onReady: (): void => {
  // Simulate slow file loading to trigger timeout modal
  const SIMULATED_DELAY = 5000; // 5 seconds
  window.setTimeout(() => {
    window.electronAPI.log('info', 'CryptPad editor is ready and document loaded successfully');
    fileLoaded.value = true;
  }, SIMULATED_DELAY);
},
```

**Expected behavior:**
1. Open a CryptPad-supported file (e.g., `.xlsx`, `.md`)
2. After 3 seconds, a timeout modal should appear asking if you want to wait or go back
3. Click "Wait" button
4. After ~2 more seconds, the file should finish loading
5. Click "Close" to dismiss the modal and start editing (it should not redirect you)

This tests that the timeout handling properly continues waiting when the user chooses to wait, correctly detects when the file finally loads, and updates the button accordingly.

**Remember to remove both testing modifications before committing.**

### DIFF
--- a/client/src/locales/en-US.json
+++ b/client/src/locales/en-US.json
@@ -2512,13 +2512,10 @@
                 "impossibleToOpen": "Cannot edit this file",
                 "genericError": "Failed to open the file",
                 "unsupportedFileType": "File not supported by Parsec",
-                "editionNotAvailable": "Document editing is unavailable",
-                "tooLongToOpen": "The file is taking too long to open"
+                "editionNotAvailable": "Document editing is unavailable"
             },
             "informationEditDownload": "If you want to edit this file, you can download it and open it locally on your device.",
             "informationPreviewDownload": "If you want to preview this file, you can download it and open it locally on your device.",
-            "tooLongToOpenOnWeb": "You can try downloading it and opening with default app (right click on file > 'Download').",
-            "tooLongToOpenOnDesktop": "You can try opening it with default app (right click on file > 'Open with default app').",
             "unknownFileExtension": "This file type extension could not be identified. Please check that it's correctly written (.txt, .pdf, .doc, etc.).",
             "noContentFileType": "Could not open this file in the Parsec editor as its type could not be identified.",
             "noFolderPreview": "Cannot preview folders.",
@@ -2577,9 +2574,12 @@
         "globalTitle": "Cannot open file",
         "errors": {
             "titles": {
-                "networkOffline": "Network connection lost"
+                "networkOffline": "Network connection lost",
+                "tooLongToOpen": "The file seems to be taking too long to open"
             },
             "networkOfflineMessage": "Your network connection has been lost. Please check your internet connection and try again. Any unsaved changes may be lost.",
+            "tooLongToOpenOnWeb": "You can try downloading it and opening with default app (right click on file > 'Download').",
+            "tooLongToOpenOnDesktop": "You can try opening it with default app (right click on file > 'Open with default app').",
             "contentInfoMissing": "The file content information is empty or missing.",
             "fileExtensionMissing": "The file type (extension) is missing or empty.",
             "fileNameMissing": "The file name is empty or missing.",
@@ -2589,12 +2589,15 @@
         "advices": {
             "title": "Some tips to help resolve this issue",
             "advice1": "Check that the file can be opened from an external application.",
-            "advice2": "Re-import the file (check versions).",
-            "advice3": "If the file used to open properly, try restoring a previous version (through file history)"
+            "advice2": "If the file used to open properly, try restoring a previous version (through file history)"
         },
         "actions": {
             "backToFiles": "Back to files",
-            "retry": "Reload page"
+            "retry": "Reload page",
+            "openWithDefaultApp": "Open with default app",
+            "downloadFile": "Download file",
+            "wait": "Wait",
+            "close": "Close"
         },
         "saving": {
             "unsaved": "Changes unsaved",
--- a/client/src/locales/fr-FR.json
+++ b/client/src/locales/fr-FR.json
@@ -2513,13 +2513,10 @@
                 "impossibleToOpen": "Impossible d'éditer ce fichier",
                 "genericError": "Impossible d'ouvrir ce fichier",
                 "unsupportedFileType": "Type de fichier non pris en charge par Parsec",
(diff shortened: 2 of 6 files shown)


client/src/locales/fr-FR.json
```diff
@@ -2581,9 +2577,12 @@
         "globalTitle": "Impossible d'ouvrir le fichier",
         "errors": {
             "titles": {
-                "networkOffline": "Connexion réseau perdue"
+                "networkOffline": "Connexion réseau perdue",
+                "tooLongToOpen": "Le fichier met trop de temps à s'ouvrir"
```
fabienSvtr: I think something like "Le fichier met plus de temps que prévu à s’ouvrir"

client/src/locales/en-US.json
```diff
@@ -2580,9 +2576,12 @@
         "globalTitle": "Cannot open file",
         "errors": {
             "titles": {
-                "networkOffline": "Network connection lost"
+                "networkOffline": "Network connection lost",
+                "tooLongToOpen": "The file seems to take too long to open"
```
```suggestion
                "tooLongToOpen": "The file seems to be taking too long to open"
```

client/src/views/files/handler/editor/FileEditor.vue
```diff
@@ -315,33 +349,78 @@ async function openFileWithCryptpad(): Promise<boolean> {
   };
 
   try {
+    // Start CryptPad (non-blocking)
     await cryptpadApi.open(cryptpadConfig);
-    window.electronAPI.log('info', 'CryptPad editor initialized successfully');
-    return true;
+
+    // Set up timeout mechanism with restart capability
```
mmmarcos: I would add more details here
```suggestion
    // Set up timeout for loading files
    // If the file does not open before timeout, a dialog is displayed 
    // to ask the user if it wants to keep waiting or go back to files.
```

client/src/views/files/handler/editor/FileEditor.vue
```diff
@@ -315,33 +349,78 @@ async function openFileWithCryptpad(): Promise<boolean> {
...
+    const LOADING_TIMEOUT_MS = 30000;
+    let shouldContinueWaiting = true;
+
+    // TODO: Fix DOCX timeout false positives - OnlyOffice doesn't reliably call onReady for 'doc' type
+    // See: https://github.com/Scille/parsec-cloud/issues/11753
+    // For now, we ignore timeout for DOCX files to avoid false positive "corrupted file" modals
+    const shouldSkipTimeout = documentType.value === CryptpadDocumentType.Doc;
+
+    if (shouldSkipTimeout) {
+      window.electronAPI.log('info', 'Skipping timeout for DOCX file (known OnlyOffice onReady issue)');
```
mmmarcos: I don't think the parenthesis are needed here
```suggestion
      window.electronAPI.log('info', 'Skipping loading timeout for DOCX file');
```

client/src/views/files/handler/editor/FileEditor.vue
```diff
@@ -315,33 +343,80 @@ async function openFileWithCryptpad(): Promise<boolean> {
...
+        window.electronAPI.log('info', 'CryptPad editor initialized successfully');
+        return true;
+      }
+
+      // Timeout occurred - show modal
+      window.electronAPI.log('warn', 'CryptPad loading timeout - file appears to be corrupted or too large');
+      const result = await openTimeoutModal();
+
+      // Check if file loaded while modal was open
+      if (fileLoaded.value) {
```
Max-7, line 380: Everything below could be changes into `shouldContinueWaiting = result === 'close'` to make use of the while's condition (right now, `while (true)` would do the same thing) and remove some code.

client/src/views/files/handler/editor/FileEditor.vue
```diff
@@ -187,35 +167,82 @@ async function loadCryptpad(): Promise<boolean> {
...
+  if (role === MsModalResult.Confirm) {
+    if (fileLoaded.value) {
+      return 'close';
+    } else {
+      return 'wait';
+    }
+  }
+
+  // If user clicks back button (cancel), navigate back
+  if (role === MsModalResult.Cancel) {
```
Max-7: Should be an `else`

client/src/views/files/handler/editor/FileEditor.vue
```diff
@@ -43,51 +43,45 @@
...
-const informationManager: InformationManager = inject(InformationManagerKey)!;
 const fileEditorRef = useTemplateRef('fileEditor');
 const documentType = ref<CryptpadDocumentType | null>(null);
 const cryptpadInstance = ref<Cryptpad | null>(null);
 const fileUrl = ref<string | null>(null);
 const error = ref('');
 const eventDistributor: EventDistributor = inject(EventDistributorKey)!;
 let eventCbId: null | string = null;
+const fileLoaded = ref(false);
+let activeModal: HTMLIonModalElement | null = null;
```
Max-7: This seems useless currently.

client/src/views/files/handler/editor/EditorIssueModal.vue
```diff
@@ -0,0 +1,172 @@
...
+import { checkmarkCircle } from 'ionicons/icons';
+import { MsModal, MsModalResult, MsReportText, MsReportTheme, Translatable } from 'megashark-lib';
+import { computed, Ref } from 'vue';
+
+const props = defineProps<{
+  status: EditorIssueStatus;
+  fileLoaded?: Ref<boolean>;
+}>();
+
+const title = computed(() => {
```
Max-7: Could group everything in one computed instead of having multiple. Something like
```
computed<{ title: Translatable, message?: Translatable, theme: MsReportTheme, button?: Translatable }>(() => {
    switch (props.status) { 
        ... 
    }
})
```

client/src/views/files/handler/editor/FileEditor.vue

PR #11826, opened 2025-12-04
[MS] Fix invalid folder path error message

### LINKED ISSUE #11823: Invalid folder name in error message
### Parsec version

3.7.0

- [x] Any
- [ ] macOS
- [ ] Linux
- [ ] Windows
- [ ] Web

### Bug description

When logged-in while offline (no Internet connection) and a folder is not available (not in local cache) the following error message is displayed:
```
Unable to list content of folder `My Folder`
```

Clicking again will make the folder name to be added to the error message as if it was part of the path (`My Folder/My Folder`):
```
Unable to list content of folder `My Folder/My Folder`
```

This can continue on an on if the user keeps clicking:
Image

### How should the application behave

_No response_

### Reproduction steps

```bash
1. Make sure you are offline (turn off WiFi or unplug cable)
2. Login to an existing organization
3. Navigate into sub-folders until you reach a folder displaying error message "Unable to list content of folder `My Folder`".
4. Continue clicking in the same folder
```

### Relevant output:

_No response_

fabienSvtr:
https://github.com/user-attachments/assets/e8cd3b3f-bb3c-418a-a6b9-1025f1fa7efb

closes #11823


client/src/locales/en-US.json
```diff
@@ -783,7 +783,8 @@
             "copyFailedMaxRecursion": "Too many entries to copy.",
             "copyFailedOneFailed": "One or more entries could not be copied.",
             "copyFailedSourceDoesNotExist": "Entry {source} does not exist.",
-            "seeInExplorerFailed": "Failed to open the path."
+            "seeInExplorerFailed": "Failed to open the path.",
```
mmmarcos: Not related to the PR, but this could be more explicit isn't it?

```suggestion
            "seeInExplorerFailed": "Failed to open the selected entry in file explorer.",
```

client/src/locales/en-US.json
```diff
@@ -783,7 +783,8 @@
...
+            "navigateMissingPath": "Failed to open the folder, check your internet network"
```
mmmarcos: I wouldn't mention internet network. As stated in the issue, the message seems to be displayed even if the user is online (e.g. when data is not yet fully available locally)

```suggestion
            "navigateMissingPath": "Failed to open the selected folder."
```

client/src/locales/en-US.json
fabienSvtr: Yes, much more explicit.

mmmarcos:
Just tested it again, the issue still persists: 
- on click to open a folder MyFolder (while offline)
- an error is displayed (OK)
- the page stays in the current directory, i.e. it does not enter MyFolder (OK)
- **navigation is broken, all subsequent actions consider that the app has entered MyFolder (KO)**

mmmarcos (approved):
It fixes the issue. Did not review the code though.

PR #11841, opened 2025-12-05
feat(pki): Provide pki_list_untrusted

### LINKED ISSUE #11838: Make `pki_list` only list request, no verification done

FirelightFlagboy:
Simpler than `pki_enrollment_list` because it does not verify the items and not require a cert ref.

Closes #11838


FirelightFlagboy:
> It's just the client impl here ?

Yep only client, I always do the binding in a second part

PR #11869, opened 2025-12-08
[MS] Fix skipped account_login tests

Ironicbay:


client/tests/e2e/specs/account_login.spec.ts
```diff
@@ -148,12 +142,6 @@ msTest.skip('Account create registration device and auto-register', async ({ bro
   await home.locator('.organization-list').locator('.organization-card').nth(0).click();
   await fillIonInput(home.locator('#password-input').locator('ion-input'), 'P@ssw0rd.');
   await home.locator('.login-card-footer').locator('.login-button').click();
-  await answerQuestion(home, false, {
```
Max-7, line 151: If you have to remove code from the tests, it means the feature is not working properly. Either keep the skip or fix the feature.

PR #11875, opened 2025-12-09
refactor(test-pki): Commit test-pki cert

FirelightFlagboy:
Also update CI to use generated test-pki


AureliaDolo (approved):
Maybe the def test_pki( fixture could be updated to look for the certs here by default

.github/workflows/ci-python.yml
```diff
@@ -34,6 +34,7 @@ env:
     -vv
     -x
   postgresql-version: 16
+  TEST_PKI_DIR: ${{ github.workspace }}/.test-pki
```
touilleMan: Remove this: we shoudn't need an environ variable to use the default configuration (otherwise it would mean we cannot run the test by just doing `py.test server/tests` when developping...)

.github/workflows/ci-rust.yml
```diff
@@ -62,6 +62,7 @@ jobs:
     # https://github.com/Scille/parsec-cloud/pkgs/container/parsec-cloud%2Fparsec-testbed-server
     env:
       TESTBED_SERVER: http://localhost:6777
+      TEST_PKI_DIR: ${{ github.workspace }}/.test-pki
```
touilleMan: same, to remove

.github/workflows/ci-rust.yml
```diff
@@ -241,6 +235,8 @@ jobs:
           - name: 🏁 Windows
             os: windows-2025
     name: "${{ matrix.name }}: 🦀 Rust tests"
+    env:
+      TEST_PKI_DIR: ${{ github.workspace }}/.test-pki
```
touilleMan: same, to remove

libparsec/crates/platform_pki/test-pki/Cert/alice.crt
```diff
@@ -0,0 +1,20 @@
+-----BEGIN CERTIFICATE-----
```
touilleMan, line 1: Avoid putting things in the root folder please :pray: 
Even hidden, this cost real estate coding space (and cognitive energy to filter it out) when listing what's in the codebase

Instead we can put this code in `libparsec/crates/platform_pki/tests` (and consider fair to have any other crate/component to access them from there)

Typically the python tests would do something like:

```python
# server/tests/common/pki.py
PROJECT_DIR = Path(__file__).parent / "../../.."
PKI_CERTIFICATES_DIR = PROJECT_DIR / "libparsec/crates/pki/tests/certificates"
```

libparsec/crates/platform_pki/src/test_fixture.rs
```diff
@@ -97,9 +97,22 @@ mod platform_implementation {
     pub fn test_pki() -> TestPKI {
         let test_pki_dir = std::env::var_os("TEST_PKI_DIR")
```
touilleMan, line 98: I don't think this environ variable is still useful

libparsec/crates/platform_pki/test-pki/Cert/alice.crt
touilleMan, line 1: move the certificates in `libparsec_platform_pki`

libparsec/crates/platform_pki/test-pki/Cert/alice.crt
```diff
@@ -0,0 +1,20 @@
...
+MIIDRDCCAiygAwIBAgIUUdAsXBYueLwRuzFgE/l9Uy3BylMwDQYJKoZIhvcNAQEL
```
touilleMan: Also use 1024bits keys in the certificates (smaller and much faster than 2048bits)

PR #11887, opened 2025-12-09
[PKI] Deduplicate certificate storage.

### LINKED ISSUE #11871: [PKI] Memory : store cert content only in cert db
```git
-    submitter_der_x509_certificate: bytes = field(repr=False)
+    submitter_der_x509_fingerprint: bytes = field(repr=False)
``` 

and add test case in pki enrollment list

AureliaDolo:
The certificate was stored in MemoryEnrollment and in MemoryPkiCertificate.
Now MemoryEnrollment contains only the cert's fingerprint.

Fix #11871

### DIFF
--- a/server/parsec/components/memory/datamodel.py
+++ b/server/parsec/components/memory/datamodel.py
@@ -509,6 +509,9 @@ class MemoryOrganization:
         # in theory it should not have been stored if too long
         return (leaf, sorted_trustchain[1:])
 
+    async def get_cert(self, fingerprint: bytes) -> MemoryPkiCertificate | None:
+        return self.pki_certificates.get(fingerprint)
+
 
 @dataclass(slots=True)
 class MemorySequesterService:
@@ -792,7 +795,8 @@ class MemoryPkiCertificate:
 @dataclass(slots=True)
 class MemoryPkiEnrollment:
     enrollment_id: PKIEnrollmentID
-    submitter_der_x509_certificate: bytes = field(repr=False)
+    # references the cert stored in MemoryPkiCertificate
+    submitter_der_x509_fingerprint: bytes = field(repr=False)
 
     submit_payload_signature: bytes = field(repr=False)
     submit_payload_signature_algorithm: PkiSignatureAlgorithm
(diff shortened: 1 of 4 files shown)


server/parsec/components/memory/datamodel.py
```diff
@@ -505,6 +505,10 @@ async def get_trustchain(self, leaf_fingerprint: bytes) -> list[MemoryPkiCertifi
         # in theory it should not have been stored if too long
         return sorted_trustchain
 
+    async def get_cert(self, fingerprint: bytes) -> MemoryPkiCertificate | None:
+        print("####get####", fingerprint[:5])
```
AureliaDolo: todo remove

server/parsec/components/memory/pki.py
```diff
@@ -102,53 +102,63 @@ async def submit(
...
-                        assert enrollment.submitter_accepted_user_id is not None
-                        assert enrollment.submitter_accepted_device_id is not None
-                        user = org.users[enrollment.submitter_accepted_user_id]
-                        if not user.is_revoked:
-                            return PkiEnrollmentSubmitBadOutcome.X509_CERTIFICATE_ALREADY_ENROLLED
-
-                # There is no need looking for older enrollments given the
-                # last one represent the current state of this x509 certificate.
-                break
+            match await org.get_cert(submitter_der_x509_certificate.fingerprint_sha256):
```
AureliaDolo, line 105: Github being github here. Most of that change is only a tab bump

As we already have the cert's fingerprint, we are not required to iter on all enrollments anymore, only on those for which the fingerprint matches, and only after checking that the cert was already here.

server/tests/api_v5/authenticated/test_pki_enrollment_list.py
```diff
@@ -213,4 +217,4 @@ async def do():
     await authenticated_http_common_errors_tester(do)
 
 
-# TODO test when leaf fingerprint not in db ->  #11871
```
AureliaDolo, line 216: Not needed because it implies corrupted DB and we don't test for that

Issue #11388, opened 2025-10-20
Better handling of revoked client


Max-7:
Right now, when connecting to a device of a user that is revoked, libparsec will sent a `RevokedSelfUser` to the GUI.
With that event, the GUI:
- displays a message
- shows the user as being offline
- creates a notification

<img width="872" height="478" alt="Image" src="https://github.com/user-attachments/assets/fc819450-2b83-4944-a998-800b4a9f2bd3" />
<img width="940" height="321" alt="Image" src="https://github.com/user-attachments/assets/f262e745-db62-479a-a09c-13547d59860b" />

Workspaces and files can still be accessed if they were downloaded previously (if they are present in the cache).

There are a few problems with this:
- the GUI can buffer the event for a bit (at most 6s I think) because of some timing issues that can happen (libparsec can send the event before the GUI loads the connected components, so we add a delay)
- the user doesn't realize they're revoked
- it can give the impression that the data are still accessible, but in reality we're only accessing what's in the cache
- the GUI doesn't keep internal states, meaning it can show a message when it receives the event, but it can keep a boolean of whether the user is revoked as we navigate

Few ideas of things that can be done:
- [ ] 1. clear the cache, so that users, workspaces and files are not present
- [ ] 2. logout automatically when receiving that event (maybe a bit violent)
- [ ] 3. disable navigation and all pages (with a grey overlay) if we're revoked
- [ ] 4. change the appearance of the GUI in a meaningful way

mmmarcos:
It's probably better to start with points 2 and 4 (above).

PR #11894, opened 2025-12-10
[MS] Better revoked user handling

Ironicbay:
Closes https://github.com/Scille/parsec-cloud/issues/11388

### DIFF
--- a/client/src/views/layouts/ConnectedLayout.vue
+++ b/client/src/views/layouts/ConnectedLayout.vue
@@ -11,7 +11,16 @@
 </template>
 
 <script lang="ts" setup>
-import { acceptTOS, getClientInfo, getTOS, listStartedClients, logout as parsecLogout } from '@/parsec';
+import {
+  acceptTOS,
+  archiveDevice,
+  AvailableDevice,
+  getClientInfo,
+  getTOS,
+  listAvailableDevices,
+  listStartedClients,
+  logout as parsecLogout,
+} from '@/parsec';
 import { getConnectionHandle, navigateTo, Routes } from '@/router';
 import { APP_VERSION } from '@/services/environment';
 import {
@@ -176,13 +185,6 @@ async function eventCallback(event: Events, data?: EventData): Promise<void> {
       );
       break;
     case Events.ClientRevoked:
-      injections.informationManager.present(
-        new Information({
-          message: 'notification.clientRevoked',
-          level: InformationLevel.Error,
-        }),
-        PresentationMode.Notification,
-      );
       await injections.informationManager.present(
         new Information({
           message: 'globalErrors.clientRevoked',
@@ -190,6 +192,15 @@ async function eventCallback(event: Events, data?: EventData): Promise<void> {
         }),
         PresentationMode.Modal,
       );
+      const clientInfo = await getClientInfo();
+      let currentDevices: Array<AvailableDevice> = [];
+      if (clientInfo.ok) {
+        currentDevices = (await listAvailableDevices(false)).filter((device) => device.userId === clientInfo.value.userId);
+      }
+      for (const device of currentDevices) {
+        await archiveDevice(device);
+      }
+      await logout();
       break;
     case Events.ClientFrozen:
       await injections.informationManager.present(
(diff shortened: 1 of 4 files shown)


client/src/views/layouts/ConnectedLayout.vue
```diff
@@ -190,6 +198,11 @@ async function eventCallback(event: Events, data?: EventData): Promise<void> {
         }),
         PresentationMode.Modal,
       );
+      const currentDevice = await getCurrentAvailableDevice();
```
Max-7: You can remove the notification part since the user will never see it.

client/tests/e2e/specs/client_events.spec.ts
```diff
@@ -16,15 +17,45 @@ msTest('Revoked event', async ({ usersPage }) => {
...
+    `You have been revoked from this organization and will be logged out.
+ If you have any issues, please contact an administrator.`,
+  );
+  await modal.locator('#next-button').click();
+  await expect(usersPage).toBeHomePage();
+  await expect(cards).toHaveCount(2);
+
+  // Login to Boby
+  await login(usersPage, 'Boby McBobFace');
+  await usersPage.waitForTimeout(3000);
```
Max-7: Instead of using `waitForTimeout` here, it's better to do `await expect(modal).toBeVisible({ timeout: 1000 })` instead. The difference is `waitForTimeout` will always wait 3000ms, while with the timeout it will return as soon as the modal is visible.

client/tests/e2e/specs/client_events.spec.ts
```diff
@@ -16,15 +17,45 @@ msTest('Revoked event', async ({ usersPage }) => {
...
   const modal = usersPage.locator('.information-modal');
   await expect(modal.locator('.ms-error').nth(0)).toBeVisible();
   await expect(modal.locator('.container-textinfo__text')).toHaveText(
-    `You have been revoked from this organization. You can still see some of the data you
- had access to, but most features of Parsec will not work.`,
+    `You have been revoked from this organization and will be logged out.
+ If you have any issues, please contact an administrator.`,
+  );
+  await modal.locator('#next-button').click();
+  await expect(usersPage).toBeHomePage();
```
Max-7, line 52: Make sure the modal closes.

client/tests/e2e/specs/client_events.spec.ts
```diff
@@ -16,15 +17,45 @@ msTest('Revoked event', async ({ usersPage }) => {
...
+  await login(usersPage, 'Boby McBobFace');
+  await usersPage.waitForTimeout(3000);
+  await expect(modal.locator('.ms-error').nth(0)).toBeVisible();
+  await expect(modal.locator('.container-textinfo__text')).toHaveText(
+    `You have been revoked from this organization and will be logged out.
+ If you have any issues, please contact an administrator.`,
   );
   await modal.locator('#next-button').click();
-  await expect(usersPage).toHaveNotification('You have been revoked from this organization.');
+  await expect(usersPage).toBeHomePage();
```
Max-7, line 65: Make sure the modal closes.

client/tests/e2e/specs/client_events.spec.ts
```diff
@@ -16,15 +17,45 @@ msTest('Revoked event', async ({ usersPage }) => {
...
+  await item2.click({ button: 'right' });
+  await menu.getByRole('listitem').nth(1).click();
+  await answerQuestion(usersPage, true, {
+    expectedTitleText: 'Revoke this user?',
+    expectedQuestionText:
+      'This will revoke Malloryy McMalloryFace, preventing them from accessing this organization. Are you sure you want to proceed?',
+    expectedPositiveText: 'Revoke',
+    expectedNegativeText: 'Cancel',
+  });
+  await expect(usersPage).toShowToast('Malloryy McMalloryFace has been revoked. They can no longer access this organization.', 'Success');
```
Max-7, line 35: Make sure both users are marked as revoked while we're at it, doesn't cost much.

client/tests/e2e/specs/client_events.spec.ts
```diff
@@ -3,6 +3,7 @@
 import { answerQuestion, expect, login, logout, msTest } from '@tests/e2e/helpers';
 
 msTest('Revoked event', async ({ usersPage }) => {
```
Max-7, line 5: I think the default test timeout is going to be a bit short, better to increase it to 45s or a minute.

PR #11895, opened 2025-12-10
feat(binding): Expose `remove_device_data`

### LINKED ISSUE #11892: Expose binding `remove_device_data`
`l/c/p/s/w/cleanup.rs`

FirelightFlagboy:
Closes #11892

### DIFF
--- a/bindings/generator/api/common.py
+++ b/bindings/generator/api/common.py
@@ -176,6 +176,10 @@ class Handle(U32BasedType):
     pass
 
 
+class AnyhowError(StrBasedType):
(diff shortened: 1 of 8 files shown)


bindings/generator/api/common.py
```diff
@@ -176,6 +176,10 @@ class Handle(U32BasedType):
     pass
 
 
+class AnyhowError(StrBasedType):
```
AureliaDolo, line 179: I'm not sure about that: That means that the front will kinda only have a string to work from instead of a real error. 

But that would require to rework the error returned by remove_device_data.

Anyhow if a shark is on board with that I'm ok

PR #11917, opened 2025-12-12
refactor(storage): Add "proper" error for `remove_device_data`

FirelightFlagboy:


FirelightFlagboy:
> It moves the issue one step deeper, maybe it would be worth asking a shark if it's enough details
> 
> (fun fact: the error in delete_database was already a js error [docs.rs/indexed-db/0.4.2/src/indexed_db/factory.rs.html#78-88](https://docs.rs/indexed-db/0.4.2/src/indexed_db/factory.rs.html#78-88))

Hard to have more information than "I've fail to remove the data" :shrug:

FirelightFlagboy:
Good enought for @Max-7

PR #11921, opened 2025-12-12
feat(pki): Client provide intermediate certs during `pki_accept`

### LINKED ISSUE #11671: Client should provide the list of intermediate certificates required to verify pki trust during `pki_accept`
Similar to #11670

FirelightFlagboy:
Closes #11671


FirelightFlagboy:
> Are they some tests to update with intermediate certs?

Nop, because that would mean have a CI configured with a PKI

PR #11941, opened 2025-12-16
[PKI] Add missing errors to protocol

AureliaDolo:


server/parsec/components/memory/pki.py
FirelightFlagboy, line 1: **issue:** Should not be included in the base code

You can either:

- Create a pytest fixture with a context manager
- Manipulate the component during the test
- monkeypatch

server/parsec/components/memory/datamodel.py
```diff
@@ -490,10 +490,12 @@ async def get_trustchain(
...
-        try:
-            leaf = self.pki_certificates[leaf_fingerprint]
-        except KeyError:
-            return None
+        match await self.get_cert(leaf_fingerprint):
+            case None:
+                return None
+            case leaf:
+                pass
+
```
AureliaDolo, line 498: Slight refactor to be able to mock get cert to mimic DB corruption in the CertificateNotFound error test.

PR #11943, opened 2025-12-17
[MS] Moved cryptpad into an iframe

Max-7:


client/src/services/cryptpad.ts
```diff
@@ -3,206 +3,120 @@
...
+    autosaveInterval: number;
+    mode: CryptpadCommAPI.OpenModes;
+    locale: string;
   }
+}
 
-  async open(config: CryptpadConfig): Promise<void> {
-    await this.init();
+export import CryptpadEditors = CryptpadCommAPI.Editors;
+export import CryptpadOpenMode = CryptpadCommAPI.OpenModes;
```
NicoTuxx: choose between both plural or singular for local variables.

client/src/services/cryptpad.ts
```diff
@@ -213,5 +127,129 @@ export function isFileEditable(name: string): boolean {
   if (!fileContentType) {
     return false;
   }
-  return isEnabledCryptpadDocumentType(fileContentType.type);
+  return isCryptpadEnabledForDocumentType(fileContentType.type);
+}
+
+function cryptpadLog(level: 'debug' | 'warn' | 'error' | 'info', message: string): void {
+  window.electronAPI.log(level, `[Cryptpad Ev List] ${message}`);
```
NicoTuxx: I'm totally for specific logs filter, but `Ev List` is not really clear, should be more specific

client/src/services/cryptpad.ts
```diff
@@ -213,5 +127,129 @@ export function isFileEditable(name: string): boolean {
...
+  function sendMessageToFrame(command: CryptpadCommAPI.Commands, data?: any): void {
+    if (!frame.contentWindow) {
+      throw new CryptpadError(CryptpadErrorCodes.FrameNotLoaded);
+    }
+    frame.contentWindow.postMessage({ command: command, ...data }, CRYPTPAD_SERVER);
+  }
+
+  try {
+    await new Promise<void>((resolve, reject) => {
+      const abortCtrlr = new AbortController();
```
NicoTuxx:
```suggestion
      const abortController = new AbortController();
```
It's not to be picky, but you're shortenning a lot of variables, it makes code less readable & easy to understand.
Readability should always prime on code shorteness, chars are free.

client/src/views/files/handler/editor/FileEditor.vue
```diff
@@ -73,14 +64,13 @@ import { checkmarkCircle } from 'ionicons/icons';
...
-const cryptpadInstance = ref<Cryptpad | null>(null);
-const fileUrl = ref<string | null>(null);
+const editorFrame = useTemplateRef<HTMLIFrameElement>('editorFrame');
+const documentType = ref<CryptpadEditors>(CryptpadEditors.Unsupported);
 const error = ref('');
 const eventDistributor: EventDistributor = inject(EventDistributorKey)!;
 let eventCbId: null | string = null;
-const fileLoaded = ref(false);
+const loadFinished = ref(false);
+let abortCtrl: AbortController | undefined = undefined;
```
NicoTuxx:
```suggestion
let abortController: AbortController | undefined = undefined;
```

NicoTuxx (changes requested):
First review comments, but I'm still testing it:

I don't know if we should do something about it, but with the new cryptpad frame implementation, we loose the ability to use any cryptpad server compared to before.
Maybe we should keep a way to use non-parsec custom server or show a specific error message for this case?

NicoTuxx:
FYI, this is what is currently happening when the server is not our custom one:

https://github.com/user-attachments/assets/db131eb9-b566-4a8d-92f7-4ac2c35560a1

Normally it was working even with a non-custom server, it may be something we will need to keep for commercial purpose, this matter should be discuss / address.
At least it should use our custom modal or page, but not this Cryptpad error animation obviously.

NicoTuxx:
Second problem: with an unknown or non-responding server, the implentation is degrading the current user experience level.
We should not see the gray step.
https://github.com/user-attachments/assets/c0c1a9da-f0b8-474a-8818-eceaa1a7da9e

Max-7:
> Second problem: with an unknown or non-responding server, the implentation is degrading the current user experience level. We should not see the gray step. https://github.com/user-attachments/assets/c0c1a9da-f0b8-474a-8818-eceaa1a7da9e

Yeah I noticed this case and then forgot about it, my bad. I added a spinner while the frame itself is loaded.

PR #11948, opened 2025-12-17
feat(server): Check submit payload with trusted anchor

### LINKED ISSUE #11939: Use `load_submit_payload` for `pki_submit` on server side

FirelightFlagboy:
Closes #11939


server/parsec/_parsec_pyi/pki.pyi
```diff
@@ -80,6 +80,8 @@ class X509CertificateInformation:
 class TrustAnchor:
     @classmethod
     def try_from_pem(cls, raw_pem: bytes) -> TrustAnchor: ...
+    @classmethod
```
touilleMan, line 83: `try_from_...` is not a Pythonic name, it should just be `from_pem` since Python doesn't have the concept of cannot-fail-function, so the default assumption is that the function will raise a `ValueError` if the input can be invalid

I the codebase, we already have types such as `UserProfile::from_str` or `UserID::from_hex` that follow this naming convention

server/tests/common/pki.py
```diff
@@ -22,49 +41,66 @@ class TestPki:
...
+            if cert.certificate.der == der:
+                return cert
+        for cert in self.root.values():
+            if cert.certificate.der == der:
+                return cert
+        raise KeyError
+
 
 @dataclass
 class Certificate:
```
touilleMan: This class is going to end up in the global scope, so it should be clear what this is about
```suggestion
class X509Certificate:
```
or 
```suggestion
class PKICertificate:
```

server/tests/common/pki.py
```diff
@@ -22,49 +41,66 @@ class TestPki:
...
+
+@dataclass
+class Key:
+    der: bytes
+    path: Path
+    obj: PrivateKeyTypes
+
+
 @dataclass
 class PartialCertificate:
```
touilleMan: Same as above

server/tests/common/pki.py
```diff
@@ -141,3 +180,113 @@ async def read_file(file: Path) -> tuple[Path, bytes]:
...
+        submitter_human_handle=human_handle,
+        submit_payload_signature=submit_payload_signature,
+        submit_payload_signature_algorithm=sign_algo,
+        submit_payload=submit_payload,
+        submitted_on=now,
+    )
+
+
+@pytest.fixture
+async def existing_enrollment(
```
touilleMan: same as above

server/tests/common/pki.py
```diff
@@ -22,49 +41,66 @@ class TestPki:
...
-            der_key=partial.der_key,
-            key_path=partial.key_path,
+            certificate=partial.certificate,
+            key=partial.key,
             cert_info=cert_info,
         )
 
 
+@dataclass
+class Cert:
```
touilleMan: same as above

server/tests/common/pki.py
FirelightFlagboy: done

server/tests/common/pki.py
FirelightFlagboy: done

server/tests/common/pki.py
FirelightFlagboy: done

server/parsec/_parsec_pyi/pki.pyi
FirelightFlagboy: done

AureliaDolo (approved):

PR #11950, opened 2025-12-17
RFC 1025 TOTP protected local device

touilleMan:


docs/rfcs/1025-totp-protected-local-device.md
```diff
@@ -0,0 +1,393 @@
...
+
+This RFC introduce the use of [Multi-Factor Authentication](https://en.wikipedia.org/wiki/Multi-factor_authentication)
+(MFA) in the form of [Time-based One-Time Password](https://en.wikipedia.org/wiki/Time-based_one-time_password)(TOTP)
+as an additional layer of protection on the local device.
+
+Basically:
+
+1. Alice starts her Parsec client
+2. Alice selects the device she wants to use
+3. The Parsec client then prompts alice Alice to enter a TOTP code.
```
AureliaDolo:
```suggestion
3. The Parsec client then prompts Alice to enter a TOTP code.
```

docs/rfcs/1025-totp-protected-local-device.md
```diff
@@ -0,0 +1,393 @@
...
+Most of those solution are local-only, which is both a blessing and a curse:
+
+- pro: Offline login is possible
+- con: Offline attack is possible (typically when the user's machine gets stolen)
+
+The use of a TOTP aims at involving the Parsec server in the login operation,
+this way no secret can be obtained from the stolen machine (unless, of course,
+the attacker also has access to the user's TOTP secret).
+
+This is not a one-size-fits-all solution since its removes the possibility to
```
AureliaDolo:
```suggestion
This is not a one-size-fits-all solution since it removes the possibility to
```

docs/rfcs/1025-totp-protected-local-device.md
```diff
@@ -0,0 +1,393 @@
...
+4. The server checks the TOTP code
+
+The key point here is there is two separate challenge-response occurring:
+first using login/password, and only then a second one involving the TOTP.
+
+However this is not something we can do with Parsec: we want to protect the device keys
+file with the TOTP, however it is the device signing key (which is part of the device
+keys file) that is used to authenticate to the Parsec server.
+
+So instead we must first to the TOTP check, and have its result used as a secret
```
AureliaDolo:
```suggestion
So instead we must first do the TOTP check, and have its result used as a secret
```

docs/rfcs/1025-totp-protected-local-device.md
```diff
@@ -0,0 +1,393 @@
...
+
+- TOTP application (e.g. Google Authenticator)
+- Email
+- SMS
+
+This RFC consider the TOTP code are only going to be obtained from a TOTP application.
+
+However later development on the server side can be done to add additional ways.
+This would typically require a new anonymous API command (`totp_info` ?) so that
+the client.
```
AureliaDolo:
```suggestion
the client can request the totp configuration.
```
```suggestion
the client can request that a TOTP code is sent to their configured option (mail or SMS).
```
I'm not sure what you meant to say though ^^ Anyhow it's a discussion for another RFC.

docs/rfcs/1025-totp-protected-local-device.md
```diff
@@ -0,0 +1,393 @@
...
+
+### 2.4 - How is used the secret obtained from the TOTP challenge ?
+
+Once the TOTP challenge succeeded, the Parsec server send to the client a secret.
```
AureliaDolo:
```suggestion
Once the TOTP challenge succeeded, the Parsec server sends a secret to the client .
```

docs/rfcs/1025-totp-protected-local-device.md
```diff
@@ -0,0 +1,393 @@
...
+This secret has then to be used in conjuction with the secret obtained from the primary
+protection (i.e. OS keyring) in order to decrypt the device keys file.
+
+This is done by considering both secret as separate keys and hence doing two decryption operations:
```
AureliaDolo:
```suggestion
This is done by considering both secrets as separate keys and hence doing two decryption operations:
```

docs/rfcs/1025-totp-protected-local-device.md
```diff
@@ -0,0 +1,393 @@
...
+                    // must be called to get it.
+                    // This is to ensure the opaque key is not used before the end-user has
+                    // sucessfully configured his authenticator app with the TOTP secret.
+                    {
+                        "name": "opaque_key_id",
+                        "type": "UUID"
+                    },
+                    {
+                        // TOTP secret to configure in the user's authenticator app.
+                        // Make sure you do that
```
AureliaDolo: Is something missing here ?

docs/rfcs/1025-totp-protected-local-device.md
```diff
@@ -0,0 +1,393 @@
...
+
+> [!TODO]:
+>
+> - Authentication API, complex to use during bootstrap organization since it is
+>   currently done in a single `bootstrap_organization()` call (hence cannot
+>   pass a save strategy with a TOTP configured).
+> - Also complex during claim operation. Need to add a new
+>   `claimer_XXX_finalize_totp_upload_opaque_key` to be called before `claimer_XXX_finalize_save_local_device`
+> - Possible solution: don't both with TOTP during bootstrap/claim. Instead pass
+>   in the server configuration if TOTP protection is required, and display a popup
```
AureliaDolo: At which level is the TOTP configuration set ? By organization or by server ? Can it be allowed but optional ?

docs/rfcs/1025-totp-protected-local-device.md
```diff
@@ -0,0 +1,393 @@
...
+                        "type": "SecretKey"
+                    }
+                ]
+            },
+            {
+                // Also returned if the opaque key ID is invalid
+                "status": "invalid_one_time_password"
+            },
+            {
+                // Too many attempts has been done for this opaque key ID
```
AureliaDolo:
```suggestion
                // Too many attempts have been done for this opaque key ID
```

Can a device end up being blocked by too many attempts ?

docs/rfcs/1025-totp-protected-local-device.md
```diff
@@ -0,0 +1,393 @@
...
+}
+```
+
+## 5 - Libparsec API
+
+### 5.1 - New TOTP API
+
+Upload operation is exposed to the GUI through the libparsec API.
+
+```rust
```
AureliaDolo, line 665: If possible it would be nice to also include or replace this by the js version as that part should be reviewed by sharks.

docs/rfcs/1025-totp-protected-local-device.md
```diff
@@ -0,0 +1,393 @@
...
+
+pub struct AvailableDevice {
+    pub key_file_path: PathBuf,
+    // ...
+    pub ty: AvailableDeviceType,
+    pub totp: Option<AvailableDeviceTOTPConfig>,
+}
+```
+
+### 5.3 - Changes in `Device(Save|Access)Strategy`
```
AureliaDolo: If the TOTP config will be added to each access variants, wouldn't be it easier/clearer to pass it alongside the access instead ?

docs/rfcs/1025-totp-protected-local-device.md
```diff
@@ -0,0 +1,393 @@
...
+  obtaining the final ciphertext.
+
+Put it another way, the local device decryption is done this way:
+
+```rust
+let key_from_totp_challenge: SecretKey = ...
+let key_from_primary_protection: SecretKey = ...
+let cleartext = key_from_totp_challenge.decrypt(
+    key_from_primary_protection.decrypt(ciphertext)
+);
```
FirelightFlagboy, line 128: **question:** Not using a shamir system here?

Would allow to effectively use multiple auth system with threshold

docs/rfcs/1025-totp-protected-local-device.md
touilleMan: :+1:

docs/rfcs/1025-totp-protected-local-device.md
touilleMan: :+1:

docs/rfcs/1025-totp-protected-local-device.md
touilleMan: :+1: done

docs/rfcs/1025-totp-protected-local-device.md
touilleMan: :+1:

docs/rfcs/1025-totp-protected-local-device.md
touilleMan: removed

docs/rfcs/1025-totp-protected-local-device.md
touilleMan: :+1:

docs/rfcs/1025-totp-protected-local-device.md
touilleMan: An important point here is to be able to validate both protection apart from each other, this useful so that we can inform the end-user of the issue:
- his TOTP code was invalid
- his password was invalid
- an unexpected error occured: the secret obtain from the server (with the TOTP challenge) couldn't be used for decryption

On the other hand, if the ciphertext is encrypted by a single key that itself is derived from multiple secrets, we cannot know which secret was invalid if the decryption fails.

> Would allow to effectively use multiple auth system with threshold

YAGNI, this add plenty of complexity and I don't see a real usecase for it now.
The need here is to prevent the client from decrypting the ciphertext without first achieving a challenge with the server.
Hence if we want to have multiple MFA systems that have to be used in conjunction (for instance the user has to provide a TOTP code provided by email, then another one from SMS), then it is something that is implemented has additional APIs exposed by the server, with the final API returning the secret to be used for decryption. So even in this case the secret is still a simple decryption key.

FirelightFlagboy: > I don't see a real usecase for it now.

I mean, we are talking about MFA with shamir we could extend the number of share when adding a new auth method and we can easily configure how many method to use (some org, ask for 3 methods :cold_sweat:)

docs/rfcs/1025-totp-protected-local-device.md
touilleMan: It would simplify the structure definition, but at the cost of additional complexity in it usage since all the functions taking a `Device(Save|Access)Strategy` would need to be modified to take an additional parameter (compared to the current approach where only the top caller and lower platform_device_loader function has to deal with the TOTP config)

On top of that, the TOTP config *is* related to the device acces/save strategy, so it makes sens to have it in this structure.
I guess we could modify the `Device(Save|Access)Strategy` to make it a structure that would have two fields `primary` (i.e. the enum) and `totp` (the totp config). But again I fear this would cause too much changes compared to just duplicating code in each variant type (however this approach can be tested during implementation to see if it is worth it :+1: )

docs/rfcs/1025-totp-protected-local-device.md
touilleMan: I'd rather stay with the Rust version since this code will be mainly implemented in Rust.
Beside, the change in naming are minimal (mainly switch from snake to camel case), so I don't think the shark team will struggle with this ;-p

docs/rfcs/1025-totp-protected-local-device.md
```diff
@@ -0,0 +1,528 @@
...
+> it.
+> This is a bit error prone for the end user (since he may assume the secret in
+> his authenticator app can be removed once he has disabled TOTP protection in
+> all his devices), however this is no big deal:
+>
+> - The worst case is the end user can no longer protect his device with TOTP,
+>   he is not locked out of it.
+> - The outcome is similar to what would occur if the end user had lose his
+>   authenticator app: in both case the server admin has to intervene to reset
+>   the TOTP setup.
```
mmmarcos, line 179: This is covered in the list of changes below
```
- CLI server: reset TOTP setup for a given user
```

I don't quite understand if this is impossible for some reason (there is no way for the user to reset TOTP themselves with the current design), or if it's just a workaround for the moment (contact the server admin to reset it).

I mean, why is not possible to have am authenticated command (so the user already unlocked its device with password + totp) that resets its own TOTP configuration?

Also, what happens to protected devices when the server admin resets totp configuration? can the user still login even if the secret used to protect the device has changed?

docs/rfcs/1025-totp-protected-local-device.md
```diff
@@ -0,0 +1,528 @@
...
+
+> [!NOTE]
+> In a traditional application, TOTP can be disabled and later re-enabled. In
+> which case a new TOTP secret is provided to the user.
+>
+> Here this is not possible: once the TOTP configured we cannot know if some
+> local device is protected with it.
+> So the TOTP configuration is setup once, and stays the same even if the user
+> remove TOTP protection from all his device, then later on choose to re-add
+> it.
```
mmmarcos, line 170: How can a user _remove_ TOTP protection from a device once it has been setup? This can be done locally?

docs/rfcs/1025-totp-protected-local-device.md
```diff
@@ -0,0 +1,528 @@
...
+
+- New authenticated API: `totp_setup_get_secret`
+- New authenticated API: `totp_create_opaque_key`
+- New anonymous API:  `totp_fetch_opaque_key`
+- CLI server: reset TOTP setup for a given user
+- Introduce a new `totp` field in each device file schema (`DeviceFilePassword`, `DeviceFileKeyring`, etc.)
+- Libparsec: new function `client_totp_status`
+- Liparsec: new function `client_totp_try_challenge`
+- Liparsec: new function `client_totp_create_opaque_key`
+- Liparsec: new function `client_totp_fetch_opaque_key`
```
mmmarcos:
```suggestion
- Libparsec: new function `client_totp_try_challenge`
- Libparsec: new function `client_totp_create_opaque_key`
- Libparsec: new function `client_totp_fetch_opaque_key`
```

docs/rfcs/1025-totp-protected-local-device.md
```diff
@@ -0,0 +1,528 @@
...
+- GUI: check TOTP status and setup it (using `client_totp_status` and `client_totp_try_challenge`)
+- GUI: add "enable TOTP" option in save device dialogue (using `client_totp_create_opaque_key`)
+- GUI: add "TOTP challeng" in load device dialogue (using `client_totp_fetch_opaque_key`)
```
mmmarcos:
```suggestion
- GUI: add "TOTP challenge" in load device dialogue (using `client_totp_fetch_opaque_key`)
```

docs/rfcs/1025-totp-protected-local-device.md
```diff
@@ -0,0 +1,528 @@
...
+(MFA) in the form of [Time-based One-Time Password](https://en.wikipedia.org/wiki/Time-based_one-time_password)(TOTP)
+as an additional layer of protection on the local device.
+
+Basically:
+
+1. Alice starts her Parsec client
+2. Alice selects the device she wants to use
+3. The Parsec client then prompts Alice to enter a TOTP code.
+   The TOTP code is then is transmitted to the server.
+4. The server checks the TOTP code and returns a second decryption key.
```
mmmarcos:
```suggestion
4. The server checks the TOTP code and returns the _secondary_ decryption key.
```

docs/rfcs/1025-totp-protected-local-device.md
```diff
@@ -0,0 +1,528 @@
...
+5. The Parsec client prompts Alice to unlock the device according to its main
+   protection (i.e. password). By doing this the client obtain a first decryption key.
```
mmmarcos:
```suggestion
   protection (e.g. password). By doing this the client obtain the _primary_ decryption key.
```

docs/rfcs/1025-totp-protected-local-device.md
```diff
@@ -0,0 +1,528 @@
...
+6. The Parsec client use both first and second decryption keys to decrypt the
```
mmmarcos:
```suggestion
6. The Parsec client use both _primary_ and _secondary_ decryption keys to decrypt the
```

docs/rfcs/1025-totp-protected-local-device.md
```diff
@@ -0,0 +1,528 @@
...
+successful, then `totp_setup_get_secret` always returns `already_setup` (i.e. the
+client can no longer obtain the TOTP secret since he is supposed to have it stored
+in his authenticator app).
+
+> [!NOTE]
+> In a traditional application, TOTP can be disabled and later re-enabled. In
+> which case a new TOTP secret is provided to the user.
+>
+> Here this is not possible: once the TOTP configured we cannot know if some
+> local device is protected with it.
```
mmmarcos, line 167: > we cannot know if some local device is protected with it.

You mean is not possible to know if the device is protected with this particular TOTP or with another? But I guess we can at least know if it is protected with TOTP or not?

docs/rfcs/1025-totp-protected-local-device.md
```diff
@@ -0,0 +1,528 @@
...
+- New authenticated API: `totp_create_opaque_key`
+- New anonymous API:  `totp_fetch_opaque_key`
+- CLI server: reset TOTP setup for a given user
+- Introduce a new `totp` field in each device file schema (`DeviceFilePassword`, `DeviceFileKeyring`, etc.)
+- Libparsec: new function `client_totp_status`
+- Liparsec: new function `client_totp_try_challenge`
+- Liparsec: new function `client_totp_create_opaque_key`
+- Liparsec: new function `client_totp_fetch_opaque_key`
+- GUI: check TOTP status and setup it (using `client_totp_status` and `client_totp_try_challenge`)
+- GUI: add "enable TOTP" option in save device dialogue (using `client_totp_create_opaque_key`)
```
mmmarcos: We should probably use the term "MFA" (or "2FA") instead of "TOTP" in the GUI and user-facing interfaces.
```suggestion
- GUI: add "enable MFA" option in save device dialogue (using `client_totp_create_opaque_key`)
```

docs/rfcs/1025-totp-protected-local-device.md
```diff
@@ -0,0 +1,528 @@
...
+### 2.5 - Per user TOTP config
+
+Since TOTP is used here to protect a local device, the simpler approach would be to
+have a different TOTP configuration for each device.
+However this means the end user would have to store in his authenticator app one
+TOTP secret per device, which is cumbersome and an error prone (the label for a
+TOTP secret is supposed to be something short, here it would end up being something
+like `Parsec - <organization_id> - <human_handle_email> - <device_label>`).
+
+So instead the TOTP should be configured on a per-user base:
```
Max-7: `Per user` or `per user per organization` or `per user per server` or `per computer`? Can we have only one entry "Parsec" on the app for all devices, whatever the org, the server, and web or laptop or desktop?

docs/rfcs/1025-totp-protected-local-device.md
```diff
@@ -0,0 +1,528 @@
...
+### 2.1 Where TOTP is to be used
+
+Currently a local device can be protected by multiple solutions, for instance:
+
+- Password
+- OS keyring
+- PKI (i.e. smartcard)
+- OpenBao (i.e. SSO-based authentication)
+
+Most of those solution are local-only, which is both a blessing and a curse:
```
AureliaDolo:
```suggestion
Most of those solutions are local-only, which is both a blessing and a curse:
```

docs/rfcs/1025-totp-protected-local-device.md
```diff
@@ -0,0 +1,528 @@
...
+
+- pro: Offline login is possible
+- con: Offline attack is possible (typically when the user's machine gets stolen)
+
+The use of a TOTP aims at involving the Parsec server in the login operation,
+this way no secret can be obtained from the stolen machine (unless, of course,
+the attacker also has access to the user's TOTP secret).
+
+This is not a one-size-fits-all solution since its removes the possibility to
+login while offline.
```
AureliaDolo:
```suggestion
This is not a one-size-fits-all solution since it removes the possibility to
access locally stored data while offline.
```

I find the phrase "offline login" a bit confusing :/

docs/rfcs/1025-totp-protected-local-device.md
```diff
@@ -0,0 +1,528 @@
...
+
+However since the TOTP server communication is only required during login (i.e. to
+obtain the device keys file's decryption key), offline operation are still
+possible after this step.
+In a nutshell, with TOTP Parsec can still be used fine in a poor connection
+configuration (e.g. traveling in a train), but cannot be used when no connection
+is possible for extended periods of time (e.g. in a submarine ^^).
```
AureliaDolo, line 58: If once connected to parsec, someone manages to never log out, they could be offline for an extended period of time ? There is no timeout ?

docs/rfcs/1025-totp-protected-local-device.md
```diff
@@ -0,0 +1,528 @@
...
+
+### 2.2 - Differences with traditional TOTP-based authentication
+
+Typical TOTP-based authentication goes a as follow:
```
AureliaDolo:
```suggestion
Typical TOTP-based authentication goes as follow:
```

docs/rfcs/1025-totp-protected-local-device.md
```diff
@@ -0,0 +1,528 @@
...
+
+### 2.3 - Different TOTP usages
+
+TOTP code can be obtained from different sources, typically:
+
+- TOTP application (e.g. Google Authenticator)
+- Email
+- SMS
+
+This RFC consider the TOTP code are only going to be obtained from a TOTP application.
```
AureliaDolo:
```suggestion
This RFC consider the TOTP codes are only going to be obtained from a TOTP application.
```
```suggestion
This RFC consider the TOTP code is only going to be obtained from a TOTP application.
```

docs/rfcs/1025-totp-protected-local-device.md
```diff
@@ -0,0 +1,528 @@
...
+);
+let device = LocalDevice::load(cleartext);
+```
+
+> [!NOTE]
+> An alternative solution would be to derive the ciphertext decryption key from both secrets
+> (i.e. `KeyDerivation::from(<key_from_primary_protection>).derive_secret_key_from_uuid(<secret_from_totp_challenge>).decrypt(<ciphertext>)`).
+>
+> We choose against it since having two separate encryption operations means we can know
+> which operation has failed (typically useful to prompt the user his password is invalid).
```
AureliaDolo, line 137: From a users POV, could the following happen ?
- they enter an invalid TOTP
- they enter a valid password
- they are asked for the TOTP again
- they do not need to enter the password again OR they are asked for their password again

docs/rfcs/1025-totp-protected-local-device.md
AureliaDolo: For the whatever the server I don't think it would be possible (or we'd better implement the MFA for parsec account).

docs/rfcs/1025-totp-protected-local-device.md
```diff
@@ -0,0 +1,528 @@
...
+have a different TOTP configuration for each device.
+However this means the end user would have to store in his authenticator app one
+TOTP secret per device, which is cumbersome and an error prone (the label for a
+TOTP secret is supposed to be something short, here it would end up being something
+like `Parsec - <organization_id> - <human_handle_email> - <device_label>`).
+
+So instead the TOTP should be configured on a per-user base:
+
+1. A `totp_setup_get_secret` command should be issued to obtain the TOTP secret.
+   The secret is displayed to the end user that configure his authenticator app with it.
```
AureliaDolo:
```suggestion
   The secret is displayed to the end user that configures his authenticator app with it.
```

docs/rfcs/1025-totp-protected-local-device.md
```diff
@@ -0,0 +1,528 @@
...
+   one-time-password as parameter.
+
+Whenever the TOTP challenge in `totp_create_opaque_key` or `totp_fetch_opaque_key` is
+successful, then `totp_setup_get_secret` always returns `already_setup` (i.e. the
+client can no longer obtain the TOTP secret since he is supposed to have it stored
+in his authenticator app).
+
+> [!NOTE]
+> In a traditional application, TOTP can be disabled and later re-enabled. In
+> which case a new TOTP secret is provided to the user.
```
AureliaDolo, line 164: Here to mimic that, we should provide a way to re create a new device with TOTP (and remove the old without TOTP)

docs/rfcs/1025-totp-protected-local-device.md
```diff
@@ -0,0 +1,528 @@
...
+> So the TOTP configuration is setup once, and stays the same even if the user
+> remove TOTP protection from all his device, then later on choose to re-add
+> it.
+> This is a bit error prone for the end user (since he may assume the secret in
+> his authenticator app can be removed once he has disabled TOTP protection in
+> all his devices), however this is no big deal:
+>
+> - The worst case is the end user can no longer protect his device with TOTP,
+>   he is not locked out of it.
+> - The outcome is similar to what would occur if the end user had lose his
```
AureliaDolo, line 177:
```suggestion
> - The outcome is similar to what would occur if the end user had lost his
```

docs/rfcs/1025-totp-protected-local-device.md
```diff
@@ -0,0 +1,528 @@
...
+            {
+                // No opaque key exists with this ID.
+                // This status is only returned if the one-time-password is valid,
+                // and can be used as a trick to ensure the end-user has correctly
+                // configured his authenticator app with the TOTP secret (in such
+                // case we send a dummy `opaque_key_id`).
+                "status": "opaque_key_not_found"
+            },
+            {
+                // Too many attempts have been done for this opaque key ID
```
AureliaDolo: What's the policy on what counts as too many attempts ?

docs/rfcs/1025-totp-protected-local-device.md
```diff
@@ -0,0 +1,528 @@
...
+> - However it is complex to use during bootstrap organization since it is
+>   currently done in a single `bootstrap_organization()` call (hence cannot
+>   pass a save strategy with a TOTP configured).
+> - Another possible solution: don't bother with TOTP during bootstrap/claim. Instead pass
+>   in the server configuration if TOTP protection is required, and display a popup
+>   on login if the current device is not protected accordingly.
+
+### 5.3 - Changes in `list_available_devices()`
+
+Listing available device should provide information about the need for MFA:
```
AureliaDolo: pedantic:
```suggestion
Listing available device should provide information about the need for TOTP:
```

docs/rfcs/1025-totp-protected-local-device.md
```diff
@@ -0,0 +1,528 @@
...
+    ...
+}
+```
+
+## 6 - List of changes
+
+- New authenticated API: `totp_setup_get_secret`
+- New authenticated API: `totp_create_opaque_key`
+- New anonymous API:  `totp_fetch_opaque_key`
+- CLI server: reset TOTP setup for a given user
```
AureliaDolo, line 749: By an organization administrator ?

docs/rfcs/1025-totp-protected-local-device.md
touilleMan: :+1: done.

docs/rfcs/1025-totp-protected-local-device.md
touilleMan: :+1: done.

docs/rfcs/1025-totp-protected-local-device.md
touilleMan: Again, this is something we can implement by having the server gives the opaque key to the client only when the client has provided codes from multiple MFA sources (instead of having the server returning a different secret for each MFA source, and recombined them into the final opaque key on the client side)

docs/rfcs/1025-totp-protected-local-device.md
touilleMan: Yes this is done locally in the same way the user can change his device protection from, for example, password to keyring.

docs/rfcs/1025-totp-protected-local-device.md
touilleMan: :+1: done.

docs/rfcs/1025-totp-protected-local-device.md
touilleMan: :+1: done.

docs/rfcs/1025-totp-protected-local-device.md
touilleMan: :+1: done.

docs/rfcs/1025-totp-protected-local-device.md
touilleMan: :+1: done.

docs/rfcs/1025-totp-protected-local-device.md
touilleMan: :+1: done.

docs/rfcs/1025-totp-protected-local-device.md
touilleMan: We can know a device keys file is protected by TOTP, however we cannot know how a given device is protected (we don't sent a notification to the server when the device keys file is modified, and doing this would prevent changing auth method while offline and would also be unreliable since the modification and the actual change wouldn't be atomic...).

Another example of the issue:
1. We have a device keys file protected by TOTP
2. We lose the machine it is on
3. Should we consider the TOTP is no longer in use (but then what happen if we recover the machine ?), or that the TOTP will never be not in use ?

docs/rfcs/1025-totp-protected-local-device.md
touilleMan: :+1: done.

docs/rfcs/1025-totp-protected-local-device.md
touilleMan: :+1: done.

docs/rfcs/1025-totp-protected-local-device.md
touilleMan: I've added a note to stress what "offline login" means ^^

docs/rfcs/1025-totp-protected-local-device.md
touilleMan: > If once connected to parsec, someone manages to never log out, they could be offline for an extended period of time ? There is no timeout ?

yes there is no timeout for the login. Note This would be a rather artificial measure since once logged in the machine has decrypted the device keys file and hence the thing the TOTP is supposed to protect is already owned by the client...

docs/rfcs/1025-totp-protected-local-device.md
touilleMan: :+1: done.

docs/rfcs/1025-totp-protected-local-device.md
touilleMan: :+1: done.

docs/rfcs/1025-totp-protected-local-device.md
touilleMan: :+1: done.

docs/rfcs/1025-totp-protected-local-device.md
touilleMan: Good point ! That was a much more complex that I've first think ^^

I've updated part `2.2 - Differences with traditional TOTP-based authentication` and `3.3 - Access the TOTP-protected opaque key to load a local device`

docs/rfcs/1025-totp-protected-local-device.md
touilleMan: :+1: done.

docs/rfcs/1025-totp-protected-local-device.md
touilleMan: This is too complex to do since local device keys file protection is something done purely in local (and even if we would notify the server about a change, we wouldn't be 100% sure since the device keys file modification and the server notification cannot be made atomic)

docs/rfcs/1025-totp-protected-local-device.md
touilleMan: :+1: done.

docs/rfcs/1025-totp-protected-local-device.md
touilleMan: The user must first pass the TOTP challenge, and only then the decryption is done (since the opaque key obtained from the TOTP is provided in the `DeviceAccessStrategy`)

From a GUI point of view I guess we want to have a first dialogue with the TOTP challenge, and only then a second dialogue with the password. This way the GUI remembers the opaque key from the TOTP challenge and if the password is invalid the user only has to retry entering another password wtithout having to do the TOTP challenge all over again

@max do you agree ?

docs/rfcs/1025-totp-protected-local-device.md
touilleMan: It could be interesting to allow an organization administrator to reset the TOTP configuration for other users, however this is not strictly needed.

On the other hand the server admin must be able to do that (since otherwise an organization with a single admin would be stuck if this admin needs to reset his TOTP setup...)

I've added documentation about this :+1:

docs/rfcs/1025-totp-protected-local-device.md
touilleMan: Per-user as in "user certificate", so each human has a single TOTP secret for each organization he is part of

I've made this more explicit in the RFC :+1: 

> Can we have only one entry "Parsec" on the app for all devices, whatever the org, the server, and web or laptop or desktop?

This is not possible to do that whatever the server since it would mean different server not trusting each other having to share and agree on a TOTP secret...

One TOTP per server will be possible with the Parsec account feature, but this is a feature for another time :trollface:

docs/rfcs/1025-totp-protected-local-device.md
touilleMan: We discussed this with @Max-7, it would be better to have this configured on a per-organization basis:
- Server-wide configuration is convenient for anything that occurs before the device decryption. However here we want to re-encrypt the device with another protection, so obviously this is not an issue
- Per-organization allows to use this feature on the SAAS server

docs/rfcs/1025-totp-protected-local-device.md
```diff
@@ -0,0 +1,702 @@
...
+- PKI (i.e. smartcard)
+- OpenBao (i.e. SSO-based authentication)
+
+Most of those solutions are local-only, which is both a blessing and a curse:
+
+- pro: Offline login is possible
+- con: Offline attack is possible (typically when the user's machine gets stolen)
+
+> [!NOTE]
+> "Offline login" sounds like an oxymore, here it refers to the fact we can
```
FirelightFlagboy: **typo:** 
```suggestion
> "Offline login" sounds like an oxymoron, here it refers to the fact we can
```

docs/rfcs/1025-totp-protected-local-device.md
```diff
@@ -0,0 +1,702 @@
...
+        },
+        "reps": [
+            {
+                "status": "ok",
+                "fields": [
+                    {
+                        // TOTP secret to configure in the user's authenticator app.
+                        "name": "totp_secret",
+                        "type": "String"
+                    }
```
FirelightFlagboy, line 241: **issue:** TOPT cannot be defined with only a String

It's defined with a reference time, an interval, a algorithm and a common secret

docs/rfcs/1025-totp-protected-local-device.md
```diff
@@ -0,0 +1,702 @@
...
+                "status": "already_setup"
+            }
+        ]
+    }
+]
+```
+
+> [!NOTE]
+> This at this point `totp_setup_get_secret` can be called to obtain the TOTP secret,
+> there is no point in having a protection against brute force attempts here.
```
FirelightFlagboy, line 301: **question:** Why do we talk about brute force protection when it's the authenticated API that is used to setup TOPT

docs/rfcs/1025-totp-protected-local-device.md
```diff
@@ -0,0 +1,702 @@
...
+        },
+        "reps": [
+            {
+                "status": "ok",
+                "fields": [
+                    {
+                        // TOTP secret to configure in the user's authenticator app.
+                        "name": "totp_secret",
+                        "type": "String"
+                    }
```
FirelightFlagboy, line 342: **issue:** Same issue regarding the configuration of TOTP

docs/rfcs/1025-totp-protected-local-device.md
```diff
@@ -0,0 +1,702 @@
...
+
+```rust
+enum ClientTOTPOperationError {
+    Offline,
+}
+
+enum TOTPStatus {
+    Stalled {
+        // Secret to display to the end-user so that he can configure his authenticator app
+        totp_secret: String,
```
FirelightFlagboy: **issue:** Same issue regarding configuring TOPT

docs/rfcs/1025-totp-protected-local-device.md
```diff
@@ -0,0 +1,702 @@
...
+    }
+    AlreadySetup,
+}
+
+async fn client_totp_status() -> Result<TOTPStatus, ClientTOTPOperationError>
+
+// Return true if the provided one-time-password has succeeded the server's TOTP challenge
+async fn client_totp_try_challenge(one_time_password: String) -> Result<bool, ClientTOTPOperationError>
```
FirelightFlagboy: **nitpick:** Rendering is borked on GitHub code review
```suggestion
async fn client_totp_status() -> Result<TOTPStatus, ClientTOTPOperationError> {}

// Return true if the provided one-time-password has succeeded the server's TOTP challenge
async fn client_totp_try_challenge(one_time_password: String) -> Result<bool, ClientTOTPOperationError> {}
```

docs/rfcs/1025-totp-protected-local-device.md
```diff
@@ -0,0 +1,702 @@
...
+### 5.2 - Create/fetch TOTP-protected opaque key
+
+```rust
+async fn client_totp_create_opaque_key(one_time_password: String) -> Result<(UUID, SecretKey), ClientTOTPOperationError>
+
+enum ClientTOTPFetchOpaqueKeyError {
+    Offline,
+    NotFound,
+}
+async fn client_totp_fetch_opaque_key(opaque_key_id: UUID, one_time_password: String) -> Result<SecretKey, ClientTOTPFetchOpaqueKeyError>
```
FirelightFlagboy: **nitpick:** Render borked by GitHub code review
```suggestion
async fn client_totp_create_opaque_key(one_time_password: String) -> Result<(UUID, SecretKey), ClientTOTPOperationError> {}

enum ClientTOTPFetchOpaqueKeyError {
    Offline,
    NotFound,
}
async fn client_totp_fetch_opaque_key(opaque_key_id: UUID, one_time_password: String) -> Result<SecretKey, ClientTOTPFetchOpaqueKeyError> {}
```

docs/rfcs/1025-totp-protected-local-device.md
touilleMan: I don't understand what you mean :thinking: 

This function is obviously missing a body, but adding a dummy `{}` doesn't change much (the "body" still doesn't return the correct type). I could add a `{ todo!() }` but it seems needless since this code won't be actually compiled anyway (and having a `todo!()` may look like the RFC is missing something, so it's also less clear...).

Besides the rendering seems good even without body: https://github.com/Scille/parsec-cloud/blob/6c379c43ce8a7e0e64a3c7592d7d1c141afc3bd0/docs/rfcs/1025-totp-protected-local-device.md#52---createfetch-totp-protected-opaque-key so I don't think there is something to change here

docs/rfcs/1025-totp-protected-local-device.md
touilleMan: see https://github.com/Scille/parsec-cloud/pull/11950#discussion_r2792470763

docs/rfcs/1025-totp-protected-local-device.md
touilleMan: :+1: done

docs/rfcs/1025-totp-protected-local-device.md
FirelightFlagboy: It's an issue on GitHub side, their code render doesn't seems to like the lack of a body borking the render of the texts that follows

docs/rfcs/1025-totp-protected-local-device.md
touilleMan: RFC 6238 defines defaults values for reference time/interval/algorithm
So I'd rather omit them and stick to the default:
- This simplifies the implementation (typically no need to define an enum for all the different algorithms)
- Also simplifies usage as the server admin doesn't have to configure what the TOTP should be based on
- Using the defaults should be "good enough" and is the most compatible solution (as supporting other algorithms than SHA1 is optional in the RFC)
- We will be able to add optional field to customize the configuration if those are needed in the future. Though it is highly unlikely anybody cares about the algorithm used in the TOTP, at least not until SHA1 is proven to be broken for this usage (and at this point we may even want to "change the default" instead of allowing a broken configuration to be still used, so again no need for customization of the config ^^).

docs/rfcs/1025-totp-protected-local-device.md
touilleMan: I agree brute force attack on the TOTP is unlikely once we are authenticated (since the TOTP is supposed to protect a device, but we *have* access to a device given we are able to send authenticated request...)

However this is the wrong way to think: the TOTP protects multiple devices, but being authenticated means we have access to a single device. So I'd rather consider the TOTP is still sensitive in this case and avoid any possible brute force attack.

docs/rfcs/1025-totp-protected-local-device.md
touilleMan: I'd rather stick with the defaults for simplicity (see https://github.com/Scille/parsec-cloud/pull/11950#discussion_r2792540234)

docs/rfcs/1025-totp-protected-local-device.md
touilleMan: I'd rather stick with the defaults for simplicity (see https://github.com/Scille/parsec-cloud/pull/11950#discussion_r2792540234)

docs/rfcs/1025-totp-protected-local-device.md
AureliaDolo: I wonder if it's worth considering what happens if we want to add other MFA, and make them cumulative with TOTP

docs/rfcs/1025-totp-protected-local-device.md
```diff
@@ -0,0 +1,702 @@
...
+   protection (e.g. password). By doing this the client obtain the _primary_
+   decryption key.
+6. The Parsec client use both _primary_ and _secondary_ decryption keys to
+   decrypt the device keys file.
+
+## 2 - Overview
+
+### 2.1 Where TOTP is to be used
+
+Currently a local device can be protected by multiple solutions, for instance:
```
AureliaDolo:
```suggestion
Currently a local device can be protected by one of the following multiple solutions:
```

To explicit that it is only one of these at a time.

docs/rfcs/1025-totp-protected-local-device.md
```diff
@@ -0,0 +1,702 @@
...
+
+- Password
+- OS keyring
+- PKI (i.e. smartcard)
+- OpenBao (i.e. SSO-based authentication)
+
+Most of those solutions are local-only, which is both a blessing and a curse:
+
+- pro: Offline login is possible
```
AureliaDolo, line 37: Maybe offline data access is better ? I don't know :/

docs/rfcs/1025-totp-protected-local-device.md
```diff
@@ -0,0 +1,702 @@
...
+
+So instead we must first do the TOTP check, and have its result used as a secret
+required to decrypt the device keys file.
+
+This means the TOTP check is sensitive to brute force attack.
+
+> [!NOTE]
+> Considering an attacker making one TOTP attempt every 30s (so every
+> time the TOTP code changes), he would have a ~65% chance of succeeding the
+> challenge after one year.
```
AureliaDolo, line 84: I don't get how/why there would be only one attempt per 30 sec.

> he would have a ~65% chance of succeeding the
> challenge after one year.

That doesn't sound bad _bad_ but doesn't sound good either.

docs/rfcs/1025-totp-protected-local-device.md
```diff
@@ -0,0 +1,702 @@
...
+                "status": "ok",
+                "fields": [
+                    {
+                        "name": "opaque_key_id",
+                        "type": "TOTPOpaqueKeyID"
+                    },
+                    {
+                        // Note the server knows about this key as it is not encrypted!
+                        //
+                        // This is because TOPT's goal is to protect against unauthorized
```
AureliaDolo:
```suggestion
                        // This is because TOTP's goal is to protect against unauthorized
```

docs/rfcs/1025-totp-protected-local-device.md
```diff
@@ -0,0 +1,702 @@
...
+            ]
+        },
+        "reps": [
+            {
+                "status": "ok",
+                "fields": [
+                    {
+                        // Note the server knows about this key as it is not encrypted!
+                        //
+                        // This is because TOPT's goal is to protect against unauthorized
```
AureliaDolo:
```suggestion
                        // This is because TOTP's goal is to protect against unauthorized
```

docs/rfcs/1025-totp-protected-local-device.md
```diff
@@ -0,0 +1,702 @@
...
+>
+> - The worst case is the end user can no longer protect his device with TOTP,
+>   he is not locked out of it.
+> - The outcome is similar to what would occur if the end user had lose his
+>   authenticator app: in both case the server admin has to intervene to reset
+>   the TOTP setup.
+
+### 2.6 - Reset TOTP setup
+
+When a user lose his TOTP authenticator app, his TOTP setup has to be reset.
```
AureliaDolo, line 183: What happens if some/all devices are TOTP protected even after the reset?

docs/rfcs/1025-totp-protected-local-device.md
touilleMan: done :+1:

docs/rfcs/1025-totp-protected-local-device.md
touilleMan: :+1: done.

docs/rfcs/1025-totp-protected-local-device.md
touilleMan: I don't think it is better: "data access" looks like we are accessing manifests & blocks (i.e. accessing encrypted files), while here we are only interested in the protection of the device keys file.

docs/rfcs/1025-totp-protected-local-device.md
touilleMan: > I don't get how/why there would be only one attempt per 30 sec.

The one-time-password changes every 30s, so this simplifies my stat since I can consider each attempt to be independent of each others.

> That doesn't sound bad bad but doesn't sound good either.

This "one attempt per 30s" is just a random example to have an idea of what happen with a sneaky attacker.
With such a low rate we would notice it, and the cost to run this attack is virtually null.

Obviously if there is no rate limit, the attacker can choose instead to crank up hip attack rate: for instance doing 1 attack per second would go 30 time faster, so the 65% success threshold would be achieved in 10 days... ouch !

so tl;dr: we definitely need some rate limiting here !

docs/rfcs/1025-totp-protected-local-device.md
touilleMan: :+1: done

docs/rfcs/1025-totp-protected-local-device.md
touilleMan: :+1: done

docs/rfcs/1025-totp-protected-local-device.md
touilleMan: No opaque key is lost during TOTP setup reset, so once the TOTP setup is completed again the user can decrypt his devices by succeeding the TOTP challenge with his new authenticator app

Put it another way: TOTP setup reset only changes the TOTP secret shared between the authenticator app and the server.

touilleMan: I've added a word about this in this §

docs/rfcs/1025-totp-protected-local-device.md
AureliaDolo: Upon thinking a bit more about that: chances of getting the right code are one in 10^6. Every 30 secs the code resets so the one attempt per 30 sec chance has the same success rate as a one-attempt-ever attack (not a very involved attack). 
One attacker would have to succeed in a 30 sec window or not at all. They can renew the attack every 30 secs but they would have to start over. So to have 1/2 chances of succeeding, they'll need 5×10^5 attempts in a 30 sec window so 16 666.66 attempts per second. That would be quite noticeable. And they'll need to start over every 30 sec.

I think your numbers consider that the TOTP code doesn't change ^^

touilleMan: - Since there is one attempt every 30s, each attempt is independent.
- So for each attempt, there is a 1/10⁶ chance of success. Hence a 1 - 1/10⁶ chance of failure.
- In one year there is (365 * 24 * 3600) / 30 = 1051200 attempts
- So the probability to have absolutely no success in one year is (1 - 1/10⁶)¹⁰⁵¹²⁰⁰ so roughly 0.349
- The probability to have at least one success is (1 - <probability of no success>) hence the 0.650 i.e. 65%

PR #11951, opened 2025-12-17
fix(pki): Regen `test-pki` with supported RSA key

FirelightFlagboy:


libparsec/crates/platform_pki/tests/units/shared.rs
```diff
@@ -0,0 +1,65 @@
...
+    let binding = crate::shared::Certificate::from_der(der_certificate);
+
+    let untrusted_cert = binding
+        .to_end_certificate()
+        .expect("Should be a valid certificate data");
+
+    verify_certificate(
+        &untrusted_cert,
+        &trusted_roots,
+        &[],
```
AureliaDolo, line 36: nit: add intermediate certs

misc/gen_test_pki.py
```diff
@@ -99,31 +99,35 @@ class CertificateConfig:
         name="black_mesa",
         type=CertificateType.Root,
         subject={COMMON_NAME: "Black Mesa CA", ORGANIZATION: "Black Mesa"},
-        key_algorithm=RSAKeyAlgorithm(length=1024),
+        key_algorithm=RSAKeyAlgorithm(
+            # Before changing, make sure that webpki support that size.
+            # Last time I've checked, webpki support RSA key between 2048-8192 in size
```
AureliaDolo, line 104: nit: add link to the doc

libparsec/crates/platform_pki/tests/units/shared.rs
FirelightFlagboy: I've added a test that make use of intermediate cert using `mallory` cert

misc/gen_test_pki.py
FirelightFlagboy: Done

PR #11960, opened 2025-12-18
[MS] Removed the splashscreen

Max-7:


mmmarcos (approved):
Works for me, although there's still some minor "glitch" when maximizing the window.

Didn't review the code though.

Issue #9161, opened 2024-12-11
[🚀 | Feature request]: CLI create folder command


Landeers:
### What you would like to accomplish

Add a command to create empty folder

### How the feature should work

`parsec-cli workspace folder create` ?

mmmarcos:
This is how it is implemented in the GUI (thanks @Max-7!):

> Currently, the GUI iterates over the folder, and for each file, adds a job to import it into Parsec. The job takes a [`File`](https://developer.mozilla.org/en-US/docs/Web/API/File), a workspace, and a destination path. The job does as follow:
> - create the destination folder if it doesn't exist
> - open a temporary file matching Parsec's prevent sync pattern (`._${crypto.randomUUID()}`) in the destination folder (write: true, truncate: true, create: true)
> - resize the file to the input file's size
> - read the input file, write to the tmp file until everything's been written
> - close the temporary file's file descriptor
> - rename the temporary file to the final file
> 
> The temporary file is used to handle interruptions properly (if the user cancels the job). Reading/writing takes some time, but renaming is atomic, so if we get an interruption, we can simply delete the temporary file and nothing gets imported. It's even more important when replacing an existing file, because deletion is not possible. Without this, the imported file may only be partially imported and in an unknown state.
> 
> Currently, the GUI has a job for every file, but it should instead group by operation: importing a folder is one operation and should result in one job, not as many jobs as they are files. All files get imported in a temporary file, we keep a list of those, and at the end we just rename them all (which technically doesn't make it atomic but the GUI gets to chose when to process the cancel so it's not much of a problem).
> 
> Another aspect to take into account is what to do when importing a file that already exist: either rename the current file, `File (2).ext` or overwrite the existing one.

Max-7:
Currently being re-worked in the GUI by #11933. The specific logic for the import. Looks something like:

```typescript
async function importFiles(files: Array<File>, workspaceHandle: WorkspaceHandle, destination: FsPath) {
  // Transaction allows us to remember all the files we imported, associated their original name
  // and path with the temporary path in which they were importer
  const transaction = new OperationTransaction(workspaceHandle);

  try {
    // Iterate over the files to import
    for (const file of files) {
      // For each one, get a temporary path matching the exclude sync pattern
      const tmpPath = await getTemporaryPath(destination);
      let fdW: FileDescriptor | null = null;
      try {
        const fdWResult = await openFile(workspaceHandle, tmpPath, { write: true, createNew: true });
    
[...]

PR #11967, opened 2025-12-19
feat(cli): Add flag --parents to command workspace import

mmmarcos:
The flag `--parents`/`-p` will make the command to create parent directories if they not exists. No errors if they already exists (similar to `mkdir -p`).

```shell
parsec-cli workspace import --parents /path/to/myfile.txt
```

Related to #9161

### DIFF
--- a/cli/src/commands/workspace/import.rs
+++ b/cli/src/commands/workspace/import.rs
@@ -9,10 +9,20 @@ use crate::utils::StartedClient;
 crate::clap_parser_with_shared_opts_builder!(
     #[with = config_dir, device, password_stdin, workspace]
     pub struct Args {
-        /// Local file to copy
+        /// Local file to import (e.g. "myfile.txt")
         pub(crate) src: PathBuf,
-        /// Workspace destination path
+        /// Destination path in the workspace (e.g. "/path/to/myfile.txt")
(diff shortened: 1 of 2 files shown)


cli/src/commands/workspace/import.rs
```diff
@@ -9,10 +9,20 @@ use crate::utils::StartedClient;
...
-        /// Local file to copy
+        /// Local file to import (e.g. "myfile.txt")
         pub(crate) src: PathBuf,
-        /// Workspace destination path
+        /// Destination path in the workspace (e.g. "/path/to/myfile.txt")
+        ///
+        /// The path is absolute from the workspace root directory. The command will fail if the
+        /// parent directories do not exist, unless the `parents` option is enabled.
+        ///
+        /// If the destination file already exists, its content will be replaced.
```
AureliaDolo, line 19: It's an issue for another day, but ideally there should be a warning (and the option to cancel the operation) if there is already a destination file.

mmmarcos: Added to task list in #9161.

Note however that this may break current user workflows, but [every change breaks someone's workflow](https://xkcd.com/1172/) ^^

PR #11994, opened 2025-12-30
[MS] Add workspace categories menu component for filtering workspaces

### LINKED ISSUE #11993: Add workspace categories to simplify previewing

2 things:
- Add workspace categories menu
- Update sidebar

Add categories like Recents, All workspaces, Favorites. Later, I will be added:
In sidebar:
- Shared with me
- Trashed workspaces
- Archived workspaces

In workspace page (menu):
- Shared with me

Image

_No response_

fabienSvtr:
https://github.com/user-attachments/assets/35bb3b5b-d9d7-4e86-ae89-c8cc4627fc25

closes #11993


client/src/views/workspaces/WorkspacesPage.vue
Ironicbay, line 1: It would be nice to have an alternative text for External users since they can't create workspaces
"You do not have access to any workspace yet. Workspaces that ~you create or~ have been shared with you will be listed here."
Screenshot 2025-12-31 at 12 34 52

client/src/locales/en-US.json
```diff
@@ -1011,7 +1018,7 @@
     "SideMenu": {
         "organization": "Organization",
         "workspaces": "My workspaces",
-        "recentWorkspaces": "Recent",
+        "recentWorkspaces": "Recents",
```
```suggestion
        "recentWorkspaces": "Recent",
```

client/src/views/menu/MainMenu.vue
```diff
@@ -690,6 +682,14 @@ onUnmounted(async () => {
   watchWindowWidthCancel();
 });
 
+async function onWorkspaceMenuClick(selectedMenu: WorkspaceMenu): Promise<void> {
+  workspaceMenuSelected.value = selectedMenu;
+  await navigateTo(Routes.Workspaces, {
+    params: { handle: getConnectionHandle() },
+    query: { filter: selectedMenu },
```
Ironicbay: I'd rather have a clean URL with `WorkspaceMenu.All` instead of `filter=all`. `All` shouldn't be a visible filter, it's a default value.

As a matter of fact, why even use the router for this? Wouldn't it be easier to just filter what workspace is shown depending on what tab the user clicked? This way we don't have to mess with the navigation (which is the source of many bugs as you're already well aware), as of this implementation switching tabs on the web client needlessly populates the browser history.

client/src/locales/fr-FR.json
```diff
@@ -698,6 +699,13 @@
         "filter": {
             "title": "Filtrer",
             "roles": "Rôle"
+        },
+        "categoriesMenu": {
+            "all": "Tous",
```
```suggestion
            "all": "Tous les espaces",
```

client/src/locales/en-US.json
```diff
@@ -632,6 +632,7 @@
         "createWorkspace": "New workspace",
         "itemCount": "No items | 1 item | {count} items",
         "noWorkspaces": "You do not have access to any workspace yet. Workspaces that you create or have been shared with you will be listed here.",
+        "noWorkspacesExternal": "You have not joined any workspace yet. Shared workspaces will appear here when you are invited.",
```
Ironicbay: I think the original wording works well.
```suggestion
        "noWorkspacesExternal": "You do not have access to any workspaces. Workspaces that have been shared with you will be listed here.",
```

client/src/locales/en-US.json
```diff
@@ -632,6 +632,7 @@
         "createWorkspace": "New workspace",
         "itemCount": "No items | 1 item | {count} items",
         "noWorkspaces": "You do not have access to any workspace yet. Workspaces that you create or have been shared with you will be listed here.",
```
Ironicbay: Might as well fix this one 
```suggestion
        "noWorkspaces": "You do not have access to any workspaces yet. Workspaces that you create or have been shared with you will be listed here.",
```

client/src/locales/en-US.json
```diff
@@ -699,6 +700,13 @@
         "filter": {
             "title": "Filter",
             "roles": "Role"
+        },
+        "categoriesMenu": {
+            "all": "All workspaces",
+            "recents": "Recently viewed",
+            "favorites": "Pinned",
+            "noRecentWorkspaces": "You have not consulted any workspace yet. Recent consulted workspaces will appear here.",
```
```suggestion
            "noRecentWorkspaces": "You have not consulted any workspaces yet. Recently consulted workspaces will be listed here.",
```

client/src/locales/en-US.json
```diff
@@ -699,6 +700,13 @@
...
+            "noFavoriteWorkspaces": "You have not pinned any workspace yet. Pin a workspace to have it appear here."
```
```suggestion
            "noFavoriteWorkspaces": "You have not pinned any workspaces yet. Pin a workspace to have it be listed here."
```

client/src/locales/fr-FR.json
```diff
@@ -631,6 +631,7 @@
         "createWorkspace": "Nouvel espace de travail",
         "itemCount": "Aucun élément | 1 élément | {count} éléments",
         "noWorkspaces": "Vous n'avez pas encore accès à un espace de travail. Les espaces de travail que vous créerez ou qui vous seront partagés s'afficheront ici.",
+        "noWorkspacesExternal": "Vous n'avez pas encore rejoint d'espace de travail. Les espaces de travail partagés apparaîtront ici lorsque vous serez invité.",
```
Ironicbay: Same here, the original wording works fine, there's technically no invitation for workspaces, only sharing / unsharing
```suggestion
        "noWorkspacesExternal": "Vous n'avez pas encore accès à un espace de travail. Les espaces de travail qui vous seront partagés s'afficheront ici.",
```

client/src/locales/fr-FR.json
```diff
@@ -698,6 +699,13 @@
         "filter": {
             "title": "Filtrer",
             "roles": "Rôle"
+        },
+        "categoriesMenu": {
+            "all": "Tous",
+            "recents": "Vu récemment",
```
Ironicbay: Up to you if you want to write `Vu` or `Vus`, singular implies an uncountable subject (workspace as a concept, or data), plural implies the subjects are the workspaces. Line 1021 you're using plural so it's mostly for the sake of consistency.
```suggestion
            "recents": "Vus récemment",
```

client/src/components/workspaces/WorkspaceCategoriesMenu.vue
```diff
@@ -0,0 +1,164 @@
...
+
+defineProps<{
+  activeMenu?: WorkspaceMenu;
+}>();
+
+defineEmits<{
+  (e: 'updateMenu', value: WorkspaceMenu): void;
+}>();
+
+const workspaceMenuList = computed(() => [
```
Ironicbay: This is just a static list, you can remove the `computed`

client/src/views/menu/MainMenu.vue
```diff
@@ -690,6 +676,13 @@ onUnmounted(async () => {
   watchWindowWidthCancel();
 });
 
+async function onCategoryMenuChange(menu: WorkspaceMenu): Promise<void> {
+  if (getCurrentRoute.name !== Routes.Workspaces) {
```
Ironicbay: Correct call for this would be
```suggestion
  if (getCurrentRoute().value.name !== Routes.Workspaces) {
```

but you can use the existing function `currentRouteIs`:
```suggestion
  if (!currentRouteIs(Routes.Workspaces)) {
```

client/src/views/menu/MainMenu.vue
```diff
@@ -424,6 +420,7 @@ const emits = defineEmits<{
   (event: 'sidebarWidthChanged', value: number): void;
 }>();
 
+const workspaceMenuSelected = workspaceMenuState;
```
Ironicbay: You can remove this const and use `workspaceMenuState` instead.

client/src/views/workspaces/WorkspacesPage.vue
```diff
@@ -218,6 +266,7 @@ const searchFilterContent = ref('');
 const workspaceFilters = ref<WorkspacesPageFilters>({ owner: true, manager: true, contributor: true, reader: true });
 const querying = ref(true);
 
+const workspaceMenuSelected = workspaceMenuState;
```
Ironicbay: You can remove this const and use workspaceMenuState instead.

client/src/views/workspaces/WorkspacesPage.vue
```diff
@@ -483,6 +551,11 @@ function onMsSorterChange(event: MsSorterChangeEvent): void {
   sortByAsc.value = event.sortByAsc;
 }
 
+async function onWorkspaceMenuClick(selectedMenu: WorkspaceMenu): Promise<void> {
+  setWorkspaceMenu(selectedMenu);
+  await navigateTo(Routes.Workspaces, { params: { handle: getConnectionHandle() } });
+}
+
```
Ironicbay: You don't need to navigate anymore, this function can be deleted and you can just call `setWorkspaceMenu` instead

client/src/views/menu/MainMenu.vue
```diff
@@ -690,6 +676,13 @@ onUnmounted(async () => {
   watchWindowWidthCancel();
 });
 
+async function onCategoryMenuChange(menu: WorkspaceMenu): Promise<void> {
+  if (getCurrentRoute.name !== Routes.Workspaces) {
+    await navigateTo(Routes.Workspaces);
+  }
+  workspaceMenuSelected.value = menu;
```
Ironicbay:
```suggestion
  setWorkspaceMenu(menu);
```

Ironicbay (changes requested):

client/src/services/workspaceMenuState.ts
```diff
@@ -0,0 +1,14 @@
+// Parsec Cloud (https://parsec.cloud) Copyright (c) BUSL-1.1 2016-present Scille SAS
+
+import { WorkspaceMenu } from '@/views/workspaces/types';
+import { ref } from 'vue';
+
+export const workspaceMenuState = ref<WorkspaceMenu>(WorkspaceMenu.All);
```
Max-7, line 6: Either remove the function or don't export this variable. Doing both is confusing.

client/src/views/workspaces/WorkspacesPage.vue
```diff
@@ -452,7 +503,24 @@ async function refreshWorkspacesList(): Promise<void> {
 const filteredWorkspaces = computed(() => {
   const filter = searchFilterContent.value.toLocaleLowerCase();
   return Array.from(workspaceList.value)
-    .filter((workspace) => workspace.currentName.toLocaleLowerCase().includes(filter) && !isWorkspaceFiltered(workspace.currentSelfRole))
+    .filter((workspace) => {
+      switch (workspaceMenuSelected.value) {
+        case WorkspaceMenu.Recents:
+          return recentDocumentManager
+            .getWorkspaces()
+            .map((w) => w.id)
```
Max-7: No point in mapping (creating a new list), just use find() instead of map+includes

client/tests/e2e/helpers/assertions.ts
```diff
@@ -250,7 +250,7 @@ export const expect = baseExpect.extend({
 
   async toBeWorkspacePage(page: Page): Promise<AssertReturnType> {
     try {
-      await expect(page).toHaveURL(/\/\d+\/workspaces$/);
+      await expect(page).toHaveURL(/\/\d+\/workspaces(\?filter=(all|favorites|recents))?$/);
```
Max-7: Why is that needed? I haven't seen anything having to do with the routing.

client/src/components/workspaces/WorkspaceCategoriesMenu.vue
```diff
@@ -0,0 +1,164 @@
...
+import { WorkspaceMenu } from '@/views/workspaces/types';
+import { IonIcon, IonText } from '@ionic/vue';
+import { rocket, star, time } from 'ionicons/icons';
+import { useWindowSize } from 'megashark-lib';
+import { computed } from 'vue';
+
+const { isLargeDisplay } = useWindowSize();
+
+defineProps<{
+  activeMenu?: WorkspaceMenu;
```
Max-7: Why is it optional?

Max-7 (changes requested):

client/tests/e2e/helpers/assertions.ts
fabienSvtr: I used it when I was using the router and the URL. I just forgot to remove it.

client/src/components/workspaces/WorkspaceCategoriesMenu.vue
fabienSvtr: At first, I wasn't sure whether we would use activeMenu, but after thinking about it, yes, we do use it, so I removed that option.

client/src/components/workspaces/WorkspaceCategoriesMenu.vue
```diff
@@ -0,0 +1,163 @@
...
+import { WorkspaceMenu } from '@/views/workspaces/types';
+import { IonIcon, IonText } from '@ionic/vue';
+import { rocket, star, time } from 'ionicons/icons';
+import { useWindowSize } from 'megashark-lib';
+
+const { isLargeDisplay } = useWindowSize();
+
+defineProps<{
+  activeMenu: WorkspaceMenu;
+}>();
```
Ironicbay, line 34: `activeMenu` is always set to `getWorskpaceMenu()` so you can ditch the prop.
```suggestion
const activeMenu = computed(() => getWorkspaceMenu());
```

and in `WorkspacesPage.vue`:
```
 <workspace-categories-menu @update-menu="setWorkspaceMenu" />
```

Max-7: I don't think the computed here would work properly since only the value is returned, not the ref. So the computed() would not know when to call the function.
I also like the idea of working with a prop, instead of the component accessing the value itself, it makes everything less tangled (you can test the component by itself, you don't need the global state).

client/src/services/workspaceMenuState.ts
```diff
@@ -0,0 +1,14 @@
+// Parsec Cloud (https://parsec.cloud) Copyright (c) BUSL-1.1 2016-present Scille SAS
+
+import { WorkspaceMenu } from '@/views/workspaces/types';
+import { ref } from 'vue';
+
+const workspaceMenuState = ref<WorkspaceMenu>(WorkspaceMenu.All);
+
+export function setWorkspaceMenu(menu: WorkspaceMenu): void {
```
Max-7: Probably better to remove the functions and just export the variable. The getter strips the ref, which means it can't be watched.

PR #12002, opened 2026-01-05
Async enrollment postgres

touilleMan:


server/parsec/components/postgresql/async_enrollment_accept.py
```diff
@@ -0,0 +1,441 @@
...
+    accept_payload_signature_pki_algorithm = $accept_payload_signature_pki_algorithm,
+    submit_payload_signature_pki_author_x509_certificate = $submit_payload_signature_pki_author_x509_certificate,
+    accept_payload_signature_openbao_signature = $accept_payload_signature_openbao_signature,
+    accept_payload_signature_openbao_author_entity_id = $accept_payload_signature_openbao_author_entity_id
+WHERE
+    _id = $enrollment_internal_id
+""")
+
+
+_q_insert_user_and_device = Q(f"""
```
mmmarcos: For `pki_accept` this was implemented re-using [`_q_insert_user_and_device`](https://github.com/Scille/parsec-cloud/blob/b68046e9e993f0a16b71e93ebc9755c9eb7b66e7/server/parsec/components/postgresql/user_create_user.py#L31) from `user_create_user.py`:

https://github.com/Scille/parsec-cloud/blob/b68046e9e993f0a16b71e93ebc9755c9eb7b66e7/server/parsec/components/postgresql/pki_accept.py#L35

I know that we've started to prefer each component/command to define its own set of queries (specially when there are slightly differences between the needs of each command), BUT for complex queries like this one, the duplication introduces a significant risk of error IMO.

Since there seems to be no difference, why not reusing-it? The only potential issue I see is that it has to be clear when a query can be reused... we could move them to the `queries` directory which already contains "common" queries.

server/parsec/components/async_enrollment.py
```diff
@@ -165,21 +168,104 @@ class AsyncEnrollmentListItem:
...
+        ]
+    except ValueError:
+        return "INVALID_DER_X509_CERTIFICATE"
+    x509_trustchain = [author_der_x509_certificate, *intermediate_der_x509_certificates]
+
+    # Validate the x509 trustchain:
+    # - Should not be longer than 10 items (including leaf and root)
+    # - Each intermediate certificate should be used exactly once
+
+    if len(intermediate_der_x509_certificates) > 8:  # 8 since we exclude leaf and root
```
AureliaDolo: We have the MAX_INTERMEDIATE_CERTIFICATES_DEPTH (the value is up to discussion)

server/parsec/components/async_enrollment.py
```diff
@@ -165,21 +168,104 @@ class AsyncEnrollmentListItem:
...
+            issuer=certificate.issuer(),
+            subject=certificate.subject(),
+            sha256_fingerprint=sha256(der).digest(),
+        )
+
+    def __repr__(self) -> str:
+        return f"{self.__class__.__qualname__}(subject={self.subject}, issuer={self.issuer})"
+
+
+def _validate_x509_trustchain(
```
AureliaDolo, line 201: You may want to have a look at build_trustchain https://github.com/Scille/parsec-cloud/blob/b68046e9e993f0a16b71e93ebc9755c9eb7b66e7/server/parsec/components/pki.py#L331

server/parsec/components/async_enrollment.py
touilleMan: done, this also makes simpler to test this error case

server/parsec/components/async_enrollment.py
touilleMan: I had a look at it, and think it would be simpler to avoid using dictionary when checking the intermediary certificates:
- the number of items is very small (and async enrollement APIs are not used often), so there is no performance concern
- `list(trustchain.values())` relies on the fact Python dictionaries retain insertion order, this is something implicit (and it has not been always the case), so I'd rather avoid it

On top of that, the fact `PkiCertificate` gets its `signed_by` attribute updated add complexity (since the object need to be mutable) and is not strictly needed (the fingerprint can be retrieved by iterating over the trustchain in the code doing the SQL query, which means quadratic complexity but again this is not an issue since the number of certificates is small)

server/parsec/components/postgresql/async_enrollment_accept.py
touilleMan: You are totally right on this, I've updated the code to directly use `_q_insert_user_and_device` :+1:

PR #12016, opened 2026-01-06
refactor(parsec-cli): Workspace import will return an error on non-file src

FirelightFlagboy:
It's the opposite of what we want for #9161, but it provide a base where non-file (mostly folder) are explicitly indicated as not supported.

### DIFF
--- a/cli/src/commands/workspace/import.rs
+++ b/cli/src/commands/workspace/import.rs
@@ -1,7 +1,12 @@
-use std::{collections::HashSet, path::PathBuf, sync::Arc, vec};
+use std::{
+    collections::HashSet,
+    path::{Path, PathBuf},
+    sync::Arc,
+    vec,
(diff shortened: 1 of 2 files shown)


cli/src/commands/workspace/import.rs
```diff
@@ -69,6 +74,22 @@ pub async fn workspace_import(args: Args, client: &StartedClient) -> anyhow::Res
             },
         }?;
     }
+    let src_metadata = src.metadata()?;
+    let src_file_type = src_metadata.file_type();
+    if src_file_type.is_file() {
+        import_file(client, workspace, &src, dest, files_to_sync).await
+    } else {
+        anyhow::bail!("Only supporting importing file")
```
mmmarcos, line 82: TIL about the `bail` macro 👍

PR #12023, opened 2026-01-07
feat(parsec-cli): Allow to prevent overwriting a file during workspace import

FirelightFlagboy:
Closes #112021

### DIFF
--- a/cli/src/commands/workspace/import.rs
+++ b/cli/src/commands/workspace/import.rs
@@ -5,7 +5,7 @@ use std::{
     vec,
 };
 
-use libparsec::{anyhow::Context, EntryName, EntryNameError, FsPath, OpenOptions, VlobID};
+use libparsec::{EntryName, EntryNameError, FsPath, OpenOptions, VlobID, WorkspaceOpenFileError};
 use libparsec_client::{EventBus, WorkspaceOps};
 use tokio::{io::AsyncReadExt, task::JoinSet};
 
@@ -14,7 +14,7 @@ use crate::utils::StartedClient;
 crate::clap_parser_with_shared_opts_builder!(
     #[with = config_dir, device, password_stdin, workspace]
     pub struct Args {
-        /// Local file to import (e.g. "myfile.txt")
+        /// Local file or folder to import (e.g. "myfile.txt")
         pub(crate) src: PathBuf,
         /// Destination path in the workspace (e.g. "/path/to/myfile.txt")
         ///
@@ -28,8 +28,22 @@ crate::clap_parser_with_shared_opts_builder!(
         /// No error if parent directories already exist (similar to `mkdir -p`)
         #[clap(long, short, action)]
         pub(crate) parents: bool,
+        /// Control how existing files are updated.
+        ///
+        /// (similar to `cp --update=...`)
+        #[clap(long, value_enum, default_value_t)]
+        update: UpdateMode,
     }
 );
+
+#[derive(Debug, Default, Clone, Copy, PartialEq, Eq, clap::ValueEnum)]
+enum UpdateMode {
+    /// Existing files in destination are replaced
+    All,
+    /// Existing files in destination are not replaced but raise and error instead.
+    #[default]
+    NoneFail,
+}
 
 crate::build_main_with_client!(
     main,
@@ -47,6 +61,7 @@ pub async fn workspace_import(args: Args, client: &StartedClient) -> anyhow::Res
         dest,
         parents,
         workspace: wid,
+        update,
         ..
     } = args;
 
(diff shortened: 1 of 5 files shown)


cli/src/commands/workspace/import.rs
```diff
@@ -28,9 +30,23 @@ crate::clap_parser_with_shared_opts_builder!(
...
+        update: UpdateMode,
     }
 );
 
+#[derive(Debug, Default, Clone, Copy, PartialEq, Eq, clap::ValueEnum)]
+enum UpdateMode {
+    /// Existing files in destination are replaced
+    #[default]
+    All,
+    /// Existing files in destination are not replaced
```
mmmarcos: For folder import, files are skipped or command fail? if skipped, I would explicitly mention it
```suggestion
    /// Existing files in destination are not replaced (skipped)
```

cli/src/commands/workspace/import.rs
```diff
@@ -28,9 +30,23 @@ crate::clap_parser_with_shared_opts_builder!(
...
+        #[clap(long, value_enum, default_value_t)]
+        update: UpdateMode,
     }
 );
 
+#[derive(Debug, Default, Clone, Copy, PartialEq, Eq, clap::ValueEnum)]
+enum UpdateMode {
+    /// Existing files in destination are replaced
+    #[default]
+    All,
```
mmmarcos: IMO `None` is a better default option because it has no collateral damage.

cli/src/commands/workspace/import.rs
```diff
@@ -28,9 +30,23 @@ crate::clap_parser_with_shared_opts_builder!(
         /// No error if parent directories already exist (similar to `mkdir -p`)
         #[clap(long, short, action)]
         pub(crate) parents: bool,
+        /// Control how existing file are updated.
+        ///
+        /// (similar to `cp --update=...`)
```
AureliaDolo, line 33:
```suggestion
        /// (similar to `cp --update={all, none}`)
```
Maybe rephrase that as cp --update as more option than we support.

cli/src/commands/workspace/import.rs
```diff
@@ -28,9 +30,23 @@ crate::clap_parser_with_shared_opts_builder!(
         /// No error if parent directories already exist (similar to `mkdir -p`)
         #[clap(long, short, action)]
         pub(crate) parents: bool,
+        /// Control how existing file are updated.
```
AureliaDolo:
```suggestion
        /// Control how existing files are updated.
```

cli/src/commands/workspace/import.rs
FirelightFlagboy: I've put the ellipsis in case we support more in the future.

Anyway, the `parsec-cli --help` provide an explanation for the enum variant

cli/src/commands/workspace/import.rs
FirelightFlagboy: That would be a breaking change

cli/src/commands/workspace/import.rs
FirelightFlagboy: We reuse `import_file` that return an error in case of existing file, like how `cp --update=none` does

mmmarcos: That is not how `cp --update=none` works:
```bash
$ touch a.txt
$ cp a.txt a.txt --update=none
$ echo $?
0
$ cp a.txt a.txt --update=none-fail
cp: not replacing 'a.txt'
$ echo $?
1
```

So I suggest one of the following:
1. keep current behavior, rename option to `none-fail`
2. keep current option, update behavior to match `cp`
3. support both `none` and `none-fail`

More tests may be needed depending on the option you prefer.

Note that, when `cp` multiple files (like our import folder), `none-fail` will does not abort on error: it prints a message for each existing file but it will process all the files. 

In any case, error or skip, all of this should be clearly documented in the option.

cli/src/commands/workspace/import.rs
```diff
@@ -108,15 +130,19 @@ async fn import_file(workspace: Arc<WorkspaceOps>, src: &Path, dest: FsPath) ->
...
+                // that the opened file did not exist beforehand
+                // (so not updating something that exist)
+                create_new: update == UpdateMode::None,
             },
         )
         .await
-        .with_context(|| {
-            format!(
+        .map_err(|e| match e {
+            WorkspaceOpenFileError::EntryExistsInCreateNewMode { .. } if update == UpdateMode::None => anyhow::anyhow!("{dest} already exist and we are not allowed to update it with current update mode"),
```
mmmarcos: nit: i would use a little less verbose message
```suggestion
            WorkspaceOpenFileError::EntryExistsInCreateNewMode { .. } if update == UpdateMode::None => anyhow::anyhow!("Skipping {dest}: already exists"),
```

cli/src/commands/workspace/import.rs
FirelightFlagboy: I've renamed `none` to `none-fail` and make it the default since it appear that no one is using it because of the lake of folder support

cli/src/commands/workspace/import.rs
FirelightFlagboy: done

cli/src/commands/workspace/import.rs
```diff
@@ -256,10 +287,7 @@ fn import_folder(
             let entry_name = match filename_to_entryname(&entry_path) {
                 Some(Ok(v)) => v,
                 Some(Err(e)) => {
-                    log::warn!(
-                        "Skipping {} because it has an invalid filename ({e})",
-                        entry_path.display()
-                    );
+                    log::warn!("Skipping {}: invalid filename ({e})", entry_path.display());
```
mmmarcos: I will change it to print error.

cli/src/commands/workspace/import.rs
```diff
@@ -268,15 +296,15 @@ fn import_folder(
...
                 });
             } else if entry_ft.is_dir() {
                 import_tasks.spawn(async move {
-                    import_folder(dup_workspace, entry_path, entry_dest).await
+                    import_folder(dup_workspace, entry_path, entry_dest, update).await
                 });
             } else {
                 log::warn!(
-                    "Skipping {} because it has an unsupported file type ({entry_ft:?})",
+                    "Skipping {}: unsupported file type ({entry_ft:?})",
```
mmmarcos: I will change it to print error.

cli/src/commands/workspace/import.rs
```diff
@@ -98,7 +126,78 @@ pub async fn workspace_import(args: Args, client: &StartedClient) -> anyhow::Res
...
+            path: src.clone(),
+            file_type: src_file_type,
+        })
+    }
+}
+
+#[derive(Debug, thiserror::Error)]
+enum ImportFileError {
+    #[error("already existing")]
+    AlreadyExisting,
```
mmmarcos: I'd rename
```suggestion
    #[error("already exists")]
    AlreadyExists,
```

mmmarcos (approved):
LGTM.

Since the command now supports importing folders, we can update the comment for `src` argument:

```rust
/// The local file **or folder** to import (e.g. "myfile.txt")
```

Also in `mod.rs`:
```rust
/// Import a local file **or folder** to a remote workspace
```

PR #12053, opened 2026-01-14
Simplify errors in `libparsec_platform_pki`

touilleMan:


libparsec/crates/platform_pki/src/shared/mod.rs
```diff
@@ -76,10 +75,7 @@ pub fn verify_message<'message, 'a>(
     certificate
         .verify_signature(verifier, &signed_message.message, &signed_message.signature)
         .map(|_| signed_message.message.as_ref())
-        .map_err(|e| match e {
-            WebPkiError::InvalidSignatureForPublicKey => VerifySignatureError::InvalidSignature,
-            e => VerifySignatureError::UnexpectedError(e),
-        })
+        .map_err(VerifySignatureError::InvalidSignature)
```
FirelightFlagboy, line 78: **issue:** Webpki can return errors that are not caused by an invalid signature and should be taken into account like it was done before (instead of being discarded)

libparsec/crates/platform_pki/src/windows/find_in_store.rs
FirelightFlagboy, line 1: **todo:** revert change

libparsec/crates/platform_pki/src/x509/mod.rs
FirelightFlagboy, line 1: **todo:** revert change

libparsec/crates/platform_pki/src/shared/mod.rs
FirelightFlagboy: NVM, better now after re-reading

libparsec/crates/platform_pki/src/test_fixture.rs
FirelightFlagboy, line 1: **todo:** revert change

libparsec/crates/platform_pki/tests/units/shared.rs
FirelightFlagboy, line 1: **todo:** revert change

libparsec/crates/platform_pki/tests/units/x509.rs
FirelightFlagboy, line 1: **todo:** revert change

libparsec/crates/platform_pki/src/lib.rs
FirelightFlagboy, line 1: **todo:** revert change

libparsec/crates/platform_pki/src/test_fixture.rs
touilleMan: `use libparsec_types::prelude::*;` is the no brainer one-size-fits-all standard way ® of importing stuff from `libparsec_types` and it is used across all the codebase.

I agree this change is not directly related to the main goal of this PR, however it is such a small change that I don't see the gain in putting it in its own PR :/

touilleMan:
> We've discussed with @AureliaDolo earlier this morning, and for errors conversion, we would prefer that the error implement From over repeating the .map_err(|e| match e ...

We can discuss about it tomorrow, however I don't really like implementing conversion types for errors:
- Errors are part of the signature of a function, so in practice most function should end up with its own error type. This means there is not much factorization gain to get here (i.e. conversion between two given errors types is most of the time only done once in the whole codebase)
- So adding conversion between errors has basically one pro: it makes error handling more compact in the function code
- However it big cons is it push the developer to avoid thinking about error handling when writing the function code. Typically consider a function A that calls a sub-function B twice for two different things (e.g. requesting a database), then we won't even think to have two different kind of errors for function A.
- We have already bitten by this with the `Internal(#[from] anyhow::Error)`. Though it is convenient, we already end up incorrectly returning a `Internal` error variant by mistake due to a `err.into()` converting in error into a `anyhow::Error` (and good luck to spot the mistake ^^)
- Plus, when an error conversion becomes useless (e.g. function A no longer calls function B) we can easily end up with dead code

In the end I think automatic conversion between errors should only be considered when there is strong rational for it  (e.g. `Offline(#[from] ConnectionError),`)

libparsec/crates/platform_pki/src/shared/mod.rs
```diff
@@ -89,13 +85,23 @@ pub fn create_local_pending(
...
+    } = encrypt_message(key.as_ref(), cert_ref).map_err(|err| match err {
+        EncryptMessageError::CannotOpenStore(err) => CreateLocalPendingError::CannotOpenStore(err),
+        EncryptMessageError::NotFound => CreateLocalPendingError::NotFound,
+        EncryptMessageError::CannotGetCertificateInfo(err) => {
+            CreateLocalPendingError::CannotGetCertificateInfo(err)
+        }
+        EncryptMessageError::CannotAcquireKeypair(err) => {
+            CreateLocalPendingError::CannotAcquireKeypair(err)
+        }
+        EncryptMessageError::CannotEncrypt(err) => CreateLocalPendingError::CannotEncrypt(err),
```
AureliaDolo, line 103: I get that you'd rather be explicit each time, but in this case, I find that moving that bloc to a `impl From<EncryptMessageError> for CreateLocalPendingError` would work the same, except it would make that function easier to read/review/maintain. And there is no additional information provided, the variant and data are only moved from one error to the other.

Same goes for other the other map_err.

AureliaDolo:
> Typically consider a function A that calls a sub-function B twice for two different things (e.g. requesting a database), then we won't even think to have two different kind of errors for function A.

I don't get that example. Do you mean something like 
```
enum ErrorA {
    ErrorBDuringStep1VariantX
    ErrorBDuringStep1VariantY
    ErrorBDuringStep2VariantX
    ErrorBDuringStep2VariantY
}
```
instead of
```
enum ErrorA {
    ErrorBVariantX
    ErrorBVariantY
}
```
or do you mean
```
enum ErrorA {
    ErrorBDuringStep1
    ErrorBDuringStep2
}
```
I'm not sure we ever will end up with that much details as in the first example.


> We have already bitten by this with the Internal(#[from] anyhow::Error)

IMHO in that the issue is more in the `anyhow` than in the `into`. If we cast any error to anyhow it's an issue so we should avoid `return SomethingError::Internal(error.into())` not the `function_that_can_fail()?` pattern.

> So adding conversion between errors has basically one pro: it makes error handling more compact in the function code

Again IMHO, the bigger pro, is not just that the code is more compact, but also easier to read, to review and most important to maintain.

touilleMan:
> I don't get that example. Do you mean something like

I meant

```
enum ErrorA {
    ErrorBDuringStep1
    ErrorBDuringStep2
}
```

But having things such as `ErrorBDuringStep1VariantX`/etc. may be the right approach, at all depend on the context of the function being implemented.

That's my point actually: I think we should put the error conversion code in the function body given this code is very dependent of the context of the function.

> IMHO in that the issue is more in the anyhow than in the into. If we cast any error to anyhow it's an issue so we should avoid return SomethingError::Internal(error.into()) not the function_that_can_fail()? pattern.

The point here is the fact that the end developer always uses the `?` operator whenever he cans (it's just too convenient ^^), and from there it is hard to discover the error was improperly handled (you have to jump to the error conversion code, and think about its meaning in the context of the function...). So if we want him to be careful about error conversion then we shouldn't provide the `?` operation.

> Again IMHO, the bigger pro, is not just that the code is more compact, but also easier to read, to review and most important to maintain.

The code as a whole is no more compact if you count the function body AND the `impl<BError> for AError` (since the conversion between the two is mostly only used once, obviously the rational is different if we endup with a generic kind of error used everywhere and always handled the same way)

So we end up with a question: is the whole codebase more readable to have less code in the function body ?
I tend to say no:
- when the conversion code is simple it is no big deal to have a more verbose code (it is indented, it is a pattern very common in the codebase, we are already used to it ^^)
- but when the code is not that simple it's very important to have it placed in the context it is meant to be used ! Otherwise you have to jump from the function body to the error declaration, find the `impl` declaration corresponding to the current error being matched (typically not fun to do during code review where there is no rust-analyzer to provide the type of the error being converted). Whenever the code change we can leave a `impl` behind that will no longer be used

libparsec/crates/platform_pki/src/shared/mod.rs
touilleMan: `create_local_pending` (and others payload-related functions) is a function related to the old pki enrollment, so it is going to be removed soon

I think it's very revealing when you say :stuck_out_tongue_winking_eye: :

> except it would make that **function** easier to read/review/maintain.

You consider only the function, not the error conversion code. However, for me, the key point is the fact that function and error conversion are two tightly coupled pieces of code, so I'd rather have them together in the function than scattered across the code where they risk silently drafting apart.

libparsec/crates/platform_pki/src/shared/mod.rs
```diff
@@ -175,10 +184,15 @@ pub fn validate_payload<'message, 'cert>(
...
-    )?;
+    )
+    .map_err(|err| match err {
+        VerifyCertificateError::Untrusted(err) => ValidatePayloadError::Untrusted(err),
+    })?;
     let trusted_cert = verified_path.end_entity();
 
-    verify_message(signed_message, trusted_cert).map_err(Into::into)
+    verify_message(signed_message, trusted_cert).map_err(|err| match err {
+        VerifySignatureError::InvalidSignature(..) => ValidatePayloadError::InvalidSignature,
```
FirelightFlagboy: **issue:** You discard data that could be usefull for debugging issues

libparsec/crates/platform_pki/src/shared/mod.rs
```diff
@@ -175,10 +184,15 @@ pub fn validate_payload<'message, 'cert>(
         &intermediate_certs,
         now,
         KeyUsage::client_auth(),
-    )?;
+    )
+    .map_err(|err| match err {
+        VerifyCertificateError::Untrusted(err) => ValidatePayloadError::Untrusted(err),
```
FirelightFlagboy, line 189: **nitpick:** poor error conversion

libparsec/crates/platform_pki/src/shared/mod.rs
FirelightFlagboy: **nitpick:** poor error conversion

libparsec/crates/platform_pki/src/shared/mod.rs
```diff
@@ -207,10 +221,28 @@ pub fn get_validation_path_for_cert(
     cert_ref: &X509CertificateReference,
     now: DateTime,
 ) -> Result<ValidationPathOwned, GetValidationPathForCertError> {
-    let trusted_anchor = crate::list_trusted_root_certificate_anchors()?;
-    let intermediate_certs = crate::list_intermediate_certificates()?;
+    let trusted_anchor =
+        crate::list_trusted_root_certificate_anchors().map_err(|err| match err {
+            ListCertificatesError::CannotOpenStore(err) => {
+                GetValidationPathForCertError::CannotOpenStore(err)
+            }
```
FirelightFlagboy, line 228: **nitpick:** poor error conversion

libparsec/crates/platform_pki/src/shared/mod.rs
```diff
@@ -207,10 +221,28 @@ pub fn get_validation_path_for_cert(
...
+        })?;
+    let intermediate_certs = crate::list_intermediate_certificates().map_err(|err| match err {
+        ListCertificatesError::CannotOpenStore(err) => {
+            GetValidationPathForCertError::CannotOpenStore(err)
+        }
```
FirelightFlagboy, line 233: **nitpick:** poor error conversion

libparsec/crates/platform_pki/src/shared/mod.rs
```diff
@@ -221,7 +253,12 @@ pub fn get_validation_path_for_cert(
         &intermediate_certs,
         now,
         KeyUsage::client_auth(),
-    )?;
+    )
+    .map_err(|err| match err {
+        VerifyCertificateError::Untrusted(err) => {
+            GetValidationPathForCertError::InvalidCertificateUntrusted(err)
+        }
```
FirelightFlagboy, line 260: **nitpick:** poor error conversion

FirelightFlagboy (changes requested):

libparsec/crates/platform_pki/src/shared/mod.rs
touilleMan: :+1: done.

PR #12055, opened 2026-01-14
PKI support for async enrollment 

touilleMan:


libparsec/crates/platform_pki/src/shared/mod.rs
FirelightFlagboy, line 1: **chore:** Revert and use `validate_payload`, the added function duplicate code and will make it harder to test for specific error case

touilleMan: `validate_payload` and `verify_message` are currently factorization bricks to implement `load_submit_payload` & `load_submit_accept`. However those two functions are scheduled to be removed since pki_enrollment is being replaced by async_enrollment.

Hence why I choose to create `verify_message2` that is basically a refactoring of `validate_payload` (main change is removing `SignedMessage` structure since passing more parameters is simpler for the caller).

However, I agree there is currently a dangerous code duplication between `verify_message2` and `get_validation_path_for_cert` for verifying the certificate trustchain.

So I'm going to:
- change the body of `verify_message2` to use `verify_certificate`
- change `verify_certificate` to no longer take `key_usage` as parameter (and rename it to `verify_certificate_`. This is because we always use `KeyUsage::client_auth()` as value, and it is not obvious that this is the correct value in our case (since the PKI feature has nothing to do with client/server interaction).

touilleMan:
> Are you planning to add more tests in another PR ?

Yes and no :/

The difficulty here is that the support of PKI is implemented in the `libparsec` main crate that is not designed to be testable.
Most of the time this is okay though since this crate only contains simple wrappers.

However for the current feature we are far from there, so obviously it would be better to have some tests here!

And to make matters worse, we currently don't have tests on the platform_pki...

So my plan is as follow:
1. Implement test in `libparsec_platform_pki`. Those tests are actually using the OS API to access the PKI.
2. Implement testbed support in `libparsec_platform_pki` to be able to mock PKI operations for tests
3. Move `libparsec/src/async_enrollment.rs`'s `strategy` mode (i.e. the complex code) to its own crate (lets call it `libparsec_async_enrollment_strategy`)
4. Add test to `libparsec_async_enrollment_strategy`, those tests relies on the testbed mock to do OpenBao/PKI operations \o/

libparsec/src/async_enrollment.rs
```diff
@@ -179,6 +208,127 @@ mod strategy {
...
+                                | PKIGetValidationPathForCertError::CannotGetCertificateInfo(_)
+                                | PKIGetValidationPathForCertError::InvalidCertificateDer(_)
+                                | PKIGetValidationPathForCertError::InvalidCertificateDateTimeOutOfRange(_)
+                                | PKIGetValidationPathForCertError::InvalidCertificateUntrusted(_)
+                            ) => SubmitAsyncEnrollmentError::PKIUnusableX509CertificateReference(err.into()),
+                            PKIGetValidationPathForCertError::CannotOpenStore(e) => {
+                                SubmitAsyncEnrollmentError::PKICannotOpenCertificateStore(e.into())
+                            }
+
+                        })?;
```
FirelightFlagboy, line 263: **nitpick:** spacing

libparsec/src/async_enrollment.rs
```diff
@@ -179,6 +208,127 @@ mod strategy {
...
+            (SecretKey, AsyncEnrollmentLocalPendingIdentitySystem),
+            SubmitAsyncEnrollmentError,
+        > {
+            let cert_ref = self.certificate_reference.clone();
+            Box::pin(pretend_future_is_send_on_web(async move {
+                // 1. Generate a random secret key
+                let key = SecretKey::generate();
+
+                // 2. Encrypt it with the PKI certificate's public key
+                let encrypted = libparsec_platform_pki::encrypt_message(key.as_ref(), &cert_ref)
```
FirelightFlagboy, line 304: **note:** Out-Of-Scope of this PR:

We use the same certificate for both signature & encryption OP, we should improved that to take into consideration the `keyUsage` of the certificate.
(We may need to store 2 cert ref: One ref for signature & one for encryption but they could be equal if the `keyUsage` permit it).

libparsec/src/async_enrollment.rs
```diff
@@ -400,12 +566,312 @@ mod strategy {
...
+                    ))
+                })?;
+
+                let cert_email = cert_info.email().ok_or_else(|| {
+                    AcceptAsyncEnrollmentError::BadSubmitPayload(anyhow::anyhow!(
+                        "Submitter X509 Certificate does not contain an email address"
+                    ))
+                })?;
+
+                if *cert_email != expected_author.to_string() {
```
FirelightFlagboy: **nitpick:** Use `cert_info.emails().find(...)` instead

libparsec/src/async_enrollment.rs
touilleMan: done.

libparsec/src/async_enrollment.rs
touilleMan: done :+1:

libparsec/src/async_enrollment.rs
touilleMan: That's a good point indeed :+1: 

I've created issue https://github.com/Scille/parsec-cloud/issues/12092 describing this need

PR #12090, opened 2026-01-23
OpenBAO list emails for the entity

touilleMan:


libparsec/crates/openbao/src/identity.rs
```diff
@@ -0,0 +1,92 @@
...
+                OpenBaoListEntityEmailsError::BadURL(anyhow::anyhow!(
+                    "URL should not be a cannot-be-a-base"
+                ))
+            })?;
+            path.pop_if_empty();
+            path.push("v1");
+            path.push("identity");
+            path.push("entity");
+            path.push("id");
+            path.push(openbao_entity_id);
```
FirelightFlagboy: **nitpick:** Use [`extend`](https://docs.rs/url/latest/url/struct.PathSegmentsMut.html#method.extend)

touilleMan: done :+1:

PR #12105, opened 2026-01-26
[MS] Replaced checkboxes, fixed default files sorting, fix user list checkbox

Max-7:


client/src/components/workspaces/WorkspaceFilterPopover.vue
Ironicbay, line 1: Texts are now duplicated on this popover 

Image

client/src/components/files/explorer/FileListItem.vue
Ironicbay, line 1: Checkbox hover border is cropped on small display

Image

PR #12109, opened 2026-01-27
feat: enable OpenDocument files in Editics

NicoTuxx:


client/tests/e2e/helpers/types.ts
```diff
@@ -54,6 +54,9 @@ export enum ImportDocuments {
   Docx = 1 << 0,
   Xlsx = 1 << 1,
   Pptx = 1 << 2,
+  Odt = 1 << 9,
```
Max-7: Why not at the end?

Max-7: Move at th end, it's better if the numbers follow each other.

Max-7: I don't think it's necessary to change the tests. You'll have to edit all tests just to change a number, we'll never use them since we're not opening Cryptpad in the tests and you'll have to add files that are basically never going to be used.

Adding some tests to check what types are handled or not handled could be useful but we can do that in another PR and there are better ways to go about it.

Max-7 (approved):

PR #12111, opened 2026-01-28
Move rename out of platform_device_loader to platform_filesystem

### LINKED ISSUE #12009: Rename content
This will be notably used by `archive_device` since it only need to change the file extension of the device file.

```rust
pub async fn rename_file(old: &Path, new: &Path) -> Result<(), RenameFileError> {
  todo!()
}
```

Error:

- StorageNotAvailable
- InvalidParent (Parent is not a folder)
- NotFound (`old` not found, or new parent is missing)
- Internal (other specific errors)

AureliaDolo:
Fix #12009


libparsec/crates/platform_device_loader/src/web/wrapper.rs
FirelightFlagboy, line 1: **question:** What's missing to remove that file?

libparsec/crates/platform_filesystem/src/native/rename.rs
```diff
@@ -1 +1,9 @@
 // Parsec Cloud (https://parsec.cloud) Copyright (c) BUSL-1.1 2016-present Scille SAS
+
+use std::path::Path;
+
+use crate::RenameFileError;
+
+pub async fn rename_file(old: &Path, new: &Path) -> Result<(), RenameFileError> {
+    Ok(tokio::fs::rename(old, new).await?)
```
FirelightFlagboy, line 8: **note:** Like previously, the rename operation does not take care of ensuring that new.parent exist beforehand.

libparsec/crates/platform_device_loader/src/web/wrapper.rs
AureliaDolo: It didn't show up as unused :shrug: 

And I'm removing it here #12118

libparsec/crates/platform_filesystem/src/native/rename.rs
AureliaDolo: with the save I had some trouble because the error was not the same between windows and linux. So that why I had to check.

FirelightFlagboy: NB: My comment is just a note, no change requested here ;)

FirelightFlagboy (approved):

PR #12112, opened 2026-01-28
Document the issue with Snap & Ubuntu 25.04+ in the installation documentation

touilleMan:


docs/userguide/installation.rst
```diff
@@ -61,6 +41,22 @@ Parsec is available for Linux, Mac, and Windows operating systems.
...
+
+        .. code-block:: shell
+
+          sudo tee /etc/apparmor.d/local/fusermount3 << 'EOF'
+          # Allow Parsec packaged as a Snap to communicate with fusermount3 through unix sockets
+          # see https://bugs.launchpad.net/ubuntu/+source/apparmor/+bug/2139081
+          unix (send, receive) peer=(label="snap.parsec.parsec"),
+          EOF
+
+          sudo apparmor_parser -r /etc/apparmor.d/fusermount3
```
FirelightFlagboy, line 58: **typo:** `/etc/apparmor.d/local/fuser..` or `/etc/apparmor.d/fuser...`?

touilleMan: that's not a typo, `/etc/apparmor.d/fusermount3` is the main config that itself loads the one in the local folder if it exists

touilleMan:
> Missing translation

:+1: don.

docs/userguide/installation.rst
```diff
@@ -61,6 +41,22 @@ Parsec is available for Linux, Mac, and Windows operating systems.
...
+        On Ubuntu 25.04 and later, a bug may prevent your workspaces from being exposed in the file system (see `bug report <https://bugs.launchpad.net/ubuntu/+source/apparmor/+bug/2139081>`_).
+
+        As a workaround, you can run the following command while Parsec is not running:
+
+        .. code-block:: shell
+
+          sudo tee /etc/apparmor.d/local/fusermount3 << 'EOF'
+          # Allow Parsec packaged as a Snap to communicate with fusermount3 through unix sockets
+          # see https://bugs.launchpad.net/ubuntu/+source/apparmor/+bug/2139081
+          unix (send, receive) peer=(label="snap.parsec.parsec"),
```
FirelightFlagboy: **question:** Does apparmor allows for wildcard?

I think for users that rename the installed snap (e.g. by adding a version suffix like @mmmarcos)

touilleMan: Good catch !

After further testing I can confirm the name of the install is used in the name of the peer.
So I've added a wildcard as you suggested (i.e. `snap.parsec*.parsec`) 👍

FirelightFlagboy: Does it still work if you install parsec with:

```shell
sudo snap install --classic parsec parsec_v3
```

?

FirelightFlagboy:
CF question above, before merging

docs/userguide/installation.rst
FirelightFlagboy: With an internal discussion, the wildcard seems to also allows for renamed app

PR #12117, opened 2026-01-29
Fix OpenBao route pattern used to sign/verify with a given OpenBao entity

touilleMan:


touilleMan:
> LGTM but missing the context, why do we need to use entity-.. instead of user-.. ?

We want each OpenBao entity to only have access to its own key, so to do that we could configure OpenBao server with just `transit/keys/{{identity.entity.id}}`.

However the entity ID is a UUID so it's not very clear that e.g. `v1/transit/keys/217b7a03-b4d0-aff6-eaa8-4e1aa0573193`  is the key for the entity with ID `217b7a03-b4d0-aff6-eaa8-4e1aa0573193`. Hence the need for a suffix.

The suffix was `user-` initially, but got replaced by `entity-` since we use the OpenBao entity ID. But unfortunately we forgot to change this everywhere...

PR #12134, opened 2026-02-02
Replace `submitter_forget_async_enrollment` by `submitter_cancel_async_enrollment`

touilleMan:


FirelightFlagboy (changes requested):
Need to discuss the API first in a RFC

libparsec/crates/client/src/invite/async_enrollment/submitter_cancel.rs
```diff
@@ -0,0 +1,79 @@
...
+            // We basically ignore the server answer and go idempotent here:
+            // - Most likely case is `enrollment_id` has been obtained by listing
+            //   the pending enrollment's local files. In this case we have to
+            //   remove our enrollment's local file no matter what.
+            // - If the enrollment ID is a dummy one, the server returns a not found,
+            //   but so will `remove_pending_async_enrollment()`.
+            // - Last case is a very weird one: the enrollment ID is a valid one (hence
+            //   the server cancels it), but it doesn't exist locally. This leads us
+            //   to return a not found error without informing the user the enrollment
+            //   has been cancelled, but we don't really care about this given how
```
AureliaDolo, line 49: Could this lead to issues if the client was able to get someone else's enrollment ID and then attempts to cancels it ?

Max-7: @touilleMan mentionned that this id was not public: only the user, the org admin and the server admin have access to it.

AureliaDolo: So if someone has  fraudulent access to this id, that means that either the admin's or Alice's device is compromised ?

libparsec/crates/protocol/schema/anonymous_cmds/async_enrollment_cancel.json5
```diff
@@ -0,0 +1,32 @@
...
+        "req": {
+            "fields": [
+                {
+                    // The enrollment ID to cancel
+                    "name": "enrollment_id",
+                    "type": "AsyncEnrollmentID"
+                }
+            ]
+        },
+        "reps": [
```
FirelightFlagboy, line 16: **question:** Given how we use response status, do we really need anything other than `OK`?

touilleMan: I copied the statuses from the info command so we don't add new oracle here
Regarding if those statuses are needed, I'd rather keep them given they don't cost much in the codebase and it's better to have the GUI decoupled from the API (as we may want to provide more precise errors in the future, for instance for the CLI)

libparsec/crates/client/src/invite/async_enrollment/submitter_cancel.rs
touilleMan: You description in the RFC's update is spot on 👍

FirelightFlagboy (approved):

PR #12142, opened 2026-02-03
Add cancel cmd to rfc 1023 async enrollment.

AureliaDolo:
In preparation for #12134

### DIFF
--- a/docs/rfcs/1023-async-enrollment.md
+++ b/docs/rfcs/1023-async-enrollment.md
@@ -75,13 +75,8 @@ This is notably to reflect two fundamental differences:
     enrollment.
 
 > [!NOTE]
-> We don't provide a clean way for Alice to cancel her enrollment submission after step 5
-> (as we consider this rather unusual).
-> So instead Alice is only expected to remove the enrollment info from her machine's filesystem
-> (hence there is no longer an enrollment pending from her point of view), and an admin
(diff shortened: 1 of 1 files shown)


docs/rfcs/1023-async-enrollment.md
```diff
@@ -75,13 +75,10 @@ This is notably to reflect two fundamental differences:
...
-> (as we consider this rather unusual).
-> So instead Alice is only expected to remove the enrollment info from her machine's filesystem
-> (hence there is no longer an enrollment pending from her point of view), and an admin
-> should manually cancel the enrollment (but accepting it by mistake is not an issue either,
-> since Alice no longer knows about her user & device secret keys, so the admin will just
-> have to remove this phantom Alice user first before the new enrollment can take place).
+> Alice can cancel her request while her enrollment has not been accepted.
+> For that she needs only her enrollment ID. That means that anyone with
+> that enrollment ID could cancel Alice's request without her knowledge.
+> This is considered as acceptable because enrollment Id is a UUID.
```
FirelightFlagboy: **suggestion:** Few suggestions:

- I think the paragraph should be moved out of a note: It's a proper element of the API not a detail.
- I do not like the phrasing "It's OK because it's a UUID" we had a reflection about it, that should be written:
  - Why it's OK to only use the enrollmentID
  - What are the limitation of only using an UUID
  - Why we accept those limitation
 
  > That being written would help to justify during audit

PR #12151, opened 2026-02-05
[MS] Create folder move/copy modal

### LINKED ISSUE #7524: Add possibility to create folder on Copy/Move modal
Image

fabienSvtr:
https://github.com/user-attachments/assets/eae139fe-a947-478e-a95d-3c04e53eaf05

closes #7524


client/src/locales/en-US.json
```diff
@@ -764,6 +764,7 @@
         "itemSelectedCount": "No selected items | 1 selected item | {count} selected items",
         "emptyFolder": "This folder is empty.",
         "loading": "Loading the folder...",
+        "newFolder": "New folder | New folder ({count})",
```
Ironicbay: This locale is not used, you're using `SideMenu.tabbar.newFolder`, and the `count` variable doesn't seem necessary either

client/src/locales/fr-FR.json
```diff
@@ -763,6 +763,7 @@
         "itemSelectedCount": "Aucun élément sélectionné | 1 élément sélectionné | {count} éléments sélectionnés",
         "emptyFolder": "Ce dossier est vide.",
         "loading": "Chargement du dossier...",
+        "newFolder": "Nouveau dossier | Nouveau dossier ({count})",
```
Ironicbay: Same thing

client/src/components/files/explorer/FolderSelectionModal.vue
```diff
@@ -68,14 +79,44 @@
...
+          <ion-button
+            fill="clear"
+            @click="confirmNewFolder()"
+            class="new-folder-confirm"
+          >
+            <ion-icon
+              :icon="checkmark"
+              class="button-icon"
+            />
+          </ion-button>
```
Ironicbay, line 119: This part should be tweaked to match the actual behaviour and add some more function: 
- The ✔️ button should only be visible or enabled when `newFolderName.length > 0`. Currently, if a user validates with an empty folder name the operation is cancelled, whereas the user could expect the placeholder name to be applied (`New folder`)
- There should be a ❌ button to explicitly cancel the new folder operation anytime

client/src/components/files/explorer/FolderSelectionModal.vue
```diff
@@ -164,6 +247,61 @@ const topbarWidthWatchCancel = watch([windowWidth, pathLength], () => {
...
+  if (folderListRef.value) {
+    await nextTick();
+    folderListRef.value.scrollTop = 0;
+  }
+}
+
+async function confirmNewFolder(): Promise<void> {
+  if (isCreatingFolder.value) {
+    isCreatingFolder.value = true;
+  }
```
Ironicbay: `if (isCreatingFolder.value)` is `true` only if `isCreatingFolder.value === true`

client/src/components/files/explorer/FolderSelectionModal.vue
```diff
@@ -164,6 +247,61 @@ const topbarWidthWatchCancel = watch([windowWidth, pathLength], () => {
...
+  const result = await createFolder(workspaceInfo.value.handle, path);
+
+  if (!result.ok) {
+    switch (result.error.tag) {
+      case WorkspaceCreateFolderErrorTag.EntryExists: {
+        error.value = { key: 'FoldersPage.errors.createFolderAlreadyExists', data: { name: trimmedName } };
+        break;
+      }
+      default:
+        cancelNewFolder();
```
Ironicbay, line 286: Add error handling in the default case: the user should be informed that something went wrong instead of silently cancelling the operation. You could set `error.value` to a generic error text, and remove the call to `cancelNewFolder()` in this default case if you choose to add a ❌ cancel button as I suggested in a previous comment.

client/src/components/files/explorer/FolderSelectionModal.vue
```diff
@@ -164,6 +247,61 @@ const topbarWidthWatchCancel = watch([windowWidth, pathLength], () => {
...
+    }
+  } else {
+    await update();
+    const entry = currentEntries.value.find((e) => e[0]?.name === trimmedName)?.[0] as EntryStat;
+    await enterFolder(entry);
```
Ironicbay, line 291: I would remove this automatic navigation to have it behave like any folder creation operation: the user remains in the base folder and is then free to choose to navigate there. Forcing the navigation could be jarring.

client/src/components/files/explorer/FolderSelectionModal.vue
```diff
@@ -164,6 +247,61 @@ const topbarWidthWatchCancel = watch([windowWidth, pathLength], () => {
...
+    isCreatingFolder.value = false;
+  }
+}
+
+function onNewFolderNameChange(): Promise<void> {
+  newFolderName.value.trim();
+  error.value = '';
+  return Promise.resolve();
```
Ironicbay: Make the function `async` and remove this line

client/src/components/files/explorer/FolderSelectionModal.vue
```diff
@@ -185,7 +323,7 @@ async function update(): Promise<void> {
   if (result.ok) {
     const newEntries: [EntryStat, boolean][] = [];
     const filteredEntries = result.value
-      .filter((entry) => !entry.isConfined())
+      .filter((entry) => !entry.isConfined() && !entry.isFile())
```
Ironicbay: Why filtering out files there? Having files displayed can help navigating the workspace, plus it makes the following check redundant

client/src/components/files/explorer/FolderSelectionModal.vue
```diff
@@ -68,14 +79,44 @@
...
+            :image="Folder"
+            class="new-folder-image"
+          />
+          <ms-input
+            v-model="newFolderName"
+            @update:model-value="onNewFolderNameChange"
+            ref="newFolderInput"
+            placeholder="SideMenu.tabbar.newFolder"
+            @keydown.enter.prevent.stop="confirmNewFolder()"
+            @keydown.esc.prevent="cancelNewFolder()"
```
Ironicbay: The encapsulating modal has `canDismiss: true` and is already dismissed when pressing `esc`. So if the user tries to press this key when setting a new folder name, everything closes and they are brought back to the folders page.

client/src/components/files/explorer/FolderSelectionModal.vue
```diff
@@ -258,16 +396,15 @@ async function back(): Promise<void> {
 
 async function onPathChange(node: RouterPathNode): Promise<void> {
   forwardStack.splice(0, forwardStack.length);
+  isCreatingFolder.value = false;
```
Ironicbay: Any reason to set this instead of calling `cancelNewFolder()` as you did in the function `enterFolder()`? Just for consistency ^^

client/src/components/files/explorer/FolderSelectionModal.vue
fabienSvtr: It overwhelms the list, whether the user has several files and folders, it's clearer to have only the folders.

client/src/components/files/explorer/FolderSelectionModal.vue
fabienSvtr: Yes, I removed it.

client/src/components/files/explorer/FolderSelectionModal.vue
fabienSvtr: No, I updated it! ^^

client/src/components/files/explorer/FolderSelectionModal.vue
fabienSvtr: I'm not very keen on this idea. The user will probably want to create a folder to move/copy items to. If they want to create a folder without navigating to it, they will do so directly on the folders/files page.

client/src/components/files/explorer/FolderSelectionModal.vue
fabienSvtr: - No, since the ✔️ button can create a new folder with "New folder" as a default name.
- The ❌ button is unnecessary, the user can still navigates in folders or move/copy even if the new folder is on process

NicoTuxx: After trying different apps, it seems very common to name the folder automatically "New folder" if `newFolderName.length === 0` instead of cancelling the request (which can be really annoying btw).

For the ❌ I'm unsure, it depends on how we see the feature: 
- for cases like the one in this PR, where the folder creation is directly integrated in the file list, the moment you clicked the "New folder" button, the operation is complete & just let you rename it
- for cases with a cancel button, the presentation is more with a modal or bottomsheet with cancel/submit buttons apart from the file list.

I would stick to what we have here, without a ❌ button at all.

client/src/components/files/explorer/FolderSelectionModal.vue
NicoTuxx: I'm not sure the cancelling is that much important, it's just a folder creation, for me we could just show the error, without cancelling at all.

client/src/components/files/explorer/FolderSelectionModal.vue
NicoTuxx: After trying some apps, when you're in the specific case of a copy/move modal, the default behavior is more often to move the user automatically to the created folder, I would keep it that way.

client/src/components/files/explorer/FolderSelectionModal.vue
NicoTuxx: It's easier to know where you are if you see your files too, you could make a compromise like Google Drive, by showing them but with a disabled/greyed style.

client/src/components/files/explorer/FolderSelectionModal.vue
```diff
@@ -70,14 +81,43 @@
...
+        >
+          <ms-image
+            :image="Folder"
+            class="new-folder-image"
+          />
+          <ms-input
+            v-model="newFolderName"
+            @update:model-value="onNewFolderNameChange"
+            ref="newFolderInput"
+            placeholder="SideMenu.tabbar.newFolder"
```
Ironicbay:
```suggestion
            placeholder="FoldersPage.createFolder"
```

client/src/components/files/explorer/FolderSelectionModal.vue
```diff
@@ -166,6 +248,62 @@ const topbarWidthWatchCancel = watch([windowWidth, pathLength], () => {
...
+async function confirmNewFolder(): Promise<void> {
+  let trimmedName = newFolderName.value.trim();
+
+  if (!workspaceInfo.value) {
+    cancelNewFolder();
+    return;
+  }
+
+  if (trimmedName.length === 0) {
+    trimmedName = I18n.translate('FoldersPage.newFolder');
```
Ironicbay:
```suggestion
    trimmedName = I18n.translate('FoldersPage.createFolder');
```

client/src/components/files/explorer/FolderSelectionModal.vue
```diff
@@ -166,6 +248,62 @@ const topbarWidthWatchCancel = watch([windowWidth, pathLength], () => {
...
+  }
+
+  const path = await Path.join(selectedPath.value, trimmedName);
+  const result = await createFolder(workspaceInfo.value.handle, path);
+  console.log('Create folder result', trimmedName);
```
Ironicbay: Remove this log

client/src/locales/en-US.json
```diff
@@ -764,6 +764,7 @@
         "itemSelectedCount": "No selected items | 1 selected item | {count} selected items",
         "emptyFolder": "This folder is empty.",
         "loading": "Loading the folder...",
+        "newFolder": "New folder",
```
Ironicbay: You can remove this locale if you choose to use `FoldersPage.createFolder`

client/src/locales/fr-FR.json
```diff
@@ -763,6 +763,7 @@
         "itemSelectedCount": "Aucun élément sélectionné | 1 élément sélectionné | {count} éléments sélectionnés",
         "emptyFolder": "Ce dossier est vide.",
         "loading": "Chargement du dossier...",
+        "newFolder": "Nouveau dossier",
```
Ironicbay: Same

Issue #11955, opened 2025-12-18
Cleanup filesystem errors between native & web


AureliaDolo:
to remove deps on `error_set`

FirelightFlagboy:
Will be done during the sub-tasks of #11954

AureliaDolo:
Nit: consider returning an error here https://github.com/Scille/parsec-cloud/pull/12091#discussion_r2736341829 (but that would change the behavior

AureliaDolo:
Previous PR was a partial fix.

PR #12153, opened 2026-02-05
[platform_device_loader] rework errors.

AureliaDolo:
Part of #11955

### DIFF
--- a/libparsec/crates/platform_device_loader/src/pki.rs
+++ b/libparsec/crates/platform_device_loader/src/pki.rs
@@ -12,10 +12,29 @@ use crate::{get_local_pending_dir, LOCAL_PENDING_EXT};
 
 #[derive(Debug, thiserror::Error)]
 pub enum SavePkiLocalPendingError {
-    #[error(transparent)]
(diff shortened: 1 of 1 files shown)


libparsec/crates/platform_device_loader/src/device/list.rs
```diff
@@ -12,6 +12,7 @@ use crate::{get_devices_dir, load_available_device, AvailableDevice, DEVICE_FILE
 
 #[derive(Debug, thiserror::Error)]
 pub enum ListAvailableDeviceError {
+    // No space left cannot occur
     #[error("Device storage is not available")]
     StorageNotAvailable,
```
FirelightFlagboy: **question:** Why not removing the variant and using `unreachable` then?

AureliaDolo: Because the web specific storage not available could still happen.

Should I move that to internal ?

FirelightFlagboy: So the comment is missleading

FirelightFlagboy (approved):

PR #12172, opened 2026-02-09
Erase organization command in server CLI

touilleMan:


docs/hosting/administration/erase_organization.rst
```diff
@@ -0,0 +1,88 @@
...
+Where & how the data are stored
+-------------------------------
+
+For any given organization, data are split as follow:
+
+- A PostgreSQL database containing the certificates (e.g. user/devices/workspaces) and
+  encrypted workspaces metadata (e.g. content of folder, list of blocks for each file version).
+- A blockstore (e.g. S3) containing the blocks (i.e. encrypted pieces of data that compose the files).
+- On top of that, each Parsec client having access to the organization has an encrypted local database
+  containing a copy of the certificates, metadata for the workspaces its has access to, and a subset
```
AureliaDolo:
```suggestion
  containing a copy of the certificates, metadata for the workspaces it has access to, and a subset
```

docs/hosting/administration/erase_organization.rst
```diff
@@ -0,0 +1,88 @@
...
+needlessly occupied space.
+
+This can be done by manually removing from the blockstore the top level directory named after the organization.
+For example, if the organization was named ``CoolOrg``, remove the ``CoolOrg/`` prefix from the bucket.
+
+.. note::
+
+   Blockstores have their own backup strategy. Typically AWS S3 allows for a bucket to
+   have an history so that a data removal can be cancelled.
+   You should pay attention to this to ensure the blocks have actually been removed.
```
AureliaDolo, line 74: Do we expect them to be removed from the history too ?

docs/hosting/administration/erase_organization.rst
```diff
@@ -0,0 +1,88 @@
...
+Step 3: Clients cleanup
+-----------------------
+
+From a Parsec client point of view, having the organization erased from the server is
+equivalent to being offline (i.e. the client cannot connect to the server).
+For this reason, an end-user is still able to use his Parsec client to work on the
+organization using the local cache (e.g. creating a new file in a workspace or
+reading an existing file that is in cache).
+
+To prevent the users from accessing the local cache, the local configuration and data should be manually removed:
```
AureliaDolo, line 88: by the user themselves ?

docs/hosting/administration/erase_organization.rst
```diff
@@ -0,0 +1,88 @@
...
+
+1. Removing everything (certificates & metadata) related to the organization in the PostgreSQL database.
+2. Removing the blocks from the blockstore.
+3. Removing the remaining data from the clients.
+
+Step 1: remove from PostgreSQL
+------------------------------
+
+Erasing an organization from Parsec is both an uncommon and a (obviously !) a destructive operation.
+As such it is not available from the Administration API but instead must be triggered from the server CLI directly.
```
AureliaDolo, line 35: It's maybe an idea for another day, but maybe it could be worth to propose an option to configure a minimum number of admins to trigger the deletion (with a shamir) to avoid a costly mistake.

docs/hosting/administration/erase_organization.rst
```diff
@@ -0,0 +1,88 @@
...
+
+.. code-block:: bash
+
+   # On Parsec server
+   parsec erase_organization --organization <OrgName> --db <database_url>
+
+.. warning::
+
+   This operation cannot be undone. Make sure you have a backup of any data you may need before
```
AureliaDolo, line 44: nit: maybe add the command to make a backup in the warning message ?

docs/hosting/administration/erase_organization.rst
```diff
@@ -0,0 +1,88 @@
...
+   proceeding.
+
+.. note::
+
+   After erasing an organization, it is possible to create a new organization with the same name
```
AureliaDolo, line 49: What would be the expected error on the client side if they haven't deleted their local data and a new organization with the same name is created ? 

If their parsec client would just look offline that may be confusing if the same entity re creates an organization and expect to enroll the same people.

server/parsec/cli/erase_organization.py
```diff
@@ -0,0 +1,139 @@
...
+        f"You are about to entirely erase the {display_org} organization from the database, this action cannot be undone."
+    )
+
+    display_bucket_path = click.style(f"{organization_id.str}/", fg="yellow")
+    click.echo("Notes:")
+    click.echo(
+        "- No trace of the organization will remain, so it will be possible to re-create another organization with the same name."
+    )
+    click.echo(
+        f"- The organization's blocks won't be erased from the blockstore, you should instead manually remove the {display_bucket_path} top level directory from it."
```
AureliaDolo:
```suggestion
        f"- The organization's blocks won't be erased from the blockstore, you should manually remove the {display_bucket_path} top level directory from it."
```

docs/hosting/administration/erase_organization.rst
```diff
@@ -0,0 +1,88 @@
...
+In practice this means:
+
+1. Removing everything (certificates & metadata) related to the organization in the PostgreSQL database.
+2. Removing the blocks from the blockstore.
+3. Removing the remaining data from the clients.
+
+Step 1: remove from PostgreSQL
+------------------------------
+
+Erasing an organization from Parsec is both an uncommon and a (obviously !) a destructive operation.
```
FirelightFlagboy: **typo:** 
```suggestion
Erasing an organization from Parsec is both an uncommon and a (obviously !) destructive operation.
```

server/tests/cli/test_erase_organization.py
```diff
@@ -0,0 +1,48 @@
...
+        lambda: runner.invoke(
+            cli,
+            "erase_organization --organization MinimalOrgTemplate --db MOCKED --with-testbed minimal",
+            input="MinimalOrgTemplate\n",
+            env={"DEBUG": "1"},
+        )
+    )
+    assert result.exit_code == 0, result.output
+    assert "Removing from database..." in result.output
+
```
AureliaDolo, line 22: Maybe add check that you can re create an org with the same name ?

Edit: or not as it's tested below

server/parsec/cli/erase_organization.py
```diff
@@ -0,0 +1,139 @@
...
+        "--dev",
+        cls=DevOption,
+        is_flag=True,
+        is_eager=True,
+        help=(
+            "Equivalent to `--debug --db=MOCKED --with-testbed=coolorg --organization CoolorgOrgTemplate`"
+        ),
+    )
+)
+def erase_organization(
```
FirelightFlagboy, line 68: **nitpick:** Can you add a flag to remove block in the blockstore?

As you've said in the document, the block data is encrypted and with the decryption key lost, that data is not usable.

But that data is still billable, I would like an option that remove the blockstore data alongside the organization in the database.

docs/hosting/administration/erase_organization.rst
touilleMan: :+1: done.

docs/hosting/administration/erase_organization.rst
touilleMan: This can also be done by an administrator that has remote access to the machine. Here we don't specify *how* things should be done but *what* should be done.

docs/hosting/administration/erase_organization.rst
touilleMan: This is overkill: the administrator is supposed to have a backup strategy for PostgreSQL. So if something bad happen, he should be able to do a rollback.
(and obviously the administrator has to take into account the backup strategy when he wants to permanently remove an organization, but this is outside of the scope of Parsec)

docs/hosting/administration/erase_organization.rst
touilleMan: Again, the administrator is responsible for the backup strategy, so we cannot provide a one-size-fits-all command here (typically "backup any  data" may refer to do a PostgreSQL snapshot and storing it on cold storage, or it may refer to simply have a user start a Parsec client and copy/paste the files still needed)

server/parsec/cli/erase_organization.py
touilleMan: :+1: done.

server/parsec/cli/erase_organization.py
touilleMan: I consider adding such flag, but decided otherwise:

If the blockstore is correctly configured (typically S3 bucket with history enabled), the removal of blocks using the S3 API shouldn't lead to an actual lost of data.

This is important since by construction a S3 bucket doesn't need backups (as you have a 99.999... guaranteed to retrieve your data), so the only thing you have to protect against is accidental deletion/overwrite (due to bug or attacker taking control of the server).

(also note access control in S3 doesn't differentiate between CREATE/OVERWRITE/DELETE of a block)

And, on the other hand, if the administrator has chosen to backup his blockstore (to protect against avoid OVH-style catastrophic datacenter fire :trollface: ), then by construction using S3 API still doesn't lead to actual lost of data (the administrator also has to manually remove data from the backup)

So in the end I think it's just better to let the administrator handle the removal manually.

docs/hosting/administration/erase_organization.rst
touilleMan: The goal here is to permanently erase all data from a given organization, so we expect the administrator to remove the history.
However this is not Parsec responsibility since blockstore configuration is specific to the administrator needs (see https://github.com/Scille/parsec-cloud/pull/12172#discussion_r2786688237).

docs/hosting/administration/erase_organization.rst
touilleMan: :+1: done.

docs/hosting/administration/erase_organization.rst
touilleMan: The client will get an error complaining the organization is not found on the server (even if a new organization with the same name is created, since they don't have the same root verify key)

I've added a word about this in the note :+1:

docs/hosting/administration/erase_organization.rst
```diff
@@ -0,0 +1,98 @@
+.. Parsec Cloud (https://parsec.cloud) Copyright (c) BUSL-1.1 2016-present Scille SAS
+
+.. _doc_hosting_erase_organization:
+
+Erase an organization
+=====================
+
+Where & how the data are stored
+-------------------------------
```
mmmarcos, line 9:
```suggestion
Where is the organization's data stored?
----------------------------------------
```

docs/hosting/administration/erase_organization.rst
```diff
@@ -0,0 +1,98 @@
...
+
+For any given organization, data are split as follow:
+
+- A PostgreSQL database containing the certificates (e.g. user/devices/workspaces) and
+  encrypted workspaces metadata (e.g. content of folder, list of blocks for each file version).
+- A blockstore (e.g. S3) containing the blocks (i.e. encrypted pieces of data that compose the files).
+- On top of that, each Parsec client having access to the organization has an encrypted local database
+  containing a copy of the certificates, metadata for the workspaces it has access to, and a subset
+  of the blocks (depending on local cache configuration).
```
mmmarcos, line 18:
```suggestion
- A PostgreSQL database containing certificates (e.g. users, devices, workspaces) and
  encrypted workspace metadata (e.g. folder contents, block list for each file version).
- An S3 blockstore containing the encrypted file blocks.
- On each Parsec client having access to the organization: An encrypted local database
  containing a copy of the certificates, workspace metadata it has access to, and a subset
  of the encrypted file blocks (depending on local cache configuration).
```

docs/hosting/administration/erase_organization.rst
```diff
@@ -0,0 +1,98 @@
...
+
+Erasing data
+------------
+
+When no longer in use (or for legal reason) an organization can be erased from the Parsec server.
+
+In practice this means:
```
mmmarcos, line 25:
```suggestion
When no longer in use (or for legal reason) an organization can be erased from the Parsec server
following these steps:
```

docs/hosting/administration/erase_organization.rst
```diff
@@ -0,0 +1,98 @@
...
+
+1. Removing everything (certificates & metadata) related to the organization in the PostgreSQL database.
+2. Removing the blocks from the blockstore.
+3. Removing the remaining data from the clients.
```
mmmarcos, line 29: I'd use "delete" (or erase) since "remove" may indicate persistence.
```suggestion
1. Delete certificates & metadata from the PostgreSQL database.
2. Delete the encrypted file blocks from the blockstore.
3. Delete remaining local data from Parsec clients.
```

docs/hosting/administration/erase_organization.rst
```diff
@@ -0,0 +1,98 @@
...
+
+Step 1: remove from PostgreSQL
+------------------------------
```
mmmarcos, line 32: I'd make these steps (1, 2, 3) subsections of "Erasing data"

```suggestion
```
```suggestion
Step 1: Delete certificates from PostgreSQL
^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```

docs/hosting/administration/erase_organization.rst
```diff
@@ -0,0 +1,98 @@
...
+
+Erasing an organization from Parsec is both an uncommon and (obviously !) a destructive operation.
```
mmmarcos, line 34:
```suggestion
Erasing an organization from Parsec is both an uncommon and (obviously !) destructive operation.
```

docs/hosting/administration/erase_organization.rst
```diff
@@ -0,0 +1,98 @@
...
+As such it is not available from the Administration API but instead must be triggered from the server CLI directly.
+
+.. code-block:: bash
+
+   # On Parsec server
+   parsec erase_organization --organization <OrgName> --db <database_url>
```
mmmarcos, line 40:
```suggestion
   parsec erase_organization --organization <organization_name> --db <database_url>
```

docs/hosting/administration/erase_organization.rst
mmmarcos: There is already a section describing how to make a backup, we should link to it. See my next comment/suggestion.

docs/hosting/administration/erase_organization.rst
```diff
@@ -0,0 +1,98 @@
...
+
+.. warning::
+
+   This operation cannot be undone. Make sure you have a backup of any data you may need before
+   proceeding.
```
mmmarcos, line 45:
```suggestion
   This operation cannot be undone. Make sure you have a backup before proceeding. 
   See the :ref:`Backup and Restore section <doc_hosting_maintenance_backup_restore>`.
```

docs/hosting/administration/erase_organization.rst
```diff
@@ -0,0 +1,98 @@
...
+
+   After erasing an organization, it is possible to create a new organization with the same name
+   since no trace of the previous one remains.
+
+   Even if they share the same name, the erased and the new organization are strictly unrelated
+   since they have a different root verify key (i.e. root key used to verify all certificates
+   in the organization).
+
+   Typically this means a Parsec client trying to access the erased organization will complain
+   it doesn't exist on the server even if a new organization with the same name exist.
```
mmmarcos, line 57:
```suggestion
   Even if they share the same name, the organizations are strictly unrelated
   since they have a different root verify key (used to verify the organization certificates). 
   If a Parsec client tries to access the erased organization it will complain
   that it doesn't exist on the server even if a new organization with the same name exist.
```

docs/hosting/administration/erase_organization.rst
```diff
@@ -0,0 +1,98 @@
...
+
+Step 2: Blockstore cleanup
+--------------------------
```
mmmarcos, line 60:
```suggestion
Step 2: Delete blocks from the Blockstore
^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```

docs/hosting/administration/erase_organization.rst
```diff
@@ -0,0 +1,98 @@
...
+
+Once step 1 done, the blocks' decryption keys has been lost. In other words, everything
+stored in the blockstore and related to the organization is irrecoverable.
+Hence removing those data from the blockstore should be seen as an optional step to reclaim
```
mmmarcos, line 64:
```suggestion
stored in the blockstore related to the organization is irrecoverable.
Hence removing those blocks can be seen as an optional step to reclaim
```

docs/hosting/administration/erase_organization.rst
mmmarcos: ⚠️ I think we should move up this whole note from the step 1, maybe to the first or second section. 

Otherwise one may think that it is OK to create an org with same name _just after_ step 1 (postgresql), _before_ step 2 (blockstore cleanup).

IMO, regardless of Parsec internals, the erasing procedure should be treated almost as a" transaction", step 1 to 3 should be performed as a whole before attempting to create a new organization with same name.

docs/hosting/administration/erase_organization.rst
```diff
@@ -0,0 +1,98 @@
...
+For example, if the organization was named ``CoolOrg``, remove the ``CoolOrg/`` prefix from the bucket.
+
+.. note::
+
+   Blockstores have their own backup strategy. Typically AWS S3 allows for a bucket to
+   have an history so that a data removal can be cancelled.
+   You should pay attention to this to ensure the blocks have actually been removed.
+
+Step 3: Clients cleanup
+-----------------------
```
mmmarcos, line 77:
```suggestion
Step 3: delete local data from Parsec clients
^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```

docs/hosting/administration/erase_organization.rst
```diff
@@ -0,0 +1,98 @@
...
+
+Once the organization erased from the server, the Parsec client will display an error
+about the organization not being found on the server.
+However the client can still work in offline mode (if the server is not reachable, the
+client cannot know the organization has been erased from the server!),
+
+For this reason, an end-user is still able to use his Parsec client to work on the
+organization using the local cache (e.g. creating a new file in a workspace or
+reading an existing file that is in cache).
```
mmmarcos, line 86:
```suggestion
For this reason, users are still able to use the Parsec client to work on the
organization for actions relying on the local cache such as importing a new file
or reading an existing file that is in cache.
```

docs/hosting/administration/erase_organization.rst
```diff
@@ -0,0 +1,98 @@
...
+
+To prevent the users from accessing the local cache, the local configuration and data should be manually removed:
+
+- Linux:
+   - config: ``$XDG_DATA_HOME or $HOME/.local/share/parsec3/<device_id>`` (e.g. ``/home/alice/.local/share/parsec3/e68b7131394749a4bbd279bd087e6ae6``)
+   - data: ``$XDG_CONFIG_HOME or $HOME/.config/parsec3/libparsec/devices/<device_id>`` (e.g. ``/home/alice/.config/parsec3/libparsec/devices/e68b7131394749a4bbd279bd087e6ae6``)
```
mmmarcos, line 92:
```suggestion
   - config: ``$XDG_DATA_HOME`` or ``$HOME/.local/share/parsec3/<device_id>`` (e.g. ``/home/alice/.local/share/parsec3/e68b7131394749a4bbd279bd087e6ae6``)
   - data: ``$XDG_CONFIG_HOME`` or ``$HOME/.config/parsec3/libparsec/devices/<device_id>`` (e.g. ``/home/alice/.config/parsec3/libparsec/devices/e68b7131394749a4bbd279bd087e6ae6``)
```

docs/hosting/administration/erase_organization.rst
mmmarcos: I'd rephrase it like this:
```suggestion
The local cache can be safely erased by manually deleting the local configuration and data for the device having access to the organization:
```

IMO we should add a note or warning to explain how to identify which device is linked to the organization (device id is displayed in the "My Devices" when logged-in... maybe we should consider displaying it in the Home page as tooltip or detail button ?)

mmmarcos (commented):
Some suggestions regarding the docs, mostly about wording or rephrasing.

I didn't review the code.

docs/hosting/administration/erase_organization.rst
mmmarcos: Actually, in other sections of the docs we suggest using the CLI for that, e.g.:

> The ``<DEVICE_ID>`` mentioned below refer to the device used to access the 
> organization. You can find out which is your device ID by running:
> 
>   .. code-block:: bash
> 
>     parsec-cli device list

PR #12173, opened 2026-02-09
feat(pki): Make `encrypt_message` platform agnostic

### LINKED ISSUE #12140: PKI Make `encrypt_message` platform agnostic
`encrypt` message only need to use the public key of the certificate and doesn't need to manipulate the private key.

FirelightFlagboy:
Since we only need the public key from the certificate to perform the encryption

Closes #12140

### DIFF
--- a/libparsec/crates/platform_device_loader/src/native/mod.rs
+++ b/libparsec/crates/platform_device_loader/src/native/mod.rs
@@ -96,7 +96,10 @@ pub(crate) async fn save_device_pki(
     let secret_key = SecretKey::generate();
 
     // Encrypt the key using the public key related to a certificate from the store
-    let (algorithm, encrypted_key) = encrypt_message(secret_key.as_ref(), certificate_ref)
+    let der = libparsec_platform_pki::get_der_encoded_certificate(certificate_ref)
+        .await
+        .map_err(|e| SaveDeviceError::Internal(e.into()))?;
+    let (algorithm, encrypted_key) = encrypt_message(der.as_ref(), secret_key.as_ref())
         .await
         .map_err(|e| SaveDeviceError::Internal(e.into()))?;
 
--- a/libparsec/crates/platform_pki/src/x509/mod.rs
+++ b/libparsec/crates/platform_pki/src/x509/mod.rs
@@ -3,8 +3,9 @@
 mod distinguished_name;
 mod extensions;
 
+use rsa::pkcs8::DecodePublicKey;
 use x509_cert::{
-    der::{Decode, Error as DERError, SliceReader},
+    der::{Decode, Encode, Error as DERError, SliceReader},
     Version,
 };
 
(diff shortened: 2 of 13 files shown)


libparsec/crates/platform_pki/src/x509/mod.rs
```diff
@@ -121,3 +122,38 @@ impl TryFrom<x509_cert::Certificate> for X509CertificateInformation {
...
+pub(crate) fn get_certificate_public_key(
+    raw_cert: &[u8],
+) -> Result<PublicKey, GetCertificatePublicKeyError> {
+    let spki = x509_cert::Certificate::decode(&mut SliceReader::new(raw_cert)?)
+        .map(|v| v.tbs_certificate.subject_public_key_info)?;
+
+    match spki.algorithm.oid {
+        x509_cert::der::oid::db::rfc5912::RSA_ENCRYPTION => {
+            // TODO: Use RsaPublicKey::try_from(spki::SubjectPublicKeyInfo) once rsa is update and
+            // use the same version of spki.
```
AureliaDolo, line 145: Is there an issue ?

libparsec/crates/platform_device_loader/src/native/mod.rs
```diff
@@ -90,7 +90,10 @@ pub(crate) async fn save_device_pki(
     let secret_key = SecretKey::generate();
 
     // Encrypt the key using the public key related to a certificate from the store
-    let (algorithm, encrypted_key) = encrypt_message(secret_key.as_ref(), certificate_ref)
+    let der = libparsec_platform_pki::get_der_encoded_certificate(certificate_ref)
```
touilleMan, line 99: Why change `encrypt_message` signature ?
This makes the code more complex since the caller has to call two functions (and handle two errors types), and providing a certificate object instead of a reference doesn't seem to have a benefit since the `get_der_encoded_certificate` is always called before `encrypt_message` 

The only gain I see is in the tests where we can test the encryption without having to do the reference-to-certificate resolution.
However I don't think this is worth it:
- As said before, this increase complexity of the exposed API
- This kind of test can also be achieved by calling an internal method (this is part of the reasons we only use unit tests after all ^^). Typically `encrypt_message`'s body is split into two private functions call `get_der_encoded_certificate` and `do_encrypt_message` and the test calls `do_encrypt_message`
- Last but not least, testing only encryption without reference-to-certificate resolution is only a temporary test due to the fact we don't support reference-to-certificate outside of windows yet. So I'd rather keep the test doing full encryption + reference-to-certificate resolution as is, mark it windows only for the moment, and have another temporary test on just `do_encrypt_message` for the other platforms

FirelightFlagboy: That change is required for the integration of Pki Web, I plan to simplify the exposed API but I need to streamlined thing first

libparsec/crates/platform_pki/src/x509/mod.rs
FirelightFlagboy: I've update the message

PR #12175, opened 2026-02-09
Unify `BootstsrapToken` and `InviteToken` into `AccessToken`

touilleMan:
Split between `BootstsrapToken` and `InviteToken` doesn't add much in term of readability and increase the amount of code.
On top of that:
- `InviteToken` is already being used in the shamir for the reveal token (see `libparsec/crates/protocol/schema/authenticated_cmds/shamir_recovery_setup.json5`) , so for things unrelated to invitation.
- We are going to need a new type of token for the TOTP feature

This shows we need a generic type representing an access token.


AureliaDolo (approved):
Seems like you'll have to regen the testbed server, otherwise LGTM

Issue #12179, opened 2026-02-10
CSPN Packaging change


FirelightFlagboy:
The CSPN build will be a Windows artifact generated with dedicated conf/env files and  (NOT by modifying sources, if statements, etc.):

- [ ] Add vite mode files (`alpha`, `rc`, `prod`) for `cspn` mode
- [ ] Add electron config files (`alpha`, `rc`, `prod`) for `cspn` mode
- [ ] Update workflow matrix that build electron artifact to build a windows artifact for CSPN

To be disabled:
- Document Editors (CryptPad)
- Auto Update (github)
- Sentry
- Stripe
- BMS
- Parsec Sign
- Customer Area

PR #12182, opened 2026-02-11
CSPN app version

Ironicbay:
Closes https://github.com/Scille/parsec-cloud/issues/12179

### DIFF
--- a/.github/workflows/package-desktop.yml
+++ b/.github/workflows/package-desktop.yml
@@ -439,7 +439,7 @@ jobs:
             grep -q -e "${app_file}" "${latest_file}"
 
             # Update artifact file
-            yq -i ".${{ matrix.platform }}.${arch}.client.${ext} = \"${app_file}\"" $artifact_file
+            yq -i ".${{ matrix.platform }}.${arch}.${{ matrix.hardened && 'hardened-' || '' }}client.${ext} = \"${app_file}\"" $artifact_file
 
             # Add to files_to_upload
             echo client/electron/dist/$app_file | tee -a $GITHUB_OUTPUT


misc/releaser.py
FirelightFlagboy, line 1: **issue:** We've decided against using special CSPN version.

CSPN artifact will be packaged alongside a "normal" release, so will use the same version

client/src/services/environment.ts
```diff
@@ -242,6 +251,9 @@ const CRYPTPAD_SERVER_VARIABLE = 'PARSEC_APP_DEFAULT_CRYPTPAD_SERVER';
 const DEFAULT_CRYPTPAD_SERVER = 'https://cryptpad-dev.parsec.cloud';
 
 function isEditicsEnabled(): boolean {
+  if (isCSPN()) {
```
Max-7: You shouldn't need to have those conditions. It should be a build only thing, not mentionned in the code. For this specific build, editics, stripe, sentry and other should be disabled.

client/electron/src/envVariables.ts
```diff
@@ -1,7 +1,7 @@
 // Parsec Cloud (https://parsec.cloud) Copyright (c) BUSL-1.1 2016-present Scille SAS
 
 export const Env = {
-  DISABLE_SENTRY: false,
-  DISABLE_UPDATES: false,
-  ENABLE_CUSTOM_BRANDING: process.env.PARSEC_APP_ENABLE_CUSTOM_BRANDING === 'true',
+  DISABLE_SENTRY: process.env.PARSEC_APP_IS_CSPN === 'true',
+  DISABLE_UPDATES: process.env.PARSEC_APP_IS_CSPN === 'true',
+  ENABLE_CUSTOM_BRANDING: process.env.PARSEC_APP_IS_CSPN !== 'true' && process.env.PARSEC_APP_ENABLE_CUSTOM_BRANDING === 'true',
```
FirelightFlagboy: **question:** `process.env` correspond to what?

Does it's value is fixed at the build time or it represent the user's envs var?

.github/workflows/package-desktop.yml
```diff
@@ -369,6 +378,7 @@ jobs:
             ${{ matrix.platform == 'linux-gnu' && 'appimage' || '' }} \
             ${{ inputs.nightly_build && '--nightly' || '' }} \
             ${{ matrix.platform == 'macos' && '--sign' || ''}}
+            ${{ (matrix.platform == 'windows' && matrix.cspn == 'true') && '--cspn' || '' }}
```
FirelightFlagboy: **nitpick:** We may want to enable cspn build for other platform, we cloud skip platform check

```suggestion
            ${{ matrix.cspn == 'true' && '--cspn' || '' }}
```

FirelightFlagboy:
Need for the workflow to create specific `lastest-{platform}-cspn.yml` for electron updater

client/electron/src/updater.ts
```diff
@@ -95,16 +95,17 @@ class CustomGithubProvider extends GitHubProvider {
     const { machine } = require('node:os');
     const arch = process.env['TEST_UPDATER_ARCH'] || this.options.buildMachineArch || machine();
 
+    let platform: string = this.runtimeOptions.platform;
     switch (this.runtimeOptions.platform) {
       case 'linux':
-        return `-linux-${arch}`;
+        platform = 'linux';
```
mmmarcos: Redundant case, can be removed.

client/electron/src/updater.ts
```diff
@@ -95,16 +96,16 @@ class CustomGithubProvider extends GitHubProvider {
     const { machine } = require('node:os');
     const arch = process.env['TEST_UPDATER_ARCH'] || this.options.buildMachineArch || machine();
 
+    let platform: string = this.runtimeOptions.platform;
+    // rename "special" platform to more sensible name
```
mmmarcos:
```suggestion
    // rename "special" platforms to a more meaningful name
```

PR #12187, opened 2026-02-12
Remove platform_pki example dependency on legacy verify_message.

### LINKED ISSUE #12185: rename `verify_message2` -> `verify_message` in `libparsec_platform_pki`
The old one is still used in pki examples.

AureliaDolo:
Fix #12185


libparsec/crates/platform_pki/examples/verify_message.rs
FirelightFlagboy, line 1: **polish:** The goal of these examples are twofold:

1. Be able to interact at a low level with the PKI without all of libparsec
2. Provide example that could be used for test

Following step 2, I would like that you use the `verify_message` expose in the lib instead even tho it's more complex and does more than previously.

PR #12177, opened 2026-02-10
Remove pki_enrollment.

AureliaDolo:
Fix #12054

PR #12188, opened 2026-02-12
TOTP implementation on server

touilleMan:


server/parsec/components/postgresql/migrations/0021_totp.sql
AureliaDolo, line 1: We're gonna have to fight over that file number ^^ #12177

AureliaDolo (commented):
Still have the PG impl to review but for now LGTM

server/parsec/components/postgresql/migrations/0021_totp.sql
touilleMan: Whoever merges first win ^^

server/parsec/components/postgresql/migrations/0021_totp.sql
```diff
@@ -0,0 +1,31 @@
+-- Parsec Cloud (https://parsec.cloud) Copyright (c) BUSL-1.1 2016-present Scille SAS
+
+-------------------------------------------------------
+--  Migration
+--
+-- TOTP support
+-------------------------------------------------------
+
+ALTER TABLE user_
+ADD COLUMN totp_setup_completed BOOLEAN NOT NULL DEFAULT FALSE,
```
mmmarcos, line 10: I wouldn't name it "completed" since it seems to involve multiple steps. 

I know that there are actually more than 1 step ("get_secret", "confirm") but if I'm not mistaken the "get_secret" API is idempotent (calling it multiple times return the same secret), so the totp is either *enabled* ("get_secret" ok + "confirm" ok) or not (no "confirm" yet).

IMO, "enabled" seems simple enough to convey the meaning of the field:

```suggestion
ADD COLUMN totp_enabled BOOLEAN NOT NULL DEFAULT FALSE,
```

server/parsec/components/memory/totp.py
```diff
@@ -0,0 +1,343 @@
...
+        match (user_id, user_email):
+            case (None, None):
+                return TOTPResetBadOutcome.NO_USER_ID_NOR_EMAIL
+            case (UserID() as user_id, None):
+                try:
+                    user = org.users[user_id]
+                except KeyError:
+                    return TOTPResetBadOutcome.USER_NOT_FOUND
+            case (None, EmailAddress() as user_email):
+                # Multiple user can have the same email, but at most one of them is active
```
mmmarcos, line 306: nit
```suggestion
                # Multiple users may have the same email, but only one of them can be active
```

server/parsec/components/memory/totp.py
```diff
@@ -0,0 +1,343 @@
...
+        try:
+            opaque_key, throttle = user.totp_opaque_keys[opaque_key_id]
+        except KeyError:
+            return TOTPFetchOpaqueKeyBadOutcome.OPAQUE_KEY_NOT_FOUND
+
+        match throttle.wait_until:
+            case None:
+                pass  # No throttle needed
+            case wait_until if now > wait_until:
+                pass  # Throttle needed, but we are already passed it
```
mmmarcos, line 269:
```suggestion
                pass  # Throttle needed, but we have already passed it
```

server/parsec/components/memory/datamodel.py
```diff
@@ -492,6 +493,28 @@ class MemoryUserProfileUpdate:
...
+    last_attempt: DateTime | None = None
+    failed_attemps: int = 0
+
+    @property
+    def wait_until(self) -> None | DateTime:
+        if self.failed_attemps <= 0:
+            return None
+        else:
+            assert self.last_attempt is not None
+            return self.last_attempt.add(seconds=2 ** (self.failed_attemps - 1))
```
mmmarcos: Should we move this up to the generic totp component? (to avoid having it duplicated in the totp postgresql implementation).

I know it's a minor one-line calculation, and probably covered by tests, but I cannot turn a blind eye on duplicated code, it's stronger than me ^^

server/parsec/components/totp.py
```diff
@@ -0,0 +1,472 @@
...
+    authenticated_cmds,
+)
+from parsec.api import api
+from parsec.client_context import AnonymousClientContext, AuthenticatedClientContext
+from parsec.components.email import SendEmailBadOutcome, send_email
+from parsec.config import BackendConfig
+from parsec.types import BadOutcomeEnum
+
+
+def compute_totp_one_time_password(
```
mmmarcos, line 36: Just curious: why implementing it here instead of using a library?

I see you've added tests for your implementation, but are they strong enough? ^^

server/parsec/components/totp.py
```diff
@@ -0,0 +1,472 @@
...
+        totp_reset_url=totp_reset_url,
+        server_url=server_url,
+    )
+    text = jinja_env.get_template("email/totp_reset.txt.j2").render(
+        organization_id=organization_id.str,
+        totp_reset_url=totp_reset_url,
+    )
+
+    message = MIMEMultipart("alternative")
+    message["Subject"] = f"[Parsec] TOTP reset for {organization_id.str}"
```
mmmarcos: The GUI uses "MFA" instead of "TOTP".

```suggestion
    message["Subject"] = f"[Parsec] Your MFA has been reset for {organization_id.str}"
```

server/parsec/templates/email/totp_reset.txt.j2
```diff
@@ -0,0 +1,7 @@
+Your TOTP authentication has been reset for the {{ organization_id }} organization on Parsec.
```
mmmarcos:
```suggestion
Your MFA has been reset for the {{ organization_id }} organization in Parsec.
```

server/parsec/templates/email/totp_reset.txt.j2
```diff
@@ -0,0 +1,7 @@
...
+
+To set up a new TOTP device, please open the following link in the Parsec application:
```
mmmarcos: I wound't mention "Parsec application", since it should be possible to open the link in web (so without the app).

```suggestion
To set up a new MFA for your device, please open the following link in Parsec:
```

server/parsec/templates/email/totp_reset.txt.j2
```diff
@@ -0,0 +1,7 @@
...
+
+{{ totp_reset_url }}
+
+For more information please refer to Parsec documentation: https://docs.parsec.cloud
```
mmmarcos: We should open an issue to link to the specific MFA section in the docs (once there is one ^^)

server/parsec/templates/email/totp_reset.html.j2
```diff
@@ -0,0 +1,31 @@
+{% extends "email/base.html.j2" %}
+{% block document_title %}Parsec TOTP Reset{% endblock %}
+
+{% block preheader %}Your TOTP authentication has been reset.{% endblock %}
```
mmmarcos:
```suggestion
{% block preheader %}Your MFA has been reset for the {{ organization_id }} organization in Parsec.{% endblock %}
```

server/parsec/templates/email/totp_reset.html.j2
```diff
@@ -0,0 +1,31 @@
...
+
+{% block title %}
+TOTP authentication reset for {{ organization_id }}
+{% endblock %}
+
+{% block description %}
+<p style="line-height: 28.9px; color: #51517B; font-size: 16px;">
+Your TOTP authentication has been reset for the <b>{{ organization_id }}</b> organization on Parsec.
```
mmmarcos:
```suggestion
Your MFA has been reset for the <b>{{ organization_id }}</b> organization in Parsec.  
```

server/parsec/templates/email/totp_reset.html.j2
```diff
@@ -0,0 +1,31 @@
...
+You can now set up a new TOTP device by following the link below.
```
mmmarcos:
```suggestion
Use the link below to set up a new MFA for your device.
```

server/parsec/templates/email/totp_reset.html.j2
```diff
@@ -0,0 +1,31 @@
...
+</p>
+{% endblock %}
+
+{% block main_content %}
+{% from "email/macros.html.j2" import gen_contents, render_link, render_fallback_link %}
+{{
+  gen_contents([
+    namespace(
+      title="Set up TOTP",
+      text="To set up a new TOTP device, please open the following link in the Parsec application:",
```
mmmarcos:
```suggestion
      title="Set up MFA",
      text="To set up MFA, please open the following link in Parsec:",
```

server/parsec/templates/email/totp_reset.html.j2
```diff
@@ -0,0 +1,31 @@
...
+      extras=[
+        render_link(totp_reset_url, "Set up TOTP"),
```
mmmarcos:
```suggestion
        render_link(totp_reset_url, "Set up MFA"),
```

server/parsec/templates/email/totp_reset.html.j2
```diff
@@ -0,0 +1,31 @@
+{% extends "email/base.html.j2" %}
+{% block document_title %}Parsec TOTP Reset{% endblock %}
+
+{% block preheader %}Your TOTP authentication has been reset.{% endblock %}
+
+{% block title %}
+TOTP authentication reset for {{ organization_id }}
```
mmmarcos:
```suggestion
Your MFA has been reset for {{ organization_id }}
```

server/parsec/templates/email/totp_reset.html.j2
```diff
@@ -0,0 +1,31 @@
+{% extends "email/base.html.j2" %}
+{% block document_title %}Parsec TOTP Reset{% endblock %}
```
mmmarcos:
```suggestion
{% block document_title %}Your MFA has been reset in Parsec{% endblock %}
```

server/parsec/cli/export_email.py
```diff
@@ -201,3 +203,35 @@ def account_delete(
     )
 
     write_mail_file_to_filesystem(message, output_dir, "account_delete_validation_email")
+
+
+@export_email.command(short_help="Export TOTP reset email")
```
mmmarcos, line 208: Should we use here the "user-facing" term (MFA)? Even if this CLI is for server admin, it might be easier to understand for on-premise deployments.

```suggestion
@export_email.command(short_help="Export MFA reset email")
```

server/parsec/cli/export_email.py
```diff
@@ -201,3 +203,35 @@ def account_delete(
...
+@mail_templates_shared_options
+@click.option(
+    "--organization-id", type=OrganizationID, default=DEFAULT_ORGANIZATION_ID, show_default=True
+)
+@click.option(
+    "--totp-reset-url",
```
mmmarcos, line 214:
```suggestion
    "--mfa-reset-url",
```

server/parsec/cli/export_email.py
```diff
@@ -201,3 +203,35 @@ def account_delete(
...
+    default=DEFAULT_TOTP_RESET_URL,
+    show_default=True,
+)
+def totp_reset(
```
mmmarcos, line 218: Should we use here the "user-facing" term (MFA)? Even if this CLI is for server admin, it might be easier to understand for on-premise server admins.

```suggestion
def mfa_reset(
```

server/parsec/components/postgresql/totp_setup_confirm.py
```diff
@@ -0,0 +1,197 @@
...
+            return TOTPSetupConfirmBadOutcome.AUTHOR_NOT_FOUND
+        case AuthNoLockBadOutcome.AUTHOR_REVOKED:
+            return TOTPSetupConfirmBadOutcome.AUTHOR_REVOKED
+
+    return await _totp_setup_confirm(
+        conn, user_internal_id=db_auth.user_internal_id, one_time_password=one_time_password
+    )
+
+
+_q_get_organization_user_token = Q("""
```
mmmarcos: This is the same exact query used in `totp_setup_get_secret`, is it possible to put it somewhere in common to avoid duplicated code? 

Or maybe adding a `totp_check_token.py` file with this query and all the token-related checks before actually performing `totp_setup_confirm` and `totp_setup_get_secret`?

Or, if feelinig lazy, maybe just leave the query in one command (e.g. here) and import it from the other?

.cspell/custom-words.txt
```diff
@@ -151,6 +151,7 @@ headerpage
 Hexagone
 HKCR
 homoglyph
+HOTP
```
mmmarcos: Unused?
```suggestion
```

mmmarcos, line 154: unused?
```suggestion
```

server/parsec/components/memory/totp.py
touilleMan: :+1: done.

server/parsec/components/totp.py
touilleMan: The code is dead simple, and it is very easy to verify using the test vector provided in the RFC

So I think it is better do that instead of adding yet another dependency which has to be trusted, updated etc.

server/parsec/components/memory/totp.py
touilleMan: :+1: done.

server/parsec/templates/email/totp_reset.txt.j2
mmmarcos: https://github.com/Scille/parsec-cloud/issues/12249

.cspell/custom-words.txt
touilleMan: it is used in a comment in `compute_totp_one_time_password`'s body:
```python
# TOTP is just HOTP (RFC 4226) with a counter derived from the time
```

server/parsec/components/totp.py
touilleMan: done.

server/parsec/templates/email/totp_reset.txt.j2
touilleMan: Note here "device" in "TOTP device" doesn't refer to a Parsec device but to the fact the TOTP secret si to be configured on an authenticator app.

However this wording is confusing, so what about
```suggestion
Use the link below to re-enable MFA.
```

server/parsec/templates/email/totp_reset.txt.j2
touilleMan: What about also adding a word about what MFA stands for ?
```suggestion
Your MFA (Multi-Factor Authentication) has been reset for the {{ organization_id }} organization in Parsec.
```

server/parsec/templates/email/totp_reset.html.j2
touilleMan: done.

server/parsec/templates/email/totp_reset.html.j2
touilleMan: done.

server/parsec/templates/email/totp_reset.html.j2
touilleMan: done.

server/parsec/templates/email/totp_reset.html.j2
touilleMan: done.

server/parsec/templates/email/totp_reset.html.j2
touilleMan: done.

server/parsec/templates/email/totp_reset.html.j2
touilleMan: done.

server/parsec/templates/email/totp_reset.html.j2
touilleMan: done.

server/parsec/cli/export_email.py
touilleMan: This command is only for development purpose, so I'd rather stick with `TOTP` wording here

server/parsec/cli/export_email.py
touilleMan: same as https://github.com/Scille/parsec-cloud/pull/12188#discussion_r2840296017

server/parsec/cli/export_email.py
touilleMan: This is not for server admin, this is only for debugging purpose to easily generate the email

I think the `export` in the name is misleading, I've opened issue https://github.com/Scille/parsec-cloud/issues/12250 to rename those render-email-for-dev-purpose commands

server/parsec/components/postgresql/totp_setup_confirm.py
touilleMan: > This is the same exact query used in totp_setup_get_secret, is it possible to put it somewhere in common to avoid duplicated code?
> ...
> Or, if feelinig lazy, maybe just leave the query in one command (e.g. here) and import it from the other?

You are right (and I'm lazy ^^), so now set confirm import the query from setup get secret :raised_hands:

server/parsec/templates/email/totp_reset.txt.j2
touilleMan: I've created issue https://github.com/Scille/parsec-cloud/issues/12251 about documenting MFA.
I've also removed this note about the documentation in the mail (since it is very ~diabolical~ misleading to link to a documentation not containing what the user is looking for ^^) and replaced by a TODO related to this issue

server/parsec/components/memory/datamodel.py
touilleMan: I'm willing to limit your suffering, so I've factorized this code :smile: 

See `compute_wait_until()` in `server/parsec/components/totp.py`

server/parsec/components/postgresql/migrations/0021_totp.sql
touilleMan: > I know that there are actually more than 1 step ("get_secret", "confirm") but if I'm not mistaken the "get_secret" API is idempotent

You are correct, so there is in the end three states:
1. The TOTP has never been configured
2. The TOTP is configured
3. The TOTP has been configured in the past, and its configuration has been reset

States 1 & 3 are basically the same (especially since you can create opaque keys *before* configuring TOTP, only accessing the opaque key require to have TOTP configured)

>  so the totp is either enabled ("get_secret" ok + "confirm" ok) or not (no "confirm" yet).

yes

> IMO, "enabled" seems simple enough to convey the meaning of the field:

I don't like "enabled" since it looks like TOTP is globaly enabled (since it is how TOTP works for most website) while here it is just about the fact it can be used.

That's why I wanted to have `setup_completed` here: it refers to the fact `totp_setup_complete` has been successfully called (and no totp setup reset has been done since then)

server/parsec/components/totp.py
```diff
@@ -0,0 +1,479 @@
...
+    msg = struct.pack(">Q", counter)
+    h = hmac.new(secret, msg, hashlib.sha1).digest()
+    # A portion of the HMAC is extracted and displayed to the user as a six- to eight-digit code;
+
+    # Step 2: Generate a 4-byte string (Dynamic Truncation)
+    # The last nibble (4 bits) of the result is used as a pointer...
+    pointer = h[-1] & 0x0F
+    # ...to a 32-bit integer, in the result byte array...
+    code = struct.unpack(">I", h[pointer : pointer + 4])[0]
+    # ...and masks out the 31st bit.
```
FirelightFlagboy: **typo:** Here it's a position, not an index, so it should be 32 since we start the position at 1 (when we refer to the first bit, we said it's in the 1st place not 0st place)

```suggestion
    # ...and masks out the 32st bit.
```

server/parsec/components/memory/totp.py
```diff
@@ -0,0 +1,343 @@
...
+        if user.totp_setup_completed:
+            return TOTPSetupGetSecretBadOutcome.ALREADY_SETUP
+
+        if user.totp_secret is None:
+            user.totp_secret = generate_totp_secret()
+
+        return user.totp_secret
+
+    @override
+    async def setup_confirm(
```
FirelightFlagboy, line 78: **question:** Why the implementation does not check that `user.topt_reset_token is None`?

By reading the code, it fells like it is possible to re-configure the TOPT after a reset by using the route without the the token

server/tests/administration/test_reset_user_totp.py
```diff
@@ -0,0 +1,296 @@
...
+    assert result["email_sent_status"] == "SENT_AS_REQUESTED"
+    assert email_totp_reset_letterbox.count() == 1
+
+    # Check that the email was sent
+    to_addr, message = await email_totp_reset_letterbox.get_next()
+    assert to_addr == minimalorg.alice.human_handle.email
+    message_as_str = message.as_string()
+    assert f"Your MFA has been reset for {minimalorg.organization_id}" in message_as_str
+    assert result["totp_reset_url_as_http_redirection"] in message_as_str
+    print(message.as_string())
```
FirelightFlagboy: **chore:** Debug print
```suggestion
```

server/tests/api_v5/authenticated/test_totp_setup_confirm.py
```diff
@@ -0,0 +1,62 @@
...
+    )
+    assert rep == authenticated_cmds.latest.totp_setup_confirm.RepOk()
+
+    # Now we cannot get the TOTP secret anymore
+
+    outcome = await minimalorg.alice.totp_setup_get_secret()
+    assert outcome == authenticated_cmds.latest.totp_setup_get_secret.RepAlreadySetup()
+
+    outcome = await minimalorg.alice.totp_setup_get_secret()
+    assert outcome == authenticated_cmds.latest.totp_setup_get_secret.RepAlreadySetup()
```
FirelightFlagboy: **question:** Why executing the same query a second time?

server/parsec/components/totp.py
touilleMan: :+1: done.

server/tests/administration/test_reset_user_totp.py
touilleMan: :+1: done.

server/tests/api_v5/authenticated/test_totp_setup_confirm.py
touilleMan: Good catch, this is an error, I've removed it :+1:

server/parsec/components/memory/totp.py
touilleMan: It is possible to reconfigure the TOTP after a reset without the token.

The need for the reset token is because a user that has lost his authenticator app is most likely locked out of his device.

We could in theory be extra zealous and only allow authenticated setup confirm for the initial setup (i.e. checking both `totp_setup_completed == False` and `totp_reset_token == None`) but that would create a more complex internal logic (currently there is only two state `configured/unconfigured` and we would switch to three `initial_unconfigured/configured/reset_unconfigured`) and potentially add another error status (otherwise it would be strange to get a `ALREADY_SETUP` status when calling `totp_setup_confirm` but a `NOT_SETUP` when calling `fetch_opaque_key` :/).

Right now the only "risk" we have is the fact an attacker with access to a device can reset the TOTP without having access to the reset link. This is a rather small issue considering the attacker has already access to a device :scream:

PR #12196, opened 2026-02-13
TOTP support for libparsec

touilleMan:


libparsec/crates/totp/src/lib.rs
```diff
@@ -0,0 +1,231 @@
+// Parsec Cloud (https://parsec.cloud) Copyright (c) BUSL-1.1 2016-present Scille SAS
+
+use libparsec_client_connection::{AnonymousCmds, AuthenticatedCmds, ConnectionError};
+use libparsec_protocol::{anonymous_cmds, authenticated_cmds};
+use libparsec_types::prelude::*;
+
+#[derive(Debug, PartialEq, Eq)]
+pub enum TOTPSetupStatus {
+    /// The TOTP setup is not yet confirmed: the secret should be displayed
+    /// to the end-user so that he can configure his authenticator app.
```
mmmarcos:
```suggestion
    /// TOTP setup is not yet set up: the secret should be displayed
    /// to the user so that it can configure its authenticator app.
```

libparsec/crates/totp/src/lib.rs
```diff
@@ -0,0 +1,231 @@
...
+    Stalled {
+        base32_totp_secret: String,
+    },
+    AlreadySetup,
```
mmmarcos: I believe "stalled" means something that is _blocked_ for some reason. Besides, the name requires that little extra explanation to understand what it really means.

Moreover, I can understand `AlreadySetup` for a bad outcome (sounds like an error), but for a status why not just use something like "Enabled"?

Some suggestions:
```suggestion
    Disabled {
        base32_totp_secret: String,
    },
    Enabled,
```
```suggestion
    Unconfirmed {
        base32_totp_secret: String,
    },
    Confirmed,
```
```suggestion
    NotYetConfigured {
        base32_totp_secret: String,
    },
    Configured,
```

libparsec/crates/totp/src/lib.rs
```diff
@@ -0,0 +1,231 @@
...
+            unknown_status
+        ))),
+    }
+}
+
+#[derive(Debug, thiserror::Error)]
+pub enum TotpSetupConfirmAuthenticatedError {
+    #[error("The one-time-password is invalid")]
+    InvalidOneTimePassword,
+    #[error("The TOTP has already been setup")]
```
mmmarcos, line 89:
```suggestion
    #[error("The TOTP has already been set up")]
```

libparsec/crates/totp/src/lib.rs
```diff
@@ -0,0 +1,231 @@
...
+        } => Err(TotpSetupStatusAnonymousError::Internal(anyhow::anyhow!(
+            "Unknown error status `{}` from server",
+            unknown_status
+        ))),
+    }
+}
+
+#[derive(Debug, thiserror::Error)]
+pub enum TotpSetupConfirmAuthenticatedError {
+    #[error("The one-time-password is invalid")]
```
mmmarcos, line 87:
```suggestion
    #[error("The one-time password is invalid")]
```

libparsec/crates/totp/src/lib.rs
```diff
@@ -0,0 +1,231 @@
...
+            ..
+        } => Err(TotpSetupConfirmAuthenticatedError::Internal(
+            anyhow::anyhow!("Unknown error status `{}` from server", unknown_status),
+        )),
+    }
+}
+
+#[derive(Debug, thiserror::Error)]
+pub enum TotpSetupConfirmAnonymousError {
+    #[error("The one-time-password is invalid")]
```
mmmarcos, line 122:
```suggestion
    #[error("The one-time password is invalid")]
```

libparsec/crates/totp/src/lib.rs
```diff
@@ -0,0 +1,231 @@
...
+
+#[derive(Debug, thiserror::Error)]
+pub enum TotpSetupStatusAuthenticatedError {
+    #[error("Cannot communicate with the server: {0}")]
+    Offline(#[from] ConnectionError),
+    #[error(transparent)]
+    Internal(#[from] anyhow::Error),
+}
+
+pub async fn totp_setup_status_authenticated(
```
mmmarcos, line 28: This puzzled me at first since I didn't recall a "status" command... why not naming it to match the command that is actually requested?

```suggestion
pub async fn totp_setup_get_secret_authenticated(
```

libparsec/crates/totp/src/lib.rs
```diff
@@ -0,0 +1,231 @@
...
+pub enum TotpSetupStatusAnonymousError {
+    #[error("The authentication token is invalid")]
+    BadToken,
+    #[error("Cannot communicate with the server: {0}")]
+    Offline(#[from] ConnectionError),
+    #[error(transparent)]
+    Internal(#[from] anyhow::Error),
+}
+
+pub async fn totp_setup_status_anonymous(
```
mmmarcos, line 60: Idem?
```suggestion
pub async fn totp_setup_get_secret_anonymous(
```

cli/src/commands/user/mod.rs
```diff
@@ -1,17 +1,21 @@
...
+pub mod totp_reset;
 
 #[derive(clap::Subcommand)]
 pub enum Group {
     /// List users
     List(list::Args),
     /// Revoke a user
     Revoke(revoke::Args),
+    /// Reset TOTP for a user (administration command)
+    TotpReset(totp_reset::Args),
```
mmmarcos: IMO, it's better to use the user-facing term here since both organization admins AND server admins (specially for on-premise) will mostly use MFA.

```suggestion
    /// Reset the MFA setup for a user (administration command)
    MfaReset(totp_reset::Args),
```

cli/src/commands/user/totp_reset.rs
```diff
@@ -0,0 +1,132 @@
+// Parsec Cloud (https://parsec.cloud) Copyright (c) BUSL-1.1 2016-present Scille SAS
+
+use libparsec::{OrganizationID, ParsecAddr};
+
+use crate::utils::*;
+
+crate::clap_parser_with_shared_opts_builder!(
+    #[with = addr, token, organization]
+    pub struct Args {
+        /// User ID (hex) of the user to reset TOTP for (mutually exclusive with --user-email)
```
mmmarcos, line 11:
```suggestion
        /// User ID (hex) of the user whose MFA will be reset (mutually exclusive with --user-email)
```

cli/src/commands/user/totp_reset.rs
```diff
@@ -0,0 +1,132 @@
...
+        #[arg(long, group = "user_ref")]
+        user_id: Option<String>,
+        /// Email of the user to reset TOTP for (mutually exclusive with --user-id)
```
mmmarcos, line 14:
```suggestion
        /// Email of the user whose MFA will be reset (mutually exclusive with --user-id)
```

cli/src/commands/user/totp_reset.rs
```diff
@@ -0,0 +1,132 @@
...
+        #[arg(long, group = "user_ref")]
+        user_email: Option<String>,
+        /// Request the server to send the TOTP reset email
```
mmmarcos:
```suggestion
        /// Request the server to send the MFA reset email
```

cli/src/commands/user/totp_reset.rs
```diff
@@ -0,0 +1,132 @@
...
+        user_id.as_deref(),
+        user_email.as_deref(),
+        send_email,
+    )
+    .await?;
+
+    let msg = match res.email_sent_status.as_ref() {
+        "NOT_SENT_AS_REQUESTED" => {
+            format!(
+                "TOTP reset for user {GREEN}{}{RESET}\nReset URL: {YELLOW}{}{RESET}",
```
mmmarcos, line 106:
```suggestion
                "Reset MFA for user {GREEN}{}{RESET}\nReset URL: {YELLOW}{}{RESET}",
```

cli/src/commands/user/totp_reset.rs
```diff
@@ -0,0 +1,132 @@
...
+                res.user_id, res.totp_reset_url,
+            )
+        }
+        "SENT_AS_REQUESTED" => {
+            format!(
+                "TOTP reset for user {GREEN}{}{RESET}\n\
```
mmmarcos, line 112:
```suggestion
                "Reset MFA for user {GREEN}{}{RESET}\n\
```

cli/src/commands/user/totp_reset.rs
```diff
@@ -0,0 +1,132 @@
...
+                Reset URL: {YELLOW}{}{RESET}\n\
+                An email with the reset URL has been sent to {YELLOW}{}{RESET}",
+                res.user_id, res.totp_reset_url, res.user_email,
+            )
+        }
+        email_sent_err => {
+            format!(
+                "TOTP reset for user {GREEN}{}{RESET}\n\
```
mmmarcos, line 120:
```suggestion
                "Reset MFA for user {GREEN}{}{RESET}\n\
```

touilleMan: done.

cli/src/commands/user/totp_reset.rs
touilleMan: done.

cli/src/commands/user/totp_reset.rs
touilleMan: done.

cli/src/commands/user/totp_reset.rs
touilleMan: done.

cli/src/commands/user/totp_reset.rs
touilleMan: done.

cli/src/commands/user/totp_reset.rs
touilleMan: done.

cli/src/commands/user/mod.rs
touilleMan: done.

libparsec/crates/totp/src/lib.rs
touilleMan: done.

libparsec/crates/totp/src/lib.rs
touilleMan: done.

libparsec/crates/totp/src/lib.rs
touilleMan: done.

libparsec/crates/totp/src/lib.rs
touilleMan: done.

libparsec/crates/totp/src/lib.rs
touilleMan: You are right, stalled is not the right word.

However I don't like enabled/disabled since it may refer to the fact TOTP is globally enabled, which is not what it is about here.

So what about:
- keeping `totp_setup_status` as the name of the function (to underline the fact this is about TOTP setup)
- Use the `Confirmed/Unconfirmed` as you suggested

libparsec/crates/totp/src/lib.rs
touilleMan: It is named to help to consumer of the API: when the GUI needs to protect a new device with TOTP, it needs first to check if the TOTP setup is complete or not (hence `totp_setup_status`)

libparsec/crates/totp/src/lib.rs
touilleMan: see https://github.com/Scille/parsec-cloud/pull/12196#discussion_r2840769636

mmmarcos (approved):

bindings/generator/api/device.py
```diff
@@ -36,10 +39,13 @@ class OpenBao:
         openbao_preferred_auth_id: str
         openbao_entity_id: str
 
+    class TOTP:
+        totp_opaque_key_id: TOTPOpaqueKeyID
+        next: Box["AvailableDeviceType"]
```
FirelightFlagboy: **question:** Why not using `__future__.annotations` here?

bindings/generator/api/totp.py
```diff
@@ -0,0 +1,136 @@
...
+    TOTPOpaqueKeyID,
+    UserID,
+    Variant,
+    VariantItemUnit,
+)
+
+
+class TOTPSetupStatus(Variant):
+    class Unconfirmed:
+        base32_totp_secret: str
```
FirelightFlagboy, line 20: **question:** Why we provide the secret in b32 over b64 or hex?

cli/src/commands/user/totp_reset.rs
FirelightFlagboy: Suggestion not applied

cli/src/commands/user/totp_reset.rs
FirelightFlagboy: Suggestion not applied

cli/src/commands/user/totp_reset.rs
FirelightFlagboy: Suggestion not applied

cli/src/commands/user/totp_reset.rs
```diff
@@ -0,0 +1,132 @@
...
+
+crate::clap_parser_with_shared_opts_builder!(
+    #[with = addr, token, organization]
+    pub struct Args {
+        /// User ID (hex) of the user to reset TOTP for (mutually exclusive with --user-email)
+        #[arg(long, group = "user_ref")]
+        user_id: Option<String>,
+        /// Email of the user to reset TOTP for (mutually exclusive with --user-id)
+        #[arg(long, group = "user_ref")]
+        user_email: Option<String>,
```
FirelightFlagboy: **polish:** Use `EmailAddress`

FirelightFlagboy: **polish:** Use `clap::ArgGroup` to mark the group as required (by default arg in group are exclusive)

clap will provide a better error message and helper with arg group

bindings/generator/api/totp.py
touilleMan: This is the standard way of encoding the TOTP secret for `otpauth://` URL (@Max-7 asked me provide the secret in this form to avoid having the GUI to roll its own base32 encoding)

bindings/generator/api/device.py
touilleMan: That's a good point, however I'd rather not modify this code since it is getting removed in the next PR

cli/src/commands/user/totp_reset.rs
touilleMan: :+1: done.

I've also improved the parameter checks and the behavior when the server returns an invalid body

touilleMan: :+1: done (I've also used `UserID`  for --user-id)

cli/src/commands/user/totp_reset.rs
touilleMan: After further discussion, we decided against using MFA here (since the CLI is much lower level than the GUI, and hence using the same term than the HTTP administration API is better)
Though I've added a word about MFA being synonym to TOTP in the command help description :+1:

libparsec/crates/platform_device_loader/tests/units/list.rs
```diff
@@ -376,13 +376,13 @@ async fn list_devices(tmp_path: TmpPath) {
...
-    let password_path = tmp_path.join("devices/foo/bar/spam/whatever.keys");
-    let pki_path = tmp_path.join("devices/foo/bar/spam/whatever3.keys");
-    let recovery_path = tmp_path.join("devices/foo/whatever.keys");
-    let account_vault_path = tmp_path.join("devices/whatever.keys");
-    let openbao_path = tmp_path.join("devices/whatever2.keys");
+    let password_path: PathBuf = tmp_path.join("devices/foo/bar/spam/whatever.keys");
+    let pki_path: PathBuf = tmp_path.join("devices/foo/bar/spam/whatever3.keys");
+    let recovery_path: PathBuf = tmp_path.join("devices/foo/whatever.keys");
+    let account_vault_path: PathBuf = tmp_path.join("devices/whatever.keys");
+    let openbao_path: PathBuf = tmp_path.join("devices/whatever2.keys");
```
FirelightFlagboy, line 385: **question:** Why the explicit type here?

`PathBuf::join` explicitly return `PathBuf`, so I do not understand why it's required here.

FirelightFlagboy (dismissed):

cli/src/commands/user/totp_reset.rs
touilleMan: see https://github.com/Scille/parsec-cloud/pull/12196#discussion_r2845658746

cli/src/commands/user/totp_reset.rs
touilleMan: see https://github.com/Scille/parsec-cloud/pull/12196#discussion_r2845658746

libparsec/crates/platform_device_loader/tests/units/list.rs
touilleMan: This is due to rust-analyzer struggling with `tmp_path` fixture:
- this fixture is defined in `libparsec_test_fixtures`
- `libparsec_test_fixtures` depends on `libparsec_platform_device_loader`
- `libparsec_platform_device_loader`'s tests depend on `libparsec_test_fixtures`

So this is totally valid, but rust-analyzer sees this as a circle dependency, stops it type inference too soon, and we end up with `tmp_path` not having any type...

libparsec/crates/totp/tests/unit/mod.rs
```diff
@@ -0,0 +1,12 @@
...
+mod totp_create_opaque_key;
+mod totp_fetch_opaque_key;
+mod totp_setup_confirm_anonymous;
+mod totp_setup_confirm_authenticated;
+mod totp_setup_status_anonymous;
+mod totp_setup_status_authenticated;
+mod utils;
+
+#[cfg(target_arch = "wasm32")]
+libparsec_tests_lite::platform::wasm_bindgen_test_configure!(run_in_browser);
```
FirelightFlagboy: **todo:** Should also be run in shared worker

libparsec/crates/totp/tests/unit/totp_setup_status_anonymous.rs
```diff
@@ -0,0 +1,64 @@
...
+    });
+
+    let result = totp_setup_status_anonymous(&cmds, user_id, token)
+        .await
+        .unwrap();
+    p_assert_eq!(
+        result,
+        TOTPSetupStatus::Unconfirmed {
+            // cspell: disable-next-line
+            base32_totp_secret: "KMZWG4RTOQ".to_string(),
```
FirelightFlagboy: **nitpick:** 
```suggestion
            base32_totp_secret: ::data_encoding::BASE32.encode(expected_secret),
```

libparsec/crates/totp/tests/unit/totp_setup_status_authenticated.rs
```diff
@@ -0,0 +1,56 @@
...
+            }
+        }
+    });
+
+    let result = totp_setup_status_authenticated(&client).await.unwrap();
+    p_assert_eq!(
+        result,
+        TOTPSetupStatus::Unconfirmed {
+            // cspell: disable-next-line
+            base32_totp_secret: "KMZWG4RTOQ".to_string(),
```
FirelightFlagboy: **nitpick:** 
```suggestion
            base32_totp_secret: ::data_encoding::BASE32.encode(expected_secret),
```

libparsec/crates/types/src/addr.rs
```diff
@@ -1237,6 +1237,13 @@ pub enum ParsecAnonymousAddr {
 }
 
 impl ParsecAnonymousAddr {
+    pub fn new(server_addr: ParsecAddr, organization_id: OrganizationID) -> Self {
```
FirelightFlagboy, line 1243: **polish:** Convey the returned ano addr type

```suggestion
    pub fn new_org_bootstrap_addr(server_addr: ParsecAddr, organization_id: OrganizationID) -> Self {
```

touilleMan: The fact we use an `OrganizationBootstrap` variant here is an implementation detail, hence this comment:
```rust
        // Small hack: we benefit from the fact a bootstrap organization can be
        // constructed from the parameters we have at hand.
```

This `ParsecAnonymousAddr::new` method is only here to create a `ParsecAnonymousAddr` from a couple `ParsecAddr + OrganizationID` (since they are the basic blocks needed for this type of addr), if we want to be extra zealous we would introduce a new variant type for this. However this is cumbersome since `ParsecOrganizationBootstrapAddr` can be used instead, and we never convert a `ParsecAnonymousAddr` back into it variant.

libparsec/crates/totp/tests/unit/mod.rs
touilleMan: :+1: done.

should we create an issue to modify the `wasm_bindgen_test_configure` in all other crates to always use shared worker ?

libparsec/crates/types/src/addr.rs
FirelightFlagboy: The function does not indicate the returned variant nor explain why it's Org bootstrap that was choosen over PkiEnroll or Org

libparsec/crates/totp/tests/unit/totp_setup_status_anonymous.rs
touilleMan: I made the change, but I'm not really convinced by it: the code is not clearer and this test no longer detects if an update breaks data_encoding (since it is used also used in the tested code).

libparsec/crates/totp/tests/unit/totp_setup_status_authenticated.rs
touilleMan: same as https://github.com/Scille/parsec-cloud/pull/12196#discussion_r2847154796

libparsec/crates/platform_device_loader/src/strategy.rs
```diff
@@ -144,6 +144,16 @@ pub enum DeviceSaveStrategy {
...
     },
+    /// TOTP consists of an opaque key stored server-side and only accessible
+    /// through a TOTP-challenge.
+    /// However this security cannot be used alone (otherwise the server would
+    /// be able to decrypt the device!), hence the `next` field to combine with
+    /// another layer of security.
+    TOTP {
+        totp_opaque_key_id: TOTPOpaqueKeyID,
+        totp_opaque_key: SecretKey,
+        next: Box<DeviceSaveStrategy>,
```
AureliaDolo: It should not happen, but in theory, we could have a TOTP in a TOTP in a TOTP ... ?

libparsec/crates/types/src/addr.rs
touilleMan: > The function does not indicate the returned variant nor explain why it's Org bootstrap that was choosen over PkiEnroll or Org

This is because this is irrelevant here, `ParsecAnonymousAddr` is not a type that is used to manipulate different types, but instead to provide a common ground between those types.
Put it another way: `ParsecAnonymousAddr` is in practice an opaque type (we could define it as `struct ParsecAnonymousAddr(ParsecAnonymousAddrInternal)` with `ParsecAnonymousAddr` being a private enum)

touilleMan: Since you seem to require something more explicit, I've added a commit to make `ParsecAnonymousAddr` internals private ;-p

libparsec/crates/platform_device_loader/src/strategy.rs
touilleMan: In theory yes, but there is some check to return an error in this case.
Note this is no big deal in any way since this code is replaced by the next PR (https://github.com/Scille/parsec-cloud/pull/12252) where nested TOTP protection is no longer possible to express

PR #12210, opened 2026-02-16
refactor(CONTRIBUTING): Split long lines

FirelightFlagboy:
When looking at `CONTRIBUTING.md` GitHub did not render lines that where too long (and I also do not like horizontal scroll)


AureliaDolo:
It has merged so fast :heart_eyes:

Issue #12054, opened 2026-01-14
Async Enrollment: remove PKI enrollment (i.e. non generic version of the async enrollment)


touilleMan:
- [x] Remove `test_pki` fixture from `libparsec_platform_pki`
- [ ] #12185
- [x] remove `pki_enrollment_*` commands schemas & server implementation
- [x] remove `pki_enrollment_*_payload` & `local_pending_enrollment` schemas in `libparsec_types`
- [x] remove `pki_enrollment_*` commands in `liparsec_client::Client`
- [x] remove bindings code
- [x] remove reference in the GUI

PR #12213, opened 2026-02-16
Remove pki_enrollment protocol.

AureliaDolo:
Part of #12054


client/src/parsec/types.ts
AureliaDolo, line 1: This was not autogenerated

PR #12215, opened 2026-02-16
Delete local_pending_enrollment.

AureliaDolo:
Part of #12054


libparsec/crates/platform_pki/src/shared/mod.rs
FirelightFlagboy, line 1: **polish:** dead code

AureliaDolo: I though I'd got it all ^^

PR #12217, opened 2026-02-16
Update testbed test to remove dep on pki enrollement

AureliaDolo:
Part of #12054


AureliaDolo:
After discussion with @touilleMan, better to use ping as it's its purpose ^^

AureliaDolo:
> Now it's even more strange about why we used pki cmds in the first place
 
Maybe because it was unused ?

PR #12218, opened 2026-02-16
TOTP: Protocol & device keys file schemas

touilleMan:


libparsec/crates/platform_device_loader/tests/units/list.rs
```diff
@@ -90,17 +93,19 @@ async fn list_devices(tmp_path: TmpPath) {
...
+        "68747470733a2f2f7061727365632e696e76616c69642faf6f7267616e697a6174696f"
+        "6e5f6964a7436f6f6c4f7267a7757365725f6964d802a11cec00100000000000000000"
+        "000000a96465766963655f6964d802de10a11cec0010000000000000000000ac68756d"
+        "616e5f68616e646c6592b4616c696365407061727365632e696e76616c6964b2416c69"
+        "636579204d63416c69636546616365ac6465766963655f6c6162656caf4d7920646576"
+        "31206d616368696e65af6b657972696e675f73657276696365af6b657972696e675f73"
+        "657276696365ac6b657972696e675f75736572ac6b657972696e675f75736572aa6369"
+        "7068657274657874c40c3c636970686572746578743eb2746f74705f6f70617175655f"
+        "6b65795f6964d8028fdb73524fdd495194e877a5fafbe0a1"
     )
```
FirelightFlagboy, line 109: **issue:** You are updating the blob instead of creating a new one

libparsec/crates/protocol/schema/anonymous_cmds/totp_fetch_opaque_key.json5
```diff
@@ -0,0 +1,78 @@
...
+                // This status is returned if:
+                // - The one-time-password is invalid (obviously)
+                // - The user is revoked or frozen
+                // - No opaque key exists with the provided ID
+                // - TOTP setup hasn't been completed (or has been reset)
+                "status": "invalid_one_time_password"
+            },
+            {
+                // Too many attempts with an invalid one time password have
+                // been done this opaque key ID.
```
FirelightFlagboy: **typo:** 
```suggestion
                // Too many attempts with an invalid one time password have
                // been done to this opaque key ID.
```

libparsec/crates/protocol/schema/anonymous_cmds/totp_fetch_opaque_key.json5
```diff
@@ -0,0 +1,78 @@
...
+                // Throttling follow a simple `wait_time_in_seconds = 2 ^ number_of_failed_attemps`
```
FirelightFlagboy, line 58: **issue:** Are you saying that I can do a DOS by gobbling the whole server memory with a ludicrously large number ?

libparsec/crates/types/schema/local_device/device_file_account_vault.json5
FirelightFlagboy, line 1: **polish:** The account device type is not in production, I think you can make the field a required option

libparsec/crates/platform_device_loader/tests/units/list.rs
touilleMan: This is fine since this test is not about schema stability (we have the tests in `libparsec/crates/types/tests/unit/local_device_file.rs` for that)
On the other hand I cannot simply add yet-another item in this test since each one of them must correspond to a different device (two device keys files for the same device would only lead to a single item listed).

libparsec/crates/protocol/schema/anonymous_cmds/totp_fetch_opaque_key.json5
touilleMan: That's a funny way to put it :laughing: 

~The server is in Python so it shouldn't be afraid of bignums, however you are right that uncapped data control by an attacker is never a good thing.~

EDIT: There is no risk for this since the failed attempts counter is only updated once an actual attempt has been made (i.e. if the server returns the `invalid_one_time_password` status, not the `throttled` status).
So, for instance, to increase the fail attempt counter from 30 to 31 the attacker has first to wait for the whole 2**30s==34years to be elapsed o_O
I've added a word about this though

libparsec/crates/protocol/schema/anonymous_cmds/totp_fetch_opaque_key.json5
touilleMan: done. :+1:

libparsec/crates/types/schema/local_device/device_file_account_vault.json5
touilleMan: I'd rather avoid this:
- The schema is already marked as introduced in Parsec 3.5.0, I don't feel comfortable changing this info given it is precisely here to help debugging (and having a schema kind of lying about when it was introduced and what it contains is typically what we don't want to have when debugging :laughing: ).
- There not much gain here since the field will be optional anyway: the main gain is to avoid a test case to check the old format with the field missing

so I don't think it is not worth it

PR #12233, opened 2026-02-19
[MS] Update redirect links & added tests

Max-7:
- Restore tests for org information
- Restore async enrollment for org information
- Fixed translations
- Allows opening Parsec app with a `parsec3://` empty link without the app displaying an error
- Adds file link handling
- Adds tests for web redirect
- Slight refactor of how web redirect handles links

### DIFF
--- a/client/src/components/organizations/OrganizationUserInformation.vue
+++ b/client/src/components/organizations/OrganizationUserInformation.vue
@@ -114,7 +114,6 @@
         </div>
 
         <div
-          v-if="false"
           button
           class="invitation-card-list-item"
           @click="goToInvitations(InvitationView.AsyncEnrollmentRequest)"
--- a/client/src/locales/en-US.json
+++ b/client/src/locales/en-US.json
@@ -1728,8 +1728,8 @@
                 "pkiCorrupted": "PKI (corrupted)"
             },
             "linkCopiedToClipboard": {
-                "success": "PKI link has been copied to clipboard.",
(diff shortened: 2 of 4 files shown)


client/src/views/home/HomePage.vue
```diff
@@ -379,12 +380,14 @@ async function handleQuery(): Promise<void> {
...
+  if (query.fileLink) {
+    handleParsecLink(query.fileLink, informationManager);
+  } else if (query.claimLink) {
+    openJoinByLinkModal(query.claimLink);
   } else if (query.bootstrapLink) {
-    await openCreateOrganizationModal(query.bootstrapLink);
+    openCreateOrganizationModal(query.bootstrapLink);
   } else if (query.asyncEnrollmentLink) {
-    await handleAsyncEnrollment(query.asyncEnrollmentLink);
+    handleAsyncEnrollment(query.asyncEnrollmentLink);
```
Ironicbay, line 390: Why remove the `awaits` ?

Ironicbay (approved):

client/src/views/home/HomePage.vue
Max-7: `await` blocks the execution until the fonctions returns (so until the modal is closed). In the case of link handling, it's done very early, which prevents the component from fully loading. So the modal opens, but the page behind it stays mostly empty (the devices don't appear).

PR #12234, opened 2026-02-19
[CLI] check that recipient_shared >= 1 in share recovery creation

### LINKED ISSUE #11522: [CLI] shamir crate does not check if `recipient_shared >= 1`
### Parsec version

master

- [ ] Any
- [ ] macOS
- [x] Linux
- [ ] Windows
- [ ] Web

### Bug description

I've create an orga with a single admin (and a standard user) and tried to create a shamir device for it. 
I use the CLI without providing the recipient list so it default to the list of admin in the organization.
Since we only have a single admin the default list is empty but the CLI does not stop with an error at that point (no recipients)

### Reproduction steps

1. Create an organization with a single user (or a single admin)
2. Try to create the shamir device from the CLI without passing a list of recipient (will use the admin list by default)
3. The CLI will ask for an invalid threshold (value between 1 and 0, such as $1 \leq x \leq 0$)
...

### Relevant output:

```
✔ Poll server for new certificates
... Creating shared recovery setup
The threshold is the minimum number of recipients that one must gather to recover the account
Choose a threshold between 1 and 0:
```

BluNfa:
Fixes #11522

Check added after creating the default recipient list to ensure that it is not empty. If it is, returns an error.

### DIFF
--- a/cli/src/commands/shared_recovery/create.rs
+++ b/cli/src/commands/shared_recovery/create.rs
@@ -62,13 +62,15 @@ pub async fn create_shared_recovery(args: Args, client: &StartedClient) -> anyho
             .map(|human| *recipient_info.get(human).expect("human should be here"))
             .collect()
     } else {
-        users
+        let admins: Vec<_> = users
             .iter()
             .filter(|info| {
                 info.current_profile == UserProfile::Admin && info.id != client.user_id()
             })
             .map(|info| info.id)
-            .collect()
+            .collect();
+        anyhow::ensure!(!admins.is_empty(), "No default recipient available");
+        admins
     };
 
     let per_recipient_shares: HashMap<UserID, NonZeroU8> = if let Some(weights) = weights {
--- a/cli/tests/integration/shared_recovery/create.rs
+++ b/cli/tests/integration/shared_recovery/create.rs
@@ -119,3 +119,19 @@ async fn create_shared_recovery_default(tmp_path: TmpPath) {
         .unwrap();
     p.exp_eof().unwrap();
 }
+
+#[rstest::rstest]
+#[tokio::test]
+async fn create_shared_recovery_no_recipient(tmp_path: TmpPath) {
+    let (_, TestOrganization { alice, .. }, _) = bootstrap_cli_test(&tmp_path).await.unwrap();
+
+    // Alice is the only admin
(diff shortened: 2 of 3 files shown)


cli/src/commands/shared_recovery/create.rs
```diff
@@ -62,13 +62,17 @@ pub async fn create_shared_recovery(args: Args, client: &StartedClient) -> anyho
             .map(|human| *recipient_info.get(human).expect("human should be here"))
             .collect()
     } else {
-        users
+        let admins: Vec<UserID> = users
```
FirelightFlagboy: **nitpick:** You can omit the inner vec type as it will be infered by the compiler


```suggestion
        let admins: Vec<_> = users
```

cli/src/commands/shared_recovery/create.rs
```diff
@@ -62,13 +62,17 @@ pub async fn create_shared_recovery(args: Args, client: &StartedClient) -> anyho
...
             .iter()
             .filter(|info| {
                 info.current_profile == UserProfile::Admin && info.id != client.user_id()
             })
             .map(|info| info.id)
-            .collect()
+            .collect();
+        if admins.is_empty() {
+            return Err(anyhow::anyhow!("No default recipient available"));
+        }
```
FirelightFlagboy: **nitpick:** Since we return `anyhow::Result`, we can use the following QOL:


```suggestion
        anyhow::ensure!(!admins.is_empty(), "No default recipient available");
```

BluNfa:
Thanks for the feedback, I applied the suggestions

FirelightFlagboy:
Do you want to squash before merging ?

BluNfa:
Just squashed, it loses the "co-authored" part in logs generated when accepting suggestion from code review, but I guess it's fine since, in the end, it will be your commit on the main

FirelightFlagboy:
> gs generated when accepting suggestion from code review, but I guess it's fine since, in the end, it will be your commit on the main

No problem. If you wanted to keep the co-authored, you would have to manually squash and edit the resulting commit message

BluNfa:
Ah yes, I didn't realise it was just in the commit message, I thought it was the same type of field as the author. It fine with me to merge like this, but I keep it in mind for the next time 😊

PR #12235, opened 2026-02-19
Display to end-user action URL (e.g. invitation) as HTTP redirection URL

touilleMan:


server/tests/administration/test_create_organization.py
```diff
@@ -251,9 +253,8 @@ async def test_overwrite_existing(
     )
     assert response.status_code == 200, response.content
     body = response.json()
-    assert body == {"bootstrap_url": ANY}
+    assert body == {"bootstrap_url": ANY, "bootstrap_url_as_http_redirection": ANY}
     new_bootstrap_token = ParsecOrganizationBootstrapAddr.from_url(body["bootstrap_url"]).token
-    assert new_bootstrap_token != bootstrap_token.hex
```
FirelightFlagboy, line 256: **issue:** Revert that, that check is important

FirelightFlagboy:
With this PR, we need to also test when the server does not have a webapp configured

server/tests/administration/test_create_organization.py
touilleMan: good catch ! done. :+1:

touilleMan:
> Missing newsfragment

Done :+1:

PR #12236, opened 2026-02-19
docs(RFC): Add RFC for the mutual auth for Idopte

FirelightFlagboy:
Rendered: https://github.com/Scille/parsec-cloud/blob/rfc-pki-web-idopte-auth/docs/rfcs/1026-pki-web-idopte-mutual-auth.md

### DIFF
--- a/docs/rfcs/1026-pki-web-idopte-mutual-auth.md
+++ b/docs/rfcs/1026-pki-web-idopte-mutual-auth.md
@@ -1,0 +1,174 @@
+<!-- Parsec Cloud (https://parsec.cloud) Copyright (c) BUSL-1.1 2016-present Scille SAS -->
+
+# PKI Web Idopte Mutual Authentication
+
+## Overview
+
+Describe how we would add the required bits to be able to execute the mutual authentication to Idopte middleware.
+
+## Background & Motivation
+
+To provide PKI support in a web context, we choose the use Idopte solution.
+
+Idopte provide a middleware installed on the client side that provide an HTTP API & JS SDK to provide PKCS11 API in a web context. (see https://www.idopte.com/doc_ug.php)
+
+The middleware API require to be authenticated with it.
+The process will authenticate both parties, The backend server & the middleware using TLS certificates.
+
+### Terms
+
+- **Server**: The Parsec backend server
+- **Client**: The Parsec web application
+- **Service**: The Idopte SCWS middleware/driver installed on the client device.
+
+## Goals and Non-Goals
+
+We want to define the required API and approach to authenticate with the middleware.
+
+## Design
+
+Below is the workflow to authenticate with SCWS:
+
+```mermaid
+sequenceDiagram
+  autonumber
+  participant M as SCWS
+  participant C as Client
+  participant S as Server
+
+  C ->> M: Look to start the authentication process with a certificate & challenge
+  M ->> C: Respond with the signed challenge and a new challenge for the server
+  C ->> S: Transmit the signed middleware challenge and server challenge to server
+  S ->> S: Verify the middleware signature against known Idopte public key and original challenge
+  S ->> S: Sign the provided challenge
+  S ->> C: Transmit the signed challenge that was provided by the middleware
+  C ->> M: Send the server signature to complete the authentication
+```
+
+
+> [!NOTE]
+> Omitted from the diagram:
+>
+> - The certificate first sent to the middleware is obtained by the client.
+> - The server know the valid public key used by Idopte.
+> - The server is configured with the private key related to the certificate that the client use.
+
+## Operations
+
+### Configuring the Server to Work with Idopte
+
+We want the configuration of SCWS to be optional (to not require every on-prem clients to be forced to have a contract with Idopte).
+
+The server needs to have 2 things:
+
+- The public key that Idopte is using
+- The private key that was used to generate the certificate
+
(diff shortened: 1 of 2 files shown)


FirelightFlagboy:
Will update the API command to the json5 format if they look good enough

docs/rfcs/1026-pki-web-idopte-mutual-auth.md
```diff
@@ -0,0 +1,140 @@
...
+Idopte provide a middleware installed on the client side that provide an HTTP API & JS SDK to provide PKCS11 API in a web context.
+
+The middleware API require to be authenticated with it.
+The process will authenticate both parties, The backend server & the middleware using TLS certificates.
+
+### Terms
+
+- **Server**: The Parsec backend server
+- **Client**: The Parsec web application
+- **Middleware**: The Idopte middleware/driver installed on the client device.
```
touilleMan: From the `scwsapi.js`, it seems Idopte themselves call the middleware the "SCWS service" so I think we should follow the same wording

docs/rfcs/1026-pki-web-idopte-mutual-auth.md
```diff
@@ -0,0 +1,140 @@
...
+
+The server needs to have 2 things:
+
+- The public key that Idopte is using
+- The private key that was used to generate the certificate
+
+Naturally, we would add 2 options on the server CLI:
+
+```
+--idopte-public-keys <PEM_FILE>
```
touilleMan: should replace all "idopte" prefix with "scws" (since Idopte is the company and SCWS the actual product we want to integrate)

docs/rfcs/1026-pki-web-idopte-mutual-auth.md
```diff
@@ -0,0 +1,140 @@
...
+```
+
+> [!NOTE]
+> The certificate is a base64 blob but should not be processed in any way
+> and should be sent as is to the middleware.
+
+The client would then retrieve the certificate using the following anonymous command:
+
+```yaml
+cmd: idopte_certificate
```
touilleMan: A dedicated command is needed here (instead of returning this certificate in `idopte_challenge`) since this certificate in needed at the very first step of authentication ? (i.e. before `idopte_challenge` is called)

FirelightFlagboy: Yes, the client to be able to start the auth process need to provide a valid certificate and a challenge that will be signed by the service

docs/rfcs/1026-pki-web-idopte-mutual-auth.md
FirelightFlagboy: done

docs/rfcs/1026-pki-web-idopte-mutual-auth.md
FirelightFlagboy: done

docs/rfcs/1026-pki-web-idopte-mutual-auth.md
```diff
@@ -0,0 +1,140 @@
...
+
+## Overview
+
+Describe how we would add the required bits to be able to execute the mutual authentication to Idopte middleware.
+
+## Background & Motivation
+
+To provide PKI support in a web context, we choose the use Idopte solution.
+
+Idopte provide a middleware installed on the client side that provide an HTTP API & JS SDK to provide PKCS11 API in a web context.
```
AureliaDolo:
```suggestion
Idopte provide a middleware installed on the client side that provide an HTTP API & JS SDK to provide PKCS11 API in a web context. (see https://www.idopte.com/doc_ug.php)
```

docs/rfcs/1026-pki-web-idopte-mutual-auth.md
```diff
@@ -0,0 +1,140 @@
...
+
+```
+--scws-service-cert <FILE>
+```
+
+> [!NOTE]
+> The certificate is a base64 blob but should not be processed in any way
+> and should be sent as is to the middleware.
+
+The client would then retrieve the certificate using the following anonymous command:
```
AureliaDolo, line 94: We would need to limit the number of request one can make.

FirelightFlagboy: Why, the certificate is public, it's not a resource that need protection.

If you think about an user flooding the server, it's not the right solution

AureliaDolo: > If you think about an user flooding the server, it's not the right solution

That what I had in mind.

docs/rfcs/1026-pki-web-idopte-mutual-auth.md
```diff
@@ -0,0 +1,140 @@
...
+```
+--scws-public-keys <PEM_FILE>
+--scws-service-key <PEM_FILE>
+```
+
+> [!IMPORTANT]
+> The Idopte public keys files can contain multiple public key.
+> All should be available to verify Idopte challenge signature.
+
+### How the Client Retrieve the Certificate
```
AureliaDolo:
```suggestion
### How does the Client retrieve the Certificate
```
```suggestion
### How the Client retrieves the Certificate
```

FirelightFlagboy:
> There is the cmd format that need to be updated (yaml -> json5) but otherwise LGTM

Should be good now :)

PR #12253, opened 2026-02-23
refactor(editor): upgrade CryptPad to winter version

NicoTuxx:
Changes:
- Add ManualSave command to CryptPad communication API
- Add CryptpadSession interface (controller + manualSave)
- Auto-save on initial unsaved changes (OO conversion artifacts)
- Save automatically when leaving the editor (modal only on failure)
- Skip save attempt when document never loaded (corrupted files)
- Add ref on dynamic component for parent-to-child save calls
- Remove file-type-specific timeout handling

### DIFF
--- a/client/src/services/cryptpad.ts
+++ b/client/src/services/cryptpad.ts
@@ -14,6 +14,7 @@ namespace CryptpadCommAPI {
     Open = 'editics-open',
     OpenResult = 'editics-open-result',
     Event = 'editics-event',
+    Save = 'editics-save',
   }
 
   export enum Events {
@@ -135,11 +136,16 @@ function cryptpadLog(level: 'debug' | 'warn' | 'error' | 'info', message: string
   window.electronAPI.log(level, `[Cryptpad] ${message}`);
 }
 
+export interface CryptpadSession {
+  controller: AbortController;
+  save: () => void;
+}
+
 export async function openDocument(
   options: OpenDocumentOptions,
   handlers: CryptpadEventHandlers,
   frame: HTMLIFrameElement,
-): Promise<AbortController | undefined> {
+): Promise<CryptpadSession | undefined> {
   const CRYPTPAD_SERVER = Env.getDefaultCryptpadServer();
 
   function sendMessageToFrame(command: CryptpadCommAPI.Commands, data?: any): void {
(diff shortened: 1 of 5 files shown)


client/src/services/cryptpad.ts
```diff
@@ -14,6 +14,7 @@ namespace CryptpadCommAPI {
     Open = 'editics-open',
     OpenResult = 'editics-open-result',
     Event = 'editics-event',
+    ManualSave = 'editics-manual-save',
```
Max-7: You can just call this `save`, no point in calling it `manualSave`. It probably won't even be manual most of the time.

client/src/views/files/handler/editor/FileEditor.vue
```diff
@@ -376,6 +378,26 @@ async function openTimeoutModal(): Promise<'wait' | 'close'> {
   // Modal was dismissed (close button) - just close without navigating
   return 'close';
 }
+
+async function manualSave(): Promise<boolean> {
+  if (!session) {
+    return false;
+  }
+  return new Promise<boolean>((resolve) => {
+    const SAVE_TIMEOUT_MS = 10000;
```
Max-7: Should check that `pendingResolve` doesn't already has a value. If we're calling save() multiple time before one has resolved, it may become a bit unstable.

Max-7: 10s is a long time to wait for a user who wants to leave the document.

client/src/views/files/handler/editor/FileEditor.vue
```diff
@@ -225,6 +228,11 @@ async function loadEditor(): Promise<void> {
                 } else {
                   emits('onSaveStateChange', SaveState.Error);
                 }
+                // Resolve pending manual save promise
+                if (pendingSaveResolve) {
```
Max-7, line 232: Not a big fan of doing a bit of logic inside onSave() and another bit of logic inside `onHasUnsavedChanges`. Ideally we would want `onSave` to just do the saving part, and the state to change only with `onHasUnsavedChanges`. But I know this can be a bit of a mess and we're already emitting `onSaveStateChange` here so I'm not sure what really can be done.

client/src/views/files/handler/editor/FileEditor.vue

client/src/services/cryptpad.ts
NicoTuxx: done

Max-7 (approved):

PR #12274, opened 2026-03-02
Add missing newsfragments for version 3.8.0

Ironicbay:


newsfragments/12021.api.rst
FirelightFlagboy, line 1: **todo:** Revert, this is not an API change since the CLI is exposed to the client (given it's a special client)

newsfragments/12152.feature.rst
FirelightFlagboy, line 1: **nitpick:** For me it's a feature 🤷

newsfragments/12235.misc.rst
FirelightFlagboy, line 1: **polish:** It's not a bug, use a different kind

newsfragments/9633.misc.rst
FirelightFlagboy, line 1: **nitpick:** It's an added feature to the CLI client

PR #12280, opened 2026-03-03
[MS] Adds a search feature for files inside a workspace

Max-7:
The input field is disabled with a `v-if`, it shouldn't be a problem for Playwright since it's not interset into the DOM and the user can't interact with it.


client/src/components/files/explorer/FileSearchResults.vue
```diff
@@ -0,0 +1,58 @@
+<!-- Parsec Cloud (https://parsec.cloud) Copyright (c) BUSL-1.1 2016-present Scille SAS -->
+
+<template>
+  <div v-show="active">
+    <ms-spinner />
+  </div>
+  <div v-if="searchResults.length === 0">NO FILES FOUND</div>
```
NicoTuxx: should use i18n

client/src/components/files/explorer/FileSearchResults.vue
```diff
@@ -0,0 +1,58 @@
...
+  <div v-else>
+    {{ searchResults.length }} RESULTS FOUND
```
NicoTuxx: should use i18n

client/src/components/files/explorer/FileSearchResults.vue
```diff
@@ -0,0 +1,58 @@
...
+.file-name {
+  font-weight: bold;
+  text-transform: uppercase;
+}
+
+.file-path {
+  font-style: italic;
+  font-size: 1rem;
+  color: #7a7a7a;
+}
```
NicoTuxx: should use lib color variables

client/src/components/files/explorer/FileSearchResults.vue
Max-7: I've removed the CSS but I'd rather not add I18n entries into the translation files because we won't use anything from that template, it's going to get completely reworked, and chances are when we do, we'll forget to remove the entries from the locales.

NicoTuxx (approved):

PR #12284, opened 2026-03-04
POC File search in GUI for workspace & workspace history

touilleMan:


mmmarcos:
File search was implemented in:

- #12415

@touilleMan should we close this PR?

PR #12314, opened 2026-03-05
Server static files `Cache-Control` header

touilleMan:


server/parsec/asgi/__init__.py
```diff
@@ -42,10 +43,31 @@
...
+        # apply a one-size-fits-all configuration here.
+        response.headers["Cache-Control"] = "max-age=31536000, public, immutable"
+        return response
+
+
 # This class is used to serve the static files of the web app (SPA)
 # and redirect to the index.html if file not found on server.
 # Useful when the user refresh the page or access a route directly.
-class StaticFilesWithSPARedirect(StaticFiles):
+class StaticFilesWithSPARedirect(StaticFilesWithCacheControl):
```
FirelightFlagboy: **issue:** Does not seems to handle the case where the content of `index.html` change

touilleMan: o_O very good point !

I've fixed this by having it inherit of `StaticFiles` (so no cache for this)

server/parsec/asgi/__init__.py
```diff
@@ -81,21 +104,42 @@ def app_factory(
     templates = Jinja2Templates(env=backend.config.jinja_env)
 
     if with_client_web_app:
```
FirelightFlagboy, line 106: **nitpick:** Would be benefical to add a small comment explaining that we do not set a cache control for `index.html` and the difference of the const `BASE_URL` vs `ASSESTS_URL`

touilleMan: good point :+1: , done.

touilleMan:
> Need to update the custom brandind documentation

I've updated `docs/hosting/custom_branding.rst` (and the translation) :+1:

server/parsec/asgi/__init__.py
```diff
@@ -81,21 +104,42 @@ def app_factory(
...
+            class NoCustomDirAlways404:
+                async def __call__(self, scope: Scope, receive: Receive, send: Send) -> None:
+                    raise HTTPException(status_code=404)
+
+            custom_assets_app = NoCustomDirAlways404()
+        app.mount(WEB_APP_CUSTOM_ASSETS_URL, custom_assets_app)
 
         app.mount(
             WEB_APP_BASE_URL, StaticFilesWithSPARedirect(directory=with_client_web_app, html=True)
         )
```
FirelightFlagboy, line 134: **question:** How does fastapi handle route dispatch ?

With the PR we have two URLs having the same base: `/client` (BASE_URL & ASSESTS_URL).
Does fastapi does the FDFU (first defined, first used) ?

client/vite.config.ts
```diff
@@ -140,7 +141,25 @@ const config: UserConfigFnObject = (_env: ConfigEnv) => ({
...
+      name: 'serve-dist-custom',
+      configureServer(server) {
+        server.middlewares.use('/custom', (req, res, _next) => {
+          const filePath = path.join(__dirname, 'dist/custom', req.url ?? '/');
+          if (fs.existsSync(filePath) && fs.statSync(filePath).isFile()) {
+            res.end(fs.readFileSync(filePath));
+          } else {
+            res.statusCode = 404;
+            res.end();
+          }
```
FirelightFlagboy, line 158: **nitpick:** You can return a promise, so use the async node API

docs/hosting/custom_branding.rst
```diff
@@ -7,12 +7,40 @@ Parsec application can be partially customized to better fit your brand.
...
 
-- In web mode, the files have to be placed inside a ``custom`` folder in the server's static files directory. A file should be reachable by doing ``/custom/<file>`` on the server.
-- In desktop mode, the files have to be placed inside a ``custom`` folder in the configuration directory. The configuration directory will depend on the OS.
+Native app
+~~~~~~~~~~
 
-    - Windows: ``%APPDATA%/parsec3/app``
-    - Linux ``~/.config/parsec3/app``
-    - MacOS ``~/Library/Application Support/parsec3/app``
+In native mode, Parsec always looks for custom branding files at startup.
```
FirelightFlagboy: **nitpick:** 
```suggestion
The Parsec desktop application always looks for custom branding files at startup.
```

client/src/services/environment.ts
```diff
@@ -144,10 +144,34 @@ function isAccountAutoLoginEnabled(): boolean {
...
+  // Allowed values:
+  // - `false`: Custom branding is disabled
+  // - `true,cacheBustingSuffix=...`: Custom branding is enabled, note the mandatory cache
+  //   busting suffix parameter that will be added to the resources names (e.g. using
+  //   `cacheBustingSuffix=-xxx` means `custom_en-US-xxx.json` will be fetched).
+  if (obj === null || obj.content === 'false') {
+    return [false, ''];
+  }
+  const expectedPrefix = 'true,cacheBustingSuffix=';
+  if (!obj.content.startsWith(expectedPrefix)) {
```
FirelightFlagboy, line 172: **polish:** I would only enable the custom branding when `cacheBustingSuffix` is defined

docs/hosting/custom_branding.rst
touilleMan: :+1: done.

client/src/services/environment.ts
touilleMan: This makes thinks less clear:
- For the user customizing `custom-branding`, setting it to an empty value becomes unclear if it disable it or not.
- The code reading the `custom-branding` value has to return a `cacheBustingSuffix` no matter what, and an empty value sometime the right one (typically in native mode we don't need cache busting). So having a different special case for empty `cacheBustingSuffix` is error prone.
- Using an empty `cacheBustingSuffix` in web mode can still be useful, typically in dev mode where we won't rename the custom files every time we modify them (and instead rely on disabling the cache in the browser)

So that's why I settle on this extra clear boolean+suffix configuration pattern

client/vite.config.ts
touilleMan: :+1: done.

(I also removed the extra check on exists + isFile, since having an error during `fs.promises.readFile` seems good enough to consider the file is not available ^^)

server/parsec/asgi/__init__.py
touilleMan: >  Does fastapi does the FDFU (first defined, first used) ?

yes, see `Route priority` in https://starlette.dev/routing/

I've added a comment about this in the code :+1:

client/src/services/environment.ts
```diff
@@ -144,10 +144,34 @@ function isAccountAutoLoginEnabled(): boolean {
  Custom branding
 */
 
-const CUSTOM_BRANDING_ENV_VARIABLE = 'PARSEC_APP_ENABLE_CUSTOM_BRANDING';
+function isCustomBrandingEnabled(): [boolean, string] {
+  // On non-web platforms, custom branding is always considered enabled (without
```
Max-7: It does need to be disabled for CSPN builds though, which was the point of the original `ENABLE_CUSTOM_BRANDING` variable. @mmmarcos, your thoughts?

client/src/services/resourcesManager.ts
```diff
@@ -2,14 +2,40 @@
...
-import axios from 'axios';
+import axios, { AxiosInstance } from 'axios';
 
 enum Resources {
-  LogoFull = 'logo.svg',
-  LogoIcon = 'logo_icon.svg',
-  HomeSidebar = 'home_sidebar.png',
-  TranslationEnUs = 'custom_en-US.json',
-  TranslationFrFr = 'custom_fr-FR.json',
+  LogoFull,
```
Max-7: I would force a string value, like `LogoFull = 'logo-full'`, because it makes debugging a bit easier (for example if you want to print the value in `resolveResource()`, seeing `logo-full` helps more than seeing `0`, and it would also solve your problem when iterating over the values (it prevents the (ridiculous) reverse mapping).

Max-7 (changes requested):

client/src/services/environment.ts
mmmarcos: This change has no impact on the current CSPN (based on v3.7.10). However, for future CSPN releases, we would probably need to update the Security Target document (Cible) to explain how this interface works (loading of custom resources on Windows client).

So, I would keep the `ENABLE_CUSTOM_BRANDING` as a flag to be able to force disabling it. Is there a reason why not keeping it?

client/src/services/resourcesManager.ts
touilleMan: :+1: done.

(I've also replaced the enum by a dict with `type Resources = typeof Resources[keyof typeof Resources];` since it gives better type informations)

client/src/services/environment.ts
touilleMan: I've modified the code to keep the `PARSEC_APP_ENABLE_CUSTOM_BRANDING` env variable (and document it is not needed on web)

Max-7 (approved):

PR #12329, opened 2026-03-06
Libparsec misc fixes

touilleMan:


libparsec/crates/client/src/workspace_history/transactions/fd_read.rs
```diff
@@ -83,12 +83,18 @@ pub async fn fd_read(
...
-    let involved_blocks = manifest
-        .blocks
-        .iter()
-        // TODO: use binary search to avoid O(n) complexity
-        .skip_while(|block| block.offset + block.size.get() <= offset)
-        .take_while(|block| block.offset < end);
+    let involved_blocks = {
+        let start_idx = manifest
+            .blocks
+            .partition_point(|block| block.offset + block.size.get() <= offset);
```
FirelightFlagboy, line 89: **question:** 

1. Do we have a garantee about `blocks` being partitioned by `block.offset + block.size` ? (We do in sorted insertion ?)
2. Do we have a garantee about non-overlapping block ?

touilleMan: yes and yes, see the comment line 80 ;-)

```rust
    // File manifest blocks comes with two important guarantees:
    // - The blocks are sorted by offset, with no overlap.
    // - Each block has a `blocksize` size, except for the last one that can be smaller.
```

PR #12348, opened 2026-03-09
Workspace chain of command: RFC 1027

touilleMan:


FirelightFlagboy (commented):
LGTM but would like to have more details about how we would handle concurrency

AureliaDolo (approved):
LGTM but I have a few questions :
- How would that work (roughly) from a GUI point of view ?
- In the sharing and roles modal, if applicable there will be a button "self promote" that one may find if they happen to be looking for it?
- Or would it be something in the notification panel ("workspace xxx has no remaining owners, you are eligible to take over it") 
- Or a modal when you enter a workspace in this situation ? (maybe here it would be worth knowing if you are the only one able to take this role or if someone else might be available) 
- Either way, should there be an api to check for that situation ?

touilleMan:
> LGTM but would like to have more details about how we would handle concurrency

Done

> How would that work (roughly) from a GUI point of view ?

I've added a part about that in the overview. Also I don't think we need a dedicated event exposed to the GUI but instead just to ensure the `WorkspacesSelfListChanged` gets triggered whenever the workspace becomes orphaned

FirelightFlagboy (commented):
LGTM

docs/rfcs/1026-workspace-chain-of-command.md
```diff
@@ -0,0 +1,234 @@
+<!-- Parsec Cloud (https://parsec.cloud) Copyright (c) BUSL-1.1 2016-present Scille SAS -->
+
+# [RFC 1027] Workspace Chain of Command
+
+## 1 - Overview
+
+When all OWNERs of a workspace are revoked, the workspace becomes orphaned: no one can manage roles, perform key rotation etc.
+This RFC introduces a **self-promotion** mechanism allowing the highest-ranked remaining user to claim the OWNER role.
```
mmmarcos:
```suggestion
This RFC introduces a **self-promotion** mechanism allowing one of the highest-ranked remaining user to claim the OWNER role.
```

docs/rfcs/1026-workspace-chain-of-command.md
```diff
@@ -0,0 +1,234 @@
...
+
+From the GUI point of view, this feature would work as follow:
+
+1. In the workspace list, an orphaned workspace (where the current user is eligible to self-promote) is shown with a
+   warning indicator signaling that the workspace has no active owner.
+2. A dedicated action (e.g. a button or context-menu entry labelled **"Claim ownership"** or **"Become owner"**)
+   is displayed alongside the workspace entry.
+3. When the user triggers that action, a confirmation dialog explains the situation
+   ("All owners of this workspace have been revoked. As the highest-ranked remaining member,
```
mmmarcos:
```suggestion
   ("All owners of this workspace have been revoked. As one of the highest-ranked remaining member,
```

docs/rfcs/1026-workspace-chain-of-command.md
```diff
@@ -0,0 +1,234 @@
...
+                    "type": "Bytes"
+                }
+            ]
+        },
+        "reps": [
+            {
+                "status": "ok"
+            },
+            {
+                // Author is not among the workspace's active members with the highest role.
```
mmmarcos: ⚠️ 
```suggestion
                // Author is not among the workspace's active members with the highest role.
                // Or author is External which does not allow to be Owner of any workspace.
```

docs/rfcs/1026-workspace-chain-of-command.md
```diff
@@ -0,0 +1,234 @@
...
+    client: Handle,
+    realm_id: VlobID,
+) -> Result<(), ClientSelfPromoteToWorkspaceOwner> {
+    // ...
+}
+```
+
+> [!NOTE]
+> We introduce this new function instead of using `client_share_workspace` to
+> simplify the usage of this API (intent in the function name, more precise
```
mmmarcos:
```suggestion
> simplify the use of this API (intent in the function name, more precise
```

docs/rfcs/1026-workspace-chain-of-command.md
```diff
@@ -0,0 +1,234 @@
...
+
+No change needed: the self-promotion uses the existing `RealmRoleCertificate`
+
+## 5 - Certificate Validation
+
+The `RealmRoleCertificate` validation algorithm should be updated (both on client and server side).
+
+This is a breaking change (an older Parsec client will consider a self-promotion certificate to be invalid), though this is considered acceptable since:
+
+- Revoking all OWNERs in a workspace is a unlikely event, so there should be ample time for all clients to update to a newer version of Parsec.
```
mmmarcos:
```suggestion
- Revoking all OWNERs in a workspace is a unlikely event, so there should be ample time for all users to update to a newer version of Parsec.
```

docs/rfcs/1026-workspace-chain-of-command.md
```diff
@@ -0,0 +1,234 @@
...
+- The issue is present only if some clients use the newer and some others the older version. In such case it should be easy enough
```
mmmarcos:
```suggestion
- The issue is present only if some users use the newer version and others the older version. In such case it should be easy enough
```

mmmarcos (approved):

docs/rfcs/1026-workspace-chain-of-command.md
touilleMan: :+1: done.

docs/rfcs/1026-workspace-chain-of-command.md
touilleMan: :+1: done.

docs/rfcs/1026-workspace-chain-of-command.md
touilleMan: :+1: done.

docs/rfcs/1026-workspace-chain-of-command.md
touilleMan: :+1: done.

docs/rfcs/1026-workspace-chain-of-command.md
touilleMan: :+1: done.

docs/rfcs/1026-workspace-chain-of-command.md
touilleMan: OUTSIDER can do self-promotion, I've added the following note about this:
```
> - Under normal condition, a user with OUTSIDER profile cannot be given OWNER
>   or MANAGER role. However self-promotion is allowed for OUTSIDER.
>   This is because it is only allowed once only OUTSIDERs remain in the
>   workspace (highly unlikely), and OUTSIDERs can already be OWNER/MANAGER of
>   a workspace (if they have become OUTSIDER *after* the role was given to them).
```

FirelightFlagboy (changes requested):
bump the RFC version (the seat 1026 is already taken)

FirelightFlagboy (approved):

PR #12399, opened 2026-03-16
Add CryptPad server URL to `server_config` command

### LINKED ISSUE #12398: Update `server_config` command to include CryptPad server URL

mmmarcos:
Closes #12398

### DIFF
--- a/libparsec/crates/protocol/schema/anonymous_server_cmds/server_config.json5
+++ b/libparsec/crates/protocol/schema/anonymous_server_cmds/server_config.json5
@@ -20,6 +20,11 @@
                     {
                         "name": "openbao",
                         "type": "OpenBaoConfig"
+                    },
+                    {
+                        "name": "cryptpad",
+                        "type": "NonRequiredOption<CryptPadConfig>",
+                        "introduced_in": "3.9"
                     }
                 ]
             }
@@ -164,6 +169,26 @@
                         // see https://openbao.org/api-docs/auth/jwt/#oidc-authorization-url-request
                         "name": "mount_path",
                         "type": "String"
+                    }
+                ]
+            },
+            {
+                "name": "CryptPadConfig",
+                "discriminant_field": "type",
+                "variants": [
+                    {
+                        "name": "Disabled",
+                        "discriminant_value": "DISABLED"
+                    },
+                    {
+                        "name": "Enabled",
+                        "discriminant_value": "ENABLED",
+                        "fields": [
+                            {
+                                "name": "server_url",
+                                "type": "String"
+                            }
+                        ]
                     }
                 ]
             }
--- a/libparsec/crates/protocol/tests/anonymous_server_cmds/v5/server_config.rs
+++ b/libparsec/crates/protocol/tests/anonymous_server_cmds/v5/server_config.rs
@@ -39,6 +39,8 @@ pub fn req() {
 pub fn rep_ok() {
     let raw_expected = [
         (
+            // Legacy format from Parsec < 3.9, missing the `cryptpad` field
+            //
             // Generated from Parsec 3.5.3-a.0+dev
             // Content:
             //   status: 'ok'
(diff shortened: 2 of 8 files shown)


libparsec/crates/protocol/tests/anonymous_server_cmds/v5/server_config.rs
```diff
@@ -39,30 +39,58 @@ pub fn req() {
...
                 account: anonymous_server_cmds::server_config::AccountConfig::Disabled,
                 organization_bootstrap: anonymous_server_cmds::server_config::OrganizationBootstrapConfig::WithBootstrapToken,
                 openbao: anonymous_server_cmds::server_config::OpenBaoConfig::Disabled,
+                cryptpad: anonymous_server_cmds::server_config::CryptPadConfig::Enabled {
+                    server_url: "https://cryptpad.parsec.invalid".to_string(),
+                },
             }
         ),
         (
             // Legacy format from Parsec < 3.8, missing the `transit_mount_path` field
```
mmmarcos, line 63: Not sure what this line actually means:
- Does it mean that the binary somehow did not contain `transit_mount_path`?
- Or is it just there to explain that `transit_mount_path` was added later and it is OK if its value is None?

In any case, I updated the binary for this test case with the `test_expected_payload_cooker.py` script to match what's expected.

libparsec/crates/protocol/tests/anonymous_server_cmds/v5/server_config.rs
FirelightFlagboy, line 1: **issue:** You should not update existing blob (to check for regression of the schema), but add new one

libparsec/crates/protocol/schema/anonymous_server_cmds/server_config.json5
```diff
@@ -166,6 +170,26 @@
                         "type": "String"
                     }
                 ]
+            },
+            {
```
FirelightFlagboy, line 175: **issue:** Should add an `introduced_in` field

FirelightFlagboy, line 175: **nitpick:** Could be reduced to an Optional

If present, cryptpad is enable else disabled

mmmarcos: Here or in the fields above?

BTW, the only example of `introduced_in` I found is the `transit_mount_path` field in this same command which uses "3.8" as value. Is that Parsec version? shouldn't be related to the API protocol version instead?

mmmarcos: Yup, I wanted to discuss with @Max-7 and @NicoTuxx if there may be other CryptPad-related configuration options that we may want to use right away.

FirelightFlagboy: My comment is on line 174, so here.

The content of `introduced_in` is only used for documentation purpose (I've tried in the past to make it have a functional meaning, but ...).
Currently, it seems the value of such field correspond to the version of Parsec and not the version of the API

libparsec/crates/protocol/tests/anonymous_server_cmds/v5/server_config.rs
mmmarcos: I understand but I might be missing something: if I do not update existing blob the tests does not pass. 

My guess is that serialization does not convert the _absence_ of the `cryptpad` field into a `Disabled` state.
```
cryptpad: anonymous_server_cmds::server_config::CryptPadConfig::Disabled,
```

Does the `introduced_field` has something to do with it? I tried adding it to the `cryptpad` field but serialization test still fails.

FirelightFlagboy: > Does the introduced_field has something to do with it? 

I don't think so.

To fix your problem, look how `transit_mount_path` type is defined :wink:

mmmarcos: Done :pray:

PR #12410, opened 2026-03-17
Workspace chain of command: protocol schemas

touilleMan:

PR #12411, opened 2026-03-17
Workspace chain of command: Server implementation

touilleMan:
close #12410


server/parsec/components/memory/realm.py
```diff
@@ -423,6 +427,133 @@ async def unshare(
...
+                author_device = org.devices[author]
+            except KeyError:
+                return RealmSelfPromoteToOwnerStoreBadOutcome.AUTHOR_NOT_FOUND
+            author_user_id = author_device.cooked.user_id
+
+            author_user = org.users[author_user_id]
+            if author_user.is_revoked:
+                return RealmSelfPromoteToOwnerStoreBadOutcome.AUTHOR_REVOKED
+
+            if author_user.current_profile == UserProfile.OUTSIDER:
```
AureliaDolo, line 486: Is there a special case for when the only people left in a workspace are outsiders ? Or are we best effort on that feature and if there are only outsiders they are forced to live in that workspace for ever and ever.

server/parsec/components/memory/realm.py
```diff
@@ -423,6 +427,133 @@ async def unshare(
...
+                        case RealmRole.READER:
+                            return 1
+                        case RealmRole.CONTRIBUTOR:
+                            return 2
+                        case RealmRole.MANAGER:
+                            return 3
+                        case RealmRole.OWNER:
+                            return 4
+                        case unknown:
+                            # TODO: Implement `Enum` on `RealmRole` so we can use `assert_never` here
```
AureliaDolo: is an issue missing or is it already in another PR ? ^^

server/parsec/components/memory/realm.py
```diff
@@ -423,6 +427,133 @@ async def unshare(
...
+            try:
+                realm = org.realms[certif.realm_id]
+            except KeyError:
+                return RealmSelfPromoteToOwnerStoreBadOutcome.REALM_NOT_FOUND
+
+            realm_topic = ("realm", certif.realm_id)
+
+            async with org.topics_lock(write=[realm_topic]) as (realm_topic_last_timestamp,):
+
+                def _role_to_priority(role: RealmRole) -> int:
```
AureliaDolo: That function is duplicated in the postgre version. Is there a reason to not share it ?

(I guess that if RealmRole becomes an enum we could implement that function in this class)

server/parsec/components/memory/realm.py
AureliaDolo: I haven't looked at the libparsec part yet but I assume that would be a type you'd want to share and define in libparsec_types ?

server/parsec/components/memory/realm.py
AureliaDolo: (or add repr(u8) and then boom magic ! 🪄 )

server/parsec/components/memory/realm.py
```diff
@@ -423,6 +427,133 @@ async def unshare(
...
+
+                if realm.last_vlob_timestamp is not None:
+                    last_timestamp = max(
+                        common_topic_last_timestamp,
+                        realm_topic_last_timestamp,
+                        realm.last_vlob_timestamp,
+                    )
+                else:
+                    last_timestamp = max(common_topic_last_timestamp, realm_topic_last_timestamp)
+                if last_timestamp >= certif.timestamp:
```
AureliaDolo, line 542: Should it be > ?  Just in (the improbable) case there is a certif with exactly the same timestamp ?

server/parsec/components/memory/realm.py
touilleMan: see the RFC ^^

> In the unlikely event where a workspace requiring self-promotion ends up with
> only OUTSIDERs users, then an organization ADMIN should first promote one of
> them to a STANDARD profile.
> The newly STANDARD user can then proceed to the self-promotion, then typically
> share the workspace to a more legitimate user and then be reverted back into
> OUTSIDER with non-OWNER access to the workspace.

server/parsec/components/memory/realm.py
touilleMan: > That function is duplicated in the postgre version. Is there a reason to not share it ?

👍 done.

> (I guess that if RealmRole becomes an enum we could implement that function in this class)

I'd rather not do that for the moment since we only need this priority thing here. If this becomes more common we may re-consider this ;-)

server/parsec/components/memory/realm.py
touilleMan: Certificate timestamps must be strictly growing, the only exception is for certificates that are uploaded together (i.e. new user means uploading both a user certificate and a device certificate).
This is important since otherwise the last certificate timestamp is a point in time whose state can evolve if a new certificate with the same timestamp is later uploaded.

server/parsec/components/memory/realm.py
touilleMan: I have created issue https://github.com/Scille/parsec-cloud/issues/12725 and added a link to it in the TODO comments :+1:

server/parsec/components/memory/realm.py
AureliaDolo: So admin intervention ^

server/tests/api_v5/authenticated/test_realm_self_promote_to_owner.py
```diff
@@ -0,0 +1,385 @@
...
+) -> None:
+    match kind:
+        case "not_a_member":
+            # In wksp1, Bob is READER and Alice is the active OWNER
+            await bob_becomes_admin_and_changes_alice(coolorg, backend, new_alice_profile=None)
+            author = coolorg.mallory
+
+        case "lower_role_than_highest":
+            # In wksp1, Bob is READER and Alice is the active OWNER
+            # Update Bob role twice, this way he used to be higher than Mallory
```
AureliaDolo, line 130: I don't get what's the impact of bob having been manager

AureliaDolo (approved):

server/tests/api_v5/authenticated/test_realm_self_promote_to_owner.py
touilleMan: This test is about a workspace that have lost its OWNER (so somebody should self-promote) but our current user is not the highest remaining role (so he shouldn't be the one to self-promote)

To cover more things in this test, we use a user that used to have the higher role, this way we also check if the SQL request is broken and doesn't only check the last role for a given user

server/parsec/components/memory/realm.py
touilleMan: I have reconsidered this (given I had to put two other role-to-priority functions in libparsec :laughing: ), so next PR introduce ordering for `RealmRole` and `UserProfile` \o/

PR #12415, opened 2026-03-17
[MS] Implements file search

Max-7:


client/src/components/files/explorer/DocumentFilterPopover.vue
```diff
@@ -0,0 +1,229 @@
...
+import { ref, toRaw } from 'vue';
+
+const props = defineProps<{
+  filters: DocumentFilters;
+}>();
+
+const updatedFilters = ref<DocumentFilters>(structuredClone(toRaw(props.filters)));
+
+async function applyFilters(): Promise<void> {
+  await popoverController.dismiss({ filters: updatedFilters.value }, MsModalResult.Confirm);
```
Ironicbay, line 186: I would change this to behave similarly to the already existing filters in the app: update them as the user checks filters instead of having to validate the popover.

client/src/locales/en-US.json
```diff
@@ -1047,6 +1048,30 @@
             "seeInExplorer": "Show in explorer",
             "details": "Details",
             "edit": "Edit"
+        },
+        "search": {
+            "title": "Search results",
+            "resultsCount": "{count} result | {count} results",
+            "currentWorkspaceOnly": "Results only in the current workspace.",
```
Ironicbay, line 1055: This is misleading: either the text needs to change to specify the `current directory`, or the search needs to be workspace-wide regardless of the current directory.

client/src/locales/fr-FR.json
```diff
@@ -1047,6 +1048,30 @@
             "seeInExplorer": "Afficher dans l'explorateur",
             "details": "Détails",
             "edit": "Éditer"
+        },
+        "search": {
+            "title": "Résultats de la recherche",
+            "resultsCount": "{count} résultat | {count} résultats",
+            "currentWorkspaceOnly": "Résultats uniquement dans l’espace de travail actuel.",
```
Ironicbay, line 1055: Same issue.

client/src/locales/en-US.json
```diff
@@ -1047,6 +1048,30 @@
...
+                "documents": "Document",
+                "spreadsheets": "Spreadsheet",
+                "presentations": "Presentation",
+                "pdfDocuments": "PDF",
+                "audios": "Audio",
+                "videos": "Video",
+                "images": "Image",
+                "texts": "Text"
+            },
+            "placeholder": "Search a file in this workspace",
```
Ironicbay, line 1070: I would change this to either `Search` or `Search for a file, folder...`

client/src/locales/en-US.json
```diff
@@ -1047,6 +1048,30 @@
...
+            "noResults": {
+                "title": "No results match your search in this workspace",
+                "subtitle": "Check the orthography or try using simpler search terms, then filter the results."
```
Ironicbay, line 1073:
```suggestion
                "subtitle": "Check the spelling or try using simpler search terms, then filter the results."
```

client/src/locales/fr-FR.json
```diff
@@ -1047,6 +1048,30 @@
...
+                "pdfDocuments": "PDF",
+                "audios": "Audio",
+                "videos": "Vidéo",
+                "images": "Image",
+                "texts": "Texte"
+            },
+            "placeholder": "Rechercher un fichier dans cet espace",
+            "noResults": {
+                "title": "Aucun résultat ne correspond à votre recherche dans cet espace",
+                "subtitle": "Vérifier l'orthographe ou essayez d'utiliser des termes de recherche plus simples, puis de filtrer les résultats. Vous ne trouvez toujours pas ce que vous cherchez ?"
```
Ironicbay, line 1073:
```suggestion
                "subtitle": "Vérifiez l'orthographe ou essayez d'utiliser des termes de recherche plus simples, puis de filtrer les résultats."
```

client/src/views/files/FoldersPage.vue
```diff
@@ -1529,7 +1599,7 @@ async function selectAll(): Promise<void> {
 }
 
 async function unselectAll(): Promise<void> {
-  manualSelection.value = true;
+  manualSelection.value = false;
```
Ironicbay: This clashes with the intended behaviour in small display. There is an `Unselect` button which unselects all and keeps the manual selection mode enabled. Use `onSelectionCancel()` instead and keep `unselectAll()` as it was.

client/src/components/files/explorer/DocumentFilterPopover.vue
Max-7: I specifically asked @fabienSvtr about that and he chose to do it this way. No backsies!

client/src/locales/en-US.json
Max-7: The search is workspace-wide regarless of the directory.

client/src/views/files/FoldersPage.vue
Max-7: Yep good catch, my mistake. I tried fixing the selection with this and decided using `onSelectionCancel` instead which already sets the manualSelection, but I forgot to reset this value.

client/src/components/files/explorer/FileSearchResultItem.vue
```diff
@@ -0,0 +1,163 @@
...
+            :title="searchItem.stats.name"
+            @click="onClick"
+          >
+            {{ searchItem.stats.name }}
+          </ion-text>
+          <ion-text
+            class="label-path can-highlight body-sm"
+            :title="searchItem.parent"
+          >
+            {{ searchItem.parent }}
```
Ironicbay, line 39: I would change this to either add the workspace name before, and / or add a `/` at the end (for non-root directories).

This would change the following to:
```
wksp1/

wksp1/zippy_azure_parrotfish/

...
```

Image

Max-7: @fabienSvtr if you want to implement this change, just be mindful that we highlight texts that match the search inside the elements that have the class `can-highlight`. So if you just add the workspace name to the ion-text, the workspace is going to get highlighted if it matches the search, which is not what we want.

fabienSvtr: I added workspace name with specific style to make it stand out.
image

fabienSvtr:
Probably a bad rebase deletes the commit to show the workspace. Patched by this PR: https://github.com/Scille/parsec-cloud/pull/12487

PR #12416, opened 2026-03-17
[MS] Get Cryptpad server from parsec server

### LINKED ISSUE #12407: Use CryptPad server URL from server config

Max-7:
Closes #12407

### DIFF
--- a/server/parsec/cli/testbed.py
+++ b/server/parsec/cli/testbed.py
@@ -58,6 +58,7 @@ from parsec.cli.utils import cli_exception_handler
 from parsec.components.memory.user import MemoryDatamodel, MemoryUserComponent
 from parsec.config import (
     BackendConfig,
+    CryptPadConfig,
     LogLevel,
     MockedBlockStoreConfig,
     MockedDatabaseConfig,
@@ -998,6 +999,7 @@ async def testbed_backend_factory(
                 ),
             ],
         ),
+        cryptpad_config=CryptPadConfig(server_url="https://cryptpad-dev.parsec.cloud"),
         # Disable the rate limit
         email_rate_limit_cooldown_delay=0,
         email_rate_limit_max_per_hour=0,


client/src/locales/en-US.json
```diff
@@ -2826,7 +2826,8 @@
                 "networkOffline": "Network connection lost",
                 "tooLongToOpen": "The file seems to be taking too long to open",
                 "frameLoadFailed": "Could not load the editor. Please check your network connection.",
-                "openFailed": "Could not open the file."
+                "openFailed": "Could not open the file.",
+                "cryptpadNotAvailable": "Your organization does not support file edition."
```
```suggestion
                "cryptpadNotAvailable": "Document editing is not available for your organization."
```

client/src/locales/fr-FR.json
```diff
@@ -2825,7 +2825,8 @@
                 "networkOffline": "Connexion réseau perdue",
                 "tooLongToOpen": "Le fichier semble mettre trop de temps à s'ouvrir",
                 "frameLoadFailed": "Impossible de charger l'éditeur. Veuillez vérifier votre connexion.",
-                "openFailed": "Impossible d'ouvrir ce fichier."
+                "openFailed": "Impossible d'ouvrir ce fichier.",
+                "cryptpadNotAvailable": "Votre organisation ne supporte pas l'édition de fichier."
```
```suggestion
                "cryptpadNotAvailable": "La fonction d'édition de documents n'est pas disponible pour votre organisation."
```

client/src/views/files/handler/editor/FileEditor.vue
Ironicbay: These advices aren't relevant in this new case, and might not be relevant in all previous cases as well. You could maybe add a flag to filter them out at least in this new scenario.

Image

mmmarcos: > These advices aren't relevant in this new case, and might not be relevant in all previous cases as well. You could maybe add a flag to filter them out at least in this new scenario.
> 
> Image

IMO we must be able to identify the following situations:
- CrytPad Server URL is defined but non reachable.
  - The current message is OK, maybe with tips like:
    - Check your Internet connection or try later
    - Contact an administrator
- CryptPad Server URL is disabled/None
  - Just inform the user about it (without the tips below), maybe "enabled" is more suitable: 
  - `Document editing is not enabled for your organization.`

PR #12421, opened 2026-03-18
Workspace archiving: Add RFC 1028

touilleMan:


docs/rfcs/1028-workspace-archiving-in-parsec-v3.md
```diff
@@ -0,0 +1,321 @@
...
+    ]
+}
+```
+
+> [!NOTE]
+> This change is not strictly needed: instead the self-promotion need status could be computed
+> from the certificates lazily (i.e. each time `Client::list_workspaces()` is called).
+> However this would go against `Client::list_workspaces()` design, as it has been designed
+> around the fact it should get its info from the `LocalUserManifest` that is itself
+> pre-computed by a monitor whenever new certificates are fetched.
```
AureliaDolo, line 143: Sounds like a note in the wrong rfc

docs/rfcs/1028-workspace-archiving-in-parsec-v3.md
```diff
@@ -0,0 +1,321 @@
...
+
+## 0 - Overview
+
+This RFC ports the workspace archiving feature described in
+[RFC 1002](./1002-workspace-archiving-and-deletion.md) to Parsec v3.
+
+The key difference with Parsec v2 is that Parsec v3 clients **eagerly fetch certificates**
+from the server as soon as they are available.
+As a result, the archiving config can then be deduced from the certificates without
+the need for a dedicated server command is it was the case in Parsec v2.
```
mmmarcos:
```suggestion
the need for a dedicated server command as it was the case in Parsec v2.
```

docs/rfcs/1028-workspace-archiving-in-parsec-v3.md
```diff
@@ -0,0 +1,321 @@
+<!-- Parsec Cloud (https://parsec.cloud) Copyright (c) BUSL-1.1 2016-present Scille SAS -->
+
+# Workspace archiving in Parsec v3
+
+## 0 - Overview
```
mmmarcos, line 5: It might worth mentioning somewhere in the RFC that realm sharing configuration (i.e. workspace roles) is not lost when archiving/deleting.

In other words: going from `AVAILABLE` to `ARCHIVED`/`DELETION_PLANNED` and back again to `AVAILABLE,` does not impact the existing user workspace roles.

docs/rfcs/1028-workspace-archiving-in-parsec-v3.md
```diff
@@ -0,0 +1,321 @@
...
+            "variants": [
+                {
+                    // The realm can be read/written, this is the default
+                    // status if no realm archiving certificate has been
+                    // issued.
+                    "name": "Available",
+                    "discriminant_value": "AVAILABLE"
+                },
+                {
+                    // The realm is now in read-only.
```
mmmarcos, line 61: I would mention that this has precedence over the workspace roles. Something like:
```suggestion
                    // The realm is now in read-only.
                    // This has precedence over workspace roles, so users
                    // with write-access (e.g. CONTRIBUTOR) would not be
                    // able to make changes on an archived realm.
```

docs/rfcs/1028-workspace-archiving-in-parsec-v3.md
```diff
@@ -0,0 +1,321 @@
...
+                        "type": "DateTime"
+                    }
+                ]
+            }
+        ]
+    }
+]
+```
+
+### 2.2 - Commands updated with `realm_archived` / `realm_deleted` statuses
```
mmmarcos:
```suggestion
### 2.2 - Commands updated with `realm_archived`/`realm_deleted` status
```

docs/rfcs/1028-workspace-archiving-in-parsec-v3.md
```diff
@@ -0,0 +1,321 @@
...
+| `realm_update_archiving`   |          | ✔️      |
+| `vlob_create`              | ✔️       | ✔️      |
+| `vlob_read_batch`          |          | ✔️      |
+| `vlob_read_versions`       |          | ✔️      |
+| `vlob_update`              | ✔️       | ✔️      |
+| `vlob_poll_changes`        |          | ✔️      |
+| `block_read`               |          | ✔️      |
+| `block_create`             | ✔️       | ✔️      |
+
+The two new reply statuses to add to the affected commands are:
```
mmmarcos:
```suggestion
The two new response status to add to the affected commands are:
```

docs/rfcs/1028-workspace-archiving-in-parsec-v3.md
```diff
@@ -0,0 +1,415 @@
...
+the actual data cleanup on the server may not happen immediately (it is performed by a background
+routine). This creates a window where the deletion date has passed but the data is still
+physically present on the server.
+
+The following rules apply during this window:
+
+1. **GUI considers the realm as deleted as soon as the deadline is reached.**
+   The client does not query the server to determine whether the data has actually been cleaned
+   up (as it would not work when offline). Instead, it compares the current time against the
+   deletion deadline and, once reached, no longer allow any access to the workspace.
```
AureliaDolo, line 314: If there is still data in local cache should it be cleaned there?

docs/rfcs/1028-workspace-archiving-in-parsec-v3.md
```diff
@@ -0,0 +1,415 @@
...
+
+2. **The server continues to give read-only access.**
+   Until the cleanup routine has actually removed the data, the server treats the realm the same
+   way as before the deletion deadline (i.e. `realm_archived` is returned for write commands,
+   read commands succeed).
+
+3. **An owner can still issue a new `RealmArchivingCertificate` to revert the status.**
+   For example, an owner can switch the configuration back to `Available` or `Archived` in order
+   to "un-delete" a realm that was marked for deletion by mistake. Because the GUI considers
```
AureliaDolo, line 327: Isn't that the purpose of the minimum_archiving_period ?

docs/rfcs/1028-workspace-archiving-in-parsec-v3.md
```diff
@@ -0,0 +1,415 @@
...
+   the realm deleted, this recovery action is only available through the Parsec CLI
+   (see section 5).
+
+> [!NOTE]
+> Since the client can no longer access the workspace once the deletion deadline as been reached,
+> this create a typical pattern where only then the users realized the deletion was a mistake.
```
AureliaDolo: Maybe we could add flashing bright red warnings when a user accesses a workspace that's set up to be deleted to avoid that

docs/rfcs/1028-workspace-archiving-in-parsec-v3.md
```diff
@@ -0,0 +1,415 @@
...
+> The rational being that the blockstore is expected to have its own backup system configured
+> in a way Parsec cannot bypass (otherwise an attacker getting access to a Parsec server could
+> erase all the data and their backup from the blockstore...).
+> So instead the server administrator is expected to manually remove the deleted workspaces'
+> data from the blockstore.
+
+This command aims at removing two kinds of workspaces:
+
+- Orphan workspace, i.e. Workspaces that are only shared with revoked users.
+- Workspaces that have been marked for deletion.
```
AureliaDolo, line 383:
```suggestion
- Workspaces that have been marked for deletion and for which the expiration date has passed.
```

docs/rfcs/1028-workspace-archiving-in-parsec-v3.md
```diff
@@ -0,0 +1,415 @@
...
+> Since the client can no longer access the workspace once the deletion deadline as been reached,
+> this create a typical pattern where only then the users realized the deletion was a mistake.
+> Hence the need for a convenient way to undo the deletion using the CLI.
+
+## 5 - CLI commands
+
+### 5.1 - Parsec CLI `workspace archive` command
+
+A new `workspace archive` subcommand is added to the Parsec CLI. It allows an owner to change
+the archiving configuration of a workspace:
```
AureliaDolo, line 345: how should we manage a setup with a deletion date that's already passed ?

docs/rfcs/1028-workspace-archiving-in-parsec-v3.md
touilleMan: This is the correct RFC for this comment ^^

It is a bit hard to understand though, this is because `LocalUserManifest` works in a peculiar way (it is never uploaded to the server, and instead it main job is to work as a cache on the info computed from the certificates)

I've updated the doc to make this more clear

docs/rfcs/1028-workspace-archiving-in-parsec-v3.md
touilleMan: done :+1:

docs/rfcs/1028-workspace-archiving-in-parsec-v3.md
touilleMan: done :+1:

docs/rfcs/1028-workspace-archiving-in-parsec-v3.md
touilleMan: done :+1:

docs/rfcs/1028-workspace-archiving-in-parsec-v3.md
touilleMan: done :+1:

docs/rfcs/1028-workspace-archiving-in-parsec-v3.md
touilleMan: done :+1:

docs/rfcs/1028-workspace-archiving-in-parsec-v3.md
touilleMan: good point ! done.

docs/rfcs/1028-workspace-archiving-in-parsec-v3.md
touilleMan: `minimum_archiving_period`'s goal is to prevent a workspace deletion to occur before we realize we in fact shouldn't delete it.

The issue addressed here is: what if a workspace is ready for deletion (but haven't been actually deleted yet), and realize we shouldn't delete it ?

This is slightly different: `minimum_archiving_period` tries to minimize the issue, but we still have to address what should happen if the issue occurs nevertheless.

Without addressing this issue we could end up in a very inelegant situation where the workspace data *are* available (typically obtained from a PostgreSQL backup), but cannot be used since the workspace is supposed to be deleted :/


I've improved the documentation to stress this point :+1:

docs/rfcs/1028-workspace-archiving-in-parsec-v3.md
touilleMan: If the deletion date doesn't respect `minimum_archiving_period`, then the server rejects the archiving request.

docs/rfcs/1028-workspace-archiving-in-parsec-v3.md
touilleMan: I think the GUI people are planning on showing a big red button on the workspaces list for the one scheduled to be deleted.

docs/rfcs/1028-workspace-archiving-in-parsec-v3.md
touilleMan: Good point ! I've added a work about the need for this cleanup :+1:

PR #12446, opened 2026-03-24
[MS] Reduce the severity of win registry logs

Max-7:


client/electron/src/winRegistry.ts
```diff
@@ -32,17 +32,17 @@ export class WinRegistry {
...
       if (longPathValue.type !== 'REG_DWORD') {
-        console.warn('`LongPathsEnabled`has incorrect type.');
+        console.info('`LongPathsEnabled`has incorrect type.');
         return false;
       }
       console.debug(`'${KEY}\\LongPathsEnabled' has value ${longPathValue.value}`);
       return (longPathValue.value as any as number) === 1;
     } catch (e: any) {
-      console.error(`Error while trying to obtain 'LongPathsEnabled' value: ${e.toString()}`);
+      console.info(`Error while trying to obtain 'LongPathsEnabled' value: ${e.toString()}`);
```
Ironicbay, line 45: Maybe switch this one to `warn`?

Max-7: I'm worried Sentry might create an alert for a warn. But ultimately it's not a really important error. It's only useful if someone tells us "Hey I can't open files with a long path" so we can look at the logs and see that there's indeed a message related to that.

PR #12449, opened 2026-03-24
feat(pki): Impl sign for Pkey

FirelightFlagboy:

### DIFF
--- a/libparsec/crates/platform_pki/src/lib.rs
+++ b/libparsec/crates/platform_pki/src/lib.rs
@@ -107,6 +107,15 @@ mod platform {
     }
 
     pub struct PrivateKey;
+
+    impl PrivateKey {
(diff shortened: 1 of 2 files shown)


libparsec/crates/platform_pki/src/lib.rs
```diff
@@ -100,6 +100,15 @@ mod platform {
     }
 
     pub struct PrivateKey;
+
+    impl PrivateKey {
+        pub async fn sign(
+            &self,
+            message: &[u8],
```
AureliaDolo:
```suggestion
            _message: &[u8],
```

FirelightFlagboy: I've used `expect` instead since the function signature is used for documentation, If find that prettier

PR #12450, opened 2026-03-24
[MS] Set persistent storage when creating a device

Max-7:
After discuting it with @fabienSvtr, we concluded that it was simply easier to request the permission when login into a device.

Here's how it looks on Firefox:
Screenshot 2026-03-24 112331

### DIFF
--- a/client/src/services/storageManager.ts
+++ b/client/src/services/storageManager.ts
@@ -1,12 +1,32 @@
 // Parsec Cloud (https://parsec.cloud) Copyright (c) BUSL-1.1 2016-present Scille SAS
 
-import { isElectron } from '@/parsec';
+import { isElectron, isWeb } from '@/parsec';
 import { Storage } from '@ionic/storage';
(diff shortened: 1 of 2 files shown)


client/src/services/storageManager.ts
```diff
@@ -1,13 +1,29 @@
...
 import { Storage } from '@ionic/storage';
 import { DateTime } from 'luxon';
 import { I18n, Locale, Theme } from 'megashark-lib';
 
 export const StorageManagerKey = 'storageManager';
 export const ThemeManagerKey = 'themeManager';
 
+export async function persistStorage(): Promise<void> {
+  async function _persist(): Promise<void> {
+    if (isWeb() && navigator.storage && !(await navigator.storage.persisted())) {
```
FirelightFlagboy: **note:** Reminder that `persisted` cannot be use to know if the `persistent data` permission was granted, but if the storage is persistent

PR #12451, opened 2026-03-24
Add `--organization-initial-minimum-archiving-period` option to serve…

touilleMan:


server/parsec/cli/run.py
```diff
@@ -304,6 +304,15 @@ def handle_parse_result(
     default=True,
     type=bool,
 )
+@click.option(
+    "--organization-initial-minimum-archiving-period",
+    envvar="PARSEC_ORGANIZATION_INITIAL_MINIMUM_ARCHIVING_PERIOD",
+    show_envvar=True,
+    help="Minimum archiving period in seconds used to configure newly created organizations",
+    default=2592000,
```
FirelightFlagboy: **nitpick:** Would be more clear is the value is decomposed into a multiplication
```suggestion
    default=3600 * 24 * 30,
```

touilleMan: :+1: done.

PR #12464, opened 2026-03-25
Workspace archive server

touilleMan:


server/parsec/components/memory/realm.py
```diff
@@ -520,6 +526,111 @@ async def rename(
...
+            async with org.topics_lock(write=[realm_topic]) as (realm_topic_last_timestamp,):
+                if realm.get_current_role_for(author_user_id) != RealmRole.OWNER:
+                    return RealmUpdateArchivingStoreBadOutcome.AUTHOR_NOT_ALLOWED
+
+                # Check if realm is already deleted (DeletionPlanned with past deletion date)
+                if realm.archivings:
+                    current_config = realm.archivings[-1].cooked.configuration
+                    if (
+                        current_config != RealmArchivingConfiguration.AVAILABLE
+                        and current_config != RealmArchivingConfiguration.ARCHIVED
```
mmmarcos, line 590: why not checking for deletion planned here?
```suggestion
                        current_config == RealmArchivingConfiguration.DELETION_PLANNED
```

server/parsec/components/postgresql/realm_update_archiving.py
```diff
@@ -0,0 +1,225 @@
...
+
+    if certif.configuration == RealmArchivingConfiguration.AVAILABLE:
+        configuration_str = "AVAILABLE"
+        deletion_date = None
+    elif certif.configuration == RealmArchivingConfiguration.ARCHIVED:
+        configuration_str = "ARCHIVED"
+        deletion_date = None
+    else:
+        configuration_str = "DELETION_PLANNED"
+        deletion_date = certif.configuration.deletion_date
```
mmmarcos, line 194: What if the certificate does not update the current configuration? (e.g. certif.configuration == ARCHIVED for a realm that is already archived)

I guess that doesn't break anything, but should we drop the certificate in that case?

server/parsec/components/postgresql/realm_update_archiving.py
```diff
@@ -0,0 +1,225 @@
...
+
```
mmmarcos, line 197:
```suggestion

    # For convenience, the certificate is accepted even if it doesn't change the 
    # realm's current status. This shouldn't cause any problems because it is
    # always the last certificate that is taken into account.
    # Dropping the certificate here should be feasible, but it would require 
    # updating the certificate validation client-side and adding new error status.
```

server/parsec/components/postgresql/realm_update_archiving.py
mmmarcos: See https://github.com/Scille/parsec-cloud/pull/12464#discussion_r2999440844

server/parsec/components/memory/realm.py
touilleMan: `RealmArchivingConfiguration.DELETION_PLANNED` doesn't exist
This is because `RealmArchivingConfiguration` is not an enum but a variant where `RealmArchivingConfiguration.AVAILABLE` and `RealmArchivingConfiguration.ARCHIVED` are both a singleton (so comparison is possible) but for deletion planned we cannot do that given it contains `deletion_date` field

The alternative would be to wrap `RealmArchivingConfiguration` in a hierarchy of classes (so we could do `isinstance(current_config, RealmArchivingConfigurationDeletionPlanned)`, but this is much more cumbersome :/

server/parsec/components/postgresql/realm_update_archiving.py
touilleMan: :+1: I've added this comment to the RFC

Issue #12355, opened 2026-03-10
Create an `Archive` page


fabienSvtr:
## Required
To archive a workspace, we need a specific route which lead to a workspace page.

## Archive workspace page
This page should include:
- All archived workspaces [(mockups)](https://www.figma.com/design/XW3SNWHV8BhPsPJpsOCV2h/Parsec-App?node-id=36632-66360&t=4NsrQZMNIHPlcEe9-11) 
- Workspace cards [(mockups)](https://www.figma.com/design/XW3SNWHV8BhPsPJpsOCV2h/Parsec-App?node-id=36623-190432&t=4NsrQZMNIHPlcEe9-4) 
  - Workspace name
  - Read only mention
  - The time since they are archived
  - Options
- Contextual menu: [(mockups)](https://www.figma.com/design/XW3SNWHV8BhPsPJpsOCV2h/Parsec-App?node-id=36623-190934&t=4NsrQZMNIHPlcEe9-11) 
    - Restore
    - Delete the workspace

## Action of archiving
- Create a modal to explain what the archive feature does [(mockups)](https://www.figma.com/design/XW3SNWHV8BhPsPJpsOCV2h/Parsec-App?node-id=36641-59215&t=4NsrQZMNIHPlcEe9-11)

PR #12466, opened 2026-03-25
Workspace archive libparsec (1/2: display archive status)

touilleMan:

PR #12479, opened 2026-03-26
[MS] Archive workspace page

Ironicbay:
Closes https://github.com/Scille/parsec-cloud/issues/12355

### DIFF
--- a/client/src/components/files/explorer/FileGridDisplay.vue
+++ b/client/src/components/files/explorer/FileGridDisplay.vue
@@ -5,7 +5,7 @@
     ref="fileDropZone"
     :show-drop-message="true"
     @files-added="$emit('filesAdded', $event)"
-    :is-reader="ownRole === WorkspaceRole.Reader"
+    :is-reader="isReader"
     @drop-as-reader="$emit('dropAsReader')"
   >
     <div
@@ -24,7 +24,7 @@
           @open-item.stop="$emit('openItem', folder, $event)"
           @menu-click="onMenuClick"
           @files-added="onFilesAdded"
-          :is-workspace-reader="ownRole === WorkspaceRole.Reader"
+          :is-workspace-reader="isReader"
           @drop-as-reader="$emit('dropAsReader')"
         />
         <file-card
@@ -55,13 +55,13 @@ import FileCard from '@/components/files/explorer/FileCard.vue';
 import FileCardProcessing from '@/components/files/explorer/FileCardProcessing.vue';
 import FileDropZone from '@/components/files/explorer/FileDropZone.vue';
 import { EntryCollection, EntryModel, FileModel, FileOperationCurrentFolder, FolderModel } from '@/components/files/types';
-import { EntryName, WorkspaceRole } from '@/parsec';
+import { EntryName } from '@/parsec';
 import { useTemplateRef } from 'vue';
 
 const props = defineProps<{
   files: EntryCollection<FileModel>;
   folders: EntryCollection<FolderModel>;
-  ownRole: WorkspaceRole;
+  isReader: boolean;
   selectionEnabled?: boolean;
   fileOperations: Array<FileOperationCurrentFolder>;
 }>();
--- a/client/src/components/files/explorer/FileListDisplay.vue
+++ b/client/src/components/files/explorer/FileListDisplay.vue
@@ -5,7 +5,7 @@
     ref="fileDropZone"
     @files-added="$emit('filesAdded', $event)"
     :show-drop-message="true"
-    :is-reader="ownRole === WorkspaceRole.Reader"
+    :is-reader="isReader"
     @drop-as-reader="$emit('dropAsReader')"
     class="file-drop-zone"
   >
@@ -91,7 +91,7 @@
             :entry="folder"
             :own-profile="ownProfile"
             :show-checkbox="someSelected || selectionEnabled === true"
-            :is-workspace-reader="ownRole === WorkspaceRole.Reader"
+            :is-workspace-reader="isReader"
             @open-item.stop="$emit('openItem', folder, $event)"
             @menu-click="onMenuClick"
             @files-added="onFilesAdded"
@@ -128,7 +128,7 @@ import FileDropZone from '@/components/files/explorer/FileDropZone.vue';
 import FileListItem from '@/components/files/explorer/FileListItem.vue';
 import FileListItemProcessing from '@/components/files/explorer/FileListItemProcessing.vue';
 import { EntryCollection, EntryModel, FileModel, FileOperationCurrentFolder, FolderModel } from '@/components/files/types';
-import { EntryName, UserProfile, WorkspaceRole } from '@/parsec';
+import { EntryName, UserProfile } from '@/parsec';
 import { IonIcon, IonLabel, IonList, IonListHeader, IonText } from '@ionic/vue';
 import { arrowDown, arrowUp } from 'ionicons/icons';
 import { MsCheckbox, MsSorterChangeEvent, useWindowSize } from 'megashark-lib';
@@ -139,7 +139,7 @@ const props = defineProps<{
   ownProfile: UserProfile;
   files: EntryCollection<FileModel>;
   folders: EntryCollection<FolderModel>;
-  ownRole: WorkspaceRole;
+  isReader: boolean;
   fileOperations: Array<FileOperationCurrentFolder>;
   selectionEnabled?: boolean;
   currentSortProperty: SortProperty;
--- a/client/src/components/small-display/SmallDisplayFileContextMenu.vue
+++ b/client/src/components/small-display/SmallDisplayFileContextMenu.vue
@@ -21,7 +21,7 @@
           </ion-item>
           <ion-item
             button
-            v-if="!multipleFiles && role !== WorkspaceRole.Reader && isEditable && isFile"
+            v-if="!multipleFiles && !isReadOnly && isEditable && isFile"
             @click="onClick(FileAction.Edit)"
             class="ion-no-padding list-group-item"
           >
@@ -35,7 +35,7 @@
           </ion-item>
           <ion-item
             button
-            v-if="!multipleFiles && role !== WorkspaceRole.Reader"
+            v-if="!multipleFiles && !isReadOnly"
             @click="onClick(FileAction.Rename)"
             class="ion-no-padding list-group-item"
           >
@@ -49,7 +49,7 @@
           </ion-item>
           <ion-item
             button
-            v-if="role !== WorkspaceRole.Reader"
+            v-if="!isReadOnly"
             @click="onClick(FileAction.MoveTo)"
             class="ion-no-padding list-group-item"
           >
@@ -64,7 +64,7 @@
 
           <ion-item
             button
-            v-if="role !== WorkspaceRole.Reader"
+            v-if="!isReadOnly"
             @click="onClick(FileAction.MakeACopy)"
             class="ion-no-padding list-group-item"
           >
@@ -126,7 +126,7 @@
             button
             @click="onClick(FileAction.ShowHistory)"
             class="ion-no-padding list-group-item"
-            v-show="!multipleFiles && role !== WorkspaceRole.Reader"
+            v-show="!multipleFiles && !isReadOnly"
           >
             <ion-icon
               class="list-group-item__icon"
@@ -184,7 +184,7 @@
         </ion-item-group>
         <ion-item-group
           class="list-group"
-          v-if="role !== WorkspaceRole.Reader"
+          v-if="!isReadOnly"
         >
           <ion-item
             button
@@ -207,14 +207,14 @@
 
 <script setup lang="ts">
 import FolderParentIcon from '@/assets/images/folder-parent.svg?raw';
-import { isDesktop, WorkspaceRole } from '@/parsec';
+import { isDesktop } from '@/parsec';
 import { FileAction } from '@/views/files/types';
 import { IonContent, IonIcon, IonItem, IonItemGroup, IonList, IonPage, IonText, modalController } from '@ionic/vue';
 import { arrowRedo, copy, create, download, eye, informationCircle, link, open, time, trashBin } from 'ionicons/icons';
 import { EyeOpenIcon, MsImage, RenameIcon } from 'megashark-lib';
 
 defineProps<{
-  role: WorkspaceRole;
+  isReadOnly: boolean;
   multipleFiles?: boolean;
   isFile: boolean;
   isEditable?: boolean;
--- a/client/src/components/workspaces/WorkspaceCard.vue
+++ b/client/src/components/workspaces/WorkspaceCard.vue
@@ -6,6 +6,7 @@
     :class="{
       'workspace-hovered': isHovered || menuOpened,
       'workspace-card-item--hidden': isHidden,
+      'workspace-card-item--archived': workspace.isArchived,
     }"
     @click="$emit('click', workspace, $event)"
     @mouseenter="isHovered = true"
@@ -14,6 +15,7 @@
   >
     <div class="workspace-card-content">
       <div
+        v-show="!workspace.isArchived"
         class="workspace-favorite-icon"
         :class="{
           'workspace-favorite-icon__on': isFavorite,
@@ -35,6 +37,7 @@
         <ion-text class="workspace-card-content__update subtitles-sm">
           <span v-if="false">{{ $msTranslate(formatTimeSince(workspace.lastUpdated, '--', 'short')) }}</span>
           <ion-icon
+            v-if="!workspace.isArchived"
             class="cloud-overlay"
             :class="workspace.availableOffline ? 'cloud-overlay-ok' : 'cloud-overlay-ko'"
             :icon="workspace.availableOffline ? cloudDone : cloudOffline"
@@ -49,7 +52,7 @@
 
         <div
           class="workspace-hidden subtitles-sm"
-          v-if="isHidden"
+          v-if="isHidden && !workspace.isArchived"
         >
           <ion-icon
             class="cloud-overlay"
@@ -57,16 +60,28 @@
           />
           <ion-text>{{ $msTranslate('WorkspacesPage.Workspace.hidden') }}</ion-text>
         </div>
+        <ion-icon
+          v-if="workspace.isArchived"
+          class="archived-icon"
+          :icon="archive"
+        />
       </div>
     </div>
     <div class="workspace-card-bottom">
       <workspace-role-tag
+        v-if="!workspace.isArchived"
         class="workspace-card-bottom__role"
         :role="workspace.currentSelfRole"
       />
+      <ion-text
+        v-else
+        class="button-small archived-label"
+      >
+        {{ $msTranslate('WorkspacesPage.archiveWorkspace.readOnly') }}
+      </ion-text>
       <div class="workspace-card-bottom__icons">
         <div
-          v-show="clientProfile !== UserProfile.Outsider"
+          v-show="clientProfile !== UserProfile.Outsider && !workspace.isArchived"
           class="icon-share-container"
           @click.stop="$emit('shareClick', workspace, $event)"
         >
@@ -94,7 +109,7 @@ import { formatFileSize } from '@/common/file';
 import { WorkspaceRoleTag } from '@/components/workspaces';
 import { UserProfile, WorkspaceInfo } from '@/parsec';
 import { IonIcon, IonText } from '@ionic/vue';
-import { cloudDone, cloudOffline, ellipsisHorizontal, eyeOff, shareSocial, star } from 'ionicons/icons';
+import { archive, cloudDone, cloudOffline, ellipsisHorizontal, eyeOff, shareSocial, star } from 'ionicons/icons';
 import { formatTimeSince } from 'megashark-lib';
 import { ref } from 'vue';
 
@@ -136,6 +151,7 @@ async function onOptionsClick(event: Event): Promise<void> {
   overflow: hidden;
   padding: 0.25rem;
   transition: all 0.15s ease-in-out;
+  min-width: 14rem;
 
   @include ms.responsive-breakpoint('sm') {
     max-width: 100%;
@@ -238,21 +254,6 @@ async function onOptionsClick(event: Event): Promise<void> {
     }
   }
 
-  .workspace-hidden {
-    background: var(--parsec-color-light-secondary-white);
-    color: var(--parsec-color-light-secondary-contrast);
-    text-align: left;
-    display: flex;
-    align-items: center;
-    gap: 0.375rem;
-    padding: 3px 0.5rem;
-    border-radius: var(--parsec-radius-8);
-
-    ion-icon {
-      font-size: 0.875rem;
-    }
-  }
-
   &-info {
     display: flex;
     align-items: center;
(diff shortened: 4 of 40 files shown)


client/src/components/files/explorer/FileGridDisplay.vue
```diff
@@ -55,13 +55,13 @@ import FileCard from '@/components/files/explorer/FileCard.vue';
...
 import { EntryCollection, EntryModel, FileModel, FileOperationCurrentFolder, FolderModel } from '@/components/files/types';
-import { EntryName, WorkspaceRole } from '@/parsec';
+import { EntryName } from '@/parsec';
 import { useTemplateRef } from 'vue';
 
 const props = defineProps<{
   files: EntryCollection<FileModel>;
   folders: EntryCollection<FolderModel>;
-  ownRole: WorkspaceRole;
+  isReader: boolean;
```
Max-7, line 64: I don't see the point of this change.

client/src/components/files/explorer/FileListDisplay.vue
```diff
@@ -139,7 +139,7 @@ const props = defineProps<{
   ownProfile: UserProfile;
   files: EntryCollection<FileModel>;
   folders: EntryCollection<FolderModel>;
-  ownRole: WorkspaceRole;
+  isReader: boolean;
```
Max-7, line 142: Same here, I don't get why you'd want to change that.

client/src/components/small-display/SmallDisplayFileContextMenu.vue
```diff
@@ -207,14 +207,14 @@
...
-import { isDesktop, WorkspaceRole } from '@/parsec';
+import { isDesktop } from '@/parsec';
 import { FileAction } from '@/views/files/types';
 import { IonContent, IonIcon, IonItem, IonItemGroup, IonList, IonPage, IonText, modalController } from '@ionic/vue';
 import { arrowRedo, copy, create, download, eye, informationCircle, link, open, time, trashBin } from 'ionicons/icons';
 import { EyeOpenIcon, MsImage, RenameIcon } from 'megashark-lib';
 
 defineProps<{
-  role: WorkspaceRole;
+  readOnly: boolean;
```
Max-7: Same here.

Max-7 (changes requested):
Looking at the mockups (https://www.figma.com/design/XW3SNWHV8BhPsPJpsOCV2h/Parsec-App?node-id=36632-66360) it is supposed to be a new page (with its own route), not a state of the existing page.

Archived workspaces are not handled front side like we do with visibility. Libparsec does the heavy lifting and you have to use it. It has been (at least partially) implemented by @touilleMan in https://github.com/Scille/parsec-cloud/pull/12466, if something is missing you can discuss the API with him and mock it in the meantime.

client/src/parsec/workspace.ts
```diff
@@ -32,7 +32,14 @@ import {
   WorkspaceRole,
 } from '@/parsec/types';
 import { generateNoHandleError } from '@/parsec/utils';
-import { MountpointUnmountError, ParsecWorkspacePathAddrAndRedirectionURL, WorkspaceStopError, libparsec } from '@/plugins/libparsec';
+import {
+  ClientArchiveWorkspaceError,
```
Max-7: Re-export the errors and tags in parsec/types.ts. We want to avoid imports directly from libparsec outside of parsec/. So a component should not import from `@/plugins/libparsec`, always from `@/parsec`

client/src/parsec/workspace.ts
```diff
@@ -266,6 +273,29 @@ export async function unmountWorkspace(workspace: WorkspaceInfo | StartedWorkspa
   return { ok: true, value: null };
 }
 
+export async function archiveWorkspace(workspace: WorkspaceInfo): Promise<Result<null, ClientArchiveWorkspaceError>> {
```
Max-7: Need only a workspaceId here, not the whole object.

client/src/parsec/workspace.ts
```diff
@@ -266,6 +273,29 @@ export async function unmountWorkspace(workspace: WorkspaceInfo | StartedWorkspa
...
+  }
+  // const clientResult = await getClientInfo();
+  // if (clientResult.ok) {
+  //   const deletionDate = DateTime.now().plus({ milliseconds: clientResult.value.serverOrganizationConfig.minimumArchivingPeriod as unknown as number });
+  //   return await libparsec.clientArchiveWorkspace(handle, workspace.id, { tag: RealmArchivingConfigurationTag.DeletionPlanned, deletionDate: deletionDate });
+  // }
+  return await libparsec.clientArchiveWorkspace(handle, workspace.id, { tag: RealmArchivingConfigurationTag.Archived });
+}
+
+export async function restoreWorkspace(workspace: WorkspaceInfo): Promise<Result<null, ClientArchiveWorkspaceError>> {
```
Max-7: Same here, just pass the id.

client/src/views/files/FoldersPage.vue
```diff
@@ -522,10 +452,94 @@ const workspaceName = computed(() => {
   return workspaceInfo.value?.currentName ?? '';
 });
 
+const isReadOnly = asyncComputed(async (): Promise<boolean> => {
+  if (ownRole.value === parsec.WorkspaceRole.Reader) {
+    return true;
+  }
+  const workspacesResult = await listWorkspaces();
```
Max-7: Why would you need to relist the workspaces here? You already have `workspaceInfo`.

client/src/views/files/FoldersPage.vue
```diff
@@ -1702,6 +1716,10 @@ async function performFolderAction(action: FolderGlobalAction): Promise<void> {
 }
 
 async function openGlobalContextMenu(event: Event): Promise<void> {
+  if (isReadOnly.value) {
```
Max-7: You have updated `_openGlobalContextMenu` to handle that.

client/src/parsec/workspace.ts
Max-7, line 1: I think it's probably easier to have some separation  when we list the workspaces, by making the current `listWorkspaces` function filter out the archived workspace, and adding a new `listArchivedWorkspaces` function. This way the code that currently interact with the workspaces doesn't have to be updated.

We also overwrite `WorkspaceInfo` in `parsec/types.ts`, you could add a `isArchived` function on that interface that you set in parsec/workspaces.ts when listing a workspace, it makes it easier to deal with in the components.

client/src/components/workspaces/WorkspaceCard.vue
```diff
@@ -106,6 +122,7 @@ const props = defineProps<{
   clientProfile: UserProfile;
   isFavorite: boolean;
   isHidden: boolean;
+  isArchived: boolean;
```
Max-7: You have the `WorkspaceInfo` already.

client/src/components/workspaces/utils.ts
```diff
@@ -147,6 +150,7 @@ export async function openWorkspaceContextMenu(
         clientRole: workspace.currentSelfRole,
         isFavorite: workspaceAttributes.isFavorite(workspace.id),
         isHidden: workspaceAttributes.isHidden(workspace.id),
+        isArchived: workspace.archivingConfiguration.tag === RealmArchivingConfigurationTag.Archived,
```
Max-7: This should probably be updated to pass the full `workspaceInfo`, since we need the name, the role, and not the archive status.

client/src/components/workspaces/utils.ts
```diff
@@ -213,12 +218,62 @@ export async function openWorkspaceContextMenu(
...
+    new Information({
+      message: {
+        key: result.ok ? 'WorkspacesPage.archiveWorkspace.archive.success' : 'WorkspacesPage.archiveWorkspace.archive.fail',
+        data: { workspace: workspace.currentName },
+      },
+      level: result.ok ? InformationLevel.Success : InformationLevel.Error,
+    }),
+    PresentationMode.Toast,
+  );
+  if (!result.ok) {
```
Max-7: libparsec function that fail are automatically logged, no need to do it yourself.

client/src/components/workspaces/utils.ts
```diff
@@ -213,12 +218,62 @@ export async function openWorkspaceContextMenu(
...
+    new Information({
+      message: {
+        key: result.ok ? 'WorkspacesPage.archiveWorkspace.restore.success' : 'WorkspacesPage.archiveWorkspace.restore.fail',
+        data: { workspace: workspace.currentName },
+      },
+      level: result.ok ? InformationLevel.Success : InformationLevel.Error,
+    }),
+    PresentationMode.Toast,
+  );
+  if (!result.ok) {
```
Max-7: Same here, it's already logged.

client/src/parsec/workspace.ts
```diff
@@ -266,6 +273,29 @@ export async function unmountWorkspace(workspace: WorkspaceInfo | StartedWorkspa
   return { ok: true, value: null };
 }
 
+export async function archiveWorkspace(workspace: WorkspaceInfo): Promise<Result<null, ClientArchiveWorkspaceError>> {
+  console.log(`Archive ${workspace.currentName}`);
+  const handle = getConnectionHandle();
+  if (!handle) {
+    return generateNoHandleError<ClientArchiveWorkspaceError>();
+  }
+  // const clientResult = await getClientInfo();
```
Max-7: Remove the comments.

client/src/parsec/workspace.ts
```diff
@@ -266,6 +273,29 @@ export async function unmountWorkspace(workspace: WorkspaceInfo | StartedWorkspa
...
+  // if (clientResult.ok) {
+  //   const deletionDate = DateTime.now().plus({ milliseconds: clientResult.value.serverOrganizationConfig.minimumArchivingPeriod as unknown as number });
+  //   return await libparsec.clientArchiveWorkspace(handle, workspace.id, { tag: RealmArchivingConfigurationTag.DeletionPlanned, deletionDate: deletionDate });
+  // }
+  return await libparsec.clientArchiveWorkspace(handle, workspace.id, { tag: RealmArchivingConfigurationTag.Archived });
+}
+
+export async function restoreWorkspace(workspace: WorkspaceInfo): Promise<Result<null, ClientArchiveWorkspaceError>> {
+  console.log(`Restore ${workspace.currentName}`);
```
Max-7: Remove the log.

client/src/views/header/HeaderPage.vue
```diff
@@ -282,6 +283,18 @@ const showSecurityChecklistSmallDisplay = computed(() => {
   return isSmallDisplay.value && securityWarningsCount.value > 0 && securityWarnings.value;
 });
 
+const isArchived = asyncComputed(async (): Promise<boolean> => {
```
Max-7, line 325: I was wondering why you didn't just call `getWorkspaceInfo` but I realized that it returns a `StartedWorkspaceInfo` instead of `WorkspaceInfo`. I think this is a good opportunity to re-write that: have `getWorkspaceInfo` take a handle, call `libparsec.workspaceInfo()` to get the id, `libparsec.clientListWorkspaces` to list the workspaces and match the id, and return the complete WorkspaceInfo. And remove all exports of `StartedWorkspaceInfo`. It would make dealing with workspaces way easier.

client/src/views/workspaces/WorkspaceArchivePage.vue
```diff
@@ -0,0 +1,217 @@
...
+const workspaceList: Ref<Array<WorkspaceInfo>> = ref([]);
+const workspaceAttributes = useWorkspaceAttributes();
+const { isLargeDisplay } = useWindowSize();
+const informationManager: Ref<InformationManager> = inject(InformationManagerKey)!;
+const eventDistributor: Ref<EventDistributor> = inject(EventDistributorKey)!;
+const storageManager: StorageManager = inject(StorageManagerKey)!;
+
+// clientProfile is required as a WorkspaceCard prop but only to show or hide
+// the share button, we can leave it as Outsider
+const clientProfile: Ref<UserProfile> = ref(UserProfile.Outsider);
```
Max-7: You don't need to declare it at all, just pass the value to `workspace-card`

client/src/views/menu/MainMenu.vue
```diff
@@ -481,6 +500,15 @@ const securityWarningsCount = computed(() => {
 });
 
 const { isSmallDisplay, windowWidth } = useWindowSize();
+const isReadOnly = computed(() => {
+  if (currentWorkspace.value) {
```
Max-7, line 505: I've noticed that the `currentWorkspace` in this component is problematic: it only looks at recent workspaces.

client/src/parsec/workspace.ts
```diff
@@ -94,14 +106,67 @@ export async function listWorkspaces(
   return generateNoHandleError<ClientListWorkspacesError>();
 }
 
-export async function getWorkspaceInfo(workspaceHandle: WorkspaceHandle): Promise<Result<StartedWorkspaceInfo, WorkspaceInfoError>> {
+export async function listAvailableWorkspaces(
+  handle: ConnectionHandle | null = null,
+): Promise<Result<Array<WorkspaceInfo>, ClientListWorkspacesError>> {
+  return listWorkspaces(handle, false);
```
Max-7: I think it's better to not modify `listWorkspaces` and to filter out the workspaces here. Right now we only have archived workspaces, but this can evolve over time and adding more parameters will become difficult.

```
const result = await listWorkspaces();
if (result.ok) {
  result.value = result.value.filter((workspace) => !workspace.isArchived);
}
```

client/src/parsec/workspace.ts
```diff
@@ -94,14 +106,67 @@ export async function listWorkspaces(
...
+
+export async function getStartedWorkspaceInfo(workspaceHandle: WorkspaceHandle): Promise<Result<StartedWorkspaceInfo, WorkspaceInfoError>> {
   const result = await libparsec.workspaceInfo(workspaceHandle);
   if (result.ok) {
     (result.value as StartedWorkspaceInfo).handle = workspaceHandle;
   }
   return result as Result<StartedWorkspaceInfo, WorkspaceInfoError>;
 }
 
+export async function getWorkspaceInfo(
```
Max-7: You don't need to start anything, you already have the handle so we know it's started, and `listWorkspaces` does the start if needed anyway. So this function is just listing the workspace, find the right one and return it.
```
const result = await listWorkspaces(handle);
if (result.ok) {
  const workspace = result.value.find((wk) => wk.handle === handle);
  if (workspace) {
    return { ok: false, { tag: WorkspaceInfoError.Internal, error: 'not-found' } };
  }
  return { ok: true, value: workspace };
}
```

client/src/parsec/workspace.ts
```diff
@@ -94,14 +106,67 @@ export async function listWorkspaces(
...
+  return listWorkspaces(handle, false);
+}
+
+export async function listArchivedWorkspaces(
+  handle: ConnectionHandle | null = null,
+): Promise<Result<Array<WorkspaceInfo>, ClientListWorkspacesError>> {
+  return listWorkspaces(handle, true);
+}
+
+export async function getStartedWorkspaceInfo(workspaceHandle: WorkspaceHandle): Promise<Result<StartedWorkspaceInfo, WorkspaceInfoError>> {
```
Max-7: The goal is to remove the notion of `StartedWorkspaceInfo` completely: WorkspaceInfo should be sufficient. It contains the mountpoints, it contains the handle. You may need to add a call to `libparsec.workspaceInfo()` inside `listWorkspaces` to get the mountpoints.
Unless I'm missing something, you should be able to just replace `getWorkspaceInfo`, make it return a `WorkspaceInfo` instead of a `StartedWorkspaceInfo`, and it should almost invisible to the app (maybe some typing).

client/src/views/workspaces/WorkspaceContextMenu.vue
```diff
@@ -194,28 +216,63 @@
...
-  workspaceName: WorkspaceName;
+const props = defineProps<{
+  workspace: WorkspaceInfo;
   clientProfile: UserProfile;
-  clientRole: WorkspaceRole;
   isFavorite: boolean;
   isHidden: boolean;
 }>();
+
+const isArchived = computed(() => {
```
Max-7: Not needed anymore

client/src/views/workspaces/SmallDisplayWorkspaceContextMenu.vue
```diff
@@ -165,29 +187,64 @@
...
-  workspaceName: WorkspaceName;
+const props = defineProps<{
+  workspace: WorkspaceInfo;
   clientProfile: UserProfile;
-  clientRole: WorkspaceRole;
   isFavorite: boolean;
   isHidden: boolean;
 }>();
+
+const isArchived = computed(() => {
```
Max-7: Not needed, you can just check `workspace.isArchived`

client/src/plugins/libparsec/definitions.ts
Max-7, line 1: You should not touch anything in this file. As mentionned at the beginning, it is auto-generated. If needed, the interface are re-exported in `parsec/types.ts` to shadow the ones defined in `definitions.ts`.

bindings/electron/src/index.d.ts
Max-7, line 1: Should not touch anything in this file.

client/src/parsec/types.ts
```diff
@@ -414,11 +417,7 @@ interface WorkspaceInfo extends ParsecWorkspaceInfo {
   created?: DateTime;
   availableOffline: boolean;
   handle: WorkspaceHandle;
-  mountpoints: [MountpointHandle, SystemPath][];
```
Max-7, line 424: Why did you remove the mounpoints?

client/src/views/menu/MainMenu.vue
```diff
@@ -496,10 +523,12 @@ const watchSidebarWidthCancel = watch(sidebarWidth, async (value: number) => {
...
 });
 
-const currentWorkspace = computed(() => {
-  for (const wk of recentDocumentManager.getWorkspaces()) {
-    if (currentRouteIsWorkspaceRoute(wk.handle)) {
-      return wk;
+const currentWorkspace = asyncComputed(async () => {
+  const workspacesResult = await listWorkspaces();
+  if (workspacesResult.ok) {
+    const wInfo = workspacesResult.value.find((wi) => currentRouteIsWorkspaceRoute(wi.handle));
```
Max-7: You can use `getWorkspaceInfo` with `getWorkspaceHandle` instead.

client/src/plugins/libparsec/definitions.ts
```diff
@@ -550,12 +541,14 @@ export interface WorkspaceHistoryFileStat {
 }
 
 export interface WorkspaceInfo {
+    client: Handle
```
Max-7: We don't need a client handle in WorkspaceInfo.

client/src/components/workspaces/WorkspaceCard.vue
```diff
@@ -33,8 +35,9 @@
 
       <div class="workspace-card-content-info">
         <ion-text class="workspace-card-content__update subtitles-sm">
-          <span v-if="false">{{ $msTranslate(formatTimeSince(workspace.lastUpdated, '--', 'short')) }}</span>
+          <span v-if="workspace.isArchived">{{ $msTranslate(formatTimeSince(workspace.lastUpdated, '--', 'short')) }}</span>
```
Max-7: Don't use `lastUpdated`, it is a mocked value and always default to the current date time.

client/src/parsec/types.ts
```diff
@@ -415,10 +418,7 @@ interface WorkspaceInfo extends ParsecWorkspaceInfo {
   availableOffline: boolean;
   handle: WorkspaceHandle;
   mountpoints: [MountpointHandle, SystemPath][];
-}
-
-interface StartedWorkspaceInfo extends ParsecStartedWorkspaceInfo {
-  handle: WorkspaceHandle;
+  isArchived?: boolean;
```
Max-7: Don't mark it as optional, the value should always be set.

client/src/parsec/workspace.ts
```diff
@@ -94,12 +103,38 @@ export async function listWorkspaces(
...
+): Promise<Result<WorkspaceInfo, WorkspaceInfoError | ClientListWorkspacesError>> {
+  const result = await listWorkspaces();
+  if (result.ok) {
+    const workspace = result.value.find((wk) => wk.handle === workspaceHandle);
+    if (workspace) {
+      return { ok: true, value: workspace };
+    }
+    return { ok: false, error: { tag: WorkspaceInfoErrorTag.Internal, error: 'not-found' } };
+  }
+  return result;
```
Max-7: I think it would be better to convert the error here: we're calling `getWorkspaceInfo`, it should return only `WorkspaceInfoError`. How the function is implemented should not be the caller's problem.

`return { ok: false, error: { tag: WorkspaceInfoErrorTag.Internal, error: result.error.error }`;

Which also mean removing the `ClientListWorkspaceError` from the return type, which also means it can be removed from `initializeWorkspace` return type.

client/src/views/files/FoldersPage.vue
```diff
@@ -522,10 +452,87 @@ const workspaceName = computed(() => {
   return workspaceInfo.value?.currentName ?? '';
 });
 
+const isReadOnly = asyncComputed(async (): Promise<boolean> => {
+  if (ownRole.value === parsec.WorkspaceRole.Reader || (workspaceInfo.value && workspaceInfo.value.isArchived === true)) {
```
Max-7: You can use the `?` operator here, which makes it a bit simpler
`retun ownRole.value === parsec.WorkspaceRole.Reader || workspaceInfo.value?.isArchived === true`
but I'm nitpicking.

client/src/views/header/HeaderPage.vue
Max-7, line 1: This file is extremely messy. I've opened https://github.com/Scille/parsec-cloud/issues/12564 to address it.

client/src/views/header/HeaderPage.vue
```diff
@@ -282,6 +298,18 @@ const showSecurityChecklistSmallDisplay = computed(() => {
...
+  const workspaceHandle = getWorkspaceHandle();
+  const workspacesResult = await listArchivedWorkspaces();
```
Max-7: You can call `getWorkspaceInfo` now \o/

client/src/views/header/WorkspaceSwitchModal.vue
```diff
@@ -0,0 +1,140 @@
...
+<template>
+  <div class="workspace-switch-modal-container">
+    <ion-radio-group
+      v-model="selectedRoute"
+      class="workspace-switch-content"
+    >
+      <ion-radio
+        class="switch-item"
+        :value="Routes.Workspaces"
+        @click="$event.preventDefault()"
```
Max-7: `@click.prevent`

client/src/views/header/WorkspaceSwitchModal.vue
```diff
@@ -0,0 +1,140 @@
...
+          v-if="selectedRoute === Routes.Workspaces"
+          :icon="checkmarkCircle"
+          class="switch-item__check-icon"
+        />
+      </ion-radio>
+
+      <ion-radio
+        class="switch-item"
+        :value="Routes.Archived"
+        @click="$event.preventDefault()"
```
Max-7: `@click.prevent`

client/src/views/workspaces/WorkspaceArchivePage.vue
```diff
@@ -0,0 +1,281 @@
...
+  }
+  querying.value = false;
+}
+
+async function onWorkspaceClick(workspace: WorkspaceInfo): Promise<void> {
+  await navigateToWorkspace(workspace.handle);
+}
+
+async function onOpenWorkspaceContextMenu(workspace: WorkspaceInfo, event: Event, onFinished?: () => void): Promise<void> {
+  await openWorkspaceContextMenu(
```
Max-7: Since what you can do on an archive workspace is ultra specific, it would probably be better to have another function. This way you can get rid of the storageManager and the workspaceAttributes that are basically imported here just to be able to call this function.

It could be done in another PR if you're getting tired of this one.

client/src/parsec/types.ts
```diff
@@ -422,10 +420,8 @@ interface WorkspaceInfo extends ParsecWorkspaceInfo {
   availableOffline: boolean;
   handle: WorkspaceHandle;
   mountpoints: [MountpointHandle, SystemPath][];
-}
-
-interface StartedWorkspaceInfo extends ParsecStartedWorkspaceInfo {
-  handle: WorkspaceHandle;
+  isArchived: boolean;
+  archivedOn?: DateTime;
```
Max-7: How are you getting that?

PR #12498, opened 2026-03-30
Backport: Remove `optimized` feature from libsodium-rs since it is not portable

FirelightFlagboy:


newsfragments/12445.bugfix.rst
```diff
@@ -0,0 +1 @@
+Fix Linux Snap version crashing during login with CPU not supporting AVX-512 instructions
```
mmmarcos:
```suggestion
Fix Linux snap crash during login on a CPU not supporting AVX-512 instructions
```

PR #12512, opened 2026-03-31
Workspace archive read only mountpoint

touilleMan:


touilleMan:
merged through #12514

PR #12543, opened 2026-04-07
feat(server): Impl `scws_service_certificate` protocol

FirelightFlagboy:
- Ascii sort components of `Backend` class


server/parsec/components/memory/factory.py
AureliaDolo, line 1: What tool/settings do you use for the formatting ? So we can all use it ^^

FirelightFlagboy: My left & right hands :sweat_smile:

AureliaDolo: ow

AureliaDolo: May I suggest to add a comment at the beginning of each sorted section to remind the next person to keep it alphabetized ? ^^

touilleMan:
replaced by https://github.com/Scille/parsec-cloud/pull/12638

PR #12560, opened 2026-04-07
refactor(async-enroll): Use new PKI API for `_verify_accept_payload`

FirelightFlagboy:
Closes #12552


libparsec/src/async_enrollment.rs
```diff
@@ -586,8 +586,7 @@ mod strategy {
 
     #[derive(Debug)]
     struct AcceptFinalizeAsyncEnrollmentPKIIdentityStrategy {
-        #[expect(dead_code, reason = "Will be used by either #12550 or #12551")]
-        system: libparsec_platform_pki::PkiSystem,
+        system: Arc<libparsec_platform_pki::PkiSystem>,
```
AureliaDolo, line 605: I don't get why you'd need the arc here ?

FirelightFlagboy: We need the Arc because the trait return a boxed future that need to be `'static` but we need to pass a copy of the PKI system

PR #12570, opened 2026-04-08
Rework `libparsec_platform_pki` 1: public api

touilleMan:


FirelightFlagboy (changes requested):
You are not using the same API across platform, use a trait to help you with that

FirelightFlagboy (commented):
Missing windows impl

touilleMan:
> You are not using the same API across platform

That is the point of the first commit: platform-agnostic types are defined to easily understand the exposed interface without having to dig in the platform-specific code.

> use a trait to help you with that

A trait is not needed here (there is no need for dynamic dispatch, nor for strong decoupling between), just like we don't use trait in e.g. `libparsec_mountpoint` crate even if we provide multiple implementations for the `Mountpoint` type.

> Missing windows impl

see https://github.com/Scille/parsec-cloud/pull/12572 

I put this on its own PR since those changes are more important (see the commit messages for explanation, note they are multi-lines so Github doesn't show them in full) as it concern the only existing implementation and not just a stub implementation like for scws/unix.

libparsec/crates/platform_pki_2_electric_boogaloo/src/shared/info.rs
```diff
@@ -0,0 +1,156 @@
...
+    let submitter_der_x509_certificate_der =
+        X509CertificateDer::from_slice(submitter_der_x509_certificate);
+    let submitter_der_x509_certificate_end =
+        X509EndCertificate::try_from(&submitter_der_x509_certificate_der).map_err(|err| {
+            GetRootX509CertificateInfoFromTrustchainError::InvalidCertificateDer(err.into())
+        })?;
+
+    let intermediate_der_x509_certificates_der = intermediate_der_x509_certificates
+        .map(X509CertificateDer::from_slice)
+        .collect::<Vec<_>>();
```
AureliaDolo, line 39: do you need to collect here ?

libparsec/crates/platform_pki_2_electric_boogaloo/src/shared/info.rs
```diff
@@ -0,0 +1,156 @@
...
+    let mut intermediate_der_x509_certificates_end = intermediate_der_x509_certificates_der
+        .iter()
+        .map(X509EndCertificate::try_from)
+        .collect::<Result<Vec<_>, _>>()
+        .map_err(|err| {
+            GetRootX509CertificateInfoFromTrustchainError::InvalidCertificateDer(err.into())
+        })?;
+
+    let mut current = submitter_der_x509_certificate_end;
+    while let Some(i) = intermediate_der_x509_certificates_end
```
AureliaDolo, line 49: In the legacy-ish pki we kept a record of the intermediate certs that are actually part of the trustchain to 
- detect circular trustchain and cut it off
- add a limit to the number of recursion we allow (cut-off was at 100 certs)
- warn if some cert were provided but not used in the trustchain

The two last points are more nice to have but the first one can happen.

libparsec/crates/platform_pki_2_electric_boogaloo/src/lib.rs
```diff
@@ -0,0 +1,60 @@
...
+    CannotOpenStore(std::io::Error),
+    #[error("Cannot get certificate info: {0}")]
+    CannotGetCertificateInfo(std::io::Error),
+}
+
+// TODO: https://github.com/Scille/parsec-cloud/issues/11215
+// This is specific to windows, it cannot be replicated on other platform.
+// Instead, we likely need to go the manual way and show a custom dialog on the client side with a
+// list of certificate that we retrieve from the platform certstore.
+pub fn show_certificate_selection_dialog_windows_only(
```
AureliaDolo, line 50: I'm not sure we need to keep that

libparsec/crates/platform_pki_2_electric_boogaloo/src/pki_certificate.rs
```diff
@@ -0,0 +1,148 @@
...
+//     CannotAcquirePubkey(#[from] crate::x509::GetCertificatePublicKeyError),
+//     #[error("Cannot encrypt message: {0}")]
+//     CannotEncrypt(#[from] rsa::errors::Error),
+// }
+
+// #[derive(Debug, thiserror::Error)]
+// pub enum PkiCertificateVerifyError {
+//     #[error("Invalid signature: {0}")]
+//     InvalidSignature(webpki::Error),
+// }
```
AureliaDolo, line 54: dead code

libparsec/crates/platform_pki_2_electric_boogaloo/src/pki_system.rs
```diff
@@ -0,0 +1,147 @@
...
+/// be ripped from the web page code).
+pub struct PkiScwsConfig {
+    pub parsec_addr: libparsec_types::ParsecAddr,
+    pub proxy: libparsec_platform_http_proxy::ProxyConfig,
+}
+
+pub enum AvailablePkiCertificate {
+    Valid {
+        reference: X509CertificateReference,
+        friendly_name: String, // May be different that `X509CertificateDetails.name` TODO: give an example
```
AureliaDolo, line 42: todo

libparsec/crates/platform_pki_2_electric_boogaloo/src/shared/info.rs
touilleMan: Yes, this is because `intermediate_der_x509_certificates_end` contains a list of structures that contains references on `intermediate_der_x509_certificates_der` (basically `X509EndCertificate` contains a `&X509CertificateDer`)

libparsec/crates/platform_pki_2_electric_boogaloo/src/lib.rs
touilleMan: This will be removed later on, but for the moment this is still required since the GUI uses it

libparsec/crates/platform_pki_2_electric_boogaloo/src/pki_system.rs
touilleMan: :+1: done. I've documented what this field is about

libparsec/crates/platform_pki_2_electric_boogaloo/src/pki_certificate.rs
touilleMan: :+1: done.

libparsec/crates/platform_pki_2_electric_boogaloo/src/shared/info.rs
touilleMan: Can you point me to the code on master ? I don't see such thing :/

This is implemented for the server code :
https://github.com/Scille/parsec-cloud/blob/29759abf3b03f74328f9f60530bb612ba92f21e2/server/parsec/components/async_enrollment.py#L233-L255
(note server-side certificate "validation" is has no actual cryptographic verification and is just some initial checks considering the client is going to do the actual validation)

I agree that having a circular detection and limit are important here, however I think this is already handled by `X509EndCertificate::verify_for_usage` that is used here (as this method can return a `Error::MaximumPathDepthExceeded` error) @FirelightFlagboy can you confirm that ?

AureliaDolo: I can't find it either :/ But I don't expect it to still be in master. It would have been part of the pki implementation that was done last October, before the async enrollment.

FirelightFlagboy (approved):

PR #12586, opened 2026-04-10
Rework `libparsec_platform_pki` 10: fix bindings

### LINKED ISSUE #12501: Use new PKI API
Use new PKI API added to `libparsec_plt_pki`

> Look for usage in `libparsec/src/async_enrollment.rs`

touilleMan:


libparsec/src/pki.rs
```diff
@@ -1,21 +1,137 @@
...
+
+    *guard = Some(std::sync::Arc::new(pki_system));
+    Ok(())
+}
+
+async fn get_pki_system() -> anyhow::Result<std::sync::Arc<libparsec_platform_pki::PkiSystem>> {
+    let guard = PKI_SYSTEM.lock().await;
+    guard
+        .as_ref()
+        .ok_or_else(|| anyhow::anyhow!("PKI system not initialized"))
```
AureliaDolo, line 62: should we make an attempt to initialized the pki system here ?

AureliaDolo (approved):

libparsec/src/pki.rs
touilleMan: Cannot do that since there is two different functions to initialize `pki_init_for_scws` vs `pki_init_for_native` (and it requires some parameters)

On top of that the `pki_init_...` is always supposed to be called before `get_pki_system`, so if that's not the case we'd rather have an error than trying to do some kind of hidden init ^^

PR #12601, opened 2026-04-13
Patch hostname address to use "app.parsec.cloud"

### LINKED ISSUE #12377: Update server address from device
### The problem

Some Parsec links provided by libparsec (join request link, file links) depend on the server address stored in the local device. 

This posed problems with devices created on the `saas-v3.parsec.cloud` domain that should (now) generate links with the `app.parsec.cloud` domain.

### Possible solutions

1. Allow to update server address in the GUI in order to prevent broken links. 
   - The good news is that libparsec already provides an `overwrite_server_address` function that is exposed via CLI and bindings. 
   - On desktop app, an advanced user could call the corresponding CLI command to update the server address. However, it is way more complicated on web, since the user needs to access the browser's console and call specific functions from bindings.
   - For security reasons, Parsec could eventually:
     - require device re-authentication?
     - suggest to create a recovery device before proceeding?
     - store a copy of the original device as a backup?
     - ...

2. Patch libparsec so that it always uses `app.parsec.cloud` instead of `saas-v3.parsec.cloud`.

mmmarcos:
Alternative to #12563

Some Parsec links provided by libparsec (join request link, file links) depend on the server address stored in the local device.

This posed problems with devices created on the legacy `saas-v3.parsec.cloud` domain that should (now) generate links with the `app.parsec.cloud` domain.

Closes #12377

### DIFF
--- a/libparsec/crates/types/tests/unit/addr.rs
+++ b/libparsec/crates/types/tests/unit/addr.rs
@@ -6,7 +6,10 @@ use std::str::FromStr;
 
 use libparsec_tests_lite::prelude::*;
 
-use crate::prelude::*;
+use crate::{
+    addr::{PARSEC_SAAS_SERVER_HOSTNAME_CURRENT, PARSEC_SAAS_SERVER_HOSTNAME_LEGACY},
+    prelude::*,
+};
 
 const ORG: &str = "MyOrg";
 const TOKEN: &str = "a0000000000000000000000000000001";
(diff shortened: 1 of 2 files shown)


libparsec/crates/types/tests/unit/addr.rs
```diff
@@ -258,6 +258,11 @@ fn good_addr_base(testbed: &dyn Testbed) {
     "https://example.com?foo=bar",
     |outcome| { p_assert_eq!(outcome, Ok("parsec3://example.com".parse().unwrap())); },
 )]
+// See: https://github.com/Scille/parsec-cloud/issues/12377
```
touilleMan: This test function only test for `ParsecAddr`, it would be better to test for ALL possible Addr types.

You can do that by using `#[rstest_reuse::apply(all_addr)]`, also you can use `testbed.assert_addr_ok_with_expected(<source>, <expected>)`

mmmarcos: Done

touilleMan (approved):

FirelightFlagboy (approved):

mmmarcos:
I've tested by adding these lines to `/etc/hosts`:
```
127.0.0.1       saas-v3.parsec.cloud
127.0.0.1       app.parsec.cloud
```
Starting the testbed server
```shell
./make rts
```
Starting the client with `TESTBED_SERVER` pointing to "saas-v3.parsec.cloud"
```shell
cd client/
export TESTBED_SERVER='parsec3://saas-v3.parsec.cloud:6770?no_ssl=true'
npm run electron:open
```
All displayed URLs and links copied (such as `invite`, `async_enrollment`, `path`) are shown with `app.parsec.cloud` domain.

PR #12610, opened 2026-04-14
[MS] Workspace trash page

### LINKED ISSUE #12356: Create a `Trash` page
## Required
To delete a workspace, we need a specific route which lead to the trash workspace page.

## Archive workspace page
This page should include:
- All trashed workspaces [(mockups)](https://www.figma.com/design/XW3SNWHV8BhPsPJpsOCV2h/Parsec-App?node-id=36616-187321&t=4NsrQZMNIHPlcEe9-11) 
- Workspace cards [(mockups)](https://www.figma.com/design/XW3SNWHV8BhPsPJpsOCV2h/Parsec-App?node-id=36623-190432&t=4NsrQZMNIHPlcEe9-4) 
  - Workspace name
  - Role mention (owner)
  - The time remaining before complete deletion
  - Options
- Contextual menu: [(mockups)](https://www.figma.com/design/XW3SNWHV8BhPsPJpsOCV2h/Parsec-App?node-id=36656-82406&t=4NsrQZMNIHPlcEe9-11) 
    - Restore

## Action of archiving
- Create a modal window to explain how the delete feature works [(mockups)](https://www.figma.com/design/XW3SNWHV8BhPsPJpsOCV2h/Parsec-App?node-id=36641-59229&t=4NsrQZMNIHPlcEe9-11)

⚠️ For now, we cannot delete permanently the workspace since we sent to the server a `deletion_date: $minimum_archiving_period` and we cannot override this value from the GUI (only a server admin can do it). So the deletion will be done only after the delay. **The default delay is 30 days**.

Ironicbay:
Closes https://github.com/Scille/parsec-cloud/issues/12356


client/src/views/header/HeaderPage.vue
```diff
@@ -596,7 +637,7 @@ async function openWorkspacesSwitchModal(): Promise<void> {
     expandToScroll: false,
     initialBreakpoint: isLargeDisplay.value ? undefined : 1,
     componentProps: {
-      currentRoute: currentRouteIs(Routes.Workspaces) ? Routes.Workspaces : Routes.Archived,
+      currentRoute: getCurrentRouteName(),
```
Max-7: This means you can in theory pass `Routes.Home` or `null` as a prop, but the component is defined to accept `Routes.Workspaces | Routes.Archived | Routes.Trash | undefined`. The linter is not smart enough to check for props when using `modalController.create`, but it is still a problem. Check here that you're passing the right value.

client/src/components/workspaces/utils.ts
```diff
@@ -343,6 +347,35 @@ async function restoreWorkspace(
...
+        data: { workspace: workspace.name },
+      },
+      level: result.ok ? InformationLevel.Success : InformationLevel.Error,
+    }),
+    PresentationMode.Toast,
+  );
+  if (result.ok) {
+    await eventDistributor.dispatchEvent(Events.WorkspaceTrashSync, { workspaceId: workspace.id });
+  } else {
+    console.log(result);
```
Max-7: No need to log, all calls to libparsec are automatically logged.

Max-7 (changes requested):

Max-7 (approved):

PR #12620, opened 2026-04-15
refactor(electron): Set feature flags in publish configuration

FirelightFlagboy:
Previously we read the `process.env.HARDENED` to try to know is the build was hardened, but that value is not fixed on build but at runtime.

To fix that we needed to have a file with hard-coded values. Initially I though of using a JSON file in `src` but `electron-builder` does not provided a simple way to overwrite it. Instead, I've extended the generated `publishConfig` JSON to include a `features` section. That JSON is read upon startup (with a fallback to default values if missing (e.g: during dev)) and wrapped by a class.

Other:

- Invert the meaning of feature flags `DISABLE_UPDATES` to `enableUpdates()` and `DISABLE_SENTRY` to `enableSentry()`:

  It's simpler to work of additive features

- Remove `--verbose` flag in `cargo build` command

### DIFF
--- a/client/electron/src/features.ts
+++ b/client/electron/src/features.ts
@@ -1,0 +1,35 @@
+// Parsec Cloud (https://parsec.cloud) Copyright (c) BUSL-1.1 2016-present Scille SAS
+
+import { CustomPublishOptions, ParsecFeatures } from '../assets/publishConfig';
+
+const DEFAULT_FEATURES: ParsecFeatures = {
+  hardened: false,
+};
+
+export default class FeaturesFlag {
+  private data: ParsecFeatures;
+
+  constructor() {
+    let data: CustomPublishOptions | undefined = undefined;
+    try {
+      data = require('../assets/publishConfig.json');
+    } catch {}
+    this.data = data ? data.features : DEFAULT_FEATURES;
+
+    console.debug('Configured features:', this.data);
+  }
+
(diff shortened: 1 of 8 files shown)


client/electron/src/features.ts
```diff
@@ -0,0 +1,33 @@
...
+    let data: CustomPublishOptions | undefined = undefined;
+    try {
+      data = require('../assets/publishConfig.json');
+    } catch {}
+    this.data = data ? data.features : DEFAULT_FEATURES;
+
+    console.debug('Configured features:', this.data);
+  }
+
+  enableSentry(): boolean {
```
mmmarcos: Since this is mostly used in `if` statements I would rename it

```suggestion
  sentryEnabled(): boolean {
```
or
```suggestion
  isSentryEnabled(): boolean {
```

client/electron/src/features.ts
```diff
@@ -0,0 +1,33 @@
...
+    return !this.hardened();
+  }
+
+  enableUpdates(): boolean {
```
mmmarcos:
```suggestion
  updatesEnabled(): boolean {
```
```suggestion
  isAutoUpdateEnabled(): boolean {
```
```suggestion
  isAppUpdateEnabled(): boolean {
```

mmmarcos (approved):
LGTM

PR #12624, opened 2026-04-15
docs: restructuring deployment guide

mmmarcos:
- Added requirements for server deployment
- Added description for required and optional components
- Moved "TLS certificates" and "env files" sections to "Preparation"
- Add description for the used env variables
- Renamed some sections and try to limit section depth

### DIFF
--- a/docs/hosting/deployment/index.rst
+++ b/docs/hosting/deployment/index.rst
@@ -14,70 +14,94 @@
 Server Deployment
 =================
 
-This section covers how to deploy the Parsec server.
+This section describes how to deploy the **Parsec Server** on Linux.
 
-.. _Parsec: https://parsec.cloud
+The steps and requirements described in this section may vary based on your specific needs.
+It is recommended to deploy and observe performance on a pilot project prior to using in production.
 
-Requirements
-************
+.. note::
 
-- `Docker`_ and the `docker-compose`_ plugin.
-- :ref:`Parsec client <doc_userguide_install_parsec>`.
-- :ref:`Parsec CLI <doc_hosting_install_cli>`.
+  Some setup and administrative operations must be performed with **Parsec CLI** on Linux.
+  Please refer to the :ref:`Install Parsec CLI on Linux <doc_hosting_install_cli>` section.
 
-.. _Docker: https://www.docker.com
-.. _docker-compose: https://docs.docker.com/compose/
+Deployment options
+==================
 
-Preamble
-********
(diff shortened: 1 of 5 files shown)


docs/hosting/deployment/index.rst
```diff
@@ -14,95 +14,117 @@
...
+Prerequisites
+=============
 
-  .. note::
+The Parsec Server depends on the following components:
 
-    The `Parsec`_ server need access to an ``S3 object storage``-like service, not necessarily ``AWS S3``
+- A `PostgreSQL`_ database to store Parsec metadata.
+- An S3-like object storage (e.g. `OpenStack Swift`_ or `Amazon S3`_) to store encrypted data.
+- An `SMTP server`_ to allow sending emails from the Parsec Server.
```
AureliaDolo, line 42: It's a nitpick and it could be tested, but the server should not have too much issue not having a smtp server. Obviously it will not send mail, but it should not return errors (only failed email sent status). 
But I get why it could be considered as mandatory as there is no proper way to disable it.

AureliaDolo (approved):

docs/hosting/deployment/index.rst
mmmarcos: Actually, there is a way to disable it by using "MOCKED" as the stmp host:
https://github.com/Scille/parsec-cloud/blob/d403cf43e09afa9f674da86a2acf09100824d896/server/parsec/cli/run.py#L400

However, I am not sure it's handled correctly because the [`_mocked_send_email`](https://github.com/Scille/parsec-cloud/blob/master/server/parsec/components/email.py#L71) function prints everything to `stderr` (not a configurable file) and it seems to store messages ad infinitum.

AureliaDolo: Because it's (only) used to test that the mails are sent properly, so it's not expected to last long

docs/hosting/deployment/index.rst
```diff
@@ -14,95 +14,117 @@
...
+Deployment options
+==================
 
-Preamble
-********
+Parsec offers the following deployment options:
 
-The `Parsec`_ server depends on the following external components in order to work properly:
+- **Direct installation on Linux server**
+- **Container-Based deployment**
```
FirelightFlagboy: **chore:** Missing link to section

mmmarcos: done

FirelightFlagboy (approved):

PR #12625, opened 2026-04-16
Fix realm deletion: org stat & key bundle not cleared

touilleMan:


server/parsec/components/memory/realm.py
```diff
@@ -1246,9 +1246,15 @@ async def delete_2_do_delete_metadata(
         # All checks are good, now we do the actual deletion
 
         realm.is_deleted = True
-        # For simplicity, we don't actually remove vlobs/blocks/keys bundle, as
-        # the `is_deleted` flag should be enough to simulate this behavior from
-        # the client point of view.
+        realm.vlobs = {}
+        for key_rotation in realm.key_rotations:
+            # Simulate what is done in PostgreSQL: keep the certificate
+            # bu remove the keys bundle and keys bundle accesses
```
mmmarcos:
```suggestion
            # but remove the keys bundle and keys bundle accesses
```

mmmarcos (approved):

server/parsec/components/memory/realm.py
touilleMan: :+1: done.

PR #12627, opened 2026-04-16
SCWS: dynamic import of scwsapi.js

touilleMan:


client/src/plugins/libparsec/web_shared_worker.ts
```diff
@@ -71,6 +71,23 @@ async function maybeInit(): Promise<LibParsecPlugin> {
...
+    // Hack ahead:
+    // As part of `pkiInitForScws`, we need to dynamically load `scwsapi.js` in the
+    // context of the ShareWorker.
+    // However this cannot be done using `web_sys`, so we intercept the call to `pkiInitForScws`,
+    // do the import, and finally put the `SCWS` object in the global scope where the actual
+    // `pkiInitForScws` will access it.
+    const originalPkiInitForScws = module.pkiInitForScws.bind(module);
+    module.pkiInitForScws = async (...args: Parameters<typeof module.pkiInitForScws>) => {
+      const scwsapi = await import(/* @vite-ignore */ args[2]);
+      (self as any).SCWS = scwsapi.SCWS;
```
FirelightFlagboy: **nitpick:** Use [globalThis](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/globalThis)

FirelightFlagboy (approved):

AureliaDolo (approved):

client/src/parsec/async_enrollment.ts
```diff
@@ -73,7 +74,43 @@ import { toRaw } from 'vue';
...
+      const scwsapiAppCertificateTag = document.querySelector(`meta[name="${SCWS_APP_CERTIFICATE}"`) as HTMLMetaElement | null;
+      if (!scwsapiAppCertificateTag?.content) {
+        console.log(`Not meta tag '${SCWS_APP_CERTIFICATE}' present`);
+        return { ok: false, error: { tag: PkiSystemInitErrorTag.NotAvailable, error: `Not meta tag '${SCWS_APP_CERTIFICATE}' present` } };
+      }
+      const parsecServerTag = document.querySelector(`meta[name="${WEB_PARSEC_SERVER}"`) as HTMLMetaElement | null;
+      if (!parsecServerTag?.content) {
+        console.log(`Not meta tag '${WEB_PARSEC_SERVER}' present`);
+        return { ok: false, error: { tag: PkiSystemInitErrorTag.NotAvailable, error: `Not meta tag '${WEB_PARSEC_SERVER}' present` } };
+      }
```
FirelightFlagboy, line 97: **question:** Why adding a new meta instead of using current location?

Max-7: Because of `no_ssl=true`, it doesn't appear in the location :/

FirelightFlagboy: But you will be in http :thinking:

PR #12633, opened 2026-04-16
[MS] Add history for archived and deleted workspaces

Ironicbay:


client/src/components/workspaces/utils.ts
```diff
@@ -274,6 +274,9 @@ export async function openArchivedWorkspaceContextMenu(
 
   if (data !== undefined) {
     switch (data.action) {
+      case WorkspaceAction.ShowHistory:
+        await navigateTo(Routes.History, { query: { documentPath: '/', workspaceHandle: workspace.handle, readOnly: true } });
```
Max-7: The `readOnly` url params is never used for history I think, better to not add it.

client/src/views/workspaces/WorkspaceHistoryPage.vue
```diff
@@ -210,6 +224,9 @@ const headerButtonsRef = useTemplateRef<HTMLDivElement>('headerButtons');
 const topbarRightRef = useTemplateRef<HTMLDivElement>('topbarRight');
 const { windowWidth, isLargeDisplay, isSmallDisplay } = useWindowSize();
 const pathOpener = usePathOpener();
+const isArchived = ref(false);
```
Max-7: Is there a point to dupplicating the variables instead of just doing `workspaceInfo.isArchived`?

client/src/views/workspaces/WorkspaceHistoryPage.vue
```diff
@@ -210,6 +224,9 @@ const headerButtonsRef = useTemplateRef<HTMLDivElement>('headerButtons');
...
+const isTrashed = ref(false);
+const isManager = ref(false);
```
Max-7: Better as a computed.

client/src/views/workspaces/WorkspaceHistoryPage.vue
```diff
@@ -81,6 +83,16 @@
                 :max-shown="2"
                 :available-width="breadcrumbsWidth"
               />
+              <ion-text
```
Max-7, line 91: The location is a bit weird, it feels a bit like it's part of the path.

Max-7 (changes requested):

Max-7 (approved):

PR #12638, opened 2026-04-17
PKI Web SCWS Mutual Authentication: server minimal implementation

touilleMan:


server/parsec/cli/run.py
```diff
@@ -663,6 +680,28 @@ async def run_cmd(
...
             )
 
+        if (
+            scws_idopte_public_keys_file is not None
+            and scws_web_application_private_key_file is not None
+        ):
+            idopte_public_keys_pem = await asyncio.to_thread(scws_idopte_public_keys_file.read_text)
+            web_application_private_key_pem = await asyncio.to_thread(
+                scws_web_application_private_key_file.read_text
+            )
```
FirelightFlagboy, line 690: **polish:** Useless async, you are effectively doing sequential reading.

To have a meaningful effect, you should use `asyncio.gather` and wait for both future at the same time.

touilleMan: I agree this is needlessly complex, however I must do it this way to avoid code smell from ruff that doesn't like a synchronous file read within an async function.
(also I'd rather avoid doing a `asyncio.gather` here since it would further add complexity to the code for no real benefit since reading is already fast and this is done only once at startup)

FirelightFlagboy (approved):

PR #12645, opened 2026-04-20
[MS] Use new pki handle API

Max-7:


client/src/views/home/HomePage.vue
```diff
@@ -846,13 +850,11 @@ async function onOrganizationSelected(device: AvailableDevice): Promise<void> {
...
-        device,
-        constructAccessStrategy(
-          device,
-          PrimaryProtectionStrategy.useSmartcard(toRaw((device.ty as AvailableDeviceTypePKI).certificateRef)),
-        ),
-      );
+      const certResult = await openCertificate((device.ty as AvailableDeviceTypePKI).certificateRef);
+      if (!certResult.ok) {
+        return;
+      }
```
Ironicbay, line 857: Wouldn't we want to raise an error there?

client/src/views/home/HomePage.vue
```diff
@@ -659,8 +659,12 @@ async function finalizeRequest(request: AsyncEnrollmentRequest): Promise<void> {
       return;
     }
     const identitySystem = request.enrollment.identitySystem as AvailablePendingAsyncEnrollmentIdentitySystemPKI;
-    identityStrategy = makeAcceptPkiIdentityStrategy(serverAddr, toRaw(identitySystem.certificateRef));
-    primaryProtection = PrimaryProtectionStrategy.useSmartcard(toRaw(identitySystem.certificateRef));
+    const certResult = await openCertificate(identitySystem.certificateRef);
+    if (!certResult.ok) {
+      return;
+    }
```
Ironicbay, line 665: Same question

client/src/views/home/HomePage.vue
```diff
@@ -1007,6 +1009,9 @@ async function login(device: AvailableDevice, access: DeviceAccessStrategy): Pro
   }
 
   const result = await parsecLogin(device, access);
+  if (access.primaryProtection.tag === DevicePrimaryProtectionStrategyTag.PKI) {
+    await closeCertificate(access.primaryProtection.pkiPrivateKeyHandle);
```
Ironicbay, line 1023: Does this function as a classic file opening? Does this call cover both `openCertificate` calls in this file?

client/src/components/devices/ChooseAuthentication.vue
```diff
@@ -271,7 +272,10 @@ function getSaveStrategy(): DeviceSaveStrategy | undefined {
     return constructSaveStrategy(PrimaryProtectionStrategy.useOpenBao(openBaoClient.value.getConnectionInfo()));
   } else if (authentication.value === DevicePrimaryProtectionStrategyTag.PKI) {
     if (certif.value) {
-      return constructSaveStrategy(PrimaryProtectionStrategy.useSmartcard(toRaw(certif.value.handle)));
+      const certResult = await openCertificate(certif.value.reference);
```
Ironicbay, line 275: Bouncing off my previous comment, does this call need a corresponding `closeCertificate()` call?

client/tests/e2e/specs/document_context_menu.spec.ts
Ironicbay, line 1: Is this relevant to the PR?

Max-7: Yeah, can't merge otherwise because of flaky tests. It's not ideal but I've spent too much time on that specific file on those specific tests.

client/src/views/home/HomePage.vue
Max-7: Yeah it does, we want to close the handle when possible.
It gets a bit tricky because the handle gets forwarded: we have cases where we create a handle during the org creation or the join process to save the device, but then use the same "PrimatyProtectionStrategy" for automatically login to the newly created device. So I think this covers all cases.

client/src/components/devices/ChooseAuthentication.vue
Max-7: It can't be here since this returns the newly created save strategy. The handle has to stay valid while this save strategy is used.

client/src/views/home/HomePage.vue
Max-7: Yeah probably. I've added one using the function to handle login error, adding an error for pki which was missing.

client/src/views/home/HomePage.vue
Max-7: It's a weird case that should not really happen, like a few lines above. We just selected the certificate, we know it exist.
We still have the automatic log from the call to libparsec to investigate, but I wouldn't know what to display to the user because it should never happen.

Ironicbay (approved):

PR #12648, opened 2026-04-20
docs: Add section "Application architecture"

mmmarcos:
Also: 
- minor changes to `install_cli`
- add `sphinxmermaid` to support Mermaid diagrams (not used in this PR, but would be useful for the next ones)


docs/hosting/install_cli.rst
```diff
@@ -7,18 +7,19 @@ Install Parsec CLI (Linux)
 ==========================
 
 In order to deploy and maintain Parsec Server, you would need to perform some operations with
-Parsec CLI for Linux. These operations can be performed from another machine.
+Parsec :abbr:`CLI (Command-Line Interface)` for Linux. These operations can be performed from another machine.
```
FirelightFlagboy, line 10: **thought:** abbr is nice, but does it exist a systematique way to do that?

I cannot see myself putting repeated abbr everywhere

FirelightFlagboy (approved):

docs/hosting/install_cli.rst
mmmarcos: Good point, I don't think there is a way to add it systematically. 
I guess we can just use it when introducing a new term / at the beginning of a section (but not necessarily at each occurrence).

PR #12651, opened 2026-04-21
Fix cago deny smells

touilleMan:


deny.toml
```diff
@@ -123,13 +117,10 @@ allow = [
...
     "BSD-2-Clause",
     "BSD-3-Clause",
-    "Unicode-DFS-2016",
     "ISC",
     "Zlib",
     "MPL-2.0",
     "Unicode-3.0",
-    "OpenSSL",
-    "CDLA-Permissive-2.0",
 ]
```
FirelightFlagboy, line 127: **question:** What is the reason to remove license that was previously authorized?

If the response is no dependencies are using those licenses, revert the change.

FirelightFlagboy (changes requested):

deny.toml
touilleMan: `cargo deny` was issuing warning with them.
I personally don't care if we keep them or not, but having warnings creates unnecessary noise and it's really simple to re-add them if we later use a crate that need use those licenses

mmmarcos: https://github.com/Scille/parsec-cloud/pull/12651/changes/1b9ac8133024fcd55eaff64a38dded5024e6f4d4

FirelightFlagboy (approved):

PR #12653, opened 2026-04-21
[MS] Use minimumArchivingPeriod server setting

Ironicbay:


client/src/components/workspaces/utils.ts
```diff
@@ -351,9 +352,14 @@ async function restoreWorkspace(
 }
 
 async function trashWorkspace(workspace: WorkspaceInfo, informationManager: InformationManager, eventDistributor: EventDistributor) {
+  let archivingDays: number = 30;
+  const clientResult = await getClientInfo();
+  if (clientResult.ok) {
+    archivingDays = Number(clientResult.value.serverOrganizationConfig.minimumArchivingPeriod) / (60 * 60 * 24); // seconds to days
```
Max-7, line 358: What if this value is 0? Or 60? I think it's a mistake to try to convert it to days and we should instead do `DateTime.utc().plus({ seconds: minimumArchivePeriod })` and display the date directly, like `The workspace will be deleted on April, 21st 2026 at 6PM`. It's much more versatile and handle all cases.

The deletion date can also be passed to `trashWorkspace` to avoid calling getClientInfo a second time and recalculating everything.

client/src/parsec/types.ts
```diff
@@ -470,6 +470,7 @@ interface OrganizationInfo {
   organizationAddr: ParsecOrganizationAddr;
   organizationId: OrganizationID;
   creationDate?: DateTime;
+  minimumArchivingPeriod?: bigint;
```
Max-7: Type it as number and do the conversion (if needed) in getOrganizationInfo() directly. This way when we're using it in a component, we don't have to worry about the conversion.

Max-7 (approved):

PR #12654, opened 2026-04-21
[MS] Use server config to allow/forbid authentication methods

Max-7:
See the changes in main.ts to simulate allowed auth methods.

### DIFF
--- a/client/src/locales/en-US.json
+++ b/client/src/locales/en-US.json
@@ -1430,6 +1430,7 @@
                 "unavailable": "This method is not allowed by this server."
             }
         },
+        "forbidden": "This organization does not allow this authentication method.",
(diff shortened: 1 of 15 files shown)


client/src/locales/en-US.json
```diff
@@ -1429,6 +1429,7 @@
                 "unavailable": "This method is not allowed by this server."
             }
         },
+        "forbidden": "This organization does not allow this authentication method.",
```
Ironicbay, line 1433: **nitpick**: the text is a bit too long and overflows in the `AuthenticationCard`.
You could replace it with `Unavailable on this organization` / `Indisponible sur cette organisation`

Ironicbay (approved):

PR #12655, opened 2026-04-21
fix(sign-windows): Add `-Hardened` param to build & sign hardened build

FirelightFlagboy:
- Update cmd script to pass extra argument to powershell script
- Python script call cmd script with param `-Hardened` when needed


mmmarcos (approved):
LGTM, didn't test it though. @Max-7 I'm counting on you ;)

PR #12657, opened 2026-04-21
[MS] Update lost access - recovery UI

fabienSvtr:
## What's changed?

The user can decided which recovery method he wants among:
- Other device/browser connected
- Recovery files

I added the mention of recovery a `session`  because the `device` word is misleading for the user because he can have multiple sessions (e.g: Firefox + Chrome + client app).

https://github.com/user-attachments/assets/675badac-7ab9-44bd-a769-db426495581f

closes #12568


client/src/views/home/HomePage.vue
```diff
@@ -49,7 +49,7 @@
               <template v-else-if="state === HomePageState.CustomerArea">
                 <client-area-login-page />
               </template>
-              <template v-else-if="state === HomePageState.Login">
+              <template v-else-if="state === HomePageState.Login || state === HomePageState.ForgottenPassword">
```
Max-7: Can't you just remove the `ForgottenPassword` state? I don't think it serves any purpose.

client/src/views/home/HomePage.vue
```diff
@@ -444,16 +438,16 @@ async function onInvitationClicked(invitation: AccountInvitation): Promise<void>
   await openJoinByLinkModal(invitationAddr);
 }
 
-async function onJoinOrganizationClicked(): Promise<void> {
+async function onJoinOrganizationClicked(isCustomText?: boolean): Promise<void> {
```
Max-7: `isCustomText` doesn't mean anything. Reading the variable name, my guess is I can pass other arguments to customize the texts. That's not what happens. It's about whether to show texts related to the recovery of a device, or to adding a new device.

client/src/views/home/HomePage.vue
```diff
@@ -900,7 +894,7 @@ async function handleLoginError(device: AvailableDevice, error: ClientStartError
       });
       if (answer === Answer.Yes) {
         selectedDevice.value = device;
-        state.value = HomePageState.ForgottenPassword;
+        await openForgottenPasswordModal(device);
```
Max-7: I don't understand why a function called `openForgottenPasswordModal` is called when we're using the keyring currently.

client/src/views/home/HomePage.vue
```diff
@@ -1071,9 +1065,34 @@ async function backToPreviousPage(): Promise<void> {
   }
 }
 
-function onForgottenPasswordClicked(device?: AvailableDevice): void {
+async function onForgottenPasswordClicked(device?: AvailableDevice): Promise<void> {
+  await openForgottenPasswordModal(device);
+}
+
+async function openForgottenPasswordModal(device?: AvailableDevice): Promise<void> {
```
Max-7: It's not about a password being forgotten, it's about the access being lost.

client/src/views/home/HomePage.vue
```diff
@@ -1071,9 +1065,34 @@ async function backToPreviousPage(): Promise<void> {
...
   selectedDevice.value = device;
```
Max-7: I'm not sure this is needed.

client/src/views/home/HomePage.vue
```diff
@@ -1071,9 +1065,34 @@ async function backToPreviousPage(): Promise<void> {
...
+    showBackdrop: true,
+    componentProps: {
+      device: device,
+      informationManager,
+    },
+  });
+  await modal.present();
+  const { role, data } = await modal.onWillDismiss();
+
+  if (role === MsModalResult.Confirm && data?.recoveryMethod === 'connected-device') {
```
Max-7: Have enums for this, not raw strings.

client/src/views/devices/ImportRecoveryDeviceShamirPage.vue
Max-7, line 1: Should not be here.

client/src/views/devices/ChooseRecoveryMethodModal.vue
Max-7, line 1: Wrong name for this file.

Max-7, line 1: It should get split.
First, we display a modal that gives a choice between the recovery methods and returns the chosen one. Then we open a modal specific to that recovery method. Trying to do everything in the same file is impossible to do properly with the number of states.

Max-7 (changes requested):

client/src/views/home/HomePage.vue
fabienSvtr: Sure, I replaced with `isRecoveryDevice`, I think it's more comprehensive.

client/src/views/home/HomePage.vue
fabienSvtr: I replaced with: `openRecoveryDeviceModal`

client/src/views/home/HomePage.vue
fabienSvtr: See my previous comment

client/src/views/home/HomePage.vue
```diff
@@ -444,16 +438,16 @@ async function onInvitationClicked(invitation: AccountInvitation): Promise<void>
   await openJoinByLinkModal(invitationAddr);
 }
 
-async function onJoinOrganizationClicked(): Promise<void> {
+async function onJoinOrganizationClicked(isRecoveryDevice?: boolean): Promise<void> {
```
Max-7: Still not well named. The purpose of this variable is to know whether or not we should display a text related to recovering an organization, not to know if a device is a recovery device.

client/src/views/devices/ChooseRecoveryMethodModal.vue
```diff
@@ -0,0 +1,347 @@
...
+      </template>
+    </div>
+  </ms-modal>
+</template>
+
+<script setup lang="ts">
+import RecoveryDeviceIcon from '@/assets/images/recovery-device-icon.svg?raw';
+import RecoveryFileIcon from '@/assets/images/recovery-file-icon.svg?raw';
+import RecoveryTrustIcon from '@/assets/images/recovery-trusted-user-icon.svg?raw';
+import { RecoveryMethodStep } from '@/views/devices/types';
```
Max-7: An internal step and the chosen method of recovery are two different things.

If I'm calling your modal to get the chosen method, I'm getting a `RecoveryMethodStep`, and I've no idea what to do with it. It contains values such as `Done` or `MethodChoice`.

Those are separate and the internal state should not be exposed.

client/src/views/devices/ChooseRecoveryMethodModal.vue
```diff
@@ -0,0 +1,347 @@
...
+    case RecoveryMethodStep.MethodChoice:
+      return { disabled: false, label: 'TextInputModal.cancel' };
+    case RecoveryMethodStep.ConnectedDevice:
+      return { disabled: false, label: 'HomePage.topbar.back', onClick: goBack };
+    default:
+      return undefined;
+  }
+});
+
+const confirmButton = computed(() => {
```
Max-7: Why have three computed that do the same thing? Put everything in one.

client/src/views/devices/ChooseRecoveryMethodModal.vue
```diff
@@ -0,0 +1,347 @@
...
+        disabled: false,
+        label: 'ImportRecoveryDevicePage.modal.actions.next',
+        onClick: nextStep,
+      };
+    default:
+      return { disabled: true };
+  }
+});
+
+async function nextStep(): Promise<boolean> {
```
Max-7, line 180: This seems over-complicated to me.

client/src/views/devices/ImportRecoveryDevice.vue
```diff
@@ -0,0 +1,516 @@
...
+    case RecoveryMethodStep.Authentication:
+      return { disabled: false, label: 'HomePage.topbar.back', onClick: goBack };
+    case RecoveryMethodStep.Done:
+      return undefined;
+    default:
+      return undefined;
+  }
+});
+
+const confirmButton = computed(() => {
```
Max-7: Again, three computed instead of one.

client/src/views/devices/ImportRecoveryDevice.vue
```diff
@@ -0,0 +1,516 @@
...
+              id="secret-key-input"
+              placeholder="ImportRecoveryDevicePage.secretKeyPlaceholder"
+              v-model="secretKey"
+              @change="checkSecretKeyValidity()"
+            />
+          </div>
+        </div>
+      </template>
+
+      <template v-else-if="currentStep === RecoveryMethodStep.Authentication && importedDevice && serverConfig">
```
Max-7: `serverConfig` should not be mandatory here.

Max-7 (changes requested):

client/src/views/home/HomePage.vue
fabienSvtr: Ok so I choose `showRecoveryText`, it may be more relevant.

client/src/views/home/HomePage.vue
```diff
@@ -36,7 +37,7 @@
                   @join-organization-click="onJoinOrganizationClicked"
                   @join-organization-with-link-click="openJoinByLinkModal"
                   @bootstrap-organization-with-link-click="openCreateOrganizationModal"
-                  @recover-click="onForgottenPasswordClicked"
+                  @recover-click="openRecoveryDeviceModal"
```
Max-7: Not `RecoveryDevice`, we're talking about organization recovery here. `recovery device` is one of the method.

Max-7 (approved):

PR #12660, opened 2026-04-22
[MS] Use WorkspacesSelfListChanged for archived/trashed workspaces

Ironicbay:


client/src/views/header/HeaderPage.vue
```diff
@@ -355,21 +365,7 @@ const workspaceSwitchModalTitle = computed(() => {
...
-  const workspaceHandle = getWorkspaceHandle();
-  if (workspaceHandle) {
-    const wInfo = await getWorkspaceInfo(workspaceHandle);
-    if (wInfo.ok) {
-      return { isArchived: wInfo.value.isArchived, isTrashed: wInfo.value.isTrashed };
-    }
-  }
-  return { isArchived: false, isTrashed: false };
-});
+const currentWorkspace: Ref<WorkspaceInfo | undefined> = ref(undefined);
```
Max-7, line 357: Remove `workspaceName`, it's pointless to have an additional variable if we can get the info using this variable.

Max-7 (changes requested):

Max-7 (approved):

PR #12662, opened 2026-04-22
[CLI] Rename remove to forget-local for clarity.

### LINKED ISSUE #8857: Client CLI: device remove is unclear
The CLI command `parsec device remove` is unclear since "remove" can mean multiple things:
1- remove from the server
2- remove from the client

The command does 2 (especially since 1 is not possible in Parsec !), but a regular user has no way of knowing this is what is going to happen.

So I think we should improve the documentation of this command to further stress the fact this command only work on the capacity to connect as this device on this local machine, typically:
- Explain it in the command documentation
- Explain what is going to happen in the confirmation message:
https://github.com/Scille/parsec-cloud/blob/df094f28a0e8886598135bb2f28336367f542d6e/cli/src/commands/device/remove.rs#L38

On top of that we may want to rename the command (`forget_local_device` ?)

AureliaDolo:
Fix #8857

### DIFF
--- a/newsfragments/8857.fix.rst
+++ b/newsfragments/8857.fix.rst
@@ -1,0 +1,1 @@
+[CLI] Rename ``device remove`` to ``device forget_local``.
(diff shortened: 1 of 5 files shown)


newsfragments/8857.fix.rst
```diff
@@ -0,0 +1 @@
+[CLI] Rename device remove to device forget_local.
```
FirelightFlagboy: **nitpick:** 
```suggestion
[CLI] Rename ``device remove`` to ``device forget_local``.
```

FirelightFlagboy (approved):

PR #12673, opened 2026-04-23
[MS] forces auth & totp according to server rules

Max-7:
Put this at the top of the main.ts:
```typescript
(window as any).TESTING_MOCK_ALLOWED_PROTECTION_METHODS = [{
  primary: 'AdvisoryDeviceFilePrimaryProtectionOpenBao', withTotp: true,
}];
```
which will triggers both authentication change & totp setup. You can then change the primary or withTotp to see how it reacts.


client/src/locales/en-US.json
```diff
@@ -1513,7 +1513,19 @@
                     "notSetup": "Multi-factor authentication is not set up.",
                     "wrongAuthentication": "The authentication is incorrect. Please try again."
                 }
+            },
+            "modalRequired": {
+                "title": "Multi-factor authentication required",
+                "subtitle": "Your organization requires multi-factor authentication. Please set it up to continue using this organization.",
```
```suggestion
                "subtitle": "Your organization requires multi-factor authentication. Please set it up to access this organization.",
```

client/src/locales/fr-FR.json
```diff
@@ -1515,7 +1515,19 @@
                     "notSetup": "L'authentification multifacteur n'est pas configurée.",
                     "wrongAuthentication": "L'authentification est incorrecte. Veuillez réessayer."
                 }
+            },
+            "modalRequired": {
+                "title": "Authentification multifacteur requise",
+                "subtitle": "Votre organisation requiert une authentification multifacteur. Veuillez la configurer pour continuer à utiliser cette organisation.",
```
```suggestion
                "subtitle": "Votre organisation requiert une authentification multifacteur. Veuillez la configurer pour accéder à cette organisation.",
```

client/src/views/totp/ActivateTotpModal.vue
Ironicbay, line 1: Modal title in `Steps.PromptAuthentication` states `password` for any current authentication method

Image

Ironicbay, line 1: You can remove the `ms-report-text` on lines 39-44

client/src/views/profile/AuthenticationPage.vue
Ironicbay, line 1: Small detail, the display is a bit broken when trying to change authentication with only one available: here I'm authenticated with SSO, and only SSO is available for authentication. The button is still clickable but it's confusing. Should this even be allowed? (changing between SSO authentications?)

Image

Ironicbay (approved):

client/src/views/profile/AuthenticationPage.vue
Max-7: Yeah it's vonluntary, for two reasons:
- in the future we'll add additional SSOs, so you need to be able to switch between them (ProConnect -> Github for example)
- you may want to use another account on the same SSO (a bit hard to test since you need to erase navigation data from the provider).

PR #12674, opened 2026-04-23
[CLI] Ask for password confirmation during organization bootstrap.

### LINKED ISSUE #10019: [🐛 | Bug]: CLI boostrap organization password prompt doesn't ask for double check
### Parsec version tested on:

3.3.3

### Platforms tested on:

Any

### Bug description:

[Currently when doing ](https://github.com/Scille/parsec-cloud/blob/050fceb7664512f9f1f1ca863a165f9a999711a6/cli/src/commands/organization/bootstrap.rs#L77-L83)

should be using `choose_password()` instead

### Relevant output:

_No response_

AureliaDolo:
Fix #10019

### DIFF
--- a/cli/src/commands/organization/bootstrap.rs
+++ b/cli/src/commands/organization/bootstrap.rs
@@ -76,7 +76,7 @@ pub async fn main(args: Args) -> anyhow::Result<()> {
     let human_handle = HumanHandle::new(email, &label)
         .map_err(|e| anyhow::anyhow!("Cannot create human handle: {e}"))?;
 
-    let password = read_password(if password_stdin {
+    let password = choose_password(if password_stdin {
         ReadPasswordFrom::Stdin
     } else {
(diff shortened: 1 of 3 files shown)


FirelightFlagboy (commented):
I want a test that pass the password through stdin with `--password-stdin` flag.

When the flag is enable, it should ask/read the password only one time from the stdin

AureliaDolo:
> I want a test that pass the password through stdin with --password-stdin flag.

That's what the existing tests are doing.

Although I'm not sure why the --password-stdin flag became with_password as a assert_cmd argument

FirelightFlagboy (approved):
The reason it's called `with_password` is be give the macro more meaning:

> Call this command with that password.

The macro does what it need to provide the password to the command (e.g.: By using the `--password-stdin` and writing the password on the stdin)

PR #12688, opened 2026-04-27
[MS] Load SCWS in the worker context

Max-7:


client/src/parsec/async_enrollment.ts
```diff
@@ -82,30 +82,26 @@ const _ASYNC_ENROLLMENT_PARSEC_API = {
...
-        // Import scwsapi dynamically and make it available in the global scope
-        const scwsapi = await import(scwsapiLocationTag.content);
-        (globalThis as any).SCWS = scwsapi.SCWS;
-        (globalThis as any).WEB_APPLICATION_CERTIFICATE = scwsapiAppCertificateTag.content;
-      } catch (err: any) {
-        console.error(`Failed to import scwsapi: ${err.toString()}`);
-        return { ok: false, error: { tag: PkiSystemInitErrorTag.NotAvailable, error: `Failed to import scwsapi (${err.toString()})` } };
-      }
-      return await libparsec.pkiInitForScws(window.getConfigDir(), parsecServerTag.content);
+      return await (libparsec.pkiInitForScws as any)(
```
touilleMan, line 109: Should add a comment do explain what is the dark magic behind this call

i.e. that a cast to any is needed since SCWS api location/app certificate are not part of the function signature, and that is done this way due to a middlelayer (fancy word for "a big ugly hack" :trollface: ) in `trampoline-web.ts` that is tasked with dynamically importing the `scwsapi.js` and plugging it to the (along wit the app certificate) in the global scope (which cannot be done directly in libparsec since the Rust bindings crate `js_sys` doesn't expose the needed Javascript API)

client/src/plugins/libparsec/trampoline-web.ts
```diff
@@ -300,6 +300,24 @@ async function withoutSharedWorker(): Promise<any> {
...
 
+      // SCWS needs to be dynamically loaded.
+      // We're loading it in TypeScript and attaching it to the global context.
+      if (name === 'pkiInitForScws') {
+        return async (configDir: string, parsecAddr: string, scwsLocation: string, certificate: string) => {
+          try {
+            console.log('Loading SCWS...');
+            const scwsapi = await import(/* @vite-ignore */ scwsLocation);
+            (globalThis as any).SCWS = scwsapi.SCWS;
+            (globalThis as any).WEB_APPLICATION_CERTIFICATE = certificate;
```
FirelightFlagboy: **nitpick:** Would use the `SCWS_` prefix for that
```suggestion
            (globalThis as any).SCWS_WEBAPP_CERT = certificate;
```

FirelightFlagboy (approved):

PR #12705, opened 2026-04-28
feat(pki): Impl `PlatformPkiSystem::list_user_certificates` for web platform

FirelightFlagboy:


libparsec/crates/platform_pki/src/pki_system.rs
```diff
@@ -56,7 +56,7 @@ pub enum AvailablePkiCertificate {
 }
 
 impl AvailablePkiCertificate {
-    pub fn load_der(der: &[u8]) -> Self {
+    pub fn load_der(friendly_name: Option<String>, der: &[u8]) -> Self {
```
AureliaDolo, line 59: I don't get from where the friendly name may come

AureliaDolo (approved):

libparsec/crates/platform_pki/src/pki_system.rs
FirelightFlagboy: From the caller :raised_eyebrow:

AureliaDolo: yes but how does the caller got it if they haven't parsed the cert yet ?

FirelightFlagboy: The friendly named does not come from the certificate it's a label associated with the certificate alongside it.

It's used by the user to better identify the certificate instead of relying on the subject that can be hard to read for non-technical person

PR #12714, opened 2026-04-29
Correct `client_archive_workspace` command to take a period of archiving

touilleMan:
close https://github.com/Scille/parsec-cloud/pull/12710

Providing a deletion date doesn't work when the minimal archiving period is used since this period is checked against the realm archiving certificate timestamp that hasn't been created yet (and hence will be newer thant the date used to determine the deletion date)


libparsec/crates/client/src/certif/realm_archive.rs
```diff
@@ -10,6 +10,28 @@ use super::{
...
+/// typically when the caller want to use the minimal archiving period (which
+/// would always be too short since the certificate timestamp is obtained
+/// later on).
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub enum RequestedRealmArchivingConfiguration {
+    Available,
+    Archived,
+    DeletionPlanned {
+        // Don't use `Duration` since it is cumbersome to expose to the bindings
+        archiving_period_in_seconds: u32,
```
FirelightFlagboy, line 31: **issue:** Use a 64 bits

https://en.wikipedia.org/wiki/Year_2038_problem

FirelightFlagboy (changes requested):

libparsec/crates/client/src/certif/realm_archive.rs
touilleMan: nop, this represents a period that is going to be *added* to the current time

So with a u32, we can at most ask the workspace to be deleted after an archiving period of... 136 years, should be enough ^^

Note I use a u32 here since it simplifies the conversion later on since `Duration::seconds()` takes a i64 (so there is no need to deal with overflows)

FirelightFlagboy (approved):

FirelightFlagboy (approved):

PR #12716, opened 2026-04-29
Fix unclear `minimum_archiving_period` name

touilleMan:
Right now there is no information that this thing is only deleting a workspace, so replacing it by `realm_minimum_archiving_period_before_deletion` better explains that


mmmarcos (approved):
LGTM

PR #12719, opened 2026-04-29
Remove after-sign and MACOS_SIGN_OPTIONS variable

Ironicbay:
After updating the certificate for notarization, this warning showed up on build:
```
Please specify notarization Team ID in the `APPLE_TEAM_ID` env var instead of `notarize.teamId`
```
Looking it up, I found this variable to be set as `MACOS_SIGN_OPTIONS` in `client/electron/package.js` which seems to be overriding the default global variable set in `package-desktop.yml`: 
```
export APPLE_TEAM_ID='${{ secrets.MACOS_NOTARIZATION_TEAM_ID }}'
```

After removing it, the notarization succeeded without warning, so I deemed it unneccessary.

--------

Regarding the `after-sign.js` removal, it simply doesn't seem to do anything. Even when changing the values set inside, the notarization proceeds as usual. So my guess it that nothing broke when it was implemented, and it might have been considered necessary for the process.


client/electron/scripts/after-sign.js
```diff
@@ -1,26 +0,0 @@
-// Parsec Cloud (https://parsec.cloud) Copyright (c) BUSL-1.1 2016-present Scille SAS
-
-const { notarize } = require('@electron/notarize');
```
FirelightFlagboy, line 3: **question:** Should we also remove that deps?

Ironicbay: It's the dependency being used to notarize so no, I believe it's implicitly used with the `--sign` flag in `package-desktop.yml`. In any case, the build fails without it.

mmmarcos (commented):
Without context, it is not clear why this code was redundant/unnecessary. It is done somewhere else? what was the issue? etc.

Please edit the commit message (and PR description) to briefly explain it.

FirelightFlagboy (approved):

mmmarcos (approved):

PR #12722, opened 2026-04-29
[SCWS] Implment get_validation_path for PlatformPkiCertificate

AureliaDolo:


FirelightFlagboy (changes requested):
You need to check the status first, `getTrust` will not fail if the certificate is not trusted.

And the trust state providing you with certificates does not indicate that they're trusted

libparsec/crates/platform_pki/src/web/pki_certificate.rs
```diff
@@ -42,7 +42,48 @@ impl PlatformPkiCertificate {
...
         &self,
     ) -> Result<X509ValidationPathOwned, PkiCertificateGetValidationPathError> {
-        unimplemented!("platform not supported")
+        let trust = self.0.get_trust().await.map_err(|_| {
+            PkiCertificateGetValidationPathError::Internal(anyhow::anyhow!("unable to get trust"))
+        })?;
+        if trust.status().is_err() {
+            return Err(PkiCertificateGetValidationPathError::Untrusted);
+        }
+        let mut certs: Vec<_> = trust.cert_path().into_iter().collect();
```
FirelightFlagboy: **nitpick:** Would have use a VecDeque instead.

It provide both `pop_front` & `pop_back` that simplify the logic to extract the leaf & root certificate.

And the remaining values can be converted back into a vec

libparsec/crates/platform_pki/src/web/pki_certificate.rs
```diff
@@ -42,7 +44,67 @@ impl PlatformPkiCertificate {
...
+        let now = DateTime::now();
+
+        let path = verify_certificate(&end_cert, &intermediates, &anchor, now)
+            .map_err(|_| PkiCertificateGetValidationPathError::Untrusted)?;
+
+        let intermediates = path
+            .intermediate_certificates()
+            .map(|cert| cert.der().into_owned())
+            .collect();
+        let root = path.anchor().to_owned();
```
FirelightFlagboy: **polish:** You have already done that step by calling `getTrust`

FirelightFlagboy (changes requested):

FirelightFlagboy (approved):

PR #12733, opened 2026-05-01
Workspace chain of command: new `CertificateOps::get_realm_can_self_promote_to_owner|self_promote_to_owner` methods

touilleMan:


libparsec/crates/client/src/certif/mod.rs
```diff
@@ -543,6 +547,26 @@ impl CertificateOps {
         realm_rename::rename_realm(self, realm_id, new_name).await
     }
 
+    /// Returns `true` if all OWNERs of the workspace are revoked and the current
+    /// user has the highest remaining role (so they can self-promote to OWNER).
+    pub async fn get_realm_can_self_promote_to_owner(
```
AureliaDolo, line 552: I assume the unused warning are resolved further down the PR train ?

libparsec/crates/client/src/certif/realm_self_promote_to_owner.rs
```diff
@@ -0,0 +1,240 @@
...
+) -> Result<bool, CertifGetRealmCanSelfPromoteToOwnerError> {
+    ops.store
+        .for_read(async |store| {
+            // OUTSIDER cannot become OWNER, hence cannot self-promote
+            if store.get_current_self_profile().await? == UserProfile::Outsider {
+                return Ok(false);
+            }
+
+            // The following code is very similar to what is in `add.rs`'s `check_realm_role_certificate_consistency`.
+            // However we cannot factorize them easily since the `store` object is of different types
```
AureliaDolo, line 31: Maybe one day we'll have a CertificatesStoreGuard trait, that would be implemented by both CertificatesStoreReadGuard and CertificatesStoreWriteGuard ? ^^

AureliaDolo (approved):

libparsec/crates/client/src/certif/mod.rs
touilleMan: Exactly ^^

libparsec/crates/client/src/certif/realm_self_promote_to_owner.rs
touilleMan: No we will put yet more crazy macros instead ! :trollface: 

More seriously, I agree that `CertificatesStoreReadGuard` vs `CertificatesStoreWriteGuard` are not that great (since they both share the ton of read-only methods), however async trait are also tricky and cumbersome...
At least that was the case when this code was implemented, we might want to re-assert this now :thinking:

AureliaDolo: I think I've implemented async traits once or twice without any issues ^^

PR #12742, opened 2026-05-04
Add knip npm dependency linter

mmmarcos:


FirelightFlagboy (approved):
LGTM for the configuration of knip, nothing to say about the change in deps listing

client/package.json
```diff
@@ -59,57 +57,44 @@
...
         "file-type": "^21.3.2",
         "fs-extra": "^11.3.3",
+        "ionicons": "^8.0.9",
         "luxon": "^3.5.0",
         "megashark-lib": "git+https://github.com/Scille/megashark-lib.git#6dbfbd41ef0116f39d456525fecee2d88726cf9b",
         "native-file-system-adapter": "^3.0.1",
         "pdfjs-dist": "^5.4.624",
+        "tslib": "^2.8.1",
         "qrcode-vue3": "^1.7.1",
-        "uuid": "^14.0.0",
```
Max-7, line 67: Why is this removed when it's used? See https://github.com/Scille/parsec-cloud/blob/0595f9d8674ba80d3ca25e58b092f2099453124d/client/src/services/eventDistributor.ts#L19 for example. And we have `@types/uuid` as a devDependencies too.

client/knip.config.ts
```diff
@@ -0,0 +1,60 @@
...
+    'src/services/smallDisplayWarning.ts',
+    'src/views/about/ChangesModal.vue',
+    'src/views/home/SummaryStep.vue',
+  ],
+  ignoreDependencies: [
+    '@electron/notarize', // reported as unused dev dependency
+    '@ionic/cli', // not used in code
+    'app-builder-lib', // electron-builder utility, only used in partial imports for typing
+    'electron-publish', // only used in electron/assets
+    'regedit', // reported as unused dependency from electron/package.json
```
Max-7: Used but imported dynamically only on Windows, which may explain why it's not reported properly.

client/package.json
```diff
@@ -59,57 +57,44 @@
...
         "vue-router": "^4.6.4"
     },
     "devDependencies": {
         "@capacitor/cli": "^7.4.4",
-        "@intlify/eslint-plugin-vue-i18n": "^4.1.0",
         "@ionic/cli": "^7.2.1",
         "@playwright/test": "^1.59.1",
         "@sentry/vite-plugin": "^5.2.0",
         "@types/adm-zip": "^0.5.7",
-        "@types/luxon": "^3.3.4",
```
Max-7, line 80: It seems to be declared twice: once in dependencies, once in devDependencies. We should keep this one, and remove the once in dependencies.

Max-7 (changes requested):

client/package.json
mmmarcos: Done

client/knip.config.ts
mmmarcos: Added this as comment.

client/package.json
mmmarcos: Seems that the knip config was outdated, I've re-added `uuid`. 

However, `@types/uuid` was still reported as unused, so I've removed it. Do you know if still needed? (check below)

BTW, your comment made me realize that I forgot to add knip to the CI, that's why it was not failing. I've just add a step to run it in the CI.

client/knip.config.ts
```diff
@@ -0,0 +1,62 @@
...
+    'src/parsec/mock_generator.ts',
+    'src/plugins/libparsec/trampoline-native.ts',
+    'src/services/screenshot.ts',
+    'src/services/smallDisplayWarning.ts',
+    'src/views/about/ChangesModal.vue',
+    'src/views/home/SummaryStep.vue',
+  ],
+  // Exclude dependencies reported as unused
+  ignoreDependencies: [
+    '@electron/notarize', // reported as unused dev dependency
```
FirelightFlagboy: **nitpick:** 
```suggestion 
    // reported as unused dev dependency
    // but used during signature of electron artifact for MacOS 
    '@electron/notarize',
```

client/knip.config.ts
```diff
@@ -0,0 +1,62 @@
...
+    '@ionic/cli', // not used in code
```
FirelightFlagboy: **praise:** 
```suggestion
	// not used in code but in script commands
    '@ionic/cli', 
```

client/package.json
Max-7: It may not be needed anymore, types may be shipped with the package now. If the linter doesn't complain, we're good.

Max-7 (approved):

client/knip.config.ts
mmmarcos: done

client/knip.config.ts
mmmarcos: done

PR #12745, opened 2026-05-04
[MS] Workspace deletion features and fixes

Ironicbay:


client/src/locales/en-US.json
```diff
@@ -786,7 +788,7 @@
             "noFavoriteWorkspaces": "You have not set starred any workspaces yet. Starred workspaces will be listed here.",
             "noHiddenWorkspaces": "You have not hidden any workspaces yet. Hidden workspaces will be listed here.",
             "noArchivedWorkspaces": "Your archived workspaces can be consulted here in read-only. Sharing is disabled unless the workspace is restored. Archived workspace owners can restore them.",
-            "noTrashedWorkspaces": "Your deleted workspaces can be consulted here in read-only. They will be permanently deleted after a set period. Deleted workspace owners can restore them before this date.",
+            "noTrashedWorkspaces": "Workspaces scheduled for deletion can be consulted here in read-only for {duration}, after which they will be permanently deleted. Owners of workspaces set for deletion can restore them before this date.",
```
```suggestion
            "noTrashedWorkspaces": "Workspaces scheduled for deletion can be consulted here in read-only for {duration}, after that time they will be permanently deleted. Workspace Owners can restore them before this date.",
```

client/src/locales/en-US.json
```diff
@@ -863,18 +865,35 @@
...
+            "yesBin": "Move to bin",
+            "yesDelete": "Permanently delete",
+            "info": "Workspaces scheduled for deletion will remain accessible in read-only for {duration}, after which they will be permanently deleted. Workspace owners can restore them before this date.",
             "loading": "Loading your deleted workspaces...",
             "backToWorkspace": "Back to my workspaces",
             "count": "1 workspace | {count} workspaces",
+            "question": {
+                "title": "Important information",
+                "subtitleBin": "If you delete the workspace {workspace}, its data, metadata and history will be deleted from the organization. It will become inaccessible to all users. The workspace will be moved to the bin and permanently deleted after {deletionDelay}.",
+                "subtitleDelete": "If you delete the workspace {workspace}, its data, metadata and history will be deleted from the organization. It will become inaccessible to all users.",
```
```suggestion
                "subtitleDelete": "If you delete the workspace {workspace}, its data, metadata and history will be deleted from the organization. It will no longer be accessible to any users.",
```

client/src/locales/en-US.json
```diff
@@ -863,18 +865,35 @@
...
+            "label": "Input '{workspace}' to proceed",
+            "yesBin": "Move to bin",
+            "yesDelete": "Permanently delete",
+            "info": "Workspaces scheduled for deletion will remain accessible in read-only for {duration}, after which they will be permanently deleted. Workspace owners can restore them before this date.",
             "loading": "Loading your deleted workspaces...",
             "backToWorkspace": "Back to my workspaces",
             "count": "1 workspace | {count} workspaces",
+            "question": {
+                "title": "Important information",
+                "subtitleBin": "If you delete the workspace {workspace}, its data, metadata and history will be deleted from the organization. It will become inaccessible to all users. The workspace will be moved to the bin and permanently deleted after {deletionDelay}.",
```
```suggestion
                "subtitleBin": "If you delete the workspace {workspace}, its data, metadata and history will be deleted from the organization. It will no longer be accessible to any users. The workspace will be moved to the bin and permanently deleted after {deletionDelay}.",
```

mmmarcos (commented):
Some minor suggestions

client/src/views/workspaces/SmallDisplayWorkspaceContextMenu.vue
```diff
@@ -152,7 +152,13 @@
               :icon="trash"
             />
             <ion-text class="button-large list-group-item__label">
-              {{ $msTranslate('WorkspacesPage.workspaceContextMenu.actionDelete') }}
+              {{
+                $msTranslate(
+                  workspaceDeletionDelay > 0
```
Max-7: It makes everything way too complicated for a simple string. Can't we just display something like `Mark workspace for deletion`? It's not okay to have to query the server to get some small variable just to display a slightly different text.

client/src/components/workspaces/utils.ts
```diff
@@ -661,3 +720,22 @@ export function compareWorkspaceRoles(role1: WorkspaceRole, role2: WorkspaceRole
   }
   return 0;
 }
+
+export function formatWorkspaceDeletionDelay(duration: number | undefined): Translatable {
```
Max-7, line 746: I'm really not a fan of trying to format dates/durations manually. It doesn't seem like I18n has something, but luxon can do it:
```typescript
function formatDuration(seconds: number): Translatable {
  const duration = Duration.fromObject({seconds: seconds}, {locale: I18n.getLocale()}).rescale();
  return valueAsTranslatable(duration.toHuman());
}
```
It can be customized a bit to change the formatting, and it can be moved into megashark:translate.ts. It has the drawback that it will not be refreshed if the locale changes.

client/src/views/workspaces/WorkspaceArchivePage.vue
```diff
@@ -116,6 +117,11 @@ async function refreshArchivedWorkspacesList(): Promise<void> {
       PresentationMode.Toast,
     );
   }
+
+  const infoResult = await getClientInfo();
+  if (infoResult.ok) {
+    workspaceDeletionDelay.value = Number(infoResult.value.serverOrganizationConfig.realmMinimumArchivingPeriodBeforeDeletion);
```
Max-7: Wasn't it typed as a number already?

Max-7 (commented):
I think there's something to check on the variable `realmMinimumArchivingPeriodBeforeDeletion`. It's defined as an always present u64 in definitions.ts, then gets redefined as an optional number in parsec/types.ts, then gets converted to Number everytime it's used.

client/src/views/workspaces/WorkspaceArchivePage.vue
Ironicbay: It's a `bigint`

client/src/views/workspaces/SmallDisplayWorkspaceContextMenu.vue
Ironicbay: I chose to use the `number` as prop instead of a `boolean` since there were already 2 optional `booleans` in `openWorkspaceContextMenu()` and `openArchivedWorkspaceContextMenu()`. It was easier to add a different type (`number`) than to handle every call to these functions case by case to differentiate between each boolean.

client/src/components/workspaces/utils.ts
```diff
@@ -338,6 +357,7 @@ async function restoreWorkspace(workspace: WorkspaceInfo, informationManager: In
 }
 
 async function trashWorkspace(workspace: WorkspaceInfo, informationManager: InformationManager) {
+  const { isLargeDisplay } = useWindowSize();
```
Max-7: Nop, cannot use a composable inside a free function. You probably will have warnings in your console.

client/src/components/workspaces/utils.ts
```diff
@@ -337,43 +341,108 @@ async function restoreWorkspace(workspace: WorkspaceInfo, informationManager: In
...
-    }),
-    PresentationMode.Toast,
+      placeholder: I18n.valueAsTranslatable(workspace.name),
+      okButtonText:
+        minimumArchivingPeriodInSeconds > 0 ? 'WorkspacesPage.trashWorkspace.yesBin' : 'WorkspacesPage.trashWorkspace.yesDelete',
+    },
+    isLargeDisplay,
   );
+
+  if (workspaceName !== workspace.name) {
```
Max-7: It could, the validator uses `localeCompare`, and this is a strict equality. Use `localeCompare` here too, to be safe.

client/src/components/workspaces/utils.ts
```diff
@@ -337,43 +341,108 @@ async function restoreWorkspace(workspace: WorkspaceInfo, informationManager: In
...
+      data: { workspace: workspace.name },
+    };
+    level = InformationLevel.Success;
+  } else {
+    switch (result.error.tag) {
+      case ClientArchiveWorkspaceErrorTag.Offline:
+        message = 'WorkspacesPage.trashWorkspace.error.offline';
+        break;
+      case ClientArchiveWorkspaceErrorTag.ArchivingPeriodTooShort:
+        message = 'WorkspacesPage.trashWorkspace.error.tooShort';
```
Max-7: I think it's better to use the same error as offline, at least for now. The user has no control over the duration, so it may be more confusing.

client/src/views/menu/MainMenu.vue
```diff
@@ -282,6 +282,7 @@
 
             <!-- Trashed -->
             <ion-text
+              v-show="workspaceDeletionDelay > 0"
```
Max-7: It should also check if we have trashed workspaces, since the server config may change. So either the deletion is superior to zero, or we have already trashed workspaces. I think you already have the workspace list in this component so it should not be too bothersome to add.

client/src/components/workspaces/utils.ts
```diff
@@ -337,43 +341,108 @@ async function restoreWorkspace(workspace: WorkspaceInfo, informationManager: In
...
+        key: 'WorkspacesPage.trashWorkspace.label',
         data: { workspace: workspace.name },
       },
-      level: result.ok ? InformationLevel.Success : InformationLevel.Error,
-    }),
-    PresentationMode.Toast,
+      placeholder: I18n.valueAsTranslatable(workspace.name),
+      okButtonText:
+        minimumArchivingPeriodInSeconds > 0 ? 'WorkspacesPage.trashWorkspace.yesBin' : 'WorkspacesPage.trashWorkspace.yesDelete',
+    },
```
Max-7, line 410: If you're rebased on master, you can add `theme: MsReportTheme.Error` to have a red button on the modal (you may have to delete the node_modules and npm install and restart the web server)

Max-7 (approved):

PR #12762, opened 2026-05-05
refactor(pki): Add `X509Pkcs11URI` during `AvailablePkiCertificate::load_der`

FirelightFlagboy:


AureliaDolo (approved):

libparsec/crates/platform_pki/src/windows/pki_system.rs
```diff
@@ -117,12 +119,23 @@ fn open_certificate(
...
     certificate_ref
-        .get_uri::<X509WindowsCngURI>()
-        .and_then(|uri| {
-            log::trace!("Looking for cert with uri: {uri}");
-            let mut uri = uri.clone();
-            find_in_store::find_cert_in_store(store, find_in_store::CertFilter::cert_id(&mut uri))
-                .next()
+        .get_uri::<X509Pkcs11URI>()
+        .cloned()
+        .map(Either::Left)
```
touilleMan, line 125: Since `Either` is only used once here, better just introduce a dedicate type in the body of this function, which allows better naming

libparsec/crates/types/src/pki/cert_ref.rs
```diff
@@ -171,17 +170,17 @@ pub struct X509Pkcs11URI {
     ///
     /// Used to link together a certificate & private key.
     /// That information is provided/configured by the smartcard so it may not be present.
-    pub id: Option<Bytes>,
+    pub id: Option<Vec<u8>>,
```
touilleMan, line 173: 👍

libparsec/crates/platform_pki/src/windows/pki_system.rs
FirelightFlagboy: I've remove Either from types

PR #12763, opened 2026-05-05
[MS] Removed capacitor-electron

Max-7:


client/scripts/copy_to_electron.cjs
```diff
@@ -0,0 +1,19 @@
...
+const DEST = path.join(WORKDIR, 'electron', 'app');
+
+if (!fs.existsSync(SRC)) {
+  console.error(`Source directory not found: ${SRC}`);
+  console.error('Run `npm run native:build` first');
+  process.exit(1);
+}
+
+console.log(`>>> cp -r ${path.relative(WORKDIR, SRC)} ${path.relative(WORKDIR, DEST)}`);
+fs.cpSync(SRC, DEST, { recursive: true });
```
FirelightFlagboy, line 18: **nitpick:** Use a symlink like how it's done in bindings/web

client/electron/src/setup.ts
FirelightFlagboy, line 1: **polish:** I would also rename the class `ElectronCapacitorApp`

client/capacitor.config.ts
```diff
@@ -15,13 +15,6 @@ const config: CapacitorElectronConfig = {
       backgroundColor: '#121212',
     },
   },
-  electron: {
-    trayIconAndMenuEnabled: true,
-    splashScreenEnabled: true,
-    deepLinkingEnabled: true,
-    deepLinkingCustomProtocol: 'parsec3',
-    customUrlScheme: 'parsec-desktop',
```
NicoTuxx, line 23: Why is this value deleted? Was it not applied? we need the `parsec-desktop` urlScheme

FirelightFlagboy: It's defined in `client/electron/capacitor.config.ts`

client/electron/src/setup.ts
Max-7: Yeah I agree, I renamed it to ParsecApp.

client/scripts/copy_to_electron.cjs
Max-7: Keeping it as a copy for now since it's what the script in capacitor-electron was doing. We can still change it later if needed.

FirelightFlagboy (approved):

mmmarcos (approved):

PR #12773, opened 2026-05-06
refactor(pki): Check values of cert info on windows with example

FirelightFlagboy:
Use `list_user_certificates` example to compare the value of issuer & serial from different sources:

- `CERT_CONTEXT->pCertInfo`
- Certificate DER


FirelightFlagboy:
Superseded #12774

PR #12774, opened 2026-05-06
fix(pki): Correct wrong endianess for serial number on Windows

FirelightFlagboy:
Windows expect the serial number to be provided in little endian (LE) as indicated by the field description of `SerialNumber` in: https://learn.microsoft.com/en-us/windows/win32/api/wincrypt/ns-wincrypt-cert_info

Since we extract the serial number from the certificate DER we de not have to convert the serial from LE to Big Endian (BE), but we still need to convert it during `open_certificate` from BE to LE.

We do the conversion by simply reversing the bytes order using `slice.reverse()`

I've also updated the `list_user_certificates` example to show the friendly name + certificate reference in full (rmp + base64).

### DIFF
--- a/libparsec/crates/platform_pki/examples/utils/mod.rs
+++ b/libparsec/crates/platform_pki/examples/utils/mod.rs
@@ -1,6 +1,6 @@
 // Parsec Cloud (https://parsec.cloud) Copyright (c) BUSL-1.1 2016-present Scille SAS
 
-use std::path::PathBuf;
+use std::{fmt::Display, path::PathBuf};
 
 use anyhow::Context;
 use bytes::Bytes;
@@ -8,7 +8,7 @@ use clap::{
     builder::{NonEmptyStringValueParser, TypedValueParser},
     error::{Error, ErrorKind},
 };
-use libparsec_types::X509CertificateHash;
+use libparsec_types::{X509CertificateHash, X509CertificateReference};
 
 // Not all examples uses `Base64Parser` so it is not always used.
 #[allow(dead_code)]
@@ -63,6 +63,7 @@ where
     }
 }
 
+#[allow(dead_code)]
 #[derive(Debug, Clone)]
 pub struct CertificateSRIHashParser;
 
@@ -107,5 +108,19 @@ impl ContentOpts {
                 .map(Into::into),
             (Some(_), Some(_)) | (None, None) => unreachable!("Handled by clap"),
         }
+    }
+}
+
+#[allow(dead_code)]
+/// Wrapper used to encode a certificate reference to be displayed.
+///
+/// It encode the reference using the rmp format & base64
+pub struct EncodedCertRef(pub X509CertificateReference);
+
+impl Display for EncodedCertRef {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        let blob = rmp_serde::to_vec(&self.0).expect("Cannot encode cert reference");
+
+        data_encoding::BASE64.encode_display(&blob).fmt(f)
     }
 }
--- a/libparsec/crates/platform_pki/tests/units/mod.rs
+++ b/libparsec/crates/platform_pki/tests/units/mod.rs
@@ -7,5 +7,7 @@ mod shared;
 mod sign;
 #[cfg(feature = "test-with-testbed")]
 mod testbed;
+#[cfg(windows)] // TODO: libparsec_platform_pki only supports Windows so far
+mod uri;
 mod utils;
 mod x509;
--- a/libparsec/crates/types/src/pki/cert_ref.rs
+++ b/libparsec/crates/types/src/pki/cert_ref.rs
@@ -180,6 +180,13 @@ pub struct X509Pkcs11URI {
     /// Subject of the certificate.
     pub der_subject: Vec<u8>,
     /// Serial number of the certificate.
+    ///
+    /// It is a positive integer assigned by the CA (cf. https://datatracker.ietf.org/doc/html/rfc5280#section-4.1.2.2).
+    /// It is stored here as an unsigned integer in big-endian, as it is the format used in X509 certificate format
+    /// (i.e. ASN.1 DER's INTEGER type, cf. https://datatracker.ietf.org/doc/html/rfc5280#section-4.1 and https://en.wikipedia.org/wiki/X.690)
+    ///
+    /// Be aware that Windows provide you the serial number in little-endian if the value is coming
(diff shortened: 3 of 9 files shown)


AureliaDolo (approved):

touilleMan (commented):
This seems really sensitive stuff, can we add a test for it ?

libparsec/crates/platform_pki/examples/utils/mod.rs
```diff
@@ -109,3 +110,14 @@ impl ContentOpts {
         }
     }
 }
+
+#[allow(dead_code)]
+pub struct EncodedCertRef(pub X509CertificateReference);
+
+impl Display for EncodedCertRef {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        let blob = rmp_serde::to_vec(&self.0).expect("Cannot encode cert reference");
```
touilleMan, line 122: RMP ? Why do we need to encode un msgack here ?
This seems rather odd since X509 stuff uses ASN1

I guess this is because `X509CertificateReference` is our own type that we only serialize as part of our schemas (hence using msgpack for this).

However I think there should be a comment about this here (typically stating this "encoded" format is not something standard and is only here for display purpose)

libparsec/crates/types/src/pki/cert_ref.rs
```diff
@@ -180,6 +180,8 @@ pub struct X509Pkcs11URI {
     /// Subject of the certificate.
     pub der_subject: Vec<u8>,
     /// Serial number of the certificate.
+    ///
+    /// Should be in big-endian format
```
touilleMan: Why should it be in big-endian format ?

I guess it is because it is how it is stored in the DER format, however this should be documented here (otherwise it begs the question "why aren't we just using following Windows and using little endian everywhere ?")

Also add a word about the fact, contrary to X509 DER format, Windows stores it in little-endian format (with link to  `https://learn.microsoft.com/en-us/windows/win32/api/wincrypt/ns-wincrypt-cert_info`), as it gives a better picture of why we need to document the endianess here

libparsec/crates/platform_pki/examples/utils/mod.rs
FirelightFlagboy: You guessed right `cert_ref` is our custom type and not part of the standard describe in RFC-5280.

> typically stating this "encoded" format is not something standard

Depends on what is standard for you, for me it's following the parsec standard regarding encoding of binary data structure (rmp + b64)

libparsec/crates/types/src/pki/cert_ref.rs
FirelightFlagboy: done

libparsec/crates/types/src/pki/cert_ref.rs
```diff
@@ -180,6 +180,13 @@ pub struct X509Pkcs11URI {
     /// Subject of the certificate.
     pub der_subject: Vec<u8>,
     /// Serial number of the certificate.
+    ///
+    /// Should be in big-endian format to follow the RFC-5280 and networking convention.
```
touilleMan:
```suggestion
    /// It is a positive integer assigned by the CA (cf. https://www.rfc-editor.org/rfc/rfc5280#section-4.1.2.2).
    /// It is stored here as an unsigned integer in big-endian, as it is the format used in X509 certificate format (i.e. ASN.1 DER's INTEGER type, cf. https://datatracker.ietf.org/doc/html/rfc5280#section-4.1 and https://en.wikipedia.org/wiki/X.690)
    ///
    /// Be aware that Windows CNG API provides the serial number in little-endian if the value is coming from [`CERT_INFO`](https://learn.microsoft.com/en-us/windows/win32/api/wincrypt/ns-wincrypt-cert_info)
```

FirelightFlagboy: done

libparsec/crates/platform_pki/examples/utils/mod.rs
touilleMan: > Depends on what is standard for you

From the point of view of the user of the CLI, something "standard" is a format largely used (e.g. X509 certificate is standard, Parsec certificate is not standard)

Seeing this `EncodedCertRef` we could be wondering if it is some largely agreed upon format (just like using the SHA256 of a x506 certificate is the standard way of identifying it), that's why I'd rather have a small comment explaining this is just a convenient way to display this opaque structure.

libparsec/crates/types/src/pki/cert_ref.rs
```diff
@@ -180,6 +180,14 @@ pub struct X509Pkcs11URI {
     /// Subject of the certificate.
     pub der_subject: Vec<u8>,
     /// Serial number of the certificate.
+    ///
+    /// Should be in big-endian format to follow the
```
touilleMan:
```suggestion
    /// It is a positive integer assigned by the CA (cf. https://www.rfc-editor.org/rfc/rfc5280#section-4.1.2.2).
    /// It is stored here as an unsigned integer in big-endian, as it is the format used in X509 certificate format (i.e. ASN.1 DER's INTEGER type, cf. https://datatracker.ietf.org/doc/html/rfc5280#section-4.1 and https://en.wikipedia.org/wiki/X.690)
    ///
    /// Be aware that Windows CNG API provides the serial number in little-endian if the value is coming from [`CERT_INFO`](https://learn.microsoft.com/en-us/windows/win32/api/wincrypt/ns-wincrypt-cert_info)
```

libparsec/crates/platform_pki/examples/utils/mod.rs
FirelightFlagboy: done

libparsec/crates/types/src/pki/cert_ref.rs
FirelightFlagboy: done

libparsec/crates/platform_pki/tests/units/mod.rs
```diff
@@ -7,5 +7,7 @@ mod shared;
 mod sign;
 #[cfg(feature = "test-with-testbed")]
 mod testbed;
+#[cfg(windows)]
```
touilleMan: If this is a test in the shared module, why is it only for windows?

(If this is a limitation in how the test is implemented, then add the `cfg(windows)` to the test function itself and add a comment explaining what is the issue with a TODO explaining what is needed (typically Unix support for a PkiSystem ?)

touilleMan (approved):

libparsec/crates/platform_pki/tests/units/mod.rs
FirelightFlagboy: done

PR #12778, opened 2026-05-07
[MS] Adds some log when loading cryptpad and use https protocol for the server if missing

Max-7:


client/src/services/environment.ts
```diff
@@ -307,8 +307,12 @@ async function getCryptpadServer(): Promise<string | null> {
...
     return null;
   }
-  CryptpadServerCache.set(connHandle, result.value.cryptpad.serverUrl);
-  return result.value.cryptpad.serverUrl;
+  let serverUrl = result.value.cryptpad.serverUrl;
+  if (!result.value.cryptpad.serverUrl.startsWith('https://')) {
+    serverUrl = `https://${result.value.cryptpad.serverUrl}`;
+  }
+  CryptpadServerCache.set(connHandle, serverUrl);
+  return serverUrl;
```
NicoTuxx, line 315: If I'm not wrong, if your serverUrl starts with `http://XXX` you will get `https://http//XXX`

NicoTuxx (changes requested):

NicoTuxx (approved):

PR #12781, opened 2026-05-07
Backport39 shorten archiving period env var

touilleMan:


FirelightFlagboy (changes requested):
Missing `backport:` prefix to commit msg.

Prefer to squash the commits to simply the ack

FirelightFlagboy (approved):

PR #12786, opened 2026-05-09
feat(cli): merge recipients and weights in creation of shared-recovery

### LINKED ISSUE #9131: [CLI] Merge the options `recipients` and `weights` of the command `shared-recovery create`
Currently, the options need to be passed sequentially like so:

https://github.com/Scille/parsec-cloud/blob/5bc12a1df3d8001e1636b1859a05c3e1a8288483/cli/tests/integration/shared_recovery/create.rs#L24-L29

That could be improved if we could merge both option in a single one.

The format could be `($Email:$NonZeroUint),*`, we could also allow to skip to `:$NonZeroUint` in wich case the weight for the recipient would be set to 1

BluNfa:
Fixes #9131

Options ``recipients`` and ``weights`` of the command ``shared-recovery create`` are merged into one option ``recipients``, which have the following format:
- either -r email:weight.
- or -r email, then weight is set to 1.

The test using inconsistent recipient and weights values is removed and replaced by a test using explicit weights.

### DIFF
--- a/cli/src/commands/shared_recovery/create.rs
+++ b/cli/src/commands/shared_recovery/create.rs
@@ -9,21 +9,53 @@ use crate::utils::{
     GREEN_CHECKMARK,
 };
 
-// TODO: should provide the recipients and their share count as a single parameter
-//       e.g. `--recipients=foo@example.com=2,bar@example.com=3`
+const ONE: NonZeroU8 = NonZeroU8::new(1).expect("always valid");
+
+#[derive(Debug, Clone, PartialEq)]
+pub struct WeightedEmail {
+    pub email: EmailAddress,
+    pub weight: NonZeroU8,
+}
+
+#[derive(Debug, thiserror::Error, PartialEq, Eq)]
+pub enum WeightedEmailParseError {
+    #[error("Invalid email address")]
+    InvalidEmail,
+    #[error("Invalid weight")]
+    InvalidWeight,
+}
+
+impl std::str::FromStr for WeightedEmail {
+    type Err = WeightedEmailParseError;
+
+    fn from_str(s: &str) -> Result<Self, Self::Err> {
+        let (email, weight) = match s.split_once(':') {
+            Some((email, weight)) => (
+                email,
+                weight
+                    .parse::<NonZeroU8>()
+                    .map_err(|_| WeightedEmailParseError::InvalidWeight)?,
+            ),
+            None => (s, ONE), // default weight is 1
+        };
+
+        let email = email
+            .parse::<EmailAddress>()
+            .map_err(|_| WeightedEmailParseError::InvalidEmail)?;
+
+        Ok(WeightedEmail { email, weight })
+    }
+}
+
 crate::clap_parser_with_shared_opts_builder!(
     #[with = config_dir, device, password_stdin]
     pub struct Args {
-        /// Share recipients, if missing organization's admins will be used instead
+        /// Share recipients, with their weight.
+        /// If missing organization's admins will be used instead.
         /// Author must not be included as recipient.
-        /// User email is expected.
-        #[arg(short, long, num_args = 1..=255)]
-        recipients: Option<Vec<EmailAddress>>,
-        /// Share weights. Requires Share recipient list.
-        /// Must have the same length as recipients.
-        /// Defaults to one per recipient.
-        #[arg(short, long, requires = "recipients", num_args = 1..=255)]
-        weights: Option<Vec<NonZeroU8>>,
+        /// email or email:weight format is expected (weight defaults to one if not provided).
+        #[arg(long, num_args = 1..=255)]
+        recipients: Option<Vec<WeightedEmail>>,
         /// Threshold number of shares required to proceed with recovery.
         #[arg(short, long, requires = "recipients")]
         threshold: Option<NonZeroU8>,
@@ -38,7 +70,6 @@ crate::build_main_with_client!(main, create_shared_recovery);
 pub async fn create_shared_recovery(args: Args, client: &StartedClient) -> anyhow::Result<()> {
     let Args {
         recipients,
-        weights,
         threshold,
         no_confirmation,
         ..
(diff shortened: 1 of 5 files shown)


BluNfa:
While working on this issue, I was confused by the definition of the threshold. Is it a number of shares:
https://github.com/Scille/parsec-cloud/blob/c56e8e4c2add0c03f00811cc1d2c48a03ef0b993/cli/src/commands/shared_recovery/create.rs#L27
or a number of recipients;
https://github.com/Scille/parsec-cloud/blob/c56e8e4c2add0c03f00811cc1d2c48a03ef0b993/cli/src/commands/shared_recovery/create.rs#L94

It's the same when all the weights are 1, but in other cases I think the meanings are different.

cli/src/commands/shared_recovery/create.rs
```diff
@@ -9,21 +9,42 @@ use crate::utils::{
...
+    email: EmailAddress,
+    weight: NonZeroU8,
+}
+
+impl FromStr for WeightedEmail {
+    type Err = String;
+
+    fn from_str(s: &str) -> Result<Self, Self::Err> {
+        let (email, weight) = match s.split_once(':') {
+            Some((name, number)) => (name, number),
```
FirelightFlagboy: **nitpick:** Would have put the number parsing in the match arm so we could return directly `1` for the default case

cli/src/commands/shared_recovery/create.rs
```diff
@@ -9,21 +9,42 @@ use crate::utils::{
...
-// TODO: should provide the recipients and their share count as a single parameter
-//       e.g. `--recipients=foo@example.com=2,bar@example.com=3`
+#[derive(Debug, Clone)]
+struct WeightedEmail {
+    email: EmailAddress,
+    weight: NonZeroU8,
+}
+
+impl FromStr for WeightedEmail {
+    type Err = String;
```
FirelightFlagboy: **issue:** We do not want an error string

You could either:

- Use `anyhow`
- Use a error enum with `thiserror` (this approach would allow for simpler error testing for the parser)

newsfragments/9131.feature.rst
```diff
@@ -0,0 +1 @@
+[CLI] Merge options ``recipients`` and ``weights`` of the command ``shared-recovery create``.
```
FirelightFlagboy: **nitpick:** 
```suggestion
[CLI] Merge options ``--recipients`` and ``--weights`` of the command ``shared-recovery create`` into ``--recipients``.
```

FirelightFlagboy (changes requested):

FirelightFlagboy:
> I was confused by the definition of the threshold. Is it a number of shares [...] or a number of recipients

The **threshold** is the number of shares required to rebuild the shared key.

When the weight is 1 for everyone we can make the shortcut saying that it correspond to the number of person to contact.

When the weight is not 1 for everyone it mean that your trust some people more (the ones that have more share) than the other (the ones that have less share or does not have any).

It allows you to do some combination depending on your strategy to recover that device:

Alice trust its manager Bob more that its co-workers Mallory, Dave & Clara.
Alice crate a shared recovery device shared amongst all of the people above with a threshold of **3**.
But since Alice trust Bob more the other it decide to give it a weight of **2**.

It mean that when it's time to recover Alice's device, Alice will need either:

- Asking Bob and another co-worker (2 people contacted) to retrieve 3 shares (2+1)
- Or asking the 3 co-workers (3 people contacted) to retrieve 3 shares (1+1+1)

cli/src/commands/shared_recovery/create.rs
```diff
@@ -48,44 +68,37 @@ pub async fn create_shared_recovery(args: Args, client: &StartedClient) -> anyho
...
+
         recipients
             .iter()
-            .map(|human| *recipient_info.get(human).expect("human should be here"))
+            .map(|recipient| {
+                (
+                    users_ids
+                        .get(&recipient.email)
+                        .copied()
+                        .expect("A user is missing"),
```
FirelightFlagboy: **polish:** Would be better to return a proper error over panicing to provide a better experience to the user

cli/src/commands/shared_recovery/create.rs
```diff
@@ -48,44 +68,37 @@ pub async fn create_shared_recovery(args: Args, client: &StartedClient) -> anyho
...
             .collect()
     } else {
-        let admins: Vec<_> = users
+        let admins: HashMap<_, _> = users
             .iter()
             .filter(|info| {
                 info.current_profile == UserProfile::Admin && info.id != client.user_id()
             })
-            .map(|info| info.id)
+            .map(|info| (info.id, NonZeroU8::new(1).expect("always valid")))
```
FirelightFlagboy: **note:** `NonZeroU8::new` & `Option::expect` are `const`

https://doc.rust-lang.org/std/num/type.NonZeroU8.html#compile-time-creation

FirelightFlagboy (commented):
Can you add test for `WeightedEmail::from_str`?

cli/src/commands/shared_recovery/create.rs
BluNfa: I added an error enum with ``InvalidEmail`` and ``InvalidWeight``.

cli/src/commands/shared_recovery/create.rs
BluNfa: Replaced by ``const ONE``

BluNfa:
I added some tests for ``WeightedEmail::from_str``, inspired by the doc of [FromStr](https://doc.rust-lang.org/std/str/trait.FromStr.html). Tbh I'm not sure if this sufficient, and if the tests should be placed in ``cli/tests/integration/shared_recovery/create.rs``. From what I have seen, custom struct implementations and their tests are often located in libparsec, I don't know if it should apply for this one.

BluNfa:
> > I was confused by the definition of the threshold. Is it a number of shares [...] or a number of recipients
> 
> The **threshold** is the number of shares required to rebuild the shared key.
> 
> When the weight is 1 for everyone we can make the shortcut saying that it correspond to the number of person to contact.
> 
> When the weight is not 1 for everyone it mean that your trust some people more (the ones that have more share) than the other (the ones that have less share or does not have any).
> 
> It allows you to do some combination depending on your strategy to recover that device:
> 
> Alice trust its manager Bob more that its co-workers Mallory, Dave & Clara. Alice crate a shared recovery device shared amongst all of the people above with a threshold of **3**. But since Alice trust Bob more the other it decide to give it a weight of **2**.
> 
> It mean that when it's time to recover Alice's device, Alice will need either:
> 
> * Asking Bob and another co-worker (2 people contacted) to retrieve 3 shares (2+1)
> * Or asking the 3 co-workers (3 people contacted) to retrieve 3 shares (1+1+1)

Thank you for the very clear explanation! I find the following description https://github.com/Scille/parsec-cloud/blob/c56e8e4c2add0c03f00811cc1d2c48a03ef0b993/cli/src/commands/shared_recovery/create.rs#L94 a bit misleading, as in your example the threshold is 3, but the minimum number of recipients is 2 (Bob and another co-worker).
I would rather have said "minimum number of shares" or "minimum weight"

cli/src/commands/shared_recovery/create.rs
```diff
@@ -48,44 +79,38 @@ pub async fn create_shared_recovery(args: Args, client: &StartedClient) -> anyho
...
+                (
+                    users_ids
+                        .get(&recipient.email)
+                        .copied()
+                        .ok_or_else(|| "A user is missing")
+                        .unwrap(),
+                    recipient.weight,
+                )
+            })
             .collect()
```
FirelightFlagboy: **nitpick:** That block still panic instead of returning the error value


```suggestion
                    users_ids
                        .get(&recipient.email)
                        .copied()
                        .ok_or_else(|| anyhow::anyhow!("A user with email {} not found", recipient.email)
                        .map(|id| (id, recipient.weight))
            })
.collect::<Result<Vec<_>, _>>()?
```

cli/tests/integration/shared_recovery/create.rs
```diff
@@ -1,10 +1,41 @@
...
     testenv_utils::{TestOrganization, DEFAULT_DEVICE_PASSWORD},
 };
 
+use parsec_cli::commands::shared_recovery::create::WeightedEmail;
+use parsec_cli::commands::shared_recovery::create::WeightedEmailParseError;
+
+#[rstest::rstest]
+#[tokio::test]
+
+async fn weighted_email_from_str() {
```
FirelightFlagboy, line 16: **chore:** 
```suggestion
#[tokio::test]
async fn weighted_email_from_str() {
```

FirelightFlagboy, line 16: **todo:** Need a test when the weight is not provided

FirelightFlagboy:
> Thank you for the very clear explanation! I find the following description
> 
> https://github.com/Scille/parsec-cloud/blob/c56e8e4c2add0c03f00811cc1d2c48a03ef0b993/cli/src/commands/shared_recovery/create.rs#L94
> a bit misleading, as in your example the threshold is 3, but the minimum number of recipients is 2 (Bob and another co-worker).
> I would rather have said "minimum number of shares" or "minimum weight"

Yes, it's misleading. You can update the text with "minimum number of shares" if you want

FirelightFlagboy:
Missing https://github.com/Scille/parsec-cloud/pull/12786#discussion_r3257204201

But overall we are good

BluNfa:
> Missing [#12786 (comment)](https://github.com/Scille/parsec-cloud/pull/12786#discussion_r3257204201)
> 
> But overall we are good

Ah my bad, I read too quickly and fought it was about testing the failure case with ``:`` but no weight.
Test is added 😊

FirelightFlagboy (approved):

FirelightFlagboy:
Thanks for the PR :+1:

PR #12794, opened 2026-05-11
[MS] Added tests for sync and timestamp on archived and trashed workspaces

Ironicbay:

PR #12795, opened 2026-05-11
[MS] Handle WorkspaceUpdate event on archived and trashed pages

Ironicbay:


Max-7 (changes requested):
Looks good, but you need to add at least a test for each, similar to what you've done in #12794:
- Bob on the Trash page, Alice trashes a workspace, Bob sees the workspace appear
- Bob on the Archived page, Alice archives a workspace, Bob sees the workspace appear

Max-7 (approved):

PR #12822, opened 2026-05-19
refactor(pki): Enable auto-resync for win cert store

FirelightFlagboy:
Configure auto-resync flag on a Windows certificate store.

That flag will make the store to resync at the start of enumeration or find if needed. If the feature is not supported by the store, we just skip the configuration.

### DIFF
--- a/libparsec/crates/platform_pki/src/windows/pki_system.rs
+++ b/libparsec/crates/platform_pki/src/windows/pki_system.rs
@@ -43,12 +43,16 @@ impl PlatformPkiSystem {
         _config_dir: &std::path::Path,
         _scws_config: Option<PkiScwsConfig>,
     ) -> Result<Self, PkiSystemInitError> {
-        CertStore::open_current_user(LEAF_CERTIFICATE_STORE)
+        let store = CertStore::open_current_user(LEAF_CERTIFICATE_STORE)
             .map(|store| Self {
                 my_cert_store: store,
             })
             .context("Cannot open windows cert store")
-            .map_err(PkiSystemInitError::Internal)
+            .map_err(PkiSystemInitError::Internal)?;
+        configure_auto_resync_store(&store.my_cert_store)
+            .context("Failed to configure auto resync for certificate store")
+            .map_err(PkiSystemInitError::Internal)?;
+        Ok(store)
     }
 
     pub async fn open_certificate(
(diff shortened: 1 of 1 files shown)


libparsec/crates/platform_pki/src/windows/pki_system.rs
```diff
@@ -313,3 +321,36 @@ fn ask_user_to_select_certificate(store: &CertStore) -> Option<CertContext> {
         }
     }
 }
+
+/// Try our best to configure auto-resync for a given certificate store.
+///
+/// If case auto-resync is not supported, we just ignore the configuration step.
```
touilleMan:
```suggestion
/// If auto-resync is not supported, we just ignore the configuration step.
```

libparsec/crates/platform_pki/src/windows/pki_system.rs
```diff
@@ -313,3 +321,36 @@ fn ask_user_to_select_certificate(store: &CertStore) -> Option<CertContext> {
...
+///
+/// ## Errors
+///
+/// Will error if the configuration fail and it's not about the driver that does not support that feature.
```
touilleMan:
```suggestion
/// Will error if auto-resync is supported, but the configuration fails nevertheless.
```

libparsec/crates/platform_pki/src/windows/pki_system.rs
```diff
@@ -313,3 +321,36 @@ fn ask_user_to_select_certificate(store: &CertStore) -> Option<CertContext> {
         }
     }
 }
+
+/// Try our best to configure auto-resync for a given certificate store.
```
touilleMan, line 321: Should explain why this is needed:

```
/// Whithout auto-resync, a certificate created after opening the store will be ignored.
/// This is typically an issue when a user stats Parsec first (hence opening the certificate store)
/// and only then plug his personal Smartcard to the machine.
```

touilleMan: Also can you add a word about when this feature is expected to be non-supported ?
(ideally linking to a documentation, or a sarcastic comment noticing the lack of information about when the feature is supported :trollface: )

touilleMan (approved):

libparsec/crates/platform_pki/src/windows/pki_system.rs
FirelightFlagboy: done

PR #12836, opened 2026-05-20
[MS] Removed @ionic/cli

Max-7:


mmmarcos (approved):
nit: you can also remove this comment at L231 and enable dependency type (uncomment L232)
https://github.com/Scille/parsec-cloud/blob/5b2442107a9807b412b350a371292cf09ee25f1d/.github/dependabot.yml#L231-L232

FirelightFlagboy (approved):

PR #12839, opened 2026-05-21
fix(pki): Convert origin into `ParsecAddr` before init

FirelightFlagboy:
Convert `window.origin` into `ParsecAddr` before calling `libparsec.pkiInitForScws` as it expect a `ParsecAddr` in the bindings.


Max-7 (approved):
Good from the GUI POV, but can't tell anything about the code.

AureliaDolo (approved):

PR #12852, opened 2026-05-22
docs: Add Parsec Server config for OpenBao

mmmarcos:
The recently added section [Single Sign-On with OpenBao](https://docs.parsec.cloud/en/latest/hosting/openbao/index.html) explained how to configure OpenBao server BUT did not mention what it needs to be done on the Parsec Server side.

This PR adds that.

Only ProConnect is mentioned (even though there is also Hexagone) as I think is sufficient for the moment.

### DIFF
--- a/docs/hosting/openbao/index.rst
+++ b/docs/hosting/openbao/index.rst
@@ -6,17 +6,20 @@
 Single Sign-On with OpenBao
 ===========================
 
-This section explains how to enable :abbr:`SSO (Single Sign-On)` in Parsec via `OpenBao`_.
+This section explains how to enable :abbr:`SSO (Single Sign-On)` in Parsec via `OpenBao`_,
+an open-source secrets manager from the Linux Foundation's `OpenSSF`_ community.
 
+.. _OpenBao: https://openbao.org
+.. _OpenSSF: https://openssf.org
 
 Overview
 ========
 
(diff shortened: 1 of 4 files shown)


AureliaDolo (approved):

docs/hosting/openbao/index.rst
```diff
@@ -179,4 +191,69 @@ Verify the setup
...
+
+And then restart the server.
+
+Parsec Server on Linux
+----------------------
+
+If you deployed Parsec Server directly on Linux, edit your run script to add
+``parsec-openbao.env``:
+
+.. admonition:: parsec-server.docker.yaml
```
FirelightFlagboy: **typo:**

FirelightFlagboy (commented):
I don't like how it's done (need extra manual steps to have a working test infra), but good for now

docs/hosting/openbao/index.rst
mmmarcos: on which word? :thinking:

FirelightFlagboy: filename, GitHub is missleading here, my comment is only for line 237

mmmarcos: Fixed

FirelightFlagboy (approved):

PR #12858, opened 2026-05-26
Check if a filename is confined in the GUI

### LINKED ISSUE #11059: Confined files are displayed in the import dialog when using `Import folder`
### Parsec version

3.5.0-a7

- [ ] Any
- [x] macOS
- [ ] Linux
- [ ] Windows
- [ ] Web

### Bug description

When importing a folder, here on MacOS with `.DS_Store` files, these files are imported regardless of if their pattern match confined files patterns.

### Reproduction steps

```bash
1. Import a folder containing files matching confined patterns
2. Confined files appear on the operations manager
```

### Relevant output:

_No response_

Max-7:
Closes #11059

### DIFF
--- a/client/src/locales/fr-FR.json
+++ b/client/src/locales/fr-FR.json
@@ -2310,7 +2310,10 @@
             "fileTooLong": "Le nom du fichier est trop long, la limite est de {limit} caractères.",
             "folderTooLong": "Le nom du dossier est trop long, la limite est de {limit} caractères.",
             "fileAlreadyExists": "Un fichier du même nom existe déjà.",
-            "folderAlreadyExists": "Un dossier du même nom existe déjà."
+            "folderAlreadyExists": "Un dossier du même nom existe déjà.",
+            "isConfined": "Ce nom est réservé.",
+            "unauthorizedExtension": "L'extension de fichier n'est pas autorisée. L'extension doit être '{allowedExtensions}'. | L'extension de fichier n'est pas autorisée. L'extension doit être parmi '{allowedExtensions}'.",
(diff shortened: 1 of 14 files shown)


libparsec/src/client.rs
```diff
@@ -759,3 +759,10 @@ pub async fn client_pki_get_addr(
     let client = borrow_client(client)?;
     Ok(client.pki_get_addr().await)
 }
+
+pub fn is_filename_confined(client: Handle, filename: &str) -> bool {
+    let Ok(client) = borrow_client(client) else {
+        return false;
+    };
+    client.config().prevent_sync_pattern.is_match(filename)
```
FirelightFlagboy: **issue:** This check is insufficient

For the function name it's enough does the filename correspond to a confined filename.
But that function may be abused to determine if a file is confined which for that check is not enough (to verify such case, you need to perform a recursive check on each parents)

FirelightFlagboy (changes requested):
Following last review and issue described in #11059, it cannot be merge as is

FirelightFlagboy (approved):

client/src/locales/fr-FR.json
```diff
@@ -2310,7 +2310,8 @@
             "fileTooLong": "Le nom du fichier est trop long, la limite est de {limit} caractères.",
             "folderTooLong": "Le nom du dossier est trop long, la limite est de {limit} caractères.",
             "fileAlreadyExists": "Un fichier du même nom existe déjà.",
-            "folderAlreadyExists": "Un dossier du même nom existe déjà."
+            "folderAlreadyExists": "Un dossier du même nom existe déjà.",
+            "isConfined": "This name is restricted."
```
Ironicbay: Needs translation

Ironicbay (approved):
It is still possible to create files and folders in the GUI with confined names, the operation is not prevented. Not sure if this is blocking or if it would be treated in another PR.

What is there is good to me 👍

Ironicbay (approved):

PR #12860, opened 2026-05-26
Fix error on Edge by forcing create:true when listing files

### LINKED ISSUE #11736: Error when arriving on page on Edge
### Parsec version

3.6.0

- [ ] Any
- [ ] macOS
- [ ] Linux
- [ ] Windows
- [x] Web

### Bug description

Image

Caused by `listAvailableDevices` returning an error with `DOMException` instead of an empty list. The problem is inside libparsec, not the GUI, since an error is returned (it's not caused by an exception).

### How should the application behave

_No response_

### Reproduction steps

```bash
1.
2.
3.
...
```

### Relevant output:

_No response_

Max-7:
Closes #11736

You may need to exec this code in Edge console to fully clear the data. For some reason, I haven't been able to find any interface option to do that properly. Or test using a private window.

```
const root = await navigator.storage.getDirectory();

for await (const [name, handle] of root.entries()) {
  await root.removeEntry(name, { recursive: true });
}
```

### DIFF
--- a/libparsec/crates/platform_filesystem/src/web/common/internal.rs
+++ b/libparsec/crates/platform_filesystem/src/web/common/internal.rs
@@ -39,7 +39,13 @@ impl Storage {
             "Listing file entries in {} with extension {extension}",
             dir.display()
         );
-        let dir = match self.root_dir.get_directory_from_path(dir, None).await {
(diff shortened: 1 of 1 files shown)


FirelightFlagboy (approved):

NicoTuxx (approved):
work on Edge

PR #12868, opened 2026-05-28
Include VC Redistributable when building the Windows installer

### LINKED ISSUE #9341: [🐛 | Bug]: Missing VCRUNTIME140.dll
### Parsec version tested on:

3.2.3

### Platforms tested on:

Windows

### Bug description:

See : https://scille.slack.com/archives/C20C5ECJH/p1734706587170129

On fresh Windows installation, Microsoft Visual C++ Redistributable is not present but is necessary to launch the application.
It can download from https://aka.ms/vs/17/release/vc_redist.x64.exe.

### Relevant output:

dumpbin /dependents <parsec install dir>/resources/app.asar.unpacked/build/src/libparsec.node
```
File Type: DLL

  Image has the following dependencies:

    bcryptprimitives.dll
    api-ms-win-core-synch-l1-2-0.dll
    kernel32.dll
    advapi32.dll
    shell32.dll
    ole32.dll
    user32.dll
    ws2_32.dll
    crypt32.dll
    psapi.dll
    secur32.dll
    ntdll.dll
    bcrypt.dll
    VCRUNTIME140.dll
    api-ms-win-crt-string-l1-1-0.dll
    api-ms-win-crt-heap-l1-1-0.dll
    api-ms-win-crt-utility-l1-1-0.dll
    api-ms-win-crt-time-l1-1-0.dll
    api-ms-win-crt-runtime-l1-1-0.dll
    api-ms-win-crt-math-l1-1-0.dll
    api-ms-win-crt-filesystem-l1-1-0.dll
    api-ms-win-crt-stdio-l1-1-0.dll
    api-ms-win-crt-convert-l1-1-0.dll
    api-ms-win-crt-environment-l1-1-0.dll

  Image has the following delay load dependencies:

    winfsp-x64.dll
```

### LINKED ISSUE #12775: Add VC Redistributable to Windows installer
Launching Parsec can fail with a critical error saying that it can't find the module, with a path to libparsec.node. This is due to missing runtime dependencies, either WinFSP or the VCRuntime.

WinFSP is unlikely since it's bundled with our installer, but we should probably also bundle the VC Redistributable. Most Windows machine already have those installed, but on a fresh installation, they can be missing.

I don't think Microsoft has fixed URLs for each version, only a permalink to the latest version. We could still download it during the build process, but we would have to skip the sha1 check. Or we could self-host the binary.

Max-7:
Closes #12775 
Closes #9341

### DIFF
--- a/client/electron/assets/installer.nsh
+++ b/client/electron/assets/installer.nsh
@@ -1,5 +1,6 @@
 !define WINFSP_VERSION "2.1.25156"
 !define WINFSP_FILE "winfsp-${WINFSP_VERSION}.msi"
+!define VC_REDIST_FILE "vc_redist.x64.exe"
 
 !macro installWinFSP
     File /oname=$PLUGINSDIR\${WINFSP_FILE} ${PROJECT_DIR}\build\${WINFSP_FILE}
@@ -25,6 +26,12 @@
     ${EndIf}
 !macroend
 
+!macro installVCRedist
+    File /oname=$PLUGINSDIR\${VC_REDIST_FILE} ${PROJECT_DIR}\build\${VC_REDIST_FILE}
+    ExecWait '"$PLUGINSDIR\${VC_REDIST_FILE}" /install /passive /norestart'
+!macroend
+
 !macro customInstall
     !insertmacro mayInstallWinFSP
+    !insertmacro installVCRedist
 !macroend
--- a/client/electron/package.cjs
+++ b/client/electron/package.cjs
@@ -152,7 +152,15 @@ const options = {
     buildResources: 'assets',
   },
 
-  files: ['assets/**/*', '!assets/installer.nsh', 'build/**/*', '!build/**/*.js.map', '!build/**/*.msi', 'app/**/*'],
+  files: [
+    'assets/**/*',
+    '!assets/installer.nsh',
+    'build/**/*',
+    '!build/**/*.js.map',
+    '!build/**/*.msi',
+    '!build/vc_redist.x64.exe',
+    'app/**/*',
+  ],
 
   publish: publishConfig,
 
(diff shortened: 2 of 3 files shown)


client/electron/assets/installer.nsh
```diff
@@ -25,6 +26,12 @@
...
 !macroend
 
+!macro installVCRedist
+    File /oname=$PLUGINSDIR\${VC_REDIST_FILE} ${PROJECT_DIR}\build\${VC_REDIST_FILE}
+    ExecWait '"$PLUGINSDIR\${VC_REDIST_FILE}" /install /passive /norestart'
+!macroend
+
 !macro customInstall
     !insertmacro mayInstallWinFSP
+    !insertmacro installVCRedist
```
FirelightFlagboy, line 36: **issue:** This always install VC, it should check if it's not already installed

FirelightFlagboy (changes requested):
It's a choice to include the VC lib in the installer, some other installers provide a link to download the VC lib instead 🤷

client/electron/assets/installer.nsh
Max-7: Honestly, the check doesn't even work properly with WinFSP, I always get a pop-up, so why bother?
You get a pop-up, you click next, if it needs to be installed it's installed. One option could also be to run it as `/quiet` instead of `/passive`, which doesn't even display a pop-up (appart maybe from Windows UAC). I put `/passive` because it's what we're doing with WinFSP and it seems a little more transparent.

Max-7:
> It's a choice to include the VC lib in the installer, some other installers provide a link to download the VC lib instead 🤷

Where do you put the link? In the installer somewhere? I'm afraid the user won't read and skip the installation. In my opinion, it's just better to bundle that in.

PR #12871, opened 2026-05-29
[MS] Added numbering in file name when creating multiple with default name

### LINKED ISSUE #12782: Unable to create a new document if one already exist with "Nouveau document" name
https://github.com/user-attachments/assets/be53ed07-5fbe-4e6e-aa4c-6a93d951f560

Ironicbay:
Closes https://github.com/Scille/parsec-cloud/issues/12782

### DIFF
--- a/client/src/common/file.ts
+++ b/client/src/common/file.ts
@@ -1,6 +1,6 @@
 // Parsec Cloud (https://parsec.cloud) Copyright (c) BUSL-1.1 2016-present Scille SAS
 
-import { Path } from '@/parsec';
+import { EntryName, entryStat, FsPath, Path, WorkspaceHandle } from '@/parsec';
 import { File, Translatable } from 'megashark-lib';
 
 // Shorten a file name by adding ellipsis in the middle if it is above a certain length.
@@ -84,6 +84,44 @@ export function formatFileSize(bytesize: number | bigint): Translatable {
     [Math.pow(1024, 4), 'common.filesize.terabytes'],
   ];
   return size(Number(bytesize), SYSTEM);
+}
+
+export async function findAvailableFileName({
+  handle,
+  fileName,
+  path,
+  maxIterations,
+}: {
+  handle: WorkspaceHandle;
+  fileName: EntryName;
+  path: FsPath;
+  maxIterations?: number;
+}): Promise<EntryName | undefined> {
+  const name = Path.filenameWithoutExtension(fileName);
+  const ext = Path.getFileExtension(fileName);
+  let numberedFileName = name;
+
+  let iteration = 1;
+  const loopLimit = maxIterations ?? 10;
+
+  while (iteration < loopLimit) {
+    const tmpName = ext.length > 0 ? `${numberedFileName}.${ext}` : numberedFileName;
+    const distPath = await Path.join(path, tmpName);
+    const stat = await entryStat(handle, distPath);
+    if (!stat.ok) {
+      break;
+    }
+    iteration += 1;
+    if (iteration === loopLimit) {
+      return undefined;
+    }
+    numberedFileName = `${name} (${iteration})`;
+  }
+
+  if (ext.length > 0) {
+    return `${numberedFileName}.${ext}`;
+  }
+  return numberedFileName;
 }
 
 /* File icons */
--- a/client/src/views/files/FoldersPage.vue
+++ b/client/src/views/files/FoldersPage.vue
@@ -272,6 +272,7 @@ import {
 
 import DocumentNew from '@/assets/images/add-document.svg?raw';
 import ListFolderError from '@/assets/images/list-folder-error.svg?raw';
+import { findAvailableFileName } from '@/common/file';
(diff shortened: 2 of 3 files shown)


client/src/services/fileOperation/operations.ts
```diff
@@ -20,11 +20,11 @@ import {
   WorkspaceHandle,
   WorkspaceHistory,
   WorkspaceHistoryEntryStatFile,
-  WorkspaceMoveEntryErrorTag,
   WorkspaceRemoveEntryErrorTag,
   writeFile,
 } from '@/parsec';
 import { stringifyError } from '@/parsec/utils';
+import { numberDuplicateName } from '@/services/fileOperation';
```
Max-7: Import from `@/services/fileOperation/utiles.ts` instead. Otherwise, it may create a circular import where operation.ts import from index.ts which itself import from operation. I don't  think it's the case here but it could avert problems later on.

client/src/services/fileOperation/utils.ts
```diff
@@ -0,0 +1,33 @@
+// Parsec Cloud (https://parsec.cloud) Copyright (c) BUSL-1.1 2016-present Scille SAS
+
+import { EntryName, entryStat, FsPath, Path, WorkspaceHandle } from '@/parsec';
+
+export async function numberDuplicateName(
```
Max-7: Since it's quite a lot of arguments, it may benefits from having an anonymous object:

```typescript
async function numberDuplicateName({
  handle, fileName, ext, path, maxIterations, withExtension
}: {
  handle: WorkspaceHandle, fileName: EntryName, ext: string, path: FsPath, maxIterations: number, withExtension?: boolean
}) {
}
```

which can be called with `numberDuplicateName({ handle: 4, fileName: 'myfile.txt', withExtension: true})`. It makes be a bit easier to read the function call (you don't have to wonder what the `true` at the end does).

client/src/services/fileOperation/operations.ts
```diff
@@ -224,37 +224,28 @@ export class OperationTransaction {
...
-    }
-    if (moveResult.error.tag === WorkspaceMoveEntryErrorTag.DestinationExists && counter < MAX_COUNTER) {
-      counter += 1;
-    } else {
-      return undefined;
-    }
+  if (!newName) {
+    return undefined;
   }
+  const newDestination = await Path.join(parentDir, newName);
```
Max-7: Your replacement code does not do the same thing.

In the case of file operation, the move is the important part because it is atomic: no operation can be done in the meantime. It's not really about finding a name, it's about avoiding collisions with existing file names.
It is a very weird edge-case but we're talking about file manipulation, it should be as foolproof as possible.

Max-7 (changes requested):

client/src/services/fileOperation/utils.ts
Max-7: Not a fan of the name. Wouldn't something like `findAvailableFileName` be more appropriate?

client/src/services/fileOperation/utils.ts
```diff
@@ -0,0 +1,40 @@
+// Parsec Cloud (https://parsec.cloud) Copyright (c) BUSL-1.1 2016-present Scille SAS
+
+import { EntryName, entryStat, FsPath, Path, WorkspaceHandle } from '@/parsec';
+
+export async function findAvailableFileName({
```
Max-7: Move it to `components/file/utils.ts` or `common/file.ts`, we already have too many files for files. And it's not used by fileOperation anyway.

client/src/services/fileOperation/utils.ts
```diff
@@ -0,0 +1,40 @@
...
+  path,
+  maxIterations,
+  withExtension,
+}: {
+  handle: WorkspaceHandle;
+  fileName: EntryName;
+  ext: string;
+  path: FsPath;
+  maxIterations: number;
+  withExtension?: boolean;
```
Max-7: Having both this argument and an ext argument makes the function a bit weird to use. I think you can remove `withExtension`.

I also find passing an extension but not having it in the return value weird. I understand that you did that to make it easier to set the selection range later, but it makes the function tailored to that specific case, which makes it less useful.

For a function like that, I would expect to pass it a file name with the extension, and for the function to return a file name with the added number if needed and the extension, so

`findAvailableFile({handle: 3, fileName: 'MyFile.txt', path: '/'})` would result in `MyFile (2).txt`

client/src/services/fileOperation/utils.ts
```diff
@@ -0,0 +1,40 @@
...
+  ext,
+  path,
+  maxIterations,
+  withExtension,
+}: {
+  handle: WorkspaceHandle;
+  fileName: EntryName;
+  ext: string;
+  path: FsPath;
+  maxIterations: number;
```
Max-7: I think we can make maxIterations optional.

Max-7 (changes requested):

Max-7 (approved):

PR #12912, opened 2026-06-08
fix(python): Remove unused `pbr` deps

FirelightFlagboy:


mmmarcos:
Should we remove this code?
https://github.com/Scille/parsec-cloud/blob/c847bcb7699c98f19c2d941f1784c33a697712af/server/parsec/components/swift_blockstore.py#L9-L12

FirelightFlagboy:
> Should we remove this code?
> https://github.com/Scille/parsec-cloud/blob/c847bcb7699c98f19c2d941f1784c33a697712af/server/parsec/components/swift_blockstore.py#L9-L12

Good catch, I didn't find looked at that code.

But it seems to show that we are lacking tests in the block store territory.

PR #12923, opened 2026-06-11
[MS] Update log viewing and reporting

Max-7:


client/src/locales/fr-FR.json
```diff
@@ -2172,7 +2172,15 @@
...
     "LogDisplayModal": {
         "title": "Logs",
-        "noLog": "Aucun log disponible."
+        "noLog": "Aucun log disponible.",
+        "download": "Télécharger",
+        "copy": "Copier",
+        "close": "Fermer",
+        "copySuccess": "Copié!",
+        "copyFailed": "Échec",
+        "downloadSuccess": "Téléchargés!",
```
mmmarcos: singular?
```suggestion
        "downloadSuccess": "Téléchargé!",
```

client/src/components/misc/LogDisplay.vue
```diff
@@ -41,69 +62,124 @@
...
+async function copyLogs(): Promise<boolean> {
+  return await Clipboard.writeText(logs.value);
+}
+
+async function downloadLogs(): Promise<boolean> {
+  try {
+    const blob = new Blob([logs.value], { type: 'text/plain' });
+    const url = window.URL.createObjectURL(blob);
+    downloadLinkRef.value?.setAttribute('href', url);
+    downloadLinkRef.value?.setAttribute('download', I18n.translate('LogDisplayModal.downloadFileName'));
```
mmmarcos: Do we really ever need to translate the file name?

I suggest:
- using the `.log` extension which is supported by most (all?) editors and can be opened with notepad on Windows.
- adding a timestamp to the file name to avoid the `Copy of file (1).log` syndrome

Example: `2026-06-30_11-09-00_parsec.log`

mmmarcos (approved):
I didn't review the code in detail, but it works fine 👍

Just a couple of suggestions on the log file name that shouldn't be too costly to implement.

PR #12926, opened 2026-06-11
Fix `libparsec_platform_pki` on web

touilleMan:


Cargo.lock
```diff
@@ -4322,7 +4322,7 @@ dependencies = [
 [[package]]
 name = "scwsapi-sys"
 version = "0.4.0"
```
FirelightFlagboy: **nitpick:** Forget to bump the patch version

Cargo.lock
```diff
@@ -4305,7 +4305,7 @@ checksum = "94143f37725109f92c262ed2cf5e59bce7498c01bcc1502d7b9afe439a4e9f49"
 [[package]]
 name = "scwsapi"
 version = "0.8.0"
```
FirelightFlagboy: **nitpick:** Forget to bump the patch version

FirelightFlagboy (approved):
Nitpick are not important, you can merge as is

Cargo.lock
touilleMan: done.

Cargo.lock
touilleMan: done.

PR #12931, opened 2026-06-12
fix: add .gitattributes to normalize line endings to LF for all text files

NicoTuxx:


AureliaDolo (approved):

FirelightFlagboy (approved):

.gitattributes
```diff
@@ -0,0 +1,2 @@
+# Normalize line endings to LF for all text files
+* text=auto eol=lf
```
mmmarcos, line 2: Not good: only scripts intended for Linux should have LF. Windows script should keep CRLF.

mmmarcos (changes requested):

NicoTuxx:
was a bad idea

PR #12944, opened 2026-06-15
[MS] Bundle .css and .js at build time

Max-7:
Bundle all the CSS and JS into one big file, avoiding a lot of HTTP requests when loading the app (~10 requests instead of ~90), add minification, and adds compression with brotli.


NicoTuxx (dismissed):
lgtm

NicoTuxx (approved):
LGTM

Working properly!

PR #12950, opened 2026-06-16
Fix mixed-line-ending pre-commit rule

touilleMan:
Without this option it seems this check doesn't reject `\r\n` (e.g. see fabc17c2db on `releases/3.9` that hasn't been rejected)

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -12,6 +12,7 @@ repos:
     rev: v6.0.0
     hooks:
       - id: mixed-line-ending
+        args: [--fix=lf]
         alias: eol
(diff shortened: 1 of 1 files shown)


FirelightFlagboy (changes requested):
I don't mind adding that extra config, but It's already done during `CI Q&A` during the step "Patch pre-commit for line-ending".

If it wasn't already included in the base configuration it's likely because there is a reason that would impact Windows devs.

Anyway, the Windows devs are expected to configure git with [`autocrlf`](https://git-scm.com/book/en/v2/Customizing-Git-Git-Configuration#_core_autocrlf)

PR #12958, opened 2026-06-16
Fix PKI certificate usage for encryption

touilleMan:


Max-7 (approved):
GUI side is good.

mmmarcos (approved):
libparsec changes looks ok

libparsec/crates/platform_pki/src/x509/extensions/mod.rs
```diff
@@ -32,7 +32,8 @@ impl Default for Extensions {
 impl Extensions {
     /// if key usage is not specified, returns true
     pub fn can_encrypt(&self) -> bool {
-        self.key_usage.0.is_empty() || self.key_usage.data_encipherment()
+        // see https://datatracker.ietf.org/doc/html/rfc5280#section-4.2.1.3
+        self.key_usage.0.is_empty() || self.key_usage.key_encipherment()
```
FirelightFlagboy: **issue:** Should only add `key_encipherment` check
```suggestion
        // see https://datatracker.ietf.org/doc/html/rfc5280#section-4.2.1.3
        self.key_usage.0.is_empty() || self.key_usage.data_encipherment() || self.key_usage.key_encipherment()
```

FirelightFlagboy (changes requested):

libparsec/crates/platform_pki/src/x509/extensions/mod.rs
FirelightFlagboy: We use the private key to encrypt the symmetric key, so `key_encipherment` was missing.
But `data_encipherment` key usage is still valid, as we can consider the key to be raw data.

FirelightFlagboy (approved):

PR #12959, opened 2026-06-17
[MS] Multiple fixes for PKI

Max-7:


client/src/views/home/UserJoinOrganizationModal.vue
```diff
@@ -163,6 +163,7 @@
             ref="authChoice"
             :server-config="serverConfig"
             :server-addr="serverAddr"
+            @auth-reset="fieldsUpdated = true"
```
Ironicbay, line 166: Is this variable doing anything? Looking at the file it's just instanciated as false and set to `false` in `canGoForward`.

client/src/locales/en-US.json
```diff
@@ -347,6 +347,7 @@
             "requestCancelError": "Could not cancel the request.",
             "requestDeleted": "Your request to join the organization has been deleted.",
             "joinButton": "Add to my organizations",
+            "confirmMissingCertificate": "Cannot access the certificate used to make this request.",
```
```suggestion
            "confirmMissingCertificate": "Failed to access the certificate used to make this request.",
```

client/src/views/home/DeviceJoinOrganizationModal.vue
```diff
@@ -223,6 +224,7 @@ const authChoiceRef = useTemplateRef<InstanceType<typeof ChooseAuthentication>>(
 const cancelled = ref(false);
 const serverConfig = ref<ServerConfig | undefined>(undefined);
 const serverAddr = ref<ParsecAddr>('');
+const fieldsUpdated = ref(false);
```
Ironicbay, line 227: Same question as before, what are these variables set for?

client/src/main.ts
```diff
@@ -44,6 +44,8 @@ import LongPathsSupportModal from '@/views/about/LongPathsSupportModal.vue';
 import IncompatibleEnvironmentModal from '@/views/home/IncompatibleEnvironmentModal.vue';
 import { Answer, I18n, Locale, MegaSharkPlugin, Obj, StripeConfig, ThemeManager, askQuestion } from 'megashark-lib';
 
+(window as any).TESTING_PKI = true;
```
Ironicbay: Forgot to delete this line

client/src/views/home/UserJoinOrganizationModal.vue
Max-7: Yeah, `canGoForward` is a computed, it watches variables used inside its scope. When the auth is reset, it doesn't change any variable inside the computed, so it doesn't know it has to recompute `canGoForward`. But since this variable changes state and its state is watched inside the computed, it forces a re-compute and an update of `canGoForward`. It's a dirty trick, this part of the code is among the oldest and it deserves a make-over but we don't really have the time, so it'll have to do for the time being.

Ironicbay (approved):

PR #12964, opened 2026-06-17
[MS] Implement a performance monitor

Ironicbay:


client/src/services/performanceMonitor.ts
```diff
@@ -0,0 +1,103 @@
...
+  maxTime: number;
+}
+
+interface MonitoredFunction {
+  name: string;
+  start: number;
+}
+
+export class PerformanceMonitor {
+  private stats: Map<string, Omit<PerformanceStats, 'avgTime'>> = new Map();
```
Max-7: I think it should be `averageTime`, not `avgTime`

Max-7 (approved):

Issue #12943, opened 2026-06-15
Add a function to libparsec to get global upload progress


Max-7:
The GUI can poll the function every X seconds and get:
- the number of files left to upload
- the volume of data left to upload

It should just be a query in the table `vlobs` and one in the table `chunks`.

Max-7:
Done by #12981

PR #12981, opened 2026-06-18
Outbound sync stats

AureliaDolo:
Parf of https://github.com/Scille/parsec-cloud/issues/12943

### DIFF
--- a/libparsec/crates/client/src/client/mod.rs
+++ b/libparsec/crates/client/src/client/mod.rs
@@ -60,6 +60,7 @@ use crate::{
         start_workspaces_refresh_list_monitor, Monitor,
     },
     user::UserOps,
+    workspace::WorkspaceGetOutboundSyncBacklogError,
     workspace_history::{WorkspaceHistoryOps, WorkspaceHistoryOpsStartError},
 };
 use libparsec_client_connection::AuthenticatedCmds;
@@ -109,6 +110,7 @@ pub use shamir_recovery_list::{
 };
 
 pub type ClientStartWorkspaceHistoryError = WorkspaceHistoryOpsStartError;
+pub type ClientGetOutboundSyncBacklogError = WorkspaceGetOutboundSyncBacklogError;
 
 // Should not be `Clone` given it manages underlying resources !
 pub struct Client {
@@ -412,15 +414,18 @@ impl Client {
         workspace_list::list_workspaces(self).await
     }
 
+    /// Get how much bytes and how many files are awaiting synchronization for started workspaces only
+    pub async fn get_outbound_sync_backlog(
+        &self,
+    ) -> Result<ClientGetOutboundSyncBacklog, ClientGetOutboundSyncBacklogError> {
+        workspace_outbound_sync_backlog::get_outbound_sync_backlog(self).await
+    }
+
     /// Create a new workspace.
     ///
     /// A new workspace starts its life locally (hence this function can be called while
     /// offline), and is later bootstrapped to be accessible to other users/devices (this
     /// is typically done by a monitor reacting to an event from this function).
-    pub async fn get_outbound_sync_backlog(&self) -> ClientGetOutboundSyncBacklog {
-        workspace_outbound_sync_backlog::get_outbound_sync_backlog(self).await
-    }
-
     pub async fn create_workspace(
         &self,
         name: EntryName,
--- a/libparsec/crates/client/src/client/mod.rs
+++ b/libparsec/crates/client/src/client/mod.rs
@@ -18,6 +18,7 @@ mod workspace_bootstrap;
 mod workspace_create;
 mod workspace_list;
 mod workspace_needs;
+mod workspace_outbound_sync_backlog;
 mod workspace_refresh_list;
 mod workspace_rename;
 mod workspace_self_promote;
@@ -40,6 +41,9 @@ pub use self::{
     workspace_create::ClientCreateWorkspaceError,
     workspace_list::WorkspaceInfo,
     workspace_needs::ClientProcessWorkspacesNeedsError,
+    workspace_outbound_sync_backlog::{
+        ClientGetOutboundSyncBacklog, ClientGetOutboundSyncBacklogItem,
+    },
     workspace_refresh_list::ClientRefreshWorkspacesListError,
     workspace_rename::ClientRenameWorkspaceError,
     workspace_self_promote::ClientSelfPromoteToWorkspaceOwnerError,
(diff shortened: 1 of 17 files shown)


libparsec/crates/client/src/workspace/store/mod.rs
```diff
@@ -384,15 +393,29 @@ impl WorkspaceStore {
         )
         .await?;
 
-        // 4) All set !
+        // 4) Reload the root manifest given startup writes may have updated it
```
AureliaDolo: I don't get why this step is added @touilleMan

libparsec/crates/client/src/workspace/store/mod.rs
```diff
@@ -319,6 +326,7 @@ pub(super) struct WorkspaceStore {
 
     data: data::WorkspaceStoreData,
     prevent_sync_pattern: PreventSyncPattern,
+    chunk_storage_overhead: u64,
```
AureliaDolo:
```suggestion
    // size (in bytes) of an empty chunk.
    chunk_storage_overhead: u64,
```

That's it ?

libparsec/crates/client/src/client/mod.rs
```diff
@@ -413,6 +417,10 @@ impl Client {
     /// A new workspace starts its life locally (hence this function can be called while
     /// offline), and is later bootstrapped to be accessible to other users/devices (this
     /// is typically done by a monitor reacting to an event from this function).
+    pub async fn get_outbound_sync_backlog(&self) -> ClientGetOutboundSyncBacklog {
+        workspace_outbound_sync_backlog::get_outbound_sync_backlog(self).await
+    }
+
```
AureliaDolo:
```suggestion
    /// Get how much bytes and how many files are awaiting synchronization
    pub async fn get_outbound_sync_backlog(&self) -> ClientGetOutboundSyncBacklog {
        workspace_outbound_sync_backlog::get_outbound_sync_backlog(self).await
    }

   /// Create a new workspace.
   ///
   /// A new workspace starts its life locally (hence this function can be called while
   /// offline), and is later bootstrapped to be accessible to other users/devices (this
   /// is typically done by a monitor reacting to an event from this function).
```

AureliaDolo: The diff is a bit broken, but get_outbound_sync_backlog was added between create_workspace and its doc, so I moved it up and added a bit of doc.

touilleMan: 👍

touilleMan: I think it is worth adding in the comment that only the started workspaces will be taken into account

libparsec/crates/client/src/client/workspace_outbound_sync_backlog.rs
```diff
@@ -0,0 +1,43 @@
...
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub struct ClientGetOutboundSyncBacklogItem {
+    pub realm_id: VlobID,
+    pub pending_entries: u64,
+    pub pending_bytes: u64,
+}
+
+#[derive(Debug, Clone, PartialEq, Eq, Default)]
+pub struct ClientGetOutboundSyncBacklog {
+    pub total_pending_entries: u64,
```
touilleMan: Should add a comment that only the started workspaces are taken into account, and that `total_pending_entries`/`total_pending_bytes` naming can be a bit misleading (maybe we want to rename those fields with a  `started_workspaces_...` prefix  to remove any ambiguity ?)

libparsec/crates/client/src/workspace/store/cache.rs
```diff
@@ -132,6 +134,7 @@ pub(super) struct CurrentViewCache {
     ///   should always be preferred over data coming from the database).
     /// - the root manifest is guaranteed to be always present (even after a clear !).
     pub manifests: ManifestsHashMap,
+    pub outbound_sync_backlog: WorkspaceOutboundSyncBacklog,
```
touilleMan: `manifests` and `outbound_sync_backlog` fields must be kept in sync, so we can no longer expose them as public but instead has to only allow modifying them through methods

libparsec/crates/client/src/workspace/store/mod.rs
```diff
@@ -428,6 +436,21 @@ impl WorkspaceStore {
             .with_current_view_cache(|cache| cache.manifests.root_manifest().clone())
     }
 
+    pub async fn get_outbound_sync_backlog(
+        &self,
+    ) -> Result<WorkspaceOutboundSyncBacklog, WorkspaceGetOutboundSyncBacklogError> {
+        self.data
+            .with_storage(|maybe_storage| async move {
+                if let Some(ref mut storage) = maybe_storage {
```
touilleMan:
```suggestion
        self.data.with_storage(|maybe_storage| async move {
                let storage = maybe_storage
                    .as_mut()
                    .ok_or_else(|| WorkspaceGetOutboundSyncBacklogError::Stopped)?;


                storage.get_outbound_sync_backlog().await
                    .map_err(WorkspaceGetOutboundSyncBacklogError::Internal)
        }.await
```

libparsec/crates/client/src/workspace/store/mod.rs
```diff
@@ -111,6 +113,12 @@ pub(super) enum EnsureManifestExistsWithParentError {
     Internal(#[from] anyhow::Error),
 }
 
+#[derive(Debug, thiserror::Error)]
+pub enum WorkspaceGetOutboundSyncBacklogError {
+    #[error(transparent)]
```
touilleMan, line 118: `Stopped` is missing

libparsec/crates/client/src/client/workspace_outbound_sync_backlog.rs
```diff
@@ -0,0 +1,47 @@
...
+pub struct ClientGetOutboundSyncBacklog {
+    pub total_pending_entries_for_started_workspaces: u64,
+    pub total_pending_bytes_for_started_workspaces: u64,
+    pub per_workspace: Vec<ClientGetOutboundSyncBacklogItem>,
+}
+
+pub async fn get_outbound_sync_backlog(
+    client: &Client,
+) -> Result<ClientGetOutboundSyncBacklog, ClientGetOutboundSyncBacklogError> {
+    let workspaces = client.workspaces.lock().await;
```
touilleMan: This locks `client.workspaces` for the whole time (so in the meantime we cannot do workspace start/stop and list workspaces).

Given this function does sqlite requests on multiple workspaces, I guess the lock can be kept taken for a long time, so we should instead try to release it as soon as possible:

```suggestion
    let workspaces: Vec<Arc<WorkspaceOps>> = {
        let workspaces = client.workspaces.lock().await;
        workspaces.clone() 
    };
```

touilleMan (approved):

PR #13004, opened 2026-06-23
[MS] Adds profile filtering in workspace sharing modal

Max-7:
image


client/src/locales/fr-FR.json
Ironicbay, line 1: You need to change the locale @ line 70
`Ayant accès ({count})` / `Déjà partagé ({count})` ...

Ironicbay (approved):

PR #13009, opened 2026-06-24
Ensure PrivateKey decrypt checks small order points

touilleMan:


libparsec/crates/crypto/Cargo.toml
```diff
@@ -148,6 +146,7 @@ hex-literal = { workspace = true }
 rmp-serde = { workspace = true }
 rstest = { workspace = true, features = ["async-timeout"] }
 curve25519-dalek = { workspace = true }
+crypto_box = { workspace = true }
```
FirelightFlagboy: **chore:** Already listed in `dependencies`

FirelightFlagboy (changes requested):

libparsec/crates/crypto/src/rustcrypto/private.rs
FirelightFlagboy, line 1: **nitpick:** Would be simpler to copy how it's done in `generate_shared_secret_key`

Since we can easily convert from one type to another

libparsec/crates/crypto/Cargo.toml
touilleMan: done. :+1:

libparsec/crates/crypto/src/rustcrypto/private.rs
touilleMan: That was my initial implementation, but:
- I disliked having a type conversion for every decryption
- I'd rather stay as close as possible to what does libsodium since it's battle-tested implementation we are mimicking
- In the long run I plan to replace this code with [RustCrypto's `crypto_box::SecretKey::unseal`](https://docs.rs/crypto_box/latest/crypto_box/struct.SecretKey.html#method.unseal] (that has been added since the time `libparsec_crypto` was written)

FirelightFlagboy (approved):

PR #13010, opened 2026-06-24
[MS] Search in all workspaces

Max-7:
- Refactored search to use an async generator
- Adds a def variable to create multiple workspaces
- Adds search accross all workspaces


client/src/components/files/explorer/FileSearchResultItem.vue
Ironicbay, line 1: Right-clicking an item when `disabledContextMenu === true` sets the hover color which remains indefinetely.

`menuOpened` shouldn't be set to `true` if `disabledContextMenu` is `true`.

Ironicbay (approved):

Issue #5557, opened 2023-10-19
Provide the server's version in the client


FirelightFlagboy:
It would be nice if we could have the server's version info displayed accessible somewhere in the client.

That would help troubleshooting / testing part where we need to know the current server version.

PR #13017, opened 2026-06-25
Api to get server version in client

AureliaDolo:
Backend for #5557

### DIFF
--- a/libparsec/crates/client_connection/src/anonymous_server_cmds.rs
+++ b/libparsec/crates/client_connection/src/anonymous_server_cmds.rs
@@ -67,22 +67,7 @@ impl AnonymousServerCmds {
     where
         T: ProtocolRequest<API_LATEST_MAJOR_VERSION> + Debug + 'static,
     {
-        #[cfg(feature = "test-with-testbed")]
-        let request = {
-            match self.send_hook.high_level_send(request).await {
-                crate::testbed::HighLevelSendResult::Resolved(rep) => return rep,
-                crate::testbed::HighLevelSendResult::PassToLowLevel(req) => req,
-            }
-        };
-
-        let request_body = request.api_dump().expect("Unexpected serialization error");
-
-        // Split non-generic code out of `send` to limit the amount of code generated
-        // by monomorphization
-        let api_version = api_version_major_to_full(T::API_MAJOR_VERSION);
-        let (response_body, _) = self.internal_send(api_version, request_body).await?;
-
-        Ok(T::api_load_response(&response_body)?)
+        self.send_verbose(request).await.map(|(rep, _)| rep)
     }
 
     /// Returns response, server version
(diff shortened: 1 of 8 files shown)


server/tests/api_v5/anonymous_server/test_server_config.py
AureliaDolo, line 1: Add that file to version_updater

FirelightFlagboy (approved):

mmmarcos (commented):
The `server_version` field and the new api version should be added to `version.rs`:

https://github.com/Scille/parsec-cloud/blob/51bba09049f201a4b102ad2f8139a6f3ee9c8d0a/libparsec/crates/protocol/src/version.rs#L54-L60

FirelightFlagboy:
CF https://github.com/Scille/parsec-cloud/pull/13017#pullrequestreview-4572169673

misc/version_updater.py
```diff
@@ -450,6 +450,9 @@ def set_tool_version(tool: Tool, version: str) -> None:
     ROOT_DIR / "server/parsec/_version.py": {
         Tool.Parsec: [ReplaceRegex(r'^__version__ = ".*"$', '__version__ = "{version}"')]
     },
+    ROOT_DIR / "server/tests/api_v5/anonymous_server/test_server_config.py": {
```
touilleMan: This seems rather weird to modify this test for each version bump (tests tend to change often, so the replace rule may break)

Instead I think having the test importing `parsec._version.__version__` and checking the result against it is enough

libparsec/crates/client_connection/src/anonymous_server_cmds.rs
FirelightFlagboy, line 1: **todo:** Reuse `send_versbose` in `send`

FirelightFlagboy (changes requested):

libparsec/crates/client_connection/src/anonymous_server_cmds.rs
```diff
@@ -64,13 +64,24 @@ impl AnonymousServerCmds {
     }
 
     pub async fn send<T>(&self, request: T) -> ConnectionResult<<T>::Response>
+    where
+        T: ProtocolRequest<API_LATEST_MAJOR_VERSION> + Debug + 'static,
+    {
+        let (rep, _) = self.send_verbose(request).await?;
+        Ok(rep)
```
FirelightFlagboy: **nitpick:** Would have used `.map` instead

libparsec/src/config.rs
```diff
@@ -175,6 +178,8 @@ pub async fn get_server_config(
                     }
                 ),
             },
+            server_version
+
```
FirelightFlagboy: **chore:** spacing

FirelightFlagboy (approved):

touilleMan (approved):

libparsec/crates/client_connection/src/anonymous_server_cmds.rs
```diff
@@ -101,10 +113,16 @@ impl AnonymousServerCmds {
         #[cfg(not(feature = "test-with-testbed"))]
         let resp = req.send().await?;
 
+        let server_version = resp
+            .headers()
+            .get("Server")
+            .map_or("unknown", |v| v.to_str().unwrap_or("invalid"))
```
touilleMan: nit: wrapping the value to make it more clear it is not a "real" value from the server

```suggestion
            .map_or("<unknown>", |v| v.to_str().unwrap_or("<invalid>"))
```

PR #13034, opened 2026-06-29
[MS] Display server version in organization info

### LINKED ISSUE #5557: Provide the server's version in the client

Ironicbay:
Following https://github.com/Scille/parsec-cloud/pull/13017
Closes https://github.com/Scille/parsec-cloud/issues/5557

### DIFF
--- a/client/src/views/organizations/OrganizationInformationPage.vue
+++ b/client/src/views/organizations/OrganizationInformationPage.vue
@@ -52,7 +52,10 @@
             :user-info="userInfo"
           />
           <!-- ------------- Information ------------- -->
-          <organization-configuration-information :org-info="orgInfo" />
+          <organization-configuration-information
+            :org-info="orgInfo"
+            :server-config="serverConfig"
+          />
 
           <!-- ------------- Storage list ------------- -->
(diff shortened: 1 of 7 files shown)


client/src/views/organizations/OrganizationInformationPage.vue
```diff
@@ -121,6 +125,7 @@ import { Ref, onMounted, onUnmounted, ref } from 'vue';
 const { isSmallDisplay } = useWindowSize();
 
 const orgInfo: Ref<OrganizationInfo | null> = ref(null);
+const serverVersion: Ref<string | null> = ref(null);
```
Max-7: Store the whole server config. It contains stuff like "do we have openbao", "do we have cryptpad", "is totp mandatory", ... I've got a feeling this will come handy to have here, so no point in restricting ourselves in only storing the server version. And then pass the whole server-config to the sub-component.

Max-7 (approved):

PR #13040, opened 2026-06-30
[MS] Fix revoked user selection and prevent user selection for members

Ironicbay:


client/src/components/users/UserCard.vue
```diff
@@ -120,6 +120,7 @@ const emits = defineEmits<{
 const props = defineProps<{
   user: UserModel;
   showCheckbox: boolean;
+  isAdmin: boolean;
```
Max-7: The prop should be more about the selection being enabled/disabled. It keeps the logic in the parent, and it prevents confusion (does `isAdmin` refers to `user` or the current client?). And if it is not enabled, we can indeed hide the checkbox, but also not emit the signal on click with `@click="selectionEnabled && $emit('select', user, !user.isSelected)"`.

client/src/views/users/UserListDisplay.vue
```diff
@@ -103,6 +105,7 @@ const props = defineProps<{
   selectionEnabled?: boolean;
   sortBy: SortProperty;
   sortAscending: boolean;
+  isAdmin: boolean;
```
Max-7: I think the `selectionEnabled` props is not properly named, it's more about the selection being active. I would rename it, and use a selectionEnabled (or selectionDisabled) instead of isAdmin to show/hide the checkboxes and emit/not emit the signals.

Max-7 (changes requested):

client/src/components/users/UserListItem.vue
```diff
@@ -60,7 +60,7 @@
         :user-avatar="user.humanHandle.label"
         class="user-mobile-avatar"
         :class="{
-          'hide-avatar': showCheckbox || isHovered,
+          'hide-avatar': activeCheckbox || isHovered,
```
Max-7: `activeCheckbox || (isHovered && allowSelection)`, otherwise the avatar disapears in small display and it looks weird.

Max-7 (approved):

PR #13044, opened 2026-07-03
[MS] Download multiple files

### LINKED ISSUE #12980: Add direct downlad of folders and multiple files
[FileSystemDirectoryHandle](https://developer.mozilla.org/en-US/docs/Web/API/FileSystemDirectoryHandle) and [FileSystemWritableFileStream](https://developer.mozilla.org/en-US/docs/Web/API/FileSystemWritableFileStream) are completely supported in all recent browsers, which means we could use it to download a whole file arborescence without using an archive.

We can still keep the archive and have two separate things: `Download` and `Download as archive`.

Max-7:
Closes #12980


client/src/locales/en-US.json
```diff
@@ -1181,6 +1182,7 @@
             "delete": "Delete",
             "share": "Share",
             "download": "Download",
+            "downloadAsArchive": "Download as a ZIP file",
```
Ironicbay, line 1189: This overflows in the tabbar menu, not sure what to replace it with. `Download ZIP` is too long as well.

client/src/views/files/FoldersPage.vue
```diff
@@ -1622,25 +1639,35 @@ async function downloadEntries(entries: EntryModel[]): Promise<void> {
...
-    await downloadArchive({
-      entries: entries,
-      archiveName: `${workspaceInfo.value.name}_${currentFolder.value === '/' ? 'ROOT' : currentFolder.value}.zip`,
-      workspaceHandle: workspaceInfo.value.handle,
-      workspaceId: workspaceInfo.value.id,
-      informationManager: informationManager.value,
-      fileOperationManager: fileOperationManager.value,
+  function _getArchiveName(): EntryName {
+    if (!workspaceInfo.value) {
+      return 'unknown.zip';
```
Ironicbay: Maybe `archive.zip`?

Ironicbay (commented):
Works great, two points to address before merging in the file operations window:
1. `Download files` operation needs to display a non-clickable checkmark instead of the clickable folder (as is the case for `Download` and `Download archive`)
Image

2. The `Download files` progress is not displayed correctly, indicating 0% until it's done.

Ironicbay (approved):

PR #13047, opened 2026-07-03
fix(electron): Update CSP rules to allow for custom cryptpad server

FirelightFlagboy:

### DIFF
--- a/client/electron/src/cspRules.ts
+++ b/client/electron/src/cspRules.ts
@@ -20,11 +20,33 @@ export function setupContentSecurityPolicy(customScheme: string): void {
 
   const CspRules: Array<[CspDirective, Array<string>]> = [
     [CspDirective.DefaultSrc, [customProtocol, 'devtools:']],
-    [CspDirective.ScriptSrc, [customProtocol, "'unsafe-inline'", 'https://*.stripe.com']],
+    [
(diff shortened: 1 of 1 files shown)


client/electron/src/cspRules.ts
```diff
@@ -20,11 +20,33 @@ export function setupContentSecurityPolicy(customScheme: string): void {
...
     [CspDirective.DefaultSrc, [customProtocol, 'devtools:']],
-    [CspDirective.ScriptSrc, [customProtocol, "'unsafe-inline'", 'https://*.stripe.com']],
+    [
+      CspDirective.ScriptSrc,
+      [
+        customProtocol,
+        "'unsafe-inline'",
+        'https://*.stripe.com',
+        // For custom cryptpad server:
+        'https:',
```
Max-7, line 30: This shouldn't be needed. This is used specifically to import scripts from another origin than the one we loaded the page from. Cryptpad doesnt do that: we load an iframe from the cryptpad server, so that iframe has the server's origin.
It's dangerous to allow script loading from any https server or unsafe-eval.

Max-7 (changes requested):

FirelightFlagboy:
No longer needed indeed

PR #13062, opened 2026-07-06
[MS] Fix UI workspace sharing modal

fabienSvtr:
Small updates to change:
- the user counter with and without filters.
- the css for the button (gather what it is common)
- `const` name to be more relevant
- user selection (it can be clicked on the container rather than only the checkbox)

image

### DIFF
--- a/client/src/locales/fr-FR.json
+++ b/client/src/locales/fr-FR.json
@@ -67,7 +67,8 @@
             "description": "Peut gérer l'accès des utilisateurs et rechiffrer l'espace de travail. Peut promouvoir d'autres utilisateurs au rôle de Propriétaire."
         },
         "none": "Non partagé",
-        "members": "Ayant accès ({count})",
+        "haveAccess": "Ayant accès ({count})",
+        "haveAccessFiltered": "Ayant accès ({count}/{total})",
         "suggestion": "Recommandés"
     },
     "link": {
@@ -624,7 +625,7 @@
             "contributor": "Contributeur",
             "reader": "Lecteur",
             "buttonSelect": "Sélectionner",
-            "buttonFinish": "Terminer",
+            "buttonCancel": "Annuler",
             "counter": "{count} utilisateur sélectionné | {count} utilisateurs sélectionnés",
             "chooseRole": "Choisir un rôle",
             "noChanges": "Les membres sélectionnés étaient déjà {role}.",
--- a/client/src/views/workspaces/WorkspaceSharingModal.vue
+++ b/client/src/views/workspaces/WorkspaceSharingModal.vue
@@ -89,7 +89,7 @@
       <ms-report-text
         :theme="MsReportTheme.Info"
         id="only-user-warning"
-        v-if="countTotalSharedUsers < 2"
+        v-if="countAllSharedUsers < 2"
       >
         <i18n-t
           keypath="WorkspaceSharing.onlyUserWarning"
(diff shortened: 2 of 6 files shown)


client/src/locales/fr-FR.json
```diff
@@ -67,7 +67,8 @@
             "description": "Peut gérer l'accès des utilisateurs et rechiffrer l'espace de travail. Peut promouvoir d'autres utilisateurs au rôle de Propriétaire."
         },
         "none": "Non partagé",
-        "members": "Ayant accès ({count})",
+        "haveAccess": "Ayant accès ({count})",
+        "haveAccessFiltered": "Ayant accès (filtré: {count})",
```
```suggestion
        "haveAccessFiltered": "Ayant accès (filtrés : {count})",
```

client/src/views/workspaces/WorkspaceSharingModal.vue
```diff
@@ -137,10 +137,27 @@
                 @change="selectAllMembers()"
                 v-show="showCheckboxes && selectableFilteredMembers.length > 0"
               />
-              {{ $msTranslate({ key: 'workspaceRoles.members', data: { count: countSharedUsers }, count: countSharedUsers }) }}
+              <span v-if="search.length > 0 || !profileFilters.All">
```
Ironicbay: **suggestion:** I would just base this on wether users are actually being filtered or not.
```suggestion
              <span v-if="countFiltersSharedUsers !== countAllSharedUsers">
```

client/src/views/workspaces/WorkspaceSharingModal.vue
Ironicbay: I would change the locale `WorkspaceSharing.batchSharing.buttonFinish` from `Finish`/`Terminer` to `Cancel`/`Annuler` (and maybe adapt the style), since the action is finished as the user clicks an option in the dropdown. Functionally this `Finish` button just cancels the selecting operation.

client/src/views/workspaces/WorkspaceSharingModal.vue
```diff
@@ -346,8 +372,12 @@ const filteredNotSharedUserRoles = computed(() => {
 });
 
 const selectedUsers = computed(() => userRoles.value.filter((user) => user.isSelected === true));
-const countSharedUsers = computed(() => filteredSharedUserRoles.value.length + (clientInfo.value ? 1 : 0));
-const countTotalSharedUsers = computed(
+const countFiltersSharedUsers = computed(() => {
+  const currentUserCounted =
+    clientInfo.value && (profileFilters.value.All || profileFilters.value[clientInfo.value.currentProfile]) ? 1 : 0;
```
Ironicbay: This doesn't take into account the search input, only the user profile filter. You can use the `currentUserMatchSearch()` function below, which checks the `clientInfo` value along with the search input for current user.

```suggestion
    currentUserMatchSearch() && (profileFilters.value.All || profileFilters.value[clientInfo.value!.currentProfile]) ? 1 : 0;
```

I added `!` to `clientInfo.value` since it's already checked in `currentUserMatchSearch()`.

Ironicbay (changes requested):

client/src/views/workspaces/WorkspaceSharingModal.vue
fabienSvtr: Yes, exaclty, I've updated the entire footer modal following these information.

Ironicbay (approved):

PR #13065, opened 2026-07-07
Replace `CertificateOps::encrypt_for_realm` by `CertificateOps::get_latest_realm_key_for_encryption`

touilleMan:
This is to be able to encrypt multiple pieces of data with the same key index (needed to implement
the `cryptpad_register_session` API which contains two encrypted keys to handle read&write and
read-only access).


FirelightFlagboy (changes requested):
Rebase PR

touilleMan:
> Rebase PR

done.

FirelightFlagboy (approved):

PR #13072, opened 2026-07-08
Cryptpad support: 3 server

touilleMan:


server/parsec/components/memory/cryptpad.py
```diff
@@ -0,0 +1,122 @@
...
+
+                case RealmRole.READER:
+                    if encrypted_candidate_edit_key is not None:
+                        return CryptpadRegisterSessionBadOutcome.AUTHOR_NOT_ALLOWED
+
+                    edit_allowed = False
+
+                case RealmRole.OWNER | RealmRole.MANAGER | RealmRole.CONTRIBUTOR:
+                    if encrypted_candidate_edit_key is None:
+                        return CryptpadRegisterSessionBadOutcome.AUTHOR_NOT_ALLOWED
```
mmmarcos, line 83: I am not sure if this is relevant, but this feels like a different kind of error (more like a "bad data" or "wrong config")

server/parsec/components/postgresql/migrations/0026_cryptpad.sql
```diff
@@ -0,0 +1,21 @@
+-- Parsec Cloud (https://parsec.cloud) Copyright (c) BUSL-1.1 2016-present Scille SAS
+
+-------------------------------------------------------
+--  Migration
+--
+-- Cryptpad support
```
mmmarcos: CryptPad is already "supported" in a way, this is about collaborative session.

```suggestion
-- Support for CryptPad collaborative editing session
```

mmmarcos (approved):

server/parsec/components/postgresql/migrations/0026_cryptpad.sql
touilleMan: :+1: done.

server/parsec/components/memory/cryptpad.py
touilleMan: I followed the same logic than `realm_role_certificate`:
https://github.com/Scille/parsec-cloud/blob/7fbc2cd441e990babe738fe562146193ee8c86e4/libparsec/crates/protocol/schema/authenticated_cmds/realm_share.json5#L30-L33

I think it is "good enough" since those error cases are roughly handled the same way from the client point of view[^1], but I can change it if you want :+1: 

[^1]: i.e. the client sends a `realm_role_certificate` but concurrently loses its role, or the client send a `realm_role_certificate` with an edit key but concurrently gets its role changed to READER. In both case the client is not allowed to start the edition session.
In theory from there the client could poll the certificates to try to understand what has changed (so that the client can automatically do another attempt without the edit key if it detects only the write access has been concurrently removed during the previous attempt). However the concurrent role change is such a nice case that this is not worth the complexity compared to just display an error message to the end-user and let him manually retry.

mmmarcos: I agree that is not worth it. Better stay consistent. Also we can always add a new error later if needed.

mmmarcos (approved):

PR #13074, opened 2026-07-08
Cryptpad support: 5 libparsec

touilleMan:


bindings/generator/api/workspace.py
```diff
@@ -815,3 +815,52 @@ async def workspace_decrypt_path_addr(
...
 ) -> Result[FsPath, WorkspaceDecryptPathAddrError]:
     raise NotImplementedError
+
+
+class CryptpadSessionKeys(Structure):
+    view_key: str
+    edit_key: str | None
+
+
+class WorkspaceRegisterCryptpadSessionError(ErrorVariant):
```
mmmarcos, line 825: `NotAllowed` missing?

libparsec/crates/client/src/workspace/transactions/register_cryptpad_session.rs
```diff
@@ -0,0 +1,255 @@
...
+
+pub use crate::CryptpadSessionKeys;
+use crate::{
+    workspace::WorkspaceOps, CertifGetLatestRealmKeyForEncryptionError, CertifPollServerError,
+    CertifValidateCryptpadSessionKeysError, EncryptionUsage, InvalidCertificateError,
+    InvalidCryptpadSessionKeysError, InvalidKeysBundleError,
+};
+
+#[derive(Debug, thiserror::Error)]
+pub enum WorkspaceRegisterCryptpadSessionError {
```
mmmarcos, line 16: `NotAllowed` missing?

libparsec/crates/client/src/workspace/transactions/register_cryptpad_session.rs
```diff
@@ -0,0 +1,255 @@
...
+                .map_err(|err| match err {
+                    CertifGetLatestRealmKeyForEncryptionError::Stopped => {
+                        WorkspaceRegisterCryptpadSessionError::Stopped
+                    }
+                    CertifGetLatestRealmKeyForEncryptionError::Offline(err) => {
+                        WorkspaceRegisterCryptpadSessionError::Offline(err)
+                    }
+                    CertifGetLatestRealmKeyForEncryptionError::NotAllowed => {
+                        WorkspaceRegisterCryptpadSessionError::NoRealmAccess
+                    }
```
mmmarcos, line 82: Is this accurate? user may not be allowed to edit, but still have access to realm. Also, if user _is_ allowed to edit, and the edit key is not sent, currently the server returns a `NotAllowed`, which mapped to `NoRealmAccess` may be even more puzzling.

mmmarcos (approved):

client/src/plugins/libparsec/definitions.ts
```diff
@@ -7449,6 +7526,12 @@ export interface LibParsecPlugin {
         entry_id: VlobID,
         mode: OpenOptions
     ): Promise<Result<FileDescriptor, WorkspaceOpenFileError>>
+    workspaceRegisterCryptpadSession(
+        workspace: Handle,
+        vlob_id: VlobID,
```
Max-7, line 7531: Is `vlob_id` the entry id? Why not call it `entry_id`?

client/src/plugins/libparsec/definitions.ts
```diff
@@ -7449,6 +7526,12 @@ export interface LibParsecPlugin {
         entry_id: VlobID,
         mode: OpenOptions
     ): Promise<Result<FileDescriptor, WorkspaceOpenFileError>>
+    workspaceRegisterCryptpadSession(
```
Max-7, line 7529: How does the function work? I'm a bit confused by the fact that you pass it some keys and it returns some keys.

client/src/plugins/libparsec/definitions.ts
touilleMan: I've renamed it `document_id` since this doesn't necessary correspond to a vlob ID (i.e. we never check that a vlob ID of corresponding ID exists in the database)

client/src/plugins/libparsec/definitions.ts
touilleMan: This function does two things:
- if no session exists for the file: create a new session
- if a session exists: reuse the session

Hence the key being passed are are *only* if the session needs to be created (hence the returned keys that must be used since they differs if the session already exist)

I have have added some documentation about this as `WorkspaceOps::register_cryptpad_session` docstring :+1:

bindings/generator/api/workspace.py
touilleMan: This is the `NoRealmAccess` error (for consistency with the other errors in WorkspaceOS, e.g. ReadChunkOrBlockError/PopulateCacheFromLocalStorageOrServerError/WorkspaceCreateFileError)

libparsec/crates/client/src/workspace/transactions/register_cryptpad_session.rs
touilleMan: same as above ^^

libparsec/crates/client/src/workspace/transactions/register_cryptpad_session.rs
touilleMan: A user without the write right on the workspace is still allowed to create a cryptpad session, it is just not allowed to provide an edit key in the creation.

> Also, if user is allowed to edit, and the edit key is not sent, currently the server returns a NotAllowed, which mapped to NoRealmAccess may be even more puzzling.

We talked about it in https://github.com/Scille/parsec-cloud/pull/13072#discussion_r3545690050 ^^

Basically if a user is allowed to edit and doesn't send the edit key this means two things:
- either there is a bug in our code, but this is a very obvious incorrect usage of the API so don't feel this is a serious issue
- either the user has just received the new role and hasn't processed it yet (hence the client still believe the user has a READER role). Since this has a very low chance to occurs (you have to start a session right when you get your role changed...) I think it is good to make it behave in a generic way which also cover the case you concurrently lost your role (which is a equally unlikely event ^^) in order to simplify the code. In both case a dialogue should be displayed to the end user and it's up to the end user to manually retry (which this time should work with the up to date role and hence behave correctly)

mmmarcos (approved):

PR #13080, opened 2026-07-09
[MS] Add useFileActions composable to isolate file action functions

### LINKED ISSUE #13060: Factorize context menus
The composable for context menu for both workspaces and files should include the actions, so they can be used anywhere.
Currently the composable only shows the context menu (so `Delete file` for example), but then the function to delete the file is located inside FoldersPage, which makes that context menu unusable from somewhere else (like a search page or workspace pages).

Ironicbay:
Closes https://github.com/Scille/parsec-cloud/issues/13060

### DIFF
--- a/client/src/views/files/FoldersPage.vue
+++ b/client/src/views/files/FoldersPage.vue
@@ -241,7 +241,6 @@
 import { entryNameValidator } from '@/common/validators';
 import * as parsec from '@/parsec';
 import {
-  Answer,
   DisplayState,
   DocumentImport,
   EmptyFolder,
@@ -260,7 +259,6 @@ import {
   MsSpinner,
   RenameIcon,
   Translatable,
-  askQuestion,
   asyncComputed,
   getTextFromUser,
   openSpinnerModal,
@@ -289,18 +287,14 @@ import {
   NewFilePopover,
   SortProperty,
   copyPathLinkToClipboard,
-  selectFolder,
 } from '@/components/files';
 import { FileSearch } from '@/components/files/handler/viewer/types';
 import { EntrySyncStatus, FileOperationCurrentFolder } from '@/components/files/types';
 import SmallDisplaySelectionHeader from '@/components/header/SmallDisplaySelectionHeader.vue';
 import { WorkspaceRoleTag } from '@/components/workspaces';
-import { showWorkspace } from '@/components/workspaces/utils';
 import {
   ClientInfo,
   EntryName,
-  EntryStatFile,
-  FsPath,
   Path,
   WorkspaceCreateFolderErrorTag,
   WorkspaceID,
@@ -312,6 +306,7 @@ import {
   listWorkspaces,
 } from '@/parsec';
 import { Routes, currentRouteIs, getCurrentRouteQuery, getDocumentPath, getWorkspaceHandle, navigateTo, watchRoute } from '@/router';
+import { useFileActions } from '@/services/contextMenu/fileActions';
 import { isFileEditable } from '@/services/cryptpad';
 import {
   EntrySyncData,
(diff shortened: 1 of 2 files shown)


client/src/services/contextMenu/fileContextOperations.ts
```diff
@@ -0,0 +1,501 @@
...
+  const workspaceAttributes = useWorkspaceAttributes();
+  const pathOpener = usePathOpener();
+  const fileOperationManager: Ref<FileOperationManager> = inject(FileOperationManagerKey)!;
+  const informationManager: Ref<InformationManager> = inject(InformationManagerKey)!;
+  const eventDistributor: Ref<EventDistributor> = inject(EventDistributorKey)!;
+  const storageManager: StorageManager = inject(StorageManagerKey)!;
+
+  async function deleteEntries(
+    entries: EntryModel[],
+    workspaceInfo: WorkspaceInfo | null,
```
Max-7: Having the `workspaceInfo` as a parameter to `useFileActions` would be ideal but it is probably a bit complicated, seing where `useFileActions` is used.
But we can at least make it mandatory: there's no reason why it would ever be null, the caller has to account for that, the function can just treat it as always defined since it's not doing anything otherwise.

client/src/services/contextMenu/fileContextOperations.ts
```diff
@@ -0,0 +1,501 @@
...
+          PresentationMode.Toast,
+        );
+      }
+    }
+  }
+
+  async function renameEntries(
+    entries: EntryModel[],
+    workspaceInfo: WorkspaceInfo | null,
+    currentPath: string,
```
Max-7: I think `currentPath` is useless here, it's just `await Path.parent(entry)`. We may have been using it in FoldersPage because we had it, but some other places (like the sidebar) will not.

client/src/services/contextMenu/fileContextOperations.ts
```diff
@@ -0,0 +1,501 @@
...
+                : 'FoldersPage.errors.deleteMultipleSomeFailed',
+            level: InformationLevel.Error,
+          }),
+          PresentationMode.Toast,
+        );
+      }
+    }
+  }
+
+  async function renameEntries(
```
Max-7: The function was originally called `renameEntries` and took an Array to be generic and have the same signature as all the other functions, but it is currently a bit confusing because if you pass more than one file, it does nothing. I think we can rename it and change `entries` to `entry`.

client/src/services/contextMenu/fileContextOperations.ts
```diff
@@ -0,0 +1,501 @@
...
+        }),
+        PresentationMode.Toast,
+      );
+      return;
+    }
+
+    await fileOperationManager.value.move(workspaceInfo.handle, entries, folder, dupPolicy);
+  }
+
+  async function showDetails(entries: EntryModel[], workspaceInfo: WorkspaceInfo | null, userInfo: ClientInfo | null): Promise<void> {
```
Max-7: Instead of the whole ClientInfo, just take a  mandatory `UserProfile`.

client/src/views/files/FoldersPage.vue
```diff
@@ -1073,15 +1058,9 @@ async function listFolder(options?: { selectFile?: EntryName; sameFolder?: boole
...
-    return;
-  }
-  const entry = entries[0];
-  searchPattern.value = '';
-  await cancelSearch();
-  const parent = await parsec.Path.parent(entry.path);
-  await navigateTo(Routes.Documents, {
-    query: { documentPath: parent, workspaceHandle: workspaceInfo.value?.handle, selectFile: entry.name },
+  await fileActions.showEnclosingFolder(entries, workspaceInfo.value, async () => {
+    searchPattern.value = '';
```
Max-7: I think the `onFinished` is overkill, it should work by just cancelling the search before or even after calling the function, I don't think it necessarily has to be cancelled at a specific time.

client/src/views/files/FoldersPage.vue
```diff
@@ -1340,395 +1319,54 @@ function getSelectedEntries(): EntryModel[] {
...
-        new Information({
-          message:
-            errorsEncountered === entries.length
-              ? 'FoldersPage.errors.deleteMultipleAllFailed'
-              : 'FoldersPage.errors.deleteMultipleSomeFailed',
-          level: InformationLevel.Error,
-        }),
-        PresentationMode.Toast,
-      );
+  await fileActions.deleteEntries(entries, workspaceInfo.value, async () => {
```
Max-7: Same here, probably overkill, you can probably just reset the search after calling the function.

Max-7 (commented):
Looks good overall, but the API can be simplified and cleaned. It has historically been a bit complicated for various reasons, but we can take this opportuniy to simplify it.

- `workspaceInfo` can be mandatory, it's not the function's job to check that it's not null, it's the caller.
- the functions that only work on a single entry can be renamed and their parameters changed
- `onFinished` functions should not be necessary as they are a bit confusing. What's the difference between them and just adding instructions after the function call?
- `currentPath` should usually not be needed if it can be infered from the entry.

Max-7 (approved):
Maybe just rename the file: it contains `useFileActions` but is named `fileContextOperations`, and we already have `fileOperations` which is slightly different, so something like `fileActions.ts` seems better.

PR #13083, opened 2026-07-12
Cryptpad support: 7 Correct manifest merge to handle the new `origin` field in `FileManifest`

touilleMan:


libparsec/crates/client/src/workspace/merge.rs
```diff
@@ -143,7 +143,38 @@ pub(super) fn merge_local_file_manifest(
         return MergeLocalFileManifestOutcome::Merged(new_local);
     }
 
-    // 4) Merge data and ensure the sync is still needed
+    // 4) Special case if both local and remote have been edited from the same
```
AureliaDolo, line 146: The case when a file is edited both in a cryptpad session and outside of it should be considered the same as without the cryptpad session right ?

AureliaDolo (approved):

libparsec/crates/client/src/workspace/merge.rs
touilleMan: Yes totally (since we can only merge the conflict if there is a single source of truth, which a cryptpad session is)

FirelightFlagboy (approved):

PR #13084, opened 2026-07-12
Cryptpad support: 8 libparsec bindings `workspace_save_and_sync_file_with_cryptpad`

touilleMan:


libparsec/crates/client/src/workspace/transactions/save_and_sync_file_with_origin.rs
```diff
@@ -0,0 +1,250 @@
...
+    let mut opened_file = opened_file.lock().await;
+    let manifest: &mut LocalFileManifest = Arc::make_mut(&mut opened_file.manifest);
+    manifest.origin = origin;
+    opened_file.flush_needed = true;
+    opened_file.modified_since_opened = true;
+
+    Ok(())
+}
+
+fn map_sync_error(err: WorkspaceSyncError) -> WorkspaceSaveAndSyncFileWithOriginError {
```
AureliaDolo, line 213: Is there a reason for that to not be a impl From<WorkspaceSyncError> for WorkspaceSaveAndSyncFileWithOriginError ?

libparsec/crates/client/src/workspace/transactions/save_and_sync_file_with_origin.rs
```diff
@@ -0,0 +1,250 @@
...
+    match err {
+        WorkspaceSyncError::Offline(e) => WorkspaceSaveAndSyncFileWithOriginError::Offline(e),
+        WorkspaceSyncError::ServerBlockstoreUnavailable => {
+            WorkspaceSaveAndSyncFileWithOriginError::ServerBlockstoreUnavailable
+        }
+        WorkspaceSyncError::Stopped => WorkspaceSaveAndSyncFileWithOriginError::Stopped,
+        WorkspaceSyncError::NotAllowed => WorkspaceSaveAndSyncFileWithOriginError::NoRealmAccess,
```
AureliaDolo, line 223: Maybe add a comment kinda like this one https://github.com/Scille/parsec-cloud/pull/13074#discussion_r3567097303 so we are not surprised by that in a few months/years

AureliaDolo (approved):

libparsec/crates/client/src/workspace/transactions/save_and_sync_file_with_origin.rs
touilleMan: done. (I've added a comment in `libparsec/crates/client/src/workspace/transactions/save_and_sync_file_with_origin.rs`, `libparsec/crates/client/src/workspace/transactions/register_cryptpad_session.rs` and `libparsec/crates/protocol/schema/authenticated_cmds/cryptpad_register_session.json5` ^^ )

libparsec/crates/client/src/workspace/transactions/save_and_sync_file_with_origin.rs
touilleMan: What do we get by doing this ?

From my point of view:

- replace `.await.map_err(map_sync_error)?` by `.await?`
- remove `map_sync_error()` but add a `impl From<WorkspaceSyncError> for WorkspaceSaveAndSyncFileWithOriginError {`

So we would getting slightly better ergonomic and the cost of not controlling where this error handling occurs (i.e. we don't explicitly apply it), which is an issue considering the comment I added about why the `NoRealmAccess` error doesn't have to handle the read-only-realm case is precisely about a specific ordering of the errors handling:
```
        // Note `WorkspaceSyncError` doesn't discriminate between "no access" and "only read access"
        // but it's fine here since we call `open_file_by_id` (that does such discrimination)
        // before doing the sync operation.
```

PR #13093, opened 2026-07-15
[MS] Migrate to node 24

### LINKED ISSUE #12689: Migrate to Node.js v24
We should consider migration to Node.js v24 (LTS) which, among other things (v8 engine upgrade, performance improvements, security enhancements) provides npm v11 and security options to manage dependencies (`min-release-age`, `allow-git`, ...)

https://nodejs.org/en/about/previous-releases

### Tasks list

- Check migration guide [Node.js v22 to v24](https://nodejs.org/en/blog/migrations/v22-to-v24)
- Bump node `misc/versions.toml` & run `version_updater.py`
- Check that node is updated in CI 
- Check that npm is updated to v11
- Enable options in `.npmrc`
  - https://github.com/Scille/parsec-sign/blob/master/.npmrc#L19
  - https://github.com/Scille/megashark-lib/blob/master/.npmrc#L19
  - https://github.com/Scille/parsec-cloud/blob/master/.npmrc#L19

Max-7:
Closes #12689


.github/workflows/package-server.yml
```diff
@@ -42,7 +42,7 @@ concurrency:
   cancel-in-progress: true
 
 env:
-  node-version: 22.22.2
+  node-version: 24.18.0
```
FirelightFlagboy: **polish:** To remove, not used in the workflow

Ironicbay (approved):

PR #13096, opened 2026-07-15
chore(docs): Migrate from `poetry` to `uv`

FirelightFlagboy:


misc/version_updater.py
```diff
@@ -220,9 +220,7 @@ def set_tool_version(tool: Tool, version: str) -> None:
         ]
     },
     ROOT_DIR / ".github/workflows/_releaser_nightly_build.yml": {Tool.Python: [PYTHON_GA_VERSION]},
-    ROOT_DIR / ".github/workflows/ci-docs.yml": {
-        Tool.Poetry: [POETRY_GA_VERSION],
-    },
+    ROOT_DIR / ".github/workflows/ci-docs.yml": {},
```
mmmarcos: Is there any reason to keep the line at all?
```suggestion
```

docs/README.md
mmmarcos, line 1: I would add a step below to activate the venv. 
Or mention that `make` should be run with `uv run`:

```
$ uv run make html
```

mmmarcos (approved):

Issue #12999, opened 2026-06-22
Add rich text when appropriate


Max-7:
With the addition of https://github.com/Scille/megashark-lib/pull/354, rich text can now be used. Go through translations and adds it where it would benefit.

Figma:
- https://www.figma.com/design/XW3SNWHV8BhPsPJpsOCV2h/%E2%98%81%EF%B8%8F-Parsec-App?node-id=36641-59215&t=kFWLbzIy0ITHECqD-11
- https://www.figma.com/design/XW3SNWHV8BhPsPJpsOCV2h/%E2%98%81%EF%B8%8F-Parsec-App?node-id=36623-194440&t=kFWLbzIy0ITHECqD-11
- https://www.figma.com/design/XW3SNWHV8BhPsPJpsOCV2h/%E2%98%81%EF%B8%8F-Parsec-App?node-id=36665-83223&t=kFWLbzIy0ITHECqD-11
- https://www.figma.com/design/XW3SNWHV8BhPsPJpsOCV2h/%E2%98%81%EF%B8%8F-Parsec-App?node-id=39049-73896&t=kFWLbzIy0ITHECqD-11

PR #13106, opened 2026-07-16
[MS] Add rich text to archive and delete workspace modals

Ironicbay:
Linked to https://github.com/Scille/parsec-cloud/issues/12999

### DIFF
--- a/client/src/components/invitations/InviteModal.vue
+++ b/client/src/components/invitations/InviteModal.vue
@@ -18,7 +18,7 @@
   >
     <div class="email-input-container">
       <ms-report-text :theme="MsReportTheme.Info">
-        <ion-text>{{ $msTranslate('UsersPage.CreateUserInvitationModal.subtitle') }}</ion-text>
+        <ms-rich-text text="UsersPage.CreateUserInvitationModal.subtitle" />
       </ms-report-text>
       <ms-input
         @change="onInputChange"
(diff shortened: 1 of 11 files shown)


Max-7 (commented):
Looks good. A few locations we may want to change:
- when restoring a trashed workspace, the dialog contains the workspace name.
- when deleting a file, the dialog contains the file name
- when revoking a user, the dialog contains the user's name and could benefit from a \n
- when creating an invite, the dialog has `by separating the email addresses with ';'`, maybe this can be changed with rich text instead. 
- refusing a async enrollment request has a blue button, should probably be red. A bit outside the scope of this PR but may as well fix that while we're at it.
- when deleting an invitation, the dialog displays the email (and could benetif from a \n)
- multiple locations where we used `i18n-t` could be replaced by rich text (maybe not all, but at least some). Since this is outside of a modal, it could be considered outside the scope of this PR.

client/src/components/invitations/InviteModal.vue
```diff
@@ -18,7 +18,7 @@
   >
     <div class="email-input-container">
       <ms-report-text :theme="MsReportTheme.Info">
-        <ion-text>{{ $msTranslate('UsersPage.CreateUserInvitationModal.subtitle') }}</ion-text>
+        <ms-rich-text :text="'UsersPage.CreateUserInvitationModal.subtitle'" />
```
Max-7: Just `text="UsersPage..."`, you don't need the `:` and the quotes here.

client/src/locales/en-US.json
```diff
@@ -880,8 +880,8 @@
         },
         "restoreWorkspace": {
             "title": "Restore this workspace",
-            "subtitleArchived": "You are about to restore the workspace {workspace}. Previous read and write accesses will be restored. Are you sure you want to proceed?",
-            "subtitleTrashed": "You are about to restore the workspace {workspace}. Its planned deletion will be cancelled. Are you sure you want to proceed?",
+            "subtitleArchived": "You are about to restore the workspace <b>{workspace}</b>. Previous read and write accesses will be restored.<br/>Are you sure you want to proceed?",
```
Max-7: `archiveWorkspace.subtitle` has two <br>'s before the "Are you sure". We should be consistent.

Max-7 (approved):

PR #13136, opened 2026-07-24
refactore(hosting): Rework docker testing infra

FirelightFlagboy:
- Replace unmaintained `minio` with `rustfs`
- Rename `gen-admin-token` to `gen-secrets`
- Generate secrets in `secrets/` folders
- Add pre/post service to configure `docker-compose` environment
- Rename service to be shorter
- Add `healthcheck` so `depends_on` fields can be updated to wait for healthy services.


docs/hosting/deployment/server-deploy.rst
```diff
@@ -131,26 +131,25 @@ These variables can be stored in a file and sourced before running the server.
...
+.. _creating_secrets:
 
-To be able to perform admin tasks (like creating an organization) on the server, an administration token is required.
-Below you will find a simple script to generate a token:
+Creating the secrets
+^^^^^^^^^^^^^^^^^^^^
 
-.. admonition:: gen-admin-token.sh
-   :collapsible: closed
+The deployment of the services require the use of some secret values:
```
mmmarcos:
```suggestion
Deploying the services requires certain secret values:
```

docs/hosting/deployment/server-deploy.rst
```diff
@@ -131,26 +131,25 @@ These variables can be stored in a file and sourced before running the server.
...
 
-   .. literalinclude:: gen-admin-token.sh
-     :language: bash
-     :linenos:
+- An administration token for admin operations on the server.
+- A random seed for *parsec-account* for non-existing accounts.
```
mmmarcos: Parsec Account is not yet available, this should be commented or excluded from docs.

docs/hosting/deployment/server-deploy.rst
```diff
@@ -131,26 +131,25 @@ These variables can be stored in a file and sourced before running the server.
...
+- Access credential for the object-storage service ``rustfs`` used in the ``docker-compose`` stack.
```
mmmarcos:
```suggestion
- Access credential for ``rustfs``, which is the object-storage service used in the ``docker-compose`` stack.
```

docs/hosting/deployment/server-deploy.rst
```diff
@@ -131,26 +131,25 @@ These variables can be stored in a file and sourced before running the server.
...
 
-The script will generate a random token (``openssl rand -hex 32``) and create the env file ``parsec-admin-token.env``.
+For that, we provide a simple scripts that generate such secrets and save them locally:
```
mmmarcos:
```suggestion
For that, we provide a simple script that generate these secrets and save them locally:
```

docs/hosting/deployment/server-deploy.rst
```diff
@@ -190,6 +189,20 @@ Create the file ``parsec-s3.env`` with the following content to set the URL for
      :language: bash
      :linenos:
 
+.. note::
+
+   The env file re-use environment variables defined in ``rustfs.env`` and ``secrets/rustfs.env`` (cf: :ref:`Creating the secrets <creating_secrets>`).
+   Those variables are used to configured the ``rustfs`` service that provide the S3-like API when using the ``docker-compose`` approach.
```
mmmarcos:
```suggestion
   Those variables are used to configured the ``rustfs`` service that provide the object-storage API when using the ``docker-compose`` approach.
```

docs/hosting/deployment/server-deploy.rst
```diff
@@ -251,22 +264,22 @@ You can use the following Docker Compose file to deploy Parsec Server for testin
...
-+---------------------+-----------------------------------------------------------------------------------+
-| ``parsec-server``   | The Parsec Server                                                                 |
-+---------------------+-----------------------------------------------------------------------------------+
-| ``parsec-proxy``    | A Nginx proxy server, used as an example to configure a reverse proxy.            |
-|                     |                                                                                   |
-|                     | Learn more about                                                                  |
-|                     | :ref:`using Parsec behind a reverse proxy<doc_hosting_deployment_behind_a_proxy>` |
-+---------------------+-----------------------------------------------------------------------------------+
++--------------+-----------------------------------------------------------------------------------+
+| Service name | Description                                                                       |
```
mmmarcos, line 282: Just curious: why not keeping the `parsec` prefix? Aren't these services exposed "globally" in Docker? If so, better to be able to tell them apart from other non-Parsec services.

docs/hosting/deployment/parsec-s3.env
```diff
@@ -23,4 +23,4 @@
 #  where <raid_type> is RAID0/RAID1/RAID5, <node> is an integer and
 # `<config>` is one of the previous s3/swift/POSTGRESQL/MOCKED configuration.
 
-PARSEC_BLOCKSTORE=s3:parsec-s3\:9000:region1:parsec:S3_ROOT_USER:S3_ROOT_PASS
+PARSEC_BLOCKSTORE=s3:$RUSTFS_ESCAPED_ADDRESS:$RUSTFS_REGION:$BUCKET_NAME:$RUSTFS_ACCESS_KEY:$RUSTFS_SECRET_KEY
```
mmmarcos, line 26: It should be the other way around: the variables defined here should be used for `rustfs`.

Reason: `PARSEC_BLOCKSTORE` is supposed to be set by the server administrators according to their needs (`rustfs` being our choice for the docker compose deploy). The variable is documented here (see above) and in the CLI. 
`RUSTFS_` variables won't necessarily make sense to admins.

docs/hosting/deployment/server-deploy.rst
FirelightFlagboy: - Shorter to writen when chaining docker compose commands
- We set `container_name` in the compose file (otherwise docker compose default to `{base_dir}_{service}`)

docs/hosting/deployment/parsec-s3.env
FirelightFlagboy: If admin sys want to further customize, (which they need to do as it's a testing infra and not production ready) they would ultimately change that value.

For the testing infra, it's simpler that way as we cannot do extensive parsing on a `.env` file

docs/hosting/deployment/server-deploy.rst
FirelightFlagboy: It's still required by the server

docs/hosting/deployment/server-deploy.rst
```diff
@@ -33,6 +33,20 @@ Parsec offers the following deployment options:
 - :ref:`Container-Based deployment <doc_hosting_deployment_with_docker>`
 - :ref:`Direct installation on Linux server <doc_hosting_deployment_with_linux>`
 
+TL,DR - Quick deploy with docker
+================================
+
+We provide a detailed explanation of the required configuration below.
+But if you just want to deploy locally a parsec server you can do the following command:
```
mmmarcos:
```suggestion
But if you just want to deploy a Parsec server locally you can do it with the following commands:
```

docs/hosting/deployment/server-deploy.rst
```diff
@@ -33,6 +33,20 @@ Parsec offers the following deployment options:
 - :ref:`Container-Based deployment <doc_hosting_deployment_with_docker>`
 - :ref:`Direct installation on Linux server <doc_hosting_deployment_with_linux>`
 
+TL,DR - Quick deploy with docker
```
mmmarcos: nit
```suggestion
TL;DR - Quick deploy with Docker
```

docs/hosting/deployment/server-deploy.rst
mmmarcos: Still, since it is not used, it would be better to use some dummy value and avoid mentioning here.

Or, alternatively, something like:

```suggestion
- A random seed for *parsec-account* (this is an unfinished feature, the seed is not being used but it is still required by the server)
```

mmmarcos (approved):

PR #13152, opened 2026-07-27
Install electron dependencies on CI and checks package-lock

Max-7:


.github/workflows/ci-web.yml
```diff
@@ -65,7 +65,9 @@ jobs:
         with:
           node-version: ${{ env.node-version }}
           cache: npm
-          cache-dependency-path: client/package-lock.json
+          cache-dependency-path: |
+            client/package-lock.json
+            client/electron/package-lock.json
```
FirelightFlagboy: **issue:** Exclude electron from the cache

The electron node_modules take ~500Mo, which is too much given we are already at the 10Gb limit of the Github cache.

Maybe [workspaces](https://docs.npmjs.com/cli/v8/using-npm/workspaces) could help deduplicate dependencies

FirelightFlagboy (changes requested):

FirelightFlagboy (approved):

PR #13155, opened 2026-07-27
feat(cli): Add command to mount a workspace

### LINKED ISSUE #13137: [CLI] Provide mountpoint for workspace
Add a command to mount a workspace on the filesystems like how the desktop applications does.

The CLI already is able to do that for sequester export (`parsec-cli mount-realm-export`).

FirelightFlagboy:
- Add `Mountpoint::mount_at_path` to allow specify a mount path over relying on `MountpointMountStrategy`

Closes #13137

### DIFF
--- a/cli/src/commands/workspace/mount.rs
+++ b/cli/src/commands/workspace/mount.rs
@@ -1,0 +1,65 @@
+use anyhow::Context;
+use libparsec_platform_mountpoint::Mountpoint;
+
+use crate::utils::StartedClient;
+use std::path::PathBuf;
+
+crate::clap_parser_with_shared_opts_builder!(
+    #[with = config_dir, device, password_stdin, workspace]
+    pub struct Args {
+        /// Path where to mount the workspace
+        #[arg(value_hint = clap::ValueHint::DirPath)]
+        mount_dir: PathBuf,
+    }
+);
+
+crate::build_main_with_client!(
+    main,
+    mount_workspace,
+    libparsec::ClientConfig {
+        with_monitors: true, // Enable monitor to keep workspace's content updated
+        ..Default::default()
+    }
+    .into()
+);
+
+pub async fn mount_workspace(args: Args, client: &StartedClient) -> anyhow::Result<()> {
+    log::trace!("Mounting workspace");
+    let Args {
+        workspace,
+        mount_dir,
+        ..
+    } = args;
+    log::debug!(
+        "Starting workspace {workspace} with device {}",
+        client.device_id()
+    );
+    let wksp = client
+        .start_workspace(workspace)
+        .await
+        .context("Failed to start workspace")?;
+
+    let mountpoint = Mountpoint::mount_at_path(wksp, mount_dir)
+        .await
+        .context("Failed to mount workspace at specified directory")?;
+
+    println!("Workspace mounted, send SIGINT signal (Ctrl-C) to stop");
+
+    let (tx, mut rx) = tokio::sync::mpsc::channel(1);
+    ctrlc::set_handler(move || {
+        let _ = tx.try_send(());
+    })
+    .expect("Failed to set Ctrl-C handler");
+    rx.recv().await.expect("Ctrl-C handler failed");
+
+    println!("Signal received, stopping...");
+    let mut unmount_option = libparsec_platform_mountpoint::UnmountOptions::default();
+    unmount_option.remove_dir = false;
+    mountpoint
+        .unmount_with_options(unmount_option)
+        .await
+        .context("Failed to unmount workspace")?;
+
+    client.stop_workspace(workspace).await;
+    Ok(())
+}
--- a/cli/src/commands/workspace/mod.rs
+++ b/cli/src/commands/workspace/mod.rs
@@ -3,6 +3,7 @@ pub mod create;
 pub mod import;
 pub mod list;
 pub mod list_users;
+pub mod mount;
 pub mod share;
 pub mod sync;
 
@@ -22,6 +23,8 @@ pub enum Group {
     Share(share::Args),
     /// Sync workspace data with the server
     Sync(sync::Args),
+    /// Mount a workspace on the filesystem
+    Mount(mount::Args),
 }
 
 pub async fn dispatch_command(command: Group) -> anyhow::Result<()> {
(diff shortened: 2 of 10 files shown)


cli/src/commands/workspace/mount.rs
```diff
@@ -0,0 +1,60 @@
...
+use libparsec_platform_mountpoint::Mountpoint;
+
+use crate::utils::StartedClient;
+use std::path::PathBuf;
+
+crate::clap_parser_with_shared_opts_builder!(
+    #[with = config_dir, device, password_stdin, workspace]
+    pub struct Args {
+        /// Path where to mount the workspace
+        #[arg(long, env = "PARSEC_WORKSPACE_MOUNT_DIR", default_value_os_t = std::env::current_dir().expect("Cannot get current directory"))]
```
mmmarcos: The `current_dir` should not be the default value. Otherwise, running the command without specifying a path (`parsec-cli workspace mount -d <device> -w <wid>`) will replace the contents of the current working directory (where the command is executed) with the content of the workspace. The data in the current directory becomes inaccessible. 

If we want to specify a default value, the only possible value would be the same path as the one used by Parsec desktop app.

---

That said, IMO the path should not be an option at all. 

The `workspace import` command expects `SRC` and `DEST` as arguments:
```
Usage: parsec-cli workspace import [OPTIONS] --workspace <WORKSPACE> <SRC> <DEST>
```

The `workspace mount` command should follow the same logic for `mount_dir`:
```
Usage: parsec-cli workspace mount [OPTIONS] --workspace <WORKSPACE> <MOUNT_DIR>
```

cli/src/commands/workspace/mount.rs
```diff
@@ -0,0 +1,60 @@
...
+    log::debug!(
+        "Starting workspace {workspace} with device {}",
+        client.device_id()
+    );
+    let wksp = client
+        .start_workspace(workspace)
+        .await
+        .context("starting workspace")?;
+
+    let mountpoint = Mountpoint::mount_at_path(wksp, mount_dir)
```
mmmarcos, line 42: This seems a little odd: mount does not create the directory, but unmount removes it.

```shell
# 1. Try to mount at ~/myws (dir does not exists)
$ parsec-cli workspace mount -d <DEVICE> -w <WID> --mount-dir ~/myws
Error: mounting workspace

Caused by:
    No such file or directory (os error 2)

# 2. It requires the directory to exist, so let's create it
$ mkdir ~/myws

# 3. Try to mount again, now it works
$ parsec-cli workspace mount -d <DEVICE> -w <WID> --mount-dir ~/myws
Workspace mounted, send SIGINT signal (Ctrl-C) to stop

^CSignal received, stopping ...

# 4. But after stop, the directory is removed, so a subsequent calls do not work
$ parsec-cli workspace mount -d <DEVICE> -w <WID> --mount-dir ~/myws
Error: mounting workspace

Caused by:
    No such file or directory (os error 2)
```

On mount, the `mount_dir` should be created if it does not exists. If it exists and it is not empty, the command should fail early. 

The checks should be performed here so the command can print more user friendly messages (`Info: creating directory <mount-dir>`, `Error: cannot mount at <mount-dir>, the directory is not empty`).

mmmarcos (changes requested):

cli/src/commands/workspace/mount.rs
FirelightFlagboy: The command is similar to `mount` that does not create the mount directory by default.

Having an "unclean" directory does not have any impact (outside of hiding the directory content in favor of the workspace)

cli/src/commands/workspace/mount.rs
FirelightFlagboy: done

cli/src/commands/workspace/mount.rs
mmmarcos: > The command is similar to `mount` that does not create the mount directory by default.

But is not similar to `umount` which does not remove it.

It's the asymmetry that is odd here: if parsec-cli does not create the mount directory on mount, it should not remove it on unmount.

> Having an "unclean" directory does not have any impact (outside of hiding the directory content in favor of the workspace)

Hiding the directory content _is_ an impact (specially if it defaults to the current directory) and it may not be unwanted by the user. The CLI should at least inform the user (like `Warning: mounting the workspace on a non-empty directory (its content won't be accessible until unmount)`).

FirelightFlagboy: > It's the asymmetry that is odd here: if parsec-cli does not create the mount directory on mount, it should not remove it on unmount.

Fixed since the previous push

cli/src/commands/workspace/mount.rs
```diff
@@ -0,0 +1,63 @@
...
+        .await
+        .context("mounting workspace")?;
```
mmmarcos:
```suggestion
        .context("Failed to mount workspace at specified directory")?;
```

cli/src/commands/workspace/mount.rs
```diff
@@ -0,0 +1,63 @@
...
+        let _ = tx.try_send(());
+    })
+    .expect("Failed to set Ctrl-C handler");
+    rx.recv().await.expect("Ctrl-C handler failed");
+
+    println!("Signal received, stopping ...");
+    mountpoint
+        .unmount_with_options(false)
+        .await
+        .context("unmounting workspace")?;
```
mmmarcos:
```suggestion
        .context("Failed to unmount workspace")?;
```

cli/src/commands/workspace/mount.rs
```diff
@@ -0,0 +1,63 @@
...
+        ..
+    } = args;
+    log::debug!(
+        "Starting workspace {workspace} with device {}",
+        client.device_id()
+    );
+    let wksp = client
+        .start_workspace(workspace)
+        .await
+        .context("starting workspace")?;
```
mmmarcos:
```suggestion
        .context("Failed to start workspace")?;
```

cli/src/commands/workspace/mount.rs
```diff
@@ -0,0 +1,63 @@
...
+    println!("Workspace mounted, send SIGINT signal (Ctrl-C) to stop");
+
+    let (tx, mut rx) = tokio::sync::mpsc::channel(1);
+    ctrlc::set_handler(move || {
+        let _ = tx.try_send(());
+    })
+    .expect("Failed to set Ctrl-C handler");
+    rx.recv().await.expect("Ctrl-C handler failed");
+
+    println!("Signal received, stopping ...");
```
mmmarcos:
```suggestion
    println!("Signal received, stopping...");
```

mmmarcos (approved):

PR #13176, opened 2026-07-29
chore(mountpoint): Update fuser to 0.17

FirelightFlagboy:
- Update include proper types for most elements in `trait FileSystem`
- No longer need to spawn our own thread, `Session::spawn` does that for ourself
- Remove `SessionUnmounter` in favor of `BackgroundSession::umount_and_join`


AureliaDolo (approved):
LGTM, you may want a shark's eye for the js part though

PR #13185, opened 2026-07-31
[MS] Refactorized workspace actions and added workspade global contex…

Max-7:


client/src/services/contextMenu/workspaceContextMenu.ts
```diff
@@ -0,0 +1,223 @@
...
+        showBackdrop: false,
+        dismissOnSelect: true,
+      });
+
+      await popover.present();
+      action = (await popover.onDidDismiss()).data?.action;
+    } else {
+      const modal = await modalController.create({
+        component: SmallDisplayWorkspaceGlobalContextMenu,
+        cssClass: 'workspace-global-context-sheet-modal',
```
Ironicbay: I don't think this class exists. The modal doesn't show correctly

Ironicbay (approved):

PR #13187, opened 2026-07-31
feat(cli): Add command `sequester generate-service-certificate`

### LINKED ISSUE #13116: [Sequester] generate the service certificate with the client CLI
Should be a offline command in the CLI

see in https://github.com/Scille/parsec-cloud/pull/13079#discussion_r3552000203_

FirelightFlagboy:
- Also refactor `assert_cmd` macros to allow for any input implementing `AsRef<OsStr>` simplifying passing `Path` notably.

Closes #13116

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2207,7 +2207,7 @@ dependencies = [
  "portable-atomic",
  "portable-atomic-util",
  "serde_core",
- "windows-sys 0.52.0",
+ "windows-sys 0.61.2",
 ]
 
 [[package]]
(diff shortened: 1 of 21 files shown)


AureliaDolo (approved):
Would be nice to update the doc here : https://github.com/Scille/parsec-cloud/blob/master/docs/hosting/sequester/index.rst#sequester-service-key-and-certificate , either in this PR or another

FirelightFlagboy:
> Would be nice to update the doc here : [`master`/docs/hosting/sequester/index.rst#sequester-service-key-and-certificate](https://github.com/Scille/parsec-cloud/blob/master/docs/hosting/sequester/index.rst#sequester-service-key-and-certificate) , either in this PR or another

It's planned, but note that we would need to keep the old way documented (for historical reason)

PR #13208, opened 2026-08-05
feat(cli): Show long version using `shadow-rs`

FirelightFlagboy:


libparsec/crates/crypto/src/lib.rs
FirelightFlagboy, line 1: **thought:** Maybe it too much for `crypto` as for that crate a would like to know which backend it use and such feature can easily be done manually

cli/tests/integration/version.rs
```diff
@@ -1,7 +1,30 @@
+use predicates::boolean::PredicateBooleanExt;
+
 #[test]
-fn version() {
-    crate::assert_cmd_success!("--version").stdout(
+fn short_version() {
```
touilleMan: I find it weird that the ouput is determined by using short or long style of option

Typically short options are useful when typing by hand to save typing time while long one are good when writting script to keep readability high.

Would it be possible to instead trigger the detailed output when the option is provided twice ? 

This is what Python does:
```shell
$ python --version
Python 3.12.3

$ python -V
Python 3.12.3

$ python --version --version
Python 3.12.3 (main, Jun 19 2026, 12:46:00) [GCC 13.3.0]

$ python -V -V
Python 3.12.3 (main, Jun 19 2026, 12:46:00) [GCC 13.3.0]
```

touilleMan: Another possibility would be to always display the long detailed version, this is probably better since it ensure a user will always provide us the long detailed version when we ask him (instead of having to be extra clear that `parsec -V` is not equivalent to `parsec --version`/`parsec --version --version`)

Also we don't loose much since if the end user needs only the short version he can just do `parsec --version | head -n 1`

touilleMan (approved):

cli/tests/integration/version.rs
FirelightFlagboy: > Would it be possible to instead trigger the detailed output when the option is provided twice ?

The version flags are provided by `clap` and does not allow overriding the behavior

PR #13221, opened 2026-08-06
refactor server email templates

mmmarcos:
- Add missing tests for `render_email` command
- Add email greeting to base templates (e.g. "Hi Alice")
  - currently only supported for the async enrollment accepted email
- Add `base.txt.j2` for text templates to reuse common texts
- Add missing `description` block to Parsec Account emails
- Add missing `title` blocks for _device_ and _shamir_ invitation emails
- Update async enrollment accepted email to use the "namespace approach"
  to render steps (to be consistent with `invite` and `totp-reset` emails)
- Add `README.md` describing template syntax

### DIFF
--- a/server/parsec/templates/email/base.txt.j2
+++ b/server/parsec/templates/email/base.txt.j2
@@ -1,0 +1,6 @@
+{% if user_label is defined %}Hi {{ user_label }},{% endif %}
+{% block main_content required %}{% endblock %}
+
+Thank you,
+The Parsec team
+https://parsec.cloud/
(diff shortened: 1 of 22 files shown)


server/parsec/templates/email/base.txt.j2
```diff
@@ -0,0 +1,6 @@
+{% if user_label is defined %}Hi {{ user_label }},{% endif %}
```
FirelightFlagboy, line 1: **question:** Why not making `user_label` required?

It would require to update every call site, maybe it's doable in that PR?

mmmarcos: I tried. It requires to update every impacted component which is relatively straightforward for the memory implementations, but needs to modify SQL requests in most cases to request the label from human handle (which I wanted to avoid ^^).

FirelightFlagboy (approved):

PR #13236, opened 2026-08-07
[MS] Fixes some hotkeys and adds a few

Ironicbay:


client/src/views/files/FoldersPage.vue
```diff
@@ -741,28 +743,48 @@ async function defineShortcuts(): Promise<void> {
   );
   hotkeys.add(
     { key: 'enter', modifiers: Modifiers.None, platforms: Platforms.MacOS, disableIfModal: true, route: Routes.Documents },
```
Max-7, line 745: The platform doesn't work, pressing Enter on web triggers the renaming. But it's not your fault, you only added the check for how many entries are selected, which works properly.

Max-7 (approved):

PR #13256, opened 2026-08-10
refactor(cli): Migrate invite sub command to new UI

FirelightFlagboy:


cli/src/commands/invite/greet.rs
FirelightFlagboy, line 1: **todo:** missed some `println`

AureliaDolo (approved):

PR #13265, opened 2026-08-12
refactor(cli): Move `mount_realm_export` to `sequester::mount`

FirelightFlagboy:
- Alias previous command to new path with an added a warning


cli/src/macros.rs
```diff
@@ -306,3 +306,33 @@ macro_rules! clap_parser_with_shared_opts_builder {
         }
     };
 }
+
+/// Macro to used in the previous command module location.
+/// Will call the command from its new location with an added warning message.
```
AureliaDolo: Should we plan to drop that support someday ? ie treat the old command as unknown.

Should we also add the version in which the command is been deprecated/replaced ?

AureliaDolo: Also it's not the first time we just rename a command (forget local device has been renamed recently) should, so I'm not sure that  having the warning is really useful. Documenting it in the changelog (via the newsfragment) should be enough ?

FirelightFlagboy: > Should we plan to drop that support someday ?

Surely

> Should we also add the version in which the command is been deprecated/replaced ?

That can be useful in conjonction to plan the support

> Documenting it in the changelog should be enough ?

Changelog should be enough

AureliaDolo (approved):

AureliaDolo (approved):

PR #13278, opened 2026-08-14
docs(cli): Add CLI generated manuals

FirelightFlagboy:
Needed to add a custom sphinx extension to be able to process the roff files and perform some document transformation


AureliaDolo (commented):
Not sure exactly where it's originating from, but the capitalization for the command names as title is weird 

Image

Also if that's possible it would be nice to omit the synopsis/option/description from the toc tree in cli/manual.rs

FirelightFlagboy:
> Not sure exactly where it's originating from, but the capitalization for the command names as title is weird

It's originate from the custom extension, when we convert the filename into a title: 

https://github.com/Scille/parsec-cloud/blob/2eee5a449a8c41bee716a3d45c9b572227983f88/docs/_ext/manpages.py#L56-L64

The alternative would be to use slug (e.g. `parsec-cli-sequester-mount`)

FirelightFlagboy:
> Also if that's possible it would be nice to omit the synopsis/option/description from the toc tree in cli/manual.rs

Done

AureliaDolo:
> The alternative would be to use slug (e.g. parsec-cli-sequester-mount)

And that would not be ideal as it would be confusing, with the subcommands and the commands with multiple words.

FirelightFlagboy:
> And that would not be ideal as it would be confusing, with the subcommands and the commands with multiple words.

I will investigate if it's possible to generate the title from the first few word of the synopsis

AureliaDolo (approved):

PR #13292, opened 2026-08-18
[CLI] Add keyring support for bootstrap organization.

### LINKED ISSUE #10021: [🐛 | Bug]: CLI support keyring in bootstrap organization
### Parsec version tested on:

3.3.3

### Platforms tested on:

Any

### Bug description:

Currently only password is supported

https://github.com/Scille/parsec-cloud/blob/050fceb7664512f9f1f1ca863a165f9a999711a6/cli/src/commands/organization/bootstrap.rs#L55

### Relevant output:

_No response_

AureliaDolo:
Fix #10021

### DIFF
--- a/cli/src/commands/organization/bootstrap.rs
+++ b/cli/src/commands/organization/bootstrap.rs
@@ -2,42 +2,42 @@
 use std::{fmt::Write as _, io::Write as _, path::PathBuf, sync::Arc};
 
 use libparsec::{
-    AvailableDevice, ClientConfig, DeviceLabel, DevicePrimaryProtectionStrategy,
-    DeviceSaveStrategy, EmailAddress, HumanHandle, ParsecOrganizationBootstrapAddr, Password, Url,
+    AvailableDevice, ClientConfig, DeviceLabel, DeviceSaveStrategy, EmailAddress, HumanHandle,
+    ParsecOrganizationBootstrapAddr, Url,
 };
 
 use crate::{ui::compat::AvailableDeviceDisplay, utils::*};
 
-#[derive(clap::Parser)]
-pub struct Args {
-    /// Bootstrap address
-    /// (e.g: parsec3://127.0.0.1:6770/Org?no_ssl=true&action=bootstrap_organization&token=59961ba6dcc9b018d2fdc9da1c0c762b716a27cff30594562dc813e4b765871a
-    /// or http://127.0.0.1:6770/Org?no_ssl=true&action=bootstrap_organization&token=59961ba6dcc9b018d2fdc9da1c0c762b716a27cff30594562dc813e4b765871a)
-    #[arg(short, long, value_hint = clap::ValueHint::Url)]
-    addr: Url,
-    /// Device label
-    #[arg(short, long, value_hint = clap::ValueHint::Hostname)]
-    device_label: DeviceLabel,
-    /// User fullname
-    #[arg(short, long, value_hint = clap::ValueHint::Username)]
-    label: String,
-    /// User email
-    #[arg(short, long, value_hint = clap::ValueHint::EmailAddress)]
-    email: EmailAddress,
-    /// Sequester authority verify key path
-    #[arg(long, value_hint = clap::ValueHint::FilePath)]
-    sequester_key: Option<PathBuf>,
-    /// Read the password from stdin instead of TTY
-    #[arg(long, default_value_t)]
-    password_stdin: bool,
-}
+crate::clap_parser_with_shared_opts_builder!(
+    #[with = password_stdin, auth]
+    pub struct Args {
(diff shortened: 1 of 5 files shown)


cli/src/commands/organization/bootstrap.rs
```diff
@@ -30,14 +30,17 @@ pub struct Args {
     /// Read the password from stdin instead of TTY
     #[arg(long, default_value_t)]
     password_stdin: bool,
+    /// Authentication method for the new device
+    #[arg( long, value_hint = clap::ValueHint::Other, default_value = "password")]
+    auth: NewAuthMethod,
```
FirelightFlagboy: **todo:** Refactor to include the arg in `clap_parser_with_shared_opts_builder`.

You will need it for the previous command, this one and the next (`invite claim`)

cli/src/commands/organization/bootstrap.rs
```diff
@@ -3,41 +3,41 @@ use std::{fmt::Write as _, io::Write as _, path::PathBuf, sync::Arc};
...
-    #[arg(long, default_value_t)]
-    password_stdin: bool,
-}
+crate::clap_parser_with_shared_opts_builder!(
+    #[with = password_stdin, auth]
+    pub struct Args {
+        /// Bootstrap address
+        /// (e.g: parsec3://127.0.0.1:6770/Org?no_ssl=true&action=bootstrap_organization&token=59961ba6dcc9b018d2fdc9da1c0c762b716a27cff30594562dc813e4b765871a
+        /// or http://127.0.0.1:6770/Org?no_ssl=true&action=bootstrap_organization&token=59961ba6dcc9b018d2fdc9da1c0c762b716a27cff30594562dc813e4b765871a)
+        #[arg(long, value_hint = clap::ValueHint::Url)]
```
AureliaDolo, line 17: I had to remove the short here, it was conflicting with --auth. I can do the opposite though, but IMHO it made sense that way.

FirelightFlagboy: I would remove short option for both:

- Previously it denoted `--addr` but now `--auth` meaning a breaking change for existing script
- Why `-a` for `--auth` and not `--addr`

FirelightFlagboy (changes requested):
Need a newsfragment for breaking change

cli/src/commands/organization/bootstrap.rs
```diff
@@ -78,13 +76,25 @@ pub async fn main(ui: crate::Ui, args: Args) -> anyhow::Result<()> {
...
+                totp_protection: None,
+                primary_protection: DevicePrimaryProtectionStrategy::Password { password },
+            }
         }
-    })?;
+        NewAuthMethod::Keyring => DeviceSaveStrategy {
+            totp_protection: None,
+            primary_protection: DevicePrimaryProtectionStrategy::Keyring,
+        },
+    };
```
FirelightFlagboy: **polish:** Would refactor that block into `NewAuth` to expose a function `obtain_save_strategy()` (or similar) to have a single point of implementation

cli/src/commands/organization/bootstrap.rs
```diff
@@ -78,13 +76,25 @@ pub async fn main(ui: crate::Ui, args: Args) -> anyhow::Result<()> {
     let human_handle = HumanHandle::new(email, &label)
         .map_err(|e| anyhow::anyhow!("Cannot create human handle: {e}"))?;
 
-    let password = choose_password(if password_stdin {
-        ReadPasswordFrom::Stdin
-    } else {
-        ReadPasswordFrom::Tty {
-            prompt: "New device password:",
+    let strategy = match auth {
```
FirelightFlagboy: **nitpick:** Since this pattern already appear 2 times and given `libparsec::DeviceSaveStrategy` can "easily" be converted to `libparsec_platform_device::DeviceSaveStrategy`, I would make `NewAuth` implement a method to return `libparsec::DeviceSaveStrategy`

cli/src/commands/device/import_recovery_device.rs
```diff
@@ -42,7 +42,7 @@ pub async fn main(ui: crate::Ui, args: Args) -> anyhow::Result<()> {
             prompt: "Enter passphrase for the recovery file:",
         }
     })?;
-    let strategy = auth.get_save_strategy(password_stdin)?;
+    let strategy = auth.get_client_save_strategy(password_stdin)?;
```
FirelightFlagboy: **typo:** Change appear in the wrong PR?

FirelightFlagboy (approved):

cli/src/commands/device/import_recovery_device.rs
AureliaDolo: yep

Issue #10028, opened 2025-03-28
Cient CLI: standardize on doing `--authentication=<flag>` to setup the device authentication


touilleMan:
see https://github.com/Scille/parsec-cloud/pull/10017#discussion_r2018150218

We need to standardize all commands that save a device file (i.e. `invite claim`, `organization bootstrap`, `device change-authentication`)

A possibility would be to provide a `--authentication` param that could take the folling values:
- `--authentication=password`
- `--authentication=keyring`

(later on we may also introduce `--authentication=smartcard`)

related to #10021 

https://github.com/Scille/parsec-cloud/blob/050fceb7664512f9f1f1ca863a165f9a999711a6/cli/src/commands/invite/claim.rs#L31-L32

AureliaDolo:
Done in https://github.com/Scille/parsec-cloud/pull/13295

PR #13294, opened 2026-08-19
[CLI] Use shared auth opt for invite claim.

AureliaDolo:
Part of #10028

### DIFF
--- a/cli/src/commands/invite/claim.rs
+++ b/cli/src/commands/invite/claim.rs
@@ -21,7 +21,7 @@ use libparsec::{
     },
     ClientConfig, ParsecInvitationAddr, Url,
 };
-use libparsec_client::{DeviceSaveStrategy, ShamirRecoveryClaimFinalizeCtx};
+use libparsec_client::ShamirRecoveryClaimFinalizeCtx;
 
 use crate::{
     ui::{compat::AvailableDeviceDisplay, Color},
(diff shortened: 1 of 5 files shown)


FirelightFlagboy (changes requested):
Missing newsfragment for breaking change (remove of `--use-keyring` flag)

cli/tests/integration/invitations/user.rs
FirelightFlagboy, line 1: **polish:** You can also use `#[value("foo", "bar")]` to simplify creating the test matrix

cli/src/commands/invite/claim.rs
```diff
@@ -498,29 +484,14 @@ async fn step5_shamir(
...
 async fn save_user(
     ui: &crate::Ui,
     ctx: UserClaimFinalizeCtx,
-    save_mode: SaveMode,
+    auth: NewAuthMethod,
+    password_stdin: bool,
 ) -> anyhow::Result<()> {
     let key_file = ctx.get_default_key_file();
-    let save_strategy = get_save_strategy(save_mode)?;
+    let save_strategy = auth.get_save_strategy(password_stdin)?;
```
FirelightFlagboy: **nitpick:** Would call that on the caller site since we don't need to provide extra argument specific from that function

FirelightFlagboy (approved):

FirelightFlagboy (approved):

PR #13295, opened 2026-08-19
[CLI] Use shared option --auth for device change change_authentication

AureliaDolo:
Part of #10028


FirelightFlagboy (changes requested):
Missing newsfragment for breaking change of removed options

FirelightFlagboy (approved):

FirelightFlagboy (approved):

PR #13308, opened 2026-08-25
[CLI] Add export workspace command

### LINKED ISSUE #13140: [CLI] Download file from a workspace
Basically the inverse of `workspace import`, allows the user to download files/folders from a workspace.

The command would take the following inputs:

- device ID
- Workspace ID
- Entry Path
- Output Folder
- Update mode

AureliaDolo:
Fix #13140

### DIFF
--- a/cli/src/commands/workspace/export.rs
+++ b/cli/src/commands/workspace/export.rs
@@ -1,0 +1,129 @@
+use std::path::PathBuf;
+
+use libparsec::{FsPath, OpenOptions};
+use tokio::io::AsyncWriteExt;
+
+use crate::utils::StartedClient;
+
+const CHUNK_SIZE: usize = 4096;
+
+crate::clap_parser_with_shared_opts_builder!(
+    #[with = config_dir, device, password_stdin, workspace]
+    pub struct Args {
+        /// File to export (e.g. "myfile.txt")
+        #[arg(value_hint = clap::ValueHint::FilePath)]
+        pub(crate) src: FsPath,
+        /// Destination path (e.g. "/path/to/myfile.txt")
+        ///
+        /// The command will fail if the
+        /// parent directories do not exist, unless the `parents` option is enabled.
+        ///
+        /// If the destination file already exists, its content will be replaced.
+        #[arg(value_hint = clap::ValueHint::DirPath)]
+        pub(crate) dest: PathBuf,
+        /// If specified, create parent directories as needed
+        ///
+        /// No error if parent directories already exist (similar to `mkdir -p`)
+        #[clap(long, short, action)]
+        pub(crate) parents: bool,
+        /// Control how existing files are updated.
+        ///
+        /// (similar to `cp --update=...`)
+        #[clap(long, value_enum, default_value_t)]
+        update: UpdateMode,
+    }
+);
+
+#[derive(Debug, Default, Clone, Copy, PartialEq, Eq, clap::ValueEnum)]
+enum UpdateMode {
+    /// Existing files in destination are replaced
+    All,
+    /// Existing files in destination are not replaced but raise an error instead.
+    #[default]
+    NoneFail,
+}
+
+crate::build_main_with_client!(
+    main,
+    workspace_export,
+    libparsec::ClientConfig {
+        with_monitors: true,
+        ..Default::default()
+    }
+    .into()
+);
+
+pub async fn workspace_export(
+    _ui: crate::Ui,
+    args: Args,
+    client: &StartedClient,
+) -> anyhow::Result<()> {
+    let Args {
+        src,
+        dest,
+        workspace: wid,
+        update,
+        parents,
+        ..
+    } = args;
+
+    log::trace!(
+        "workspace_export: {wid}:{src} -> {dst}",
+        src = src,
+        dst = dest.display()
+    );
+
+    let workspace = client.start_workspace(wid).await?;
+
+    let file = workspace
+        .open_file(src.clone(), OpenOptions::read_only())
+        .await
+        .map_err(|e| match e {
+            libparsec::WorkspaceOpenFileError::EntryNotFound => {
+                anyhow::anyhow!("File {src} not found in Parsec workspace.")
+            }
+            libparsec::WorkspaceOpenFileError::EntryNotAFile { .. } => {
+                anyhow::anyhow!("{src} is not a file.")
+            }
+            _ => e.into(),
+        })?;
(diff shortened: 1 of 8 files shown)


cli/src/commands/workspace/export.rs
```diff
@@ -0,0 +1,106 @@
...
+
+crate::clap_parser_with_shared_opts_builder!(
+    #[with = config_dir, device, password_stdin, workspace]
+    pub struct Args {
+        /// File to export (e.g. "myfile.txt")
+        #[arg(value_hint = clap::ValueHint::FilePath)]
+        pub(crate) src: FsPath,
+        /// Destination path (e.g. "/path/to/myfile.txt")
+        ///
+        ///  The command will fail if the
```
FirelightFlagboy: **typo:** 
```suggestion
        /// The command will fail if the
```

cli/src/commands/workspace/export.rs
```diff
@@ -0,0 +1,106 @@
...
+        /// parent directories do not exist, unless the `parents` option is enabled.
```
FirelightFlagboy, line 19: **todo:** Missing `parents` option per the option description

cli/src/commands/workspace/export.rs
```diff
@@ -0,0 +1,106 @@
...
+        dst = dest.display()
+    );
+
+    let workspace = client.start_workspace(wid).await?;
+
+    let file = workspace.open_file(src, OpenOptions::read_only()).await?;
+    let stats = workspace.fd_stat(file).await?;
+    let size = stats.size;
+
+    if matches!(update, UpdateMode::NoneFail) && tokio::fs::try_exists(&dest).await? {
```
FirelightFlagboy: **todo:** `UpdateMode` impl `Eq` so use equality operator

cli/src/commands/workspace/export.rs
```diff
@@ -0,0 +1,106 @@
...
+
+    log::trace!(
+        "workspace_export: {wid}:{src} -> {dst}",
+        src = src,
+        dst = dest.display()
+    );
+
+    let workspace = client.start_workspace(wid).await?;
+
+    let file = workspace.open_file(src, OpenOptions::read_only()).await?;
```
FirelightFlagboy: **todo:** We likely want to check if the provided path is a file beforehand to provide better error message since this command only support exporting file and not folder recursively

cli/src/commands/workspace/export.rs
```diff
@@ -0,0 +1,106 @@
...
+        // read and decrypt chunk
+        let bytes_read = workspace
+            .fd_read(file, offset, CHUNK_SIZE as u64, &mut read_buf)
+            .await?;
+
+        // write to dst file
+        write_buf
+            .write_all(&read_buf[..bytes_read.try_into()?])
+            .await?;
+        write_buf.flush().await?;
```
FirelightFlagboy: **todo:** Move the flush outside of the loop

cli/tests/integration/workspace/export.rs
```diff
@@ -0,0 +1,118 @@
...
+                &wid.hex(),
+                &remote_path,
+                &local_path,
+                "--update",
+                &update
+            )
+            .stdout(predicates::str::is_empty());
+        }
+        _ => unimplemented!(),
+    }
```
FirelightFlagboy: **todo:** Factory up to stdout/stderr

FirelightFlagboy (changes requested):

cli/src/commands/workspace/export.rs
```diff
@@ -0,0 +1,131 @@
...
+    let size = stats.size;
+
+    if (update == UpdateMode::NoneFail) && tokio::fs::try_exists(&dest).await? {
+        return Err(anyhow::anyhow!("File already exists."));
+    }
+
+    if parents {
+        if let Some(p) = dest.parent() {
+            if !p.exists() {
+                return Err(anyhow::anyhow!("Parent directory does not exist."));
```
FirelightFlagboy: **issue:** If `--parents` option is provided, it's expected that you create the missing parents

FirelightFlagboy (changes requested):

cli/src/commands/workspace/export.rs
```diff
@@ -0,0 +1,133 @@
...
+    }
+
+    if let Some(p) = dest.parent() {
+        if !p.exists() {
+            if !parents {
+                return Err(anyhow::anyhow!("Parent directory does not exist."));
+            } else {
+                tokio::fs::create_dir_all(p).await?
+            }
+        }
```
FirelightFlagboy: **nitpick:** I would simplify that block and only keep `create_dir_all` similarly to how it's done in `workspace import`.

It's not required to check if the parents exists since eitherway it will fail at the next step.

cli/src/commands/workspace/export.rs
```diff
@@ -0,0 +1,129 @@
...
+
+    if (update == UpdateMode::NoneFail) && tokio::fs::try_exists(&dest).await? {
+        return Err(anyhow::anyhow!("File already exists."));
+    }
+
+    if parents {
+        if let Some(p) = dest.parent() {
+            tokio::fs::create_dir_all(p).await?
+        }
+    }
```
FirelightFlagboy, line 101: **nitpick:** 
```suggestion
    if parents && let Some(p) = dest.parent() {
        tokio::fs::create_dir_all(p).await?
    }
```

FirelightFlagboy (approved):

cli/src/commands/workspace/export.rs
AureliaDolo: I've tried that, but it doesn't works because if let some is it's own can of worms https://play.rust-lang.org/?version=stable&mode=debug&edition=2024&gist=716f43bd6f755fc79fce453d167366d3

FirelightFlagboy: I'm confused, the code work, what's the issue ?

https://play.rust-lang.org/?version=stable&mode=debug&edition=2024&gist=f315989d6bceee0664449e3ca411de39

AureliaDolo: image

So I guess I got confused with this error message

FirelightFlagboy: You code is lacking an extra `&` :wink:

AureliaDolo: Nooo 
image

And we are still in 2021 :cry:

PR #11213, opened 2025-09-26
[MS] Fixed back button after logout

Max-7:
#11150

PR #13327, opened 2026-09-01
[MS] Fix back button error homepage and multi-handle navigation errors

### LINKED ISSUE #11150: [MS] Black screen when trying to go back to a logged out organization
### Parsec version

3.5.0-rc0

- [ ] Any
- [x] macOS
- [ ] Linux
- [ ] Windows
- [x] Web

### Bug description

Similar to https://github.com/Scille/parsec-cloud/issues/11095

After browsing a workspace inside an organization, and then logging out, the user can use the mouse `back` button which might try to take them back to the logged out organization.

https://github.com/user-attachments/assets/03e083e0-5e04-45d6-bdb6-67e2aa228131

### Reproduction steps

```bash
1. Login to an org
2. Populate the router history by browsing a workspace and subfolders
3. Log out
4. Click mouse `back` button
```

### Relevant output:

_No response_

Ironicbay:
Follows #11213
Closes #11150

### DIFF
--- a/client/src/router/types.ts
+++ b/client/src/router/types.ts
@@ -1,6 +1,15 @@
 // Parsec Cloud (https://parsec.cloud) Copyright (c) BUSL-1.1 2016-present Scille SAS
 
-import { ConnectionHandle, EntryName, FsPath, ParsecAccount, ParsecWorkspacePathAddr, WorkspaceHandle, WorkspaceName } from '@/parsec';
+import {
+  ConnectionHandle,
+  EntryName,
+  FsPath,
+  getClientInfo,
+  ParsecAccount,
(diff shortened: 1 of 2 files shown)


client/src/router/types.ts
```diff
@@ -250,6 +259,23 @@ router.beforeEach((to, from) => {
     routeBeforeFileHandler = undefined;
     return;
   }
+  if (to.params.handle) {
```
fabienSvtr: Since we have `if (!to.name || !to.params.handle) {` (l.258), I don't think this first `if` is necessary.

fabienSvtr (approved):

Max-7 (approved):

Issue #8352, opened 2024-09-12
Automatically delete organizations on the trial server after 15 days


Max-7:
Create a server command that could be a `crontab` to delete organizations after 15 days

The command need to remove:

- metadata (pg)
- data (s3)

PR #13331, opened 2026-09-02
feat(server): Add CLI task to list organizations

FirelightFlagboy:
In preparation of #8352, I needed to bootstrap the `tasks` cli group and investigate if it's doable to perform operation without relying on a running backend.

Also investigate how to support `pydantic` directly in `server` crate allowing to skip defining `*Field` annotations and making intergration of parsec types simpler on python side with pydantic (Usefull for the added command since it rely on pydantic & parsec's types)


server/Cargo.toml
```diff
@@ -16,6 +16,8 @@ name = "parsec"
 crate-type = ["cdylib"]
 
 [features]
+default = ["pydantic-support"]
```
touilleMan, line 19: This crate is only needed by the Python code, and I guess we cannot use it if we disable this `pydantic-support` feature.

So that makes this `pydantic-support` feature mandatory (hence why it is in `default`), hence we can just remove it

AureliaDolo (approved):

PR #13338, opened 2026-09-03
[CLI] Display long available device id

### LINKED ISSUE #8602: [CLI] Command `device remove` use a fixed short_id for device when displaying confirmation message

AureliaDolo:
So forget local device can display the full ID.
Fix #8602


FirelightFlagboy (changes requested):
The goal of the refactor was also to normalize how to show things to user.

Better to choose one version only, I would prefer the long format

AureliaDolo:
From what I understand of the issue, the goal was to just display the long id during the confirmation process. It may be worth having a look if there are other places where the id is displayed for confirmation.

I'd rather keep the short id, at least in the list that is shown when a device id is missing from a command. What to do with the real list available device is up to discussion.

Or we could get rid of the short ids completely ? Or never show them but still be able to accept them as input (or there maybe there is some auto completion trick that can be done ?)

FirelightFlagboy:
For displaying short id, you cannot use an hardcoded value because it may not display enough information to discriminate one device from another.

You need the length of the id to be adapted accordingly following the global context.
See how it's done for `format_device`: 

https://github.com/Scille/parsec-cloud/blob/653e19072c60af621d8c18e12045754926e987b0/cli/src/utils.rs#L54-L61

FirelightFlagboy (approved):

PR #13376, opened 2026-09-14
refactor(server): Add pydantic support for `OrganizationID` & `DateTime`

FirelightFlagboy:
Experiment adding pydantic's support directly to Rust types to not have to rely on `*Field` annotations.

During development in the server, I lost track how we added support of pydantic for custom rust types. After reading the pydantic documentation I found that I needed to implement specific method to add support.

And with those method implemented, we no longer need to define `*Field` annotations and directly use the type instead.

I've check the generated `openapi.json` spec and it's correct.


mmmarcos:
The scope of this PR is not clear: 
- if you consider adding support for other types here, you could add a checklist in the description and mark it as draft PR in the meantime.
- if you prefer to stick to `OrganizationID` and `Datetime` here, then the checklist for other types could go to an issue to track progress in refactor
  - BTW, there are other route handlers still using `raw_organization_id` (currently, only the `administration_get_organization` route is updated).

Anyway, kudos for this approach :+1:

FirelightFlagboy:
That would be too much to do the rest in the same PR.

This PR is mostly to have feedback, if it's something we want to continue on or not, so I did minimal changes.

mmmarcos:
The server code is a little bit simplified at the expense of more complex type declaration on the Rust side, do you see any particular benefit/drawbacks of this approach?

FirelightFlagboy:
> do you see any particular benefit/drawbacks of this approach?

As indicated in the PR description, it to not rely on `*Field` annotations.
It simplify usage on the python side since you do not need to use an annotated type.

But at the cost that required methods must be implemented on the Rust side.

It's more verbose on the Rust side, but for documentation it's similar thanks to `pydantic_json_schema!` macro

mmmarcos:
+1 to simplify usage

PR #13379, opened 2026-09-14
feat(server): Add tasks to remove old organization

### LINKED ISSUE #8352: Automatically delete organizations on the trial server after 15 days

FirelightFlagboy:
Will be used to remove ephemeral organization older than a given time

Closes #8352


server/tests/cli/tasks/test_rm_old_organizations.py
```diff
@@ -0,0 +1,74 @@
...
+
+from parsec.cli.options import Duration
+from parsec.cli.tasks import delete_old_organization
+from parsec.cli.testbed import TestbedBackend
+from parsec.components.organization import Organization, OrganizationGetBadOutcome
+from tests.common.client import CoolorgRpcClients
+
+
+@pytest.mark.parametrize("kind", ("interval", "date"))
+def test_delete_old_organization_cmd(kind: str, db_args: list[str], testbed: TestbedBackend):
```
FirelightFlagboy, line 18: **issue:** When run with PG enabled, this tests tries to delete every testbed orga.

This is caused because every orga use the same creation dates (2000-01-01T00:00:00).

Futhermore, we duplicate testbed orga, so even tho we would use different dates for templates, it would only reduce the targeted orga to delete

PR #13380, opened 2026-09-15
Adds RFC for multiple SSO handling by the GUI

Max-7:


docs/rfcs/1030-generic-sso-provider-display.md
```diff
@@ -0,0 +1,225 @@
...
+### 1. Data the GUI expects to receive
+
+This RFC assumes the GUI receives, for each configured SSO provider, an object shaped roughly
+like:
+
+```ts
+interface SsoProviderDisplay {
+  id: string; // opaque identifier, passed back unchanged to select this provider
+  name: string; // display name, e.g. "ProConnect", "Acme Corp SSO"
+  mountPath: string; // currently implemented
```
FirelightFlagboy, line 68: **nitpick:** Use a better comment

docs/rfcs/1030-generic-sso-provider-display.md
```diff
@@ -0,0 +1,225 @@
...
+  link?: string; // URL to documentation/help about this provider, if any
```
FirelightFlagboy, line 69: **nitpick:** 
```suggestion
  helpLink?: string; // URL to documentation/help about this provider, if any
```

docs/rfcs/1030-generic-sso-provider-display.md
```diff
@@ -0,0 +1,225 @@
...
+  icon?: string; // something the GUI can render as an image (see 4.)
```
FirelightFlagboy, line 70: **nitpick:** Better specify the type of icon, can be a link or raw data
```suggestion
  iconURL?: string; // something the GUI can render as an image (see 4.)
```

URL seems to better fit because the navigator can then cache the icon, and you can still embeded icon using the `data:` scheme

docs/rfcs/1030-generic-sso-provider-display.md
```diff
@@ -0,0 +1,225 @@
...
+
+## Design
+
+### 1. Data the GUI expects to receive
+
+This RFC assumes the GUI receives, for each configured SSO provider, an object shaped roughly
+like:
+
+```ts
+interface SsoProviderDisplay {
```
FirelightFlagboy, line 65: **thought:** Should the object should also store a color set for background & text ?

E.g.:

- For GitHub: black background + white text
- For Pro-Connect: blue background + white text

docs/rfcs/1030-generic-sso-provider-display.md
```diff
@@ -0,0 +1,225 @@
...
+Parsec's design system rather than by a provider-supplied image, and the icon/name are laid out
+*inside* that shared chrome. This is a deliberate change from today's ProConnect card, whose whole
+clickable surface *is* a brand-supplied bitmap. It keeps a list of many, visually unrelated
+providers from looking like a pile of random buttons of different sizes and styles.
+
+Sensible fallbacks are needed since data comes from an arbitrary server config:
+
+- no `icon` -> a generic placeholder icon (e.g. a lock/key `ion-icon`) shipped with the GUI,
+- no `link` -> the secondary line is simply not rendered,
+- missing/empty id/name/mountPath -> the button is not displayed.
```
FirelightFlagboy, line 102: **polish:** Maybe you could consider displaying a warning message for empty value ?

Note that libparsec will not break the contract for required field, so only empty value variant is possible

docs/rfcs/1030-generic-sso-provider-display.md
AureliaDolo: I assume it would depend on if we want to display our colors or theirs. From L52 I would guess that we would rather use our colors.

docs/rfcs/1030-generic-sso-provider-display.md
```diff
@@ -0,0 +1,225 @@
...
+  id: string; // opaque identifier, passed back unchanged to select this provider
+  name: string; // display name, e.g. "ProConnect", "Acme Corp SSO"
+  mountPath: string; // currently implemented
+  link?: string; // URL to documentation/help about this provider, if any
+  icon?: string; // something the GUI can render as an image (see 4.)
+}
+```
+
+The exact wire format (in particular whether `name` is a plain string or needs to support multiple
+languages, and the concrete encoding of `icon`) is for the libparsec RFC to define. This RFC only
```
AureliaDolo, line 75: I don't get why the name would need to support multiple languages.

docs/rfcs/1030-generic-sso-provider-display.md
FirelightFlagboy: That could potentially conflict with the color of the icon tho, e.g.:

Our blue logo with no outline on a blue background

AureliaDolo: I guess that if the admin provides an icon without a background, there is nothing that can be done to alter the icon on the fly. Maybe it would be worth encouraging to provide an icon with a background.

Or I don't know how it could be done in practice but maybe it would be nice to have somewhere in the app settings a page to test the rendering for the sso config. The ideal would be to have that as a standalone so the admin doesn't have to actually install the parsec app, but that would be one more thing.

docs/rfcs/1030-generic-sso-provider-display.md
Max-7: I wondered about that but I'm not sure if it'll always be stricly a `helpLink` or even whether we want to present it as a help link (a link "More info" or a link "Help").

If we take the example of ProConnect, what link should be provided? I see [https://www.proconnect.gouv.fr/](ProConnect.gouv.fr) or [https://proconnect.crisp.help/fr/](ProConnect.crisp.help) and weirdly, the one dedicated for help is more confusing.

docs/rfcs/1030-generic-sso-provider-display.md
Max-7: It's discussed in 4. whether we want to provide a URL or embed the raw icon instead, and it's for libparsec to decide:
```
Whatever format libparsec ends up choosing for icon (embedded data URI vs. a URL served by the
Parsec server vs. something else)
```

Renaming the attribute here would mean it's already decided.

docs/rfcs/1030-generic-sso-provider-display.md
Max-7: I've wondered about the colors but I see two problems with it:
- it could clash with the theme (even worse if we get dark mode)
- it basically lets the server admin make design decisions

Displaying a generic button that works for all SSOs is going to be complicated enough I assume, adding colors will make it even worse. We can iterate with @fabienSvtr and re-evaluate if he deems it necessary but I'm fairly sure it's a bad idea.

docs/rfcs/1030-generic-sso-provider-display.md
Max-7: Well that's for libparsec RFC to define, as discussed in 5.

It has more to do with the fact that we don't like to display raw strings in the GUI, everything should be going through the translation system. I don't know if there are cases out there of SSOs that change their name depending on the language, and there probably aren't, but you never know (something named AccountConnect becoming CompteConnexion in French because some legislator took a stance and made themselve the big defender of the French language, annoying hundred of engineers in the process).

While it's probably not needed for the name, it may become more of a problem for the link.

docs/rfcs/1030-generic-sso-provider-display.md
Max-7: I've updated the RFC with my comment on why we're avoiding colors in our object.

docs/rfcs/1030-generic-sso-provider-display.md
Max-7: I've updated the RFC to add something specially on the need for the link to be translated.

docs/rfcs/1030-generic-sso-provider-display.md
```diff
@@ -0,0 +1,237 @@
...
+Parsec's design system rather than by a provider-supplied image, and the icon/name are laid out
+*inside* that shared chrome. This is a deliberate change from today's ProConnect card, whose whole
+clickable surface *is* a brand-supplied bitmap. It keeps a list of many, visually unrelated
+providers from looking like a pile of random buttons of different sizes and styles.
+
+Sensible fallbacks are needed since data comes from an arbitrary server config:
+
+- no `icon` -> a generic placeholder icon (e.g. a lock/key `ion-icon`) shipped with the GUI,
+- no `link` -> the secondary line is simply not rendered,
+- missing/empty id/name/mountPath -> the button is not displayed. This should not happen as
```
FirelightFlagboy: **polish:** libparsec will not breaking the contract, if a value must be provide, it will be provided or fail
```suggestion
- empty id/name/mountPath -> the button is not displayed. This should not happen as
```

touilleMan:
Regarding how libparsec & server side currently handle SSO:

## Current situation

SSO configuration is transmited from the client to the server using the `server_config` command (server anonymous API), and SSO are not hardcoded there:

https://github.com/Scille/parsec-cloud/blob/a25b4727eeba163d90393cae0dcaa2bd3447416b/libparsec/crates/protocol/schema/anonymous_server_cmds/server_config.json5#L172-L185


However the server CLI itself gets the SSO configuration through hardcoded parameters:

https://github.com/Scille/parsec-cloud/blob/a25b4727eeba163d90393cae0dcaa2bd3447416b/server/parsec/cli/run.py#L270-L289

Similarly, libparsec cook the `server_config` response to turn the generic SSO ID into hardcoded values:

https://github.com/Scille/parsec-cloud/blob/a25b4727eeba163d90393cae0dcaa2bd3447416b/libparsec/src/config.rs#L167-L177

## Why ?

The choice has been made to keep the API SSO agnostic as future-proofing (i.e. anticipate this RFC ^^), while the server CLI parameters & libparsec was using hardcoded options to simplify code and stay close to the actual needs of the GUI (i.e. YAGNI)

## Evolutions

Since all SSO info has to be provided from the server, we must update the `server_config` API, server CLI parameters and libparsec.

### `server_config` API

```json5
            {
                "name": "OpenBaoAuthConfig",
                "fields": [
                    // Existing fields
                    {
                        // Opaque identifier, passed back unchanged to select this provider
                        "name": "id",
                        "type": "String"
                    },
                    {
                        // Basically OIDC authentication is going to be done with
                        // `<openbao_server_url>/<openbao_mount_path>/oidc/auth_url`
                        // see https://openbao.org/api-docs/auth/jwt/#oidc-authorization-url-request
                        "name": "mount_path",
                        "type": "String"
                    },
                    // New fields
                    {
                        // Display name, e.g. "ProConnect", "Acme Corp SSO"
                        // Should default to the ID if not provided
                        "name": "name",
                        "type": "String",
                        "introduced_in": 5.7
                    },
                    {
                        // URL to documentation/help about this provider, if any
                        "name": "link",
                        "type": "RequiredOption<String>",
                        "introduced_in": 5.7
                    },
                    {
                        // Something the GUI can render as an image
                        "name": "icon",
                        "type": "RequiredOption<String>",
                        "introduced_in": 5.7
                    }
                ]
            },
```

### server CLI parameters

Considering the amount of items to configure a single SSO (and the fact we can configure multiple SSO), we should use JSON payload as argument of the parameter:

```shell
parsec run ... --openbao-auth '{"id": "PRO_CONNECT", "name": "ProConnect", ...}'
```

### libparsec

The exposed `OpenBaoAuthConfig` should no longer be an enum, instead the `OpenBaoAuthConfig` structure from `server_config` API  can directly be used.

docs/rfcs/1030-generic-sso-provider-display.md
```diff
@@ -0,0 +1,237 @@
...
+- **Arbitrary link**: clicking a provider opens `link` in the system browser via the existing
+  `Env.Links.openUrl()` helper (never an in-app webview), which is already how `proConnect.link`
+  behaves today. This doesn't introduce a new mechanism, but it does widen what can be linked from
+  inside the Parsec client to whatever an org's server config sets - the server config is already a
+  trust boundary the client relies on today (it already dictates the OpenBao server URL, mount
+  paths, etc.), so this isn't a new trust boundary, just a broader use of an existing one.
+- **Arbitrary icon**: must strictly be rendered as an image resource, never as raw markup, to
+  prevent an SSO entry's icon field from becoming an XSS vector (see 4. above).
+- **Social-engineering surface**: because `name`/`link` are free text controlled by whoever
+  configures the server, a compromised or malicious org admin could label an entry misleadingly
```
mmmarcos, line 201:
```suggestion
  configures the server, a compromised or malicious server admin could label an entry misleadingly
```

docs/rfcs/1030-generic-sso-provider-display.md
```diff
@@ -0,0 +1,237 @@
...
+  (e.g. a `name` designed to look like "Reset your password" pointing to a credential-harvesting
+  page). This is not fundamentally new (the same admin already controls the OpenBao server URL used
+  for real authentication), but the attack becomes slightly cheaper/more visible since it's now a
+  UI element instead of a URL, so this is worth flagging to the security team when the
+  libparsec/server side of this feature is designed.
```
mmmarcos, line 206: On the GUI-side, maybe we could think of a way to clearly infrom the user that this action requires special attention, something like:
```
You are about to be redirected to the SSO auth page, in case you have 
any concerns about the page (mispelled texts, weird logos, strange URLs...) 
please do not proceed and contact an administrator of your organization." 
```

It's common to see this type of warning when transferring money on Bank websites for example, not sure about the real impact on security though as it still depends on the user to be careful.

docs/rfcs/1030-generic-sso-provider-display.md
```diff
@@ -0,0 +1,237 @@
...
+  implementation may need adjusting once that follow-up RFC settles the real data contract
+  (especially icon encoding and whether `name`/`link` support localization).
+
+## Remarks & open questions
+
+- What identifier does the GUI pass back to select a provider once `OpenBaoAuthConfigTag` is no
+  longer a closed enum? This RFC assumes an opaque `id` round-tripped from the list entry, but the
+  concrete type is for the libparsec RFC to define.
+- Does ProConnect keep a dedicated, brand-compliant button, or move fully to the generic renderer
+  like every other provider (see Risks)?
```
mmmarcos, line 231: This should be addressed in the RFC by checking the official docs.

mmmarcos (commented):
LGTM overall 👍

docs/rfcs/1030-generic-sso-provider-display.md
FirelightFlagboy: I think for "well-known" providers, we could have special tailored button for them.

The goal of this RFC is to not limit to only those providers

PR #13381, opened 2026-09-15
Add missing 3.10.0 newsfragments

Ironicbay:


FirelightFlagboy (approved):

newsfragments/13310.feature.rst
```diff
@@ -0,0 +1 @@
+[Desktop] Merged the title bar into the window
```
mmmarcos: I believe this is called a "unified title bar". 
BTW does it work on Linux/snap?
```suggestion
[Desktop] Combined app window and title into a unified title bar
```

newsfragments/13309.feature.rst
```diff
@@ -0,0 +1 @@
+[MacOS] Added commands to the native menu in the desktop app
```
mmmarcos: Wouldn't be better to specify which commands?

newsfragments/13225.bugfix.rst
```diff
@@ -0,0 +1 @@
+Added proper error management when trying to drag&drop files in an empty workspace as a `Reader`
```
mmmarcos:
```suggestion
Improve error management when trying to drag&drop files in an empty workspace as a `Reader`
```

PR #13383, opened 2026-09-15
Update changelog to group CLI-related changes into a specific section

Ironicbay:


FirelightFlagboy (changes requested):
Since it's specific to the CLI create a breaking change section under CLI.

I would propose to skip translating the breaking change

FirelightFlagboy (approved):

PR #13384, opened 2026-09-16
[MS] Add Monaco as the default text editor

Max-7:
Replace Cryptpad with Monaco for editing text files, as it's going to be needed with the incoming changes on collaborative editing.

It also includes a bit of refactoring:
- handle more text file extensions
- renamed `[MOCKED-ELECTRON-LOG]` to `[PARSEC-LOG]` (this was forgotten in #13173)
- adds a spinner modal when opening a document. I thought this was already the case but apparently not
- updated viewers API to handle saves/readonly. This is a bit weird because `Viewer` and `Editor` are misnamed (`editor` means cryptpad, `viewer` means something we display ourselve). We may need to refactor this part in the future.


client/src/services/viewers.ts
```diff
@@ -114,7 +146,8 @@ export function getStreamUrl(workspaceHandle: number, path: string, size: number
 }
 
 export async function initViewers(): Promise<void> {
-  await _initPdf();
+  _initPdf();
+  _initMonaco();
```
Ironicbay, line 150: Why aren't these `awaited`?

client/src/views/files/handler/viewer/TextViewer.vue
```diff
@@ -0,0 +1,262 @@
...
+      scrollBeyondLastLine: false,
+      unicodeHighlight: { ambiguousCharacters: false, invisibleCharacters: false },
+      minimap: { enabled: false },
+      glyphMargin: false,
+      codeLens: false,
+      quickSuggestions: false,
+    });
+    if (!isReadOnly.value) {
+      subscription = editor.onDidChangeModelContent((_event: monaco.editor.IModelContentChangedEvent) => {
+        console.log(_event);
```
Ironicbay: Is this log wanted?

Ironicbay (approved):

client/src/services/viewers.ts
Max-7: We don't need to wait for them to be initialized, we're not using them until waaaay later. So we just fire and forget.

PR #13406, opened 2026-09-21
[MS] Rework streaming worker

Max-7:
- Made the worker stateless
- Better handling of Range requests, especially when the protocol is not HTTP (like on Electron where it's parsec-desktop)
- Better response with content-type

Tested on Chrome, Electron and Firefox, with video, images and sound, with a file up to 900MB. Everything works properly. It can be a bit slow, but that's linked to Parsec block downloading being unoptimized.


client/public/streaming-worker.js
```diff
@@ -1,15 +1,62 @@
...
-const pendingRequests = new Map();
+// This worker serves decrypted file content to <video>, <audio> and <img> elements.
+//
+// It is stateless on purpose: browsers terminate idle service workers (after ~30s
+// without events) and wipe their global state, so nothing can be kept in memory
+// between two requests. Instead, every read is delegated to the tab that made the
+// request (the only place with access to libparsec) through a dedicated one-shot
+// MessageChannel:
+//
+//   worker -> tab (client.postMessage):  { type: 'READ', workspaceHandle, filePath, offset, size, historyHandle }
```
FirelightFlagboy, line 11: **polish:** Since, we use a service worker, you could directly open a new connection to it and thus removing the passtrough on `viewers.ts`

Max-7: That was my original design but because Electron and web works differently, it adds unecessary complexity. This does add a level of indirection but it works the same way on all platforms, and if we ever notice that performance is an issue, we'll rework this part.

Ironicbay (approved):

PR #13409, opened 2026-09-21
[MS] Rework download

Max-7:
Completely change how downloads are handled.

We used to try and get a file handle using the [File System API](https://developer.mozilla.org/en-US/docs/Web/API/File_System_API). With that handle, we can write files directly on the user's drive. Sadly, only Chrome-based browsers implement this API. We used [native-file-system-adapter](https://github.com/jimmywarting/native-file-system-adapter) which basically replaced the native handle by a fake in memory one to keep the same API. This kinda works but it has some heavy restrictions (whole file in memory, some bug when a service worker is present, some limitation with directory handles, ...).

Instead, we use the streaming-worker to serve the file over HTTP as if the file was downloaded from a server: we make a request to `/parsec-download` with a list of files, the worker intercepts that request, get the files from Parsec, and stream it back to the browser. For a user, it looks like any other download.

I've removed the `Download as a ZIP file`, to keep only `Download`. If multiple files are selected, we automatically create a zip file and stream it.

I haven't renamed `streaming-worker` yet, we'll do that later one it's merged back into master, to make the review and the conflict resolution easier.


client/public/vendor/zip-native.min.js
FirelightFlagboy, line 1: **issue:** Where does this lib come from ? Why do we need to vendor it ?

FirelightFlagboy (dismissed):

client/public/vendor/zip-native.min.js
Max-7: We were getting it from the npm repository before. But now it's only used in the service worker that sits in the public/ repository because it can't be packaged with everything (issues with rolldown, see #13018). Since the library needs to be available to the worker, it has to be shipped in that same directory. We could keep getting it from npm and do a copy at build time but it's way more annoying than just ship it this way.

FirelightFlagboy: By looking at the changelog, they've included some fixes to the multi input it seems

Max-7: I'm relunctant to change this now (could cause problems for the next RCs and make the acknowledge difficult) but I've opened #13412, we'll look at it when all this gets on master.

Ironicbay (approved):

PR #13411, opened 2026-09-22
refactor(postgresql): Add `on delete` constraints for account tables

FirelightFlagboy:
Use the `on delete cascade` constraint to have all linked items removed for a given account. Currently we do not remove account, but that would be useful if it need to be.


server/parsec/components/postgresql/migrations/0027_account_delete_constraints.sql
```diff
@@ -0,0 +1,39 @@
+-- Parsec Cloud (https://parsec.cloud) Copyright (c) BUSL-1.1 2016-present Scille SAS
+
+-------------------------------------------------------
+--  Migration
+--
+-- Add delete constraints for account tables to simplify removal of account
+--
+-- NOTE: With an inline REFERENCES like that (no explicit CONSTRAINT name given),
+-- Postgres auto-generates the name using the pattern: `<table>_<column>_fkey`
```
mmmarcos, line 9: Couldn't find this in the official docs, which makes me think that could be changed in the future.

We should probably define our own naming convention for constraints as a best practice.

server/parsec/components/postgresql/migrations/0027_account_delete_constraints.sql
```diff
@@ -0,0 +1,39 @@
+-- Parsec Cloud (https://parsec.cloud) Copyright (c) BUSL-1.1 2016-present Scille SAS
+
+-------------------------------------------------------
+--  Migration
+--
+-- Add delete constraints for account tables to simplify removal of account
```
mmmarcos, line 6: Shouldn't we update account delete tests to check that associated `vault`/`vault_item`/`vault_authentication_method` has been deleted?

mmmarcos (commented):
LGTM

server/parsec/components/postgresql/migrations/0027_account_delete_constraints.sql
FirelightFlagboy: We do not delete the account when we delete it :sweat_smile: we just disable it.

The tests does not need to be updated as nothing is deleted so.

But I've added a tests for that

server/parsec/components/postgresql/migrations/0027_account_delete_constraints.sql
FirelightFlagboy: I've updated the datamodel to give constraint names

server/tests/account/test_account_delete.py
```diff
@@ -0,0 +1,102 @@
...
+            )
+            assert row_id is not None, (
+                f"Table {table} is missing entry linked to account#{account_id}"
+            )
+
+    # 2. Wipe the account
+    async with account.pool.acquire() as conn:
+        await conn.execute("DELETE FROM account WHERE _id = $1", account_id)
+
+    # 3. Verify that the wipe as cascaded
```
mmmarcos, line 68:
```suggestion
    # 3. Verify that the wipe has cascaded
```

server/tests/account/test_account_delete.py
```diff
@@ -0,0 +1,102 @@
...
+    async with account.pool.acquire() as conn:
+        for table, id in itertools.chain(
+            (("account", account_id),),
+            (("vault", id) for id in vault_ids),
+            (("vault_item", id) for id in vault_item_ids),
+            (("vault_authentication_method", id) for id in vault_authentication_method_ids),
+        ):
+            row_id = await conn.fetchval(f"SELECT _id FROM {table} WHERE _id = $1", id)
+            assert row_id is None, f"Table {table} still have a line linked to deleted account"
```
mmmarcos, line 77:
```suggestion
            assert row_id is None, f"Table {table} still has a row linked to the deleted account"
```

server/tests/account/test_account_delete.py
```diff
@@ -0,0 +1,102 @@
...
+
+        for table in ("account_delete_validation_code", "account_recover_validation_code"):
+            row_id = await conn.fetchval(
+                f"SELECT account FROM {table} WHERE account = $1", account_id
+            )
+            assert row_id is None, f"Table {table} still have a line linked to deleted account"
```
mmmarcos, line 83:
```suggestion
            assert row_id is None, f"Table {table} still has a row linked to the deleted account"
```

mmmarcos (approved):

PR #13413, opened 2026-09-22
[MS] Hide Hidden workspace category button when no workspace is hidden

Ironicbay:


client/src/views/workspaces/WorkspacesPage.vue
```diff
@@ -347,6 +349,7 @@ const search = ref<FileSearch | undefined>(undefined);
 let searchAborter: AbortController | undefined = undefined;
 const searchPattern = ref('');
 const searchInputValue = ref('');
+const someHiddenWorkspaces = ref(false);
```
Max-7: Use a computed.

Max-7 (changes requested):
You could also add a test or edit an existing test to check if the button is properly hidden.

Max-7 (approved):

